PR #4879, opened 2021-07-28
EVG-15058 add DefaultSectionToRepo function

ablack12:
[EVG-15058](https://jira.mongodb.org/browse/EVG-15058)

### Description 
Created constants for each of the sections, and unset or deleted anything relevant (projectRef variables that aren't set use the value from the repoRef). Also logged, and made sure to log even if we were somehow only able to complete part of the update.

### Testing 
Added unit test. This isn't used anywhere yet to test elsewhere.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -327,7 +327,7 @@ const (
 	ProjectRefGeneralSection        = "general"
 	ProjectRefAccessSection         = "access"
 	ProjectRefVariablesSection      = "variables"
-	ProjectRefGithubSection         = "github"
+	ProjectRefGithubAndCQSection    = "github_and_commit_queue"
 	ProjectRefNotificationsSection  = "notifications"
 	ProjectRefAliasSection          = "alias"
 	ProjectRefWorkstationsSection   = "workstations"
@@ -1306,7 +1306,7 @@ func (projectRef *ProjectRef) Upsert() error {
 
 // DefaultSectionToRepo modifies a subset of the project ref to use the repo values instead.
 // This subset is based on the pages used in Spruce.
-func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller string) error {
+func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, userId string) error {
 	before, err := GetProjectSettingsEvents(p)
 	if err != nil {
 		return errors.Wrap(err, "error getting before project settings event")
@@ -1329,7 +1329,7 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 					projectRefDefaultLoggerKey:           1,
 					projectRefCedarTestResultsEnabledKey: 1,
 					projectRefPatchingDisabledKey:        1,
-					projectRefTaskSyncKey:                1, // TODO: look into whether this works for individual settings
+					projectRefTaskSyncKey:                1,
 					ProjectRefDisabledStatsCacheKey:      1,
 					ProjectRefFilesIgnoredFromCacheKey:   1,
 				},
@@ -1363,7 +1363,7 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 		if err != nil {
 			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
 		}
-	case ProjectRefGithubSection:
+	case ProjectRefGithubAndCQSection:
 		err = db.Update(ProjectRefCollection,
 			bson.M{ProjectRefIdKey: p.Id},
 			bson.M{
@@ -1389,7 +1389,7 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 		}
 		if catcher.HasErrors() {
 			// log the changes before returning whatever errors were caught
-			_ = getAndLogProjectModified(p.Id, caller, before)
+			_ = getAndLogProjectModified(p.Id, userId, before)
 			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
 		}
 
@@ -1409,7 +1409,7 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 		}
 		if catcher.HasErrors() {
 			// log the changes before returning whatever errors were caught
-			_ = getAndLogProjectModified(p.Id, caller, before)
+			_ = getAndLogProjectModified(p.Id, userId, before)
 			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
 		}
 	case ProjectRefAliasSection:
@@ -1422,8 +1422,17 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 		}
 		if catcher.HasErrors() {
 			// log any changes before returning whatever errors were caught
-			_ = getAndLogProjectModified(p.Id, caller, before)
+			_ = getAndLogProjectModified(p.Id, userId, before)
 			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefWorkstationsSection:
+		err = db.Update(ProjectRefCollection,
+			bson.M{ProjectRefIdKey: p.Id},
+			bson.M{
+				"$unset": bson.M{projectRefWorkstationConfigKey: 1},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
 		}
 	case ProjectRefTriggersSection:
 		err = db.Update(ProjectRefCollection,
@@ -1433,15 +1442,6 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 					projectRefTriggersKey:            1,
 					projectRefPatchTriggerAliasesKey: 1, // is this where we decided patch trigger aliases would go?
 				},
-			})
-		if err != nil {
-			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
-		}
-	case ProjectRefWorkstationsSection:
-		err = db.Update(ProjectRefCollection,
-			bson.M{ProjectRefIdKey: p.Id},
-			bson.M{
-				"$unset": bson.M{projectRefWorkstationConfigKey: 1},
 			})
 		if err != nil {
 			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
@@ -1459,23 +1459,23 @@ func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller stri
 		return errors.Errorf("invalid section '%s'", section)
 	}
 
-	return getAndLogProjectModified(p.Id, caller, before)
+	return getAndLogProjectModified(p.Id, userId, before)
 }
 
-func getAndLogProjectModified(id, caller string, before *ProjectSettingsEvent) error {
+func getAndLogProjectModified(id, userId string, before *ProjectSettingsEvent) error {
 	p, err := FindOneProjectRef(id)
 	if err != nil {
 		return errors.Wrapf(err, "error getting new project ref")
 	}
 	if p == nil {
-
+		return errors.Errorf("project ref '%s' not found", id)
 	}
 	// still modify the project as logged before erroring
 	after, err := GetProjectSettingsEvents(p)
 	if err != nil {
 		return errors.Wrap(err, "error getting after project settings event")
 	}
-	return errors.Wrapf(LogProjectModified(p.Id, caller, before, after), "error logging project modified")
+	return errors.Wrapf(LogProjectModified(p.Id, userId, before, after), "error logging project modified")
 }
 
 // getBatchTimeForVariant returns the Batch Time to be used for this variant
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1170,6 +1170,17 @@ func FindMergedProjectRefsForRepo(repoRef *RepoRef) ([]ProjectRef, error) {
 	return projectRefs, nil
 }
 
+func GetProjectSettingsEventById(projectId string) (*ProjectSettingsEvent, error) {
+	pRef, err := FindOneProjectRef(projectId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error finding project ref")
+	}
+	if pRef == nil {
+		return nil, errors.Errorf("couldn't find project ref")
+	}
+	return GetProjectSettingsEvents(pRef)
+}
+
 func GetProjectSettingsEvents(p *ProjectRef) (*ProjectSettingsEvent, error) {
 	hook, err := FindGithubHook(p.Owner, p.Repo)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+}
+
+func getAndLogProjectModified(id, caller string, before *ProjectSettingsEvent) error {
+	p, err := FindOneProjectRef(id)
+	if err != nil {
+		return errors.Wrapf(err, "error getting new project ref")
+	}
+	if p == nil {
+
+	}
```
sophstad, line 1472: Is this intentionally left empty?

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+			bson.M{ProjectRefIdKey: p.Id},
+			bson.M{
+				"$unset": bson.M{
+					projectRefPRTestingEnabledKey:      1,
+					projectRefGithubChecksEnabledKey:   1,
+					projectRefGithubTriggerAliasesKey:  1,
+					projectRefGitTagVersionsEnabledKey: 1,
+					ProjectRefGitTagAuthorizedUsersKey: 1,
+					ProjectRefGitTagAuthorizedTeamsKey: 1,
+					projectRefCommitQueueKey:           1,
```
malikchaya2, line 1377: The projectRefCommitQueueKey includes Enabled, MergeMethod, Message. 

The UI commit queue section also has the following fields that probably need to be defaulted as well:
- patch type
- commit queue patch definitions.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+			// log the changes before returning whatever errors were caught
+			_ = getAndLogProjectModified(p.Id, caller, before)
+			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefAliasSection:
+		catcher := grip.NewBasicCatcher()
+		// remove only non-internal aliases
+		for _, a := range before.Aliases {
+			if !utility.StringSliceContains(evergreen.InternalAliases, a.Alias) {
+				catcher.Add(RemoveProjectAlias(a.ID.Hex()))
```
malikchaya2, line 1420: This seems to be doing the same thing that's done on 1384. Should this be specific to the Patch Aliases section (as opposed to project aliases)?

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+			}
+		}
+		if catcher.HasErrors() {
+			// log any changes before returning whatever errors were caught
+			_ = getAndLogProjectModified(p.Id, caller, before)
+			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefTriggersSection:
```
malikchaya2, line 1428: nit: would it be possible to move this down to be between the virtual workstation and periodic builds section so that it's in the same order as the UI?

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefWorkstationsSection:
+		err = db.Update(ProjectRefCollection,
+			bson.M{ProjectRefIdKey: p.Id},
+			bson.M{
+				"$unset": bson.M{projectRefWorkstationConfigKey: 1},
```
malikchaya2, line 1444: Similar comment as above, we'd need to separate this into WorkstationSetupCommand.Command and WorkstationSetupCommand.Directory if we refactor to one update function.

model/project_ref.go
```diff
@@ -316,6 +321,20 @@ const (
 	maxBatchTime             = 153722867 // math.MaxInt64 / 60 / 1_000_000_000
 )
 
+type ProjectRefSection string
+
+const (
+	ProjectRefGeneralSection        = "general"
+	ProjectRefAccessSection         = "access"
+	ProjectRefVariablesSection      = "variables"
+	ProjectRefGithubSection         = "github"
```
malikchaya2, line 330: This is technically both a github and commit queue section on the UI.

ablack12: Yeah; I thought this simplified it since commit queue might just be github PRs in the future but eh I'll be clearer.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
 
+// DefaultSectionToRepo modifies a subset of the project ref to use the repo values instead.
+// This subset is based on the pages used in Spruce.
+func (p *ProjectRef) DefaultSectionToRepo(section ProjectRefSection, caller string) error {
+	before, err := GetProjectSettingsEvents(p)
+	if err != nil {
+		return errors.Wrap(err, "error getting before project settings event")
+	}
+
+	switch section {
```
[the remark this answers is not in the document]
ablack12: > and pass in that struct containing the changes you want to make

I'm not sure what you mean by this. Should we be passing in a struct with the relevant ProjectRef fields already unset and then just call update? Or a map of things we want to change? I'm not sure it works really to make this very general, since each section updates different or multiple collections. 
We could make each one it's own function. I liked the idea of keeping it all together for organizational purposes.

model/project_ref.go
ablack12: Whoops nope

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+	// check if the sections cover all the settings
+	case ProjectRefGeneralSection:
+		err = db.Update(ProjectRefCollection,
```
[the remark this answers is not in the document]
ablack12: Ah, I don't think that should default to repo (similar to how branch should not be defaulted to repo). I think DisplayName should be required.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+					ProjectRefBatchTimeKey:               1,
+					ProjectRefRemotePathKey:              1,
+					projectRefSpawnHostScriptPathKey:     1,
+					projectRefDispatchingDisabledKey:     1,
+					ProjectRefDeactivatePreviousKey:      1,
+					projectRefRepotrackerDisabledKey:     1,
+					projectRefDefaultLoggerKey:           1,
+					projectRefCedarTestResultsEnabledKey: 1,
+					projectRefPatchingDisabledKey:        1,
+					projectRefTaskSyncKey:                1, // TODO: look into whether this works for individual settings
```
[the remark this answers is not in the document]
ablack12: nope sorry I already looked into this

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				"$unset": bson.M{
+					ProjectRefPrivateKey:    1,
+					ProjectRefRestrictedKey: 1,
+					ProjectRefAdminsKey:     1,
+				},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefVariablesSection:
```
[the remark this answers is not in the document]
ablack12: This removes only the branch level variables. The repo level variables are stored under the RepoRefid.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+		err = db.Update(ProjectVarsCollection,
+			bson.M{ProjectRefIdKey: p.Id},
+			bson.M{
+				"$unset": bson.M{
+					projectVarsMapKey:    1,
+					privateVarsMapKey:    1,
+					restrictedVarsMapKey: 1,
```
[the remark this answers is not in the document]
ablack12: We'll be removing restricted vars in the future. We don't really need to include this here but it seems we might as well for now.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
+		}
+	case ProjectRefGithubSection:
+		err = db.Update(ProjectRefCollection,
+			bson.M{ProjectRefIdKey: p.Id},
+			bson.M{
+				"$unset": bson.M{
```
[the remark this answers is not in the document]
ablack12: Github webhooks isn't a setting that the user can modify; if the owner/repo/branch is reachable by Evergreen then it's true, otherwise it's false.
Github patch definitions and git tag version definitions are handled by the alias section below, i.e. we remove all aliases relating to Evergreen processes.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+					projectRefCommitQueueKey:           1,
+				},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
+		}
+		catcher := grip.NewBasicCatcher()
+		for _, a := range before.Aliases {
+			// remove only internal aliases
+			if utility.StringSliceContains(evergreen.InternalAliases, a.Alias) {
```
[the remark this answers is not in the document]
ablack12: Aliases that don't have one of these labels are Patch Aliases, which is defined in a different section (ProjectRefAliasSection)

model/project_ref.go
ablack12: This is doing the _opposite_ thing as 1384: removes the project alias if it's _not_ an internal alias. Patch Aliases are still project aliases; they are defined at the project level and can be used for patch creation. For example our test alias: https://evergreen.mongodb.com/projects##evergreen

model/project_ref.go
ablack12: Patch Type is actually outdated; commit queues now accept both PR and CLI. I don't think we need to ask Lara to modify the design but we should drop a comment that it's not relevant anymore.

Commit queue patch definitions are handled in the alias section below; these are aliases with the __commit_queue alias label.

ablack12:
Thanks for the refactor suggestion! I found it difficult/non-intuitive to extrapolate the other collections into a "save or default" kind of function but I think this is much clearer and will handle save for a lot of the sections without needing too much other handling. I think that can be covered in EVG-15060

model/project_ref.go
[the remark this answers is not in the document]
malikchaya2: I guess it makes sense in case someone already has restricted vars defined.

model/project_ref.go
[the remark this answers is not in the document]
malikchaya2: Would it be possible to add a comment here so that I remember that next time I look at this logic?n

model/project_ref.go
malikchaya2: haha yes, I missed that! thanks!

malikchaya2: Would it be possible to update the comment above? I think the comment is the same.

malikchaya2: nope, that's not true. never mind :)

malikchaya2: Would it maybe be possible to elaborate in the comment that non-internal aliases represent patch aliases?

model/project_ref.go
malikchaya2: Do we not want to do it as part of case ProjectRefGithubAndCQSection: so that when someone clicks on the "default page to repo" button for the commit queue page that gets defaulted as well?

model/project_ref.go
malikchaya2: ah I hear that! thanks!

model/project_ref.go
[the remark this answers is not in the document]
malikchaya2: Should they maybe be part of this section since it's part of the github page, and the user would expect that do be defaulted as well when the page is defaulted?

model/project_ref.go
```diff
@@ -1255,6 +1315,176 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				catcher.Add(RemoveProjectAlias(a.ID.Hex()))
+			}
+		}
+		if catcher.HasErrors() {
+			// log the changes before returning whatever errors were caught
+			_ = getAndLogProjectModified(projectId, userId, before)
+			return errors.Wrapf(catcher.Resolve(), "error defaulting to repo for section '%s'", section)
```
malikchaya2, line 1457: Do we maybe want to remove the return from here and let the return at the end of the function handle it, since that resolves the catcher as well?

model/project_ref.go
```diff
@@ -1255,6 +1315,176 @@ func (projectRef *ProjectRef) Upsert() error {
...
+	modified, err := saveProjectRefForSection(projectId, nil, section)
+	if err != nil {
+		return errors.Wrapf(err, "error defaulting project ref to repo for section '%s'", section)
+	}
+
+	// Handle sections that modify collections outside of the project ref.
+	// Handle errors at the end so that we can still log the project as modified, if applicable.
+	catcher := grip.NewBasicCatcher()
+	switch section {
+	case ProjectRefVariablesSection:
```
malikchaya2, line 1435: do we want to set modified to true here since the saveProjectRefForSection doesn't do it? Or do we specifically only want to log it the ProjectRef collection was touched as opposed to just the project settings page in general?

model/project_ref.go
```diff
@@ -316,6 +321,20 @@ const (
...
 
+type ProjectRefSection string
+
+const (
+	ProjectRefGeneralSection        = "general"
+	ProjectRefAccessSection         = "access"
+	ProjectRefVariablesSection      = "variables"
+	ProjectRefGithubAndCQSection    = "github_and_commit_queue"
+	ProjectRefNotificationsSection  = "notifications"
+	ProjectRefAliasSection          = "alias"
```
malikchaya2, line 332: The section in the UI is specifically only for patch aliases. Would it be possible to change this to `PatchAliasSection  = "patch_alias"`? I think that will make things clearer since we have other kinds of aliases sprinkled throughout other pages.

model/project_ref.go
```diff
@@ -1255,6 +1315,176 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				"$set": bson.M{
+					projectRefTriggersKey: p.Triggers,
+				},
+			})
+	case ProjectRefAliasSection:
+		err = db.Update(ProjectRefCollection,
+			bson.M{ProjectRefIdKey: projectId},
+			bson.M{
+				"$set": bson.M{
+					projectRefPatchTriggerAliasesKey: p.PatchTriggerAliases,
```
malikchaya2, line 1394: I think that we decided to put patchTriggerAliases in the github section.

model/project_ref.go
```diff
@@ -1255,6 +1315,176 @@ func (projectRef *ProjectRef) Upsert() error {
...
+			})
+	case ProjectRefVariablesSection:
+		// this section doesn't modify the project ref
+		return false, nil
+	default:
+		return false, errors.Errorf("invalid section")
+	}
+
+	if err != nil {
+		return false, errors.Wrapf(err, "error defaulting to repo for section '%s'", ProjectRefGithubAndCQSection)
```
malikchaya2, line 1411: You probably meant for that to be ProjectRefSection instead of ProjectRefGithubAndCQSection

model/project_ref.go
```diff
@@ -1255,6 +1315,176 @@ func (projectRef *ProjectRef) Upsert() error {
...
+					privateVarsMapKey:    1,
+					restrictedVarsMapKey: 1,
+				},
+			})
+
+		catcher.Wrapf(err, "error defaulting to repo for section '%s'", section)
+	case ProjectRefGithubAndCQSection:
+		for _, a := range before.Aliases {
+			// remove only internal aliases
+			if utility.StringSliceContains(evergreen.InternalAliases, a.Alias) {
```
malikchaya2, line 1450: should this be specifying the aliases with the __commit_queue alias label?

malikchaya2: Or, are all the internal aliases going to be Github or Commit queue aliases, and therefor all be defaulted on this page? 

I'm guessing this is only the case if we do indeed move the patchTriggerAliases to the github page.

model/project_ref.go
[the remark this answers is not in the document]
malikchaya2: moving this conversation to a different thread and resolving this one.

model/project_ref.go
ablack12: sure thing

model/project_ref.go
ablack12: Sorry I should've taken notes on our conversation. Didn't we decide to put it in the same section it relied on? (Which I thought was this one because we defined the alias here and then used it in patch trigger aliases but I'm pretty confused now). I'm going to ping Lara and see what she wrote down.

model/project_ref.go
ablack12: yes thank you! this error in general is actually wrong

model/project_ref.go
ablack12: yes, thank you!

model/project_ref.go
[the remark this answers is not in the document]
ablack12: For sure

model/project_ref.go
ablack12: Yes, if you look at the InternalAliases list it includes commit queue aliases, github PR aliases, github check aliases, and git tag aliases, which are all part of this section.

model/project_ref.go
```diff
@@ -1255,6 +1304,180 @@ func (projectRef *ProjectRef) Upsert() error {
...
+				},
+			})
+		if err != nil {
+			return errors.Wrapf(err, "error defaulting to repo for section '%s'", section)
+		}
+		catcher := grip.NewBasicCatcher()
+		for _, a := range before.Aliases {
+			// remove only internal aliases
+			if utility.StringSliceContains(evergreen.InternalAliases, a.Alias) {
+				catcher.Add(RemoveProjectAlias(a.ID.Hex()))
```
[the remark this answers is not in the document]
ablack12: I'm okay with this since we'll be logging the changes in the event log

model/project_ref.go
ablack12: yep I did this wrong, fixed

model/project_ref.go
ablack12: I talked to Lara, she also remembers it being in this section for the same reason but can't find her notes on the subject. We may need to re have the design conversation with her to iron it back out; for now let's leave it here and we can re-visit if necessary.

malikchaya2 (approved):

PR #4840, opened 2021-07-15
EVG-8289 Support loading mainline commits on spruce

khelif96:
[EVG-8289](https://jira.mongodb.org/browse/EVG-8289)

### Description 
On spruce We incorrectly used a patch interchangeably with a version. This is problematic because while every patch can have a version not every version has a patch (Mainline commits). This PR rectifies that misrepresentation and adds resolvers that return a version properly as well as adding any fields we may need to represent the data that belongs to a version or its tasks. These will be used by the UI to allow it to properly display versions that do not have patches

This pr also maintains backwards compatibility with the current version of spruce to ensure we can revert if necessary/ and or roll out these changes individually

PR #4880, opened 2021-07-28
EVG-15134 Only resolve statuses for base level patches

khelif96:
[EVG-15134](https://jira.mongodb.org/browse/EVG-15134)

### Description 
This reverts a significant performance regression  introduced in https://github.com/evergreen-ci/evergreen/pull/4840 
Patch status will only be calculated when viewing a patch/version page and will not be calculated elsewhere such as the mypatches pages.
Calculating patch status on the my patches page is an expensive operation because it needs to query for every visible patches tasks to determine if a patch should have an aborted status. Which lead to significant load on the db. 

Note: the status on the my patches page has the potential to not be in sync with the patch status on the version page, which represents the same behavior as before  the change in #4840 was introduced so this is not a functionality regression.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -971,42 +971,41 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
 	return *obj.Id, nil
 }
 
-func (r *patchResolver) Status(ctx context.Context, obj *restModel.APIPatch) (string, error) {
-	failedAndAbortedStatuses := append(evergreen.TaskFailureStatuses, evergreen.TaskAborted)
-	opts := data.TaskFilterOptions{
-		Statuses:        failedAndAbortedStatuses,
-		FieldsToProject: []string{task.DisplayStatusKey},
-	}
-	tasks, _, err := r.sc.FindTasksByVersion(*obj.Id, opts)
-	if err != nil {
-		return "", InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch tasks for patch: %s ", err.Error()))
-	}
-	status := *obj.Status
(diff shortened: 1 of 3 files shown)


graphql/resolvers.go
```diff
@@ -971,42 +971,41 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
...
-	if len(obj.ChildPatches) > 0 {
-		for _, cp := range obj.ChildPatches {
-			// add the child patch tasks to tasks so that we can consider their status
-			childPatchTasks, _, err := r.sc.FindTasksByVersion(*cp.Id, opts)
-			if err != nil {
-				return "", InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch tasks for patch: %s ", err.Error()))
-			}
-			tasks = append(tasks, childPatchTasks...)
+
+	if evergreen.IsFinishedPatchStatus(*patch.Status) {
```
bsamek, line 980: I would have expected this code to look like https://github.com/evergreen-ci/evergreen/blob/c7d7bb97a84b3db534e5149d33a1070dbd86d183/graphql/resolvers.go#L974-L1010 if this is a revert, but I don't see `IsFinishedPatchStatus` there.

I don't need to be blocking though, since I don't have context on this change.

bsamek (commented):
I would expect

graphql/resolvers.go
khelif96: I added this in to try and improve the performance a little over the original implementation so it would only try to resolve  the patch status when the patch has completed.

arjunpatel100 (approved):

ablack12 (approved):

PR #4877, opened 2021-07-28
EVG-15099  Update priority for child patches when updating the parent 

malikchaya2:
[EVG-15099](https://jira.mongodb.org/browse/EVG-15099)

### Description 
If someone sets the priority for a patch, it should also set the priority for child patches 

### Note 
This will only update the patch priority if the child patch has already been created. In the case where the child patch waits on the parent to finish in order to finalize, the child patch won't have that priority. I don't know if this is significant enough to warrant a bigger refactor to be able to store the priority and then add it when the child gets created. 

### Testing 
Manually tested on staging.

PR #4881, opened 2021-07-30
EVG-14974: setPatchPriority accepts multiple patch IDs

sophstad:
[EVG-14974](https://jira.mongodb.org/browse/EVG-14974)

### Description 
- Add `SetVersionPriority` resolver which accepts multiple version IDs to be updated
- Will remove `SetPatchPriority` once all usage on the front end has been converted to `SetVersionPriority`

### Testing 
- Add resolver tests for `setVersionPriority`
- Keep tests for `setPatchPriority` to ensure backwards compatibility


graphql/resolvers.go
```diff
@@ -1979,16 +1979,21 @@ func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abo
 	return &patchID, nil
 }
 
-func (r *mutationResolver) SetPatchPriority(ctx context.Context, patchID string, priority int) (*string, error) {
+func (r *mutationResolver) SetPatchPriority(ctx context.Context, patchIds []string, priority int) ([]string, error) {
```
khelif96, line 1982: I know this pr isn't complete yet but this would result in a non backwards compatible api change.  I think this is a good opportunity to rename this resolver to `SetVersionPriority` since that is a more accurate description of what it does.

khelif96: In order to keep this backwards compatible i would keep this resolver as is and create a second resolver called `SetVersionPriority` that accepts the updated parameters. In a follow up PR we can remove the `SetPatchPriority` resolver when the front end changes have been deployed.

malikchaya2 (commented):
Should we refactor this considering your refactor?
https://github.com/evergreen-ci/evergreen/pull/4877

malikchaya2 (commented):
Alternatively, since I already refactored ModifyVersion to modify child patches as well, do we want to re-use that instead?

sophstad:
Confirmed that #4877 disables downstream tasks since this is accomplished by setting priority to -1. Closing as this already works!

PR #4882, opened 2021-07-30
EVG-15111: event_log ttl db cleanup job

hhoke:
[EVG-15111](https://jira.mongodb.org/browse/EVG-15111)

### Description 
I wanted to create a cron job to TTL TASK, SCHEDULER, and PATCH r_type items for event_log, because a TTL index cannot be compound. Most of the effect of this could be accomplished by simply using a TTL index with a partial filter match for r_type of TASK, because they are much more numerous than the others (1.5TB over the past six years, compared with <100 GB for SCHEDULER and PATCH). However, I want to be consistent and as general as possible, to prevent future problems. While I was in the codebase, I figured I should depastify some existing data cleanup code. The resulting amboy job takes in a collection and a TTL days parameter, and uses deletion filters specific to each collection name.

### Testing 
  I will run this on staging after this passes code review but before I merge, monitoring the contents of event_log. There are significantly fewer TASK-type events in the staging DB, so I may need to create some if a test of the bulk write capacity is indicated.


units/db_cleanup_general.go
```diff
@@ -0,0 +1,178 @@
...
+	cleanupBatchSize = 100 * 1000
+)
+
+func init() {
+	registry.AddJobType(dbCleanupJobName, func() amboy.Job {
+		return makeDbCleanupJob()
+	})
+}
+
+type dataCleanup struct {
```
Kimchelly, line 36: Rather than multiplex all the different data cleanup jobs into 1 job, can we instead keep them as 3 separate jobs and just embed a common job object in all of them, and dependency inject the relevant TTL cleanup function? Muxing them all into 1 implementation makes them hard to change in the future (e.g. if one needs to have some special behavior that's not shared with the other ones).

I would do something like this to inject the relevant deletion logic while still re-using all the other common logic:
```go
type DataCleanupJobBase struct {
    CollectionName string `bson:"collection_name"`
}

func (b *DataCleanupJobBase) runWithDeleteFn(ctx context.Context, deleteFn func(ctx context.Context, env evergreen.Environment, ts time.Time, batchSize int)) {
    // ... all the logic common to all the Run methods
    numDeleted, err := deleteFn(ctx, j.env, ts, batchSize)
    // ... all the remaining logic common to all the Run methods
}

type testLogCleanupJob struct {
    job.Base `bson:"metadata"`
    DataCleanup DataCleanupJobBase `bson:"data_cleanup"`
}

func (j *testLogCleanupJob) Run(ctx context.Context) {
    // do whatever initial job populating logic is necessary...

    j.DataCleanup.runWithDeleteFn(ctx, model.DeleteTestLogsWithLimit)
}
```

Kimchelly (changes requested):

units/db_cleanup_general.go
```diff
@@ -0,0 +1,178 @@
...
+
+	j.SetDependency(dependency.NewAlways())
+
+	return j
+}
+
+func NewDbCleanupJob(ts time.Time, collectionName string, ttlDays int) amboy.Job {
+	j := makeDbCleanupJob()
+	j.SetID(fmt.Sprintf("%s.%s", dbCleanupJobName, ts.Format(TSFormat)))
+	j.UpdateTimeInfo(amboy.JobTimeInfo{MaxTime: time.Minute})
```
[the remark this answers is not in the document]
hhoke: For my crontab job (not on evergreen) I used 500 iterations of 7k jobs, with a second sleep time, and it took about 15 minutes to finish. (7k*500)/15 is about 200k items/ minute, so it checks out.

units/db_cleanup_general.go
```diff
@@ -0,0 +1,178 @@
...
+		var filter map[string]interface{}
+		if j.CollectionName == testresult.Collection {
+			filter = bson.M{"_id": bson.M{"$lt": primitive.NewObjectIDFromTimestamp(ts)}}
+		} else if j.CollectionName == "event_log" {
+			rTypes := []string{
+				"TASK",
+				"SCHEDULER",
+				"PATCH",
+			}
+			filter = bson.M{"_id": bson.M{"$lt": primitive.NewObjectIDFromTimestamp(ts)}, "r_type": bson.M{"$in": rTypes}}
```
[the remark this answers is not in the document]
hhoke: I'll use ts since that's more conservative.

units/db_cleanup_general.go
hhoke: I agree the if/else case switching is inelegant, and may present problems. I think the best solution is to define the filters where the function is called, and pass them in. This should provide more abstraction without requiring largely identical deletion functions

units/db_cleanup_general.go
```diff
@@ -0,0 +1,178 @@
...
+func init() {
+	registry.AddJobType(dbCleanupJobName, func() amboy.Job {
+		return makeDbCleanupJob()
+	})
+}
+
+type dataCleanup struct {
+	job.Base       `bson:"metadata" json:"metadata" yaml:"metadata"`
+	CollectionName string `bson:"collection_name" json:"collection_name" yaml:"collection_name"`
+	TTLDays        int    `bson:"ttl_days" json:"ttl_days" yaml:"ttl_days"`
```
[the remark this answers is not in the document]
hhoke: I'll make it a duration

units/db_cleanup_general.go
Kimchelly: The problem is that you can't serialize a function object into BSON, so when it round-trips through the DB, the TTLFilter will be ignored. I think my proposal doesn't have any copy-pasted code - if you want, you can consolidate the deletion functions by passing the `FilterFunc` to the common base job as I originally suggested.

ybrill:
I realize this is late in the game (and maybe we talked about this already), but could the we achieve the TTLs inside the DB with a partial TTL indexes?
Something like
```
db.event_log.createIndex( { "ts": 1 }, { expireAfterSeconds: 7776000, partialFilterExpression: { r_type: { $in: ["TASK", "SCHEDULER", "PATCH"] } } } )
```

hhoke:
> I realize this is late in the game (and maybe we talked about this already), but could the we achieve the TTLs inside the DB with partial TTL indexes?
> Something like
> 
> ```
> db.event_log.createIndex( { "ts": 1 }, { expireAfterSeconds: 7776000, partialFilterExpression: { r_type: { $in: ["TASK", "SCHEDULER", "PATCH"] } } } )
> ```

One [cannot](https://docs.mongodb.com/manual/core/index-partial/) use `{ $in: ["TASK", "SCHEDULER", "PATCH"] }` in a partial filter expression. We could just use `r_type: {$eq: "TASK}`, and get most of the effect.

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+const (
+	cleanupBatchSize = 100 * 1000
+)
+
+type FilterFunc func(time.Time) map[string]interface{}
+
+type DataCleanupJobBase struct {
+	CollectionName string        `bson:"collection_name" json:"collection_name" yaml:"collection_name"`
+	TTL            time.Duration `bson:"ttl" json:"ttl" yaml:"ttl"`
+	Errors         []error       `bson:"errors" json:"errors" yaml:"errors"`
```
Kimchelly, line 24: Errors are actually non-marshallable - they're more like an interface (for anything that implements the `Error` method) so you would have to store these as their string representations if you want them to persist in the DB.

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+		select {
+		case <-ctx.Done():
+			break LOOP
+		default:
+			if time.Since(startAt) >= 50*time.Second {
+				break LOOP
+			}
+			opStart := time.Now()
+			num, err := b.deleteCollectionWithLimit(ctx, timestamp, cleanupBatchSize, filterFunc)
+			b.Errors = append(b.Errors, err)
```
Kimchelly, line 57: Is there a reason you accumulate the errors but don't use it anywhere? It seems like this could be replaced with `j.AddError`.

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+		"run_end_at":         time.Now(),
+		"num_batches":        batches,
+		"num_docs":           numDocs,
+		"time_spent_seconds": timeSpent.Seconds(),
+	}, b.Errors
+}
+
+func (b *DataCleanupJobBase) deleteCollectionWithLimit(ctx context.Context, ts time.Time, limit int, filterFunc FilterFunc) (int, error) {
+	if limit > 100*1000 {
+		panic("cannot delete more than 100k documents in a single operation")
```
Kimchelly, line 87: I would avoid panicking - in go, panicking is typically a disaster case and not a normal error case - I would just return an error.

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+			opStart := time.Now()
+			num, err := b.deleteCollectionWithLimit(ctx, timestamp, cleanupBatchSize, filterFunc)
+			b.Errors = append(b.Errors, err)
+
+			batches++
+			numDocs += num
+			timeSpent += time.Since(opStart)
+			if num < cleanupBatchSize {
+				break
```
Kimchelly, line 63: I know this is copy-pasted from the original, but is this supposed to be `break LOOP` to stop the infinite loop?

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+type DataCleanupJobBase struct {
+	CollectionName string        `bson:"collection_name" json:"collection_name" yaml:"collection_name"`
+	TTL            time.Duration `bson:"ttl" json:"ttl" yaml:"ttl"`
+	Errors         []error       `bson:"errors" json:"errors" yaml:"errors"`
+
+	env evergreen.Environment
+}
+
+func (b *DataCleanupJobBase) RunWithDeleteFn(ctx context.Context, filterFunc FilterFunc) (message.Fields, []error) {
+
```
Kimchelly, line 30: nit: remove random whitespace

units/db_cleanup_general.go
```diff
@@ -0,0 +1,101 @@
...
+type DataCleanupJobBase struct {
+	CollectionName string        `bson:"collection_name" json:"collection_name" yaml:"collection_name"`
+	TTL            time.Duration `bson:"ttl" json:"ttl" yaml:"ttl"`
+	Errors         []error       `bson:"errors" json:"errors" yaml:"errors"`
+
+	env evergreen.Environment
+}
+
+func (b *DataCleanupJobBase) RunWithDeleteFn(ctx context.Context, filterFunc FilterFunc) (message.Fields, []error) {
```
Kimchelly, line 29: It doesn't seem like this method needs to be exported since no other package uses it - can you make this `runWithDeleteFn`?

Kimchelly (changes requested):

units/db_cleanup_general.go
hhoke: there's no `j` to add the error to, though

hhoke: I return them at the end, and then add them to j.Errors e.g. db_cleanup_event_log.go   line 85

Kimchelly (approved):

PR #4883, opened 2021-07-30
EVG-13490 add versionMetadata field which resolves a version

khelif96:
[EVG-13490](https://jira.mongodb.org/browse/EVG-13490)

### Description 
Adds a versionMetadata field which resolves versions for tasks. Decided to do it in a backwards compatible way. This will be renamed to version in a follow up PR. Any fields marked deprecated will also be removed in a follow up pr.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2427,6 +2427,7 @@ func (r *taskResolver) FailedTestCount(ctx context.Context, obj *restModel.APITa
 	return failedTestCount, nil
 }
 
+// TODO: DEPRICATE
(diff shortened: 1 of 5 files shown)


malikchaya2 (approved):

graphql/resolvers.go
```diff
@@ -2482,7 +2482,7 @@ func (r *taskResolver) FailedTestCount(ctx context.Context, obj *restModel.APITa
 	return failedTestCount, nil
 }
 
-// TODO: DEPRICATE
+// TODO: deprecated
```
malikchaya2, line 2485: super nit: do we want the verb form deprecate since it's a todo?

khelif96: we probably do lol but i will remove this soon anyways so i think its alright

PR #4884, opened 2021-07-30
EVG-14777: Handle display tasks when skipping stranded BFG creation

hadjri:
[EVG-14777](https://jira.mongodb.org/browse/EVG-14777)

### Description 
Change was made previously to skip Jira BFG ticket creation if a task's description details was 'stranded' since this is indicative only of a host being taken by AWS and not a real issue.  Additional change has been introduced to skip BFG ticket creation for failed display tasks if the execution task failures underneath it only have 'stranded' as their task details description to handle display tasks as well.


trigger/task.go
```diff
@@ -328,7 +328,23 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 	var payload interface{}
 	if sub.Subscriber.Type == event.JIRAIssueSubscriberType {
 		// We avoid creating BFG ticket in the case that the task is stranded to reduce noise for the Build Baron
-		if t.task.Details.Description == evergreen.TaskDescriptionStranded {
+		// If task is display, we skip ticket creation if all execution task failures are only 'stranded'
+		shouldSkipTicket := false
+		if t.task.DisplayOnly {
+			for _, executionTask := range t.task.ExecutionTasksFull {
```
ablack12, line 334: Sorry that this is confusing. UI added a few fields to the Task struct that are really only used for the UI / aren't stored in the database and should be removed (discussing this with Mohamed). Instead you'll need to get all execution tasks from the DB and then go through the loop (can use WithFields to just get the details though since that's all you need here).

ablack12 (changes requested):

ablack12 (approved):

PR #4886, opened 2021-07-30
EVG-15029: Make task_execution required for /rest/v2/tasks/{task_id}/annotation

hadjri:
[EVG-15029](https://jira.mongodb.org/browse/EVG-15029)

### Description 
Change has been made to require task_execution in the req body for this endpoint and to error otherwise.  The code to search for the latest execution has been removed.

### Testing
Additional unit test cases were added to ensure that an error is thrown if task execution is omitted and that an error is thrown if invalid data type is passed in.


malikchaya2 (commented):
Can we please add a test or edit existing tests in annotations_test.go?

rest/route/annotations.go
```diff
@@ -247,12 +247,22 @@ func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
...
 			StatusCode: http.StatusBadRequest,
 		}
 	}
+	if taskExecutionsAsString == "" {
+		return gimlet.ErrorResponse{
+			Message:    "task execution cannot be empty",
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+	if taskExec, err := strconv.Atoi(taskExecutionsAsString); err != nil {
```
malikchaya2, line 263: Can we please return if there's an error?

rest/route/annotations.go
```diff
@@ -247,12 +247,22 @@ func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
 func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request) error {
 	var err error
 	h.taskId = gimlet.GetVars(r)["task_id"]
+	taskExecutionsAsString := gimlet.GetVars(r)["task_execution"]
```
malikchaya2, line 250: to keep this consistent with other parse functions, we can just call this `execution`.

malikchaya2 (commented):
For future PRs, it's preferable to keep the "Testing" section in the description. If no testing was done, you can use that section to explain your reasoning for why you don't think testing is necessary.

rest/route/annotations_test.go
```diff
@@ -376,17 +376,18 @@ func TestAnnotationByTaskPutHandlerParse(t *testing.T) {
 	execution0 := 0
 	execution1 := 1
 	a := &model.APITaskAnnotation{
-		Id:     utility.ToStringPtr("1"),
-		TaskId: utility.ToStringPtr("TaskFailedId"),
-		Note:   &model.APINote{Message: utility.ToStringPtr("task-1-note_0")},
+		Id:            utility.ToStringPtr("1"),
+		TaskId:        utility.ToStringPtr("TaskFailedId"),
+		Note:          &model.APINote{Message: utility.ToStringPtr("task-1-note_0")},
+		TaskExecution: utility.ToIntPtr(1),
```
malikchaya2, line 382: Can we please also add a test where an annotation doesn't have an execution and make sure that it errors?

rest/route/annotations.go
```diff
@@ -247,12 +247,27 @@ func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
...
+		}
+	}
+	taskExec, err := strconv.Atoi(taskExecutionsAsString)
+	if err != nil {
+		return gimlet.ErrorResponse{
+			Message:    "cannot convert task execution to integer",
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+	h.annotation.TaskExecution = &taskExec
```
ablack12, line 270: TaskExecution doesn't need to be a pointer anymore. This was only necessary before because we needed to know if it wasn't provided, but now it will always be provided so we can just set it at 263 directly.

hadjri:
> For future PRs, it's preferable to keep the "Testing" section in the description. If no testing was done, you can use that section to explain your reasoning for why you don't think testing is necessary.

Got it.  Re-added testing section

malikchaya2 (commented):
Now that execution is required, we might want to do something like this for execution as well: 
https://github.com/evergreen-ci/evergreen/pull/4886/files#diff-e69d34eba4380397dd99b9bf9191455c05769d39cf249f17db77845af58c09faR313-R322

rest/route/annotations.go
```diff
@@ -247,12 +247,26 @@ func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
...
 	var err error
 	h.taskId = gimlet.GetVars(r)["task_id"]
+	taskExecutionsAsString := gimlet.GetVars(r)["execution"]
 	if h.taskId == "" {
 		return gimlet.ErrorResponse{
 			Message:    "task ID cannot be empty",
 			StatusCode: http.StatusBadRequest,
 		}
 	}
+	if taskExecutionsAsString == "" {
```
malikchaya2, line 257: I spoke to David about this, and he said that they only ever use the body of the request to set the execution and that reading the docs for the endpoint, wasn't even aware that they could send it as part of the url. We should instead of erring here, check if it's set in the body and only err if it's not set anywhere. We should also update the docs to say that the execution can be set in the url.

hadjri:
> Now that execution is required, we might want to do something like this for execution as well:
> https://github.com/evergreen-ci/evergreen/blob/483b65805b60872bd90a52a1cb03384964e59c38/rest/route/annotations.go#L313-L321

Would you mean checking to see if the task's Execution value is the same as what was passed in in the request? The code at this point should guarantee that the task execution is not nil at the end of the method if no error is thrown.

rest/route/annotations.go
```diff
@@ -247,12 +247,26 @@ func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
 func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request) error {
 	var err error
 	h.taskId = gimlet.GetVars(r)["task_id"]
+	taskExecutionsAsString := gimlet.GetVars(r)["execution"]
```
malikchaya2, line 250: I think you might need `r.URL.Query().Get("execution")` instead of GetVars here.

malikchaya2:
> Would you mean checking to see if the task's Execution value is the same as what was passed in in the request? The code at this point should guarantee that the task execution is not nil at the end of the method if no error is thrown.

Yes, that's what I meant. We want to make sure that someone doesn't pass in ?execution=1 in the url but execution: 0 in the body.

rest/route/annotations.go
ablack12: Can we instead only allow it in the body and not in the URL?

ablack12: Or is this inconsistent with other routes ( / can we make them all consistent)

hadjri: Updated doc: https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#create-or-update-a-new-task-annotation

malikchaya2: It would definitely make it cleaner and less ambiguous to only take the execution in the body, but yes, taking the execution in the url seems to be the convention based on other routes.

rest/route/annotations.go
```diff
@@ -265,8 +266,7 @@ func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request)
 	}
 
 	// check if the task exists
-	// t, err := task.FindOne(task.ById(h.taskId))
-	t, err := task.FindByIdExecution(h.taskId, h.annotation.TaskExecution)
+	t, err := task.FindOne(task.ById(h.taskId))
```
malikchaya2, line 269: We need this to check both. This fixes a bug where `if !evergreen.IsFailedTaskStatus(t.Status) ` on line 279 doesn't allow for annotation to be filed if a task fails for execution 0 but succeeds for execution 1.

rest/route/annotations.go
```diff
@@ -297,10 +297,30 @@ func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request)
 		}
 	}
 
-	// set TaskExecution to the latest execution
-	if h.annotation.TaskExecution == nil {
-		h.annotation.TaskExecution = &t.Execution
+	if h.annotation.TaskExecution == 0 && taskExecutionsAsString == "" {
```
malikchaya2, line 300: it can be 0 because someone set it to the 0 execution. I think we do need a pointer still because a user can either set it in the body or the url and either is valid. cc @ablack12

rest/route/annotations.go
malikchaya2: Commented below. We still need it because it can be null in the body but provided in the url, or vice versa.

rest/route/annotations.go
hadjri: True, because as it stands if execution is omitted from the url it will be treated as 0.

rest/route/annotations.go
hadjri: Added back to find by execution

rest/route/annotations.go
```diff
@@ -264,8 +265,33 @@ func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request)
...
+			h.annotation.TaskExecution = &taskExecution
+		} else if *h.annotation.TaskExecution != taskExecution {
+			return gimlet.ErrorResponse{
+				Message:    "Task execution must equal the task execution specified in the annotation",
+				StatusCode: http.StatusBadRequest,
+			}
+		}
+	}
+
+	if h.annotation.TaskExecution == nil && taskExecutionsAsString == "" {
```
malikchaya2, line 287: At this point, h.annotation.TaskExecution because of the assignment on `278` will already be equal to either what was in the body or the url. So you should be able to just check if ` h.annotation.TaskExecution == nil`.

hadjri: h.annotation.TaskExecution will only be set if the execution string from the url is not empty so line 278 would not be hit if the string is empty.  This line serves to throw an error only if both fields are never set so I think this boolean statement should remain.

malikchaya2: We do still need the boolean statement, but I'm wondering if we need both sides of the && in order to check "has the user not specified it anywhere" considering the check on 277.

hadjri: > We do still need the boolean statement, but I'm wondering if we need both sides of the && in order to check "has the user not specified it anywhere" considering the check on 277.

Simplified the statement

malikchaya2 (approved):

PR #4890, opened 2021-07-30
EVG-15118 Modifying a version should also modify the child patches 

malikchaya2:
[EVG-15118](https://jira.mongodb.org/browse/EVG-15118)

### Description 
This takes care of scheduling/unscheduling tasks for child patches when it's done for the parent patch. It made more sense to do it ModifyVersionHandler where it handles but scheduling/unscheduling and set priority so that code does not have to be repeated (especially with all the casing). Restart is handled separately because we need to take input from the user. 

### Testing 
Will test in staging

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -766,11 +766,7 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 				return ResourceNotFound.Send(ctx, fmt.Sprintf("patch not found %s: %s", patchID, err.Error()))
 			}
 			if p.IsParent() {
-				childPatchIds := []string{}
-				if p != nil && err != nil {
-					childPatchIds = p.Triggers.ChildPatches
-				}
-				for _, childPatchId := range childPatchIds {
+				for _, childPatchId := range p.Triggers.ChildPatches {
 					err = ModifyVersionHandler(ctx, dataConnector, childPatchId, modifications)
 					if err != nil {
 						return errors.Wrap(mapHTTPStatusToGqlError(ctx, httpStatus, err), fmt.Sprintf("error modifying child patch '%s'", patchID))
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -753,7 +753,7 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 		return mapHTTPStatusToGqlError(ctx, httpStatus, err)
 	}
 
-	if version.Requester == evergreen.PatchVersionRequester || version.Requester == evergreen.GithubPRRequester {
+	if evergreen.IsPatchRequester(version.Requester) {
 		// restart is handled through graphql because we need the user to specify
 		// which downstream tasks they want to restart
 		if modifications.Action != Restart {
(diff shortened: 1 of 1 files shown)


graphql/util.go
```diff
@@ -764,6 +752,25 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 	if err != nil {
 		return mapHTTPStatusToGqlError(ctx, httpStatus, err)
 	}
+
+	if modifications.Action != Restart {
```
ablack12, line 756: can we include  the reason you listed in the description for restart being different as a comment here? And can we check if the version requester is for a patch before running the patch query? (Then we can error if p == nil also)

ablack12: Also i'm assuming there's no such thing as nested child patches, in which case we could also check that ParentPatchID == "" so we don't call this unnecessarily for child patches.

ablack12 (changes requested):

graphql/util.go
malikchaya2: done. 

> we could also check that ParentPatchID == ""
I wrapped it in `if p.IsParent()`

graphql/util.go
```diff
@@ -764,6 +752,31 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 	if err != nil {
 		return mapHTTPStatusToGqlError(ctx, httpStatus, err)
 	}
+
+	if version.Requester == evergreen.PatchVersionRequester || version.Requester == evergreen.GithubPRRequester {
```
ablack12, line 756: can we use IsPatchRequester instead? (unless we explicitly don't want to include commit queue, although then we should explain that in a comment)

graphql/util.go
```diff
@@ -764,6 +752,31 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
...
+		// restart is handled through graphql because we need the user to specify
+		// which downstream tasks they want to restart
+		if modifications.Action != Restart {
+			//do the same for child patches
+			p, err := patch.FindOneId(patchID)
+			if err != nil {
+				return ResourceNotFound.Send(ctx, fmt.Sprintf("error finding patch %s: %s", patchID, err.Error()))
+			}
+			if p == nil {
+				return ResourceNotFound.Send(ctx, fmt.Sprintf("patch not found %s: %s", patchID, err.Error()))
```
ablack12, line 766: This shouldn't have err.Error() since it's nil

ablack12: can this be "patch '%s' not found" instead?

ablack12 (changes requested):

graphql/util.go
malikchaya2: done

graphql/util.go
malikchaya2: done

graphql/util.go
ablack12: oh gotcha i misunderstood

ablack12 (approved):

PR #4895, opened 2021-08-02
EVG-15163 Fix cannot annotate tasks that have been retried

malikchaya2:
[EVG-15163](https://jira.mongodb.org/browse/EVG-15163)

### Description 
This fetches a task by execution so that the status restriction applies to the execution in question and not the latest one. 
### Testing 
Tested in staging and made sure it doesn't fail with a task that first failed and then didn't.

### DIFF
--- a/rest/route/annotations.go
+++ b/rest/route/annotations.go
@@ -254,8 +254,19 @@ func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request)
 		}
 	}
 
+	body := util.NewRequestReader(r)
+	defer body.Close()
+	err = json.NewDecoder(body).Decode(&h.annotation)
+	if err != nil {
+		return gimlet.ErrorResponse{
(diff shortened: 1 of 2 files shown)


rest/route/annotations.go
```diff
@@ -255,7 +255,7 @@ func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request)
 	}
 
 	// check if the task exists
-	t, err := task.FindOne(task.ById(h.taskId))
+	t, err := task.FindByIdExecution(h.taskId, h.annotation.TaskExecution)
```
ablack12, line 258: h.annotation is nil at this point, since we don't read it in until 275

ablack12 (changes requested):

rest/route/annotations.go
malikchaya2: yeah I just realized that, my bad. I'm working on a fix.

malikchaya2 (commented):
Making some changes, please hold off on the review.

ablack12 (approved):

PR #4900, opened 2021-08-04
PoC:stepbacks given analogous priority bump to patch tasks

hhoke:

### Description 
Quick proposal for change per https://mongodb.slack.com/archives/G979SL0HM/p1627661411002300
Made into draft pull for readability

### Testing 


bsamek:
We currently give patch tasks higher priority than mainline commits, which gives people faster turnaround times. How can we mitigate the risk that patch build wait times slow down?

ablack12:
Since we decided to do this with a new configurable factor, I think we can close this POC.

PR #4901, opened 2021-08-04
Evg 14823

hadjri:
[EVG-14823](https://jira.mongodb.org/browse/EVG-14823)

### Description 
This change was deployed but a week later it was raised that the change was triggering a race detector in one of our tests causing it to fail at random times, and we rolled back.  Previously the "successful task" and "failed task" lifecycle tests in TestTaskLifecycleEndpoints were using the same task, but since this new change makes it so that the DisplayTask can be modified when running an end task command, it introduced the possibility of a data race which we saw intermittently.  Since it is not important for the purpose of the test, change has been made to have the "failed task" lifecycle test to use a different task/host from the "successful task" test to prevent a data race.

### Testing 
Test-service is passing locally and in this PR which should indicate that this test failure should not occur again.


ybrill (approved):

ybrill:
Don't forget to edit the title (to the standard `EVG-xxxx <description>`) before you merge!

PR #4903, opened 2021-08-04
EVG-15155  Fix downstream task's patch breadcrumb link is showing the wrong patch number

malikchaya2:
[EVG-15155](https://jira.mongodb.org/browse/EVG-15155)

### Description 
This sets the task's patch number to the parent's patch number so that the link is correct. This won't fix the old ones we have in staging, but it will be correct for new ones. (this is not used in prod yet)

### Testing 
Tested by creating a new patch with a child patch in staging and making sure that the patchNumber is correct.


malikchaya2 (commented):
I actually want to rethink my approach to see how I can do this without fetching the patch for every task. Please hold of on reviewing.

malikchaya2:
This is ready for review

sophstad (approved):
makes sense to me!

PR #4904, opened 2021-08-04
EVG-13902 Create resolver for manifest

khelif96:
[EVG-13902](https://jira.mongodb.org/browse/EVG-13902)

### Description 
This will resolve the manifest field for a version 
### Testing 
  graphql tests

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2913,13 +2913,7 @@ func (r *versionResolver) Manifest(ctx context.Context, v *restModel.APIVersion)
 	}
 	modules := map[string]interface{}{}
 	for key, module := range m.Modules {
-		modules[key] = &manifest.Module{
-			Branch:   module.Branch,
-			Repo:     module.Repo,
-			Revision: module.Revision,
-			Owner:    module.Owner,
-			URL:      module.URL,
-		}
+		modules[key] = module
 	}
 	versionManifest.Modules = modules
 
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2898,7 +2898,10 @@ type versionResolver struct{ *Resolver }
 func (r *versionResolver) Manifest(ctx context.Context, v *restModel.APIVersion) (*Manifest, error) {
 	m, err := manifest.FindFromVersion(*v.Id, *v.Project, *v.Revision, *v.Requester)
 	if err != nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Error finding manifest for version %s : %s", *v.Id, err.Error()))
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error fetching manifest for version %s : %s", *v.Id, err.Error()))
+	}
+	if m == nil {
+		return nil, nil
 	}
 	versionManifest := Manifest{
 		ID:              m.Id,
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -20,6 +20,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/distro"
 	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
+	"github.com/evergreen-ci/evergreen/model/manifest"
 	"github.com/evergreen-ci/evergreen/model/patch"
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/model/testresult"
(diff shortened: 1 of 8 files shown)


graphql/resolvers.go
```diff
@@ -2894,6 +2895,36 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
 
+func (r *versionResolver) Manifest(ctx context.Context, v *restModel.APIVersion) (*Manifest, error) {
+	m, err := manifest.FindFromVersion(*v.Id, *v.Project, *v.Revision, *v.Requester)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error fetching manifest for version %s : %s", *v.Id, err.Error()))
+	}
+	if m == nil {
+		return nil, nil
+	}
+	versionManifest := Manifest{
```
ablack12, line 2906: It's not clear to me why you're re-writing this to a new manifest. Why can't we just return m as a pointer?

khelif96: They are separate types. with the same name which is slightly confusing. 

`manifest.FindFromVersion` returns a `*manifest.Manifest`
And this is converting it to the generated graphql type also called Manifest. 
https://github.com/evergreen-ci/evergreen/pull/4904/files#diff-6da92836c5e2838b7c88391920b32232247de308e0b015ad02cb5154561ac4b2R317

khelif96: I don't think we should be using the Manifest db type for a graphql resolver.

ablack12: Why are we using interface{} instead of also defining Modules? 
Would it make sense to make an APIManifest instead? Is that how we avoid having things like version be redefined? (sorry, my memory on this process is shakey, but this looks unusual)

ablack12: In my memory it only makes sense to have a graphql type instead of an API struct when the structure itself looks different, but this is just a copy.

khelif96: So the way i understood the abstraction was 
We have a DBstruct that defines our database schema that we use in go to represent db documents.

We abstract this into an APIstruct that we  use in the API layer so we don't send db documents back to the user in our rest API. For the most part they contain the same fields that we use in our DBStructs, along side some other fields and utility functions to handle manipulating them. As well as converting to and from db structs. (BuildFromService, toService)

For our graphql api we define types in our `schema.graphql` and use `gqlgen` to generate go types for them. Usually they match or are similar enough to some struct we already have. so we tell `gqlgen` [to use that existing struct instead of creating one](https://github.com/evergreen-ci/evergreen/blob/a45252099c481a54987891b9094fa36486fc847e/gqlgen.yml) . We usually map our APIstructs to our graphql types since they are very similar and offer a layer of protection against returning and manipulating db types directly. 

We didn't have an APIManifest/ didn't need one for our rest API so it seemed unnecessary to create one if its only going to be used by our graphql resolvers. 

This is simply doing the `buildFromService` step that we usually do in our normal APIModels. And converting the db models directly to the desired end type.

khelif96: I used `interface{}` for modules because that was the default type provided by gqlgen to represent [map relationships](https://gqlgen.com/reference/scalars/#map). 
Technically the graphql spec [doesn't support free form maps with key value](https://github.com/graphql/graphql-spec/issues/101) pairs since it requires queries to be strongly typed. 

gqlgen provides this type as a utility to allow this type of relationship. And by default it is an interface{}
This was the only way to represent an identical structure to what modules is. (a map with `key: object` pairs)
The alternative would have been to convert it into an array of objects where each key:value pair is its own object. 

something like this
```
modules : [
    {
        key: "someModuleKey",
        value: {
             branch: "some_branch",
             repo: "some_repo",
             ....
         }
}
```

and then converting it back to an object client side.  

So using an interface{} seemed like the simplest way to accomplish this. But if you think the other way would have been preferable i don't mind switching.

ablack12: Wow okay that all makes sense. This map[string]item thing is _upsetting_. I'm surprised I haven't seen this before; I guess we don't store object maps as often as I thought. Interface should be fine and is preferable to changing a structure that users expect / defined themselves, assuming it'll still be printed correctly.

graphql/resolvers.go
```diff
@@ -2894,6 +2895,36 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+		ID:              m.Id,
+		Revision:        m.Revision,
+		Project:         m.ProjectName,
+		Branch:          m.Branch,
+		IsBase:          m.IsBase,
+		ModuleOverrides: m.ModuleOverrides,
+	}
+	modules := map[string]interface{}{}
+	for key, module := range m.Modules {
+		modules[key] = &manifest.Module{
```
ablack12, line 2916: sorry last question, can we just do module[key] = module? Or do we need a new pointer?

ablack12 (approved):

graphql/resolvers.go
khelif96: Yeah i was hoping to avoid it as well but this seemed like the most straight forward solution.

PR #4905, opened 2021-08-04
EVG-14366 Allow FindPatchesByProject to search by ProjectIdentifiers

bynn:
[EVG-14366](https://jira.mongodb.org/browse/EVG-14366)

### Description 
FindPatchesByProject can now search even if it's given an identifier

### Testing 
unittest

### DIFF
--- a/rest/data/patch.go
+++ b/rest/data/patch.go
@@ -39,7 +39,11 @@ type DBPatchConnector struct{}
 // FindPatchesByProject uses the service layer's patches type to query the backing database for
 // the patches.
 func (pc *DBPatchConnector) FindPatchesByProject(projectId string, ts time.Time, limit int) ([]restModel.APIPatch, error) {
+	apiPatches := []restModel.APIPatch{}
 	id, err := model.GetIdForProject(projectId)
+	if id == "" {
+		return apiPatches, nil
+	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "problem fetching project with id %s", projectId)
 	}
@@ -47,7 +51,6 @@ func (pc *DBPatchConnector) FindPatchesByProject(projectId string, ts time.Time,
 	if err != nil {
 		return nil, errors.Wrapf(err, "problem fetching patches for project %s", id)
 	}
-	apiPatches := []restModel.APIPatch{}
 	for _, p := range patches {
 		apiPatch := restModel.APIPatch{}
 		err = apiPatch.BuildFromService(p)
--- a/rest/data/patch.go
+++ b/rest/data/patch.go
@@ -41,9 +41,6 @@ type DBPatchConnector struct{}
 func (pc *DBPatchConnector) FindPatchesByProject(projectId string, ts time.Time, limit int) ([]restModel.APIPatch, error) {
 	apiPatches := []restModel.APIPatch{}
 	id, err := model.GetIdForProject(projectId)
-	if id == "" {
-		return apiPatches, nil
-	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "problem fetching project with id %s", projectId)
 	}
@@ -355,7 +352,7 @@ func (hp *MockPatchConnector) FindPatchesByProject(projectId string, ts time.Tim
 		}
 	}
 	if id == "" {
-		return patchesToReturn, nil
+		return nil, errors.Errorf("failed to find project '%s'", projectId)
 	}
 	for i := len(hp.CachedPatches) - 1; i >= 0; i-- {
 		p := hp.CachedPatches[i]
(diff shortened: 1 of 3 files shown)


rest/data/patch.go
```diff
@@ -342,7 +346,7 @@ func (hp *MockPatchConnector) FindPatchesByProject(projectId string, ts time.Tim
 	}
 	for i := len(hp.CachedPatches) - 1; i >= 0; i-- {
 		p := hp.CachedPatches[i]
-		if *p.ProjectId == projectId && !p.CreateTime.After(ts) {
+		if (*p.ProjectId == projectId || *p.ProjectIdentifier == projectId) && !p.CreateTime.After(ts) {
```
ablack12, line 349: instead can we can just check if *p.ProjectId == id

ablack12 (changes requested):

rest/data/patch.go
bynn: done

rest/data/patch.go
```diff
@@ -327,10 +331,20 @@ func (p *DBPatchConnector) GetPatchRawPatches(patchID string) (map[string]string
...
-	CachedPriority   map[string]int64
-	CachedRawPatches map[string]string
+	CachedPatches     []restModel.APIPatch
+	CachedProjectRefs []restModel.APIProjectRef
+	CachedAborted     map[string]string
+	CachedPriority    map[string]int64
+	CachedRawPatches  map[string]string
+}
+
+func (hp *MockPatchConnector) GetIdForProject(identifier string) (string, error) {
```
ablack12, line 341: I don't think you should make a separate connector function for this. The mock functions are meant to mock the function of the same name for the _real_ connector, and GetIdForProject is part of the FindPatchesByProject connector function, not its own function. I would embed this logic below.

ablack12 (changes requested):
almost there!

rest/data/patch.go
bynn: done

rest/data/patch.go
```diff
@@ -39,11 +39,18 @@ type DBPatchConnector struct{}
 // FindPatchesByProject uses the service layer's patches type to query the backing database for
 // the patches.
 func (pc *DBPatchConnector) FindPatchesByProject(projectId string, ts time.Time, limit int) ([]restModel.APIPatch, error) {
-	patches, err := patch.Find(patch.PatchesByProject(projectId, ts, limit))
+	apiPatches := []restModel.APIPatch{}
+	id, err := model.GetIdForProject(projectId)
+	if id == "" {
```
ablack12, line 44: this will also be true in the error case, so we should check the user first

rest/data/patch.go
```diff
@@ -340,9 +348,18 @@ func (hp *MockPatchConnector) FindPatchesByProject(projectId string, ts time.Tim
 	if limit <= 0 {
 		return patchesToReturn, nil
 	}
+	var id string
+	for _, p := range hp.CachedProjectRefs {
+		if *p.Id == projectId || *p.Identifier == projectId {
+			id = *p.Id
+		}
+	}
+	if id == "" {
```
ablack12, line 357: should error here

ablack12 (changes requested):

rest/data/patch.go
bynn: done

rest/data/patch.go
bynn: done

bynn:
This ticket changes the behavior of the api because it now returns an error when projects don't exist when it previously just returned 200 with StatusOk

rest/data/patch_test.go
```diff
@@ -144,8 +163,13 @@ func (s *PatchConnectorFetchByProjectSuite) TestFetchTooFew() {
...
 }
 
-func (s *PatchConnectorFetchByProjectSuite) TestFetchNonexistentFail() {
-	patches, err := s.ctx.FindPatchesByProject("zzz", s.time, 1)
+func (s *PatchConnectorFetchByProjectSuite) TestProjectNonexistentFail() {
+	_, err := s.ctx.FindPatchesByProject("zzz", s.time, 1)
+	s.Error(err)
+}
+
+func (s *PatchConnectorFetchByProjectSuite) TestPatchNonexistentFail() {
```
ablack12, line 171: "PatchNonexistentFail" seems wrong; we want it to succeed. Maybe "EmptyPatchesOkay" ?

ablack12 (approved):
LGTM with nit. (the failing batchtime test is just occasionally flakey so you can ignore it)

malikchaya2 (approved):

rest/data/patch_test.go
bynn: done

PR #4906, opened 2021-08-04
EVG-15005: add index hint for host termination query

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15005

### Description 
* Add hint to use status-prefixed index for host termination query.
* Revendor anser.
* Fix all queries that used `db.FindOne`/`db.FindAll` directly to instead use `db.Q`. Using `db.Q` is preferable, because it has better support for building queries with options when most of the options are not utilized. It seems like most of these queries probably should have used `db.FindOneQ`/`db.FindAllQ`, but opted instead to use the more verbose query functions `db.FindOne`/`db.FindAll`.

### Testing 
I verified in the mongo shell that running the host termination query with the hint utilizes the correct index. I also checked that staging/prod both have this index. Empirically based on the explain output, the query planner typically uses the index that I set for the query.


model/testutil/indexes.go
```diff
@@ -27,5 +29,10 @@ func AddTestIndexes(collection string, unique, sparse bool, key ...string) error
 		Keys:    spec,
 		Options: options.Index().SetUnique(unique).SetSparse(sparse),
 	}))
+
+	catcher.Add(db.EnsureIndex(host.Collection, mongo.IndexModel{
```
ybrill, line 33: is this necessary?

units/host_termination_test.go
```diff
@@ -84,8 +91,8 @@ func TestFlaggingDecommissionedHosts(t *testing.T) {
 		Convey("only hosts in the database who are marked decommissioned"+
 			" should be returned", func() {
 
-			// reset the db
 			require.NoError(t, db.ClearCollections(host.Collection), "error clearing hosts collection")
+			setupHostTerminationQueryIndex(t)
```
ybrill, line 95: why did you want to have the index in place for these tests?

Kimchelly: This is required because if you give a hint, the index must exit or the query will fail. We don't have a way of distinguishing a query run in prod from a query run in tests, so there's no way to omit the hint from the query.

ybrill: ah, got it.
Does it work with only the first key ("status" and not "spawn_options.spawned_by_task"), then?

Separately, would an alternative be to only set the hint when `ExecutionEnvironmentType == production`?

Kimchelly: No, index patterns have to be exact matches.

I would prefer to avoid adding prod-based conditional code execution as much as possible. Switching on prod/non-prod tends to be a hack-ish solution that becomes more kludgy the more you lean on it. If we ever have to introduce more query hints than just this one-off, it might be worth considering doing test setup for indexes somehow.

units/host_termination_test.go
```diff
@@ -27,6 +28,12 @@ func init() {
 	}
 }
 
+func setupHostTerminationQueryIndex(t *testing.T) {
+	require.NoError(t, db.EnsureIndex(host.Collection, mongo.IndexModel{
```
ybrill, line 32: > index patterns have to be exact matches.

Do we need to add the second key here?

Kimchelly: Yes, because that index is the only index available.

ybrill (approved):

PR #4908, opened 2021-08-05
EVG-15107, EVG-15165: Add command and admin settings to download and run the agent in a pod

jessfeng:
[EVG-15107](https://jira.mongodb.org/browse/EVG-15107)

[EVG-15165](https://jira.mongodb.org/browse/EVG-15165)

### Description 
- Implement curl and agent command with retry to download agent and run it in the pod
- Add admin settings for `PodInit` 
- NOTE: EVG-15165 is WIP because of the HTML page that I have yet to write + test

### Testing 
- Unit tested each OS for curl and agent command
- (Will do - test admin settings on local-evergreen and staging)


model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/pkg/errors"
+)
+
+const (
+	executable = "./evergreen"
+	shell      = "CMD-SHELL "
```
Kimchelly, line 14: This has to be `"CMD-SHELL"` I think (no extra space).

model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/pkg/errors"
+)
+
+const (
+	executable = "./evergreen"
```
Kimchelly, line 13: Rather than define a constant for this, I would just use
```go
fmt.Sprintf("./%s, p.BinaryPath())
```
on L45 since the entire agent command download/execution script is dependent on all the commands being relative to the working directory. Making assumptions about relative/absolute paths is generally risky as it's a common source of bugs (e.g. doing something in a relative path when the "working directory" isn't exactly what you expect it to be). Therefore, I would try to localize relative path dependencies like this to be as close as possible to other dependencies on relative paths `curl -LO` and `chmod ./evergreen` depend on relative paths, so the dependency on invoking the evergreen binary in a relative path should also be localized to that spot rather than be defined as a constant visible to the entire package.

model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+	cmds, err := p.curlCommands(settings, retryArgs)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	return strings.Join(cmds, " && "), nil
+}
+
+// CurlCommandWithDefaultRetry is the same as CurlCommandWithRetry using the
+// default retry parameters.
+func (p *Pod) CurlCommandWithDefaultRetry(settings *evergreen.Settings) (string, error) {
```
Kimchelly, line 38: There's no need to have two separate `CurlCommandWithRetry` and `CurlCommandWithDefaultRetry` methods. You can just make one `CurlCommand` method that uses the default retry values.

model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+	return filepath.Join(string(p.TaskContainerCreationOpts.Arch), p.BinaryName())
+}
+
+func (p *Pod) curlCommands(settings *evergreen.Settings, curlArgs string) ([]string, error) {
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting service flags")
+	}
+
+	curlCmd := shell
```
Kimchelly, line 97: The [example in the docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html) says that you have to pass the arguments in a string slice with the first element in the slice being "CMD" or "CMD-SHELL", like this:
```go
[ "CMD-SHELL", "curl -f http://localhost/ || exit 1" ]
```

The first element is the designation of the type of command execution you want (`"CMD-SHELL"`), and the second argument is the shell script itself.

Therefore, it seems like you want the `curlCmd` to be just the script that you build in L98-104 and `"CMD-SHELL"` should instead be the first element in the slice on L108.

model/pod/podutil_test.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const workingDir = "/data/mci"
```
Kimchelly, line 12: Does this need to be a global constant across all tests in this package? It seems like you could test with any random meaningless directory and set working dir to something like "/some/working/dir" or whatever within each test.

model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+	} else {
+		curlCmd += fmt.Sprintf("curl -LO '%s'%s", p.ClientURL(settings), curlArgs)
+	}
+
+	agentCmd := strings.Join(p.AgentCommand(settings), " ")
+
+	return []string{
+		curlCmd,
+		agentCmd,
+		fmt.Sprintf("chmod +x %s", p.BinaryName()),
```
Kimchelly, line 111: `chmod +x (Pod).TaskContainerCreationOpts.WorkingDir`.

model/pod/podutil.go
```diff
@@ -0,0 +1,117 @@
...
+		curlCmd += fmt.Sprintf("(curl -LO '%s'%s || curl -LO '%s'%s)", p.S3ClientURL(settings), curlArgs, p.ClientURL(settings), curlArgs)
+	} else {
+		curlCmd += fmt.Sprintf("curl -LO '%s'%s", p.ClientURL(settings), curlArgs)
+	}
+
+	agentCmd := strings.Join(p.AgentCommand(settings), " ")
+
+	return []string{
+		curlCmd,
+		agentCmd,
```
Kimchelly, line 110: You have to run `chmod` before you invoke the agent command because `chmod` is what makes the agent binary an actual executable.

Kimchelly (changes requested):
Can you include the Jira link to EVG-15165 as well?

model/pod/podutil.go
Kimchelly: Sorry, maybe I should have made this comment clearer after we talked in office hours - you have to chmod the executable within the working directory. I think it might be alright to just switch it back to what it originally was (`fmt.Sprintf("chmod +x %s", p.BinaryName())`), since we're using relative paths for everything in this shell script.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+
+// CurlCommand returns the command to curl the evergreen client and retries the request with the default retry parameters.
+func (p *Pod) CurlCommand(settings *evergreen.Settings) ([]string, error) {
+	var retryArgs string
+	if curlDefaultNumRetries != 0 && curlDefaultMaxSecs != 0 {
+		retryArgs = " " + curlRetryArgs(curlDefaultNumRetries, curlDefaultMaxSecs)
+	}
+	cmds, err := p.curlCommands(settings, retryArgs)
+	if err != nil {
+		return []string{}, errors.WithStack(err)
```
Kimchelly, line 26: `return nil, errors.WithStack(err)`

model/pod/podutil.go
```diff
@@ -9,40 +9,29 @@ import (
...
 )
 
-// CurlCommandWithRetry returns the command to curl the evergreen client and retries the request.
-func (p *Pod) CurlCommandWithRetry(settings *evergreen.Settings, numRetries, maxRetrySecs int) (string, error) {
+// CurlCommand returns the command to curl the evergreen client and retries the request with the default retry parameters.
+func (p *Pod) CurlCommand(settings *evergreen.Settings) ([]string, error) {
 	var retryArgs string
-	if numRetries != 0 && maxRetrySecs != 0 {
-		retryArgs = " " + curlRetryArgs(numRetries, maxRetrySecs)
+	if curlDefaultNumRetries != 0 && curlDefaultMaxSecs != 0 {
```
Kimchelly, line 21: I think we don't need this check since we're not passing in parameter to `CurlCommand` and the constants are already non-zero - you can just construct the `curlRetryArgs` without this if check:
```go
retryArgs := " " + curlRetryArgs(curlDefaultNumRetries, curlDefaultMaxSecs)
```

config_test.go
```diff
@@ -360,6 +360,19 @@ func (s *AdminSuite) TestJiraConfig() {
 	s.Equal(config, settings.Jira)
 }
 
+func (s *AdminSuite) TestPodinitConfig() {
+	config := PodInitConfig{
+		S3BaseURL: "s3_base_url",
+	}
+
+	err := config.Set()
+	s.NoError(err)
```
Kimchelly, line 369: This should probably be `s.Require().NoError()`, which is the test suite equivalent of `require.NoError`, since all the checks after it in the test will definitely fail if this line fails.

config_test.go
```diff
@@ -360,6 +360,19 @@ func (s *AdminSuite) TestJiraConfig() {
...
+	settings, err := GetConfig()
+	s.NoError(err)
```
Kimchelly, line 371: `s.Require().NoError()`

config_test.go
```diff
@@ -360,6 +360,19 @@ func (s *AdminSuite) TestJiraConfig() {
...
+	s.NotNil(settings)
```
Kimchelly, line 372: `s.Require().NotNil(settings)`, or else L373 could panic if it's nil.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+	return []string{
+		fmt.Sprintf("./%s", p.BinaryName()),
+		"agent",
+		fmt.Sprintf("--api_server=%s", settings.ApiUrl),
+		fmt.Sprintf("--mode=pod"),
+		fmt.Sprintf("--pod_id=%s", p.ID),
+		fmt.Sprintf("--pod_secret=%s", p.Secret),
+		fmt.Sprintf("--log_prefix=%s", filepath.Join(p.TaskContainerCreationOpts.WorkingDir, "agent")),
+		fmt.Sprintf("--working_directory=%s", p.TaskContainerCreationOpts.WorkingDir),
+		"--cleanup",
```
Kimchelly, line 42: I think we actually don't need this `--cleanup` flag. This flag tells the agent to clean up some leftover state in between executions of different tasks (e..g left over files in the file system). However, since pods don't have to be cleaned up between different tasks (since they're dedicated to the one task that runs it), we don't need this flag.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	cmds, err := p.curlCommands(settings, retryArgs)
+	if err != nil {
+		return []string{}, errors.WithStack(err)
+	}
+	return []string{"CMD-SHELL", strings.Join(cmds, " && ")}, nil
+}
+
+// AgentCommand returns the arguments to start the agent.
+func (p *Pod) AgentCommand(settings *evergreen.Settings) []string {
```
Kimchelly, line 32: I think this method doesn't need to be exported for now since no other package depends on it, so it can be `agentCommand`.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+	return strings.Join([]string{
+		strings.TrimSuffix(settings.ApiUrl, "/"),
+		strings.TrimSuffix(settings.ClientBinariesDir, "/"),
+		p.ExecutableSubPath(),
+	}, "/")
+}
+
+// S3ClientURL returns the URL in S3 where the Evergreen client version can be
+// retrieved for this server's particular Evergreen build version.
+func (p *Pod) S3ClientURL(settings *evergreen.Settings) string {
```
Kimchelly, line 58: `s3ClientURL` since no other package needs to use this function right now.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+		fmt.Sprintf("--pod_secret=%s", p.Secret),
+		fmt.Sprintf("--log_prefix=%s", filepath.Join(p.TaskContainerCreationOpts.WorkingDir, "agent")),
+		fmt.Sprintf("--working_directory=%s", p.TaskContainerCreationOpts.WorkingDir),
+		"--cleanup",
+	}
+}
+
+// ClientURL returns the URL used to get the latest Evergreen client version
+// directly from the Evergreen server.
+func (p *Pod) ClientURL(settings *evergreen.Settings) string {
```
Kimchelly, line 48: `clientURL` since no other package needs to use this function right now.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+// BinaryName returns the file name of the binary.
+func (p *Pod) BinaryName() string {
+	name := "evergreen"
+	if p.TaskContainerCreationOpts.OS == OSWindows {
+		return name + ".exe"
+	}
+	return name
+}
+
+// ExecutableSubPath returns the directory containing the compiled agents.
```
Kimchelly, line 75: clientURLSubpath returns the URL path to the compiled agent.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+func (p *Pod) ExecutableSubPath() string {
+	return filepath.Join(string(p.TaskContainerCreationOpts.Arch), p.BinaryName())
```
Kimchelly, line 77: I think this actually has to join both OS and arch with an underscore because binaries are stored in this path in S3 and in the app server:

The general format of the URL is this:
```go
clients/${GOOS}_${GOARCH}/evergreen (or evergreen.exe for Windows)
```
So I would include the `p.TaskContainerCreationOpts.OS` in the path.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+func (p *Pod) BinaryName() string {
+	name := "evergreen"
+	if p.TaskContainerCreationOpts.OS == OSWindows {
+		return name + ".exe"
+	}
+	return name
+}
+
+// ExecutableSubPath returns the directory containing the compiled agents.
+func (p *Pod) ExecutableSubPath() string {
```
Kimchelly, line 76: `clientURLSubpath` - the significance of this field is slightly different than on EC2 hosts, but it's referring to a portion of the URL path that tells it where the agent is located.

For example, if you're downloading the agent for `GOOS="linux"` and `GOARCH="amd64"`, the path to the compiled binary is:
```go
linux_amd64/evergreen
```
And for `GOOS="windows"` `GOARCH="amd64"`, the path would be:
```go
windows_amd64/evergreen.exe
```

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// Constants representing default curl retry arguments.
+const (
+	curlDefaultNumRetries = 10
+	curlDefaultMaxSecs    = 100
+)
+
+// CurlCommand returns the command to curl the evergreen client and retries the request with the default retry parameters.
+func (p *Pod) CurlCommand(settings *evergreen.Settings) ([]string, error) {
```
Kimchelly, line 19: I would rename this to be `AgentScript` because it's doing multiple things to provision the agent (curl, change permissions, run the agent), not just curling it.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/pkg/errors"
+)
+
+// Constants representing default curl retry arguments.
+const (
+	curlDefaultNumRetries = 10
+	curlDefaultMaxSecs    = 100
+)
+
+// CurlCommand returns the command to curl the evergreen client and retries the request with the default retry parameters.
```
Kimchelly, line 18: AgentScript returns the script to provision and run the agent in the pod's container.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+	var curlCmd string
+	if !flags.S3BinaryDownloadsDisabled && settings.PodInit.S3BaseURL != "" {
+		// Attempt to download the agent from S3, but fall back to downloading from
+		// the app server if it fails.
+		curlCmd = fmt.Sprintf("(curl -LO '%s'%s || curl -LO '%s'%s)", p.S3ClientURL(settings), curlArgs, p.ClientURL(settings), curlArgs)
+	} else {
+		curlCmd = fmt.Sprintf("curl -LO '%s'%s", p.ClientURL(settings), curlArgs)
+	}
+
+	agentCmd := strings.Join(p.AgentCommand(settings), " ")
```
Kimchelly, line 94: This logic should be in `AgentScript` instead of `curlCommands` because it's the thing that actually runs the agent and has nothing to do with downloading the agent.

model/pod/podutil.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return []string{
+		curlCmd,
+		fmt.Sprintf("chmod +x %s", p.TaskContainerCreationOpts.WorkingDir),
```
Kimchelly, line 98: I would move this into `AgnetScript` for the same reason as above.

Kimchelly (changes requested):

model/pod/podutil.go
jessfeng: Sounds good, but some of the Archs already have the OS appended in front (ie. `evergreen.ArchWindowsAmd64` =  `windows_amd64`), so should I just add it for the Archs that don't have the OS specified?

Kimchelly: Yeah, the `evergreen.Arch` types are for distros specifically and the only allowed pod architectures are the ones in the DB model package for pods.

Kimchelly (approved):

PR #4911, opened 2021-08-05
EVG-15110: increased checks and logging to prevent collision of host drawdown jobs

hhoke:
[EVG-15110](https://jira.mongodb.org/browse/EVG-15110)

### Description 
1) We appeared to be having a bunch of collisions due to the fact that drawdown jobs are named only after the start time, resulting in drastically reduced effectiveness system-wide
2) Drawdown jobs were being run for static distros, which should never be the case. I had assumed the host allocator only ran for spawnable distros, but there is now a check built in to make sure we only create host drawdown jobs for spawnable distros.
3) A huge portion of drawdown jobs end up killing 0 hosts. I added logging to help figure out why.

### Testing 
  none


ybrill:
We definitely run the host allocator for static distros too. I think the reason we don't try to spawn them is [the return here](https://github.com/evergreen-ci/evergreen/blob/db3f306dac84b094a7a55942a0c6f03995a916a3/scheduler/utilization_based_host_allocator.go#L151-L153).

units/host_allocator.go
```diff
@@ -305,7 +306,7 @@ func (j *hostAllocatorJob) setTargetAndTerminate(ctx context.Context, numUpHosts
 			DistroID:     distro.Id,
 			NewCapTarget: newCapTarget,
 		}
-		err := amboy.EnqueueUniqueJob(ctx, j.env.RemoteQueue(), NewHostDrawdownJob(j.env, drawdownInfo, utility.RoundPartOfMinute(1).Format(TSFormat)))
+		err := amboy.EnqueueUniqueJob(ctx, j.env.RemoteQueue(), NewHostDrawdownJob(j.env, drawdownInfo, utility.RoundPartOfMinute(1).Format(TSFormat)+distro.Id))
```
ybrill, line 309: an alternative to doing it this way is to encorporate the distro id in the id and scope [here](https://github.com/evergreen-ci/evergreen/blob/db3f306dac84b094a7a55942a0c6f03995a916a3/units/host_drawdown.go#L67-L68)

ybrill (commented):
perhaps we should also log when `checkTerminationExemptions` returns true?

hhoke:
> perhaps we should also log when `checkTerminationExemptions` returns true?

I can't think of a reason to do this -- we already log or have a descriptive error for most of the cases where checkTerminationExemptions returns true, and j.Terminated keeps a reliable record of how many hosts have ben terminated.

units/host_allocator.go
ybrill: This definitely _works_, but I think it's more typical to assemble the id in the job's "constructor". Doing it there has the advantage that the id is consistent no matter where the job is created. 
A couple of examples [here](https://github.com/evergreen-ci/evergreen/blob/ea2993ee93fc6fc77fd6dbc6ef33063f74d6ec4f/units/host_allocator.go#L62), [here](https://github.com/evergreen-ci/evergreen/blob/ea2993ee93fc6fc77fd6dbc6ef33063f74d6ec4f/units/scheduler.go#L50), [here](https://github.com/evergreen-ci/evergreen/blob/ea2993ee93fc6fc77fd6dbc6ef33063f74d6ec4f/units/provisioning_user_data_done.go#L64), etc.

hhoke: I changed it to fit this format

ybrill (approved):

PR #4912, opened 2021-08-06
EVG-14973  Hide child patches from showing up as a patch

malikchaya2:
[EVG-14973](https://jira.mongodb.org/browse/EVG-14973)

### Description 
Add a parent_patch user and use that for child patches so that they don't show up on the myPatches page. 

### Testing 
Tested by adding the parent_patch user to staging, creating a patch with child patches, and making sure that they show up in the right places.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -12,6 +12,7 @@ import (
 const (
 	User            = "mci"
 	GithubPatchUser = "github_pull_request"
+	ParentPatchUser = "parent_patch"
 
 	HostRunning         = "running"
 	HostTerminated      = "terminated"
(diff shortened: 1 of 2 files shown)


arjunpatel100:
Viewing the patch details page for a child patch may have a breadcrumb pointing the the parent_patch user patches page. I wonder if we should show a breadcrumb link the parent patch instead which would have a proper user.

ablack12:
> Viewing the patch details page for a child patch may have a breadcrumb pointing the the parent_patch user patches page. I wonder if we should show a breadcrumb link to the parent patch instead.

I'm okay with this; the user shouldn't really be able to access this page anyway.

ablack12 (approved):

malikchaya2:
> Viewing the patch details page for a child patch may have a breadcrumb pointing the the parent_patch user patches page. I wonder if we should show a breadcrumb link to the parent patch instead.

I agree with Annie. To see this page someone would specifically have to know about it and seek it out.

arjunpatel100 (approved):

sophstad (approved):

PR #4913, opened 2021-08-06
EVG-14112: Introduce post_error_fails_task

hadjri:
[EVG-14112](https://jira.mongodb.org/browse/EVG-14112)

### Description 
Currently failures in post task commands are ignored by default.  A change has been introduced to allow users to enforce in config that failures during post tasks cause the task to fail (similar to [pre_error_fails_task](https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#pre-post-and-timeout))

### Testing 
Added unit tests to ensure that these new configurations let post tasks to return an error if they are set.


model/project.go
```diff
@@ -491,6 +492,7 @@ type TaskGroup struct {
 	// data about the task group
 	MaxHosts              int             `yaml:"max_hosts" bson:"max_hosts"`
 	SetupGroupFailTask    bool            `yaml:"setup_group_can_fail_task" bson:"setup_group_can_fail_task"`
+	TeardownGroupFailTask bool            `yaml:"teardown_group_can_fail_task" bson:"teardown_group_can_fail_task"`
```
ablack12, line 495: TeardownGroup is actually something else and won't work for this, so this is misleading. It's actually TeardownTask (i.e. post tasks) that we're allowing to fail (not the best naming scheme)

ablack12 (changes requested):
You should be able to test this in staging by adding a post task with a shell.exec command "exit 1" (which will always fail), and then verify that post_error_fails_task does cause the task to fail (and that it doesn't fail without that)

agent/agent.go
```diff
@@ -618,29 +623,37 @@ func (a *Agent) endTaskResponse(tc *taskContext, status string, message string)
...
 		tc.logger.Execution().Error(errors.Wrap(err, "error fetching task group for post-task commands"))
-		return
+		return nil
 	}
 	if taskGroup.TeardownTask != nil {
-		err = a.runCommands(postCtx, tc, taskGroup.TeardownTask.List(), runCommandsOptions{})
-		tc.logger.Task().Error(message.WrapError(err, message.Fields{
-			"message": "Error running post-task command.",
-		}))
+		opts.shouldSetupFail = taskGroup.TeardownGroupFailTask
```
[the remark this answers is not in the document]
hadjri: shouldSetupFail is part of the runCommandsOptions struct.  Currently it is set as flagged to true if pre_error_fails_task is true because this will set SetupGroupFailTask and this property is referring to that. We can introduce a new property in the struct called something like shouldTeardownFail, or we could rename this property to something more generic that can apply to pre and post task commands, like shouldTriggerFailure.

hadjri:
> You should be able to test this in staging by adding a post task with a shell.exec command "exit 1" (which will always fail), and then verify that post_error_fails_task does cause the task to fail (and that it doesn't fail without that)

Tested in staging and observed that once a post task fails, the entire task will fail if post_error_fails_task is set: https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_test_model_event_patch_db3f306dac84b094a7a55942a0c6f03995a916a3_6111591697b1d349228bf2dc_21_08_09_16_34_46/logs?execution=0&sortBy=STATUS&sortDir=ASC

Compared to currently in prod all post tasks will execute and do not have an impact on whether the task fails or not: https://evergreen.mongodb.com/task/evergreen_race_detector_test_model_event_8b69f0ddb6666271114214e0a9b57620a2e27e61_21_08_09_16_18_52

agent/agent.go
```diff
@@ -618,29 +623,37 @@ func (a *Agent) endTaskResponse(tc *taskContext, status string, message string)
...
+		err = a.runCommands(postCtx, tc, taskGroup.TeardownTask.List(), opts)
+		if err != nil {
+			tc.logger.Task().Error(message.WrapError(err, message.Fields{
```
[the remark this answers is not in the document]
hadjri: Can you specify where the double logging occurs? It looks like runCommands only returns an error and doesn't log it out, and this log message has been in the codebase for some time.

config.go
```diff
@@ -88,7 +88,6 @@ type Settings struct {
 	Notify              NotifyConfig              `yaml:"notify" bson:"notify" json:"notify" id:"notify"`
 	Plugins             PluginConfig              `yaml:"plugins" bson:"plugins" json:"plugins"`
 	PluginsNew          util.KeyValuePairSlice    `yaml:"plugins_new" bson:"plugins_new" json:"plugins_new"`
-	PodInit             PodInitConfig             `yaml:"pod_init" bson:"pod_init" json:"pod_init" id:"pod_init"`
```
ablack12, line 91: Is this from the merge?

agent/agent.go
```diff
@@ -618,29 +625,37 @@ func (a *Agent) endTaskResponse(tc *taskContext, status string, message string)
...
 		tc.logger.Execution().Error(errors.Wrap(err, "error fetching task group for post-task commands"))
-		return
+		return nil
 	}
 	if taskGroup.TeardownTask != nil {
-		err = a.runCommands(postCtx, tc, taskGroup.TeardownTask.List(), runCommandsOptions{})
-		tc.logger.Task().Error(message.WrapError(err, message.Fields{
-			"message": "Error running post-task command.",
-		}))
+		opts.shouldSetupFail = taskGroup.TeardownTaskCanFailTask
```
ablack12, line 643: (moving conversation from a previous thread here, since it's attached to outdated code)
I think we should rename this field then. It isn't going to be clear to us in the future why teardown is affecting setup (or so it reads here).

agent/agent.go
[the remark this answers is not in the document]
ablack12: Oh I think I said the wrong thing but there is still double logging, because we're logging here, and we're logging in FinishTask if runPostTaskCommands returns an error (I think we can just remove that log)

agent/agent.go
```diff
@@ -550,12 +550,19 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 	detail := a.endTaskResponse(tc, status, message)
 	switch detail.Status {
 	case evergreen.TaskSucceeded:
-		tc.logger.Task().Info("Task completed - SUCCESS.")
-		a.runPostTaskCommands(ctx, tc)
+		if err := a.runPostTaskCommands(ctx, tc); err != nil {
```
ablack12, line 553: I think the issue with this is that most users won't be using post_error_fails_task, but now they suddenly have to wait for post task to know if the task has been successful or not. Maybe the more consistent user experience would be to log this before post task still, and then log "Post task completed -- FAILURE" if we're allowing that (this may be mildly confusing for that use case but it'll be more rare anyway)

ablack12 (changes requested):
Were you able to test this in staging? I would like to verify that it works end to end before merging.

config.go
hadjri: Yes, looks like something didn't go totally right when I was resolving merge conflicts in this file.  Fixing.

hadjri:
> Were you able to test this in staging? I would like to verify that it works end to end before merging.

Tested in staging with a post command (shell.exec) just running exit 1 in script.  The task is marked as failed even though all tests passed due to this post task failing alongside post_err_fails_task being enabled.  Will also update documentation when change is merged. https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_test_model_event_patch_c6ffa29e590ee1ce9137bbd6827cc2767feff144_611289ad9dbe3250b56347ec_21_08_10_14_14_50/logs?execution=0

model/project.go
```diff
@@ -489,16 +490,17 @@ type TaskGroup struct {
...
-	SetupGroup            *YAMLCommandSet `yaml:"setup_group" bson:"setup_group"`
-	TeardownGroup         *YAMLCommandSet `yaml:"teardown_group" bson:"teardown_group"`
-	SetupTask             *YAMLCommandSet `yaml:"setup_task" bson:"setup_task"`
-	TeardownTask          *YAMLCommandSet `yaml:"teardown_task" bson:"teardown_task"`
-	Timeout               *YAMLCommandSet `yaml:"timeout,omitempty" bson:"timeout"`
-	Tasks                 []string        `yaml:"tasks" bson:"tasks"`
-	Tags                  []string        `yaml:"tags,omitempty" bson:"tags"`
+	MaxHosts                int             `yaml:"max_hosts" bson:"max_hosts"`
+	SetupGroupFailTask      bool            `yaml:"setup_group_can_fail_task" bson:"setup_group_can_fail_task"`
+	TeardownTaskCanFailTask bool            `yaml:"teardown_task_can_fail_task" bson:"teardown_task_can_fail_task"`
```
ablack12, line 495: nit: can you keep the setupgroup variables together and move this before or after (or next to teardown)

ablack12 (changes requested):
Sorry for the long wait on the review! Feel free to ping me if it's been more than 24 hours.

This is just about there, we'll just need to add post error to this merge function https://github.com/evergreen-ci/evergreen/blob/cd549ca996e22bf1b812023ed403c19bc84be478/model/project_parser.go#L1207 (there's a ticket planned to create a linter so that we catch when things haven't been added to this merge function in the future)

ablack12 (approved):

hadjri:
Updated documentation: https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#pre-post-and-timeout

PR #4914, opened 2021-08-06
EVG-14332 add notification for failed or blocked task

ablack12:
[EVG-14332](https://jira.mongodb.org/browse/EVG-14332)

### Description 
Notify if the task is failed or blocked. (Could've made this just notify on blocked; I thought a bit about the reporter's use case / general use case for this and I think failed + blocked does make the most sense, but the work to add "just blocked" would be similar to this if it's ever requested in the future.)

### Testing 
 Tested in staging on both legacy and spruce for blocked and failed (slack and email).

Spruce
image

Legacy
image

Slack
image

Email
image

### DIFF
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -128,7 +128,6 @@ func LogTaskRestarted(taskId string, execution int, userId string) {
 	logTaskEvent(taskId, TaskRestarted, TaskEventData{Execution: execution, UserId: userId})
 }
 
-// TODO: should we also log what task blocked it? (Would only want to log it then if it's blocked for the first time I think)
 func LogTaskBlocked(taskId string, execution int) {
 	logTaskEvent(taskId, TaskBlocked, TaskEventData{Execution: execution})
 }
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -10,6 +10,7 @@ import (
 func init() {
 	registry.AddType(ResourceTypeTask, taskEventDataFactory)
 	registry.AllowSubscription(ResourceTypeTask, TaskFinished)
+	registry.AllowSubscription(ResourceTypeTask, TaskBlocked)
 }
 
 const (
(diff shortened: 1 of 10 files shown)


model/event/task_event.go
```diff
@@ -126,6 +128,11 @@ func LogTaskRestarted(taskId string, execution int, userId string) {
 	logTaskEvent(taskId, TaskRestarted, TaskEventData{Execution: execution, UserId: userId})
 }
 
+// TODO: should we also log what task blocked it? (Would only want to log it then if it's blocked for the first time I think)
```
bynn, line 131: I think it would be nice to log it!

bynn (approved):

model/event/task_event.go
ablack12: Sorry this is actually an old todo--I decided not to do this, because we don't display this on the event log page. I think this would clutter that page since blocked information is already easily available on the main page, and would be confusing if we didn't log every time and then also log when something was unblocked.

malikchaya2 (commented):
editing to remove comment

model/task/task.go
```diff
@@ -1767,15 +1767,26 @@ func (t *Task) MarkUnscheduled() error {
...
 	}
 
-	return UpdateAllMatchingDependenciesForTask(t.Id, dependency.Id, unattainable)
+	if err := updateAllMatchingDependenciesForTask(t.Id, dependencyId, unattainable); err != nil {
+		return err
+	}
+
+	// only want to log the task as blocked if it wasn't already blocked
+	if !wasBlocked && unattainable {
+		event.LogTaskBlocked(t.Id, t.Execution)
```
malikchaya2, line 1787: Does this mean that the task will never run? Can it at some point become attainable again once the dependency gets unblocked?

ablack12: This just means that it's currently blocked. It can become attainable again if the dependency is unblocked, however we aren't displaying this log message anywhere, we're just using it to notify the user that the task _is_ blocked so they can intervene if they want (I don't think adding notifications for a task becoming unblocked is really the ask of this ticket).

malikchaya2: Ah that makes sense. I was thinking that we may not want to notify if something is blocked temporarily (and will be unblocked a minute later), but if the user may be able to intervene, maybe they actually want that.

malikchaya2 (approved):

PR #4916, opened 2021-08-09
EVG-15197  Don't error if child version doesn't exist when modifying

malikchaya2:
[EVG-15197](https://jira.mongodb.org/browse/EVG-15197)

### Description 
This checks if a child patch is finalized before making modifications to avoid errors. 

### Testing 
Tested by creating a patch with a not finalized child patch in staging and changing the priority with no errors.

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -775,7 +775,7 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 						return ResourceNotFound.Send(ctx, fmt.Sprintf("child patch '%s' not found ", childPatchId))
 					}
 					// only modify the child patch if it is finalized
-					if p.Activated {
+					if p.Version != "" {
 						err = ModifyVersionHandler(ctx, dataConnector, childPatchId, modifications)
 						if err != nil {
 							return errors.Wrap(mapHTTPStatusToGqlError(ctx, httpStatus, err), fmt.Sprintf("error modifying child patch '%s'", patchID))
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -767,10 +767,21 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 			}
 			if p.IsParent() {
 				for _, childPatchId := range p.Triggers.ChildPatches {
-					err = ModifyVersionHandler(ctx, dataConnector, childPatchId, modifications)
+					p, err := patch.FindOneId(childPatchId)
 					if err != nil {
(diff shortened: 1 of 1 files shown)


ablack12 (commented):
How could the child patch not be finalized but the parent patch is finalized?

graphql/util.go
```diff
@@ -767,10 +767,21 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
...
+					p, err := patch.FindOneId(childPatchId)
 					if err != nil {
-						return errors.Wrap(mapHTTPStatusToGqlError(ctx, httpStatus, err), fmt.Sprintf("error modifying child patch '%s'", patchID))
+						return ResourceNotFound.Send(ctx, fmt.Sprintf("error finding child patch %s: %s", childPatchId, err.Error()))
 					}
+					if p == nil {
+						return ResourceNotFound.Send(ctx, fmt.Sprintf("child patch '%s' not found ", childPatchId))
+					}
+					// only modify the child patch if it is finalized
+					if p.Activated {
```
ablack12, line 778: more specifically what you're checking here is "does a version exist" which you can check by looking at p.Version instead

malikchaya2:
> How could the child patch not be finalized but the parent patch is finalized?

Users can set it up for the child to wait on the parent. It can be helpful in cases where they may want to pass artifacts or values from the parent to the child.

graphql/util.go
malikchaya2: Is that more fool-proof that checking the Activated field?

ablack12: Theoretically it's the same but just more general; the version definitely exists if the version field is set, whereas I don't remember if Activated can be changed. They probably both work though.

graphql/util.go
```diff
@@ -767,10 +767,21 @@ func ModifyVersionHandler(ctx context.Context, dataConnector data.Connector, pat
 			}
 			if p.IsParent() {
 				for _, childPatchId := range p.Triggers.ChildPatches {
-					err = ModifyVersionHandler(ctx, dataConnector, childPatchId, modifications)
+					p, err := patch.FindOneId(childPatchId)
```
ablack12, line 770: can we get all of the patches with one query and then iterate through the patches instead?

ablack12 (approved):
LGTM, left a performance improvement suggestion but not blocking

graphql/util.go
malikchaya2: That's a good point. There is another place where we do this so I will create another ticket to write a getchildPatches query and use it where appropriate.

graphql/util.go
malikchaya2: checking p.version seems more accurate. I'll change it.

PR #4917, opened 2021-08-09
EVG-15002 Silence unknown github event error

bynn:
[EVG-15002](https://jira.mongodb.org/browse/EVG-15002)

### Description 
do not grip anything for unknown event type `pull_request_reveiw_thread`

### Testing 
unit test

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -88,7 +88,13 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 		}
 	}
 
-	gh.event, _ = github.ParseWebHook(gh.eventType, body)
+	gh.event, err = github.ParseWebHook(gh.eventType, body)
+	if err != nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Message:    err.Error(),
+		}
+	}
 
 	return nil
 }
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -74,7 +74,7 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 		}
 	}
 
-	if gh.eventType != "pull_request_review_thread" {
+	if gh.eventType != "pull_request_review_thread" && gh.eventType != "workflow_job" {
 		body, err := github.ValidatePayload(r, gh.secret)
 		if err != nil {
 			grip.Error(message.WrapError(err, message.Fields{
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -74,33 +74,33 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rest/route/github.go
```diff
@@ -90,12 +90,14 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 
 	gh.event, err = github.ParseWebHook(gh.eventType, body)
 	if err != nil {
-		grip.Error(message.WrapError(err, message.Fields{
-			"source":  "github hook",
-			"msg_id":  gh.msgID,
-			"event":   gh.eventType,
-			"message": "rejecting github webhook",
-		}))
+		if gh.eventType != "pull_request_review_thread" {
```
ybrill, line 93: since we know from the beginning that we're not going to parse this event maybe we can return early on line 76? 
I could also imagine that this wouldn't be an error: we could return a 200 to tell GitHub the message was processed successfully (we've pulled everything we want from it, which in this case is nothing)

bynn: i thought it would be more clear to wrap the validate and parse in an if statement rather than return early.

bynn: I think you're right and from the looks of it, the gh.event being nil will ultimately give us the error we want in Run.

rest/route/github.go
```diff
@@ -88,19 +88,7 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
...
-			"msg_id":  gh.msgID,
-			"event":   gh.eventType,
-			"message": "rejecting github webhook",
-		}))
-		return gimlet.ErrorResponse{
-			StatusCode: http.StatusBadRequest,
-			Message:    err.Error(),
-		}
-	}
+	gh.event, _ = github.ParseWebHook(gh.eventType, body)
```
ablack12, line 91: If we don't return an error here then we'll continue to Run() right? Which would be a problem. I would think we should still error here but not log it.

ybrill: I mean, I don't think it's *technically* a problem: Run() will skip to the end of the switch and return an empty response.
Still, I agree stylistically: it's probably best to avoid relying on this for the sake of clarity and in case Run() changes.

Perhaps the following can be a path forward: we could store the event type and body in the `githubHookApi`. In the run function we can check the eventType against a list of the events of interest (`utility.StringSliceContains(eventsOfInterestSlice, gh.eventType)`) and return 200 if we're not interested. Then we call ParseWebHook and take it from there.

ablack12: I do think this is a real error though, and returning 200 for a response we can't parse is incorrect. I would prefer to continue treating this as a real error but just not logging it back to Splunk. I don't think this issue is worth the engineering bandwidth to do anything too complicated.

bynn: parse returns error again

ybrill (approved):

ablack12 (approved):

PR #4918, opened 2021-08-09
EVG-15032 add display task ID to tasks

ablack12:
[EVG-15032](https://jira.mongodb.org/browse/EVG-15032)

### Description 
Adds displayTaskId as a pointer so we know that if it's nil we can't rely on it not being an execution task and we need to go check. I'm not a huge fan of this since for many tasks it will be populated but empty, but the alternative is to _also_ introduce a pointer to a boolean which I'm not sure is really better.

nil -> unset
empty string -> not an execution task
populated -> execution task 

### Testing 
Fixed unit tests. Going to add a few new ones.


model/task/task.go
```diff
@@ -177,6 +177,9 @@ type Task struct {
 	ResetWhenFinished  bool     `bson:"reset_when_finished,omitempty" json:"reset_when_finished,omitempty"`
 	DisplayTask        *Task    `bson:"-" json:"-"` // this is a local pointer from an exec to display task
 
+	// DisplayTaskId is populated for execution tasks created after this change and is explicitly empty for non execution tasks.
+	DisplayTaskId *string `bson:"display_task_id,omitempty" json:"display_task_id,omitempty"`
```
malikchaya2, line 181: Can you please explain more about how  for many `tasks it will be populated but empty`? Why is that the case and can we fix that?

malikchaya2: Is that taken care of by simply check if = ""?

model/task/task.go
```diff
@@ -2393,31 +2390,66 @@ func (t *Task) IsPartOfSingleHostTaskGroup() bool {
...
 func (t *Task) IsPartOfDisplay() bool {
-	dt, err := t.GetDisplayTask()
-	if err != nil {
-		grip.Error(message.WrapError(err, message.Fields{
-			"message":        "unable to get display task",
-			"execution_task": t.Id,
-		}))
-		return false
+	// if display task ID is nil, we need to check manually if we have an execution task
+	if t.DisplayTaskId == nil {
```
malikchaya2, line 2394: what if it's not nil, but it's ""? Do we not need to check in that case?

model/task/task.go
```diff
@@ -2393,31 +2390,66 @@ func (t *Task) IsPartOfSingleHostTaskGroup() bool {
...
+			}
+		}
 	}
 	if err != nil {
 		return nil, err
 	}
+
+	if t.DisplayTaskId == nil {
+		// Cache display task ID for future use. If we couldn't find the display task,
+		// we cache the empty string to show that it doesn't exist.
```
malikchaya2, line 2445: So does and empty string reliably mean that it doesn't exist and is therefor not an execution task?  If so, what did you mean by `for many tasks it will be populated but empty`? Are there cases where it's empty but doesn't reliably mean that it's not an execution task?

model/lifecycle.go
```diff
@@ -1262,6 +1268,7 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 		TriggerEvent:        v.TriggerEvent,
 		CommitQueueMerge:    buildVarTask.CommitQueueMerge,
 		IsGithubCheck:       isGithubCheck,
+		DisplayTaskId:       utility.ToStringPtr(""), // this will be overridden if the task is an execution task
```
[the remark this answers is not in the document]
ablack12: The pointer only exists to distinguish between "explicitly set, and doesn't exist" and "this task hasn't checked if display task ID exists yet". (i.e. tasks made before this change and tasks made after)

model/task/task.go
ablack12: If it's explicitly set as a pointer to the empty string, then we know that there's no display task for this task, so we don't have to look it up from the DB (this is handled at line 2406)

model/task/task.go
ablack12: I think this is explained in the above comment and the PR comment (i.e. that a pointer to the empty string means that no display task exists, whereas a nil pointer means that we haven't checked to populate this field yet). I can add to this comment to help it be clear.

model/task/task.go
ablack12: Yes, that is the behavior -- by "It will be populated but empty" I meant that it will be set as a pointer to an empty string. The case where it's nil is the case that isn't reliable.

ablack12 (commented):
Feel free to DM me if this doesn't explain it!

malikchaya2 (approved):

PR #4920, opened 2021-08-09
EVG-15105: support container-level secrets

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15105

### Description 
* Support container-level secrets. I intentionally didn't include the container-level status, because it would be more complex to maintain it and keep it in sync. Since we can assume that the only container running in a pod is the agent container (for now), it doesn't really matter if we track the container-level status. We can add it in later if it becomes relevant to the pod lifecycle.
* Revendor Cocoa to include container-specific settings.

### Testing 
Updated the Amboy/DB/REST tests.


rest/model/pod.go
```diff
@@ -125,6 +125,9 @@ func (p *APIPod) ToService() (*pod.Pod, error) {
 	}
 	timing := p.TimeInfo.ToService()
 	resources := p.Resources.ToService()
+	if err != nil {
```
julianedwards, line 128: Where is this error coming from?

julianedwards (changes requested):

cloud/pod_util.go
```diff
@@ -34,9 +35,9 @@ func MakeECSPodCreator(c cocoa.ECSClient, v cocoa.Vault) (cocoa.ECSPodCreator, e
 // ExportPod exports the pod DB model to its equivalent cocoa.ECSPod backed by
 // the given ECS client and secret vault.
 func ExportPod(p *pod.Pod, c cocoa.ECSClient, v cocoa.Vault) (cocoa.ECSPod, error) {
-	status, err := ExportPodStatus(p.Status)
+	stat, err := ExportECSPodStatusInfo(p)
```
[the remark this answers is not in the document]
Kimchelly: Makes sense, done.

rest/model/pod.go
Kimchelly: Sorry that's from an old piece of code. I removed it.

julianedwards (approved):

PR #4922, opened 2021-08-09
EVG-15192 Merge getTasksByVersion queries into one

khelif96:
[EVG-15192](https://jira.mongodb.org/browse/EVG-15192)

### Description 
Instead of performing two queries to fetch tasks and to fetch the matching task count. Perform only one query by merging the separate pipelines using `$facet`

### Testing 
Reused existing tests since this is a refactor instead of a feature change.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2933,9 +2933,9 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 				sortFields = append(sortFields, bson.E{Key: singleSort.Key, Value: singleSort.Order})
 			}
 		}
-	} else {
-		sortFields = append(sortFields, bson.E{Key: IdKey, Value: 1})
 	}
+	sortFields = append(sortFields, bson.E{Key: IdKey, Value: 1})
+
 	sortAndPaginatePipeline = append(sortAndPaginatePipeline, bson.M{
 		"$sort": sortFields,
 	})
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2958,7 +2958,7 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 		})
 	}
 
-	// Use a $facet to perform seperate aggregations for $count and to sort and paginate the results in the same query
+	// Use a $facet to perform separate aggregations for $count and to sort and paginate the results in the same query
 	tasksAndCountPipeline := bson.M{
 		"$facet": bson.M{
 			"count": []bson.M{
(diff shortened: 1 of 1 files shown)


model/task/task.go
```diff
@@ -2920,31 +2920,31 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 			},
 		})
 	}
-	countPipeline := []bson.M{}
-	countPipeline = append(countPipeline, pipeline...)
-	countPipeline = append(countPipeline, bson.M{"$count": "count"})
+
+	sortAndPaginatePipeline := []bson.M{}
```
ablack12, line 2924: I think this function in general needs many more comments so that people besides you (i.e. me, gulp) can understand this in the future. To that end, can you drop a comment here explaining why you're starting a new pipeline?

ablack12 (approved):
So obsessed with this! Do we have a future way to see if this speeds things up by any chance?

model/task/task.go
```diff
@@ -2920,31 +2920,31 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
...
 			if singleSort.Key == DisplayStatusKey || singleSort.Key == BaseTaskStatusKey {
-				pipeline = append(pipeline, addStatusColorSort((singleSort.Key)))
+				sortAndPaginatePipeline = append(sortAndPaginatePipeline, addStatusColorSort((singleSort.Key)))
 				sortFields = append(sortFields, bson.E{Key: "__" + singleSort.Key, Value: singleSort.Order})
 			} else {
 				sortFields = append(sortFields, bson.E{Key: singleSort.Key, Value: singleSort.Order})
 			}
 		}
+	} else {
+		sortFields = append(sortFields, bson.E{Key: IdKey, Value: 1})
```
ablack12, line 2937: Why do we now only do this in the else case?

khelif96: it seemed wrong, It would always apply a sort on the task id. Even though there were other sorts applied.  What do you think?

ablack12: I think it does seem reasonable to sort alphabetically after other sorts have been applied (like if we sort by status, sort task ID alphabetically within that, right?)

arjunpatel100: the IdKey helps stabilize the sort if the sort condition is a tie. including this conditionally could cause our gql tests to sometimes fail https://docs.mongodb.com/manual/reference/operator/aggregation/sort/#sort-consistency

arjunpatel100 (approved):

sophstad (approved):
very cool

PR #4923, opened 2021-08-10
EVG-15059 get project settings

malikchaya2:
[EVG-15059](https://jira.mongodb.org/browse/EVG-15059)

### Description 
@malikchaya2 this is the draft. Looks bigger than it is because I renamed ProjectSettingsEvent so we could use it. The thing here we may want to keep is FindPatchAliasesInProject (and I found two places where we aren't getting aliases right whoops).

### Testing 


ablack12:
I linked the relevant branch for this to the ticket, so I think we can close this PR.

PR #4926, opened 2021-08-10
EVG-14997: Merge test results audit

julianedwards:
EVG-14997: https://jira.mongodb.org/browse/EVG-14997

This replaces task DB functions that merge test results with the "NoMerge" equivalent versions of those functions where test results are not really needed. In a subsequent PR, I will add back the changes to merge cedar test results and we want to avoid making unnecessary calls to cedar (and downloading potentially large test result sets) when we do not need to.

Since this affects the task lifecycle, triggers, and UI I am putting multiple people on the PR for more eyeballs.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -1156,21 +1156,16 @@ func FindOneOld(query db.Q) (*Task, error) {
 	return task, err
 }
 
-// FindOld returns all task from the old tasks collection that satisfies the query.
-func FindOld(query db.Q) ([]Task, error) {
+// FindOldNoMerge returns all task from the old tasks collection that satisfies
+// the query without merging test results.
+func FindOldNoMerge(query db.Q) ([]Task, error) {
 	tasks := []Task{}
 	err := db.FindAllQ(OldCollection, query, &tasks)
 	if adb.ResultsNotFound(err) {
 		return nil, nil
 	}
-	for i, task := range tasks {
-		if err = task.MergeNewTestResults(); err != nil {
-			return nil, errors.Wrap(err, "error merging new test results")
-		}
-		tasks[i] = task
-	}
 
-	// remove display tasks from results
+	// Remove display tasks from results.
 	for i := len(tasks) - 1; i >= 0; i-- {
 		t := tasks[i]
 		if t.DisplayOnly {
@@ -1180,13 +1175,13 @@ func FindOld(query db.Q) ([]Task, error) {
 	return tasks, err
 }
 
-// FindOldWithDisplayTasks finds display and execution tasks in the old collection
-func FindOldWithDisplayTasks(query db.Q) ([]Task, error) {
-	tasks := []Task{}
-	err := db.FindAllQ(OldCollection, query, &tasks)
-	if adb.ResultsNotFound(err) {
-		return nil, nil
+// FindOld returns all task from the old tasks collection that satisfies the query.
+func FindOld(query db.Q) ([]Task, error) {
+	tasks, err := FindOldNoMerge(query)
+	if err != nil {
+		return nil, err
 	}
+
 	for i, task := range tasks {
 		if err = task.MergeNewTestResults(); err != nil {
 			return nil, errors.Wrap(err, "error merging new test results")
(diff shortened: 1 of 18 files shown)


ablack12 (commented):
Is the default typically to find without merge? I think the reason we've accidentally use test results in so many places is because it's not typical that a "FindOne" would do any extra work.  If we don't want this to recur in the future, we should rename the no merge to just "FindOne", and then rename the other to "FindOneWithMerge" so we only use it when we explicitly need test results

khelif96 (approved):

model/task_lifecycle.go
```diff
@@ -153,7 +153,7 @@ func SetActiveStateById(id, user string, active bool) error {
 // originalStepbackTask is only specified if we're first activating the generator for a generated task.
 func activatePreviousTask(taskId, caller string, originalStepbackTask *task.Task) error {
 	// find the task first
-	t, err := task.FindOne(task.ById(taskId))
+	t, err := task.FindOneId(taskId)
```
ablack12, line 156: this doesn't seem like it would need test results

model/task_lifecycle.go
```diff
@@ -272,7 +272,7 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 			if t.DisplayOnly {
 				execTasks := map[string]string{}
 				for _, et := range t.ExecutionTasks {
-					execTask, err = task.FindOne(task.ById(et))
+					execTask, err = task.FindOneId(et)
```
ablack12, line 275: in this case we only get the execTask long enough to log statuses, so I don't think we need to merge test results

model/task_lifecycle.go
```diff
@@ -311,7 +311,7 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 }
 
 func AbortTask(taskId, caller string) error {
-	t, err := task.FindOne(task.ById(taskId))
```
ablack12, line 314: do we need test results here (and if we do can we comment why)

model/task_lifecycle.go
```diff
@@ -396,7 +396,7 @@ func DeactivatePreviousTasks(t *task.Task, caller string) error {
 // otherwise. Note that the setting is obtained from the top-level
 // project, if not explicitly set on the task.
 func getStepback(taskId string) (bool, error) {
-	t, err := task.FindOne(task.ById(taskId))
+	t, err := task.FindOneId(taskId)
```
ablack12, line 399: We shouldn't need test results here; we're only using the task to find the correct project information

rest/route/annotations.go
```diff
@@ -384,7 +383,7 @@ func (h *createdTicketByTaskPutHandler) Parse(ctx context.Context, r *http.Reque
 	h.execution = execution
 
 	// check if the task exists
-	t, err := task.FindOne(task.ById(h.taskId))
+	t, err := task.FindOneId(h.taskId)
```
ablack12, line 386: this won't need test results

service/task.go
```diff
@@ -410,7 +410,7 @@ func (uis *UIServer) taskPage(w http.ResponseWriter, r *http.Request) {
 }
 
 func getAbortedBy(abortedByTaskId string) (*abortedByDisplay, error) {
-	abortedTask, err := task.FindOne(task.ById(abortedByTaskId))
+	abortedTask, err := task.FindOneId(abortedByTaskId)
```
ablack12, line 413: This wouldn't need test results

model/task/db.go
```diff
@@ -1180,25 +1175,67 @@ func FindOld(query db.Q) ([]Task, error) {
 	return tasks, err
 }
 
-// FindOldWithDisplayTasks finds display and execution tasks in the old collection
-func FindOldWithDisplayTasks(query db.Q) ([]Task, error) {
+// FindOld returns all task from the old tasks collection that satisfies the query.
```
Kimchelly, line 1178: Can you also mention that on top of finding the task, it also merges test results?

model/task_lifecycle.go
julianedwards: This is the bad naming I was talking about—`task.FindOneId` does not merge test results.

ablack12: Oh haha nice. Can we add NoMerge then to be consistent with the other no merge functions?

julianedwards: It is misleading, I know. I will try and fix it in the next PR.

ablack12: Or else `FindOneNoMerge(task.ById` is probably preferable.

julianedwards: Yeah, since I think I am going to get rid of merge functions in the next PR I think I would rather leave it as is.

PR #4931, opened 2021-08-11
EVG-14977 Unhide Adding and creating Patch Trigger Aliases

malikchaya2:
[EVG-14977 ](https://jira.mongodb.org/browse/EVG-14977)

### Description 
This was kept hidden till now so that users don't start using it before it is ready. While the feature is not yet fully ready to role out, at this point, a user using it won't break it, and they'll already see the downstream tasks on the patch page. 

### Testing 
none

### DIFF
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -484,7 +484,7 @@ Evergreen Projects
       </div>
       <div>
         <!--  Github Trigger Aliases  -->
-        <div id="github-trigger-alias" style="display: none;">
+        <div id="github-trigger-alias">
           <div class="h4">Add Github Trigger Alias </div>
           <div class="muted small">Add previously defined patch trigger aliases that add tasks from other projects to patches</div>
           <div class="form-group">
@@ -856,7 +856,7 @@ Evergreen Projects
         </div>
       </div>
       <!-- Patch Trigger Aliases -->
-      <div class="variables" style="display: none;">
+      <div class="variables">
         <div class="form-group">
           <div class="col-header col-lg-6 form-control-static">
             <h3>Patch Trigger Aliases</h3>


malikchaya2:
> I don't really have the context to verify the time is ripe to unhide these.

That makes sense. I'll add some more context in case it helps you or Annie. 
We have the following left before the project is fully ready: 

- letting users select aliases on the UI when configuring a patch 
- restarting downstream tasks when the parent is restarted 
- testing 

The following is good to go 

- Creating aliases
- applying aliases to patches created via the cli 
- applying aliases to github prs 
- viewing downstream tasks

ablack12 (commented):
Should we add to the comment of the section to say what's available to use and what's not available to use, or do you think we'll do this on a case by case basis?

ybrill:
I wonder about turning it on prior to the completion of [EVG-15119](https://jira.mongodb.org/browse/EVG-15119). It would be ideal if through testing we can ferret out the road bumps before users see them.

On the other hand, is there a compelling reason to enable it now? For example, is there a team waiting to use it that will be unblocked by enabling it?

malikchaya2:
> I wonder about turning it on prior to the completion of [EVG-15119](https://jira.mongodb.org/browse/EVG-15119). It would be ideal if through testing we can ferret out the road bumps before users see them.
> 
> On the other hand, is there a compelling reason to enable it now? For example, is there a team waiting to use it that will be unblocked by enabling it?

I'm already working on EVG-15119 so it wouldn't be a big deal to wait.

ablack12:
@malikchaya2 are we ready to unhide this?

malikchaya2:
The backend testing is done. The UI testing is not, but I don't know if this should be held up for that.

ablack12:
I'm fine to go ahead 👍

malikchaya2:
> I'm fine to go ahead 👍 Do we have documentation for this written up already? Without that, this option may be confusing.

We have the [CLI --list ](https://github.com/evergreen-ci/evergreen/wiki/Using-the-Command-Line-Tool/_compare/ddb6f1d563c01ea57a6a5e64aa3dd4427ded8544...4ba59f925cb5d5075b18636ec4a257d6293a2609)command documented, and the [downstream_expansion command](https://github.com/evergreen-ci/evergreen/wiki/Project-Commands/_compare/906fc57e09fff978194ace06dd1e2e3f42c5efa5...a8028981e537dce38d9bcd4e72201d3885dc326d) documented. 

There is currently no plan to add more documentation. Do you think more documentation is needed?

ablack12:
Ah gotcha -- yeah I think it would be helpful to add a section to this wiki page https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings explaining what the options are, how to set it up, etc. If people have questions  they usually go to the wiki first (and/or it's a convenient place for us to direct users). Might be reasonable to ticket this as a one pointer.

malikchaya2:
Sounds good, I created https://jira.mongodb.org/browse/EVG-15504

PR #4932, opened 2021-08-11
EVG-14997: Fetch test results lazily

julianedwards:
EVG-14997: https://jira.mongodb.org/browse/EVG-14997

- Lazily fetch test results.
- Remove all test results merging in task DB functions.
- Create new `PopulateTestResults` which populates the task's `LocalTestResults` field (even if it is a display task).
- Add a new local field, `testResultsPopulated`, to the `Task` struct to avoid populating test results more than once.
- Call `PopulateTestResults` where needed (this is the riskiest part of the PR).

### DIFF
--- a/graphql/ui_plugin_build_baron.go
+++ b/graphql/ui_plugin_build_baron.go
@@ -269,23 +269,27 @@ func BbGetProject(settings *evergreen.Settings, projectId string) (evergreen.Bui
 func BbGetTask(taskId string, executionString string) (*task.Task, error) {
 	execution, err := strconv.Atoi(executionString)
 	if err != nil {
-		return nil, errors.Wrap(err, "Invalid execution number")
+		return nil, errors.Wrap(err, "invalid execution number")
 	}
+
 	t, err := task.FindOneIdOldOrNew(taskId, execution)
 	if err != nil {
-		return nil, errors.Wrap(err, "problem finding task")
+		return nil, errors.Wrap(err, "finding task")
 	}
 	if t == nil {
-		return nil, errors.Errorf("No task found for taskId: %s and execution: %d", taskId, execution)
+		return nil, errors.Errorf("no task found for task id: %s and execution: %d", taskId, execution)
 	}
+
 	if t.DisplayOnly {
 		t.LocalTestResults, err = t.GetTestResultsForDisplayTask()
 		if err != nil {
-			return nil, errors.Wrapf(err, "Problem finding test results for display task '%s'", t.Id)
+			return nil, errors.Wrapf(err, "finding test results for display task '%s'", t.Id)
 		}
 	}
+
 	return t, nil
 }
+
 func (js *JiraSuggest) GetTimeout() time.Duration {
 	// This function is never called because we are willing to wait forever for the fallback handler
 	// to return JIRA ticket results.
--- a/graphql/ui_plugin_build_baron.go
+++ b/graphql/ui_plugin_build_baron.go
@@ -42,7 +42,6 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool, 
 	t, err := task.FindOne(task.ById(taskId))
 	if err != nil {
 		return taskNotFound, err
-
 	}
 	if t == nil {
 		taskNotFound = true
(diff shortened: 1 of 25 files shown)


model/task/task.go
```diff
@@ -208,6 +208,8 @@ type Task struct {
 
 	CanSync       bool             `bson:"can_sync" json:"can_sync"`
 	SyncAtEndOpts SyncAtEndOptions `bson:"sync_at_end_opts,omitempty" json:"sync_at_end_opts,omitempty"`
+
+	testResultsPopulated bool
```
ablack12, line 212: can we comment this so we understand in the future why it's local?

model/task/task.go
```diff
@@ -2047,21 +2058,47 @@ func (t *Task) makeArchivedTask() *Task {
...
 
-// MergeNewTestResults returns the task with both old (embedded in
-// the tasks collection) and new (from the testresults collection) test results
-// merged in the Task's LocalTestResults field.
-func (t *Task) MergeNewTestResults() error {
+// PopulateTestResults returns the task with both old (embedded in the tasks
+// collection) and new (from the testresults collection) OR Cedar test results
+// merged into the Task's LocalTestResults field.
+func (t *Task) PopulateTestResults() error {
+	if t.testResultsPopulated {
```
ablack12, line 2065: Does it not work to use len(LocalTestResults) > 0 or do you want to avoid repeatedly fetching if there are no results? Also, if the task is only started, I imagine we would want to get test results again in the future if the task has finished since they might have changed?

julianedwards: I think there is a case where we want to merge legacy test results and test results from the test results collection. We don't set populated until the task is finished (see L2068).

ybrill (commented):
Will this fix [the issue Maria raised](https://jira.mongodb.org/browse/EVG-14997?focusedCommentId=3992167&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3992167)? It looks like we're still only calling MergeTestResultsBulk there.
Other calls to MergeTestResultsBulk also still look like they won't get cedar test results.

model/task/task.go
ablack12: Are you referring to `if !evergreen.IsFinishedTaskStatus(t.Status) && t.Status != evergreen.TaskStarted {` ? So we do set it if the task is started but not finished.

service/task.go
```diff
@@ -923,75 +922,31 @@ func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, proj
...
+				TimeTaken: et.TimeTaken,
+				Status:    et.ResultStatus(),
 			})
 		}
 
-		// Check cedar first.
-		cedarTestResults := uis.getCedarTestResults(ctx, projCtx.Task, taskId)
-		for _, tr := range cedarTestResults {
-			testResults = append(testResults, task.ConvertCedarTestResult(tr))
+		for _, tr := range testResults {
```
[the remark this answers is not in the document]
julianedwards: Ah good catch, we actually do not need the `testResults` variable anymore. We just load all the display task's test results into `LocalTestResults`.

julianedwards:
@ybrill per our conversation this morning, I think we can merge this PR without making the changes to support the task history drawer since we are going to hold off turning on test results for mongodb-mongo-* projects.

graphql/ui_plugin_build_baron.go
```diff
@@ -269,23 +268,20 @@ func BbGetProject(settings *evergreen.Settings, projectId string) (evergreen.Bui
...
 	}
-	if t.DisplayOnly {
-		t.LocalTestResults, err = t.GetTestResultsForDisplayTask()
-		if err != nil {
-			return nil, errors.Wrapf(err, "Problem finding test results for display task '%s'", t.Id)
-		}
-	}
-	return t, nil
+
+	return t, errors.Wrap(t.PopulateTestResults(), "populating test results")
```
khelif96, line 282: We should split this up. We typically don't return a value along side errors. We prefer to return nil values when theres an error since we propagate that error to the UI. 
```suggestion
        err = t.PopulateTestResults()
        if err != nil {
             return nil, errors.Wrap(err, "Error while populating test results")
         }
	return t, nil
```

khelif96 (changes requested):

ybrill (approved):

khelif96 (approved):

PR #4933, opened 2021-08-12
TIG-3231: Iframe for new trend charts, plus url and sizing fixed.

ancostas:
[TIG-3231](https://jira.mongodb.org/browse/TIG-3231)

### Description 
Adds back the new trend charts iframe, but now handles sizing and url trusting properly. I'm using the same sizing spruce uses for these trend charts (100% of parent container width, 1000px height), for reference.

### Testing 
I was able to get a local version working and saw that this didn't affect the log section showing up. I also did a temporary hack to enable viewing this new panel without it being configured, and that also didn't affect the log section. I think it should be good to go, now.


ablack12 (approved):

ancostas:
evergreen merge

PR #4935, opened 2021-08-12
EVG-15190 use display task ID and facet to get version tasks

ablack12:
[EVG-15190](https://jira.mongodb.org/browse/EVG-15190)

### Description 
If the task has display task ID explicitly set to empty or displayTask is true, we don't have to do the lookup to find out if the task is an execution task, so we use facet to filter those out, and we only lookup execution tasks on the remaining tasks.

### Testing 
Added a unit test.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2856,7 +2856,7 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 		// display only is false (i.e. we don't know if it's a display task or not).
 		facet := bson.M{
 			"$facet": bson.M{
-				// we skip lookup for anything we already know is not part of a display task
+				// We skip lookup for anything we already know is not part of a display task
 				"id_empty": []bson.M{
 					{
 						"$match": bson.M{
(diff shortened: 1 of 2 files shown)


model/task/task.go
```diff
@@ -2852,23 +2852,60 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
...
+
+		// Recombine the tasks so that we can continue the pipeline on the joined tasks
+		recombineTasks := []bson.M{
+			{"$project": bson.M{
+				"tasks": bson.M{
+					"$setUnion": []string{"$no_id", "$id_empty"},
+				}},
+			},
+			{"$unwind": "$tasks"},
+			{"$replaceRoot": bson.M{"newRoot": "$tasks"}},
```
khelif96, line 2902: what does this do/ is it necessary?

khelif96 (commented):
This is pretty awesome and a super creative solution i just had one question, apart from that it LGTM

model/task/task.go
ablack12: Without replace root, we have the list of tasks nested under tasks, so every following part of the pipeline will need to reference "tasks.<whatever> instead of just having the list of tasks at the top level; this means that the rest of the pipeline would have to look differently based on whether or not IncludeExecutionTasks is true or not. Replace root is an inexpensive way to reorient the data so that it's in the same format as before we filtered out the exec tasks.

khelif96 (approved):

PR #4936, opened 2021-08-12
EVG-15210: Separate task annotation settings from BB settings

hadjri:
[EVG-15210](https://jira.mongodb.org/browse/EVG-15210)

### Description 
Build baron settings and task annotation settings are not directly related therefore it makes more sense for each of them to be set as their own entity while we are migrating these settings to the project page.  An additional flag has been added on the Admin page to toggle whether or not plugin settings developed for this project should be activated (PluginAdminPageDisabled)


graphql/ui_plugin_build_baron.go
```diff
@@ -52,11 +52,24 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
 	settings := env.Settings()
 	queue := env.RemoteQueue()
 	bbProject, ok := BbGetProject(settings, t.Project)
+	var webHook evergreen.WebHook
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return false, errors.Wrap(err, "error retrieving admin settings")
+	}
+	if flags.PluginAdminPageDisabled {
+		projectRef, err := model.FindOneProjectRef(t.Project)
```
ablack12, line 61: need to check if projectRef is nil. Also, we'll want to use `FindMergedProjectRef` so that we include the logic of the repo settings. (I think we aren't doing this right in a lot of places; I'll probably audit existing usages to make this more clear)

graphql/ui_plugin_build_baron.go
```diff
@@ -79,8 +92,21 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
...
 func IsWebhookConfigured(t *task.Task) bool {
-	bbProject, _ := BbGetProject(evergreen.GetEnvironment().Settings(), t.Project)
-	webHook := bbProject.TaskAnnotationSettings.FileTicketWebHook
+	var webHook evergreen.WebHook
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return false
+	}
+	if flags.PluginAdminPageDisabled {
+		projectRef, err := model.FindOneProjectRef(t.Project)
```
ablack12, line 101: same project ref comment

graphql/ui_plugin_build_baron.go
```diff
@@ -52,11 +52,24 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
 	settings := env.Settings()
 	queue := env.RemoteQueue()
 	bbProject, ok := BbGetProject(settings, t.Project)
+	var webHook evergreen.WebHook
```
ablack12, line 55: This is a big chunk of copied code now (with IsWebhooKConfigured). Can we refactor IsWebhookConfigured to return the webhook so that we can use it here?

model/project_parser.go
```diff
@@ -49,36 +50,37 @@ const LoadProjectError = "load project error(s)"
...
+	CallbackTimeout        *int                          `yaml:"callback_timeout_secs,omitempty" bson:"callback_timeout_secs,omitempty"`
+	Modules                []Module                      `yaml:"modules,omitempty" bson:"modules,omitempty"`
+	BuildVariants          []parserBV                    `yaml:"buildvariants,omitempty" bson:"buildvariants,omitempty"`
+	Functions              map[string]*YAMLCommandSet    `yaml:"functions,omitempty" bson:"functions,omitempty"`
+	TaskGroups             []parserTaskGroup             `yaml:"task_groups,omitempty" bson:"task_groups,omitempty"`
+	Tasks                  []parserTask                  `yaml:"tasks,omitempty" bson:"tasks,omitempty"`
+	ExecTimeoutSecs        *int                          `yaml:"exec_timeout_secs,omitempty" bson:"exec_timeout_secs,omitempty"`
+	Loggers                *LoggerConfig                 `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
+	CreateTime             time.Time                     `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
+	TaskAnnotationSettings evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
```
ablack12, line 83: Can't omitempty a struct, so a pointer to a struct is better

plugin/build_baron.go
```diff
@@ -36,7 +37,21 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
 	}
 
 	for projName, proj := range bbpOptions.Projects {
-		webhookConfigured := proj.TaskAnnotationSettings.FileTicketWebHook.Endpoint != ""
+		var webHook evergreen.WebHook
```
ablack12, line 40: same thought with "can we use isWebhookConfigured"

service/templates/admin.html
```diff
@@ -133,6 +133,17 @@ <h2 class="modal-title">[[modalTitle]]</h2>
 											</tr>
 										</thead>
 										<tbody>
+											<tr>
```
ablack12, line 136: Can we move this lower down in the list? The top list of flags are our "things are bad and we need to shut them off quick" kinds of actions.

service/templates/admin.html
```diff
@@ -133,6 +133,17 @@ <h2 class="modal-title">[[modalTitle]]</h2>
...
+												<td>Plugin settings on admin page</td>
```
ablack12, line 137: Another person looking at this may be confused about what it means, since the plugin settings will be visible on the admin page regardless of this toggle. Maybe "Use admin-level plugin settings" or etc?

ablack12 (changes requested):

model/project_parser.go
ablack12: Also, we need to add this to the mergeUnique function https://github.com/evergreen-ci/evergreen/blob/cd549ca996e22bf1b812023ed403c19bc84be478/model/project_parser.go#L1192

plugin/build_baron.go
```diff
@@ -36,7 +37,21 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
...
+		flags, err := evergreen.GetServiceFlags()
+		if err != nil {
+			return errors.Wrap(err, "error retrieving admin settings")
+		}
+		if flags.PluginAdminPageDisabled {
+			projectRef, err := model.FindOneProjectRef(projName)
```
ablack12, line 46: We're checking project ref; should we also be checking parser project if we're going to be allowing this to be defined in the yaml? (to do this we would probably get the parser project for the most recent version on this project)

ablack12 (changes requested):

ablack12:
Adding Chaya as a reviewer since she has the most context on task annotations.

malikchaya2 (approved):
Thanks Annie! 

This looks good to me once Annie's requested changes are addressed. Thank you for doing this!

graphql/ui_plugin_build_baron.go
```diff
@@ -52,24 +52,14 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
...
-		webHook = projectRef.TaskAnnotationSettings.FileTicketWebHook
-	} else {
-		webHook = bbProject.TaskAnnotationSettings.FileTicketWebHook
-	}
 	if !ok {
 		return taskNotFound, errors.Errorf("error finding build baron plugin for task '%s'", taskId)
 	}
 
+	webHook, ok := IsWebhookConfigured(t.Project)
+	if !ok {
```
ablack12, line 60: I'm not sure this should be treated as an error; it isn't in the original code. We just only hit 63 if it has been.

graphql/ui_plugin_build_baron.go
```diff
@@ -91,23 +81,35 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
...
+func IsWebhookConfigured(project string) (evergreen.WebHook, bool) {
 	var webHook evergreen.WebHook
 	flags, err := evergreen.GetServiceFlags()
 	if err != nil {
-		return false
+		return evergreen.WebHook{}, false
 	}
 	if flags.PluginAdminPageDisabled {
-		projectRef, err := model.FindOneProjectRef(t.Project)
+		parserProject, err := model.ParserProjectFindOneById(project)
```
ablack12, line 91: Parser projects are actually created for every version, not for every project. What you'll want to do instead is look up the most recent valid version for the project (https://github.com/evergreen-ci/evergreen/blob/65cdf890e49b342699e7e56e331a4917b4ae0773/model/version_db.go#L68) and then use that version ID to look up the parser project.

ablack12 (changes requested):

graphql/ui_plugin_build_baron.go
```diff
@@ -78,10 +78,39 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
...
-	webHook := bbProject.TaskAnnotationSettings.FileTicketWebHook
-	return webHook.Endpoint != ""
+func IsWebhookConfigured(project string) (evergreen.WebHook, bool) {
+	var webHook evergreen.WebHook
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return evergreen.WebHook{}, false
+	}
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(project, -1)
```
ablack12, line 88: Actually sorry! I told you something unnecessary. In the case that we know the task (it looks to me like this is everytime we use IsWebhookConfigured) we can just pass in t.Version and use that to look up the ParserProject directly, saving us a DB call (this won't work for BuildBaronPlugin Configure though; you may need to look up the Version outside if the flag is set and then pass in the ID. Should also check that version isn't nil)

plugin/build_baron.go
ablack12: Can we address this?

ablack12 (changes requested):

model/project_ref.go
```diff
@@ -105,6 +105,8 @@ type ProjectRef struct {
 	// This is a temporary flag to enable individual projects to use repo settings
 	UseRepoSettings bool   `bson:"use_repo_settings" json:"use_repo_settings" yaml:"use_repo_settings"`
 	RepoRefId       string `bson:"repo_ref_id" json:"repo_ref_id" yaml:"repo_ref_id"`
+
+	TaskAnnotationSettings evergreen.AnnotationsSettings `bson:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
```
ablack12, line 109: The unspoken flow of this section is that we only have it not in the first section if it needs a comment; in this case I think a comment is reasonable though (can you also move above 105; those are logically very different and I think make sense to keep at the end)

plugin/build_baron.go
```diff
@@ -36,7 +37,19 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
...
 	for projName, proj := range bbpOptions.Projects {
-		webhookConfigured := proj.TaskAnnotationSettings.FileTicketWebHook.Endpoint != ""
+		webHook := proj.TaskAnnotationSettings.FileTicketWebHook
+		flags, err := evergreen.GetServiceFlags()
+		if err != nil {
+			return errors.Wrap(err, "error getting service flags")
+		}
+		if flags.PluginAdminPageDisabled {
+			wh, ok := IsWebhookConfigured(projName, "")
+			if !ok {
```
ablack12, line 47: I don't think we should check this here; we already check if there's an error at line 53 and 56 but it's not actually an error otherwise (also we're wrapping an error that doesn't exisst)

plugin/build_baron.go
```diff
@@ -105,3 +118,75 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
 		},
 	}, nil
 }
+
+func IsWebhookConfigured(project string, version string) (evergreen.WebHook, bool) {
```
ablack12, line 122: Now that this function is a bit more complicated, it would be helpful to add a comment around what version is for and the lookup hierarchy

ablack12 (approved):
A couple of comment suggestions and an error that needs to be removed; I don't need to look again though!

plugin/build_baron.go
```diff
@@ -119,6 +115,9 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
 	}, nil
 }
 
+// IsWebhookConfigured webhook will can be retrieved from project or admin config depending on PluginAdminPageDisabled flag
```
ablack12, line 118: Actually sorry can you clean this up a bit? And with multiline comments, it's best to use full sentences/punctuation rather than relying on linebreaks

PR #4939, opened 2021-08-16
EVG-15215: Remove dashboard logic from plugin structure

hadjri:
[EVG-15215](https://jira.mongodb.org/browse/EVG-15215)

### Description 
The “dashboard” property in the admin plugin settings config appears to not be used anywhere in the plugin code.  It should be ok to remove.  On approval of this PR the dashboard property can be removed in the Evergreen Admin UI.


ablack12 (commented):
I feel like I might have been the one to say this was okay but I don't fully remember why; I think maybe because there isn't a file in the plugin package that uses the dashboard route? @julianedwards do you have context into whether this is used anymore? Seems like older branches.

hadjri:
> I feel like I might have been the one to say this was okay but I don't fully remember why; I think maybe because there isn't a file in the plugin package that uses the dashboard route? @julianedwards do you have context into whether this is used anymore? Seems like older branches.

Yes, there is no dashboard plugin registered in the plugin package when evergreen starts up and the web service is initialized.   There is a dashboard plugin REST route (/dashboard/tasks/project/{project_id}/version/{version_id}) but it does not appear to reference any plugin config.

julianedwards:
I do not have any context on this dashboard route.

ablack12 (approved):
🤞  I'm still confused about how this code became dead but; I guess let's try this and if it causes a problem we can revert

PR #4944, opened 2021-08-17
EVG-15256: include secret and task definition prefixes

Kimchelly:
JIra: https://jira.mongodb.org/browse/EVG-15256

### Description 
* Include required prefixes for secrets and task definitions. We need these because our IAM role only has permissions to use resources containing those prefixes.
* Set agent's container name.

### Testing 
Updated tests.

### DIFF
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -173,6 +173,10 @@ func ExportECSContainerResources(info pod.ContainerResourceInfo) cocoa.ECSContai
 	return *res
 }
 
+// agentContainerName is the standard name for the container that's running the
+// agent in a pod.
+const agentContainerName = "evg-agent"
+
 // ExportPodContainerDef exports the ECS pod container definition into the
 // equivalent cocoa.ECSContainerDefintion.
 func ExportPodContainerDef(settings *evergreen.Settings, p *pod.Pod) (*cocoa.ECSContainerDefinition, error) {
@@ -181,7 +185,7 @@ func ExportPodContainerDef(settings *evergreen.Settings, p *pod.Pod) (*cocoa.ECS
 		return nil, errors.Wrap(err, "building agent script")
 	}
 	return cocoa.NewECSContainerDefinition().
-		SetName("evg-agent").
+		SetName(agentContainerName).
 		SetImage(p.TaskContainerCreationOpts.Image).
 		SetMemoryMB(p.TaskContainerCreationOpts.MemoryMB).
 		SetCPU(p.TaskContainerCreationOpts.CPU).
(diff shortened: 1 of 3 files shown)


cloud/pod_util.go
```diff
@@ -171,17 +173,25 @@ func ExportECSContainerResources(info pod.ContainerResourceInfo) cocoa.ECSContai
...
+	if err != nil {
+		return nil, errors.Wrap(err, "building agent script")
+	}
 	return cocoa.NewECSContainerDefinition().
-		SetImage(opts.Image).
-		SetMemoryMB(opts.MemoryMB).
-		SetCPU(opts.CPU).
-		SetWorkingDir(opts.WorkingDir).
-		SetEnvironmentVariables(exportEnvVars(opts.EnvVars, opts.EnvSecrets)), nil
+		SetName("evg-agent").
```
julianedwards, line 184: Should "evg-agent" be a const somewhere?

cloud/pod_util.go
```diff
@@ -171,17 +173,25 @@ func ExportECSContainerResources(info pod.ContainerResourceInfo) cocoa.ECSContai
 	return *res
 }
 
-// ExportPodContainerDef exports the ECS pod container definition into the equivalent cocoa.ECSContainerDefintion.
-func ExportPodContainerDef(opts pod.TaskContainerCreationOptions) (*cocoa.ECSContainerDefinition, error) {
+// ExportPodContainerDef exports the ECS pod container definition into the
+// equivalent cocoa.ECSContainerDefintion.
+func ExportPodContainerDef(settings *evergreen.Settings, p *pod.Pod) (*cocoa.ECSContainerDefinition, error) {
```
julianedwards, line 178: A little late to the game here, but why do we pass in the evg settings? Can't we just fetch those from the global env?

Kimchelly: This is the standard way to use the Evergreen settings (especially in Amboy jobs), because calling `evergreen.GetEnvironment()` in arbitrary locations is difficult to debug since it depends on setting global state. It also makes testing difficult because if you want to inject the environment into a test as a dependency, this is harder because you have to always `evergreen.SetEnvironment` to ensure that all calls to `evergreen.GetEnvironment()` shares one global state (and later remember to un-set it or else it could affect future tests). It's usually preferable to call `evergreen.GetEnvironment()` once when the job starts running and pass that around.

cloud/pod_util.go
Kimchelly: Sure, I made it a constant.

julianedwards (approved):

PR #4947, opened 2021-08-17
EVG-15271: Fetch aliases available for a given patch

sophstad:
[EVG-15271](https://jira.mongodb.org/browse/EVG-15271)

### Description 
- Add resolver to return all trigger aliases that can be run as part of a patch
- Will attach variant/task data to `triggerAliases` field in a followup PR
- Let me know if the newlines should be removed (I kept them since they didn't break any lint rules and I think are generally good)

### Testing 
- Add query unit test
- Add `patch_trigger_aliases` field to local data for use in dev environment


graphql/schema.graphql
```diff
@@ -506,6 +506,11 @@ type FileDiff {
   description: String!
 }
 
+type TriggerAlias {
```
ablack12, line 509: Evergreen does have the concept of trigger aliases that aren't for patches, so I think to keep this consistent/clear this should still be named PatchTriggerAlias.

malikchaya2 (approved):

graphql/resolvers.go
```diff
@@ -972,6 +972,19 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
 	return *obj.Id, nil
 }
 
+func (r *patchResolver) TriggerAliases(ctx context.Context, obj *restModel.APIPatch) ([]*PatchTriggerAlias, error) {
```
ablack12, line 975: can we also rename this?

ablack12: (don't need to block merging on another review from me after this fix though)

sophstad: For sure, wasn't sure if that'd be redundant since it's a field within the `Patch` object. Will do, thanks!

PR #4948, opened 2021-08-17
EVG-15183 Change stepback tasks to have a configurable priority above mainline commits

bynn:
[EVG-15183](https://jira.mongodb.org/browse/EVG-15183)

### Description 
Add stepback task factor to admin page

### Testing 
the factor saves on local evergreen

### DIFF
--- a/model/distro/distro.go
+++ b/model/distro/distro.go
@@ -233,6 +233,7 @@ type PlannerSettings struct {
 	MainlineTimeInQueueFactor int64         `bson:"mainline_time_in_queue_factor" json:"mainline_time_in_queue_factor" mapstructure:"mainline_time_in_queue_factor"`
 	ExpectedRuntimeFactor     int64         `bson:"expected_runtime_factor" json:"expected_runtime_factor" mapstructure:"expected_runtime_factor"`
 	GenerateTaskFactor        int64         `bson:"generate_task_factor" json:"generate_task_factor" mapstructure:"generate_task_factor"`
+	StepbackTaskFactor        int64         `bson:"stepback_task_factor" json:"stepback_task_factor" mapstructure:"stepback_task_factor"`
 
 	maxDurationPerHost time.Duration
 }
@@ -363,6 +364,13 @@ func (d *Distro) GetMainlineTimeInQueueFactor() int64 {
 		return 1
 	}
 	return d.PlannerSettings.MainlineTimeInQueueFactor
+}
+
+func (d *Distro) GetStepbackTaskFactor() int64 {
+	if d.PlannerSettings.StepbackTaskFactor <= 0 {
+		return 1
+	}
+	return d.PlannerSettings.StepbackTaskFactor
 }
 
 func (d *Distro) GetExpectedRuntimeFactor() int64 {
@@ -695,6 +703,7 @@ func (d *Distro) GetResolvedPlannerSettings(s *evergreen.Settings) (PlannerSetti
 		MainlineTimeInQueueFactor: ps.MainlineTimeInQueueFactor,
 		ExpectedRuntimeFactor:     ps.ExpectedRuntimeFactor,
 		GenerateTaskFactor:        ps.GenerateTaskFactor,
+		StepbackTaskFactor:        ps.StepbackTaskFactor,
 		maxDurationPerHost:        evergreen.MaxDurationPerDistroHost,
 	}
 
--- a/model/distro/distro.go
+++ b/model/distro/distro.go
@@ -747,8 +747,8 @@ func (d *Distro) GetResolvedPlannerSettings(s *evergreen.Settings) (PlannerSetti
 		resolved.GenerateTaskFactor = config.GenerateTaskFactor
 	}
 
-	// StepbackTaskFactor is always undefined in distro
-	resolved.GenerateTaskFactor = config.StepbackTaskFactor
+	// StepbackTaskFactor isn't configurable by distro
+	resolved.StepbackTaskFactor = config.StepbackTaskFactor
 
 	if catcher.HasErrors() {
 		return PlannerSettings{}, errors.Wrapf(catcher.Resolve(), "cannot resolve PlannerSettings for distro '%s'", d.Id)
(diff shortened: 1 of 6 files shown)


ablack12 (commented):
This is a good start but we aren't actually using StepbackTaskFactor in this PR to change the priority of stepback tasks; some logic needs to be added to [this function](https://github.com/evergreen-ci/evergreen/blob/52e68f9fb10dfed06a2cab56f28191c994d2326e/scheduler/planner.go#L248) (sorry that earlier I said this doesn't need to be added to distro; I don't think it necessarily needs to be added to the distro page but we may still need it to be added to the distro struct that is used here)

scheduler/planner.go
```diff
@@ -231,6 +235,10 @@ func (unit *Unit) RankValue() int64 {
 		// give generators a boost so people don't have to wait twice.
 		priority = priority * unit.distro.GetGenerateTaskFactor()
 	}
+	if stepbackTask {
+		// give stepback tasks a priority over mainline
+		priority = priority * unit.distro.GetStepbackTaskFactor()
```
bynn, line 240: should this be `+=`? I saw that the generate tasks was the only one to not use the `+=` so I copied that

model/distro/distro.go
```diff
@@ -695,6 +703,7 @@ func (d *Distro) GetResolvedPlannerSettings(s *evergreen.Settings) (PlannerSetti
 		MainlineTimeInQueueFactor: ps.MainlineTimeInQueueFactor,
 		ExpectedRuntimeFactor:     ps.ExpectedRuntimeFactor,
 		GenerateTaskFactor:        ps.GenerateTaskFactor,
+		StepbackTaskFactor:        ps.StepbackTaskFactor,
```
ablack12, line 706: We need to add an if statement below to use the admin factor if it's not defined for the distro (or, since we aren't adding this to the distro page at all, we could just not add it to planner settings and set it immediately to the admin setting)

scheduler/planner.go
ablack12: We use += farther down because we're adding it to cachedValue specifically; here we're just modifying the base priority that we'll be using in the below calculations.

ablack12 (changes requested):

model/distro/distro.go
bynn: done

model/distro/distro.go
```diff
@@ -365,6 +366,13 @@ func (d *Distro) GetMainlineTimeInQueueFactor() int64 {
 	return d.PlannerSettings.MainlineTimeInQueueFactor
 }
 
+func (d *Distro) GetStepbackTaskFactor() int64 {
```
ablack12, line 369: we can get rid of this / remove d.PlannerSettings.StepbackTaskFactor

model/distro/distro.go
```diff
@@ -738,6 +746,10 @@ func (d *Distro) GetResolvedPlannerSettings(s *evergreen.Settings) (PlannerSetti
 	if resolved.GenerateTaskFactor == 0 {
 		resolved.GenerateTaskFactor = config.GenerateTaskFactor
 	}
+
+	// StepbackTaskFactor is always undefined in distro
```
ablack12, line 750: since we're removing it from distro entirely, we can maybe say "isn't configurable by distro"

scheduler/planner.go
```diff
@@ -231,6 +235,10 @@ func (unit *Unit) RankValue() int64 {
 		// give generators a boost so people don't have to wait twice.
 		priority = priority * unit.distro.GetGenerateTaskFactor()
 	}
+	if stepbackTask {
+		// give stepback tasks a priority over mainline
```
ablack12, line 239: I'm actually thinking through this math a bit more. if my priority is initially 10:
For a mainline commit with factor 20, priority is now 200. Then when we get to MainlineCommitFactor, which let's say is 10, we're adding an overall priority of 2000 ish.
For patches with 10 initial priority, and a mainline commit with factor 20, we're adding an overall priority of 200 (so the same factor produces very different results).

All this to say, maybe we should handle stepback in the same line as mainline commits, maybe by adding to the mainline commit factor or etc.

ablack12 (changes requested):

model/distro/distro.go
bynn: done

model/distro/distro.go
bynn: done

bynn: done

scheduler/planner.go
bynn: done

model/distro/distro.go
bynn: not done

scheduler/planner.go
```diff
@@ -249,9 +253,15 @@ func (unit *Unit) RankValue() int64 {
 		// of a bump, to avoid running older builds first.
 		avgLifeTime := timeInQueue / time.Duration(length)
 
+		mainlinePriority := int64(1)
 		if avgLifeTime < time.Duration(7*24)*time.Hour {
-			unit.cachedValue += priority * unit.distro.GetMainlineTimeInQueueFactor() * int64((7*24*time.Hour - avgLifeTime).Hours())
+			mainlinePriority = unit.distro.GetMainlineTimeInQueueFactor() * int64((7*24*time.Hour - avgLifeTime).Hours())
```
ablack12, line 258: we should do += since if GetMainlineTimeInQueueFactor is 0 (i think it is right now actually) we'll be setting priority to 0 too

ablack12 (approved):
One last change, then LGTM!

scheduler/planner.go
bynn: All the Get__Factor functions returns 1 if the factor is less than or equal to 0
Should I still make this change? I didn't know if adding the extra 1 would make a difference

ablack12: Ah okay that's fine then

ablack12: (Does look like tests need to be fixed though)

bynn: it was off by 1 because the cachedValue would be 1 when it would normally be 0 so I went ahead and still had to make this change!

ablack12 (approved):

PR #4950, opened 2021-08-18
EVG-14997: fetch display task before setting has_cedar_results

julianedwards:


model/task/task.go
```diff
@@ -1058,7 +1058,11 @@ func (t *Task) SetHasCedarResults(hasCedarResults, failedResults bool) error {
 	}
 
 	if !t.DisplayOnly && t.IsPartOfDisplay() {
-		return t.DisplayTask.SetHasCedarResults(hasCedarResults, failedResults)
+		displayTask, err := t.GetDisplayTask()
+		if err != nil {
+			return errors.Wrap(err, "getting display task")
+		}
+		return displayTask.SetHasCedarResults(hasCedarResults, failedResults)
```
ablack12, line 1065: If you want this to destructively modify t, you'll want to still use t.DisplayTask here

ablack12 (approved):

model/task/task.go
julianedwards: `GetDisplayTask` returns a pointer to the display task, so it will modify it either way.

PR #4951, opened 2021-08-18
EVG-14964 Add originalStatus

malikchaya2:
[EVG-14964](https://jira.mongodb.org/browse/EVG-14964)

### Description 
This adds the OriginalStatus field so that we can overwrite the display status to be  known-issue when needed. Doing that helps keep the cache in sync with how it's done on the tasks page. Once this is merged, I will make the necessary ui changes. 

### Testing 
graphql on staging
image

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2591,20 +2591,13 @@ func (r *taskResolver) CanSetPriority(ctx context.Context, obj *restModel.APITas
 }
 
 func (r *taskResolver) Status(ctx context.Context, obj *restModel.APITask) (string, error) {
-	annotation, err := annotations.FindOneByTaskIdAndExecution(*obj.Id, obj.Execution)
-	if err != nil {
-		return "", InternalServerError.Send(ctx, fmt.Sprintf("error finding task annotation: %s", err.Error()))
-	}
-
-	if annotation != nil && len(annotation.Issues) > 0 {
-		return evergreen.TaskKnownIssue, nil
-
-	}
 	return *obj.DisplayStatus, nil
 }
 
 func (r *taskResolver) OriginalStatus(ctx context.Context, obj *restModel.APITask) (*string, error) {
-	return obj.Status, nil
+	// this is needed for the task (as opposed to tasks) so that we don't overwrite
+	// the Status field in the cache, since tasks handles the status value differently
+	return obj.DisplayStatus, nil
 }
 
 func (r *taskResolver) LatestExecution(ctx context.Context, obj *restModel.APITask) (int, error) {
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2591,20 +2591,20 @@ func (r *taskResolver) CanSetPriority(ctx context.Context, obj *restModel.APITas
 }
 
 func (r *taskResolver) Status(ctx context.Context, obj *restModel.APITask) (string, error) {
+	annotation, err := annotations.FindOneByTaskIdAndExecution(*obj.Id, obj.Execution)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


graphql/schema.graphql
```diff
@@ -765,11 +767,13 @@ type Task {
   spawnHostLink: String
   startTime: Time
   status: String!
+  originalStatus: String
   taskGroup: String
   taskGroupMaxHosts: Int
   timeTaken: Duration
   totalTestCount: Int!
-  version: String! @deprecated(reason: "version is deprecated. Use versionMetadata instead.")
+  version: String!
```
khelif96, line 775: I think the directive has to be placed on the same line as the field. https://www.apollographql.com/docs/apollo-server/schema/directives/

rest/model/task.go
```diff
@@ -247,6 +248,20 @@ func (at *APITask) BuildFromService(t interface{}) error {
 				PRClosed:   v.AbortInfo.PRClosed,
 			},
 		}
+
+		annotation, err := annotations.FindOneByTaskIdAndExecution(id, v.Execution)
```
khelif96, line 252: Can we do this elsewhere? 
This  db query will be called every time we call buildFromService. Even in places where it is not needed. 
 Since we only need this in the graphql resolver we should do this work there.

khelif96 (changes requested):

graphql/schema.graphql
malikchaya2: oof, my linter does it automatically on save.

rest/model/task.go
malikchaya2: yes, I can move it to the status resolver

graphql/resolvers.go
```diff
@@ -2591,9 +2591,22 @@ func (r *taskResolver) CanSetPriority(ctx context.Context, obj *restModel.APITas
...
 func (r *taskResolver) Status(ctx context.Context, obj *restModel.APITask) (string, error) {
+	annotation, err := annotations.FindOneByTaskIdAndExecution(*obj.Id, obj.Execution)
+	if err != nil {
+		return "", InternalServerError.Send(ctx, fmt.Sprintf("error finding task annotation: %s", err.Error()))
+	}
+
+	if annotation != nil && len(annotation.Issues) > 0 {
+		return evergreen.TaskKnownIssue, nil
+
+	}
```
khelif96, line 2602: We can make some optimizations so we don't this lookup unnecessarily
`DisplayStatus` might already be `KnownIssue` if a task is returned by the `GetTasksByVersion` query so we should check if the status is already Known Issue and if so don't do this query. 

We should also only do this check if a task is in a state where there could be annotations 
eg. Only a failure status. We are already planning on making the same change in the the getTasksByVersion query https://jira.mongodb.org/browse/EVG-15279 

tbh we might even be able to omit this logic and only add the `OriginalStatus` field  and leave this as is. What do you think?

khelif96 (changes requested):

graphql/resolvers.go
```diff
@@ -2594,6 +2594,12 @@ func (r *taskResolver) Status(ctx context.Context, obj *restModel.APITask) (stri
 	return *obj.DisplayStatus, nil
 }
 
+func (r *taskResolver) OriginalStatus(ctx context.Context, obj *restModel.APITask) (string, error) {
+	// this is needed for the task (as opposed to tasks) so that we don't overwrite
+	// the Status field in the cache, since tasks handles the status value differently
+	return *obj.DisplayStatus, nil
+}
```
khelif96, line 2601: This does the same thing as the status resolver

malikchaya2: yes. We need the same thing without overwriting the cache.

PR #4954, opened 2021-08-18
EVG-14343 Add nil check for agent panic

bsamek:
[EVG-14343](https://jira.mongodb.org/browse/EVG-14343)


agent/agent.go
```diff
@@ -731,7 +731,7 @@ func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupC
 	}
 
 	if a.shouldKill(tc, ignoreTaskGroupCheck) {
-		if tc.task.ID != "" {
+		if tc.task.ID != "" && tc.taskConfig != nil {
```
ybrill, line 734: an alternative that occurs to me is that on the case where `tc.taskConfig == nil` we could call `KillSpawnedProcs` with the empty string as the workdir (since we check for the empty string [here](https://github.com/evergreen-ci/evergreen/blob/ef19ea1fa0e2ca32b375a46402a2b7bae0bb463f/agent/util/subtree_linux.go#L160-L162) and [here](https://github.com/evergreen-ci/evergreen/blob/ef19ea1fa0e2ca32b375a46402a2b7bae0bb463f/agent/util/subtree_darwin.go#L91-L93))

bsamek: We have multiple implementations of `getPIDsToKill` and would have to remember to correctly handle an empty working directory in all of them. On the other hand, the disadvantage of the solution I proposed is that we wouldn't run `killProcs` as completely as we would before.

I pushed a different solution. In my new code, we avoid running `killProcs` in the `runPostGroupCommands` function, since there's no reason to. Let me know if that makes sense.

agent/agent.go
```diff
@@ -658,10 +658,13 @@ func (a *Agent) runPostTaskCommands(ctx context.Context, tc *taskContext) error
...
 func (a *Agent) runPostGroupCommands(ctx context.Context, tc *taskContext) {
 	defer a.removeTaskDirectory(tc)
-	defer a.killProcs(ctx, tc, true)
 	if tc.taskConfig == nil {
 		return
 	}
+	// Only killProcs if tc.taskConfig is not nil. This avoids passing an
+	// empty working directory to killProcs, and is okay because this
+	// killProcs is only for the processes run in runPostGroupCommands.
+	defer a.killProcs(ctx, tc, true)
```
ablack12, line 667: I think even if we don't do this in runPost GroupCommands, it still makes sense to keep the nil check in killProcs. Is there a reason we aren't concerned about this in the other cases?

bsamek: We've only seen this panic happen in the `runPostGroupCommands` case, but yeah, it does seem like reasonable defensive programming to keep the nil check. I put it back.

bsamek:
@ablack12 @ybrill FYI I added a line to the .gitignore file for ag and ripgrep users, since I found myself using an .ignore. It's not covered in this ticket but seems trivial.

agent/agent.go
ybrill: will it _always_ be what we want to completely skip killProcs when it's nil?

bsamek: I don't know, but I think this is a future code maintainability problem. I explained this in my other comment. We have multiple implementations of `getPIDsToKill`, and I don't think it should be the responsibility of a function to handle odd input, since it puts the burden on the function of understanding what an empty directory means.

ybrill (approved):

ablack12 (approved):

PR #4957, opened 2021-08-19
EVG-14219 Continue on error in stats collector

bsamek:
[EVG-14219](https://jira.mongodb.org/browse/EVG-14219)

### Description 
Currently the stats collector doesn't run later commands if an earlier one
fails. This adds the ContinueOnError jasper option so that they all run.

### Testing 
Since this is internal jasper behavior, I did not test this.


bsamek:
Adding @Kimchelly in case I misunderstood the option.

Kimchelly (approved):

PR #4958, opened 2021-08-19
EVG-15060 save project settings for section

ablack12:
[EVG-15060](https://jira.mongodb.org/browse/EVG-15060)

### Description 
In order to use the existing connectors for this function, we need to split the operations by the type of connector it's part of. I made GetProjectSettingsBySection to handle all of the DBProjectConnector operations, and then separately made a new alias function and modified the subscription function, so these can all be called by the resolver. 
(The alternative to this would be to move all the logic for the connectors into model functions and not use connectors at all, but I think that's messier in the long run.)

### Testing 
Added tests to the new project settings function and the alias function.

### DIFF
--- a/rest/data/aliases.go
+++ b/rest/data/aliases.go
@@ -131,7 +131,7 @@ func (pc *DBAliasConnector) UpdateAliasesForSection(projectId string, updatedAli
 		if shouldSkipAliasForSection(section, originalAlias.Alias) {
 			continue
 		}
-		id := originalAlias.ID.String()
+		id := originalAlias.ID.Hex()
 		if _, ok := aliasesIdMap[id]; !ok {
 			catcher.Add(model.RemoveProjectAlias(id))
 		}
--- a/rest/data/aliases.go
+++ b/rest/data/aliases.go
@@ -1,6 +1,7 @@
 package data
 
 import (
+	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/model"
 	restModel "github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/evergreen-ci/utility"
@@ -107,6 +108,49 @@ func (d *DBAliasConnector) UpdateProjectAliases(projectId string, aliases []rest
 	return catcher.Resolve()
 }
 
+// should be called in the resolver
+func (pc *DBAliasConnector) UpdateAliasesForSection(projectId string, updatedAliases []restModel.APIProjectAlias,
+	originalAliases []model.ProjectAlias, section model.ProjectRefSection) (bool, error) {
+	aliasesIdMap := map[string]bool{}
+	aliasesToUpdate := []restModel.APIProjectAlias{}
+
+	for _, a := range updatedAliases {
+		if shouldSkipAliasForSection(section, utility.FromStringPtr(a.Alias)) {
+			continue
+		}
+		aliasesToUpdate = append(aliasesToUpdate, a)
+		aliasesIdMap[utility.FromStringPtr(a.ID)] = true
+	}
+	if err := pc.UpdateProjectAliases(projectId, aliasesToUpdate); err != nil {
+		return false, errors.Wrap(err, "error updating project aliasesToUpdate")
+	}
+	catcher := grip.NewBasicCatcher()
+	// delete any aliasesToUpdate that were in the list before but are not now
+	for _, originalAlias := range originalAliases {
+		// only look at the relevant aliases to update
+		if shouldSkipAliasForSection(section, originalAlias.Alias) {
+			continue
+		}
+		id := originalAlias.ID.String()
+		if _, ok := aliasesIdMap[id]; !ok {
+			catcher.Add(model.RemoveProjectAlias(id))
+		}
+	}
+	return true, catcher.Resolve()
+}
+
+func shouldSkipAliasForSection(section model.ProjectRefSection, alias string) bool {
+	// if we're updating internal aliases, skip non-internal aliases
+	if section == model.ProjectRefGithubAndCQSection && !utility.StringSliceContains(evergreen.InternalAliases, alias) {
+		return true
+	}
+	// if we're updating patch aliases, skip internal aliases
+	if section == model.ProjectRefPatchAliasSection && utility.StringSliceContains(evergreen.InternalAliases, alias) {
+		return true
+	}
+	return false
+}
+
 //  GetMatchingGitTagAliasesForProject returns matching git tag aliases that match the given git tag
 func (d *DBAliasConnector) HasMatchingGitTagAliasAndRemotePath(projectId, tag string) (bool, string, error) {
 	aliases, err := model.FindMatchingGitTagAliasesInProject(projectId, tag)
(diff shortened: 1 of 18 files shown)


rest/data/aliases.go
```diff
@@ -107,6 +108,49 @@ func (d *DBAliasConnector) UpdateProjectAliases(projectId string, aliases []rest
...
+		}
+		aliasesToUpdate = append(aliasesToUpdate, a)
+		aliasesIdMap[utility.FromStringPtr(a.ID)] = true
+	}
+	if err := pc.UpdateProjectAliases(projectId, aliasesToUpdate); err != nil {
+		return false, errors.Wrap(err, "error updating project aliasesToUpdate")
+	}
+	catcher := grip.NewBasicCatcher()
+	// delete any aliasesToUpdate that were in the list before but are not now
+	for _, originalAlias := range originalAliases {
```
malikchaya2, line 129: what does this loop do? where is it removing it from and why is it necessary after UpdateProjectAliases()?

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
 	return nil
 }
 
+// SaveProjectSettingsForSection passes in the existing state of the project page, but saves only the pieces
+// related to the specified section. This handles project ref, webhooks, auth, and project variables.
+func (pc *DBProjectConnector) SaveProjectSettingsForSection(ctx context.Context, projectId string, changes *restModel.APIProjectSettings,
```
malikchaya2, line 85: same note about using a connector here.

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
...
+	// used to verify that if something is set to nil in the request, we properly validate using the merged project ref
+	mergedProjectRef, err := model.GetProjectRefMergedWithRepo(*newProjectRef)
+	if err != nil {
+		return errors.Wrapf(err, "error merging project ref")
+	}
+
+	catcher := grip.NewBasicCatcher()
+	modified := false
+	switch section {
+	// aliases and notifications use a different connector and should be handled separated
```
malikchaya2, line 106: Is it possible to add aliases and notifications to this switch statement if we don't use connectors?

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
...
+		modified = true
+		if !before.ProjectRef.IsRestricted() && mergedProjectRef.IsRestricted() {
+			catcher.Wrap(before.ProjectRef.MakeRestricted(), "error making project restricted")
+		}
+		if before.ProjectRef.IsRestricted() && !mergedProjectRef.IsRestricted() {
+			catcher.Wrap(before.ProjectRef.MakeUnrestricted(), "error making unrestricted")
+		}
+
+	case model.ProjectRefVariablesSection:
+		// Remove any variables that only exist in the original settings. (It's possible we can send this information
```
malikchaya2, line 131: what do you mean by "(It's possible we can send this information from the UI in the future instead)"?

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
...
+				toDelete = append(toDelete, key)
+			}
+		}
+		changes.Vars.VarsToDelete = toDelete
+		if err = pc.UpdateProjectVars(projectId, &changes.Vars, false); err != nil { // destructively modifies vars
+			return errors.Wrapf(err, "Database error updating variables for project '%s'", projectId)
+		}
+	}
+	// save project ref only after the rest of the section has been validated.
+	// How do we know here if ProjectRef should be nil or not? Or should we just pass in the project and update regardless of whether or not it's changed (like the API route?)
```
malikchaya2, line 145: I don't know if this answers the question, but ideally we shouldn't let a user press save if they didn't change anything.

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
...
+			if _, ok := changes.Vars.Vars[key]; !ok {
+				toDelete = append(toDelete, key)
+			}
+		}
+		changes.Vars.VarsToDelete = toDelete
+		if err = pc.UpdateProjectVars(projectId, &changes.Vars, false); err != nil { // destructively modifies vars
+			return errors.Wrapf(err, "Database error updating variables for project '%s'", projectId)
+		}
+	}
+	// save project ref only after the rest of the section has been validated.
```
malikchaya2, line 144: what do you mean by this? Where is it validated?

rest/data/project.go
```diff
@@ -80,6 +80,81 @@ func (pc *DBProjectConnector) UpdateRepo(repoRef *model.RepoRef) error {
...
+	mergedProjectRef, err := model.GetProjectRefMergedWithRepo(*newProjectRef)
+	if err != nil {
+		return errors.Wrapf(err, "error merging project ref")
+	}
+
+	catcher := grip.NewBasicCatcher()
+	modified := false
+	switch section {
+	// aliases and notifications use a different connector and should be handled separated
+	case model.ProjectRefGeneralSection:
```
malikchaya2, line 107: should modified be set to true in this section and in the ProjectRefVariablesSection?

rest/data/aliases.go
```diff
@@ -107,6 +108,49 @@ func (d *DBAliasConnector) UpdateProjectAliases(projectId string, aliases []rest
...
+
+	for _, a := range updatedAliases {
+		if shouldSkipAliasForSection(section, utility.FromStringPtr(a.Alias)) {
+			continue
+		}
+		aliasesToUpdate = append(aliasesToUpdate, a)
+		aliasesIdMap[utility.FromStringPtr(a.ID)] = true
+	}
+	if err := pc.UpdateProjectAliases(projectId, aliasesToUpdate); err != nil {
+		return false, errors.Wrap(err, "error updating project aliasesToUpdate")
```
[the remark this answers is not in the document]
ablack12: ack find and replace fails me

rest/data/aliases.go
ablack12: This deletes any aliases that have been deleted by the user, i.e. they existed originally (in the before ProjectSettings struct) but don't exist anymore. It's removing those aliases from the alias collection.

rest/data/project.go
ablack12: Here we're building toDelete based on what existed before vs what exists now. We could just build toDelete by sending that information over from the frontend (i.e. adding the variable key to toDelete if the delete button is pressed)

rest/data/project.go
ablack12: They may have changed other things but not the ProjectRef itself. In the variables section case for example, is it possible we won't pass in a project ref at all since we didn't have to search for it yet? (This is the only section that doesn't need project ref though; maybe we should just always get it)

rest/data/project.go
ablack12: I think this is a future note; we discussed validating in the resolver

rest/data/aliases.go
```diff
@@ -107,6 +108,49 @@ func (d *DBAliasConnector) UpdateProjectAliases(projectId string, aliases []rest
 	return catcher.Resolve()
 }
 
+// should be called in the resolver
+func (pc *DBAliasConnector) UpdateAliasesForSection(projectId string, updatedAliases []restModel.APIProjectAlias,
```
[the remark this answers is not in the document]
malikchaya2: I was wrong about this. Resolving the comment.

rest/data/project.go
ablack12: Rearranged for clarity

rest/data/project.go
ablack12: yes thank you!

rest/data/aliases.go
malikchaya2: ah

rest/data/project.go
malikchaya2: ah I see. I agree with the approach of putting more in the backend, and re-assessing if we need less when building out the UI.

rest/data/project.go
malikchaya2: Isn't there potential for this to be true in ProjectRefGeneralSection and ProjectRefAccessSection as well, where they may update something that's not part of the projectRef?

malikchaya2: I wonder if we can maybe maintain a boolean on the front end. I don't think I fully understand the implications though and how necessary it is to know...

malikchaya2 (approved):

PR #4959, opened 2021-08-19
EVG-15212: Only reference a project’s configuration when interacting with build baron configs

hadjri:
[EVG-15212](https://jira.mongodb.org/browse/EVG-15212)

### Description 
Currently all build baron configs are derived from the admin config.  Change has been made to introduce the option of getting build baron settings from the project OR admin config depending on the state of the PluginAdminPageDisabled service flag.


model/project_parser.go
```diff
@@ -81,6 +81,7 @@ type ParserProject struct {
 	Loggers                *LoggerConfig                  `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
 	CreateTime             time.Time                      `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
+	BuildBaronProject      *evergreen.BuildBaronProject   `yaml:"build_baron_project,omitempty" bson:"build_baron_project,omitempty"`
```
ablack12, line 84: Nit: can we just call this BuildBaron or BuildBaronSettings? Since this is part of a Project configuration, Project as part of the variable name is a little redundant/reads more like a bool.

model/project_parser.go
```diff
@@ -1195,7 +1197,7 @@ func (pp *ParserProject) mergeOrderedUnique(toMerge *ParserProject) error {
 
 // mergeUnique merges fields that are non-lists.
 // These fields can only be defined 1 yaml.
-// These fields are: [stepback, batch time, pre error fails task, OOM tracker, display name, command type, callback/exec timeout, task annotations]
+// These fields are: [stepback, batch time, pre error fails task, OOM tracker, display name, command type, callback/exec timeout]
```
ablack12, line 1200: You removed task annotations from this list, but I imagine you meant to add build baron

model/project_ref.go
```diff
@@ -1360,6 +1365,22 @@ func saveProjectRefForSection(projectId string, p *ProjectRef, section ProjectRe
 					projectRefTriggersKey: p.Triggers,
 				},
 			})
+	case ProjectRefBuildBaronSection:
```
ablack12, line 1368: This probably isn't right--this function is used for updating the new UI page and we haven't decided where Build baron / task annotation are going to live on this new page. I think we should just add a TODO with the ticket number for designing this.

plugin/build_baron.go
```diff
@@ -178,6 +178,29 @@ func BbGetConfig(settings *evergreen.Settings) map[string]evergreen.BuildBaronPr
 }
 
 func BbGetProject(settings *evergreen.Settings, projectId string) (evergreen.BuildBaronProject, bool) {
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return evergreen.BuildBaronProject{}, false
+	}
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectId, -1)
+		if err != nil || lastGoodVersion == nil {
```
ablack12, line 187: if lastGoodVersion == nil, I think we'd want to continue to looking in the project ref instead

plugin/build_baron.go
```diff
@@ -178,6 +178,29 @@ func BbGetConfig(settings *evergreen.Settings) map[string]evergreen.BuildBaronPr
...
+			return evergreen.BuildBaronProject{}, false
+		}
+		parserProject, err := model.ParserProjectFindOneById(lastGoodVersion.Id)
+		if err != nil {
+			return evergreen.BuildBaronProject{}, false
```
ablack12, line 192: Can we add a comment to explain what's happening / the bool return value, now that this function is more complex?

service/ui.go
```diff
@@ -89,16 +89,15 @@ func NewUIServer(env evergreen.Environment, queue amboy.Queue, home string, fo T
 	}
 
 	uis := &UIServer{
-		Settings:           *settings,
-		env:                env,
-		queue:              queue,
-		Home:               home,
-		clientConfig:       evergreen.GetEnvironment().ClientConfig(),
-		CookieStore:        sessions.NewCookieStore([]byte(settings.Ui.Secret)),
-		buildBaronProjects: plugin.BbGetConfig(settings),
```
ablack12, line 98: Why did you get rid of this? I'm not sure we can do this.

plugin/build_baron.go
```diff
@@ -178,6 +178,29 @@ func BbGetConfig(settings *evergreen.Settings) map[string]evergreen.BuildBaronPr
 }
 
 func BbGetProject(settings *evergreen.Settings, projectId string) (evergreen.BuildBaronProject, bool) {
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return evergreen.BuildBaronProject{}, false
+	}
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectId, -1)
```
ablack12, line 186: Similar to the last PR, there are some places where we call BbGetProject where we can pass in the version ID from the task, and then we only need to call this if that's empty.

ablack12 (changes requested):
We should write a unit test for BbGetConfig

service/ui.go
hadjri: I removed buildBaronProjects from the UIServer because it was not referenced anywhere in the codebase, and we will not need this to live in UIServer moving forward.

hadjri:
Apologies, disregard re-requested review for the moment, adding BbGetConfig unit test

hadjri:
Added.

model/project.go
```diff
@@ -60,6 +60,7 @@ type Project struct {
 	ExecTimeoutSecs        int                            `yaml:"exec_timeout_secs,omitempty" bson:"exec_timeout_secs"`
 	Loggers                *LoggerConfig                  `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
+	BuildBaronSettings     *evergreen.BuildBaronSettings  `yaml:"build_baron_project,omitempty" bson:"build_baron_project,omitempty"`
```
ablack12, line 63: Since this is a new field, we can also change the yaml and bson tags to match the variable name

model/project_parser.go
```diff
@@ -81,6 +81,7 @@ type ParserProject struct {
 	Loggers                *LoggerConfig                  `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
 	CreateTime             time.Time                      `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
+	BuildBaronSettings     *evergreen.BuildBaronSettings  `yaml:"build_baron_project,omitempty" bson:"build_baron_project,omitempty"`
```
ablack12, line 84: Same with yaml/bson tags here

model/project_ref.go
```diff
@@ -105,6 +105,8 @@ type ProjectRef struct {
 	// TaskAnnotationSettings holds settings for the file ticket button in the Task Annotations to call custom webhooks when clicked
 	TaskAnnotationSettings evergreen.AnnotationsSettings `bson:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
 
+	BuildBaronSettings evergreen.BuildBaronSettings `bson:"build_baron_project,omitempty" json:"build_baron_project,omitempty" yaml:"build_baron_project,omitempty"`
```
ablack12, line 108: Same with bson/yaml/json here

model/project_ref.go
```diff
@@ -337,6 +340,8 @@ const (
 	ProjectPageWorkstationsSection   = "workstations"
 	ProjectPageTriggersSection       = "triggers"
 	ProjectPagePeriodicBuildsSection = "periodic-builds"
+	ProjectRefBuildBaronSection      = "build_baron"
```
ablack12, line 343: Can you remove these variables since they aren't used yet?

plugin/build_baron.go
```diff
@@ -102,9 +102,13 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
...
 				DataFunc: func(context UIContext) (interface{}, error) {
-					enabled := len(bbp.opts.Projects[context.ProjectRef.Id].TicketSearchProjects) > 0
+					bbSettings, ok := BbGetProject(evergreen.GetEnvironment().Settings(), context.ProjectRef.Id, "")
+					if !ok {
+						return nil, nil
+					}
+					enabled := len(bbSettings.TicketSearchProjects) > 0
 					if !enabled {
-						enabled = len(bbp.opts.Projects[context.ProjectRef.Identifier].TicketSearchProjects) > 0
+						enabled = len(bbSettings.TicketSearchProjects) > 0
```
ablack12, line 111: Looks like we don't need this if statement anymore, since we're setting it to the same thing as 109

plugin/build_baron.go
```diff
@@ -127,10 +131,9 @@ func IsWebhookConfigured(project string, version string) (evergreen.WebHook, boo
 	if flags.PluginAdminPageDisabled {
 		if version == "" {
 			lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(project, -1)
-			if err != nil || lastGoodVersion == nil {
-				return evergreen.WebHook{}, false
+			if err == nil && lastGoodVersion != nil {
```
ablack12, line 134: Now if version == "" even after checking lastGoodVersion, we make an unnecessary parserProject DB query. I think you want to add an `if version != ""` and rearrange the ifs/elses

plugin/build_baron.go
```diff
@@ -177,7 +180,35 @@ func BbGetConfig(settings *evergreen.Settings) map[string]evergreen.BuildBaronPr
 	return projects
 }
 
-func BbGetProject(settings *evergreen.Settings, projectId string) (evergreen.BuildBaronProject, bool) {
+// BbGetProject build baron settings will be retrieved from either project or admin config depending on PluginAdminPageDisabled flag
```
ablack12, line 183: Thanks for writing a function comment! A couple of thoughts: 
Shouldn't rely on line breaks to indicate the end of a sentence, since some comments truly are just long; should use periods.
Can you try to shorten this comment up? It's not easy to skim / understand quickly what's going on. For example 186 can be simplified to "returns build baron settings and ok if found". 185 can probably be included in 184 ("project parser config takes precedence (found using version ID if given)")

plugin/build_baron_test.go
```diff
@@ -154,3 +163,55 @@ func TestBuildBaronPluginConfigureBFSuggestion(t *testing.T) {
...
+	assert.NoError(t, myProject2.Insert())
+	assert.NoError(t, myProjectParser.Insert())
+
+	env := evergreen.GetEnvironment().Settings()
+	flags := evergreen.ServiceFlags{
+		PluginAdminPageDisabled: true,
+	}
+	assert.NoError(t, evergreen.SetServiceFlags(flags))
+
+	bbProj, _ := BbGetProject(env, testTask.Project, testTask.Version)
```
ablack12, line 213: should check ok value too

ablack12 (changes requested):
Sorry this review is so late! Feel free to ping me if I've forgotten to review something

PR #4960, opened 2021-08-19
EVG-15294 Support pagination for mainline commit waterfall builds.

khelif96:
[EVG-15294](https://jira.mongodb.org/browse/EVG-15294)

### Description 
This will help support pagination when viewing mainline commits on the waterfall and on the task history pages.
One ask from the mainline commits project is to be able to be linked to  a mainline commit at a specific point of time. Since commits are continuously being added to the front of the waterfall.
 Its difficult to be able to use a normal pagination strategy that depends on offsets and limits based on the amount of content on the screen. Since what might be on page 1 one day might be on page 6 the next. 

Luckily we have a field called `Order` Number which represents the position of a commit relative to the others in the waterfall. Using this we can paginate based off of what commit is currently in view. 
`skipOrderNumber` is a param of the mainline commits query that returns the commits prior to the orderNumber passed in
`nextPageOrderNumber` returns what the last commit in view is so when it is passed into the query for `skipOrderNumber` it will return the next pages worth of commits
`prevPageOrderNumber` returns the order number of the previous pages last active commit. Since mainline commits will always return an active commit as the last commit. 
This value will be nil if there are no newer commits.

### Testing 
Wrote unit tests and tested in staging
Demo of pagination
https://user-images.githubusercontent.com/4605522/130142284-f43aaf0f-8d73-4a3e-9964-0d85b7d76ca3.mov

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -1322,7 +1322,7 @@ func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution 
 			}))
 		}
 
-		if cedarBaseTestResults != nil && len(cedarBaseTestResults) > 0 {
+		if len(cedarBaseTestResults) > 0 {
 			for _, t := range cedarBaseTestResults {
 				baseTestStatusMap[t.TestName] = t.Status
 				baseTestStatusMap[t.DisplayTestName] = t.Status
@@ -1359,7 +1359,7 @@ func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution 
 	var totalTestCount int
 	var filteredTestCount int
 
-	if cedarTestResults != nil && len(cedarTestResults) > 0 {
+	if len(cedarTestResults) > 0 {
 		filteredTestResults, testCount := FilterSortAndPaginateCedarTestResults(FilterSortAndPaginateCedarTestResultsOpts{
 			GroupID:          groupIdParam,
 			Limit:            limitParam,
@@ -2846,6 +2846,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 	})
 	var mainlineCommits MainlineCommits
 	activatedVersionCount := 0
+
+	mostRecentVersion, err := model.GetMostRecentMainlineCommit(projectId)
+
+	if err != nil {
+		// This shouldn't realy happen, but if it does, we should return an error and log it
+		grip.Warning(message.WrapError(err, message.Fields{
+			"message":   "Error getting most recent version",
+			"projectId": projectId,
+		}))
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting most recent mainline commit: %s", err.Error()))
+	}
+
+	// If a commit is the most recent version, we return nil for PrevPageOrderNumber since there aren't any other newer versions
+	if mostRecentVersion.RevisionOrderNumber != versions[0].RevisionOrderNumber {
+		// PrevPageOrderNumber is the order number of the first commit on the screen + the amount of commits on the previous page
+		mainlineCommits.PrevPageOrderNumber = utility.ToIntPtr(versions[0].RevisionOrderNumber + limit)
+	}
+
 	for _, v := range versions {
 		if activatedVersionCount == limit {
 			break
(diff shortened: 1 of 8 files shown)


graphql/resolvers.go
```diff
@@ -2846,6 +2846,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 	})
 	var mainlineCommits MainlineCommits
 	activatedVersionCount := 0
+
+	mostRecentVersion, err := model.GetMostRecentMainlineCommit(projectId)
```
ablack12, line 2850: Can you add a comment about why we need this?

graphql/resolvers.go
```diff
@@ -2846,6 +2846,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+
+	if err != nil {
+		// This shouldn't realy happen, but if it does, we should return an error and log it
```
ablack12, line 2853: really*

graphql/resolvers.go
```diff
@@ -2846,6 +2846,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+		grip.Warning(message.WrapError(err, message.Fields{
+			"message":   "Error getting most recent version",
+			"projectId": projectId,
+		}))
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting most recent mainline commit: %s", err.Error()))
+	}
+
+	// If a commit is the most recent version, we return nil for PrevPageOrderNumber since there aren't any other newer versions
+	if mostRecentVersion.RevisionOrderNumber != versions[0].RevisionOrderNumber {
```
ablack12, line 2862: I imagine it's unlikely for versions to be empty but for defensive programming we should probably check that versions isn't empty somewhere (before most recent versions maybe)

graphql/resolvers.go
```diff
@@ -2846,6 +2846,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+		// PrevPageOrderNumber is the order number of the first commit on the screen + the amount of commits on the previous page
```
ablack12, line 2863: I actually don't really understand _why_ this number is this.  In your description you mention that limit is fickle; why can we use limit to calculate this?

graphql/resolvers.go
```diff
@@ -2856,42 +2874,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+		// If the version was created before we started caching activation status we must manually verify it and cache that value.
 		if v.Activated == nil {
-			defaultSort := []task.TasksSortOrder{
-				{Key: task.DisplayNameKey, Order: 1},
-			}
-			opts := data.TaskFilterOptions{
-				Sorts: defaultSort,
-			}
-			tasks, _, err := r.sc.FindTasksByVersion(v.Id, opts)
+			err = setVersionActivationStatus(r.sc, &v)
```
ablack12, line 2879: nice refactor 👍

graphql/resolvers.go
```diff
@@ -2856,42 +2874,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
-		// into a single MainlineCommitVersion and then append it to our returned list
 		if utility.FromBoolPtr(v.Activated) {
 			activatedVersionCount++
-			mainlineCommits.NextPageOrderNumber = &v.RevisionOrderNumber
+			mainlineCommits.NextPageOrderNumber = utility.ToIntPtr(v.RevisionOrderNumber)
 			mainlineCommitVersion.Version = &apiVersion
 
 		} else {
+			// If a version is not activated we roll up all the unactivated versions that are sequentially near each other
+			// into a single MainlineCommitVersion and then append it to our returned list
```
ablack12, line 2894: nit: use punctuation for multi-sentence comments.

graphql/util.go
```diff
@@ -1307,3 +1307,28 @@ func applyVolumeOptions(ctx context.Context, volume host.Volume, volumeOptions r
...
+	}
+	opts := data.TaskFilterOptions{
+		Sorts: defaultSort,
+	}
+	tasks, _, err := sc.FindTasksByVersion(version.Id, opts)
+	if err != nil {
+		return errors.Wrapf(err, "error getting tasks for version %s", version.Id)
+	}
+	if !task.AnyActiveTasks(tasks) {
+		err = version.SetNotActivated()
```
ablack12, line 1323: can we make these one line? `if err = ... ; err != nil {`

ablack12 (changes requested):
So is PrevPageOrderNumber something that will be used if people want to page to the newer commits?

graphql/util.go
```diff
@@ -1307,3 +1307,26 @@ func applyVolumeOptions(ctx context.Context, volume host.Volume, volumeOptions r
...
+	}
+	opts := data.TaskFilterOptions{
+		Sorts: defaultSort,
+	}
+	tasks, _, err := sc.FindTasksByVersion(version.Id, opts)
+	if err != nil {
+		return errors.Wrapf(err, "error getting tasks for version %s", version.Id)
+	}
+	if !task.AnyActiveTasks(tasks) {
+		if err = version.SetNotActivated(); err != nil {
```
ablack12, line 1323: simplification suggestion: can just return `errors.Wrapf(version.SetNotActivated, "Error updating version activated status for '%s'", version.Id`, i.e. can remove the second Sprintf (handled by Wrapf) and just return regardless of error, since Wrapf will return nil if err is nil. 

Then, we can skip the else if entirely (since everything else falls into this case) and just `return errors.Wrapf(version.SetActivated, ....)`

graphql/resolvers.go
ablack12: some of these comments haven't been addressed

model/version.go
```diff
@@ -373,6 +373,23 @@ func VersionGetHistory(versionId string, N int) ([]Version, error) {
...
+		},
+	}
+	pipeline := []bson.M{{"$match": match}, {"$sort": bson.M{VersionRevisionOrderNumberKey: -1}}, {"$limit": 1}}
+	res := []Version{}
+
+	if err := db.Aggregate(VersionCollection, pipeline, &res); err != nil {
+		return nil, errors.Wrapf(err, "error aggregating versions")
+	}
+
+	return &res[0], nil
```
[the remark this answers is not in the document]
ablack12: ^ still need to address this

PR #4500, opened 2021-03-17
EVG-14263 use mongo driver for test_names aggregation

ablack12:
I'm not sure if this will be faster but we are logging how long the query takes, I can make a graph at some point and see if it improved.

Jonathan recommended using the go driver so that we could cancel if the query takes too long, so we don't get another 71 minute thing. Theoretically that's what SetSocketTimeout was supposed to do, but I couldn't really figure out how anser worked.

PR #4961, opened 2021-08-20
EVG-14028 set maxtime for TaskHistory queries

ybrill:
[EVG-14028](https://jira.mongodb.org/browse/EVG-14028)

### Description 
Now that MaxTime is part of anser I redid the effort from #4500 to use it instead of using the driver directly. 

Another consideration is that we will no longer be passing the request's context to the mongo driver, but rather the application context. On the one hand this is bad because it means that if a call to `iter.Next()` (getMore) is taking a long time the operation won't get killed when the request dies but rather when maxtime passes on the server side (i.e. a minute after it starts). On the other hand, routinely killing an operation on the server by killing the driver's context apparently isn't good for server performance (I wrote more about this in [the anser PR](https://github.com/mongodb/anser/pull/43#issuecomment-897110393)). My judgement is that the combination of maxtime on the server side with checking `ctx.Err()` between calls to Next should be sufficient.

I thought of watching the ctx in a separate goroutine and closing the iterator/cursor when we see the ctx is cancelled. This would have all the advantages: the cursor is closed immediately when the request is over even if Next is blocked on getting a slow batch of results from the server. However, [the driver says](https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo#Cursor) a cursor isn't goroutine safe so I hesitate to do this.

### Testing 
  I'll try out the task_history page in staging when staging becomes available 😄

### DIFF
--- a/model/task_history.go
+++ b/model/task_history.go
@@ -261,8 +261,6 @@ func (iter *taskHistoryIterator) GetChunk(v *Version, numBefore, numAfter int, i
 	}
 	defer session.Close()
 
-	session.SetSocketTimeout(time.Minute)
-
 	versionsBefore, exhausted, err := iter.findAllVersions(v, numBefore, true, include)
 	if err != nil {
 		return chunk, errors.WithStack(err)
@@ -335,7 +333,7 @@ func (iter *taskHistoryIterator) GetChunk(v *Version, numBefore, numAfter int, i
 		{"$group": groupStage},
 		{"$sort": bson.M{task.RevisionOrderNumberKey: -1}},
 	}
-	agg := database.C(task.Collection).Pipe(pipeline)
+	agg := database.C(task.Collection).Pipe(pipeline).MaxTime(time.Minute)
 	var aggregatedTasks []bson.M
 	if err = agg.All(&aggregatedTasks); err != nil {
 		return chunk, errors.WithStack(err)
@@ -357,8 +355,6 @@ func (self *taskHistoryIterator) GetDistinctTestNames(numCommits int) ([]string,
 		return nil, errors.Wrap(err, "problem getting database session")
 	}
 	defer session.Close()
-
-	session.SetSocketTimeout(time.Minute)
 
 	pipeline := mdb.C(task.Collection).Pipe(
 		[]bson.M{
@@ -395,9 +391,9 @@ func (self *taskHistoryIterator) GetDistinctTestNames(numCommits int) ([]string,
 			{"$group": bson.M{"_id": fmt.Sprintf("$%v.%v", testResultsKey, task.TestResultTestFileKey)}},
 		},
 	)
+	pipeline.MaxTime(time.Minute)
 
 	var output []bson.M
-
 	if err = pipeline.All(&output); err != nil {
 		return nil, errors.WithStack(err)
 	}
--- a/model/task_history.go
+++ b/model/task_history.go
@@ -30,6 +30,8 @@ const (
 	// this regex either matches against the exact 'test' string, or
 	// against the 'test' string at the end of some kind of filepath.
 	testMatchRegex = `(\Q%s\E|.*(\\|/)\Q%s\E)$`
+
+	taskHistoryMaxTime = 90 * time.Second
 )
 
 type taskHistoryIterator struct {
@@ -334,7 +336,7 @@ func (iter *taskHistoryIterator) GetChunk(v *Version, numBefore, numAfter int, i
 		{"$group": groupStage},
 		{"$sort": bson.M{task.RevisionOrderNumberKey: -1}},
 	}
-	agg := database.C(task.Collection).Pipe(pipeline).MaxTime(time.Minute)
+	agg := database.C(task.Collection).Pipe(pipeline).MaxTime(taskHistoryMaxTime)
 	var aggregatedTasks []bson.M
 	if err = agg.All(&aggregatedTasks); err != nil {
 		return chunk, errors.WithStack(err)
@@ -392,7 +394,7 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, numCo
 			{"$group": bson.M{"_id": fmt.Sprintf("$%v.%v", testResultsKey, task.TestResultTestFileKey)}},
 		},
 	)
-	pipeline.MaxTime(time.Minute)
+	pipeline.MaxTime(taskHistoryMaxTime)
 
 	var names []string
 	res := bson.M{}
--- a/model/task_history.go
+++ b/model/task_history.go
@@ -400,10 +400,13 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, numCo
 	for iter.Next(&res) {
 		names = append(names, res["_id"].(string))
 		if ctx.Err() != nil {
-			_ = iter.Close()
-			return nil, errors.Wrap(ctx.Err(), "iterating results")
+			catcher := grip.NewBasicCatcher()
+			catcher.Add(errors.Wrap(ctx.Err(), "iterating results"))
+			catcher.Add(iter.Close())
+			return nil, catcher.Resolve()
 		}
 	}
+	sort.Strings(names)
 
 	return names, nil
 }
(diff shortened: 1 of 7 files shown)


model/task_history.go
```diff
@@ -389,35 +391,20 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, env e
...
-			names = append(names, val)
+	pipeline.MaxTime(time.Minute)
+
+	var names []string
+	res := bson.M{}
+	iter := pipeline.Iter()
+	for iter.Next(&res) {
+		names = append(names, res["_id"].(string))
+		if ctx.Err() != nil {
+			_ = iter.Close()
```
Kimchelly, line 403: Is this supposed to ignore the error? The typical way to deal with closing resources during an error case is to accumulate them in a grip catcher.

model/task_history.go
```diff
@@ -389,35 +391,20 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, env e
...
-		select {
-		case <-ctx.Done():
-			return nil, errors.New("context cancelled")
-		default:
-			res := bson.M{}
-			if err := cursor.Decode(&res); err != nil {
-				return nil, errors.Wrapf(err, "error decoding result")
-			}
-			// remove quotes from value
-			val := strings.Replace(res["_id"].(string), "\"", "", -1)
```
Kimchelly, line 413: Is this string replacement not necessary anymore?

model/task_history.go
```diff
@@ -389,35 +391,20 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, env e
...
+		names = append(names, res["_id"].(string))
+		if ctx.Err() != nil {
+			_ = iter.Close()
+			return nil, errors.Wrap(ctx.Err(), "iterating results")
 		}
 	}
-	if err := cursor.Err(); err != nil {
-		return nil, errors.Wrap(err, "error reading cursor")
-	}
-	sort.Strings(names)
```
Kimchelly, line 420: Same question about this sort.

model/task_history.go
ybrill: I'm not sure. It was added in #4500. When I reverted that commit this went away too. I didn't put it back because although it wasn't clear what necessitated it there it was clear that it was [a result of the changes there](https://github.com/evergreen-ci/evergreen/pull/4500/files#r597719059).

model/task_history.go
ybrill: This was another instance of the same thing: the changes from #4500 made them always come back in a different order, [for some reason](https://github.com/evergreen-ci/evergreen/pull/4500#issuecomment-802322659). When I reverted it the sort was removed too.

When I get a hold of staging I'll see if this is still an issue.

ybrill: ah, I'm wrong: Annie said it happened before too.
I'll add it back

model/task_history.go
ybrill: that's an excellent idea.
done

model/task_history.go
ybrill: done

model/task_history.go
```diff
@@ -337,7 +334,7 @@ func (iter *taskHistoryIterator) GetChunk(v *Version, numBefore, numAfter int, i
 		{"$group": groupStage},
 		{"$sort": bson.M{task.RevisionOrderNumberKey: -1}},
 	}
-	agg := database.C(task.Collection).Pipe(pipeline)
+	agg := database.C(task.Collection).Pipe(pipeline).MaxTime(time.Minute)
```
ablack12, line 337: In the last PR we purposefully changed this to be larger, re some advice from John:
> I would set this slightly larger than the load balancer timeout (90 secs?) so that we don't accidentally kill something that might have finished

model/task_history.go
```diff
@@ -389,35 +391,23 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, env e
...
+
+	var names []string
+	res := bson.M{}
+	iter := pipeline.Iter()
+	for iter.Next(&res) {
+		names = append(names, res["_id"].(string))
+		if ctx.Err() != nil {
+			catcher := grip.NewBasicCatcher()
+			catcher.Add(errors.Wrap(ctx.Err(), "iterating results"))
+			catcher.Add(iter.Close())
```
ablack12, line 405: are we able to just defer iter.Close() ? I imagine we should do this in the non error case too

model/task_history.go
ablack12: This was just a legitimate issue with how MongoDB was returning the test results; they were all being printed as 
` "\"TestName"\"` or something when I checked with staging. Testing in staging (when you can haha) will determine if this is still necessary.

ablack12 (changes requested):
Can we change this PR title to include an imperative verb? Looking back at this in the future, it won't be very clear what's going on.

model/task_history.go
ybrill: done

model/task_history.go
ybrill: This is an interesting question.

iter.Close() passes through to [close the cursor](https://github.com/mongodb/anser/blob/7c3b6870adfb65e15f5b130bae0adbb7efc67718/db/wrapper.go#L342). In the happy path where where the cursor gets exhausted I think the cursor gets closed automatically by the server. It was hard to find anything in the mongodb docs 😞  (I did find [this](https://docs.mongodb.com/manual/core/tailable-cursors/#:~:text=by%20default%2C%20mongodb%20will%20automatically%20close%20a%20cursor%20when%20the%20client%20has%20exhausted%20all%20results%20in%20the%20cursor)), and the internet wasn't terribly helpful, but I found [this](https://jira.mongodb.org/browse/SERVER-8968?focusedCommentId=298093&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-298093) from a while ago. I determined it empirically on our db, though
```
mci:SECONDARY> var cur = db.tasks.find().limit(1000);
mci:SECONDARY> cur.isClosed()
false
mci:SECONDARY> cur.forEach((res)=>{});
mci:SECONDARY> cur.isClosed()
true
```

ybrill:
> Can we change this PR title to include an imperative verb? Looking back at this in the future, it won't be very clear what's going on.

done

model/task_history.go
ybrill: I did eventually get to test this in staging. It was a little tricky because we already switched test results over to cedar and the query is limited to the last 10 commits, but it worked out by matching on a date in the past.
It returned this for the `test-service` task:
```
[
  "TestActivateVersion",
  "TestAssignNextAvailableTaskWithDispatcherSettingsVersionLegacy",
  "TestAssignNextAvailableTaskWithDispatcherSettingsVersionTunable",
  "TestCedar",
  "TestCheckHostHealth",
  "TestCheckHostWrapper",
  "TestConsistentTaskAssignment",
  "TestGetBuildInfo",
  "TestGetBuildStatus",
  "TestGetDisplayTaskInfo",
  "TestGetDockerfile",
  "TestGetHostDNS",
  "TestGetHostFromCache",
  "TestGetRecentVersions",
  "TestGetTaskInfo",
  "TestGetTaskStatus",
  "TestGetVariantsWithCommand",
  "TestGetVersionInfo",
  "TestGetVersionInfoViaRevision",
  "TestGetVersionStatus",
  "TestModifyHostStatusWithUpdateStatus",
  "TestNextTask",
  "TestPatchListModulesEndPoints",
  "TestProjectRoutes",
  "TestServiceStatusEndPoints",
  "TestStuckHostEndpoints",
  "TestTaskHistory",
  "TestTaskHistory/TestGetVersionsInWindow",
  "TestTaskHistory/TestSurroundingVersions",
  "TestTaskHistory/TestSurroundingVersionsSort",
  "TestTaskLifecycleEndpoints",
  "TestValidateTaskEndDetails",
  "TestVerifyAliasExists"
]
```
which is what I'd expect.

ablack12 (approved):

Kimchelly (approved):

PR #4963, opened 2021-08-20
EVG-15054 add functions to attach/detach repo

ablack12:
[EVG-15054](https://jira.mongodb.org/browse/EVG-15054)

### Description 
I think the most intuitive behavior for  DetachToRepo would be to set everything that currently defaults to the repo value.
If we don't do this, we risk putting the project in a bad state (for example, if PRTestingEnabled is set to true in the project, but PR aliases were defaulted to repo, then we have an invalid state if we don't copy the PR aliases over). Additionally, if we reattach, values we thought were set to nil (because they were showing the empty value) will suddenly be defaulting to repo.

### Testing 
So much unit testing 🙃


model/repo_ref.go
```diff
@@ -74,49 +50,18 @@ func (r *RepoRef) Insert() error {
 }
 
 // Upsert updates the project ref in the db if an entry already exists,
-// overwriting the existing ref. If no project ref exists, one is created
+// overwriting the existing ref. If no project ref exists, one is created.
+// Ensures that fields that aren't relevant to repos aren't set.
 func (r *RepoRef) Upsert() error {
```
ablack12, line 55: This is a slightly unrelated change but, I realized how prone this is to bugs. The only reason I handled all of the sets individually is because I didn't want the fields that didn't relate to repos to be set; however there are _much_ fewer of them so it makes sense to handle them separately instead.

model/project_ref.go
```diff
@@ -420,6 +446,96 @@ func (p *ProjectRef) AddToRepoScope(user *user.DBUser) error {
...
+	for _, a := range before.Aliases {
+		if utility.StringSliceContains(evergreen.InternalAliases, a.Alias) {
+			hasInternalAliases[a.Alias] = true
+		} else { // if it's not an internal alias, it's a patch alias. Only add repo patch aliases if no patch aliases exist for the project.
+			hasPatchAlias = true
+		}
+	}
+	repoAliasesToCopy := []ProjectAlias{}
+	for _, internalAlias := range evergreen.InternalAliases {
+		// if the branch doesn't have the internal alias set, add any that exist for the repo
```
malikchaya2, line 514: I may be wrong about this, but I believe it is possible to have a mixed state where a project uses both repo level and project level aliases, and if there are two with the same name, then the project one will take precedence. If that is the case, do we want to maybe copy over all of them unless it's overwritten?

model/project_ref.go
```diff
@@ -420,6 +446,96 @@ func (p *ProjectRef) AddToRepoScope(user *user.DBUser) error {
 	return nil
 }
 
+// DetachFromRepo removes the branch from the relevant repo scopes, and updates the project to not point to the repo.
+// Any values that previously defaulted to repo will have the repo value explicitly set.
+func (p *ProjectRef) DetachFromRepo(u *user.DBUser) error {
```
malikchaya2, line 451: Do we need to do something about notifications?

model/project_ref.go
```diff
@@ -420,6 +446,96 @@ func (p *ProjectRef) AddToRepoScope(user *user.DBUser) error {
...
+
+	// catch any resulting errors so that we log before returning
+	catcher := grip.NewBasicCatcher()
+	if mergedVars != nil {
+		_, err = mergedVars.Upsert()
+		catcher.Wrap(err, "error saving merged vars")
+	}
+
+	if len(before.Subscriptions) == 0 {
+		// Save repo subscriptions as project subscriptions if none exist
```
malikchaya2, line 488: Why do we only do this if none exist? Can we have a situation where a project relies on both repo level and project level subscriptions?

model/project_ref.go
ablack12: yep! subscriptions == notifications

model/project_ref.go
ablack12: nope; the project either defaults to repo subscriptions, or defines its own (this is only different for variables)

model/project_ref.go
ablack12: I think based on https://github.com/evergreen-ci/evergreen/blob/3c58dd6911ad37d81323cc4d3216d6d52e030788/model/project_aliases.go#L104 that's actually technically true 🤔   but shouldn't be based on the UI design (it's supposed to be that either you default to repo, or you define your own, which is more generally consistent), so I think you've just found a bug! We may want to add some fields to ProjectRef to indicate whether or not we should be defaulting instead of doing it automatically if no aliases exist; this seems ticket-sized / will affect other things so I'm going to leave this as is but create a ticket. https://jira.mongodb.org/browse/EVG-15307

malikchaya2: Does this mean that if someone defaults aliases to repo, we don't want to let them define additional aliases that are specific to the project, and they will need to add it to the repo if they want to be able to use it in their project?

malikchaya2 (approved):

model/project_ref.go
malikchaya2: discussed offline

PR #4965, opened 2021-08-21
EVG-14955 Update evaluate to return the merged project

bynn:
[EVG-14955](https://jira.mongodb.org/browse/EVG-14955)

### Description 
added type to opts

### Testing 
local test
image

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -559,7 +559,6 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 // If reading from a version config, LoadProjectForVersion should be used to persist the resulting parser project.
 // opts is used to look up files on github if the main parser project has an Include.
 func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
-	catcher := grip.NewBasicCatcher()
 	intermediateProject, err := createIntermediateProject(data)
 	if err != nil {
 		return nil, errors.Wrapf(err, LoadProjectError)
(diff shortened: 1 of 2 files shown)


model/project_parser.go
```diff
@@ -558,6 +559,7 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 // If reading from a version config, LoadProjectForVersion should be used to persist the resulting parser project.
 // opts is used to look up files on github if the main parser project has an Include.
 func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
+	catcher := grip.NewBasicCatcher()
```
ablack12, line 562: Are you using a catcher so that you can still attempt to hit Translate? I think translate will fail if any of the above things went wrong, so I'm not sure it's worth doing that. It seems like enough to just return the parser project at whatever state it's at, with the error.

model/project_parser.go
```diff
@@ -585,43 +590,58 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
 	if p != nil {
 		*project = *p
 	}
+	catcher.Wrapf(err, LoadProjectError)
 	project.Identifier = identifier
-	return intermediateProject, errors.Wrap(err, LoadProjectError)
+	return intermediateProject, catcher.Resolve()
 }
 
+const (
```
ablack12, line 598: This is a good strategy! I think it's worth being more descriptive though, so it's clear looking back what's happening here. "Type" should maybe be called something more descriptive (RequestSource? ReadFileFrom? Not actually sure) or add a comment to Type to explain (also add a comment before the constants themselves).

ablack12 (changes requested):

model/project_parser.go
bynn: i changed it to ReadFileFrom

model/project_parser.go
```diff
@@ -567,15 +568,15 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
 		opts.RemotePath = path.FileName
 		yaml, err := retrieveFile(ctx, opts)
 		if err != nil {
-			return nil, errors.Wrapf(err, LoadProjectError)
+			return intermediateProject, errors.Wrapf(err, LoadProjectError)
```
ablack12, line 571: Actually sorry, can you add a comment at the top about why we return intermediateProject as it is, even in error? (because we may want to see how far we got with the merging process before we ran into an issue) This behavior is non-standard so it's good for future engineers to know that this was a purposeful choice.

model/project_parser.go
ablack12: Can we update these constants to make more sense for that? ReadFileFrom "mainline" doesn't make as much sense as "github" (local still also probably makes sense, i'm not actually sure what Patch should look like yet but we can push that off until the relevant ticket)

ablack12 (commented):
You'll be able to test this using evergreen-local (`make build` so that you have an updated evergreen binary in `clients/darwin_amd64/evergreen `, pass in an evergreen local configuration file with -c, and then call `evaluate` with an evergreen yaml that has an include; let me know if you need help setting that up

model/project_parser.go
bynn: done

model/project_parser.go
bynn: done

ablack12 (approved):
Sorry for the delayed review -- feel free to ping me about things!

PR #4966, opened 2021-08-21
EVG-15166: Remove premature return statement from stats iterators

hadjri:
[EVG-15166](https://jira.mongodb.org/browse/EVG-15166)

### Description 
Currently When we're failing to process test stats it's preventing us from processing task stats. Since we return early when we hit an error in the hourly stats, we never actually get to processing daily task results. In one observed case when there was a test name long enough that one of these indexes exceeds the WT 1024 byte limit, and this will continue to cause an error until Evergreen upgrades to Mongo 4.4 from 4.0.  A change has been made to add logging to notify if a certain test stat has failed to process and continue processing the rest of the stats instead of stopping and returning once a single error is found.
### Testing 
Deployed change to staging and enabled the cache historical statistics, and saw that there was not any [unexpected behavior](index="evergreen-staging" "Generating daily test stats") in the logs and that task stats were [properly generated](https://evergreen-staging.corp.mongodb.com/rest/v2/projects/mci/test_stats?after_date=2021-06-01&before_date=2021-08-22&tasks=test-agent) still.

### DIFF
--- a/units/cache_historical_test_data.go
+++ b/units/cache_historical_test_data.go
@@ -225,8 +225,8 @@ func (c *cacheHistoricalJobContext) updateHourlyAndDailyStats(ctx context.Contex
 		})
 		if err != nil {
 			grip.Warning(message.WrapError(err, message.Fields{
+				"message":    "error iterating over hourly stats",
 				"project_id": c.ProjectID,
-				"job_time":   c.JobTime,
 			}))
 		}
 	}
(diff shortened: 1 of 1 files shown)


bsamek (approved):
I hadn't read this code before. The one concern I have is that it may mean we do more work. Prior to this change, we would not have executed https://github.com/evergreen-ci/evergreen/blob/e29a626d90903b9f65f7e87fd8a7db7526f986f8/units/cache_historical_test_data.go#L149. This change causes us to run https://github.com/evergreen-ci/evergreen/blob/e29a626d90903b9f65f7e87fd8a7db7526f986f8/units/cache_historical_test_data.go#L150, which we do want, but also `GenerateDailyTestStatsFromHourly`, which presumably we don't want to run if `GenerateHourlyTestStats` fails. However, reading the code I can't identify a specific risk, so I think this should be fine. I can keep an eye on db performance after deploying the change.

PR #4968, opened 2021-08-23
EVG-15286: Schedule downstream patches

sophstad:
[EVG-15286](https://jira.mongodb.org/browse/EVG-15286)

### Description 
- `SchedulePatch` resolver accepts list of patch trigger aliases to run from UI
- Call `ProcessTriggerAliases` from `SchedulePatch` to create intents for child patches
- Update parent patch document with child patch IDs

### Testing 
- Add unit tests and tested on staging


model/patch_lifecycle.go
```diff
@@ -446,8 +446,13 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		if err = tasksToInsert.InsertUnordered(sessCtx); err != nil {
 			return nil, errors.Wrapf(err, "error inserting tasks for version '%s'", patchVersion.Id)
 		}
+		if p.IsParent() {
```
malikchaya2, line 449: Is there a way to add a check(s) to this conditional to increase our chances of this not being done if there are no new aliases being added?

graphql/util.go
```diff
@@ -408,6 +408,14 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
 			return errors.Wrap(err, "Error setting patch variants and tasks"), http.StatusInternalServerError, "", ""
 		}
 
+		projectRef, err := model.FindOneProjectRef(project.Identifier)
```
ablack12, line 411: need to check if project ref is nil also (although we also do this in the `if` case, so probably we should just factor this out so it can be used in both places rather than repeating code)

graphql/util.go
```diff
@@ -408,6 +408,14 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
...
+		if err != nil {
+			return errors.Wrap(err, "unable to find project ref"), http.StatusInternalServerError, "", ""
+		}
+		if err := units.ProcessTriggerAliases(ctx, p, projectRef, evergreen.GetEnvironment(), patchTriggerAliases); err != nil {
```
ablack12, line 415: I'm not sure it's a great idea to call this here. ProcessTriggerAliases creates and runs jobs right away, which I don't think should be done as part of graphql. Should ProcessTriggerAliases itself be a job, or do we need this to run before FinalizePatch (given some of the comments in that function)? In which case I think we should add some comments here to make it clear why we need to do this in the moment. cc @malikchaya2

model/patch/patch.go
```diff
@@ -578,6 +578,18 @@ func (p *Patch) SetActivated(ctx context.Context, versionId string) error {
 	return err
 }
 
+// SetChildPatches appends the IDs of downstream patches to the db
+func (p *Patch) SetChildPatches(ctx context.Context) error {
+	triggersKey := bsonutil.GetDottedKeyName(TriggersKey, TriggerInfoChildPatchesKey)
+	_, err := evergreen.GetEnvironment().DB().Collection(Collection).UpdateOne(ctx,
```
ablack12, line 584: This is a long form syntax that we don't typically use; UpdateOne in this package is better (see the below function for an example)

model/patch/patch.go
```diff
@@ -578,6 +578,18 @@ func (p *Patch) SetActivated(ctx context.Context, versionId string) error {
...
+		bson.M{IdKey: p.Id},
+		bson.M{
+			"$push": bson.M{triggersKey: bson.M{"$each": p.Triggers.ChildPatches}},
```
ablack12, line 587: I would use $addToSet just to verify we don't race / double-add patches

ablack12 (changes requested):

graphql/util.go
malikchaya2: This can't be done in finishPatch where it was being done until now (for the CLI). However, I think we may be able to move this into the if p.IsParent() {} block in finalize patch.

sophstad: Yes definitely worth pointing this out and I'll add some commentst! For context, Chaya and I met with Jonathan and determined that it pretty much is necessary to create and run child patch jobs in one sweep. When downstream tasks are created with `--triggerAlias` with the CLI, the intent can be created immediately and then the job can be run later. This code supports a different flow, though, where a patch has been created without the flag and a user would like to add downstream tasks using the UI. In that case, I think it's pretty much necessary to immediately create the job and run it. We acknowledged that this may result in some slowness, but the CLI remains available for a speedier option.

Let me know if this makes sense or if I misunderstood! Also if you have any other thoughts about relocating this I'm all ears.

sophstad: Oh sorry, just remembered my thinking here -- I didn't call `ProcessTriggerAliases` in `FinalizePatch` because `finishPatch` (which is invoked by the CLI) calls `FinalizePatch`. This seemed to me like the easiest way to avoid a duplicate operation.

ablack12: It might be worth duplicating anyway; can a user change the aliases they passed in from the CLI?
I'm fine with not moving it into FinalizePatch if that's too messy; I'd like to still add comments here though in that case.

sophstad: No, I believe we decided that for the current scope the UI will only allow adding new downstream tasks. Modifying CLI-specified options could be added later if we hear from users. I think I'll keep the location for now, comments incoming 👍 thanks!

model/patch_lifecycle.go
sophstad: Updated `FinalizePatch` to accept a list of the child patch IDs created by `ProcessTriggerAlias`; this list can be used to check if the document should be updated.

model/patch_lifecycle.go
```diff
@@ -446,8 +446,13 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		if err = tasksToInsert.InsertUnordered(sessCtx); err != nil {
 			return nil, errors.Wrapf(err, "error inserting tasks for version '%s'", patchVersion.Id)
 		}
+		if p.IsParent() && len(childPatchIds) > 0 {
+			if err = p.SetChildPatches(); err != nil {
```
ablack12, line 450: Sorry I didn't ask about this earlier but: these patches are created from running ProcessTriggerAliases so there's not really a reason we need to do this within the transaction / that it's worth passing through just for that. Can we do this before calling ProcessTriggerAliases in the graphql function instead, if childPatchIds length > 0?

ablack12 (changes requested):

graphql/util.go
```diff
@@ -311,7 +311,7 @@ func hasEnqueuePatchPermission(u *user.DBUser, existingPatch *restModel.APIPatch
 
 // SchedulePatch schedules a patch. It returns an error and an HTTP status code. In the case of
 // success, it also returns a success message and a version ID.
-func SchedulePatch(ctx context.Context, patchId string, version *model.Version, patchUpdateReq PatchVariantsTasksRequest, parametersModel []*restModel.APIParameter) (error, int, string, string) {
+func SchedulePatch(ctx context.Context, patchId string, version *model.Version, patchUpdateReq PatchVariantsTasksRequest, parametersModel []*restModel.APIParameter, patchTriggerAliases []string) (error, int, string, string) {
```
khelif96, line 314: nit. Can we refactor this to ingest an options object. This now has 6 different parameters and it is a little unwieldy/error prone if you mix up the order of a passed in param.

ablack12 (approved):

malikchaya2 (approved):

PR #4970, opened 2021-08-23
EVG-15278 Return parentPatch if available

malikchaya2:
[EVG-15278](https://jira.mongodb.org/browse/EVG-15278)

### Description 
The original bug reported in the ticket was actually not really a bug. The version resolver accidentally left the parent patch status out of the calculation, so the patch status always showed the child status _only_ instead of a collective status. This mimicked the patch not waiting on the parent, but it wasn't actually the case. Once the resolver was fixed, this "bug" went away. However, in my investigation I noticed something that either has hidden problems or is bound to cause a problem at some point, so I fixed it here. 

### Testing 
I noticed that the parent patch was null in splunk logs where I didn't expect it to be, and after this change, it no longer showed up as null in that case.

### DIFF
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -173,7 +173,7 @@ func checkPatchStatus(p *patch.Patch) (bool, *patch.Patch, bool, error) {
 	// make sure the parent is done, if not, wait for the parent
 	if p.IsChild() {
 		if !evergreen.IsFinishedPatchStatus(parentPatch.Status) {
-			return isReady, nil, false, nil
+			return isReady, parentPatch, false, nil
 		}
 	}
 	childrenStatus, err := getChildrenOrSiblingsReadiness(childrenOrSiblings)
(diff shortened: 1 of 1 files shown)


trigger/patch.go
```diff
@@ -173,7 +173,7 @@ func checkPatchStatus(p *patch.Patch) (bool, *patch.Patch, bool, error) {
 	// make sure the parent is done, if not, wait for the parent
 	if p.IsChild() {
 		if !evergreen.IsFinishedPatchStatus(parentPatch.Status) {
-			return isReady, nil, false, nil
+			return isReady, parentPatch, false, nil
```
ablack12, line 176: Should we also do this for 170, 181, 184?

malikchaya2: probably a good idea. done. thanks!

ablack12 (approved):

PR #4972, opened 2021-08-23
EVG-14767: Support regex when allowing CORS

hadjri:
[EVG-14767](https://jira.mongodb.org/browse/EVG-14767)

### Description 
A change has been made to allow regexes when adding CORS, because currently we only apply if a request url exactly matches what we have in the allowedOrigins defined on the admin page.  Now exact strings and regexes will both match and apply CORS.

### Testing 
Added unit test certifying behavior

### DIFF
--- a/util/strings.go
+++ b/util/strings.go
@@ -94,3 +94,18 @@ func CoalesceString(in ...string) string {
 func CoalesceStrings(inArray []string, inStrs ...string) string {
 	return CoalesceString(CoalesceString(inArray...), CoalesceString(inStrs...))
 }
+
+// StringContainsSliceRegex determines if a string matches a regex in a slice
+func StringContainsSliceRegex(slice []string, item string) bool {
+	if len(slice) == 0 {
+		return false
+	}
+
+	for idx := range slice {
+		matched, err := regexp.MatchString(slice[idx], item)
+		if err == nil && matched {
+			return true
+		}
+	}
+	return false
+}
--- a/util/strings.go
+++ b/util/strings.go
@@ -97,12 +97,8 @@ func CoalesceStrings(inArray []string, inStrs ...string) string {
 
 // StringContainsSliceRegex determines if a string matches a regex in a slice
 func StringContainsSliceRegex(slice []string, item string) bool {
-	if len(slice) == 0 {
-		return false
-	}
-
-	for idx := range slice {
-		matched, err := regexp.MatchString(slice[idx], item)
+	for _, sliceItem := range slice {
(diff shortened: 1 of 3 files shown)


util/strings_test.go
```diff
@@ -38,3 +38,10 @@ func TestCoalesce(t *testing.T) {
 	assert.Equal("one", CoalesceStrings(nil, "", "one", "two"))
 	assert.Equal("", CoalesceStrings(nil, "", ""))
 }
+
+func TestStringContainsSliceRegex(t *testing.T) {
+	assert := assert.New(t)
+	domains := []string{".*.corp.mongodb.com", "https://something.mongodb.com"}
+	assert.Equal(true, StringContainsSliceRegex(domains, "https://patch-analysis-ui.server-tig.staging.corp.mongodb.com"))
+	assert.Equal(true, StringContainsSliceRegex(domains, "https://something.mongodb.com"))
```
ablack12, line 46: Can we add some fall cases to this test?

util/strings_test.go
```diff
@@ -38,3 +38,10 @@ func TestCoalesce(t *testing.T) {
 	assert.Equal("one", CoalesceStrings(nil, "", "one", "two"))
 	assert.Equal("", CoalesceStrings(nil, "", ""))
 }
+
+func TestStringContainsSliceRegex(t *testing.T) {
+	assert := assert.New(t)
```
ablack12, line 43: Adding assert to the top of a function is actually an outdated practice; can you just pass in t for each of the asserts below?

util/strings.go
```diff
@@ -94,3 +94,18 @@ func CoalesceString(in ...string) string {
 func CoalesceStrings(inArray []string, inStrs ...string) string {
 	return CoalesceString(CoalesceString(inArray...), CoalesceString(inStrs...))
 }
+
+// StringContainsSliceRegex determines if a string matches a regex in a slice
+func StringContainsSliceRegex(slice []string, item string) bool {
+	if len(slice) == 0 {
```
ablack12, line 100: we don't really need to check this; it's implied by 110 since we just skip the for loop

util/strings.go
```diff
@@ -94,3 +94,18 @@ func CoalesceString(in ...string) string {
...
+		return false
+	}
+
+	for idx := range slice {
```
ablack12, line 104: you can use the notation `idx, sliceItem := range slice {` and then you can use `sliceItem` instead of `slice[idx]` (the only time we don't want to do this is if you're making modifications to slice)

ablack12 (changes requested):

PR #4973, opened 2021-08-23
EVG-15302: pass service flags to helper functions

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15302

### Description 
Pass the service flags into the helper functions that depend on them to avoid having dependencies on global testing state in unit tests. It also avoids the issue where arbitrary functions can ask for the global environment state.

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -301,7 +301,14 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 		ec2Settings.UserData = expanded
 	}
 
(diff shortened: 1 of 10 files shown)


julianedwards (commented):
Why do we only care that the service flags aren't stale? Shouldn't we call `evergreen.GetConfig()` to have all the latest settings?

Kimchelly:
The standard way to do things in Evergreen is to always use the latest service flags because they usually are for critical operations (e.g. the service is down), but always use the cached admin settings for everything else because you might not want the effects to take place immediately (e.g. for backward/forward compatibility reasons). Using the cached settings also makes it easier to do admin settings changes/migrations before the app server is actually rolled out so that when it's rolled out, it uses the new data correctly.

julianedwards (approved):

PR #4977, opened 2021-08-24
EVG-15222 allow immutable project ID to be user specified

ablack12:
[EVG-15222](https://jira.mongodb.org/browse/EVG-15222)

### Description 
Originally we didn't allow this to be user-specified _because_ it is immutable; once an ID is taken it can never be used as an identifier for a different project, even if the identifier for this project is changed. In theory this was fine since we transitioned IDs to be evergreen internal and identifiers to be used externally; however, there are some dev prod teams that also rely on Evergreen internals and this has caused problems. They would like their projects to be able to specify something immutable because they don't have any use case for changing it. 

Although it's a risk to allow users to do something that they might want to change later and can't, this was already the behavior of identifier, and our users are engineers so they should understand the significance of an immutable project ID.

I recognize that this modal could be prettier, but since we're not supporting this page soon I don't think it's worth the time.
Screen Shot 2021-08-24 at 11 49 27 AM

### Testing 
Tested with local that I could specify my immutable ID:
Screen Shot 2021-08-24 at 11 49 10 AM

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -351,7 +351,18 @@ func (projectRef *ProjectRef) Insert() error {
 }
 
 func (p *ProjectRef) Add(creator *user.DBUser) error {
-	p.Id = mgobson.NewObjectId().Hex()
+	if p.Id != "" {
+		// verify that this is a unique ID
+		conflictingRef, err := FindOneProjectRef(p.Id)
+		if err != nil {
+			return errors.Wrap(err, "error checking for conflicting project ref")
+		}
+		if conflictingRef != nil {
+			return errors.New("ID already being used as ID or identifier for another project")
+		}
+	} else {
+		p.Id = mgobson.NewObjectId().Hex()
+	}
 
 	// if a hidden project exists for this configuration, use that ID
 	if p.Owner != "" && p.Repo != "" && p.Branch != "" {
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -1378,16 +1378,34 @@ mciModule.controller(
 mciModule.directive("adminNewProject", function () {
   return {
     restrict: "E",
-    template: '<div class="row">' +
-      '<div class="col-lg-12">' +
-      "Enter project name " +
-      '<form style="display: inline" ng-submit="addProject()">' +
-      '<input type="text" id="project-name" placeholder="project name" ng-model="newProject.identifier">' +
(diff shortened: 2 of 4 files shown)


model/project_ref.go
```diff
@@ -351,7 +351,18 @@ func (projectRef *ProjectRef) Insert() error {
 }
 
 func (p *ProjectRef) Add(creator *user.DBUser) error {
-	p.Id = mgobson.NewObjectId().Hex()
+	if p.Id != "" {
```
malikchaya2, line 354: Do we want to check that it's in a certain format? Or is any length/combination of letter, numbers, symbols, etc, fine?

public/static/js/projects.js
```diff
@@ -1378,16 +1378,34 @@ mciModule.controller(
...
+    template: '' +
+      '<div class="row">' +
+        '<div class="col-lg-12">' +
+         "Enter project identifier " +
+          '<input type="text" id="project-name" placeholder="project name" ng-model="newProject.identifier">' +
+        "</div>" +
+      '</div>' +
+      '<div class="row">' +
+        '<div class="col-lg-12">' +
+          "Optionally enter immutable project ID " +
```
malikchaya2, line 1390: Do we want to maybe add a warning to only do this if they have a specific reason to?

public/static/js/projects.js
```diff
@@ -1378,16 +1378,34 @@ mciModule.controller(
...
-      "Enter project name " +
-      '<form style="display: inline" ng-submit="addProject()">' +
-      '<input type="text" id="project-name" placeholder="project name" ng-model="newProject.identifier">' +
-      ' <input type="checkbox" id="copy-project" ng-model="newProject.copyProject">' +
-      " Duplicate current project" +
-      "</form>" +
-      '<button type="submit" class="btn btn-primary" style="float: right; margin-left: 10px;" ng-click="addProject()">Create Project</button>' +
-      "</div>" +
-      "</div>",
+    template: '' +
```
malikchaya2, line 1381: How does this change the scope/design of Projects Track Multiple Branches? Do we need additional tickets or design additions?

public/static/js/projects.js
ablack12: Yeah i was hoping that "Evergreen internally" was strong enough but it's probably worth being more clear

PR #4981, opened 2021-08-25
EVG-14956 Verify that PatchedConfig stores the merged config

bynn:
[EVG-14956](https://jira.mongodb.org/browse/EVG-14956)

### Description 
read from correct source for patches for creating and finalizing patches

### Testing 
make patch on staging
https://spruce-staging.corp.mongodb.com/version/61268bf29dbe324c960ed003/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

cli command
https://spruce-staging.corp.mongodb.com/version/612847a697b1d30357e29332/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

change is only on non-main yaml
https://spruce-staging.corp.mongodb.com/version/612e935f97b1d3077001e0f7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


model/patch_lifecycle.go
```diff
@@ -280,7 +282,26 @@ func MakePatchedConfig(ctx context.Context, env evergreen.Environment, p *patch.
 func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, githubOauthToken string) (*Version, error) {
 	// unmarshal the project YAML for storage
 	project := &Project{}
-	opts := GetProjectOpts{}
+	projectRef, err := FindOneProjectRef(p.Project)
```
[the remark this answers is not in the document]
ablack12: Looks good but I think the opts construction should still be moved to a helper function, possibly a method on patch like GetLoadProjectOptsForPatch (seems identical to the opts construction in GetPatchedProject)

model/patch_lifecycle.go
```diff
@@ -173,16 +176,33 @@ func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken str
 
 	// apply remote configuration patch if needed
 	if !(p.IsGithubPRPatch() || p.IsPRMergePatch()) && p.ConfigChanged(path) {
+		opts.ReadFileFrom = ReadFromPatchDiff
```
ablack12, line 179: right now, p.ConfigChanged() only checks if the main yaml was changed (in the case you tested in staging, evergreen.yml). What if _only_ an include file is changed? (ex.yml) 

I think what LoadProjectInto could do is in the loop `for _, path := range intermediateProject.Include`, check if `p.ConfigChanged(path)` (i.e. check it again for each of the included filenames), and if so change ReadFileFrom to ReadFromPatchDiff

model/project_parser.go
```diff
@@ -640,6 +668,35 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 	}
 }
 
+func GetFileFromPatchDiff(ctx context.Context, opts GetProjectOpts) (string, error) {
```
ablack12, line 671: It looks like this is a subset of GetPatchedProject. Can we refactor GetPatchedProject to use GetFileFromPatchDiff?

ablack12: Also, I'm not sure this function needs to be public: are all the calls from within the model package? If we do keep it public, we need a function comment.

ablack12 (changes requested):
This is looking really good!

model/project_parser.go
bynn: done

model/task_lifecycle.go
```diff
@@ -844,7 +844,7 @@ func getBuildStatus(buildTasks []task.Task) string {
 
 	// finished but failed
 	for _, t := range buildTasks {
-		if evergreen.IsFailedTaskStatus(t.Status) {
+		if evergreen.IsFailedTaskStatus(t.Status) || t.Aborted {
```
ablack12, line 847: Is this related to your other ticket?

model/patch_lifecycle.go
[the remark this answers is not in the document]
ablack12: Did this not get committed? I'm still seeing FindOne

model/project_parser.go
```diff
@@ -567,6 +567,11 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
 	// return intermediateProject even if we run into issues to show merge progress
 	for _, path := range intermediateProject.Include {
 		opts.RemotePath = path.FileName
+		if opts.PatchOpts.patch != nil {
```
ablack12, line 570: can combine these if statements with &&

ablack12: Should also add a comment to explain why we're doing this. (And in the else case, we should change it to ReadFromPatch, so that with multiple includes we switch back if necessary)

model/project_parser.go
```diff
@@ -640,6 +673,35 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 	}
 }
 
+func getFileFromPatchDiff(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
```
ablack12, line 676: This function name isn't quite right; we _aren't_ getting the file from patch diff here; we're getting it from github, and then only erroring if it's a problem that we don't have it. getFileForPatch is maybe more accurate, or something like that

model/project_parser.go
```diff
@@ -616,6 +628,27 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 			return nil, errors.Wrap(err, "error reading project config")
 		}
 		return fileContents, nil
+	case ReadFromPatch:
+		configFile, err := thirdparty.GetGithubFile(ctx, opts.Token, opts.Ref.Owner,
```
ablack12, line 632: Can this section also be using GetFileFromPatchDiff?

ablack12 (changes requested):

model/project_parser.go
bynn: done

model/project_parser.go
bynn: yes

model/project_parser.go
bynn: i think we would want to limit that behavior only if it's reading from a patch

model/project_parser.go
```diff
@@ -567,10 +567,13 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
...
 	for _, path := range intermediateProject.Include {
 		opts.RemotePath = path.FileName
-		if opts.PatchOpts.patch != nil {
-			if opts.PatchOpts.patch.ConfigChanged(path.FileName) {
+		// read from the patch diff if a change has been made in any of the include files
+		if opts.ReadFileFrom == ReadFromPatch || opts.ReadFileFrom == ReadFromPatchDiff {
+			if opts.PatchOpts.patch != nil && opts.PatchOpts.patch.ConfigChanged(path.FileName) {
 				opts.ReadFileFrom = ReadFromPatchDiff
 			}
+		} else {
```
ablack12, line 575: I think you want this else to line up with the if statement at 572; here you're setting ReadFromLocal to ReadFromPatch, for example

ablack12 (changes requested):

model/project_parser.go
bynn: yes

ablack12 (approved):
🚀

PR #4982, opened 2021-08-26
EVG-15274 add display task ID to existing execution tasks

ablack12:
[EVG-EVG-15274](https://jira.mongodb.org/browse/EVG-15274)

### Description 
When I made this display task ID change, I didn't think to update tasks that already exist but are only now being made into execution tasks. But also I decided to just refactor this section (which per Julian's retro he mentioned we shouldn't do whoops 😅  but it's a small ish refactor; I'll call out the actual change for the fix in a comment)

### Testing 
Added a unit test. I'm honestly having trouble replicating this bug on staging; I'm not sure if it's the way the tasks are generated on the server project or what. I do suspect this will fix the problem, I've read over it a few times and it does seem like _a_ bug if not _the_ bug so I think it's worth deploying and seeing what happens, but let me know what you think.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -899,7 +899,7 @@ func createTasksForBuild(project *Project, buildVariant *BuildVariant, b *build.
 		}
 
 		// update existing exec tasks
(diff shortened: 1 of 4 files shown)


malikchaya2 (approved):
I don't think I have enough context to understand what exactly this fixes, but nothing stands out as being problematic :)

PR #4986, opened 2021-08-30
EVG-15345 Add childVersions to the Version resolver

malikchaya2:
[EVG-15345](https://jira.mongodb.org/browse/EVG-15345)

### Description 
This adds childVersions to the version resolver 

### Testing 
Tested on the graphql playground on staging.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3099,6 +3099,7 @@ func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVer
 	if len(childPatchIds) > 0 {
 		childVersions := []*restModel.APIVersion{}
 		for _, cp := range childPatchIds {
+			// this calls the graphql Version query resolver
 			v, err := r.Query().Version(ctx, cp)
 			if err != nil {
 				//before erroring due to the version being nil or not found,
@@ -3113,10 +3114,10 @@ func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVer
 						Message:    fmt.Sprintf("patch with id %s not found", cp),
 					}
 				}
-				if p.Version == "" {
-					return nil, nil
+				if p.Version != "" {
+					//only return the error if the version is activated (and we therefore expect it to be there)
+					return nil, InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch child version: %s ", err.Error()))
 				}
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch child version: %s ", err.Error()))
 			}
 			childVersions = append(childVersions, v)
 		}
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3088,6 +3088,28 @@ func (r *versionResolver) Patch(ctx context.Context, v *restModel.APIVersion) (*
 	return apiPatch, nil
 }
 
+func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVersion) ([]*restModel.APIVersion, error) {
+	if !evergreen.IsPatchRequester(*v.Requester) {
+		return nil, nil
+	}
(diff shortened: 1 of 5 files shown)


graphql/resolvers.go
```diff
@@ -3088,6 +3088,28 @@ func (r *versionResolver) Patch(ctx context.Context, v *restModel.APIVersion) (*
...
+		return nil, nil
+	}
+	childPatchIds, err := r.sc.GetChildPatchIds(*v.Id)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Couldn't find a patch with id: `%s` %s", *v.Id, err.Error()))
+	}
+	if len(childPatchIds) > 0 {
+		childVersions := []*restModel.APIVersion{}
+		for _, cp := range childPatchIds {
+			v, err := r.Query().Version(ctx, cp)
```
khelif96, line 3102: Can we use `r.sc.FindVersionById(cp)` here instead. 
Just as an edge case check, Are we sure all child patches here will have a corresponding version? i.e all of the child patches are  activated..

khelif96 (approved):
Approved with one comment

graphql/resolvers.go
malikchaya2: > Can we use r.sc.FindVersionById(cp) here instead.
discussed over slack and decided against it 

> Just as an edge case check, Are we sure all child patches here will have a corresponding version? i.e all of the child patches are activated..

done

graphql/resolvers.go
```diff
@@ -3101,6 +3101,21 @@ func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVer
 		for _, cp := range childPatchIds {
 			v, err := r.Query().Version(ctx, cp)
```
khelif96, line 3102: nit. Can we add a comment indicating this calls the gql resolver and not a db query.

graphql/resolvers.go
```diff
@@ -3101,6 +3101,21 @@ func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVer
...
 			if err != nil {
+				//before erroring due to the version being nil or not found,
+				// fetch the child patch to see if it's activated
+				p, err := patch.FindOneId(cp)
+				if err != nil {
+					return nil, err
+				}
+				if p == nil {
+					return nil, gimlet.ErrorResponse{
```
khelif96, line 3111: This should be a graphql resolver error not a gimlet error.

graphql/resolvers.go
```diff
@@ -3101,6 +3101,21 @@ func (r *versionResolver) ChildVersions(ctx context.Context, v *restModel.APIVer
...
+						StatusCode: http.StatusNotFound,
+						Message:    fmt.Sprintf("patch with id %s not found", cp),
+					}
+				}
+				if p.Version == "" {
```
khelif96, line 3116: This runs in a for loop. Why do we return nil here? If only one child patch is unactivated this would return nil for all of them.

khelif96 (changes requested):
This is on the right track. Since a version could have many child patches that are different states. We should probably handle that. If one child patch is `created` and unactivated (Doesn't have a corresponding version) and another child patch does have a version. This would return nil for all of them and not resolve any of the child patches.

graphql/resolvers.go
malikchaya2: shoots! you're right, thanks for spotting that!

graphql/resolvers.go
malikchaya2: thanks! copy pasta, lol

graphql/resolvers.go
malikchaya2: done

graphql/resolvers.go
malikchaya2: done

khelif96 (approved):

PR #4987, opened 2021-08-30
EVG-15353: Use ConverToOld when converting test results

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15353

Not using this function caused a bug in the UI because we were not setting the `TaskID` field.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2106,8 +2106,8 @@ func (t *Task) populateNewTestResults() error {
 	if err != nil {
 		return errors.Wrap(err, "finding test results")
 	}
-	for _, result := range newTestResults {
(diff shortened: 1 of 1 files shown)


ybrill (approved):

ybrill:
It occurred to me to point out that we're passing in a pointer to a loop's iterator variable (which is getting reset between iterations), but that it works here because ConvertToOld just copies out the values.
It's not clear to me why the parameter was a pointer in the first place.

ybrill (approved):
I had been thinking ConvertToOld could take a TestResult value, but this works too.
I'm good either way.

julianedwards:
Good catch!

PR #4988, opened 2021-08-31
EVG-15338 print saving project error correctly

ablack12:
[EVG-15338](https://jira.mongodb.org/browse/EVG-15338)

### Description 
Kind of a bandaid, but fixing the backend to always return the error in the same format is more complicated, and this is soon-to-be legacy anyway.

### Testing
Tested with evergreen local.

### DIFF
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -618,7 +618,11 @@ mciModule.controller(
             $scope.isDirty = false;
           },
           function (resp) {
-            $scope.saveMessage = "Couldn't save project: " + resp.data.error;
+              var message = resp.data;
+            if (resp.data.error !== undefined) {
+                message = resp.data.error;
+            }
+            $scope.saveMessage = "Couldn't save project: " + message;
           }
         );
     };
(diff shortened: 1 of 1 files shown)


public/static/js/projects.js
```diff
@@ -618,7 +618,11 @@ mciModule.controller(
             $scope.isDirty = false;
           },
           function (resp) {
-            $scope.saveMessage = "Couldn't save project: " + resp.data.error;
+              var message = resp.data;
+            if (resp.data.error !== undefined) {
+                message = resp.data.error;
+            }
```
arjunpatel100, line 624:
```suggestion
            const message = resp.data.error ?? resp.data;
```

arjunpatel100 (approved):

malikchaya2 (approved):
Approving, but also suggest implementing Arjun's suggested change.

public/static/js/projects.js
ablack12: this is so awesome

PR #4990, opened 2021-08-31
EVG-15357 Include display name for build variants

khelif96:
[EVG-15357](https://jira.mongodb.org/browse/EVG-15357)

### Description 
Add display names when fetching buildvariant names for the task history query
Also renamed the queries to be more reflective of what they do. This is safe to do in a non backwards compatible way since its not used by anything yet.
Tries to fetch the latest display names for each build variant in order to avoid returning duplicate build variants that may have had their names changed.


model/task/db.go
```diff
@@ -815,42 +820,37 @@ func FindUniqueBuildVariantNamesByTask(projectId string, taskName string) ([]str
...
-	unwindAndSort := []bson.M{
-		{
-			"$unwind": "$build_variants",
-		},
-		{
-			"$sort": bson.M{
-				"build_variants": 1,
-			},
+
+	project := bson.M{
```
ablack12, line 828: Are we sure we don't want to still unwind? It's possible for build variants to change display names; if multiple build variants match a display name, is it okay that we're only keeping the first one?

model/task/db.go
```diff
@@ -803,8 +803,13 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
 	return result, nil
 }
 
-// FindUniqueBuildVariantNamesByTask returns  a list of unique build variants names for a given task name
-func FindUniqueBuildVariantNamesByTask(projectId string, taskName string) ([]string, error) {
+type TaskHistoryBuildVariantHeader struct {
```
ablack12, line 806: Why Header?

model/task_history_headers_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/db"
+	"github.com/evergreen-ci/evergreen/model/build"
+	"github.com/evergreen-ci/evergreen/model/task"
+	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFindUniqueBuildVariantNamesByTask(t *testing.T) {
+	Convey("Should return unique build variants for tasks", t, func() {
```
ablack12, line 15: Convey is actually an outdated way of writing go tests; we do stick to asserts whenever possible.
If you want to  split this test into subtests, I suggest this method: https://github.com/evergreen-ci/evergreen/blob/4829a262cd4a60ae42b6a547b2c5fbaca5fb2bff/model/generate_test.go#L921

ablack12 (changes requested):

model/task/db.go
```diff
@@ -803,8 +803,13 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
+	BuildVariant string `bson:"build_variant"`
+	DisplayName  string `bson:"display_name"`
+}
+
+// FindUniqueBuildVariantNamesByTask returns  a list of unique build variants names and their display names for a given task name
```
arjunpatel100, line 811: nit: extra space between `returns  a`

model/task/db.go
```diff
@@ -803,54 +803,83 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
 		},
 	}
 
-	pipeline = append(pipeline, group)
-	pipeline = append(pipeline, unwindAndSort...)
+	pipeline = append(pipeline, groupByBuildVariant)
 
-	type taskBuildVariants struct {
-		BuildVariants []string `bson:"build_variants"`
+	// // reorganize the results to get the build variant names and a corresponding build id
```
ablack12, line 844: double comment?

model/task/db.go
```diff
@@ -803,54 +803,83 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
-			"$sort": bson.M{
-				"build_variants": 1,
+
+	pipeline = append(pipeline, sortByOrderNumber)
+
+	// group the build variants by unique build variant names and get a build id for each
+	groupByBuildVariant := bson.M{
+		"$group": bson.M{
+			"_id": bson.M{
+				"build_variant": "$" + BuildVariantKey,
```
ablack12, line 834: here we use "build_variant" but at 848 we use BuildVariantKey, can we make these consistent?

ablack12: (Does it work to set _id to $BuildVariantKey directly?)

ablack12 (approved):
A couple of comments but overall LGTM!

model/task/db.go
khelif96: I don't think so. I tried and the query would fail i think there needs to be an expression instead of just a value.

arjunpatel100 (approved):

PR #4992, opened 2021-09-01
EVG-15213: Reference a project’s configuration when interacting with perf plugin configs

hadjri:
[EVG-15213](https://jira.mongodb.org/browse/EVG-15213)

### Description 
Currently all perf plugin configs are derived from the admin config. Change has been made to introduce the option of getting perf plugin settings from the project or admin config depending on the state of the PluginAdminPageDisabled service flag. Since the admin page settings simply displays a list of projects for which to enable the perf plugin, the new property on the individual project configs is a boolean to toggle whether or not the perf plugin should be enabled.


graphql/resolvers.go
```diff
@@ -2628,19 +2628,37 @@ func (r *taskResolver) GeneratedByName(ctx context.Context, obj *restModel.APITa
...
 	if err != nil {
 		return false, err
 	}
-	if perfPluginSettings, exists := evergreen.GetEnvironment().Settings().Plugins[perfPlugin.Name()]; exists {
-		err := mapstructure.Decode(perfPluginSettings, &perfPlugin)
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(*obj.ProjectId, -1)
+		if err == nil && lastGoodVersion != nil {
+			parserProject, err := model.ParserProjectFindOneById(lastGoodVersion.Id)
+			if err == nil && parserProject != nil && *parserProject.PerfEnabled {
```
ablack12, line 2639: If PerfEnabled is explicitly set to false, we should return false, not check the ProjectRef. 
Can just return `utility.FromBoolPtr(parserProject.PerfEnabled)` in the not nil case (we typically prefer this over explicitly dereferencing in the function; Resolvers get away with it because of how they can require obj fields to be populated, and here you _did_ do a nil check but it's still not generally preferred)

graphql/resolvers.go
```diff
@@ -2628,19 +2628,37 @@ func (r *taskResolver) GeneratedByName(ctx context.Context, obj *restModel.APITa
...
+				return true, nil
+			}
+		}
+		project, err := model.FindMergedProjectRef(*obj.ProjectId)
```
ablack12, line 2643: can use FindProjectById as well (although it should use `true` here since we do want the merged one)

plugin/perf.go
```diff
@@ -62,3 +64,26 @@ func (pp *PerfPlugin) GetPanelConfig() (*PanelConfig, error) {
...
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return false
+	}
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectRef.Id, -1)
+		if err == nil && lastGoodVersion != nil {
+			parserProject, err := model.ParserProjectFindOneById(lastGoodVersion.Id)
+			if err == nil && parserProject != nil {
+				return *parserProject.PerfEnabled
```
ablack12, line 78: utility.FromBoolPtr is preferable

plugin/perf.go
```diff
@@ -62,3 +64,26 @@ func (pp *PerfPlugin) GetPanelConfig() (*PanelConfig, error) {
...
+func isPerfEnabled(projectRef model.ProjectRef, projects []string) bool {
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return false
+	}
+	if flags.PluginAdminPageDisabled {
+		lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectRef.Id, -1)
+		if err == nil && lastGoodVersion != nil {
+			parserProject, err := model.ParserProjectFindOneById(lastGoodVersion.Id)
+			if err == nil && parserProject != nil {
```
ablack12, line 77: I believe we only want to return early if parserProject.PerfEnabled is specified, right? same to the one that's called in the task resolver.
is there a way to make this a helper function that could be used both here and in the resolver?

ablack12 (changes requested):

ablack12 (approved):
LGTM with function move

PR #4993, opened 2021-09-02
EVG-14963: Add host create details when host is externally terminated

hadjri:
[EVG-14963](https://jira.mongodb.org/browse/EVG-14963)

### Description 
Currently hosts that are spawned by a task that are terminated externally cause host.list to timeout since they are not included in the host details property on a task, and when an external termination occurs the terminated host is not included in the list of hosts associated with a given task.  A change has been made to add a host's details to a given task when it is terminated externally.

### DIFF
--- a/units/host_monitoring_external_termination.go
+++ b/units/host_monitoring_external_termination.go
@@ -138,7 +138,7 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
 		if h.RunningTask != "" {
-			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, ctx.Err()); err != nil {
+			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, errors.New("error adding host create error details")); err != nil {
 				grip.Error(message.WrapError(err, message.Fields{
 					"message":      "error adding host create error details",
 					"cloud_status": cloudStatus.String(),
--- a/units/host_monitoring_external_termination.go
+++ b/units/host_monitoring_external_termination.go
@@ -137,8 +137,8 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		if cloudStatus != cloud.StatusTerminated && (h.UserHost || h.StartedBy != evergreen.User) {
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
-		if h.RunningTask != "" {
-			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, errors.New("host was externally terminated")); err != nil {
+		if h.SpawnOptions.SpawnedByTask {
+			if err := task.AddHostCreateDetails(h.SpawnOptions.TaskID, h.Id, errors.New("host was externally terminated")); err != nil {
 				grip.Error(message.WrapError(err, message.Fields{
 					"message":      "error adding host create error details",
 					"cloud_status": cloudStatus.String(),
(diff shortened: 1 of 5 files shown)


units/host_monitoring_external_termination.go
```diff
@@ -136,10 +137,19 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		if cloudStatus != cloud.StatusTerminated && (h.UserHost || h.StartedBy != evergreen.User) {
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
-
+		if h.RunningTask != "" {
+			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, ctx.Err()); err != nil {
```
ablack12, line 141: It's really likely that there isn't an error in context (so this will noop): I think here instead we're going to want to pass in our own error that tells the user that the host was terminated externally

ablack12 (changes requested):

units/host_monitoring_external_termination.go
```diff
@@ -136,10 +137,19 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		if cloudStatus != cloud.StatusTerminated && (h.UserHost || h.StartedBy != evergreen.User) {
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
-
+		if h.RunningTask != "" {
+			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, errors.New("error adding host create error details")); err != nil {
```
ablack12, line 141: Sorry, I think this still isn't right:
The error you're passing into AddHostCreateDetails is the error we'll be returning to the user. The error we want to tell them isn't that we couldn't add host create details (although logging this to grip is reasonable); the thing we _want_ to tell them about is that the host was terminated externally.

units/host_status.go
```diff
@@ -186,6 +187,11 @@ func (j *cloudHostReadyJob) setCloudHostStatus(ctx context.Context, m cloud.Mana
 		event.LogHostTerminatedExternally(h.Id, h.Status)
 
 		catcher := grip.NewBasicCatcher()
+		if h.RunningTask != "" {
+			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, j.Error()); err != nil {
```
ablack12, line 191: same idea here: there's likely no error in the job that we want to log (or even if there is, it's not what we want to tell the user). We want to tell the user that the host has been terminated externally.

ablack12 (changes requested):

units/host_monitoring_external_termination.go
hadjri: Ah apologies, had mistaken this error for the error we want to show if we fail to update the details themselves.  Fixed.

units/host_monitoring_external_termination.go
```diff
@@ -136,10 +137,19 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		if cloudStatus != cloud.StatusTerminated && (h.UserHost || h.StartedBy != evergreen.User) {
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
-
+		if h.RunningTask != "" {
+			if err := task.AddHostCreateDetails(h.RunningTask, h.Id, errors.New("host was externally terminated")); err != nil {
```
ablack12, line 141: Just realized: h.RunningTask is set for the host that's running the task; we don't want the host that is _running_ the task but rather the host that was _created by_ the task (same below). Looking at what we query for when looking for uphosts, I think we instead want to use `h.SpawnOptions.TaskId` ( and we can check h.SpawnOptions.SpawnedByTask)

ablack12 (approved):

PR #4994, opened 2021-09-03
EVG-14927: Treat aborted tasks as unscheduled

hadjri:
[EVG-14927](https://jira.mongodb.org/browse/EVG-14927)

### Description 
Currently when a new PR commit comes in before the previous patch finishes, the previous commit's tests in evergreen will get aborted if they are still running. This sends a Slack notification saying the patch has failed although this is not necessary. 
 Change has been made to not treat these as failures to avoid any confusion.

### Testing
Changes were deployed to staging and a patch was run and manually aborted, and a Slack notification was placed on it to myself.  Previously we would expect to see a Slack message generated and the status on the task to be "failed", but this time when the task was aborted the status was changed to "unsipatched" [here](https://evergreen-staging.corp.mongodb.com/api/rest/v2/tasks/evg_ubuntu1604_test_model_event_patch_c6ffa29e590ee1ce9137bbd6827cc2767feff144_611289ad9dbe3250b56347ec_21_08_10_14_14_50?execution=4) and the task did not fail and indicated an [abort](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_test_model_event_patch_c6ffa29e590ee1ce9137bbd6827cc2767feff144_611289ad9dbe3250b56347ec_21_08_10_14_14_50/logs?execution=4) in the logs instead of a failure.


ablack12 (commented):
I think this isn't how we want to approach this ticket. The task _should_ be failed; this is changing a deeper Evergreen behavior that is likely to confuse people and mess up scripts that rely on this behavior. In the comment I made on the ticket, I said that we should consider aborts unscheduled _for notifications_, which may be where this idea came from. Rather what I meant was that at the place where we decide whether or not to send a notification (https://github.com/evergreen-ci/evergreen/blob/e49d6b501d32f12093d24aa790278d527e5c1f69/trigger/task.go#L447) we should consider whether or not the task was aborted. 

We should also add this handling to version failure. That one is a little trickier: we may need to iterate through all tasks for the version, and if all failed tasks are also aborted, then don't send the notification.

trigger/version.go
```diff
@@ -193,7 +193,23 @@ func (t *versionTriggers) versionGithubCheckOutcome(sub *event.Subscription) (*n
...
+		if versionTask.Status == evergreen.TaskFailed {
+			if versionTask.Aborted {
+				skipNotification = true
+			} else {
+				skipNotification = false
+				break
+			}
+		}
+	}
+	if t.data.Status != evergreen.VersionFailed || skipNotification {
```
ablack12, line 212: We should do this before 196; if the version isn't failed, we already know we don't have an aborted task

trigger/version.go
```diff
@@ -193,7 +193,23 @@ func (t *versionTriggers) versionGithubCheckOutcome(sub *event.Subscription) (*n
...
+	versionTasks, err := task.FindAll(task.ByVersion(t.version.Id))
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting tasks in version")
+	}
+	skipNotification := false
+	// If the only failures in the version come from aborted tasks then we skip sending a notification
+	for _, versionTask := range versionTasks {
+		if versionTask.Status == evergreen.TaskFailed {
+			if versionTask.Aborted {
+				skipNotification = true
```
ablack12, line 205: We don't need to store a skipNotification variable; we can just return here immediately if we hit this case

ablack12 (changes requested):

trigger/version.go
```diff
@@ -193,7 +193,23 @@ func (t *versionTriggers) versionGithubCheckOutcome(sub *event.Subscription) (*n
 }
 
 func (t *versionTriggers) versionFailure(sub *event.Subscription) (*notification.Notification, error) {
-	if t.data.Status != evergreen.VersionFailed {
+	versionTasks, err := task.FindAll(task.ByVersion(t.version.Id))
```
ablack12, line 196: If we query by Version ID and Aborted == true, then we can use task.Count() and just check if count > 0, rather than iterating through all tasks (should put this query in the task/db.go file)

ablack12 (changes requested):

trigger/version.go
hadjri: I think we need this variable to ensure all failures are aborted tasks.  If we find a task that is failed and not aborted, then a notification may still need to be sent.

trigger/version.go
ablack12: Given your comment on the above thread we actually can't do this (alas), disregard

hadjri: Sure I think this would work if we also check that the number of aborted tasks equals the number of failed tasks.

ablack12: True, although this would require two queries which I don't think is an improvement. We could filter by version and the failed status though to limit how many tasks we have to iterate through.

model/task/db.go
```diff
@@ -345,6 +345,22 @@ func ByVersion(version string) db.Q {
...
+		VersionKey: version,
+		AbortedKey: true,
+	})
+}
+
+// FailedTasksByVersion produces a query that returns all failed tasks for the given version.
+func FailedTasksByVersion(version string) db.Q {
+	return db.Query(bson.M{
+		VersionKey: version,
+		StatusKey:  evergreen.TaskFailed,
```
ablack12, line 360: Should use $in evergreen.FailureTaskStatuses (and we should write a small unit test for at least one of these, with Count).

ablack12 (changes requested):

trigger/version.go
```diff
@@ -196,6 +196,22 @@ func (t *versionTriggers) versionFailure(sub *event.Subscription) (*notification
 	if t.data.Status != evergreen.VersionFailed {
 		return nil, nil
 	}
+	failedTasks, err := task.FindAll(task.FailedTasksByVersion(t.version.Id))
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting failed tasks in version")
+	}
+	skipNofitication := false
```
ablack12, line 203: nit: notification 😁

ablack12 (approved):
LGTM with spelling

PR #4995, opened 2021-09-03
EVG-15047 add get project settings resolver

ablack12:
[EVG-15047](https://jira.mongodb.org/browse/EVG-15047)

### Description 
Wrote this so that it could be used for projects and for repos (the only difference is which collection we look up the initial project from). Decided to just always look up the project since we'd have to do that anyway to verify we're using the internal ID. 
Made it so we have to check both collections to know if it's project or repo, but we could change this to take in an optional boolean from the UI if we already know if it's repo or not.

### Testing 
Many tests; much playground.


gqlgen.yml
```diff
@@ -96,11 +96,64 @@ models:
...
+  ProjectVars:
+    model: github.com/evergreen-ci/evergreen/rest/model.APIProjectVars
+    fields:
+      privateVars:
+        resolver: true
+  ProjectAlias:
+    model: github.com/evergreen-ci/evergreen/rest/model.APIProjectAlias
+  ProjectSubscription:
+    model: github.com/evergreen-ci/evergreen/rest/model.APISubscription
+  SubscriberWrapper:
```
ablack12, line 140: I think I used a wrapper because if I put the resolver: true on `target` it seemed like `target` wasn't going to be populated correctly, and it's necessary to figure out the individual subscriber; I'm not sure if that's true though or if I just got too deep.

graphql/tests/projectSettings/queries/project-settings-subscriptions.graphql
```diff
@@ -0,0 +1,49 @@
+{
```
malikchaya2, line 1: I think you will also need to add data to https://github.com/evergreen-ci/evergreen/blob/main/graphql/tests/task/results.json to complete these tests

gqlgen.yml
malikchaya2: Would it work if we call it something like ProjectSubscriber?

graphql/resolvers.go
```diff
@@ -291,6 +306,125 @@ func (r *projectResolver) IsFavorite(ctx context.Context, at *restModel.APIProje
...
+	return hook != nil, nil
+}
+
+func (r *projectSettingsResolver) Vars(ctx context.Context, a *restModel.APIProjectSettings) (*restModel.APIProjectVars, error) {
+	vars, err := model.FindOneProjectVars(utility.FromStringPtr(a.ProjectRef.Id))
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error finding project vars for '%s': %s", *a.ProjectRef.Id, err.Error()))
+	}
+	if vars == nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("vars for '%s' don't exist", *a.ProjectRef.Id))
```
malikchaya2, line 324: Should this maybe be ResourceNotFound instead of InternalServerError?

graphql/resolvers.go
```diff
@@ -291,6 +306,125 @@ func (r *projectResolver) IsFavorite(ctx context.Context, at *restModel.APIProje
...
+	}
+	return hook != nil, nil
+}
+
+func (r *projectSettingsResolver) Vars(ctx context.Context, a *restModel.APIProjectSettings) (*restModel.APIProjectVars, error) {
+	vars, err := model.FindOneProjectVars(utility.FromStringPtr(a.ProjectRef.Id))
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error finding project vars for '%s': %s", *a.ProjectRef.Id, err.Error()))
+	}
+	if vars == nil {
```
malikchaya2, line 323: Do we expect all projects to have vars?

graphql/resolvers.go
```diff
@@ -291,6 +306,125 @@ func (r *projectResolver) IsFavorite(ctx context.Context, at *restModel.APIProje
...
+	aliases, err := model.FindAliasesForProject(utility.FromStringPtr(a.ProjectRef.Id))
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error finding aliases for project: %s", err.Error()))
+	}
+	res := []*restModel.APIProjectAlias{}
+	for _, alias := range aliases {
+		apiAlias := restModel.APIProjectAlias{}
+		if err = apiAlias.BuildFromService(alias); err != nil {
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("problem building APIPProjectAlias %s from service: %s",
+				alias.ID.Hex(), err.Error()))
```
malikchaya2, line 343: would alias.alias be more easily identifiable by a user than alias.ID?

graphql/resolvers.go
```diff
@@ -1063,6 +1226,11 @@ func (r *projectResolver) Patches(ctx context.Context, obj *restModel.APIProject
 	return &Patches{Patches: patchPointers, FilteredPatchCount: *count}, nil
 }
 
+//
```
malikchaya2, line 1229: leftover comments

graphql/resolvers.go
```diff
@@ -291,6 +306,125 @@ func (r *projectResolver) IsFavorite(ctx context.Context, at *restModel.APIProje
...
+		}
+		res.JiraIssueSubscriber = sub
+	case event.JIRACommentSubscriberType:
+		res.JiraCommentSubscriber = a.Target.(*string)
+	case event.EmailSubscriberType:
+		res.EmailSubscriber = a.Target.(*string)
+	case event.SlackSubscriberType:
+		res.SlackSubscriber = a.Target.(*string)
+	case event.EnqueuePatchSubscriberType:
+		// do nothing
```
malikchaya2, line 420: Why do we want to do nothing in this case?

graphql/resolvers.go
```diff
@@ -800,6 +934,33 @@ func (r *queryResolver) MyHosts(ctx context.Context) ([]*restModel.APIHost, erro
 	return apiHosts, nil
 }
 
+func (r *queryResolver) ProjectSettings(ctx context.Context, identifier string) (*restModel.APIProjectSettings, error) {
+	// first, check project collection
+	res := &restModel.APIProjectSettings{}
+	projectRef, err := model.FindOneProjectRef(identifier)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error looking in project collection: %s", err.Error()))
+	}
```
malikchaya2, line 943: nit: can we maybe add a 
// then, check the repo collection 
here to close the loop?

graphql/resolvers.go
```diff
@@ -800,6 +934,33 @@ func (r *queryResolver) MyHosts(ctx context.Context) ([]*restModel.APIHost, erro
...
+		}
+	}
+	if projectRef == nil {
+		return nil, ResourceNotFound.Send(ctx, "project/repo doesn't exist")
+	}
+	apiProjectRef := restModel.APIProjectRef{}
+	if err = apiProjectRef.BuildFromService(projectRef); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error building APIProjectRef from service: %s", err.Error()))
+	}
+	res.ProjectRef = apiProjectRef
```
malikchaya2, line 960: Ah, I like this setup!

gqlgen.yml
khelif96: Yeah i think it would make sense to name it ProjectSubscriber

graphql/resolvers.go
```diff
@@ -291,6 +306,125 @@ func (r *projectResolver) IsFavorite(ctx context.Context, at *restModel.APIProje
...
+				alias.Alias, err.Error()))
+		}
+		res = append(res, &apiAlias)
+	}
+	return res, nil
+}
+
+func (r *projectVarsResolver) PrivateVars(ctx context.Context, a *restModel.APIProjectVars) ([]*string, error) {
+	res := []*string{}
+	for privateAlias, isPrivate := range a.PrivateVars {
```
khelif96, line 352: Just out of curiosity wouldn't isPrivate be true for everything inside a variable named PrivateVars?

graphql/resolvers.go
```diff
@@ -972,15 +1133,17 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
 	return *obj.Id, nil
 }
 
-func (r *patchResolver) PatchTriggerAliases(ctx context.Context, obj *restModel.APIPatch) ([]*PatchTriggerAlias, error) {
+func (r *patchResolver) PatchTriggerAliases(ctx context.Context, obj *restModel.APIPatch) ([]*restModel.APIPatchTriggerDefinition, error) {
```
khelif96, line 1136: Nice catch matching this to an Api struct!

graphql/schema.graphql
```diff
@@ -797,15 +801,185 @@ type GroupedProjects {
...
+
+type ProjectSubscription {
+  id: String!
+  resourceType: String!
+  trigger: String!
+  selectors: [Selector]!
+  regexSelectors: [Selector]!
+  subscriber: SubscriberWrapper
+  ownerType: String!
+  triggerData: StringMap
```
khelif96, line 836: Just to double check triggerData is an unpredictable  free form map right? Which is why we have to use to StringMap.

graphql/schema.graphql
```diff
@@ -797,15 +801,185 @@ type GroupedProjects {
...
+  variantTags: [String!]!
+  taskTags: [String!]!
+}
+
+type ProjectSubscription {
+  id: String!
+  resourceType: String!
+  trigger: String!
+  selectors: [Selector]!
+  regexSelectors: [Selector]!
```
khelif96, line 833: Lets make Selector non nullable as well if the array is non nullable
`[Selector!]!`

khelif96 (approved):
Just had a few questions. But LGTM as long as @malikchaya2  suggestion to rename `SubscriptionWrapper` is also resolved

graphql/resolvers.go
ablack12: Theoretically, but it's possible that when updating this variable in the backend we would set `PrivateVars[varName] = false` to unset something. Just in case this seemed the most careful.

graphql/schema.graphql
ablack12: triggerData is a map[string]string which I assumed is what StringMap was for. Is this right?

ablack12: https://github.com/evergreen-ci/evergreen/blob/68fd43a436971e7c8df60fdaa322e2f27c8e1c28/model/event/subscriptions.go#L86

graphql/resolvers.go
sophstad: +1, thanks! 😄

model/github_hook.go
```diff
@@ -52,7 +52,7 @@ func (h *GithubHook) Remove() error {
 
 func FindGithubHook(owner, repo string) (*GithubHook, error) {
 	if len(owner) == 0 || len(repo) == 0 {
-		return nil, errors.New("Owner and repository must not be empty strings")
+		return nil, nil
```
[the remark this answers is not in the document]
malikchaya2: that makes sense, thanks!

graphql/resolvers.go
malikchaya2: that makes sense!

malikchaya2 (approved):

sophstad (approved):

graphql/schema.graphql
khelif96: Yup makes sense 👍

PR #4997, opened 2021-09-03
EVG-15279 use facet to conditionally lookup annotations

ablack12:
[EVG-15279 ](https://jira.mongodb.org/browse/EVG-15279 )

### Description 
Use facet so we don't have to lookup annotations for successful tasks.

### Testing 
Added a test that succeeds before and after changes.


khelif96 (approved):
Awesome Thanks for doing this!

PR #4999, opened 2021-09-06
EVG-15373 Edit restart version to take in child patches

malikchaya2:
[EVG-15373](https://jira.mongodb.org/browse/EVG-15373)

### Description 
Edit restartVersion to also restart childVersions 

### Testing 
Tested by restarting a some tasks and child tasks on staging 

### Spruce PR 
https://github.com/evergreen-ci/spruce/pull/868


graphql/util.go
```diff
@@ -674,13 +674,13 @@ type VersionModifications struct {
 	Active   bool                      `json:"active"`
 	Abort    bool                      `json:"abort"`
 	Priority int64                     `json:"priority"`
-	TaskIds  []string                  `json:"task_ids"`
+	TaskIds  []*model.TaskToRestart    `json:"task_ids"`
 }
 
 func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectRef, modifications VersionModifications) (int, error) {
 	switch modifications.Action {
 	case Restart:
```
arjunpatel100, line 682: This switch case is called into from the Legacy UI Restart Modal and changing the type for TaskIds will break it: https://github.com/evergreen-ci/evergreen/blob/main/public/static/js/version_admin.js#L50 

In order to retain backwards compatibility with Spruce and the Legacy UI, it's probably best to make an entirely new switch statement here called `RestartVersions` and remove `case Restart` once the Legacy UI and Spruce UI no longer rely on it.

arjunpatel100 (changes requested):

graphql/resolvers.go
```diff
@@ -1994,14 +1994,14 @@ func (r *mutationResolver) ScheduleUndispatchedBaseTasks(ctx context.Context, pa
...
-func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abort bool, taskIds []string) (*string, error) {
+func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abort bool, taskIds []*model.TaskToRestart) (*string, error) {
 	if len(taskIds) == 0 {
 		return nil, InputValidationError.Send(ctx, "`taskIds` array is empty. You must provide at least one task id")
 	}
 	modifications := VersionModifications{
 		Action:  Restart,
 		Abort:   abort,
-		TaskIds: taskIds,
+		TaskIds: taskIds, //change this to be able to take in the new type
```
ablack12, line 2004: Is this an open TODO?

graphql/schema.graphql
```diff
@@ -143,6 +143,12 @@ type Mutation {
   clearMySubscriptions: Int!
 }
 
+input TaskToRestart {
```
ablack12, line 146: Should this be PatchesToRestart instead of TaskToRestart, since each TaskToRestart object corresponds to one patch rather than one task? Or something like that; I hesitate to commit to this naming because it's not immediately clear. For example, RestartVersions passes in one version ID but then iterates through taskIds. This is bound to cause confusion in the future. Also taskIds is confusing as a variable name for the object, since there's also an inner taskIds.

model/lifecycle.go
```diff
@@ -373,6 +380,18 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 		return errors.Wrap(err, "unable to find version")
 	}
 	return errors.Wrap(version.UpdateStatus(evergreen.VersionStarted), "unable to change version status")
+
+}
+
+// RestartVersions restarts completed tasks associated with a set of versionId.
```
ablack12, line 386: this comment also will need modifying, since the restart isn't related to the version IDs but rather the set of patches

ablack12 (changes requested):

graphql/schema.graphql
```diff
@@ -143,6 +143,12 @@ type Mutation {
...
+  patchId: String
```
khelif96, line 147:
```suggestion
  versionId: String
```

graphql/schema.graphql
```diff
@@ -83,7 +83,7 @@ type Mutation {
   schedulePatch(patchId: String!, configure: PatchConfigure!): Patch!
   schedulePatchTasks(patchId: String!): String
   unschedulePatchTasks(patchId: String!, abort: Boolean!): String
-  restartPatch(patchId: String!, abort: Boolean!, taskIds: [String!]!): String
+  restartPatch(patchId: String!, abort: Boolean!, taskIds: [TaskToRestart!]!): String
```
khelif96, line 86: We should do this in a backwards compatible way. 
Lets create a new resolver call restartVersion() with a call signature that matches the pattern we want.

We can add the `@deprecated` directive to this field so we know we will be able to remove it soon.

graphql/resolvers.go
```diff
@@ -1994,14 +1994,14 @@ func (r *mutationResolver) ScheduleUndispatchedBaseTasks(ctx context.Context, pa
 	return scheduledTasks, nil
 }
 
-func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abort bool, taskIds []string) (*string, error) {
+func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abort bool, taskIds []*model.TaskToRestart) (*string, error) {
```
khelif96, line 1997: Can we do this in a backwards compatible way. 
We should create a new resolver called RestartVersion

khelif96 (changes requested):

graphql/resolvers.go
malikchaya2: good point. I'll do that.

graphql/resolvers.go
malikchaya2: oops, no.

graphql/util.go
malikchaya2: I added a check. Does that work?

graphql/schema.graphql
malikchaya2: done

graphql/schema.graphql
malikchaya2: done

graphql/schema.graphql
malikchaya2: done

model/lifecycle.go
malikchaya2: done

sophstad (approved):

graphql/resolvers.go
```diff
@@ -2173,6 +2173,22 @@ func (r *mutationResolver) RestartPatch(ctx context.Context, patchID string, abo
 	return &patchID, nil
 }
 
+func (r *mutationResolver) RestartVersions(ctx context.Context, patchID string, abort bool, versionsToRestart []*model.VersionToRestart) (*string, error) {
+	if len(versionsToRestart) == 0 {
+		return nil, InputValidationError.Send(ctx, "`taskIds` array is empty. You must provide at least one task id")
```
ablack12, line 2178: need to update this comment; may want to make it more user friendly

graphql/util.go
```diff
@@ -670,17 +670,23 @@ const (
...
-	Action   VersionModificationAction `json:"action"`
-	Active   bool                      `json:"active"`
-	Abort    bool                      `json:"abort"`
-	Priority int64                     `json:"priority"`
-	TaskIds  []string                  `json:"task_ids"`
+	Action            VersionModificationAction `json:"action"`
+	Active            bool                      `json:"active"`
+	Abort             bool                      `json:"abort"`
+	Priority          int64                     `json:"priority"`
+	VersionsToRestart []*model.VersionToRestart `json:"task_ids"`
```
ablack12, line 677: can we change the json for this?

model/lifecycle.go
```diff
@@ -48,6 +48,11 @@ type RestartResults struct {
 	ItemsErrored   []string
 }
 
+type VersionToRestart struct {
+	VersionId *string  `json:"patchId"`
```
ablack12, line 52: can we modify the json for this? Also we typically use snake case with json

model/lifecycle.go
```diff
@@ -373,6 +380,18 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 		return errors.Wrap(err, "unable to find version")
 	}
 	return errors.Wrap(version.UpdateStatus(evergreen.VersionStarted), "unable to change version status")
+
+}
+
+// RestartVersions restarts selected tasks for a set of versions,
```
ablack12, line 386: nit: end with period

graphql/util.go
malikchaya2: oops. yes, thanks!

model/lifecycle.go
```diff
@@ -373,6 +380,18 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 		return errors.Wrap(err, "unable to find version")
 	}
 	return errors.Wrap(version.UpdateStatus(evergreen.VersionStarted), "unable to change version status")
+
+}
+
+// RestartVersions restarts selected tasks for a set of versions.
+// If abortInProgress is true for any version, it also sets the abort flag on any in-progress tasks.
+func RestartVersions(taskIds []*VersionToRestart, abortInProgress bool, caller string) error {
```
khelif96, line 388: nit. Maybe we should name the first parameter versionsToRestart. Since its an array of objects that contain taskIds

graphql/schema.graphql
```diff
@@ -144,6 +145,12 @@ type Mutation {
   clearMySubscriptions: Int!
 }
 
+input VersionToRestart {
+  versionId: String
```
khelif96, line 149: This should be required.

graphql/schema.graphql
```diff
@@ -84,7 +84,8 @@ type Mutation {
   schedulePatch(patchId: String!, configure: PatchConfigure!): Patch!
   schedulePatchTasks(patchId: String!): String
   unschedulePatchTasks(patchId: String!, abort: Boolean!): String
-  restartPatch(patchId: String!, abort: Boolean!, taskIds: [String!]!): String
+  restartVersions(patchId: String!, abort: Boolean!, versionsToRestart: [VersionToRestart!]!): String
```
khelif96, line 87: We should name the first param `versionId`.

khelif96, line 87: This is a good opportunity to update the apollo cache. If we return a `Version` or `[Version]` we can include the Version updates along with all the tasks status updates In the response and avoid an extra request on the front end to refetch fresh data.

khelif96 (changes requested):
Had a suggestion for improving our response type for this mutation.

graphql/schema.graphql
malikchaya2: The best way to do this would be to edit ModifyVersionHandler to return the version so that it could be returned here. Since that is a bigish refactor because the function is pretty nested for different actions, and this ticket is holding up the release of the project, how would you feel about creating a ticket for this and doing it at a later point?

khelif96: We can do either way. But It would be easier to implement this prior to the release since that will avoid any backwards compatibility issues that may come with updating this route later on Since this route is not yet used by anything. 
Could we simply do a findVersionById() here and return the versions. https://github.com/evergreen-ci/evergreen/pull/4999/files#diff-a3ce50f20b746b196f29c6d414e685b0110c24b768bcf83756de8d842c98f6dcR2189
it would accomplish the same effect and would not require refactoring any functions.

malikchaya2: Aren't we doing this to avoid a db call though?

malikchaya2: did it

ablack12 (approved):

khelif96 (approved):

graphql/schema.graphql
khelif96: Thanks! For clarity we will be making the db call either way. But doing it this way allows us to avoid an extra network call to fetch fresh data, Since this will return fresh data as part of the initial response.

arjunpatel100 (approved):

PR #5001, opened 2021-09-07
EVG-14601 Update statuses after tasks are aborted

bynn:
[EVG-14601](https://jira.mongodb.org/browse/EVG-14601)

### Description 
updates status on setActiveState now

### Testing 
https://spruce-staging.corp.mongodb.com/version/6137a8299dbe322c5de92d93/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -133,8 +133,7 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 			return errors.Wrap(err, "problem updating display task")
 		}
 	} else {
-		err := UpdateBuildAndVersionStatusForTask(t)
-		if err != nil {
+		if err := UpdateBuildAndVersionStatusForTask(t); err != nil {
 			return errors.Wrap(err, "problem updating build and version status for task")
 		}
 	}
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -128,14 +128,14 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		return nil
 	}
 
-	if err := UpdateBuildAndVersionStatusForTask(t); err != nil {
-		return errors.Wrap(err, "problem updating build and version status for task")
-	}
-
 	if t.IsPartOfDisplay() {
 		if err := UpdateDisplayTaskForTask(t); err != nil {
 			return errors.Wrap(err, "problem updating display task")
 		}
+	}
+
+	if err := UpdateBuildAndVersionStatusForTask(t); err != nil {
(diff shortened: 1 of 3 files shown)


model/task_lifecycle.go
```diff
@@ -134,6 +134,7 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		}
 	}
 
+	UpdateBuildAndVersionStatusForTask(t)
```
ablack12, line 137: We should return this so that we don't suppress the error

ablack12 (changes requested):
Can you change this PR title to have an imperative verb describing what action you took (for example, "update build/version status with task activation")? The title you have here is the problem, not the solution, which we try to avoid.
https://docs.google.com/document/d/1LJHPjLb_bUZxv-l_9Y61kitzAg7u-13I5DjOhZ-deBA/edit#heading=h.fua24zxppw7l

model/task_lifecycle.go
```diff
@@ -134,7 +134,10 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		}
 	}
 
-	UpdateBuildAndVersionStatusForTask(t)
+	err := UpdateBuildAndVersionStatusForTask(t)
+	if err != nil {
```
ablack12, line 138: nit; combine in one line with 137 `if err = ... ; err != nil {`

ablack12 (approved):
LGTM with nit

model/task_lifecycle.go
```diff
@@ -132,6 +132,10 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		if err := UpdateDisplayTaskForTask(t); err != nil {
 			return errors.Wrap(err, "problem updating display task")
 		}
+	} else {
```
ablack12, line 135: Sorry for not catching this: I think we should do this regardless of IsPartOfDisplay; if the execution task status changed, the display task status will also have changed, so the build/version status will have changed.

ablack12 (changes requested):

model/task_lifecycle.go
```diff
@@ -128,6 +128,10 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		return nil
 	}
 
+	if err := UpdateBuildAndVersionStatusForTask(t); err != nil {
```
ablack12, line 131: What was the logic for moving this above the IsPartOfDisplay check? I would think it wouldn't work right until the display task status is updated at 136.

bynn: yeah.

ablack12 (approved):

PR #5002, opened 2021-09-07
EVG-14777: Find tasks by execution when creating BFG

hadjri:
[EVG-14777](https://jira.mongodb.org/browse/EVG-14777)

### Description 
A change was initially made to prevent BFG tickets from being created when task hosts are taken away by AWS.  However in some instances where a task fails and is immediately retried a race condition occurs where the execution tasks for the previous execution have been moved into the old tasks collection for evergreen.  Since we check execution tasks to determine whether to create BFGs for display tasks, it occasionally happens that we are unable to retrieve execution tasks due to this race condition and send out a BFG when we should return early instead.  Change has been made to look up tasks by execution to avoid this issue.


ablack12 (approved):
LGTM with nit

PR #5006, opened 2021-09-08
EVG-12547 stepback earlier tasks in single host task group

ablack12:
[EVG-12547](https://jira.mongodb.org/browse/EVG-12547)

### Description 
Tasks in single host task groups typically need to run sequentially on the same host in order to work. When we stepback a failed task (to see if it succeeds on a previous version), we currently only activate the one task, which means that single-host task group tasks are bound to fail. Instead we should activate the earlier tasks in the group (we don't need to activate the later ones to determine stepback).

### Testing 
Unit test.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -801,7 +801,7 @@ func evalStepback(t *task.Task, caller, status string, deactivatePrevious bool) 
 		}
 
 		if t.IsPartOfSingleHostTaskGroup() {
-			// stepback earlier task group tasks as well
+			// Stepback earlier task group tasks as well because these need to be run sequentially.
 			catcher := grip.NewBasicCatcher()
 			tasks, err := task.FindTaskGroupFromBuild(t.BuildId, t.TaskGroup)
 			if err != nil {
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -796,11 +796,31 @@ func evalStepback(t *task.Task, caller, status string, deactivatePrevious bool) 
 		if err != nil {
 			return errors.WithStack(err)
 		}
-		if shouldStepBack {
-			if err = doStepback(t); err != nil {
-				return errors.Wrap(err, "Error during step back")
-			}
+		if !shouldStepBack {
+			return nil
 		}
+
+		if t.IsPartOfSingleHostTaskGroup() {
+			// stepback earlier task group tasks as well
+			catcher := grip.NewBasicCatcher()
+			tasks, err := task.FindTaskGroupFromBuild(t.BuildId, t.TaskGroup)
+			if err != nil {
+				return errors.Wrapf(err, "can't get task group for task '%s'", t.Id)
+			}
+			if len(tasks) == 0 {
+				return errors.Errorf("no tasks in task group '%s' for task '%s'", t.TaskGroup, t.Id)
+			}
+			for _, tgTask := range tasks {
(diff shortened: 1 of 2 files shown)


model/task_lifecycle_test.go
```diff
@@ -3614,6 +3614,151 @@ tasks:
...
+	prevT3FromDb, err := task.FindOneId(prevT3.Id)
+	assert.NoError(t, err)
+	assert.False(t, prevT3FromDb.Activated)
+
+	// stepping back t3 should now also stepback t3 and not error on earlier activated tasks
+	assert.NoError(t, evalStepback(&t3, "", evergreen.TaskFailed, false))
+	prevT3FromDb, err = task.FindOneId(prevT3.Id)
+	assert.NoError(t, err)
+	assert.True(t, prevT3FromDb.Activated)
+}
```
hadjri, line 3760: It may be worthwhile to add some testing on "unhappy path" error assertions as well, i.e. an error returned when stepping back task group for an earlier task, or  len(tasks) == 0.

ablack12: There is a test for stepping back an earlier task in the task group (although this isn't an error case so I'm not sure if I'm misunderstanding that suggestion). The len(tasks) == 0 is really an impossible state that shouldn't happen.

hadjri: I was referring to the case on line 814 where doStepback for an earlier task can fail to ensure that's handled correctly.  Up to you if you think it's necessart.  LGTM.

hadjri (approved):

model/task_lifecycle_test.go
ablack12: Gotcha-- doStepback has it's own tests so I don't think it needs to be handled specifically here.

model/task_lifecycle.go
```diff
@@ -796,11 +796,31 @@ func evalStepback(t *task.Task, caller, status string, deactivatePrevious bool)
...
 		}
-		if shouldStepBack {
-			if err = doStepback(t); err != nil {
-				return errors.Wrap(err, "Error during step back")
+		if !shouldStepBack {
+			return nil
+		}
+
+		if t.IsPartOfSingleHostTaskGroup() {
+			// stepback earlier task group tasks as well
```
malikchaya2, line 804: I think that adding "because they need to run sequentially on the same host in order to work" to this comment might help people reading the code in the future, but up to you :)

model/task_lifecycle.go
```diff
@@ -796,11 +796,31 @@ func evalStepback(t *task.Task, caller, status string, deactivatePrevious bool)
...
+			if err != nil {
+				return errors.Wrapf(err, "can't get task group for task '%s'", t.Id)
+			}
+			if len(tasks) == 0 {
+				return errors.Errorf("no tasks in task group '%s' for task '%s'", t.TaskGroup, t.Id)
 			}
+			for _, tgTask := range tasks {
+				catcher.Wrapf(doStepback(&tgTask), "error stepping back task group task '%s'", tgTask.DisplayName)
+				if tgTask.Id == t.Id {
+					break // don't need to stepback later tasks in the group
```
malikchaya2, line 816: are we certain that they are stored in this order? with the earliest being the first in the array and the latest being the last in the array? (I'm guessing yes considering the tests pass...)

ablack12: Yep: task groups are defined and stored in this way.

PR #5007, opened 2021-09-08
EVG-15053 add function to change a project's repo

ablack12:
[EVG-15053](https://jira.mongodb.org/browse/EVG-15053)

### Description 
Create function to change owner/repo for project -- validate the new input, remove the project from the previous scope and add it to the new scope ( if the project is attached to a repo), and update the project ref.

### Testing 
Added a unit test.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -553,6 +553,38 @@ func (p *ProjectRef) DetachFromRepo(u *user.DBUser) error {
 	return catcher.Resolve()
 }
 
+func (p *ProjectRef) ChangeOwnerRepo(u *user.DBUser) error {
+	before, err := GetProjectSettingsById(p.Id, false)
+	if err != nil {
+		return errors.Wrap(err, "error getting before project settings event")
+	}
+
+	allowedOrgs := evergreen.GetEnvironment().Settings().GithubOrgs
+	if err := p.ValidateOwnerAndRepo(allowedOrgs); err != nil {
+
+	}
+	if p.UseRepoSettings {
+		if err := p.RemoveFromRepoScope(); err != nil {
+			return errors.Wrapf(err, "error removing project from old repo scope")
+		}
+		p.RepoRefId = "" // will reassign this in add
+		if err := p.AddToRepoScope(u); err != nil {
+			return errors.Wrapf(err, "error addding project to new repo scope")
+		}
+	}
+	update := bson.M{
+		"$set": bson.M{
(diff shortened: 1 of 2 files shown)


model/project_ref.go
```diff
@@ -553,6 +553,38 @@ func (p *ProjectRef) DetachFromRepo(u *user.DBUser) error {
 	return catcher.Resolve()
 }
 
+func (p *ProjectRef) ChangeOwnerRepo(u *user.DBUser) error {
```
malikchaya2, line 556: The way I thought this function would work would be to take in a project, and take in the the repo we want it to be under, and change that project to now be associated with the new repo. This only takes in a user, and I'm a bit confused about how a graphql resolver would call this.

model/project_ref_test.go
```diff
@@ -266,6 +268,52 @@ func TestGetActivationTimeWithCron(t *testing.T) {
...
+		Id: "myRepo",
+	}}
+	assert.NoError(t, repoRef.Upsert())
+	u := &user.DBUser{Id: "me",
+		SystemRoles: []string{GetViewRepoRole("myRepo")},
+	}
+	assert.NoError(t, u.Insert())
+	pRef.Owner = "newOwner"
+	pRef.Repo = "newRepo"
+	assert.NoError(t, pRef.ChangeOwnerRepo(u))
```
malikchaya2, line 301: Would it maybe makes sense to change the flow a bit to do this part in the function? Like have the function take in the new owner and repo, and set it in the function, as opposed to having whoever calls this(most likely a graphql resolver) set it first on the projectRef and then calling ChangeOwnerRepo? 

(this is a comment on the overall setup, not the test specifically).

ablack12: I thought about that -- I'm not sure, it really depends on how information is passed to the resolver I guess. The other Edit functions assume that the ProjectRef itself has been modified and then just updates the relevant pieces so I thought this would be more consistent. (I think this also covers the above question.) If you think it's better the other way though I can change it to take in owner/repo.

malikchaya2: Either way works, and I hear the argument for consistency. I'll approve and leave it up to you to decide.

malikchaya2 (approved):

model/project_ref_test.go
ablack12: I'll leave it and it can be a quick change if it turns out easier for the UI to have it work the other way 👍

PR #5009, opened 2021-09-09
EVG-14178 Log cannot map pull request to project at a lower level

bynn:
[EVG-14178](https://jira.mongodb.org/browse/EVG-14178)

### Description 
Since `FindOneProjectRefByRepoAndBranchWithPRTesting` already logs a debug message, I just deleted it

### Testing 
n\a

### DIFF
--- a/rest/data/patch_intent.go
+++ b/rest/data/patch_intent.go
@@ -19,13 +19,16 @@ type DBPatchIntentConnector struct{}
 
 func (p *DBPatchIntentConnector) AddPatchIntent(intent patch.Intent, queue amboy.Queue) error {
 	patchDoc := intent.NewPatch()
-	_, err := model.FindOneProjectRefByRepoAndBranchWithPRTesting(patchDoc.GithubPatchData.BaseOwner,
+	projectRef, err := model.FindOneProjectRefByRepoAndBranchWithPRTesting(patchDoc.GithubPatchData.BaseOwner,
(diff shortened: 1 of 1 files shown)


rest/data/patch_intent.go
```diff
@@ -19,20 +19,14 @@ type DBPatchIntentConnector struct{}
...
+	_, err := model.FindOneProjectRefByRepoAndBranchWithPRTesting(patchDoc.GithubPatchData.BaseOwner,
 		patchDoc.GithubPatchData.BaseRepo, patchDoc.GithubPatchData.BaseBranch)
 	if err != nil {
 		return gimlet.ErrorResponse{
 			StatusCode: http.StatusInternalServerError,
 			Message:    "failed to fetch project_ref",
 		}
 	}
-	if projectRef == nil {
-		return gimlet.ErrorResponse{
```
ybrill, line 31: I think it's still valuable to return here. For example, if the project has PR testing disabled we don't want to continue to create the patch.

bynn: returns nil

ablack12: I think we should still log this to grip; it's possible that we will want this information if users ask why a PR isn't being made, but logging at the Debug level might be more accurate (and then returning nil).

ablack12: Oh just saw your comment that we already log at the debug level so disregard.

ybrill (approved):

PR #5010, opened 2021-09-09
EVG-14593 Add new notification type for task started

bynn:
[EVG-14593](https://jira.mongodb.org/browse/EVG-14593)

### Description 
new trigger for notifications task-started
made a ticket for the front end portion
https://jira.mongodb.org/browse/EVG-15390

### Testing 
added to unit test
image
image

### DIFF
--- a/trigger/payloads.go
+++ b/trigger/payloads.go
@@ -27,7 +27,7 @@ const (
 	evergreenSuccessColor    = "#4ead4a"
 	evergreenFailColor       = "#ce3c3e"
 	evergreenSystemFailColor = "#ce3c3e"
-	evergreenRunningColor    = "#FEF7E3"
+	evergreenRunningColor    = "#ffdd49"
 
 	// slackAttachmentsLimit is a limit to the number of extra entries to
 	// attach to a Slack message. It does not count the link to Evergreen,
--- a/trigger/payloads.go
+++ b/trigger/payloads.go
@@ -27,6 +27,7 @@ const (
 	evergreenSuccessColor    = "#4ead4a"
 	evergreenFailColor       = "#ce3c3e"
 	evergreenSystemFailColor = "#ce3c3e"
+	evergreenRunningColor    = "#FEF7E3"
 
 	// slackAttachmentsLimit is a limit to the number of extra entries to
 	// attach to a Slack message. It does not count the link to Evergreen,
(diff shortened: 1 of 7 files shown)


ablack12 (commented):
There's some more stuff you need here. You need to add a way to subscribe to the notification, and then add a new kind of event log for when a task starts (assuming we don't have one already) -- we process these events and call triggers based off of them. You also likely will need to tweak [this function|https://github.com/evergreen-ci/evergreen/blob/e8db8e20cbd608fff5005b26ae152bc01af49c90/trigger/task.go#L293] to verify that the message is formatted correctly. This PR has some unnecessary stuff mixed in but might help: https://github.com/evergreen-ci/evergreen/pull/4914/files

You can also add this to the legacy UI and then test this in staging.

bynn:
I can't seem to figure out why nothing is sending.
I checked the database and the Subscription is being made properly and the start event is also being logged correctly.

trigger/task.go
```diff
@@ -293,6 +295,8 @@ func (t *taskTriggers) makeData(sub *event.Subscription, pastTenseOverride, test
 	} else if data.PastTenseStatus == evergreen.TaskSucceeded {
 		slackColor = evergreenSuccessColor
 		data.PastTenseStatus = "succeeded"
+	} else if data.PastTenseStatus == evergreen.TaskStarted {
+		slackColor = evergreenSuccessColor
```
ablack12, line 299: I'm thinking we should add a new yellow color for started.

ablack12 (changes requested):
Looks great!!! Only one change needed that I see

trigger/payloads.go
```diff
@@ -27,6 +27,7 @@ const (
 	evergreenSuccessColor    = "#4ead4a"
 	evergreenFailColor       = "#ce3c3e"
 	evergreenSystemFailColor = "#ce3c3e"
+	evergreenRunningColor    = "#FEF7E3"
```
ablack12, line 30: Not sure if you've tested this color but it seems a little dull. It looks like we use this in the UI (as an option) https://github.com/ablack12/evergreen/blob/6e1a16313ed11e8cce2934bcc2402b161c458c8e/public/static/less/global_styles.less#L66
This 
image
versus 
image

bynn: image

ablack12 (approved):

PR #5012, opened 2021-09-09
EVG-15267 use directories consistently for applying module patches

ablack12:
[EVG-15267](https://jira.mongodb.org/browse/EVG-15267)

### Description 
Commands aren't my strong suite so I'm only partially convinced this makes sense but:
It seems like when we clone the module, we setup the command directory and the module directory using different paths than when we apply the patch. This works for the non `${workdir}` case but not for the other case I think because of how we order things. By using the same scheme as when we clone the module, this works for both cases. (I think.)

Adding Kim to review because she'll probably understand this the best.

### Testing 
[Patch](https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu1604_unit_tests_patch_92f8580018375186b1129c7599ee2db76c1605be_613a83e09dbe3251f4f7e817_21_09_09_22_00_30/logs?execution=0&logtype=agent) with `prefix ../../src`
[Patch](https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu1604_unit_tests_patch_92f8580018375186b1129c7599ee2db76c1605be_613a852797b1d342a4a20007_21_09_09_22_05_56/logs?execution=0&logtype=agent) with `prefix ${workdir}/src`
Note that these succeed regardless of git.get_project since it's a pre_task command for this project, but there are no errors in the logs.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -778,7 +778,7 @@ func (c *gitFetchProject) applyPatch(ctx context.Context, logger client.LoggerPr
 				continue
 			}
 
-			dir = filepath.Join(c.Directory, expandModulePrefix(conf, module.Name, module.Prefix, logger), module.Name)
+			dir = filepath.ToSlash(filepath.Join(expandModulePrefix(conf, module.Name, module.Prefix, logger), module.Name))
 		}
 
 		if len(patchPart.PatchSet.Patch) == 0 {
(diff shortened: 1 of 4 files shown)


malikchaya2 (commented):
I don't think I have enough context here, so I'll leave this one for @Kimchelly :)

Kimchelly:
I also don't have any context on this because I never deal with the working directory.

ablack12:
This is failing tests anyway :( back to the drawing board.

PR #5013, opened 2021-09-10
EVG-15232: check for start by evergreen user

hhoke:
[EVG-15232](https://jira.mongodb.org/browse/EVG-15232)

### Description 
Idle host count was off by a constant number. I spawned 3 hosts manually, and this number appeared to go up by 3. Consequently, we need to filter out all hosts that are not started by  evergreen.User, which this PR accomplishes.

### Testing 
 Modified existing test `TestHostStatsByDistro` by adding a host without a StartedBy value.


model/host/stats.go
```diff
@@ -110,6 +110,8 @@ func statsByDistroPipeline() []bson.M {
 	return []bson.M{
 		{
 			"$match": bson.M{
+				// don't count user-spawned hosts EVG-15232
+				StartedByKey: evergreen.User,
```
ybrill, line 114: once we have this do we still need [the check](https://github.com/evergreen-ci/evergreen/blob/dd3199cc8f9ee0d3316f8b1251a3d956a3b3ab83/model/host/stats.go#L116-L117) that it wan't started by a task? If it was started by a task wouldn't the StartedBy field be the task id?

hhoke: Yes, I think it would be.

ybrill (approved):

PR #5017, opened 2021-09-10
EVG-15379: Generate test log URLs at the model level and use display test names

julianedwards:
EVG-15379: https://jira.mongodb.org/browse/EVG-15379

I tested this in staging along with my changes in lobster [here](https://github.com/evergreen-ci/lobster/pull/114).


model/task/task.go
```diff
@@ -324,14 +325,70 @@ func (tr *TestResult) GetLogTestName() string {
...
 		return tr.DisplayTestName
 	}
 
 	return tr.TestFile
 }
 
+// GetLogURL returns the external or internal log URL for this test result.
+func (tr TestResult) GetLogURL(raw bool) string {
+	root := evergreen.GetEnvironment().Settings().ApiUrl
+	deprecatedLobsterURL := "https://logkeeper.mongodb.org"
```
Kimchelly, line 339: Should this be renamed to `deprecatedLogkeeperURL`? Lobster is just the front end for viewing HTML logs.

rest/model/test.go
```diff
@@ -29,139 +28,72 @@ type APITest struct {
...
-				"/test_log/%s/%d?test_name=%s&group_id=%s&text=true",
-				url.PathEscape(v.TaskID),
-				v.Execution,
-				url.QueryEscape(v.TestFile),
-				url.QueryEscape(v.GroupID),
-			))
-		} else {
-			dispString := fmt.Sprintf("/test_log/%s?text=true", *at.Logs.LogId)
-			at.Logs.RawDisplayURL = &dispString
+		if v.LogID != "" {
```
Kimchelly, line 68: What's the significance of not allowing LogID to be an empty string?

service/templates/task.html
```diff
@@ -470,15 +470,15 @@ <h4 class="text-center text-muted">No test results to show</h4>
                           </td>
                           <td class="col-lg-3">
                             <div class="url-link nowrap" id="html-link">
-                                <a ng-href="[[getURL(test.test_result, false)]]" ng-hide="[[hideURL(test.test_result, false)]]">
+                                <a ng-href="[[test.test_result.url]]">
                                   HTML
                                 </a>
                               </div>
                               <div id="separator">
-                                <b ng-show="[[hasBothURL(test)]]">&middot;</b>
```
Kimchelly, line 478: Is it alright to assume that both the raw and HTML URLs will always be present for a task (including display tasks)?

trigger/task.go
```diff
@@ -774,15 +774,15 @@ func (t *taskTriggers) shouldIncludeTest(sub *event.Subscription, previousTask *
...
+	oldTestResult, ok := t.oldTestResults[test.GetDisplayTestName()]
 	// a new test in an existing task is defined as a regression
 	if !ok {
 		return true, nil
 	}
 
 	if isTestStatusRegression(oldTestResult.Status, test.Status) {
 		// try to find a stepback alert
-		alertForStepback, err := alertrecord.FindByTaskRegressionTestAndOrderNumber(sub.ID, test.TestFile, currentTask.DisplayName, currentTask.BuildVariant, currentTask.Project, previousTask.RevisionOrderNumber)
+		alertForStepback, err := alertrecord.FindByTaskRegressionTestAndOrderNumber(sub.ID, test.GetDisplayTestName(), currentTask.DisplayName, currentTask.BuildVariant, currentTask.Project, previousTask.RevisionOrderNumber)
```
Kimchelly, line 785: Is using `GetDisplayTestName` instead of `TestFile` backwards compatible with non-Cedar test results?

Kimchelly (changes requested):
I reviewed the back-end changes, but can't meaningfully review the UI/GraphQL modifications.

graphql/resolvers.go
```diff
@@ -1392,14 +1392,6 @@ func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution
 			if err = apiTest.BuildFromService(&t); err != nil {
 				return nil, InternalServerError.Send(ctx, err.Error())
 			}
-			if err = util.CheckURL(utility.FromStringPtr(apiTest.Logs.HTMLDisplayURL)); apiTest.Logs.HTMLDisplayURL != nil && err != nil {
```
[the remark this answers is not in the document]
julianedwards: These fields no longer exist. I removed them since they duplicated `URL` and `URLRaw`. They were there because sometimes we needed to prepend the api URL (`https://evergreen.mongodb.com`).

model/task/task.go
julianedwards: Good catch

rest/route/service_test.go
```diff
@@ -688,25 +688,9 @@ func TestTestPaginator(t *testing.T) {
...
-						Logs: model.TestLogs{
-							URL:            utility.ToStringPtr(""),
-							URLRaw:         utility.ToStringPtr(""),
-							LogId:          utility.ToStringPtr(""),
-							RawDisplayURL:  utility.ToStringPtr("/test_log//0?test_name=&group_id=&text=true"),
-							HTMLDisplayURL: utility.ToStringPtr("/test_log//0?test_name=&group_id=#L0"),
-						},
-					}
+					nextModelTest := &model.APITest{}
+					_ = nextModelTest.BuildFromService(&cachedTests[i])
```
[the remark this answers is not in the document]
julianedwards: They populate the `APITest` with the expected output. It would only fail if an unrecognized type is passed in, but since we know both `string` and `*testresult.TestResult` are accepted there is no point of checking the error.

service/templates/task.html
julianedwards: Well, for test logs in cedar we don't know if they exist until we actually hit the route.

trigger/task.go
julianedwards: Yes! If the display name DNE, `GetDisplayTestName` returns `TestFile`.

rest/model/test.go
julianedwards: I just don't want to clutter the API response. LogID is only really useful when fetching test logs that are stored in evg (which `GetLogURL` already takes care of since it populates the url fields).

model/task/task.go
julianedwards: Actually, this is correct. Lobster was hosted in previously hosted in logkeeper and we moved it over to evergreen.

Kimchelly: Sure but were the raw logs also using lobster?

julianedwards: Yeah, I replace it for raw logs as well.

Kimchelly: My question is about the variable name - lobster isn't used at all for raw logs is it? That would make this a deceptive field by implying that lobster is somehow tied to the raw logs.

Kimchelly (approved):
I'm gonna LGTM this mod the one open comment thread.

model/task/task.go
julianedwards: Ah, yes I see what you are saying now. This is actually a bug—raw logs from logkeeper still need to use the `logkeeper.mongodb.org` route. Thanks for the pushback!

julianedwards: But I still think the variable name should be `deprecatedLobsterURL` cause `logkeeper.mongodb.org` itself is not deprecated (we still download valid logs from there) it more so that we no longer want to use the lobster hosted there.

Kimchelly (approved):

rest/model/test.go
```diff
@@ -29,139 +29,75 @@ type APITest struct {
...
 }
 
 // TestLogs is a struct for storing the information about logs that will be
 // written out as part of an APITest.
 type TestLogs struct {
-	URL            *string `json:"url"`
-	LineNum        int     `json:"line_num"`
-	URLRaw         *string `json:"url_raw"`
-	LogId          *string `json:"log_id"`
-	RawDisplayURL  *string `json:"url_raw_display"`
```
arjunpatel100, line 42: We should propagate these forward until spruce is relies on the new variables

arjunpatel100 (approved):

PR #5018, opened 2021-09-13
EVG-15203: Support scheduling settings in project yaml

hadjri:
[EVG-15203](https://jira.mongodb.org/browse/EVG-15203)

### Description 
Scheduling settings can currently only be configured by clicking in the Evergreen UI on the project page.  Support has been added to enable the toggle of the deactivate_previous flag in the project yaml, and in the event that both are set we take the value of what is set in the yaml config.  There are now several places in the codebase where we retrieve the project parser and compare it to the project ref, so some refactoring was done to abstract that away to a helper function to improve readability.


model/project.go
```diff
@@ -61,6 +61,8 @@ type Project struct {
 	Loggers                *LoggerConfig                  `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
 	PerfEnabled            bool                           `yaml:"perf_enabled,omitempty" bson:"perf_enabled,omitempty"`
+	// ProjectRef override fields
```
ablack12, line 64: This comment isn't immediately clear. Are you trying to indicate that fields below this comment can be overridden by the parser project? (In which case could you separate this more from 64 and expand the comment?)

model/project_parser.go
```diff
@@ -83,6 +83,10 @@ type ParserProject struct {
 	CreateTime             time.Time                      `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
 	PerfEnabled            *bool                          `yaml:"perf_enabled,omitempty" bson:"perf_enabled,omitempty"`
+
+	// ProjectRef override fields
```
ablack12, line 87: Same comment

model/project_parser_db.go
```diff
@@ -54,6 +54,26 @@ func ParserProjectFindOneById(id string) (*ParserProject, error) {
 	return ParserProjectFindOne(ParserProjectById(id))
 }
 
+func ParserProjectFindOneByVersion(id string, version string) (*ParserProject, error) {
```
ablack12, line 57: Let's rename ID to projectId, or else this reads like version Id. Can you add a function comment to indicate what happens if version isn't passed in?

model/project_parser_db.go
```diff
@@ -54,6 +54,26 @@ func ParserProjectFindOneById(id string) (*ParserProject, error) {
...
+		parserProject, err := ParserProjectFindOneById(version)
+		if err != nil {
+			return nil, err
+		}
+		if parserProject == nil {
+			return nil, errors.Errorf("Unable to find parser project for project '%s'", id)
+		}
+		return parserProject, nil
+	}
+	return nil, errors.Errorf("Unable to find last good version for project '%s'", id)
```
ablack12, line 74: We should error here on 60 instead; it's more clear than waiting to error, and we can wrap the actual problem. Then at 64 we can just assume version != ""

service/api_task.go
```diff
@@ -223,7 +228,11 @@ func (as *APIServer) EndTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// mark task as finished
-	err = model.MarkEnd(t, APIServerLockTitle, finishTime, details, projectRef.ShouldDeactivatePrevious())
+	deactivatePrevious := projectRef.ShouldDeactivatePrevious()
+	if utility.FromBoolPtr(projectParser.DeactivatePrevious) {
```
ablack12, line 232: I might have a preference for making a ShouldDeactivatePrevious function for project parser to be consistent. This was meant to make the code a bit more readable.

ablack12 (changes requested):

ablack12 (approved):

PR #5021, opened 2021-09-13
EVG-14123: Prevent build baron config errors from stopping the app from starting

hadjri:
[EVG-14123](https://jira.mongodb.org/browse/EVG-14123)

### Description 
Currently when a validation error occurs for a project's build baron configuration, the UI server errors out and fails to come up.  Change has been made to log out errors notifying of any misconfiguration and then omitting the build baron project configuration from the Build Baron plugin options struct rather than returning an error.
### Testing 
Tested locally with malformed build baron configs and the app started successfully.  Unit test refactoring.


plugin/build_baron.go
```diff
@@ -47,45 +50,78 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
 		}
 		webhookConfigured := webHook.Endpoint != ""
 		if !webhookConfigured && proj.TicketCreateProject == "" {
-			return fmt.Errorf("ticket_create_project and taskAnnotationSettings.FileTicketWebHook endpoint cannot both be blank")
+			grip.Error(message.WrapError(err, message.Fields{
```
malikchaya2, line 53: @ablack12 do you think these errors should be grip.Critical? If it gets to this point and this is the case, we probably entered something bad into the db manually. I wouldn't want it to fail silently and only notice after a user gets frustrated.

plugin/build_baron.go
```diff
@@ -47,45 +50,78 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
...
+				"message":      "ticket_create_project and taskAnnotationSettings.FileTicketWebHook endpoint cannot both be blank",
+				"project_name": projName,
+			}))
+			hasValidationError = true
 		}
 		if !webhookConfigured && len(proj.TicketSearchProjects) == 0 {
```
malikchaya2, line 59: Are we sure that there aren't places where we rely on this not being empty (because until now, this would always be > 0 or else the app wouldn't start) ? If yes, we may need to add some extra checks. Same with the other checks that we are removing. Now that we know longer have these guarantees, so we need additional checks anywhere?

plugin/build_baron.go
```diff
@@ -47,45 +50,78 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
 		}
 		webhookConfigured := webHook.Endpoint != ""
 		if !webhookConfigured && proj.TicketCreateProject == "" {
-			return fmt.Errorf("ticket_create_project and taskAnnotationSettings.FileTicketWebHook endpoint cannot both be blank")
```
malikchaya2, line 50: now that we no longer have the guarantee that they won't both be defined, should we change how the IsWebhookConfigured() function behaves? For example, if a project has both the webhook configured and the TicketCreateProject defined, then we don't know where to file the ticket if someone clicks on the button. So maybe we should add a check like this to IsWebhookConfigured() and error if both are defined.

plugin/build_baron.go
ablack12: I think it's fine to make this critical (which means that the errors go to evergreen-ops-alerts @hadjri). 

Really all of this validation should exist for the project routes/pages instead of here; once we've implemented this for the legacy UI and routes and removed the ability to define this on the admin page, I believe we can remove this Configure loop entirely. (In my mind we would be doing this ticket after that but we may just want to make a new ticket to ensure validation for plugins is done in the right places and to remove this validation once we've removed the admin page option)

plugin/build_baron.go
```diff
@@ -47,45 +50,78 @@ func (bbp *BuildBaronPlugin) Configure(conf map[string]interface{}) error {
...
+				grip.Error(message.WrapError(err, message.Fields{
+					"message":      fmt.Sprintf(`Failed to parse bf_suggestion_server for project "%s"`, projName),
+					"project_name": projName,
+				}))
+				hasValidationError = true
 			}
 			if proj.BFSuggestionUsername == "" && proj.BFSuggestionPassword != "" {
-				return errors.Errorf(`Failed validating configuration for project "%s": `+
-					"bf_suggestion_password must be blank if bf_suggestion_username is blank", projName)
+				grip.Error(message.WrapError(err, message.Fields{
```
ablack12, line 75: message.WrapError isn't what you want in some of these cases; if err doesn't exist, this will no op.

plugin/build_baron.go
hadjri: That makes sense to me.  Since this PR's been opened and the logic for validating the build baron plugin will likely remain similar to what currently exists (just probably need to move this logic somewhere else), I've created https://jira.mongodb.org/browse/EVG-15419 and added it to the project scope.  I'll address the remaining PR comments here.

plugin/build_baron.go
hadjri: We will still have a guarantee that they wont both be defined, since we are not removing the build baron validation coming from the admin page until we remove the ability to define plugin configs on the admin page.  I think your comment is right and should be implemented in https://jira.mongodb.org/browse/EVG-15419, however I think the scope for now is to stop the app from crashing when a bad build baron config is processed.

plugin/build_baron.go
hadjri: Same comment as [here](https://github.com/evergreen-ci/evergreen/pull/5021#discussion_r710201201), I think this should now be implemented in https://jira.mongodb.org/browse/EVG-15419

plugin/build_baron.go
malikchaya2: Where do we have the guarantee that they won't both be defined? 
While the main scope is to stop the app from crashing, I wouldn't want to potentially leave anything in a potentially bad state.

hadjri: We have kept line 107 that checks if both are defined and sends a critical error if so. None of the previously existing validation checks have changed, only how we handle the errors has changed.

malikchaya2: right, but the way we used to handle it is to not start the app, so if someone had both defined, no user would be clicking a "file ticket" button because we wouldn't have an app running at all. Now that the way we handle it no longer provides that guarantee, I think we should add a check to IsWebhookConfigured() so that when a user tries to file a ticket in that state, they  gets an error saying that it's not configured properly, instead of it picking one and filing it with one, when it may not be the one they intended. 
A critical error will alert us, but nothing will get fixed until someone on our team sees it, understands whats wrong, reaches out to the team, and fixes it.

graphql/ui_plugin_build_baron.go
```diff
@@ -54,7 +54,9 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
 	}
 
 	webHook, _ := plugin.IsWebhookConfigured(t.Project, t.Version)
-	if webHook.Endpoint != "" {
+	if webHook.Endpoint != "" && bbProject.TicketCreateProject != "" {
```
malikchaya2, line 57: I think I would prefer for this check to be in IsWebhookConfigured so that it covers more use cases.

graphql/ui_plugin_build_baron.go
```diff
@@ -53,8 +53,11 @@ func BbFileTicket(context context.Context, taskId string, execution int) (bool,
 		return taskNotFound, errors.Errorf("error finding build baron plugin for task '%s'", taskId)
 	}
 
-	webHook, _ := plugin.IsWebhookConfigured(t.Project, t.Version)
-	if webHook.Endpoint != "" {
+	webHook, ok, err := plugin.IsWebhookConfigured(t.Project, t.Version)
+	if err != nil {
+		return taskNotFound, err
```
malikchaya2, line 58: I think it might be helpful to wrap the error with more explanation here.

malikchaya2 (approved):

PR #5022, opened 2021-09-13
EVG-15396: Fix historical test stats aggregation to include display test names

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-15396

- Use display test name in historical test stat calculations.
- Exclude stats of the two affected build variants during the affected period from historical test stat queries.

### DIFF
--- a/model/stats/db.go
+++ b/model/stats/db.go
@@ -801,7 +801,7 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
 	// with jasper (SERVER-54315), test names are now random UUIDs with
 	// with a human-readble display name. For the period between
 	// 09/01/21-09/13/21, the historical test stats calculations failed to
-	// for this (EVG-15396). We should ignored any test stats affected by
+	// for this (EVG-15396). We should ignore any test stats affected by
 	// this bug.
 	affectedVariants := []string{
 		"enterprise-rhel-80-64-bit",
--- a/model/stats/db.go
+++ b/model/stats/db.go
@@ -801,8 +801,8 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
 	// with jasper (SERVER-54315), test names are now random UUIDs with
 	// with a human-readble display name. For the period between
 	// 09/01/21-09/13/21, the historical test stats calculations failed to
-	// for this (EVG-15396). We should ignore any test stats affected by
-	// this bug.
+	// account for this (EVG-15396). We should ignore any test stats
+	// affected by this bug.
 	affectedVariants := []string{
 		"enterprise-rhel-80-64-bit",
 		"enterprise-rhel-80-64-bit-dynamic-required",
--- a/model/stats/db.go
+++ b/model/stats/db.go
@@ -800,7 +800,7 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
 	// After cutting over the below build variants to use resmoke spawned
 	// with jasper (SERVER-54315), test names are now random UUIDs with
 	// with a human-readble display name. For the period between
-	// 09/01/21-09/13/21, the historical test stats aggregation failed to
+	// 09/01/21-09/13/21, the historical test stats calculations failed to
 	// for this (EVG-15396). We should ignored any test stats affected by
 	// this bug.
 	affectedVariants := []string{
--- a/model/stats/db.go
+++ b/model/stats/db.go
@@ -601,46 +601,6 @@ func (filter StatsFilter) buildMatchStageForTest() bson.M {
 		DbTestStatsIdRequesterKeyFull: bson.M{"$in": filter.Requesters},
 	}
 
-	// After cutting over the below build variants to use resmoke spawned
-	// with jasper (SERVER-54315), test names are now random UUIDs with
-	// with a human-readble display name. For the period between
-	// 09/01/21-09/14/21, the historical test stats calculations failed to
-	// account for this (EVG-15396). We should ignore any test stats
-	// affected by this bug.
-	//
-	// TODO: (EVG-15396) Remove this code once all affected test stats TTL
-	// on 03/14/2022.
-	affectedVariants := []string{
-		"enterprise-rhel-80-64-bit",
-		"enterprise-rhel-80-64-bit-dynamic-required",
-	}
-	if len(filter.BuildVariants) == 0 || len(utility.StringSliceIntersection(filter.BuildVariants, affectedVariants)) > 0 {
-		or := []bson.M{
-			{
-				"$and": []bson.M{
-					{DbTaskStatsIdBuildVariantKeyFull: bson.M{"$in": affectedVariants}},
-					{DbTaskStatsIdDateKeyFull: bson.M{
-						"$lt": time.Date(2021, time.September, 1, 0, 0, 0, 0, time.UTC),
-						"$gt": time.Date(2021, time.September, 14, 0, 0, 0, 0, time.UTC),
-					}},
-				},
-			},
-			{
-				DbTaskStatsIdBuildVariantKeyFull: bson.M{"$nin": affectedVariants},
-			},
-		}
-		if filter.StartAt != nil {
-			match["$and"] = []bson.M{
-				{"$or": or},
-				{"$or": filter.buildTestPaginationOrBranches()},
-			}
-		} else {
-			match["$or"] = or
-		}
-	} else if filter.StartAt != nil {
-		match["$or"] = filter.buildTestPaginationOrBranches()
-	}
-
 	if len(filter.Tests) > 0 {
 		match[DbTestStatsIdTestFileKeyFull] = BuildMatchArrayExpression(filter.Tests)
 	}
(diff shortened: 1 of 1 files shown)


model/stats/db.go
```diff
@@ -228,16 +228,26 @@ func getHourlyTestStatsPipeline(projectId string, requester string, start time.T
...
+					testresult.EndTimeKey:         1}}},
 			"as": "testresults"}},
 		{"$unwind": "$testresults"},
 		{"$project": bson.M{
-			dbTestStatsIdTestFileKey: "$testresults." + testresult.TestFileKey,
+			// Use the display test name if there is one.
+			dbTestStatsIdTestFileKey: bson.M{
+				"$cond": bson.M{
+					"if": bson.M{
+						"$ne": []string{bsonutil.GetDottedKeyName("$testresults", testresult.DisplayTestNameKey), ""},
```
Kimchelly, line 245: I'm not that familiar with this syntax since usually, it's expressed as `"field": {"$ne": <value>}`. Is it okay to use a string slice instead of an interface slice (i.e. `Array`) like the other places do in the test stats code?

model/stats/db.go
```diff
@@ -786,6 +796,34 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
 		DbTestStatsIdProjectKeyFull:   filter.Project,
 		DbTestStatsIdRequesterKeyFull: bson.M{"$in": filter.Requesters},
 	}
+
+	// After cutting over the below build variants to use resmoke spawned
```
Kimchelly, line 800: Can you add a TODO to remove this once the data has been deleted by the TTL?

model/stats/db.go
```diff
@@ -786,6 +796,34 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
...
+	// with jasper (SERVER-54315), test names are now random UUIDs with
+	// with a human-readble display name. For the period between
+	// 09/01/21-09/13/21, the historical test stats calculations failed to
+	// for this (EVG-15396). We should ignore any test stats affected by
+	// this bug.
+	affectedVariants := []string{
+		"enterprise-rhel-80-64-bit",
+		"enterprise-rhel-80-64-bit-dynamic-required",
+	}
+	if len(filter.Tasks) == 0 || len(utility.StringSliceIntersection(filter.Tasks, affectedVariants)) > 0 {
```
Kimchelly, line 810: Why is this comparing the task filters against the buildvariant names? I would have thought it would compare against `filter.BuildVariants`.

model/stats/db.go
```diff
@@ -786,6 +796,34 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
 		DbTestStatsIdProjectKeyFull:   filter.Project,
 		DbTestStatsIdRequesterKeyFull: bson.M{"$in": filter.Requesters},
 	}
+
+	// After cutting over the below build variants to use resmoke spawned
+	// with jasper (SERVER-54315), test names are now random UUIDs with
+	// with a human-readble display name. For the period between
+	// 09/01/21-09/13/21, the historical test stats calculations failed to
```
Kimchelly, line 803: ... calculations failed to account for this?

Kimchelly (changes requested):

model/stats/db.go
julianedwards: Done.

model/stats/db.go
```diff
@@ -786,6 +796,34 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
...
+	// account for this (EVG-15396). We should ignore any test stats
+	// affected by this bug.
+	affectedVariants := []string{
+		"enterprise-rhel-80-64-bit",
+		"enterprise-rhel-80-64-bit-dynamic-required",
+	}
+	if len(filter.Tasks) == 0 || len(utility.StringSliceIntersection(filter.Tasks, affectedVariants)) > 0 {
+		match["$or"] = []bson.M{
```
ybrill, line 814: if there is pagination will this get overwritten [later](https://github.com/evergreen-ci/evergreen/blob/dd3199cc8f9ee0d3316f8b1251a3d956a3b3ab83/model/stats/db.go#L800)?

model/stats/db.go
```diff
@@ -786,6 +796,34 @@ func (filter StatsFilter) buildMatchStageForTask() bson.M {
...
+			{
+				"$and": []bson.M{
+					{DbTaskStatsIdBuildVariantKeyFull: bson.M{"$in": affectedVariants}},
+					{DbTaskStatsIdDateKeyFull: bson.M{
+						"$lt": time.Date(2021, time.September, 1, 0, 0, 0, 0, time.UTC),
+						"$gt": time.Date(2021, time.September, 13, 0, 0, 0, 0, time.UTC),
```
ybrill: Should this be tomorrow? (assuming we deploy this tomorrow)

model/stats/db.go
ybrill: I thought it works this way because it's [the aggregation version](https://docs.mongodb.com/manual/reference/operator/aggregation/ne/)

model/stats/db.go
julianedwards: Done.

model/stats/db.go
julianedwards: This should work in the same way that string slices are marshaled into valid BSON arrays when in our schema structs. I am double checking in staging.

Kimchelly (approved):

model/stats/db.go
julianedwards: Good catch, I fixed it.

model/stats/db.go
```diff
@@ -590,6 +600,47 @@ func (filter StatsFilter) buildMatchStageForTest() bson.M {
...
+	// affected by this bug.
+	//
+	// TODO: (EVG-15396) Remove this code once all affected test stats TTL
+	// on 03/14/2022.
+	affectedVariants := []string{
+		"enterprise-rhel-80-64-bit",
+		"enterprise-rhel-80-64-bit-dynamic-required",
+	}
+	if len(filter.BuildVariants) == 0 || len(utility.StringSliceIntersection(filter.BuildVariants, affectedVariants)) > 0 {
+		or := []bson.M{
```
ybrill, line 618: This looks good to me.
Maybe an alternative could be to $not the $and, like
```
"$not": []bson.M{{"$and": []bson.M{
	{DbTaskStatsIdBuildVariantKeyFull: bson.M{"$in": affectedVariants}},
	{DbTaskStatsIdDateKeyFull: bson.M{
		"$lt": time.Date(2021, time.September, 1, 0, 0, 0, 0, time.UTC),
		"$gt": time.Date(2021, time.September, 14, 0, 0, 0, 0, time.UTC),
	}},
}}},
```

Kimchelly (approved):

model/stats/db.go
```diff
@@ -590,6 +600,32 @@ func (filter StatsFilter) buildMatchStageForTest() bson.M {
...
+	// affected by this bug.
+	//
+	// TODO: (EVG-15396) Remove this code once all affected test stats TTL
+	// on 03/14/2022.
+	affectedVariants := []string{
+		"enterprise-rhel-80-64-bit",
+		"enterprise-rhel-80-64-bit-dynamic-required",
+	}
+	if len(filter.BuildVariants) == 0 || len(utility.StringSliceIntersection(filter.BuildVariants, affectedVariants)) > 0 {
+		match["$not"] = []bson.M{
```
ybrill, line 618: the [doc](https://docs.mongodb.com/manual/reference/operator/aggregation/not/) just provides a syntax with a single expression in the $not.
Does specifying multiple expressions also work?

julianedwards: This is a single expression since it is an array of bson.M of length 1.

Kimchelly: I don't think the aggregation form of `$not` can be combined with `$match` - I tried it in the mongo shell but it say it's invalid syntax. I think you can only use the non-aggregation form of `$not` in the `$match` query.

Using `$not` in place of `$and` produces an error.
```
db.c.aggregate([{"$match": {"$not": [{"foo": 1}]}}])

uncaught exception: Error: command failed: {
	"operationTime" : Timestamp(1631634685, 1),
	"ok" : 0,
	"errmsg" : "unknown top level operator: $not",
	"code" : 2,
	"codeName" : "BadValue",
	"$clusterTime" : {
		"clusterTime" : Timestamp(1631634685, 1),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	}
}
```
However, this usage of `$not` is valid:
```
db.c.aggregate([{"$match": {"foo": {"$not": {"$eq": 1}}}])
```

Can we revert the `$not` change and just use the `$and` thing that you used previously?

ybrill: ah, got it. Sorry: I got confused counting brackets 😄

ybrill (approved):

model/stats/db.go
ybrill: Nice catch @Kimchelly. Reverting my suggestion is fine by me.

Kimchelly (approved):

PR #5023, opened 2021-09-13
EVG-5312 decommission hosts running task groups carefully

ablack12:
[EVG-5312](https://jira.mongodb.org/browse/EVG-5312)

### Description 
We've had a few build failures where a host is decommissioned in-between running tasks in a task group, and then later tasks in the group fail.

I wrote drawdown and termination to be slightly different because I think drawdown is meant more to deallocate when we don't need things (so it's safe to just not deallocate), whereas for termination we may truly want to terminate the host, so in this case I just mark the task group for reset rather than preventing the termination of the host. We can try and implement them more similarly though if this doesn't seem reasonable.

### DIFF
--- a/units/host_termination.go
+++ b/units/host_termination.go
@@ -10,6 +10,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
+	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/amboy"
 	"github.com/mongodb/amboy/dependency"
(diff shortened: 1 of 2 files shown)


units/host_termination.go
```diff
@@ -180,9 +181,29 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 		} else {
 			return
 		}
+	} else {
+		// Consider if the host is in between running a single-host task group
+		if j.host.LastGroup != "" {
```
hhoke, line 186: It looks like this only gets set in `ClearRunningAndSetLastTask(`, so I'm assuming this is to avoid that situation we had yesterday in evergreen-users.

hhoke (approved):

units/host_termination.go
ablack12: It's actually to avoid a build failure with task groups, not user hosts. This is set for EndTask for every task. https://github.com/evergreen-ci/evergreen/blob/763cae970ea8b113126184a5916c5124859fc581/model/host/host.go#L1154

PR #5024, opened 2021-09-14
EVG-14958 Verify validate project configs

bynn:
[EVG-14958](https://jira.mongodb.org/browse/EVG-14958)

### Description 
validate to use local read

### Testing 
include file doesn't exist
image

include file has duplicate build variant 
image

function doesn't exist
image

duplicate uncommitted file
image

### DIFF
--- a/operations/validate.go
+++ b/operations/validate.go
@@ -6,10 +6,12 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/validator"
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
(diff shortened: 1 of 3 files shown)


ablack12 (approved):

ablack12 (changes requested):
oops sorry i button mashed that somehow

bynn:
this seems functional right now but is there a reason why i cant do this completely locally by also moving the validator commands from api.go?

ablack12 (approved):
You're right that this CLI command doesn't do anything that can't be run locally, however we have an upcoming project to improve the validator in ways that _will_ require some database lookup, which can't be done locally. That being said we may still want to refactor this eventually to send the project ref to the client side, since marshalling/unmarshalling yaml is expensive and this won't be the typical use case. cc @hadjri 

Also remember to increment  the client version string.

PR #5025, opened 2021-09-14
EVG-15330 Split out tasklog resolvers into individual resolvers

khelif96:
[EVG-15330](https://jira.mongodb.org/browse/EVG-15330)

### Description 
With the previous setup if there was a failure fetching any logs (system, agent, task logs, event logs) the entire resolver would fail and return nothing. This refactors that resolver so that each log type is fetched in its own resolver. And if there is a failure it would be constrained to the resolver. It would also reduce any over-fetching we do on the UI since we can now only conditionally fetch each log type when the user requests it

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -73,6 +73,9 @@ func (r *Resolver) Project() ProjectResolver {
 func (r *Resolver) Annotation() AnnotationResolver {
 	return &annotationResolver{r}
 }
+func (r *Resolver) TaskLogs() TaskLogsResolver {
+	return &taskLogsResolver{r}
+}
 
 func (r *Resolver) ProjectSettings() ProjectSettingsResolver {
 	return &projectSettingsResolver{r}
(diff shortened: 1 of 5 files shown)


graphql/resolvers.go
```diff
@@ -1671,13 +1675,84 @@ func (r *queryResolver) TaskFiles(ctx context.Context, taskID string, execution
...
+	}
+	systemLogPointers := []*apimodels.LogMessage{}
+	for i := range systemLogs {
+		systemLogPointers = append(systemLogPointers, &systemLogs[i])
+	}
+
+	return systemLogPointers, nil
+}
+func (r *taskLogsResolver) EventLogs(ctx context.Context, obj *TaskLogs) ([]*restModel.TaskAPIEventLogEntry, error) {
 	const logMessageCount = 100
```
arjunpatel100, line 1750: it would be nice if we could share this variable between the resolvers so it's only set in one place

arjunpatel100 (approved):

graphql/resolvers.go
khelif96: Only think i can think is to pass it in the response to share it with the other resolvers. :/

PR #5027, opened 2021-09-14
EVG-15355 continue adding project admins on error

bynn:
[EVG-15355](https://jira.mongodb.org/browse/EVG-15355)

### Description 
does not return on error while adding admins

### Testing 
unit test

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1885,22 +1885,30 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
 		viewRole = GetViewRepoRole(p.RepoRefId)
 	}
 
+	catcher := grip.NewBasicCatcher()
 	for _, addedUser := range toAdd {
 		adminUser, err := user.FindOneById(addedUser)
 		if err != nil {
-			return errors.Wrapf(err, "error finding user '%s'", addedUser)
+			catcher.Wrapf(err, "error finding user '%s'", addedUser)
+			continue
 		}
 		if adminUser == nil {
-			return errors.Errorf("no user '%s' found", addedUser)
+			catcher.Errorf("no user '%s' found", addedUser)
+			continue
 		}
 		if err = adminUser.AddRole(role.ID); err != nil {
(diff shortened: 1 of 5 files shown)


model/project_ref.go
```diff
@@ -1885,23 +1885,31 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
...
 		if viewRole != "" {
 			if err = adminUser.AddRole(viewRole); err != nil {
-				return errors.Wrapf(err, "error adding role %s to user %s", viewRole, addedUser)
+				catcher.Wrapf(err, "error adding role %s to user %s", viewRole, addedUser)
+				continue
 			}
 		}
 	}
+	if err = catcher.Resolve(); err != nil {
+		return errors.Wrap(err, "error adding users")
```
ablack12, line 1911: let's say "error adding some users" since any users that aren't in this error _were_ added

ablack12 (changes requested):
We'll want to extend this change to the toRemove users below, and also to the repo ref method: https://github.com/evergreen-ci/evergreen/blob/14bc0f57bbb164f4377bb26e767fdfcde52953a5/model/repo_ref.go#L234

model/project_ref.go

ablack12 (changes requested):
Hi sorry to make this more complicated (and theoretically this was actually already wrong): if we're adding/removing roles for some users, we _should_ update the existing admin list (see https://github.com/evergreen-ci/evergreen/blob/dc484d82582b2fd4dc151054a2d06c41fc9fb077/rest/data/project.go#L142: we want to get down to SaveProjectPageForSection so we save the admin list, and then log the project as modified), even though we errored for some users.

This means that we have UpdateAminRoles should probably remove from p.Admins any users it failed to add, and similarly add to p.Admins any users it failed to remove. Then we can continue on error in SaveProjectSettingsForSection and save the admins list in the state that it currently is in. Let me know if you have questions with this.

ablack12 (approved):
This looks good and we can fix the other discrepancies we discussed in the future ticket

PR #5029, opened 2021-09-15
EVG-15052 create copyProject function for routes and UI

ablack12:
[EVG-15052](https://jira.mongodb.org/browse/EVG-15052)

### Description 
I'm torn on the best way to go about this. All of the functionality for copy project already exists for the rest route, but across multiple data connectors. It would be nice to have a function that covered as much crossover as possible, so that the route and graphql can share it. Graphql didn't seem like a neutral package so I created a new file in rest/data (although this is kind of cheating, I can't think of a better spot for it). I also only included the pieces that we could lump under "InternalServiceError".

### Testing 
Passes existing copy tests.


ablack12:
If we like this approach, I can also probably refactor SaveProjectBySection to be in data/util.go and use all the connectors

malikchaya2 (approved):
I'm a fan of this setup, and I think having rest/data/util.go will be useful for future use cases as well.

ablack12:
Created https://jira.mongodb.org/browse/EVG-15420

PR #5031, opened 2021-09-15
EVG-15277 Fix github pr childpatch link

malikchaya2:
[EVG-15277](https://jira.mongodb.org/browse/EVG-15277)

### Description 
This links a child patch status to the parent downstream tasks tab instead of the child patch's own page

### Testing 
[tested on staging](https://github.com/evergreen-ci/commit-queue-sandbox/pull/327)


trigger/payloads.go
```diff
@@ -470,8 +470,11 @@ func buildLink(uiBase string, buildID string, hasPatch bool) string {
 	return url
 }
 
-func versionLink(uiBase string, versionID string, hasPatch bool) string {
+func versionLink(uiBase string, versionID string, hasPatch bool, isChild bool) string {
```
ablack12, line 473: It's not great to have multiple booleans in a row for function signatures, since it's easy to get these mixed up, so it's probably worth an input struct now.

malikchaya2: true, fixed.

trigger/payloads.go
```diff
@@ -453,9 +453,19 @@ func buildLink(uiBase string, buildID string, hasPatch bool) string {
...
+	uiBase    string
+	versionID string
+	hasPatch  bool
+	isChild   bool
+}
+
+func versionLink(i versionLinkInput) string {
+	url := fmt.Sprintf("%s/version/%s", i.uiBase, url.PathEscape(i.versionID))
+	if i.isChild {
+		url += "/downstream-tasks"
```
ablack12, line 466: Does this only exist on spruce? Do we need to do anything intelligent about not doing this if the user isn't on spruce? (I guess this kind of version does need to be on the new UI though).

ablack12 (approved):
LGTM with question but not blocking

trigger/payloads.go
malikchaya2: This only exists on Spruce if we want to show it as "downstream tasks", as an extension of the main patch. I guess we can technically show the child patch itself decoupled from the main patch that kicked it off if the user is not on Spruce, but I think always leading to Spruce probably makes more sense. I worry that showing the child patch in a legacy patch page may confuse the user, and they may accidentally think that they are viewing the main patch and make decisions based on that, so I would caution against that approach.

ablack12: That makes sense; since this is a pretty special case I'm fine with /downstream-tasks just not working in legacy right now and hopefully that'll prompt users to switch to Spruce (if they aren't already opting in). If it becomes a problem in the future we can always do something to redirect them or explain why the page doesn't work for legacy.

PR #5035, opened 2021-09-17
EVG-14024: Check task secret for additional routes

hadjri:
[EVG-14024](https://jira.mongodb.org/browse/EVG-14024)

### Description 
Some routes in api.go can unexpectedly be curled without any authentication needed.  Change has been made to check the task secret for routes that were not checking any host/task secret before and can potentially return sensitive data.
### Testing 
Deployed to staging and tested these routes locally on tasks which I have permission for without any error.


ablack12 (approved):
We should also do this with s3Copy I believe (we can see that the secret is passed in here: https://github.com/evergreen-ci/evergreen/blob/83945041f84a60db700e61e3b1fd09c08860ec37/agent/command/s3_copy.go#L172 which verifies that the change is safe.

PR #5036, opened 2021-09-17
EVG-15385: Attach trigger name to patch

sophstad:
[EVG-15385](https://jira.mongodb.org/browse/EVG-15385)

### Description
- Add `SetTriggerAliases` field to ensure that we are logging the alias associated with each child patch (the `ChildPatches` and `Aliases` field within [`TriggerInfo`](https://github.com/evergreen-ci/evergreen/blob/main/model/patch/patch.go#L193) are zipped/have 1 to 1 correspondence)
- Add `childPatchAliases` GraphQL field to patch to return trigger name with each child patch ID

### Testing 
- Add DB and GraphQL unit tests

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -419,6 +419,10 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version, 
 			if err = p.SetChildPatches(); err != nil {
 				return errors.Wrapf(err, "error attaching child patches '%s'", p.Id.Hex()), http.StatusInternalServerError, "", ""
 			}
+			p.Triggers.Aliases = patchUpdateReq.PatchTriggerAliases
+			if err = p.SetTriggerAliases(); err != nil {
+				return errors.Wrapf(err, "error attaching trigger aliases '%s'", p.Id.Hex()), http.StatusInternalServerError, "", ""
+			}
 		}
 
 		requester := p.GetRequester()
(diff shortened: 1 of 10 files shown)


graphql/util.go
```diff
@@ -419,6 +419,10 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
 			if err = p.SetChildPatches(); err != nil {
 				return errors.Wrapf(err, "error attaching child patches '%s'", p.Id.Hex()), http.StatusInternalServerError, "", ""
 			}
+			p.Triggers.Aliases = patchUpdateReq.PatchTriggerAliases
```
malikchaya2, line 422: I thought we did this in a previous pr. did we do something similar somewhere else? It's not important, I'm just wondering....

model/patch/patch.go
```diff
@@ -578,6 +578,17 @@ func (p *Patch) SetActivated(ctx context.Context, versionId string) error {
 	return err
 }
 
+// SetAliases appends the names of invoked trigger aliases to the DB
```
malikchaya2, line 581: nit: SetTriggerAliases

malikchaya2 (approved):
Approved with super slight nit and random question

graphql/util.go
sophstad: Almost! We set `p.Triggers.ChildPatches` [here](https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go#L525), added in https://github.com/evergreen-ci/evergreen/pull/4947. So this will ensure the two fields are properly zipped!

malikchaya2: ah yes!

PR #5037, opened 2021-09-17
EVG-15360 rename FindOneProjectRef to be branch-specific, and audit usage

ablack12:
[EVG-15360](https://jira.mongodb.org/browse/EVG-15360)

### Description 
FindOneProjectRef is often used when FindMergedProjectRef should be used, just because the function is habit. Renamed to FindBranchProjectRef to encourage developers to consider which function they really need. Audited usage of FindOneProjectRef for places where we should really be using FindMergedProjectRef.

### Testing 
Ensured existing tests are unbroken.


model/task_lifecycle.go
```diff
@@ -96,13 +96,6 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		}
 
 		if t.Requester == evergreen.MergeTestRequester {
-			projRef, err := FindOneProjectRef(t.Project)
```
hadjri, line 99: Why did these checks get removed instead of changing the method name to FindBranchProjectRef here?

rest/route/project.go
```diff
@@ -979,8 +979,8 @@ func (p *GetPatchTriggerAliasHandler) Parse(ctx context.Context, r *http.Request
 }
 
 func (p *GetPatchTriggerAliasHandler) Run(ctx context.Context) gimlet.Responder {
-	proj, err := dbModel.FindOneProjectRef(p.projectID)
-	if err != nil {
+	proj, err := dbModel.FindMergedProjectRef(p.projectID)
+	if err != nil || proj == nil {
```
hadjri, line 983: Maybe it would be preferable to case errors separately on err != nil and proj == nil with different messages to be more clear.

service/api_plugin_manifest.go
```diff
@@ -18,12 +18,16 @@ import (
...
+	projectRef, err := model.FindMergedProjectRef(task.Project)
 	if err != nil {
-		as.LoggedError(w, r, http.StatusBadRequest,
-			errors.Wrapf(err, "projectRef not found for project %s", task.Project))
+		as.LoggedError(w, r, http.StatusInternalServerError,
+			errors.Wrapf(err, "error finding project '%s'", task.Project))
 		return
 	}
+	if projectRef == nil {
+		as.LoggedError(w, r, http.StatusBadRequest,
```
hadjri, line 28: Should this be a 404 instead?

service/api_task.go
```diff
@@ -504,7 +504,7 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
 		}
 
 		projectRef, err := model.FindMergedProjectRef(nextTask.Project)
-		if err != nil {
+		if err != nil || projectRef == nil {
```
hadjri, line 507: Same comment as above

graphql/util.go
```diff
@@ -367,7 +367,7 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
 	// can't interrupt the db operations here
 	newCxt := context.Background()
 
-	projectRef, err := model.FindOneProjectRef(project.Identifier)
+	projectRef, err := model.FindMergedProjectRef(project.Identifier)
```
hadjri, line 370: Just for my clarity can you explain what scenarios call for returning a proj ref merged with repo settings vs. just getting the branch level settings?

hadjri (changes requested):

graphql/util.go
hadjri: Thanks!

service/api_plugin_manifest.go
hadjri: SGTM

hadjri (approved):

PR #5038, opened 2021-09-17
EVG-14930 Collapse unmatching commits along side active ones

khelif96:
[EVG-14930](https://jira.mongodb.org/browse/EVG-14930)

### Description 
Collapse inactive commits along side active ones. I figured out how to go forward but haven't gotten as far as thinking about navigating to the previous page.  This is an expensive calculation that involves checking every task for every activated version to determine if we should fold it up or not. 

https://user-images.githubusercontent.com/4605522/133855079-302a3470-f584-4f9c-a990-49ae724c9db3.mov

### Testing 
 On staging


graphql/resolvers.go
```diff
@@ -3262,8 +3273,50 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+				return versions[i].RevisionOrderNumber > versions[j].RevisionOrderNumber
+			})
+		}
+	}
+	// Gather some analytics data
+	endTime := time.Now()
+	grip.Info(message.Fields{
+		"message":                 "mainlineCommits",
+		"commitsChecked":          len(versions),
+		"timeTaken":               endTime.Sub(startTime).String(),
```
arjunpatel100, line 3316: Our resolvers offer an extension to show query duration in the http response. Do you think that could sufficient instead of logging additional details?

arjunpatel100 (approved):

graphql/resolvers.go
khelif96: We don't actually log that information anywhere, Its only used when testing locally. I wanted to log this additional information so i can understand how varying amounts of commits affect query performance and scalability.

arjunpatel100 (approved):

graphql/resolvers.go
```diff
@@ -3231,8 +3233,18 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 		}
 		mainlineCommitVersion := MainlineCommitVersion{}
 
+		shouldNotBeRolledUp := true
```
ablack12, line 3236: can we move this whole thing to be in an utility.FromBoolPtr(v.Activated) loop rather than checking individually at 3238 and 3247? that would also make it shorter to combine if statements at 3237.

graphql/resolvers.go
```diff
@@ -3262,8 +3273,50 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+			} else {
+				skipOrderNumber = utility.FromIntPtr(mainlineCommits.NextPageOrderNumber)
+			}
+
+			opts = model.MainlineCommitVersionOptions{
+				Activated:       true,
+				Limit:           limit,
+				SkipOrderNumber: skipOrderNumber,
+			}
+			activatedVersions, err = model.GetMainlineCommitVersionsWithOptions(projectId, opts)
```
ablack12, line 3293: Should we be checking hasMatchingTask for these?

graphql/resolvers.go
```diff
@@ -3231,8 +3233,18 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 		}
 		mainlineCommitVersion := MainlineCommitVersion{}
 
+		shouldNotBeRolledUp := true
+		if buildVariantOptions != nil {
+			if utility.FromBoolPtr(v.Activated) && (len(buildVariantOptions.Statuses) != 0 || len(buildVariantOptions.Tasks) != 0 || len(buildVariantOptions.Variants) != 0) {
+				activatedCommitsCheckedCount++
```
ablack12, line 3239: Is this just for debugging purposes? I think it might be easier to move all of the length checks into `hasMatchingTasks` to simplify things

khelif96: Yeah it is, This collects the amount of activated mainline commits we have checked. I think its better to keep them separate since the values aren't really related. The only way they are related is that `hasMatchingTasks` coincidentally gets run when ever we increment this value.

PR #5039, opened 2021-09-17
EVG-15051 Create a createProject function

malikchaya2:
[EVG-15051](https://jira.mongodb.org/browse/EVG-15051)

### Description 
This puts CreateProject in a place where it can be shared by the resolvers and routes. 

### Testing 
Passes existing tests.

### DIFF
--- a/rest/data/util.go
+++ b/rest/data/util.go
@@ -48,3 +48,14 @@ func CopyProject(ctx context.Context, sc Connector, projectToCopy *model.Project
 	}
 	return apiProjectRef, nil
 }
+
+func CreateProject(ctx context.Context, sc Connector, projectRef *model.ProjectRef) (*model.ProjectRef, error) {
+	u := gimlet.GetUser(ctx).(*user.DBUser)
+
+	if err := sc.CreateProject(projectRef, u); err != nil {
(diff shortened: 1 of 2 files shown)


rest/data/util.go
```diff
@@ -48,3 +48,14 @@ func CopyProject(ctx context.Context, sc Connector, projectToCopy *model.Project
 	}
 	return apiProjectRef, nil
 }
+
+func CreateProject(ctx context.Context, sc Connector, projectRef *model.ProjectRef) (*model.ProjectRef, error) {
```
ablack12, line 52: I don't think we need this. Can't the route and the resovler already just use sc.CreateProject? Really the only bit that would be duplicated is getting the user. The reason this was necesssary for CopyProject is because it used multiple different kind of connectors.

ablack12: I think the thing that _can_ be done for this ticket is moving the getting and validating of the project input into CreateProject, but again we shouldn't need a new function here for it since we only need the ProjectConnector.

malikchaya2: yeah, I was questioning wether we need this as I was writing it. Since you're having the same though, I'll close it.

PR #5041, opened 2021-09-20
EVG-15436: Do not store URL in xunit test results

julianedwards:
[EVG-15436](https://jira.mongodb.org/browse/EVG-15436)

We should not store internal test log URLs for test results since we generate them [here](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L341).

### DIFF
--- a/agent/command/results_xunit_parser.go
+++ b/agent/command/results_xunit_parser.go
@@ -139,8 +139,6 @@ func (tc testCase) toModelTestResultAndLog(conf *internal.TaskConfig) (task.Test
 		log.Task = conf.Task.Id
 		log.TaskExecution = conf.Task.Execution
 
-		// update the URL of the result to the expected log URL
-		res.URL = log.URL()
 		res.LogTestName = log.Name
 	}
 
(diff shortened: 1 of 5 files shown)


arjunpatel100 (approved):

Kimchelly (commented):
This change seems fine, I have just a couple questions.
1. Does this affect legacy xunit test results at all? Or are all projects that use xunit also using new test results at this point?
2. If we're not storing the URL anymore, is keeping the BSON tag on the URL field still necessary (or is it necessary for backwards compatibility)?
3. Can we delete the `(*TestLog).URL()` method?

julianedwards:
> This change seems fine, I have just a couple questions.
> 
> 1. Does this affect legacy xunit test results at all? Or are all projects that use xunit also using new test results at this point?
> 2. If we're not storing the URL anymore, is keeping the BSON tag on the URL field still necessary (or is it necessary for backwards compatibility)?
> 3. Can we delete the `(*TestLog).URL()` method?

1. No, this will not affect legacy test results. The only reason we were storing the URL field was so the UI would know when to hide the log link. We can no longer do that since we do not know if logs are in cedar until we hit the route (though maybe we could add a field to cedar test results, `has_logs` or something, but that is a future thing). We generate logs lazily now, so even if xunit tests are stored in legacy, their log URLs will be generated correctly.

2. Necessary for external logs (such as legacy resmoke storing logs in resmoke).

3. Yes, done.

I also added an if clause that will prepend the evg root to internal logs stored in the URL field.

Kimchelly (approved):

PR #5044, opened 2021-09-21
CORPSEC-6283

hhoke:
[CORPSEC-6283](https://jira.mongodb.org/browse/CORPSEC-6283)

### Description 
Double slash seems to have been causing an open redirect issue, see ticket. We now make sure that the path does not have any leading `//`.

### Testing 
  Deployed to staging, problem URL in ticket no longer redirects to external site.


hhoke:
see https://github.com/urfave/negroni/pull/259

PR #5045, opened 2021-09-21
EVG-15341: Return 404 when error encountered finding merged project ref

hadjri:
[EVG-15341](https://jira.mongodb.org/browse/EVG-15341)

### Description 
Currently all routes that use urlVarsToProjectScopes in the middleware such as this [endpoint](https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-versions-for-a-project) return a 500 internal server error when a merged project ref is unable to be found. Change has been made to return a 404 in this case instead.
### Testing 
Deployed to staging and confirmed that the route now returns 404 for an invalid project.  Attached screenshot.

### DIFF
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -631,7 +631,7 @@ func urlVarsToProjectScopes(r *http.Request) ([]string, int, error) {
 
 	projectRef, err := model.FindMergedProjectRef(projectID)
 	if err != nil {
-		return nil, http.StatusInternalServerError, errors.WithStack(err)
+		return nil, http.StatusNotFound, errors.WithStack(err)
 	}
 	if projectRef == nil {
 		return nil, http.StatusNotFound, errors.Errorf("error finding the project '%s'", projectID)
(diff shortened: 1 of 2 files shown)


hadjri:
Screen Shot 2021-09-21 at 12 51 33 PM

rest/route/middleware.go
```diff
@@ -631,7 +631,7 @@ func urlVarsToProjectScopes(r *http.Request) ([]string, int, error) {
 
 	projectRef, err := model.FindMergedProjectRef(projectID)
 	if err != nil {
-		return nil, http.StatusInternalServerError, errors.WithStack(err)
+		return nil, http.StatusNotFound, errors.WithStack(err)
```
ablack12, line 634: I actually have PR up (although I need to fix a test) to change FindMergedProjectRef to return `nil, nil` if not found, instead of an error, which is more consistent. In that case, this _should_ be http.StatusInternalServerError, but since that PR is going to take some time I can just change this back in that PR once this is merged.

ablack12 (approved):

PR #5046, opened 2021-09-21
EVG-15356: Send log to splunk when user offboards

hadjri:
[EVG-15356](https://jira.mongodb.org/browse/EVG-15356)

### Description 
There is currently no record in Splunk when we offboard users.  Added logging to track who we offboard and when.

### DIFF
--- a/rest/route/host.go
+++ b/rest/route/host.go
@@ -400,6 +400,13 @@ func (ch *offboardUserHandler) Run(ctx context.Context) gimlet.Responder {
 	opts := model.APIHostParams{
 		UserSpawned: true,
 	}
+	if !ch.dryRun {
+		grip.Info(message.Fields{
+			"message": "executing user offboarding",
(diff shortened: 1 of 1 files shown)


rest/route/host.go
```diff
@@ -400,6 +400,13 @@ func (ch *offboardUserHandler) Run(ctx context.Context) gimlet.Responder {
 	opts := model.APIHostParams{
 		UserSpawned: true,
 	}
+	if !ch.dryRun {
+		grip.Info(message.Fields{
+			"message": "executing user offboarding",
+			"context": "user offboarding",
```
ablack12, line 406: Context here just seems to be a repeat of message so I'm not sure that's useful. It might be worth logging this at 438 instead so we can also log the `toTerminate` response.

ablack12 (changes requested):

ablack12 (approved):

PR #5047, opened 2021-09-21
EVG-15442: Attach tasks/variants to alias

sophstad:
[EVG-15442](https://jira.mongodb.org/browse/EVG-15442)

### Description 
- Fetch the variants/tasks that would be invoked by a given patch trigger alias

### Testing 
- Added GraphQL unit test and tested in staging playground


graphql/schema.graphql
```diff
@@ -532,11 +532,17 @@ type ChildPatchAlias {
 type PatchTriggerAlias {
   alias: String!
   childProject: String!
-  taskSpecifiers: [TaskSpecifier]
-  status: String
-  parentAsModule: String
+  variantsTasks: [VariantTask]!
 }
 
+type PatchTriggerDefinition {
```
ablack12, line 538: I don't love that we have two different data structures to represent the same thing, just different fields. Can we combine these into one and just populate the fields we need? (only populate variant tasks for this patch resolver?)

rest/model/patch.go
```diff
@@ -73,6 +74,12 @@ type APIChildPatchAlias struct {
 	PatchID *string `json:"patch_id"`
 }
 
+type APIPatchTriggerAlias struct {
```
ablack12, line 77: Do we need to create an API type for something that is only being used in the UI? Mohamed would know better but I would think we could just use a graphql type instead.

graphql/resolvers.go
```diff
@@ -1137,18 +1137,39 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
...
 	}
 
-	aliases := []*restModel.APIPatchTriggerDefinition{}
-	for _, alias := range project.PatchTriggerAliases {
-		aliases = append(aliases, &restModel.APIPatchTriggerDefinition{
-			Alias:        utility.ToStringPtr(alias.Alias),
-			ChildProject: utility.ToStringPtr(alias.ChildProject)})
+	project := &model.Project{}
+	opts := model.GetProjectOpts{}
+	if _, err = model.LoadProjectInto(ctx, []byte(*obj.PatchedConfig), opts, *obj.ProjectId, project); err != nil {
```
ablack12, line 1148: Parsing yaml can be really expensive; if `len(projectRef.PatchTriggerAliases) == 0` we should return early so we don't need to go through this.

ablack12 (changes requested):

rest/model/patch.go
sophstad: Good point, however I think I'll go ahead with combining the fields like you suggested above. Perhaps in that case it's appropriate to maintain `APIPatchTriggerDefinition` as an API type as you configured it in https://github.com/evergreen-ci/evergreen/pull/4995?

graphql/resolvers.go
```diff
@@ -1138,17 +1138,42 @@ func (r *patchResolver) ID(ctx context.Context, obj *restModel.APIPatch) (string
...
+
+	project := &model.Project{}
+	opts := model.GetProjectOpts{}
+	if _, err = model.LoadProjectInto(ctx, []byte(*obj.PatchedConfig), opts, *obj.ProjectId, project); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error unmarshaling project config: %v", err.Error()))
+	}
+
 	aliases := []*restModel.APIPatchTriggerDefinition{}
-	for _, alias := range project.PatchTriggerAliases {
+	for _, alias := range projectRef.PatchTriggerAliases {
```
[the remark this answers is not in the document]
sophstad: We do not know the child patch IDs at this point! The use case for this ticket is to show users a list of the variants/tasks that will be run if they invoke an alias from the UI configure page (i.e. giving them more information so they can decide whether to run it). Hence the only information we have available at this point is the alias name. Let me know if that answers your question!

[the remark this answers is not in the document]
ablack12: That makes sense but I think we are still using the wrong project. We'll need to find latest version with valid project for each of the child projects. 
Using your test to illustrate what I'm thinking: you used the task specifier "lint" which does exist for both Evergreen and Spruce, so this works even though we're using the Evergreen parent project to choose tasks/variants with. If we were to use a task that _only_ exists for Spruce, such as e2e_test, we wouldn't find this matching task in the Evergreen project, only in the Spruce project.

graphql/resolvers.go
ablack12: If this is only used for the Configure Page (I think it is, Sophie would know) then I imagine it's not a concern since it would only increase the time for that one page, right? 
Otherwise if we also need this for the regular patch page, if the patch has been finalized we can read the project from the parser project collection, which is less expensive. I think this isn't the scenario though.

graphql/resolvers.go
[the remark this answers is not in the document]
ablack12: Do you think there's ever a use case to have the same project defined for multiple aliases? I'm imagining no, but if it does seem like something that would be done, we should keep a map of project ID to project outside of this loop so we can cache projects and not look them up if we've already done so.

graphql/resolvers.go
sophstad: Yes, it's used only for the configure page. This data will not be needed anywhere else because once the child patch is scheduled, its variants/tasks are available via the Patch resolver's `childPatches` field.

I agree that this could be expensive, but I'm not very in tune with the usage patterns of the configure page. It seems to me like this may be something that a user looks at once, schedules the patch, and doesn't look at again. In that case the caching may not be particularly helpful, but it's tough to speculate. Would really appreciate any other insights y'all have!

trigger/payloads_test.go
```diff
@@ -121,7 +121,7 @@ func (s *payloadSuite) TestEvergreenWebhook() {
 	s.NoError(err)
 	s.Require().NotNil(m)
 
-	s.Len(m.Body, 584)
+	s.Len(m.Body, 601)
```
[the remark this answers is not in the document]
ablack12: Yeah that's the gist; it's been around long enough that i guess it's never been worth fixing but it it is a little ridiculous 😅

rest/model/project.go
```diff
@@ -79,6 +79,7 @@ type APIPatchTriggerDefinition struct {
 	TaskSpecifiers []APITaskSpecifier `json:"task_specifiers"`
 	Status         *string            `json:"status,omitempty"`
 	ParentAsModule *string            `json:"parent_as_module,omitempty"`
+	VariantsTasks  []VariantTask      `json:"variants_tasks"`
```
ablack12, line 82: nit: can we add omitempty since this won't be common?

graphql/resolvers.go
ablack12: Sorry our comments kind of raced in a weird way 😂  The configure page isn't used often, and most projects won't define patch trigger aliases, so I personally think a mild slowdown on this page is fine.

graphql/resolvers.go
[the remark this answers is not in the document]
sophstad: Good point! I've discussed this with Chaya before and while we don't envision it as a common use case, it's permissible and could potentially be useful. I will add this!

graphql/resolvers.go
khelif96: Cool SGTM. Thanks for the explanation

graphql/resolvers.go
```diff
@@ -1147,11 +1147,16 @@ func (r *patchResolver) PatchTriggerAliases(ctx context.Context, obj *restModel.
...
 	}
 
+	projectCache := map[string]*model.Project{}
 	aliases := []*restModel.APIPatchTriggerDefinition{}
 	for _, alias := range projectRef.PatchTriggerAliases {
-		_, project, err := model.FindLatestVersionWithValidProject(alias.ChildProject)
-		if err != nil {
-			return nil, InternalServerError.Send(ctx, errors.Wrapf(err, "Problem getting last known project for '%s'", alias.ChildProject).Error())
+		project, projectCached := projectCache[alias.ChildProject]
+		if projectCached == false {
```
ablack12, line 1154: nit: `!projectCached`

ablack12 (approved):
LGTM with nit 🚀

PR #5049, opened 2021-09-22
EVG-15203: Add nil check for project parser on ShouldDeactivatePrevious

hadjri:
[EVG-15203](https://jira.mongodb.org/browse/EVG-15203)

### Description 
Addressing the [following alert](https://mongodb.slack.com/archives/C866SR2LR/p1632332425051500) caused by a panic created when project parser was nil.


service/api_task.go
```diff
@@ -229,7 +229,7 @@ func (as *APIServer) EndTask(w http.ResponseWriter, r *http.Request) {
 
 	// mark task as finished
 	deactivatePrevious := projectRef.ShouldDeactivatePrevious()
-	if projectParser.ShouldDeactivatePrevious() {
+	if projectParser != nil && projectParser.ShouldDeactivatePrevious() {
```
ablack12, line 232: I'm not sure we should ever be getting here (although this is still a reasonable check). I'm thinking what's actually happening is that we're erroring on 208 but not returning.  Should we be returning? It looks like we _are_ hitting errors but I would think that we shouldn't be. We may want to add a grip log to ParserProjectFindOneVersion to detail if we're using the passed in version or looking one up, and what the version ID is.
https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen%20metadata.level%3E%3D70%20500&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=1632331860&latest=1632333686&workload_pool=standard_perf&sid=1632333699.411703

model/project_parser_db.go
```diff
@@ -55,13 +55,21 @@ func ParserProjectFindOneById(id string) (*ParserProject, error) {
...
 	if version == "" {
 		lastGoodVersion, err := FindVersionByLastKnownGoodConfig(projectId, -1)
 		if err != nil || lastGoodVersion == nil {
 			return nil, errors.Wrapf(err, "Unable to retrieve last good version for project '%s'", projectId)
 		}
 		version = lastGoodVersion.Id
+		msg["lookup_version"] = true
 	}
+	msg["version"] = version
+	grip.Info(msg)
```
ablack12, line 72: Let's only log for error (and log under grip.Debug) or else we'll be logging like crazy.

ablack12 (approved):
LGTM after logging fix

PR #5050, opened 2021-09-22
EVG-15225: remove single quotes from agent curl command for Windows

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15225

### Description
* Remove quotation marks from URL being curled because Windows batch scripts don't allow single quotes. The quotes shouldn't be necessary for the bash/batch script to have valid syntax.
* Fix [a test](https://github.com/evergreen-ci/evergreen/pull/5050/files#diff-2567873892e2fae0771c76fc6d952f51827126e55f8840e253c9e26736268d20R22) that breaks if the `host.running_tasks` unique index exists (which is created by [this test](https://github.com/evergreen-ci/evergreen/blob/0ac37d2251e9fa6833e214944c77058955ea43f3/units/host_monitoring_idle_termination_test.go#L45)). @hhoke 

### Testing 
Updated the unit tests.


hhoke (approved):
LGTM

PR #5051, opened 2021-09-22
EVG-15321 modify delete user route to require resource ID

ablack12:
[EVG-15321](https://jira.mongodb.org/browse/EVG-15321)

### Description 
We made this route for Mana but they were told that we would accept a resource ID, and apparently we don't, which means that no matter what we're deleting a ton of resources for users. We should only delete roles relevant to the resource, and we should ignore the basic roles so that the user can remain a basic user of Evergreen.

### Testing 
Fixed unit test.

### DIFF
--- a/rest/route/user.go
+++ b/rest/route/user.go
@@ -269,6 +269,10 @@ func (h *userPermissionsDeleteHandler) Run(ctx context.Context) gimlet.Responder
 	}
 
 	rolesForResource, err := h.rm.FilterForResource(rolesToCheck, h.resourceId, h.resourceType)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(
+			errors.Wrapf(err, "unable to filter user roles for resource ID '%s'", h.resourceId))
+	}
 	rolesToRemove := []string{}
 	for _, r := range rolesForResource {
 		rolesToRemove = append(rolesToRemove, r.ID)
--- a/rest/route/user.go
+++ b/rest/route/user.go
@@ -168,6 +168,7 @@ func (h *userPermissionsPostHandler) Run(ctx context.Context) gimlet.Responder {
 
 type deletePermissionsRequest struct {
 	ResourceType string `json:"resource_type"`
+	ResourceId   string `json:"resource_id"`
 }
 
 const allResourceType = "all"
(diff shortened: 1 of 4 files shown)


rest/route/user.go
```diff
@@ -245,27 +257,33 @@ func (h *userPermissionsDeleteHandler) Run(ctx context.Context) gimlet.Responder
...
+		}
+	}
+	if len(rolesToCheck) == 0 {
+		gimlet.NewJSONResponse(struct{}{})
 	}
-	applicableScopes, err := h.rm.FilterScopesByResourceType(scopesIds, h.resourceType)
+
+	rolesForResource, err := h.rm.FilterForResource(rolesToCheck, h.resourceId, h.resourceType)
 	if err != nil {
-		grip.Error(message.WrapError(err, message.Fields{
```
hadjri, line 254: Why are we no longer logging to Splunk here?

rest/route/user.go
```diff
@@ -245,27 +257,33 @@ func (h *userPermissionsDeleteHandler) Run(ctx context.Context) gimlet.Responder
...
-		scopesIds = append(scopesIds, role.Scope)
+	rolesToCheck := []gimlet.Role{}
+	// don't remove basic access, just special access
+	for _, r := range roles {
+		if !utility.StringSliceContains(evergreen.BasicAccessRoles, r.ID) {
+			rolesToCheck = append(rolesToCheck, r)
+		}
+	}
+	if len(rolesToCheck) == 0 {
+		gimlet.NewJSONResponse(struct{}{})
```
hadjri, line 268: What does this line do? Do we need to specify a return statement here?

hadjri (commented):
LGTM outside of a couple clarifying questions.

rest/route/user.go
ablack12: It's a bit redundant; we already log errors from rest routes.

rest/route/user.go
ablack12: This just returns an empty successful response; we also do this at 294 at the end of the function but in this case if there are no roles to check we can skip the next query. We should probably also do this in the update function.

hadjri (approved):

PR #5052, opened 2021-09-22
EVG-15430 add route to get users for role

ablack12:
[EVG-15430](https://jira.mongodb.org/browse/EVG-15430)

### Description 
Add a route for mana to use to return all of the users that have a given role.

### Testing 
Added unit test and tested on local.


hadjri (approved):
LGTM

PR #5053, opened 2021-09-23
EVG-15384 Support pre-flight options requests for all POST routes.

khelif96:
[EVG-15384](https://jira.mongodb.org/browse/EVG-15384)

### Description 
This adds support for [pre-flight OPTION](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) requests for all POST request routes on the `/rest/v2` endpoints.

These requests are used to verify that CORS requirements are satisfied, Before the actual POST requests are sent. 

I've also added a header that should prevent the client from sending this type of request more then once every 10 minutes (Which is the max time i can set)

This is dependent on revendoring https://github.com/evergreen-ci/gimlet/pull/98 
### Testing 
Tested in staging and wrote unit tests

### DIFF
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -799,9 +799,8 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
 func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		requester := r.Header.Get("Origin")
-		grip.Debug(message.Fields{
+		grip.DebugWhen(requester != "", message.Fields{
 			"op":              "addCORSHeaders",
-			"request":         r,
 			"requester":       requester,
 			"allowed_origins": allowedOrigins,
 			"adding_headers":  util.StringContainsSliceRegex(allowedOrigins, requester),
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -19,7 +19,6 @@ import (
 	"github.com/evergreen-ci/evergreen/rest/data"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/gimlet"
-	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
@@ -802,10 +801,12 @@ func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.Handler
 		requester := r.Header.Get("Origin")
 		grip.Debug(message.Fields{
 			"op":              "addCORSHeaders",
+			"request":         r,
 			"requester":       requester,
 			"allowed_origins": allowedOrigins,
-			"adding_headers":  utility.StringSliceContains(allowedOrigins, requester),
+			"adding_headers":  util.StringContainsSliceRegex(allowedOrigins, requester),
 			"settings_is_nil": evergreen.GetEnvironment().Settings() == nil,
+			"headers":         r.Header,
 		})
 		if len(allowedOrigins) > 0 {
 			// Requests from a GQL client include this header, which must be added to the response to enable CORS
@@ -815,6 +816,7 @@ func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.Handler
 				w.Header().Add("Access-Control-Allow-Credentials", "true")
 				w.Header().Add("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PATCH, PUT")
 				w.Header().Add("Access-Control-Allow-Headers", fmt.Sprintf("%s, %s, %s", evergreen.APIKeyHeader, evergreen.APIUserHeader, gqlHeader))
+				w.Header().Add("Access-Control-Max-Age", "600")
 			}
 		}
 		next(w, r)
--- a/rest/route/options.go
+++ b/rest/route/options.go
@@ -7,7 +7,7 @@ import (
 	"github.com/evergreen-ci/gimlet"
 )
 
-// optionsHandler is a RequestHandler for resolving options requests
+// optionsHandler is a RequestHandler for resolving options requests.
 type optionsHandler struct {
 }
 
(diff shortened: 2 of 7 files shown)


rest/route/service.go
```diff
@@ -209,4 +213,36 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
...
+
+func (h *optionsHandler) Parse(ctx context.Context, r *http.Request) error {
+	return nil
+}
+
+// Execute returns an empty json response to the options request
+// The options request only looks for a 200 status code so the response value is not relevant
+func (h *optionsHandler) Run(ctx context.Context) gimlet.Responder {
+	return gimlet.NewJSONResponse(struct{}{})
 }
```
khelif96, line 248: Any suggestions where this should live?

rest/route/middleware.go
```diff
@@ -802,10 +801,12 @@ func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.Handler
 		requester := r.Header.Get("Origin")
 		grip.Debug(message.Fields{
 			"op":              "addCORSHeaders",
+			"request":         r,
 			"requester":       requester,
 			"allowed_origins": allowedOrigins,
-			"adding_headers":  utility.StringSliceContains(allowedOrigins, requester),
+			"adding_headers":  util.StringContainsSliceRegex(allowedOrigins, requester),
 			"settings_is_nil": evergreen.GetEnvironment().Settings() == nil,
+			"headers":         r.Header,
```
ablack12, line 809: Is it worth logging headers before we add headers? Maybe we should move this log to the end. We should also probably start only logging this if requester != "" cause it's been noisy.

rest/route/middleware.go
```diff
@@ -802,10 +801,12 @@ func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.Handler
 		requester := r.Header.Get("Origin")
 		grip.Debug(message.Fields{
 			"op":              "addCORSHeaders",
+			"request":         r,
```
ablack12, line 804: What does this add for us?

rest/route/middleware.go
```diff
@@ -815,6 +816,7 @@ func AddCORSHeaders(allowedOrigins []string, next http.HandlerFunc) http.Handler
 				w.Header().Add("Access-Control-Allow-Credentials", "true")
 				w.Header().Add("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PATCH, PUT")
 				w.Header().Add("Access-Control-Allow-Headers", fmt.Sprintf("%s, %s, %s", evergreen.APIKeyHeader, evergreen.APIUserHeader, gqlHeader))
+				w.Header().Add("Access-Control-Max-Age", "600")
```
ablack12, line 819: I don't think we need to enforce that users can only make a request once every ten minutes. What's motivating this change? The UI uses this too so I think this will cause problems.

rest/route/service.go
```diff
@@ -209,4 +212,36 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/versions/{version_id}/builds").Version(2).Get().Wrap(viewTasks).RouteHandler(makeGetVersionBuilds(sc))
 	app.AddRoute("/versions/{version_id}/restart").Version(2).Post().Wrap(checkUser, editTasks).RouteHandler(makeRestartVersion(sc))
 	app.AddRoute("/versions/{version_id}/annotations").Version(2).Get().Wrap(checkUser, viewAnnotations).RouteHandler(makeFetchAnnotationsByVersion(sc))
+
+	// Add an options method to every POST request to handle pre-flight Options requests
+	// These requests must not check for credentials and just validate whether a route exists
+	// And allows requests from a origin.
+	for _, route := range app.Routes() {
+		if route.HasMethod("POST") {
```
ablack12, line 220: Awesome solution!

rest/route/service.go
```diff
@@ -209,4 +212,36 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/versions/{version_id}/builds").Version(2).Get().Wrap(viewTasks).RouteHandler(makeGetVersionBuilds(sc))
 	app.AddRoute("/versions/{version_id}/restart").Version(2).Post().Wrap(checkUser, editTasks).RouteHandler(makeRestartVersion(sc))
 	app.AddRoute("/versions/{version_id}/annotations").Version(2).Get().Wrap(checkUser, viewAnnotations).RouteHandler(makeFetchAnnotationsByVersion(sc))
+
+	// Add an options method to every POST request to handle pre-flight Options requests
```
ablack12, line 216: Nit: needs period

rest/route/service.go
ablack12: I think you can make a new rest/route/options.go

ablack12 (changes requested):

rest/route/middleware.go
khelif96: This is logging the request headers that we are receiving not the ones we set. 

Sure, yeah we can add that conditional.

rest/route/middleware.go
khelif96: This was logging the entire request that was being sent i was hoping to use it to identify some request metadata that i may have missed but i can remove it.

rest/route/middleware.go
khelif96: This is only used by pre flight requests. Which are just used to identify if a CORS policy will be violated prior to making a request.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age

This will end up reducing load on evergreen (Although a very minuscule drop) . Since this will prevent 2 requests from being made for every POST request.

rest/route/options.go
```diff
@@ -0,0 +1,31 @@
...
+// Handler returns a pointer to a new optionsHandler.
+func (h *optionsHandler) Factory() gimlet.RouteHandler {
+	return &optionsHandler{}
+}
+
+func (h *optionsHandler) Parse(ctx context.Context, r *http.Request) error {
+	return nil
+}
+
+// Execute returns an empty json response to the options request
```
ablack12, line 27: nit: end sentences with periods

ablack12 (approved):
LGTM with nit 🚀

PR #5055, opened 2021-09-24
EVG-14959 Create stub versions for merging errors

bynn:
[EVG-14959](https://jira.mongodb.org/browse/EVG-14959)

### Description 
Merging errors create a stub version now

### Testing 
added a unit test 

created one on staging 
https://evergreen-staging.corp.mongodb.com/version/sandbox_9db9a2dd3155a2f1145cca6fb9f16db4136513d4#/0

### DIFF
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -399,7 +399,8 @@ func (repoTracker *RepoTracker) GetProjectConfig(ctx context.Context, revision s
 		_, ymlFmtErr := errors.Cause(err).(thirdparty.YAMLFormatError)
 		_, noFileErr := errors.Cause(err).(thirdparty.FileNotFoundError)
 		parsingErr := strings.Contains(err.Error(), "translating project")
-		if apiReqErr || noFileErr || ymlFmtErr || parsingErr {
+		mergingErr := strings.Contains(err.Error(), "merging project configs")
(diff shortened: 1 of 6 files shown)


repotracker/repotracker.go
```diff
@@ -399,7 +399,8 @@ func (repoTracker *RepoTracker) GetProjectConfig(ctx context.Context, revision s
 		_, ymlFmtErr := errors.Cause(err).(thirdparty.YAMLFormatError)
 		_, noFileErr := errors.Cause(err).(thirdparty.FileNotFoundError)
 		parsingErr := strings.Contains(err.Error(), "translating project")
-		if apiReqErr || noFileErr || ymlFmtErr || parsingErr {
+		mergingErr := strings.Contains(err.Error(), "merging project configs")
```
ablack12, line 402: Although I'm the one who did this in the first place, we should avoid hardcoding errors when we can. Instead it makes sense to make a constant for this error and for translating (like [this one](https://github.com/evergreen-ci/evergreen/blob/04bd7460eadc19cb52b146e5ccbfe3d4edbc39bc/model/project_parser.go#L25) and use it to wrap the original error and to check the error here.

ablack12 (changes requested):

repotracker/repotracker.go
bynn: done

ablack12 (approved):

PR #5056, opened 2021-09-24
EVG-15461: terminate intent pods

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15461

### Description 
Intent pods don't have any backing resources in ECS or Secrets Manager, so there's no way to create a Cocoa pod from an Evergreen intent pod. During pod termination, intent pods can skip the deletion of the backing cloud resources, so all they have to do is be marked as terminated.

### Testing 
Updated the job tests.


julianedwards (approved):
LGTM % nit

PR #5058, opened 2021-09-27
EVG-8326 Create override dependencies mutation

khelif96:
[EVG-8326](https://jira.mongodb.org/browse/EVG-8326)

### Description 
This will override task dependencies for blocked tasks


graphql/resolvers.go
```diff
@@ -2169,6 +2169,21 @@ func (r *mutationResolver) SetTaskPriority(ctx context.Context, taskID string, p
 	return apiTask, err
 }
 
+func (r *mutationResolver) OverrideTaskDependencies(ctx context.Context, taskID string) (*restModel.APITask, error) {
+	currentUser := MustHaveUser(ctx)
+	t, err := task.FindOneIdAndExecutionWithDisplayStatus(taskID, nil)
+	if err != nil {
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("error finding task %s: %s", taskID, err.Error()))
```
ablack12, line 2267: should this be InternalServerError?

ablack12 (approved):

ablack12 (changes requested):
Disregard this comment -- somehow the PR I was reviewing didn't have your updated permission changes. Let me look again.

graphql/resolvers.go
```diff
@@ -299,6 +299,97 @@ func (r *taskResolver) ReliesOn(ctx context.Context, at *restModel.APITask) ([]*
 	return dependencies, nil
 }
 
+func (r *taskResolver) DependsOn(ctx context.Context, at *restModel.APITask) ([]*Dependency, error) {
```
ablack12, line 302: What is this being used for?

graphql/resolvers.go
```diff
@@ -299,6 +299,97 @@ func (r *taskResolver) ReliesOn(ctx context.Context, at *restModel.APITask) ([]*
...
+	currentUser := MustHaveUser(ctx)
+	i, err := at.ToService()
+	if err != nil {
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("Error getting service model for APITask %s: %s", *at.Id, err.Error()))
+	}
+	t, ok := i.(*task.Task)
+	if !ok {
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert APITask %s to Task", *at.Id))
+	}
+	if t.OverrideDependencies {
```
ablack12, line 377: Are you returning false here because dependencies have already been overridden? Is it necessary to convert the whole task document? Seems like we've used `utility.FromBoolPtr` when we haven't needed the whole struct.

graphql/resolvers.go
```diff
@@ -299,6 +299,97 @@ func (r *taskResolver) ReliesOn(ctx context.Context, at *restModel.APITask) ([]*
...
+		return false, nil
+	}
+	requiredPermission := gimlet.PermissionOpts{
+		ResourceType:  "project",
```
ablack12, line 381: evergreen.ProjectResourceType

graphql/resolvers.go
```diff
@@ -299,6 +299,97 @@ func (r *taskResolver) ReliesOn(ctx context.Context, at *restModel.APITask) ([]*
...
+		Permission:    evergreen.PermissionTasks,
+		RequiredLevel: 30,
```
ablack12, line 383: TasksAdmin.Value https://github.com/evergreen-ci/evergreen/blob/92821d037c441ad06df39d0f41caa4daec9afbcf/globals.go#L790

graphql/resolvers.go
khelif96: This will be replacing the `reliesOn` field. For some reason when we implemented a dependencies resolver the `dependsOn` field was renamed to `reliesOn`. This is resolving that and renaming it back to `dependsOn`

model/task/task.go
```diff
@@ -1501,7 +1501,7 @@ func (t *Task) GetDisplayStatus() string {
 		if !t.Activated {
 			return evergreen.TaskUnscheduled
 		}
-		if t.Blocked() {
+		if t.Blocked() && !t.OverrideDependencies {
```
SupaJoon, line 1504: what does OverrideDependencies represent?

khelif96: OverrideDependencies is a flag on the task that gets toggled here. It indicates that the tasks dependencies should be ignored. 
https://github.com/evergreen-ci/evergreen/blob/9dce0b8be4d937b2b6582e0254dc6cf2aecdfb7c/model/task/task.go#L511-L523

SupaJoon (approved):

graphql/resolvers.go
```diff
@@ -2293,14 +2285,17 @@ func (r *mutationResolver) OverrideTaskDependencies(ctx context.Context, taskID
...
 	}
 	if t == nil {
 		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("cannot find task with id %s", taskID))
 	}
 	if err = t.SetOverrideDependencies(currentUser.Username()); err != nil {
 		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error overriding dependencies for task %s: %s", taskID, err.Error()))
 	}
+	// Clear DisplayStatus so GetDisplayStatus recalculates it.
+	t.DisplayStatus = ""
+	t.DisplayStatus = t.GetDisplayStatus()
```
ablack12, line 2298: Instead of having to clear the display status, can we just use FindOneIdAndExecution at 2286 instead of "with display status"?  Doesn't seem like we're using it.

ablack12 (changes requested):

graphql/resolvers.go
khelif96: Nice catch. Done.

graphql/resolvers.go
```diff
@@ -2283,7 +2283,7 @@ func (r *mutationResolver) SetTaskPriority(ctx context.Context, taskID string, p
 
 func (r *mutationResolver) OverrideTaskDependencies(ctx context.Context, taskID string) (*restModel.APITask, error) {
 	currentUser := MustHaveUser(ctx)
-	t, err := task.FindOneIdAndExecutionWithDisplayStatus(taskID, nil)
+	t, err := task.FindByIdExecution(taskID, nil)
```
ablack12, line 2286: Actually sorry last thought -- I think we do need to be passing in execution here. A user may want to override dependencies on an old execution of a task; this function will always only return the most recent

khelif96: We actually don't allow any task actions if the task is not on the latest execution. Including overriding dependencies, I've added the same safety check to my pr. 
https://evergreen-staging.corp.mongodb.com/task/evg_ubuntu1604_test_cloud_patch_ccbdfbe1ca4c5e72baed4d0505b6b058566c89b4_61083e9a97b1d31e62c81c7b_21_08_02_18_51_19/0 
https://github.com/evergreen-ci/evergreen/blob/921e132cd5094bbae58151c550ac915110ca2c83/service/templates/task.html#L118

ablack12: Thanks for finding!

ablack12 (approved):

PR #5059, opened 2021-09-27
EVG-15456 check for nil cmds when validating

ablack12:
[EVG-15456](https://jira.mongodb.org/browse/EVG-15456)

### Description 
Didn't make cmds length an error because we check for this elsewhere and add an error so it would be duplicate info; at this point we should just make sure we don't panic.

### Testing 
Tested with local evergreen.
image


bynn (approved):

bynn:
looks good with agent version!

PR #5060, opened 2021-09-27
EVG-15420 refactor save project settings for section to use multiple connectors

ablack12:
[EVG-15420](https://jira.mongodb.org/browse/EVG-15420)

### Description 
I realized I can use the DBConnector to have these functions still be methods, which is preferable. Renamed the file to better reflect the contents.

### Testing 
Added to the unit test.


rest/data/project_settings.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/evergreen-ci/gimlet"
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+)
+
+// This file is used to combine operations across data connectors, to avoid
+// duplicated connector usage across the codebase.
+
+func (sc *DBConnector) CopyProject(ctx context.Context, projectToCopy *model.ProjectRef, newProject string) (*restModel.APIProjectRef, error) {
```
ablack12, line 18: This function was not modified.

rest/data/project_settings.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, errors.Wrapf(err, "error copying subscriptions from project '%s'", oldIdentifier)
+	}
+	// set the same admin roles from the old project on the newly copied project.
+	if err := sc.UpdateAdminRoles(projectToCopy, projectToCopy.Admins, nil); err != nil {
+		return nil, errors.Wrapf(err, "Database error updating admins for project '%s'", newProject)
+	}
+	return apiProjectRef, nil
+}
+
+func (sc *DBConnector) SaveProjectSettingsForSection(ctx context.Context, projectId string, changes *restModel.APIProjectSettings,
```
ablack12, line 53: Most of this function was just moved -- only added the last two cases

rest/data/project_settings_test.go
```diff
@@ -0,0 +1,391 @@
...
+			assert.NoError(t, err)
+			assert.NotNil(t, varsFromDb)
+			assert.Equal(t, varsFromDb.Vars["it"], "me")
+			assert.Equal(t, varsFromDb.Vars["banana"], "phone")
+			assert.Equal(t, varsFromDb.Vars["hello"], "")
+			assert.False(t, varsFromDb.PrivateVars["it"])
+			assert.False(t, varsFromDb.PrivateVars["hello"])
+			assert.True(t, varsFromDb.PrivateVars["banana"])
+		},
+		model.ProjectPageNotificationsSection: func(t *testing.T, ref model.ProjectRef) {
```
ablack12, line 279: This is the only thing I added to this file (and the subscription added  to the DB below)

rest/route/project.go
```diff
@@ -462,10 +462,7 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
 	if err = h.sc.UpdateAdminRoles(h.newProjectRef, adminsToAdd, adminsToDelete); err != nil {
 		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "Database error updating admins for project '%s'", h.project))
 	}
-	for i := range h.apiNewProjectRef.Subscriptions {
-		h.apiNewProjectRef.Subscriptions[i].OwnerType = utility.ToStringPtr(string(event.OwnerTypeProject))
-		h.apiNewProjectRef.Subscriptions[i].Owner = utility.ToStringPtr(h.project)
-	}
+
```
ablack12, line 465: This is actually handled in SaveSubscriptions

ablack12 (commented):
Added some comments to help with reviewing.

ablack12:
evergreen retry

rest/route/project.go
malikchaya2: in saveProject it only sets the OwnerType to event.OwnerTypeProject if isProjectOwner. And it sets the owner to h.newProjectRef.Id instead of h.project. 
Also, in SaveSubscription it only sets it if certain conditions are met, and here it sets it for everything, potentially leaving something without an owner and ownerType. 

I don't have a full understanding of this, I just want to make sure we are not accidentally overlooking something.

malikchaya2 (commented):
`I realized I can use the DBConnector to have these functions still be methods, which is preferable. Renamed the file to better reflect the contents.`

Good point

rest/route/project.go
ablack12: We do call SaveSubscriptions with `isProjectOwner=true` on line 468 so I think that's covered. The only other condition I'm seeing for setting it is that the interface converts cleanly (which needs to happen to save the document anyway). Is there something else you're seeing?

malikchaya2 (approved):

PR #5062, opened 2021-09-28
EVG-15125: support shared pod secrets

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15125

### Description 
* Support pods that may or may not own their secrets from creation to deletion. This is necessary since future work will include making fewer calls to Secrets Manager, so secrets will be pre-generated and shared between multiple pods.
* Revendor Cocoa.

### Testing 
Updated DB, REST, and cloud util tests.

### DIFF
--- a/model/pod/pod.go
+++ b/model/pod/pod.go
@@ -18,7 +18,7 @@ type Pod struct {
 	Status Status `bson:"status"`
 	// Secret is the shared secret between the server and the pod for
 	// authentication when the host is provisioned.
-	Secret string `bson:"secret" json:"secret"`
+	Secret Secret `bson:"secret" json:"secret"`
 	// TaskCreationOpts are options to configure how a task should be
 	// containerized and run in a pod.
 	TaskContainerCreationOpts TaskContainerCreationOptions `bson:"task_creation_opts,omitempty" json:"task_creation_opts,omitempty"`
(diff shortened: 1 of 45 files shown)


rest/model/pod.go
```diff
@@ -88,8 +97,8 @@ func (p *APICreatePod) splitEnvVars() (envVars map[string]string, secrets map[st
 // APIPod represents a pod to be used and returned from the REST API.
 type APIPod struct {
 	ID                        *string                            `json:"id"`
-	Status                    *APIPodStatus                      `json:"status"`
-	Secret                    *string                            `json:"secret,omitempty"`
+	Status                    APIPodStatus                       `json:"status"`
+	Secret                    APIPodSecret                       `json:"secret,omitempty"`
```
julianedwards, line 101: Should this be a pointer if we want to omit empty?

model/pod/pod.go
```diff
@@ -196,6 +196,33 @@ func (o TaskContainerCreationOptions) IsZero() bool {
...
 }
 
+// Secret is a sensitive secret that a pod can access. The secret is managed
+// in an integrated secrets storage service.
+type Secret struct {
+	// Name is the friendly name of the secret.
+	Name string `bson:"name,omitempty" json:"name,omitempty" yaml:"name,omitempty"`
+	// ExternalID is the unique external resource identifier for a secret that
+	// already exists in the secrets storage service.
+	ExternalID string `bson:"external_id,omitempty" json:"external_id,omitempty" yaml:"external_id,omitempty"`
```
julianedwards, line 206: Should there be a validation on `ExternalID` only existing when `Owned` is `false`? It seems like it would be invalid to own the secret and also have an external ID.

Kimchelly: Maybe, but I want to keep it flexible by not assuming that this is how secrets will be shared yet (if we could make this assumption, I would get rid of `Owned` entirely and just decide whether the secret is owned based on the presence/absence of `ExternalID`). There's also no good place to validate it right now since the only way to create a pod is through a temporary REST route which will go away once pods are used in the task lifecycle.

rest/model/pod.go
Kimchelly: I guess it could be a pointer but it probably doesn't matter that much since the `APIPodSecret` has only pointer fields. I removed omitempty from all the struct fields in `APIPod`.

julianedwards (approved):

PR #5064, opened 2021-09-28
EVG-14960 Update LoadProjectInto to take opts pointer

bynn:
[EVG-14960](https://jira.mongodb.org/browse/EVG-14960)

### Description 
Update LoadProjectInto to take opts pointer

### Testing 
compiles

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -530,9 +530,8 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (*Project, *P
 	}
 	p := &Project{}
 	// opts empty because project yaml with `include` will not hit this case
-	opts := GetProjectOpts{}
 	ctx := context.Background()
-	pp, err = LoadProjectInto(ctx, []byte(v.Config), opts, id, p)
+	pp, err = LoadProjectInto(ctx, []byte(v.Config), nil, id, p)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "error loading project")
 	}
@@ -566,7 +565,7 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 // and sets the project's identifier field to identifier. Tags are evaluated. Returns the intermediate step.
 // If reading from a version config, LoadProjectForVersion should be used to persist the resulting parser project.
 // opts is used to look up files on github if the main parser project has an Include.
-func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
+func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
 	intermediateProject, err := createIntermediateProject(data)
 	if err != nil {
 		return nil, errors.Wrapf(err, LoadProjectError)
@@ -583,7 +582,7 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
 				opts.ReadFileFrom = ReadFromPatch
 			}
 		}
-		yaml, err := retrieveFile(ctx, opts)
+		yaml, err := retrieveFile(ctx, *opts)
 		if err != nil {
 			return intermediateProject, errors.Wrapf(err, LoadProjectError)
 		}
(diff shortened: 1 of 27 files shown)


model/project_parser.go
```diff
@@ -584,7 +583,7 @@ func LoadProjectInto(ctx context.Context, data []byte, opts GetProjectOpts, iden
 				opts.ReadFileFrom = ReadFromPatch
 			}
 		}
-		yaml, err := retrieveFile(ctx, opts)
```
ablack12, line 587: Let's skip the loop entirely if opts isn't nil

model/testutil/api.go
```diff
@@ -55,7 +55,7 @@ func SetupAPITestData(testConfig *evergreen.Settings, taskDisplayName string, va
 	// Unmarshal the project configuration into a struct
 	project := &model.Project{}
 	ctx := context.Background()
-	opts := model.GetProjectOpts{
```
ablack12, line 58: I think it's awkward that we can have opts but not have ReadFileFrom set (I know this is a dummy case but). If opts != nil in LoadProjectInto, should we default ReadFileFrom? Or error log if it's empty (and add something to this case?)

rest/data/project.go
```diff
@@ -599,7 +599,7 @@ tasks:
 - name: t1
 `
 	p := &model.Project{}
-	opts := model.GetProjectOpts{
```
ablack12, line 602: Do we need a ReadFileFrom here? (Although I recognize this is just a mock)

rest/route/version_create.go
```diff
@@ -60,7 +60,7 @@ func (h *versionCreateHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.NewJSONErrorResponse(errors.Errorf("project '%s' doesn't exist", h.ProjectID))
 	}
 	p := &model.Project{}
-	opts := model.GetProjectOpts{
```
ablack12, line 63: Do we need a LoadFileFrom here?

units/periodic_builds.go
```diff
@@ -120,7 +120,7 @@ func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.Peri
 		return "", errors.Wrap(err, "error decoding config file")
 	}
 	proj := &model.Project{}
-	opts := model.GetProjectOpts{
+	opts := &model.GetProjectOpts{
```
ablack12, line 123: LoadFileFrom?

ablack12 (changes requested):
Most of my comments are around having a ReadFileFrom which I realize isn't part of the PR description but is the purpose of the ticket; if you wanted to make a different PR for that then disregard those comments

model/project_parser.go
```diff
@@ -574,6 +574,9 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 
 	// return intermediateProject even if we run into issues to show merge progress
 	for _, path := range intermediateProject.Include {
+		if opts == nil {
+			return nil, errors.Wrapf(err, LoadProjectError)
```
ablack12, line 578: Errors.Wrap here will return nil since error is nil. I'm not actually sure we should error here since that's kind of debilitating for users and really is more of an error on our side. Maybe logging at the Error or Critical level (which goes to ops-alerts) is better so that we're aware of the problem.

rest/route/version_create.go
```diff
@@ -61,7 +61,8 @@ func (h *versionCreateHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 	p := &model.Project{}
 	opts := &model.GetProjectOpts{
```
ablack12, line 63: Is it true that we might be reading from Github here? Is there a way to test this route (or trace it back)? If h.Config already contains the full yaml, then we really should juts be passing in nil opts.

ablack12 (changes requested):

model/project_parser.go
ablack12: ah yes haha

model/project_parser.go
bynn: critical logging

model/project_parser.go
```diff
@@ -567,14 +566,19 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
...
 	intermediateProject, err := createIntermediateProject(data)
 	if err != nil {
 		return nil, errors.Wrapf(err, LoadProjectError)
 	}
 
 	// return intermediateProject even if we run into issues to show merge progress
 	for _, path := range intermediateProject.Include {
+		if opts == nil {
+			grip.Critical(message.Fields{
+				"message": "Trying to open include files with empty opts",
```
ablack12, line 579: Can we do message.WithError(errors.WithStack(<include an error here>)) ? 
Without the stack trace, we really don't have any information about how to fix this.

ablack12: We also want to break out of the loop if we hit this.

bynn: message.WithError didn't exist and the WithStack wanted an error so this is what i came up with.

ablack12 (approved):

PR #5065, opened 2021-09-28
EVG-15212: Fix build baron HTML error for disabled projects

hadjri:
[EVG-15212](https://jira.mongodb.org/browse/EVG-15212)

### Description 
Was observed [here](https://mongodb.slack.com/archives/C0V896UV8/p1630698656157000) that a user was unexpectedly seeing build baron UI panels showing up on all tasks regardless of whether or not they were failed or not.  Rolled back the change and tested locally, found that the UI panel was rendering due to a null pointer exception on the frontend.  Before the rollback a change was made to return a nil object in the case we are unable to retrieve the build baron config settings when rendering the UI panel config.  Change has been made to return a struct with enabled: false to prevent conf.enabled from throwing a null pointer exception when this occurs.  The initial EVG-15212 changes have also been un-reverted.
### Testing 
Checking out the old branch with this change showed the same HTML error where build baron panels were shown when they should not.  Added this change and tested locally and in staging and the issue was gone.

### DIFF
--- a/plugin/build_baron.go
+++ b/plugin/build_baron.go
@@ -138,10 +138,13 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
 					template.HTML(`<script type="text/javascript" src="/static/plugins/buildbaron/js/task_build_baron.js"></script>`),
 				},
 				DataFunc: func(context UIContext) (interface{}, error) {
-					enabled := len(bbp.opts.Projects[context.ProjectRef.Id].TicketSearchProjects) > 0
-					if !enabled {
-						enabled = len(bbp.opts.Projects[context.ProjectRef.Identifier].TicketSearchProjects) > 0
+					bbSettings, ok := BbGetProject(evergreen.GetEnvironment().Settings(), context.ProjectRef.Id, "")
+					if !ok {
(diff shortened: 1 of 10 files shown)


plugin/build_baron.go
```diff
@@ -138,10 +138,13 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
 					template.HTML(`<script type="text/javascript" src="/static/plugins/buildbaron/js/task_build_baron.js"></script>`),
 				},
 				DataFunc: func(context UIContext) (interface{}, error) {
-					enabled := len(bbp.opts.Projects[context.ProjectRef.Id].TicketSearchProjects) > 0
-					if !enabled {
-						enabled = len(bbp.opts.Projects[context.ProjectRef.Identifier].TicketSearchProjects) > 0
+					bbSettings, ok := BbGetProject(evergreen.GetEnvironment().Settings(), context.ProjectRef.Id, "")
+					if !ok {
```
ablack12, line 142: if !ok, can we just set enabled to false? Then we can still hit 148 and not have to redefine the dummy struct

ablack12 (approved):

PR #5066, opened 2021-09-29
EVG-15263: Use updated cedar test results route

julianedwards:
EVG-15263: https://jira.mongodb.org/browse/EVG-15263

- Revendor timber.
- Update code to use new test results route with filtering, sorting and pagination.
- Remove graphql filtering, sorting, and pagination code.
- Add fields `HasCedarResults` and `CedarResultsFailed` to the `APITask` struct (used by graphql funcs).
- Update CLI buildlogger to fetch latest execution if none specified.

I tested the test results API, legacy UI, and spruce in staging.


apimodels/test_results.go
```diff
@@ -28,52 +54,112 @@ type CedarTestResult struct {
...
+		return nil, errors.Wrap(err, "unmarshaling test results from Cedar")
 	}
 
 	return testResults, nil
 }
 
+// GetMultiPageCedarTestResults makes a request to Cedar for a task's test
+// results and returns an io.ReadCloser that will continue fetching and reading
+// subsequent pages of test results if paginated.
+func GetMutliPageCedarTestResults(ctx context.Context, opts GetCedarTestResultsOptions) (io.ReadCloser, error) {
```
Kimchelly, line 116: GetMultiPageCedarTestResults

apimodels/test_results.go
```diff
@@ -28,52 +54,112 @@ type CedarTestResult struct {
...
+	stats := &CedarTestResultsStats{}
+	if err = json.Unmarshal(data, stats); err != nil {
+		return nil, errors.Wrap(err, "unmarshaling test results stats from cedar")
+	}
+
+	return stats, nil
+}
+
+// GetCedarTestResultsFailedSample makes a request to Cedar for a task's failed
+// test result sample (the first 10 failed tests). This route ignores
```
Kimchelly, line 145: I would avoid details like "the first 10 failed tests". Would it be possible to phrase this in a way that doesn't reveal specifics about the current implementation details, or just remove the parentheses comment?

apimodels/test_results.go
```diff
@@ -28,52 +54,112 @@ type CedarTestResult struct {
...
+// filtering, sorting, and pagination query parameters.
```
Kimchelly, line 146: If there are two methods (`GetCedarTestResultsFailedSample` and `GetCedarTestResultsStats`) that ignore certain parameters in `GetCedarTestResultsOptions`, would it be preferable to split the `GetCedarTestResultsOptions` into two structs so that methods don't selectively ignore some parameters?

graphql/resolvers.go
```diff
@@ -1422,200 +1422,131 @@ func (r *queryResolver) PatchTasks(ctx context.Context, patchID string, sorts []
...
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding base task for task %s: %s", taskID, err))
+	}
+
+	var sortBy, cedarSortBy string
 	if sortCategory != nil {
 		switch *sortCategory {
 		case TestSortCategoryStatus:
+			cedarSortBy = apimodels.CedarTestResultsSortByStatus
 			sortBy = testresult.StatusKey
-			break
```
Kimchelly, line 1430: Thank you for this (no change). 🙏

model/task/task.go
```diff
@@ -86,8 +86,8 @@ type Task struct {
 	TaskGroupOrder     int                 `bson:"task_group_order,omitempty" json:"task_group_order,omitempty"`
 	Logs               *apimodels.TaskLogs `bson:"logs,omitempty" json:"logs,omitempty"`
 	MustHaveResults    bool                `bson:"must_have_results,omitempty" json:"must_have_results,omitempty"`
-	HasCedarResults    bool                `bson:"has_cedar_results,omitempty" json:"has_cedar_results,omitempty"`
-	CedarResultsFailed bool                `bson:"cedar_results_failed,omitempty" json:"cedar_results_failed,omitempty"`
+	HasCedarResults    bool                `bson:"has_cedar_results" json:"has_cedar_results"`
+	CedarResultsFailed bool                `bson:"cedar_results_failed" json:"cedar_results_failed"`
```
Kimchelly, line 90: Is it necessary to change these to be non-omitempty in the DB? It seems like it would only have to be added to the REST model, which will still correctly see a false boolean in the case where it's omitted in the DB model.

Kimchelly (changes requested):

model/task/task.go
julianedwards: I guess it really doesn't matter, saves room in the DB so I will add them back.

apimodels/test_results.go
```diff
@@ -4,22 +4,48 @@ import (
...
+	Stats   CedarTestResultsStats `json:"stats"`
+	Results []CedarTestResult     `json:"results"`
+}
+
+// CedarTestResultsStats represents the expected test results stats format
+// returned from Cedar.
+type CedarTestResultsStats struct {
+	TotalCount    int `json:"total_count"`
+	FailedCount   int `json:"failed_count"`
+	FilteredCount int `json:"filtered_count"`
```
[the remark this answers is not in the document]
julianedwards: Nope, changing it to an optional int.

apimodels/test_results.go
julianedwards: This is more for future proofing, it is possible that in the future we want to get test stats on filtered results or filter the failed test sample.

Kimchelly (approved):

rest/model/task.go
```diff
@@ -74,6 +74,8 @@ type APITask struct {
 	CanSync                 bool                `json:"can_sync,omitempty"`
 	SyncAtEndOpts           APISyncAtEndOptions `json:"sync_at_end_opts"`
 	Ami                     *string             `json:"ami"`
+	HasCedarResults         bool                `json:"has_cedar_results"`
+	CedarResultsFailed      bool                `json:"cedar_results_failed"`
```
khelif96, line 78: nit. I'm assuming this is only used because we want to access this field from a child resolver without fetching the db task.
 https://github.com/evergreen-ci/evergreen/pull/5066/files#diff-a3ce50f20b746b196f29c6d414e685b0110c24b768bcf83756de8d842c98f6dcR2703 

Would we be able to specify `json:"-" for these fields` so they doesn't show up in an api response. We had previously spoken to the app team about reducing the amount of fields we add to the apiModels.

khelif96 (commented):
LGTM

rest/model/task.go
julianedwards: Yeah, or we can add an `omitempty` so it only appears if there are cedar test results or if the cedar test results failed. Would it ever be useful outside of the resolvers to use these fields?

khelif96: I don't think we wan't/need to expose these fields to users. Since we are only using them internally to check if the values are populated to support some other functionality we don't ever return them or expose them to the user.

julianedwards: SGTM, I will put up another PR to set the json tags accordingly.

PR #5067, opened 2021-09-29
EVG-15405: Use merged parser project and project ref struct for properties that are shared between them

hadjri:
[EVG-15405](https://jira.mongodb.org/browse/EVG-15405)

### Description 
With the introduction of the Expose Configuration Options to Users project, there are many places where we retrieve a parser project, check if certain properties are defined, and if not default to the settings on the project ref.  Initially it seemed like it would be cleaner to merge the parser project ref into a single entity on the backend, and refer to this merged struct for all project settings related operations.  The work involved in this however would likely require a larger refactoring of both our database schemas and would fundamentally change the way we view projects from a backend standpoint which is out of scope for this project.  

To improve code cleanliness and readability, a new struct has been introduced that holds all of the shared properties between a project ref and a parser project, and when these properties' configs are required elsewhere in the project, we return this merged struct which will have the appropriate value (whether it comes from parser project or project ref).

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -113,12 +113,6 @@ type ProjectRef struct {
 	Hidden *bool `bson:"hidden,omitempty" json:"hidden,omitempty"`
 }
 
-type MergedProjectConfig struct {
-	DeactivatePrevious     *bool                         `bson:"deactivate_previous,omitempty" json:"deactivate_previous,omitempty" yaml:"deactivate_previous"`
-	PerfEnabled            *bool                         `bson:"perf_enabled,omitempty" json:"perf_enabled,omitempty" yaml:"perf_enabled,omitempty"`
-	TaskAnnotationSettings evergreen.AnnotationsSettings `bson:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
-}
-
 type CommitQueueParams struct {
 	Enabled     *bool  `bson:"enabled" json:"enabled"`
 	MergeMethod string `bson:"merge_method" json:"merge_method"`
@@ -410,12 +404,17 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
 	return patch.PatchTriggerDefinition{}, false
 }
 
-func (p *ProjectRef) GetMergedConfig(version string) *MergedProjectConfig {
+func (p *ProjectRef) GetProjectParserMergedProjectRef(version string) *ProjectRef {
 	lookupVersion := false
 	if version == "" {
 		lastGoodVersion, err := FindVersionByLastKnownGoodConfig(p.Id, -1)
 		if err != nil || lastGoodVersion == nil {
-			return nil
+			grip.Error(message.WrapError(err, message.Fields{
+				"message":    fmt.Sprintf("Unable to retrieve last good version for project '%s'", p.Id),
+				"project_id": p.Id,
+				"version":    version,
+			}))
+			return p
 		}
 		version = lastGoodVersion.Id
 		lookupVersion = true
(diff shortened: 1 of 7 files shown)


model/project_ref.go
```diff
@@ -404,6 +410,49 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
 	return patch.PatchTriggerDefinition{}, false
 }
 
+func (p *ProjectRef) GetMergedConfig(version string) *MergedProjectConfig {
```
ablack12, line 413: I'm not sure about making another new struct, especially since so many things will be shared between them. Could we instead destructively modify p (with a function like IncludeParserProject or something)? Or return a new merged project ref that incorporates the ParserProject version of the variables.

model/project_ref.go
```diff
@@ -404,6 +410,49 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
...
+	if version == "" {
+		lastGoodVersion, err := FindVersionByLastKnownGoodConfig(p.Id, -1)
+		if err != nil || lastGoodVersion == nil {
+			return nil
+		}
+		version = lastGoodVersion.Id
+		lookupVersion = true
+	}
+	parserProject, err := ParserProjectFindOneById(version)
+	if err != nil {
```
ablack12, line 424: I don't know that we should be suppressing these errors.

ablack12 (changes requested):

model/project_ref.go
```diff
@@ -404,6 +404,47 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
 	return patch.PatchTriggerDefinition{}, false
 }
 
+func (p *ProjectRef) GetProjectParserMergedProjectRef(version string) *ProjectRef {
```
ablack12, line 407: Can we shorten this title somehow? Maybe "MergeWithParserProject"; we don't need to say ProjectRef again since this is a function of ProjectRef.

model/project_ref.go
ablack12: What I meant by this is that we should return the error; an error here is a database error, not a "this project doesn't exist" instance so we should error on it and the user or job or etc can try again.

model/project_ref.go
```diff
@@ -404,6 +404,47 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
...
+				"version":    version,
+			}))
+			return p
+		}
+		version = lastGoodVersion.Id
+		lookupVersion = true
+	}
+	parserProject, err := ParserProjectFindOneById(version)
+	if err != nil {
+		grip.Debug(message.Fields{
```
ablack12, line 424: Should use message.WrapError instead of putting error directly into the fields

model/project_ref.go
ablack12: Should probably add a short comment also to explain what happens with the input.

model/project_ref.go
```diff
@@ -404,6 +404,47 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
...
+			"message":        fmt.Sprintf("Error retrieving parser project for version '%s'", version),
+			"project_id":     p.Id,
+			"version":        version,
+			"lookup_version": lookupVersion,
+			"err":            err.Error(),
+		})
+		return p
+	}
+	pRef := p
```
ablack12, line 433: Can we write some sort of test to verify that this doesn't destructively modify the input? (Since this is pointers I think it will)

ablack12 (changes requested):

model/project_ref.go
hadjri: This does destructively modify the input.  However isn't that what we are aiming to do here?

ablack12: Gotcha -- we can do that; typically when a method returns the data structure, it's implied that the input isn't changed, and we return a modified version. If we _intend_ to modify p, then there's no reason to return a different projectRef; this function can just return an error.

service/api_task.go
```diff
@@ -223,7 +223,13 @@ func (as *APIServer) EndTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// mark task as finished
-	deactivatePrevious := utility.FromBoolPtr(projectRef.GetProjectParserMergedProjectRef(t.Version).DeactivatePrevious)
+	mergedProjectRef, err := projectRef.MergeWithParserProject(t.Version)
+	if err != nil || mergedProjectRef == nil {
```
ablack12, line 227: We can't handle these errors together. If err == nil and mergedProjectRef == nil, then 228 will be nil because there is no error to wrap, and we'll be silently erroring.

model/project_ref.go
```diff
@@ -1420,7 +1429,11 @@ func IsPerfEnabledForProject(projectId string) bool {
 	if err != nil || projectRef == nil {
 		return false
 	}
-	return utility.FromBoolPtr(projectRef.GetProjectParserMergedProjectRef("").PerfEnabled)
+	mergedProjectRef, err := projectRef.MergeWithParserProject("")
+	if err != nil || mergedProjectRef == nil {
+		return false
+	}
+	return utility.FromBoolPtr(mergedProjectRef.PerfEnabled)
```
ablack12, line 1436: FYI We have an IsPerfEnabled (and similar functions) to make usage more readable but either work

service/api_task.go
hadjri: Good catch, thanks

ablack12 (approved):
LGTM -- depending honestly on how often we use the ProjectRef+ParserProject structure, maybe the main "FindMergedProjectRef" should just also handle adding in parser project. This is also something we can revisit as you continue though.

PR #5069, opened 2021-09-30
EVG-15413 Hide jira comment subscription option on waterfall page

bynn:
[EVG-15413](https://jira.mongodb.org/browse/EVG-15413)

### Description 
Disallow jira comment subscriptions to all tasks in a project

### Testing 
unit test

### DIFF
--- a/model/event/subscriptions.go
+++ b/model/event/subscriptions.go
@@ -315,18 +315,19 @@ func (s *Subscription) Validate() error {
 		catcher.Add(errors.New("must specify at least 1 selector"))
 	}
 	if s.ResourceType == "" {
-		catcher.Add(errors.New("subscription type is required"))
+		catcher.New("subscription type is required")
 	}
 	if s.Trigger == "" {
-		catcher.Add(errors.New("subscription trigger is required"))
+		catcher.New("subscription trigger is required")
 	}
 	if !IsValidOwnerType(string(s.OwnerType)) {
-		catcher.Add(errors.Errorf("%s is not a valid owner type", s.OwnerType))
+		catcher.Errorf("%s is not a valid owner type", s.OwnerType)
 	}
+	// Disallow creating a JIRA comment type subscriber for every task in a project
 	if s.OwnerType == OwnerTypeProject &&
 		s.ResourceType == ResourceTypeTask &&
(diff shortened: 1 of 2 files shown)


ablack12:
Okay wow I didn't even know you could do this for the waterfall 😮 this looks like you're omitting it completely though; theoretically we only want to omit it for "any task finishes" and "any task fails" since those are the noisy ones (although if no one is using it we may as well omit it). 

We do also need to do this for the project page (this is actually where the issue came from originally).

model/event/subscriptions.go
```diff
@@ -323,6 +323,11 @@ func (s *Subscription) Validate() error {
 	if !IsValidOwnerType(string(s.OwnerType)) {
 		catcher.Add(errors.Errorf("%s is not a valid owner type", s.OwnerType))
 	}
+	if s.OwnerType == OwnerTypeProject &&
+		s.ResourceType == ResourceTypeTask &&
+		s.Subscriber.Type == JIRACommentSubscriberType {
+		catcher.Add(errors.New("JIRA comment subscription not allowed for all tasks in the project"))
```
ablack12, line 329: We should probably disallow this for jira ticket creation as well. 

Also nit: catcher.New()

model/event/subscriptions.go
```diff
@@ -323,6 +323,11 @@ func (s *Subscription) Validate() error {
 	if !IsValidOwnerType(string(s.OwnerType)) {
 		catcher.Add(errors.Errorf("%s is not a valid owner type", s.OwnerType))
 	}
+	if s.OwnerType == OwnerTypeProject &&
```
ablack12, line 326: Can you add a comment about why we're validating this?

ablack12 (changes requested):

ablack12:
Sorry if I made it sound like the waterfall change you made was wrong; we _should_ still disallow doing this from the waterfall level, I just didn't know that was a thing at the time!

model/event/subscriptions.go
bynn: done and leaving the option in as discussed

model/event/subscriptions.go
bynn: done

ablack12 (approved):

PR #5072, opened 2021-10-01
EVG-15513: Attach child project identifier to patch trigger alias

sophstad:
[EVG-15513](https://jira.mongodb.org/browse/EVG-15513)

### Description 
- Add `childProjectIdentifier` field to be returned with patch trigger aliases

### Testing 
- Update unit test
- Tested on staging & playground

### DIFF
--- a/rest/model/project.go
+++ b/rest/model/project.go
@@ -74,12 +74,13 @@ func (t *APITriggerDefinition) BuildFromService(h interface{}) error {
 }
 
 type APIPatchTriggerDefinition struct {
-	Alias          *string            `json:"alias"`
-	ChildProject   *string            `json:"child_project"`
-	TaskSpecifiers []APITaskSpecifier `json:"task_specifiers"`
-	Status         *string            `json:"status,omitempty"`
-	ParentAsModule *string            `json:"parent_as_module,omitempty"`
-	VariantsTasks  []VariantTask      `json:"variants_tasks,omitempty"`
+	Alias                  *string            `json:"alias"`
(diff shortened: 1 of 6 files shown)


rest/model/project.go
```diff
@@ -74,12 +74,13 @@ func (t *APITriggerDefinition) BuildFromService(h interface{}) error {
...
 
 type APIPatchTriggerDefinition struct {
-	Alias          *string            `json:"alias"`
-	ChildProject   *string            `json:"child_project"`
-	TaskSpecifiers []APITaskSpecifier `json:"task_specifiers"`
-	Status         *string            `json:"status,omitempty"`
-	ParentAsModule *string            `json:"parent_as_module,omitempty"`
-	VariantsTasks  []VariantTask      `json:"variants_tasks,omitempty"`
+	Alias                  *string            `json:"alias"`
+	ChildProject           *string            `json:"child_project"`
```
malikchaya2, line 78: Is this the Id? should we maybe add "Id" to this?

sophstad: We could, but I believe that would introduce a breaking change and would require a little more coordination around releasing since the UI is already requesting `childProject`. I am open to it if you think it's worth doing!

malikchaya2: hmm. Should we maybe add another field called childProjectId that also holds the id  (and add a quick comment) so that we can eventually make the change next time we touch the UI? I think just calling it childProject might end up being confusing.

malikchaya2 (approved):

PR #5074, opened 2021-10-04
EVG-15397 Add superuser check directive.

khelif96:
[EVG-15397](https://jira.mongodb.org/browse/EVG-15397)

### Description 
This adds a directive which verifies if a query is made by a user with super user permissions

To use it you simply need to add `@superUserOnly` to the end of a query definition in the `schema.graphql`.

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -1,3 +1,6 @@
+directive @superUserOnly on FIELD_DEFINITION 
+directive @permissions(permissions: [String!]) on FIELD_DEFINITION
+
 type Query {
   userPatches(
     limit: Int = 0
@@ -11,7 +14,7 @@ type Query {
   taskAllExecutions(taskId: String!): [Task!]!
   patch(id: String!): Patch!
   version(id: String!): Version!
-  projects: [GroupedProjects]!
+  projects: [GroupedProjects]! @superUserOnly
   project(projectId: String!): Project!
   patchTasks(
     patchId: String!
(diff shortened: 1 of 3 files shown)


ablack12 (commented):
This looks good to me but why is this part of the projects track multiple branches project? There aren't queries that only super users should be able to use.

malikchaya2:
> This looks good to me but why is this part of the projects track multiple branches project? There aren't queries that only super users should be able to use.

It ended up in the epic because this is the first time that we need specific middleware for graphql. Now that it is set up, we can build on it and try to create more specific ones.

malikchaya2:
the CreateProject mutation that I have a[ pr up for](https://github.com/evergreen-ci/evergreen/pull/5073/files) should have this directive. If this ends up being merged after that, let's add it to that(though this will probably be merged first so no worries) :)

SupaJoon (approved):

malikchaya2 (approved):

malikchaya2 (commented):
Is there a way to test directives?

graphql/schema.graphql
```diff
@@ -13,7 +13,7 @@ type Query {
   taskAllExecutions(taskId: String!): [Task!]!
   patch(id: String!): Patch!
   version(id: String!): Version!
-  projects: [GroupedProjects]! 
+  projects: [GroupedProjects]!  @superUserOnly
```
ablack12, line 16: I'm not sure if you did this for testing / what this is used for but I wouldn't think this is necessarily superUser only.

khelif96: nice catch

malikchaya2 (approved):
This looks great! Since we are using requireSuperUser, can we implement Brian's suggestion of changing the naming on the app side? Either here or in a different ticket?

PR #5075, opened 2021-10-04
EVG-15126 Improve error message from /addtional patches route

bynn:
[EVG-15126](https://jira.mongodb.org/browse/EVG-15126)

### Description 
I changed it to not return an internal error but I'm not sure if this is correct 

### Testing 

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -3,6 +3,7 @@ package data
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ablack12 (commented):
Not returning an internal error is the right idea because we retry automatically for that error code; however, searching string errors is a bit hacky and we try to avoid it when possible. Something we can do instead is in GetAdditionalPatches, we can return a gimlet.ErrorResponse with the correct error code included.

bynn:
function now returns gimlet errors

ablack12 (approved):

PR #5076, opened 2021-10-04
EVG-14837: Query cedar to determine if perf data exists for a task

hhoke:

### Description 
IsPerfPluginEnabled now takes advantage of a new rest endpoint in cedar/timber that tells us the number of perf results.

### Testing 
  I was unable to test in cedar staging, as there are no perf results in evergreen staging.


graphql/resolvers.go
```diff
@@ -2871,13 +2871,26 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
 				return false, err
 			}
 			for _, projectName := range perfPlugin.Projects {
-				if projectName == pRef.Id || projectName == pRef.Identifier {
-					return true, nil
+				if projectName != pRef.Id || projectName != pRef.Identifier {
+					return false, nil
```
khelif96, line 2875: This seems like it would always fail on the first project if it does not match.

graphql/resolvers.go
```diff
@@ -2871,13 +2871,26 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
 			}
 		}
+		opts := apimodels.GetCedarPerfCountOptions{
+			BaseURL: evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:  *obj.Id,
+		}
+
+		result, err := apimodels.CedarPerfResultsCount(ctx, opts)
+		if err != nil {
+			return false, err
```
khelif96, line 2886: This should be returning a graphql error. `InternalServerError.Send(ctx, fmt.Sprintf("error requesting perf data from cedar: ", err))`

khelif96 (changes requested):

graphql/resolvers.go
hhoke: absolutely, forgot to change the || to an &&. Thanks for the catch!

apimodels/perf.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/evergreen-ci/timber"
+	perf "github.com/evergreen-ci/timber/perf"
+	"github.com/pkg/errors"
+)
+
+// GetCedarPerfCountOptions represents the arguments for getting a count
+// of perf results for a given task id from cedar
```
julianedwards, line 14: Missing period.

apimodels/perf.go
```diff
@@ -0,0 +1,46 @@
...
+func (opts GetCedarPerfCountOptions) convert() perf.GetOptions {
+	return perf.GetOptions{
+		Cedar: timber.GetOptions{
+			BaseURL: fmt.Sprintf("https://%s", opts.BaseURL),
+		},
+		TaskID: opts.TaskID,
+	}
+}
+
+type CedarPerfCount struct {
```
julianedwards, line 29: Add documentation.

graphql/resolvers.go
```diff
@@ -2871,13 +2871,26 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
-				if projectName == pRef.Id || projectName == pRef.Identifier {
-					return true, nil
+				if projectName != pRef.Id || projectName != pRef.Identifier {
+					return false, nil
 				}
 			}
 		}
+		opts := apimodels.GetCedarPerfCountOptions{
+			BaseURL: evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:  *obj.Id,
```
julianedwards, line 2881: Can this ever be the ID of a task in the old collection? Cedar does not recognize those, we always have to use the task id from the new collection (which is stored in old tasks under the field `old_task_id`).

julianedwards (changes requested):

graphql/resolvers.go
hhoke: Not sure. Maybe @khelif96 knows, if not I can try and figure it out.

khelif96: No it shouldn't but that is a good point. We should include execution in this request. Since the existence of perf data may change on a per execution basis.

graphql/resolvers.go
```diff
@@ -3030,13 +3030,27 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
 				return false, err
 			}
 			for _, projectName := range perfPlugin.Projects {
-				if projectName == pRef.Id || projectName == pRef.Identifier {
-					return true, nil
+				if projectName != pRef.Id && projectName != pRef.Identifier {
+					return false, nil
 				}
```
khelif96, line 3035: This logic doesn't look correct. 
The `pRef.Id` and `Identifier` are not always consistent with each other which is why we had the `or` statement. So we only needed to match on one of those values. 

When we return a value from a resolver the request ends so any code below will be unreachable. 

So if the first value checked in `perfPlugin.Projects` satisfies the conditional the resolver will return false and not allow any other elements to be checked.

graphql/resolvers.go
```diff
@@ -3019,6 +3019,9 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
 	if flags.PluginAdminPageDisabled {
 		return model.IsPerfEnabledForProject(*obj.ProjectId), nil
 	} else {
+		if !evergreen.IsFinishedTaskStatus(utility.FromStringPtr(obj.Status)) {
+			return false, nil
+		}
```
khelif96, line 3024: This can be moved up to before line 3015 so we can avoid doing any work for this resolver if we know there won't be any perf results.

julianedwards (approved):

PR #5077, opened 2021-10-04
EVG-15428 Update SaveProjectSettingsForSection to handle admin list correctly

bynn:
[EVG-15428](https://jira.mongodb.org/browse/EVG-15428)

### Description 
the updateAdmin function now modifies the admins list 

### Testing 
unit test

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1892,19 +1892,23 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
 		adminUser, err := user.FindOneById(addedUser)
 		if err != nil {
 			catcher.Wrapf(err, "error finding user '%s'", addedUser)
+			p.removeFromAdmins(addedUser)
 			continue
 		}
 		if adminUser == nil {
 			catcher.Errorf("no user '%s' found", addedUser)
+			p.removeFromAdmins(addedUser)
 			continue
 		}
 		if err = adminUser.AddRole(role.ID); err != nil {
 			catcher.Wrapf(err, "error adding role %s to user %s", role.ID, addedUser)
+			p.removeFromAdmins(addedUser)
 			continue
 		}
 		if viewRole != "" {
 			if err = adminUser.AddRole(viewRole); err != nil {
 				catcher.Wrapf(err, "error adding role %s to user %s", viewRole, addedUser)
+				p.removeFromAdmins(addedUser)
 				continue
 			}
 		}
@@ -1934,6 +1938,14 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
 		return errors.Wrap(err, "error updating some admins")
 	}
 	return nil
+}
+
+func (p *ProjectRef) removeFromAdmins(user string) {
+	for i, name := range p.Admins {
+		if name == user {
+			p.Admins = append(p.Admins[:i], p.Admins[i+1:]...)
+		}
+	}
 }
 
 func (p *ProjectRef) AuthorizedForGitTag(ctx context.Context, githubUser string, token string) bool {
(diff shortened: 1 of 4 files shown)


model/project_ref.go
```diff
@@ -1892,19 +1892,23 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
...
 		}
 		if err = adminUser.AddRole(role.ID); err != nil {
 			catcher.Wrapf(err, "error adding role %s to user %s", role.ID, addedUser)
+			p.removeFromAdmins(addedUser)
 			continue
 		}
 		if viewRole != "" {
 			if err = adminUser.AddRole(viewRole); err != nil {
 				catcher.Wrapf(err, "error adding role %s to user %s", viewRole, addedUser)
+				p.removeFromAdmins(addedUser)
```
ablack12, line 1911: Hmm theoretically we shouldn't remove them here since they _did_ get the admin role successfully and that's the more important one.

model/repo_ref.go
```diff
@@ -275,6 +278,14 @@ func (r *RepoRef) UpdateAdminRoles(toAdd, toRemove []string) error {
 	return nil
 }
 
+func (r *RepoRef) removeFromAdmins(user string) {
```
ablack12, line 281: You shouldn't need this function; RepoRef inherits the methods from ProjectRef since it just wraps it (this might not be true since it's private but you could just make it public). I'd also say "removeFromAdminsList" to make it clear that this is only local.

model/project_ref.go
```diff
@@ -1936,6 +1940,14 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
 	return nil
 }
 
+func (p *ProjectRef) removeFromAdmins(user string) {
```
ablack12, line 1943: We also need to handle the case that removing an admin doesn't work (i.e. we need an "addToAdminsList")

rest/data/project_settings.go
```diff
@@ -91,8 +91,9 @@ func (sc *DBConnector) SaveProjectSettingsForSection(ctx context.Context, projec
 			// For repos, we need to use the repo ref functions, as they update different scopes/roles.
 			repoRef := &model.RepoRef{ProjectRef: *newProjectRef}
 			if err = repoRef.UpdateAdminRoles(adminsToAdd, adminsToDelete); err != nil {
-				return errors.Wrap(err, "error updating repo admin roles")
+				catcher.Wrap(err, "error updating repo admin roles")
 			}
+			newProjectRef.Admins = repoRef.Admins
```
ablack12, line 96: Now that we're potentially always updating, we should move modified = true above the if/else statement

rest/data/project_settings_test.go
```diff
@@ -84,6 +84,52 @@ func TestSaveProjectSettingsForSectionForRepo(t *testing.T) {
 			assert.NotNil(t, newAdminFromDB)
 			assert.Contains(t, newAdminFromDB.Roles(), model.GetRepoAdminRole(ref.Id))
 		},
+		"Access with error": func(t *testing.T, ref model.RepoRef) {
```
ablack12, line 87: This test doe a lot of things that aren't relevant to what you need I think. This seems to test if markiing something restricted works, and it _also_ adds an admin. We should check the adding an admin case, removing an admin case, and the failure case. Right now I'm only seeing the success case but I'm not sure what error you're expecting at 99 (same with below)

bynn: test made clearer

rest/data/project_settings.go
bynn: done

model/repo_ref.go
bynn: done

model/project_ref.go
bynn: no need for a function but added using append

ablack12 (approved):

PR #5078, opened 2021-10-05
EVG-15434 Check if dbUser is nil after converting

bynn:
[EVG-15434](https://jira.mongodb.org/browse/EVG-15434)

### Description 
added an internal error which i think is the most appropriate 

### Testing 
na

### DIFF
--- a/rest/route/user.go
+++ b/rest/route/user.go
@@ -159,6 +159,9 @@ func (h *userPermissionsPostHandler) Run(ctx context.Context) gimlet.Responder {
 	if !valid {
 		return gimlet.NewTextInternalErrorResponse("unexpected type of user found")
 	}
+	if dbuser == nil {
+		return gimlet.NewTextInternalErrorResponse("error converting gimlet user to DBUser")
+	}
 	if err = dbuser.AddRole(newRole.ID); err != nil {
 		return gimlet.NewTextInternalErrorResponse(err.Error())
 	}


malikchaya2 (approved):
This looks good to me, but let's also double check with Annie since she is the one that spotted this.

rest/route/user.go
```diff
@@ -159,6 +159,9 @@ func (h *userPermissionsPostHandler) Run(ctx context.Context) gimlet.Responder {
 	if !valid {
 		return gimlet.NewTextInternalErrorResponse("unexpected type of user found")
 	}
+	if dbuser == nil {
+		return gimlet.NewTextInternalErrorResponse("error converting gimlet user to DBUser")
```
ablack12, line 163: Can you change the error to match the one at 148 closer? Really it is the same error, it's just that interfaces can be not nil while the underlying structure _is_ actually nil (#justgothings) https://tour.golang.org/methods/12

rest/route/user.go
```diff
@@ -159,6 +159,12 @@ func (h *userPermissionsPostHandler) Run(ctx context.Context) gimlet.Responder {
 	if !valid {
 		return gimlet.NewTextInternalErrorResponse("unexpected type of user found")
 	}
+	if dbuser == nil {
+		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+			Message:    fmt.Sprintf("unable to convert gimlet user '%s' to DBUser", h.userID),
```
ablack12, line 164: I think the message itself is still incorrect though

ablack12: no worries!

PR #5080, opened 2021-10-05
EVG-15295 Update LoadProjectInto to handle includes with modules

bynn:
[EVG-15295](https://jira.mongodb.org/browse/EVG-15295)

### Description 
if modules not empty, it will call retrieveFileForModule 

### Testing
must specify path for cli command
image

defined post in ex.yml so it would error if properly loaded 
image

same error by pulling from github using patch
image

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -589,6 +589,7 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 			}
 		}
 		var yaml []byte
+		opts.Identifier = identifier
 		if path.Module != "" {
 			opts.RemotePath = path.FileName
 			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module)
@@ -633,6 +634,7 @@ type GetProjectOpts struct {
 	Revision     string
 	Token        string
 	ReadFileFrom string
+	Identifier   string
 }
 
 type PatchOpts struct {
@@ -681,11 +683,11 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 		}
 		configFile, err := thirdparty.GetGithubFile(ctx, opts.Token, opts.Ref.Owner, opts.Ref.Repo, opts.RemotePath, opts.Revision)
 		if err != nil {
-			return nil, errors.Wrapf(err, "error fetching project file for '%s' at '%s'", opts.Ref.Id, opts.Revision)
+			return nil, errors.Wrapf(err, "error fetching project file for '%s' at '%s'", opts.Identifier, opts.Revision)
 		}
 		fileContents, err := base64.StdEncoding.DecodeString(*configFile.Content)
 		if err != nil {
-			return nil, errors.Wrapf(err, "unable to decode config file for '%s'", opts.Ref.Id)
+			return nil, errors.Wrapf(err, "unable to decode config file for '%s'", opts.Identifier)
 		}
 		return fileContents, nil
 	}
@@ -710,7 +712,6 @@ func retrieveFileForModule(ctx context.Context, opts GetProjectOpts, modules Mod
 	repoOwner, repoName := module.GetRepoOwnerAndName()
 	moduleOpts := GetProjectOpts{
 		Ref: &ProjectRef{
-			Id:    module.Name,
 			Owner: repoOwner,
 			Repo:  repoName,
 		},
@@ -718,6 +719,7 @@ func retrieveFileForModule(ctx context.Context, opts GetProjectOpts, modules Mod
 		Revision:     module.Branch,
 		Token:        opts.Token,
 		ReadFileFrom: ReadfromGithub,
+		Identifier:   moduleName,
 	}
 	return retrieveFile(ctx, moduleOpts)
 }
--- a/operations/validate.go
+++ b/operations/validate.go
@@ -28,7 +28,7 @@ func Validate() cli.Command {
 			Usage: "include long validation checks (only applies if the check is over some threshold, in which case a warning is issued)",
 		}, cli.StringSliceFlag{
 			Name:  joinFlagNames(localModulesFlagName, "lm"),
-			Usage: "specify a local modules as MODULE_NAME=PATH pairs",
+			Usage: "specify local modules as MODULE_NAME=PATH pairs",
 		}),
 		Before: mergeBeforeFuncs(setPlainLogger, requirePathFlag),
 		Action: func(c *cli.Context) error {


model/project.go
```diff
@@ -315,6 +315,16 @@ func (l *ModuleList) IsIdentical(m manifest.Manifest) bool {
 	return reflect.DeepEqual(manifestModules, projectModules)
 }
 
+func (l *ModuleList) GetModuleByName(find string) (*Module, error) {
```
ablack12, line 318: kind of a strange variable name since "find" is really a verb; maybe just "name"?

model/project.go
```diff
@@ -315,6 +315,16 @@ func (l *ModuleList) IsIdentical(m manifest.Manifest) bool {
...
+	for _, module := range *l {
+		if module.Name == find {
+			return &module, nil
+		}
+	}
+
+	return nil, errors.New("No such module on this project.")
```
ablack12, line 325: This error message would be more helpful with the actual name: "module '<module_name>' doesn't exist"

model/project_parser.go
```diff
@@ -581,15 +581,20 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-			if opts.PatchOpts.patch != nil && opts.PatchOpts.patch.ConfigChanged(path.FileName) {
-				opts.ReadFileFrom = ReadFromPatchDiff
-			} else {
-				opts.ReadFileFrom = ReadFromPatch
+		var yaml []byte
+		if path.Module != "" {
+			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module)
+		} else {
+			// read from the patch diff if a change has been made in any of the include files
+			if opts.ReadFileFrom == ReadFromPatch || opts.ReadFileFrom == ReadFromPatchDiff {
```
ablack12, line 589: I think this should still be done outside of the if/else, bc retrieveFileForModule will also eventually need to consider the differences here)

model/project_parser.go
```diff
@@ -685,6 +690,34 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
...
+		}
+		ghToken, err := conf.GetGithubOauthToken()
+		if err != nil {
+			return nil, errors.Wrap(err, "can't get Github OAuth token from configuration")
+		}
+		opts.Token = ghToken
+	}
+	configFile, err := thirdparty.GetGithubFile(ctx, opts.Token, repoOwner, repoName, opts.RemotePath, module.Branch)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error fetching project file for '%s' at '%s'", opts.Ref.Id, opts.Revision)
```
ablack12, line 712: this should be different for the module

model/project_parser.go
```diff
@@ -685,6 +690,34 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 	}
 }
 
+func retrieveFileForModule(ctx context.Context, opts GetProjectOpts, modules ModuleList, moduleName string) ([]byte, error) {
```
ablack12, line 693: Right now this does have a lot of copied code from retrieveFile, which I don't think is necessary. If we change GetProjectOpts to accept Owner and Repo strings (instead of Ref) then `retrieveFileForModule` can just set owner/repo and use retrieveFile, which would save you from repeating 699 - 719

bynn: everything uses retrieveFile

model/project_parser.go
bynn: done

model/project_parser.go
bynn: it should print module name and branch now

model/project.go
```diff
@@ -315,6 +315,16 @@ func (l *ModuleList) IsIdentical(m manifest.Manifest) bool {
...
 }
 
+func GetModuleByName(moduleList ModuleList, moduleName string) (*Module, error) {
+	for _, module := range moduleList {
+		if module.Name == moduleName {
+			return &module, nil
+		}
+	}
+
+	return nil, errors.Errorf("Module '%s' doesn't exist.", moduleName)
```
ablack12, line 325: remove period

model/project_parser.go
```diff
@@ -589,7 +588,28 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 				opts.ReadFileFrom = ReadFromPatch
 			}
 		}
-		yaml, err := retrieveFile(ctx, *opts)
+		var yaml []byte
+		if path.Module != "" {
+			module, err := GetModuleByName(intermediateProject.Modules, path.Module)
```
ablack12, line 593: you can still have a retrieveFileForModule that itself calls retrieveFile so that you can move this stuff out and make the function more readable.

model/project_parser.go
```diff
@@ -589,7 +588,28 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+			moduleOpts := GetProjectOpts{
+				Ref: &ProjectRef{
+					Id:    module.Name,
+					Owner: repoOwner,
+					Repo:  repoName,
+				},
+				RemotePath:   path.FileName,
+				Revision:     module.Branch,
+				Token:        opts.Token,
+				ReadFileFrom: ReadfromGithub,
```
ablack12, line 607: We will eventually need to figure out how to handle the other cases but for now we can do just this one.

ablack12 (changes requested):
lmk if you need help testing this

model/project_parser.go
bynn: checks for local modules first

model/project_parser.go
```diff
@@ -643,6 +627,7 @@ const (
 type GetProjectOpts struct {
 	Ref          *ProjectRef
 	PatchOpts    *PatchOpts
+	LocalModules *map[string]string
```
ablack12, line 630: We don't typically make maps pointers (not really a reason to)

model/project_parser.go
```diff
@@ -705,6 +690,37 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
 	}
 }
 
+func retrieveFileForModule(ctx context.Context, opts GetProjectOpts, modules ModuleList, moduleName, path string) ([]byte, error) {
+	// Look through any given local modules first
+	if opts.LocalModules != nil {
+		if val, ok := (*opts.LocalModules)[moduleName]; ok {
+			moduleOpts := GetProjectOpts{
+				RemotePath:   fmt.Sprintf("%s/%s", val, path),
```
ablack12, line 698: val here actually is the path to module, right?

operations/validate.go
```diff
@@ -14,6 +15,10 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
+const (
+	localModuleFlagName = "module"
```
ablack12, line 19: Can you define this in the same place as longFlagName?

operations/validate.go
```diff
@@ -60,17 +73,29 @@ func Validate() cli.Command {
...
+			return validateFile(path, ac, quiet, long, localModuleMap)
 		},
 	}
 }
+func getLocalModulesFromInput(localModulePaths []string) (*map[string]string, error) {
+	moduleMap := make(map[string]string)
+	catcher := grip.NewBasicCatcher()
+	for _, module := range localModulePaths {
+		pair := strings.Split(module, "=")
+		if len(pair) < 2 {
```
ablack12, line 90: really it's if it != 2 exactly

operations/validate.go
```diff
@@ -60,17 +73,29 @@ func Validate() cli.Command {
...
+			catcher.Add(errors.Errorf("problem parsing local module '%s'", module))
```
ablack12, line 91: catcher.Errorf (and maybe explicitly say why it's an invalid format)

ablack12 (commented):
This is looking good! Let me know if you want to discuss testing.

model/project_parser.go
bynn: i thought it would be easier and faster to check this instead of an empty map

model/project_parser.go
bynn: yes, name clarification

operations/validate.go
bynn: done

operations/validate.go
bynn: done

operations/validate.go
bynn: done

model/project_parser.go
ablack12: This doesn't look updated

operations/validate.go
```diff
@@ -25,13 +26,21 @@ func Validate() cli.Command {
 		}, cli.BoolFlag{
 			Name:  joinFlagNames(longFlagName, "l"),
 			Usage: "include long validation checks (only applies if the check is over some threshold, in which case a warning is issued)",
+		}, cli.StringSliceFlag{
+			Name:  joinFlagNames(localModulesFlagName, "lm"),
+			Usage: "specify a local module as a MODULE_NAME=PATH pair",
```
ablack12, line 31: since this takes in a list, should use plurals ("modules", "pairs")

operations/validate.go
```diff
@@ -60,17 +69,29 @@ func Validate() cli.Command {
...
 		},
 	}
 }
+func getLocalModulesFromInput(localModulePaths []string) (*map[string]string, error) {
+	moduleMap := make(map[string]string)
+	catcher := grip.NewBasicCatcher()
+	for _, module := range localModulePaths {
+		pair := strings.Split(module, "=")
+		if len(pair) != 2 {
+			catcher.Errorf("expected only 1 '=' sign while parsing local module '%s'", module)
```
ablack12, line 87: nit: should type out one

operations/validate.go
```diff
@@ -60,17 +69,29 @@ func Validate() cli.Command {
...
 }
+func getLocalModulesFromInput(localModulePaths []string) (map[string]string, error) {
+	moduleMap := make(map[string]string)
+	catcher := grip.NewBasicCatcher()
+	for _, module := range localModulePaths {
+		pair := strings.Split(module, "=")
+		if len(pair) != 2 {
+			catcher.Errorf("expected only one '=' sign while parsing local module '%s'", module)
+		}
+		moduleMap[pair[0]] = pair[1]
```
ablack12, line 89: should do this in an else if so we don't panic

model/project_parser.go
```diff
@@ -589,7 +588,12 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
 		}
-		yaml, err := retrieveFile(ctx, *opts)
+		var yaml []byte
+		if path.Module != "" {
+			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module, path.FileName)
+		} else {
+			yaml, err = retrieveFile(ctx, *opts)
+		}
 		if err != nil {
 			return intermediateProject, errors.Wrapf(err, LoadProjectError)
```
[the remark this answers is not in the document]
bynn: does not wraf LoadProjectError anymore 
i hope this doesnt cause any confusions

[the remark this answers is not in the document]
ablack12: could do `errors.Wrapf(err, "%s: failed to retrieve file '%s'", LoadProjectError, path.FileName)`

model/project_parser.go
```diff
@@ -685,6 +691,37 @@ func retrieveFile(ctx context.Context, opts GetProjectOpts) ([]byte, error) {
...
+	}
+	// Retrieve from github
+	module, err := GetModuleByName(modules, moduleName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "can't get module for module name '%s'", moduleName)
+	}
+	repoOwner, repoName := module.GetRepoOwnerAndName()
+	moduleOpts := GetProjectOpts{
+		Ref: &ProjectRef{
+			Id:    module.Name,
```
ablack12, line 713: This isn't the ID actually, but do we use this? I think we can just remove

PR #5082, opened 2021-10-06
EVG-15323 route to get users for resource ID

ablack12:
[EVG-15323](https://jira.mongodb.org/browse/EVG-15323)

### Description 
Mana 2 needs a route that will return all users that have access to a particular resource (ex: a particular project) and their highest available permissions.

### Testing 
Unit test galore

### DIFF
--- a/rest/route/user.go
+++ b/rest/route/user.go
@@ -294,6 +294,110 @@ func (h *userPermissionsDeleteHandler) Run(ctx context.Context) gimlet.Responder
 	return gimlet.NewJSONResponse(struct{}{})
 }
 
+////////////////////////////////////////////////////////////////////////
+//
+// GET /users/permissions
+
+type UsersPermissionsInput struct {
+	ResourceId   string `json:"resource_id"`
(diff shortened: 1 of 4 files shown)


rest/route/user.go
```diff
@@ -294,6 +294,110 @@ func (h *userPermissionsDeleteHandler) Run(ctx context.Context) gimlet.Responder
...
+	for _, u := range usersWithRoles {
+		for _, userRole := range u.SystemRoles {
+			permissions, ok := permissionsMap[userRole]
+			if ok {
+				res[u.Username()] = getMaxPermissions(res[u.Username()], permissions)
+			}
+		}
+	}
+
+	return gimlet.NewJSONResponse(res)
```
bynn, line 381: would we want to return an error if we find nothing? or is the empty response fine?

ablack12: Typically finding nothing is still supposed to be a successful response, as long as the input isn't malformed.

PR #5086, opened 2021-10-07
EVG-15045 add SaveProjectSettingsForSection resolver

ablack12:
[EVG-15045](https://jira.mongodb.org/browse/EVG-15045 )

### Description 
Add inputs necessary for ProjectSettings and add SaveProjectSettingsForSection. Changed backend function to return the current state of the project settings (in case only some of the saves were successful or something). Also modified the ProjectSettings resolver to use an isRepo boolean for consistency. 

### Testing 
Added unit tests -- didn't test every section since the backend tests are pretty robust but if there are edge cases we can think of I can add them.

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -363,7 +363,7 @@ input ProjectSettingsInput {
 }
 
 input ProjectInput {
-  id: String
+  id: String!
   identifier: String
   displayName: String
   enabled: Boolean
--- a/graphql/generated.go
+++ b/graphql/generated.go
@@ -6719,7 +6719,7 @@ input ProjectSettingsInput {
 }
 
 input ProjectInput {
-  id: String
+  id: String!
   identifier: String
   displayName: String
   enabled: Boolean
@@ -33372,7 +33372,7 @@ func (ec *executionContext) unmarshalInputProjectInput(ctx context.Context, obj 
 		switch k {
 		case "id":
 			var err error
-			it.Id, err = ec.unmarshalOString2ᚖstring(ctx, v)
+			it.Id, err = ec.unmarshalNString2ᚖstring(ctx, v)
 			if err != nil {
 				return it, err
 			}


graphql/resolvers.go
```diff
@@ -1026,22 +1026,25 @@ func (r *queryResolver) MyHosts(ctx context.Context) ([]*restModel.APIHost, erro
 	return apiHosts, nil
 }
 
-func (r *queryResolver) ProjectSettings(ctx context.Context, identifier string) (*restModel.APIProjectSettings, error) {
+func (r *queryResolver) ProjectSettings(ctx context.Context, identifier string, isRepo bool) (*restModel.APIProjectSettings, error) {
```
malikchaya2, line 1029: Would the workflow for this on the ui be: 
ProjectSettings = ProjectSettings(ctx, sandbox, false)
RepoSettings = ProjectSettings(ctx, sandbox, true)

instead of  doing it this way? 
ProjectSettings = ProjectSettings(ctx, sandbox)
RepoSettings = RepoSettings(ctx, sandbox)

graphql/resolvers.go
```diff
@@ -2150,6 +2153,16 @@ func (r *taskQueueItemResolver) Requester(ctx context.Context, obj *restModel.AP
 	return TaskQueueItemTypeCommit, nil
 }
 
+func (r *mutationResolver) SaveProjectSettingsForSection(ctx context.Context, obj *restModel.APIProjectSettings, section string, isRepo bool) (*restModel.APIProjectSettings, error) {
```
malikchaya2, line 2156: I'm afraid that the isRepo bool as opposed to a SaveProjectSettingsForSection and SaveRepotSettingsForSection (and same for the ProjectSettings resolver) will require a steeper learning curve for new engineers reading the code. It's not a huge difference but I feel like separating them makes the code a lot easier to follow.

graphql/resolvers.go
ablack12: yep - we can make a RepoSettings one if you want instead but then we have to repeat all of the resolvers I'm assuming

graphql/resolvers.go
ablack12: I'm a little anti duplicating code and they are pretty much identical. We can change it but then follow up work will need to follow the same convention. You'll be working with it the most though so whichever convention you prefer.

graphql/tests/saveProjectSettingsForSection/queries/varsSection.graphql
```diff
@@ -0,0 +1,20 @@
...
+            projectRef: {
+                id: "1234"
+            }
+            vars: {
+                vars: {hello: "world", goodbye: "now"},
+                privateVarsList: ["hello", "goodbye"]
+            }
+        }
+        section: "variables",
+        isRepo: false
```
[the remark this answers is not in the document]
ablack12: Nah I just haven't written isRepo tests yet; it works the same for each, just looks up a different collection. I can add another test (unless you do want the function names to be different in which case I guess that can be handled in a different ticket)

graphql/resolvers.go
sophstad: Hahaha too much pressure! I'm ambivalent but it does feel little more straightforward to work two differently named functions like Chaya suggested. Maybe they can call the same util function with a boolean to avoid duplication? 😄

graphql/schema.graphql
```diff
@@ -363,7 +363,7 @@ input ProjectSettingsInput {
 }
 
 input ProjectInput {
-  id: String!
+  id: String
```
malikchaya2, line 366: Why the change to make this not required?

ablack12: This fails for CreateProject, because it also uses ProjectInput and ID wouldn't be passed in in that case.

ablack12: Should I switch so create project uses like a CreateProjectInput or something?

malikchaya2: Ah that makes sense. duplicating it feels overkill, I wouldn't do that. maybe it's okay for it not to be required. @sophstad any thoughts on this?

PR #5087, opened 2021-10-08
EVG-15481 filter out basic roles from permissions route

ablack12:
[EVG-15481](https://jira.mongodb.org/browse/EVG-15481)

### Description 
This route returns all projects that the user has permissions for, and what the highest permissions are. Since every user by default has basic project access to almost every project, this return is super long and pretty unnecessary. Mana 2 wants to filter the basic project access out.

### Testing 
Added basic roles to test and verified the result didn't have them.

### DIFF
--- a/rest/route/user.go
+++ b/rest/route/user.go
@@ -445,7 +445,9 @@ func (h *userPermissionsGetHandler) Run(ctx context.Context) gimlet.Responder {
 	if u == nil {
 		return gimlet.NewJSONErrorResponse(errors.New("user not found"))
 	}
-	permissions, err := rolemanager.PermissionSummaryForRoles(ctx, u.Roles(), h.rm)
+	rolesToSearch, _ := utility.StringSliceSymmetricDifference(u.SystemRoles, evergreen.BasicAccessRoles)
+	// filter out the roles that everybody has automatically
+	permissions, err := rolemanager.PermissionSummaryForRoles(ctx, rolesToSearch, h.rm)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


rest/route/user.go
```diff
@@ -445,7 +445,9 @@ func (h *userPermissionsGetHandler) Run(ctx context.Context) gimlet.Responder {
 	if u == nil {
 		return gimlet.NewJSONErrorResponse(errors.New("user not found"))
 	}
-	permissions, err := rolemanager.PermissionSummaryForRoles(ctx, u.Roles(), h.rm)
+	rolesToSearch, _ := utility.StringSliceSymmetricDifference(u.SystemRoles, evergreen.BasicAccessRoles)
```
malikchaya2, line 448: I'm a bit confused about how this works because I couldn't find where u.SystemRoles is populated. 
Will u.SystemRoles contain every single available role?

ablack12: Does that also help answer the second question? This function returns u.SystemRoles minus the basic roles.

ablack12: Nope it doesn't: you can read what happens [here](https://github.com/evergreen-ci/evergreen/blob/5b01d31f9ab5d44c2a959ac45741efcc77aeae12/vendor/github.com/evergreen-ci/utility/slice.go#L40). It returns two lists: the first is everything that is in A that is not in B, and the second is everything that is in B that is not in A. Here's the case you're describing: https://play.golang.org/p/tm0LVDrr5Zm

PR #5089, opened 2021-10-08
EVG-15462: add method to drop a collection

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15462

### Description 
* Add method to drop a collection for test setup/teardown when the test adds both docs and indexes. This makes it easier to ensure that the DB state is actually cleared after a test, especially to get rid of indexes that could affect an unrelated test.
* Make `DropAllIndexes` succeed even if the collection does not exist. Before, if the collection did not already exist, using `DropAllIndexes` would fail.

### Testing 
Updated existing DB tests that create indexes.

### DIFF
--- a/db/db_utils.go
+++ b/db/db_utils.go
@@ -148,6 +148,8 @@ func EnsureIndex(collection string, index mongo.IndexModel) error {
 	return errors.WithStack(err)
 }
 
+const errorCodeNamespaceNotFound = 26
+
 // DropAllIndexes drops all indexes in the specified collections, returning an
 // error immediately if dropping the indexes in any one of them fails.
 func DropAllIndexes(collections ...string) error {
@@ -162,7 +164,7 @@ func DropAllIndexes(collections ...string) error {
 			if mongoErr, ok := err.(driver.Error); ok && mongoErr.NamespaceNotFound() {
 				continue
 			}
-			if cmdErr, ok := err.(mongo.CommandError); ok && cmdErr.HasErrorCode(26) {
+			if cmdErr, ok := err.(mongo.CommandError); ok && cmdErr.HasErrorCode(errorCodeNamespaceNotFound) {
 				continue
 			}
 			return errors.Wrapf(err, "dropping indexes in collection '%s'", coll)


db/db_utils.go
```diff
@@ -131,13 +148,27 @@ func EnsureIndex(collection string, index mongo.IndexModel) error {
...
-	return errors.WithStack(err)
+	for _, coll := range collections {
+		if _, err := env.DB().Collection(coll).Indexes().DropAll(ctx); err != nil {
+			// DropAll errors if the collection does not exist, so make this
+			// idempotent by ignoring the error for the case of a nonexistent
+			// collection.
+			if mongoErr, ok := err.(driver.Error); ok && mongoErr.NamespaceNotFound() {
+				continue
+			}
+			if cmdErr, ok := err.(mongo.CommandError); ok && cmdErr.HasErrorCode(26) {
```
ybrill, line 165: what does this one mean?

model/testutil/indexes.go
```diff
@@ -4,17 +4,13 @@ import (
...
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-// AddTestIndexes drops all existing indexes and adds indexes for a given
-// collection
+// AddTestIndexes adds indexes in a given collection for testing.
 func AddTestIndexes(collection string, unique, sparse bool, key ...string) error {
-	catcher := grip.NewBasicCatcher()
-	catcher.Add(db.DropAllIndexes(collection))
```
ybrill, line 17: why don't we want to drop existing indexes anymore?

db/db_utils.go
```diff
@@ -131,13 +148,27 @@ func EnsureIndex(collection string, index mongo.IndexModel) error {
 	return errors.WithStack(err)
 }
 
-// DropIndex takes in a collection and a slice of keys and drops those indexes
-func DropAllIndexes(collection string) error {
+// DropAllIndexes drops all indexes in the specified collections, returning an
```
ybrill, line 151: is there ever a case where we'd use this instead of just dropping the collections?

db/db_utils.go
Kimchelly: It's mongo driver is pretty opaque about which error is which and when each will be returned. The [docs for driver.Error](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.7.3/x/mongo/driver#Error) seem to imply it's some catch-all for command execution errors. However, if you read the [docs for mongo.CommandError](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.7.3/mongo#CommandError), it's _also_ for command execution errors.

To be clear, error-checking for specific cases in MongoDB is basically always hacky like this. Anytime I've seen a piece of code check for specific error codes, it always has to check many cases because the error-handling is not unified very well.

PR #5090, opened 2021-10-08
EVG-14930 Fetch unmatching commits 

khelif96:
[EVG-14930](https://jira.mongodb.org/browse/EVG-14930)

### Description 
This works the same way as https://github.com/evergreen-ci/evergreen/pull/5038 But is a little more refined and less error prone.
We will stop fetching mainline commits after we have searched 300 commits. To prevent endlessly checking the versions collections

### Testing 
I wrote some programmatic graphql tests. Instead of the static comparison based ones we use elsewhere

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3294,7 +3294,7 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 				shouldCollapse = true
 			}
 		}
-		// If a version matches our filter critera we append it directly to our returned list of mainlineCommits
+		// If a version matches our filter criteria we append it directly to our returned list of mainlineCommits
 		if !shouldCollapse {
 			matchingVersionCount++
 			mainlineCommits.NextPageOrderNumber = utility.ToIntPtr(v.RevisionOrderNumber)
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3222,31 +3222,17 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 		limit = utility.FromIntPtr(options.Limit)
 	}
 	opts := model.MainlineCommitVersionOptions{
-		Activated:       true,
 		Limit:           limit,
 		SkipOrderNumber: utility.FromIntPtr(options.SkipOrderNumber),
 	}
 
-	activatedVersions, err := model.GetMainlineCommitVersionsWithOptions(projectId, opts)
+	versions, err := model.GetMainlineCommitVersionsWithOptions(projectId, opts)
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Error getting activated versions: %s", err.Error()))
 	}
 
-	opts = model.MainlineCommitVersionOptions{
-		Activated:       false,
-		SkipOrderNumber: utility.FromIntPtr(options.SkipOrderNumber),
-	}
-	unactivatedVersions, err := model.GetMainlineCommitVersionsWithOptions(projectId, opts)
-	if err != nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Error getting unactivated versions: %s", err.Error()))
-	}
-
-	versions := append(activatedVersions, unactivatedVersions...)
-	sort.Slice(versions, func(i, j int) bool {
-		return versions[i].RevisionOrderNumber > versions[j].RevisionOrderNumber
-	})
 	var mainlineCommits MainlineCommits
-	activatedVersionCount := 0
+	matchingVersionCount := 0
 
 	// We only want to return the PrevPageOrderNumber if a user is not on the first page
 	if options.SkipOrderNumber != nil {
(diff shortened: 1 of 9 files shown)


graphql/resolvers.go
```diff
@@ -3266,10 +3252,23 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+	for matchingVersionCount < limit {
+		// If we no longer have any more versions to check break out and return what we have
+		if index >= len(versions) {
 			break
 		}
+		// If we have checked more versions then the MaxMainlineCommitVersionLimit then break out and return what we have
+		if versionsCheckedCount >= model.MaxMainlineCommitVersionLimit {
+			// Return an error if we did not find any versions that match
+			if matchingVersionCount == 0 {
+				return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Error getting activated versions: exceeded max limit of %d versions", model.MaxMainlineCommitVersionLimit))
```
ablack12, line 3266: This error won't really make sense to users. Maybe "matching version not found in %d most recent versions" ? But also I think this case would get caught by 3259, right? If versions is empty even the first loop we'd break at 3259.

graphql/resolvers.go
```diff
@@ -3284,10 +3283,26 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 			}
 		}
 		mainlineCommitVersion := MainlineCommitVersion{}
-
+		shouldCollapse := false
+		if !utility.FromBoolPtr(v.Activated) {
+			shouldCollapse = true
+		} else if buildVariantOptions != nil && (len(buildVariantOptions.Tasks) > 0 || len(buildVariantOptions.Variants) > 0 || len(buildVariantOptions.Statuses) > 0) {
```
ablack12, line 3289: Can we have HasMatchingTasks handle this (since it's implicitly true that it does if we aren't filtering)? Or at least make a buildVariantOptions helper method for this, like IsPopulated()

graphql/resolvers.go
```diff
@@ -3284,10 +3283,26 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+			}
+			hasTasks, err := task.HasMatchingTasks(v.Id, opts)
+			if err != nil {
+				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error checking if version has tasks: %s", err.Error()))
+			}
+			if !hasTasks {
+				shouldCollapse = true
+			}
+		}
 		// If a version is activated we append it directly to our returned list of mainlineCommits
```
ablack12, line 3303: this comment should now be updated and possibly moved up

graphql/resolvers.go
```diff
@@ -3266,10 +3252,23 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 		}
 	}
 
-	for _, v := range versions {
-		if activatedVersionCount == limit {
+	index := 0
+	versionsCheckedCount := 0
+	for matchingVersionCount < limit {
```
ablack12, line 3257: Can we add a comment here to describe the loop, i.e. something about how we're looping through each version returned from GetMainlineCommitVersionsWithOptions and will call the query again if we run out? We don't use while loops often so i think it'll help us for the future

model/task/task.go
```diff
@@ -3077,6 +3077,42 @@ func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) ([]*
 	return StatusCount, nil
 }
 
+type HasMatchingTasksOptions struct {
+	TaskNames []string
+	Variants  []string
+	Statuses  []string
+}
+
+// HasMatchingTasks returns true if the version has tasks with the given statuses,
```
ablack12, line 3086: nit: comma

graphql/resolvers.go
```diff
@@ -3316,6 +3331,21 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 		if mainlineCommitVersion.Version != nil || mainlineCommitVersion.RolledUpVersions != nil {
 			mainlineCommits.Versions = append(mainlineCommits.Versions, &mainlineCommitVersion)
 		}
+		index++
+		// If we have exhausted all of our versions we should fetch some more.
+		if index == len(versions) {
```
ablack12, line 3336: Right now we'll still be doing this query even if this is the version that gets us equal to limit, so we should either move this to the top of the loop or add an if to make sure we aren't doing that

graphql/resolvers.go
```diff
@@ -3266,10 +3252,23 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
-	for _, v := range versions {
-		if activatedVersionCount == limit {
+	index := 0
+	versionsCheckedCount := 0
+	for matchingVersionCount < limit {
+		// If we no longer have any more versions to check break out and return what we have
+		if index >= len(versions) {
 			break
 		}
+		// If we have checked more versions then the MaxMainlineCommitVersionLimit then break out and return what we have
```
ablack12, line 3262: nit: than

ablack12 (approved):

PR #5091, opened 2021-10-08
EVG-15543 use timestamp for project activation

ablack12:
[EVG-15543](https://jira.mongodb.org/browse/EVG-15543)

### Description 
If we have a cron for tasks/variants to be activated at 8 pm daily, the 8 pm project-activation job should get the most recent version (from before 8 pm), which is set to be activated at 8 pm (this job runs every 10 minutes). However, if the job takes a minute to get there, and a new commit is pushed at 8:01 pm, we'll instead look at _this_ version since we just look at the most recent. Since this commit was pushed after 8 pm, the cron for it will be set to 8 pm for the NEXT day -- this is correct, however now we miss our window for activating the current day.

By using the timestamp that the job represents, we avoid this race. 

### Testing 
Added a unit test to verify I didn't break the query.

### DIFF
--- a/model/version_activation.go
+++ b/model/version_activation.go
@@ -11,9 +11,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-func DoProjectActivation(id string) (bool, error) {
-	// fetch the most recent, non-ignored version to activate
-	activateVersion, err := VersionFindOne(VersionByMostRecentNonIgnored(id))
+func DoProjectActivation(id string, ts time.Time) (bool, error) {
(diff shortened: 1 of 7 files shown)


malikchaya2:
> The query uses `VersionCreateTimeKey: bson.M{"$lte": ts},` i.e. it does search for anything before the given timestamp, and then sorts so we get the most recent.

Ah, I was confused by the test but I understand now.

ablack12:
Oh gotcha -- yeah I wanted to test that specific edge case.

PR #5094, opened 2021-10-11
EVG-14856 Refactor execution task resolver to perform one query instead of several

khelif96:
[EVG-14856](https://jira.mongodb.org/browse/EVG-14856)

### Description 
Refactors the execution tasks resolver to perform between 1-2 db calls instead of several for each execution task.  
For an improvement of  O(N) -> O(1) db calls in the best case scenario where all executions line up.
And O(2N) -> O(2) when the executions are mismatching. 

### Testing 
  Unit tests

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3092,37 +3092,23 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
 	return apiTask, nil
 }
 func (r *taskResolver) ExecutionTasksFull(ctx context.Context, obj *restModel.APITask) ([]*restModel.APITask, error) {
-	i, err := obj.ToService()
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting service model for APITask %s: %s", *obj.Id, err.Error()))
-	}
-	t, ok := i.(*task.Task)
-	if !ok {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert APITask %s to Task", *obj.Id))
-	}
-	if len(t.ExecutionTasks) == 0 {
+	if len(obj.ExecutionTasks) == 0 {
 		return nil, nil
 	}
-	executionTasks := []*restModel.APITask{}
-	for _, execTaskID := range t.ExecutionTasks {
-		execT, err := task.FindOneIdAndExecutionWithDisplayStatus(execTaskID, &t.Execution)
-		if err != nil {
-			// The task is not found, possibly because the execution is out of sync with the display task. Get the latest instead.
-			execT, err = task.FindOneIdAndExecutionWithDisplayStatus(execTaskID, nil)
-			if err != nil {
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while getting execution task with id: %s : %s", execTaskID, err.Error()))
-			}
-		}
-		if execT != nil {
-			apiTask := &restModel.APITask{}
-			if err = apiTask.BuildFromService(execT); err != nil {
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert task: %s to APITask", execT.Id))
-			}
-			executionTasks = append(executionTasks, apiTask)
-		}
+	tasks, err := task.ByExecutionTasksAndMaxExecution(obj.ExecutionTasks, obj.Execution)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding execution tasks for task: %s : %s", *obj.Id, err.Error()))
 	}
-
-	return executionTasks, nil
+	apiTasks := []*restModel.APITask{}
+	for _, t := range tasks {
+		apiTask := &restModel.APITask{}
+		err = apiTask.BuildFromService(&t)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


model/task/db.go
```diff
@@ -812,6 +812,70 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
+					"$in": missingTasks,
+				},
+				ExecutionKey: bson.M{
+					"$lte": execution,
+				},
+			},
+		}
+		oldTaskPipeline = append(oldTaskPipeline, match)
+		// filter by the old taskids that have the highest execution
+		sort := bson.M{
```
SupaJoon, line 861: What's the reason for sorting old tasks and not the "new" tasks?

graphql/resolvers.go
```diff
@@ -3092,37 +3092,23 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
...
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding execution tasks for task: %s : %s", *obj.Id, err.Error()))
+	}
+	apiTasks := []*restModel.APITask{}
+	for _, t := range tasks {
+		apiTask := &restModel.APITask{}
+		err = apiTask.BuildFromService(&t)
 		if err != nil {
-			// The task is not found, possibly because the execution is out of sync with the display task. Get the latest instead.
-			execT, err = task.FindOneIdAndExecutionWithDisplayStatus(execTaskID, nil)
```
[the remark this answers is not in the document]
khelif96: They do. This is what https://github.com/evergreen-ci/evergreen/pull/5094/files#diff-c3e271afbbec24ae0ffa3c35d94434a355d20028b609ede4dc0048e514fae202R846 does.

graphql/resolvers.go
```diff
@@ -3092,37 +3092,23 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
...
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while getting execution task with id: %s : %s", execTaskID, err.Error()))
-			}
-		}
-		if execT != nil {
-			apiTask := &restModel.APITask{}
-			if err = apiTask.BuildFromService(execT); err != nil {
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert task: %s to APITask", execT.Id))
-			}
-			executionTasks = append(executionTasks, apiTask)
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert task %s to APITask : %s", t.Id, err))
```
ablack12, line 3107: err.Error()

model/task/db.go
```diff
@@ -812,6 +812,63 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
 	return result, nil
 }
 
+// ByExecutionTasksAndMaxExecution returns the tasks corresponding to the passed in taskIds and execution,
+// or the most recent executions of those tasks if they do not have a matching execution
+func ByExecutionTasksAndMaxExecution(taskIds []*string, execution int) ([]Task, error) {
```
ablack12, line 817: Can you call this FindBy... ? Typically when our DB functions are just By, they only return the query and then they can be used with Find, Count, whatever, but this includes the Find inside the function.

model/task/db.go
```diff
@@ -812,6 +812,63 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
+			IdKey: bson.M{
+				"$in": taskIds,
+			},
+			ExecutionKey: bson.M{
+				"$lte": execution,
+			},
+		},
+	}
+	pipeline = append(pipeline, match)
+	// Get the taskIds that were not found in the previous match stage
```
ablack12, line 830: does it make sense to move this comment after the aggregation?

model/task/db.go
```diff
@@ -812,6 +812,63 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
+	result := []Task{}
+	if err := Aggregate(pipeline, &result); err != nil {
+		return nil, errors.Wrap(err, "Error finding tasks in task collection")
+	}
+	taskMap := map[string]bool{}
+	for _, t := range result {
+		taskMap[t.Id] = true
+	}
+	missingTasks := []string{}
```
ablack12, line 839: You can shorten this up by building a list of foundTaskIds and then call `missingTasks, _ := utility.StringSliceSymmetricDifference(taskIds, foundTaskIds)` because the first return value is all strings that exist in the first list (taskIds) but not in the second (found task Ids).

model/task/db.go
```diff
@@ -812,6 +812,63 @@ func GetRecentTaskStats(period time.Duration, nameKey string) ([]StatusItem, err
...
+		if err := db.Aggregate(OldCollection, oldTaskPipeline, &oldTasks); err != nil {
+			return nil, errors.Wrap(err, "error finding tasks in old tasks collection")
+		}
+
+		result = append(result, oldTasks...)
+	}
+	if len(result) == 0 {
+		return nil, nil
+	}
+	return result, nil
```
ablack12, line 869: Should we sort this on anything?  / does it not matter that we aren't doing WithDisplayStatus anymore 🤔

khelif96: I don't think so we don't sort execution tasks anywhere else. The only place these results need to be sorted is on the execution tasks tab but that is handled client side.

PR #5095, opened 2021-10-11
EVG-15478, EVG-15491: update cross compilation

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-15478
https://jira.mongodb.org/browse/EVG-15491

### Description 
* Remove the install flag (`-i`), since it's deprecated and not necessary.
* Remove compilation with race detector enabled since Evergreen does not compile when the race detector is enabled.
* Disable cgo for non-race detector cases so that we don't depend on glibc. Doing so makes it possible to run the evergreen binary on non-glibc systems such as Alpine Linux, which is nice because user-provided containers might be based on non-glibc images.
    * There's some risk that this will disable some feature that we need, but I don't think this will be particularly problematic in reality. The standard library doesn't use cgo much (except in the net and os/user packages) and the only library we use that uses cgo much is gopsutil. Most Evergreen devs build with cgo disabled. Also, cross-compiling to a different OS/arch disables cgo, so our dists are already compiled without cgo on all platforms (minus Linux). The fact that people have successfully deployed custom builds (and the fact cross compiling disables cgo) makes me think that Linux also does not depend on cgo for any specific functionality.
    * From a future development standpoint, there's also the chance that someone will want to implement something and it requires cgo. I can't currently imagine a scenario where this is necessary since we never do low-level things that require C bindings. If we ever have such a need, we could revisit compiling with cgo, but I don't see any need for it currently.
* Use trimpath to remove absolute file paths.
* As a side effect of disabling cgo and using trimpath, our compiled builds are now reproducible.

### Testing 
People already compiled and deploy to staging with cgo disabled and I've deployed to staging with trimpath.

### DIFF
--- a/makefile
+++ b/makefile
@@ -35,6 +35,12 @@ export GOROOT := $(shell cygpath -m $(GOROOT))
 endif
 
 export GO111MODULE := off
+ifneq (,$(RACE_DETECTOR))
+# cgo is required for using the race detector.
+export CGO_ENABLED=1
+else
+export CGO_ENABLED=0
+endif
 # end go runtime settings
 
 


cmd/build-cross-compile/build-cross-compile.go
```diff
@@ -33,7 +19,6 @@ func main() {
 		buildName string
 		output    string
 		goBin     string
-		race      bool
```
ybrill, line 36: is the idea here that we don't need to build with `-race` to be able to test with `-race`?

Kimchelly: 1. `go build` with race does the same as what race detector goes when you run `go test` - it prints a loud warning when your program races. https://go.dev/blog/race-detector
2. We've never built with race detection enabled.

ybrill: 1. so go build will actually run the program to detect races? Or the resultant executable exits when it sees a race?
2. weren't we calling go build with `-race` in the case where `RACE_DETECTOR` was set before this change?

makefile
```diff
@@ -35,6 +35,12 @@ export GOROOT := $(shell cygpath -m $(GOROOT))
 endif
 
 export GO111MODULE := off
+ifneq (,$(RACE_DETECTOR))
+# cgo is required for using the race detector.
```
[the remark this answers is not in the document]
Kimchelly: The go binary refuses to run your test if you have `-race` and `CGO_ENABLED=0`.

PR #5097, opened 2021-10-12
EVG-15207: Merge aliases from parser project on a key name basis when calling alias DB

hadjri:
[EVG-15207](https://jira.mongodb.org/browse/EVG-15207)

### Description 
Change has been made to add support for defining aliases in the project yaml and incorporate it in the project parser, and then when we use aliases defined on a project we take the aliases in the DB and the aliases defined in the yaml and merge them. The project parser takes priority in that if the project yaml has any keys with this name, we ignore any keys with that name in the project ref.
### Testing 
Added unit tests for merge behavior.

### DIFF
--- a/model/project_parser_db.go
+++ b/model/project_parser_db.go
@@ -1,8 +1,6 @@
 package model
 
 import (
-	"fmt"
-
 	"github.com/evergreen-ci/evergreen/db"
 	"github.com/mongodb/anser/bsonutil"
 	adb "github.com/mongodb/anser/db"
@@ -79,7 +77,7 @@ func ParserProjectByVersion(projectId string, version string) (*ParserProject, e
 				"project_id": projectId,
 				"version":    version,
 			}))
-			return nil, errors.Wrap(err, fmt.Sprintf("Unable to retrieve last good version for project '%s'", projectId))
+			return nil, errors.Wrapf(err, "Unable to retrieve last good version for project '%s'", projectId)
 		}
 		version = lastGoodVersion.Id
 		lookupVersion = true
@@ -92,7 +90,7 @@ func ParserProjectByVersion(projectId string, version string) (*ParserProject, e
 			"version":        version,
 			"lookup_version": lookupVersion,
 		}))
-		return nil, errors.Wrap(err, fmt.Sprintf("Error retrieving parser project for version '%s'", version))
+		return nil, errors.Wrapf(err, "Error retrieving parser project for version '%s'", version)
 	}
 	return parserProject, nil
 }


model/project_aliases.go
```diff
@@ -71,22 +71,30 @@ type ProjectAlias struct {
...
-	out := []ProjectAlias{}
+func FindAliasesForProject(projectID string, version string) ([]ProjectAlias, error) {
+	var out []ProjectAlias
 	q := db.Query(bson.M{
 		projectIDKey: projectID,
 	})
 	err := db.FindAllQ(ProjectAliasCollection, q, &out)
 	if err != nil {
-		return nil, err
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project aliases")
```
ablack12, line 81: we can still return nil for these error cases

model/project_aliases.go
```diff
@@ -71,22 +71,30 @@ type ProjectAlias struct {
 type ProjectAliases []ProjectAlias
 
 // FindAliasesForProject fetches all aliases for a given project
-func FindAliasesForProject(projectID string) ([]ProjectAlias, error) {
-	out := []ProjectAlias{}
+func FindAliasesForProject(projectID string, version string) ([]ProjectAlias, error) {
```
ablack12, line 74: Right now we use this for repo projects as well, because they store aliases the same way; however, repos don't have parser projects (because they're just a logical grouping of settings not tied to one specific project) so it now may make sense to have a FindAliasesForRepo that only queries the alias collection.

model/project_aliases.go
```diff
@@ -96,13 +104,48 @@ func findAliasInProject(projectID, alias string) ([]ProjectAlias, error) {
 	if err != nil {
 		return []ProjectAlias{}, errors.Wrap(err, "error finding project aliases")
 	}
+	parserProject, err := ParserProjectByVersion(projectID, version)
```
ablack12, line 107: Does it make sense to flip these? If it exists for parser project then we never need to query the alias collection.

model/project_aliases.go
```diff
@@ -71,22 +71,30 @@ type ProjectAlias struct {
...
 	err := db.FindAllQ(ProjectAliasCollection, q, &out)
 	if err != nil {
-		return nil, err
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project aliases")
+	}
+	parserProject, err := ParserProjectByVersion(projectID, version)
+	if err != nil {
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project parser")
+	}
+	if parserProject != nil && parserProject.Aliases != nil {
```
ablack12, line 87: i recommend checking len(parserProject.Aliases)

model/project_aliases.go
```diff
@@ -71,22 +71,30 @@ type ProjectAlias struct {
...
+		return mergeAliasMaps(parserProjectAliases, dbAliases), nil
 	}
-
 	return out, nil
 }
 
 // findAliasInProject finds all aliases with a given name for a project.
 // Typically FindAliasInProjectOrRepo should be used.
-func findAliasInProject(projectID, alias string) ([]ProjectAlias, error) {
+func findAliasInProject(projectID, alias string, version string) ([]ProjectAlias, error) {
```
ablack12, line 97: similarly should maybe have findAliasesInRepo

model/project_aliases.go
```diff
@@ -96,13 +104,48 @@ func findAliasInProject(projectID, alias string) ([]ProjectAlias, error) {
 	if err != nil {
 		return []ProjectAlias{}, errors.Wrap(err, "error finding project aliases")
 	}
+	parserProject, err := ParserProjectByVersion(projectID, version)
+	if err != nil {
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project parser")
```
ablack12, line 109: should return nil for project aliases

model/project_aliases_test.go
```diff
@@ -2,6 +2,8 @@ package model
 
 import (
 	"fmt"
+	"github.com/evergreen-ci/evergreen"
```
ablack12, line 5: should move these imports down to the github.com block

model/project_parser_db.go
```diff
@@ -65,6 +66,36 @@ func ParserProjectFindOne(query db.Q) (*ParserProject, error) {
...
 
+// ParserProjectByVersion finds a parser project with a given version.
+// If version is empty the last known good config will be returned
+func ParserProjectByVersion(projectId string, version string) (*ParserProject, error) {
+	lookupVersion := false
+	if version == "" {
+		lastGoodVersion, err := FindVersionByLastKnownGoodConfig(projectId, -1)
+		if err != nil || lastGoodVersion == nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message":    fmt.Sprintf("Unable to retrieve last good version for project '%s'", projectId),
```
ablack12, line 77: Sorry for not having this thought the first time around: wrapping the message here is a bit unnecessary since projectId is listed in the fields; I would just put the "unable to retrieve last good version for project" bit (but it would be good to wrap the actual error returned at 81)

ablack12: same below

ablack12 (changes requested):

model/project_aliases.go
```diff
@@ -71,38 +71,110 @@ type ProjectAlias struct {
 type ProjectAliases []ProjectAlias
 
 // FindAliasesForProject fetches all aliases for a given project
```
ablack12, line 73: Should update this comment for the version string

model/project_aliases.go
```diff
@@ -71,38 +71,110 @@ type ProjectAlias struct {
...
+	err := db.FindAllQ(ProjectAliasCollection, q, &out)
 	if err != nil {
 		return nil, err
 	}
 
 	return out, nil
 }
 
+// findAliasInRepo finds all aliases with a given name for a repo.
+// Typically FindAliasInRepo should be used.
```
ablack12, line 110: FindAliasInProjectOrRepo I think

model/project_aliases.go
```diff
@@ -71,38 +71,110 @@ type ProjectAlias struct {
...
+func findAliasInRepo(projectID, alias string) ([]ProjectAlias, error) {
```
ablack12, line 111: can we change this to repoId for consistency

model/project_aliases.go
```diff
@@ -71,38 +71,110 @@ type ProjectAlias struct {
...
+	var out []ProjectAlias
+	q := db.Query(bson.M{
+		projectIDKey: projectID,
+		aliasKey:     alias,
+	})
+	err := db.FindAllQ(ProjectAliasCollection, q, &out)
+	if err != nil {
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project aliases")
```
ablack12, line 119: can we return nil here?

model/project_aliases.go
ablack12: Want to make sure you saw this ^ feel free to DM for more clarification!

model/project_aliases.go
```diff
@@ -71,38 +71,110 @@ type ProjectAlias struct {
...
+		return nil, errors.Wrap(err, "error finding project aliases")
+	}
+	parserProject, err := ParserProjectByVersion(projectID, version)
+	if err != nil {
+		return []ProjectAlias{}, errors.Wrap(err, "error finding project parser")
+	}
+	if parserProject != nil && len(parserProject.Aliases) > 0 {
+		parserProjectAliases := aliasesToMap(parserProject.Aliases)
+		dbAliases := aliasesToMap(out)
+		return mergeAliasMaps(parserProjectAliases, dbAliases), nil
```
ablack12, line 90: I like this implementation! Can mergeAliasMaps instead take in just the lists and be "mergeAliases" and create the maps inside, to move more of the implementation out?

model/project_parser.go
```diff
@@ -85,6 +85,7 @@ type ParserProject struct {
 	TaskAnnotationSettings *evergreen.AnnotationsSettings `yaml:"task_annotation_settings,omitempty" bson:"task_annotation_settings,omitempty"`
 	BuildBaronSettings     *evergreen.BuildBaronSettings  `yaml:"build_baron_settings,omitempty" bson:"build_baron_settings,omitempty"`
 	PerfEnabled            *bool                          `yaml:"perf_enabled,omitempty" bson:"perf_enabled,omitempty"`
+	Aliases                []ProjectAlias                 `yaml:"aliases,omitempty" bson:"aliases,omitempty"`
```
ablack12, line 88: We discussed in slack potentially breaking this up into different kinds of aliases and merging this ourselves (GithubAliases, GitTagAliases, etc) -- did we decide not to do this? In which case we'll just need to be very clear in our documentation what aliases should be named to apply to these different things.

model/project_aliases.go
ablack12: Actually I haven't found a place where we're using the version string. Do we need to input this or should we always use the empty version?

model/project_aliases.go
hadjri: Yep you're right.  I was thinking about FindAliasesForProject but here we are just looking for a single key.

rest/data/aliases.go
```diff
@@ -30,14 +30,14 @@ func (d *DBAliasConnector) FindProjectAliases(projectId, repoId string, aliasesT
 		}
 	}
 	if projectId != "" {
-		aliases, err = model.FindAliasesForProject(projectId)
```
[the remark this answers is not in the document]
hadjri: Took a look into its usages and it is only used in REST routes.

model/project_aliases.go
```diff
@@ -70,8 +70,8 @@ type ProjectAlias struct {
 
 type ProjectAliases []ProjectAlias
 
-// FindAliasesForProject fetches all aliases for a given project and optional version
-func FindAliasesForProject(projectID string, version string) ([]ProjectAlias, error) {
+// FindAliasesForProject fetches all aliases for a given project
+func FindAliasesForProject(projectID string) ([]ProjectAlias, error) {
 	var out []ProjectAlias
 	q := db.Query(bson.M{
```
ablack12, line 76: We can use FindAliasesForProjectFromDB here now to remove duplicate code

model/project_ref.go
```diff
@@ -1384,7 +1384,7 @@ func GetProjectSettings(p *ProjectRef) (*ProjectSettings, error) {
 	if projectVars == nil {
 		projectVars = &ProjectVars{}
 	}
-	projectAliases, err := FindAliasesForProject(p.Id, "")
+	projectAliases, err := FindAliasesForProject(p.Id)
```
ablack12: This should use FromDB; we use GetProjectSettings to get all settings for the project page.

rest/data/aliases.go
```diff
@@ -61,7 +61,7 @@ func (d *DBAliasConnector) FindProjectAliases(projectId, repoId string, aliasesT
 
 // CopyProjectAliases finds the aliases for a given project and inserts them for the new project.
 func (d *DBAliasConnector) CopyProjectAliases(oldProjectId, newProjectId string) error {
-	aliases, err := model.FindAliasesForProject(oldProjectId, "")
+	aliases, err := model.FindAliasesForProject(oldProjectId)
```
ablack12: This should use FromDB; used to copy a project page.

service/project.go
```diff
@@ -707,7 +707,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	origProjectAliases, _ := model.FindAliasesForProject(id, "")
+	origProjectAliases, _ := model.FindAliasesForProject(id)
```
ablack12: Anything that's a method of UIServer is for the UI and so should use FindAlilasesForProjectFromDB; this includes this call and the following

model/project_aliases.go
```diff
@@ -72,33 +72,145 @@ type ProjectAliases []ProjectAlias
 
 // FindAliasesForProject fetches all aliases for a given project
 func FindAliasesForProject(projectID string) ([]ProjectAlias, error) {
```
ablack12, line 74: Maybe this should say "FindAllAliasesForProject" or "FindAliasesForProjectWithYamlAliases" ? I feel like both of those names aren't great but to get across that this isn't what it used to be. (Alternatively maybe just a comment would do the trick)

ablack12 (commented):
This is another instance used by the New UI that should use the non-merged function.
https://github.com/evergreen-ci/evergreen/blob/b9a385e2d40b17d85dbdf168d99bac11ebe1d2a5/graphql/resolvers.go#L431

model/project_aliases.go
hadjri: I think both would be good; changed name to FindAllAliasesForProject and added comment.

model/project_parser_db.go
```diff
@@ -65,6 +67,36 @@ func ParserProjectFindOne(query db.Q) (*ParserProject, error) {
...
+	lookupVersion := false
+	if version == "" {
+		lastGoodVersion, err := FindVersionByLastKnownGoodConfig(projectId, -1)
+		if err != nil || lastGoodVersion == nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message":    "Unable to retrieve last good version for project",
+				"project_id": projectId,
+				"version":    version,
+			}))
+			return nil, errors.Wrap(err, fmt.Sprintf("Unable to retrieve last good version for project '%s'", projectId))
```
ablack12, line 82: errors.Wrapf instead of fmt.

model/project_parser_db.go
```diff
@@ -65,6 +67,36 @@ func ParserProjectFindOne(query db.Q) (*ParserProject, error) {
...
+	}
+	parserProject, err := ParserProjectFindOneById(version)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
+			"message":        "Error retrieving parser project for version",
+			"project_id":     projectId,
+			"version":        version,
+			"lookup_version": lookupVersion,
+		}))
+		return nil, errors.Wrap(err, fmt.Sprintf("Error retrieving parser project for version '%s'", version))
```
ablack12, line 95: same here

PR #5098, opened 2021-10-12
EVG-15293 Add a requireEditProjectSettings middleware to gqlgen

malikchaya2:
[EVG-15293](https://jira.mongodb.org/browse/EVG-15293)

### Description 
This adds a @requireEditProjectSettings directive that checks if a user is authorized to edit a project. After discussing middlewares with Annie I realized that it would be nice to have a more generic middleware that takes in the Resource type, Setting type and Value so we basically only ever need one middleware. However, since all the work was already done, and this is all we need for now, this is good for now and we can make it more generic when we add a requireViewProjectSettings or next time we add another one. 

### Testing 
Added a test.


malikchaya2:
NOTE: this is not working as expected. It turns out the value does not get passed in dynamically. I'll have to go back to the drawing board.

khelif96:
Will reopen this pr when i come back to it.

PR #5099, opened 2021-10-13
EVG-15362 Update ipv4_address to hold private addresses instead

bynn:
[EVG-15362](https://jira.mongodb.org/browse/EVG-15362)

### Description 
10.* addresses are private addresses so I updated them to save the private one
adding a ipv4_address_1 index should prevent hanging according to changes on staging

### Testing 
called command on staging

### DIFF
--- a/cloud/ec2_util.go
+++ b/cloud/ec2_util.go
@@ -320,7 +320,7 @@ func cacheHostData(ctx context.Context, h *host.Host, instance *ec2.Instance, cl
 		return errors.New("instance missing public dns name")
 	}
 	if instance.PrivateIpAddress == nil {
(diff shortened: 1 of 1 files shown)


cloud/ec2_util.go
```diff
@@ -319,14 +319,14 @@ func cacheHostData(ctx context.Context, h *host.Host, instance *ec2.Instance, cl
 	if instance.PublicDnsName == nil {
 		return errors.New("instance missing public dns name")
 	}
-	if instance.PublicIpAddress == nil {
+	if instance.PrivateIpAddress == nil {
 		return errors.New("instance missing public ip address")
```
ybrill, line 323: don't forget to change the error string!

PR #5100, opened 2021-10-13
EVG-15040 add CopyProject mutation

ablack12:
[EVG-15040](https://jira.mongodb.org/browse/EVG-15040)

### Description 
This is something of a work in progress -- I wanted to refactor the CopyProject connector function to do the project id / identifier validation, to avoid duplicate code. This also wasn't caught for CreateProject but we added the ability to specify project ID if the user wants (shown below). If this way of handling it seems good I can also add this to CreateProject.

image

### Testing 
Not tested yet

### DIFF
--- a/graphql/generated.go
+++ b/graphql/generated.go
@@ -6730,6 +6730,7 @@ input CreateProjectInput {
   identifier: String!
   owner: String!
   repo: String!
+  id: String
 }
 
 input CopyProjectInput {
@@ -32976,6 +32977,12 @@ func (ec *executionContext) unmarshalInputCreateProjectInput(ctx context.Context
 		case "repo":
 			var err error
 			it.Repo, err = ec.unmarshalNString2ᚖstring(ctx, v)
+			if err != nil {
+				return it, err
+			}
+		case "id":
+			var err error
+			it.Id, err = ec.unmarshalOString2ᚖstring(ctx, v)
 			if err != nil {
 				return it, err
 			}
--- a/graphql/tests/project/queries/create-project.graphql
+++ b/graphql/tests/project/queries/create-project.graphql
@@ -1,11 +1,13 @@
 mutation {
   createProject(
     project: {
-      identifier: "catVideos", 
+      identifier: "catVideos",
+      id: "grumpyCat",
       owner: "theCat", 
       repo: "catGenerator"
       }
   ){
+    id
     identifier
     repo
   }
--- a/graphql/tests/project/results.json
+++ b/graphql/tests/project/results.json
@@ -31,6 +31,7 @@
       "result": {
         "data": {
           "createProject": {
+            "id": "grumpyCat",
             "identifier": "catVideos",
             "repo": "catGenerator"
           }


graphql/schema.graphql
```diff
@@ -316,6 +317,12 @@ input ProjectInput {
   repo: String!
 }
 
+input CopyProjectInput {
+  projectIdToCopy: String
+  newProjectIdentifier: String!
```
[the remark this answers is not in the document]
ablack12: Right, we do need to support this which is what I meant by the original description; it was introduced after the designs were created which is why you don't see it there. I don't think we necessarily need to ask Lara to update the design since it's a small thing but we should maybe drop a comment so we remember it should be there?

graphql/schema.graphql
```diff
@@ -316,6 +317,12 @@ input ProjectInput {
   repo: String!
 }
 
+input CopyProjectInput {
+  projectIdToCopy: String
```
malikchaya2, line 321: Since we need this to be able to copy the project and it missing produces and error, this should probably be required.

rest/data/project_settings.go
```diff
@@ -15,18 +17,48 @@ import (
...
+	}
+
+	// verify project with new name doesn't exist
+	if err = sc.verifyUniqueProject(opts.NewProjectIdentifier); err != nil {
+		return nil, err
+	}
+
 	oldId := projectToCopy.Id
+	if opts.NewProjectId != "" {
+		if err = sc.verifyUniqueProject(opts.NewProjectId); err != nil {
```
malikchaya2, line 45: I agree that we should add this to CreateProject as well. We can also do it in a separate ticket if you prefer that.

rest/data/project_settings.go
```diff
@@ -45,11 +77,29 @@ func (sc *DBConnector) CopyProject(ctx context.Context, projectToCopy *model.Pro
...
 	if err := sc.UpdateAdminRoles(projectToCopy, projectToCopy.Admins, nil); err != nil {
-		return nil, errors.Wrapf(err, "Database error updating admins for project '%s'", newProject)
+		return nil, errors.Wrapf(err, "Database error updating admins for project '%s'", opts.NewProjectIdentifier)
 	}
 	return apiProjectRef, nil
 }
 
+func (sc *DBConnector) verifyUniqueProject(name string) error {
+	_, err := sc.FindProjectById(name, false)
+	if err == nil {
```
malikchaya2, line 87: we should probably double check that the error is a StatusNotFound error, because the error can also come from FindMergedProjectRef.

graphql/resolvers.go
```diff
@@ -1092,6 +1092,23 @@ func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.
...
+	if err != nil {
+		apiErr, ok := err.(gimlet.ErrorResponse)
+		if ok {
+			if apiErr.StatusCode == http.StatusBadRequest {
+				return nil, InputValidationError.Send(ctx, apiErr.Message)
+			}
+			// include StatusNotFound as an internal error because we determine this input
+			return nil, InternalServerError.Send(ctx, apiErr.Message)
+		}
+		return nil, InternalServerError.Send(ctx, err.Error())
```
khelif96, line 1106: We have a util function that does this mapping. 
https://github.com/evergreen-ci/evergreen/blob/33049b7427a97f529980546ceb391d85a67f14ed/graphql/util.go

rest/data/project.go
```diff
@@ -68,6 +78,24 @@ func (pc *DBProjectConnector) UpdateProject(projectRef *model.ProjectRef) error
 	return nil
 }
 
+func (sc *DBProjectConnector) VerifyUniqueProject(name string) error {
+	_, err := sc.FindProjectById(name, false)
+	if err == nil {
```
malikchaya2, line 83: It seems to me like it would be possible to get an error that's not a statusNotfound error. In that case we would assume that we are good to go ahead and create the project (since err != nil), but that error may be some other error, not an error due to the project not being found.

rest/data/project_settings.go
malikchaya2: even with FindBranchProjectRef, we may get an error from here: https://github.com/evergreen-ci/evergreen/blob/main/db/query.go#L78. 
err won't be nil so we won't return.

graphql/schema.graphql
```diff
@@ -354,6 +355,19 @@ input PatchesInput {
   includeCommitQueue: Boolean! = false
 }
 
+input CreateProjectInput {
+  identifier: String!
+  owner: String!
+  repo: String!
+  id: String
```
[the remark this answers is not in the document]
ablack12: sure thing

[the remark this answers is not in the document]
ablack12: this was a good call, I hadn't updated the generated code

PR #5103, opened 2021-10-13
EVG-15320: Link to new UI commit queue page in CLI confirmation of commit queue patch submission

hadjri:
[EVG-15320](https://jira.mongodb.org/browse/EVG-15320)

### Description 
Changed enqueue patch command in the commit queue CLI to link to the new UI commit queue page for the project in the CLI.

### DIFF
--- a/operations/commit_queue.go
+++ b/operations/commit_queue.go
@@ -291,7 +291,7 @@ func enqueuePatch() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem creating a commit queue patch")
 			}
-			patchDisp, err := getAPIPatchDisplay(mergePatch, false, conf.UIServerHost)
+			patchDisp, err := getAPIPatchDisplay(mergePatch, false, "https://spruce.mongodb.com")
 			if err != nil {
 				grip.Errorf("can't print patch display for new patch '%s'", mergePatch.Id)
 			}
@@ -473,7 +473,7 @@ func listCLICommitQueueItem(item restModel.APICommitQueueItem, ac *legacyClient,
 	if p.Author != "" {
 		grip.Infof("Author: %s", p.Author)
 	}
-	disp, err := getPatchDisplay(p, false, uiServerHost)
+	disp, err := getPatchDisplay(p, false, uiServerHost, "")
 	if err != nil {
 		grip.Error(message.WrapError(err, "\terror getting patch display"))
 		return
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -153,7 +153,7 @@ func (p *patchParams) validateSubmission(diffData *localDiff) error {
 }
 
 func (p *patchParams) displayPatch(conf *ClientSettings, newPatch *patch.Patch) error {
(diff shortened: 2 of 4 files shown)


hadjri:
Screen Shot 2021-10-18 at 1 30 41 PM

operations/commit_queue.go
```diff
@@ -176,7 +176,12 @@ func mergeCommand() cli.Command {
 			client := conf.setupRestCommunicator(ctx)
 			defer client.Close()
 
-			return params.mergeBranch(ctx, conf, client, ac)
+			settings, err := client.GetSettings(ctx)
+			if err != nil {
+				return errors.Wrap(err, "problem retrieving admin settings")
+			}
+
+			return params.mergeBranch(ctx, conf, client, ac, settings.Ui.UIv2Url)
```
ablack12, line 184: I'm not sure there's a reason to get UIv2Url before entering mergeBranch since we're passing in client and ctx; can we just retrieve it at the time it's necessary?

operations/patch_util.go
```diff
@@ -176,6 +176,30 @@ func (p *patchParams) displayPatch(conf *ClientSettings, newPatch *patch.Patch)
...
+func (p *patchParams) displayCommitQueuePatch(conf *ClientSettings, newPatch *patch.Patch, uiV2Url string) error {
+	patchDisp, err := getCommitQueuePatchDisplay(newPatch, p.ShowSummary, uiV2Url)
+	if err != nil {
+		return err
+	}
+
+	grip.Info("Patch successfully created.")
+	grip.Info(patchDisp)
+
+	if p.Browse {
```
ablack12, line 188: Could move this to a helper function so we can call it here and displayPatch (rather than having the same code in both places)?

operations/patch_util.go
```diff
@@ -461,7 +508,7 @@ func getAPIPatchDisplay(apiPatch *restmodel.APIPatch, summarize bool, uiHost str
 		return "", errors.New("service patch is not a Patch")
 	}
 
-	return getPatchDisplay(&servicePatch, summarize, uiHost)
+	return getCommitQueuePatchDisplay(&servicePatch, summarize, uiHost)
```
ablack12, line 511: is getAPIPatchDisplay only used for commit queue, in which case should it also be called getAPICommitQueuePatchDisplay?

operations/patch_util.go
```diff
@@ -451,6 +475,29 @@ func getPatchDisplay(p *patch.Patch, summarize bool, uiHost string) (string, err
 	return out.String(), nil
 }
 
+// getCommitQueuePatchDisplay returns a human-readable summary representation of a patch object
+// which can be written to the terminal, and links to the new UI commit queue page for
+// the project in question in the CLI
+func getCommitQueuePatchDisplay(p *patch.Patch, summarize bool, uiHost string) (string, error) {
```
ablack12, line 481: Rather than having another function for this, getPatchDisplay can just use `p.GetCommitQueueURL` if `IsCommitQueuePatch` is true.

operations/patch_util.go
hadjri: Yes, we can rename to that clearer function name

PR #5105, opened 2021-10-14
EVG-15267 Decompose git.get_project into smaller functions

bsamek:
[EVG-15267](https://jira.mongodb.org/browse/EVG-15267)

### Description 
This is to prepare for working on this ticket, since the function was very
long and hard to read.

### Testing 
This mostly decomposes functions so should be covered by existing tests.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -395,9 +395,7 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
 		return errors.Wrap(err, "error expanding github parameters")
 	}
 
-	var projectMethod string
-	var projectToken string
-	projectMethod, projectToken, err = getProjectMethodAndToken(c.Token, conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion), conf.Distro.CloneMethod)
+	projectMethod, projectToken, err := getProjectMethodAndToken(c.Token, conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion), conf.Distro.CloneMethod)
 	if err != nil {
 		return errors.Wrap(err, "failed to get method of cloning and token")
 	}
@@ -438,16 +436,15 @@ func (c *gitFetchProject) fetchSource(ctx context.Context,
 	logger client.LoggerProducer,
 	conf *internal.TaskConfig,
 	jpm jasper.Manager,
-	opts cloneOpts) (string, error) {
-
-	stdErr := noopWriteCloser{&bytes.Buffer{}}
+	opts cloneOpts) error {
 
 	gitCommands, err := c.buildCloneCommand(ctx, conf, logger, opts)
 	if err != nil {
-		return "", err
+		return err
 	}
 	fetchScript := strings.Join(gitCommands, "\n")
 
+	stdErr := noopWriteCloser{&bytes.Buffer{}}
 	fetchSourceCmd := jpm.CreateCommand(ctx).Add([]string{"bash", "-c", fetchScript}).Directory(conf.WorkDir).
 		SetOutputSender(level.Info, logger.Task().GetSender()).SetErrorWriter(stdErr)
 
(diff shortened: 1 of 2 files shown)


agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
 	var err error
-	// expand the github parameters before running the task
-	if err = util.ExpandValues(c, conf.Expansions); err != nil {
-		return errors.Wrap(err, "error expanding github parameters")
+	logger.Execution().Infof("Fetching module: %s", moduleName)
+
+	var module *model.Module
+	module, err = conf.Project.GetModuleByName(moduleName)
+	if err != nil {
+		return errors.Errorf("Couldn't get module %s: %v", moduleName, err)
```
ablack12, line 495: Can we wrap this error?

agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
+		if module.Ref != "" {
+			revision = module.Ref
+			c.logModuleRevision(logger, revision, moduleName, "ref field in config file")
+		} else {
+			revision = module.Branch
+			c.logModuleRevision(logger, revision, moduleName, "branch field in config file")
+		}
+	}
+	var owner, repo string
+	owner, repo, err = thirdparty.ParseGitUrl(module.Repo)
```
ablack12, line 544: Right now if owner/repo is empty we continue. Do we still want to validate this?

agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
-		dir:                c.Directory,
-		token:              projectToken,
-		shallowClone:       c.ShallowClone && !conf.Distro.DisableShallowClone,
-		recurseSubmodules:  c.RecurseSubmodules,
-		mergeTestRequester: conf.Task.Requester == evergreen.MergeTestRequester,
+
+	var modulePatch *patch.ModulePatch
+	if p != nil {
+		// find module among the patch's Patches
+		for i := range p.Patches {
```
ablack12, line 552: We already do this at 508 with p.FindModule so i think we can just store that variable outside of the if and skip this altogether

agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
-		errOutput := stdErr.String()
-		if errOutput != "" {
-			if opts.token != "" {
-				errOutput = strings.Replace(errOutput, opts.token, "[redacted oauth token]", -1)
-			}
-			logger.Execution().Info(errOutput)
-		}
-		if err != nil {
-			return errors.Wrap(err, "problem with git command")
+			logger.Execution().Error(err)
```
ablack12, line 675: should we be continuing here?

agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
 		}
 	}
 
+	// Apply additional patches for commit queue batch execution.
```
ablack12, line 679: Is there a reason we handle modules in between fetching / applying the additional patches? (If we aren't sure I respect not re-factoring this until later if it might be breaking but it does seem like it makes more sense to have these happen together)

agent/command/git.go
```diff
@@ -393,240 +434,260 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
 	if conf.Task.Requester == evergreen.MergeTestRequester && !conf.Task.CommitQueueMerge {
 		for _, patchId := range additionalPatches {
-			logger.Task().Infof("applying changes from previous commit queue patch '%s'", patchId)
-			newPatch, err := comm.GetTaskPatch(ctx, td, patchId)
+			err := c.applyAdditionalPatches(ctx, conf, comm, logger, td, patchId)
```
ablack12, line 682: nit: applyAdditionalPatch seems more accurate unless we want to pass in all of them at once

agent/command/git.go
bsamek: I added this to `ParseGitUrl`

agent/command/git.go
bsamek: I wondered about this too and was avoiding changing it purely because this is a refactor.

PR #5106, opened 2021-10-14
EVG-15603 avoid marking display task as aborted when aborting for failed task

ablack12:
[EVG-15603 ](https://jira.mongodb.org/browse/EVG-15603 )

### Description 
When aborting a version due to a failed task, we take precautions not to mark the failure itself as aborted; we should also do this for the display task itself, if it's an execution task.

### Testing 
Added a unit test.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1949,6 +1949,8 @@ func AbortVersion(versionId string, reason AbortInfo) error {
 	}
 	if reason.TaskID != "" {
 		q[IdKey] = bson.M{"$ne": reason.TaskID}
+		// if the aborting task is part of a display task, we also don't want to mark it as aborted
+		q[ExecutionTasksKey] = bson.M{"$ne": reason.TaskID}
 	}
 	ids, err := findAllTaskIDs(db.Query(q))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


model/task/task.go
```diff
@@ -1949,6 +1949,8 @@ func AbortVersion(versionId string, reason AbortInfo) error {
 	}
 	if reason.TaskID != "" {
 		q[IdKey] = bson.M{"$ne": reason.TaskID}
+		// if the aborting task is part of a display task, we also don't want to mark it as aborted
+		q[ExecutionTasksKey] = bson.M{"$ne": reason.TaskID}
```
malikchaya2, line 1953: Apologies if this is a stupid question, but why is this on ExecutionTasks and not on DisplayTask? I thought we want to avoid marking the display task as aborted if one of its execution tasks has a legitimate failure.

ablack12: The example in the test might also help! The display task has three execution tasks; the second one et2 fails, but we don't mark the display task as aborted because it has ExecutionTasks = ["et1", "et2", "et3"]

malikchaya2 (approved):

PR #5107, opened 2021-10-15
EVG-15609: fix AWSVPC admin settings

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15609

### Description 
* Fix an issue where existing AWSVPC subnets/security groups couldn't be modified.
* Fix an issue where all admin settings do not zero out the cached value for a zero field when the field's BSON tag is omitempty.

### Testing
Tested locally and ran admin settings tests.


Kimchelly:
I don't really know why the array of objects is different (probably something to do with copying object references vs. copying immutable string types), but as long as it works, I don't think it matters since Angular is going out the door.

PR #5115, opened 2021-10-20
EVG-15440 Log Graphql panics and errors to Splunk

khelif96:
[EVG-15440](https://jira.mongodb.org/browse/EVG-15440)

### Description 
We were never logging any errors or panics to splunk for our graphql routes. Like we do for our rest routes. Which has made it difficult to debug issues.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2649,7 +2649,6 @@ func (r *mutationResolver) SaveSubscription(ctx context.Context, subscription re
 		if t == nil {
 			return false, ResourceNotFound.Send(ctx, fmt.Sprintf("cannot find task with id %s", id))
 		}
-		break
 	case "build":
 		b, buildErr := r.sc.FindBuildById(id)
 		if buildErr != nil {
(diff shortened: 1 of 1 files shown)


bynn (approved):

graphql/errors.go
```diff
@@ -33,13 +35,20 @@ func (err GqlError) Send(ctx context.Context, message string) *gqlerror.Error {
...
 	default:
-		return gqlerror.ErrorPathf(graphql.GetResolverContext(ctx).Path(), message)
+		return gqlerror.ErrorPathf(graphql.GetFieldContext(ctx).Path(), message)
 	}
 }
 
-func formError(ctx context.Context, message string, code GqlError) *gqlerror.Error {
+func formError(ctx context.Context, msg string, code GqlError) *gqlerror.Error {
+	queryPath := graphql.GetFieldContext(ctx).Path()
+	grip.Warning(message.Fields{
```
ablack12, line 44: This is a lower level than error, if that's what you're going for.

graphql/http_handler.go
```diff
@@ -30,5 +31,18 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 		})
 		return errors.New("internal server error")
 	})
+
+	srv.SetErrorPresenter(func(ctx context.Context, err error) *gqlerror.Error {
+		fieldCtx := graphql.GetFieldContext(ctx)
+		queryPath := fieldCtx.Path()
+		args := fieldCtx.Args
+		grip.Error(message.Fields{
```
ablack12, line 39: In this case we can use message.WrapError() to wrap the error rather than setting it as "message" which would give us stack trace / other information if available.

PR #5116, opened 2021-10-20
EVG-15219 Add sections on project page to configure plugins

hadjri:
[EVG-15219](https://jira.mongodb.org/browse/EVG-15219)

### Description 

### Testing 

### DIFF
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -999,6 +999,85 @@ Evergreen Projects
             </div>
           </div>
         </div>
+        <div id="plugin-info">
+          <div class="h3">Performance Plugin</div>
+          <div class="form-group">
+            <div class="col-lg-4 col-header">
+              <label class="control-label">Enabled&nbsp;&nbsp;
+                <input type="checkbox" name="perf_enabled" ng-model="settingsFormData.perf_enabled"/>
+              </label>
+              <div class="muted small">When checked, performance plugin will be enabled.</div>
+            </div>
+          </div>
+          <div class="h3">Build Baron Plugin</div>
+          <div class="form-group">
+            <div class="col-lg-4"> <label class="control-label"> Ticket create project </label> </div>
+          </div>
+          <div class="form-group">
+            <div class="col-lg-4">
+              <input class="form-control" type="text" ng-model="settingsFormData.build_baron_settings.ticket_create_project" placeholder="Ticket create project">
+            </div>
+          </div>
+          <div class="form-group">
+            <div class="col-lg-4"> <label class="control-label"> Ticket search project </label> </div>
+          </div>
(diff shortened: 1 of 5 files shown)


malikchaya2 (commented):
This looks great! 

Right now, someone can add both a webhook and ticket search project, and press save. We don't want to let them do that. If they fill out one, we should disable the other. Or we should at least show an error and not let them press save.

service/templates/projects.html
```diff
@@ -999,6 +999,85 @@ <h3>Virtual Workstations Commands</h3>
...
+          <div class="form-group">
+            <div class="col-lg-4"> <label class="control-label"> Ticket create project </label> </div>
+          </div>
+          <div class="form-group">
+            <div class="col-lg-4">
+              <input class="form-control" type="text" ng-model="settingsFormData.build_baron_settings.ticket_create_project" placeholder="Ticket create project">
+            </div>
+          </div>
+          <div class="form-group">
+            <div class="col-lg-4"> <label class="control-label"> Ticket search project </label> </div>
```
ablack12, line 1022: Now that we're showing this somewhere where non-admins can see it and won't understand it, I think maybe it's time to change this to something more helpful 😅  maybe "Project to Create Tickets In" / "Projects to Search"? Should also use proper capitalization for headings

service/templates/projects.html
```diff
@@ -999,6 +999,85 @@ <h3>Virtual Workstations Commands</h3>
...
+              <input class="form-control" ng-model="jira_display_text" type="text" placeholder="display text">
+            </div>
+            <div class="col-lg-2">
+              <button class="plus-button btn btn-primary" ng-disabled="!(jira_field) || !(jira_display_text)" type="button" ng-click="addJiraField()">
+                <i class="fa fa-plus"></i>
+              </button>
+            </div>
+          </div>
+          <div class="form-group">
+            <div class="col-lg-4"> <label class="control-label"> Webhook </label> </div>
```
ablack12, line 1070: I think we should put a small description under this but I'm not actually sure what it would be; @malikchaya2 any suggestions?

malikchaya2: That's a good idea. 
Maybe something like: 

To use a custom setup for creating build failure tickets, specify a webhook endpoint and secret.

hadjri:
> This looks great!
> 
> Right now, someone can add both a webhook and ticket create project, and press save. We don't want to let them do that. If they fill out one, we should disable the other. Or we should at least show an error and not let them press save. Or at the very least show an error on save and not save it on the backend.
> 
> Also: should we let people specify a ticket search project if they are not also specifying a ticket create project?

That makes sense, made the change to disable webhook inputs if build baron configs are set, and vice versa.  Also yes, we should not let people submit ticket search project without a ticket create project.  Save changes button is now disabled unless both or neither are defined.

PR #5117, opened 2021-10-21
EVG-15531 Group the list of Projects a user can view by their repo id

malikchaya2:
[EVG-15531](https://jira.mongodb.org/browse/EVG-15531)

### Description 
This groups an array of projects by the Repo they are part of. This is needed by the UI so that it can show projects in a grouped fashion on the menu.

### Testing 
Added a test.

### DIFF
--- a/model/project_ref_test.go
+++ b/model/project_ref_test.go
@@ -779,6 +779,13 @@ func TestGroupProjectsByRepo(t *testing.T) {
 
 	assert.Equal(2, len(groupedProjects["gimlet"]))
 	assert.Equal(3, len(groupedProjects["mongo"]))
+
+	assert.Equal("projectB", groupedProjects["mongo"][0].Id)
+	assert.Equal("projectC", groupedProjects["mongo"][1].Id)
+	assert.Equal("projectD", groupedProjects["mongo"][2].Id)
+
+	assert.Equal("projectE", groupedProjects["gimlet"][0].Id)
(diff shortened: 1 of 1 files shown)


rest/data/project.go
```diff
@@ -96,6 +96,16 @@ func (sc *DBProjectConnector) VerifyUniqueProject(name string) error {
 	return nil
 }
 
+// GroupProjectsByRepo takes in an array of projects and groups them in a map based on the repo they are part of
+func (sc *DBProjectConnector) GroupProjectsByRepo(projects []model.ProjectRef) map[string][]model.ProjectRef {
```
ablack12, line 100: Because this function doesn't actually use the database, it shouldn't be a connector function at all. I would either do this work directly in the resolver, do the project fetching here, or add it as a regular model/project_ref.go function.

rest/data/project.go
```diff
@@ -96,6 +96,16 @@ func (sc *DBProjectConnector) VerifyUniqueProject(name string) error {
...
+	groupedProject := make(map[string][]model.ProjectRef)
+	for _, project := range projects {
+		repoProjects := groupedProject[project.Repo]
```
ablack12, line 103: Theoretically there can be multiple repos with the same name; it's the owner/repo pair that's guaranteed to be unique, or the RepoRefId (This may be better because we don't want to group projects that aren't defaulting to repo; it implies that they have a connection to each other where they do not).

PR #5141, opened 2021-10-27
EVG-14740 Update patch to work locally

bynn:
[EVG-14740](https://jira.mongodb.org/browse/EVG-14740)

### Description 
only grip a warning when we can't get commit event from github
not sure if this is what we would want though

### Testing 
image
image


model/patch_lifecycle.go
```diff
@@ -291,7 +300,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 
 	_, err = thirdparty.GetCommitEvent(githubCtx, githubOauthToken, projectRef.Owner, projectRef.Repo, p.Githash)
 	if err != nil {
-		return nil, errors.Wrap(err, "Couldn't fetch commit information")
+		grip.Warning(errors.Wrap(err, "Couldn't fetch commit information").Error())
```
ablack12, line 303: Can we log this warning with grip.WithError and grip.Message to include more information (owner/repo/githash/patch ID)? If there is a real issue this would help us with debugging

model/patch_lifecycle.go
```diff
@@ -149,9 +149,18 @@ func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken str
 	}
 	opts.RemotePath = path
 	opts.PatchOpts.env = env
-	projectFileBytes, err = getFileForPatchDiff(ctx, *opts)
-	if err != nil {
-		return nil, "", errors.Wrapf(err, "could not fetch remote configuration file")
+
+	if githubOauthToken == evergreen.Localhost {
+		opts.ReadFileFrom = ReadFromLocal
+		projectFileBytes, err = ioutil.ReadFile(path)
```
ablack12, line 155: If we require the config file to be changed to be able to patch locally, do we hit https://github.com/evergreen-ci/evergreen/blob/17b6f385ea0f76f5173772c66c303b14ebc06ea4/model/project_parser.go#L745 instead? That would be preferable to modifying this code just to handle the local case.

config.go
```diff
@@ -576,10 +576,17 @@ func (s *Settings) GetGithubOauthStrings() ([]string, error) {
...
+const (
+	Localhost = "localhost"
+)
+
 func (s *Settings) GetGithubOauthToken() (string, error) {
 	if s == nil {
 		return "", errors.New("not defined")
 	}
+	if s.DomainName == Localhost {
+		return Localhost, nil
```
ablack12, line 588: I'd prefer not to add code to prod that specifically cases on running locally. Some alternatives:
1) store a fake token in https://github.com/evergreen-ci/evergreen/blob/990f2a8d3f8a44d939aadea5ad9aae8d749e609d/testdata/local/admin.json#L5 so that we do have a token to return (even though it won't work), or 
2) Don't error if token doesn't exist when it's called for GetPatchedConfig, just log, and then only call GetCommitEvent if the token isn't empty.

ablack12 (changes requested):

model/patch_lifecycle.go
bynn: i looked into this for a while but configChanged doesnt return the correct result (i think) because module name always seems to be populated even at the creation of patchDoc. im not sure what would be a better alternative

model/patch_lifecycle.go
```diff
@@ -300,7 +300,14 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 
 	_, err = thirdparty.GetCommitEvent(githubCtx, githubOauthToken, projectRef.Owner, projectRef.Repo, p.Githash)
 	if err != nil {
-		grip.Warning(errors.Wrap(err, "Couldn't fetch commit information").Error())
+		grip.Warning(message.Fields{
+			"error":    err,
```
ablack12, line 304: should use message.WrapError rather than putting it as a field (formats it better)

model/patch_lifecycle.go
```diff
@@ -300,7 +300,14 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+			"message":  "Couldn't fetch commit information",
+			"patch id": p.Id,
```
ablack12, line 306: patch_id

model/patch_lifecycle.go
ablack12: Hmm that shouldn't be true -- one of them will always be the empty string because it's the current repo. Can we add some logging to ConfigChanged to walk through this together today?

units/patch_intent.go
```diff
@@ -580,8 +580,9 @@ func (j *patchIntentProcessor) buildCliPatchDoc(ctx context.Context, patchDoc *p
 	commit, err := thirdparty.GetCommitEvent(ctx, githubOauthToken, projectRef.Owner,
 		projectRef.Repo, patchDoc.Githash)
 	if err != nil {
-		return errors.Wrapf(err, "could not find base revision '%s' for project '%s'",
-			patchDoc.Githash, projectRef.Id)
+		grip.Warning(
```
ablack12, line 583: I'm concerned about not erroring at all; this really could be an error. I'd like to explore the "ConfigChanged" idea.

bynn:
closed because this ticket is project sized

PR #5145, opened 2021-10-29
EVG-15712 stop polling generate.tasks for doc too large

ablack12:
[EVG-15712](https://jira.mongodb.org/browse/EVG-15712)

### Description 
Re-opening to address Brian's comments.

### DIFF
--- a/agent/command/generate.go
+++ b/agent/command/generate.go
@@ -114,20 +114,23 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
 			if err != nil {
 				return true, err
 			}
-			if generateStatus.Finished {
-				return false, nil
+
+			var generateErr error
+			if len(generateStatus.Errors) > 0 {
+				generateErr = errors.New(strings.Join(generateStatus.Errors, ", "))
 			}
-			generateErr := errors.New(generateStatus.Error)
 			if !generateStatus.ShouldRetry {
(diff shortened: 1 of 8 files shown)


agent/command/generate.go
```diff
@@ -114,13 +114,20 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
 			if err != nil {
 				return true, err
 			}
+
+			var generateErr error
 			if len(generateStatus.Errors) > 0 {
-				fullErr := errors.New(strings.Join(generateStatus.Errors, ", "))
+				generateErr = errors.New(strings.Join(generateStatus.Errors, ", "))
+			}
+			if !generateStatus.ShouldRetry {
```
bsamek, line 122: I think this relies on a perfectly lockstep deploy. Consider:
1. We deploy. The app servers and agents both (eventually) get the new code.
2. We revert. The app servers revert, but not yet all agents.
In that state `ShouldRetry` will always be false. Agents that have this code will always stop at 123 and never get to 125.

bsamek (changes requested):

agent/command/generate.go
ablack12: I see what you're saying. I'll flip this boolean so that we retry with the default value.

bsamek (approved):

PR #5146, opened 2021-10-29
EVG-15661 improve fetching expansions for variants (with matrices)

ablack12:
[EVG-15661](https://jira.mongodb.org/browse/EVG-15661)

### Description 
Didn't consider that variants can be created from matrix definitions. Included this from translateProject.

### Testing 
I'm a bit put out that I had to slightly learn matrices to unit test this.

### DIFF
--- a/model/project_matrix.go
+++ b/model/project_matrix.go
@@ -247,6 +247,15 @@ func evaluateAxisTags(ase *axisSelectorEvaluator, axis string, selectors []strin
 	return out, errs
 }
 
+func GetVariantsWithMatrices(ase *axisSelectorEvaluator, axes []matrixAxis, bvs []parserBV) ([]parserBV, []error) {
+	if ase == nil {
+		ase = NewAxisSelectorEvaluator(axes)
+	}
+	regularBVs, matrices := sieveMatrixVariants(bvs)
+	matrixBVs, errs := buildMatrixVariants(axes, ase, matrices)
+	return append(regularBVs, matrixBVs...), errs
+}
+
 // buildMatrixVariants takes in a list of axis definitions, an axisSelectorEvaluator, and a slice of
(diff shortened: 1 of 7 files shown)


malikchaya2 (approved):
I don't see anything concerning here, but I am not familiar with matrices...

model/project_matrix.go
```diff
@@ -247,6 +247,15 @@ func evaluateAxisTags(ase *axisSelectorEvaluator, axis string, selectors []strin
 	return out, errs
 }
 
+func GetVariantsWithMatrices(ase *axisSelectorEvaluator, axes []matrixAxis, bvs []parserBV) ([]parserBV, []error) {
+	if ase == nil {
```
bsamek, line 251: I think this `if` statement is unnecessary. `GetVariantsWithMatrices` is only called twice. In `TranslateProject` we construct an `ase` immediately before from `pp.Axes` and then pass it to the function. In `FindExpansionsForVariant` we pass `nil` but then also pass `pp.Axes`. I think you can remove the first argument to `GetVariantsWithMatrices` since you're always passing `pp.Axes`. You can then always construct the `ase` in `GetVariantsWithMatrices`. And in `TranslateProject` remove constructing the `ase` before calling `GetVariantsWithMatrices`.

bsamek (changes requested):

model/project_matrix.go
ablack12: We do need ase in TranslateProject for the NewVariantSelectorEvaluator. I wanted to save some complexity in TranslateProject by passing it through since it's an expensive function; I can change to just always passing it in instead though if you prefer.

bsamek: Oh I missed that. That makes sense, thanks. LGTM.

bsamek (approved):

PR #5147, opened 2021-10-29
EVG-15473 Error on empty configuration file found before attempting to validate

hadjri:
[EVG-15473](https://jira.mongodb.org/browse/EVG-15473)

### Description 
When a patches' config file cannot be retrieved from GitHub in a PR patch, a validation error shows up in the UI.  In splunk this error comes up as "invalid patched config syntax: project 'jasper' must specify at least one buildvariant".  Added additional error handling in the case that we retrieve an empty file from GitHub and we do not apply local patch differences that we produce a more useful error rather than raising false alarms for the validity of the config file.

### Testing 
Tested validation on an empty config file and confirmed that we received the same error seen here: https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen%22%20e229a9b7523bee394f446f33b52c0e79ee68705b&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=0&latest=&sid=1635443577.3015334


model/task_lifecycle.go
```diff
@@ -1446,6 +1446,7 @@ func UpdateDisplayTaskForTask(t *task.Task) error {
 	}
 
 	update := bson.M{
+		task.DisplayStatusKey: statusTask.ResultStatus(),
```
ablack12, line 1449: Does this change belong here?

model/patch_lifecycle.go
```diff
@@ -162,6 +162,9 @@ func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken str
 			return nil, "", errors.Wrapf(err, "Could not patch remote configuration file")
 		}
 	}
+	if string(projectFileBytes) == "" {
```
ablack12, line 165: Can we move this check into MakePatchedConfig?

ablack12 (changes requested):

ablack12 (approved):

PR #5148, opened 2021-11-01
EVG-15394 allow periodic build definitions without alias

ablack12:
[EVG-15394](https://jira.mongodb.org/browse/EVG-15394)

### Description 
We don't actually require an alias here, especially since users can define a completely separate yaml. This is consistent with [documentation](https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#periodic-builds).

### Testing 
Verified I was able to save a periodic build in the DB without error and that it no longer shows up as required on the UI.
image


bynn (approved):

hadjri (approved):
LGTM

PR #5150, opened 2021-11-01
EVG-14655 return clearer errors for set-module

ablack12:
[EVG-14655](https://jira.mongodb.org/browse/EVG-14655)

### Description 
I think the difficulty with the error is that it doesn't say what project you're using or specify that the project itself may be the problem, so I added some detail.

### Testing 
tested using local.
```
$ ../evergreen/clients/darwin_amd64/evergreen -c ~/.evergreen-local.yml set-module -i 5f74d99ab2373627c047c5e5 -m dsi  -y
module 'dsi' unknown or not found
Project 'evergreen' has no configured modules. Specify different project or see the evergreen configuration file for module configuration.

$ clients/darwin_amd64/evergreen set-module -i 61689cf1c9ec443f692bd8b0 -m dsi -y
module 'dsi' unknown or not found
Could not find module named 'dsi' for project 'spruce'. Specify different project or select correct module from:
	evergreen
```

### DIFF
--- a/operations/patch_module.go
+++ b/operations/patch_module.go
@@ -84,13 +84,14 @@ func PatchSetModule() cli.Command {
 			moduleBranch, err := getModuleBranch(module, proj)
 			if err != nil {
 				grip.Error(err)
-				mods, merr := ac.GetPatchModules(patchID, existingPatch.Project)
+				mods, projectIdentifier, merr := ac.GetPatchModules(patchID, existingPatch.Project)
 				if merr != nil {
 					return errors.Wrap(merr, "errors fetching list of available modules")
 				}
 
 				if len(mods) != 0 {
-					grip.Noticef("known modules includes:\n\t%s", strings.Join(mods, "\n\t"))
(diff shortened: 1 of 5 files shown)


operations/patch_module.go
```diff
@@ -84,13 +84,18 @@ func PatchSetModule() cli.Command {
...
+						"see the evergreen configuration file for module configuration.",
+						projectIdentifier)
+				}
 				if len(mods) != 0 {
-					grip.Noticef("known modules includes:\n\t%s", strings.Join(mods, "\n\t"))
+					return errors.Errorf("Could not find module named '%s' for project '%s'; specify different project or select correct module from:\n\t%s",
+						module, projectIdentifier, strings.Join(mods, "\n\t"))
 				}
 
 				return errors.Errorf("could not set specified module: \"%s\"", module)
```
malikchaya2, line 101: It seems like this line would never be reached since it will always hit either `if len(mods) == 0` or  `if len(mods) != 0` and will always return an error.

ablack12: yep, fixed

malikchaya2 (approved):

PR #5152, opened 2021-11-02
EVG-14837: use perf count cedar endpoint

hhoke:

### Description 
Query perf endpoint to avoid trying to fetch perf results inappropriately. This includes the integration test, and a revendor of timber. Everything else has been covered in prior PR #5076, which covers up to commit 015a44a, and was reverted due to bugs that are fixed and tested here in 8456432.

### Testing 
I tested this with the included integration test. If you request that I remove this, that's fine, but I wanted to document the testing procedure I used.


int_tests/int-test-perf.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"fmt"
+
+	"github.com/evergreen-ci/evergreen/apimodels"
+)
+
+func main() {
+	ctx := context.Background()
+	opts := apimodels.GetCedarPerfCountOptions{
+		//BaseURL:   evergreen.GetEnvironment().Settings().Cedar.BaseURL,
```
julianedwards, line 13: Nit: remove comment.

int_tests/int-test-perf.go
```diff
@@ -0,0 +1,29 @@
+package main
```
julianedwards, line 1: Do we need to keep this file?

apimodels/perf.go
```diff
@@ -0,0 +1,52 @@
...
+			BaseURL: fmt.Sprintf("https://%s", opts.BaseURL),
+		},
+		TaskID:    opts.TaskID,
+		Execution: utility.ToIntPtr(opts.Execution),
+		Count:     true,
+	}
+}
+
+// CedarPerfCount holds one element, NumberOfResults, matching the json returned by
+// cedar's perf count rest route.
```
julianedwards, line 34: Nit: `Cedar's`

graphql/resolvers.go
```diff
@@ -3090,14 +3093,33 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+		}
+		opts := apimodels.GetCedarPerfCountOptions{
+			BaseURL:   evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:    *obj.Id,
+			Execution: obj.Execution,
+		}
+
+		result, err := apimodels.CedarPerfResultsCount(ctx, opts)
+		if err != nil {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("error requesting perf data from cedar: %s", err))
```
julianedwards, line 3115: `Cedar`

julianedwards (changes requested):

int_tests/int-test-perf.go
hhoke: No. I'll get rid of it.

khelif96 (approved):
LGTM pending Julians review.

apimodels/perf.go
```diff
@@ -0,0 +1,52 @@
+package apimodels
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/evergreen-ci/timber"
+	perf "github.com/evergreen-ci/timber/perf"
```
Kimchelly, line 9: This doesn't have to be explicitly labeled as "perf" since the package is already called perf.

graphql/resolvers.go
```diff
@@ -3090,14 +3093,33 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+		}
+		if result.NumberOfResults == 0 {
+
```
Kimchelly, line 3118: nit: remove whitespace

graphql/resolvers.go
```diff
@@ -3090,14 +3093,33 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+			BaseURL:   evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:    *obj.Id,
+			Execution: obj.Execution,
+		}
+
+		result, err := apimodels.CedarPerfResultsCount(ctx, opts)
+		if err != nil {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("error requesting perf data from cedar: %s", err))
+		}
+		if result.NumberOfResults == 0 {
```
Kimchelly, line 3117: This is more of a question about the perf plugin behavior - if there's no results, does that necessarily mean that the perf plugin is disabled? It seems like the perf plugin could be enabled but it just had no results.

graphql/resolvers.go
```diff
@@ -3090,14 +3093,33 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+					break
 				}
 			}
+			if !projectMatches {
+				return false, nil
+			}
+		}
+		opts := apimodels.GetCedarPerfCountOptions{
+			BaseURL:   evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:    *obj.Id,
```
Kimchelly, line 3109: I would use `utility.FromStringPtr` for this since this could panic.

Kimchelly (changes requested):

graphql/resolvers.go
khelif96: This field is used to determine if we should attempt to show the perfPlugin on the UI. If it has no results it won't actually display anything and wouldn't be useful for a user to see.

apimodels/perf.go
hhoke: Agreed. I got a linting error I resolved by adding this for some reason a while back, but this is no longer the case...

Kimchelly (approved):
LGTM assuming tests pass.

julianedwards (approved):

PR #5153, opened 2021-11-02
EVG-15764 Limit taskNamesByBuildVariant query to last 50 versions

khelif96:
[EVG-15764](https://jira.mongodb.org/browse/EVG-15764)

### Description 
This adds the same limits as #5144 to the taskNamesByBuildVariant query

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3249,7 +3249,7 @@ func (r *queryResolver) TaskNamesForBuildVariant(ctx context.Context, projectId 
 		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while getting repository for '%s': %s", projectId, err.Error()))
 	}
 	if repo == nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("could not find repository '%s'", projectId))
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("could not find repository '%s'", pid))
 	}
 	buildVariantTasks, err := task.FindTaskNamesByBuildVariant(projectId, buildVariant, repo.RevisionOrderNumber)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


model/task/db.go
```diff
@@ -963,6 +963,10 @@ func FindTaskNamesByBuildVariant(projectId string, buildVariant string) ([]strin
 				{DisplayTaskIdKey: bson.M{"$exists": false}},
 				{DisplayTaskIdKey: ""},
 			},
+			"$and": []bson.M{
+				{RevisionOrderNumberKey: bson.M{"$gte": repoOrderNumber - VersionLimit}},
```
ablack12, line 967: I think technically this should just be $gt to respect the version limit.

ablack12 (changes requested):

model/task/db.go
khelif96: done.

graphql/resolvers.go
```diff
@@ -3240,7 +3240,18 @@ func (r *queryResolver) BbGetCreatedTickets(ctx context.Context, taskID string)
...
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Could not find project with id: %s", projectId))
+	}
+	repo, err := model.FindRepository(pid)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while getting repository for '%s': %s", projectId, err.Error()))
+	}
+	if repo == nil {
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("could not find repository '%s'", projectId))
+	}
+	buildVariantTasks, err := task.FindTaskNamesByBuildVariant(projectId, buildVariant, repo.RevisionOrderNumber)
```
ablack12, line 3254: Need to use pid here as well

ablack12 (approved):
LGTM with last comment

bynn (approved):

PR #5155, opened 2021-11-03
EVG-15037 Add a graphql resolver that given a user, returns a list of grouped projects that they permissions for

malikchaya2:
[EVG-15037 ](https://jira.mongodb.org/browse/EVG-15037 )

### Description 
This groups by repoRefId so that it only groups projects by repo if it's defaulted to the repo 

### Testing 
e2e test


graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
...
-		if err = apiProjectRef.BuildFromService(p); err != nil {
-			return nil, InternalServerError.Send(ctx, fmt.Sprintf("error building APIProjectRef from service: %s", err.Error()))
-		}
-
-		if projs, ok := groupsMap[groupName]; ok {
-			groupsMap[groupName] = append(projs, &apiProjectRef)
-		} else {
-			groupsMap[groupName] = []*restModel.APIProjectRef{&apiProjectRef}
-		}
+	groupedProjects, err := GroupProjects(allProjs, false)
```
ablack12, line 1505: sorry if i've asked this before: what is this resolver used for?

graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
...
+	projectIds, err := usr.GetViewableProjects()
 
-	for groupName, groupedProjects := range groupsMap {
-		gp := GroupedProjects{
-			Name:     groupName,
-			Projects: groupedProjects,
-		}
-		groupsArr = append(groupsArr, &gp)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error getting viewable projects for '%s': '%s'", usr.DispName, err))
```
ablack12, line 1517: err.Error()

graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
...
-		}
+	groupedProjects, err := GroupProjects(allProjs, false)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error grouping project: %s", err.Error()))
 	}
+	return groupedProjects, nil
+}
 
-	groupsArr := []*GroupedProjects{}
+func (r *queryResolver) ViewableProjects(ctx context.Context) ([]*GroupedProjects, error) {
```
ablack12, line 1512: What will this resolver be used for? Should it not also show projects that don't default? (grouped by the empty string or something)

graphql/util.go
```diff
@@ -1290,3 +1290,45 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		var groupName string
+		if onlyDefaultedToRepo {
+			groupName = p.RepoRefId
+		} else {
+			groupName = strings.Join([]string{p.Owner, p.Repo}, "/")
```
ablack12, line 1302: i would suggest `fmt.Sprintf("%s/%s", p.Owner, p.Repo)`

ablack12 (changes requested):

graphql/resolvers.go
malikchaya2: It's used here: 
https://github.com/evergreen-ci/spruce/blob/main/src/pages/Commits.tsx#L131

graphql/resolvers.go
malikchaya2: It does, it's be grouped under "".

graphql/resolvers.go
ablack12: Ah is this for the mainline commits dropdown or?

graphql/resolvers.go
ablack12: Oh I think onlyDefaultedToRepo confused me; that implies to me that we're only returning projects that use the repo default. Can you add some comments to that function?

graphql/util.go
```diff
@@ -1290,3 +1290,45 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
+
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		var groupName string
+		if onlyDefaultedToRepo {
```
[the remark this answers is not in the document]
ablack12: disregard, I understand what's happening now. I think setting groupName here and then setting it again below is a bit confusing. Can I suggest setting groupName to `owner/repo` here unless `onlyDefaultedToRepo && !p.UseRepoSettings` in which case we set to the empty string? Then we can skip the re-processing later on

graphql/resolvers.go
khelif96, line 1512: I don't think a separate resolver is necessary these seem to be doing the same thing. Your version of the resolver actually seems more accurate  since its checking a users permissions and identifying what projects they have permission to view. Could we add the permission check to the projects resolver above and consolidate the two?

khelif96 (changes requested):

graphql/resolvers.go
malikchaya2: It's part of the mainline commits ui [EVG-13855](https://jira.mongodb.org/browse/EVG-13855) @khelif96 can you maybe provide more details?

graphql/resolvers.go
malikchaya2: oops, thanks!

graphql/resolvers.go
malikchaya2: I'm not sure what you mean by repo default. I set it up this way based on your comment here: [5117](https://github.com/evergreen-ci/evergreen/pull/5117#discussion_r733743503). What kind of comment would be helpful here?

malikchaya2: My resolver only groups projects by repo if they are defaulted to the repo, whereas the old one groups all of them, whether they are defaulted or not.

graphql/util.go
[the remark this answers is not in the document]
malikchaya2: that won't work because the sorting needs to happen based on the repoRefId for the onlyDefaultedToRepo. It sorts it into the based on the repoRefId and then later on changes it to the human readable form.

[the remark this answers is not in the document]
malikchaya2: In other words, I can't skip the re-proccessing because we need it to be the repoRefId at first and the displayname (owner/repo) later.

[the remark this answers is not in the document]
ablack12: But repoRefId is created based on owner/repo; it's essentially the same thing.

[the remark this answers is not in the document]
ablack12: That is, when we assign a repo ref to a project ref, we look for a corresponding owner/repo pair, and if it exists we assign the RepoRefId and if it doesn't we make a new one. 
https://github.com/evergreen-ci/evergreen/blob/f2137ddff03f27ca606dffe523b0d5a3d762c25c/model/project_ref.go#L455

graphql/resolvers.go
malikchaya2: after a conversation with Sophie, I realized that we shouldn't be fetching the entire project Ref here, so this won't be relevant anymore (unless that's also the case where you are using this)

graphql/util.go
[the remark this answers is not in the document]
malikchaya2: ahh good point, so I can use owner/repo as the key instead. I'll update it. thanks!

graphql/resolvers.go
khelif96: Its used to fetch the projects for the project selector on the waterfall. And will also be added to the projects patches page in the near future.

khelif96: I do think we should incorporate the logic chaya added to determine what projects a user can view. Otherwise this just returns all the projects available in evergreen

ablack12: Chaya's logic shows what projects a user can view _settings_ for. Theoretically this should return just the projects that a user has permission to see (tasks view permission) but I imagine that should be a separate ticket.

graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
...
+	usr := MustHaveUser(ctx)
+	projectIds, err := usr.GetViewableProjects()
```
ablack12, line 1514: I wonder if these names should be changed to GetViewableProjectSettings or something like that, to distinguish between "user can view tasks for this project" vs "users can view admin settings for this project"

graphql/schema.graphql
```diff
@@ -1014,7 +1016,9 @@ type BaseTaskInfo {
 }
 
 type GroupedProjects {
-  name: String!
+  repoDisplayName: String!
```
ablack12, line 1019: Can we call this groupDisplayName instead? Otherwise it's confusing why this isn't just repo.DisplayName

graphql/util.go
```diff
@@ -1290,3 +1290,63 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
 }
+
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		var groupName string
+		groupName = fmt.Sprintf("%s/%s", p.Owner, p.Repo)
+
+		if onlyDefaultedToRepo && p.RepoRefId == "" {
```
ablack12, line 1301: can we check UseRepoSettings instead of RepoRefId here?

graphql/util.go
```diff
@@ -1290,3 +1290,63 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+			groupsMap[groupName] = append(projs, &apiProjectRef)
+		} else {
+			groupsMap[groupName] = []*restModel.APIProjectRef{&apiProjectRef}
+		}
+	}
+
+	groupsArr := []*GroupedProjects{}
+
+	for groupName, groupedProjects := range groupsMap {
+		repoRef, err := model.FindOneRepoRef(*groupedProjects[0].RepoRefId)
```
ablack12, line 1320: We should only do this if UseRepoSettings

graphql/util.go
```diff
@@ -1290,3 +1290,63 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+		if err != nil {
+			return nil, err
+		}
+
+		apiRepoRef := restModel.APIProjectRef{}
+		if repoRef != nil {
+			if err := apiRepoRef.BuildFromService(repoRef.ProjectRef); err != nil {
+				return nil, errors.Wrap(err, "error building APIProjectRef from service")
```
ablack12, line 1328: should distinguish this error from above somehow

graphql/util.go
```diff
@@ -1290,3 +1290,63 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
+
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
```
ablack12, line 1294: Can we add a function comment here to explain what's happening? (And can this be a private function?)

graphql/util.go
```diff
@@ -1290,3 +1290,63 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		var groupName string
+		groupName = fmt.Sprintf("%s/%s", p.Owner, p.Repo)
```
ablack12, line 1299: can we do groupName := instead of declaring the var

graphql/util.go
ablack12: Also should use utility.FromStringPtr

ablack12 (changes requested):

graphql/resolvers.go
malikchaya2: good point. done.

graphql/schema.graphql
malikchaya2: I like that. done.

graphql/util.go
malikchaya2: I did this in my head but apparently forgot to write it in code :smile:
thank you!

graphql/util.go
malikchaya2: done

graphql/util.go
malikchaya2: oops, leftover from before the refactor. thanks!

graphql/util.go
malikchaya2: done

graphql/resolvers.go
ablack12: Can this also be ViewableProjectSettings ?

graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
...
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error getting viewable projects for '%s': '%s'", usr.DispName, err.Error()))
 	}
 
-	sort.SliceStable(groupsArr, func(i, j int) bool {
-		return groupsArr[i].Name < groupsArr[j].Name
-	})
+	projects, err := model.FindProjectRefsByIds(projectIds)
+	if err != nil {
+		return nil, err
```
ablack12, line 1522: Should this use the graphql format InternalServerError ?

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+
+// GroupProjects takes a list of projects and groupes them by their repo. If onlyDefaultedToRepo is true,
+// it only groups projects that are defaulted to the repo
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		groupName := fmt.Sprintf("%s/%s", p.Owner, p.Repo)
+
+		if onlyDefaultedToRepo && p.UseRepoSettings {
```
ablack12, line 1302: !UseRepoSettings, right? If it does use repo settings then we _do_ want it to be in the group

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+
+	for groupName, groupedProjects := range groupsMap {
+		gp := GroupedProjects{
+			Name:             &groupName, //deprecated
+			GroupDisplayName: groupName,
+			Projects:         groupedProjects,
+		}
+
+		if groupedProjects[0].UseRepoSettings {
+			repoRef, err := model.FindOneRepoRef(utility.FromStringPtr(groupedProjects[0].RepoRefId))
```
ablack12, line 1328: Should check nil

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+				return nil, err
+			}
+
+			apiRepoRef := restModel.APIProjectRef{}
+			if repoRef != nil {
+				if err := apiRepoRef.BuildFromService(repoRef.ProjectRef); err != nil {
+					return nil, errors.Wrap(err, "error building the repo's ProjectRef from service")
+				}
+			}
+			if repoRef != nil {
```
ablack12, line 1339: can this be in the above if statement?

model/project_ref_test.go
```diff
@@ -786,29 +786,6 @@ func TestDefaultRepoBySection(t *testing.T) {
 	}
 }
 
-func TestGroupProjectsByRepo(t *testing.T) {
```
ablack12, line 789: Can we make a new test (or adapt this one) for GroupProjects?

malikchaya2: There is a graphql test for it 
https://github.com/evergreen-ci/evergreen/blob/1e8af407957a6e03874fb1c101009e681f0f9166/graphql/tests/project/results.json

graphql/util.go
malikchaya2: yes :see_no_evil:

graphql/util.go
malikchaya2: yes 🙈

graphql/resolvers.go
malikchaya2: Since the resolver "Projects" returns projects, I'm worried that people will expected "ViewableProjectSettings" to return ProjectSettings (as opposed to projects).

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
+
+// GroupProjects takes a list of projects and groupes them by their repo. If onlyDefaultedToRepo is true,
```
khelif96, line 1294: groups*

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+// it only groups projects that are defaulted to the repo
+func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*GroupedProjects, error) {
+	groupsMap := make(map[string][]*restModel.APIProjectRef)
+
+	for _, p := range projects {
+		groupName := fmt.Sprintf("%s/%s", p.Owner, p.Repo)
+
+		if onlyDefaultedToRepo && p.UseRepoSettings {
+			groupName = ""
+		}
```
khelif96, line 1304: This would change the group name we show on the project select dropdown. Is this new name appropriate/ something we should show to all the users?

graphql/util.go
```diff
@@ -1290,3 +1290,67 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+
+			apiRepoRef := restModel.APIProjectRef{}
+			if repoRef != nil {
+				if err := apiRepoRef.BuildFromService(repoRef.ProjectRef); err != nil {
+					return nil, errors.Wrap(err, "error building the repo's ProjectRef from service")
+				}
+			}
+			if repoRef != nil {
+				gp.Repo = &apiRepoRef
+				if repoRef.DisplayName != "" {
```
khelif96, line 1341: Can you explain what this is doing?
This looks pretty expensive if we have to do this query for every project we are returning. (There appears to be > 100 in the dropdown)

khelif96 (changes requested):

graphql/util.go
malikchaya2: right now it shows "/". I'm not a fan of that either. I figured I'll leave it blank for now and we can decide when we build the UI.

graphql/resolvers.go
```diff
@@ -1502,37 +1502,31 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
 
-	groupsMap := make(map[string][]*restModel.APIProjectRef)
```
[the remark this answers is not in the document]
malikchaya2: Can you please create a separate ticket for that?

graphql/testdata/projects1.graphql
```diff
@@ -1,6 +1,6 @@
 {
   projects {
```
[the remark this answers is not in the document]
malikchaya2: We should probably do that as part of the ticket we create to add the filtering to this resolver so that we can transfer the tests.

graphql/util.go
malikchaya2: This will only be fetched for projects that default to the repo. I suspect that there won't be too many, especially to start. It fetches the repoRef so that we can access the repoId (so that we can use it for routing, so that the repo itself is clickable), and so that we can use the repoDisplayName if the user set it instead of owner/repo.

graphql/util.go
malikchaya2: oy. thank you!

graphql/schema.graphql
```diff
@@ -1014,7 +1016,9 @@ type BaseTaskInfo {
 }
 
 type GroupedProjects {
-  name: String!
+  repoDisplayName: String! 
+  name: String @deprecated
```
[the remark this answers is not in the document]
malikchaya2: https://jira.mongodb.org/browse/EVG-15819

graphql/util.go
malikchaya2: it's there now

graphql/util.go
malikchaya2: I'm remembering now why I used RepoRefId instead of UseRepoSettings. The tests only work with RepoRefId. I can explain more. Can we maybe use RepoRefId here since it has the same affect?

graphql/util.go
ablack12: It's also not for every project, it's for every owner/repo pair.

graphql/resolvers.go
ablack12: hmm I see your point. Maybe ViewableProjectRef?  "Project" is just a bit vague.

graphql/util.go
ablack12: Sorry I meant we should check if repoRef is nil (maybe just logging a warning idk)

graphql/resolvers.go
malikchaya2: That sounds good, I'll update it to ViewableProjectRefs

graphql/util.go
malikchaya2: done

graphql/util.go
```diff
@@ -1323,15 +1323,22 @@ func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
...
+			repoRef, err := model.FindOneRepoRef(repoRefId)
 			if err != nil {
 				return nil, err
 			}
 
 			apiRepoRef := restModel.APIProjectRef{}
-			if repoRef != nil {
+			if repoRef == nil {
+				grip.Warning(message.WrapError(err, message.Fields{
+					"message":   "repoRef not found ",
```
ablack12, line 1337: nit: extra space, and by convention we use snake case (repo_ref_id) for splunk. I think we can log this at the actual Error level also, since this does indicate a real error.

graphql/util.go
```diff
@@ -1323,15 +1323,22 @@ func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
...
-			repoRef, err := model.FindOneRepoRef(utility.FromStringPtr(groupedProjects[0].RepoRefId))
+		project := groupedProjects[0]
+		if project.UseRepoSettings {
+			repoRefId := utility.FromStringPtr(project.RepoRefId)
+			repoRef, err := model.FindOneRepoRef(repoRefId)
 			if err != nil {
 				return nil, err
 			}
 
 			apiRepoRef := restModel.APIProjectRef{}
```
ablack12, line 1334: Can we initialize this after the repoRef == nil check?

ablack12 (approved):
LGTM with some small fixes

khelif96:
I think you need to merge in main and rerun gqlgen

malikchaya2:
> I think you need to merge in main and rerun gqlgen

done

graphql/schema.graphql
```diff
@@ -1016,7 +1018,9 @@ type BaseTaskInfo {
 }
 
 type GroupedProjects {
-  name: String!
+  groupDisplayName: String! 
+  name: String @deprecated(reason: "name is deprecated. Use groupDisplayName instead.")
```
khelif96, line 1022: can you re add the `!` since that field should always be present.

graphql/tests/project/queries/viewableProjects.graphql
```diff
@@ -0,0 +1,11 @@
...
+  viewableProjectRefs {
+    groupDisplayName
+    projects{
+      identifier
+      repo
+      owner 
+      repoRefId
+    }
+  } 
+}
```
khelif96, line 11: nit. eol

graphql/util.go
```diff
@@ -1290,3 +1290,71 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+					gp.GroupDisplayName = repoRef.DisplayName
+				}
+			}
+		}
+
+		groupsArr = append(groupsArr, &gp)
+	}
+
+	sort.SliceStable(groupsArr, func(i, j int) bool {
+
```
khelif96, line 1356: nit extra line.

graphql/util.go
```diff
@@ -1290,3 +1290,71 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+	groupsArr := []*GroupedProjects{}
+
+	for groupName, groupedProjects := range groupsMap {
+		gp := GroupedProjects{
+			Name:             &groupName, //deprecated
+			GroupDisplayName: groupName,
+			Projects:         groupedProjects,
+		}
+		project := groupedProjects[0]
+		if project.UseRepoSettings {
```
khelif96, line 1327: Why do we only care about the first project in this array?

graphql/util.go
```diff
@@ -1290,3 +1290,71 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
...
+			repoRefId := utility.FromStringPtr(project.RepoRefId)
+			repoRef, err := model.FindOneRepoRef(repoRefId)
+			if err != nil {
+				return nil, err
+			}
+
+			if repoRef == nil {
+				grip.Error(message.WrapError(err, message.Fields{
```
khelif96, line 1335: I'm not sure on go conventions if err is nil here would that throw a nil pointer exception?

malikchaya2: we know that err isn't nil because of the check above. But either way, I don't think it will since we are not accessing err.Error.

graphql/util.go
malikchaya2: because all the projects in the group have the same information, so we only need one. The first one is arbitrary, it can be any.

graphql/schema.graphql
malikchaya2: sure

graphql/util.go
khelif96: err is nil since the check above would have returned an error if error wasn't nil.

khelif96 (approved):

graphql/util.go
malikchaya2: ah yes, I read that backwards. I think it'll just log err as nil and log the rest of the fields. I'll double check before merging though.

graphql/util.go
```diff
@@ -1290,3 +1290,70 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
+
+// GroupProjects takes a list of projects and groups them by their repo. If onlyDefaultedToRepo is true,
+// it only groups projects that are defaulted to the repo
```
sophstad, line 1295: This comment threw me off slightly—it suggests that projects that don't default to repo are not returned, but they're actually grouped under `""` right? That feels worth mentioning here.

sophstad (changes requested):
Just a few minor fixes, nice work on this Chaya!!

graphql/util.go
malikchaya2: good point, I updated the comment.

graphql/integration_spec.json
```diff
@@ -118,7 +118,7 @@
         ],
         "patches": [
           {
-            "name": "",
+             "name": "",
```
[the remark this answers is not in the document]
malikchaya2: fixed

graphql/schema.graphql
```diff
@@ -6,6 +6,7 @@ type Query {
   patch(id: String!): Patch!
   version(id: String!): Version!
   projects: [GroupedProjects]!
+  ViewableProjectRefs: [GroupedProjects]!
```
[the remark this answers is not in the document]
malikchaya2: done

graphql/util.go
sophstad: Missing a git push I think :)

malikchaya2: my bad, pushed, should be there now

sophstad (approved):

PR #5156, opened 2021-11-03
EVG-14660 Update notifications to use new ui link when possible

bynn:
[EVG-14660](https://jira.mongodb.org/browse/EVG-14660)

### Description 
links now have `?redirect_spruce_users=true`

### Testing 
on staging 
image
https://evergreen-staging.corp.mongodb.com//task/evg_lint_generate_lint_e47bb73265e2eb443ea129a3b444336a3da1d53b_21_10_22_21_07_36/1/?redirect_spruce_users=true

caused a user not found panic once but im not sure what triggered it

### DIFF
--- a/service/task.go
+++ b/service/task.go
@@ -154,7 +154,7 @@ func (uis *UIServer) taskPage(w http.ResponseWriter, r *http.Request) {
 	if r.FormValue("redirect_spruce_users") == "true" {
 		if u := gimlet.GetUser(r.Context()); u != nil {
 			usr, ok := u.(*user.DBUser)
-			if ok && usr.Settings.UseSpruceOptions.SpruceV1 {
+			if ok && usr != nil && usr.Settings.UseSpruceOptions.SpruceV1 {
 				http.Redirect(w, r, fmt.Sprintf("%s/task/%s", uis.Settings.Ui.UIv2Url, projCtx.Task.Id), http.StatusTemporaryRedirect)
 				return
 			}
--- a/service/task.go
+++ b/service/task.go
@@ -18,6 +18,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/evergreen/model/user"
 	"github.com/evergreen-ci/evergreen/plugin"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/gimlet"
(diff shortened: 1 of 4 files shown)


trigger/payloads.go
```diff
@@ -438,7 +438,7 @@ func getFailedTestsFromTemplate(t task.Task) ([]task.TestResult, error) {
 }
 
 func taskLink(uiBase string, taskID string, execution int) string {
-	return fmt.Sprintf("%s/task/%s/%d", uiBase, url.PathEscape(taskID), execution)
```
ablack12, line 441: should this consider hasPatch @khelif96 ?

ablack12 (approved):

ablack12 (commented):
Looks like the test failure is relevant.

trigger/payloads.go
khelif96: I don't think so since mainline commits are now supported on spruce we don't need to check if a patch exists.

trigger/payloads.go
```diff
@@ -438,7 +438,7 @@ func getFailedTestsFromTemplate(t task.Task) ([]task.TestResult, error) {
...
+	return fmt.Sprintf("%s/task/%s/%d/?redirect_spruce_users=true", uiBase, url.PathEscape(taskID), execution)
```
khelif96, line 441: The image you shared has a malformed link. (Two forward slashes after mongodb.com) Can you update the link to remove the extra forward slash. 
image

khelif96: We have some logic in gimlet that parses out the double forward slash but we shouldn't be generating links like that anyways.  https://github.com/evergreen-ci/gimlet/pull/97

khelif96 (approved):
Apart from the one comment LGTM

trigger/payloads.go
ablack12: also shouldn't put a slash before ? for correct syntax https://help.marketruler.com/wiki/What_is_the_correct_syntax_for_query_strings%3F

bynn: done

trigger/payloads.go
```diff
@@ -438,7 +438,7 @@ func getFailedTestsFromTemplate(t task.Task) ([]task.TestResult, error) {
 }
 
 func taskLink(uiBase string, taskID string, execution int) string {
-	return fmt.Sprintf("%s/task/%s/%d/?redirect_spruce_users=true", uiBase, url.PathEscape(taskID), execution)
+	return fmt.Sprintf("%s/task/%s/%d?redirect_spruce_users=true", uiBase, url.PathEscape(taskID), execution)
```
khelif96, line 441: This doesn't seem to change the double forward slash at the beginning of the path. Can you double check on staging and validate that it is no longer the case. We had a similar case like this in the past and it is not a valid link. https://github.com/evergreen-ci/gimlet/pull/97#discussion_r713430345

bynn: there was an extra / in the admin page that the prod page didn't have so i fixed it (updated the picure)

bynn: the fix was changing base url on the admin page. if we want a validation we would make it there but annie said there's no need since we never change it

service/task.go
```diff
@@ -151,10 +152,12 @@ func (uis *UIServer) taskPage(w http.ResponseWriter, r *http.Request) {
 	projCtx := MustHaveProjectContext(r)
 
 	if r.FormValue("redirect_spruce_users") == "true" {
-		user := MustHaveUser(r)
-		if user.Settings.UseSpruceOptions.SpruceV1 {
-			http.Redirect(w, r, fmt.Sprintf("%s/task/%s", uis.Settings.Ui.UIv2Url, projCtx.Task.Id), http.StatusTemporaryRedirect)
-			return
+		if u := gimlet.GetUser(r.Context()); u != nil {
+			usr, ok := u.(*user.DBUser)
+			if ok && usr.Settings.UseSpruceOptions.SpruceV1 {
```
ablack12, line 157: we actually do need to check here that usr isn't nil again, because of how nil interfaces work in go  (https://tour.golang.org/methods/12)

ablack12 (approved):
LGTM after comment addressed

service/task.go
bynn: done

PR #5158, opened 2021-11-03
EVG-14292 only log notifications when insert error isn't duplicate error

ablack12:
[EVG-14292](https://jira.mongodb.org/browse/EVG-14292)

### Description 
Right now we actually typically log these errors twice because we also add it to the job; we really only need to log here when the error is unexpected (because we also log notifications). It appears that we [do expect duplicate insert errors](https://github.com/evergreen-ci/evergreen/blob/9316e9aad79073b2e37dcca13314f76cfa56e4fa/model/notification/db.go#L103), so we shouldn't log these specifically.

### DIFF
--- a/units/event_notifier.go
+++ b/units/event_notifier.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/model/event"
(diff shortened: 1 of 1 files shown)


ybrill (approved):
did you consider also not adding it to the job's errors?

bynn (approved):

ablack12:
I considered it but wasn't sure if we'd want it for some really strange debugging. I guess given the assumption that it shouldn't be important it seems reasonable though.

ybrill (approved):

PR #5159, opened 2021-11-03
EVG-15779 add logging when creating github hook

ablack12:
[EVG-15779](https://jira.mongodb.org/browse/EVG-15779)

### Description 
Given logging and the fact that we _do_ have a github hook ID in the DB, I kind of feel like this was a github issue, or that errors are now returned to us in ways we don't know how to process, so I'd like to log this body.

### DIFF
--- a/model/github_hook.go
+++ b/model/github_hook.go
@@ -6,15 +6,14 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/mongodb/grip"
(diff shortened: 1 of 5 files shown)


malikchaya2 (approved):
Is this meant to give us more information so that we can figure out what's wrong? 
If so, should we maybe add the ticket ID to the message to make it more robust for people viewing it in splunk?

PR #5162, opened 2021-11-04
EVG-15791: resolve baseTaskId in baseTaskMetadata gql type

SupaJoon:
[EVG-15791](https://jira.mongodb.org/browse/EVG-15791)

### Description 
This will be used to generate a spruce link on the task details metadata card


khelif96 (changes requested):
Can you use the `baseTask` type on `task` instead `baseTaskMetadata`. `baseTaskMetadata` should be deprecated since it is a subset of `baseTask` and we should be using `baseTask` directly instead.

PR #5164, opened 2021-11-04
EVG-15625, EVG-15626, EVG-15627: use Go modules

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-15625
https://jira.mongodb.org/browse/EVG-15626
https://jira.mongodb.org/browse/EVG-15627

#### PR Reviewer Note
Since the diff is gigantic, I don't think GitHub can even load it. If you want to view it locally:
```sh
    # If you use origin instead of upstream, replace upstream with origin.
    git fetch upstream pull/5164/head:5164
    git checkout 5164
```

For reading the diff (minus the part where I blew up the vendor directory), you can use this:
```sh
git diff $(git merge-base "$(git branch --show-current)" main) -- ':^vendor'
```

Or if you prefer viewing a side-by-side diff in your browser, you can install [webdiff](https://github.com/danvk/webdiff/) and install with `pip install webdiff`:
```sh
git webdiff $(git merge-base "$(git branch --show-current)" main) -- ':^vendor'
```

### Description 

There's no way to make this PR smaller without leaving something in a half-broken vendor-dependent state. Basically every change in this PR is risky. I won't merge this until next week, and I'll fix the inevitable merge conflicts that arise from other PRs before it gets merged.

* Replace vendor directory and glide with Go modules. I removed all the revendoring tools because they will not apply in Go modules and people will have to know how to use Go modules tooling rather than just blindly rely on a single make target to do what they need. I'm going to document it in [EVG-15450](https://jira.mongodb.org/browse/EVG-15450), but I don't think it matters much except for PLT since others don't mess with dependency versions that often.
* Remove GOPATH dependencies in our dev/build process.
* Upgrade all evergreen-ci/monogdb repos to latest version and fix up backward incompatibilities introduced because Evergreen is using stale copies of some repos.
* Fix `github.com/sabhiram/go-git-ignore`, which is now `github.com/sabhiram/go-gitignore`.
* Remove `evg-lint` linter until [EVG-15453](https://jira.mongodb.org/browse/EVG-15453). This is only temporary and it only catches very specific things (e.g. testify `TearDownSuite` capitalization and `defer`s in for loops).
* Move mgo hacks into a subpackage of Evergreen. We're using [a special branch of mgo](https://github.com/go-mgo/mgo/tree/60f23985327354fe2d4dde32ba05598bb047a114) and the vendored copy of mgo has been manually modified to support the mongo-driver (and prevent our DB tests from failing), so we have no choice but to move it into Evergreen since it's not really an official version of mgo anymore.
    * I read over the license and it seems to allow this, as long as we keep the license with the code (I'm not a license lawyer though).
* Fix graphql `gqlgen` target and GraphQL tooling to not depend on vendoring the gqlgen tool. I don't really know if it auto-generated the proper `graphql/generated.go` file properly, but the version we were pinned to in glide (v0.11.2) is not compatible with Go modules. However, upgrading gqlgen is risky since it has no stable releases, so every new version has some risk of backward incompatibility. We've upgraded from v0.11.2 to v0.14.0 and it seems like it generated a syntactically-valid `graphql/generated.go`, but I'm not really qualified to say if it's correct or not.
    * Because `gqlgen` is a CLI utility that we need to run for auto-generation, I followed [the recommended way to track Go tool dependencies in a repo](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) and made a tools file for it (`cmd/gqlgen/gqlgen.go`).
* Remove codegen hacks in the makefile, which relied on `go get`'ing `goimports` into the `GOPATH`. It now just relies on it like a regular module.

### Testing 
Evergreen is unable to test a diff this large, so I can't verify that the tests pass in CI.

I ran all the local non-integration tests and the linter. I also deployed it to staging and verified that I could run a task.

I ran `gqlgen` for the auto-generated graphql and `generate-rest-model` for the auto-generated REST routes to verify that they at least seem to do something sensible.

I ran `generate-lint` to see if it produced a reasonable-looking generate.tasks JSON flie.


khelif96:
Can we merge this into a temporary branch and run a patch build against it? Is that possible?

khelif96:
Thanks for updating gqlgen! The staging server still works and responds to gql requests.  I have not tested code generation or any of our tests yet.

Kimchelly:
I don't know how far the [projects track multiple branches](https://jira.mongodb.org/browse/PM-610) project has gone along, but as far as I'm aware, it's not possible to run Evergreen tests against an alternative branch.

It may (?) be possible to make a new Evergreen project just for testing this PR, but it would have to be against my fork of Evergreen, and would have to include the other things like AWS keys, GitHub keys, etc. that we include in CI integration tests. Not sure how well that would go.

bsamek:
@khelif96 @Kimchelly We could merge and then revert this commit. That would give us a build of this commit so that we could review the test results.

Kimchelly:
We could do this, but I'm just waiting on @bsamek to okay commit/revert as a valid testing strategy.

bsamek:
> We could do this, but I'm just waiting on @bsamek to okay commit/revert as a valid testing strategy.

Yes go for it.

> Can I get some reviews on this? The longer this sits, the more merge conflicts are going to come up.

I can review this tomorrow.

PR #5165, opened 2021-11-04
EVG-15307 consider patch aliases as overridden even if one doesn't match

ablack12:
[EVG-15307](https://jira.mongodb.org/browse/EVG-15307)

### Description 
I realized there's a bug with how patch aliases work for project/repo; right now we check "do we have a matching alias" and if we don't we check the next level (parser project -> project -> repo). This works for internal aliases like __github_alias, however patch aliases can be named anything. If the project overrides the repo by defining patch alias "alias1" and "alias2", we shouldn't continue to check the repo when "alias3" doesn't show up.

Alternatively we can change the design to have patch aliases work like project variables 😬  But maybe let's start with this and reconsider the design if needed?

### Testing 
Added a unit test case and verified existing tests pass.

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -70,23 +70,6 @@ type ProjectAlias struct {
 
 type ProjectAliases []ProjectAlias
 
-// FindAllAliasesForProject fetches aliases for a project from the aliases collection
-// and merges them with the aliases defined in the project yaml.
-func FindAllAliasesForProject(projectID string) ([]ProjectAlias, error) {
-	dbAliases, err := FindAliasesForProjectFromDb(projectID)
-	if err != nil {
-		return nil, err
-	}
-	parserProject, err := ParserProjectByVersion(projectID, "")
-	if err != nil {
-		return nil, errors.Wrap(err, "error finding project parser")
-	}
-	if parserProject != nil {
-		return mergeAliases(parserProject, dbAliases), nil
-	}
-	return dbAliases, nil
-}
-
 // FindAliasesForProjectFromDb fetches all aliases for a given project without merging with aliases from the parser project
 func FindAliasesForProjectFromDb(projectID string) ([]ProjectAlias, error) {
 	var out []ProjectAlias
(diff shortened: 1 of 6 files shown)


malikchaya2 (commented):
I thought that the way we set it up was to always use repo level aliases unless the branch has overwritten it. So if a repo has the following aliases: 
 alias1 = value1
alias2 = value2 
alias3 = value 3, 

and the branch owner wants a custom alias for 1 and 2 and nothing for 3, they would need to define it as: 
 alias1 = custom1
alias2 = custom2 
alias3 = "".

Do we want to instead change the design to use only branch level aliases by default, and a mix like described above only if a boolean is toggled?

ablack12:
Referencing the design: "by default" or not isn't really what I'm fixing. It's that we either override ALL repo patch aliases or we use all patch aliases (I think you're thinking of the project variable usage, which reads from both).

 So in the example you discussed, "if the branch owner wants a custom alias for 1 and 2 and nothing for 3":
 they should just have to define alias1 and alias2, and alias3 is overridden by default.
image

Alternatively we could change the design to work how variables work.

malikchaya2:
> Alternatively we could change the design to work how variables work.

What is the benefit of the defaulting behavior being different for patch aliases and project variables?

hadjri (changes requested):
Code logic looks sound.  Can we add some more detailed comments in the large block line 34-58 describing the pattern in which we override and why returning boolean parameters when retrieving patch aliases are necessary?

model/project_aliases.go
```diff
@@ -221,24 +222,24 @@ func mergeAliases(parserProject *ParserProject, databaseAliases []ProjectAlias)
...
 // If the project has no aliases, the repo is checked for aliases.
 func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
-	aliases, err := findAliasInProject(projectID, alias)
+	aliases, shouldExit, err := findMatchingAliasForParserProject(projectID, alias)
 	if err != nil {
-		return aliases, errors.Wrapf(err, "error finding aliases for project '%s'", projectID)
+		return aliases, errors.Wrapf(err, "error checking parser project for project '%s' for aliases", projectID)
 	}
-	if len(aliases) > 0 {
+	if shouldExit {
```
malikchaya2, line 229: shouldExist could be true while aliases is empty. Should we not be fetching it from somewhere else in that case?

ablack12: Do you mean shouldExit? ShouldExit is true when aliases is empty in the case I described earlier, where other patch aliases _are_ defined.

ablack12: If I'm misunderstanding, can you provide an example of what you mean?

ablack12:
@hadjri can you indicate with a comment where you're referring to? I do have this comment above the relevant functions
```
//  other aliases in the category are defined, in which case we shouldn't check other sources.
```

ablack12:
closed on accidnet 😂

malikchaya2 (approved):

model/project_aliases.go
malikchaya2: Ah, I misread it

hadjri (approved):

PR #5166, opened 2021-11-04
EVG-15457 Add hasVersion resolver

khelif96:
[EVG-15457](https://jira.mongodb.org/browse/EVG-15457)

### Description 
This will serve as an "introspection" query for patch and version ids. The client will be able to pass in an id to this query and it will return a boolean indicating whether or not a version exists for this task. Which can be used to determine if we need to redirect the user. 

This is in response to 
https://github.com/evergreen-ci/spruce/pull/961#discussion_r742113228
So we can avoid requests to the patches resolver that might result in superfluous errors being logged from using that resolver to validate ids.
### Testing 
Unit tests


graphql/resolvers.go
```diff
@@ -184,6 +184,20 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
...
+	// a valid patch or version id.
+	p, _ := patch.FindOneId(id)
+	if p != nil {
+		return PatchOrVersionTypePatch, nil
+	}
+	v, _ := r.sc.FindVersionById(id)
+	if v != nil {
+		return PatchOrVersionTypeVersion, nil
+	}
+	return "", ResourceNotFound.Send(ctx, fmt.Sprintf("Unable to find patch or version %s", id))
```
ablack12, line 198: Would it be helpful to save the error and still return it here?

graphql/resolvers.go
```diff
@@ -184,6 +184,20 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
 	return &apiHost, nil
 }
 
+func (r *queryResolver) IsPatchOrVersion(ctx context.Context, id string) (PatchOrVersionType, error) {
```
ablack12, line 187: Could this instead be HasPatch and return a boolean or something? Technically patches can have versions and the ID would work for both; also "PatchOrVersionTypePatch" is wordy. Don't have the full use case picture though.

sophstad: I prefer this approach as well!

khelif96: The way i imagined this working was the following.
Client sends the id to `IsPatchOrVersion` 
GQL returns `PATCH` or `VERSION` or `ERROR`
- if `PATCH` is returned 
  1) We can safely request data about the patch and send out an initial request which returns data about if the patch is configured (And has a version) and its commit queue status. 
     - if the Patch is configured and has a Version we know we can safely request the version data that we need to render the page
     - If the patch is on the commit queue but not yet running we can redirect to the commit queue page
     - If the Patch is not configured we can redirect to the configure page
- If `VERSION` is returned 
  1) We can safely request the necessary version data we need to render the page since we know the version exists
- if `ERROR` is returned
  1) We can throw an error because we know there is no version or patch in evergreen with the associated ID and we don't have to send out any subsequent requests to determine this at a later stage in loading the page.

khelif96: `PatchOrVersionTypePatch` This was autogenerated by gqlgen.

graphql/resolvers.go
```diff
@@ -189,6 +189,20 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
...
+	// We do not check the error here because we only want to return an error if the id is not
+	// a valid patch or version id.
+	p, _ := patch.FindOneId(id)
+	if p != nil {
+		return PatchOrVersionTypePatch, nil
+	}
+	v, _ := r.sc.FindVersionById(id)
+	if v != nil {
+		return PatchOrVersionTypeVersion, nil
+	}
```
SupaJoon, line 202: nit: can we can put these condition on one line

SupaJoon (approved):

graphql/resolvers.go
ablack12: I guess I'm still not sure why this can't be a boolean instead. "IsPatchOrVersion" just feels misleading; patches can also have versions. 

If "the function was "HasVersion": 
* If `TRUE` is returned, do the "is version is returned" step.
* If `FALSE` is returned, do the "is patch is returned" step (because if it doesn't have a version then it does have a patch).
* If `ERROR` is returned (which we should return if we can't validate the patch / version thing): do the error step.

This eliminates the need for "PatchOrVersionType" at all.

khelif96: Great point! I didn't consider that. Just made that change.

sophstad (approved):

graphql/resolvers.go
```diff
@@ -189,6 +189,18 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
 	return &apiHost, nil
 }
 
+func (r *queryResolver) HasVersion(ctx context.Context, id string) (bool, error) {
+	// We do not check the error here because we only want to return an error if the id is not
+	// a valid patch or version id.
+	if v, _ := r.sc.FindVersionById(id); v != nil {
```
ablack12, line 195: Can we case on the error? If it's not StatusNotFound we should still error (or at least log an error)

graphql/resolvers.go
```diff
@@ -3048,7 +3060,7 @@ func (r *taskResolver) CanSchedule(ctx context.Context, obj *restModel.APITask)
 	if err != nil {
 		return false, err
 	}
-	return *canRestart == false && !obj.Aborted, nil
+	return !utility.FromBoolPtr(canRestart) && !obj.Aborted, nil
```
ablack12, line 3063: This seems like an unrelated change but it seems better to just have canRestartTask not return a pointer

ablack12 (changes requested):

graphql/resolvers.go
khelif96: Yeah this is unrelated i made this change  to suppress a random warning i saw while scrolling through this file. Theres a ticket tracking refactoring that function here. https://jira.mongodb.org/browse/EVG-15698

graphql/resolvers.go
```diff
@@ -192,11 +192,33 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
...
 	// a valid patch or version id.
-	if v, _ := r.sc.FindVersionById(id); v != nil {
+	v, err := r.sc.FindVersionById(id)
+	if err != nil {
+		apiErr, ok := err.(gimlet.ErrorResponse)
+		if !ok {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("Error finding version %s: %s", id, err.Error()))
+		}
+		if apiErr.StatusCode != http.StatusNotFound {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("Error finding version %s: %s", id, err.Error()))
```
ablack12, line 202: can use apiErr.Message here now instead of err.Error() (this just removes the status code)

graphql/resolvers.go
```diff
@@ -189,6 +189,40 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
 	return &apiHost, nil
 }
 
+func (r *queryResolver) HasVersion(ctx context.Context, id string) (bool, error) {
+	// We do not check the error here because we only want to return an error if the id is not
```
ablack12, line 193: need to update this comment

graphql/resolvers.go
```diff
@@ -189,6 +189,40 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
...
+		return true, nil
+	}
+	if patch.IsValidId(id) {
+		p, err := patch.FindOneId(id)
+		if err != nil {
+			apiErr, ok := err.(gimlet.ErrorResponse)
+			if !ok {
+				return false, InternalServerError.Send(ctx, fmt.Sprintf("Error finding version %s: %s", id, err.Error()))
+			}
+			if apiErr.StatusCode != http.StatusNotFound {
```
ablack12, line 215: Don't need to check status code here because you aren't using a connector.
Also actually sorry I didn't realize before; if you don't want to do the StatusNotFound drama above, you can use the actual DB function get the version

khelif96: Thanks the drama has been dropped!

graphql/resolvers.go
```diff
@@ -189,6 +189,27 @@ func (r *volumeResolver) Host(ctx context.Context, obj *restModel.APIVolume) (*r
...
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("Error finding version %s: %s", id, err.Error()))
+	}
+	if v != nil {
+		return true, nil
+	}
+
+	if patch.IsValidId(id) {
+		p, err := patch.FindOneId(id)
+		if err != nil {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("Error finding version %s: %s", id, err.Error()))
```
ablack12, line 204: error finding patch

ablack12 (approved):
LGTM with nit

PR #5169, opened 2021-11-05
EVG-15334 Handle 16 mb limit on unpaginated requests

khelif96:
[EVG-15334](https://jira.mongodb.org/browse/EVG-15334)

### Description 
We would hit  the 16 mb document limit if we try to load a large version with allot of tasks in the `$facet` stage. 
When we would make non limited/paginated requests to this query all of the returned tasks would be embedded in one document causing it to exceed the 16 mb document limit mongodb imposes.  

We can get around this by not using the `$facet` if the results are not paginated. Instead computing the total count in go (Because we have access to all of the results in this scenario).  

This allows us to preserve the advantages the `$facet` provided (Not needing to perform 2 queries to fetch the count without the pagination) While also avoiding the document size limit in the scenarios where we would run into them.

### Testing 
Staging and existing unit tests

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -888,15 +888,6 @@ func FindUniqueBuildVariantNamesByTask(projectId string, taskName string, repoOr
 		},
 	}
 
-	// sort by most recent version to get the most recent display names for the build variants first
-	sortByOrderNumber := bson.M{
-		"$sort": bson.M{
-			CreateTimeKey: -1,
-		},
-	}
-
-	pipeline = append(pipeline, sortByOrderNumber)
-
 	// group the build variants by unique build variant names and get a build id for each
 	groupByBuildVariant := bson.M{
 		"$group": bson.M{
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -3029,30 +3029,27 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 		return nil, 0, err
 	}
 
-	type TasksAndCount struct {
-		Tasks []Task           `bson:"tasks"`
-		Count []map[string]int `bson:"count"`
-	}
-	TaskAndCountResults := []TasksAndCount{}
-	TaskResults := []Task{}
 	var results []Task
 	var count int
 
 	// If there is no limit applied we should just return the tasks and compute the total count in go.
 	// This avoids hitting the 16 MB limit on the aggregation pipeline in the $facet stage https://jira.mongodb.org/browse/EVG-15334
 	if opts.Limit > 0 {
-		err = cursor.All(ctx, &TaskAndCountResults)
+		type TasksAndCount struct {
+			Tasks []Task           `bson:"tasks"`
+			Count []map[string]int `bson:"count"`
+		}
+		taskAndCountResults := []TasksAndCount{}
+		err = cursor.All(ctx, &taskAndCountResults)
 		if err != nil {
 			return nil, 0, err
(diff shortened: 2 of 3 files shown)


model/task/db.go
```diff
@@ -888,15 +888,6 @@ func FindUniqueBuildVariantNamesByTask(projectId string, taskName string, repoOr
 		},
 	}
 
-	// sort by most recent version to get the most recent display names for the build variants first
-	sortByOrderNumber := bson.M{
```
khelif96, line 892: Not related to 16mb limit but its a performance optimization for a query used on the task history page. 
This sort was preventing the use of an index and was very slow on large projects like `mongodb-mongo-*`.

Its no longer necessary since we are limiting to the 50 most recent versions we don't have to worry about older build variant names being shown.

model/task_build_variant_names_test.go
```diff
@@ -64,28 +64,11 @@ func TestFindUniqueBuildVariantNamesByTask(t *testing.T) {
 		RevisionOrderNumber: 1,
 	}
 	assert.NoError(t, t3.Insert())
-	b4 := build.Build{
```
khelif96, line 67: Removed a test case that was testing to see if this query could handle build variant names changing. 
This is no longer necessary since we only fetch build variant names from the most recent 50 versions as opposed to all time.

SupaJoon (approved):

model/task/task.go
```diff
@@ -2969,80 +2969,102 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
...
-	err = cursor.All(ctx, &results)
-	if err != nil {
-		return nil, 0, err
+
+	type TasksAndCount struct {
+		Tasks []Task           `bson:"tasks"`
+		Count []map[string]int `bson:"count"`
+	}
+	TaskAndCountResults := []TasksAndCount{}
+	TaskResults := []Task{}
```
ablack12, line 3037: Can just define this in these variables  in the if/else statements they're used in

model/task/task.go
```diff
@@ -2969,80 +2969,102 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
...
 	}
-	err = cursor.All(ctx, &results)
-	if err != nil {
-		return nil, 0, err
+
+	type TasksAndCount struct {
+		Tasks []Task           `bson:"tasks"`
+		Count []map[string]int `bson:"count"`
+	}
+	TaskAndCountResults := []TasksAndCount{}
```
ablack12, line 3036: We really only use uppercase variables for constants

ablack12 (changes requested):

model/task/task.go
```diff
@@ -3029,37 +3029,35 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
...
-			if len(TaskAndCountResult.Count) != 0 {
-				count = TaskAndCountResult.Count[0]["count"]
-			}
-			results = TaskAndCountResult.Tasks
+		if len(taskAndCountResults) > 0 && len(taskAndCountResults[0].Count) > 0 {
+			count = taskAndCountResults[0].Count[0]["count"]
+			results = taskAndCountResults[0].Tasks
 		}
 	} else {
+		TaskResults := []Task{}
```
ablack12, line 3052: lowercase

ablack12 (approved):
LGTM with nit

PR #5171, opened 2021-11-08
EVG-14385 Add Patch to Task resolver

minnakt:
[EVG-14385](https://jira.mongodb.org/browse/EVG-14385)

### Description 
This PR modifies `resolvers.go` so that it is possible to retrieve the corresponding `Patch` associated with a task.

~~This change was made because a `Patch` contains the `commitQueuePosition` field which can be used to determine if a task is part of a patch that is on the commit queue. We can then use this field to determine whether or not to disable the `Schedule` and `Restart` buttons on the `/task/:id` pages.~~ 

The `requester` field was used to accomplish the functionality described in the paragraph above. However, these changes are still useful as there might be a situation in the future where we need to get the `Patch` associated with a task.

### Testing 
- Added `patch.graphql` in `graphql/tests/task/queries`


graphql/resolvers.go
```diff
@@ -3211,6 +3211,14 @@ func (r *taskResolver) VersionMetadata(ctx context.Context, obj *restModel.APITa
 	return apiVersion, nil
 }
 
+func (r *taskResolver) Patch(ctx context.Context, obj *restModel.APITask) (*restModel.APIPatch, error) {
+	apiPatch, err := r.sc.FindPatchById(*obj.Version)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Couldn't find a patch with id: `%s` %s", *obj.Version, err.Error()))
```
khelif96, line 3217: We should do a check similar to https://github.com/evergreen-ci/evergreen/blob/41b7592f43d24458bb51170afecc2c5218a1a5f8/graphql/resolvers.go#L3600-L3610
to determine if the task has a patch before performing this query. 

Since some tasks might not have a patch we should avoid throwing an error since that is a valid scenario.

khelif96 (changes requested):
@ablack12  pointed out you can use the requester field that exists on the Task to determine if it is a commit queue task.
https://github.com/evergreen-ci/evergreen/blob/109f25380471ab7752337f246f3f700875c72d30/graphql/schema.graphql#L998

These changes are still good and we should keep them. 👍

graphql/resolvers.go
malikchaya2: Why do we still need the whole patch if we can disable the Schedule and Restart buttons on the /task/:id pages based on the requester?

khelif96: Just a matter of future proofing things. :) Its not necessary for this feature. But it might be useful in the future.

khelif96 (approved):

bynn (approved):

graphql/resolvers.go
malikchaya2: sounds good. Let's make sure to not, in the future, accidentally call it on all tasks because then it can end up being expensive :)

malikchaya2 (approved):

PR #5172, opened 2021-11-08
EVG-15509 make throughput configurable for volumes

malikchaya2:
[EVG-15509](https://jira.mongodb.org/browse/EVG-15509)

### Description 
Make throughput configurable 

### Testing 
Added to existing tests


cloud/ec2_util.go
```diff
@@ -408,6 +409,18 @@ func makeBlockDeviceMappings(mounts []MountPoint) ([]*ec2aws.BlockDeviceMapping,
 			if mount.Iops != 0 {
 				m.Ebs.Iops = aws.Int64(mount.Iops)
 			}
+
+			if mount.Throughput != 0 {
+				//aws only allows values between 125 and 1000
+				if mount.Throughput > 1000 || mount.Throughput < 125 {
```
ablack12, line 415: nit: 1000 (no ,)

cloud/ec2_util.go
```diff
@@ -408,6 +409,18 @@ func makeBlockDeviceMappings(mounts []MountPoint) ([]*ec2aws.BlockDeviceMapping,
...
+					return nil, errors.New("throughput must be between 125 and 1,000")
+				}
+				// This parameter is valid only for gp3 volumes.
+				if *m.Ebs.VolumeType != ec2aws.VolumeTypeGp3 {
```
ablack12, line 419: It looks to me like we set this to Gp2 at 407 and don't actually override the default until 427, so this won't ever get hit

ablack12: Also, should use utility.ToStringPtr

ablack12 (changes requested):
We should add throughput to a staging distro, create a host, and verify in AWS that throughput is what we expect (if we know how to do that, or else at least we know host creation is successful with throughput in the mapping so)

malikchaya2:
> We should add throughput to a staging distro, create a host, and verify in AWS that throughput is what we expect (if we know how to do that, or else at least we know host creation is successful with throughput in the mapping so)

I'll try in staging tomorrow and see how it goes.

cloud/ec2_util.go
malikchaya2: I don't think there is a (,), you may have seen an older version of my changes maybe?

cloud/ec2_util.go
malikchaya2: haha yes. Thanks for spotting that! 
Do you mean FromStringPtr?

malikchaya2: If yes, then done.

cloud/ec2_util.go
ablack12: sorry this is on the wrong line; I meant for line 421

malikchaya2: done

cloud/ec2_util.go
```diff
@@ -414,6 +415,17 @@ func makeBlockDeviceMappings(mounts []MountPoint) ([]*ec2aws.BlockDeviceMapping,
...
+			if mount.Throughput != 0 {
+				//aws only allows values between 125 and 1000
+				if mount.Throughput > 1000 || mount.Throughput < 125 {
+					return nil, errors.New("throughput must be between 125 and 1000")
+				}
+				// This parameter is valid only for gp3 volumes.
+				if utility.FromStringPtr(m.Ebs.VolumeType) != ec2aws.VolumeTypeGp3 {
+					return nil, errors.New("throughput is only valid for gp3 volumes")
+				}
+				m.Ebs.Throughput = aws.Int64(mount.Throughput)
```
ablack12, line 427: Can we add this validation and the 125 - 1000 check to the distro page? otherwise it's easy to save values that will break the distro https://github.com/evergreen-ci/evergreen/blob/90cb3e180815e3b4b54bde99bc2e008ddc647a41/validator/distro_validator.go#L24

malikchaya2: I don't think I can add it here because the distro doesn't store the ebs device information, and there doesn't seem to be a volume validator.

ablack12: It is part of distro; it's part of the EC2 provider settings. Go isn't great at interfaces so instead we stored this as a [list of birch documents](https://github.com/evergreen-ci/evergreen/blob/cf637f34e4a5c031cb3638046a252c76d02f19b9/model/distro/distro.go#L30), and unmarshal [based on provider](https://github.com/evergreen-ci/evergreen/blob/4a7a32041d8bd02f4784321820eb2655f7476df2/cloud/cloud.go#L238).  It looks like we do validate the EC2 provider settings. 
That being said, it looks like we do already call makeBlockDeviceMappings when we validate the EC2 settings. https://github.com/evergreen-ci/evergreen/blob/877d8c75a4cdc40e12f5faa4f6d3ceca617fbadd/cloud/ec2.go#L107 
which is called in [this validator](https://github.com/evergreen-ci/evergreen/blob/90cb3e180815e3b4b54bde99bc2e008ddc647a41/validator/distro_validator.go#L136) so we should be good. Can you verify on local if this prevents you from saving the page?

PR #5173, opened 2021-11-08
EVG-14727 Add a way for tasks to be disabled

bynn:
[EVG-14727](https://jira.mongodb.org/browse/EVG-14727)

### Description 
Tasks can now take the `disable` param 

### Testing 
on staging 
image

patch on staging with a disabled dependency 
https://spruce-staging.corp.mongodb.com/version/618d79cf9dbe324ea5829fee/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

for non patches 
https://evergreen-staging.corp.mongodb.com/version/sandbox_912ef97949fbabdc161230303776b0543f798a92

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -374,7 +374,6 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig, patchPro
 				}
 			}
 		}
-		grip.Info(tasksForVariant)
 		if len(tasksForVariant) > 0 {
 			variant.Tasks = tasksForVariant
 			variantMappings[variant.Name] = variant
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -365,7 +365,7 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
 	for _, variant := range project.BuildVariants {
 		tasksForVariant := []model.BuildVariantTaskUnit{}
 		for _, taskFromVariant := range variant.Tasks {
-			if taskFromVariant.IsDisabled() && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
+			if !taskFromVariant.IsDisabled() && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
 				if taskFromVariant.IsGroup {
 					tasksForVariant = append(tasksForVariant, model.CreateTasksFromGroup(taskFromVariant, project)...)
 				} else {
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -365,7 +365,7 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
 	for _, variant := range project.BuildVariants {
 		tasksForVariant := []model.BuildVariantTaskUnit{}
 		for _, taskFromVariant := range variant.Tasks {
-			if !utility.FromBoolPtr(taskFromVariant.Disable) && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
+			if taskFromVariant.IsDisabled() && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
 				if taskFromVariant.IsGroup {
 					tasksForVariant = append(tasksForVariant, model.CreateTasksFromGroup(taskFromVariant, project)...)
 				} else {
@@ -373,7 +373,7 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
 				}
 			}
 		}
-		if len(tasksForVariant) != 0 {
+		if len(tasksForVariant) > 0 {
 			variant.Tasks = tasksForVariant
 			variantMappings[variant.Name] = variant
 		}
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -365,7 +365,12 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig, patchPro
 	for _, variant := range project.BuildVariants {
 		tasksForVariant := []model.BuildVariantTaskUnit{}
 		for _, taskFromVariant := range variant.Tasks {
+			grip.Info(message.Fields{
+				"bynnbynn": taskFromVariant,
+				"disable":  taskFromVariant.IsDisabled(),
+			})
 			if !taskFromVariant.IsDisabled() && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
+				grip.Info("here")
 				if taskFromVariant.IsGroup {
 					tasksForVariant = append(tasksForVariant, model.CreateTasksFromGroup(taskFromVariant, project)...)
 				} else {
@@ -373,6 +378,7 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig, patchPro
 				}
 			}
 		}
+		grip.Info(tasksForVariant)
 		if len(tasksForVariant) > 0 {
 			variant.Tasks = tasksForVariant
 			variantMappings[variant.Name] = variant
@@ -381,6 +387,10 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig, patchPro
 
 	tasksList := []struct{ Name string }{}
 	for _, task := range project.Tasks {
+		grip.Info(message.Fields{
+			"bynnbynn": task,
+			"disable":  utility.FromBoolPtr(task.Disable),
+		})
 		// add a task name to the list if it's patchable and not restricted to git tags and not disabled
 		if !utility.FromBoolPtr(task.Disable) && utility.FromBoolTPtr(task.Patchable) && !utility.FromBoolPtr(task.GitTagOnly) {
 			tasksList = append(tasksList, struct{ Name string }{task.Name})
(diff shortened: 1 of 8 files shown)


bynn:
doesn't seem to work with dist on self test for some reason not sure why

model/dependencies.go
```diff
@@ -80,6 +81,12 @@ func (di *dependencyIncluder) handle(pair TVPair) (bool, error) {
 		di.included[pair] = false
 		return false, errors.Errorf("task '%s' in variant '%s' cannot be run for a '%s'", pair.TaskName, pair.Variant, di.requester)
 	}
+
+	if utility.FromBoolPtr(bvt.Disable) {
+		di.included[pair] = false
+		return false, errors.Errorf("task '%s' in variant '%s' has been disabled", pair.TaskName, pair.Variant)
```
ablack12, line 87: did you test with dependencies? I'm not sure that we want this to be an error, we just don't want to include it. (We want tasks that depend on a disabled task to be scheduled, just without that dependency)

model/dependencies.go
```diff
@@ -115,7 +122,7 @@ func (di *dependencyIncluder) expandDependencies(pair TVPair, depends []TaskUnit
 					}
 					projectTask := di.Project.FindTaskForVariant(t.Name, v.Name)
 					if projectTask != nil {
-						if projectTask.SkipOnRequester(di.requester) {
+						if utility.FromBoolPtr(projectTask.Disable) || projectTask.SkipOnRequester(di.requester) {
```
ablack12, line 125: Can we make a function for projectTasks IsDisabled?

self-tests.yml
```diff
@@ -576,6 +598,7 @@ tasks:
     depends_on:
       - name: dist
         status: success
+    disable: true
```
ablack12, line 601: I know this PR is WIP but reminder we can't commit this

validator/project_validator.go
```diff
@@ -1213,6 +1213,12 @@ func validateTaskDependencies(project *model.Project) ValidationErrors {
 					Message: fmt.Sprintf("Task '%s' depends on non-patchable task '%s'. Neither will run in patches", task.Name, dep.Name),
 				})
 			}
+			if utility.FromBoolPtr(dependent.Disable) && !utility.FromBoolPtr(task.Disable) {
+				errs = append(errs, ValidationError{
+					Level:   Warning,
+					Message: fmt.Sprintf("Task '%s' depends on disabled task '%s'. Neither will run", task.Name, dep.Name),
```
ablack12, line 1219: "Neither will run" isn't actually the behavior we want.

graphql/util.go
```diff
@@ -363,22 +363,24 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
...
 				if taskFromVariant.IsGroup {
 					tasksForVariant = append(tasksForVariant, model.CreateTasksFromGroup(taskFromVariant, project)...)
 				} else {
 					tasksForVariant = append(tasksForVariant, taskFromVariant)
 				}
 			}
 		}
-		variant.Tasks = tasksForVariant
-		variantMappings[variant.Name] = variant
+		if len(tasksForVariant) != 0 {
```
ablack12, line 374: can you say > 0 ? just easier to read quickly really

ablack12 (changes requested):
DMed about the Disable issue, going to test some more and do further review after that's figured out

model/dependencies.go
bynn: done

self-tests.yml
bynn: oops ill be sure to delete this

validator/project_validator.go
bynn: hmm im not sure what should happen in that case then

ablack12: We can zoom if you want about this. If taskA depends on taskB, I think it should treat taskB as though it doesn't exist entirely; users want to use "disable" as a way to essentially "comment out" a task without having to make tons of small updates wherever it's referenced (like dependencies).

I think this can be handled here: https://github.com/evergreen-ci/evergreen/blob/b41b8975e9f4bd0afc0251c0fa83e1bb377e4f95/model/dependencies.go#L179 we should check if it's disabled before adding it.

ablack12 (commented):
Here are some of the cases I think we need to consider, and we can probably write unit tests to help handle these:
1) if taskA is disabled, shouldn't be able to patch it (you've tested this)
2) if taskA is disabled, it shouldn't show up in a waterfall commit
3) if taskA is disabled, and taskB depends on taskA, can we still schedule taskB?
4) if taskA is disabled and has tag "test" and nothing else has tag "test", and variantA uses tag "test", does variantA still show up? Does this give a validation warning, or a validation error?

graphql/util.go
```diff
@@ -363,22 +363,24 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
...
+			variantMappings[variant.Name] = variant
+		}
 	}
 
 	tasksList := []struct{ Name string }{}
 	for _, task := range project.Tasks {
-		// add a task name to the list if it's patchable and not restricted to git tags
-		if utility.FromBoolTPtr(task.Patchable) && !utility.FromBoolPtr(task.GitTagOnly) {
+		// add a task name to the list if it's patchable and not restricted to git tags and not disabled
+		if !utility.FromBoolPtr(task.Disable) && utility.FromBoolTPtr(task.Patchable) && !utility.FromBoolPtr(task.GitTagOnly) {
```
ablack12, line 383: !IsDisabled

validator/project_validator.go
```diff
@@ -1213,6 +1213,12 @@ func validateTaskDependencies(project *model.Project) ValidationErrors {
 					Message: fmt.Sprintf("Task '%s' depends on non-patchable task '%s'. Neither will run in patches", task.Name, dep.Name),
 				})
 			}
+			if utility.FromBoolPtr(dependent.Disable) && !utility.FromBoolPtr(task.Disable) {
```
ablack12, line 1216: IsDisabled

model/dependencies.go
ablack12: Yeah this might be okay, we'll just need to test this.

model/dependencies.go
```diff
@@ -175,7 +181,13 @@ func (di *dependencyIncluder) expandDependencies(pair TVPair, depends []TaskUnit
...
 				v = pair.Variant
 			}
-			deps = append(deps, TVPair{TaskName: d.Name, Variant: v})
+			projectTask := di.Project.FindTaskForVariant(d.Name, v)
+			if projectTask != nil {
+				if projectTask.IsDisabled() {
+					continue
+				}
+				deps = append(deps, TVPair{TaskName: d.Name, Variant: v})
+			}
```
hadjri, line 190: I think you can just put "if projectTask != nil && !projectTask.IsDisabled()" for readability.

hadjri (changes requested):

model/dependencies.go
bynn: done

graphql/util.go
```diff
@@ -365,22 +365,34 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
 	for _, variant := range project.BuildVariants {
 		tasksForVariant := []model.BuildVariantTaskUnit{}
 		for _, taskFromVariant := range variant.Tasks {
-			if utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
+			grip.Info(message.Fields{
```
ablack12, line 368: remove logging :)

graphql/util.go
```diff
@@ -365,22 +365,34 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
...
 
 	tasksList := []struct{ Name string }{}
 	for _, task := range project.Tasks {
-		// add a task name to the list if it's patchable and not restricted to git tags
-		if utility.FromBoolTPtr(task.Patchable) && !utility.FromBoolPtr(task.GitTagOnly) {
+		grip.Info(message.Fields{
+			"bynnbynn": task,
+			"disable":  utility.FromBoolPtr(task.Disable),
+		})
+		// add a task name to the list if it's patchable and not restricted to git tags and not disabled
```
ablack12, line 394: can we also put this comment above?

model/lifecycle.go
```diff
@@ -756,6 +756,7 @@ func CreateTasksFromGroup(in BuildVariantTaskUnit, proj *Project) []BuildVariant
 			GroupName:        in.Name,
 			Patchable:        in.Patchable,
 			PatchOnly:        in.PatchOnly,
+			Disable:          in.Disable,
```
ablack12, line 759: Did you create a new ticket for fixing CreateTasksFromGroup?

ablack12 (commented):
Do you have an example of testing dependencies on staging?

graphql/util.go
bynn: done

graphql/util.go
bynn: done

model/lifecycle.go
bynn: yes! 
https://jira.mongodb.org/browse/EVG-15837

graphql/util.go
```diff
@@ -365,22 +365,25 @@ func GetVariantsAndTasksFromProject(ctx context.Context, patchedConfig string, p
 	for _, variant := range project.BuildVariants {
 		tasksForVariant := []model.BuildVariantTaskUnit{}
 		for _, taskFromVariant := range variant.Tasks {
-			if utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
+			// add a task name to the list if it's patchable and not restricted to git tags and not disabled
+			if !taskFromVariant.IsDisabled() && utility.FromBoolTPtr(taskFromVariant.Patchable) && !utility.FromBoolPtr(taskFromVariant.GitTagOnly) {
```
ablack12, line 369: I think we can summarize the latter half by saying SkipOnRequester(PatchRequester)

ablack12 (approved):
🚀 
you'll want to add to the documentation here https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#commit-only-patch-only-and-git-tag-only-tasks and maybe add some more detail about how it works for dependencies and etc, and link it to the ticket before closing. (May also want to change the section header to something more general)

graphql/util.go
bynn: done

PR #5175, opened 2021-11-09
EVG-15625, EVG-15626, EVG-15627: use Go modules

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-15625
https://jira.mongodb.org/browse/EVG-15626
https://jira.mongodb.org/browse/EVG-15627

### Description

This is a follow-on to #5164 . I merged and reverted and applied some small fixes. The diff is no longer gigantic compared to the base commit, so Evergreen can test it, but the GitHub diff on this PR is also not correct since it's based off of the now-reverted commit, so you'll have to diff it locally.

I can't merge main since it would produce the same gigantic-diff problem as before, so I would prefer this to be reviewed quickly if possible and I'll deal with all the merge conflicts once it's approved.

* Replace vendor directory and glide with Go modules. I removed all the revendoring tools because they will not apply in Go modules and people will have to know how to use Go modules tooling rather than just blindly rely on a single make target to do what they need. I'm going to document it in [EVG-15450](https://jira.mongodb.org/browse/EVG-15450), but I don't think it matters much except for PLT since others don't mess with dependency versions that often.
* Remove GOPATH dependencies in our dev/build process.
* Upgrade all evergreen-ci/monogdb repos to latest version and fix up backward incompatibilities introduced because Evergreen is using stale copies of some repos.
* Fix `github.com/sabhiram/go-git-ignore`, which is now `github.com/sabhiram/go-gitignore`.
* Remove `evg-lint` linter until [EVG-15453](https://jira.mongodb.org/browse/EVG-15453). This is only temporary and it only catches very specific things (e.g. testify `TearDownSuite` capitalization and `defer`s in for loops).
* Move mgo hacks into a subpackage of Evergreen. We're using [a special branch of mgo](https://github.com/go-mgo/mgo/tree/60f23985327354fe2d4dde32ba05598bb047a114) and the vendored copy of mgo has been manually modified to support the mongo-driver (and prevent our DB tests from failing), so we have no choice but to move it into Evergreen since it's not really an official version of mgo anymore.
    * I read over the license and it seems to allow this, as long as we keep the license with the code (I'm not a license lawyer though).
* Fix graphql `gqlgen` target and GraphQL tooling to not depend on vendoring the gqlgen tool. I don't really know if it auto-generated the proper `graphql/generated.go` file properly, but the version we were pinned to in glide (v0.11.2) is not compatible with Go modules. However, upgrading gqlgen is risky since it has no stable releases, so every new version has some risk of backward incompatibility. We've upgraded from v0.11.2 to v0.14.0 and it seems like it generated a syntactically-valid `graphql/generated.go`, but I'm not really qualified to say if it's correct or not.
    * Because `gqlgen` is a CLI utility that we need to run for auto-generation, I followed [the recommended way to track Go tool dependencies in a repo](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) and made a tools file for it (`cmd/gqlgen/gqlgen.go`).
* Remove codegen hacks in the makefile, which relied on `go get`'ing `goimports` into the `GOPATH`. It now just relies on it like a regular module.

### Testing 
I ran all the local non-integration tests and the linter. I also deployed it to staging and verified that I could run a task.

I ran `gqlgen` for the auto-generated graphql and `generate-rest-model` for the auto-generated REST routes to verify that they at least seem to do something sensible.

I ran `generate-lint` to see if it produced a reasonable-looking generate.tasks JSON flie.


julianedwards (approved):

khelif96:
The CI run had a bunch of setup failures Is this ready for a review or are we still waiting on some changes? https://spruce.mongodb.com/version/evergreen_6176faef5c24b2fe5f9ffd5c10116f517fee4e2a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC 

It lgtm as long as the tests still run and pass.

Kimchelly:
Yeah, my plan was to test the commit by merging it, then reduce the diff size for this PR since there's a base commit that includes the module changes. Since the tests pass in this PR, it means that the committed change + the fixups in this PR are sufficient for the tests to be green.

The merge conflicts are only because I reverted the original base commit, so I have to revert the revert, then apply this small diff on top of it.

bsamek (approved):
LGTM.

@khelif96 Could you review the graphql/ changes? There are some new things, like a new way to do contexts, that I don't understand, so am not sure how to review.

@Kimchelly If you haven't yet, could you validate that the dist works in staging? One approach could be testing that something could be saved using the copied mgo code and then read back after reverting the staging deploy. Given that we haven't tried this before, doing some kind of smoke test in staging could increase our confidence. An alternative would be to let it burn in in staging for a while, asking people to rebase on it.

Kimchelly:
> If you haven't yet, could you validate that the dist works in staging?
I did a test deploy with a patch and it was fine, and it sat for a while before someone deployed a different dist. I also checked that the patches page seemed to work properly (the patch model uses the legacy mgo ObjectId). Other than that, I believe that the DB tests would be sufficient to cover that the mgo copy-paste worked. They were failing when I tried to use the official version of mgo.

If it gives any more confidence, Cedar has already been running with Go modules enabled for multiple days without any issue.

bsamek:
> > If you haven't yet, could you validate that the dist works in staging?
> 
> I did a test deploy with a patch and it was fine, and it sat for a while before someone deployed a different dist. I also checked that the patches page seemed to work properly (the patch model uses the legacy mgo ObjectId). Other than that, I believe that the DB tests would be sufficient to cover that the mgo copy-paste worked. They were failing when I tried to use the official version of mgo.
> 
> If it gives any more confidence, Cedar has already been running with Go modules enabled for multiple days without any issue.

Sounds good, thanks!

khelif96:
I get the following error when loading the project in vscode. I know most of you use vim but do you know why this error is occuring?
image

Kimchelly:
If you type `go env GO111MODULE` it should say `"auto"` or `"on"` instead of `"off"` since most people have explicitly disabled it for Evergreen. Setting the env var should fix it.

khelif96 (approved):
LGTM the graphql changes work 

We should include some information on how to migrate over once this is merged.
VSCODE users should set
 `GO111MODULE=on` 
as well as add the following flags to their vscode settings.json file
```
    "go.toolsEnvVars": {
      "GO111MODULE": "on"
    }
```
to enable go modules for the vscode go plugin
and 
```
    "go.testEnvVars": {
      "EVGHOME": "${workspaceFolder}",
      "SETTINGS_OVERRIDE": "/Users/mohamed.khelif/test_config.yml",
      "GOMODCACHE": "/Users/mohamed.khelif/go/pkg/mod"
    },
```
to support the vscode go debugger

PR #5176, opened 2021-11-09
EVG-15818: evergreen project should use repo settings

sophstad:
[EVG-15818](https://jira.mongodb.org/browse/EVG-15818)

### Description 
- Set `use_repo_settings` to `true` for the evergreen project so that we have one local project using this setting for testing purposes. 

### Testing 
- Ran locally


malikchaya2 (commented):
good idea! 
Do we need more here? do we need to create an actual repoRef?

sophstad:
> do we need to create an actual repoRef?

@malikchaya2 Done in [`66177de`](https://github.com/evergreen-ci/evergreen/commit/66177de8f1551281d7371aa9ad7fd81e04ef4fe3#diff-c419c41c60f5cb06069eff0b933375b7115f7b98c40ec3e267f12909385fa04c)!

I think this should be adequate for now. The existing projects and repo have a good mix of explicit/null settings so this has been sufficient for testing so far.

malikchaya2 (approved):
perfect, thank you for doing this!

PR #5179, opened 2021-11-10
EVG-14648 Update s3copy to log destination path

bynn:
[EVG-14648](https://jira.mongodb.org/browse/EVG-14648)

### Description 
add logging to s3 copy 

### Testing 
will test on staging when later


agent/command/s3_copy.go
```diff
@@ -230,6 +230,8 @@ func (c *s3copy) s3Copy(ctx context.Context,
 			logger.Task().Warning("destination bucket names containing dots that are created after Sept. 30, 2020 are not guaranteed to have valid attached URLs")
 			foundDottedBucketName = true
 		}
+
+		logger.Task().Infof("successfully copied file to '%s'", s3CopyFile.Destination.Path)
```
ablack12, line 234: The ticket description implies we also want to log the source file

bynn: done

ablack12 (approved):

ablack12:
no need to test in staging i think for such a small change

PR #5180, opened 2021-11-10
EVG-15511 add route to trigger user-defined task status

ablack12:
[EVG-15511](https://jira.mongodb.org/browse/EVG-15511)

### Description 
Allow users to hit this route as a way to define their own task status.

### Testing 
Verified the status was set here, and verified here that, with should continue set, the status was still set to setup failed (but after the next command ran).


agent/command.go
```diff
@@ -147,6 +147,9 @@ func (a *Agent) runCommandSet(ctx context.Context, tc *taskContext, commandInfo
 			return errors.Wrap(ctx.Err(), "Agent stopped early")
 		}
 		tc.logger.Task().Infof("Finished %s in %s", fullCommandName, time.Since(start).String())
+		if a.endTaskResp != nil {
```
bsamek, line 150: Would it make sense to reuse the context cancellation machinery? Above on 141 we check for a cancelled context for timeouts. Since we know timeouts work correctly, it might make sense to do something similar for the task status case.

One question that the ticket doesn't answer though is whether the task should cease execution immediately if a status is set. We may wish to clarify that with the users first. I can imagine reasons that it should cease immediately (we know the status, so we don't need to keep going) and that it shouldn't (there's some addition error handling stuff the task wants to do).

agent/agent.go
```diff
@@ -473,6 +474,14 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 }
 
 func (a *Agent) handleTaskResponse(ctx context.Context, tc *taskContext, status string, message string) (bool, error) {
+	if a.endTaskResp != nil { // if the user indicated a task response, use this instead
+		if !evergreen.IsFinishedTaskStatus(a.endTaskResp.Status) {
+			tc.logger.Task().Errorf("'%s' is not a valid task status", a.endTaskResp.Status)
+		} else {
+			status = a.endTaskResp.Status
+			message = a.endTaskResp.Message
```
bsamek, line 482: If this message ends up in the UI, validate its length, maybe.

agent/status.go
```diff
@@ -105,6 +107,42 @@ func (a *Agent) statusHandler() http.HandlerFunc {
...
+	Status  string `json:"status"`
+	Message string `json:"message"`
+}
+
+func (a *Agent) endTaskHandler(w http.ResponseWriter, r *http.Request) {
+	defer func() {
+		_ = grip.GetSender().Close()
+	}()
+
+	grip.Info(message.Fields{
```
bsamek, line 120: This message ends up in a local file on the agent, I think, so it probably isn't useful to log errors like this. Instead write the error message in the response to the HTTP call so that the task can log what it gets from the call to /task_status.

agent/command.go
ablack12: I think that it makes sense for the task to cease execution immediately unless it's a pre/post task; we want to simulate how tasks affect statuses. I added logic to ensure this behavior and added a should_continue option if the user does want to override that logic.

ablack12:
Going to do an updated staging testing without all my logging but this is ready for review

ablack12:
Verified the status was set [here](https://evergreen-staging.corp.mongodb.com/task/evg_ubuntu1604_test_set_status_patch_cccc0c06f1c06be02cc2903cf89c2dc07a665af5_6192b46897b1d3076fbe5ada_21_11_15_19_26_43##comparehashes=cccc0c06f1c06be02cc2903cf89c2dc07a665af5&threads=maxonly), and verified [here](https://evergreen-staging.corp.mongodb.com/task/evg_ubuntu1604_test_set_status_patch_758c4464174ac9413e31fbe071e7740a72ad2e54_6192be8397b1d30e196a87bb_21_11_15_20_09_43##comparehashes=758c4464174ac9413e31fbe071e7740a72ad2e54&threads=maxonly) that, with should continue set, the status was still set to setup failed (but after the next command ran).

agent/agent.go
```diff
@@ -603,14 +605,48 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 
 func (a *Agent) endTaskResponse(tc *taskContext, status string, message string) *apimodels.TaskEndDetail {
 	var description string
-	var cmdType string
+	var failureType string
+	if a.endTaskResp != nil { // if the user indicated a task response, use this instead
+		tc.logger.Task().Infof("Task status set using /task_status")
```
bsamek, line 610: nit: Use a constant for `task_status` and the endpoint definition or call it something more generic.

agent/agent.go
```diff
@@ -603,14 +605,48 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
...
+		if !evergreen.IsValidTaskEndStatus(a.endTaskResp.Status) {
+			tc.logger.Task().Errorf("'%s' is not a valid task status", a.endTaskResp.Status)
+			status = evergreen.TaskFailed
+			failureType = evergreen.CommandTypeSystem
+			tc.logger.Task().Debugf("END TASK RESP: %v", a.endTaskResp)
```
bsamek, line 615: This is a pretty big struct to log, and I'm not sure it's useful. I think logging `a.endTaskResp.Status` on 612 is sufficient.

agent/agent.go
```diff
@@ -603,14 +605,48 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
...
+		} else {
+			status = a.endTaskResp.Status
+			if len(a.endTaskResp.Description) > MessageLimit {
+				tc.logger.Task().Warningf("description from task_status is too long to set (%s character limit), defaulting to command display name", MessageLimit)
```
bsamek, line 619: nit: %d for number in printf

agent/agent.go
```diff
@@ -603,14 +605,48 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
...
+	if a.endTaskResp != nil { // if the user indicated a task response, use this instead
+		tc.logger.Task().Infof("Task status set using /task_status")
+		if !evergreen.IsValidTaskEndStatus(a.endTaskResp.Status) {
+			tc.logger.Task().Errorf("'%s' is not a valid task status", a.endTaskResp.Status)
+			status = evergreen.TaskFailed
+			failureType = evergreen.CommandTypeSystem
+			tc.logger.Task().Debugf("END TASK RESP: %v", a.endTaskResp)
+		} else {
+			status = a.endTaskResp.Status
+			if len(a.endTaskResp.Description) > MessageLimit {
```
bsamek, line 618: How do we intend users to use `Description` versus `Message`?

agent/command.go
```diff
@@ -147,6 +147,11 @@ func (a *Agent) runCommandSet(ctx context.Context, tc *taskContext, commandInfo
 			return errors.Wrap(ctx.Err(), "Agent stopped early")
 		}
 		tc.logger.Task().Infof("Finished %s in %s", fullCommandName, time.Since(start).String())
+		if (options.isTaskCommands || options.failPreAndPost) && a.endTaskResp != nil && !a.endTaskResp.ShouldContinue {
+			// only error if we're running a command that should fail, and we don't want to continue to run other tasks
+			tc.logger.Task().Debug("task status has been set; triggering end task")
```
bsamek, line 152: Although it would be a duplicate, this log line should log `endTaskResp.Status` I think, in case we somehow don't get to the validation log line. This allows a user to see earlier what status they've set and would aid in debugging.

agent/status.go
```diff
@@ -105,6 +108,35 @@ func (a *Agent) statusHandler() http.HandlerFunc {
 	}
 }
 
+type TriggerEndTaskResp struct {
+	apimodels.TaskEndDetail `json:",omitempty"`
```
bsamek, line 112: This is a pretty big struct, and contains a bunch of things we don't need. I would suggest letting the user set only the fields we're interseted in and then converting to another struct if you need to.

agent/status.go
```diff
@@ -105,6 +108,35 @@ func (a *Agent) statusHandler() http.HandlerFunc {
...
+	payload, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		_, _ = w.Write([]byte(err.Error()))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	resp := TriggerEndTaskResp{}
+	if err := json.Unmarshal(payload, &resp); err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
```
bsamek, line 130: Should these be `w.Write` instead so that the caller sees it?

globals.go
```diff
@@ -302,6 +302,10 @@ func IsFailedTaskStatus(status string) bool {
 	return utility.StringSliceContains(TaskFailureStatuses, status)
 }
 
+func IsValidTaskEndStatus(status string) bool {
+	return status == TaskSucceeded || status == TaskFailed || status == TaskUndispatched
```
bsamek, line 306: I don't think we should let someone set `TaskUndispatched`.

agent/agent.go
ablack12: Sorry this was leftover debugging for myself

agent/agent.go
ablack12: Honestly good question. It looks like the only time we ever set message is [here](https://github.com/evergreen-ci/evergreen/blob/67d01e997b9c444325723bb4fc6a7e541eb0a785/agent/agent.go#L443) and I'm not sure if we display it in the UI or not. We could instead just disallow setting this and not provide in the documentation to users.

agent/status.go
ablack12: Yeah I was torn between just using this struct or putting the things we needed directly into TriggerEndTaskResp. I thought that using TaskEndDetail would result in less refactoring in the future. Right now we're only allowing description, status, and failure type though.

agent/status.go
ablack12: I wasn't actually aware this would reach the caller. Fixed but not sure if plain text is the best way to do it.

bsamek (approved):

PR #5181, opened 2021-11-10
EVG-15042 add attach to new repo resolver

ablack12:
[EVG-15042](https://jira.mongodb.org/browse/EVG-15042)

### Description 
Add attach to new repo resolver (called MoveRepo in the ticket).
I added a comment to this affect in the design, but I think this resolver/button should only be used if useRepoSettings is true; otherwise this really makes more sense to be edited with the general project section, since it doesn't require any special handling. Modified that function accordingly.

### Testing 
Added tests.


graphql/schema.graphql
```diff
@@ -369,6 +370,12 @@ input CopyProjectInput {
   newProjectId: String
 }
 
+input MoveProjectInput {
+  id: String!
```
malikchaya2, line 374: this could theoretically be either a projectId or a repoId, so can we please label it more specifically?

graphql/schema.graphql
```diff
@@ -369,6 +370,12 @@ input CopyProjectInput {
...
+  owner: String!
+  repo: String!
```
malikchaya2, line 376: can we maybe label these newOwner and newRepo so that it's clear that these are the values of the repo we will be moving it to?

model/project_ref.go
```diff
@@ -1525,24 +1524,30 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+			projectRefPatchingDisabledKey:        p.PatchingDisabled,
+			projectRefTaskSyncKey:                p.TaskSync,
+			ProjectRefDisabledStatsCacheKey:      p.DisabledStatsCache,
+			ProjectRefFilesIgnoredFromCacheKey:   p.FilesIgnoredFromCache,
+		}
+		if !isRepo && !p.UseRepoSettings {
+			setUpdate[ProjectRefOwnerKey] = p.Owner
+			setUpdate[ProjectRefRepoKey] = p.Repo
+			setUpdate[ProjectRefRepoRefIdKey] = p.RepoRefId // just in case this is outdated somehow
+		}
```
malikchaya2, line 1546: I don't fully understand why this is in this section? Is this for the case where useRepoSettings is false, aka it's not currently attached to a repo and the user want's to attach it for the first time? Shouldn't they be using the attach button in that case?

ablack12: No, it's for the case where it's unattached and they want to change their owner/repo. I think this should just work like a regular project page since owner/repo doesn't have any special meaning unless you're using repo. I'm only resetting the RepoRefId in case of programmer error (like if we were to toggle UseRepoSettings from true to false but not unset RepoRefId)

malikchaya2: In the current design and setup in this [pr](https://github.com/evergreen-ci/spruce/pull/977), if a project is not attached, we don't show them the move repo button, and instead show them the attach to repo button. Do we want to change that?

sophstad: In that case the move repo button isn't shown, but the owner and repo fields are present and so a user can attach to a new repo in that way. This pattern is based on Annie's Figma comment [here](https://www.figma.com/file/Z0rfsPjVKglRq9baDhsSux?node-id=546:374#123741858).

ablack12: To further clarify:
If a user is unattached, we show only the Attach to Repo button. They should be able to use the owner/repo boxes as normal text boxes and change them using the Save Changes on Page button. This will not set UseRepoSettings to true (they will remain unattached); they can only attach by clicking Attach to Repo.

If a user is attached, we show both the Detach to Repo button and the Move Repo button, and they can only edit owner/repo by clicking Move Repo, since this has greater consequences than just changing two textbox fields.

malikchaya2: Ah, I understand now. Thank you for explaining that.

graphql/schema.graphql
```diff
@@ -369,6 +370,12 @@ input CopyProjectInput {
   newProjectId: String
 }
 
+input MoveProjectInput {
```
malikchaya2, line 373: Why is this not defined in gqlgen.yml?

malikchaya2: never mind, it doesn't need to be :)

malikchaya2 (approved):

malikchaya2 (approved):

PR #5182, opened 2021-11-10
EVG-15680 Fix join with absolute clone directory

bsamek:
[EVG-15680](https://jira.mongodb.org/browse/EVG-15680)

### Description 
Sometimes the clone directory is absolute. In those cases we shouldn't merge it with the working directory.

### Testing 
I added a unit test for the new directory function, but I didn't see a way to test the full code path.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -577,7 +577,7 @@ func (c *gitFetchProject) fetchModuleSource(ctx context.Context,
 
 	stdErr := noopWriteCloser{&bytes.Buffer{}}
 	err = jpm.CreateCommand(ctx).Add([]string{"bash", "-c", strings.Join(moduleCmds, "\n")}).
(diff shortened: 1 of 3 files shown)


ablack12 (commented):
We join workdir all over the agent package; should we make this more of a util function and extend it to other commands, or do we have reason to believe that this is unique to this function?

bsamek:
> We join workdir all over the agent package; should we make this more of a util function and extend it to other commands, or do we have reason to believe that this is unique to this function?

That's true. I'd like to make this a separate ticket to work on later though, since there are about 50 instances total.

ablack12 (approved):

bsamek:
@ablack12 FYI I opened https://jira.mongodb.org/browse/EVG-15829 and assigned to myself for the general case.

PR #5186, opened 2021-11-10
EVG-15828: Replace strings with enums in save mutation

sophstad:
[EVG-15828](https://jira.mongodb.org/browse/EVG-15828)

### Description 
- Replace strings with enums for added type safety on the front end ahead of EVG-15642

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -1290,37 +1290,3 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
-
-func GetProjectSettingsSection(ctx context.Context, sectionName ProjectSettingsSection) model.ProjectPageSection {
-	var section model.ProjectPageSection
-	switch sectionName {
-	case ProjectSettingsSectionGeneral:
-		section = model.ProjectPageGeneralSection
-		break
-	case ProjectSettingsSectionAccess:
-		section = model.ProjectPageAccessSection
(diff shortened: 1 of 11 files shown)


graphql/util.go
```diff
@@ -1290,3 +1290,37 @@ func getAPISubscriptionsForProject(ctx context.Context, projectId string) ([]*re
 	}
 	return res, nil
 }
+
+func GetProjectSettingsSection(ctx context.Context, sectionName ProjectSettingsSection) model.ProjectPageSection {
+	var section model.ProjectPageSection
+	switch sectionName {
+	case ProjectSettingsSectionGeneral:
```
ablack12, line 1297: Would it work to switch model.ProjectPageSection to just be the same enum?

sophstad: Just looked into it briefly, this pattern has not been used elsewhere and also introduces a dependency cycle (which tbh is perhaps the reason why). I could spend some more time looking into this but it may be more trouble than it's worth!

model/project_ref.go
```diff
@@ -339,15 +339,15 @@ const (
 type ProjectPageSection string
 
 const (
```
ablack12, line 341: Can we drop a comment here that these need to stay consistent with the graphql enums?

ablack12 (changes requested):
🎉

ablack12 (approved):

PR #5188, opened 2021-11-11
EVG-15829 Fix join with workdir for absolute case

bsamek:
https://jira.mongodb.org/browse/EVG-15829

### DIFF
--- a/agent/command/util.go
+++ b/agent/command/util.go
@@ -56,3 +56,14 @@ func expandModulePrefix(conf *internal.TaskConfig, module, prefix string, logger
 	conf.ModulePaths[module] = modulePrefix
 	return modulePrefix
 }
+
+// getJoinedWithWorkDir joins the conf.WorkDir A with B like this:
+//   if B is relative, return A+B.
+//   if B is absolute, return B.
+// We use this because B might be absolute.
+func getJoinedWithWorkDir(conf *internal.TaskConfig, path string) string {
+	if filepath.IsAbs(path) {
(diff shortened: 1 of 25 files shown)


agent/command/util.go
```diff
@@ -56,3 +56,14 @@ func expandModulePrefix(conf *internal.TaskConfig, module, prefix string, logger
 	conf.ModulePaths[module] = modulePrefix
 	return modulePrefix
 }
+
+// getJoinedWithWorkDir joins the conf.WorkDir A with B like this:
+//   if B is relative, return A+B.
+//   if B is absolute, return B.
+// We use this because B might be absolute.
+func getJoinedWithWorkDir(conf *internal.TaskConfig, path string) string {
```
ablack12, line 64: Why did you decide to have this take in the whole conf instead of just passing in the string that's necessary? My guess is that it's to verify that we actually only use this function for workdir, but now we do have to pass conf through new code paths.

bsamek: I didn't think about it too deeply. I thought that (1) it makes it harder for the person calling the function to flip the parameters by mistake, since they're disparate types, and (2) forces people to only use it when they're joining with workDir, as opposed to using the function on some other two paths without really understanding what it does—it's `getJoinedWithWorkDir`, not `joinTheSecondPathUnlessItIsAbsolute`.

That said, I'm happy to switch it to two strings if you think that's better.

ablack12 (approved):

PR #5189, opened 2021-11-11
EVG-15725 Only add files if S3 Copy succeeded

malikchaya2:
[EVG-15725](https://jira.mongodb.org/browse/EVG-15725)

### Description 
Right now, we upload files to the test even if s3_copy didn't succeed, or if it succeeded because it's optional, but didn't truly copy over. This is very confusing to users, especially if a file of that name already exists in the destination bucket, but it's not truly related to the test since the copy failed. 

### Testing 
Tested in staging. The first execution failed and has 0 files and the second execution succeeded and has 1 file. 
https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_s3_fake_destination_optional_patch_7e0cce4bf0fe845220526732bae0677256643421_618d9094b237364acb006453_21_11_11_21_52_38/logs?execution=0

### DIFF
--- a/agent/command/s3_copy.go
+++ b/agent/command/s3_copy.go
@@ -207,10 +207,15 @@ func (c *s3copy) s3Copy(ctx context.Context,
 			logger.Task().Infof("s3Copy response: %s", responseString)
 		}
 
-		if err != nil {
+		if err == nil {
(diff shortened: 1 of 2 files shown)


hadjri (approved):
LGTM following uptick of agent version

PR #5198, opened 2021-11-15
EVG-14837: use perf count cedar endpoint

hhoke:

### Description 
https://github.com/evergreen-ci/cedar/pull/410 resolved a bug where the route returned 404 if the task was not found, causing a server error, instead of 0, which resolves this. There is no new code here, this just reverts a revert.

### Testing 
3b6771530c77a165c26e6461183161bdceb00a08 contains integration tests that include task IDs that previously worked, and that were broken due to the task in question not being in cedar.


hhoke:
@khelif96 What happens here if a task does not store perf results in cedar but still needs to access them?

graphql/resolvers.go
```diff
@@ -3143,14 +3146,34 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+			}
+		}
+		opts := apimodels.GetCedarPerfCountOptions{
+			BaseURL:   evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:    utility.FromStringPtr(obj.Id),
+			Execution: obj.Execution,
+		}
+		if opts.BaseURL == "" {
+			return false, nil
+		}
```
khelif96, line 3167: nit. We can move this check to the top of the file. And avoid making any queries if `evergreen.GetEnvironment().Settings().Cedar.BaseURL == ""`

khelif96:
> @khelif96 What happens here if a task does not store perf results in cedar but still needs to access them?

I guess this question is more so for the Tips teams. If there are no perf results in cedar do they fetch it from a secondary source. 

From the original ticket it looks like perf results can only exist in cedar. So that might not be a scenario we encounter. https://jira.mongodb.org/browse/EVG-14546

graphql/resolvers.go
```diff
@@ -3143,14 +3146,34 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
...
+		result, err := apimodels.CedarPerfResultsCount(ctx, opts)
+		if err != nil {
+			return false, InternalServerError.Send(ctx, fmt.Sprintf("error requesting perf data from cedar: %s", err))
```
khelif96, line 3170: We should update this check. if the error is a 404 it should return false instead of an error.

hhoke: I modified the cedar route to return 0 rather than 404 if a task does not exist in cedar.

hhoke:
@julianedwards says perf results can exist in evergreen too. Julian, how do you think we should check if perf results exist in evergreen here?

ablack12:
Can you add a title to this PR @hhoke ?

julianedwards:
> @julianedwards says perf results can exist in evergreen too. Julian, how do you think we should check if perf results exist in evergreen here?

The easiest thing to do is a fallback, first check the evg DB then check cedar (or vice versa since most perf results are in cedar).

hhoke:
As shown in https://spruce.mongodb.com/task/perf_linux_wt_standalone_query_05e859095d739f6b7a7f0fa8672fced547d96c3f_15_05_18_19_35_23/trend-charts?execution=0

The perf plugin is non-functional in spruce, and functional in the old UI. This function is for spruce, so I should only check cedar.

PR #5201, opened 2021-11-15
EVG-13840 allow specifying variant tags for project aliases

ablack12:
[EVG-13840](https://jira.mongodb.org/browse/EVG-13840)

### Description 
I took the ticket to handle the backend portion of this work but it turns out it already works (considering [these functions](https://github.com/evergreen-ci/evergreen/blob/a5a7a6a1a3559f21b357be14142871dd3eddd8b4/model/project_aliases.go#L370-L369) are always used).
Added to the legacy UI so that users could stop being blocked on this; will add comments to the figma design to make sure that we do this for the new page.

### Testing 
Tested on local to verify that the information gets saved correctly.

### DIFF
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -446,16 +446,20 @@ Evergreen Projects
           </div>
         </div>
         <div id="patch-variants-list-header" class="form-group">
-          <div class="col-lg-3"> <label class="control-label"> Variant Regex </label> </div>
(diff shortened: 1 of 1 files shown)


malikchaya2 (commented):
Can you please add a screenshot of the change? :)

malikchaya2 (approved):

PR #5202, opened 2021-11-16
EVG-15709 Clear necessary fields between task executions

malikchaya2:
[EVG-15709](https://jira.mongodb.org/browse/EVG-15709)

### Description 
This clears host_create_details and override_dependencies. The ticket comment also also asks to clear task_failure_details: I'm not sure if that refers to task.details which is already being cleared, or something else. 

### Testing 
None, because there is no existing test to add to.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1623,11 +1623,13 @@ func resetTaskUpdate(t *Task) bson.M {
 			LastHeartbeatKey:       utility.ZeroTime,
 		},
 		"$unset": bson.M{
-			DetailsKey:            "",
-			HasCedarResultsKey:    "",
-			CedarResultsFailedKey: "",
-			ResetWhenFinishedKey:  "",
-			AgentVersionKey:       "",
+			DetailsKey:              "",
+			HasCedarResultsKey:      "",
+			CedarResultsFailedKey:   "",
+			ResetWhenFinishedKey:    "",
+			AgentVersionKey:         "",
+			HostCreateDetailsKey:    "",
+			OverrideDependenciesKey: "",
 		},
 	}
 	return update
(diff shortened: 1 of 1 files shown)


model/task/task.go
```diff
@@ -1623,11 +1623,13 @@ func resetTaskUpdate(t *Task) bson.M {
 			LastHeartbeatKey:       utility.ZeroTime,
 		},
 		"$unset": bson.M{
-			DetailsKey:            "",
-			HasCedarResultsKey:    "",
-			CedarResultsFailedKey: "",
-			ResetWhenFinishedKey:  "",
-			AgentVersionKey:       "",
+			DetailsKey:              "",
```
ablack12, line 1626: Should add these updates to the t != nil block as well.

ablack12 (changes requested):
When investigating https://jira.mongodb.org/browse/EVG-14742, I decided the issue was with Details because it isn't set [here|https://github.com/evergreen-ci/evergreen/blob/d6889a75eacbfad076d95b00cda3e40331242aaf/model/task/task.go#L2040]. Although it's set where you changed, I wonder if there's a time we're hitting Archive and not ResetTask. I think adding these fields or doing the reset in Archive would be the full fix? But I guess I'm not sure. Possibly we may just want to reopen that ticket.

model/task/task.go
malikchaya2: done

malikchaya2:
> When investigating https://jira.mongodb.org/browse/EVG-14742, I decided the issue was with Details because it isn't set [here|https://github.com/evergreen-ci/evergreen/blob/d6889a75eacbfad076d95b00cda3e40331242aaf/model/task/task.go#L2040]. Although it's set where you changed, I wonder if there's a time we're hitting Archive and not ResetTask. I think adding these fields or doing the reset in Archive would be the full fix? But I guess I'm not sure. Possibly we may just want to reopen that ticket.

Should we re-open that ticket since it would benefit from some additional investigation and maybe testing?

ablack12 (approved):
Feel free to reopen the other ticket

PR #5206, opened 2021-11-16
EVG-15584: only set goos/goarch if the Go binary is available

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15584

### Description 
This is mostly to reduce noise in the task logs. If the Go binary is not in the PATH (e.g. the mongod setup targets don't set them), setting the GOOS/GOARCH for the local CLI will fail. GOOS/GOARCH only used for the local binary and cross-compilation, so they should only be set for make targets that actually use Go in some way.

### Testing 
I checked that the CI tests pass and don't output the error `make: go: No such file or directory` in the task log.

### DIFF
--- a/makefile
+++ b/makefile
@@ -69,8 +69,14 @@ unixPlatforms := linux_amd64 darwin_amd64 $(if $(STAGING_ONLY),,darwin_arm64 lin
 windowsPlatforms := windows_amd64
 
 
-goos := $(shell $(gobin) env GOOS)
-goarch := $(shell $(gobin) env GOARCH)
+goos := $(GOOS)
+goarch := $(GOARCH)
+ifeq ($(goos),)
+goos := $(shell $(gobin) env GOOS 2> /dev/null)
+endif
+ifeq ($(goarch),)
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -69,8 +69,14 @@ unixPlatforms := linux_amd64 darwin_amd64 $(if $(STAGING_ONLY),,darwin_arm64 lin
 windowsPlatforms := windows_amd64
 
 
-goos := $(shell $(gobin) env GOOS)
-goarch := $(shell $(gobin) env GOARCH)
+goos := $(GOOS)
+goarch := $(GOARCH)
+ifeq ($(goos),)
+goos := $(shell $(gobin) env GOOS 2> /dev/null)
```
julianedwards, line 75: Does `/dev/null` exist on non-unix systems?

Kimchelly: Cygwin is a Unix-like system and Make uses its `/dev/null`, so it should work on Windows tests.

I added MacOS/Windows to the CI tests just to check.

julianedwards (approved):

PR #5210, opened 2021-11-16
EVG-15863 add loop on error to get version with good config

ablack12:
[EVG-15863](https://jira.mongodb.org/browse/EVG-15863)

### Description 
If the returned version has errors, try again.
Could also return all the versions right away, but really in almost all cases error _is_ empty, so I think it's reasonable to push the additional work off unless we've actually hit a problem.

### Testing 
Verifying that existing tests pass.

### DIFF
--- a/model/version_db.go
+++ b/model/version_db.go
@@ -66,7 +66,7 @@ var VersionAll = db.Query(bson.D{})
 
 // FindVersionByLastKnownGoodConfig filters on versions with valid (i.e., have no errors) config for the given project.
 func FindVersionByLastKnownGoodConfig(projectId string, revisionOrderNumber int) (*Version, error) {
-	retryLimit := 50
+	const retryLimit = 50
 	q := bson.M{
 		VersionIdentifierKey: projectId,
 		VersionRequesterKey:  evergreen.RepotrackerVersionRequester,
--- a/model/version_db.go
+++ b/model/version_db.go
@@ -67,15 +67,14 @@ var VersionAll = db.Query(bson.D{})
 // FindVersionByLastKnownGoodConfig filters on versions with valid (i.e., have no errors) config for the given project.
 func FindVersionByLastKnownGoodConfig(projectId string, revisionOrderNumber int) (*Version, error) {
 	retryLimit := 50
+	q := bson.M{
+		VersionIdentifierKey: projectId,
+		VersionRequesterKey:  evergreen.RepotrackerVersionRequester,
+	}
+	if revisionOrderNumber >= 0 {
+		q[VersionRevisionOrderNumberKey] = bson.M{"$lt": revisionOrderNumber}
+	}
 	for i := 0; i < retryLimit; i++ {
-		q := bson.M{
-			VersionIdentifierKey: projectId,
-			VersionRequesterKey:  evergreen.RepotrackerVersionRequester,
-		}
-
-		if revisionOrderNumber >= 0 {
-			q[VersionRevisionOrderNumberKey] = bson.M{"$lt": revisionOrderNumber}
-		}
 		v, err := VersionFindOne(db.Query(q).Sort([]string{"-" + VersionRevisionOrderNumberKey}))
 		if err != nil {
 			return nil, errors.Wrapf(err, "Error finding recent valid version for '%s'", projectId)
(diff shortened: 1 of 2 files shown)


model/version_db.go
```diff
@@ -64,27 +64,32 @@ func VersionByIds(ids []string) db.Q {
...
-		},
-		VersionErrorsKey: bson.M{
-			"$exists": false,
-		},
-	}
+	retryLimit := 50
+	for i := 0; i < retryLimit; i++ {
+		q := bson.M{
+			VersionIdentifierKey: projectId,
+			VersionRequesterKey:  evergreen.RepotrackerVersionRequester,
```
ablack12, line 73: I also realized it was maybe a bug to use any other requester here, since git tag versions can technically have any config, not the one the project typically uses. I audited some places we use this and I feel like this change is very safe, but if we're nervous about it I can create a different ticket.

bsamek:
@malikchaya2 @hadjri  I made some changes to this PR. Since this addresses the outage, I'd like to merge it soon. Could you take a look?

@hadjri I made a change to a test for some code you wrote, which I think seems legit but wasn't 100% sure.

model/version_db.go
```diff
@@ -64,27 +64,29 @@ func VersionByIds(ids []string) db.Q {
 // All is a query for all versions.
 var VersionAll = db.Query(bson.D{})
 
-// FindVersionByLastKnownGoodConfig filters on versions with valid (i.e., have no errors) config for the given
-// project. Does not apply a limit, so should generally be used with a findOneRepoRefQ.
+// FindVersionByLastKnownGoodConfig filters on versions with valid (i.e., have no errors) config for the given project.
 func FindVersionByLastKnownGoodConfig(projectId string, revisionOrderNumber int) (*Version, error) {
+	retryLimit := 50
```
malikchaya2, line 69: Theoretically, if we have time, it would be a good idea to move this to a const.

model/version_db.go
```diff
@@ -64,27 +64,29 @@ func VersionByIds(ids []string) db.Q {
...
 	}
-
 	if revisionOrderNumber >= 0 {
 		q[VersionRevisionOrderNumberKey] = bson.M{"$lt": revisionOrderNumber}
 	}
-	v, err := VersionFindOne(db.Query(q).Sort([]string{"-" + VersionRevisionOrderNumberKey}))
-	if err != nil {
-		return nil, errors.Wrapf(err, "Error finding recent valid version for '%s'", projectId)
+	for i := 0; i < retryLimit; i++ {
+		v, err := VersionFindOne(db.Query(q).Sort([]string{"-" + VersionRevisionOrderNumberKey}))
```
malikchaya2, line 78: I don't think .Sort works on FindOne(). What is this trying to do?

malikchaya2 (changes requested):

model/version_db.go
bsamek: Done.

model/version_db.go
bsamek: This didn't change in this PR. This is the same as writing `db.<coll>.findOne(<query>).sort(<query>)` or  `db.<coll>.find(<query>).sort(<query>).limit(1)`. It's trying to find the document with the greatest `VersionRevisionOrderNumberKey`.

malikchaya2: Based on what I'm reading online its seems like `db.<coll>.findOne(<query>).sort(<query>)` would not work, but `db.<coll>.find(<query>).sort(<query>).limit(1)` would. However, if we know that this has been working till now, then maybe what I'm reading is outdated.

hadjri: Here sort is setting the q.sort value for the query: https://github.com/evergreen-ci/evergreen/blob/main/db/query.go#L51

So that later this property can be used in FindOneQ: https://github.com/evergreen-ci/evergreen/blob/main/db/query.go#L82

This allows us to retrieve a sorted response with most recent version revision number first and take the first element, hence Limit(1) that is applied.

hadjri (approved):

model/version_db.go
malikchaya2: aaah that makes sense, thank you for explaining that!

malikchaya2 (approved):

PR #5212, opened 2021-11-16
EVG-15774 Respect admin permissions in SetPatchPriority resolver

minnakt:
[EVG-15774](https://jira.mongodb.org/browse/EVG-15774)

### Description 
This PR resolves the issue where it became impossible for users with elevated permissions to set the priority of their patch above `100` (max priority for a regular user).

### Testing 
Tested on staging by setting my own patch to have priority `999`:

Screen Shot 2021-11-16 at 3 18 43 PM

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -618,18 +618,9 @@ func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectR
 			}
 		}
 	case SetPriority:
-		var projId string
-		if proj == nil {
-			foundProjId, err := model.GetIdForProject(version.Identifier)
-			if err != nil {
-				return http.StatusNotFound, errors.Errorf("error getting project ref: %s", err)
-			}
-			if foundProjId == "" {
-				return http.StatusNotFound, errors.Errorf("project for %s came back nil: %s", version.Branch, err)
-			}
-			projId = foundProjId
-		} else {
-			projId = proj.Id
+		projId := version.Identifier
+		if projId == "" {
+			return http.StatusNotFound, errors.Errorf("Could not find project for version %s", version.Id)
 		}
 		if modifications.Priority > evergreen.MaxTaskPriority {
 			requiredPermission := gimlet.PermissionOpts{
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -620,13 +620,14 @@ func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectR
 	case SetPriority:
 		var projId string
 		if proj == nil {
-			projId, err := model.GetIdForProject(version.Identifier)
+			foundProjId, err := model.GetIdForProject(version.Identifier)
 			if err != nil {
 				return http.StatusNotFound, errors.Errorf("error getting project ref: %s", err)
 			}
-			if projId == "" {
+			if foundProjId == "" {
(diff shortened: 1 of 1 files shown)


graphql/util.go
```diff
@@ -620,13 +620,14 @@ func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectR
...
 				return http.StatusNotFound, errors.Errorf("error getting project ref: %s", err)
 			}
-			if projId == "" {
+			if foundProjId == "" {
 				return http.StatusNotFound, errors.Errorf("project for %s came back nil: %s", version.Branch, err)
 			}
+			projId = foundProjId
 		} else {
 			projId = proj.Id
 		}
```
minnakt, line 633: The issue was that `projId` was technically being redeclared within the `if` statement because of the `:=` syntax. I think using this syntax doesn’t change the value of `projId` outside of the `if` statement, but instead declares a new variable within this scope. 

For this reason, `projId` was always an empty string `""` (the default value) when constructing the `gimlet.PermissionOpts` struct, so it was not possible to correctly determine if a user had the appropriate permissions for setting patch priority on a given project.

ablack12: good find!!! The reason this is happening in the if statement is because we're using := (which creates a new variable) instead of = (which assigns to the variable); you could continue to use projId if you wanted within the if statement and set it this way instead (`projId, err = ...`).

ablack12: Actually we shouldn't have to do this at all; version.Identifier already equals the project ID (which is confusing since it's labeled identifier), so we could just set projId directly to version.Identifier

minnakt: I'll give it a go! Thank you for reviewing my PR right before your vacation 🔥

graphql/util.go
```diff
@@ -618,17 +618,9 @@ func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectR
...
-				return http.StatusNotFound, errors.Errorf("error getting project ref: %s", err)
-			}
-			if projId == "" {
-				return http.StatusNotFound, errors.Errorf("project for %s came back nil: %s", version.Branch, err)
-			}
-		} else {
-			projId = proj.Id
+		projId := version.Identifier
+		if projId == "" {
+			return http.StatusNotFound, errors.Errorf("Could not find project for version %s", version.Id)
```
minnakt, line 623: Would this ever happen? 🤔 i.e. would `version.Identifier` ever be `""`

minnakt:
Sorry if I tagged you out of the blue 🥲 I don’t know a lot about this repo, so I tagged people who might have more context about Annie’s suggestion.

graphql/util.go
khelif96: Someone on app can verify but i don't think thats possible. But if it is its a good defensive programming technique to case on it. Otherwise i think it will just return false on line 632

khelif96 (approved):

graphql/util.go
bynn: checked in the prod database and there does seem to be some without an identifier but they seem to be weird without a project so i think it's safe to assume that it shouldnt be empty

bynn (approved):

hadjri (approved):

malikchaya2 (approved):

PR #5214, opened 2021-11-17
EVG-15202 Add variant tasks to downstream tasks 

malikchaya2:
[EVG-15202](https://jira.mongodb.org/browse/EVG-15202)

### Description 
It looked like downstream tasks wasn't populating for the api endpoint, but what was really happening was that the downstream patch had no tasks, it only had variant tasks. 

### Testing 
Added to the existing test.

### DIFF
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -50,7 +50,7 @@ type APIPatch struct {
 type DownstreamTasks struct {
 	Project      *string       `json:"project"`
 	Tasks        []*string     `json:"tasks"`
-	VariantTasks []VariantTask `json:"tasks"`
+	VariantTasks []VariantTask `json:"variant_tasks"`
 }
 
 type ChildPatch struct {
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -12,7 +12,6 @@ import (
 	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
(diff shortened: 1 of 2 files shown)


rest/model/patch.go
```diff
@@ -189,8 +191,22 @@ func (apiPatch *APIPatch) BuildFromService(h interface{}) error {
 
 	apiPatch.PatchedConfig = utility.ToStringPtr(v.PatchedConfig)
 	apiPatch.CanEnqueueToCommitQueue = v.HasValidGitInfo()
+	//this child apipatch has a child patch
+	// it has variant tasks but no tasks
+	grip.Error(message.Fields{
+		"error":    "chayaMTesting patch.go 196",
```
bynn, line 197: remove comments

rest/model/patch.go
```diff
@@ -189,8 +191,22 @@ func (apiPatch *APIPatch) BuildFromService(h interface{}) error {
...
+		"v":        v,
+		"apiPatch": apiPatch,
+	})
 
 	downstreamTasks, childPatches, err := getChildPatchesData(v)
+	grip.Error(message.Fields{
+		"error":           "chayaMTesting patch.go 196",
```
bynn, line 204: same

bynn (approved):

hadjri (approved):

rest/model/patch.go
malikchaya2: ah oops! thanks for spotting that!

PR #5215, opened 2021-11-17
EVG-13030: Resolve taskStatusCounts for Patch gql type

SupaJoon:
[EVG-13030](https://jira.mongodb.org/browse/EVG-13030)

### Description 
This field will be used  to display task counts on the user and project patches page. 
https://mongodb.invisionapp.com/console/Evergreen-Version-2---Updated-Typography-cknazaa9t9fbb01xljqq8o8q1/cknaznc5ia0su01xlly201tfo/play
### Testing 
integration tests

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3558,7 +3558,6 @@ func (r *versionResolver) BaseTaskStatuses(ctx context.Context, v *restModel.API
 	return getAllTaskStatuses(tasks), nil
 }
 
-// Returns task status counts (a mapping between status and the number of tasks with that status) for a version.
 func (r *versionResolver) TaskStatusCounts(ctx context.Context, v *restModel.APIVersion, options *BuildVariantOptions) ([]*task.StatusCount, error) {
 	stats, err := getTaskStatusCountsForVersion(*v.Id, options)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


graphql/resolvers.go
```diff
@@ -1237,6 +1237,14 @@ func (r *patchResolver) BaseTaskStatuses(ctx context.Context, obj *restModel.API
 	return getAllTaskStatuses(baseTasks), nil
 }
 
+func (r *patchResolver) TaskStatusCounts(ctx context.Context, obj *restModel.APIPatch, options *BuildVariantOptions) ([]*task.StatusCount, error) {
```
khelif96, line 1240: I don't think we should do this. Patches don't technically have tasks. The version is what contains the tasks. 

I think we should instead make it possible for a patch to resolve a version. So the caller can resolve the TaskStatusCounts from that resolver instead. 

So the call would look like this. 
```
user {
patches {
  version {
    taskStatusCounts
  }
} 
}

```

khelif96: if the patch is unconfigured and has no version  and no tasks it will just return null.

khelif96 (changes requested):
We should go with the approach we discussed originally to make this return a version and just rely on the existing resolvers instead of trying to duplicate them for the patch.

graphql/resolvers.go
SupaJoon: Sounds good! I originally thought to create the separate field to save us a query into the Version collection when querying patches but it seems useful to query any Version field from a patch.

PR #5216, opened 2021-11-17
EVG-13030: Query version on patch

SupaJoon:
[EVG-13030](https://jira.mongodb.org/browse/EVG-13030)

### Description 
Query a patches version to allow access to the taskStatusCounts from the user and project patches query

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -811,7 +811,7 @@ type Patch {
   patchNumber: Int!
   author: String!
   authorDisplayName: String!
-  version: String!
+  version: Version
   status: String!
   variants: [String!]!
   tasks: [String!]!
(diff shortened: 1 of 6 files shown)


graphql/schema.graphql
```diff
@@ -811,7 +811,7 @@ type Patch {
   patchNumber: Int!
   author: String!
   authorDisplayName: String!
-  version: String!
+  version: Version
```
khelif96, line 814: Can you validate that this isn't used on the UI?

khelif96 (approved):
LGTM pending we can make this change in a backwards compatible way.

graphql/schema.graphql
SupaJoon: It's not utilized 👍

PR #5217, opened 2021-11-18
EVG-15765 Verify that display task is not nil

bynn:
[EVG-15765](https://jira.mongodb.org/browse/EVG-15765)

### Description 
add a nil check for display task 

### Testing 
n/a

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1395,6 +1395,9 @@ func UpdateDisplayTaskForTask(t *task.Task) error {
 	if err != nil {
 		return errors.Wrapf(err, "error getting display task for task")
 	}
+	if dt == nil {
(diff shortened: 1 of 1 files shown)


model/task_lifecycle.go
```diff
@@ -1395,6 +1395,9 @@ func UpdateDisplayTaskForTask(t *task.Task) error {
 	if err != nil {
 		return errors.Wrapf(err, "error getting display task for task")
 	}
+	if dt == nil {
+		return errors.Errorf("%s may hold a display task that doesn't exist", t.Id)
```
malikchaya2, line 1399: Let's also log an error to splunk with grip.Error so that we can track it on our to see what went wrong. 
Also, I would maybe use that wording for splunk, and use wording like "display task not found for... '%s'"  or something like that.

malikchaya2 (approved):

PR #5218, opened 2021-11-18
EVG-15154 remove ExecutionTasksFull fields from task type

khelif96:
[EVG-15154](https://jira.mongodb.org/browse/EVG-15154)

this isn't stored anywhere and is only used in the ui so doesn't need to be apart of the db struct

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -172,11 +172,10 @@ type Task struct {
 	LocalTestResults []TestResult `bson:"-" json:"test_results"`
 
 	// display task fields
-	DisplayOnly        bool     `bson:"display_only,omitempty" json:"display_only,omitempty"`
-	ExecutionTasks     []string `bson:"execution_tasks,omitempty" json:"execution_tasks,omitempty"`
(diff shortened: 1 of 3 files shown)


model/task/task.go
```diff
@@ -172,11 +172,10 @@ type Task struct {
 	LocalTestResults []TestResult `bson:"-" json:"test_results"`
 
 	// display task fields
-	DisplayOnly        bool     `bson:"display_only,omitempty" json:"display_only,omitempty"`
-	ExecutionTasks     []string `bson:"execution_tasks,omitempty" json:"execution_tasks,omitempty"`
-	ExecutionTasksFull []Task   `bson:"execution_tasks_full" json:"-"` // this is a local pointer from a display task to its execution tasks
```
malikchaya2, line 177: We should also probably remove it from here: 
https://github.com/evergreen-ci/evergreen/blob/main/model/task/db.go#L67

malikchaya2 (changes requested):

bynn (approved):

model/task/task.go
khelif96: Thanks!

malikchaya2 (approved):

PR #5219, opened 2021-11-18
EVG-15837 Update CreateTasksFromGroup to disable correctly

bynn:
[EVG-15837](https://jira.mongodb.org/browse/EVG-15837)

### Description 
Previously, all taskgroup tasks were showing up on the config page even if they weren't supposed to (ie upload-cli sets patchable to false but still shows up when you create a patch.) The changes fix this problem 

### Testing 
uplad-cli doesn't show up on the config page anymore
https://spruce-staging.corp.mongodb.com/patch/61967796b2373666af059883/configure/tasks
image

### DIFF
--- a/model/lifecycle_test.go
+++ b/model/lifecycle_test.go
@@ -1993,7 +1993,7 @@ func TestCreateTasksFromGroup(t *testing.T) {
 			},
 		},
 	}
-	bvts := CreateTasksFromGroup(in, p)
+	bvts := CreateTasksFromGroup(in, p, "")
 	assert.Equal("new_dependency", bvts[0].DependsOn[0].Name)
 	assert.Equal("new_dependency", bvts[1].DependsOn[0].Name)
 }
--- a/model/lifecycle_test.go
+++ b/model/lifecycle_test.go
@@ -1993,7 +1993,7 @@ func TestCreateTasksFromGroup(t *testing.T) {
 		TaskGroups: []TaskGroup{
 			{
 				Name:  "name",
-				Tasks: []string{"first_task", "second_task"},
+				Tasks: []string{"first_task", "second_task", "third_task"},
 			},
 		},
 	}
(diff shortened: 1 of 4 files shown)


malikchaya2 (commented):
What problem is this trying to solve? What was happening before that made this change necessary?

model/lifecycle_test.go
```diff
@@ -1993,7 +1993,7 @@ func TestCreateTasksFromGroup(t *testing.T) {
 			},
 		},
 	}
-	bvts := CreateTasksFromGroup(in, p)
+	bvts := CreateTasksFromGroup(in, p, "")
```
malikchaya2, line 1996: can we maybe pass in a requester here and check that it's handled properly?

model/lifecycle_test.go
```diff
@@ -1985,15 +1985,19 @@ func TestCreateTasksFromGroup(t *testing.T) {
...
 				Name:  "name",
-				Tasks: []string{"first_task", "second_task"},
+				Tasks: []string{"first_task", "second_task", "third_task"},
 			},
 		},
 	}
-	bvts := CreateTasksFromGroup(in, p)
+	bvts := CreateTasksFromGroup(in, p, evergreen.PatchVersionRequester)
 	assert.Equal("new_dependency", bvts[0].DependsOn[0].Name)
 	assert.Equal("new_dependency", bvts[1].DependsOn[0].Name)
```
malikchaya2, line 2002: should the length here be 2, since the third task shouldn't be included? If yes, should we maybe check the length here?

malikchaya2 (changes requested):

model/lifecycle_test.go
malikchaya2: discussed over zoom

malikchaya2 (approved):

PR #5223, opened 2021-11-22
EVG-15852 Remove useRepoSettings

malikchaya2:
[EVG-15852](https://jira.mongodb.org/browse/EVG-15852)

### Description 
Remove the useRepoSettings and replace it with useRepoSettings() which checks the RepoRefId field. 

### Testing 
Existing tests should still pass.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -849,11 +849,9 @@ func (p *ProjectRef) createNewRepoRef(u *user.DBUser) (repoRef *RepoRef, err err
 	if !utility.StringSliceContains(repoRef.Admins, u.Username()) {
 		repoRef.Admins = append(repoRef.Admins, u.Username())
 	}
-	if repoRef.Id != "" {
-		return nil, errors.Wrap(err, "repo ref is already defined")
-	}
 	// some fields shouldn't be set from projects
 	repoRef.Id = mgobson.NewObjectId().Hex()
+	repoRef.RepoRefId = ""
 	// set explicitly in case no project is enabled
 	repoRef.Owner = p.Owner
 	repoRef.Repo = p.Repo
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -105,7 +105,6 @@ type ProjectRef struct {
 	BuildBaronSettings evergreen.BuildBaronSettings `bson:"build_baron_settings,omitempty" json:"build_baron_settings,omitempty" yaml:"build_baron_settings,omitempty"`
 	PerfEnabled        *bool                        `bson:"perf_enabled,omitempty" json:"perf_enabled,omitempty" yaml:"perf_enabled,omitempty"`
 
-	// This is a temporary flag to enable individual projects to use repo settings
 	RepoRefId string `bson:"repo_ref_id" json:"repo_ref_id" yaml:"repo_ref_id"`
 
 	// The following fields are used by Evergreen and are not discoverable.
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -849,6 +849,9 @@ func (p *ProjectRef) createNewRepoRef(u *user.DBUser) (repoRef *RepoRef, err err
 	if !utility.StringSliceContains(repoRef.Admins, u.Username()) {
 		repoRef.Admins = append(repoRef.Admins, u.Username())
 	}
+	if repoRef.Id != "" {
+		return nil, errors.Wrap(err, "repo ref is already defined")
+	}
 	// some fields shouldn't be set from projects
 	repoRef.Id = mgobson.NewObjectId().Hex()
 	// set explicitly in case no project is enabled
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -106,8 +106,7 @@ type ProjectRef struct {
 	PerfEnabled        *bool                        `bson:"perf_enabled,omitempty" json:"perf_enabled,omitempty" yaml:"perf_enabled,omitempty"`
 
 	// This is a temporary flag to enable individual projects to use repo settings
-	UseRepoSettings bool   `bson:"use_repo_settings" json:"use_repo_settings" yaml:"use_repo_settings"`
-	RepoRefId       string `bson:"repo_ref_id" json:"repo_ref_id" yaml:"repo_ref_id"`
+	RepoRefId string `bson:"repo_ref_id" json:"repo_ref_id" yaml:"repo_ref_id"`
 
 	// The following fields are used by Evergreen and are not discoverable.
 	// Hidden determines whether or not the project is discoverable/tracked in the UI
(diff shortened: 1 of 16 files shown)


rest/route/repo_test.go
```diff
@@ -307,190 +315,4 @@ func TestPatchHandlersWithRestricted(t *testing.T) {
 	assert.Len(t, allProjectsScope.Resources, 2)
 	assert.Contains(t, allProjectsScope.Resources, "branch1")
 	assert.Contains(t, allProjectsScope.Resources, "branch2")
-
```
malikchaya2, line 310: This test doesn't play nice with the new setup and I wasn't successful in figuring out how to re-write it.

bynn (approved):

model/project_ref.go
```diff
@@ -106,8 +106,7 @@ type ProjectRef struct {
 	PerfEnabled        *bool                        `bson:"perf_enabled,omitempty" json:"perf_enabled,omitempty" yaml:"perf_enabled,omitempty"`
 
 	// This is a temporary flag to enable individual projects to use repo settings
```
ablack12, line 108: This comment should change now (or be removed entirely) since it's referencing UseRepoSettings.

rest/model/project.go
```diff
@@ -651,7 +650,7 @@ func (p *APIProjectRef) BuildFromService(v interface{}) error {
 	p.PRTestingEnabled = utility.BoolPtrCopy(projectRef.PRTestingEnabled)
 	p.GitTagVersionsEnabled = utility.BoolPtrCopy(projectRef.GitTagVersionsEnabled)
 	p.GithubChecksEnabled = utility.BoolPtrCopy(projectRef.GithubChecksEnabled)
-	p.UseRepoSettings = projectRef.UseRepoSettings
+	p.UseRepoSettings = projectRef.UseRepoSettings()
```
ablack12, line 653: Do we need this at all now?

rest/route/project.go
```diff
@@ -433,17 +433,12 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
...
 	// if owner/repo has changed or we're toggling repo settings off, update scope
 	if h.newProjectRef.Owner != h.originalProject.Owner || h.newProjectRef.Repo != h.originalProject.Repo ||
-		(!h.newProjectRef.UseRepoSettings && h.originalProject.UseRepoSettings) {
+		(!h.newProjectRef.UseRepoSettings() && h.originalProject.UseRepoSettings()) {
 		if err = h.newProjectRef.RemoveFromRepoScope(); err != nil {
 			return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "error removing project from old repo scope"))
 		}
 		h.newProjectRef.RepoRefId = "" // if using repo settings, will reassign this in the next block
 	}
-	if h.newProjectRef.UseRepoSettings && h.newProjectRef.RepoRefId == "" {
```
ablack12, line 442: So we actually still need this block to exist; users need a way to say "I want to turn on using repo settings" (which is probably why we introduced this field to begin with womp).  I think we could use requestProjectRef to handle this, since APIProjectRef is keeping UseRepoSettings; we'll also need to set this to true (or handle AddToRepoScope) in the above block.

rest/route/project.go
```diff
@@ -671,20 +666,19 @@ func (h *projectDeleteHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.MakeJSONErrorResponder(errors.Errorf("project '%s' is already hidden", h.projectName))
 	}
 
-	if !project.UseRepoSettings {
+	if !project.UseRepoSettings() {
 		return gimlet.MakeJSONErrorResponder(
-			errors.Errorf("project '%s' must have UseRepoSettings enabled to be eligible for deletion", h.projectName))
+			errors.Errorf("project '%s' must be attached to a repo enabled to be eligible for deletion", h.projectName))
```
ablack12, line 671: should drop "enabled" from this comment

ablack12, line 671: should drop "enabled" from this comment

ablack12 (changes requested):

rest/route/project.go
malikchaya2: This comment is a duplicate, right?

ablack12: 😮  looks like it, not sure how that even happens

malikchaya2: haha me neither! I just wanted to double check!

rest/route/project.go
malikchaya2: ah wait, let me look into using requestProjectRef.

malikchaya2: Can you please elaborate on "we'll also need to set this to true (or handle AddToRepoScope) in the above block"? It seems to me that if we just swap out h.newProjectRef.UseRepoSettings for h.apiNewProjectRef.UseRepoSettings everything should continue to work. Am I misunderstanding?

malikchaya2: discussed over slack. done. (I also reset the tests)

model/project_ref.go
```diff
@@ -854,7 +851,6 @@ func (p *ProjectRef) createNewRepoRef(u *user.DBUser) (repoRef *RepoRef, err err
 	}
 	// some fields shouldn't be set from projects
 	repoRef.Id = mgobson.NewObjectId().Hex()
-	repoRef.UseRepoSettings = false
```
ablack12, line 857: I'm not sure why we don't already explicitly make sure RepoRefId is empty but it should be.

rest/route/project.go
ablack12: One last thought: before if `use_repo_settings` was set to false by the user, it was fine if we kept `repo_ref_id` set to whatever, but now we want to explicitly set it to empty. We should handle that here in UpdateProject, i.e. if UseRepoSettings is explicitly false, set RepoRefId to empty. That means I think that UseRepoSettings in APIProjectRef needs to be a boolean pointer.

ablack12 (changes requested):

model/project_ref.go
malikchaya2: I added an error if it's not empty. Is that what you meant?

rest/route/project.go
malikchaya2: like that?

malikchaya2: (Apologies for the rebase, the last two commits are new.)

model/project_ref.go
ablack12: Sorry, I meant that here the line that was deleted: `repoRef.UseRepoSettings = false` should probably be replaced with `repoRef.RepoRefId = ""` (since that's not a real field for repos).

model/project_ref.go
```diff
@@ -849,6 +849,9 @@ func (p *ProjectRef) createNewRepoRef(u *user.DBUser) (repoRef *RepoRef, err err
 	if !utility.StringSliceContains(repoRef.Admins, u.Username()) {
 		repoRef.Admins = append(repoRef.Admins, u.Username())
 	}
+	if repoRef.Id != "" {
```
ablack12, line 852: This isn't really necessary, since we create the repoRef ourselves in this function

rest/route/project.go
```diff
@@ -442,9 +442,20 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
 		}
 	}
 
-	if h.apiNewProjectRef.UseRepoSettings && h.newProjectRef.RepoRefId == "" {
-		if err = h.newProjectRef.AddToRepoScope(h.user); err != nil {
-			return gimlet.MakeJSONInternalErrorResponder(err)
+	// if the user explicitly set useRepoSettings, update the repo accordingly
+	if h.apiNewProjectRef.UseRepoSettings != nil {
+		useRepoSettings := utility.FromBoolPtr(h.apiNewProjectRef.UseRepoSettings)
+		// the user set it to false, and it was previously true - remove
```
ablack12, line 448: These comments are great!

rest/route/project.go
```diff
@@ -442,9 +442,20 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
...
+		if !useRepoSettings && h.newProjectRef.RepoRefId != "" {
+			h.newProjectRef.RepoRefId = ""
```
ablack12, line 450: We should set this after RemoveFromRepoScope or else it will noop.

hadjri (approved):

ablack12 (approved):
LGTM! Since the patch route logic is a little tricky I would suggest testing on local by toggling use_repo_settings on / off and verifying that this works as you would expect

PR #5226, opened 2021-11-22
EVG-15882 Update CreateTasksFromGroup to pass in params from project task

bynn:
[EVG-15882](https://jira.mongodb.org/browse/EVG-15882)

### Description 
https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#task-groups
these 4 params will overwrite the ones in project tasks 

### Testing 
unit test

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -749,6 +749,10 @@ func CreateTasksFromGroup(in BuildVariantTaskUnit, proj *Project, requester stri
 	}
 
 	for _, t := range tg.Tasks {
+		pt := proj.FindProjectTask(t)
+		if pt == nil {
+			return tasks
+		}
 		bvt := BuildVariantTaskUnit{
 			Name: t,
 			// IsGroup is not persisted, and indicates here that the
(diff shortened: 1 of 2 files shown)


malikchaya2 (approved):
This looks good to me. However, since this is a conceptual change, I'd suggest waiting on a review from Annie or Brian as well.

bsamek:
Swapped myself out for @ablack12

model/lifecycle.go
```diff
@@ -767,6 +771,19 @@ func CreateTasksFromGroup(in BuildVariantTaskUnit, proj *Project, requester stri
 			Activate:         in.Activate,
 			CommitQueueMerge: in.CommitQueueMerge,
 		}
+		// Overwrite limiting when tasks will run from project task
```
ablack12, line 774: We're not overwriting, we're defaulting right?

ablack12: Maybe what we should do instead is always call bvt.Populate (line 788) but then only add to tasks (line 789) if 787 is met. This would prevent us from special casing ahead of time.

ablack12 (changes requested):

model/lifecycle.go

ablack12 (approved):

PR #5227, opened 2021-11-23
EVG-15868: Move newly exposed configurations to separate collection

hadjri:
[EVG-15868](https://jira.mongodb.org/browse/EVG-15868)

### Description 
Retrieving the most recent valid parser project requires a lookup of all parser projects associated with a project, which can be expensive since some projects have a large number of parser project versions accrued over time.  This has been shown to put stress on evergreen in production since this is looked up very frequently in the codebase.  Instead, we can store the newly supported config properties as a different struct in a different collection.  These are derived from the same yaml file that generates the parser project, and will be less prone to long lists of version history. This will also allow us to directly look up the configurations instead of making a DB call to get the most recent version, and then use that version to make another DB query to retrieve its parser project.

Note: This PR is large with 50+ files edited, this is due to the change of signature for some functions that are widely used.  Many of these edits are just making sure the function signatures are consistent everywhere.  The true logic changes of this PR are mainly in these 10 files below.

model/patch_lifecycle.go
model/project.go
model/project_aliases.go
model/project_configs.go
model/project_configs_db.go
model/project_parser.go
model/project_parser_db.go
model/project_ref.go
repotracker/repotracker.go
trigger/project_triggers.go

### Testing 
Made multiple edits across several unit tests.


repotracker/repotracker.go
```diff
@@ -1110,6 +1112,19 @@ func createVersionItems(ctx context.Context, v *model.Version, metadata model.Ve
 			}
 			return errors.Wrapf(err, "error inserting version '%s'", v.Id)
 		}
+		// todo uncomment
+		//_, err = db.Collection(model.ProjectConfigsCollection).InsertOne(sessCtx, projectInfo.IntermediateConfig)
```
ablack12, line 1116: We should only insert this if IntermediateConfig isn't nil, and we should only set IntermediateConfig if anything exists for it (may want an IsEmpty function or something that checks all of the things besides ID? Not sure)

model/project_parser.go
```diff
@@ -517,35 +501,40 @@ func (pss *parserStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) e
 
 // LoadProjectForVersion returns the project for a version, either from the parser project or the config string.
 // If read from the config string and shouldSave is set, the resulting parser project will be saved.
-func LoadProjectForVersion(v *Version, id string, shouldSave bool) (*Project, *ParserProject, error) {
+func LoadProjectForVersion(v *Version, id string, shouldSave bool) (*Project, *ParserProject, *ProjectConfigs, error) {
```
ablack12, line 504: This return is getting messy; we may want to refactor this to use ProjectInfo or something instead, but this PR is already giant so if you want to make a new ticket that would make sense

ablack12 (changes requested):
I'm not actually 100% sure I read everything haha but this is a start.

A good way to test this besides unit testing may be to ensure that this works in staging with a project yaml with config options.

model/project_parser.go
```diff
@@ -793,6 +786,18 @@ func createIntermediateProject(yml []byte) (*ParserProject, error) {
 	return p, nil
 }
 
+// CreateIntermediateConfig marshals the supplied YAML into our
+// intermediate configs representation.
+func CreateIntermediateConfig(yml []byte) (*ProjectConfigs, error) {
```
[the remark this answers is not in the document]
hadjri: I thought it was imported in another package but I guess it is not. Changed to lower case.

model/project_configs.go
```diff
@@ -0,0 +1,46 @@
...
+	Enabled     *bool     `yaml:"enabled,omitempty" bson:"enabled,omitempty"`
+	Owner       *string   `yaml:"owner,omitempty" bson:"owner,omitempty"`
+	Repo        *string   `yaml:"repo,omitempty" bson:"repo,omitempty"`
+	RemotePath  *string   `yaml:"remote_path,omitempty" bson:"remote_path,omitempty"`
+	Branch      *string   `yaml:"branch,omitempty" bson:"branch,omitempty"`
+	Identifier  *string   `yaml:"identifier,omitempty" bson:"identifier,omitempty"`
+	DisplayName *string   `yaml:"display_name,omitempty" bson:"display_name,omitempty"`
+	CreateTime  time.Time `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
+
+	// These fields can be set for the ProjectRef struct on the project page, or in the project config yaml.
```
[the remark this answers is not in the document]
hadjri: These fields can still be set on the project ref struct, they are what gets picked up from inputs on the project page.

model/project_parser_merge_functions.go
```diff
@@ -180,36 +180,6 @@ func (pp *ParserProject) mergeUnique(toMerge *ParserProject) error {
 		pp.ExecTimeoutSecs = toMerge.ExecTimeoutSecs
```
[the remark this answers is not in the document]
hadjri: Sure that new naming makes sense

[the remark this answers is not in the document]
hadjri: I think as long as we keep it so that there is no way to include file paths on the project config struct, that guarantees that the configs have to exist in the top level yaml.  Would it be necessary / helpful to allow for merging of the ProjectConfig struct? The number of properties it supports is relatively small.

model/project_parser.go
hadjri: Will create ticket to track this

hadjri: https://jira.mongodb.org/browse/EVG-15923

repotracker/repotracker.go
hadjri: Implemented isEmpty for the project configs to prevent this.

model/project_ref.go
```diff
@@ -404,29 +404,37 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
...
-		if parserProject.TaskAnnotationSettings != nil {
-			p.TaskAnnotationSettings = *parserProject.TaskAnnotationSettings
+		var emptyAnnotations evergreen.AnnotationsSettings
+		if reflect.DeepEqual(p.TaskAnnotationSettings, emptyAnnotations) {
+			p.TaskAnnotationSettings = *projectConfig.TaskAnnotationSettings
 		}
-		if parserProject.WorkstationConfig != nil {
-			p.WorkstationConfig = *parserProject.WorkstationConfig
+		var emptyWorkstation WorkstationConfig
+		if reflect.DeepEqual(p.WorkstationConfig, emptyWorkstation) {
```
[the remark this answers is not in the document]
hadjri: Implemented something similar to the recursive merge of project ref and repo refs that happens elsewhere in the codebase.  We can now mix and match sub-properties of the same structs between the ref and the config yaml.

model/patch_lifecycle.go
```diff
@@ -428,6 +431,12 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		if err != nil {
 			return nil, errors.Wrapf(err, "error inserting parser project for version '%s'", patchVersion.Id)
 		}
+		if config != nil {
+			_, err = db.Collection(ProjectConfigsCollection).InsertOne(sessCtx, config)
+			if err != nil {
+				return nil, errors.Wrapf(err, "error inserting project configs for version '%s'", patchVersion.Id)
```
ablack12, line 437: nit: project config

model/project_aliases.go
```diff
@@ -140,21 +140,21 @@ func findMatchingAliasForProjectRef(projectID, alias string) ([]ProjectAlias, bo
 
 // Returns true if we have an alias match or the alias doesn't match but
 //  other aliases in the category are defined, in which case we shouldn't check other sources.
-func findMatchingAliasForParserProject(projectID, alias string) ([]ProjectAlias, bool, error) {
+func findMatchingAliasForProjectConfigs(projectID, alias string) ([]ProjectAlias, bool, error) {
```
ablack12, line 143: nit: ProjectConfig (same for the variables below)

model/project_aliases.go
```diff
@@ -173,35 +173,35 @@ func aliasesToMap(aliases []ProjectAlias) map[string][]ProjectAlias {
 	return output
 }
 
-func getFullParserProjectAliases(parserProject *ParserProject) []ProjectAlias {
-	var parserProjectAliases []ProjectAlias
-	if parserProject != nil {
-		for _, commitQueueAlias := range parserProject.CommitQueueAliases {
+func getFullProjectConfigsAliases(projectConfigs *ProjectConfig) []ProjectAlias {
```
ablack12, line 176: ProjectConfig for everything in here too

model/project_aliases_test.go
```diff
@@ -161,71 +159,71 @@ func (s *ProjectAliasSuite) TestFindAliasInProject() {
...
-	s.Len(projectAliases, 1)
-	projectAliases, err = FindAliasInProjectOrRepo("project-1", "alias-2")
-	s.NoError(err)
-	s.Len(projectAliases, 1)
-
-	projectAliases, err = FindAliasInProjectOrRepo("project-1", "nonexistent")
-	s.NoError(err)
-	s.Len(projectAliases, 0)
-}
+//func (s *ProjectAliasSuite) TestMergeAliasesWithParserProject() {
```
ablack12, line 162: Is this still commented out because of the error with creating a new collection? Did you try 
```
    _ = env.DB().RunCommand(nil, map[string]string{"create": evergreen.ProjectConfigCollection})
 ```

File: model/project_configs.go [hunk 11]
```diff
@@ -0,0 +1,76 @@
...
+	CommitQueueAliases     []ProjectAlias                 `yaml:"commit_queue_aliases,omitempty" bson:"commit_queue_aliases,omitempty"`
+	GitHubPRAliases        []ProjectAlias                 `yaml:"github_pr_aliases,omitempty" bson:"github_pr_aliases,omitempty"`
+	GitTagAliases          []ProjectAlias                 `yaml:"git_tag_aliases,omitempty" bson:"git_tag_aliases,omitempty"`
+	GitHubChecksAliases    []ProjectAlias                 `yaml:"github_checks_aliases,omitempty" bson:"github_checks_aliases,omitempty"`
+	PatchAliases           []ProjectAlias                 `yaml:"patch_aliases,omitempty" bson:"patch_aliases,omitempty"`
+	DeactivatePrevious     *bool                          `yaml:"deactivate_previous" bson:"deactivate_previous,omitempty"`
+	WorkstationConfig      *WorkstationConfig             `yaml:"workstation_config,omitempty" bson:"workstation_config,omitempty"`
+	CommitQueue            *CommitQueueParams             `yaml:"commit_queue,omitempty" bson:"commit_queue,omitempty"`
+	TaskSync               *TaskSyncOptions               `yaml:"task_sync,omitempty" bson:"task_sync,omitempty"`
+} // End of ProjectConfig struct
```
ablack12, line 35: Are we using this "End of Project Config struct" comment? This is only for the function where we test if things are merged, which we won't be doing.

model/project_configs.go
```diff
@@ -0,0 +1,76 @@
...
+// createConfig marshals the supplied YAML into our
+// intermediate configs representation.
+func createConfig(yml []byte) (*ProjectConfig, error) {
+	p := &ProjectConfig{}
+	if err := util.UnmarshalYAMLWithFallback(yml, p); err != nil {
+		yamlErr := thirdparty.YAMLFormatError{Message: err.Error()}
+		return nil, errors.Wrap(yamlErr, "error unmarshalling into project configs")
+	}
+	isEmpty := p.isEmpty()
+	if isEmpty {
```
ablack12, line 71: we can just call `if p.IsEmpty()` in one line rather than creating a variable

model/project_configs.go
```diff
@@ -0,0 +1,76 @@
...
+	return true
+}
+
+// createConfig marshals the supplied YAML into our
+// intermediate configs representation.
+func createConfig(yml []byte) (*ProjectConfig, error) {
+	p := &ProjectConfig{}
+	if err := util.UnmarshalYAMLWithFallback(yml, p); err != nil {
+		yamlErr := thirdparty.YAMLFormatError{Message: err.Error()}
+		return nil, errors.Wrap(yamlErr, "error unmarshalling into project configs")
```
ablack12, line 68: projectConfig

model/project_configs.go
```diff
@@ -0,0 +1,76 @@
...
+				return false
+			}
+		}
+	}
+	return true
+}
+
+// createConfig marshals the supplied YAML into our
+// intermediate configs representation.
+func createConfig(yml []byte) (*ProjectConfig, error) {
```
ablack12, line 64: can we call this createProjectConfig?

model/project_configs_db.go
```diff
@@ -0,0 +1,82 @@
...
+	adb "github.com/mongodb/anser/db"
+	"github.com/pkg/errors"
+)
+
+const (
+	ProjectConfigsCollection = "project_configs"
+)
+
+var (
+	ProjectConfigsIdKey         = bsonutil.MustHaveTag(ProjectConfig{}, "Id")
```
ablack12, line 16: These should all be ProjectConfig to match the struct

model/project_parser_merge_functions.go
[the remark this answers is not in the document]
ablack12: I think we'll just need to be really clear in the documentation, because users don't have a concept of "ParserProject struct" and "ProjectConfig struct", they just have a yaml file. Probably it would be okay to enforce that all of these things be defined at the top level for now and this would simplify things for us, but we'll just need to be clear.

model/project_ref_test.go
```diff
@@ -61,13 +61,13 @@ func TestFindMergedProjectRef(t *testing.T) {
...
 	parserProject := &ParserProject{
-		Id:                 "ident",
-		DeactivatePrevious: utility.TruePtr(),
-		TaskAnnotationSettings: &evergreen.AnnotationsSettings{
-			FileTicketWebhook: evergreen.WebHook{
-				Endpoint: "random2",
-			},
-		},
+		Id: "ident",
+		//DeactivatePrevious: utility.TruePtr(),
```
ablack12, line 65: Why is this commented out?

model/project_ref.go
```diff
@@ -404,32 +404,43 @@ func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerD
...
+	projectConfig, err := FindProjectConfigToMerge(p.Id, version)
 	if err != nil {
 		return err
 	}
-	if parserProject != nil {
-		if parserProject.PerfEnabled != nil {
-			p.PerfEnabled = parserProject.PerfEnabled
+	if projectConfig != nil {
+		defer func() {
+			err = recovery.HandlePanicWithError(recover(), err, "project and repo structures do not match")
```
ablack12, line 418: copy pasta, should be project ref and config.

I'm surprised that`recursivelySetUndefinedFields` actually works here, because they `do` have different structures, and this function assumes that the structs are the same (because projectRef and repoRef are the same). Also, why are we using reflect but then also assigning workstationConfig and etc explicitly?

plugin/build_baron.go
```diff
@@ -208,20 +208,12 @@ func BbGetProject(settings *evergreen.Settings, projectId string, version string
 		return evergreen.BuildBaronSettings{}, false
 	}
 	if flags.PluginAdminPageDisabled {
-		if version == "" {
-			lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectId, -1)
-			if err == nil && lastGoodVersion != nil {
-				version = lastGoodVersion.Id
-			}
+		projectConfigs, err := model.FindProjectConfigToMerge(projectId, version)
```
ablack12, line 211: projectConfig

plugin/build_baron.go
```diff
@@ -208,20 +208,12 @@ func BbGetProject(settings *evergreen.Settings, projectId string, version string
...
 		}
-		if version != "" {
-			parserProject, err := model.ParserProjectFindOneById(version)
-			if err != nil {
-				return evergreen.BuildBaronSettings{}, false
-			}
-			if parserProject != nil && parserProject.BuildBaronSettings != nil {
-				return *parserProject.BuildBaronSettings, true
-			}
+		if projectConfigs != nil && projectConfigs.BuildBaronSettings != nil {
```
ablack12, line 215: This should come after ProjectRef since that now takes precedence, but idk if you have a different ticket to handle that.

repotracker/repotracker.go
```diff
@@ -1110,6 +1114,20 @@ func createVersionItems(ctx context.Context, v *model.Version, metadata model.Ve
 			}
 			return errors.Wrapf(err, "error inserting version '%s'", v.Id)
 		}
+		if projectInfo.Config != nil {
+			_, err = db.Collection(model.ProjectConfigsCollection).InsertOne(sessCtx, projectInfo.Config)
+			if err != nil {
+				grip.Notice(message.WrapError(err, message.Fields{
+					"message": "aborting transaction",
+					"cause":   "can't insert project configs",
```
ablack12, line 1122: project config

model/project_parser.go
```diff
@@ -577,18 +561,22 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
...
 // If reading from a version config, LoadProjectForVersion should be used to persist the resulting parser project.
 // opts is used to look up files on github if the main parser project has an Include.
-func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
+func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, identifier string, project *Project) (*ParserProject, *ProjectConfig, error) {
 	intermediateProject, err := createIntermediateProject(data)
 	if err != nil {
-		return nil, errors.Wrapf(err, LoadProjectError)
+		return nil, nil, errors.Wrapf(err, LoadProjectError)
+	}
+	config, err := createConfig(data)
```
ablack12, line 569: we need to set config.Identifier (similarly to 609)

model/project_configs.go
hadjri: Gotcha, removing

model/project_configs.go
hadjri: Sure, renamed

model/project_ref_test.go
hadjri: Ah,  a byproduct of previous debugging. Uncommented

model/project_ref.go
hadjri: This works because I am assigning a new ProjectRef struct, and then explicitly setting each config property (workstation config etc) explicitly, and only then letting recursivelySetUndefinedFields do its magic since I am now comparing two project ref structs.

plugin/build_baron.go
hadjri: Actually since findMergedProjectRef already takes care of this we can remove the entire block looking up project config from line 211 - 217. Removing.

model/patch_lifecycle.go
hadjri: Renamed to project config for all mentions in the repo

model/project_aliases_test.go
hadjri: Works, thanks! I don't know why only some tests were giving me trouble with this and not all of them across the board.

model/project_configs_db.go
```diff
@@ -0,0 +1,89 @@
...
+	ProjectConfigsGitTagAliasesKey          = bsonutil.MustHaveTag(ProjectConfigs{}, "GitTagAliases")
+	ProjectConfigsGitHubChecksAliasesKey    = bsonutil.MustHaveTag(ProjectConfigs{}, "GitHubChecksAliases")
+	ProjectConfigsPatchAliasesKey           = bsonutil.MustHaveTag(ProjectConfigs{}, "PatchAliases")
+	ProjectConfigsWorkstationConfigKey      = bsonutil.MustHaveTag(ProjectConfigs{}, "WorkstationConfig")
+	ProjectConfigsCommitQueueKey            = bsonutil.MustHaveTag(ProjectConfigs{}, "CommitQueue")
+	ProjectConfigsTaskSyncKey               = bsonutil.MustHaveTag(ProjectConfigs{}, "TaskSync")
+)
+
+// FindProjectConfigToMerge returns a project config by id, or the most recent project config if id is empty
+func FindProjectConfigToMerge(projectId, id string) (*ProjectConfigs, error) {
```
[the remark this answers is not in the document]
hadjri: Is that done with code?

ablack12 (changes requested):
Just a couple of small things!

model/patch_lifecycle.go
```diff
@@ -162,16 +161,12 @@ func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken str
 			return nil, "", errors.Wrapf(err, "Could not patch remote configuration file")
 		}
 	}
-	pp, _, err := LoadProjectInto(ctx, projectFileBytes, opts, p.Project, project)
+	_, _, err = LoadProjectInto(ctx, projectFileBytes, opts, p.Project, project)
```
ablack12, line 164: The reason we use the parser project output is because it contains the included files, so using projectFileBytes potentially omits a lot of new stuff. I'm not super sure what to do about that for this project; maybe patches need a new variable to store the PatchedProjectConfig (for lack of better term) if it exists. You could also just drop a TODO here and handle it in a new ticket.

model/project_aliases.go
```diff
@@ -200,19 +200,35 @@ func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
...
+	// If nothing is defined in the DB, check the project config,
+	// unless the alias defined is a patch alias and there are patch aliases
 	// defined on the project page.
 	if len(aliases) > 0 {
 		return aliases, nil
 	}
 	shouldExit := false
 	if IsPatchAlias(alias) {
-		numPatchAliases, err := countPatchAliasesForProjectRef(projectID)
+		projectRefPatchAliases, err := countPatchAliases(projectID)
```
ablack12, line 211: nit: should still use num or count or something in the variable name since this is just an integer and not the actual aliases.

model/project_aliases.go
```diff
@@ -200,19 +200,35 @@ func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
...
 		if err != nil {
 			return nil, errors.Wrap(err, "error counting patch aliases")
 		}
-		shouldExit = numPatchAliases > 0
+		shouldExit = projectRefPatchAliases > 0
+		if !shouldExit {
```
ablack12, line 216: We could always make a new ticket to improve this if that's easier, but this block of code is pretty repetitive with FindAliasInProjectOrRepoFromDb; it would be better if we could find a way for that function to give us more information. For example, it already does this count https://github.com/evergreen-ci/evergreen/blob/ffde3b1d0dc6e70aeea094410c86a288f50e35d9/model/project_aliases.go#L132 and decides whether or not we need to look at the repo. Perhaps the repo function https://github.com/evergreen-ci/evergreen/blob/ffde3b1d0dc6e70aeea094410c86a288f50e35d9/model/project_aliases.go#L248 should also now count in the case that there's no match, and then return whether or not we should exit (bc repo patch aliases exist) or continue onto the project config. That would save us two repeated DB calls in this function.

hadjri: Sounds reasonable - refactored.

model/patch_lifecycle.go
hadjri: Added todo covering new ticket: https://jira.mongodb.org/browse/EVG-16000

model/project_aliases.go
```diff
@@ -206,30 +206,6 @@ func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
 	if len(aliases) > 0 {
```
ablack12, line 206: can we change this to if len(aliases) > 0 || shouldExit rather than two statements?

ablack12 (approved):
LGTM with nit!

PR #5229, opened 2021-11-23
EVG-1496 Modify TaskQueueDistros resolver to also return HostCount

minnakt:
[EVG-1496](https://jira.mongodb.org/browse/EVG-1496)

### Description 
This PR modifies the `TaskQueueDistros` resolver so that it returns the number of hosts associated with each distro. Previously, this function would only return the distro `ID` and `QueueCount` (the number of tasks on the queue).

### Testing 
- Modified the tests in `gql/tests/taskQueueDistros`


graphql/resolvers.go
```diff
@@ -2205,16 +2205,21 @@ func (r *queryResolver) TaskQueueDistros(ctx context.Context) ([]*TaskQueueDistr
 	distros := []*TaskQueueDistro{}
 
 	for _, distro := range queues {
+		numHosts, err := host.CountRunningHosts(distro.Distro)
+		if err != nil {
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting associated hosts: %v", err.Error()))
```
SupaJoon, line 2210:
```suggestion
			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting associated hosts: %s", err.Error()))
```

SupaJoon (approved):

graphql/schema.graphql
```diff
@@ -672,7 +672,8 @@ type TaskQueueItem {
 
 type TaskQueueDistro {
   id: ID!
-  queueCount: Int!
```
SupaJoon, line 675: we should keep this field to retain backwards compatability

SupaJoon (changes requested):

graphql/schema.graphql
khelif96: This is actually safe to remove it isn't used anywhere on the front end.

khelif96 (approved):

graphql/schema.graphql
SupaJoon: queueCount is queried here: https://github.com/evergreen-ci/spruce/blob/main/src/pages/TaskQueue.tsx#L32-L35
as part of this page: https://spruce.mongodb.com/task-queue/rhel71-power8-small

SupaJoon (approved):

PR #5230, opened 2021-11-29
EVG-15281 Remove plugin references to admin page

hadjri:
[EVG-15281](https://jira.mongodb.org/browse/EVG-15281)

### Description 
Plugin configs currently may be derived from admin or project settings depending on a service flag set.  Once it is confirmed that deriving from project settings works fine we no longer need any of the old code that references the admin settings. As part of this refactoring I have moved the place we validate BB configs which covers this ticket as well: https://jira.mongodb.org/browse/EVG-15419

### Testing 
Modified existing unit tests.  PluginAdminPageDisabled flag should be toggled in production for a day or so before these changes are to be merged into main.


graphql/resolvers.go
```diff
@@ -3154,49 +3154,24 @@ func (r *taskResolver) IsPerfPluginEnabled(ctx context.Context, obj *restModel.A
 	if !evergreen.IsFinishedTaskStatus(utility.FromStringPtr(obj.Status)) {
 		return false, nil
 	}
-	flags, err := evergreen.GetServiceFlags()
+	projectMatches := model.IsPerfEnabledForProject(*obj.ProjectId)
```
ablack12, line 3157: ProjectMatches doesn't really fit anymore; could just call IsPerfEnabled directly in the if statement

plugin/build_baron.go
```diff
@@ -17,114 +17,11 @@ func init() {
...
 }
 
-type bbPluginOptions struct {
-	Projects map[string]evergreen.BuildBaronSettings
-}
-
-type BuildBaronPlugin struct {
-	opts *bbPluginOptions
-}
+type BuildBaronPlugin struct{}
```
ablack12, line 20: Is there a reason we can't just delete this file completely? (Not sure if it breaks something but)

plugin/build_baron.go
```diff
@@ -149,28 +46,13 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
 	}, nil
 }
 
-// IsWebhookConfigured webhook will can be retrieved from project or admin config depending on PluginAdminPageDisabled flag
-// if deriving from project config, we first try to retrieve webhook config prom project parser config, otherwise we fallback to project page settings
-// version is needed to retrieve last good project config, if version is not available/empty when calling this function we must first retrieve it
+// IsWebhookConfigured retrieves webhook configuration from the project settings.
 func IsWebhookConfigured(project string, version string) (evergreen.WebHook, bool, error) {
```
ablack12, line 50: I think all of these functions should be refactored to be in the model package, since build_baron isn't really a plugin anymore but a project setting.

plugin/build_baron.go
```diff
@@ -199,44 +81,93 @@ func BbGetConfig(settings *evergreen.Settings) map[string]evergreen.BuildBaronSe
...
-	if flags.PluginAdminPageDisabled {
-		if version == "" {
-			lastGoodVersion, err := model.FindVersionByLastKnownGoodConfig(projectId, -1)
-			if err == nil && lastGoodVersion != nil {
-				version = lastGoodVersion.Id
-			}
+	return projectRef.BuildBaronSettings, true
+}
+
+func validateBbProject(projName string, proj evergreen.BuildBaronSettings) bool {
```
ablack12, line 99: Do we already do all of this validation in the project page / project patch route? We shouldn't allow users to store an invalid configuration, in which case we shouldn't need to validate here at all.

ablack12 (changes requested):

plugin/build_baron.go
hadjri: Moved to project_ref.go

plugin/build_baron.go
hadjri: This validation used to happen looping through each project on startup - now that that code is gone it does not exist anywhere but here.  However I do agree it makes sense to move this check into the modifyProject route since that is the only time we are writing to these BB configs.

plugin/build_baron.go
hadjri: The struct needs to exist for the Publish method at the top of each plugin file to register the plugin (as we do for manifest, perf, BB, etc).  This is the way we register plugins so that their associated PanelConfig can be displayed in the UI.  This can probably have its pattern change as an improvement in the future since it is a bit confusing but it would be out of scope for this ticket.  See the RegisterPlugins method for reference.

service/project.go
```diff
@@ -1009,3 +1015,87 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
 	}
 	return false, nil
 }
+
+func bbProjectIsValid(projName string, proj evergreen.BuildBaronSettings) error {
+	webHook, _, _ := model.IsWebhookConfigured(projName, "")
```
ablack12, line 1020: I think we should return an error if there's an error here.

service/project.go
```diff
@@ -1009,3 +1015,87 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+	webhookConfigured := webHook.Endpoint != ""
+
+	if !webhookConfigured && proj.TicketCreateProject == "" {
```
ablack12, line 1023: Should this be an error still? I think this made sense when we had a list of build baron projects, but now this configuration just implies a project that doesn't intend to use the build baron. You can test this a bit on Evergreen local.

service/project.go
```diff
@@ -1009,3 +1015,87 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+		msg := "ticket_create_project and taskAnnotationSettings.FileTicketWebhook endpoint cannot both be blank"
```
ablack12, line 1024: These messages shouldn't be logged critical anymore, since the project isn't able to save the invalid configuration (critical logs get sent to evergreen-ops-alerts). We can just return the error.

service/project.go
```diff
@@ -1009,3 +1015,87 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+		grip.Critical(message.Fields{
+			"message":      msg,
+			"project_name": projName,
+		})
+		return errors.Errorf(msg)
+	}
+	if !webhookConfigured && len(proj.TicketSearchProjects) == 0 {
+		msg := "ticket_search_projects cannot be empty"
```
ablack12, line 1032: We should make these a bit more user-friendly, using whatever terms the user would understand (must provide projects to search)

ablack12 (changes requested):

service/project.go
hadjri: Agreed, this should just have the error immediately return as nil.  Changed to this.

hadjri: Tested locally to confirm the modifyProject route works

operations/service.go
```diff
@@ -18,7 +18,7 @@ import (
 func Service() cli.Command {
 	return cli.Command{
 		Name:  "service",
-		Usage: "run evergreen services",
+		Usage: "run evergreen services MALIK",
```
ablack12, line 21: should change this back haha

plugin/build_baron.go
```diff
@@ -149,94 +40,13 @@ func (bbp *BuildBaronPlugin) GetPanelConfig() (*PanelConfig, error) {
...
-
-// BbGetProject retrieves build baron settings from project or admin config depending on PluginAdminPageDisabled flag.
-// Project parser config takes precedence, otherwise fallback to project page settings (found using version ID if given).
+// BbGetProject retrieves build baron settings from project settings.
+// Project page settings takes precedence, otherwise fallback to project config yaml.
 // Returns build baron settings and ok if found.
-func BbGetProject(settings *evergreen.Settings, projectId string, version string) (evergreen.BuildBaronSettings, bool) {
-	flags, err := evergreen.GetServiceFlags()
-	if err != nil {
+func BbGetProject(projectId string, version string) (evergreen.BuildBaronSettings, bool) {
```
ablack12, line 46: This function also should be in model now (and maybe renamed to make more sense, like GetBuildBaronSettings)

service/project.go
ablack12: This is only not an error if webhook not configured, ticket create project is empty, AND ticket search project is empty. Right now this will allow users to just set ticket search project but not ticket create project, which isn't a valid configuration.

service/project.go
```diff
@@ -1009,3 +1015,47 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+
+	if !webhookConfigured && proj.TicketCreateProject == "" {
+		return nil
+	}
+	if !webhookConfigured && len(proj.TicketSearchProjects) == 0 {
+		return errors.Errorf("Must provide projects to search")
+	}
+	if proj.BFSuggestionServer != "" {
+		if _, err = url.Parse(proj.BFSuggestionServer); err != nil {
+			return errors.Errorf(fmt.Sprintf(`Failed to parse bf_suggestion_server for project "%s"`, projName))
```
ablack12, line 1034: Don't need fmt.Sprintf; Errorf already does this. Also we typically use "" as the outer quotes and ' ' to wrap formatting

service/project.go
```diff
@@ -1009,3 +1015,47 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+		return errors.Errorf(fmt.Sprintf(`Failed validating configuration for project "%s": `+
+			"bf_suggestion_username and bf_suggestion_password must be blank alt_endpoint_url is blank", projName))
+	} else if proj.BFSuggestionTimeoutSecs != 0 {
+		return errors.Errorf(fmt.Sprintf(`Failed validating configuration for project "%s": `+
+			"bf_suggestion_timeout_secs must be zero when bf_suggestion_url is blank", projName))
+	}
+	// the webhook cannot be used if the default build baron creation and search is configured
+	if webhookConfigured {
+		if len(proj.TicketCreateProject) != 0 {
+			return errors.Errorf("The custom file ticket webhook and the build baron TicketCreateProject should not both be configured")
```
ablack12, line 1054: should probably drop TicketCreateProject here

operations/service.go
hadjri: hahah oops

service/project.go
hadjri: Got it, added

service/project.go
```diff
@@ -566,6 +567,11 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Errorf("expected build baron config but was actually '%T'", i))
 		return
 	}
+	err = bbProjectIsValid(projectRef.Id, buildbaronConfig)
```
ablack12, line 570: Not sure if there's currently a way to handle bbProject in the patch route. If there is then we should add validation; if not we should add it (but we could make a new ticket for that).

service/project.go
```diff
@@ -1009,3 +1015,47 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+	webHook, _, err := model.IsWebhookConfigured(projName, "")
+	if err != nil {
+		return err
+	}
+	webhookConfigured := webHook.Endpoint != ""
+
+	if !webhookConfigured && proj.TicketCreateProject == "" && len(proj.TicketSearchProjects) == 0 {
+		return nil
+	}
+	if !webhookConfigured && len(proj.TicketSearchProjects) == 0 {
```
ablack12, line 1029: Should we also be checking the other combination? !webhooksConfigured && proj.TicketCreateProject == "" ? (the "must provide project to create tickets for" case)

service/project.go
```diff
@@ -1009,3 +1015,47 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+		return errors.Errorf("Must provide projects to search")
+	}
+	if proj.BFSuggestionServer != "" {
+		if _, err = url.Parse(proj.BFSuggestionServer); err != nil {
+			return errors.Errorf("Failed to parse bf_suggestion_server for project '%s'", projName)
```
ablack12, line 1034: can we make all of the errors in this loop more human readable? what does BF suggestion server show up as for the UI?

service/project.go
```diff
@@ -1009,3 +1015,47 @@ func verifyAliasExists(alias, projectId string, newAliasDefinitions []model.Proj
...
+		if proj.BFSuggestionUsername == "" && proj.BFSuggestionPassword != "" {
+			return errors.Errorf("Failed validating configuration for project '%s': "+
+				"bf_suggestion_password must be blank if bf_suggestion_username is blank", projName)
+		}
+		if proj.BFSuggestionTimeoutSecs <= 0 {
+			return errors.Errorf("Failed validating configuration for project '%s': "+
+				"bf_suggestion_timeout_secs must be positive", projName)
+		}
+	} else if proj.BFSuggestionUsername != "" || proj.BFSuggestionPassword != "" {
+		return errors.Errorf("Failed validating configuration for project '%s': "+
```
ablack12, line 1045: > bf_suggestion_username and bf_suggestion_password must be blank alt_endpoint_url is blank

is there an if or a when missing from this sentence?

ablack12 (changes requested):

service/project.go
hadjri: I'll create a ticket for this

service/project.go
hadjri: Sounds good, added!

service/project.go
hadjri: The old UI does not have any inputs right now for BF suggestion server (because no project had anything for it). These properties were migrated over as part of the plugin work but I'm actually not sure if they are used at all anymore - there is not any use of them in the evergreen repo aside from this validation that existed. I can create another ticket to investigate this and remove them if needed.

ablack12 (approved):
apparently i never clicked this button 🙃

ablack12 (approved):
I don't see a new thing to look at so I assume you wanted me to look at it again just verify that it’s good to merge now that the admin plugins are updated? You can go ahead once merge conflicts are resolved!

service/project.go
hadjri: https://jira.mongodb.org/browse/EVG-15964

PR #5231, opened 2021-11-29
EVG-15772 Fail version when task aborts

hadjri:
[EVG-15772](https://jira.mongodb.org/browse/EVG-15772)

### Description 
Setting a version's status to failed when a task aborts instead of maintaining the "started" state
### Testing 
Tested by aborting a task in staging for a version and confirmed that the version no longer stays stuck as "started" for its status and does appropriately set to "failed": https://spruce-staging.corp.mongodb.com/version/61b3d7a6b237366caa7df321/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -3,6 +3,7 @@ package task
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/model"
 	"testing"
 	"time"
 
@@ -2863,7 +2864,7 @@ func TestAbortVersion(t *testing.T) {
 	}
 	assert.NoError(t, db.InsertMany(Collection, finishedExecTask, failingExecTask, otherExecTask, dt))
 
-	assert.NoError(t, AbortVersion("v1", AbortInfo{TaskID: "et2"}))
+	assert.NoError(t, model.AbortVersion("v1", AbortInfo{TaskID: "et2"}))
 
 	var err error
 	dt, err = FindOneId("dt")
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -3,7 +3,6 @@ package task
 import (
 	"context"
 	"fmt"
-	"github.com/evergreen-ci/evergreen/model"
 	"testing"
 	"time"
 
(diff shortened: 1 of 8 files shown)


model/task/task_test.go
```diff
@@ -3,6 +3,7 @@ package task
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/model"
```
ablack12, line 6: nit: this import should be in the below list, not this one

model/task_lifecycle.go
```diff
@@ -338,7 +338,21 @@ func AbortTask(taskId, caller string) error {
 		return err
 	}
 	event.LogTaskAbortRequest(t.Id, t.Execution, caller)
-	return t.SetAborted(task.AbortInfo{User: caller})
+	err = t.SetAborted(task.AbortInfo{User: caller})
+	if err != nil {
+		return err
+	}
+
+	err = VersionUpdateOne(
```
ablack12, line 346: I'm suspicious of just setting this like this; when we update a version there are typically other things we also need to handle (build status, patch status, etc). Can https://github.com/evergreen-ci/evergreen/blob/d22272191944362483efe21761b7eff89c596e09/model/task_lifecycle.go#L1052 be used here? (Or adapted?)

ablack12: Also I'm not sure that this should automatically be set to failed here; if one task is aborted, but other tasks are running, we still want the version status to be running. So modifying UpdateBuildAndVersionStatus to look at aborted might be something to look into.

ablack12 (changes requested):

model/task_lifecycle.go
hadjri: Thanks for the tip - the task status changes cascade to the build status, which then cascade to version statuses in that method.  It looks like this 1 line change is all that was needed to fix. I tested in staging with this new change and aborting one task fails the version, but only after all other tasks finish.

ablack12 (approved):

PR #5232, opened 2021-11-29
EVG-15022: handle hosts that start in the cloud but are stuck building

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15022

### Description 
Hosts that are "building" are in an intermediate state between an intent host and a real host running in the cloud. In the "building" state, the host makes two important changes - 1. it attempts to start the host, and then 2. it updates the DB document metadata for the host. The metadata updates are to change the host document ID from its intent host ID to the EC2 instance ID (by removing the old host doc and inserting a new one) and transition the status to "starting". The issue is that it's not always clear whether or not a host whose current state is stuck "building" has simply failed to start a host (e.g. something's wrong with the instance request parameters) or if it actually started a host successfully and the DB document was never updated with the new metadata (e.g. the app server cancelled the request to create the instance, but on EC2's side, it started successfully).

This PR addresses the case where the host starts in EC2 but the host document is stuck in "building". The current treatment of such a host is that the intent host document is simply removed. This is problematic because the cloud host's agent can still check into the app server but is unable to do anything, leaving it alive to continuously poll for the next task but unable to actually receive any task and unable to terminate until the AWS reaper catches it. To fix this, I did a couple things:

* Create a new "building-failed" state, which is when the host document is stuck "building" for too long. Hosts marked as "building-failed" will be terminated by the host termination job.
    * This is different from the "decommissioned" state, which is only valid when a host has definitely started and has an instance ID. In the case of a "building-failed", the server is uncertain whether the host is actually alive in EC2 or not.
* Instead of removing intent host documents that are stuck in "building" and stale, the host allocator instead changes their state to "building-failed" as an intermediate to be sent to the host termination job.
* The agent now passes the EC2 instance ID in the next task payload if it's available. If the host is stuck "building" but the agent is alive, the agent can pass the instance ID so the app server can perform operations (e.g. terminate, get instance status) on the instance in the cloud.
* During the next task route, deal with an intent host that has a running agent by updating the host ID with the real instance ID passed with the next task payload. Either set the host to "starting" or "decommissioned" depending on whether the host document is considered stale or not.
* Update the host termination job to allow intent hosts to be terminated and cloud hosts to be re-terminated if they're already in the terminated state.
    * The re-termination logic is to handle an edge case where the intent host could be considered stale and marked as terminated, but the agent checks in much later when the host is already marked terminated.
* Make a few small helper functions to break up a few very long and verbose functions.

### Testing 
* Unit tests for host termination job, next task route, and DB queries.
* Test in staging that agents on EC2 instances send the instance ID with their next task payload.


model/host/db.go
```diff
@@ -904,7 +907,7 @@ func FindUserDataSpawnHostsProvisioning() ([]Host, error) {
 	return hosts, nil
 }
 
-// Removes host intents that have been been uninitialized for more than 3
+// Removes host intents that have been been initializing for more than 3
```
julianedwards, line 910: Remove duplicate `been`.

julianedwards: Also, can you update this documentation to reflect the changes you made removing the condition of checking hosts that have been building for 15+ minutes?

model/host/db.go
```diff
@@ -937,16 +955,27 @@ func RemoveStaleInitializing(distroID string) error {
...
 	for _, h := range hosts {
 		ids = append(ids, h.Id)
 	}
+	if len(ids) == 0 {
+		return nil
+	}
 
-	if err := db.RemoveAll(evergreen.CredentialsCollection, bson.M{CertUserIDKey: bson.M{"$in": ids}}); err != nil {
-		return errors.Wrap(err, "could not delete credentials")
+	if _, err := db.UpdateAll(Collection, bson.M{
```
julianedwards, line 966: Why not just do an `UpdateAll` with the original query? Is it just for the event logging below?

julianedwards (changes requested):

agent/util/ec2.go
```diff
@@ -1,17 +1,23 @@
...
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 )
 
+const metadataBaseURL = "http://169.254.169.254/latest/meta-data"
```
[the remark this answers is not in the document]
Kimchelly: Done.

model/host/db.go
Kimchelly: Done.

model/host/db.go
Kimchelly: It's for the event logging.

julianedwards (approved):

PR #5234, opened 2021-11-29
EVG-15768 display error when project identifier already exists

ablack12:
[EVG-15768 ](https://jira.mongodb.org/browse/EVG-15768)

### Description 
Could make this change more robust but since this is legacy and is mostly just done by our team I think a small fix is best.

### Testing 
Changed logkeeper project to "logkeeper2" and verified that trying to create a new logkeeper returned this error.
image

### DIFF
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -228,7 +228,7 @@ mciModule.controller(
 
     $scope.findProject = function (identifier) {
       return _.find($scope.trackedProjects, function (project) {
-        return project.identifier == identifier;
+        return project.identifier == identifier || project.id == identifier;
       });
     };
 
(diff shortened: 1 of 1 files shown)


public/static/js/projects.js
```diff
@@ -228,7 +228,7 @@ mciModule.controller(
 
     $scope.findProject = function (identifier) {
       return _.find($scope.trackedProjects, function (project) {
-        return project.identifier == identifier;
+        return project.identifier == identifier || project.id == identifier;
```
malikchaya2, line 231: Why is it a problem if someone wants to use an id that is being used as an identifier somewhere? Is the concern that it will end up being confusing?

malikchaya2 (changes requested):

public/static/js/projects.js
ablack12: It isn't a concern, it's something we enforce throughout the codebase ([ById](https://github.com/evergreen-ci/evergreen/blob/0e8564e68d1eff7e6b27b99d6aab9c354b3478e7/model/project_ref.go#L1111) assumes this even).  If the user passes in identifier "evergreen" for example, but we have a project with identifier "evergreen" and ID "evergreen", how do we know which one to give them? For new projects IDs are random hashes so it's unlikely that a project would have that ID, but for old projects that's not true (and also users have the option to set the ID if they choose). Because of this, we need all IDs and identifiers to be unique.

PR #5237, opened 2021-11-30
EVG-15517: Return 200 when no test results found in cedar

julianedwards:
EVG-15517: https://jira.mongodb.org/browse/EVG-15517

This is a weird edge case where if the first execution of a task has a system or setup failure (and thus no results are produced) and a second execution successfully produces and sends test results to cedar, the `rest/v2/tasks/{task_id}/tests` route returns a 500 for the first execution because it thinks there are results in cedar. The old task will correctly indicate no cedar test results, but the project context always grabs the latest task—it seemed easier to just check the status rather than fetch the older task since this is a rare situation.

I tested this in staging by aborting a running task before it was able to send test results to cedar and rerunning it so the second execution has test results in cedar. The rest route now behaves as expected.


apimodels/test_results.go
```diff
@@ -113,29 +130,47 @@ func GetCedarTestResults(ctx context.Context, opts GetCedarTestResultsOptions) (
...
+	data, status, err := testresults.Get(ctx, timberOpts)
 	if err != nil {
-		return nil, errors.Wrap(err, "getting test results stats from Cedar")
+		return nil, 0, errors.Wrap(err, "getting test results stats from Cedar")
 	}
 
 	stats := &CedarTestResultsStats{}
-	if err = json.Unmarshal(data, stats); err != nil {
-		return nil, errors.Wrap(err, "unmarshaling test results stats from cedar")
+	if status == http.StatusOK {
```
Kimchelly, line 154: nit: same suggestion as above about inverting this check.

apimodels/test_results.go
```diff
@@ -113,29 +130,47 @@ func GetCedarTestResults(ctx context.Context, opts GetCedarTestResultsOptions) (
...
+		if err = json.Unmarshal(data, stats); err != nil {
+			return nil, 0, errors.Wrap(err, "unmarshaling test results stats from Cedar")
+		}
+	}
+
+	return stats, status, nil
+}
+
+// GetCedarTestResultsStatsWithStatusError  makes a request to Cedar for a
```
Kimchelly, line 163: nit: remove extra space.

apimodels/test_results.go
```diff
@@ -144,17 +179,35 @@ func GetCedarTestResultsStats(ctx context.Context, opts GetCedarTestResultsOptio
...
+	data, status, err := testresults.Get(ctx, timberOpts)
 	if err != nil {
-		return nil, errors.Wrap(err, "getting failed test result sample from Cedar")
+		return nil, 0, errors.Wrap(err, "getting failed test result sample from Cedar")
 	}
 
 	sample := []string{}
-	if err = json.Unmarshal(data, &sample); err != nil {
-		return nil, errors.Wrap(err, "unmarshaling failed test result sample from Cedar")
+	if status == http.StatusOK {
```
Kimchelly, line 191: nit: same thing as above.

rest/route/test.go
```diff
@@ -106,10 +106,13 @@ func (tgh *testGetHandler) Run(ctx context.Context) gimlet.Responder {
 			Limit:       tgh.limit,
 			Page:        page,
 		}
-		cedarTestResults, err := apimodels.GetCedarTestResults(ctx, opts)
+		cedarTestResults, status, err := apimodels.GetCedarTestResults(ctx, opts)
 		if err != nil {
 			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "getting test results"))
 		}
+		if status != http.StatusOK && status != http.StatusNotFound {
```
Kimchelly, line 113: Should this propagate the HTTP status code from Cedar to the user rather than unconditionally return 500?

Kimchelly (changes requested):
I'm having some difficulty understanding how this can happen. Is the problem that that `/rest/v2/tasks/{task_id}/tests` has a project context middleware that always uses the latest task ID, even if you pass a `{task_id}` parameter of an older task execution? Or something else?

rest/route/test.go
julianedwards: Yeah I went back and forth about this, but it seems like its still an internal server error since it is failing to connect to a third-party service. I am not too opinionated on this though.

Kimchelly (approved):

PR #5243, opened 2021-12-02
EVG 15901 Skip BFG creations if all subtask failures are setup-failed or stranded

Chukwurado:
[EVG-15901](https://jira.mongodb.org/browse/EVG-<number>)

### Description 
This change prevents BFGs from being created if all subtasks failures are identified as stranded or setup-failed.

### DIFF
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -355,7 +355,7 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					if executionTask.Details.Description == evergreen.TaskDescriptionStranded {
+					if executionTask.Details.Description == evergreen.TaskDescriptionStranded || executionTask.Details.Description == evergreen.TaskSetupFailed {
 						shouldSkipTicket = true
 					} else {
 						shouldSkipTicket = false
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -345,7 +345,7 @@ func (t *taskTriggers) makeData(sub *event.Subscription, pastTenseOverride, test
 func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, testNames string) (*notification.Notification, error) {
 	var payload interface{}
 	if sub.Subscriber.Type == event.JIRAIssueSubscriberType {
-		// We avoid creating BFG ticket in the case that the task is stranded to reduce noise for the Build Baron
+		// We avoid creating BFG ticket in the case that the task is setup-failed or stranded to reduce noise for the Build Baron
 		// If task is display, we skip ticket creation if all execution task failures are only 'stranded'
 		shouldSkipTicket := false
 		if t.task.DisplayOnly {
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -355,13 +355,10 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					skipTicket := executionTask.Details.Description == evergreen.TaskDescriptionStranded ||
+					shouldSkipTicket := executionTask.Details.Description == evergreen.TaskDescriptionStranded ||
 						executionTask.Details.Type == evergreen.CommandTypeSetup ||
 						executionTask.IsSystemUnresponsive()
-					if skipTicket {
-						shouldSkipTicket = true
-					} else {
-						shouldSkipTicket = false
+					if !shouldSkipTicket {
 						break
 					}
 				}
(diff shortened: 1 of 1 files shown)


ablack12 (commented):
Leaving the review for Malik, but can you re-word the PR title to start with an imperative verb (explained more [here](https://docs.google.com/document/d/1LJHPjLb_bUZxv-l_9Y61kitzAg7u-13I5DjOhZ-deBA/edit#heading=h.fua24zxppw7l)? for example, "skip BFG creation if ...."

trigger/task.go
```diff
@@ -355,7 +355,7 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					if executionTask.Details.Description == evergreen.TaskDescriptionStranded {
+					if executionTask.Details.Description == evergreen.TaskDescriptionStranded || executionTask.Details.Description == evergreen.TaskSetupFailed {
```
hadjri, line 358: We do not set setup-failed as the description of the task details, in this way it is unlike the taskDescriptionStranded property.  It is set as the task's display status, see here : https://evergreen.mongodb.com/api/rest/v2/tasks/mongodb_mongo_v5.1_enterprise_windows_powercycle_replication_smalloplog_0_enterprise_windows_b21f0146559bab88eb51d73dff6f896cf8bc5a9c_21_11_17_21_10_33?execution=1

hadjri (changes requested):

trigger/task.go
hadjri: You can check the codebase for instances where we apply evergreen.TaskSetupFailed to a task.

trigger/task.go
```diff
@@ -355,7 +355,7 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					if executionTask.Details.Description == evergreen.TaskDescriptionStranded {
+					if executionTask.Details.Description == evergreen.TaskDescriptionStranded || executionTask.GetDisplayStatus() == evergreen.TaskSetupFailed {
```
ablack12, line 358: GetDisplayStatus does a lot of work which we don't actually need to do; this could instead be handled by just checking `t.Details.Type == evergreen.CommandTypeSetup` (which is the step inside GetDisplayStatus that returns task setup failed). This is how we handle this elsewhere for task triggers https://github.com/evergreen-ci/evergreen/blob/c2fc7132326fa66cf2be16bcc719fc934580cb06/trigger/task.go#L475.
We should probably also skip the ticket if IsSystemUnresponsive is true.

trigger/task.go
```diff
@@ -355,7 +355,10 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					if executionTask.Details.Description == evergreen.TaskDescriptionStranded {
+					skipTicket := executionTask.Details.Description == evergreen.TaskDescriptionStranded ||
```
ablack12, line 358: can set shouldSkipTicket here (instead of creating an equivalent `skipTicket` variable) and then just break if shouldSkipTicket is false

Chukwurado: Oh yeah that is much cleaner, thanks!

trigger/task.go
```diff
@@ -355,10 +355,10 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
 					return nil, errors.Wrapf(err, "error getting execution task")
 				}
 				if executionTask.Details.Status == evergreen.TaskFailed {
-					if executionTask.Details.Description == evergreen.TaskDescriptionStranded {
-						shouldSkipTicket = true
-					} else {
-						shouldSkipTicket = false
+					shouldSkipTicket := executionTask.Details.Description == evergreen.TaskDescriptionStranded ||
```
ablack12, line 358: nit: should use = instead of := in this case; by using := you're creating a new `shouldSkipTicket` variable that only exists within the if statement, so the variable defined at 350 and checked at 369 will still be false. By using = you will instead assign the value to the variable defined at 350.

trigger/task.go
```diff
@@ -355,10 +355,10 @@ func (t *taskTriggers) generate(sub *event.Subscription, pastTenseOverride, test
...
+						executionTask.Details.Type == evergreen.CommandTypeSetup ||
+						executionTask.IsSystemUnresponsive()
```
hadjri, line 360: I don't think we need to check IsSystemUnresponsive; checking the details type should be good enough to determine if the display status was set to setup-failed.

hadjri (changes requested):

trigger/task.go
ablack12: IsSystemUnresponsive checks a different symptom; i recommended that he add it here https://github.com/evergreen-ci/evergreen/pull/5243#discussion_r762282172

ablack12 (approved):

hadjri (approved):

PR #5245, opened 2021-12-03
EVG 15012 Return warning errors if tasks have no timeout

Chukwurado:
[EVG-15012](https://jira.mongodb.org/browse/EVG-15012)

### Description 
This PR contains a function that returns warning validation errors if not all tasks have timeouts set.

### Testing
Modified some unit tests in `project_validator_test.go`

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -131,6 +131,7 @@ var projectSemanticValidators = []projectValidator{
 	checkTaskCommands,
 	checkTaskGroups,
 	checkLoggerConfig,
+	checkTaskTimeout,
 }
 
 var projectSettingsValidators = []projectSettingsValidator{
@@ -1897,4 +1898,22 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
 		return "", "", errors.Errorf("command '%s' was supplied parameter '%s' but is not a string argument, got %T", c.Command, paramName, i)
 	}
 	return task, bv, nil
+}
+
+// checkTaskTimeout checks if all tasks contains a timeout
+func checkTaskTimeout(project *model.Project) ValidationErrors {
+	errs := ValidationErrors{}
+	for _, task := range project.Tasks {
+		if task.ExecTimeoutSecs == 0 {
+			errs = append(errs,
+				ValidationError{
+					Message: fmt.Sprintf("task '%s' in project '%s' does not "+
+						"contain an exec_timeout_secs",
+						task.Name, project.Identifier),
+					Level: Warning,
+				},
+			)
+		}
+	}
+	return errs
 }
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -1911,11 +1911,12 @@ func checkTaskTimeout(project *model.Project) ValidationErrors {
 			errs = append(errs,
 				ValidationError{
 					Message: fmt.Sprintf("task '%s' in project '%s' does not "+
-						"have an exec_timeout_secs defined; the task will default to a timeout of 6 hours",
+						"have an exec_timeout_secs defined; the task will default to a timeout of 6 hours\n",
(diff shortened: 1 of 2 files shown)


hadjri:
Can you build this version of the CLI locally and test this with a config that should trigger these warnings and confirm that it works?

validator/project_validator.go
```diff
@@ -1898,3 +1899,21 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
 	}
 	return task, bv, nil
 }
+
+// checkTaskTimeout checks if all tasks contain a timeout
+func checkTaskTimeout(project *model.Project) ValidationErrors {
```
hadjri, line 1904: Some unit tests are gonna need to be updated to expect this warning now - a couple are failing.  Once that's done add that to the testing section of the description of this PR as well. Looks good though!

hadjri (changes requested):

validator/project_validator.go
```diff
@@ -1898,3 +1899,37 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
...
+					taskNames[0], project.Identifier),
+				Level: Warning,
+			},
+		)
+	} else if len(taskNames) > 1 {
+		errs = append(errs,
+			ValidationError{
+				Message: fmt.Sprintf("tasks '%s' in project '%s' do not "+
+					"have an exec_timeout_secs defined",
+					strings.Join(taskNames, ", "), project.Identifier),
```
hadjri, line 1929: There are some projects that might have this warning true for many many tasks in which case this could be a super long warning message - can you instead follow something like the message at this line? https://github.com/evergreen-ci/evergreen/blob/main/validator/project_validator.go#L567 Where we pick up if one or more tasks fit the warning criteria and then simply remind the user that tasks must have an exec_timeout_secs defined for them.  Can you also include part of the message to say what default timeout the tasks will fall back to if nothing is defined? (I believe its like 4 hours but not sure).

hadjri (changes requested):

validator/project_validator.go
```diff
@@ -1898,3 +1899,25 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
...
+func checkTaskTimeout(project *model.Project) ValidationErrors {
+	errs := ValidationErrors{}
+	if project.ExecTimeoutSecs > 0 {
+		return errs
+	}
+	for _, task := range project.Tasks {
+		if task.ExecTimeoutSecs == 0 {
+			errs = append(errs,
+				ValidationError{
+					Message: fmt.Sprintf("task '%s' in project '%s' does not "+
```
hadjri, line 1913: To avoid clutter we should just return a single warning if a config has any tasks without exec timeout secs defined, similar to how this message handles it: https://github.com/evergreen-ci/evergreen/blob/main/validator/project_validator.go#L567 Once we find a task without timeout secs defined we can append this warning and break the loop.

hadjri (changes requested):

validator/project_validator.go
```diff
@@ -1898,3 +1899,26 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
...
+						"have an exec_timeout_secs defined; the task will default to a timeout of 6 hours\n",
```
hadjri, line 1914: We should make this message more generic, i.e. something like "project %s has tasks without exec_timeout_secs defined; these tasks will default to a timeout of6 hours" since we are just checking if the project has ANY tasks without this specified to avoid spitting out a bunch of errors.

hadjri: Also I don't think we need the newline character - lets remove it. I dont see it on any other validation error messages.

hadjri (approved):
LGTM after my comment below is addressed!

PR #5246, opened 2021-12-03
EVG-15930: handle race between provisioning-create-host and agent

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15930

### Description
This is a follow on to [EVG-15022](https://jira.mongodb.org/browse/EVG-15022) (#5232).

The order of operations during host creation and agent startup _should_ be this:
provisioning-create-host job creates host that starts the agent on its own -> provisioning-create-host finishes -> agent checks in to next task
But in some cases appear to be this:
provisioning-create-host job creates host that starts the agent on its own -> agent checks in to next task -> provisioning-create-host finishes

The second order of operations means the provisioning-create-host job can race with the agent REST route to swap the intent host with the real host. particularly if the transaction takes a long time. I'm unsure why it takes a long time, but the Splunk logs say the transaction cases take over a minute to abort.

This issue should go away entirely with [EVG-15875](https://jira.mongodb.org/browse/EVG-15875), but in the meantime, this is a simple fix for an uncommon occurrence.

* Check if the intent host has already been swapped with the real host when host swapping fails in provisioning-create-host. If the real host already exists, then the swap succeeded and provisioning-create-host can proceed knowing the swap worked.
    * There is a minor issue with not doing the host swap in provisioning-create-host - the on-demand EC2 provider adds extra EC2 tag metadata to the in-memory host copy in `SpawnHost`. Since provisioning-create-host doesn't do the swap itself, these in-memory tags are never persisted in the DB. I don't think this is particularly important, but could be fixed if it becomes an issue. Otherwise, [EVG-15875](https://jira.mongodb.org/browse/EVG-15875) would fix this issue.
* Add logging for the host swap transaction duration. This is going to be removed at some point, but it's worth seeing some statistics on how long it takes to perform.

### Testing 
Not really testable since the provisioning-create-host job is not tested, and it's difficult to simulate the race to swap the host document.

### DIFF
--- a/units/provisioning_create_host.go
+++ b/units/provisioning_create_host.go
@@ -372,26 +372,8 @@ func (j *createHostJob) createHost(ctx context.Context) error {
 		}
 	} else {
 		// For most cases, spawning a host will change the ID, so we remove/re-insert the document
-		if err := host.UnsafeReplace(ctx, j.env, j.HostID, j.host); err != nil {
-			grip.Warning(message.WrapError(err, message.Fields{
-				"message":        "could not perform unsafe host replacement",
-				"job":            j.ID(),
-				"intent_host_id": j.HostID,
-				"host_id":        j.host.Id,
-				"distro":         j.host.Distro.Id,
-				"provider":       j.host.Provider,
-			}))
-			terminateCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-			defer cancel()
-			grip.Error(message.WrapError(cloudManager.TerminateInstance(terminateCtx, j.host, evergreen.User, "hit database error trying to update host"), message.Fields{
-				"message":        "could not terminate host after failed unsafe host replacement",
-				"host_id":        j.host.Id,
-				"intent_host_id": j.HostID,
-				"distro":         j.host.Distro.Id,
-				"provider":       j.host.Provider,
-				"job":            j.ID(),
-			}))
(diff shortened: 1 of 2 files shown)


units/provisioning_create_host.go
```diff
@@ -463,3 +445,53 @@ func (j *createHostJob) isImageBuilt(ctx context.Context) (bool, error) {
...
+// real host that was created by this job. If that fails, it checks to see if
+// the real host already exists. If both of those fail, it will attempt to
+// terminate the cloud host.
+func (j *createHostJob) tryHostReplacement(ctx context.Context, cloudMgr cloud.Manager) error {
+	err := host.UnsafeReplace(ctx, j.env, j.HostID, j.host)
+	if err == nil {
+		return nil
+	}
+
+	grip.Warning(message.WrapError(err, message.Fields{
```
julianedwards, line 459: Should the warning message be sent after we check for the host? It seems if the host was already successfully swapped there is not point of adding a warning message.

Kimchelly: It's relevant to know when the host swap has failed in provisioning create host, which is the normal way to do the swap (doing it in the agent is the abnormal route). If it silently fails to swap due to this race without any logging, it'll be more difficult to debug. It's also an uncommon occurrence, so it won't log that much compared to all the other noisy logs that jobs have.

julianedwards (approved):

PR #5247, opened 2021-12-03
EVG-15049 add project/repo event log resolvers

ablack12:
[EVG-15049](https://jira.mongodb.org/browse/EVG-15049)

### Description 
Modeled these resolvers off of the host events resolver to allow for pagination.

### Testing 
Not sure how intense the tests need to be since we test ProjectSettings / RepoSettings extensively elsewhere but I added simple tests.


graphql/schema.graphql
```diff
@@ -1081,21 +1091,44 @@ type GroupedProjects {
 }
 
 type ProjectSettings {
-  githubWebhooksEnabled: Boolean!
+  gitHubWebhooksEnabled: Boolean!
```
khelif96, line 1094: This change makes this a breaking change. As long as it isn't being used in production it should be fine.

khelif96 (approved):
LGTM. Only needs changes if the comment below is true and it is being used in production.

graphql/schema.graphql
ablack12: Not being used in production so should be fine 👍

graphql/tests/repoEvents/data.json
```diff
@@ -0,0 +1,53 @@
...
+    }
+  ],
+  "project_vars": [
+    {
+    "_id": "5678",
+    "vars": {"hello": "world", "foo":  "bar"},
+    "private_vars": {"hello":  true, "foo":  false}
+    }
+  ],
+  "event_log": [
```
sophstad, line 14: It would be helpful to add this data to the repo_ref file in testdata for local development!

sophstad (approved):

PR #5249, opened 2021-12-07
EVG-16018 Add built-in command to sign, notarize macOS artifacts

turalf:
### Description 
This PR adds a built in evergreen command to automatically sign and notarize macOS artifacts by talking the internal macOS signing and notarization service.
### Testing 
Unit and manual end to end tests have been performed on these changes. 
###
Documentation to this command is to be added on the next PR.


agent/command/macsign.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return macSign.validate()
+}
+
+func (macSign *macSign) validate() error {
+	catcher := grip.NewSimpleCatcher()
+
+	// make sure the command params are valid
+	if macSign.KeyId == "" {
+		catcher.Add(errors.New("key_id cannot be blank"))
```
ablack12, line 95: you can use catcher.New("string") instead

agent/command/macsign.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return catcher.Resolve()
+}
+
+// Apply the expansions from the relevant task config (including restricted expansion)
+// to all appropriate fields of the macSign.
+func (macSign *macSign) expandParams(conf *internal.TaskConfig) error {
+	var err error
+	if err = util.ExpandValues(macSign, conf.GetExpansionsWithRestricted()); err != nil {
```
ablack12, line 126: Please use conf.Expansions; GetExpansionsWithRestricted is something special but also we're planning to remove it.

agent/command/macsign.go
```diff
@@ -0,0 +1,209 @@
...
+	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
+
+	if !macSign.shouldRunForVariant(conf.BuildVariant.Name) {
+		logger.Task().Infof("Skipping macsign of local file %v for variant %v",
+			macSign.LocalZipFile, conf.BuildVariant.Name)
+		return nil
+	}
+
+	// expand necessary params
+	if err := macSign.expandParams(conf); err != nil {
```
ablack12, line 161: Should you expandParams before checking if you should skip the variant, just in case something needs to be populated?

agent/command/macsign.go
```diff
@@ -0,0 +1,209 @@
...
+		exitCode = exitErr.ExitCode()
+	}
+
+	if exitCode != 0 {
+		return fmt.Errorf("none zero exit code: %d: \n%s", exitCode, output)
+	}
+
+	logger.Task().Info(output)
+
+	logger.Task().Infof("Artifact - %s signed(and/or notarized) and new file created: %s", macSign.LocalZipFile, macSign.OutputZipFile)
```
ablack12, line 206: nit: missing space

agent/command/macsign_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/evergreen-ci/evergreen/model"
+	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/evergreen/util"
+	. "github.com/smartystreets/goconvey/convey"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMacSignValidateParams(t *testing.T) {
+
+	Convey("With a mac signing and notarization command", t, func() {
```
ablack12, line 21: For future reference (if ever haha) we don't actually use Convey for testing anymore; instead we use assert / require. You don't have to fix this if you don't want to though since it's already done with Convey now.

ablack12 (changes requested):

agent/command/macsign.go
```diff
@@ -0,0 +1,209 @@
...
+		catcher.New(`artifact_type needs to be either blank, "binary" or "app"`)
+	}
+	if macSign.Notarize && macSign.BundleId == "" {
+		catcher.New("if notarization is requested, bundle_id cannot be blank")
+	}
+
+	return catcher.Resolve()
+}
+
+// Apply the expansions from the relevant task config (including restricted expansion)
```
ablack12, line 122: Should remove (including restricted expansions) from this comment

agent/command/macsign.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if macSign.ServiceUrl == "" {
+		catcher.New("service_url cannot be blank")
+	}
+	if runtime.GOOS != "darwin" {
+		// do not fail, just set verifying to false.
+		macSign.Verify = false
+	}
+	if !(macSign.ArtifactType == "" || macSign.ArtifactType == "binary" || macSign.ArtifactType == "app") {
+		catcher.New(`artifact_type needs to be either blank, "binary" or "app"`)
```
ablack12, line 113: consistency nit: we typically still use "" as outer quotes, and ' ' as inner quotes

agent/command/macsign.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	if !macSign.shouldRunForVariant(conf.BuildVariant.Name) {
+		logger.Task().Infof("Skipping macsign of local file %s for variant %s",
+			macSign.LocalZipFile, conf.BuildVariant.Name)
+		return nil
+	}
+
+	// client binary
+	if macSign.ClientBinary == "" {
```
ablack12, line 180: Does it make sense to include this defaulting in in ExpandParams with the other defaults?

agent/command/macsign.go
```diff
@@ -0,0 +1,232 @@
...
+
+	var exitCode int
+
+	stdout, err := cmd.CombinedOutput()
+	output := string(stdout)
+
+	if err != nil {
+		exitErr, ok := err.(*exec.ExitError)
+		if !ok {
+			return fmt.Errorf("unexpected error %s: \n%s\n%s\n%s", err, output, runtime.GOOS, runtime.GOARCH)
```
ablack12, line 217: preference for errors.Wrapf(err, "message")

agent/command/macsign.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		exitCode = exitErr.ExitCode()
+	}
+
+	if exitCode != 0 {
+		return fmt.Errorf("none zero exit code: %d: \n%s", exitCode, output)
+	}
+
+	logger.Task().Info(output)
```
ablack12, line 227: Is this output likely to be long/useful? I think it may be better to log this under debugging, or as logger.Execution() (i.e. under the agent logs)

agent/command/macsign.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		exitErr, ok := err.(*exec.ExitError)
+		if !ok {
+			return fmt.Errorf("unexpected error %s: \n%s\n%s\n%s", err, output, runtime.GOOS, runtime.GOARCH)
+		}
+
+		exitCode = exitErr.ExitCode()
+	}
+
+	if exitCode != 0 {
```
ablack12, line 223: Is there a reason to log this separately from 217? Will output contain the error instead of err?

ablack12 (commented):
A couple of comments but generally LGTM; if this is something Evergreen is going to maintain in the future, it would help to put comments above each of the defaults and for each of the flags to give context. Otherwise I think when adding to the documentation https://github.com/evergreen-ci/evergreen/wiki/Project-Commands, maybe add a note that this command is maintained by the Build Team or something similar. (Also could you ping me after adding to the documentation just so i can check it over?)

agent/command/macsign.go
turalf: line 217 is the logging when the error is not an exit error. and this one is the logging when the error is indeed exit error but not zero. That would be useful information when debugging. e.g. if exit code is none-zero that means a problem is related to executable binary, otherwise the problem is different(hopefully error message would describe it).

agent/command/macsign.go
turalf: the output of that message will contain an information about the download url of the signed artifact(along with other useful information such as request Id of the service). This would be useful, to download the signed artifact later on or ask macos notary service owners what was the problem with the specific requestId.

turalf:
> A couple of comments but generally LGTM; if this is something Evergreen is going to maintain in the future, it would help to put comments above each of the defaults and for each of the flags to give context. Otherwise I think when adding to the documentation https://github.com/evergreen-ci/evergreen/wiki/Project-Commands, maybe add a note that this command is maintained by the Build Team or something similar. (Also could you ping me after adding to the documentation just so i can check it over?)

Make sense. I will add a documentation to the command. I will make a note in this PR that it is going to be added on the next one.

agent/command/macsign.go
ablack12: Gotcha; should 225 also log the error then, or is it expected that an exit error would have the error be in the output?

agent/command/macsign.go
ablack12: can this be fixed?

ablack12 (approved):
LGTM with nit

turalf:
https://github.com/evergreen-ci/evergreen/pull/5249#discussion_r774015468
Yes, the output would say why the error code is not zero.

PR #5250, opened 2021-12-10
EVG-15706: Remove all amboy remote management flag code

hadjri:
[EVG-15706](https://jira.mongodb.org/browse/EVG-15706)

### Description 
Removed all code related to the amboy remote management flag and the flag itself since it is not used.
### Testing 
Tested UI locally.


ablack12 (changes requested):
The changes made here are flipped, I believe. We want to leave Amboy Remote Management purposefully disabled; as such we are never running the code in the `if !AmboyRemoteManagementDisabled` and we never should.

hadjri:
> The changes made here are flipped, I believe. We want to leave Amboy Remote Management purposefully disabled; as such we are never running the code in the `if !AmboyRemoteManagementDisabled` and we never should.

My mistake - got confused because the flag name in the UI is just "Amboy Remote Management" which I had assumed meant the value was false, not true

units/stats_amboy.go
```diff
@@ -92,16 +91,6 @@ func (j *amboyStatsCollector) Run(ctx context.Context) {
 			"message": "amboy remote queue stats",
 			"stats":   remoteQueue.Stats(ctx),
 		})
-
-		serviceFlags, err := evergreen.GetServiceFlags()
```
ablack12, line 96: It doesn't look like this job really does anything anymore; we'll just be logging work that isn't actually being done. We should probably just remove the whole job.

hadjri: Makes sense, removing

ablack12 (approved):

PR #5251, opened 2021-12-13
EVG-15924 Remove s3-fake-destination-optional

malikchaya2:
[EVG-15924](https://jira.mongodb.org/browse/EVG-15924)

### Description 
This was accidentally committed to main 

### Testing 
Created a [patch](https://spruce.mongodb.com/patch/61b70caec9ec4412edcc263f/configure/tasks). s3-fake-destination-optional is no longer there, but docker-cleanup is .

### DIFF
--- a/self-tests.yml
+++ b/self-tests.yml
@@ -339,17 +339,6 @@ functions:
           set -o xtrace
           rm evergreen/bin/output.*
           rm evergreen/bin/jstests/*.xml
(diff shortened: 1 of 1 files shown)


ablack12 (approved):
LGTM; can you change the PR title to be "remove s3-fake-etc" or something like that so it meets guidelines?

PR #5254, opened 2021-12-13
EVG-15734: Upgrade Docker to v20.10.11

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15734

### Description 
Upgrade Docker to the latest tagged version and bump Jasper to be compatible with the latest version. Ordinarily, a Go module would allow multiple major versions of a transient Go module dependency (e.g. this repo can use v1 of github.com/urfave/cli but dependent Go modules can use github.com/urfave/cli/v2 without conflicting with this repo). However, because Docker does not use Go modules, we cannot have multiple different major versions of the Docker SDK (hence why the Go modules has the `+incompatible` tag on it), so Evergreen and all its dependencies must agree to use Docker v20 now.

### Testing 
Verified in staging that the container building process still works.

### DIFF
--- a/cloud/docker_client.go
+++ b/cloud/docker_client.go
@@ -1,6 +1,3 @@
-//go:build go1.7
-// +build go1.7
-
 package cloud
 
 import (
--- a/cloud/docker_client.go
+++ b/cloud/docker_client.go
@@ -1,3 +1,4 @@
+//go:build go1.7
 // +build go1.7
 
 package cloud
@@ -27,6 +28,7 @@ import (
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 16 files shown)


cloud/docker_client.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.7
 // +build go1.7
```
julianedwards, line 2: Do we need both of these lines?

Kimchelly: This is [the new syntax for build tags](https://pkg.go.dev/go/build@go1.17#hdr-Build_Constraints) as of go1.17, so `gofmt` auto-added it. It's not really a problem since go1.16 or below would just ignore this line. None of these tags should be necessary since we've been building with a version of go above 1.7 for a long while now, so I just removed all of them.

Kimchelly:
I tested the container building process in staging. It seems like both staging and prod are pinned to use a Docker client API version below the minimum necessary to pass those options (1.37).

I changed the `ContainerCreate` call to pass `nil` for the platform options because those options are only available for a particular minimum Docker API version (and are not required, as far as I can tell).

julianedwards (approved):

PR #5257, opened 2021-12-15
EVG-13397 Add warning validation errors for modules with duplicate names

Chukwurado:
[EVG-13397](https://jira.mongodb.org/browse/EVG-13397)

### Description 
This PR updates the warning message for modules with duplicate names


validator/project_validator.go
```diff
@@ -1898,3 +1899,21 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
 	}
 	return task, bv, nil
 }
+
+func checkDuplicateModuleNames(project *model.Project) ValidationErrors {
```
hadjri, line 1903: Turns out this ticket is rather old and we've since already introduced a warning to cover this!! Sorry for not seeing this earlier: https://github.com/evergreen-ci/evergreen/blob/main/validator/project_validator.go#L670 So this new function here isn't necessary anymore.  However what I think we should do is specify which module with the duplicate name will actually be used when we look up a project's modules by name.  Is it the first one, or the last one in the list? We should figure that out and then enhance the existing warning with those details.

hadjri (changes requested):

hadjri (approved):

PR #5258, opened 2021-12-15
EVG-15706 Bring back stats_amboy job

hadjri:
[EVG-15706](https://jira.mongodb.org/browse/EVG-15706)

### Description 
Bring back amboy stats job to populate amboy splunk dashboards.


ablack12 (approved):

Kimchelly:
I don't think this is a big deal, but I think adding a new file like this rather than reverting + applying the desired change deletes the original git blame on the `units/stats_amboy.go` file. Is it worth reverting + applying the relevant change diff so that the blame history gets preserved?

ablack12:
Fine with this either was -- since Github lets us walk back blames now I think it's not as important as it once was but it is easier to skip that step, so up to you Malik.

hadjri:
> Fine with this either was -- since Github lets us walk back blames now I think it's not as important as it once was but it is easier to skip that step, so up to you Malik.

Given this I think it's fine to merge.

Kimchelly:
My original point was that [you can no longer step through the blame history](https://github.com/evergreen-ci/evergreen/blame/main/units/stats_amboy.go), so anyone who reads this file has to know to go through the commit history to find the commit before the file deletion occurred in order to continue the blame. I guess this is alright for a one-time thing, but I wouldn't do it in the future when it comes to file deletions, since GitHub tends to treat the blame weirdly.

ablack12:
Ah good point Kim, yes we should consider that going forward.

PR #5261, opened 2021-12-15
EVG-15595: move globals collection and model into model package

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15595

### Description 
The db package, which contains the general DB utility helpers and Anser wrappers, also has a data model for global per-build-variant counters. Rather than keep it in the DB package, I moved it to a proper DB model package. I cleaned it up to simplify the logic in `GetNewBuildVariantBuildNumber`.

I'm not even sure that this code is necessary anymore (it populates a build variant model field, but I think that field is unused), but it might as well go in its own model package since that's how DB models are organized in the code base.

### Testing 
Added DB tests for the global package.


julianedwards (approved):

model/global/global.go
```diff
@@ -0,0 +1,12 @@
+package global
+
+// Global stores internal global tracking information for each build variant.
+type Global struct {
```
ablack12, line 4: Can this be called something less generic? A "global" collection implies this could eventually be used for other global things. That or maybe the collection name could be more specific, like "global_variant_data"

ablack12: Sorry I didn't see that this struct already exists 😭

ablack12 (commented):
I don't actually see that this is used anywhere but in this legacy route https://github.com/evergreen-ci/evergreen/blob/6ae840fc0e0b469dcef0d33569cf213bbb685199/service/rest.go#L75 so I think getting rid of this field entirely is really the solution to the tech debt rather than moving it, but we could make a new ticket for that if you prefer.

Kimchelly:
I usually hesitate to delete anything that's exposed in a public REST route, but if nobody's using it, then I'm fine with deletion.

I added a commit to delete the `BuildNumber` and the globals code.

ablack12:
> I usually hesitate to delete anything that's exposed in a public REST route, but if nobody's using it, then I'm fine with deletion.

Yeah I get that; I'm a bit torn since we don't know if anyone's relying on it. In some ways the only method for getting rid of bad legacy things without trying it and seeing if anyone notices. It looks like [this route is used|https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen%20rest%2Fv1%2Fbuilds&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-30d%40d&latest=now&workload_pool=standard_perf&sid=1640022565.1618585&display.prefs.events.offset=0] but I highly doubt they're using this part. I'm inclined to get rid of it and just see if anyone says anything, and/or if they are using it ask them to start using `order` instead.

ablack12 (approved):

Kimchelly:
How about an intermediate solution - if people are possibly using it, a safer method is to remove it from just the REST model for now, but keep the back-end code that populates the build number. Then, if nobody complains, we can remove the rest of the global build number code after a few months. That way, if someone really needs the build number and we have to revert, there's no risk that doing so will produce an inconsistent DB state (e.g. some builds are missing a build number).

ablack12:
Sure that's a good idea!

ablack12 (approved):
LGTM!

PR #5262, opened 2021-12-15
EVG-15918 Mongodb test filtering for taskTestSample resolver

khelif96:
[EVG-15918](https://jira.mongodb.org/browse/EVG-15918)

### Description 
This adds test filtering for test results stored in the db.
### Testing 
  On staging and unit tests


graphql/resolvers.go
```diff
@@ -1850,10 +1851,48 @@ func (r *queryResolver) TaskTestSample(ctx context.Context, tasks []string, test
 			}
 			testResultsToReturn = append(testResultsToReturn, tr)
 		}
-		return testResultsToReturn, nil
+	} else {
+		regexFilter := ""
+		for i, f := range testFilters {
+			regexFilter += f.TestName
+			// If there is another test filter add a pipe to separate them
+			if i < len(testFilters)-1 {
```
ablack12, line 1859: I think a slightly cleaner way to do this is add it at the beginning of the loop if `i != 0` but both work

graphql/resolvers.go
```diff
@@ -1822,13 +1822,15 @@ func (r *queryResolver) TaskTestSample(ctx context.Context, tasks []string, test
 	if len(tasks) == 0 {
 		return nil, nil
 	}
-	t, err := task.Find(task.ByIds(tasks))
+	t, err := task.FindAll(task.ByIds(tasks))
```
ablack12, line 1825: "t" implies one task, not many. Can we rename this to tasks? (Also helps us avoid the `for t := range t` thing below)

graphql/resolvers.go
```diff
@@ -1850,10 +1851,48 @@ func (r *queryResolver) TaskTestSample(ctx context.Context, tasks []string, test
...
+			filteredTestResults, err := r.sc.FindTestsByTaskId(data.FindTestsByTaskIdOpts{
+				TaskID:    t.Id,
+				Execution: t.Execution,
+				TestName:  regexFilter,
+				Statuses:  []string{evergreen.TestFailedStatus},
+				Limit:     10,
+				Page:      0,
+			})
+			if err != nil {
+				return nil, ResourceNotFound.Send(ctx, err.Error())
```
ablack12, line 1873: Shouldn't this be InternalServerError? I don't think it should be an error if no tests are found.

ablack12 (changes requested):

graphql/resolvers.go
khelif96: I'm not sure if we can since we only want to add this condition if we know there is gonna be another filter proceeding the current one.

ablack12: It works since you add it to the top of the loop, since you immediately add the testname after:
```
for i, f := range testFilters {
    if i != 0 {  // add pipe }
    regexFilter := f.TestName
}
```

ablack12 (approved):
LGTM though, the loop thing is optional

graphql/resolvers.go
```diff
@@ -1850,10 +1851,48 @@ func (r *queryResolver) TaskTestSample(ctx context.Context, tasks []string, test
...
+			if i < len(testFilters)-1 {
+				regexFilter += "|"
+			}
+		}
+		for _, t := range dbTasks {
+			filteredTestResults, err := r.sc.FindTestsByTaskId(data.FindTestsByTaskIdOpts{
+				TaskID:    t.Id,
+				Execution: t.Execution,
+				TestName:  regexFilter,
+				Statuses:  []string{evergreen.TestFailedStatus},
```
julianedwards, line 1868: If we only ever filter on failed tests, can we just pass in a `[]string` of test name filters instead of `[]TestFilter`?

khelif96: This is meant to eventually support filtering on any status for different filters. So i opted to leave TestFilter as a struct. Since we will eventually need to add it. Leaving this as a struct will allow us to be able to make a non breaking change to the api when we do update this.

PR #5263, opened 2021-12-16
EVG-15919 update admins in project ref only if successful

ablack12:
[EVG-15919](https://jira.mongodb.org/browse/EVG-15919)

### Description 
We partially handled this in https://jira.mongodb.org/browse/EVG-15428 but we didn't handle the case where we exit before even trying to add/remove admins.

### Testing 
Added a case to the test.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -382,7 +382,8 @@ func (p *ProjectRef) Add(creator *user.DBUser) error {
 				return errors.Wrapf(err, "error upserting project ref '%s'", hidden.Id)
 			}
 			if creator != nil {
-				return p.UpdateAdminRoles([]string{creator.Id}, nil)
+				_, err = p.UpdateAdminRoles([]string{creator.Id}, nil)
+				return err
 			}
 			return nil
 		}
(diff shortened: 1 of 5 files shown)


model/project_ref.go
```diff
@@ -1910,24 +1911,25 @@ func (p *ProjectRef) MakeUnrestricted() error {
 	return nil
 }
 
-func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) error {
```
hadjri, line 1913: Why is this being done for projectref only and not reporef?

ablack12: If you take a look at repo ref, we don't have to look up what the role is because we already know the structure of it, so it's not possible to have an error before we get to the loop which handles it more dynamically.

malikchaya2 (commented):
Did you end up creating a ticket for the admin role not being added when usersettings is set to false? If not, it might be convenient to add the fix here? 

But it's up to you, a separate PR works just as well. Approving this because this fix looks good.

malikchaya2 (approved):

ablack12:
> Did you end up creating a ticket for the admin role not being added when usersettings is set to false? If not, it might be convenient to add the fix here?

I actually don't remember what this is referring to 😅  Let's sync on this when you're back? Maybe we do need a new ticket.

sophstad (approved):
Tested using the new front end page, LGTM!

PR #5264, opened 2021-12-17
EVG-15979 redact private variables in the UI

ablack12:
[EVG-15979](https://jira.mongodb.org/browse/EVG-15979)

### Description 
After discussion with Sophie, determined that private variables should be redacted before sending to the UI, and the backend save function should ensure that these are still saved correctly. This assumes that private variable values cannot be modified on the UI.

To do this, changed the UpdateProjectVars function to overwrite the existing variables object (i.e. using Upsert instead of FindAndModify) bc since the UI is sending the new variables and not the diff this is actually easier to reason about.

### Testing 
Added to the frontend and backend tests.

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -142,15 +142,14 @@ func (sc *DBConnector) SaveProjectSettingsForSection(ctx context.Context, projec
 			}
 		}
 	case model.ProjectPageVariablesSection:
-		// Remove any variables that only exist in the original settings.
-		toDelete := []string{}
-		for key, _ := range before.Vars.Vars {
-			if _, ok := changes.Vars.Vars[key]; !ok {
-				toDelete = append(toDelete, key)
+		for key, value := range before.Vars.Vars {
+			// Private variables are redacted in the UI, so re-set to the real value
+			// before updating (assuming the value isn't deleted).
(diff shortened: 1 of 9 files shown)


rest/data/project_settings.go
```diff
@@ -142,15 +142,14 @@ func (sc *DBConnector) SaveProjectSettingsForSection(ctx context.Context, projec
 			}
 		}
 	case model.ProjectPageVariablesSection:
-		// Remove any variables that only exist in the original settings.
-		toDelete := []string{}
-		for key, _ := range before.Vars.Vars {
-			if _, ok := changes.Vars.Vars[key]; !ok {
-				toDelete = append(toDelete, key)
+		for key, value := range before.Vars.Vars {
```
malikchaya2, line 145: what would happen here in the case where the value was deleted?

malikchaya2 (changes requested):
leaving one small question

rest/data/project_settings.go
ablack12: Yep! changes reflects the current state of the project, and if the variable is deleted from vars it should also be deleted from private vars. (Checking changes.Vars.Vars isn't quite as clear because the value has been made empty in order to redact the private value)

malikchaya2 (approved):

PR #5265, opened 2021-12-17
EVG-15656 add gated flag to allow specifying cq patch author

ablack12:
[EVG-15656](https://jira.mongodb.org/browse/EVG-15656)

### Description 
There's a lot of detail in the ticket but this adds a flag so that the server bot that creates commit queue patches on behalf of server engineers can specify github usernames, so that the patch will correctly be linked to the real author. 

Because this is a strong ability, I'll create a new bot for the server team to use that will have this permission. If the user doesn't have their github username on their profile, the patch will be listed as authored by "Commit Queue Bot User" or something like that, rather than Ethan Zhang, so still a step up.

### Testing 
I created this patch on Brian's behalf. https://spruce-staging.corp.mongodb.com/version/61bcf01697b1d3404fe84ca3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
He also still received notifications from it.

### DIFF
--- a/service/api_patch.go
+++ b/service/api_patch.go
@@ -113,10 +113,6 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
 
 	patchID := mgobson.NewObjectId()
 	author := dbUser.Id
-	grip.Debug(message.Fields{
-		"message":       "testing user",
-		"github_author": data.GithubAuthor,
-	})
 	if data.GithubAuthor != "" {
 		opts := gimlet.PermissionOpts{
 			Resource:      pref.Id,
@@ -131,6 +127,7 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
 		specifiedUser, err := user.FindByGithubName(data.GithubAuthor)
 		if err != nil {
 			as.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "error looking for specified author"))
+			return
 		}
 		if specifiedUser != nil {
 			grip.Info(message.Fields{
--- a/service/api_patch.go
+++ b/service/api_patch.go
@@ -122,7 +122,7 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
 			Resource:      pref.Id,
 			ResourceType:  evergreen.ProjectResourceType,
 			Permission:    evergreen.PermissionPatches,
-			RequiredLevel: evergreen.PatchSubmitBot.Value,
+			RequiredLevel: evergreen.PatchSubmitAdmin.Value,
 		}
 		if !dbUser.HasPermission(opts) {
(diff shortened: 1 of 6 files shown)


service/api_patch.go
```diff
@@ -109,8 +111,46 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
...
+			grip.Info(message.Fields{
+				"message":    "overriding patch author as specified by the submitter",
+				"submitter":  dbUser.Id,
+				"new_author": data.GithubAuthor,
+				"patch_id":   patchID,
+			})
+			author = specifiedUser.Id
+		}
+		grip.DebugWhen(specifiedUser == nil, message.Fields{
+			"message":         "github user not found",
```
hadjri, line 145: Should this be a LoggedError?

service/api_patch.go
```diff
@@ -109,8 +111,46 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	patchID := mgobson.NewObjectId()
+	author := dbUser.Id
+	grip.Debug(message.Fields{
```
malikchaya2, line 116: Is this leftover from debugging?

malikchaya2, line 116: Is this leftover from debugging?

service/api_patch.go
```diff
@@ -109,8 +111,46 @@ func (as *APIServer) submitPatch(w http.ResponseWriter, r *http.Request) {
...
+			Permission:    evergreen.PermissionPatches,
+			RequiredLevel: evergreen.PatchSubmitAdmin.Value,
+		}
+		if !dbUser.HasPermission(opts) {
+			as.LoggedError(w, r, http.StatusUnauthorized, errors.New("user is not authorized to patch on behalf of other users"))
+			return
+		}
+		specifiedUser, err := user.FindByGithubName(data.GithubAuthor)
+		if err != nil {
+			as.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "error looking for specified author"))
```
malikchaya2, line 133: In this case, shouldn't we just use an anonymous author instead of erroring?

malikchaya2 (changes requested):

service/api_patch.go
ablack12: Yes :)

service/api_patch.go
ablack12: I put this here as helpful logging for us in the future, whereas I think LoggedError should only be used if we're failing the command (it's not worth failing here bc it's not the inputters fault that we can't find the github user, but the Github users fault for not including this in their user settings)

service/api_patch.go
ablack12: I think it's correct to error for database issues (this is rare anyway).

malikchaya2 (approved):

PR #5266, opened 2021-12-17
EVG-15797 Handle empty task status errors

hadjri:
[EVG-15797](https://jira.mongodb.org/browse/EVG-15797)

### Description 
Added additional error logging when a task status is set to an empty string, and in these cases set the status to failed in order to preserve UI functionality.  Changed stranded task cleanup job to pass in a hardcoded details struct instead of &t.Details since an initial investigation indicated this might be an issue.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1469,13 +1469,13 @@ func (t *Task) MarkEnd(finishTime time.Time, detail *apimodels.TaskEndDetail) er
 		"execution": t.Execution,
 		"project":   t.Project,
 		"details":   t.Details,
-		"status":    detail.Status,
 	})
 	if detail.Status == "" {
-		grip.Error(message.Fields{
-			"error":   errors.WithStack(errors.New("task end details status detected as empty, marking status as failed")),
+		grip.Error(message.WrapError(errors.WithStack(errors.New("task status is empty, marking as failed")), message.Fields{
+			"message": "empty task details",
 			"details": detail,
-		})
+			"task_id": t.Id,
+		}))
 		detail.Status = evergreen.TaskFailed
 	}
 	return UpdateOne(
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1469,7 +1469,15 @@ func (t *Task) MarkEnd(finishTime time.Time, detail *apimodels.TaskEndDetail) er
 		"execution": t.Execution,
 		"project":   t.Project,
 		"details":   t.Details,
+		"status":    detail.Status,
 	})
+	if detail.Status == "" {
+		grip.Error(message.Fields{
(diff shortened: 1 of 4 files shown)


model/task/task.go
```diff
@@ -1469,7 +1469,15 @@ func (t *Task) MarkEnd(finishTime time.Time, detail *apimodels.TaskEndDetail) er
 		"execution": t.Execution,
 		"project":   t.Project,
 		"details":   t.Details,
+		"status":    detail.Status,
```
ablack12, line 1472: We're already logging this as part of "details"

model/task/task.go
```diff
@@ -1469,7 +1469,15 @@ func (t *Task) MarkEnd(finishTime time.Time, detail *apimodels.TaskEndDetail) er
...
 	})
+	if detail.Status == "" {
+		grip.Error(message.Fields{
```
ablack12, line 1475: Can we use grip.Error(message.WrapError ? This handles the stack but formats it better.
I would also suggest adding a "message" of some sort (ticket number or "empty task details") to make it easier to create an alert on this, and adding a "task_id" section.

units/task_stranded_cleanup.go
```diff
@@ -3,6 +3,7 @@ package units
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/apimodels"
```
ablack12, line 6: This should be in the below list of inports

ablack12 (changes requested):

model/task/task.go
hadjri: Oops yes we are :)

model/task_lifecycle.go
```diff
@@ -242,6 +242,14 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 		} else if !t.IsFinished() {
 			if detail != nil {
 				grip.Debugln(msg, "marking as failed")
+				if detail.Status == "" {
```
ablack12, line 245: Do we need to to do this again here? MarkEnd calls task.MarkEnd so that log seems like it would be enough.

hadjri: True, wanted to be sure in case of any alternate chains of logic but I suppose we can remove this here.

ablack12 (approved):
Looks good to me, we should create a splunk alert on this message to ensure that we aren't just suppressing this problem now. Can you attach it to the ticket?

hadjri:
> Looks good to me, we should create a splunk alert on this message to ensure that we aren't just suppressing this problem now. Can you attach it to the ticket?

Attached newly created alert to the ticket.

PR #5267, opened 2021-12-20
EVG-15836 s3_copy should not say success if the file didn't actually copy

malikchaya2:

### Description 
Moving the s3copy command eliminates a bunch of errors that I couldn't solve in other ways. 

### Testing 
Tested on staging. 
(will add links in a few minutes) 
[On prod/main](https://spruce.mongodb.com/version/61c0924aa4cf475c62f74ac6/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC): The AWS key and secret is fake, but the task passed and a file is attached. 
[the same task with my changes on staging](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_s3_fake_destination_optional_patch_d30cf446b44d58d39e1cea3c3665de8cc8e007e5_61c09460b237360f8b651f53_21_12_20_14_34_41/logs?execution=0): It fails and prints the AWS error


agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
 
-// ParseParams decodes the S3 push command parameters that are
-// specified as part of an S3CopyPlugin command; this is required
-// to satisfy the 'Command' interface
-func (c *s3copy) ParseParams(params map[string]interface{}) error {
-	if err := mapstructure.Decode(params, c); err != nil {
-		return errors.Wrapf(err, "error decoding %v params", c.Name())
+// ParseParams decodes the S3 push command parameters
+func (s3pc *s3copy) ParseParams(params map[string]interface{}) error {
+	decoder, err := mapstructure.NewDecoder(&mapstructure.DecoderConfig{
```
ablack12, line 103: What was the reason for using this method of decoding?

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
 // the fields necessary for carrying out an S3 copy operation are present
-func (c *s3copy) validateParams() error {
-	if c.AwsKey == "" {
-		return errors.New("s3 AWS key cannot be blank")
+func (s3pc *s3copy) validate() error {
+	catcher := grip.NewSimpleCatcher()
+
+	// make sure the command params are valid
+	if s3pc.AwsKey == "" {
+		catcher.Add(errors.New("aws_key cannot be blank"))
```
ablack12, line 125: can use catcher.New instead of catcher.Add(errors.New

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
 		if s3CopyFile.Destination.Region == "" {
 			s3CopyFile.Destination.Region = endpoints.UsEast1RegionID
 		}
-		if s3CopyFile.Source.Bucket == "" {
-			return errors.New("s3 source bucket cannot be blank")
-		}
-		if s3CopyFile.Destination.Bucket == "" {
-			return errors.New("s3 destination bucket cannot be blank")
-		}
-		if s3CopyFile.Source.Path == "" {
```
ablack12, line 116: Is it not true that we need path?

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
-			return errors.New("s3 source path cannot be blank")
-		}
-		if s3CopyFile.Destination.Path == "" {
-			return errors.New("s3 destination path cannot be blank")
-		}
-		if s3CopyFile.Permissions == "" {
-			s3CopyFile.Permissions = s3.BucketCannedACLPublicRead
```
ablack12, line 123: I think we should still be defaulting permissions in case current commands rely on this.

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
-		return errors.New("s3 AWS secret cannot be blank")
+	if s3pc.AwsSecret == "" {
+		catcher.Add(errors.New("aws_secret cannot be blank"))
 	}
-	for _, s3CopyFile := range c.S3CopyFiles {
+
+	for _, s3CopyFile := range s3pc.S3CopyFiles {
+		// make sure both buckets are valid
+		if err := validateS3BucketName(s3CopyFile.Source.Bucket); err != nil {
+			return errors.Wrapf(err, "%v is an invalid bucket name", s3CopyFile.Source.Bucket)
```
ablack12, line 134: How come for these errors we don't use the catcher? Seems like we should either use it for everything or not use it at all.

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
+}
 
-		err = validateS3BucketName(s3CopyFile.Destination.Bucket)
-		if err != nil {
-			return errors.Wrapf(err, "destination bucket '%v' is invalid",
-				s3CopyFile.Destination.Bucket)
-		}
+// Apply the expansions from the relevant task config (including restricted expansion)
+// to all appropriate fields of the s3put.
+func (s3pc *s3copy) expandParams(conf *internal.TaskConfig) error {
```
ablack12, line 153: We're getting rid of restricted expansions, but anyway this command shouldn't be using them, only [s3.put and s3.get](https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#variables). I'm not sure this helper function is really adding anything as opposed to just calling ExpandValues within Execute.

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
+		errChan <- errors.WithStack(s3pc.copyWithRetry(ctx, comm, logger, conf))
 	}()
 
 	select {
 	case err := <-errChan:
-		return errors.WithStack(err)
+		return err
 	case <-ctx.Done():
-		logger.Execution().Info("Received signal to terminate execution of S3 copy command")
+		logger.Execution().Info("Received signal to terminate execution of S3 Put Command")
```
ablack12, line 189: s3 copy still

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
-
 	return nil
 }
 
-// Execute carries out the s3copy command - this is required
-// to satisfy the 'Command' interface
-func (c *s3copy) Execute(ctx context.Context,
+// Implementation of Execute.  Expands the parameters, and then copies the
+// resource from one s3 bucket to another one.
+func (s3pc *s3copy) Execute(ctx context.Context,
```
ablack12, line 163: FYI I think s3pc is used for s3put to stand for "s3 put command"

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
 	}
+	// re-validate command here, in case an expansion is not defined
+	if err := s3pc.validate(); err != nil {
+		return errors.WithStack(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	s3pc.taskdata = client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
```
ablack12, line 178: Seems like we can define this directly in copyWithRetry

agent/command/s3_copy.go
```diff
@@ -75,105 +95,140 @@ type s3Loc struct {
...
+	td := s3pc.taskdata
 
 	var foundDottedBucketName bool
 
-	for _, s3CopyFile := range c.S3CopyFiles {
+	client := utility.GetHTTPClient()
+	client.Timeout = 10 * time.Minute
+	defer utility.PutHTTPClient(client)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
ablack12, line 209: We do this in Execute and copyWithRetry; do we need it in both places?

agent/command/s3_copy.go
```diff
@@ -188,43 +243,70 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
-			S3SourcePath:        s3CopyFile.Source.Path,
-			S3DestinationRegion: s3CopyFile.Destination.Region,
-			S3DestinationBucket: s3CopyFile.Destination.Bucket,
-			S3DestinationPath:   s3CopyFile.Destination.Path,
-			S3DisplayName:       s3CopyFile.DisplayName,
-			S3Permissions:       s3CopyFile.Permissions,
-		}
-
-		responseString, err := comm.S3Copy(ctx, td, &s3CopyReq)
+		s3CopyReq.AwsKey = s3pc.AwsKey
```
ablack12, line 246: Are we able to add this when s3CopyReq is defined?

agent/command/s3_copy.go
```diff
@@ -188,43 +243,70 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
-			if s3CopyFile.Optional {
-				logger.Execution().Errorf("file '%s' is optional, continuing",
-					s3CopyFile.DisplayName)
-				continue
-			} else {
-				return errors.Wrapf(err, "failed to push %s to %s",
-					s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket)
-			}
-
+			grip.Error(errors.Wrap(err, "S3 copy failed, could not establish connection to source bucket"))
```
ablack12, line 259: Can we log this to the task or execution logs? Or else logging this isn't really helpful. Although it seems like we should be returning here.

agent/command/s3_copy.go
```diff
@@ -188,43 +243,70 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+		destOpts := pail.S3Options{
+			Credentials: pail.CreateAWSCredentials(s3CopyReq.AwsKey, s3CopyReq.AwsSecret, ""),
+			Region:      s3CopyReq.S3DestinationRegion,
+			Name:        s3CopyReq.S3DestinationBucket,
+			Permissions: pail.S3Permissions(s3CopyReq.S3Permissions),
+		}
+		destBucket, err := pail.NewS3MultiPartBucket(destOpts)
 		if err != nil {
-			return errors.WithStack(err)
+			grip.Error(errors.Wrap(err, "S3 copy failed, could not establish connection to destination bucket"))
```
ablack12, line 273: Same thought here with errors

agent/command/s3_copy.go
```diff
@@ -188,43 +243,70 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+		}
+
+		for i := 1; i <= maxS3OpAttempts; i++ {
```
ablack12, line 276: i := **0**; i **<** maxS3OpAttempts; i++ is more standard

agent/command/s3_copy.go
```diff
@@ -188,43 +243,70 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+					DestinationBucket: destBucket,
+				}
+				err = srcBucket.Copy(ctx, copyOpts)
+				if err != nil {
+					newPushLog.Status = pushLogFailed
+					if err := comm.UpdatePushStatus(ctx, td, newPushLog); err != nil {
+						return errors.Wrap(err, "updating pushlog status failed for task")
+					}
+					if s3CopyFile.Optional {
+						logger.Execution().Errorf("S3 push copy failed to copy'%s' to '%s'. File is optional, continuing \n error",
```
ablack12, line 293: nit: missing space between copy and '%s'. Also not sure what this `\n error` is for

service/api.go
```diff
@@ -440,6 +440,80 @@ func (as *APIServer) SetDownstreamParams(w http.ResponseWriter, r *http.Request)
...
+			errors.Wrapf(err, "problem querying task %s with version id %s",
+				task.Id, task.Version))
+		return
+	}
+
+	// Check for an already-pushed file with this same file path,
+	// but from a conflicting or newer commit sequence num
+	if v == nil {
+		as.LoggedError(w, r, http.StatusNotFound,
+			errors.Errorf("no version found for build '%s'", task.BuildId))
```
ablack12, line 467: Can we log the task ID instead? (same below)

service/api.go
```diff
@@ -440,6 +440,80 @@ func (as *APIServer) SetDownstreamParams(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		as.LoggedError(w, r, http.StatusInternalServerError,
+			errors.Wrapf(err, "problem querying for push log at %s (build=%s)",
+				copyToLocation, task.BuildId))
+		return
+	}
+	if newestPushLog != nil {
+		grip.Warningln("conflict with existing pushed file:", copyToLocation)
+		gimlet.WriteJSON(w, nil)
+		return
```
ablack12, line 483: Are we not returning an error here to avoid the retries? Can we drop a comment here to that effect if so?

ablack12 (changes requested):
Yay!!!

agent/command/s3_copy.go
malikchaya2: There was no specific reason, I copied it from the s3put command. Would you prefer a different method?

agent/command/s3_copy.go
malikchaya2: I moved it because I send s3copyReq to the push command, which doesn't need them, and I didn't want to send them if it's not necessary.

agent/command/s3_copy.go
malikchaya2: \n error makes it more readable in the task log. It tends to get pretty long, this puts err: on a new line so the user doesn't have to keep scrolling horizontally.

service/api.go
malikchaya2: I don't think that that is the reason. There being a conflict is not an error state. We just don't want to push again so we noop.

service/api.go
```diff
@@ -440,6 +440,80 @@ func (as *APIServer) SetDownstreamParams(w http.ResponseWriter, r *http.Request)
...
+	err := utility.ReadJSON(utility.NewRequestReader(r), pushLog)
+	if err != nil {
+		as.LoggedError(w, r, http.StatusBadRequest, err)
+		return
+	}
+
+	err = errors.Wrapf(pushLog.UpdateStatus(pushLog.Status),
+		"updating pushlog status failed for task %s", task.Id)
+	if err != nil {
+		grip.Error(err)
```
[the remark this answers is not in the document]
malikchaya2: It was set up that way originally, I don't see a reason to remove it.

service/api.go
```diff
@@ -678,11 +752,12 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.Route().Version(2).Route("/task/{taskId}/expansions").Wrap(requireTask, requireHost).Handler(as.GetExpansions).Get()
 
 	// plugins
+	app.Route().Version(2).Prefix("/task/{taskId}").Route("/newPush").Wrap(requireTask).Handler(as.NewPush).Post()
+	app.Route().Version(2).Prefix("/task/{taskId}").Route("/updatePushStatus").Wrap(requireTask).Handler(as.UpdatePushStatus).Post()
 	app.Route().Version(2).Prefix("/task/{taskId}").Route("/git/patchfile/{patchfile_id}").Wrap(requireTaskSecret).Handler(as.gitServePatchFile).Get()
 	app.Route().Version(2).Prefix("/task/{taskId}").Route("/git/patch").Wrap(requireTaskSecret).Handler(as.gitServePatch).Get()
 	app.Route().Version(2).Prefix("/task/{taskId}").Route("/keyval/inc").Wrap(requireTask).Handler(as.keyValPluginInc).Post()
 	app.Route().Version(2).Prefix("/task/{taskId}").Route("/manifest/load").Wrap(requireTask).Handler(as.manifestLoadHandler).Get()
-	app.Route().Version(2).Prefix("/task/{taskId}").Route("/s3Copy/s3Copy").Wrap(requireTaskSecret).Handler(as.s3copyPlugin).Post()
```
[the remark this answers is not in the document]
malikchaya2: When will it fail? Do you mean any tasks that started running before the deploy?

agent/command/s3_copy.go
ablack12: I was just wondering, since it's different than the existing method of decoding, but it's probably fine then.

agent/command/s3_copy.go
ablack12: Oh did you mean to log the actual error? This is just logging the word error.

service/api.go
ablack12: I think not being able to put something does seem like an error (StatusConflict probably) and that error could be parsed instead of checking TaskId != "", but if that would retry then we wouldn't want to do that.

service/api.go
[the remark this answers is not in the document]
ablack12: As it is this is just cluttering splunk, since it does'nt log in a very useful way. We should either add message.Fields to log taskId or etc, or not log.

service/api.go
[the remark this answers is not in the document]
ablack12: Yep

agent/command/s3_copy.go
malikchaya2: ah yeah this is a copy/paste mistake, I'll add it back in

malikchaya2: ah yeah this is a copy/paste mistake, I'll add it back in

agent/command/s3_copy.go
malikchaya2: no :)

agent/command/s3_copy.go
malikchaya2: True, I'm not sure why we don't. Will fix.

malikchaya2: True, I'm not sure why we don't. Will fix.

agent/command/s3_copy.go
malikchaya2: Ah, I meant to wrap the error. I'll add it, thanks!

service/api.go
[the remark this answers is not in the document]
malikchaya2: good point, fixing.

service/api.go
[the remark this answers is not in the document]
malikchaya2: good point, I'll add it back in and create another ticket to remove it after some time.

agent/command/s3_copy.go
malikchaya2: It's a copy/paste fail, fixing.

malikchaya2: It's a copy/paste fail, fixing.

service/api.go
malikchaya2: Ah good point. Yes, it probably would retry.

service/api.go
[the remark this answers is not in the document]
malikchaya2: image

agent/command/s3_copy.go
```diff
@@ -188,43 +243,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
-				logger.Execution().Errorf("file '%s' is optional, continuing",
-					s3CopyFile.DisplayName)
-				continue
-			} else {
-				return errors.Wrapf(err, "failed to push %s to %s",
-					s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket)
-			}
-
+			connectionErr := errors.Wrap(err, "S3 copy failed, could not establish connection to source bucket")
+			grip.Error(connectionErr)
```
ablack12, line 260: can we remove this grip log (and at 277)? These aren't very useful since they're just logged to the agent

agent/command/s3_copy.go
```diff
@@ -188,43 +243,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+			connectionErr := errors.Wrap(err, "S3 copy failed, could not establish connection to destination bucket")
+			grip.Error(connectionErr)
+			logger.Task().Error(connectionErr)
+			return connectionErr
+		}
+
+		for i := 0; i < maxS3OpAttempts; i++ {
+			select {
+			case <-ctx.Done():
+				return errors.New("s3 put operation canceled")
```
ablack12, line 285: s3 copy

agent/command/s3_copy.go
```diff
@@ -188,43 +243,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+					DestinationKey:    s3CopyReq.S3DestinationPath,
+					DestinationBucket: destBucket,
+				}
+				err = srcBucket.Copy(ctx, copyOpts)
+				if err != nil {
+					newPushLog.Status = pushLogFailed
+					if err := comm.UpdatePushStatus(ctx, td, newPushLog); err != nil {
+						return errors.Wrap(err, "updating pushlog status failed for task")
+					}
+					if s3CopyFile.Optional {
```
ablack12, line 298: nit: missing space between copy and '%s'

agent/command/s3_copy.go
```diff
@@ -78,102 +98,137 @@ type s3Loc struct {
...
 
+	timer := time.NewTimer(0)
+	defer timer.Stop()
+
 	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
 
 	var foundDottedBucketName bool
 
+	client := utility.GetHTTPClient()
+	client.Timeout = 10 * time.Minute
```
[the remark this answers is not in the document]
malikchaya2: It's hard to tell because of how much I shifted around. I think it might, but it would operate the same way s3_put does, so it's probably safe.

service/api.go
```diff
@@ -678,6 +758,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.Route().Version(2).Route("/task/{taskId}/expansions").Wrap(requireTask, requireHost).Handler(as.GetExpansions).Get()
 
 	// plugins
+	app.Route().Version(2).Prefix("/task/{taskId}").Route("/newPush").Wrap(requireTask).Handler(as.NewPush).Post()
```
ablack12, line 761: Can you move this up to the agent routes section?

agent/command/s3_copy.go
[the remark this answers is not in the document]
malikchaya2: I don't think it should, it's copied over from api_plugin_s3copy.go (the old code).

service/api.go
```diff
@@ -678,6 +758,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
...
+	app.Route().Version(2).Prefix("/task/{taskId}").Route("/updatePushStatus").Wrap(requireTask).Handler(as.UpdatePushStatus).Post()
```
ablack12, line 762: nit: can this route be update_push_status (and new_push)? There are a couple of outliers that snuck through but this is our general convention.

agent/command/s3_copy.go
```diff
@@ -78,102 +98,137 @@ type s3Loc struct {
...
+		}
 		if s3CopyFile.Source.Region == "" {
 			s3CopyFile.Source.Region = endpoints.UsEast1RegionID
 		}
 		if s3CopyFile.Destination.Region == "" {
 			s3CopyFile.Destination.Region = endpoints.UsEast1RegionID
 		}
 		if s3CopyFile.Source.Bucket == "" {
-			return errors.New("s3 source bucket cannot be blank")
+			catcher.New("s3 source bucket cannot be blank")
```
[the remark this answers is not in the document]
malikchaya2: I can actually remove that because it'll be covered under `len(bucket) < 3`

service/api.go
```diff
@@ -676,6 +756,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.Route().Version(2).Route("/task/{taskId}/parser_project").Wrap(requireTaskSecret).Handler(as.GetParserProject).Get()
 	app.Route().Version(2).Route("/task/{taskId}/project_ref").Wrap(requireTaskSecret).Handler(as.GetProjectRef).Get()
 	app.Route().Version(2).Route("/task/{taskId}/expansions").Wrap(requireTask, requireHost).Handler(as.GetExpansions).Get()
+	app.Route().Version(2).Prefix("/task/{taskId}").Route("/new_push").Wrap(requireTask).Handler(as.NewPush).Post()
```
ablack12, line 759: Last thing: I think both of these should use requireTaskSecret rather than requireTask, since it's a stricter check and we are passing the secret in.

ablack12 (approved):
LGTM with middleware change!

service/api.go
ablack12: Also can this just use .Route instead of .Prefix, just for consistency with the other agent routes?

hadjri (approved):

PR #5268, opened 2021-12-20
EVG-16005 Remove webhook check from created_ticket endpoint

malikchaya2:
[EVG-16005](https://jira.mongodb.org/browse/EVG-16005)

### Description 
When this endpoint was created, it was only meant to be used in conjunction with a webhook set up to be hooked up to the file ticket button. However, due to a new project to move build failure tickets, a new use case came up that warrants allowing the endpoint to be used independently. 

### Testing 
none

### DIFF
--- a/rest/route/annotations.go
+++ b/rest/route/annotations.go
@@ -8,7 +8,6 @@ import (
 	"strconv"
 
 	"github.com/evergreen-ci/evergreen"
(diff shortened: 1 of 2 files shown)


ablack12 (approved):
LGTM; should also update any matching documentation

PR #5269, opened 2021-12-20
EVG-15493: add REST route to get pod information

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15493

### Description 
* Add REST route to get a pod by ID. The data model already exists, it just has to be exposed in the API.
    * I made it available only for admins for now, since it's not a feature available for production. I'll change it to use looser permissions once the containers initiative is more production-ready.
* Fix the APIPod data model to include the required pod type field.
* Slightly fix mock pod data connector to have the same behavior as the DB connector for `FindPodByID`.

### Testing 
* Added test for REST route.
* Fixed existing REST tests to require pod type.

### DIFF
--- a/rest/model/pod.go
+++ b/rest/model/pod.go
@@ -177,16 +177,13 @@ func (t *APIPodType) BuildFromService(pt pod.Type) error {
 }
 
 // ToService converts a REST API pod type into a service-layer pod type.
-func (t *APIPodType) ToService() (*pod.Type, error) {
-	if t == nil {
-		return nil, errors.New("nonexistent pod type")
-	}
+func (t APIPodType) ToService() (*pod.Type, error) {
 	var converted pod.Type
-	switch *t {
+	switch t {
 	case PodTypeAgent:
 		converted = pod.TypeAgent
 	default:
-		return nil, errors.Errorf("unrecognized pod type '%s'", *t)
+		return nil, errors.Errorf("unrecognized pod type '%s'", t)
 	}
 	return &converted, nil
 }
@@ -225,12 +222,9 @@ func (s *APIPodStatus) BuildFromService(ps pod.Status) error {
 }
 
 // ToService converts a REST API pod status into a service-layer pod status.
-func (s *APIPodStatus) ToService() (*pod.Status, error) {
-	if s == nil {
-		return nil, errors.New("nonexistent pod status")
-	}
+func (s APIPodStatus) ToService() (*pod.Status, error) {
 	var converted pod.Status
-	switch *s {
+	switch s {
 	case PodStatusInitializing:
 		converted = pod.StatusInitializing
 	case PodStatusStarting:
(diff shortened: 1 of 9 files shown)


rest/route/pod.go
```diff
@@ -96,3 +96,49 @@ func (h *podPostHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+func (h *podGetHandler) Factory() gimlet.RouteHandler {
+	return &podGetHandler{
+		env: h.env,
+		sc:  h.sc,
+	}
+}
+
+// Parse fetches the podID and JSON payload from the HTTP request.
```
julianedwards, line 124: Which JSON payload?

rest/model/pod.go
```diff
@@ -149,6 +156,41 @@ func (p *APIPod) ToService() (*pod.Pod, error) {
...
+	default:
+		return errors.Errorf("unrecognized pod type '%s'", pt)
+	}
+	*t = converted
+	return nil
+}
+
+// ToService converts a REST API pod type into a service-layer pod type.
+func (t *APIPodType) ToService() (*pod.Type, error) {
+	if t == nil {
```
julianedwards, line 181: Do we need to check for `nil` here? It seems like panic behavior.

julianedwards (changes requested):

rest/route/pod.go
```diff
@@ -96,3 +96,49 @@ func (h *podPostHandler) Run(ctx context.Context) gimlet.Responder {
...
+	}
+}
+
+// Parse fetches the pod ID from the HTTP request.
+func (h *podGetHandler) Parse(ctx context.Context, r *http.Request) error {
+	h.podID = gimlet.GetVars(r)["pod_id"]
+	return nil
+}
+
+// Run creates a new resource based on the Request-URI and JSON payload.
```
julianedwards, line 130: I think this comment also needs updating.

rest/model/pod.go
Kimchelly: I waffled on this when I wrote it because there's some remote possibility that the type could be optional during API <-> DB conversion, so we would have to handle `*APIPodType` at some point in the future. There could be an argument that the `APIPod`'s `Type` field should be `*APIPodType` for consistency with other REST models, which always make all fields optional. However, given that it's not an optional field for now and assuming the DB will always populate it (this should be true), I changed it to be a non-pointer so the receiver is now `t APIPodType`.

On the topic of panics, this doesn't panic because you can call `ToService` even if `t` is a nil pointer. I'm pretty sure the virtual method table for a type should still be valid even on a nil pointer. Ex:
```go
import (
	"fmt"
)

type foo struct{}

func (f *foo) bar() {
	fmt.Println("I don't panic")
}

func main() {
	var f *foo
	f.bar()
}
```

rest/route/pod.go
Kimchelly: Sorry, copy-paste error. I fixed it.

rest/route/pod.go
Kimchelly: Also fixed!

julianedwards (approved):

PR #5271, opened 2021-12-20
EVG-7756 Enable ability to specify project and check project settings

Chukwurado:
[EVG-7756](https://jira.mongodb.org/browse/EVG-7756)

### Description 
Allow users to specify a project and check the project settings


operations/flags.go
```diff
@@ -59,6 +59,8 @@ const (
 	dbWriteNumFlagName  = "w"
 	dbWmodeFlagName     = "wmode"
 	dbRmodeFlagName     = "rmode"
+
+	projectId = "project_identifier"
```
ablack12, line 63: This flag already exists on line 16; projectFlagName

Chukwurado: Oh I see. Thanks!

operations/validate.go
```diff
@@ -29,13 +29,18 @@ func Validate() cli.Command {
 		}, cli.StringSliceFlag{
 			Name:  joinFlagNames(localModulesFlagName, "lm"),
 			Usage: "specify local modules as MODULE_NAME=PATH pairs",
+		}, cli.StringFlag{
+			Name:  joinFlagNames(projectFlagName, "p"),
+			Usage: "specify project identifier",
```
ablack12, line 34: Can we make this more helpful? It may not be clear to users why they would specify a project name in order to validate a yaml file. Maybe something like "specify project identifier in order to run validation requiring project settings"

operations/validate.go
```diff
@@ -29,13 +29,18 @@ func Validate() cli.Command {
...
 		}),
 		Before: mergeBeforeFuncs(setPlainLogger, requirePathFlag),
 		Action: func(c *cli.Context) error {
 			confPath := c.Parent().String(confFlagName)
 			path := c.String(pathFlagName)
 			quiet := c.Bool(quietFlagName)
 			long := c.Bool(longFlagName)
+
```
ablack12, line 42: nit: remove space

service/api.go
```diff
@@ -533,8 +533,10 @@ func (as *APIServer) listVariants(w http.ResponseWriter, r *http.Request) {
 // validateProjectConfig returns a slice containing a list of any errors
 // found in validating the given project configuration
 func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Request) {
+
```
ablack12, line 536: nit: remove space

service/api.go
```diff
@@ -560,9 +562,27 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	errs := validator.CheckYamlStrict(input.ProjectYaml)
+	errs := validator.ValidationErrors{}
+	if input.ProjectID != "" {
```
ablack12, line 566: Should we add a warning like "validation will proceed without checking project settings" if projectID != ""?

operations/validate.go
```diff
@@ -29,13 +29,18 @@ func Validate() cli.Command {
...
+			projectID := c.String(projectFlagName)
```
ablack12, line 43: If projectID isn't specified, we can check and see if they have one in their local settings.  (this is the `projects_for_directory` section in your project evergreen.yml file) https://github.com/evergreen-ci/evergreen/blob/6158b431898d1492110cec3e9c5ac98508b5829d/operations/model.go#L216

service/api.go
```diff
@@ -560,9 +562,27 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
...
+		projectRef, err := model.FindBranchProjectRef(input.ProjectID)
```
ablack12, line 567: Should use FindMergedProjectRef; we're introducing a feature that will allow users to specify default values at the repo level that will apply to all branches for that repo. FindBranchProjectRef doesn't use these default values though.

validator/project_validator.go
```diff
@@ -98,6 +98,7 @@ type ValidationInput struct {
 	ProjectYaml []byte `json:"project_yaml" yaml:"project_yaml"`
 	Quiet       bool   `json:"quiet" yaml:"quiet"`
 	IncludeLong bool   `json:"include_long" yaml:"include_long"`
+	ProjectID   string `json:"project" yaml:"project"`
```
ablack12, line 101: can we have the json/yaml tags match the variable name? i.e. project_id

ablack12 (changes requested):

service/api.go
hadjri: Should be for == "" correct?

hadjri (approved):
LGTM pending Annie's comments and given that this has been successfully tested in staging with and without the project flag.

service/api.go
```diff
@@ -560,9 +562,27 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
...
+		if err != nil {
+			validationErr.Message = "error finding project; validation will proceed without checking project settings"
```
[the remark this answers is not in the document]
ablack12: cleaner still would be:
```
validationErr = validator.ValidationError{
    Message: <message>,
    Level: Warning,
}
```

ablack12 (commented):
Also need to increment the client version string

service/api.go
```diff
@@ -560,9 +561,36 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	errs := validator.CheckYamlStrict(input.ProjectYaml)
+	errs := validator.ValidationErrors{}
+	if input.ProjectID != "" {
+		projectRef, err := model.FindMergedProjectRef(input.ProjectID, "", false)
```
ablack12, line 566: Can you actually move looking up the projectId from the local config to here? If we end up needing to lookup the project from the local config, we shouldn't error on not finding it; we should instead use the "no project specified" error. cc @hadjri feel free to LGTM this change while I'm gone

ablack12 (commented):
You'll need to merge main to resolve the conflict

service/api.go
hadjri: I think the lookup has to be done in the operations package and can't be done in the API layer - I also don't see that we are erroring on not finding it. The project id will remain empty and we will get to the "no project specified" error.

operations/validate.go
```diff
@@ -29,13 +29,25 @@ func Validate() cli.Command {
...
 			path := c.String(pathFlagName)
 			quiet := c.Bool(quietFlagName)
 			long := c.Bool(longFlagName)
+			projectID := c.String(projectFlagName)
+			if projectID == "" {
+				cwd, err := os.Getwd()
+				grip.Error(errors.Wrap(err, "unable to get current working directory"))
+				cwd, err = filepath.EvalSymlinks(cwd)
+				grip.Error(errors.Wrap(err, "unable to resolve symlinks"))
+				conf := ClientSettings{}
```
hadjri, line 48: This is initializing an empty client settings which will not be able to find the file.  Use the conf variable defined on line 59 (conf, err := NewClientSettings(confPath))

hadjri (changes requested):

hadjri (approved):
LGTM once you fix the client version to December

PR #5272, opened 2021-12-21
EVG-15899 fix this project's a panic

malikchaya2:
[EVG-15899](https://jira.mongodb.org/browse/EVG-15899)

### Description 
If a project does not yet have any versions created for it, it panics when trying to load the event log and waterfall. I added a check to make sure that there is a valid version. 

### Testing 
none

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -1203,7 +1203,7 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, er
 	}
 
 	if lastGoodVersion == nil {
-		return nil, nil, errors.WithStack(errors.Errorf("could not find a valid version for project %s", projectId))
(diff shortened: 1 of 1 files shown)


model/project.go
```diff
@@ -1202,6 +1202,10 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, er
 		}))
 	}
 
+	if lastGoodVersion == nil {
+		return nil, nil, errors.WithStack(errors.Errorf("could not find a valid version for project %s", projectId))
```
ablack12, line 1206: We can still use the error that's wrapped at 1209, just not with lastGoodVersion.Identifier, so either we could change that error statement to not use it or have this error be different but still an errors.Wrapf()

malikchaya2: done

malikchaya2: done

ablack12 (approved):

PR #5273, opened 2021-12-21
EVG-6687 Enable commit-queue delete command to not require passing a project

Chukwurado:
[EVG-6687](https://jira.mongodb.org/browse/EVG-6687)

### Description 
Make passing a project when running `command-queue delete` not required

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,7 +32,7 @@ var (
 	BuildRevision = ""
 
 	// Commandline Version String; used to control auto-updating.
-	ClientVersion = "2021-12-17"
+	ClientVersion = "2021-12-22"
 
 	// Agent version to control agent rollover.
 	AgentVersion = "2021-12-06"
--- a/config.go
+++ b/config.go
@@ -32,11 +32,10 @@ var (
 	BuildRevision = ""
 
 	// Commandline Version String; used to control auto-updating.
-	ClientVersion = "2021-12-28"
-
+	ClientVersion = "2022-01-04"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2021-12-21"
+	AgentVersion = "2021-12-06"
 )
 
 // ConfigSection defines a sub-document in the evergreen config
--- a/config.go
+++ b/config.go
@@ -32,7 +32,7 @@ var (
 	BuildRevision = ""
 
 	// Commandline Version String; used to control auto-updating.
-	ClientVersion = "2021-12-22"
+	ClientVersion = "2022-01-04"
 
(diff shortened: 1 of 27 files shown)


rest/client/interface.go
```diff
@@ -100,7 +100,7 @@ type Communicator interface {
 
 	// Commit Queue
 	GetCommitQueue(ctx context.Context, projectID string) (*restmodel.APICommitQueue, error)
-	DeleteCommitQueueItem(ctx context.Context, projectID string, item string) error
```
malikchaya2, line 103: This also needs to be edited in rest/client/mock.go

malikchaya2 (commented):
This looks great! I would suggest testing these changes locally or on staging to make sure that it still works as expected.

rest/client/methods.go
```diff
@@ -1111,7 +1112,12 @@ func (c *communicatorImpl) GetCommitQueue(ctx context.Context, projectID string)
 	return &cq, nil
 }
 
-func (c *communicatorImpl) DeleteCommitQueueItem(ctx context.Context, projectID, item string) error {
+func (c *communicatorImpl) DeleteCommitQueueItem(ctx context.Context, item string) error {
+	requestedPatch, err := patch.FindOneId(item)
```
ablack12, line 1116: need to check if requestedPatch == nil

ablack12 (changes requested):
Need to increment client version string https://github.com/evergreen-ci/evergreen/blob/cc9bd2c636945353e6748be842e441e83b58a616/config.go#L35

rest/client/methods.go
```diff
@@ -1111,7 +1112,15 @@ func (c *communicatorImpl) GetCommitQueue(ctx context.Context, projectID string)
...
+	if err != nil {
+		return errors.Wrap(err, "error finding item")
+	}
+	if requestedPatch == nil {
+		return errors.Wrap(err, "item not found")
```
hadjri, line 1121: Since error is nil here you can just return an error with the message only.

config.go
```diff
@@ -32,7 +32,7 @@ var (
 	BuildRevision = ""
 
 	// Commandline Version String; used to control auto-updating.
-	ClientVersion = "2021-12-17"
+	ClientVersion = "2021-12-22"
```
hadjri, line 35: Need to increment one more time, unfortunately its updated again since we started review.

hadjri (approved):
LGTM, go ahead and merge once my 2 comments are addressed!

PR #5274, opened 2021-12-22
EVG-15773  Do not displays "SETUP FAILURE" when there is a test failure

malikchaya2:
[EVG-15773 ](https://jira.mongodb.org/browse/EVG-15773 )

### Description 
If a task has a test failure, we want to display "failed" regardless of what the last command type was. 

### Testing 
I'm not sure how to test this, but I'm fairly confident that it will work.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -470,12 +470,9 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 	if err != nil {
 		return errors.Wrap(err, "checking for failed tests")
 	}
-	if hasFailedTests {
-		if detailsCopy.Status != evergreen.TaskFailed {
-			detailsCopy.Type = evergreen.CommandTypeTest
-		} else {
-			detailsCopy.Status = evergreen.TaskFailed
-		}
+	if hasFailedTests && detailsCopy.Status != evergreen.TaskFailed {
+		detailsCopy.Type = evergreen.CommandTypeTest
+		detailsCopy.Status = evergreen.TaskFailed
 	}
 
 	if t.Status == detailsCopy.Status {
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -473,8 +473,9 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 	if hasFailedTests {
 		if detailsCopy.Status != evergreen.TaskFailed {
 			detailsCopy.Type = evergreen.CommandTypeTest
+		} else {
+			detailsCopy.Status = evergreen.TaskFailed
 		}
-		detailsCopy.Status = evergreen.TaskFailed
 	}
 
 	if t.Status == detailsCopy.Status {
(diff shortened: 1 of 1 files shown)


model/task_lifecycle.go
```diff
@@ -472,6 +472,7 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 	}
 	if hasFailedTests {
 		detailsCopy.Status = evergreen.TaskFailed
+		detailsCopy.HasFailedTest = true
```
ablack12, line 475: I'm not sure this is really the fix; this only affects the color that we display on the new UI really, but the API will be inconsistent. We want to instead fix the underlying problem of having the test failure come from  Maybe the problem is that we always set the failure type. If we instead were to only set it if status=failed, then when we get to this point we'll just have a regular test failure.
https://github.com/evergreen-ci/evergreen/blob/70a14e5170f8a816ee2452dd40f516389cc50ebe/agent/agent.go#L663

malikchaya2: I think I found a way to fix it.

model/task_lifecycle.go
```diff
@@ -488,7 +488,9 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		if detailsCopy.Status == evergreen.TaskSucceeded && count == 0 && t.MustHaveResults {
 			detailsCopy.Status = evergreen.TaskFailed
 			detailsCopy.Description = evergreen.TaskDescriptionNoResults
-			detailsCopy.Type = evergreen.CommandTypeTest
+			if !hasFailedTests {
+				detailsCopy.Type = evergreen.CommandTypeTest
```
ablack12, line 492: I believe this ticket is about how, when there _are_ test failures, we aren't modifying the failure type (the status looks to be assigned at 474). This code block handles a different thing, where we mark things as a failure if they require test results but don't have test results. It works correctly (in this case count == 0 is the "hasFailedTests") and shouldn't be modified.

model/task_lifecycle.go
```diff
@@ -471,6 +471,9 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		return errors.Wrap(err, "checking for failed tests")
 	}
 	if hasFailedTests {
+		if detailsCopy.Status != evergreen.TaskFailed {
+			detailsCopy.Type = evergreen.CommandTypeTest
```
malikchaya2, line 475: I didn't add a description because the command type will be "test".

model/task_lifecycle.go
```diff
@@ -471,6 +471,9 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		return errors.Wrap(err, "checking for failed tests")
 	}
 	if hasFailedTests {
+		if detailsCopy.Status != evergreen.TaskFailed {
```
ablack12, line 474: nit: can we combine into one if statement? We don't need to do 477 if detailsCopy.Status = evergreen.TaskFailed already anyway, since it's redundant

ablack12 (changes requested):

model/task_lifecycle.go
malikchaya2: good point, I'll combine.

malikchaya2: good point, I'll combine.

malikchaya2: I put it in an else.

ablack12 (approved):

PR #5275, opened 2021-12-22
EVG-15217 Remove legacy plugin UI component

hadjri:
[EVG-15217](https://jira.mongodb.org/browse/EVG-15217)

### Description 
Now that plugin configs are derived from individual project configs, and not from the admin page, we do not need the UI component for plugins on the admin page.
### Testing 
Tested UI locally ensuring no errors.

### DIFF
--- a/service/templates/admin.html
+++ b/service/templates/admin.html
@@ -2112,20 +2112,6 @@ Admin Settings
 								</md-card-content>
 							</md-card>
 
-							<md-card flex=50 id="plugins" style="max-width:49%">
-								<md-card-title>
-									<md-card-title-text>
-										<span>Plugins</span>
-									</md-card-title-text>
-								</md-card-title>
-								<md-card-content>
-									<md-input-container class="control">
-										<textarea ng-model="tempPlugins" rows="3" md-select-on-focus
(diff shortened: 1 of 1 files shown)


ablack12 (commented):
oh we do still have one plugin left in this section, "dashboard". Do we know if this does something? We should ensure that an empty plugins section doesn't cause issues first before removing it from the UI.

hadjri:
> oh we do still have one plugin left in this section, "dashboard". Do we know if this does something? We should ensure that an empty plugins section doesn't cause issues first before removing it from the UI.

I'm almost certain it does nothing  - but I'll remove and monitor over the next couple days to be sure.

hadjri:
> > oh we do still have one plugin left in this section, "dashboard". Do we know if this does something? We should ensure that an empty plugins section doesn't cause issues first before removing it from the UI.
> 
> I'm almost certain it does nothing - but I'll remove and monitor over the next couple days to be sure.

Removed and nothing amiss since - looks like it was indeed unused. I have a local copy of it just in case.

ablack12 (approved):
🚀

PR #5282, opened 2021-12-23
EVG-16022 Fix setup failure notifications do not work

malikchaya2:
[EVG-16022](https://jira.mongodb.org/browse/EVG-16022)

### Description 
Right now, we deliberately don’t create notifications for setup failure. This reverses that.


bsamek (approved):
lgtm, though we should double check with @ablack12 before merging in case there's some history we're missing

ablack12 (approved):

PR #5283, opened 2021-12-28
EVG-15923 Refactor multi return arguments for project configs into single ProjectInfo

hadjri:
[EVG-15923](https://jira.mongodb.org/browse/EVG-15923)

### Description 
There are instances now across the codebase where 4 part function returns are being used (ProjectRef, ProjectParser, ProjectConfig, error).  To clean this up these methods will now be returning ProjectInfo struct which contains a thruple of the 3 aforementioned structs.
### Testing 
Modified existing unit tests

### DIFF
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -289,6 +289,14 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
 					"project_identifier": ref.Identifier,
 					"revision":           revision,
 				}))
+				if pInfo.Project == nil {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message":            fmt.Sprintf("unable to find project config for revision %s", revision),
+						"runner":             RunnerName,
+						"project":            ref.Id,
+						"project_identifier": ref.Identifier,
+					}))
+				}
 				return err
 			}
 		}
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -289,16 +289,16 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
 					"project_identifier": ref.Identifier,
 					"revision":           revision,
 				}))
-				if pInfo.Project == nil {
-					grip.Error(message.WrapError(err, message.Fields{
-						"message":            fmt.Sprintf("unable to find project config for revision %s", revision),
-						"runner":             RunnerName,
-						"project":            ref.Id,
-						"project_identifier": ref.Identifier,
-					}))
-				}
 				return err
 			}
+		} else if pInfo.Project == nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message":            fmt.Sprintf("unable to find project config for revision %s", revision),
+				"runner":             RunnerName,
+				"project":            ref.Id,
+				"project_identifier": ref.Identifier,
+			}))
+			return err
 		}
 
 		// "Ignore" a version if all changes are to ignored files
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -292,16 +292,6 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
 				return err
 			}
 		}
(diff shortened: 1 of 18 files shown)


service/api_plugin_manifest.go
```diff
@@ -45,20 +45,23 @@ func (as *APIServer) manifestLoadHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	var project *model.Project
-	project, _, _, err = model.LoadProjectForVersion(v, v.Identifier, false)
+	projectInfo, err := model.LoadProjectForVersion(v, v.Identifier, false)
+	if projectInfo.Project == nil {
```
ablack12, line 49: We should check error first

repotracker/repotracker.go
```diff
@@ -302,6 +292,16 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
 				return err
 			}
 		}
+		if pInfo.Project == nil {
+			msg := fmt.Sprintf("unable to find project config for revision %s", revision)
```
ablack12, line 296: Why do we need msg here? Wouldn't we return err instead of errors.New(msg)?

ablack12: I think Project being nil logging is also probably handled by the err != nil case above so we can probably just return.

rest/data/host_create.go
```diff
@@ -192,12 +192,15 @@ func makeProjectAndExpansionsFromTask(t *task.Task) (*model.Project, *util.Expan
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "error populating expansions")
 	}
-	params := append(proj.GetParameters(), v.Parameters...)
+	var params []patch.Parameter
```
ablack12, line 195: If projectInfo.Project == nil, we do need params to still equal v.Params, so maybe we should assign that here

trigger/project_triggers.go
```diff
@@ -38,10 +38,13 @@ func TriggerDownstreamVersion(args ProcessorArgs) (*model.Version, error) {
 	var pp *model.ParserProject
 	var pc *model.ProjectConfig
 	if args.ConfigFile != "" {
-		proj, pp, pc, err = makeDownstreamProjectFromFile(args.DownstreamProject, args.ConfigFile)
+		projectInfo, err := makeDownstreamProjectFromFile(args.DownstreamProject, args.ConfigFile)
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
+		proj = projectInfo.Project
```
ablack12, line 45: We use these variables to create a new ProjectInfo object at line 58 so this feels redundant; instead of initializing each of these individually, we can just create a projectInfo at the top and then only assign Ref at 58

repotracker/repotracker.go
```diff
@@ -289,13 +289,21 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
 					"project_identifier": ref.Identifier,
 					"revision":           revision,
 				}))
+				if pInfo.Project == nil {
```
ablack12, line 292: This makes more sense as an if else for `err != nil`; really we just want to make sure we don't have a nil pointer dereference in the case of `err == nil && project == nil` (which isn't supposed to happen but). No reason to log again here, since it's expected to have a nil project if err != nil.

ablack12 (changes requested):

repotracker/repotracker.go
```diff
@@ -289,16 +289,16 @@ func (repoTracker *RepoTracker) StoreRevisions(ctx context.Context, revisions []
...
-						"message":            fmt.Sprintf("unable to find project config for revision %s", revision),
-						"runner":             RunnerName,
-						"project":            ref.Id,
-						"project_identifier": ref.Identifier,
-					}))
-				}
 				return err
 			}
+		} else if pInfo.Project == nil {
+			grip.Error(message.WrapError(err, message.Fields{
```
ablack12, line 295: We can't wrap error here, because we've guaranteed in the if/else that there _is_ no error, so this will noop. Should just use message.Fields.

ablack12 (approved):

PR #5284, opened 2022-01-04
EVG-15955: fix gopsutil import

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15955

### Description 
* Fix import path for gopsutil to use Go module convention.
* Change process stat type to `[]string` because the return type changed in gopsutil. This changes the schema of the S3 payload stored on the Cedar side but since nobody's reading from system metrics currently, it doesn't seem like a big deal.

### Testing 
N/A


hhoke (approved):
LGTM

PR #5287, opened 2022-01-05
EVG-16024 set displayTaskId to empty for display tasks

ablack12:
[EVG-16024](https://jira.mongodb.org/browse/EVG-16024)

### Description 
We explicitly set DisplayTaskId at creation for tasks but not display tasks, which means we go through GetDisplayTask for them even though they won't ever have one. 

We could also check in IsPartOfDisplay that DisplayOnly is false, however I think populating DisplayTaskId as much as possible is best for the UI, and effectively does the same thing in IsPartOfDisplay.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1150,7 +1150,7 @@ func (t *Task) SetHasCedarResults(hasCedarResults, failedResults bool) error {
 		return err
 	}
 
-	if !t.DisplayOnly && t.IsPartOfDisplay() {
+	if t.IsPartOfDisplay() {
 		displayTask, err := t.GetDisplayTask()
 		if err != nil {
 			return errors.Wrap(err, "getting display task")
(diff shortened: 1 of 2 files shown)


model/task/task.go
```diff
@@ -1150,7 +1150,7 @@ func (t *Task) SetHasCedarResults(hasCedarResults, failedResults bool) error {
 		return err
 	}
 
-	if !t.DisplayOnly && t.IsPartOfDisplay() {
```
malikchaya2, line 1153: It looks like isPartOfDisplay checks if `t.DisplayTaskId == nil`, not if it's "". So it looks like we'd still need `!t.DisplayOnly`.

ablack12: Checking !t.DisplayOnly here is really just a shortcut; IsPartOfDisplay will always be false. That being said, with this change any new display task will have t.DisplayTaskId != nil (because it will be set to "") so we'll still shortcircuit

ablack12: I may be misunderstanding your comment, but "" _is_ not nil, so we would correctly be skipping the if statement in IsPartOfDisplay and check the `utility.FromStringPtr(t.DisplayTaskId) != ""` line directly

malikchaya2: Ah, you're right, I was misreading it.

malikchaya2 (approved):

bynn (approved):

PR #5290, opened 2022-01-05
EVG-16000 Handle inclusion of project config yaml in GetPatchedProject

hadjri:

### Description 

### Testing 

### DIFF
--- a/model/patch/patch.go
+++ b/model/patch/patch.go
@@ -138,32 +138,33 @@ type GitMetadata struct {
 
 // Patch stores all details related to a patch request
 type Patch struct {
-	Id              mgobson.ObjectId       `bson:"_id,omitempty"`
-	Description     string                 `bson:"desc"`
-	Path            string                 `bson:"path,omitempty"`
-	Project         string                 `bson:"branch"`
-	Githash         string                 `bson:"githash"`
-	PatchNumber     int                    `bson:"patch_number"`
-	Author          string                 `bson:"author"`
-	Version         string                 `bson:"version"`
-	Status          string                 `bson:"status"`
-	CreateTime      time.Time              `bson:"create_time"`
-	StartTime       time.Time              `bson:"start_time"`
-	FinishTime      time.Time              `bson:"finish_time"`
-	BuildVariants   []string               `bson:"build_variants"`
-	Tasks           []string               `bson:"tasks"`
-	VariantsTasks   []VariantTasks         `bson:"variants_tasks"`
-	SyncAtEndOpts   SyncAtEndOptions       `bson:"sync_at_end_opts,omitempty"`
-	Patches         []ModulePatch          `bson:"patches"`
-	Parameters      []Parameter            `bson:"parameters,omitempty"`
-	Activated       bool                   `bson:"activated"`
-	PatchedConfig   string                 `bson:"patched_config"`
-	Alias           string                 `bson:"alias"`
-	Triggers        TriggerInfo            `bson:"triggers"`
-	BackportOf      BackportInfo           `bson:"backport_of,omitempty"`
-	MergePatch      string                 `bson:"merge_patch"`
-	GithubPatchData thirdparty.GithubPatch `bson:"github_patch_data,omitempty"`
-	GitInfo         *GitMetadata           `bson:"git_info,omitempty"`
+	Id                   mgobson.ObjectId       `bson:"_id,omitempty"`
+	Description          string                 `bson:"desc"`
+	Path                 string                 `bson:"path,omitempty"`
+	Project              string                 `bson:"branch"`
+	Githash              string                 `bson:"githash"`
+	PatchNumber          int                    `bson:"patch_number"`
+	Author               string                 `bson:"author"`
+	Version              string                 `bson:"version"`
+	Status               string                 `bson:"status"`
+	CreateTime           time.Time              `bson:"create_time"`
+	StartTime            time.Time              `bson:"start_time"`
+	FinishTime           time.Time              `bson:"finish_time"`
+	BuildVariants        []string               `bson:"build_variants"`
+	Tasks                []string               `bson:"tasks"`
+	VariantsTasks        []VariantTasks         `bson:"variants_tasks"`
+	SyncAtEndOpts        SyncAtEndOptions       `bson:"sync_at_end_opts,omitempty"`
+	Patches              []ModulePatch          `bson:"patches"`
+	Parameters           []Parameter            `bson:"parameters,omitempty"`
+	Activated            bool                   `bson:"activated"`
+	PatchedParserProject string                 `bson:"patched_parser_project"`
+	PatchedProjectConfig string                 `bson:"patched_project_config"`
+	Alias                string                 `bson:"alias"`
+	Triggers             TriggerInfo            `bson:"triggers"`
+	BackportOf           BackportInfo           `bson:"backport_of,omitempty"`
+	MergePatch           string                 `bson:"merge_patch"`
+	GithubPatchData      thirdparty.GithubPatch `bson:"github_patch_data,omitempty"`
+	GitInfo              *GitMetadata           `bson:"git_info,omitempty"`
 	// DisplayNewUI is only used when roundtripping the patch via the CLI
 	DisplayNewUI bool `bson:"display_new_ui,omitempty"`
 	// MergeStatus is only used in gitServePatch to send the status of this
@@ -678,7 +679,7 @@ func (p *Patch) UpdateGithashProjectAndTasks() error {
 		"$set": bson.M{
 			GithashKey:       p.Githash,
 			PatchesKey:       p.Patches,
-			PatchedConfigKey: p.PatchedConfig,
+			PatchedConfigKey: p.PatchedParserProject,
 			VariantsTasksKey: p.VariantsTasks,
 			BuildVariantsKey: p.BuildVariants,
 			TasksKey:         p.Tasks,
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -109,22 +109,26 @@ func AddNewTasksForPatch(ctx context.Context, p *patch.Patch, patchVersion *Vers
 // GetPatchedProject creates and validates a project created by fetching latest commit information from GitHub
 // and applying the patch to the latest remote configuration. Also returns the condensed yaml string for storage.
 // The error returned can be a validation error.
-func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken string) (*Project, string, error) {
+func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken string) (*Project, *PatchConfig, error) {
 	if p.Version != "" {
-		return nil, "", errors.Errorf("Patch %v already finalized", p.Version)
+		return nil, nil, errors.Errorf("Patch %v already finalized", p.Version)
 	}
 
 	project := &Project{}
 	projectRef, opts, err := getLoadProjectOptsForPatch(p, githubOauthToken)
 	if err != nil {
-		return nil, "", errors.Wrap(err, "Error fetching project opts for patch")
+		return nil, nil, errors.Wrap(err, "Error fetching project opts for patch")
 	}
 	// if the patched config exists, use that as the project file bytes.
-	if p.PatchedConfig != "" {
-		if _, _, err := LoadProjectInto(ctx, []byte(p.PatchedConfig), opts, p.Project, project); err != nil {
-			return nil, "", errors.WithStack(err)
+	if p.PatchedParserProject != "" {
+		if _, _, err := LoadProjectInto(ctx, []byte(p.PatchedParserProject), opts, p.Project, project); err != nil {
+			return nil, nil, errors.WithStack(err)
 		}
-		return project, p.PatchedConfig, nil
+		patchConfig := &PatchConfig{
+			PatchedParserProject: p.PatchedParserProject,
+			PatchedProjectConfig: p.PatchedProjectConfig,
+		}
+		return project, patchConfig, nil
 	}
 
 	var cancel context.CancelFunc
@@ -151,7 +155,7 @@ func GetPatchedProject(ctx context.Context, p *patch.Patch, githubOauthToken str
 	opts.PatchOpts.env = env
 	projectFileBytes, err = getFileForPatchDiff(ctx, *opts)
 	if err != nil {
-		return nil, "", errors.Wrapf(err, "could not fetch remote configuration file")
+		return nil, nil, errors.Wrapf(err, "could not fetch remote configuration file")
 	}
 
 	// apply remote configuration patch if needed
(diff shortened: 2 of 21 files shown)


model/patch/patch.go
```diff
@@ -138,32 +138,33 @@ type GitMetadata struct {
...
+	StartTime            time.Time              `bson:"start_time"`
+	FinishTime           time.Time              `bson:"finish_time"`
+	BuildVariants        []string               `bson:"build_variants"`
+	Tasks                []string               `bson:"tasks"`
+	VariantsTasks        []VariantTasks         `bson:"variants_tasks"`
+	SyncAtEndOpts        SyncAtEndOptions       `bson:"sync_at_end_opts,omitempty"`
+	Patches              []ModulePatch          `bson:"patches"`
+	Parameters           []Parameter            `bson:"parameters,omitempty"`
+	Activated            bool                   `bson:"activated"`
+	PatchedParserProject string                 `bson:"patched_parser_project"`
```
ablack12, line 160: So unfortunately we can't change the bson tag of an existing variable, because this already exists as `patched_config` in the database, and now we'll have no way to read any existing patch document into the patch struct.

We could still just change the variable name, so that in the code we have `PatchedParserProject` and `PatchedProjectConfig`, but in the DB have `patched_config` and `patched_project_config` or we can leave the variable name as PatchedConfig (for consistency with the bson tag) and use a different variable for the new one. There are truly pros and cons to either but whichever we choose, we should add a comment here to distinguish the two.

ablack12 (commented):
I see you mentioned you tested this on staging; what kind of project config did you use? I think the use case for having this work in patches is to redefine aliases, and then confirm that they work when creating the patch (or when creating a PR). As it stands, we _would_ still be hitting this case https://github.com/evergreen-ci/evergreen/blob/4e0a6f018b9884854c6f207d07492cafdc20c060/operations/patch_util.go#L251 since we aren't checking the local config first.

model/patch/patch.go
hadjri: I've gone ahead and reverted `patched_parser_project` back to `patched_config` and using `patched_project_config` for the new project configs, while keeping the struct name as `PatchedParserProject ` although it deviates from the BSON I do think it makes the code more legible that way.

hadjri:
> I see you mentioned you tested this on staging; what kind of project config did you use? I think the use case for having this work in patches is to redefine aliases, and then confirm that they work when creating the patch (or when creating a PR). As it stands, we _would_ still be hitting this case since we aren't checking the local config first.
> 
> https://github.com/evergreen-ci/evergreen/blob/4e0a6f018b9884854c6f207d07492cafdc20c060/operations/patch_util.go#L262

Should we expect the CLI to automatically pick up on local config yaml changes and apply them? Currently project configs get inserted into DB at the same time we do so for parser projects, and then we reference them later when making API calls rather than referencing them in real time in the CLI.  There are some properties that are supported in the parser project that we lookup via API rather than reading the local config in the CLI, such as when we [list parameters](https://github.com/evergreen-ci/evergreen/blob/main/operations/list.go#L239) for a project.  If we do decide to do this I think we may need a separate ticket to read the local config yaml across the `operations` package.

ablack12:
> Should we expect the CLI to automatically pick up on local config yaml changes and apply them? Currently project configs get inserted into DB at the same time we do so for parser projects, and then we reference them later when making API calls rather than referencing them in real time in the CLI. There are some properties that are supported in the parser project that we lookup via API rather than reading the local config in the CLI, such as when we list parameters for a project. If we do decide to do this I think we may need a separate ticket to read the local config yaml across the operations package.

The evergreen patch command already picks up on local changes; the reason it stores the ParserProject in the patch document (and now also this ProjectConfig) is so that they can actually be used _for_ the patch; otherwise there's likely no purpose in saving ProjectConfig in the patch document at all, it could just be included as a regular change. While `list parameters` may not use the local parameter changes (and could be updated to do so), if you ran a task in an evergreen patch that _relied_ on the parameter value you modified in your patch, it would use the new value because of the stored PatchedParserProject. Users won't be distinguishing between ParserProject and ProjectConfig which is why I'm advocating for ProjectConfig to behave similarly as ParserProject for patches.

hadjri:
Screen Shot 2022-01-19 at 2 31 07 PM

hadjri:
Screen Shot 2022-01-19 at 2 30 20 PM

hadjri:
Tested this now in staging where a patch alias is not defined in the DB but it is defined in the project yaml - see screenshots.  Regex for alias was applied successfully indicating it was picked up on and there is no longer a validation error when specifying an alias that is not present in the DB.

model/project.go
```diff
@@ -1449,6 +1456,14 @@ func (p *Project) ResolvePatchVTs(bvs, tasks []string, requester, alias string,
 		catcher := grip.NewBasicCatcher()
 		vars, err := FindAliasInProjectOrRepo(p.Identifier, alias)
 		catcher.Add(errors.Wrap(err, "can't get alias from project"))
+		if len(vars) == 0 && len(patchDoc.PatchedProjectConfig) > 0 {
+			projectConfig, err := createProjectConfig([]byte(patchDoc.PatchedProjectConfig))
+			catcher.Add(errors.Wrap(err, "can't retrieve aliases from patched config"))
```
ablack12, line 1461: catcher.Wrap (also 1458)

model/project.go
```diff
@@ -1449,6 +1456,14 @@ func (p *Project) ResolvePatchVTs(bvs, tasks []string, requester, alias string,
 		catcher := grip.NewBasicCatcher()
 		vars, err := FindAliasInProjectOrRepo(p.Identifier, alias)
```
ablack12, line 1457: I think this function isn't the right thing to do anymore. It doesn't actually just check project/repo, it also checks Project Config (which should probably be part of the title).

model/project.go
```diff
@@ -1449,6 +1456,14 @@ func (p *Project) ResolvePatchVTs(bvs, tasks []string, requester, alias string,
...
 		catcher.Add(errors.Wrap(err, "can't get alias from project"))
+		if len(vars) == 0 && len(patchDoc.PatchedProjectConfig) > 0 {
+			projectConfig, err := createProjectConfig([]byte(patchDoc.PatchedProjectConfig))
+			catcher.Add(errors.Wrap(err, "can't retrieve aliases from patched config"))
+			if err == nil {
+				projectConfigAliases := aliasesToMap(getFullProjectConfigAliases(projectConfig))
```
ablack12, line 1463: Can we call findAliasFromProjectConfig here instead

model/project_aliases.go
```diff
@@ -209,6 +213,23 @@ func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
...
+	aliases, shouldExit, err := FindAliasInProjectOrRepoFromDb(projectID, alias)
+	if err != nil {
+		return nil, errors.Wrap(err, "error checking for existing aliases")
+	}
+	if len(aliases) > 0 || shouldExit || patchedConfig == "" {
+		return aliases, nil
+	}
+	projectConfig, err := createProjectConfig([]byte(patchedConfig))
+	if err != nil {
+		return nil, errors.Wrap(err, "error retrieving aliases from patched config")
```
ablack12, line 228: This error is actually error creating project config

rest/model/patch.go
```diff
@@ -39,7 +39,8 @@ type APIPatch struct {
 	GithubPatchData         githubPatch          `json:"github_patch_data,omitempty"`
 	ModuleCodeChanges       []APIModulePatch     `json:"module_code_changes"`
 	Parameters              []APIParameter       `json:"parameters"`
-	PatchedConfig           *string              `json:"patched_config"`
+	PatchedParserProject    *string              `json:"patched_config"`
+	PatchedProjectConfig    *string              `json:"patched_project_config"`
```
ablack12, line 43: Do we use this at all with APIPatches? Otherwise we shouldn't include it since it'll make routes like https://evergreen.mongodb.com/rest/v2/patches/61dc61bc0305b90e4255d28c crazier (I'm actually not sure if we should even be including PatchedParserProject but that's a different question)

units/patch_intent.go
```diff
@@ -277,6 +278,24 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 			}
 		}
 	}
+	alias := j.intent.GetAlias()
+
+	if alias != "" {
+		validAlias := false
+		aliases, err := model.FindAliasInProjectOrPatchedConfig(pref.Id, alias, patchDoc.PatchedProjectConfig)
```
ablack12, line 285: Can you verify that this error is visible in the CLI if the alias is invalid, and can we move this logic to a helper function?

units/patch_intent.go
```diff
@@ -277,6 +278,24 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+		if err != nil {
+			return errors.Wrap(err, "error contacting API server")
```
ablack12, line 287: This isn't the correct error anymore

rest/data/project.go
```diff
@@ -3,6 +3,7 @@ package data
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/model/patch"
```
ablack12, line 6: should move this import line to it's own section

model/project_aliases.go
```diff
@@ -209,6 +213,23 @@ func FindAliasInProjectOrRepo(projectID, alias string) ([]ProjectAlias, error) {
 	return findMatchingAliasForProjectConfig(projectID, alias)
 }
 
+// FindAliasInProjectOrPatchedConfig finds all aliases with a given name for a project.
+// If the project has no aliases, the patched config string is checked for the alias as well.
+func FindAliasInProjectOrPatchedConfig(projectID, alias, patchedConfig string) ([]ProjectAlias, error) {
```
ablack12, line 218: Can we write a test for this function?

ablack12 (changes requested):

rest/model/patch.go
hadjri: We don't - I'll remove it

model/patch_lifecycle.go
```diff
@@ -276,14 +289,21 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+	}
+	config, err := createProjectConfig([]byte(p.PatchedProjectConfig))
 	if err != nil {
 		return nil, errors.Wrapf(err,
 			"Error marshaling patched project config from repository revision “%v”",
 			p.Githash)
 	}
 	intermediateProject.Id = p.Id.Hex()
 	if config != nil {
+		config.Identifier = p.Project
```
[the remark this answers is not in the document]
hadjri: Sure!

units/patch_intent.go
hadjri: Screen Shot 2022-01-20 at 6 01 00 PM
Screen Shot 2022-01-20 at 6 01 10 PM

'test2' is defined in the yaml and thus works, 'test' is invalid and we see such in the CLI. I'll change the error to say that the alias isn't specified on the project because it actually doesn't mean it's "invalid" if we get this error.

model/patch/patch.go
```diff
@@ -678,7 +679,7 @@ func (p *Patch) UpdateGithashProjectAndTasks() error {
 		"$set": bson.M{
 			GithashKey:       p.Githash,
 			PatchesKey:       p.Patches,
-			PatchedConfigKey: p.PatchedConfig,
+			PatchedConfigKey: p.PatchedParserProject,
```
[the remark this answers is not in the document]
hadjri: This is called after the patch is updated to refresh a patch's config when a commit queue job runs to save the altered properties to the DB. Since patchedProjectConfig is a part of these props that get modified we do need to add it. I've done so.

model/project.go
hadjri: Modified name to FindAliasInProjectRepoOrConfig

hadjri: Since the current pattern is that aliases in the alias collection take precedence over aliases set in the project configs collection, I think we'd need to check both since they can both potentially contain an active alias.  If someone specifies an alias in their local config for a patch, we can have that take precedence over any alias that is in DB collections since a user would probably want whatever they have locally to be picked up immediately. I'll move that check to be first, and then we'd only check the DB with FindAliasInProjectRepoOrConfig if no matching alias is found in the local config.

ablack12: That behavior seems inconsistent; if in the project ref we have a patch alias called "myAlias' with only one task/variant, and someone tries to locally create an alias called "myAlias" with all tasks for a variant, for their patch they would see all tasks like they expect, but as soon as it's merged the alias will go back to behaving as one task/variant.

units/patch_intent.go
```diff
@@ -836,6 +821,27 @@ func (j *patchIntentProcessor) buildTriggerPatchDoc(ctx context.Context, patchDo
...
+	alias := j.intent.GetAlias()
+	if alias != "" {
+		validAlias := false
+		aliases, err := model.FindAliasInProjectOrPatchedConfig(projectId, alias, patchDoc.PatchedProjectConfig)
+		if err != nil {
+			return errors.Wrapf(err, "error retrieving aliases for project %s", projectId)
+		}
+		for _, al := range aliases {
+			if al.Alias == alias {
+				validAlias = true
```
ablack12, line 834: we can just return nil here now instead of holding onto `validAlias`, and return `errors.Errorf("alias %s is not set on project %s", alias, projectId)` at 842. We can also return immediately at the top if alias == "" so that most of the code doesn't need to be in an if statement

units/patch_intent.go
```diff
@@ -836,6 +821,27 @@ func (j *patchIntentProcessor) buildTriggerPatchDoc(ctx context.Context, patchDo
...
 
+func (j *patchIntentProcessor) verifyValidAlias(projectId string, patchDoc *patch.Patch) error {
+	alias := j.intent.GetAlias()
+	if alias != "" {
+		validAlias := false
+		aliases, err := model.FindAliasInProjectOrPatchedConfig(projectId, alias, patchDoc.PatchedProjectConfig)
+		if err != nil {
+			return errors.Wrapf(err, "error retrieving aliases for project %s", projectId)
+		}
+		for _, al := range aliases {
```
ablack12, line 832: nit: "a" is more typical as a shortening for alias

ablack12:
Adding this here https://github.com/evergreen-ci/evergreen/pull/5290#discussion_r789947981 so it doesn't get lost since it's on an outdated thread.

model/project.go
hadjri: Gotcha, that makes sense, thanks!

model/project.go
```diff
@@ -1454,14 +1454,23 @@ func (p *Project) ResolvePatchVTs(patchDoc *patch.Patch, requester, alias string
...
 		catcher := grip.NewBasicCatcher()
-		vars, err := FindAliasInProjectOrRepo(p.Identifier, alias)
-		catcher.Add(errors.Wrap(err, "can't get alias from project"))
-		if len(vars) == 0 && len(patchDoc.PatchedProjectConfig) > 0 {
-			projectConfig, err := createProjectConfig([]byte(patchDoc.PatchedProjectConfig))
-			catcher.Add(errors.Wrap(err, "can't retrieve aliases from patched config"))
-			if err == nil {
-				projectConfigAliases := aliasesToMap(getFullProjectConfigAliases(projectConfig))
-				vars = projectConfigAliases[alias]
+		vars, shouldExit, err := FindAliasInProjectOrRepoFromDb(p.Identifier, alias)
```
ablack12, line 1457: Similar to verifyValidAlias in units, can we move the logic to populate vars into a helper function?

model/project.go
```diff
@@ -1454,14 +1454,23 @@ func (p *Project) ResolvePatchVTs(patchDoc *patch.Patch, requester, alias string
...
+		catcher.Wrap(err, "can't get alias from project")
+		if !shouldExit && len(vars) == 0 {
+			if len(patchDoc.PatchedProjectConfig) > 0 {
+				projectConfig, err := createProjectConfig([]byte(patchDoc.PatchedProjectConfig))
+				catcher.Wrap(err, "can't retrieve aliases from patched config")
+				if err == nil {
+					vars, err = findAliasFromProjectConfig(projectConfig, alias)
+					if err != nil {
```
ablack12, line 1465: Similar to above, we don't need to check if err is nil, since catcher.Wrapf will noop if there is no error (same with 1472)

model/project_aliases.go
ablack12, line 218: More accurate is ProjectRepoOrPatchedConfig

ablack12 (changes requested):
I think this is almost there, just needs a bit of refatoring!

model/project.go
```diff
@@ -1447,8 +1454,10 @@ func (p *Project) ResolvePatchVTs(bvs, tasks []string, requester, alias string,
 
 	if alias != "" {
 		catcher := grip.NewBasicCatcher()
-		vars, err := FindAliasInProjectOrRepo(p.Identifier, alias)
-		catcher.Add(errors.Wrap(err, "can't get alias from project"))
+		vars, err := p.findAliasesForPatch(alias, patchDoc, catcher)
+		if err != nil {
```
ablack12, line 1458: can just catcher.Wrap without checking err

model/project.go
```diff
@@ -1542,6 +1551,25 @@ func (p *Project) IsGenerateTask(taskName string) bool {
 	return ok
 }
 
+func (p *Project) findAliasesForPatch(alias string, patchDoc *patch.Patch, catcher grip.Catcher) ([]ProjectAlias, error) {
```
ablack12, line 1554: There's really no reason to pass the catcher into this function; it can operate independently and just return an error if it hits one, since we add that error to the catcher at 1459 anyway.

model/project.go
```diff
@@ -1551,20 +1549,26 @@ func (p *Project) IsGenerateTask(taskName string) bool {
...
-			catcher.Wrap(err, "can't retrieve aliases from patched config")
-			if err == nil {
-				vars, err = findAliasFromProjectConfig(projectConfig, alias)
-				catcher.Wrapf(err, "error retrieving alias '%s' from project config", alias)
+			if err != nil {
+				return nil, errors.Wrap(err, "can't retrieve aliases from patched config")
+			}
+			vars, err = findAliasFromProjectConfig(projectConfig, alias)
+			if err != nil {
+				return nil, errors.Wrapf(err, "error retrieving alias '%s' from project config", alias)
```
ablack12, line 1565: Can we say "patched project config" to distinguish this error from 1570?

ablack12 (approved):
LGTM with nit (I would suggest testing on staging with aliases one more time just to verify since this is a large change)

PR #5291, opened 2022-01-05
EVG-15927 Refactor syntax and semantic validators

Chukwurado:
[EVG-15927](https://jira.mongodb.org/browse/EVG-15927)

### Description 
Refactor syntax and semantics validator groups into relevant error / warning groupings

### Testing 
Modified unit tests

### DIFF
--- a/service/api.go
+++ b/service/api.go
@@ -667,14 +667,15 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		errs = append(errs, validationErr)
 	}
 
-	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
 	errs = append(errs, validator.CheckProjectErrors(project, input.IncludeLong)...)
+	if len(errs) > 0 && input.Quiet {
+		errs = errs.AtLevel(validator.Error)
+		gimlet.WriteJSONError(w, errs)
+		return
+	}
+	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
 	errs = append(errs, validator.CheckProjectWarnings(project)...)
-
 	if len(errs) > 0 {
-		if input.Quiet {
-			errs = errs.AtLevel(validator.Error)
-		}
 		gimlet.WriteJSONError(w, errs)
 		return
 	}
--- a/service/api.go
+++ b/service/api.go
@@ -671,15 +671,10 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 
 	if input.Quiet {
 		errs = errs.AtLevel(validator.Error)
-		if len(errs) > 0 {
-			gimlet.WriteJSONError(w, errs)
-		} else {
-			gimlet.WriteJSON(w, validator.ValidationErrors{})
-		}
-		return
+	} else {
+		errs = append(errs, validator.CheckProjectWarnings(project, input.ProjectYaml)...)
 	}
 
-	errs = append(errs, validator.CheckProjectWarnings(project, input.ProjectYaml)...)
 	if len(errs) > 0 {
 		gimlet.WriteJSONError(w, errs)
 		return
--- a/service/api.go
+++ b/service/api.go
@@ -668,8 +668,8 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 	}
 
 	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
-	errs = append(errs, validator.CheckProjectSyntax(project, input.IncludeLong)...)
-	errs = append(errs, validator.CheckProjectSemantics(project)...)
+	errs = append(errs, validator.CheckProjectErrors(project, input.IncludeLong)...)
+	errs = append(errs, validator.CheckProjectWarnings(project)...)
 
 	if len(errs) > 0 {
 		if input.Quiet {
--- a/service/api.go
+++ b/service/api.go
@@ -668,11 +668,17 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 	}
 
 	errs = append(errs, validator.CheckProjectErrors(project, input.IncludeLong)...)
-	if len(errs) > 0 && input.Quiet {
+
+	if input.Quiet {
 		errs = errs.AtLevel(validator.Error)
-		gimlet.WriteJSONError(w, errs)
+		if len(errs) > 0 {
+			gimlet.WriteJSONError(w, errs)
+		} else {
+			gimlet.WriteJSON(w, validator.ValidationErrors{})
+		}
 		return
 	}
+
 	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
 	errs = append(errs, validator.CheckProjectWarnings(project)...)
 	if len(errs) > 0 {
--- a/service/api.go
+++ b/service/api.go
@@ -679,8 +679,7 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
-	errs = append(errs, validator.CheckProjectWarnings(project)...)
+	errs = append(errs, validator.CheckProjectWarnings(project, input.ProjectYaml)...)
 	if len(errs) > 0 {
 		gimlet.WriteJSONError(w, errs)
 		return
--- a/units/commit_queue.go
+++ b/units/commit_queue.go
@@ -609,7 +609,7 @@ func AddMergeTaskAndVariant(patchDoc *patch.Patch, project *model.Project, proje
 	project.Tasks = append(project.Tasks, mergeTask)
 	project.TaskGroups = append(project.TaskGroups, mergeTaskGroup)
 
-	validationErrors := validator.CheckProjectSyntax(project, true)
+	validationErrors := validator.CheckProjectErrors(project, true)
 	validationErrors = append(validationErrors, validator.CheckProjectSettings(project, projectRef)...)
 	catcher := grip.NewBasicCatcher()
 	for _, validationErr := range validationErrors.AtLevel(validator.Error) {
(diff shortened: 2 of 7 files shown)


validator/project_validator.go
```diff
@@ -128,11 +125,19 @@ var projectSyntaxValidators = []projectValidator{
 }
 
 // Functions used to validate the semantics of a project configuration file.
-var projectSemanticValidators = []projectValidator{
+var projectWarningValidators = []projectValidator{
 	checkTaskCommands,
 	checkTaskGroups,
 	checkLoggerConfig,
 	checkTaskTimeout,
+	checkProjectFieldsWarnings,
```
hadjri, line 133: I think it makes sense to refactor the error checkers to all begin with name "validateXyz", and have warning checkers abide by name "checkXyz" for consistency.  Since functions like checkProjectFieldsWarnings are already in the warning validator I think it's redundant to end the function name with "Warnings".  I think checkProjectFields would suffice.  Same goes for all other functions.

hadjri (changes requested):

validator/project_validator.go
```diff
@@ -197,19 +202,19 @@ func getDistrosForProject(projectID string) (ids []string, aliases []string, err
 }
 
 // verify that the project configuration semantics is valid
-func CheckProjectSemantics(project *model.Project) ValidationErrors {
+func CheckProjectWarnings(project *model.Project) ValidationErrors {
 	validationErrs := ValidationErrors{}
-	for _, projectSemanticValidator := range projectSemanticValidators {
+	for _, projectSemanticValidator := range projectWarningValidators {
```
ablack12, line 207: can we change projectSemanticValidator to projectWarningValidator?

validator/project_validator.go
```diff
@@ -197,19 +202,19 @@ func getDistrosForProject(projectID string) (ids []string, aliases []string, err
...
 	}
 	return validationErrs
 }
 
 // verify that the project configuration syntax is valid
-func CheckProjectSyntax(project *model.Project, includeLong bool) ValidationErrors {
+func CheckProjectErrors(project *model.Project, includeLong bool) ValidationErrors {
 	validationErrs := ValidationErrors{}
-	for _, projectSyntaxValidator := range projectSyntaxValidators {
+	for _, projectSyntaxValidator := range projectErrorValidators {
```
ablack12, line 217: same here

validator/project_validator.go
```diff
@@ -259,13 +264,13 @@ func CheckYamlStrict(yamlBytes []byte) ValidationErrors {
 // verify that the project configuration semantics and configuration syntax is valid
 func CheckProjectConfigurationIsValid(project *model.Project, pref *model.ProjectRef) error {
 	catcher := grip.NewBasicCatcher()
-	syntaxErrs := CheckProjectSyntax(project, false)
+	syntaxErrs := CheckProjectErrors(project, false)
```
ablack12, line 267: similarly, should rename these projectErrs / projectWarnings etc, same with the error strings

validator/project_validator.go
```diff
@@ -259,13 +264,13 @@ func CheckYamlStrict(yamlBytes []byte) ValidationErrors {
...
 	if len(syntaxErrs) != 0 {
 		if errs := syntaxErrs.AtLevel(Error); len(errs) != 0 {
 			catcher.Errorf("project contains syntax errors: %s", ValidationErrorsToString(errs))
 		}
 	}
-	semanticErrs := CheckProjectSemantics(project)
+	semanticErrs := CheckProjectWarnings(project)
```
ablack12, line 273: Are we certain now that all of the errors returned here are at the warning level? In which case I'm not sure we should be checking this at all here, since we only add errors to the catcher.

validator/project_validator.go
```diff
@@ -128,11 +125,19 @@ var projectSyntaxValidators = []projectValidator{
 }
 
 // Functions used to validate the semantics of a project configuration file.
-var projectSemanticValidators = []projectValidator{
+var projectWarningValidators = []projectValidator{
```
ablack12, line 128: @hadjri, is there ever a time we want just warnings? Right now to separate these validators out, we've had to create new validators that loop more times, which seems like an unnecessary performance hit if we do typically want 1) Only Errors and 2) Errors and Warnings.

Maybe we could have these functions instead:
CheckProjectErrors (runs all functions that return errors but filters out the warnings) and CheckProjectWarningsAndErrors (runs all functions). I don't think there's anywhere where we only want to return warnings and not errors anyway, but we could brainstorm other formats in that case, or stick with this one. (We can also discuss outside of this thread.)

ablack12 (changes requested):

validator/project_validator.go
Chukwurado: yes it returns all warnings

validator/project_validator.go
hadjri: I don't think there is anywhere at the moment that explicitly only needs warnings - I think that sounds reasonable, since there are currently a few places where we only want to check for errors and having this organization would be cleaner.

hadjri (changes requested):
Looks good after refactoring discussed above and removal of unused code!

validator/project_validator.go
hadjri: SGTM

hadjri (approved):

service/api.go
```diff
@@ -667,14 +667,15 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		errs = append(errs, validationErr)
 	}
 
+	errs = append(errs, validator.CheckProjectErrors(project, input.IncludeLong)...)
+	if len(errs) > 0 && input.Quiet {
+		errs = errs.AtLevel(validator.Error)
+		gimlet.WriteJSONError(w, errs)
+		return
+	}
 	errs = append(errs, validator.CheckYamlStrict(input.ProjectYaml)...)
```
ablack12, line 676: Should we move CheckYamlStrict to be part of CheckProjectWarnings, since it's just a warning?

service/api.go
```diff
@@ -667,14 +667,15 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		errs = append(errs, validationErr)
 	}
 
+	errs = append(errs, validator.CheckProjectErrors(project, input.IncludeLong)...)
+	if len(errs) > 0 && input.Quiet {
```
ablack12, line 671: Even if len(errs) == 0, f input is quiet we want to return here, right?

units/commit_queue.go
```diff
@@ -609,7 +609,7 @@ func AddMergeTaskAndVariant(patchDoc *patch.Patch, project *model.Project, proje
 	project.Tasks = append(project.Tasks, mergeTask)
 	project.TaskGroups = append(project.TaskGroups, mergeTaskGroup)
 
-	validationErrors := validator.CheckProjectSyntax(project, true)
+	validationErrors := validator.CheckProjectErrors(project, true)
 	validationErrors = append(validationErrors, validator.CheckProjectSettings(project, projectRef)...)
 	catcher := grip.NewBasicCatcher()
 	for _, validationErr := range validationErrors.AtLevel(validator.Error) {
```
ablack12, line 615: II know you didn't write this but I think we can just do `catcher.Extend(validationErrors.AtLevel(validator.Error))` instead of looping, or can just use the ValidationErrorsToString functioin that's used elsewhere

validator/project_validator.go
```diff
@@ -197,21 +196,21 @@ func getDistrosForProject(projectID string) (ids []string, aliases []string, err
...
-func CheckProjectSyntax(project *model.Project, includeLong bool) ValidationErrors {
+func CheckProjectErrors(project *model.Project, includeLong bool) ValidationErrors {
 	validationErrs := ValidationErrors{}
-	for _, projectSyntaxValidator := range projectSyntaxValidators {
+	for _, projectErrorValidator := range projectErrorValidators {
 		validationErrs = append(validationErrs,
-			projectSyntaxValidator(project)...)
+			projectErrorValidator(project)...)
 	}
 	for _, longSyntaxValidator := range longSyntaxValidators {
```
ablack12, line 215: Seems like longSyntaxValidators should be longErrorValidator now

ablack12 (changes requested):

service/api.go
Chukwurado: Oh correct

validator/project_validator.go
```diff
@@ -1901,14 +1878,22 @@ func parseS3PullParameters(c model.PluginCommandConf) (task, bv string, err erro
...
+			errs = append(errs,
+				ValidationError{
+					Message: fmt.Sprintf("task '%s' in project '%s' does not "+
+						"contain any commands",
+						task.Name, project.Identifier),
+					Level: Warning,
+				},
+			)
+		}
+		if task.ExecTimeoutSecs == 0 && !execTimeoutWarningAdded {
```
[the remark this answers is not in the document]
Chukwurado: Yes it should. Thanks for catching that

validator/project_validator.go
Chukwurado: Oh my mistake. CheckProjectWarnings should run even when there are no errors. I have modified api.go so that the function will run even when there are no errors. The only time errors will not run is if input.Quiet is set to false. https://github.com/evergreen-ci/evergreen/blob/9320bb372bdb550a7175d201b8e4e1035e234f99/service/api.go#L670-L688

units/commit_queue.go
Chukwurado: I got an error when I replaced the loop with `catcher.Extend(validationErrors.AtLevel(validator.Error))`. It error pretty much said `catcher.Extent()` expects an `[]error` and could not accept `validator.ValidatorErrors`

service/api.go
```diff
@@ -667,14 +667,20 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
...
+		errs = errs.AtLevel(validator.Error)
+		if len(errs) > 0 {
+			gimlet.WriteJSONError(w, errs)
+		} else {
+			gimlet.WriteJSON(w, validator.ValidationErrors{})
 		}
+		return
+	}
+
+	errs = append(errs, validator.CheckProjectWarnings(project, input.ProjectYaml)...)
```
ablack12, line 682: We can simplify this section with 
```
if input.Quiet {
    errs = append(errs, validator.CheckProjectWarnings(project, input.ProjectYaml)...)
} else {
    errs = errs.AtLevel(validator.Error)
}
``` 
and then use the same `if len(errs) > 0` logic, rather than having it twice

ablack12 (approved):
LGTM with nits!

PR #5294, opened 2022-01-07
EVG-14232 Return error when shell.exec command is specified without a script

Chukwurado:
[EVG-14232](https://jira.mongodb.org/browse/EVG-14232)

### Description 
Modify validateCommands to also return an error when command shell.exec is specified without a script.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -460,6 +460,7 @@ const (
 	HostCreateCommandName    = "host.create"
 	S3PushCommandName        = "s3.push"
 	S3PullCommandName        = "s3.pull"
+	ShellExecCommandName     = "shell.exec"
 )
 
 type SenderKey int
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -982,7 +982,7 @@ func validateCommands(section string, project *model.Project,
 				Message: fmt.Sprintf("cannot specify both command '%s' and function '%s'", cmd.Command, cmd.Function),
 			})
 		}
-		if cmd.Command == "shell.exec" && cmd.Params["script"] == nil {
+		if cmd.Command == evergreen.ShellExecCommandName && cmd.Params["script"] == nil {
 			errs = append(errs, ValidationError{
 				Level:   Warning,
 				Message: fmt.Sprintf("%s section: command '%s' specified without a script", section, cmd.Command),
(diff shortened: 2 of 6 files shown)


validator/project_validator.go
```diff
@@ -1014,18 +1020,6 @@ func validatePluginCommands(project *model.Project) ValidationErrors {
 			)
 		}
 
-		for _, c := range commands.List() {
```
hadjri, line 1017: Is this the redundant error?

validator/project_validator.go
```diff
@@ -982,6 +982,12 @@ func validateCommands(section string, project *model.Project,
 				Message: fmt.Sprintf("cannot specify both command '%s' and function '%s'", cmd.Command, cmd.Function),
 			})
 		}
+		if cmd.Command == "shell.exec" && cmd.Params["script"] == nil {
```
hadjri, line 985: Can you make this a string in globals.go alongside GenerateTasksCommandName to prevent hardcoding

validator/project_validator.go
```diff
@@ -982,6 +982,12 @@ func validateCommands(section string, project *model.Project,
...
+			errs = append(errs, ValidationError{
+				Level:   Error,
```
hadjri, line 987: An error will break the build of a project entirely that may have been working before.  I think since having no script is not a critical error we should downgrade this to a warning, just to notify the user but not break their build.

hadjri (changes requested):

validator/project_validator.go
Chukwurado: yes it is

Chukwurado: Yes it is

globals.go
```diff
@@ -460,6 +460,7 @@ const (
 	HostCreateCommandName    = "host.create"
 	S3PushCommandName        = "s3.push"
 	S3PullCommandName        = "s3.pull"
+	ShellExecCommandName     = "shell.exec"
```
ablack12, line 463: Can you also use this where the command is created, for consistency? https://github.com/evergreen-ci/evergreen/blob/293d1e320dccba8f2e8675afa328e51d9d286102/agent/command/registry.go#L55 and https://github.com/evergreen-ci/evergreen/blob/9a855967091f8f17a1a63b66e3c2d8b22cd2265a/agent/command/shell.go#L80

hadjri (changes requested):
I believe tests are failing because of here: https://github.com/evergreen-ci/evergreen/blob/820249ee0e3462a115b2f066eac5a87f55227420/validator/project_validator.go#L1016 You're setting this error as a warning initially which is correct, however this line wipes that and resets it back to an error instead of a warning.  We might want to specify Level here.

hadjri:
Agent version in needs to be updated as well.

hadjri (approved):

PR #5302, opened 2022-01-11
EVG-15693 Update rotate project variables route to handle artifact files

bynn:
[EVG-15693](https://jira.mongodb.org/browse/EVG-15693)

### Description 
added an artifact function to query matching aws secrets and rotate them as a part of the rotate vars route 

### Testing 
artifact unittest


model/artifact/db.go
```diff
@@ -104,6 +114,45 @@ func (e Entry) Upsert() error {
...
+		bson.M{
+			"$set": bson.M{
+				FilesKey: e.Files,
+			},
+			"$setOnInsert": bson.M{
+				ExecutionKey: e.Execution,
+			},
+		},
+	)
+	if err != nil && e.Execution == 0 {
```
bynn, line 135: tries again if it fails and the execution was 0 but wasnt sure which error to return

rest/route/project.go
```diff
@@ -1120,5 +1121,10 @@ func (h *projectVarsPutHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.NewJSONInternalErrorResponse(errors.Wrapf(err,
 			"error updating projects with matching keys"))
 	}
+	_, err = artifact.RotateSecrets(h.replaceVars.ToReplace, h.replaceVars.Replacement, h.replaceVars.DryRun)
```
bynn, line 1124: should i be trying to return this as well? not sure exactly how many artifact files well be changing

model/artifact/db.go
ablack12: Rather than running two Updates, can we case ahead and check if execution == 0, then match on `{"$or": {"ExecutionKey: {$exists: false}, ExecutionKey: 0}` (something like https://github.com/evergreen-ci/evergreen/blob/39faa9305d7f6447447544fc423240a670dbd5fd/model/distro/db.go#L149)?

rest/route/project.go
ablack12: Yeah this is a good point, it might take long enough that this would time out, and I'm not sure there's a way to test if that's true. Can we add a flag that's like "rotate_files": true or something to the body? I would suggest in the docs that people first rotate the secrets, and _then_ rotate the files, since that's less critical. (Or alternatively rotating the secret for the files should be it's own route, but it's probably not worth that work)

ablack12 (changes requested):

model/artifact/artifact_file_test.go
```diff
@@ -203,3 +203,15 @@ func (s *TestArtifactFileSuite) TestFindByIds() {
...
 }
+
+func (s *TestArtifactFileSuite) TestRotateSecret() {
+	changes, err := RotateSecrets("secret", "changedSecret", true)
+	s.NoError(err)
+	s.Len(changes, 3)
+	changes, err = RotateSecrets("secret", "changedSecret", false)
+	s.NoError(err)
+	s.Len(changes, 3)
+	entryFromDb, err := FindOne(ByTaskId("task1"))
```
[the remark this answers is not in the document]
bynn: done

model/artifact/db.go
```diff
@@ -77,6 +77,16 @@ func ByBuildId(id string) db.Q {
 	return db.Query(bson.M{BuildIdKey: id}).Sort([]string{TaskNameKey})
 }
 
+func BySecret(secret string) db.Q {
+	return db.Query(bson.M{
+		"files": bson.M{
+			"$elemMatch": bson.M{
+				"aws_secret": secret,
```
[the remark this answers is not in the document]
bynn: done

model/artifact/db.go
bynn: done

model/artifact/db.go
```diff
@@ -104,6 +114,45 @@ func (e Entry) Upsert() error {
...
+			TaskIdKey:    e.TaskId,
+			TaskNameKey:  e.TaskDisplayName,
+			BuildIdKey:   e.BuildId,
+			ExecutionKey: e.Execution,
+		},
+		bson.M{
+			"$set": bson.M{
+				FilesKey: e.Files,
+			},
+			"$setOnInsert": bson.M{
```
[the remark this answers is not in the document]
bynn: removed because this should never change the execution

rest/route/project.go
bynn: added rotate files and will update doc later!

model/artifact/db.go
```diff
@@ -115,37 +116,36 @@ func (e Entry) Upsert() error {
...
-	if err != nil && e.Execution == 0 {
+	var err error
+	if e.Execution == 0 {
 		err = db.Update(
 			Collection,
 			bson.M{
 				TaskIdKey:   e.TaskId,
 				TaskNameKey: e.TaskDisplayName,
 				BuildIdKey:  e.BuildId,
+				"$or": []bson.M{
```
ablack12, line 127: The only difference between these two queries _is_ the update. We can set this bson.M document at the top like 
```
update := bson.M{
	TaskIdKey:   e.TaskId,
	TaskNameKey: e.TaskDisplayName,
	BuildIdKey:  e.BuildId
}
```
and then say 
```
if e.Execution == 0 {
    update["$or"] = []bson.M{ <execution stuff > }
}
```
Then we can just call db.Update(Collection, update, etc) outside of the if statement. This would prevent the other code duplication.

rest/route/project.go
```diff
@@ -1121,7 +1122,7 @@ func (h *projectVarsPutHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.NewJSONInternalErrorResponse(errors.Wrapf(err,
 			"error updating projects with matching keys"))
 	}
-	_, err = artifact.RotateSecrets(h.replaceVars.ToReplace, h.replaceVars.Replacement, h.replaceVars.DryRun)
```
ablack12: I think this isn't right; we just shouldn't call RotateSecrets at all if RotateFiles isn't true, but RotateSecrets should still consider DryRun.

ablack12 (changes requested):

model/artifact/db.go
bynn: wow youre so smart! done

model/artifact/db.go
```diff
@@ -116,40 +116,35 @@ func (e Entry) Upsert() error {
...
-					FilesKey: e.Files,
-				},
-			},
-		)
+		update[ExecutionKey] = e.Execution
 	}
+
+	err := db.Update(
+		Collection,
+		bson.M{
```
ablack12, line 135: This should be using Update :)

rest/route/project.go
```diff
@@ -1120,5 +1122,10 @@ func (h *projectVarsPutHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.NewJSONInternalErrorResponse(errors.Wrapf(err,
 			"error updating projects with matching keys"))
 	}
+	_, err = artifact.RotateSecrets(h.replaceVars.ToReplace, h.replaceVars.Replacement, h.replaceVars.RotateFiles)
```
ablack12, line 1125: > I think this isn't right; we just shouldn't call RotateSecrets at all if RotateFiles isn't true, but RotateSecrets should still consider DryRun.

I'm not sure why this comment went away 🤔   this still needs to be addressed; rotateFiles shouldn't be treated as dry run.

ablack12 (changes requested):

rest/route/project.go
bynn: huh i dont know why i didnt see but i changed it so that it would still dryrun but only go into it if rotatefiles was set

PR #5304, opened 2022-01-11
EVG-14714 Update version status on restart

bynn:
[EVG-14714](https://jira.mongodb.org/browse/EVG-14714)

### Description 
New function update version status so we dont have to look up build every time 

### Testing 
unscheduling task now updates version and build status 
 
image

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -54,6 +54,50 @@ type VersionToRestart struct {
 	TaskIds   []string `json:"task_ids"`
 }
 
+func UpdateVersionandPatchStatusForBuild(buildIds []string) error {
+	builds, err := build.Find(build.ByIds(buildIds))
+	if err != nil {
+		return errors.Wrapf(err, "fetching builds")
+	}
+	for _, build := range builds {
+		buildStatusChanged, err := UpdateBuildStatus(&build)
+		if err != nil {
+			return errors.Wrapf(err, "updating build '%s' status", build.Id)
+		}
+		// If no build has changed status, then we can assume the version and patch statuses have also stayed the same.
+		if !buildStatusChanged {
+			continue
+		}
+
+		buildVersion, err := VersionFindOneId(build.Version)
+		if err != nil {
+			return errors.Wrapf(err, "getting version '%s' for build '%s'", build.Version, build.Id)
+		}
+		if buildVersion == nil {
+			return errors.Errorf("no version '%s' found for build '%s'", build.Version, build.Id)
+		}
+		newVersionStatus, err := UpdateVersionStatus(buildVersion)
+		if err != nil {
+			return errors.Wrapf(err, "updating version '%s' status", buildVersion.Id)
+		}
+
+		if evergreen.IsPatchRequester(buildVersion.Requester) {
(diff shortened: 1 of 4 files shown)


ablack12 (commented):
I think this isn't the right change for the ticket. The title and comments mention that restarts do work (as evidenced also by [this line](https://github.com/evergreen-ci/evergreen/blob/476075b75a44a657acaee1c4dd6e2a969f90d17d/model/lifecycle.go#L383). It looks like we instead want to handle updating statuses when we unschedule a task. It's possible though that this was already fixed in a previous PR https://github.com/evergreen-ci/evergreen/blob/5f8061f8f7315d12122d845241856244abaf3c76/model/task_lifecycle.go#L130 but you may want to test this by unscheduling a task to confirm.

ablack12 (changes requested):
Don't we also need to UpdateVersionStatus and patch status? (Should we maybe make something like `UpdateBuildAndVersionStatusForTask` that just skips the task step and handles many builds?

model/lifecycle.go
```diff
@@ -74,6 +74,13 @@ func SetVersionActivation(versionId string, active bool, caller string) error {
 		return errors.Wrapf(err, "can't set activation for builds in '%s'", versionId)
 	}
 
+	for _, build := range builds {
+		_, err := UpdateBuildStatus(&build)
```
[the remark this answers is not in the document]

model/lifecycle.go
```diff
@@ -54,6 +54,50 @@ type VersionToRestart struct {
 	TaskIds   []string `json:"task_ids"`
 }
 
+func UpdateVersionandPatchStatusForBuild(buildIds []string) error {
```
ablack12, line 57: nits: And should be capitalized, and should be Builds since we pass in multiple build Ids. 
Also, can you move this function to be next to https://github.com/evergreen-ci/evergreen/blob/5f8061f8f7315d12122d845241856244abaf3c76/model/task_lifecycle.go#L1051 since they are related?

model/lifecycle.go
```diff
@@ -54,6 +54,50 @@ type VersionToRestart struct {
...
+		buildStatusChanged, err := UpdateBuildStatus(&build)
+		if err != nil {
+			return errors.Wrapf(err, "updating build '%s' status", build.Id)
+		}
+		// If no build has changed status, then we can assume the version and patch statuses have also stayed the same.
+		if !buildStatusChanged {
+			continue
+		}
+
+		buildVersion, err := VersionFindOneId(build.Version)
```
ablack12, line 72: This may involve us updating version status n number of times if every build is for the same version. Instead, can we keep a map of versions to update outside of the loop, and then loop through those after we finish updating builds? (we can also update patches in that second loop)

ablack12 (changes requested):

model/lifecycle.go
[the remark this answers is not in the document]
ablack12: I actually didn't but this is a better change 👍

model/lifecycle.go

ablack12 (approved):
LGTM with test fix

PR #5306, opened 2022-01-12
EVG-15836 Fix s3_copy 

malikchaya2:

### Description 
This includes the reverted s3 copy change plus a fix
Note: only the "reset timer and add tests" commit is new. 

### Testing 
I edited the smoke test to copy multiple files, including two that are the same so that it triggers a noop. I then ran it in a patch: https://spruce.mongodb.com/version/61dea5e7a4cf4746d4b92f68/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


agent/command/s3_copy.go
```diff
@@ -220,11 +220,13 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
...
 		if newPushLog.TaskId == "" {
 			logger.Task().Infof("noop, this version is currently in the process of trying to push, or has already succeeded in pushing the file: '%s/%s'", s3CopyFile.Destination.Bucket, s3CopyFile.Destination.Path)
-			return nil
+			timer.Reset(backoffCounter.Duration())
+			continue
 		}
 
 		if len(s3CopyFile.BuildVariants) > 0 && !utility.StringSliceContains(
 			s3CopyFile.BuildVariants, conf.BuildVariant.Name) {
+			timer.Reset(backoffCounter.Duration())
```
ablack12, line 229: Can we just set timer.Reset(0) at the top of the for loop? The backoff counter only seems necessary for the retryLoop, and that makes it less error prone if we ever need to add another if statement or etc.

scripts/agent.yml
```diff
@@ -194,6 +194,10 @@ tasks:
           s3_copy_files:
               - {'source': {'path': 'evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}', 'bucket': 'mciuploads'},
                 'destination': {'path': 'evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}-copy', 'bucket': 'mciuploads'}}
+              - {'source': {'path': 'evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}-2', 'bucket': 'mciuploads'},
```
ablack12, line 197: Can the noop be the middle file rather than the last, to ensure that we continue past the noop?

ablack12 (changes requested):

agent/command/s3_copy.go
malikchaya2: sure

scripts/agent.yml
malikchaya2: yes, I added another file at the end

malikchaya2: new patch with the changes: https://spruce.mongodb.com/version/61dee74d3627e037b1aaf66d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

ablack12: Alternatively instead of copying 3 full files, we could do <file 1>, <file 1>, <file 2> but i'm fine with either approach

ablack12 (approved):

PR #5310, opened 2022-01-13
EVG-16064: Ensure on-demand intent hosts send their instance ID

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16064

### Description 
The EC2 instance ID wasn't being sent by agents running in on-demand hosts because I accidentally excluded it from the filter in `(*agent.Agent).getEC2InstanceID`. This caused an issue where the intent host's agent would check in to the app server but wouldn't send its instance ID. Without an instance ID, we can't do anything to the host, so we couldn't do the host doc swap to turn it into a real host to actually run tasks, nor could we terminate it during the host termination job.

* Fix the agent logic to send the instance ID for on-demand hosts. This will also fix the handful of on-demand intent hosts that are currently unable to send an instance ID.
* Retry fetching the instance ID on every agent request for the next task in case previous attempts failed. This makes it less likely that the agent remains in a bad state if the EC2 instance metadata is temporarily unavailable for whatever reason.
* Make re-termination of intent hosts that are already terminated idempotent by no-oping.
* Improve docs on some existing methods.

### Testing 
Updated existing tests for proper intent host termination.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -176,16 +176,14 @@ func (a *Agent) Start(ctx context.Context) error {
 		tryCleanupDirectory(a.opts.WorkingDirectory)
 	}
 
-	a.getEC2InstanceID(ctx)
-
 	return errors.Wrap(a.loop(ctx), "error in agent loop, exiting")
 }
 
-// getEC2InstanceID sets the agent's instance ID based on the EC2 instance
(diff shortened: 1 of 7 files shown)


julianedwards (approved):
LGTM % missing documentation

PR #5313, opened 2022-01-14
EVG-6846 Return error if a task referenced in buildvariant is also referenced in taskgroup

Chukwurado:
[EVG-6846](https://jira.mongodb.org/browse/EVG-6846)

### Description 
Return error if task referenced in a buildvariant is also referenced in a task group in the buildvariant

### Testing 
Added a unit test

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -568,6 +568,7 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 	for _, buildVariant := range project.BuildVariants {
 		buildVariantTasks := map[string]bool{}
 		revTgNameTasksMap := map[string]string{}
+		taskInTaskGroup := map[string]string{}
 		for _, task := range buildVariant.Tasks {
 			if _, ok := allTaskNames[task.Name]; !ok {
 				if task.Name == "" {
@@ -576,6 +577,7 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 							Message: fmt.Sprintf("tasks for buildvariant '%s' "+
 								"in project '%s' must each have a name field",
 								project.Identifier, buildVariant.Name),
+							Level: Error,
 						},
 					)
 				} else {
@@ -585,6 +587,7 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 								"project '%s' references a non-existent "+
 								"task '%s'", buildVariant.Name,
 								project.Identifier, task.Name),
+							Level: Error,
 						},
 					)
 				}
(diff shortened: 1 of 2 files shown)


validator/project_validator.go
```diff
@@ -585,6 +590,28 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 				}
 			}
 			buildVariantTasks[task.Name] = true
+
+			if task.IsGroup {
```
ablack12, line 594: If you look at where this is defined, the yaml /bson tags are "-" indicating that this is just something we ourselves populate to help us translate these tasks/variants later on, so this isn't really something you can use in validation.

A more useful map may be [taskId] - > taskGroupName (for all tasks in task groups). Then you can look up each buildVariantTask in the map directly (rather than looping through each task groups), and if it exists, you also have the task group name right there as the value. 

I'm also not sure what revTgNameTaskMap is for but given that we're removing the "isGroup" as it's own case it's likely not needed.

ablack12 (changes requested):
@hadjri Chukwurado likely won't have time to do this today, but before merging this change we'd want to run this past all enabled yamls to ensure that we aren't breaking something that currently exists.

validator/project_validator_test.go
```diff
@@ -3781,6 +3781,152 @@ func TestParseS3PullParameters(t *testing.T) {
...
+						Name: "task1",
+					},
+					{
+						Name: "task2",
+					},
+					{
+						Name: "task3",
+					},
+				},
+				TaskGroups: []model.TaskGroup{
```
hadjri, line 3803: Can we add another test case with multiple task groups defined?

validator/project_validator.go
```diff
@@ -585,6 +590,28 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
...
+				for _, tgTask := range tgNameTasksMap[task.Name] {
+					if _, ok := buildVariantTasks[tgTask]; ok {
+						errs = append(errs,
+							ValidationError{
+								Message: fmt.Sprintf("task '%s' in build variant '%s' is already referenced in '%s' task group",
```
hadjri, line 599: Level: Error is default when we don't specify but I think it would be more readable if we explicitly set it.

hadjri (changes requested):

validator/project_validator.go
Chukwurado: `revTgNameTaskMap` maps tasks to task groups, so essentially it is `[taskId] - > taskGroupName`. It was supposed to be a reverse map of `tgNameTaskMap` because `tgNameTaskMap` maps task groups to a slice of its tasks - `taskGroupName -> [taskId1, taskId2]`.

PR #5317, opened 2022-01-19
EVG-16039 schedule/unschedule task group tasks when activating/deactivating one

ablack12:
[EVG-16039](https://jira.mongodb.org/browse/EVG-16039)

### Description 
The Activating interaction can be seen here: task group task 2 was [scheduled by the user](https://evergreen.mongodb.com/task/mongodb_mongo_v4.4_macos_stitch_support_run_tests_a17699e50a4d75d91b5788baa7eb15916341b905_21_12_24_16_20_52/0#/log/EV) but because [task group task 1](https://evergreen.mongodb.com/task/mongodb_mongo_v4.4_macos_stitch_support_install_tests_a17699e50a4d75d91b5788baa7eb15916341b905_21_12_24_16_20_52#/log/EV) had already been scheduled and run successfully a previous day, this wasn't reset and so didn't run with task group task 1.

Also handling the deactivating case (if task group task 1 is deactivated, then task group task 2 should also be deactivated). https://jira.mongodb.org/browse/BF-23865

### Testing 
Added unit tests.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -94,12 +94,19 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 		// If the task was not activated by step back, and either the caller is not evergreen
 		// or the task was originally activated by evergreen, deactivate the task
 	} else if !evergreen.IsSystemActivator(caller) || evergreen.IsSystemActivator(t.ActivatedBy) {
-		// We are trying to deactivate this task
-		// So we check if the person trying to deactivate is evergreen.
-		// If it is not, then we can deactivate it.
-		// Otherwise, if it was originally activated by evergreen, anything can
-		// deactivate it.
 		var err error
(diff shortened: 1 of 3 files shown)


model/task_lifecycle.go
```diff
@@ -49,9 +49,19 @@ func SetActiveState(t *task.Task, caller string, active bool) error {
 			if err != nil {
 				return errors.Wrapf(err, "error getting tasks '%s' depends on", t.Id)
 			}
-			tasksToActivate = append(tasksToActivate, deps...)
+			for _, dep := range deps {
```
malikchaya2, line 52: should we be checking if it's part of a single host task group here?

ablack12: We do check at line 55 if it's in the same task group as t which I was  thinking implied that it was in a single host task group but I can actually make that more explicit

malikchaya2 (approved):

PR #5318, opened 2022-01-19
EVG-16154 merge buildvariants without using pointers

ablack12:
[EVG-16154](https://jira.mongodb.org/browse/EVG-16154)

### Description 
Because we were using pointers to merge build variants, the parserBVs we save to the map are all overwritten to be the last pp.BuildVariants. By not using pointers, we fix this problem and avoid running into it for this operation in the future.

### Testing 
Added a unit test that failed before the fix (bvNew2 didn't exist but bvNew1 existed twice).

### DIFF
--- a/model/project_parser_merge_functions.go
+++ b/model/project_parser_merge_functions.go
@@ -188,27 +188,27 @@ func (pp *ParserProject) mergeUnique(toMerge *ParserProject) error {
 func (pp *ParserProject) mergeBuildVariant(toMerge *ParserProject) error {
 	catcher := grip.NewBasicCatcher()
 
-	bvs := map[string]*parserBV{}
+	bvs := map[string]parserBV{}
 	for _, bv := range pp.BuildVariants {
 		newBv := bv
-		bvs[bv.Name] = &newBv
+		bvs[bv.Name] = newBv
 	}
 	for _, bv := range toMerge.BuildVariants {
-		if _, ok := bvs[bv.Name]; ok {
+		if currentBV, ok := bvs[bv.Name]; ok {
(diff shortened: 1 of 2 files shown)


bynn (approved):

model/project_parser_merge_functions.go
```diff
@@ -188,27 +188,27 @@ func (pp *ParserProject) mergeUnique(toMerge *ParserProject) error {
...
 				catcher.Errorf("build variant '%s' has been declared already", bv.Name)
 			} else {
-				bvs[bv.Name].Tasks = append(bvs[bv.Name].Tasks, bv.Tasks...)
-				bvs[bv.Name].DisplayTasks = append(bvs[bv.Name].DisplayTasks, bv.DisplayTasks...)
+				currentBV.Tasks = append(bvs[bv.Name].Tasks, bv.Tasks...)
+				currentBV.DisplayTasks = append(bvs[bv.Name].DisplayTasks, bv.DisplayTasks...)
+				bvs[bv.Name] = currentBV
 			}
 		} else {
-			pp.BuildVariants = append(pp.BuildVariants, bv)
```
malikchaya2, line 205: This line is confusing me a bit. Why was it here before and why do we no longer want it?

ablack12: I think it's just an oversight; we set pp.BuildVariants to empty at 209 and we don't use it at all within the loop so there's no reason to do this.

bynn: its because i wrote it when i was young and foolish 😭

malikchaya2 (approved):

PR #5319, opened 2022-01-19
EVG-16035 insert a new intent for failed hosts

ybrill:
[EVG-16035](https://jira.mongodb.org/browse/EVG-16035)

### Description 
Currently, when a task spawns a host that AWS terminates immediately after giving it to us we fail the task on `host.list`.
This will instead insert a new host intent so we try again.

We will insert a new intent only if the original host is still "starting". If the host went to running and then terminated the task could have moved on and this would be a waste. I've never actually seen a "running" instance terminate unexpectedly, but just in case.
We also skip this if the task is not running or aborted because it would be a waste.

I'm undecided about limiting the number of times we loop on this. I didn't do it yet, but if someone feels strongly I'm game to add it. I think it's safe to assume it's a transient error since we know the instance was created once successfully (so it's not a user error). Even if the error does persist it won't loop infinitely because `host.list` will eventually reach its timeout.

### Testing 
I manually terminated a task's host in staging (very carefully!) and Evergreen created a second host for the task.


model/host/host.go
```diff
@@ -1600,6 +1600,37 @@ func (h *Host) AddSSHKeyName(name string) error {
 	return nil
 }
 
+// HandleTerminatedHostSpawnedByTask should be called when a host spawned by a task
```
Kimchelly, line 1603: I would rephrase this doc to instead say something along the lines of "re-creates a new intent host if possible when this host.create host fails; otherwise if it cannot create a new host, it'll populate the reason that host.create failed".

model/host/host_intent.go
```diff
@@ -78,6 +78,37 @@ func NewIntent(d distro.Distro, instanceName, provider string, options CreateOpt
 	return intentHost
 }
 
+func newIntentFromHost(h *Host) *Host {
```
Kimchelly, line 81: `newIntentFromExisting`?

model/host/host_intent.go
```diff
@@ -78,6 +78,37 @@ func NewIntent(d distro.Distro, instanceName, provider string, options CreateOpt
...
+	newID := h.Distro.GenerateName()
+	intentHost := &Host{
```
Kimchelly, line 83: This struct initialization seems really similar to `NewIntent` in that both set the same fields in much the same way to achieve the same goal. Can we make both of them call the same common method to create the new host model to keep them in agreement on how to create an intent host?

model/host/host_test.go
```diff
@@ -4829,6 +4829,112 @@ func TestSetNewSSHKeys(t *testing.T) {
...
+				SpawnOptions: SpawnOptions{
+					SpawnedByTask: false,
+				},
+			},
+			newIntentCreated: false,
+			hostDetailsSet:   false,
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			require.NoError(t, db.ClearCollections(Collection, task.Collection))
```
Kimchelly, line 4910: I would defer clear the collections at the end of these tests to avoid leaving random test documents in people's DBs.

model/host/host_test.go
```diff
@@ -4829,6 +4829,112 @@ func TestSetNewSSHKeys(t *testing.T) {
...
+			require.NoError(t, testCase.t.Insert())
+
+			assert.NoError(t, testCase.h.HandleTerminatedHostSpawnedByTask())
+
+			intent, err := FindOne(db.Query(nil))
+			require.NoError(t, err)
+			if testCase.newIntentCreated {
+				require.NotNil(t, intent)
+				assert.Contains(t, intent.Id, "evg")
```
Kimchelly, line 4919: `assert.True(t, host.IsIntentHostId(intent.Id))`

units/host_status.go
```diff
@@ -185,8 +184,8 @@ func (j *cloudHostReadyJob) setCloudHostStatus(ctx context.Context, m cloud.Mana
 
 		catcher := grip.NewBasicCatcher()
 		if h.SpawnOptions.SpawnedByTask {
```
Kimchelly, line 186: It seems like checking if it's spawned by a task is redundant if it's already checked in the function.

units/host_monitoring_external_termination.go
```diff
@@ -134,16 +133,18 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		if cloudStatus != cloud.StatusTerminated && (h.UserHost || h.StartedBy != evergreen.User) {
 			return false, errors.New("non-agent host is not already terminated and should not be terminated")
 		}
+
 		if h.SpawnOptions.SpawnedByTask {
```
Kimchelly, line 137: It seems like checking if it's spawned by a task is redundant if it's already checked in the function.

model/host/host_test.go
```diff
@@ -4829,6 +4829,112 @@ func TestSetNewSSHKeys(t *testing.T) {
...
+					SpawnedByTask:       true,
+					TaskID:              "t0",
+					TaskExecutionNumber: 0,
+				},
+				Status: evergreen.HostStarting,
+			},
+			newIntentCreated: false,
+			hostDetailsSet:   true,
+		},
+		"HostAlreadyRuning": {
```
Kimchelly, line 4876: Running

model/host/host.go
```diff
@@ -1600,6 +1600,37 @@ func (h *Host) AddSSHKeyName(name string) error {
...
+		intent := newIntentFromHost(h)
+		grip.Info(message.Fields{
+			"message":              "inserting a host intent to replace a terminated host for a task",
+			"original_host":        h.Id,
+			"original_host_status": h.Status,
+			"new_host":             intent.Id,
+			"task":                 t.Id,
+			"task_execution":       t.Execution,
+		})
+		return intent.Insert()
```
Kimchelly, line 1628: It seems like if this insertion fails, host.create will be considered a failure since it won't get another host. It seems like this should also call `AddHostCreateDetails` in the insertion error case.

Same with the other error cases above this - if they're not transient errors and this method does not retry (at the Amboy job level), it seems like they should populate the host create details.

model/host/host.go
```diff
@@ -1600,6 +1600,37 @@ func (h *Host) AddSSHKeyName(name string) error {
...
+	}
+	if t == nil {
+		return errors.Errorf("host '%s' created by task '%s' execution '%d' that does not exist", h.Id, h.SpawnOptions.TaskID, h.SpawnOptions.TaskExecutionNumber)
+	}
+
+	if h.Status == evergreen.HostStarting && t.Status == evergreen.TaskStarted && !t.Aborted {
+		intent := newIntentFromHost(h)
+		grip.Info(message.Fields{
+			"message":              "inserting a host intent to replace a terminated host for a task",
+			"original_host":        h.Id,
```
Kimchelly, line 1622: nit: Most of the code base use `"host_id"` to reference fields containing host ID fields in grip logs (none of the other DB models include `_id`, but hosts do). Can we also make it consistent here, so it's `"original_host_id"` and `"new_host_id"`?

Kimchelly (changes requested):

ablack12:
> I'm undecided about limiting the number of times we loop on this. I didn't do it yet, but if someone feels strongly I'm game to add it. I think it's safe to assume it's a transient error since we know the instance was created once successfully (so it's not a user error). Even if the error does persist it won't loop infinitely because host.list will eventually reach its timeout.

I would like to press to limit the number of times we do this. Part of why we have host.list fail intentionally (with information) is so that we can present the user (and ourselves) with information. If we transiently fail even three times, I would guess something larger is wrong with AWS, and we could say something to the user to that effect.

model/host/host.go
```diff
@@ -1600,6 +1600,37 @@ func (h *Host) AddSSHKeyName(name string) error {
 	return nil
 }
 
+// HandleTerminatedHostSpawnedByTask should be called when a host spawned by a task
+// is found terminated.
+func (h *Host) HandleTerminatedHostSpawnedByTask() error {
```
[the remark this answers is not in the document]
ybrill: moved into the host-monitoring-external-state-check job

model/host/host.go
ybrill: done (though now it's private)

model/host/host_intent.go
ybrill: this went away

model/host/host_intent.go
ybrill: I think this is good now.

model/host/host_test.go
ybrill: done

model/host/host_test.go
ybrill: done

units/host_status.go
ybrill: removed the extra check

model/host/host_test.go
ybrill: fixed

model/host/host.go
ybrill: I think I've done this now. 
Though if the db call to get the task fails perhaps that's not transient and we could retry 🤷

model/host/host.go
ybrill: done

ybrill:
@ablack12
> I would like to press to limit the number of times we do this

This is done. I made it retry twice (so we get three tries).

model/host/host.go
```diff
@@ -309,6 +309,10 @@ type SpawnOptions struct {
 	// Retries is the number of times Evergreen should try to spawn this host.
 	Retries int `bson:"retries,omitempty" json:"retries,omitempty"`
 
+	// Respawns is the number of spawns remaining if the host is externally terminated after
```
Kimchelly, line 312: number of spawn attempts remaining

units/host_monitoring_external_termination.go
```diff
@@ -168,3 +167,61 @@ func handleExternallyTerminatedHost(ctx context.Context, id string, env evergree
 		return false, errors.Errorf("unexpected host status '%s'", cloudStatus)
 	}
 }
+
+// handleTerminatedHostSpawnedByTask re-creates a new intent host if possible when this host.create host fails;
```
Kimchelly, line 171: I would put a period at the end of this sentence, or start L172 without capitalization.

Kimchelly (approved):
LGTM mod 2 minor comments.

Should the wiki be updated to mention that host.create will retry and propagate failed attempt reasons to the user?

units/host_monitoring_external_termination.go
ybrill: done

model/host/host.go
ybrill: done

Kimchelly:
It seems like host.create/host.list failure reasons are an end user feature since we propagate internal failure reasons to host.list, which they have to be able to meaningfully interpret without resorting to #evergreen-users. I don't think that they need to know how many retries they're given, just let them know that they should expect host.list to include details for each spawn attempt failure.

ybrill:
I'm confused by this: I thought host.list will only tell them "host was externally terminated" once when it's run out of retries because even if we retry twice we will still only add host details once.

Kimchelly:
I didn't really look at host.list that carefully, but it seems like each host.create attempt failure tries to call `AddHostCreateDetails` and each call to `AddHostCreateDetails` pushes another error string to the details field. If host.list only exposes the last error in the host details, then it's probably fine, but I would otherwise expect them to receive error reasons for all host.create failed attempts.

ablack12:
Pushing a host create details to a task implies that that host has completely failed; in host list, we count everyHostCreateDetail object as one of the hosts we're waiting for. I haven't read this code closely but we shouldn't add a HostCreateDetail for every error, only if it all fails at the end.

ybrill:
That's definitely what I was aiming for: when this case occurs we either insert a new intent or push a HostCreateDetail.
Never both.

ybrill:
@ablack12 and/or @Kimchelly do you intent to look at this more?
(if not I'll just merge)

ablack12:
I'm good with it!

Kimchelly:
SGTM

PR #5322, opened 2022-01-19
EVG-15949 Introduce toggle to allow version control for project configs

hadjri:
[EVG-15949](https://jira.mongodb.org/browse/EVG-15949)

### Description 
Introduced a toggle on the project page when unset, we will not bother to attempt to unmarshal a project's config yaml into new config properties in order to improve performance and preserve resources.
### Testing 
Toggled in staging with toggle unset, confirmed no entry to project_configs DB occurred even when yaml was configured with new props.

### DIFF
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -130,6 +130,8 @@ Evergreen Projects
       <div class="col-lg-6">
         <input type="checkbox" id="version-control-enabled-checkbox" ng-model="settingsFormData.version_control_enabled" />
         <label for="version-control-enabled-checkbox">Enable Config Version Control</label>
+<!--        todo: write documentation and provide link below-->
+        <div class="muted small">Allow defining of <a href="">these properties</a> on this page in config yaml rather than exclusively in this UI.</div>
       </div>
     </div>
     <div id="dispatching-disabled" class="form-group">
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -331,7 +331,7 @@ Evergreen Projects
           <input type="checkbox" id="version-control-enabled-checkbox" ng-model="settingsFormData.version_control_enabled" />
           <label for="version-control-enabled-checkbox">Enable Config Version Control</label>
           <!--        todo: write documentation and provide link below-->
-          <div class="muted small">Allow defining of <a href="">these properties</a> on this page in config yaml rather than exclusively in this UI.</div>
+          <div class="muted small">Allow defining of <a href="https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#version-control">these properties</a> on this page in config yaml rather than exclusively in this UI.</div>
         </div>
       </div>
     </div>
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -126,6 +126,12 @@ Evergreen Projects
         </div>
       </div>
     </div>
+    <div id="version-control-enabled" class="form-group">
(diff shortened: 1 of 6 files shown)


service/templates/projects.html
```diff
@@ -126,6 +126,12 @@ <h2> Settings for [[displayName]]</h2>
         </div>
       </div>
     </div>
+    <div id="version-control-enabled" class="form-group">
+      <div class="col-lg-6">
+        <input type="checkbox" id="version-control-enabled-checkbox" ng-model="settingsFormData.version_control_enabled" />
+        <label for="version-control-enabled-checkbox">Enable Config Version Control</label>
```
ablack12, line 132: Can we add some help text here to indicate what this is?

ablack12 (approved):
We should make a ticket to also add this to the General Page project flag section https://www.figma.com/file/Z0rfsPjVKglRq9baDhsSux/Project-Settings-Sections?node-id=489%3A0 and add to Projects Track Multiple Branches

ablack12 (changes requested):
> Toggled in staging with toggle unset, confirmed no entry to project_configs DB occurred even when yaml was configured with new props.

Can you explain this a bit more?  I don't see a change to your code that prevents a new project config from being saved to the DB, but rather that we won't use it when looking for the project ref in FindMergedProjectRef.

hadjri:
> > Toggled in staging with toggle unset, confirmed no entry to project_configs DB occurred even when yaml was configured with new props.
> 
> Can you explain this a bit more? I don't see a change to your code that prevents a new project config from being saved to the DB, but rather that we won't use it when looking for the project ref in FindMergedProjectRef.

My mistake - what I had meant to check was that when the toggle was unset that nothing from the project configs showed up was returned from FindMergedProjectRef (tested here to retrieve the ref: https://evergreen-staging.corp.mongodb.com/api/ref/evg) and that when configs were set in the project yaml such as build baron settings, the api response reflected that if the toggle was changed to set.

service/templates/projects.html
```diff
@@ -126,6 +126,14 @@ <h2> Settings for [[displayName]]</h2>
...
+<!--        todo: write documentation and provide link below-->
+        <div class="muted small">Allow defining of <a href="">these properties</a> on this page in config yaml rather than exclusively in this UI.</div>
```
ablack12, line 134: Can we move this lower on the project page? Right now this is right below Enabled/Disabled which is the most important toggle on the page, whereas this feature is only applicable to some teams. I think a new section similar to Test Results may be reasonable.

service/templates/projects.html
ablack12: Are we going to wait to deploy this (or hide it) until documentation is done (given the href?)

service/templates/projects.html
hadjri: Sure thing. I can't reply to your second comment for some reason, maybe we should keep this PR open til we have some docs available? I am working on them right now

ablack12: Whichever you prefer is fine 👍

ablack12 (approved):

hadjri:
> We should make a ticket to also add this to the General Page project flag section https://www.figma.com/file/Z0rfsPjVKglRq9baDhsSux/Project-Settings-Sections?node-id=489%3A0 and add to Projects Track Multiple Branches

Ticket created: https://jira.mongodb.org/browse/EVG-16185

PR #5324, opened 2022-01-20
EVG-16161 filter bot users out of users/permissions role

ablack12:
[EVG-16161](https://jira.mongodb.org/browse/EVG-16161)

### Description 
Some permissions we may choose to only give to bots (for example, https://github.com/evergreen-ci/evergreen/blob/8e010917fc3774e4f32d8c2843c6d23cbed623a5/globals.go#L834) which is breaking mana's assumptions about our permissions. That being said, they don't really need to know about bots so we can just filter them out. Will also add to documentation.

### Testing 
Added a unit test.

### DIFF
--- a/model/user/db.go
+++ b/model/user/db.go
@@ -189,11 +189,15 @@ func FindByRole(role string) ([]DBUser, error) {
 	return res, errors.Wrapf(err, "error finding users with role '%s'", role)
 }
 
-func FindByRoles(roles []string) ([]DBUser, error) {
+// FindUsersByRoles returns only real users (i.e. omits API only users).
+func FindUsersByRoles(roles []string) ([]DBUser, error) {
 	res := []DBUser{}
 	err := db.FindAllQ(
 		Collection,
-		db.Query(bson.M{RolesKey: bson.M{"$in": roles}}),
+		db.Query(bson.M{
+			RolesKey:   bson.M{"$in": roles},
+			OnlyAPIKey: bson.M{"$ne": true},
+		}),
 		&res,
 	)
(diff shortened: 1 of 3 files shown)


model/user/db.go
```diff
@@ -189,11 +189,15 @@ func FindByRole(role string) ([]DBUser, error) {
 	return res, errors.Wrapf(err, "error finding users with role '%s'", role)
 }
 
-func FindByRoles(roles []string) ([]DBUser, error) {
+// FindUsersByRoles returns only real users (i.e. omits API only users).
+func FindUsersByRoles(roles []string) ([]DBUser, error) {
```
Kimchelly, line 193: The service-user-only queries include the word service in them to make it obvious that they only apply to the bots. Can we call this `FindHumanUsersByRoles` (or something to that effect) to distinguish it from queries that also include service users?

model/user/db.go
```diff
@@ -189,11 +189,15 @@ func FindByRole(role string) ([]DBUser, error) {
 	return res, errors.Wrapf(err, "error finding users with role '%s'", role)
 }
 
-func FindByRoles(roles []string) ([]DBUser, error) {
+// FindUsersByRoles returns only real users (i.e. omits API only users).
```
Kimchelly, line 192: Can we document that this finds real users matching the given roles?

Kimchelly (changes requested):

model/user/db.go
ablack12: Yes perfect I couldn't think of a good name for this

Kimchelly (approved):

bynn (approved):

PR #5326, opened 2022-01-20
EVG-12923 tweak single-host task group dependencies logging

ablack12:

### Description 
To help take over investigation of this, adding some logging (namely the dependency in question, and the stack trace). Also logging at a lower level.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1599,15 +1599,17 @@ func checkResetSingleHostTaskGroup(t *task.Task, caller string) error {
 		return errors.Wrapf(err, "can't get task group for task '%s'", t.Id)
 	}
 	taskSet := map[string]bool{}
-	for _, t := range tasks {
-		taskSet[t.Id] = true
-		for _, dep := range t.DependsOn {
+	for _, tgTask := range tasks {
+		taskSet[tgTask.Id] = true
+		for _, dep := range tgTask.DependsOn {
 			if taskSet[dep.TaskId] && dep.Unattainable {
-				grip.Info(message.Fields{
-					"message": "task group task was blocked on an earlier task group task after reset",
-					"task":    t.Id,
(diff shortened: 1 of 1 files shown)


malikchaya2 (approved):

bynn (approved):

model/task_lifecycle.go
```diff
@@ -1599,15 +1599,17 @@ func checkResetSingleHostTaskGroup(t *task.Task, caller string) error {
...
+	for _, tgTask := range tasks {
+		taskSet[tgTask.Id] = true
+		for _, dep := range tgTask.DependsOn {
 			if taskSet[dep.TaskId] && dep.Unattainable {
-				grip.Info(message.Fields{
-					"message": "task group task was blocked on an earlier task group task after reset",
-					"task":    t.Id,
-					"ticket":  "EVG-12923",
-				})
+				grip.Debug(message.WrapError(errors.New(
```
ybrill, line 1606: Are you certain this will include the stack? 
I would have assumed we'd need to do something like
```
grip.Debug(message.WrapError(errors.WithStack(errors.New(...
```
so [WithStack](https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L143-L153) annotates the error with a stack trace.
Alternatively, the stack trace could be added as another field in the message ([this](https://github.com/evergreen-ci/evergreen/blob/825ea809950bd90fe225b6ca662bd8f0e0816fa2/model/task_lifecycle.go#L610) is an example)

ablack12: I'm fairly certain that WrapError handles it with the extended field (I want to say [here|https://github.com/mongodb/grip/blob/main/message/error_message.go#L108] but I don't know grip very well). That being said I'm now curious so I may just try it like this.

PR #5332, opened 2022-01-25
EVG-15654 Refactor PR patch testing 

bynn:
[EVG-15654](https://jira.mongodb.org/browse/EVG-15654)

### Description 
added another checkbox to enable manual testing with 'evergreen patch' or 'evergreen retry' 
current checkbox only disables automatic testing 
having both checked off has the same effect as having the previous checkbox ticked off 

### Testing 
https://github.com/evergreen-ci/commit-queue-sandbox/pull/336#partial-pull-merging

image

image


model/patch/github.go
```diff
@@ -213,6 +224,10 @@ func (g *githubIntent) RequesterIdentity() string {
 	return evergreen.GithubPRRequester
 }
 
+func (g *githubIntent) IsCalledBy() string {
```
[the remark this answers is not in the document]
bynn: i get a duplicate field and method error for that so i changed it to IsCalledBy

model/project_ref.go
```diff
@@ -268,9 +270,17 @@ func (p *ProjectRef) IsDispatchingDisabled() bool {
 }
 
 func (p *ProjectRef) IsPRTestingEnabled() bool {
+	return p.IsAutoPRTestingEnabled() || p.IsManualPRTestingEnabled()
```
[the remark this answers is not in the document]
bynn: just for clarification:
does that mean it would be better to something like `utility.FromBoolPtr(p.PRTestingEnabled) || utility.FromBoolPtr(p.ManualPRTestingEnabled)` rather than using the functions?

rest/route/github.go
```diff
@@ -239,17 +240,17 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
-			if triggersRetry(*event.Action, *event.Comment.Body) {
+			if triggersPatch(*event.Action, *event.Comment.Body) {
 				grip.Info(message.Fields{
 					"source":    "github hook",
 					"msg_id":    gh.msgID,
 					"event":     gh.eventType,
 					"repo":      *event.Repo.FullName,
 					"pr_number": *event.Issue.Number,
 					"user":      *event.Sender.Login,
-					"message":   "retry triggered",
```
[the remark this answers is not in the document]
bynn: i think it may be helpful in the future

model/patch/cli_intent.go
```diff
@@ -169,6 +169,10 @@ func (g *cliIntent) RequesterIdentity() string {
 	return evergreen.PatchVersionRequester
 }
 
+func (g *cliIntent) IsCalledBy() string {
```
ablack12, line 172: Can you comment why this isn't populated? Not relevant to CLI?

model/patch/github.go
```diff
@@ -24,6 +24,12 @@ const (
 	GithubIntentType = "github"
 )
 
+// CalledBy can be either auto or manual.
+const (
+	CalledAutomatically = "auto"
```
ablack12, line 29: More conventional wording would be AutomatedCaller / ManualCaller or something

model/patch/github.go
```diff
@@ -77,6 +83,9 @@ type githubIntent struct {
 
 	// IntentType indicates the type of the patch intent, e.g. GithubIntentType
 	IntentType string `bson:"intent_type"`
+
+	// CalledBy indicates what created the intent
```
ablack12, line 87: "whether the intent was created automatically by Evergreen or manually" or something like that is more accurate, to distinguish it from IntentType

model/project_ref.go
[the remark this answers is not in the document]
ablack12: I think this isn't relevant since both of those functions return regular booleans

service/templates/projects.html
```diff
@@ -447,13 +447,17 @@ <h3> GitHub Pull Request Testing</h3>
...
       <div id="patch-variants-list-header" class="form-group" ng-show="prTestingEnabled || prTestingConflicts.length === 0">
-        <div class="col-lg-6">
+        <div class="col-lg-5">
           <input type="checkbox" id="prtesting-checkbox" ng-model="settingsFormData.pr_testing_enabled" />
-          <label for="prtesting-checkbox">Enable Github PR Testing</label>
+          <label for="prtesting-checkbox">Enable Automatic Github PR Testing</label>
+        </div>
+        <div class="col-lg-5">
+          <input type="checkbox" id="manualprtesting-checkbox" ng-model="settingsFormData.manual_pr_testing_enabled" />
+          <label for="prtesting-checkbox">Enable Manual Github PR Testing</label>
```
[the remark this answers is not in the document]
bynn: image

model/project_ref.go
```diff
@@ -1205,16 +1216,29 @@ func FindDownstreamProjects(project string) ([]ProjectRef, error) {
...
+				p.checkDefaultLogger()
+				return &p, nil
+			}
+		case patch.CalledAutomatically:
+			if p.IsAutoPRTestingEnabled() {
+				p.checkDefaultLogger()
+				return &p, nil
+			}
+		default:
+			if p.IsPRTestingEnabled() {
```
ablack12, line 1238: Need to also consider this at 1260. A way to move this to a helper would be IsPRTestingEnabledByCaller(caller string) so you don't need to case each time you need to check

model/patch/github.go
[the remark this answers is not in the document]
malikchaya2: in that case, maybe "GetCalledBy"?

model/project_ref.go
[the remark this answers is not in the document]
malikchaya2: oh my bad, my brain skipped over the fact that this is a function and not just the field :see_no_evil:

service/templates/projects.html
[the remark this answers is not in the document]
malikchaya2: thank you!

rest/route/github.go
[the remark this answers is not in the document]
ablack12: +1

rest/route/github.go
```diff
@@ -239,17 +240,17 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+					"message":   "patch triggered",
 				})
-				if err := gh.retryPRPatch(ctx, event.Repo.Owner.GetLogin(), event.Repo.GetName(), event.Issue.GetNumber()); err != nil {
+				if err := gh.createPRPatch(ctx, event.Repo.Owner.GetLogin(), event.Repo.GetName(), patch.CalledManually, event.Issue.GetNumber()); err != nil {
```
ablack12, line 256: "evergreen retry" should still be patch.CalledAutomatically actually in terms of workflow; I know it is user triggered but its part of the automated workflow as opposed to part of the "only if manual is checked will this work" workflow.

ablack12 (changes requested):

model/patch/github.go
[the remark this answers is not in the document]
ablack12: +1

rest/route/github.go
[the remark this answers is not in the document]
bynn: done

rest/route/github.go
bynn: done (changed what triggersPatch returns) not sure if that's what we want to do but if we add more types later it would be easiest this way but it makes sense because we should be able to retry even if only the auto is enabled

model/project_ref.go
ablack12: What I mean is, rather than doing this switch statement here to determine which function to return, you can have IsPRTestingEnabledByCaller do:
```
case patch.ManualCaller:
	return p.IsManualPRTestingEnabled()
case patch.AutomatedCaller:
	return  p.IsAutoPRTestingEnabled() 
default:
	return p.IsPRTestingEnabled()
```
and then this function call just do
```
if  p.IsPRTestingEnabledByCaller(calledBy) { 
           p.checkDefaultLogger()
	   return &p, nil
```

model/project_ref.go
```diff
@@ -1604,13 +1628,14 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
 				"$set": bson.M{
-					projectRefPRTestingEnabledKey:      p.PRTestingEnabled,
-					projectRefGithubChecksEnabledKey:   p.GithubChecksEnabled,
-					projectRefGithubTriggerAliasesKey:  p.PatchTriggerAliases,
-					projectRefGitTagVersionsEnabledKey: p.GitTagVersionsEnabled,
-					ProjectRefGitTagAuthorizedUsersKey: p.GitTagAuthorizedUsers,
-					ProjectRefGitTagAuthorizedTeamsKey: p.GitTagAuthorizedTeams,
-					projectRefCommitQueueKey:           p.CommitQueue,
+					projectRefPRTestingEnabledKey:       p.PRTestingEnabled,
+					projectRefManualPRTestingEnabledKey: p.ManualPRTestingEnabled,
```
ablack12, line 1632: FYI will need to make a small UI ticket to add this to the new UI

model/project_ref_test.go
```diff
@@ -1138,54 +1138,54 @@ func TestFindOneProjectRefByRepoAndBranchWithPRTesting(t *testing.T) {
...
+	projectRef, err = FindOneProjectRefByRepoAndBranchWithPRTesting("mongodb", "mci", "yours", patch.AutomatedCaller)
 	assert.NoError(err)
 	assert.NotNil(projectRef)
 	assert.Equal("yours", projectRef.Branch)
 	assert.True(projectRef.IsHidden())
 	firstAttemptId := projectRef.Id
 
 	// verify we return the same hidden project
-	projectRef, err = FindOneProjectRefByRepoAndBranchWithPRTesting("mongodb", "mci", "yours")
+	projectRef, err = FindOneProjectRefByRepoAndBranchWithPRTesting("mongodb", "mci", "yours", patch.AutomatedCaller)
```
ablack12, line 1188: Can you write a new test case for ManualCaller?

rest/route/github.go
```diff
@@ -258,7 +261,7 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
 						"repo":      *event.Repo.FullName,
 						"pr_number": *event.Issue.Number,
 						"user":      *event.Sender.Login,
-						"message":   "can't retry PR",
+						"message":   "can't create PR",
```
ablack12, line 264: can we also add the  triggering comment here?

rest/route/github.go
```diff
@@ -239,17 +241,18 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+			if triggerPatch {
 				grip.Info(message.Fields{
 					"source":    "github hook",
 					"msg_id":    gh.msgID,
 					"event":     gh.eventType,
 					"repo":      *event.Repo.FullName,
 					"pr_number": *event.Issue.Number,
 					"user":      *event.Sender.Login,
-					"message":   "retry triggered",
+					"message":   fmt.Sprintf("%s triggered", *event.Comment.Body),
```
ablack12, line 253: nit: "'%s' triggered" will read more clearly

rest/route/github.go
```diff
@@ -705,12 +708,19 @@ func (gh *githubHookApi) tryDequeueCommitQueueItemForPR(pr *github.PullRequest)
 	return nil
 }
 
-func triggersRetry(action, comment string) bool {
+func triggersPatch(action, comment string) (bool, string) {
```
ablack12, line 711: Can you comment this to indicate what and why we return a string value? (Should this even return a boolean now? "" == false)

service/templates/projects.html
```diff
@@ -447,13 +447,18 @@ <h3> GitHub Pull Request Testing</h3>
...
+          <div class="muted small">Patches can be run manually by commenting 'evergreen patch' on the PR.</div>
```
ablack12, line 457: can we include "even if automated testing isn't configured" just to be very clear?

rest/route/github.go
ablack12: Although actually we could return "" from retryComment so that a patch is created for either manual or automated 👀

model/project_ref_test.go
bynn: done

malikchaya2 (approved):

rest/route/github.go
ablack12: Can you still add a comment to explain the two return values then? And can we add an explicit patch.AllCallers = ""

ablack12 (commented):
Sorry small last thing!

rest/route/github.go
bynn: i added it for retrycomment but not for the other "" because we should be returning something like nil when we return false

rest/route/github.go
```diff
@@ -708,6 +708,8 @@ func (gh *githubHookApi) tryDequeueCommitQueueItemForPR(pr *github.PullRequest)
 	return nil
 }
 
+// The bool value returns whether the patch should be created or not
```
ablack12, line 711: nit: comments with multiple sentences should end with periods

ablack12 (approved):
LGTM with nit

PR #5333, opened 2022-01-25
EVG-5457 validate attach.xunit_results not used in teardown group

ablack12:
[EVG-5457](https://jira.mongodb.org/browse/EVG-5457)

### Description 
We already validate two of the three attach commands; adding the third for consistency.

### Testing 
Ran against all enabled projects to ensure this isn't introducing a new error.


Kimchelly (commented):
Should [this test](https://github.com/evergreen-ci/evergreen/blob/9fd4ebddf1d44a9e3018257d5b9de1ac47a5755b/validator/project_validator_test.go#L2110-L2137) be updated to check that all the attach commands are not allowed?

Kimchelly (approved):

PR #5335, opened 2022-01-25
EVG-15749 Error when shell.exec is missing or misspelling params

hadjri:
[EVG-15749](https://jira.mongodb.org/browse/EVG-15749)

### Description 
Validator will now fail when params is missing from shell.exec command.
### Testing 
Updated unit test. Ran this new validation on all existing enabled projects and none hit this new error.


validator/project_validator_test.go
```diff
@@ -1421,6 +1421,22 @@ func TestValidatePluginCommands(t *testing.T) {
 			So(validatePluginCommands(project), ShouldNotResemble, ValidationErrors{})
 			So(len(validatePluginCommands(project)), ShouldEqual, 1)
 		})
+		Convey("an error should be thrown if a shell.exec command is missing params", func() {
+			project := &model.Project{
+				Functions: map[string]*model.YAMLCommandSet{
```
malikchaya2, line 1426: since the ticket originally came out of a struggle with "params" being misspelled, would it maybe be possible to add a test that mimics that so that we know for sure that it covers that use case?

malikchaya2 (changes requested):

agent/command/shell.go
```diff
@@ -82,6 +82,10 @@ func (*shellExec) Name() string { return evergreen.ShellExecCommandName }
 
 // ParseParams reads in the command's parameters.
 func (c *shellExec) ParseParams(params map[string]interface{}) error {
+	if params == nil {
```
ablack12, line 85: Does ParseParams get called in the validator? Or else can we add something like https://github.com/evergreen-ci/evergreen/blob/ddcfe0c158c9db27872d9735d492d8c3906264b5/validator/project_validator.go#L951 ?

ablack12 (changes requested):

agent/command/shell.go
hadjri: ParseParams gets called in the validator - it begins with command.Render at this line and cascades down to call it: https://github.com/evergreen-ci/evergreen/blob/ddcfe0c158c9db27872d9735d492d8c3906264b5/validator/project_validator.go#L932

ablack12: Ah i guess we don't have script in parseParams bc it's just a warning. It would be nice if that could be an error too since idk what the point is without a script (could make that a separate ticket if you prefer)

ablack12 (approved):

malikchaya2 (approved):

agent/command/shell.go
```diff
@@ -82,6 +82,13 @@ func (*shellExec) Name() string { return evergreen.ShellExecCommandName }
...
+		return errors.New("params cannot be nil")
+	}
+	if params["script"] == nil {
+		return errors.Errorf("command '%s' specified without a script", evergreen.ShellExecCommandName)
```
ablack12, line 89: I think this can just return the error "script cannot be nil" since https://github.com/evergreen-ci/evergreen/blob/d90ee9d7f77d5d29f58f8231b9e5217206091f07/validator/project_validator.go#L943 handles the command name.

ablack12 (commented):
If you haven't done this already, can we re-run this validation (that includes script) against existing yamls to verify that it's not a warning for a reason?

hadjri:
> If you haven't done this already, can we re-run this validation (that includes script) against existing yamls to verify that it's not a warning for a reason?

Yes, ran the validation on existing project and we do indeed have some new breaking changes in projects (for "script cannot be nil").  Perhaps it's best to keep this as a warning, since it doesn't seem like we need to prevent a build from succeeding if a shell.exec script isnt defined?

ablack12:
Yes that sounds good, thanks for looking into it 👍  can you add a comment explaining that some existing projects rely on this being a warning so we aren't confused in the future?

PR #5337, opened 2022-01-26
EVG-15915 Fix terminated spawn host is marked as running

malikchaya2:
[EVG-15915](https://jira.mongodb.org/browse/EVG-15915)

### Description 
A user ran into an issue where a host was labeled as running when they tried to click on the trash button of an already terminating host. 

### Testing 
I tried stopping and then pausing a few hosts on staging with and without my changes. The ones that stuck around as running (despite actually being stopped) are ones spawned without my changes. 
image

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -1162,13 +1162,13 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 	if err != nil {
 		// fetch the instance to get the current status, since it may have changed
 		// since we saved the previous status
-		currentStatus, err := m.GetInstanceStatus(ctx, h)
-		if err != nil {
-			return errors.Wrap(err, "error getting instance status after stopping error")
+		currentStatus, currentStatusErr := m.GetInstanceStatus(ctx, h)
+		if currentStatusErr != nil {
+			return errors.Wrap(currentStatusErr, "error getting instance status after stopping error")
 		}
 
-		if err2 := h.SetStatus(currentStatus.String(), user, ""); err2 != nil {
-			return errors.Wrapf(err2, "failed to revert status from stopping to '%s'", currentStatus)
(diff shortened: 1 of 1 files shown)


cloud/ec2.go
```diff
@@ -1161,8 +1160,15 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 		InstanceIds: []*string{aws.String(h.Id)},
 	})
 	if err != nil {
-		if err2 := h.SetStatus(prevStatus, user, ""); err2 != nil {
-			return errors.Wrapf(err2, "failed to revert status from stopping to '%s'", prevStatus)
+		// fetch the instance to get the current status, since it may have changed
+		// since we saved the previous status
+		instance, err = m.client.GetInstanceInfo(ctx, h.Id)
```
ablack12, line 1165: Can we just call m.GetInstanceStatus? This will also handle translating the status to an Evergreen status

malikchaya2: I was looking for something like that! thank you!

cloud/ec2.go
```diff
@@ -1161,8 +1160,15 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 		InstanceIds: []*string{aws.String(h.Id)},
 	})
 	if err != nil {
-		if err2 := h.SetStatus(prevStatus, user, ""); err2 != nil {
-			return errors.Wrapf(err2, "failed to revert status from stopping to '%s'", prevStatus)
+		// fetch the instance to get the current status, since it may have changed
+		// since we saved the previous status
+		currentStatus, err := m.GetInstanceStatus(ctx, h)
+		if err != nil {
```
ablack12, line 1166: This should also be an err2 instead of err so we don't return nil at 1173 (same below)

malikchaya2: thanks, fixed

ablack12 (approved):
Solid rename 👍

PR #5338, opened 2022-01-27
EVG-16199 check pprof

ybrill:
[EVG-16199](https://jira.mongodb.org/browse/EVG-16199)

### Description 
When the number of goroutines passes a threshold we will write the goroutine report to a file. If the most recent file is too new or if we've already written out a bunch of them we skip writing the file.

My thinking is that we could apply this on top of main and deploy it to a few app servers. **I don't intend to actually merge this**.

I considered using a cron script external to the app but integrating into the evergreen app has the advantages that 1) it made it really easy to get the goroutine count (with `runtime.NumGoroutine()`) and (perhaps more importantly) 2) if the pprof request was through an API call into the app we might exacerbate the problem or not get an answer.
One concern was that maybe the app isn't running amboy jobs during these incidents, but from Splunk it appears we are.

### Testing 
I put it up on one of the staging app servers and the goroutine report was written at the appropriate intervals.

### DIFF
--- a/units/check_pprof.go
+++ b/units/check_pprof.go
@@ -1,0 +1,121 @@
+package units
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"runtime/pprof"
+	"time"
+
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+)
+
+const (
+	pprofCheckName     = "pprof-check"
+	goroutineThreshold = 1500
+	pprofPath          = "/srv/evergreen/pprof"
+	totalBytesLimit    = 1000000000
+	checkInterval      = 5 * time.Minute
+)
+
+func init() {
+	registry.AddJobType(pprofCheckName,
+		func() amboy.Job { return makePprofCheckJob() })
+}
+
+type pprofCheckJob struct {
+	job.Base `bson:"job_base" json:"job_base" yaml:"job_base"`
+}
+
+func makePprofCheckJob() *pprofCheckJob {
+	j := &pprofCheckJob{
+		Base: job.Base{
+			JobType: amboy.JobType{
+				Name:    pprofCheckName,
+				Version: 0,
+			},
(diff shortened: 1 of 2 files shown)


units/check_pprof.go
```diff
@@ -0,0 +1,118 @@
...
+		j.AddError(errors.Wrapf(err, "creating pprof file '%s'", pprofFilePath))
+		return
+	}
+
+	if err := pprof.Lookup("goroutine").WriteTo(pprofFile, 2); err != nil {
+		j.AddError(errors.Wrap(err, "writing pprof to file"))
+	}
+
+	grip.Info(message.Fields{
+		"message":        "wrote pprof to file",
```
Kimchelly, line 114: Include job ID as a field.

units/check_pprof.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	var totalSize int64
+	for _, file := range contents {
+		age := time.Now().Sub(file.ModTime())
+		if age < checkInterval {
+			grip.Info(message.Fields{
+				"message":            "not collecting pprof",
+				"reason":             "most recent file is newer then the threshold",
+				"mostRecentFile":     file.Name(),
```
Kimchelly, line 83: nit: it doesn't matter that much since this is a temporary piece of code, but snake case is more common for these JSON messages.

units/check_pprof.go
```diff
@@ -0,0 +1,118 @@
...
+				"ageInMinutes":       age.Minutes(),
+				"thresholdInMinutes": checkInterval.Minutes(),
+			})
+			return
+		}
+		totalSize += file.Size()
+	}
+
+	if totalSize > totalBytesLimit {
+		grip.Info(message.Fields{
```
Kimchelly, line 93: Include job ID.

Kimchelly (approved):
LGTM mod minor things.

units/check_pprof.go
ybrill: done

units/check_pprof.go
ybrill: done

units/check_pprof.go
ybrill: done

ybrill:
rebased on main's [HEAD](https://github.com/evergreen-ci/evergreen/commit/36ab61f308073a01f57b8f9f22d9ef2ee8efd1a0)

PR #5339, opened 2022-01-27
EVG-16128 warn users about immutable ID if copying perf project

ablack12:
[EVG-16128](https://jira.mongodb.org/browse/EVG-16128)

### Description 
Warn users from copying a project with different ID and identifier that they may want to make them the same if relying on Performance. (This is likely a temporary measure, pending https://jira.mongodb.org/browse/EVG-16179, so not currently planning on extending to the new UI).

### Testing 
image

### DIFF
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -1475,7 +1475,7 @@ mciModule.directive("adminNewProject", function () {
         '<div class="col-lg-12">' +
           "Optionally enter immutable project ID " +
           '<div class="muted small">' +
-            "(Used by Evergreen internally and defaults to a random hash; should only be user-specified with good reason. Cannot be changed!)" +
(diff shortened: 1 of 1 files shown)


public/static/js/projects.js
```diff
@@ -1321,7 +1321,7 @@ mciModule.controller(
       };
     }
 
-    function comebineDateTime(date, time) {
+    function combineDateTime(date, time) {
```
Kimchelly, line 1324: 💯

Kimchelly (approved):

public/static/js/projects.js
malikchaya2: comebine has a nice ring to it.

malikchaya2 (approved):

PR #5340, opened 2022-01-27
EVG-14645: forward arbitrary SSH args and add dry run flag

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-14645

### Description 
Allow users to pass whatever flags they want to the SSH binary as additional parameters. I also added a dry run flag to display the SSH command that will run without actually running the command, which is useful for debugging. `evergreen host rsync` has a similar dry run flag.

`evergreen host ssh -i <identity_file>` is now redundant since it's equivalent to `evergreen host ssh --host <host_id> -- -i <identity_file>`, but I kept it for backward compatibility.

### Testing 
SSH is hard to test, so I just manually tested that it still works the way it's used today and the additional arguments are forwarded as expected.

### DIFF
--- a/operations/host_spawn.go
+++ b/operations/host_spawn.go
@@ -635,7 +635,7 @@ Examples:
 			}
 
 			if dryRun {
-				fmt.Printf("Going to execute the following command: %s\n", strings.Join(args, " "))
+				fmt.Printf("Going to execute the following command:\n%s\n", strings.Join(args, " "))
 				return nil
 			}
 
@@ -1548,7 +1548,7 @@ func hostFindBy() cli.Command {
 			hostUser := utility.FromStringPtr(host.StartedBy)
 			link := fmt.Sprintf("%s/host/%s", conf.UIServerHost, hostId)
 			fmt.Printf(`
-		 ID : %s
+	     ID : %s
 	   Link : %s
 	   User : %s
 `, hostId, link, hostUser)
@@ -1714,7 +1714,7 @@ func buildRsyncCommand(opts rsyncOpts) (*jasper.Command, error) {
 		cmdWithoutDryRun := make([]string, len(args[:dryRunIndex]), len(args)-1)
 		_ = copy(cmdWithoutDryRun, args[:dryRunIndex])
 		cmdWithoutDryRun = append(cmdWithoutDryRun, args[dryRunIndex+1:]...)
-		fmt.Printf("Going to execute the following command: %s\n", strings.Join(cmdWithoutDryRun, " "))
+		fmt.Printf("Going to execute the following command:\n%s\n", strings.Join(cmdWithoutDryRun, " "))
 	}
 	logLevel := level.Info
 	stdout, err := send.NewPlainLogger("stdout", send.LevelInfo{Threshold: logLevel, Default: logLevel})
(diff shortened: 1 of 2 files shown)


operations/host_spawn.go
```diff
@@ -563,23 +563,43 @@ func hostStart() cli.Command {
 func hostSSH() cli.Command {
 	const (
 		identityFlagName = "identity_file"
+		dryRunFlagName   = "dry-run"
 	)
 
 	return cli.Command{
 		Name:  "ssh",
 		Usage: "ssh into a spawn host",
+		UsageText: `
```
ablack12, line 572: This is awesome, I didn't even know we had this.

operations/host_spawn.go
```diff
@@ -563,23 +563,43 @@ func hostStart() cli.Command {
...
+* SSH into a host (i-abcdef12345) by ID:
+	evergreen host ssh --host i-abcdef12345
+
+* As a special case, if you pass a single argument, it will be interpreted as the host ID:
+	evergreen host ssh i-abcdef12345
+
+* SSH into a host (i-abcdef12345) and pass additional arguments to the SSH binary to use verbose mode (-vvv) and run a command (echo hello world):
+	evergreen host ssh --host i-abcdef12345 -- -vvv "echo hello world"
+`,
+		ArgsUsage: "[-- other_ssh_arguments]",
```
ablack12, line 585: I don't see this in the output of `evergreen ssh -h` ; do you know where/if this gets used?

operations/host_spawn.go
```diff
@@ -607,6 +627,18 @@ func hostSSH() cli.Command {
...
 			}
+
+			// Append the arbitrary additional SSH args unless it's the special
+			// case where it's a single arg and it's the host ID.
+			if c.NArg() > 1 || c.NArg() == 1 && c.Args().Get(0) != hostID {
+				args = append(args, c.Args()...)
+			}
+
+			if dryRun {
+				fmt.Printf("Going to execute the following command: %s\n", strings.Join(args, " "))
```
ablack12, line 638: can the command be on a new line?

operations/host_spawn.go
```diff
@@ -1516,7 +1548,7 @@ func hostFindBy() cli.Command {
 			hostUser := utility.FromStringPtr(host.StartedBy)
 			link := fmt.Sprintf("%s/host/%s", conf.UIServerHost, hostId)
 			fmt.Printf(`
-	     ID : %s
+		 ID : %s
```
ablack12, line 1551: was this intentional?

malikchaya2 (approved):

operations/host_spawn.go
Kimchelly: Hmm I think it might not do anything. I swear it did something when I originally added it, but it's probably not necessary since the usage text includes examples of how to use the command, so I removed this.

operations/host_spawn.go
Kimchelly: Done.

I also did this for the rsync command to keep it consistent.

operations/host_spawn.go
Kimchelly: No, vim just auto-reindents files when files have mixed tabs/spaces used in indentation and this output string has a mix of tabs/spaces. I reverted this change.

ablack12 (approved):

PR #5343, opened 2022-01-28
EVG-16164: Add admin setting for max parallel pod request limit

hhoke:
[EVG-16164](https://jira.mongodb.org/browse/EVG-<number>)

### Description 
Add, but do not use, admin setting for max parallel pod request limit. This is analogous to the host creation rate throttle in hostinit.

### Testing 
Local tests on laptop. No new tests, but extended existing tests with new fields.


config_db.go
```diff
@@ -103,7 +103,8 @@ var (
 	hostInitMaxTotalDynamicHostsKey = bsonutil.MustHaveTag(HostInitConfig{}, "MaxTotalDynamicHosts")
 	hostInitS3BaseURLKey            = bsonutil.MustHaveTag(HostInitConfig{}, "S3BaseURL")
 
-	podInitS3BaseURLKey = bsonutil.MustHaveTag(PodInitConfig{}, "S3BaseURL")
+	podInitS3BaseURLKey       = bsonutil.MustHaveTag(PodInitConfig{}, "S3BaseURL")
+	maxParallelPodRequestsKey = bsonutil.MustHaveTag(PodInitConfig{}, "MaxParallelPodRequests")
```
Kimchelly, line 107: Can this be made consistent with the above key to be `podInitMaxParallelPodRequestsKey`?

config_podinit.go
```diff
@@ -46,13 +47,18 @@ func (c *PodInitConfig) Set() error {
...
+			maxParallelPodRequestsKey: c.MaxParallelPodRequests,
 		},
 	}, options.Update().SetUpsert(true))
 
 	return errors.Wrapf(err, "error updating section %s", c.SectionId())
 }
 
 func (c *PodInitConfig) ValidateAndDefault() error {
+	if c.MaxParallelPodRequests <= 0 {
+		// TODO: Determine empirically if this is indeed reasonable
```
Kimchelly, line 60: Can you add a Jira ticket to act on this TODO?

I would set the default if it's equal to 0, but error if the request limit is negative.

model/pod/pod.go
```diff
@@ -57,6 +57,11 @@ const (
 	StatusTerminated Status = "terminated"
 )
 
+var ActiveStatus = []Status{
```
Kimchelly, line 60: Add doc comment.

rest/model/admin.go
```diff
@@ -908,7 +910,8 @@ func (a *APIPodInitConfig) BuildFromService(h interface{}) error {
 
 func (a *APIPodInitConfig) ToService() (interface{}, error) {
 	return evergreen.PodInitConfig{
-		S3BaseURL: utility.FromStringPtr(a.S3BaseURL),
+		S3BaseURL:              utility.FromStringPtr(a.S3BaseURL),
+		MaxParallelPodRequests: a.MaxParallelPodRequests,
```
Kimchelly, line 914: I would add an assertion for this extra field in the admin REST model test like [this one](https://github.com/evergreen-ci/evergreen/blob/52a73f7fee9d3f8cf56cf52b96e811b213cc955a/rest/model/admin_test.go#L136) for `BuildFromService` and [this one](https://github.com/evergreen-ci/evergreen/blob/52a73f7fee9d3f8cf56cf52b96e811b213cc955a/rest/model/admin_test.go#L218) for `ToService`.

I would also add an assertion for [the REST route test](https://github.com/evergreen-ci/evergreen/blob/52a73f7fee9d3f8cf56cf52b96e811b213cc955a/rest/route/admin_test.go#L150).

service/templates/admin.html
```diff
@@ -602,6 +602,10 @@ <h2 class="modal-title">[[modalTitle]]</h2>
 										<label>S3 Download Base URL</label>
 										<input type="text" ng-model="Settings.pod_init.s3_base_url">
 									</md-input-container>
+									<md-input-container class="control" style="width:45%;">
+										<label>Max Total Dynamic Hosts</label>
```
Kimchelly, line 606: Max Parallel Pod Requests

Kimchelly (changes requested):

config_podinit.go
hhoke: Jira ticket is added. We do the same `<=` thing here in hostinit, so I think we should not error on negatives.

model/pod/pod.go
Kimchelly: nit: doc comments generally don't need to talk about type information like saying it's a slice of statuses, since that's type information that's readily available just by using a typed language. Since you haven't defined what "active" means, I would phrase it as something more along the lines of "ActiveStatus represents all statuses in which the pod exists in the cloud provider."

Alternatively, `ActiveStatus` could just be removed since it seems like it's not being used.

config_podinit.go
Kimchelly: I would argue that hostinit is doing the wrong thing because negatives are invalid values. Negatives should not be equivalent to the default because negative numbers tend to be associated with "unlimited" or other connotations. Realistically, if a person ever set the value to negative, it should be a different effect from setting it to 0 (which is generally understood to mean "the default").

I would correct both to make it invalid to have negative values.

rest/model/admin.go
Kimchelly: Still missing the assertion for the REST route test.

Kimchelly (changes requested):

model/pod/pod.go
hhoke: I think we might end up using something like this in the future but will delete for now.

config_podinit.go
hhoke: Sounds reasonable to me -- I've seen -1 used as a sentinel value a lot but rarely as a default.

config_hostinit.go
```diff
@@ -61,20 +62,30 @@ func (c *HostInitConfig) Set() error {
 }
 
 func (c *HostInitConfig) ValidateAndDefault() error {
-	if c.HostThrottle <= 0 {
+	catcher := grip.NewSimpleCatcher()
+	if c.HostThrottle == 0 {
 		c.HostThrottle = defaultHostThrottle
+	} else if c.HostThrottle < 0 {
+		catcher.Add(errors.New("HostThrottle cannot be negative"))
```
Kimchelly, line 69: This statement can be simplified down to `catcher.New("HostThrottle cannot be negative")`.

nit: I would just simplify the entire else-if block down to `catcher.NewWhen(c.HostThrottle < 0, "HostThrottle cannot be negative")`.

Same thing with the other negative value checks.

Kimchelly (changes requested):

Kimchelly (approved):

PR #5344, opened 2022-01-28
EVG-15935 fix unmarshal yaml strict for validator

ablack12:
[EVG-15935](https://jira.mongodb.org/browse/EVG-15935)

### Description 
Before, we passed the project yaml directly to the API server to validate, so CheckYamlStrict worked. Now however, we LoadProjectInto locally in order to capture any included files, so by the time we get to CheckYamlStrict, we have a broken yaml file.

### Testing 
Tested locally with a broken file (using --quiet and without).
```
evergreen (EVG-15935) $ egl validate self-tests.yml
Banner: This is an important notification
WARNING: error unmarshalling strictly: load project error(s): yaml: unmarshal errors:
  line 531: mapping key "commands" already defined at line 529
<continues on to validate the file>
```
```
evergreen (EVG-15935) $ egl validate self-tests.yml -q
Banner: This is an important notification
self-tests.yml is valid
```
_Alternatively_ we can just make this a breaking error. I'm a bit torn; it doesn't seem like this breaks any existing configurations, but we typically don't use unmarshal strict. If we make this a real validation error, we should start using strict _everywhere_ to avoid projects getting into a bad state.

### DIFF
--- a/operations/validate.go
+++ b/operations/validate.go
@@ -8,12 +8,14 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/evergreen-ci/evergreen/util"
+
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/validator"
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 func Validate() cli.Command {
@@ -152,9 +154,11 @@ func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model
 	project *model.Project) (*model.ParserProject, validator.ValidationErrors) {
 	errs := validator.ValidationErrors{}
 	pp, _, err := model.LoadProjectInto(ctx, data, opts, "", project)
+
 	if err != nil {
-		if opts.UnmarshalStrict {
-			// try it without unmarshal strict; if the error is from being strict, warn but don't error.
+		// If the error came from unmarshalling strict, try it again without strict to verify if
+		// it's a legitimate unmarshal error or just an error from strict (which should be a warning)
+		if strings.Contains(err.Error(), util.UnmarshalStrictError) {
 			opts.UnmarshalStrict = false
 			pp, _, err2 := model.LoadProjectInto(ctx, data, opts, "", project)
 			if err2 == nil {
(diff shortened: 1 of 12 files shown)


operations/validate.go
```diff
@@ -115,10 +116,15 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
...
 		ReadFileFrom: model.ReadFromLocal,
 	}
-	pp, _, err := model.LoadProjectInto(ctx, confFile, opts, "", project)
-	if err != nil {
-		return errors.Wrapf(err, "%s is an invalid configuration", path)
+	if !quiet {
+		opts.UnmarshalStrict = true
+	}
+	pp, validationErrs := loadProjectIntoWithValidation(ctx, confFile, opts, project)
+	grip.Info(validationErrs)
```
malikchaya2, line 123: Did you mean to leave this here?

malikchaya2 (approved):
LGTM
I don't think I know enough to have an opinion on whether it's a good idea to use strict everywhere.

operations/validate.go
ablack12: Yes, this is intentionally to communicate any errors with the user.

operations/validate.go
```diff
@@ -129,26 +135,40 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
...
 }
+
+// loadProjectIntoWithValidation returns a warning (instead of an error) if there's an error with unmarshalling strictly
+func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model.GetProjectOpts,
+	project *model.Project) (*model.ParserProject, validator.ValidationErrors) {
+	errs := validator.ValidationErrors{}
+	pp, _, err := model.LoadProjectInto(ctx, data, opts, "", project)
+	if err != nil {
+		if opts.UnmarshalStrict {
+			// try it without unmarshal strict; if the error is from being strict, warn but don't error.
```
hadjri, line 157: Perhaps we can wrap the error returned if UnmarshalYAMLStrictWithFallback fails in createIntermediateProject with a string and index this error for that specific string? Instead of running LoadProjectInto twice and specifying this new err2

hadjri (changes requested):
I think keeping this as a warning rather than a breaking error makes sense, it seems like this solution is sufficient for allowing checking the yaml structure to work with local calls to LoadProjectInto.  Changing the codebase to use strict everywhere might introduce an unnecessary level of risk. LGTM with one comment

operations/validate.go
ablack12: Good idea -- I don't think we can do away with running LoadProjectInto twice entirely, because UnmarshalStrict can error for both strict and non-strict reasons; we don't know which it is without running the same thing _without_ unmarshal strict (i.e. determining if it should be a warning or an error). You're right though that we can limit this edge case to just the unmarshalling yaml part.

hadjri (approved):

PR #5345, opened 2022-01-28

bynn:

### Description 
aded hint to the query

### Testing 
query doesnt hang in shell 
image

### DIFF
--- a/makefile
+++ b/makefile
@@ -385,7 +385,7 @@ mongodb/.get-mongodb:
 get-mongodb:mongodb/.get-mongodb
 	@touch $<
 start-mongod:mongodb/.get-mongodb
-	./mongodb/mongod --dbpath ./mongodb/db_files --port 27017 --replSet evg --smallfiles --oplogSize 10
+	./mongodb/mongod --dbpath ./mongodb/db_files --port 27017 --replSet evg  --oplogSize 10
 	@echo "waiting for mongod to start up"
 start-mongod-auth:mongodb/.get-mongodb
 	./mongodb/mongod --auth --dbpath ./mongodb/db_files --port 27017 --replSet evg --oplogSize 10
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -21,7 +21,7 @@ const (
 )
 
 var (
-	Distro_1_status_1_activated_1_priority_1 = bson.D{
+	ActivatedTasksByDistroIndex = bson.D{
 		{Key: DistroIdKey, Value: 1},
 		{Key: StatusKey, Value: 1},
 		{Key: ActivatedKey, Value: 1},
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -21,6 +21,15 @@ const (
 )
 
 var (
+	Distro_1_status_1_activated_1_priority_1 = bson.D{
+		{Key: DistroIdKey, Value: 1},
+		{Key: StatusKey, Value: 1},
+		{Key: ActivatedKey, Value: 1},
+		{Key: PriorityKey, Value: 1},
+	}
+)
+
+var (
 	// BSON fields for the task struct
 	IdKey                       = bsonutil.MustHaveTag(Task{}, "Id")
 	SecretKey                   = bsonutil.MustHaveTag(Task{}, "Secret")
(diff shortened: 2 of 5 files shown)


model/task/task.go
```diff
@@ -1039,6 +1039,7 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 // operation will select tasks from all distros.
 func UnscheduleStaleUnderwaterTasks(distroID string) (int, error) {
 	query := scheduleableTasksQuery()
+	hint := "distro_1_status_1_activated_1_priority_1"
```
Kimchelly, line 1042: What index is it using instead of this one?

Is it worth commenting why the explicit hint is necessary instead of just letting the query planner pick?

bynn: its currently using `status_1_depends_on.status_1_depends_on.unattainable_1` if we dont specify 

if you think a comment might be useful i could add that in though!

Kimchelly: I think saying why we have to tell the query planner to not plan on our behalf makes sense. Comments are valuable when explaining why we have to do something unexpected, especially special cases like overriding the built-in query planner's decision, which we rarely have to do.

ablack12: Can you build this like this instead of hardcoding? https://github.com/evergreen-ci/evergreen/blob/8e010917fc3774e4f32d8c2843c6d23cbed623a5/model/host/host.go#L1696
I agree with commenting why this is needed.

model/task/task.go
```diff
@@ -1054,8 +1053,16 @@ func UnscheduleStaleUnderwaterTasks(distroID string) (int, error) {
 		},
 	}
 
-	q := db.Query(query).Hint(hint)
-	info, err := UpdateAll(q, update)
+	// force the query to use 'distro_1_status_1_activated_1_priority_1'
```
Kimchelly, line 1056: nit: capitalization of full sentences and end with a period.

Kimchelly (approved):
LGTM mod nit.

makefile
```diff
@@ -385,7 +385,7 @@ mongodb/.get-mongodb:
 get-mongodb:mongodb/.get-mongodb
 	@touch $<
 start-mongod:mongodb/.get-mongodb
-	./mongodb/mongod --dbpath ./mongodb/db_files --port 27017 --replSet evg --smallfiles --oplogSize 10
+	./mongodb/mongod --dbpath ./mongodb/db_files --port 27017 --replSet evg  --oplogSize 10
```
hhoke, line 388: https://docs.mongodb.com/v4.2/release-notes/4.2-compatibility/index.html

hhoke (approved):
LGTM

model/task/task.go
```diff
@@ -1053,7 +1053,16 @@ func UnscheduleStaleUnderwaterTasks(distroID string) (int, error) {
 		},
 	}
 
-	info, err := UpdateAll(query, update)
+	// Force the query to use 'distro_1_status_1_activated_1_priority_1'
+	// instead of defaulting to 'status_1_depends_on.status_1_depends_on.unattainable_1'.
+	hint := bson.D{
```
ablack12, line 1058: Instead of putting this within the function, can we define this as a constant? (Maybe in db.go) Then we can also use it in the test

ablack12 (changes requested):
Congrats on passing tests!!!

model/task/task.go
bynn: done

model/task/db.go
```diff
@@ -20,6 +20,15 @@ const (
 	OldCollection = "old_tasks"
 )
 
+var (
+	Distro_1_status_1_activated_1_priority_1 = bson.D{
```
ablack12, line 24: Can we also call this something more userfriendly?  In this case ActivatedTasksByDistroIndex or UnderwaterTasksIndex or etc

ablack12 (approved):
LGTM with nit

model/task/db.go
ablack12: Really "Index" is important here; as it is this isn't any more readable than hardcoding it.

PR #5348, opened 2022-01-31
EVG-15724 handle resetting unfinished display tasks from routes

ablack12:
[EVG-15724](https://jira.mongodb.org/browse/EVG-15724)

### Description 
Display tasks can't be restarted if they're currently in progress, but that isn't obvious to users who are using rest routes. Display tasks can be marked to be restarted when the whole thing is finished, so I think that's the right thing to do here.

### Testing 
Unit test.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1452,6 +1452,9 @@ func ClearAndResetStrandedTask(h *host.Host) error {
 	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, &t.Details), "problem resetting task")
 }
 
+// ResetTaskOrDisplayTask is a wrapper for TryResetTask that handles execution and display tasks that are restarted
+// from sources separate from marking the task finished. If an execution task, attempts to restart the display task instead.
+// Marks display tasks as reset when finished and then check if it can be reset immediately.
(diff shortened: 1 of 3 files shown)


model/task_lifecycle.go
```diff
@@ -1449,18 +1449,32 @@ func ClearAndResetStrandedTask(h *host.Host) error {
...
 
-	if t.IsPartOfDisplay() {
-		dt, err := t.GetDisplayTask()
+	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, &t.Details), "problem resetting task")
+}
+
+// ResetTaskOrDisplayTask is a wrapper for TryResetTask that handles execution and display tasks that are restarted
+// from sources separate from marking the task finished. If an execution task, attempts to restart the display task instead.
+// Marks display tasks as reset when finished and then check if it can be reset immediately.
+func ResetTaskOrDisplayTask(t *task.Task, user, origin string, detail *apimodels.TaskEndDetail) error {
```
ablack12, line 1458: ResetTaskOrDisplayTask I realize is not a good name, but I can't think of a better one without completely refactoring all of the logic which is out of scope here. Let me know if you think of a better one! I wanted to move the ClearAndResetStrandedTask logic to a shared function to help with testing and code deduplication.

malikchaya2: I'm not against the name, it makes it clear what the function is doing.

malikchaya2 (approved):

PR #5349, opened 2022-02-01
EVG-16168: add field for task execution platform

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16168

### Description 
Add a field to distinguish a task running in a container from a task running in a host. This field is currently set to host mode for any newly-created execution task, although it should eventually be set based on the YAML config.

### Testing 
Updated unit tests to verify that it's set when execution tasks are first created.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -117,6 +117,9 @@ type Task struct {
 	// The host the task was run on. This value is empty for display tasks
 	HostId string `bson:"host_id" json:"host_id"`
 
+	// ExecutionMode determines the execution environment that the task runs in.
+	ExecutionMode ExecutionMode `bson:"execution_mode,omitempty" json:"execution_mode,omitempty"`
+
 	// The version of the agent this task was run on.
 	AgentVersion string `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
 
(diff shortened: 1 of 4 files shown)


model/task/task.go
```diff
@@ -214,6 +217,16 @@ type Task struct {
...
 
+// ExecutionMode indicates the type of environment that the task runs in.
+type ExecutionMode string
+
+const (
+	// ExecutionModeHost indicates that the task runs in a host.
+	ExecutionModeHost ExecutionMode = "host"
+	// ExecutionModeContainer indicates that the task runs in a container.
+	ExecutionModeContainer ExecutionMode = "container"
+)
```
hhoke, line 228: Do we foresee adding more execution modes? I think "ExecutionMode" might be a little vague, given that the information in here is basically "does this task run on a container or a host". Maybe ExecutionPlatform?

Kimchelly: I don't foresee adding more today, but it could change in the future. I find that using booleans (e.g. `IsHostTask`) in data models instead of enums is pretty inflexible to change and I've seen numerous cases where a boolean was used to represent one specific feature and then later seems like a bad decision in hindsight when it morphs into many booleans that you have to check when a single enum would have achieved the same result with less pain. Since we can't do migrations against the task collection easily, future-proofing is the easiest way to avoid making the task model more complicated than it already is.

I don't have a good name for this, so I renamed it to ExecutionPlatform.

hhoke: Agreed that enums seems like the best approach. ExecutionPlatform sounds better to me.

model/task/task.go
```diff
@@ -117,6 +117,9 @@ type Task struct {
 	// The host the task was run on. This value is empty for display tasks
 	HostId string `bson:"host_id" json:"host_id"`
 
+	// ExecutionPlatform determines the execution environment that the task runs in.
+	ExecutionPlatform ExecutionPlatform `bson:"execution_mode,omitempty" json:"execution_mode,omitempty"`
```
hhoke, line 121: execution_platform

hadjri (approved):

PR #5355, opened 2022-02-01
EVG-15936: improve host termination for intent hosts

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15936

### Description 
I was originally going to make it so that hosts that are already marked "terminated" in the DB could be terminated again for intent hosts that are marked terminated before their real hosts come. However, rather than change assumptions around the terminated host state, it seems like it's easier to continue assuming that hosts that are terminated in the DB have also been terminated in the cloud. To maintain the termination assumption, I changed the behavior of a host that checks in after the intent host is already terminated to do the host doc swap and mark the real host as decommissioned, so it actually gets terminated in the cloud.

* Treat terminated intent hosts that are actually alive in the cloud as decommissioned real hosts.
* Log if a host is terminated in the DB but the agent is clearly still alive well after it should have been terminated.
* If a cloud instance does not exist when terminating the host, assume that the cloud instance is already terminated.

### Testing 
Updated existing tests.

### DIFF
--- a/cloud/cloud_status.go
+++ b/cloud/cloud_status.go
@@ -14,16 +14,17 @@ const (
 	// fulfilled, but it's not yet done booting up.
 	StatusInitializing
 
-	// StatusFailed indicates that an attempt to start the instance has failed;
-	// Could be due to billing, lack of capacity, etc.
+	// StatusFailed indicates that an attempt to start the instance has failed.
+	// This could be due to billing, lack of capacity, etc.
 	StatusFailed
 
 	// StatusRunning means the machine is done booting, and active.
 	StatusRunning
 
-	// StatusStopped indicates that the instance is in the processing of
+	// StatusStopping indicates that the instance is in the processing of
 	// stopping but has not yet stopped completely.
 	StatusStopping
+
 	// StatusStopped indicates that the instance is shut down, but can be
 	// started again.
 	StatusStopped
--- a/cloud/cloud_status.go
+++ b/cloud/cloud_status.go
@@ -3,27 +3,32 @@ package cloud
 type CloudStatus int
 
 const (
-	//Catch-all for unrecognized status codes
+	// StatusUnknown is a catch-all for unrecognized status codes.
 	StatusUnknown = CloudStatus(iota)
 
-	//StatusPending indicates that it is not yet clear if the instance
-	//has been successfully started or not (e.g., pending spot request)
+	// StatusPending indicates that it is not yet clear if the instance has been
+	// successfully started or not (e.g. pending spot request).
 	StatusPending
 
-	//StatusInitializing means the instance request has been successfully
-	//fulfilled, but it's not yet done booting up
+	// StatusInitializing means the instance request has been successfully
+	// fulfilled, but it's not yet done booting up.
 	StatusInitializing
 
-	//StatusFailed indicates that an attempt to start the instance has failed;
-	//Could be due to billing, lack of capacity, etc.
+	// StatusFailed indicates that an attempt to start the instance has failed;
+	// Could be due to billing, lack of capacity, etc.
 	StatusFailed
 
-	//StatusRunning means the machine is done booting, and active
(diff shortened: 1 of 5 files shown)


cloud/cloud_status.go
```diff
@@ -3,27 +3,32 @@ package cloud
...
-	//StatusInitializing means the instance request has been successfully
-	//fulfilled, but it's not yet done booting up
+	// StatusInitializing means the instance request has been successfully
+	// fulfilled, but it's not yet done booting up.
 	StatusInitializing
 
-	//StatusFailed indicates that an attempt to start the instance has failed;
-	//Could be due to billing, lack of capacity, etc.
+	// StatusFailed indicates that an attempt to start the instance has failed;
+	// Could be due to billing, lack of capacity, etc.
```
julianedwards, line 18: `could` or change L17 to have a period instead of a semicolon.

cloud/cloud_status.go
```diff
@@ -3,27 +3,32 @@ package cloud
...
 	StatusFailed
 
-	//StatusRunning means the machine is done booting, and active
+	// StatusRunning means the machine is done booting, and active.
 	StatusRunning
 
+	// StatusStopped indicates that the instance is in the processing of
+	// stopping but has not yet stopped completely.
 	StatusStopping
+	// StatusStopped indicates that the instance is shut down, but can be
```
julianedwards, line 27: Nit: add new line above.

cloud/cloud_status.go
```diff
@@ -3,27 +3,32 @@ package cloud
...
-	//Could be due to billing, lack of capacity, etc.
+	// StatusFailed indicates that an attempt to start the instance has failed;
+	// Could be due to billing, lack of capacity, etc.
 	StatusFailed
 
-	//StatusRunning means the machine is done booting, and active
+	// StatusRunning means the machine is done booting, and active.
 	StatusRunning
 
+	// StatusStopped indicates that the instance is in the processing of
```
julianedwards, line 24: `StatusStopping`

service/api_task.go
```diff
@@ -461,11 +423,19 @@ func (as *APIServer) prepareHostForAgentExit(ctx context.Context, h *host.Host)
...
 		// The host should already be terminated but somehow it's still alive in
-		// the cloud, so terminate it again.
-		if err := as.terminateGhost(ctx, h, "host should already have been terminated, but is still alive"); err != nil {
-			return false, errors.Wrap(err, "terminating ghost")
+		// the cloud. If the host was terminated very recently, it may simply
+		// need time for the cloud instance to actually be terminated and fully
+		// cleaned up. Beyond this short grace period, there's likely a state
+		// mismatch where Evergreen has marked the host terminated but the host
+		// has clearly not been terminated in the cloud.
+		const waitingForTerminationGracePeriod = 10 * time.Minute
```
julianedwards, line 431: Should this constant be defined here? Seems a bit awkward that it is not a package-level variable. If it is only used here, then maybe it doesn't even need to be a variable? Excellent documentation btw, so I think that should take care of any confusion someone might have if there is no variable and you just have:
```
if time.Now().Sub(h.TerminationTime) > 10 * time.Minute {
```

julianedwards (changes requested):

cloud/cloud_status.go
Kimchelly: Fixed.

cloud/cloud_status.go
Kimchelly: Fixed.

cloud/cloud_status.go
Kimchelly: Done.

service/api_task.go
Kimchelly: Yeah, I flip-flopped on using a constant or not because I don't love using random arbitrarily-chosen magic time durations in the codebase, but since there's a comment, it's probably not a big deal. I changed it to not use the constant.

julianedwards (approved):

PR #5357, opened 2022-02-02
EVG-15836 Fix s3_copy

malikchaya2:

### Description 
This was merged to main before because it broke the smoke test. After testing, we realized that the source of that break was the added lines to the smoke test. [EVG-16241](uhttps://jira.mongodb.org/browse/EVG-16241) was created to address that. This PR merges what was merged before (the first commit), plus a removal of the additional lines to agent.yml, and a removal of a test in self-test.yml that shouldn't have been committed. 

### Testing 
Tested with patches on staging. 
[smoke test patch](https://spruce-staging.corp.mongodb.com/version/61fab1cfb237367ee0cbcfb6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[a patch with a fake key that should fail](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_copy_should_succeed_patch_1e78e1f484cda40069403e82d1887ee670f82d1d_61faae5bb237367ee0cbce2b_22_02_02_16_16_40/logs?execution=0)
[a patch that should succeed](https://spruce-staging.corp.mongodb.com/version/61faaf329dbe32539cf15a38/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (I verified that the upload was done).

### DIFF
--- a/agent/command/s3_copy.go
+++ b/agent/command/s3_copy.go
@@ -115,7 +115,7 @@ func (c *s3copy) ParseParams(params map[string]interface{}) error {
 	return c.validate()
 }
 
-// validateParams is a helper function that ensures all
+// validate is a helper function that ensures all
 // the fields necessary for carrying out an S3 copy operation are present
 func (c *s3copy) validate() error {
 	catcher := grip.NewSimpleCatcher()
@@ -157,16 +157,15 @@ func (c *s3copy) validate() error {
 	return catcher.Resolve()
 }
 
-// Implementation of Execute.  Expands the parameters, and then copies the
+// Execute expands the parameters, and then copies the
 // resource from one s3 bucket to another one.
 func (c *s3copy) Execute(ctx context.Context,
 	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
 
-	// expand necessary params
 	if err := util.ExpandValues(c, conf.Expansions); err != nil {
 		return errors.WithStack(err)
 	}
-	// re-validate command here, in case an expansion is not defined
+	// Re-validate the command here, in case an expansion is not defined.
 	if err := c.validate(); err != nil {
 		return errors.WithStack(err)
 	}
@@ -241,7 +240,6 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
 		s3CopyReq.AwsKey = c.AwsKey
 		s3CopyReq.AwsSecret = c.AwsSecret
 
-		// Now copy the file into the permanent location
 		srcOpts := pail.S3Options{
 			Credentials: pail.CreateAWSCredentials(s3CopyReq.AwsKey, s3CopyReq.AwsSecret, ""),
 			Region:      s3CopyReq.S3SourceRegion,
@@ -267,9 +265,9 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
 		}
 		destBucket, err := pail.NewS3MultiPartBucket(destOpts)
 		if err != nil {
-			connectionErr := errors.Wrap(err, "S3 copy failed, could not establish connection to destination bucket")
-			logger.Task().Error(connectionErr)
-			return connectionErr
+			bucketErr := errors.Wrap(err, "S3 copy failed, could not establish connection to destination bucket")
+			logger.Task().Error(bucketErr)
+			return bucketErr
 		}
 	retryLoop:
 		for i := 0; i < maxS3OpAttempts; i++ {
@@ -289,12 +287,12 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
 						return errors.Wrap(err, "updating pushlog status failed for task")
 					}
 					if s3CopyFile.Optional {
-						logger.Execution().Errorf("S3 push copy failed to copy '%s' to '%s'. File is optional, continuing \n error: %v",
-							s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket, err)
+						logger.Execution().Errorf("S3 push copy failed to copy '%s' to '%s'. File is optional, continuing \n error: %s",
+							s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket, err.Error())
 						timer.Reset(backoffCounter.Duration())
 						continue retryLoop
 					} else {
-						return errors.Wrapf(err, "S3 push copy failed to copy '%s' to '%s'. File is not optional, exiting \n error",
+						return errors.Wrapf(err, "S3 push copy failed to copy '%s' to '%s'. File is not optional, exiting \n error:",
 							s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket)
 					}
 				} else {
--- a/agent/command/s3_copy.go
+++ b/agent/command/s3_copy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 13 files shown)


agent/command/s3_copy.go
```diff
@@ -78,102 +98,132 @@ type s3Loc struct {
...
+
 	}
 
-	return nil
+	return catcher.Resolve()
 }
 
-// Execute carries out the s3copy command - this is required
-// to satisfy the 'Command' interface
+// Implementation of Execute.  Expands the parameters, and then copies the
```
Kimchelly, line 160: Can this still say `Execute expands the parameters, ...`? Go documentation comments conventionally start with the name of the thing being documented (i.e. `Execute`).

agent/command/s3_copy.go
```diff
@@ -78,102 +98,132 @@ type s3Loc struct {
...
+// resource from one s3 bucket to another one.
 func (c *s3copy) Execute(ctx context.Context,
 	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
 
-	// expand the S3 copy parameters before running the task
+	// expand necessary params
```
Kimchelly, line 165: Full sentence.

Is this comment necessary? It seems like it's unnecessary since you could just see without the comment that it's expanding the expansion values.

agent/command/s3_copy.go
```diff
@@ -78,102 +98,132 @@ type s3Loc struct {
...
 	if err := util.ExpandValues(c, conf.Expansions); err != nil {
 		return errors.WithStack(err)
 	}
+	// re-validate command here, in case an expansion is not defined
```
Kimchelly, line 169: Full sentence.

agent/command/s3_copy.go
```diff
@@ -78,102 +98,132 @@ type s3Loc struct {
...
+	if err := c.validate(); err != nil {
+		return errors.WithStack(err)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
Kimchelly, line 174: Should this be ignoring the passed-in `ctx` to `Execute`?

agent/command/s3_copy.go
```diff
@@ -188,43 +238,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
-			S3Permissions:       s3CopyFile.Permissions,
-		}
-
-		responseString, err := comm.S3Copy(ctx, td, &s3CopyReq)
+		s3CopyReq.AwsKey = c.AwsKey
+		s3CopyReq.AwsSecret = c.AwsSecret
 
-		if responseString != "" {
-			logger.Task().Infof("s3Copy response: %s", responseString)
+		// Now copy the file into the permanent location
```
Kimchelly, line 244: Same question about comments - is this one necessary?

agent/command/s3_copy.go
```diff
@@ -188,43 +238,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+		destOpts := pail.S3Options{
+			Credentials: pail.CreateAWSCredentials(s3CopyReq.AwsKey, s3CopyReq.AwsSecret, ""),
+			Region:      s3CopyReq.S3DestinationRegion,
+			Name:        s3CopyReq.S3DestinationBucket,
+			Permissions: pail.S3Permissions(s3CopyReq.S3Permissions),
+		}
+		destBucket, err := pail.NewS3MultiPartBucket(destOpts)
 		if err != nil {
-			return errors.WithStack(err)
+			connectionErr := errors.Wrap(err, "S3 copy failed, could not establish connection to destination bucket")
```
Kimchelly, line 270: This is not always going to be a connection error, is it? It seems like it could fail for other reasons like providing invalid credentials options.

service/api.go
```diff
@@ -705,6 +785,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.Route().Version(2).Route("/task/{taskId}/parser_project").Wrap(requireTaskSecret).Handler(as.GetParserProject).Get()
 	app.Route().Version(2).Route("/task/{taskId}/project_ref").Wrap(requireTaskSecret).Handler(as.GetProjectRef).Get()
 	app.Route().Version(2).Route("/task/{taskId}/expansions").Wrap(requireTask, requireHost).Handler(as.GetExpansions).Get()
+	app.Route().Version(2).Route("/task/{taskId}/new_push").Wrap(requireTaskSecret).Handler(as.NewPush).Post()
```
Kimchelly, line 788: I guess you don't have to rewrite it, but for future reference, the note above this block says new routes should not be added to the legacy APIServer and should go into the REST v2 routes instead.

service/api_plugin_s3copy.go
```diff
@@ -134,7 +134,8 @@ func (as *APIServer) s3copyPlugin(w http.ResponseWriter, r *http.Request) {
 		})
 
 	if err != nil {
-		grip.Error(errors.Wrap(errors.WithStack(newPushLog.UpdateStatus(model.PushLogFailed)), "updating pushlog status failed"))
+		// this is not using a constant because the code will be removed soon
```
Kimchelly, line 137: Can you attach a `// TODO (<ticket_num>)` for which ticket is reponsible for removing it?

agent/command/s3_copy.go
```diff
@@ -188,43 +238,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+					if err := comm.UpdatePushStatus(ctx, td, newPushLog); err != nil {
+						return errors.Wrap(err, "updating pushlog status failed for task")
+					}
+					if s3CopyFile.Optional {
+						logger.Execution().Errorf("S3 push copy failed to copy '%s' to '%s'. File is optional, continuing \n error: %v",
+							s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket, err)
+						timer.Reset(backoffCounter.Duration())
+						continue retryLoop
+					} else {
+						return errors.Wrapf(err, "S3 push copy failed to copy '%s' to '%s'. File is not optional, exiting \n error",
```
Kimchelly, line 297: Is this supposed to say `\n error` at the end without providing an error? It seems like the error message should just say `S3 copy failed to copy '%s' to '%s'. File is not optional, exiting."`.

agent/command/s3_copy.go
```diff
@@ -188,43 +238,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+					DestinationBucket: destBucket,
+				}
+				err = srcBucket.Copy(ctx, copyOpts)
+				if err != nil {
+					newPushLog.Status = pushLogFailed
+					if err := comm.UpdatePushStatus(ctx, td, newPushLog); err != nil {
+						return errors.Wrap(err, "updating pushlog status failed for task")
+					}
+					if s3CopyFile.Optional {
+						logger.Execution().Errorf("S3 push copy failed to copy '%s' to '%s'. File is optional, continuing \n error: %v",
```
Kimchelly, line 292: Should this verbosely print the error with `%v` or just the error string `%s`? Seems like a user wouldn't want to see things like the stack trace you woudl get if you use `%v`.

agent/command/s3_copy.go
```diff
@@ -188,43 +238,76 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
+					DestinationKey:    s3CopyReq.S3DestinationPath,
+					DestinationBucket: destBucket,
+				}
+				err = srcBucket.Copy(ctx, copyOpts)
+				if err != nil {
+					newPushLog.Status = pushLogFailed
+					if err := comm.UpdatePushStatus(ctx, td, newPushLog); err != nil {
+						return errors.Wrap(err, "updating pushlog status failed for task")
+					}
+					if s3CopyFile.Optional {
```
Kimchelly, line 291: Just as a general thing for the future, I would avoid doing 5 nested blocks of logic (for loop, for looop, select block, if block, if block) since it's difficult to reason about. Making helper functions to break down the logic into more digestible units would be nice.

Kimchelly (changes requested):
I know most of the comments are on stuff from the previous PR (which I didn't look at), but it seems worth addressing now.

agent/command/s3_copy.go
```diff
@@ -78,102 +98,132 @@ type s3Loc struct {
...
+
+	if err := decoder.Decode(params); err != nil {
+		return errors.Wrapf(err, "error decoding %s params", c.Name())
 	}
-	return nil
+
+	return c.validate()
 }
 
 // validateParams is a helper function that ensures all
```
bynn, line 118: change function name in comment

malikchaya2: thank you!

agent/command/s3_copy.go
malikchaya2: true, I'll remove. thanks!

agent/command/s3_copy.go
malikchaya2: It was modeled off of [this](https://github.com/evergreen-ci/evergreen/blob/main/agent/command/s3_get.go#L146), but I'm not sure why it's done that way...

agent/command/s3_copy.go
malikchaya2: nah, I'll remove it

agent/command/s3_copy.go
malikchaya2: good point, fixed

agent/command/s3_copy.go
malikchaya2: thanks, fixed

agent/command/s3_copy.go
malikchaya2: yes, because it wraps the error, so this puts the error on the next line, and appends "error: " to the front of it.

service/api_plugin_s3copy.go
malikchaya2: done. This whole file should be removed.

agent/command/s3_copy.go
```diff
@@ -188,43 +237,75 @@ func (c *s3copy) s3Copy(ctx context.Context,
...
-			if s3CopyFile.Optional {
-				logger.Execution().Errorf("file '%s' is optional, continuing",
-					s3CopyFile.DisplayName)
-				continue
-			} else {
-				return errors.Wrapf(err, "failed to push %s to %s",
-					s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket)
-			}
-
+			connectionErr := errors.Wrap(err, "S3 copy failed, could not establish connection to source bucket")
```
Kimchelly, line 252: Can you also rename this one to be `bucketErr`?

Kimchelly (approved):
LGTM mod one tiny thing.

agent/command/s3_copy.go
Kimchelly: Hmm, if it's already like that, I guess it's fine to leave it. My only concern was that doing it this way might end up ignoring command timeouts if the user has specified one.

PR #5360, opened 2022-02-03
EVG-16175: (partial) addition of new task status and concomitant changes for trigger, scheduler, and graphql

hhoke:
[EVG-16175](https://jira.mongodb.org/browse/EVG-16175)

### Description 
I added a TaskContainerAllocated status to differentiate between container tasks that were ready to run but waiting for a container, and that are ready to run with the container allocated. This required several downstream changes, which will be made in parts. 

### Testing 
https://spruce.mongodb.com/version/61fbe1f02fbabe23af018894/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -57,13 +57,16 @@ const (
 	TaskUnscheduled  = "unscheduled"
 	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
 	TaskWillRun = "will-run"
-
-	// TaskStarted indicates a task is running on an agent
-	TaskStarted = "started"
+	// TaskContainerAllocated indicates that a tasks's container has been requested
+	// from the cloud provider but hasn't started yet
+	TaskContainerAllocated = "container-allocated"
 
 	// TaskDispatched indicates that an agent has received the task, but
 	// the agent has not yet told Evergreen that it's running the task
 	TaskDispatched = "dispatched"
+
+	// TaskStarted indicates a task is running on an agent
+	TaskStarted = "started"
 
 	// The task statuses below indicate that a task has finished.
 	TaskSucceeded = "success"
--- a/globals.go
+++ b/globals.go
@@ -291,7 +291,7 @@ var TaskNonGenericFailureStatuses = []string{
 
 // TaskFailureStatuses represents all the ways that a task can fail, inclusive of system failures
 // and task failures.
-var TaskFailureStatuses = append(TaskNonGenericFailureStatuses, TaskFailed)
+var TaskFailureStatuses = append([]string{TaskFailed}, TaskNonGenericFailureStatuses...)
 
 var TaskUnstartedStatuses = []string{
 	TaskInactive,
--- a/globals.go
+++ b/globals.go
@@ -58,7 +58,7 @@ const (
 	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
 	TaskWillRun = "will-run"
 	// TaskContainerAllocated indicates that a tasks's container has been requested
-	// from the cloud provider but hasn't started yet
+	// from the cloud provider but hasn't started yet.
 	TaskContainerAllocated = "container-allocated"
 
 	// TaskDispatched indicates that an agent has received the task, but
--- a/globals.go
+++ b/globals.go
@@ -293,6 +293,7 @@ var TaskUnstartedStatuses []string = []string{
 	TaskInactive,
 	TaskUnstarted,
 	TaskUndispatched,
+	TaskContainerAllocated,
 }
 
 func IsUnstartedTaskStatus(status string) bool {
@@ -677,9 +678,9 @@ var (
 	}
 
 	// constant arrays for db update logic
-	AbortableStatuses   = []string{TaskStarted, TaskDispatched}
+	AbortableStatuses   = []string{TaskStarted, TaskDispatched, TaskContainerAllocated}
 	CompletedStatuses   = []string{TaskSucceeded, TaskFailed}
-	UncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict}
+	UncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict, TaskContainerAllocated}
 
 	SyncStatuses = []string{TaskSucceeded, TaskFailed}
 
--- a/globals.go
+++ b/globals.go
@@ -65,7 +65,7 @@ const (
 	// the agent has not yet told Evergreen that it's running the task
 	TaskDispatched = "dispatched"
 
-	// TaskStarted indicates a task is running on an agent
+	// TaskStarted indicates a task is running on an agent.
 	TaskStarted = "started"
 
 	// The task statuses below indicate that a task has finished.
@@ -272,14 +272,15 @@ const (
 	InvalidDivideInputError    = "$divide only supports numeric types"
 )
 
-var InternalAliases []string = []string{
+var InternalAliases = []string{
 	CommitQueueAlias,
 	GithubPRAlias,
 	GithubChecksAlias,
 	GitTagAlias,
 }
 
-var TaskSystemFailureStatuses []string = []string{
+// TaskNonGenericFailureStatuses represents some kind of specific abnormal failure mode.
+var TaskNonGenericFailureStatuses = []string{
 	TaskTimedOut,
 	TaskSystemFailed,
 	TaskTestTimedOut,
@@ -288,9 +289,11 @@ var TaskSystemFailureStatuses []string = []string{
 	TaskSystemTimedOut,
 }
 
-var TaskFailureStatuses []string = append([]string{TaskFailed}, TaskSystemFailureStatuses...)
+// TaskFailureStatuses represents all the ways that a task can fail, inclusive of system failures
+// and task failures.
+var TaskFailureStatuses = append(TaskNonGenericFailureStatuses, TaskFailed)
 
-var TaskUnstartedStatuses []string = []string{
+var TaskUnstartedStatuses = []string{
 	TaskInactive,
 	TaskUnstarted,
 	TaskUndispatched,
@@ -679,9 +682,13 @@ var (
 	}
 
 	// constant arrays for db update logic
-	AbortableStatuses   = []string{TaskStarted, TaskDispatched, TaskContainerAllocated}
-	CompletedStatuses   = []string{TaskSucceeded, TaskFailed}
-	UncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict, TaskInactive, TaskContainerAllocated}
+	// TaskAbortableStatuses have been picked up by an agent but may or may not have started.
+	TaskAbortableStatuses = []string{TaskStarted, TaskDispatched}
+	// TaskCompletedStatuses have not experienced some sort of system failure and
+	// are in a finished state.
+	TaskCompletedStatuses = []string{TaskSucceeded, TaskFailed}
+	// TaskUncompletedStatuses are all statuses that do not represent a finished state.
+	TaskUncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict, TaskInactive, TaskContainerAllocated}
 
 	SyncStatuses = []string{TaskSucceeded, TaskFailed}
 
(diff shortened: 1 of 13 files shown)


globals.go
```diff
@@ -276,20 +279,22 @@ var InternalAliases []string = []string{
 	GitTagAlias,
 }
 
-var TaskFailureStatuses []string = []string{
+var TaskSystemFailureStatuses []string = []string{
```
Kimchelly, line 282: This doesn't need the `[]string` type annotation because it's inferred to be a string slice from its initialized value. Can you remove it (and for the other `var VARNAME []string` usages in the file)?

globals.go
```diff
@@ -674,9 +679,9 @@ var (
 	}
 
 	// constant arrays for db update logic
-	AbortableStatuses   = []string{TaskStarted, TaskDispatched}
+	AbortableStatuses   = []string{TaskStarted, TaskDispatched, TaskContainerAllocated}
```
Kimchelly, line 682: My understanding of aborting a task is that a task isn't aborted until it's actually dispatched to an agent. I assume that we should keep the existing behavior of being able to unschedule until the task is dispatched, and only then allowing the task to be aborted.

globals.go
Kimchelly: I would avoid calling this `TaskSystemFailureStatuses` because there's already `TaskSystemFailed`, which generally means a purple task, whereas `TaskSetupFailed` is lavendar and understood to be distinct from a system failure.

Maybe something like `TaskNonGenericFailureStatuses`?

trigger/task.go
```diff
@@ -761,7 +761,7 @@ func (t *taskTriggers) taskRuntimeChange(sub *event.Subscription) (*notification
 }
 
 func isFailedTaskStatus(status string) bool {
-	return status == evergreen.TaskFailed || status == evergreen.TaskSystemFailed || status == evergreen.TaskTestTimedOut
+	return utility.StringSliceContains(evergreen.TaskFailureStatuses, status)
```
Kimchelly, line 764: This is the same as `evergreen.IsFailedTaskStatus`, so this function could just be removed entirely.

Kimchelly: Is there any reason why this check originally omitted other failure statuses like system unresponsive and setup failed?

globals.go
Kimchelly, line 682: Can be rename these task status variables to be prefixed with `Task` (e.g. `TaskAbortableStatuses`) to remain consistent with the other task status global variables?

globals.go
```diff
@@ -57,14 +57,17 @@ const (
...
-	TaskStarted = "started"
+	// TaskContainerAllocated indicates that a tasks's container has been requested
+	// from the cloud provider but hasn't started yet
+	TaskContainerAllocated = "container-allocated"
 
 	// TaskDispatched indicates that an agent has received the task, but
 	// the agent has not yet told Evergreen that it's running the task
 	TaskDispatched = "dispatched"
 
+	// TaskStarted indicates a task is running on an agent
```
Kimchelly, line 68: Period at the end of this sentence.

globals.go
```diff
@@ -57,14 +57,17 @@ const (
 	TaskUnscheduled  = "unscheduled"
 	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
 	TaskWillRun = "will-run"
-
-	// TaskStarted indicates a task is running on an agent
-	TaskStarted = "started"
+	// TaskContainerAllocated indicates that a tasks's container has been requested
+	// from the cloud provider but hasn't started yet
```
Kimchelly, line 61: Period at the end of this sentence.

globals.go
Kimchelly: Can you add a doc comment?

globals.go
```diff
@@ -276,20 +279,22 @@ var InternalAliases []string = []string{
...
 	TaskTimedOut,
-	TaskFailed,
 	TaskSystemFailed,
 	TaskTestTimedOut,
 	TaskSetupFailed,
 	TaskSystemUnresponse,
 	TaskSystemTimedOut,
 }
 
+var TaskFailureStatuses []string = append([]string{TaskFailed}, TaskSystemFailureStatuses...)
```
Kimchelly, line 291: Documentation.

globals.go
Kimchelly, line 682: Documentation.

globals.go
```diff
@@ -674,9 +679,9 @@ var (
 	}
 
 	// constant arrays for db update logic
-	AbortableStatuses   = []string{TaskStarted, TaskDispatched}
+	AbortableStatuses   = []string{TaskStarted, TaskDispatched, TaskContainerAllocated}
 	CompletedStatuses   = []string{TaskSucceeded, TaskFailed}
-	UncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict}
+	UncompletedStatuses = []string{TaskStarted, TaskUnstarted, TaskUndispatched, TaskDispatched, TaskConflict, TaskInactive, TaskContainerAllocated}
```
Kimchelly, line 684: Documentation.

Kimchelly (changes requested):
Nice job! Thank you for documenting things as well! \o/

Are you going to also add a status name (e.g. "container unallocated") to replace "undispatched" in a later PR?

I'm going to request app's review because my understanding of task failure statuses isn't that great.

globals.go
```diff
@@ -57,14 +57,17 @@ const (
 	TaskUnscheduled  = "unscheduled"
 	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
 	TaskWillRun = "will-run"
-
-	// TaskStarted indicates a task is running on an agent
-	TaskStarted = "started"
+	// TaskContainerAllocated indicates that a tasks's container has been requested
```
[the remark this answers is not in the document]
Kimchelly: This will eventually be part of the UI, but the UI for containers/containerized tasks won't exist until [PM-2620](https://jira.mongodb.org/browse/PM-2620). For now, as long as this change is backward compatible with the existing host task logic, it should be fine if it's not accounted for in the UI.

trigger/task.go
hhoke: There is no reason that I know of.

ablack12: My guess is that the statuses that weren't here (for example TaskSetupFailed,TaskSystemUnresponse,TaskSystemTimedOut) are outdated (I know system unresponse isn't used anymore since we use details but I'd have to do some digging to determine if the others are also outdated), but it also might be that we purposefully choose not to trigger on these things to avoid noise. If it doesn't seem like those statuses are used much then it should be fine to make this change.

ablack12 (changes requested):

graphql/resolvers.go
```diff
@@ -3278,7 +3278,7 @@ func (r *taskResolver) CanSchedule(ctx context.Context, obj *restModel.APITask)
 }
 
 func (r *taskResolver) CanUnschedule(ctx context.Context, obj *restModel.APITask) (bool, error) {
-	return obj.Activated && *obj.Status == evergreen.TaskUndispatched, nil
+	return obj.Activated && *obj.Status == evergreen.TaskUndispatched || *obj.Status == evergreen.TaskContainerAllocated, nil
```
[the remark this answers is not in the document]
hhoke: I think this should have different clustering. I think as it's now written the global doesn't make sense.

trigger/task.go
hhoke: I defer to Annie. These statuses are too common for me to risk it. I have reverted my change.

globals.go
Kimchelly: I would avoid doing appending in that order because [append can mutate the array underlying the slice](https://go.dev/ref/spec#Appending_and_copying_slices).

> If the capacity of s is not large enough to fit the additional values, append allocates a new, sufficiently large underlying array that fits both the existing slice elements and the additional values. Otherwise, append re-uses the underlying array.

ablack12: I feel like that isn't super relevant here since I assume we know the underlying array size of TaskSystemFailureStatuses because it's a constant; I am mostly biased because this looks ugly and someone will likely change it in the future anyway (so if we really don't want to change it we should add a comment to explain, although I think it would be fine to change)

Kimchelly: The issue is mostly if you ever try to append to `TaskSystemFailureStatuses` in another place, it may mutate the underlying array and produce unexpected results for those who are appending to it.
```go
import "fmt"

func main() {
	s0 := make([]string, 0, 3)
	s0 = append(s0, "one", "two")
	s1 := append(s0, "three")
	s2 := append(s0, "four")
	fmt.Println(s0) // [one two]
	fmt.Println(s1) // [one two four]
	fmt.Println(s2) // [one two four]
}

```

globals.go
Kimchelly: This still needs a period.

globals.go
```diff
@@ -269,27 +272,32 @@ const (
...
 	TaskTestTimedOut,
 	TaskSetupFailed,
 	TaskSystemUnresponse,
 	TaskSystemTimedOut,
 }
 
-var TaskUnstartedStatuses []string = []string{
+// TaskFailureStatuses represents all the ways that a task can fail, inclusive of system failures
+// and task failures.
+var TaskFailureStatuses = append(TaskNonGenericFailureStatuses, TaskFailed)
```
Kimchelly, line 294: I would avoid doing the append this way for the reason I mentioned in the previous comment.

graphql/util.go
```diff
@@ -770,7 +770,7 @@ func canRestartTask(ctx context.Context, at *restModel.APITask) (*bool, error) {
 	if err != nil {
 		return nil, err
 	}
-	nonrestartableStatuses := []string{evergreen.TaskStarted, evergreen.TaskUnstarted, evergreen.TaskUndispatched, evergreen.TaskDispatched, evergreen.TaskInactive}
+	nonrestartableStatuses := evergreen.TaskUncompletedStatuses
```
Kimchelly, line 773: It seems like this doesn't need a separate variable if it's only used on the next line.

trigger/task.go
```diff
@@ -469,7 +469,7 @@ func (t *taskTriggers) taskFailure(sub *event.Subscription) (*notification.Notif
 		return nil, nil
 	}
 
-	if !isFailedTaskStatus(t.data.Status) {
+	if !evergreen.IsFailedTaskStatus(t.data.Status) {
```
Kimchelly, line 472: Since `isFailedTaskStatus()` is unused, I would delete it.

Kimchelly (changes requested):

globals.go
ablack12: 😭  that's devastating. Thanks for the example.

graphql/util.go
ablack12: +1

trigger/build.go
```diff
@@ -30,21 +30,17 @@ func (t *buildTriggers) taskStatusToDesc() string {
...
 			success++
 
-		case evergreen.TaskFailed:
+		case t.Status == evergreen.TaskFailed:
 			failed++
 
-		case evergreen.TaskSystemFailed, evergreen.TaskTimedOut,
-			evergreen.TaskSystemUnresponse, evergreen.TaskSystemTimedOut,
-			evergreen.TaskTestTimedOut:
+		case utility.StringSliceContains(evergreen.TaskNonGenericFailureStatuses, t.Status):
```
ablack12, line 40: This case is specifically system failure statuses; TaskNonGenericFailureStatuses includes TaskSetupFailure

trigger/task.go
ablack12: I want to call out Using IsFailedTaskStatus here instead of isFailedTaskStatus is effectively the same as what we discussed here https://github.com/evergreen-ci/evergreen/pull/5360#discussion_r798748663 (i.e. changing how the private function worked) but I do think this change may be fine; I would just check how often the new statuses are used (or if that was more legacy)

trigger/build.go
```diff
@@ -70,6 +70,14 @@ func (t *buildTriggers) taskStatusToDesc() string {
 	return appendTime(t.build, desc)
 }
 
+func statusIsSystemFailure(status string) bool {
+	systemFailures := []string{evergreen.TaskSystemFailed, evergreen.TaskTimedOut,
```
Kimchelly, line 74: nit: can you put one status per line?

Kimchelly (approved):
LGTM mod nit.

ablack12 (approved):

PR #5361, opened 2022-02-03
EVG-7310 remove projectId from validation errors

ablack12:
[EVG-7310](https://jira.mongodb.org/browse/EVG-7310)

### Description 
Kind of snuck this into the epic, but right now we print projectId for most validation; this isn't super useful because in many cases there isn't a project Identifier (since people use validation from the CLI), so we just print `project ''` which is pretty useless (also in non-CLI cases we typically error with the project id anyway). 

I did add printing the project ID for the settings that rely on it (i.e. task sync).


hadjri (approved):
LGTM with unit test fixes (I assume you'll just need to change the expected error string outputs for some)

PR #5362, opened 2022-02-03
EVG-16170: only consider host tasks for the distro scheduler

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16170

### Description
* Check for host execution platform in distro scheduler queries. I'm going to build the relevant indexes in staging/prod after this is approved but before it's merged.
* Rename some task find functions to clarify that they're only for host tasks.
* Add/fix some documentation in the scheduler package.

### Testing 
Updated existing unit tests.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -107,10 +107,10 @@ type Task struct {
 	OverrideDependencies     bool         `bson:"override_dependencies,omitempty" json:"override_dependencies,omitempty"`
 
 	// DistroAliases refer to the optional secondary distros that can be
-	// associated with a task. This is used for running tasks in case there
-	// is excess capcacity in a distro. Despite the variable name, this is a
-	// distinct concept from actual distro aliases (i.e. alternative distro
-	// names).
+	// associated with a task. This is used for running tasks in case there are
+	// idle hosts in a distro with an empty primary queue. Despite the variable
+	// name, this is a distinct concept from actual distro aliases (i.e.
+	// alternative distro names).
 	// TODO (EVG-15148): rename this to represent secondary distros.
 	DistroAliases []string `bson:"distro_aliases,omitempty" json:"distro_aliases,omitempty"`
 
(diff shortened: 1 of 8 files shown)


scheduler/task_finder_test.go
```diff
@@ -261,10 +271,10 @@ func (s *TaskFinderComparisonSuite) SetupTest() {
 	defer cancel()
 	_, err := env.DB().Collection(task.Collection).Indexes().CreateMany(ctx, []mongo.IndexModel{
 		{
-			Keys: bson.D{{Key: "activated", Value: 1}, {Key: "status", Value: 1}, {Key: "priority", Value: 1}},
+			Keys: bson.D{{Key: task.ActivatedKey, Value: 1}, {Key: task.StatusKey, Value: 1}, {Key: task.PriorityKey, Value: 1}},
```
ablack12, line 274: 🚀

model/task/task.go
```diff
@@ -106,6 +106,12 @@ type Task struct {
 	NumDependents            int          `bson:"num_dependents,omitempty" json:"num_dependents,omitempty"`
 	OverrideDependencies     bool         `bson:"override_dependencies,omitempty" json:"override_dependencies,omitempty"`
 
+	// DistroAliases refer to the optional secondary distros that can be
+	// associated with a task. This is used for running tasks in case there
+	// is excess capcacity in a distro. Despite the variable name, this is a
```
hhoke, line 111: "excess load". Generally, "capacity" should be used to describe ability to absorb load, in our case the room we have for tasks, e.g. lots of hosts is "high capacity", too few is "low capacity".

model/task/task.go
```diff
@@ -1046,13 +1052,13 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 	return nil
 }
 
-// Removes tasks older than the unscheduable threshold (e.g. one
-// weeks) from the scheduler queue.
+// Removes host tasks older than the unscheduable threshold (e.g. one weeks)
```
hhoke, line 1055: "host tasks (in contrast with container tasks)"

hhoke (changes requested):
Clarifications on comments

model/task/task.go
Kimchelly: I don't think this needs clarification to contrast with container tasks, because making the host/container task parentheses comment in every single place where the code is doing something host-specific seems excessive. People should generally understand what a task running on a host is.

model/task/task.go
hhoke: Thanks for the change. I misunderstood the functionality, but it is clear now.

hhoke (approved):

PR #5363, opened 2022-02-03
EVG-15293: Restrict project settings GraphQL operations

sophstad:
[EVG-15293](https://jira.mongodb.org/browse/EVG-15293)

### Description 
- Add directive to verify user is admin when using `ProjectSettings`/`RepoSettings` queries, or `SaveProjectSettingsForSection`/`SaveRepoSettingsForSection` mutations
- Update tests to pass, and update test data to use more semantic/traceable values
- Update local data so that `admin` user can view settings for evergreen project, spruce project, and repo

### Testing 
- Add directive-specific unit tests
- Make existing GraphQL unit tests pass


graphql/tests/repoSettings/results.json
```diff
@@ -251,16 +251,17 @@
         "data": null,
         "errors": [
           {
-            "message": "repo doesn't exist",
+            "message": "user testuser does not have permission to access settings for the project evergreen",
```
sophstad, line 254: This error message is a bit less correct than before because of how the directive works. I didn't think it was worth performing an additional DB lookup in the projectRef and repoRef collections to present a correct error message, but interested in your thoughts.

graphql/resolvers.go
```diff
@@ -4081,5 +4081,23 @@ func New(apiURL string) Config {
 		}
 		return nil, Forbidden.Send(ctx, fmt.Sprintf("user %s does not have permission to access this resolver", user.Username()))
 	}
+	c.Directives.RequireProjectAdmin = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
+		args, isStringMap := obj.(map[string]interface{})
+
+		if !isStringMap {
+			return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Project not specified"))
```
ablack12, line 4088: Don't need fmt.Sprintf here since no string formatting is being done

graphql/resolvers.go
```diff
@@ -4081,5 +4081,23 @@ func New(apiURL string) Config {
...
+		if identifier, hasIdentifier := args["identifier"].(string); hasIdentifier {
+			pid, err := model.GetIdForProject(identifier)
+			if err != nil {
+				return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Could not find project with identifier: %s", identifier))
+			}
+			return HasProjectAdminPermission(ctx, pid, next)
+		} else if id, hasId := args["id"].(string); hasId {
+			return HasProjectAdminPermission(ctx, id, next)
+		}
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Could not find project"))
```
ablack12, line 4100: same formatting comment

graphql/schema.graphql
```diff
@@ -75,8 +76,8 @@ type Query {
   mainlineCommits(options: MainlineCommitsOptions!, buildVariantOptions: BuildVariantOptions): MainlineCommits
   taskNamesForBuildVariant(projectId: String!, buildVariant: String!): [String!]
   buildVariantsForTaskName(projectId: String!, taskName: String!): [BuildVariantTuple]
-  projectSettings(identifier: String!): ProjectSettings!
-  repoSettings(id: String!): RepoSettings!
+  projectSettings(identifier: String! @requireProjectAdmin): ProjectSettings!
```
ablack12, line 79: Anyone with view permission should be able to _see_ settings, edit access is only required really for the mutations.

graphql/schema.graphql
```diff
@@ -430,7 +431,7 @@ input ProjectSettingsInput {
 }
 
 input ProjectInput {
-  id: String!
+  id: String! @requireProjectAdmin
```
ablack12, line 434: I assume this is what affects the SaveProject mutation? So this would separately need to be added to the RepoRefInput, attachProjectToNewRepo, attachProjectToRepo, detachProjectFromRepo, forceRepotrackerRun inputs I think

ablack12 (changes requested):
Hype about this!!

graphql/schema.graphql
sophstad: Oh shoot! I did not know this, I think there will be some front end implications as well. Would you like the "view only" access to be handled in this ticket or separately?

ablack12: I think it could make sense to do the directive as part of this ticket, and then any front-end facing implications as part of a different ticket, but whatever you prefer 💃 

For the legacy UI I think people with view only see the page normally but can't click save; I would also prefer that for the front end since the "disabled" view is such a drag (I have a memory of approving this with Lara back in the day also but as I am aged my memory may be faulty)

graphql/resolvers.go
```diff
@@ -4081,5 +4081,23 @@ func New(apiURL string) Config {
...
 		return nil, Forbidden.Send(ctx, fmt.Sprintf("user %s does not have permission to access this resolver", user.Username()))
 	}
+	c.Directives.RequireProjectAdmin = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
+		args, isStringMap := obj.(map[string]interface{})
+
+		if !isStringMap {
+			return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Project not specified"))
+		}
+
+		if identifier, hasIdentifier := args["identifier"].(string); hasIdentifier {
```
malikchaya2, line 4091: aaah, so THAT'S how you do it! Thank you for figuring this out!

malikchaya2 (approved):
This is fantastic stuff!

graphql/schema.graphql
sophstad: Lol:

> as I am aged my memory may be faulty

That sounds good to me! Fwiw disabling all fields on the page would be fairly easy but I understand that that is not the most appealing view. And I'll add the view-only directive to this PR, thanks!

graphql/resolvers.go
```diff
@@ -4081,23 +4081,33 @@ func New(apiURL string) Config {
...
-	c.Directives.RequireProjectAdmin = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
-		args, isStringMap := obj.(map[string]interface{})
+	c.Directives.RequireProjectAccess = func(ctx context.Context, obj interface{}, next graphql.Resolver, access ProjectSettingsAccess) (res interface{}, err error) {
+		var permissionLevel evergreen.PermissionLevel
+		if access == ProjectSettingsAccessEdit {
+			permissionLevel = evergreen.ProjectSettingsEdit
+		} else if access == ProjectSettingsAccessView {
+			permissionLevel = evergreen.ProjectSettingsView
+		} else {
+			permissionLevel = evergreen.ProjectSettingsNone
```
ablack12, line 4091: Should we just return in this case?

graphql/util.go
```diff
@@ -1391,7 +1391,7 @@ func GroupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
 	return groupsArr, nil
 }
 
-func HasProjectAdminPermission(ctx context.Context, resource string, next graphql.Resolver) (res interface{}, err error) {
+func HasProjectPermission(ctx context.Context, resource string, next graphql.Resolver, permissionLevel evergreen.PermissionLevel) (res interface{}, err error) {
```
ablack12, line 1394: Can this be a private function?

graphql/resolvers.go
```diff
@@ -4081,23 +4081,33 @@ func New(apiURL string) Config {
 		}
 		return nil, Forbidden.Send(ctx, fmt.Sprintf("user %s does not have permission to access this resolver", user.Username()))
 	}
-	c.Directives.RequireProjectAdmin = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
-		args, isStringMap := obj.(map[string]interface{})
+	c.Directives.RequireProjectAccess = func(ctx context.Context, obj interface{}, next graphql.Resolver, access ProjectSettingsAccess) (res interface{}, err error) {
+		var permissionLevel evergreen.PermissionLevel
+		if access == ProjectSettingsAccessEdit {
+			permissionLevel = evergreen.ProjectSettingsEdit
```
ablack12, line 4087: Can we instead just store and pass in value since we don't actually need the description?

ablack12 (changes requested):

ablack12 (approved):

graphql/resolvers.go
```diff
@@ -4081,5 +4081,33 @@ func New(apiURL string) Config {
...
+
+		if identifier, hasIdentifier := args["identifier"].(string); hasIdentifier {
+			pid, err := model.GetIdForProject(identifier)
+			if err != nil {
+				return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Could not find project with identifier: %s", identifier))
+			}
+			return hasProjectPermission(ctx, pid, next, permissionLevel)
+		} else if id, hasId := args["id"].(string); hasId {
+			return hasProjectPermission(ctx, id, next, permissionLevel)
+		} else if projectId, hasProjectId := args["projectId"].(string); hasProjectId {
```
khelif96, line 4107: I wonder if it would make sense enforcing a consistent variable name for this field. Since we call it different things in different resolvers but its referencing the same field.

PR #5368, opened 2022-02-07
EVG-15475 Remove collection of mock connectors

hadjri:
[EVG-15475](https://jira.mongodb.org/browse/EVG-15475)

### Description 
Unit tests have been refactored to use the DBConnector and not the mock one in hopes that we can remove the mock connector entirely once this is done for the remaining mock connectors
### Testing 
Removed tests and checked they are still passing (hopefully)


rest/route/github_test.go
```diff
@@ -23,10 +23,12 @@ import (
 	"github.com/mongodb/amboy"
 	"github.com/mongodb/grip"
 	"github.com/stretchr/testify/suite"
+	"go.mongodb.org/mongo-driver/bson"
 )
 
 type GithubWebhookRouteSuite struct {
-	sc                     *data.MockConnector
+	ss                     *data.MockConnector
```
ablack12, line 30: I think the c here typically means connector; could we have ss be mock or something so it's clear? (I know we intend to get rid of this ASAP but just in case we make the mock more of a subclass of the DB connector in the future)

rest/route/github_test.go
```diff
@@ -125,6 +101,18 @@ func TestGithubWebhookRouteSuite(t *testing.T) {
 }
 
 func (s *GithubWebhookRouteSuite) TestAddIntent() {
+	s.NoError(db.Clear(model.ProjectRefCollection))
```
ablack12, line 104: can do `s.NoError(db.ClearCollections(model.ProjectRefCollection, patch.IntentCollection))` instead

rest/route/github_test.go
```diff
@@ -236,25 +227,25 @@ func (s *GithubWebhookRouteSuite) TestPushEventTriggersRepoTracker() {
...
-	}
-
-	s.NoError(err)
-	if s.Len(s.sc.MockCommitQueueConnector.Queue, 1) {
-		s.Equal("1", utility.FromStringPtr(s.sc.MockCommitQueueConnector.Queue["bth"][0].Issue))
-		s.Equal("test_module", utility.FromStringPtr(s.sc.MockCommitQueueConnector.Queue["bth"][0].Modules[0].Module))
-		s.Equal("1234", utility.FromStringPtr(s.sc.MockCommitQueueConnector.Queue["bth"][0].Modules[0].Issue))
-	}
-}
+//func (s *GithubWebhookRouteSuite) TestCommitQueueCommentTrigger() {
```
ablack12, line 230: Should this be commented out?

rest/route/github_test.go
```diff
@@ -359,26 +356,26 @@ func (s *GithubWebhookRouteSuite) TestTryDequeueCommitQueueItemForPR() {
...
-	pRef := model.ProjectRef{
-		Id:                    "my-project",
-		GitTagAuthorizedUsers: []string{"release-bot", "not-release-bot"},
-		GitTagVersionsEnabled: utility.TruePtr(),
-	}
-	v, err := s.h.createVersionForTag(context.Background(), pRef, nil, model.Revision{}, tag, "")
-	s.NoError(err)
-	s.NotNil(v)
-}
+//func (s *GithubWebhookRouteSuite) TestCreateVersionForTag() {
```
ablack12, line 359: Should this be commented out?

ablack12 (commented):
This is a great start 😁

rest/route/build_test.go
```diff
@@ -21,8 +22,8 @@ import (
 // Tests for fetch build by id
 
 type BuildByIdSuite struct {
-	sc   *data.MockConnector
-	data data.MockBuildConnector
+	sc   *data.DBConnector
+	data data.DBBuildConnector
```
ablack12, line 26: it's not really clear to me why we need both data and DBConnector; should we just get rid of data?

rest/route/github_mock.go
```diff
@@ -0,0 +1,218 @@
...
+	secret []byte
+
+	event     interface{}
+	eventType string
+	msgID     string
+	sc        data.Connector
+	settings  *evergreen.Settings
+}
+
+func makeMockGithubHooksRoute(sc data.Connector, queue amboy.Queue, secret []byte, settings *evergreen.Settings) gimlet.RouteHandler {
```
ablack12, line 35: Similarly, instead of mocking the route itself, I think we need a small mock connector

rest/route/patch_test.go
```diff
@@ -396,16 +437,18 @@ func (s *PatchRestartSuite) TestRestart() {
 	s.NotNil(res)
 
 	s.Equal(http.StatusOK, res.Status())
-	s.Equal("user1", s.sc.CachedRestartedVersions[s.objIds[0]])
+	restartedPatch, err := s.sc.FindPatchById(s.objIds[0])
+	s.NoError(err)
+	s.Equal("patch_request", *restartedPatch.Requester)
```
ablack12, line 442: nit: can you use PatchVersionRequester instead of hardcoding?

rest/route/patch_test.go
```diff
@@ -521,13 +567,13 @@ func (s *PatchesByUserSuite) TestEmptyTimeShouldSetNow() {
...
-			URL: "https://evergreen.example.net",
-			MockPatchConnector: data.MockPatchConnector{
-				CachedRawPatches: map[string]string{
-					"":        "main diff",
-					"module1": "module1 diff",
-				},
+		sc: &data.DBConnector{
+			URL:              "https://evergreen.example.net",
+			DBPatchConnector: data.DBPatchConnector{
+				//CachedRawPatches: map[string]string{
```
ablack12, line 573: Should this be commented out?

rest/route/project_events_test.go
```diff
@@ -15,44 +17,43 @@ import (
...
 func TestProjectEventsTestSuite(t *testing.T) {
 
 	suite.Run(t, new(ProjectEventsTestSuite))
 }
 
-func getMockProjectSettings(projectId string) restModel.APIProjectSettings {
-	return restModel.APIProjectSettings{
-		ProjectRef: restModel.APIProjectRef{
-			Owner:      utility.ToStringPtr("admin"),
+func getMockProjectSettings(projectId string) model.ProjectSettings {
```
ablack12, line 32: Can we call this getTestProjectSettings or something instead?

rest/route/project_events_test.go
```diff
@@ -67,21 +68,28 @@ func (s *ProjectEventsTestSuite) SetupSuite() {
...
-	s.sc = &data.MockConnector{
-		URL:                  "https://evergreen.example.net",
-		MockProjectConnector: s.data,
+	projectRef := &model.ProjectRef{
+		Id:      "mci2",
+		Enabled: utility.TruePtr(),
 	}
+	s.NoError(projectRef.Insert())
+
+	s.NoError(model.LogProjectEvent("PROJECT_ADDED", "mci2", s.event))
```
ablack12, line 92: nit: can we use EventTypeProjectAdded here instead of hardcoding?

rest/route/project_test.go
```diff
@@ -140,7 +137,9 @@ func (s *ProjectPatchByIDSuite) TestRunWithValidBbConfig() {
 	s.NotNil(resp)
 	s.NotNil(resp.Data())
 	s.Require().Equal(http.StatusOK, resp.Status())
-	s.Require().Equal("EVG", s.sc.MockProjectConnector.CachedProjects[0].BuildBaronSettings.TicketCreateProject)
+	pRef, err := s.sc.DBProjectConnector.FindProjectById("dimoxinil", false, false)
```
ablack12, line 140: Do we need to specify the connector type? i think s.sc.FindProjectById should be fine

rest/route/project_test.go
```diff
@@ -464,8 +467,8 @@ func (s *ProjectGetByIDSuite) TestRunExistingId() {
 // Tests for GET /rest/v2/projects
 
 type ProjectGetSuite struct {
-	data  data.MockProjectConnector
-	sc    *data.MockConnector
+	data  data.DBProjectConnector
```
ablack12, line 470: I also don't think we need data here either really

rest/route/project_test.go
```diff
@@ -559,45 +576,47 @@ func (s *ProjectGetSuite) TestGetRecentVersions() {
...
+	return &serviceModel.ProjectRef{
+		Owner:      "dimoxinil",
+		Repo:       "dimoxinil-enterprise-repo",
+		Branch:     "main",
+		Enabled:    utility.FalsePtr(),
+		Private:    utility.TruePtr(),
+		BatchTime:  0,
+		RemotePath: "evergreen.yml",
+		Id:         "dimoxinil",
+		//	Identifier:         "dimoxinil",
```
ablack12, line 603: Should this be commented out?

rest/route/project_test.go
```diff
@@ -957,7 +979,7 @@ func (s *ProjectPutRotateSuite) TestRotateProjectVars() {
 	s.Contains(respMap["dimoxinil"], "banana")
 	s.Contains(respMap["dimoxinil"], "lemon")
 	s.Equal(resp.Status(), http.StatusOK)
-	s.Equal("yellow", s.sc.CachedVars[0].Vars["banana"])
+	//s.Equal("yellow", s.sc.CachedVars[0].Vars["banana"])
```
ablack12, line 982: Should this be commented out?

rest/route/service_test.go
```diff
@@ -1064,19 +1081,20 @@ func TestTaskGetHandler(t *testing.T) {
...
 			Activated:    false,
 			Secret:       "initialSecret",
 			DispatchTime: timeNow,
 		}
-		sc.MockTaskConnector.CachedTasks = append(sc.MockTaskConnector.CachedTasks, testTask)
+		So(testTask.Insert(), ShouldBeNil)
 		ctx := context.Background()
 		Convey("and an error from the service function", func() {
-			sc.MockTaskConnector.StoredError = fmt.Errorf("could not reset task")
+			//sc.DBTaskConnector.StoredError = fmt.Errorf("could not reset task")
```
ablack12, line 1097: We can remove this entirely right?

ablack12 (changes requested):
This looks really good! Given the functions that are tested that use Github I'm pretty sure we can't get rid of the MockConnector entirely; I think maybe we should always use the MockConnector but for almost all of the small connectors it uses the DB ones, and just for some special cases it uses the mock ones. Let me know if this needs clarification. (Or possibly we can just have an interface for the Github functions and mock only those.) Let me know if you want to have an IRL discussion.

rest/data/interface.go
```diff
@@ -398,20 +398,24 @@ type Connector interface {
...
-	MockGetAdditionalPatches(patchId string) ([]string, error)
+type MockConnector interface {
+	// Commit queue methods
+	// GetGithubPR takes the owner, repo, and PR number.
+	GetGitHubPR(context.Context, string, string, int) (*github.PullRequest, error)
+	// if bool is true, move the commit queue item to be processed next.
+	EnqueueItem(string, restModel.APICommitQueueItem, bool) (int, error)
+	AddPatchForPr(ctx context.Context, projectRef model.ProjectRef, prNum int, modules []restModel.APIModule, messageOverride string) (string, error)
+	FindCommitQueueForProject(string) (*restModel.APICommitQueue, error)
+	EnableCommitQueue(*model.ProjectRef) error
```
ablack12, line 411: A number of these methods don't need to be mocked though, right? EnqueueItem, FindCommitQueueForProject, etc all only touch the DB commit queue

ablack12: How many functions actually require being mocked (would call Github otherwise?) I'm still not convinced we shouldn't just remove those tests. (We should maybe have an in-person sync about this)

model/commitqueue/commit_queue_test.go
```diff
@@ -31,27 +32,31 @@ var sampleCommitQueueItem = CommitQueueItem{
 
 func TestCommitQueueSuite(t *testing.T) {
 	s := new(CommitQueueSuite)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env := testutil.NewEnvironment(ctx, t)
```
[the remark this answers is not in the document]
hadjri: Adding DB operations to so many tests at once caused many context cancelled errors like this one: https://evergreen.mongodb.com/lobster/evergreen/test/evergreen_ubuntu1604_test_rest_route_patch_247e88dd4f3510209853ab20b10d0827247703c9_62148a205623435871e7bcb8_22_02_22_07_00_49/0/output.rest-route.test/#bookmarks=0%2C6229&l=1&shareLine=5190
There would be 100+ faliures for a given version with this same error, and adding these lines fixed the issue. I had thought we'd need to explicitly recycle the contexts used in each test from now on. These tests would all succeed individually but fail when running all together from the makefile.

rest/data/build_test.go
```diff
@@ -15,42 +15,32 @@ import (
 // Tests for fetch build by id route
 
 type BuildConnectorFetchByIdSuite struct {
-	ctx  Connector
-	mock bool
+	ctx Connector
 	suite.Suite
+	env *mock.Environment
```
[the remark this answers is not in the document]
hadjri: Nope, removing, thanks

rest/data/commit_queue_test.go
```diff
@@ -248,15 +248,15 @@ func (s *CommitQueueSuite) TestMockGetGitHubPR() {
 }
 
 func (s *CommitQueueSuite) TestMockEnqueue() {
-	s.ctx = &MockConnector{}
-	pos, err := s.ctx.EnqueueItem("mci", restModel.APICommitQueueItem{Source: utility.ToStringPtr(commitqueue.SourceDiff), Issue: utility.ToStringPtr("1234")}, false)
+	s.ctx = &DBConnector{}
+	pos, err := s.ctx.MockEnqueueItem("mci", restModel.APICommitQueueItem{Source: utility.ToStringPtr(commitqueue.SourceDiff), Issue: utility.ToStringPtr("1234")}, false)
```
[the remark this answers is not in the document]
hadjri: Sounds good, removed all of them except for the few that are actually needed

rest/data/patch_test.go
```diff
@@ -35,45 +33,15 @@ type PatchConnectorFetchByProjectSuite struct {
...
-	}
-
-	s.teardown = func() error {
-		return db.ClearCollections(patch.Collection, dbModel.ProjectRefCollection)
-	}
-
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env := testutil.NewEnvironment(ctx, t)
+	evergreen.SetEnvironment(env)
```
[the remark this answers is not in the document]
hadjri: Here it is used to set the global environment which is used later in subtests

rest/data/pod_test.go
```diff
@@ -1,20 +1,28 @@
...
 func TestPodConnector(t *testing.T) {
-	for tName, tCase := range map[string]func(t *testing.T, conn Connector){
-		"CreatePodSucceeds": func(t *testing.T, conn Connector) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env := testutil.NewEnvironment(ctx, t)
+	evergreen.SetEnvironment(env)
+	for tName, tCase := range map[string]func(ctx context.Context, t *testing.T){
+		"CreatePodSucceeds": func(ctx context.Context, t *testing.T) {
+			conn := &DBConnector{}
```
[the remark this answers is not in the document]
hadjri: This is residual of me trying to resolve the context cancelled bug, I thought this was necessary. Removing now.

rest/route/github_mock.go
hadjri: Mock connector is added, removed this now

rest/route/patch_test.go
hadjri: Removed, this was an old property on the mock.

rest/route/service_test.go
hadjri: Yes - removing

rest/route/project_test.go
hadjri: CachedVars are an old property on the mock connector that isnt needed anymore, deleted now.

rest/route/project_test.go
hadjri: Byproduct of debugging, uncommented now

rest/route/project_test.go
hadjri: True, removing

rest/route/github_test.go
hadjri: Uncommented and test now working

rest/route/github_test.go
hadjri: Same as above, working now

rest/data/interface.go
hadjri: I've created a new separate mock connector interface that implements these few methods that need to be mocked (the ones talking to GitHub) and implemented them all in one generic mock connector implementation called MockDBConnectorImpl

rest/data/impl.go
```diff
@@ -33,45 +33,10 @@ type DBConnector struct {
 	TaskReliabilityConnector
 	DBCommitQueueConnector
 	SchedulerConnector
+	MockCommitQueueConnector
```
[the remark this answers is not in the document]
hadjri: I've done this with the caveat being MockDBConnectorImpl is what exists instead of MockCommitQueueConnector since there are a few host / alias related methods that turned out needing to be mocked as well. They're few and far between so I put them in the same implementation for cleanliness.

rest/route/service_test.go
```diff
@@ -1,1257 +1,1327 @@
...
-	for idx := range expected {
-		m, ok := data[idx].(model.Model)
-
-		if assert.True(t, ok) {
-			assert.Equal(t, expected[idx], m)
-		}
-	}
-}
+//
+//import (
```
ablack12, line 4: Leftover testing? :)

rest/data/interface.go
```diff
@@ -407,15 +407,9 @@ type MockConnector interface {
...
-	CommitQueueRemoveItem(string, string, string) (*restModel.APICommitQueueItem, error)
-	IsItemOnCommitQueue(string, string) (bool, error)
-	CommitQueueClearAll() (int, error)
 	CreatePatchForMerge(context.Context, string, string) (*restModel.APIPatch, error)
-	IsPatchEmpty(string) (bool, error)
 	IsAuthorizedToPatchAndMerge(context.Context, *evergreen.Settings, UserRepoInfo) (bool, error)
-	GetMessageForPatch(string) (string, error)
-	ConcludeMerge(string, string) error
-	GetAdditionalPatches(patchId string) ([]string, error)
+	HasMatchingGitTagAliasAndRemotePath(string, string) (bool, string, error)
```
ablack12, line 412: This function doesn't seem to use Github; does it need to be mocked?

rest/data/interface.go
```diff
@@ -407,15 +407,9 @@ type MockConnector interface {
...
+	GetProjectFromFile(context.Context, model.ProjectRef, string, string) (model.ProjectInfo, error)
+	CreateVersionFromConfig(context.Context, *model.ProjectInfo, model.VersionMetadata, bool) (*model.Version, error)
```
ablack12, line 414: I'm also not 100% certain why this needs to be mocked; in the existing codebase the only tests I see that use this don't use the mock version.

rest/data/interface.go
```diff
@@ -399,3 +399,17 @@ type Connector interface {
 	// CompareTasks returns the order that the given tasks would be scheduled, along with the scheduling logic.
 	CompareTasks([]string, bool) ([]string, map[string]map[string]string, error)
 }
+
+type MockConnector interface {
+	// Commit queue methods
+	// GetGithubPR takes the owner, repo, and PR number.
+	GetGitHubPR(context.Context, string, string, int) (*github.PullRequest, error)
+	// if bool is true, move the commit queue item to be processed next.
+	EnqueueItem(string, restModel.APICommitQueueItem, bool) (int, error)
```
ablack12, line 408: I don't believe this function needs to be mocked; it just adds something to the commit queue

rest/data/interface.go
```diff
@@ -399,3 +399,17 @@ type Connector interface {
 	// CompareTasks returns the order that the given tasks would be scheduled, along with the scheduling logic.
 	CompareTasks([]string, bool) ([]string, map[string]map[string]string, error)
 }
+
+type MockConnector interface {
```
ablack12, line 403: Why do we still have this interface? It seems like MockDBConnector and MockDBConnectorImpl is what we want. If it seems like some amount of work to get MockDBConnector to just implement Connector (and remove this mock interface) we could break that into a different ticket since tests are passing.

rest/data/impl.go
```diff
@@ -40,38 +40,7 @@ func (ctx *DBConnector) SetURL(url string)       { ctx.URL = url }
...
-	MockCLIUpdateConnector
-	MockGenerateConnector
-	MockSubscriptionConnector
-	MockNotificationConnector
-	MockCreateHostConnector
-	MockStatsConnector
-	MockTaskReliabilityConnector
-	MockCommitQueueConnector
-	MockSchedulerConnector
+type MockDBConnector struct {
```
ablack12, line 43: This implements Connector right?

rest/data/impl.go
```diff
@@ -40,38 +40,7 @@ func (ctx *DBConnector) SetURL(url string)       { ctx.URL = url }
...
+	DBConnector
+	MockDBConnectorImpl
```
ablack12, line 45: Can we call this MockGithubConnector to be clear about what kinds of things we need to mock?

rest/route/service_test.go
hadjri: Yes this file in particular had been difficult to get to pass so I had it commented out while I work on them as the last failing tests. Uncommented now

rest/data/impl.go
hadjri: Yes

rest/data/interface.go
hadjri: You're right we don't need this interface, removing it entirely doesnt cause any errors. MockDBConnector already implements Connector.

rest/data/interface.go
hadjri: This is the only case where we are creating a git tagged version along with it, which requires authenticating with Github

hadjri: in TestCreateVersionForTag

rest/data/interface.go
hadjri: You're right I was able to remove it with a bit of refactoring

rest/data/interface.go
hadjri: It is used in conjunction with other DB calls in the createVersionForTag function but it was actually just checking the alias DB, so no need to mock. Removed

hadjri:
The MockGitHubConnector should now only mock the 5 essential connector functions that are talking to GH. The rest of the codebase is using the real DBConnector.

rest/data/mock_impl.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (mvc *MockGitHubConnectorImpl) CreateVersionFromConfig(ctx context.Context, projectInfo *model.ProjectInfo, metadata model.VersionMetadata, active bool) (*model.Version, error) {
+	return &model.Version{
+		Requester:         evergreen.GitTagRequester,
+		TriggeredByGitTag: metadata.GitTag,
+	}, nil
+}
+
+func (mtc *MockGitHubConnectorImpl) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]testresult.TestResult, error) {
```
ablack12, line 90: Do tests need to be mocked? These things don't touch github.

ablack12 (approved):
I have a question about why test things are still in the mock, but if removing them is difficult let's split it to another ticket.

I think there is one more ticket that should come out of this: a major reason why the interface is a developer pain point is because the interface breaks "jump to function definition" for editors. 

I think the only functions that actually _need_ to be in the interface are the ones that are being mocked -- every other function _could_ be removed from the DB and we could either just call the base function that it's wrapping (since a lot of them do that) or move them to be a new function in their relevant packages. A lot of routes might not even need to take in a connector anymore! Definitely an idea for future work but wanted your thoughts on it.

rest/data/mock_impl.go
hadjri: This doesn't need to be mocked per se but this test in particular has been quite difficult to mock and might need a large refactor to get it working (has to do with the fact that test ids are mgobson.ObjectIds and not strings, and this test runs into sorting issues when returning tests vs the hardcoded expected sequential list). I wrangled with it for a while and I think it's best to create a separate ticket, as you can see I reverted it to its mock behavior for now.

hadjri:
> I have a question about why test things are still in the mock, but if removing them is difficult let's split it to another ticket.
> 
> I think there is one more ticket that should come out of this: a major reason why the interface is a developer pain point is because the interface breaks "jump to function definition" for editors.
> 
> I think the only functions that actually _need_ to be in the interface are the ones that are being mocked -- every other function _could_ be removed from the DB and we could either just call the base function that it's wrapping (since a lot of them do that) or move them to be a new function in their relevant packages. A lot of routes might not even need to take in a connector anymore! Definitely an idea for future work but wanted your thoughts on it.

I definitely like this idea - the jump to function flow with the interfaces is indeed a bit cumbersome and I think it'd be a good idea to only keep the mocked functions in an interface.  Happy to create a ticket for this as well and pull it into an upcoming sprint. I optimistically think that it wouldn't be too big of an effort.

PR #5371, opened 2022-02-07
EVG-16242 Add validations for project configs in project validation

hadjri:
[EVG-16242](https://jira.mongodb.org/browse/EVG-16242)

### Description 
Adding validation for plugin configs and project aliases defined in the project yaml to a the project validation
### Testing 
Added existing unit tests.


ablack12 (changes requested):
We should also validate [periodic builds](https://github.com/evergreen-ci/evergreen/blob/ccf58c1fb6813e43c1eb9b84006b31aab596bff0/service/project.go#L596).

validator/project_validator.go
```diff
@@ -389,6 +404,108 @@ func dependencyCycleExists(node model.TVPair, allNodes []model.TVPair, visited m
...
+			Message: errorMsg,
+			Level:   Error,
+		})
+	}
+	return validationErrs
+}
+
+func validatePlugins(pc *model.ProjectConfig) ValidationErrors {
+	errs := ValidationErrors{}
+	webHook, _, err := model.IsWebhookConfigured(pc.Project, "")
```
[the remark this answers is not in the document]
hadjri: I wasn't totally sure if we should check against the project ref since it _would_ be an issue if a custom was configured on the project ref level and a BB config was specified in yaml if the BB config on the project page was empty.  This is because FindMergedProjectRef would merge the two entities and then both custom webhook and BB config would be defined.  However since this is really meant to strictly validate the project-config I think it's good to change this to use the passed in ProjectConfig.

validator/project_validator.go
```diff
@@ -217,7 +226,7 @@ func CheckProjectWarnings(project *model.Project, yamlBytes []byte) ValidationEr
 }
 
 // verify that the project configuration syntax is valid
-func CheckProjectErrors(project *model.Project, includeLong bool) ValidationErrors {
+func CheckProjectErrors(project *model.Project, projectConfig *model.ProjectConfig, includeLong bool) ValidationErrors {
```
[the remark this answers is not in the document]
hadjri: I did this since I had considered there might be places where we call CheckProjectErrors outside of the logic of the validation route where we'd want to also validate the ProjectConfig.  On second thought though I think it's best to just keep this in the validation route (similar to how we only check for project warnings in the validation route) and leave CheckProjectErrors unchanged.

validator/project_validator.go
```diff
@@ -389,6 +404,108 @@ func dependencyCycleExists(node model.TVPair, allNodes []model.TVPair, visited m
...
+
+func validatePlugins(pc *model.ProjectConfig) ValidationErrors {
+	errs := ValidationErrors{}
+	webHook, _, err := model.IsWebhookConfigured(pc.Project, "")
+	if err != nil {
+		return errs
+	}
+	webhookConfigured := webHook.Endpoint != ""
+
+	if !webhookConfigured && pc.BuildBaronSettings.TicketCreateProject == "" && len(pc.BuildBaronSettings.TicketSearchProjects) == 0 {
```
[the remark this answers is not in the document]
hadjri: Sure, I can refactor it a bit

validator/project_validator.go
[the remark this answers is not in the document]
ablack12: Sorry I think what I actually meant was that the ProjectConfig would be looked up when looking up the ProjectRef:
https://github.com/evergreen-ci/evergreen/blob/ccf58c1fb6813e43c1eb9b84006b31aab596bff0/model/project_ref.go#L2407-L2406
> This is because FindMergedProjectRef would merge the two entities and then both custom webhook and BB config would be defined.
It seems like when merging we should ignore both webhook & BB config from the project config if it's configured for the project ref (since that takes precedence) rather than erroring for validation.

validator/project_validator.go
```diff
@@ -135,6 +137,12 @@ var projectErrorValidators = []projectValidator{
 	validateAliases,
 }
 
+// Functions used to validate the syntax of project configs representing properties found on the project page.
+var projectConfigErrorValidators = []projectConfigValidator{
```
ablack12, line 141: Wanted to make sure you saw this comment https://github.com/evergreen-ci/evergreen/pull/5371#pullrequestreview-875226345

hadjri: My bad! Thanks for bubbling it back up

model/project_ref.go
```diff
@@ -2284,53 +2284,59 @@ func GetBuildBaronSettings(projectId string, version string) (evergreen.BuildBar
 	return projectRef.BuildBaronSettings, true
 }
 
-func BbProjectIsValid(projName string, proj evergreen.BuildBaronSettings) error {
-	webHook, _, err := IsWebhookConfigured(projName, "")
-	if err != nil {
-		return err
+func ValidateBbProject(projName string, proj evergreen.BuildBaronSettings, webhook *evergreen.WebHook, lookupProjectRef bool) []string {
```
ablack12, line 2287: Returning a list of strings as errors is a bit unconventional (ValidateProjectAliases I think does it but probably shouldn't). If all of these count as errors for validation, we can just use a catcher here and return a singular error, like the periodic build validation.

model/project_ref.go
```diff
@@ -2284,53 +2284,59 @@ func GetBuildBaronSettings(projectId string, version string) (evergreen.BuildBar
...
+	errs := []string{}
+	var err error
+	var webhookConfigured bool
```
ablack12, line 2290: This should default to the passed in webhook right?

rest/route/project.go
```diff
@@ -508,9 +508,12 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(catcher.Resolve(), "error validating triggers"))
 	}
 
-	err = dbModel.BbProjectIsValid(h.newProjectRef.Id, h.newProjectRef.BuildBaronSettings)
-	if err != nil {
-		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "error validating build baron config"))
+	errs := dbModel.ValidateBbProject(h.newProjectRef.Id, h.newProjectRef.BuildBaronSettings, nil, true)
```
ablack12, line 511: Should this be using newProjectRef.TaskAnnotationSettings.Webhook instead of looking it up?

service/project.go
```diff
@@ -567,9 +567,13 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Errorf("expected build baron config but was actually '%T'", i))
 		return
 	}
-	err = model.BbProjectIsValid(projectRef.Id, buildbaronConfig)
-	if err != nil {
-		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "error validating build baron config input"))
+	catcher := grip.NewSimpleCatcher()
+	errs = model.ValidateBbProject(projectRef.Id, buildbaronConfig, nil, true)
```
ablack12, line 571: Same question about the webhook

ablack12 (changes requested):

rest/route/project.go
hadjri: Yes since we actually do have access to the webhook in all places where we call ValidateBbProject, we can directly pass it in for all cases.  Since FindMergedProjectRef already prevents webhook & BB config from picking up the project config values if webhook & BB are configured for the project ref we don't need to consider that case during validation.  Changing to pass in newProjectRef.TaskAnnotationSettings.Webhook

service/project.go
hadjri: Same as above

model/project_ref.go
```diff
@@ -2284,58 +2284,57 @@ func GetBuildBaronSettings(projectId string, version string) (evergreen.BuildBar
...
-		webhookConfigured = webhook.Endpoint != ""
+		webhookConfigured = webhook != nil && webhook.Endpoint != ""
 	}
 
 	if !webhookConfigured && proj.TicketCreateProject == "" && len(proj.TicketSearchProjects) == 0 {
 		return nil
 	}
 	if !webhookConfigured && len(proj.TicketSearchProjects) == 0 {
-		errs = append(errs, "Must provide projects to search")
+		catcher.Add(errors.Errorf("Must provide projects to search"))
```
ablack12, line 2304: catcher.New

model/project_ref.go
```diff
@@ -2284,58 +2284,57 @@ func GetBuildBaronSettings(projectId string, version string) (evergreen.BuildBar
...
 	}
 	if !webhookConfigured && proj.TicketCreateProject == "" {
-		errs = append(errs, "Must provide project to create tickets for")
+		catcher.Add(errors.Errorf("Must provide project to create tickets for"))
 	}
 	if proj.BFSuggestionServer != "" {
 		if _, err = url.Parse(proj.BFSuggestionServer); err != nil {
-			errs = append(errs, fmt.Sprintf("Failed to parse bf_suggestion_server for project '%s'", projName))
+			catcher.Add(errors.Errorf("Failed to parse bf_suggestion_server for project '%s'", projName))
```
ablack12, line 2311: catcher.Errorf

validator/project_validator.go
```diff
@@ -446,11 +446,11 @@ func validateProjectConfigPlugins(pc *model.ProjectConfig) ValidationErrors {
...
 		webhook = &annotationSettings.FileTicketWebhook
 	}
-	buildBaronErrs := model.ValidateBbProject(pc.Project, *pc.BuildBaronSettings, webhook, false)
-	for _, errMsg := range buildBaronErrs {
+	err := model.ValidateBbProject(pc.Project, *pc.BuildBaronSettings, webhook)
+	if err != nil {
 		errs = append(errs,
 			ValidationError{
-				Message: errMsg,
+				Message: errors.Wrap(err, "error validating build baron config").Error(),
```
ablack12, line 453: Can we explicitly add the Level: Error here for clarity?

ablack12 (approved):
LGTM but can you manually test creating a patch and running validate with a broken project config before merging?

bynn (approved):

PR #5374, opened 2022-02-08
EVG-16088 Only surface enabled projects on project selector

khelif96:
[EVG-16088](https://jira.mongodb.org/browse/EVG-16088)

### Description 
Project selector now only shows enabled projects


graphql/resolvers.go
```diff
@@ -1595,7 +1595,7 @@ func (r *queryResolver) TaskAllExecutions(ctx context.Context, taskID string) ([
 }
 
 func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error) {
```
ablack12, line 1597: I'm surprised that this is what's used for the dropdown; I think a future project could be a resolver that just returns the grouped names (as opposed to all of the info) since that's a lot of querying but I'm not sure if the project info is also being used in other ways.

model/project_ref.go
```diff
@@ -1078,6 +1078,21 @@ func FindFirstProjectRef() (*ProjectRef, error) {
 	return projectRef, nil
 }
 
+// FindAllMergedTrackedEnabledProjectRefs returns all enabled project refs in the db
+// that are currently being tracked (i.e. their project files
+// still exist and the project is not hidden).
+// Can't hide a repo without hiding the branches, so don't need to aggregate here.
+func FindAllMergedTrackedEnabledProjectRefs() ([]ProjectRef, error) {
+	projectRefs := []ProjectRef{}
+	q := db.Query(bson.M{ProjectRefHiddenKey: bson.M{"$ne": true}, ProjectRefEnabledKey: bson.M{"$ne": false}})
```
ablack12, line 1087: This will return projects that are enabled _and_ projects that have enabled set to nil (i.e. they default to the repo, or else they're false). It would make sense then to re-iterate through the projects after 1093 to make sure that all the remaining projects _are_ enabled.

model/project_ref.go
```diff
@@ -1078,6 +1078,21 @@ func FindFirstProjectRef() (*ProjectRef, error) {
 	return projectRef, nil
 }
 
+// FindAllMergedTrackedEnabledProjectRefs returns all enabled project refs in the db
+// that are currently being tracked (i.e. their project files
+// still exist and the project is not hidden).
+// Can't hide a repo without hiding the branches, so don't need to aggregate here.
+func FindAllMergedTrackedEnabledProjectRefs() ([]ProjectRef, error) {
```
ablack12, line 1085: Can you write a unit test for this?

model/project_ref.go
```diff
@@ -1078,6 +1078,21 @@ func FindFirstProjectRef() (*ProjectRef, error) {
 	return projectRef, nil
 }
 
+// FindAllMergedTrackedEnabledProjectRefs returns all enabled project refs in the db
+// that are currently being tracked (i.e. their project files
+// still exist and the project is not hidden).
+// Can't hide a repo without hiding the branches, so don't need to aggregate here.
```
ablack12, line 1084: This isn't super relevant to this function given my comment below, i.e. that we _do_ need to consider the repo value.

model/project_ref.go
ablack12: Also can we split this out into multiple lines to make it cleaner since there are multiple query parameters?

ablack12 (changes requested):

graphql/resolvers.go
```diff
@@ -1599,8 +1599,13 @@ func (r *queryResolver) Projects(ctx context.Context) ([]*GroupedProjects, error
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
-
-	groupedProjects, err := GroupProjects(allProjs, false)
+	enabledProj := []model.ProjectRef{}
```
ablack12, line 1602: nit: can we call this enabledProjects since it's a list (and more consistent with groupedProjects)

ablack12: Also if you think this change will confuse someone in the future, I would add a comment saying we need to iterate over the merged projects which is why we don't query directly (or something)

ablack12 (approved):

PR #5377, opened 2022-02-09
EVG-16251, EVG-16260: reduce duplicate event processing and limit the max parallel events processed

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-16251
https://jira.mongodb.org/browse/EVG-16260

### Description
This is supposed to make the more efficient by reducing the duplicate events processed during times when there are a lot of unprocessed events, while also throttling the max number of events that can be simultaneously processed in parallel to avoid overloading the DB.

This change is risky, so I'm going to do the deploy after it's merged. If this doesn't work (e.g. the parallel processing limit is too high/low), the only work to rollback should be reverting this change and marking some jobs in the queue group complete.

* Process individual events instead of batches of events in the event notifier.
    * It now ignores the event processing batch size. This batch size can be removed from the admin settings at a later point if this change actually works.
* Use scopes on each event notifier to prevent duplicate event/subscriptions processing.
* Move the event notifier to the queue group and give it a temporarily hard-coded set of options that will throttle the event processing rate. The throttling is indirectly achieved due to the fact that there's a fixed number of Amboy workers per app server and fixed number of app servers. I did some calculations and it seems like under normal load, the apps can process 100 events in parallel with no issues, so I set the number of workers to achieve the same level of throughput as under normal load situations.
* Do some trivial logging changes since some durations were logged in seconds, while others were logged in nanoseconds.

### Testing 
* Updated existing tests.
* Manually tested in staging by creating a few hundred unprocessed events to verify that the events were still processed as expected and with a reasonable-looking throughput.

### DIFF
--- a/units/crons.go
+++ b/units/crons.go
@@ -93,7 +93,7 @@ func PopulateHostMonitoring(env evergreen.Environment) amboy.QueueOperation {
 	}
 }
 
-func PopulateEventAlertProcessing(env evergreen.Environment) amboy.QueueOperation {
+func PopulateEventSendJobs(env evergreen.Environment) amboy.QueueOperation {
 	return func(ctx context.Context, queue amboy.Queue) error {
 		flags, err := evergreen.GetServiceFlags()
 		if err != nil {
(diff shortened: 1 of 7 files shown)


units/crons_remote_minute.go
```diff
@@ -90,6 +90,13 @@ func (j *cronsRemoteMinuteJob) Run(ctx context.Context) {
 		catcher.Add(PopulateCommitQueueJobs(j.env)(ctx, commitQueueQueue))
 	}
 
+	eventNotifierQueue, err := j.env.RemoteQueueGroup().Get(appCtx, "service.event.notifier")
```
ablack12, line 93: Can the comment on line 77 extend to this?

units/crons.go
```diff
@@ -109,17 +109,24 @@ func PopulateEventAlertProcessing(env evergreen.Environment) amboy.QueueOperatio
...
+			return errors.WithStack(err)
 		}
-		limit := notifySettings.EventProcessingLimit
-		if limit <= 0 {
-			limit = evergreen.DefaultEventProcessingLimit
+
+		if flags.EventProcessingDisabled {
+			grip.InfoWhen(sometimes.Percent(evergreen.DegradedLoggingPercent), message.Fields{
+				"message": "alerts disabled",
+				"impact":  "not processing alerts for notifications",
```
ablack12, line 126: Can this message be different than the one for PopulateEventSendJobs?

units/crons.go
```diff
@@ -109,17 +109,24 @@ func PopulateEventAlertProcessing(env evergreen.Environment) amboy.QueueOperatio
...
 		}
 
-		err = dispatchUnprocessedNotifications(ctx, queue, flags)
-		grip.Error(message.WrapError(err, message.Fields{
-			"message": "unable to dispatch unprocessed notifications",
-		}))
-		notifySettings := env.Settings().Notify
-		if err = notifySettings.Get(env); err != nil {
-			return errors.Wrap(err, "unable to get notify settings")
+		return dispatchUnprocessedNotifications(ctx, queue, flags)
```
ablack12, line 112: I trust the process but I'm a little unclear on why we dispatch notifications both here and in the event notifier job.

Kimchelly: I didn't write this, but I'm pretty sure this is just a performance boost to reduce the waiting time between creating the notification and actually sending the notification.

units/crons.go
Kimchelly: Yes, fixed.

units/crons_remote_minute.go
Kimchelly: Done.

ybrill (approved):

ablack12 (approved):

bsamek:
@Kimchelly I removed myself as a reviewer, since I likely wouldn't be able to review this for some time. Please let me know if you'd prefer that I do review it.

PR #5381, opened 2022-02-10
EVG-16243 Remove CQ and several toggles from project configs

hadjri:
[EVG-16243](https://jira.mongodb.org/browse/EVG-16243)

### Description 
Commit queue params and several toggles related to github project configurations have been removed from the newly exposed set of configurations, since enabling these properties via version control may lead to unexpected behavior when restarting old tasks that have their config yaml version conflicting with what is set on the project page.  Two toggles have been kept: perf enabled and deactivate previous, since it looks as though keeping these under version control would not be problematic.
### Testing 
Modified existing unit tests.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -465,8 +465,8 @@ func (p *ProjectRef) MergeWithProjectConfig(version string) error {
 		reflectedRef := reflect.ValueOf(p).Elem()
 		reflectedConfig := reflect.ValueOf(pRefToMerge)
 		recursivelySetUndefinedFields(reflectedRef, reflectedConfig)
-		p.PerfEnabled = projectConfig.PerfEnabled
-		p.DeactivatePrevious = projectConfig.DeactivatePrevious
+		p.PerfEnabled = utility.ToBoolPtr(utility.FromBoolPtr(projectConfig.PerfEnabled) || utility.FromBoolPtr(p.PerfEnabled))
+		p.DeactivatePrevious = utility.ToBoolPtr(utility.FromBoolPtr(projectConfig.DeactivatePrevious) || utility.FromBoolPtr(p.DeactivatePrevious))
 	}
 	return err
 }
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -465,8 +465,6 @@ func (p *ProjectRef) MergeWithProjectConfig(version string) error {
 		reflectedRef := reflect.ValueOf(p).Elem()
 		reflectedConfig := reflect.ValueOf(pRefToMerge)
 		recursivelySetUndefinedFields(reflectedRef, reflectedConfig)
-		p.PerfEnabled = utility.ToBoolPtr(utility.FromBoolPtr(projectConfig.PerfEnabled) || utility.FromBoolPtr(p.PerfEnabled))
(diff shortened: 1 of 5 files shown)


model/project_ref.go
```diff
@@ -465,15 +459,14 @@ func (p *ProjectRef) MergeWithProjectConfig(version string) error {
...
-		if projectConfig.CommitQueue != nil {
-			pRefToMerge.CommitQueue = *projectConfig.CommitQueue
-		}
 		if projectConfig.TaskSync != nil {
 			pRefToMerge.TaskSync = *projectConfig.TaskSync
 		}
 		reflectedRef := reflect.ValueOf(p).Elem()
 		reflectedConfig := reflect.ValueOf(pRefToMerge)
 		recursivelySetUndefinedFields(reflectedRef, reflectedConfig)
+		p.PerfEnabled = utility.ToBoolPtr(utility.FromBoolPtr(projectConfig.PerfEnabled) || utility.FromBoolPtr(p.PerfEnabled))
```
ablack12, line 468: Why this change? This is actually incorrect bc if PerfEnabled is _supposed_ to be nil (bc both the project config and the project ref are nil) we'll now incorrectly set it to false explicitly.

operations/model.go
```diff
@@ -82,6 +82,7 @@ type ClientSettings struct {
 	APIKey                string              `json:"api_key" yaml:"api_key,omitempty"`
 	User                  string              `json:"user" yaml:"user,omitempty"`
 	UncommittedChanges    bool                `json:"patch_uncommitted_changes" yaml:"patch_uncommitted_changes,omitempty"`
+	AutoUpgradeCli        bool                `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
```
ablack12, line 85: I think this was mistakenly included in this PR

ablack12 (changes requested):

operations/model.go
hadjri: Yup, thanks!

model/project_ref.go
hadjri: I made this change because I noticed that once a boolean ptr is set, it cannot become nil again, at least with the current way modifying a project ref works on the UI.  In this case if someone toggled PerfEnabled true on the UI, and then later untoggled it, it would be stuck in false rather than nil.  Therefore the project config could never override the project page even though at this point the flag would be "unset" on the project page.  Since having a nil pointer is functionally equivalent to having a false pointer from a code execution perspective since we are using utility.FromBoolPtr I went with this approach.  However open to discussing design alternatives!

ablack12: Hmm I see what you're saying. On the new UI it will be possible to explicitly set the value to false but not yet, and yeah since this is the last step in getting the final merged project ref to be used it would be okay to explicitly set these to true. I'm wondering then if it's maybe just more consistent to also remove these toggles, unless you are attached to them, or else we will need to make modifications to _not_ override an explicitly false value once the new UI is finished.

hadjri: I'm not particularly attached, aliases, periodic builds and build baron configs are definitely the meat here. I'm fine to remove them and perhaps in the future if we get user requests we can add toggles for enabling the performance plugin in version control.

ablack12 (approved):

PR #5382, opened 2022-02-10
EVG-16066: prevent concurrent user-initiated spawn host modifications

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16066

### Description 
The unexpirable spawn host ended up in an unusual state because the user attempted to terminate and stop the host at the same time. The host ended up being terminated in AWS but marked as running in the DB. I changed the jobs that allow users to manually modify spawn hosts so that only one can happen at a time. For example, if the user tries to terminate the host in the UI, any attempt to stop/start the host from the UI will error while it's in progress.

* Use scopes to ensure host start/stop/terminate requests from the user do not cause host concurrency problems when modifying the host.
    * User-initiated spawn host termination now happens asynchronously in a job rather than synchronously with the UI request.
* Clean up a bunch of copy-pasted code across the spawn host start/stop/modify jobs. This is backwards compatible with the existing jobs. Once this is merged and deployed, the deletion TODOs can be done in a follow-up PR.
* Add more useful logs in host start/stop about status updates in failure cases (e.g. failure to stop the host).

### Testing 
Updated unit tests and verified that it works as expected in staging.

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -1167,7 +1167,7 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 			return errors.Wrap(currentStatusErr, "error getting instance status after stopping error")
 		}
 
-		if currentStatusErr = h.SetStatus(currentStatus.String(), user, ""); currentStatusErr != nil {
+		if currentStatusErr = h.SetStatus(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
 			return errors.Wrapf(currentStatusErr, "failed to revert status from stopping to '%s'", currentStatus)
 		}
 		return errors.Wrapf(err, "error stopping EC2 instance '%s'", h.Id)
@@ -1182,10 +1182,11 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 			if err != nil {
 				return false, errors.Wrap(err, "error getting instance info")
 			}
-			if ec2StatusToEvergreenStatus(*instance.State.Name) == StatusStopped {
+			status := ec2StatusToEvergreenStatus(*instance.State.Name)
+			if status == StatusStopped {
 				return false, nil
 			}
-			return true, errors.New("host is not stopped")
+			return true, errors.Errorf("host is not stopped, current status is '%s'", status)
 		}, utility.RetryOptions{
 			MaxAttempts: checkSuccessAttempts,
 			MinDelay:    checkSuccessInitPeriod,
@@ -1196,7 +1197,7 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 		if currentStatusErr != nil {
 			return errors.Wrap(currentStatusErr, "error getting instance status after stopping error")
 		}
-		if currentStatusErr := h.SetStatus(currentStatus.String(), user, ""); currentStatusErr != nil {
+		if currentStatusErr := h.SetStatus(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
 			return errors.Wrapf(currentStatusErr, "failed to revert status from stopping to '%s'", currentStatus)
 		}
 		return errors.Wrap(err, "error checking if spawnhost stopped")
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -1167,7 +1167,7 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 			return errors.Wrap(currentStatusErr, "error getting instance status after stopping error")
 		}
 
-		if currentStatusErr = h.SetStatus(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
+		if currentStatusErr = h.SetStatusAtomically(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
 			return errors.Wrapf(currentStatusErr, "failed to revert status from stopping to '%s'", currentStatus)
 		}
 		return errors.Wrapf(err, "error stopping EC2 instance '%s'", h.Id)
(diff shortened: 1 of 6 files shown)


malikchaya2 (commented):
I had also done this in an attempt to solve this problem. 
https://github.com/evergreen-ci/evergreen/pull/5337/files

Kimchelly:
I think the extra status check on error in #5337 could be removed since it's no longer relevant. #5337 would still leave the host state prone to races because people can press the stop/start/terminate button whenever they want and the AWS status check/update is non-atomic.

malikchaya2:
> I think the extra status check on error in #5337 could be removed since it's no longer relevant. #5337 would still leave the host state prone to races because people can press the stop/start/terminate button whenever they want and the AWS status check/update is non-atomic. This PR basically makes the AWS status changes atomic, so there's no risk of racing between start/stop/terminate requests.

would it make sense to remove it as part of this PR?

Kimchelly:
Since stopping is supposed to be a temporary state, it still makes sense to back out of the stopping state if stopping the host in AWS failed. I changed it to only set the status if the host status has not been changed since it tried to stop the host.

cloud/ec2.go
```diff
@@ -1197,7 +1197,7 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
 		if currentStatusErr != nil {
 			return errors.Wrap(currentStatusErr, "error getting instance status after stopping error")
 		}
-		if currentStatusErr := h.SetStatus(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
+		if currentStatusErr := h.SetStatusAtomically(currentStatus.String(), user, "reverting to the most up-to-date host status"); currentStatusErr != nil {
```
malikchaya2, line 1200: I don't fully understand how this change makes it only set the status if the host status has not been changed since it tried to stop the host. Where does that checking take place?

Kimchelly: `SetStatusAtomically` will only update the host status if the host status in the DB is still "stopping". If it's another state, the host's state has been changed by something other than this thread and attempting to overwrite that state change would be undesirable.

malikchaya2: Ah I see that now. Thank you for explaining it.

malikchaya2 (approved):
This looks fine to me, but I don't know this part of the codebase very well, so another pair of eyes may be a good idea.

units/spawnhost_modify.go
```diff
@@ -11,11 +11,51 @@ import (
...
+	}
+
+	var err error
+	if m.host == nil {
+		m.host, err = host.FindOneByIdOrTag(m.HostID)
+		if err != nil {
+			return errors.Wrap(err, "finding host")
+		}
+		if m.host == nil {
+			return errors.Wrap(err, "host not found")
```
ablack12, line 43: can't wrap error here

units/spawnhost_modify.go
```diff
@@ -26,12 +66,13 @@ func init() {
 }
 
 type spawnhostModifyJob struct {
-	HostID        string                 `bson:"host_id" json:"host_id" yaml:"host_id"`
-	ModifyOptions host.HostModifyOptions `bson:"modify_options" json:"modify_options" yaml:"modify_options"`
-	job.Base      `bson:"job_base" json:"job_base" yaml:"job_base"`
+	// TODO (EVG-16066): remove HostID.
+	HostID                string                 `bson:"host_id" json:"host_id" yaml:"host_id"`
+	ModifyOptions         host.HostModifyOptions `bson:"modify_options" json:"modify_options" yaml:"modify_options"`
+	CloudHostModification `bson:"base_spawn_host_modification" json:"base_spawn_host_modification" yaml:"base_spawn_host_modification"`
```
ablack12, line 72: Preference for having the bson/json tags match the variable name

units/spawnhost_modify.go
```diff
@@ -49,56 +90,32 @@ func makeSpawnhostModifyJob() *spawnhostModifyJob {
...
+		j.CloudHostModification.HostID = j.HostID
 	}
 
-	var err error
-	if j.host == nil {
-		j.host, err = host.FindOneByIdOrTag(j.HostID)
-		if err != nil {
-			j.AddError(err)
-			return
+	if err := j.CloudHostModification.modifyHost(ctx, func(mgr cloud.Manager, h *host.Host, user string) error {
```
ablack12, line 109: this if statement is a bit unclear to read; can we set the function as a variable instead and then just call it here?

units/spawnhost_start.go
```diff
@@ -44,52 +47,42 @@ func makeSpawnhostStartJob() *spawnhostStartJob {
 	return j
 }
 
+// NewSpawnhostStartJob returns a job to start a stopped spawn host.
 func NewSpawnhostStartJob(h *host.Host, user, ts string) amboy.Job {
 	j := makeSpawnhostStartJob()
 	j.SetID(fmt.Sprintf("%s.%s.%s.%s", spawnhostStartName, user, h.Id, ts))
-	j.HostID = h.Id
-	j.UserID = user
+	j.SetScopes([]string{fmt.Sprintf("%s.%s", spawnHostStatusChangeScopeName, h.Id)})
```
ablack12, line 54: Why do we set scopes here but not in the spawnhostModifyJob?

ablack12 (changes requested):

units/spawnhost_modify.go
Kimchelly: Fixed.

units/spawnhost_modify.go
Kimchelly: Ah yeah, nice catch. This is what I get for waffling on naming things too many times. Fixed.

units/spawnhost_modify.go
Kimchelly: Sure, done. I changed the other places that do this as well.

units/spawnhost_start.go
Kimchelly: Start/stop/terminate are mutually exclusive states, so I wanted to prevent races between them with the scope. Modifying the spawn host metadata (i.e. the spawn host modify job) could also fail if the host's state changed, but it seemed less important to put a scope on it.

I also waffled on setting scopes for the spawn host modify job because it would prevent people from modifying their spawn host's metadata multiple times in quick succession (e.g. add a tag, save the page, extend the expiration, save the page again) and would prevent users from doing other operations like terminating the host until the modification attempt was finished. My reasoning was that if you were modifying the host metadata, it would be important to still allow other overwriting operations like stopping/terminating the host, which would make the host modification irrelevant. If the modification fails because the user changed the host state, it'll show up in the event log, so I figured it was okay to allow multiple modification jobs regardless of the host state.

ablack12 (approved):

PR #5383, opened 2022-02-10
EVG-6496 Auto update CLI if it is set in client settings before each command

hadjri:

### Description 
New property has been added in client settings to optionally automatically update the evergreen CLI before any command is run.

### Testing 
Built CLI locally with new client settings to auto update and tested manually.

### DIFF
--- a/cmd/evergreen/evergreen.go
+++ b/cmd/evergreen/evergreen.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -15,6 +16,7 @@ import (
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/level"
 	"github.com/mongodb/grip/send"
+	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
 
@@ -100,6 +102,19 @@ func buildApp() *cli.App {
 	}
 
 	app.Before = func(c *cli.Context) error {
+		confPath := c.String("conf")
+		conf, err := operations.NewClientSettings(confPath)
+		if err != nil {
+			return errors.Wrap(err, "problem loading configuration")
+		}
+		if conf.AutoUpgradeCli {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			err = operations.DoUpdate(conf, ctx, true, false)
+			if err != nil {
+				return errors.Wrap(err, "problem auto updating CLI")
+			}
+		}
 		return loggingSetup(app.Name, c.String("level"))
 	}
 
--- a/cmd/evergreen/evergreen.go
+++ b/cmd/evergreen/evergreen.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -107,12 +108,11 @@ func buildApp() *cli.App {
 		if err != nil {
 			return errors.Wrap(err, "problem loading configuration")
 		}
-		if conf.AutoUpgradeCli {
+		if conf.AutoUpgradeCLI {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-			err = operations.DoUpdate(conf, ctx, true, false)
-			if err != nil {
-				return errors.Wrap(err, "problem auto updating CLI")
+			if err = operations.CheckAndUpdateVersion(conf, ctx, true, false); err != nil {
+				fmt.Println("Automatic CLI update failed, continuing command execution")
 			}
 		}
 		return loggingSetup(app.Name, c.String("level"))
--- a/cmd/evergreen/evergreen.go
+++ b/cmd/evergreen/evergreen.go
@@ -111,7 +111,7 @@ func buildApp() *cli.App {
 		if conf.AutoUpgradeCLI {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-			if err = operations.CheckAndUpdateVersion(conf, ctx, true, false); err != nil {
+			if err = operations.CheckAndUpdateVersion(conf, ctx, true, false, true); err != nil {
 				fmt.Println("Automatic CLI update failed, continuing command execution")
 			}
 		}
--- a/operations/model.go
+++ b/operations/model.go
@@ -82,7 +82,7 @@ type ClientSettings struct {
 	APIKey                string              `json:"api_key" yaml:"api_key,omitempty"`
 	User                  string              `json:"user" yaml:"user,omitempty"`
 	UncommittedChanges    bool                `json:"patch_uncommitted_changes" yaml:"patch_uncommitted_changes,omitempty"`
-	AutoUpgradeCli        bool                `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
+	AutoUpgradeCLI        bool                `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
 	PreserveCommits       bool                `json:"preserve_commits" yaml:"preserve_commits,omitempty"`
 	Projects              []ClientProjectConf `json:"projects" yaml:"projects,omitempty"`
 	Admin                 ClientAdminConf     `json:"admin" yaml:"admin,omitempty"`
@@ -148,7 +148,7 @@ func (s *ClientSettings) Write(fn string) error {
 // To avoid printing these messages, call getRestCommunicator instead.
 func (s *ClientSettings) setupRestCommunicator(ctx context.Context) client.Communicator {
 	c := s.getRestCommunicator(ctx)
-	printUserMessages(ctx, c, !s.AutoUpgradeCli)
+	printUserMessages(ctx, c, !s.AutoUpgradeCLI)
 	return c
 }
 
@@ -369,4 +369,9 @@ func (s *ClientSettings) SetDefaultProject(cwd, project string) {
 	}
 	s.ProjectsForDirectory[cwd] = project
 	grip.Infof("Project '%s' will be set as the one to use for directory '%s'. To disable automatic defaulting, set 'disable_auto_defaulting' to true.", project, cwd)
+}
+
+func (s *ClientSettings) SetAutoUpgradeCLI() {
+	s.AutoUpgradeCLI = true
+	grip.Info("Evergreen CLI will be automatically updated and installed before each command if a more recent version is detected.")
 }
--- a/operations/model.go
+++ b/operations/model.go
@@ -82,6 +82,7 @@ type ClientSettings struct {
 	APIKey                string              `json:"api_key" yaml:"api_key,omitempty"`
 	User                  string              `json:"user" yaml:"user,omitempty"`
 	UncommittedChanges    bool                `json:"patch_uncommitted_changes" yaml:"patch_uncommitted_changes,omitempty"`
+	AutoUpgradeCli        bool                `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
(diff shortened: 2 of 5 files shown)


operations/update.go
```diff
@@ -53,96 +53,99 @@ func Update() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem loading configuration")
 			}
+			return DoUpdate(conf, ctx, doInstall, forceUpdate)
```
Kimchelly, line 56: Should users have an option to set the auto-upgrade config setting when calling `evergreen get-update` rather than manually editing their YAML?

operations/model.go
```diff
@@ -82,6 +82,7 @@ type ClientSettings struct {
 	APIKey                string              `json:"api_key" yaml:"api_key,omitempty"`
 	User                  string              `json:"user" yaml:"user,omitempty"`
 	UncommittedChanges    bool                `json:"patch_uncommitted_changes" yaml:"patch_uncommitted_changes,omitempty"`
+	AutoUpgradeCli        bool                `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
```
Kimchelly, line 85: Can this be AutoUpgradeCLI? This is [the recommended style](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) for acronyms.

cmd/evergreen/evergreen.go
```diff
@@ -100,6 +102,19 @@ func buildApp() *cli.App {
...
 	app.Before = func(c *cli.Context) error {
+		confPath := c.String("conf")
+		conf, err := operations.NewClientSettings(confPath)
+		if err != nil {
+			return errors.Wrap(err, "problem loading configuration")
+		}
+		if conf.AutoUpgradeCli {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			err = operations.DoUpdate(conf, ctx, true, false)
```
Kimchelly, line 113: L113-115 could be simplified down to `return errors.Wrap(err, "problem auto-updating CLI")`.

cmd/evergreen/evergreen.go
```diff
@@ -100,6 +102,19 @@ func buildApp() *cli.App {
...
+			if err != nil {
+				return errors.Wrap(err, "problem auto updating CLI")
```
Kimchelly, line 115: Should this log and continue on error to perform the user's actual CLI command, even if auto-updating fails?

operations/update.go
```diff
@@ -53,96 +53,99 @@ func Update() cli.Command {
...
+		},
+	}
+}
 
-			client := conf.getRestCommunicator(ctx)
-			defer client.Close()
+func DoUpdate(conf *ClientSettings, ctx context.Context, doInstall bool, forceUpdate bool) error {
```
Kimchelly, line 61: Documentation.

I would rename this to `CheckAndUpdateVersion`.

Kimchelly (changes requested):

operations/update.go
hadjri: Sure, I like the sound of that - going to add

operations/update.go
```diff
@@ -53,12 +53,22 @@ func Update() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem loading configuration")
 			}
-			return DoUpdate(conf, ctx, doInstall, forceUpdate)
+			if conf.AutoUpgradeCLI {
+				fmt.Println("Automatic CLI updates are already run before each command, get-update command is not necessary")
+				return nil
+			}
+			if !conf.AutoUpgradeCLI && confirm(fmt.Sprint("Automatic CLI updates are not set up, configure now? (Y/n):"), false) {
```
Kimchelly, line 60: Lowercase "y" since "Y" conventionally means "yes to multiple prompts" in a CLI, whereas this is a single prompt.

Rather than make this a prompt, can it be a boolean flag on the command instead? Seems like it would be a hassle to always respond "n" if a user was prompted every time if they want to set up auto-upgrading when they're trying to manually upgrade.

operations/update.go
```diff
@@ -53,12 +53,22 @@ func Update() cli.Command {
...
+				conf.SetAutoUpgradeCLI()
+				if err := conf.Write(""); err != nil {
+					return err
```
Kimchelly, line 63: Adding context about what operation failed is helpful to users:
`return errors.Wrap(err, "setting auto-upgrade CLI option")`

operations/update.go
```diff
@@ -53,12 +53,22 @@ func Update() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem loading configuration")
 			}
-			return DoUpdate(conf, ctx, doInstall, forceUpdate)
+			if conf.AutoUpgradeCLI {
```
Kimchelly, line 56: I don't think this block should be necessary since `CheckAndUpdateVersion` already checks if the command is really up-to-date or not. There would be some benefit in being told that they have the most up-to-date version even if auto-upgrading is enabled. In this case, I would let `evergreen get-update` say the "Binary is already up to date - not updating." message to let them explicitly know they're on the newest version.

cmd/evergreen/evergreen.go
```diff
@@ -107,12 +108,11 @@ func buildApp() *cli.App {
...
 			return errors.Wrap(err, "problem loading configuration")
 		}
-		if conf.AutoUpgradeCli {
+		if conf.AutoUpgradeCLI {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
-			err = operations.DoUpdate(conf, ctx, true, false)
-			if err != nil {
-				return errors.Wrap(err, "problem auto updating CLI")
+			if err = operations.CheckAndUpdateVersion(conf, ctx, true, false); err != nil {
```
Kimchelly, line 114: One concern I have is that if this runs before every command, it'll always say "Binary is already up to date - not updating." because of `checkUpdate` at the beginning. This might be undesirable if people have things like scripts that use the CLI and don't want this extra output. Would it be possible to propagate `silent = true` to `checkUpdate` when auto-upgrades are enabled so that people only see the update logs when there's actually a new update available?

Kimchelly (changes requested):

operations/update.go
Kimchelly: I would leave a doc comment for `CheckAndUpdateVersion` since it's recommended to document exported Go methods.

cmd/evergreen/evergreen.go
hadjri: That makes sense, yes

bynn (approved):

operations/update.go
Kimchelly: I still have this open thread about documenting the function.

operations/update.go
```diff
@@ -39,12 +40,17 @@ func Update() cli.Command {
 				Name:  joinFlagNames(forceFlagName, "f"),
 				Usage: "download a new CLI even if the current CLI is not out of date",
 			},
+			cli.BoolFlag{
+				Name:  joinFlagNames(autoUpdateFlagName, "a"),
```
Kimchelly, line 44: Since the config option is called "AutoUpgradeCLI", can this also be "autoUpgrade" for consistency?

Kimchelly (approved):
LGTM mod two small comments.

operations/update.go
```diff
@@ -53,96 +59,109 @@ func Update() cli.Command {
...
-			defer client.Close()
-
-			update, err := checkUpdate(client, false, forceUpdate)
-			if err != nil {
-				return err
-			}
-			if !update.needsUpdate || len(update.binaries) == 0 {
-				return nil
+			if conf.AutoUpgradeCLI && autoUpdate {
+				fmt.Printf("automatic CLI updates are already set up, %s flag is not necessary", autoUpdateFlagName)
```
ablack12, line 63: Since this prints to the user in the CLI, can this please be a complete sentence? (same with line 70)

operations/update.go
```diff
@@ -39,12 +40,17 @@ func Update() cli.Command {
...
+				Name:  joinFlagNames(autoUpdateFlagName, "a"),
+				Usage: "setup automatic installations of a new CLI if the current CLI is out of date",
+			},
 		},
 		Before: setPlainLogger,
 		Action: func(c *cli.Context) error {
 			confPath := c.Parent().String(confFlagName)
 			doInstall := c.Bool(installFlagName)
 			forceUpdate := c.Bool(forceFlagName)
+			autoUpdate := c.Bool(autoUpdateFlagName)
```
ablack12, line 53: If autoUpdate isn't setup, should we print to the user that -a is an option so that they know about it? (Most people will never look at this help page)

operations/update.go
```diff
@@ -59,21 +59,28 @@ func Update() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem loading configuration")
 			}
-			if conf.AutoUpgradeCLI && autoUpdate {
-				fmt.Printf("automatic CLI updates are already set up, %s flag is not necessary", autoUpdateFlagName)
+			if !conf.AutoUpgradeCLI && !autoUpgrade {
+				fmt.Printf("Automatic CLI upgrades are not set up, specifying the -%s (-a for short) flag will enable automatic CLI upgrades before each command if the current CLI is out of date.", autoUpgradeFlagName)
```
ablack12, line 63: some nits: I would remove the -a for short; since it's something they only need to do once a shorthand isn't even super necessary (I would even argue against using it at all in case we want to use -a for something more common in the future). Also can you use a semicolon before "specifying" since that's the start of a new sentence?

operations/update.go
```diff
@@ -25,7 +25,7 @@ import (
 func Update() cli.Command {
 	const installFlagName = "install"
 	const forceFlagName = "force"
-	const autoUpdateFlagName = "auto_update"
+	const autoUpgradeFlagName = "auto_update"
```
ablack12, line 28: I would have this value be just "auto" (or at least "auto-upgrade" for consistency with wording and dashes)

operations/update.go
```diff
@@ -59,21 +59,28 @@ func Update() cli.Command {
...
 				}
-				fmt.Println("automatic CLI updates have successfully been setup")
+				fmt.Println("Automatic CLI upgrades have successfully been setup.")
 			}
 			return CheckAndUpdateVersion(conf, ctx, doInstall, forceUpdate, false)
 		},
 	}
 }
 
+// CheckAndUpdateVersion checks if the CLI is up to date, if there is no new CLI version it will simply notify that the current binary is up to date.
```
ablack12, line 80: nit: "if there is no new CLI version" is the start of a new sentence

operations/update.go
```diff
@@ -59,21 +59,28 @@ func Update() cli.Command {
...
+// If there is a new version available, it will be downloaded, and if doInstall is set the new binary will automatically be replaced at the current path the old binary existed in.
+// Otherwise, it will simply be downloaded and a suggested 'mv' command will be printed so the user can replace the binary at their discretion.
+// Toggling forceUpdate will download a new CLI even if the current CLI is not out of date.
```
ablack12, line 83: maybe "does not have an out-of-date CLI version string" is more clear since that's typically the use case of forcing the update (i.e. someone didn't increment the string or we did a revert).

operations/update.go
hadjri: True, I'll just remove -a completely.

operations/update.go
```diff
@@ -41,7 +41,7 @@ func Update() cli.Command {
 				Usage: "download a new CLI even if the current CLI is not out of date",
 			},
 			cli.BoolFlag{
-				Name:  joinFlagNames(autoUpgradeFlagName, "a"),
+				Name:  joinFlagNames(autoUpgradeFlagName),
```
ablack12, line 44: nit: remove joinFlagNames now that there's just the one

ablack12 (approved):

PR #5384, opened 2022-02-10
EVG-16175: (partial) addition of new task status and concomitant changes for model and service

hhoke:
[EVG-16175](https://jira.mongodb.org/browse/EVG-16175)

### Description 
I added a TaskContainerAllocated status to differentiate between container tasks that were ready to run but waiting for a container (analogous to TaskDispatched), and that are ready to run with the container allocated. TaskContainerUnallocated is the container equivalent of TaskUndispatched.

### Testing 
Standard patch test

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -591,7 +591,7 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string, 
 			tc.logger.Task().Error(errors.Wrap(err, "error running post task commands"))
 		}
 		a.runEndTaskSync(ctx, tc, detail)
-	// TODO: (PM-26160) make sure that TaskContainerUnallocated means the task has been aborted.
+	// TODO: (PM-2616) make sure that TaskContainerUnallocated means the task has been aborted.
 	case evergreen.TaskUndispatched, evergreen.TaskContainerUnallocated:
 		tc.logger.Task().Info("Task completed - ABORTED.")
 	case evergreen.TaskConflict:
--- a/model/lifecycle_test.go
+++ b/model/lifecycle_test.go
@@ -222,7 +222,6 @@ func TestBuildRestart(t *testing.T) {
 				Status:      evergreen.TaskDispatched,
 				Activated:   true,
 			}
-			// TODO: abortion doesn't apply for container tasks, so I think this should not be modified
 			So(taskTwo.Insert(), ShouldBeNil)
 
 			So(RestartBuild(b.Id, []string{"task1", "task2"}, true, evergreen.DefaultTaskActivator), ShouldBeNil)
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -217,7 +217,7 @@ var (
 					},
 					"then": evergreen.TaskTimedOut,
 				},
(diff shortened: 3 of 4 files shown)


model/lifecycle_test.go
```diff
@@ -222,6 +222,7 @@ func TestBuildRestart(t *testing.T) {
 				Status:      evergreen.TaskDispatched,
 				Activated:   true,
 			}
+			// TODO: abortion doesn't apply for container tasks, so I think this should not be modified
```
Kimchelly, line 225: Container tasks can still be aborted. What does this comment mean?

Kimchelly (changes requested):

model/task/db.go
```diff
@@ -217,6 +217,18 @@ var (
 					},
 					"then": evergreen.TaskTimedOut,
 				},
+				{
```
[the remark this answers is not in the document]
hhoke: In addition to what's here? Or instead of it?

model/lifecycle_test.go
hhoke: I was thinking we might have to make some edits to `RestartVersion` and, similarly, `restartTasksForBuild`, but I think this will be covered by the EVG-16305 build.

agent/agent.go
```diff
@@ -591,7 +591,7 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 			tc.logger.Task().Error(errors.Wrap(err, "error running post task commands"))
 		}
 		a.runEndTaskSync(ctx, tc, detail)
-	case evergreen.TaskUndispatched:
+	case evergreen.TaskUndispatched, evergreen.TaskContainerUnallocated:
```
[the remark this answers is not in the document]
Kimchelly: Sorry, PM-2616. Spelling is hard.

model/lifecycle_test.go
Kimchelly: I don't think this comment should be necessary then, since this test will presumably be updated in the EVG-16305 PR to deal with restarting/resetting tasks in a build/version. I would just remove it and assume the future PR will do the due diligence to test the change.

model/task/db.go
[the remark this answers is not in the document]
Kimchelly: In addition. The UI won't need to display these statuses anyways since nobody's using it for now.

model/task/db.go
```diff
@@ -217,6 +217,19 @@ var (
 					},
 					"then": evergreen.TaskTimedOut,
 				},
+				//TODO (PM-2620): Handle these statuses properly in the UI
```
Kimchelly, line 220: nit: space between comment delimiter `//` and `TODO` is the preferred style.

model/task/task.go
```diff
@@ -1607,6 +1609,7 @@ func ResetTasks(taskIds []string) error {
 	return err
 }
 
+//TODO: (EVG-16305): create container version of this
```
Kimchelly, line 1612: I don't think this comment is necessary since changing `ResetTasks` is already part of the ticket.

Kimchelly (approved):

PR #5387, opened 2022-02-14
EVG-16315 Fix build variants logging as pushing when being skipped

malikchaya2:
[EVG-16315](https://jira.mongodb.org/browse/EVG-16315)

### Description 
Because we logged a push event before checking if the build variant will run, in the case where a build variant was skipped, by the time it got to the build variant that was truly trying to push the file, it would no-op, because the db thought that the task already pushed. 

### Testing 
None right now, but will coordinate with the right people to get this tested once it's deployed.

### DIFF
--- a/agent/command/s3_copy.go
+++ b/agent/command/s3_copy.go
@@ -204,6 +204,15 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
 	for _, s3CopyFile := range c.S3CopyFiles {
 		timer.Reset(0)
 
+		if len(s3CopyFile.BuildVariants) > 0 && !utility.StringSliceContains(
+			s3CopyFile.BuildVariants, conf.BuildVariant.Name) {
+			continue
+		}
+
+		if ctx.Err() != nil {
+			return errors.New("s3copy operation received was canceled")
+		}
+
 		s3CopyReq := apimodels.S3CopyRequest{
 			S3SourceRegion:      s3CopyFile.Source.Region,
 			S3SourceBucket:      s3CopyFile.Source.Bucket,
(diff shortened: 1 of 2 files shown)


agent/command/s3_copy.go
```diff
@@ -223,20 +232,6 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
...
-		if len(s3CopyFile.BuildVariants) > 0 && !utility.StringSliceContains(
-			s3CopyFile.BuildVariants, conf.BuildVariant.Name) {
-			continue
-		}
-
-		if ctx.Err() != nil {
-			return errors.New("s3copy operation received was canceled")
-		}
-
-		logger.Execution().Infof("Making API push copy call to "+
```
ablack12, line 235: I kind of think it still makes sense to log here since we've now officially "taken the lock" on the push log but whatever makes most sense for your debugging

agent/command/s3_copy.go
```diff
@@ -204,6 +204,15 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
 	for _, s3CopyFile := range c.S3CopyFiles {
 		timer.Reset(0)
 
+		if len(s3CopyFile.BuildVariants) > 0 && !utility.StringSliceContains(
+			s3CopyFile.BuildVariants, conf.BuildVariant.Name) {
+			continue
+		}
+
+		if ctx.Err() != nil {
+			return errors.New("s3copy operation received was canceled")
```
ablack12, line 213: I know this was just a copy paste but can we remove "received" so that this sentence makes sense?

ablack12 (approved):

agent/command/s3_copy.go
```diff
@@ -223,20 +232,6 @@ func (c *s3copy) copyWithRetry(ctx context.Context,
...
-
-		if ctx.Err() != nil {
-			return errors.New("s3copy operation received was canceled")
-		}
-
-		logger.Execution().Infof("Making API push copy call to "+
-			"transfer %v/%v => %v/%v", s3CopyFile.Source.Bucket,
-			s3CopyFile.Source.Path, s3CopyFile.Destination.Bucket,
-			s3CopyFile.Destination.Path)
-
```
ablack12, line 239: Should we be updating push status to failed for each of the error return statements after this? (or maybe handling in a defer?) It looks like we would leave it as "pushing" if we exited anytime before the retry loop

malikchaya2: good point, will move.

ablack12 (approved):

PR #5391, opened 2022-02-14
EVG-16325 Modify validation route to accept separate project parser and project config yamls

hadjri:
[EVG-16325](https://jira.mongodb.org/browse/EVG-16325)

### Description 
Validation route now accepts a payload with a field dedicated for a parser project, and one for a project config.
### Testing 
Modified unit tests, and tested in staging.


service/api.go
```diff
@@ -637,7 +630,12 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		ReadFileFrom: model.ReadFromLocal,
 	}
 	validationErr := validator.ValidationError{}
-	if _, projectConfig, err = model.LoadProjectInto(ctx, input.ProjectYaml, opts, "", project); err != nil {
+	if _, _, err = model.LoadProjectInto(ctx, validationConfigs.ParserProjectYaml, opts, "", project); err != nil {
+		validationErr.Message = err.Error()
+		gimlet.WriteJSONError(w, validator.ValidationErrors{validationErr})
+		return
+	}
+	if _, projectConfig, err = model.LoadProjectInto(ctx, validationConfigs.ProjectConfigYaml, opts, "", project); err != nil {
```
ablack12, line 638: Can we explicitly just call CreateProjectConfig here instead of going all the way through loadProjectInto again? And can we add a comment about why we need to do this and why we can't just use the one  returned from LoadProjectInto?

validator/project_validator.go
```diff
@@ -108,10 +108,10 @@ func (v ValidationErrors) HasError() bool {
 }
 
 type ValidationInput struct {
-	ProjectYaml []byte `json:"project_yaml" yaml:"project_yaml"`
-	Quiet       bool   `json:"quiet" yaml:"quiet"`
-	IncludeLong bool   `json:"include_long" yaml:"include_long"`
-	ProjectID   string `json:"project_id" yaml:"project_id"`
+	ValidationConfigs model.ValidationConfigs `json:"validation_configs" yaml:"validation_configs"`
```
ablack12, line 111: Changing this means that any existing yaml that didn't update is going to error. We should check to see if ValidationConfigs is empty, and if so return an error asking users to update.

service/api.go
```diff
@@ -616,19 +615,13 @@ func (as *APIServer) listVariants(w http.ResponseWriter, r *http.Request) {
 func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Request) {
 	body := utility.NewRequestReader(r)
 	defer body.Close()
-
-	bytes, err := ioutil.ReadAll(body)
+	input := validator.ValidationInput{}
+	err := json.NewDecoder(body).Decode(&input)
 	if err != nil {
-		gimlet.WriteJSONError(w, fmt.Sprintf("Error reading request body: %v", err))
+		gimlet.WriteJSONError(w, fmt.Sprintf("Error while unmarshalling JSON body: %v", err))
```
ablack12, line 621: Users are going to hit this if they're somehow still using the legacy structure referenced at 628. We should make this error more clear that they need to update their CLI to continue.

ablack12 (changes requested):
Can you also try testing in staging with a non-updated CLI to verify that you get helpful errors back regarding the incorrectly formatted structure?

service/api.go
hadjri: I've actually gone ahead and decided against this idea - I've changed the way we unmarshal and remarshal the project config raw bytes in a way that will allow us to still just send one byte array. CLI updating won't be needed.

model/project_configs.go
```diff
@@ -13,10 +13,9 @@ import (
 )
 
 type ProjectConfig struct {
-	Id         string    `yaml:"_id" bson:"_id"`
-	Project    string    `yaml:"project,omitempty" bson:"project,omitempty"`
-	CreateTime time.Time `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
-
+	Id               string    `yaml:"_id" bson:"_id"`
+	Project          string    `yaml:"project,omitempty" bson:"project,omitempty"`
+	ConfigCreateTime time.Time `yaml:"config_create_time,omitempty" bson:"config_create_time,omitempty"`
```
ablack12, line 18: What's the purpose for renaming this? (Or having it at all?)

model/project_configs.go
```diff
@@ -33,6 +32,22 @@ type ProjectConfig struct {
 	PeriodicBuilds         []PeriodicBuildDefinition      `yaml:"periodic_builds,omitempty" bson:"periodic_builds,omitempty"`
 }
 
+type HeadlessProjectConfig struct {
```
ablack12, line 35: Is the reason we need this because ID is a duplicated field across ProjectConfig and ParserProject or something and so it doesn't work to have ProjectConfig within the struct used for validating strictly?

I'm pretty against having this struct. It's a lot of code duplication for something that's only used in one place, and "Headless" isn't very informative (and is a bit morbid haha).  Alternatives could be:

* Have a ProjectConfig structure called ProjectConfigFields (or something similar); have this called inline for project config. Add comments about why we're doing this. Instead of the GetHeadless / GetProject functions, could just return ProjectConfig.ProjectConfigFields or whatever.
* Just add each ProjectConfig field that we need to validate into the unmarshalWithStrict field directly. Maybe add a comment to ProjectConfig mentioning to add these there. (Similar amount of code duplication but now it's restricted to the code that needs it.)

ablack12 (changes requested):

model/project_configs.go
hadjri: It conflicts with the same field that exists in Parser Project during strict validation.

model/project_configs.go
hadjri: I had initially gone with your first solution but backed out due to the amount of refactoring in other files / tests that was required, but I get where you're coming from. I've added it back and it's actually cleaner than I expected. Also removed the "Headless" naming in favor of 'ProjectConfigFields'. Honestly that was just me being a little cheeky and deep down I knew that name wouldn't fly haha.

model/project_configs.go
```diff
@@ -13,13 +13,13 @@ import (
 )
 
 type ProjectConfig struct {
-	Id         string    `yaml:"_id" bson:"_id"`
-	Project    string    `yaml:"project,omitempty" bson:"project,omitempty"`
-	CreateTime time.Time `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
+	Id                  string `yaml:"_id" bson:"_id"`
+	ProjectConfigFields `yaml:",inline" bson:",inline"`
```
ablack12, line 17: Can we add a comment here to explain why we separated this?

model/project_configs.go
```diff
@@ -13,13 +13,13 @@ import (
...
+}
 
-	// These fields can be set for the ProjectRef struct on the project page, or in the project config yaml.
```
ablack12, line 20: Shouldn't we still have a comment like this?

model/project_configs.go
```diff
@@ -13,13 +13,13 @@ import (
...
-	// Values for the below fields set on the project page will take precedence over this struct and will
-	// be the configs used for a given project during runtime.
+type ProjectConfigFields struct {
+	Project                string                         `yaml:"project,omitempty" bson:"project,omitempty"`
+	ConfigCreateTime       time.Time                      `yaml:"config_create_time,omitempty" bson:"config_create_time,omitempty"`
```
ablack12, line 22: Can this be moved out of ProjectConfigFields so that config can be removed from the title?

model/project_parser.go
```diff
@@ -800,32 +796,45 @@ func GetProjectFromFile(ctx context.Context, opts GetProjectOpts) (ProjectInfo,
...
 			Variables interface{} `yaml:"variables,omitempty" bson:"-"`
 		}{}
 
 		if err := util.UnmarshalYAMLStrictWithFallback(yml, &strictProjectWithVariables); err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 		p = strictProjectWithVariables.ParserProject
+		c = ProjectConfig{
+			"",
```
ablack12, line 816: We don't define things like this without specifying the field; can you just set ` c = ProjectConfig{ProjectConfigFields: strictProjectWithVariables.ProjectConfigFields}` ?

operations/validate.go
```diff
@@ -132,6 +132,13 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
 		return errors.Wrapf(err, "Could not marshal parser project into yaml")
 	}
 
+	if pc != nil {
+		projectConfigYaml, err := yaml.Marshal(pc.ProjectConfigFields)
+		if err != nil {
+			return errors.Wrapf(err, "Could not marshal project config into yaml")
+		}
+		projectYaml = []byte(strings.Join([]string{string(projectYaml), string(projectConfigYaml)}, "\n"))
```
ablack12, line 140: We could use fmt.Sprintf to join these instead to shorten this up

operations/validate.go
```diff
@@ -151,10 +158,9 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
 
 // loadProjectIntoWithValidation returns a warning (instead of an error) if there's an error with unmarshalling strictly
 func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model.GetProjectOpts,
-	project *model.Project) (*model.ParserProject, validator.ValidationErrors) {
+	project *model.Project) (*model.ParserProject, *model.ProjectConfig, validator.ValidationErrors) {
 	errs := validator.ValidationErrors{}
-	pp, _, err := model.LoadProjectInto(ctx, data, opts, "", project)
-
+	pp, pc, err := model.LoadProjectInto(ctx, data, opts, "", project)
```
ablack12, line 163: Do we only really need LoadProjectInto to return projectConfig for this function? In which case can we just separately unmarshal data into ProjectConfig as part of this function (using YamlWithFallback since any strict errors will still be caught on this line)? We'd still need to add ProjectConfig to the UnmarshalStrict struct but we wouldn't need to return the unmarshalled fields.

ablack12 (changes requested):

model/project_configs.go
hadjri: Yes, was accidentally removed when I changed ProjectConfig to include the fields inline. Added back

operations/validate.go
hadjri: Ok, I'll remove the project config unmarshalling from createIntermediateProject to decrease the amount of places we perform unmarshalling

hadjri: While I was at it I did the same for LoadProjectInto since a lot of places that call it don't care about ProjectConfig

bynn (approved):

model/project_configs.go
```diff
@@ -14,12 +14,17 @@ import (
...
+	// ProjectConfigFields are the properties on the project config that do not duplicate parser project's fields to allow strict unmarshalling of a full config file.
+	// Since a config file gets split into ParserProject and ProjectConfig, strict unmarshalling does not work when duplicate fields exist (e.g. Id, CreateTime).
+	ProjectConfigFields `yaml:",inline" bson:",inline"`
+}
 
+type ProjectConfigFields struct {
 	// These fields can be set for the ProjectRef struct on the project page, or in the project config yaml.
 	// Values for the below fields set on the project page will take precedence over this struct and will
 	// be the configs used for a given project during runtime.
+	Project                string                         `yaml:"project,omitempty" bson:"project,omitempty"`
```
ablack12, line 27: I think Project doesn't belong here since it's not a user configured field but instead an Evergreen populated field; however I don't think we need this at all -- do we use it for anything? Might be worth removing this field.

operations/validate.go
```diff
@@ -132,6 +132,14 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
 		return errors.Wrapf(err, "Could not marshal parser project into yaml")
 	}
 
+	if pc != nil {
+		projectConfigYaml, err := yaml.Marshal(pc.ProjectConfigFields)
+		if err != nil {
+			return errors.Wrapf(err, "Could not marshal project config into yaml")
+		}
+		projectYamlString := fmt.Sprintf("%s\n%s", string(projectYaml), string(projectConfigYaml))
```
ablack12, line 140: sorry i know i recommended this but I wondder if bytes.Join might be the better way to do this
https://pkg.go.dev/bytes#Join

rest/data/mock_impl.go
```diff
@@ -72,11 +72,11 @@ tasks:
 		Token:        token,
 		ReadFileFrom: model.ReadFromLocal,
 	}
-	pp, pConfig, err := model.LoadProjectInto(ctx, []byte(config), opts, pRef.Id, p)
+	pp, err := model.LoadProjectInto(ctx, []byte(config), opts, pRef.Id, p)
 	return model.ProjectInfo{
 		Project:             p,
 		IntermediateProject: pp,
-		Config:              pConfig,
```
ablack12, line 79: Why not just make this nil?

rest/route/version_create.go
```diff
@@ -64,13 +64,20 @@ func (h *versionCreateHandler) Run(ctx context.Context) gimlet.Responder {
...
 	}
-	projectInfo.IntermediateProject, projectInfo.Config, err = model.LoadProjectInto(ctx, h.Config, opts, projectInfo.Ref.Id, p)
+	projectInfo.IntermediateProject, err = model.LoadProjectInto(ctx, h.Config, opts, projectInfo.Ref.Id, p)
 	if err != nil {
 		return gimlet.NewJSONErrorResponse(gimlet.ErrorResponse{
 			StatusCode: http.StatusBadRequest,
 			Message:    errors.Wrap(err, "unable to unmarshal yaml config").Error(),
 		})
 	}
+	projectInfo.Config, err = model.CreateProjectConfig(h.Config, projectInfo.Ref.Id)
```
ablack12, line 74: Should we just do this if projectInfo.Ref.VersionControlEnabled?

units/periodic_builds.go
```diff
@@ -168,10 +168,14 @@ func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.Peri
 		Token:        token,
 		ReadFileFrom: model.ReadfromGithub,
 	}
-	intermediateProject, config, err := model.LoadProjectInto(ctx, configBytes, opts, j.project.Id, proj)
+	intermediateProject, err := model.LoadProjectInto(ctx, configBytes, opts, j.project.Id, proj)
 	if err != nil {
 		return "", errors.Wrap(err, "error parsing config file")
 	}
+	config, err := model.CreateProjectConfig(configBytes, j.project.Id)
```
ablack12, line 175: Should we just do this if VersionControlEnabled?

rest/data/mock_impl.go
hadjri: Sure

model/project_configs.go
hadjri: I can take it out of the ProjectConfigFields struct but I think we still do need it - it actually acts as an "identifier" field for ProjectConfig.

units/periodic_builds.go
hadjri: Yes, thank you

model/patch_lifecycle.go
```diff
@@ -289,13 +296,13 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
 	}
-	intermediateProject, _, err := LoadProjectInto(ctx, []byte(p.PatchedParserProject), opts, p.Project, project)
+	intermediateProject, err := LoadProjectInto(ctx, []byte(p.PatchedParserProject), opts, p.Project, project)
 	if err != nil {
 		return nil, errors.Wrapf(err,
 			"Error marshaling patched parser project from repository revision “%v”",
 			p.Githash)
 	}
-	config, err := CreateProjectConfig([]byte(p.PatchedProjectConfig))
+	config, err := CreateProjectConfig([]byte(p.PatchedProjectConfig), p.Project)
```
ablack12, line 305: should this check version control?

model/project_parser.go
```diff
@@ -530,10 +530,14 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
 	p := &Project{}
 	// opts empty because project yaml with `include` will not hit this case
 	ctx := context.Background()
-	pp, pc, err := LoadProjectInto(ctx, []byte(v.Config), nil, id, p)
+	pp, err = LoadProjectInto(ctx, []byte(v.Config), nil, id, p)
 	if err != nil {
 		return ProjectInfo{}, errors.Wrap(err, "error loading project")
 	}
+	pc, err := CreateProjectConfig([]byte(v.Config), id)
```
ablack12, line 537: Should this be version controlled?

ablack12: Could pass this in as a boolean and then we don't have to look it up for cases where it's not relevant (anywhere we just use .Project)

operations/validate.go
```diff
@@ -151,28 +160,34 @@ func validateFile(path string, ac *legacyClient, quiet, includeLong bool, localM
 
 // loadProjectIntoWithValidation returns a warning (instead of an error) if there's an error with unmarshalling strictly
 func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model.GetProjectOpts,
-	project *model.Project) (*model.ParserProject, validator.ValidationErrors) {
+	project *model.Project) (*model.ParserProject, *model.ProjectConfig, validator.ValidationErrors) {
 	errs := validator.ValidationErrors{}
-	pp, _, err := model.LoadProjectInto(ctx, data, opts, "", project)
-
+	pc, err := model.CreateProjectConfig(data, "")
```
ablack12, line 165: Can we add a comment here that we validate this regardless of if version control is actually being used to verify correctness if it _is_ turned on

ablack12 (commented):
Sorry to balloon this but I do think there are a few more places to audit to make sure IsVersionControlEnabled is used in the right places. If you'd like we could split that to a different ticket.

model/patch_lifecycle.go
hadjri: If version control is disabled PatchedProjectConfig will always be nil, so checking that won't be necessary. I can however wrap this in an if statement if it is an empty string or not.

model/project_parser.go
hadjri: This was the one instance where I was unsure to check for version control since I wasn't sure if we'd want to ad a DB call looking up project ref.  I've added it now.

model/project_configs.go
hadjri: > Sorry to balloon this but I do think there are a few more places to audit to make sure IsVersionControlEnabled is used in the right places. If you'd like we could split that to a different ticket.

I've audited the usages of CreateProjectConfig and there shouldn't be any instances of it being called if a project ref has version control disabled at this point.

hadjri:
> Sorry to balloon this but I do think there are a few more places to audit to make sure IsVersionControlEnabled is used in the right places. If you'd like we could split that to a different ticket.

I've audited the usages of CreateProjectConfig and there shouldn't be any instances of it being called if a project ref has version control disabled at this point.

model/patch_lifecycle.go
```diff
@@ -302,11 +302,14 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 			"Error marshaling patched parser project from repository revision “%v”",
 			p.Githash)
 	}
-	config, err := CreateProjectConfig([]byte(p.PatchedProjectConfig), p.Project)
-	if err != nil {
-		return nil, errors.Wrapf(err,
-			"Error marshaling patched project config from repository revision “%v”",
-			p.Githash)
+	var config *ProjectConfig
+	if len(p.PatchedProjectConfig) > 0 {
```
ablack12, line 306: It doesn't look like we only create PatchedProjectConfig if VersionControl isn't enabled.

model/project_parser.go
```diff
@@ -534,9 +538,12 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
 	if err != nil {
 		return ProjectInfo{}, errors.Wrap(err, "error loading project")
 	}
-	pc, err := CreateProjectConfig([]byte(v.Config), id)
-	if err != nil {
-		return ProjectInfo{}, errors.Wrap(err, "error loading project config")
+	var pc *ProjectConfig
```
ablack12, line 541: Sorry I just realized: this is actually the legacy case  (i.e. before parser project got created). So similar to how we mention we don't handle opts for this (mentioned at line 535), we don't really need to handle project config for this. We do maybe need to handle it at 518? But we could just look it up from the DB, not read it from v.Config.  

(Although I'm not sure if LoadProjectInto needs to populate config)

ablack12 (changes requested):

model/patch_lifecycle.go
ablack12: Gotcha. I'm thinking this scenario still exists though:
1. patch gets created when version control is enabled.
2. version control is toggled off bc of problems.
3. patch is still using the version controlled config. To prevent this, we need to still check here if we intend to use version control.

I'm not going to block this on another review though.

model/project_configs_db.go
```diff
@@ -18,8 +18,8 @@ var (
 	ProjectConfigCreateTimeKey = bsonutil.MustHaveTag(ProjectConfig{}, "CreateTime")
 )
 
-// FindProjectConfigToMerge returns a project config by id, or the most recent project config if id is empty
-func FindProjectConfigToMerge(projectId, id string) (*ProjectConfig, error) {
+// FindProjectConfig returns a project config by id, or the most recent project config if id is empty
+func FindProjectConfig(projectId, id string) (*ProjectConfig, error) {
```
ablack12, line 22: Can this be called something more specific like FindConfigForProjectOrVersion? FindProjectConfig is ambiguous since the structure itself is ProjectConfig

model/project_parser.go
```diff
@@ -505,15 +505,14 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
...
-	pRef, err := FindMergedProjectRef(id, "", false)
-	if err != nil {
-		return ProjectInfo{}, errors.Wrap(err, "error finding project ref")
-	}
 	pp, err = ParserProjectFindOneById(v.Id)
 	if err != nil {
 		return ProjectInfo{}, errors.Wrap(err, "error finding parser project")
 	}
-
+	pc, err := FindProjectConfig(v.Identifier, v.Id)
```
ablack12, line 512: Why do we no longer use version control here?

hadjri: I did this because this is a DB call and should be quick, and isn't doing any unmarshalling like in other places we check for VC. To check for VC we need to add a DB query to retrieve the project ref so I thought we may as well just retrieve the project config if we are going to add 1 DB query.  That being said I'm happy to use version control here for consistency

ablack12: That's a good point. I guess it's up to us to decide what the version control flag means. Does it mean that if it's on, we save and use project configs, and will still use it for any version if it exists even if we toggle the flag off? This would mean we only need to check it in the places where we save the project config. Or does it mean that if it's off, it's off and we don't use the config even if we've stored it? 

Since we _did_ initially create this flag to make DB improvements I would actually be fine to have it mean the former thing. Whatever we decide, we just need to make sure it's documented. And we may want one more ticket to fix things for whichever behavior we decide and/or verify the behavior.

ablack12 (approved):
LGTM -- I left a longer comment about how we should consider version control going forward. Feel free to DM if you want to discuss!

model/project_parser.go
hadjri: Yeah I think it makes most sense for when it's off, don't use the config whatsoever even if we've stored it. I think it's more intuitive.

PR #5392, opened 2022-02-14
EVG-16275 Update agent to handle admin vars

bynn:
[EVG-16275](https://jira.mongodb.org/browse/EVG-16275)

### Description 
update fetch expansions for task by only populating vars with admin only vars if user has admin access 

### Testing 
execution 1 and 2 were done by someone without admin access and doesnt print the admin var
i restarted execution 3 and it printed the expansion 

https://spruce-staging.corp.mongodb.com/version/620ad57eb237361682a73cc7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/model/project_vars.go
+++ b/model/project_vars.go
@@ -216,11 +216,23 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
 	return restrictedVars
 }
 
+func (projectVars *ProjectVars) GetAdminOnlyVars() map[string]string {
+	adminOnlyVars := map[string]string{}
+
+	for k, v := range projectVars.Vars {
+		if projectVars.AdminOnlyVars[k] {
+			adminOnlyVars[k] = v
+		}
+	}
+	return adminOnlyVars
+}
+
+// Returns non-restricted and non-admin only vars for now.
 func (projectVars *ProjectVars) GetUnrestrictedVars() map[string]string {
 	safeVars := map[string]string{}
 
 	for k, v := range projectVars.Vars {
-		if !projectVars.RestrictedVars[k] {
+		if !projectVars.RestrictedVars[k] && !projectVars.AdminOnlyVars[k] {
 			safeVars[k] = v
 		}
 	}
--- a/service/api.go
+++ b/service/api.go
@@ -336,8 +336,21 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
 	if projectVars.PrivateVars != nil {
 		res.PrivateVars = projectVars.PrivateVars
 	}
-
-	u, err := user.FindOneById(t.ActivatedBy)
+	v, err := model.VersionFindOne(model.VersionById(t.Version))
+	if err != nil {
+		as.LoggedError(w, r, http.StatusInternalServerError, err)
+		return
+	}
+	if v == nil {
+		as.LoggedError(w, r, http.StatusNotFound, errors.New("version not found"))
+		return
+	}
+	var u *user.DBUser
+	if t.ActivatedBy == evergreen.StepbackTaskActivator {
+		u, err = user.FindOneById(v.Author)
+	} else {
+		u, err = user.FindOneById(t.ActivatedBy)
+	}
 	if err != nil {
 		as.LoggedError(w, r, http.StatusInternalServerError, err)
 		return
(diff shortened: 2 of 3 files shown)


model/project_vars.go
```diff
@@ -216,11 +216,23 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
...
+
+	for k, v := range projectVars.Vars {
+		if projectVars.AdminOnlyVars[k] {
+			adminOnlyVars[k] = v
+		}
+	}
+	return adminOnlyVars
+}
+
+// Returns non-restricted and non-admin only vars for now.
```
ablack12, line 230: nit: function comments should start with the name of the function

ablack12: Can you link the ticket where this will be changing (probably where you'll be removing restricted) rather than for now?

service/api.go
```diff
@@ -346,6 +345,30 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
 		as.LoggedError(w, r, http.StatusNotFound, errors.New("version not found"))
 		return
 	}
+	var u *user.DBUser
+	if t.ActivatedBy == evergreen.StepbackTaskActivator {
+		u, err = user.FindOneById(v.Author)
+	} else {
+		u, err = user.FindOneById(t.ActivatedBy)
```
ablack12, line 352: Can we only do this if t.ActivatedBy isn't empty ? (to handle the legacy case)

service/api.go
```diff
@@ -346,6 +345,30 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
+	}
+	if err != nil {
+		as.LoggedError(w, r, http.StatusInternalServerError, err)
+		return
+	}
+	if u != nil {
+		// check if user is admin
```
ablack12, line 359: is an* admin

ablack12 (changes requested):

model/project_vars.go

service/api.go
bynn: done

service/api.go
bynn: done

service/api.go
```diff
@@ -346,6 +345,30 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
+	if t.ActivatedBy == evergreen.StepbackTaskActivator {
+		u, err = user.FindOneById(v.Author)
+	} else if t.ActivatedBy != "" {
+		u, err = user.FindOneById(t.ActivatedBy)
+	}
+	if err != nil {
+		as.LoggedError(w, r, http.StatusInternalServerError, err)
+		return
+	}
+	if u != nil {
```
malikchaya2, line 358: should u being nil be treated as an error?

bynn: i dont think so because there could be a non-user activated by and we wouldnt want to error every time we get a nil user

service/api.go
```diff
@@ -346,6 +345,30 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
+		// check if user is an admin
+		authorized := u.HasPermission(gimlet.PermissionOpts{
```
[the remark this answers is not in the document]
bynn: the function only seems to exist for gimlet users and not for dbUsers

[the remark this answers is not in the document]
ablack12: gimlet.User is an interface, which DBUser implements, so you should still be able to call this function with it (note this other use case which passes in a DB user)
https://github.com/evergreen-ci/evergreen/blob/5b4d40574043d889a4f652a87da98daf2d2d080e/service/project.go#L992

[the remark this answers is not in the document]
bynn: isAdmin updated to take string

service/api.go
```diff
@@ -346,6 +345,25 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
+		u, err = user.FindOneById(t.ActivatedBy)
+	}
+	if err != nil {
+		as.LoggedError(w, r, http.StatusInternalServerError, err)
+		return
+	}
+	if u != nil {
+		// check if user is an admin
+		authorized := isAdmin(u, t.Project)
+		if authorized {
```
ablack12, line 361: Can we do this inline instead of assigning a variable? `if isAdmin()`

ablack12 (approved):
lgtm with nit

PR #5395, opened 2022-02-15
EVG-15557 Update version message with patch description

bynn:
[EVG-15557](https://jira.mongodb.org/browse/EVG-15557)

### Description 
update version message so the patch page shows update name 

### Testing 
image


graphql/util.go
```diff
@@ -235,6 +235,9 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
 	if err = p.SetDescription(patchUpdateReq.Description); err != nil {
 		return errors.Wrap(err, "Error setting description"), http.StatusInternalServerError, "", ""
 	}
+	if err = model.UpdateVersionMessage(p.Version, patchUpdateReq.Description); err != nil {
```
ablack12, line 238: We should only do this if p.Version != "" (so should move to 266)

ablack12 (changes requested):

model/version_db.go
```diff
@@ -314,6 +314,17 @@ func VersionUpdateOne(query interface{}, update interface{}) error {
 	)
 }
 
+func UpdateVersionMessage(versionId, message string) error {
```
malikchaya2, line 317: in the comment on the ticket, it says that the problem is  `that the version gets the update but the patch does not.`. Is it really the opposite, that the patch gets it but not the version?

malikchaya2: that would make sense since it shows up on "my patches" but not the actual patch page.

malikchaya2 (approved):
lgtm after Annie's comments are addressed.

graphql/util.go
bynn: done

model/version_db.go
bynn: yeah, the patch in the db is being updated but the UI uses the version page displays the version message which isnt being updated

graphql/util.go
```diff
@@ -259,6 +259,17 @@ func SchedulePatch(ctx context.Context, patchId string, version *model.Version,
 		if version == nil {
 			return errors.Errorf("Couldn't find patch for id %v", p.Version), http.StatusInternalServerError, "", ""
 		}
+
+		v, err := model.VersionFindOneId(p.Version)
```
ablack12, line 263: We don't need to call this here; you can see at line 259 that we've already checked that the  version isn't nil (it's passed into schedule patch)

rest/route/patch_test.go
```diff
@@ -691,6 +691,10 @@ buildvariants:
 		PatchedParserProject: config,
 	}
 	require.NoError(t, unfinalized.Insert())
+	// version := serviceModel.Version{
```
ablack12, line 694: Should remove this

ablack12 (changes requested):

graphql/util.go
bynn: done

rest/route/patch_test.go
bynn: sorry, done

PR #5396, opened 2022-02-15
EVG-16095 Ignore all filter for statuses

khelif96:
[EVG-16095](https://jira.mongodb.org/browse/EVG-16095)

### Description 
If a user includes "all" in a task filter they want all of the matching task statuses so we should ignore that field in the `$match` stage of the pipeline since we don't want to filter anything out.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -291,22 +291,6 @@ var TaskStatuses = []string{
 	TaskTimedOut,
 }
 
-// FilterValidTaskStatuses returns a slice of task statuses that are valid and are searchable.
-// It returns an empty array if emptyArrayForAll is true
-func FilterValidTaskStatuses(statuses []string, emptyArrayForAll bool) []string {
-	var filteredStatuses []string
-	if emptyArrayForAll && utility.StringSliceContains(statuses, TaskAll) {
-		return []string{}
-	}
-	for _, status := range statuses {
-		if !utility.StringSliceContains(TaskStatuses, status) {
-			continue
-		}
-		filteredStatuses = append(filteredStatuses, status)
-	}
-	return filteredStatuses
-}
-
 var InternalAliases = []string{
 	CommitQueueAlias,
 	GithubPRAlias,
--- a/globals.go
+++ b/globals.go
@@ -88,6 +88,9 @@ const (
 	// This is not an official task status; it is used by the front end to indicate that there is a linked issue in the annotation
 	TaskKnownIssue = "known-issue"
 
+	// This is not an official task status; it is used by the front end to indicate that the filter should apply to all of the tasks
+	TaskAll = "all"
+
 	// Task Command Types
 	CommandTypeTest   = "test"
 	CommandTypeSystem = "system"
@@ -271,6 +274,38 @@ const (
 	KeyTooLargeToIndexError    = "key too large to index"
 	InvalidDivideInputError    = "$divide only supports numeric types"
 )
+
+var TaskStatuses = []string{
+	TaskStarted,
+	TaskSucceeded,
+	TaskFailed,
+	TaskSystemFailed,
+	TaskTestTimedOut,
+	TaskSetupFailed,
+	TaskAborted,
+	TaskStatusBlocked,
+	TaskStatusPending,
+	TaskKnownIssue,
+	TaskSystemUnresponse,
+	TaskSystemTimedOut,
+	TaskTimedOut,
+}
+
+// FilterValidTaskStatuses returns a slice of task statuses that are valid and are searchable.
+// It returns an empty array if emptyArrayForAll is true
+func FilterValidTaskStatuses(statuses []string, emptyArrayForAll bool) []string {
+	var filteredStatuses []string
+	if emptyArrayForAll && utility.StringSliceContains(statuses, TaskAll) {
+		return []string{}
+	}
+	for _, status := range statuses {
+		if !utility.StringSliceContains(TaskStatuses, status) {
+			continue
+		}
+		filteredStatuses = append(filteredStatuses, status)
+	}
+	return filteredStatuses
+}
 
 var InternalAliases = []string{
 	CommitQueueAlias,
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -483,7 +483,7 @@ func generateBuildVariants(sc data.Connector, versionId string, searchVariants [
 		{Key: task.DisplayNameKey, Order: 1},
 	}
 	opts := data.TaskFilterOptions{
-		Statuses:         statuses,
+		Statuses:         evergreen.FilterValidTaskStatuses(statuses, true),
 		Variants:         searchVariants,
 		TaskNames:        searchTasks,
 		Sorts:            defaultSort,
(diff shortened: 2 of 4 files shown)


malikchaya2 (changes requested):
This looks great! 
How was it tested?

model/task/task.go
```diff
@@ -2990,11 +2990,14 @@ func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task,
 	pipeline := getTasksByVersionPipeline(versionID, opts)
 
 	if len(opts.BaseStatuses) > 0 {
-		pipeline = append(pipeline, bson.M{
-			"$match": bson.M{
-				BaseTaskStatusKey: bson.M{"$in": opts.BaseStatuses},
-			},
-		})
+		// If we are filtering for all statuses it is more performant to not include the status filter
+		if !utility.StringSliceContains(opts.BaseStatuses, "all") {
```
ablack12, line 2994: Can we make this a constant somewhere?

ablack12: Actually, if the user chooses "all" on the UI can we just not specify any base statuses? (or can that be handled in the resolver?) Since this is pretty UI specific it seems better to handle outside of this function.

ablack12 (changes requested):

model/task/task.go
```diff
@@ -3358,12 +3361,16 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
-			},
-		})
+		// If a user is filtering for all statuses, it is more performant to skip the status filter
+		if !utility.StringSliceContains(opts.Statuses, "all") {
+			pipeline = append(pipeline, bson.M{
+				"$match": bson.M{
+					DisplayStatusKey: bson.M{"$in": opts.Statuses},
+				},
+			})
+		}
```
minnakt, line 3370: Hmm.. I'm confused because it looks like we already pass `[]` when `"All"` is selected... 🤔

Screen Shot 2022-02-15 at 4 16 21 PM

maybe it's not being passed properly? because I do see `all` being passed 

Screen Shot 2022-02-15 at 4 20 14 PM

model/task/task.go
khelif96: 1. I ended up moving this check into the pipeline and included it in the guard clause here. We shouldn't be applying base status filters if include base tasks is false
https://github.com/evergreen-ci/evergreen/pull/5396/files#diff-4b29d204064065e99ec7762e466f51f0ac8461fcc5f8fc2eeda53d87a5044060R3363-R3375

model/task/task.go
khelif96: That is possible Thanks for pointing it out. I will be updating that logic as well and removing that statement since now its more efficient to include "all" in this query.

model/task/task.go
ablack12: This isn't quite what I meant; "all" is very UI specific (it's not something that the backend would ever pass in) and since this is used in a lot of places I think it makes sense to do set opts.BaseStatuses to nil before ever calling GetTasksByVersion. Minna's review seems worth looking into to address that though.

model/task/task.go
minnakt: ^^ Also, this is only for the Tasks page, I don't think we have this logic in the Commits page.

I also think this function might not work as expected? When clicking `"All"` I actually see every statuses _except_ `all` in the backend, when I think it should really just be an empty array.

model/task/task.go
khelif96: This pipeline stage is used in several places i think it makes sense to include this logic here as a general guard for everywhere `getTasksByVersionPipeline` as opposed to leaving it up to the implementer. AFAICT these queries are only used on the UI so it makes sense to tie these UI concepts to it. The entire query revolves around UI specific concepts such as display statuses, execution task fold ups, build variant display names.

khelif96:
> This looks great! How was it tested (to make sure it didn't accidentally break something)?

There are several existing tests around this query so they should hopefully catch any broken changes.

model/task/task.go
ablack12: Still, if we're going to keep this in the pipeline I'd like it to be more general than assuming globals from the UI. Can we have a list in global of all TaskStatuses and then filter those out of BaseStatuses before checking length?

globals.go
```diff
@@ -272,6 +275,38 @@ const (
...
+	TaskStatusPending,
+	TaskKnownIssue,
+	TaskSystemUnresponse,
+	TaskSystemTimedOut,
+	TaskTimedOut,
+}
+
+// FilterValidTaskStatuses returns a slice of task statuses that are valid and are searchable.
+// It returns an empty array if emptyArrayForAll is true
+func FilterValidTaskStatuses(statuses []string, emptyArrayForAll bool) []string {
```
ablack12, line 296: Can this function not be in globals? Since it's just used by graphql it could be a private function in the util

globals.go
```diff
@@ -272,6 +275,38 @@ const (
...
+	var filteredStatuses []string
+	if emptyArrayForAll && utility.StringSliceContains(statuses, TaskAll) {
+		return []string{}
```
ablack12, line 299: Also I would preference initializing filteredStatuses as filteredStatuses := []string{} and just returning that here

ablack12 (changes requested):

model/task/task.go
```diff
@@ -3364,6 +3356,17 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 			},
 		})
 	}
+
+	if opts.IncludeBaseTasks {
```
ablack12, line 3360: can you combine these into one if statement?

graphql/util.go
```diff
@@ -1406,3 +1406,20 @@ func hasProjectPermission(ctx context.Context, resource string, next graphql.Res
...
+
+// getValidTaskStatusesFilter returns a slice of task statuses that are valid and are searchable.
+// It returns an empty array if all is included as one of the entries
+func getValidTaskStatusesFilter(statuses []string) []string {
+	filteredStatuses := []string{}
+	if utility.StringSliceContains(statuses, evergreen.TaskAll) {
+		return filteredStatuses
+	}
+	filteredStatuses = utility.StringSliceIntersection(evergreen.TaskStatuses, statuses)
+	for _, status := range statuses {
```
ablack12, line 1418: Did you purposefully leave this for loop here? StringSliceIntersection already handled this so now you've added all statuses twice

khelif96: oops

ablack12 (approved):

minnakt (approved):

model/task/task.go
```diff
@@ -3364,6 +3356,15 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 			},
 		})
 	}
+
+	if opts.IncludeBaseTasks && len(opts.BaseStatuses) > 0 {
```
malikchaya2, line 3360: do we want to include this in the above` if opts.IncludeBaseTasks` ?  Do we specifically want it after the two if statements that follow?

malikchaya2 (approved):
I added a question, but it's small enough so it doesn't have to hold up the review.

model/task/task.go
khelif96: The base status filter depends on base status existing but base status does not depend on a filter existing.

PR #5399, opened 2022-02-15
EVG-16172: add admin settings for queue-specific settings

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16172

### Description 
I hard-coded the queue configuration for the event notifier, so these admin settings are meant to replace it with values that can be configured and don't require re-compiling to set them. Before this is deployed, I'm going to set the proper values in the DB's Amboy settings so that the event notifier uses the values that I replaced.

* Create admin settings to set queue-specific options for named queues in the queue group.
* Remove hard-coded event notifier queue settings.

The new settings look like this:
image

### Testing 
* Updated the standard REST tests for admin settings.
* Tested in local Evergreen and staging to verify that the admin page settings worked as expected.


environment.go
```diff
@@ -468,6 +453,27 @@ func (e *envState) getRemoteQueueGroupDBOptions() queue.MongoDBOptions {
 	return opts
 }
 
+func (e *envState) getNamedRemoteQueueOptions() map[string]queue.MongoDBQueueOptions {
+	perQueueOpts := map[string]queue.MongoDBQueueOptions{}
+	for _, namedQueue := range e.settings.Amboy.NamedQueues {
+		dbOpts := e.getRemoteQueueGroupDBOptions()
+		if namedQueue.SampleSize != 0 {
```
hhoke, line 460: It seems to me that if namedQueue.SampleSize is zero, dbOpts will default to 0. Why not just set this to namedQueue.SampleSize?

rest/model/admin.go
```diff
@@ -439,6 +450,30 @@ func (a *APIAmboyRetryConfig) ToService() (interface{}, error) {
 	}, nil
 }
 
+// APIAmboyNamedQueueConfig is the model for named Amboy queue settings.
+type APIAmboyNamedQueueConfig struct {
+	Name               *string `json:"name"`
```
hhoke, line 455: For my own understanding: What is the point of having this secondary struct, that only differs from `AmboyNamedQueueConfig` in that it has a string pointer instead of a string?

hhoke (changes requested):

rest/model/admin.go
Kimchelly: The API-level struct is required to separate the REST request/response data model from the DB data model since they're different layers of abstraction.

Passing the named queues as a slice with the name as a field instead of using a map is just simpler for JSON payloads. It's not that desirable to use a slice where a map is more appropriate, but it makes writing the JS easier and it's not a major issue since admins are the only ones using it.

environment.go
Kimchelly: Mostly just to leave open the future possibility that remote queue group default fields such as sample size can be set and respected. The options for named queues should only override the defaults if they're explicitly set and otherwise should obey whatever default is set. Doing this zero/non-zero check when following options precedence/defaulting rules is a common thing in our code bases.

hhoke (approved):

PR #5400, opened 2022-02-16
EVG-16194 add resolver for conflicting refs

ablack12:
[EVG-16194](https://jira.mongodb.org/browse/EVG-16194)

### Description 
Added backend function and resolver to get github conflicts for a project. Used this new function in some existing backend functions. 

Also fixed a bug where we query the ProjectRef collection for branch directly (whereas this can be set from the repo as well).

### Testing 
Added unit tests and ensuring existing tests pass.

### DIFF
--- a/gqlgen.yml
+++ b/gqlgen.yml
@@ -260,6 +260,8 @@ models:
     model: github.com/evergreen-ci/evergreen/rest/model.APINotificationPreferences
   GithubUserInput:
     model: github.com/evergreen-ci/evergreen/rest/model.APIGithubUser
+  GithubProjectConflicts:
+    model: github.com/evergreen-ci/evergreen/model.GithubProjectConflicts
   ClientConfig:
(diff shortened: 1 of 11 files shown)


malikchaya2 (approved):

sophstad:
Per our discussion and the ticket, I think we also wanted to update the save resolvers to return an error if a user attempts to make a change with a collision (to avoid any race conditions). Should that be included in this PR?

sophstad (commented):
Per our discussion and the ticket, I think we also wanted to update the save resolvers to return an error if a user attempts to make a change with a collision (to avoid any race conditions). Should that be included in this PR?

ablack12:
@sophstad I started trying to add it but I think it's going to be a slightly difficult change; given that it _can_ be logically broken up from this piece I think I'd like to do it as part of [EVG-16338](https://jira.mongodb.org/browse/EVG-16338) instead.

PR #5402, opened 2022-02-16
EVG-16176: add method to create new intent pod for tasks

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16176

### Description 
* Add method to create new intent pod for a task.
* Remove `Secret` field from pods. It was originally used to represent the agent secret, but it shouldn't be necessary to have a separate field for it since it already has to be stored in the secret environment variables. I removed it to reduce the duplication in the data model.

### Testing 
Updated existing unit tests and added new ones for `NewTaskIntentPod`.


hhoke (approved):
LGTM

PR #5404, opened 2022-02-16
EVG-15559 Add activated flag to test data

khelif96:
[EVG-15559](https://jira.mongodb.org/browse/EVG-15559)

### Description 
I noticed a really interesting test failure in https://github.com/evergreen-ci/spruce/pull/1093 
the `version_restart_modal` test kept failing because the modal stopped showing build variants for this version and this version only. 

upon investigation the query that returns the build variants was returning a null value which is confusing because the query and back end resolver the modal relied on were not changed at all. 

I took a look at the resolver locally and it was correctly returning the tasks. 
But when it would run in CI it would keep failing. 
I realized this was likely because of some earlier test breaking the test.
I narrowed it down to the `task_actions.ts` test. 
Once i reproed locally i was able to take a look at the struct in the db and noticed the `activated` flag was set to false in the version. 
I narrowed it down to https://github.com/evergreen-ci/evergreen/blob/14b08789b6cfe3512d98d2001777d92bc8c1d4f7/graphql/resolvers.go#L3802-L3813

Apparently since `activated` was originally nil it would hit this code path which would try to determine if the version had any activated tasks and if it didn't it would set the `activated` field to `false` and not return any tasks.

Since `task_actions` unschedules the task once the above code is ran it also marks the version as inactive and would cause it to not return tasks.

You may ask yourself how come this worked before but suddenly stopped in Mohameds pr? 

> I think it never worked and we were instead seeing a ghost value that was stored in the apollo cache. 
> The build variants resolver was also used for the build variants grid so it would have loaded the correct data at least once before `task_actions` was ran. 
> So once we stopped making that query the apollo cache never had a record of that value so it would always hit the network and return a null value.
> 

Won't this be a problem in prod?

> No, activated is set by default on new versions and has been for several months so there isn't a risk of this value being nil  and someone unscheduling a task and making this value false. 

### Testing 
Tested locally and confirmed it works after running those tests


malikchaya2 (approved):

sophstad:
🤯 nice investigating!! I hope we don't start running into more mysterious Apollo caching issues, but glad now it's on our radar if we do!

PR #5410, opened 2022-02-17
EVG-16253 upgrade mongodb version for tests

ybrill:
[EVG-16253](https://jira.mongodb.org/browse/EVG-16253)

### Description 
* Change the db versions for each variant to 4.2.18. Because mongo doesn't package a precompiled tarball for 4.2 on Ubuntu16.04 ARM I changed that variant to Ubuntu 18.04 ARM. There's also no tarball for Arch. We could experiment with using one of the tarballs for another Linux distro but (as Brian noted on the ticket) there's no compelling reason to run race-detector on Arch so it will run on Ubuntu. It made sense to me to run on Ubuntu 16.04 because the app servers run on it and we're running the non-race tests on it already.
* Use the current distro names for ubuntu1604 (small instead of test and large instead of build). It means the same thing at the end of the day, but why not use the distros' proper names?
* Set the fCV of the db to match what's running in prod. If this makes it through code review I'll open a ticket to remove it once we move the db.
* Consolidate the self-tests.yml functions to setup mongodb with and without auth.

### Testing 
I can still run tests in a patch 😺.

### DIFF
--- a/self-tests.yml
+++ b/self-tests.yml
@@ -238,53 +238,16 @@ functions:
         background: true
         working_dir: evergreen
         command: make start-mongod
-    - command: subprocess.exec
-      type: setup
-      params:
-        working_dir: evergreen
-        command: make check-mongod
-    - command: subprocess.exec
-      type: setup
-      params:
-        working_dir: evergreen
-        command: make init-rs
-    - command: subprocess.exec
-      type: setup
-      params:
         env:
+          AUTH_ENABLED: ${with_auth}
+    - command: subprocess.exec
+      type: setup
+      params:
+        working_dir: evergreen
+        command: make configure-mongod
+        env:
+          AUTH_ENABLED: ${with_auth}
           FCV: "4.0"
-        working_dir: evergreen
-        command: make set-mongo-fcv
-  setup-mongodb-auth:
-    - command: subprocess.exec
-      type: setup
-      params:
-        env:
-          MONGODB_URL: ${mongodb_url}
-          DECOMPRESS: ${decompress}
-        working_dir: evergreen
-        command: make get-mongodb
-    - command: subprocess.exec
-      type: setup
-      params:
-        background: true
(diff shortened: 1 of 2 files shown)


self-tests.yml
```diff
@@ -238,46 +238,16 @@ functions:
...
-        command: make init-rs
-    - command: subprocess.exec
-      type: setup
-      params:
-        working_dir: evergreen
-        command: make init-auth
+        command: make configure-mongod
+        env:
+          AUTH_ENABLED: ${with_auth}
+          FCV: "4.0"
```
Kimchelly, line 250: Is there a reason we're not upgrading to FCV 4.2?

self-tests.yml
```diff
@@ -625,12 +596,12 @@ buildvariants:
   - name: race-detector
     display_name: Race Detector
     run_on:
-      - archlinux-new-small
-      - archlinux-new-large
+      - ubuntu1604-small
```
Kimchelly, line 599: Is there a reason that Evergreen would switch away from archlinux for the race detector? All the other repos use archlinux and if I remember correctly, the tests behave slightly differently on archlinux than on ubuntu.

If we're switching this, we have to switch all the other repos.

Kimchelly (commented):
Are we going to upgrade all the other repos that use 4.0?

self-tests.yml
ybrill: That's interesting. How would it have benefited us to have the latest golang?
I'm game to update the tests on other repos. I'll open ticket(s), as Kim suggested.

bsamek: I think the idea was to have an early warning system about anything that might break in a newer version of Go. I don't remember that ever actually happening though.

hhoke (approved):
LGTM

ybrill:
I opened [EVG-16381](https://jira.mongodb.org/browse/EVG-16381)

PR #5413, opened 2022-02-18
EVG-15335 Update host extension error messages

bynn:
[EVG-15335](https://jira.mongodb.org/browse/EVG-15335)

### Description 
Previous behavior: 
does not allow hosts to be extended 30 past their creation dates
can only extend up to 2 weeks from the current time 

Now: 
only checks that you can't extend the host beyond 30 days after creation

### Testing 
image

### DIFF
--- a/cloud/spawn.go
+++ b/cloud/spawn.go
@@ -324,10 +324,9 @@ func ModifySpawnHost(ctx context.Context, env evergreen.Environment, host *host.
 func MakeExtendedSpawnHostExpiration(host *host.Host, extendBy time.Duration) (time.Time, error) {
 	if err := host.PastMaxExpiration(extendBy); err != nil {
 		return time.Time{}, err
-	} else {
-		newExp := host.ExpirationTime.Add(extendBy)
-		return newExp, nil
 	}
+	newExp := host.ExpirationTime.Add(extendBy)
+	return newExp, nil
 }
 
 func modifySpawnHostProviderSettings(d distro.Distro, settings *evergreen.Settings, region, volumeID string) ([]*birch.Document, error) {
--- a/cloud/spawn.go
+++ b/cloud/spawn.go
@@ -324,15 +324,10 @@ func ModifySpawnHost(ctx context.Context, env evergreen.Environment, host *host.
 func MakeExtendedSpawnHostExpiration(host *host.Host, extendBy time.Duration) (time.Time, error) {
 	if err := host.PastMaxExpiration(extendBy); err != nil {
 		return time.Time{}, err
+	} else {
+		newExp := host.ExpirationTime.Add(extendBy)
+		return newExp, nil
 	}
-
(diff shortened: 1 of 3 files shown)


malikchaya2 (commented):
Why do we want to remove the check for "can only extend up to 2 weeks from the current time"?

cloud/spawn.go
```diff
@@ -324,15 +324,10 @@ func ModifySpawnHost(ctx context.Context, env evergreen.Environment, host *host.
 func MakeExtendedSpawnHostExpiration(host *host.Host, extendBy time.Duration) (time.Time, error) {
 	if err := host.PastMaxExpiration(extendBy); err != nil {
 		return time.Time{}, err
+	} else {
```
malikchaya2, line 327: Since the if returns, I don't think putting this in an else helps.

cloud/spawn.go
```diff
@@ -324,15 +324,10 @@ func ModifySpawnHost(ctx context.Context, env evergreen.Environment, host *host.
...
+		newExp := host.ExpirationTime.Add(extendBy)
+		return newExp, nil
 	}
-
-	newExp := host.ExpirationTime.Add(extendBy)
-	remainingDuration := newExp.Sub(time.Now()) //nolint
```
malikchaya2, line 330: should we maybe move this check above `host.ExpirationTime.Add` instead of removing it?

cloud/spawn.go
```diff
@@ -324,15 +324,10 @@ func ModifySpawnHost(ctx context.Context, env evergreen.Environment, host *host.
...
-	if remainingDuration > evergreen.MaxSpawnHostExpirationDurationHours {
-		return time.Time{}, errors.Errorf("Can not extend host '%s' expiration by '%s'. Maximum host duration is limited to %s", host.Id, extendBy.String(), evergreen.MaxSpawnHostExpirationDurationHours.String())
```
malikchaya2, line 332: should we maybe move this check above `host.ExpirationTime.Add` instead of removing it?

cloud/spawn.go
bynn: done

cloud/spawn.go
bynn: im not sure if we want to have the arbitrary 2 week extension maximum and the PastMaxExpiration should be checking everything (i think)

malikchaya2: Whether or not we should have an arbitrary 2 week extension maximum feels more like a product question. I think we should double check with Annie before removing it.

ablack12: The idea behind the 2 week extension check was just to prevent users from extending too far when they don't need to; that being said since we _do_ have the one month max so there's really no reason to have two levels of prevention (this was pre-nonexpirable hosts days when we were likely trying to be careful about money here). I'm fine with removing it.

malikchaya2 (approved):

PR #5414, opened 2022-02-22
EVG-16155 prevent last-green from erroring for nonexistent variants

ablack12:
[EVG-16155](https://jira.mongodb.org/browse/EVG-16155)

### Description 
Generated variants might not exist yet for the most recent version, so erroring if they aren't present isn't correct behavior. Instead we should just let the query handle it.

### Testing 
Not sure there's a great way to get the staging lock while slack is down but I will test on staging before merging.

### DIFF
--- a/service/rest_version.go
+++ b/service/rest_version.go
@@ -580,16 +580,11 @@ func (ra *restAPI) lastGreen(w http.ResponseWriter, r *http.Request) {
 	// GET /rest/v1/projects/mongodb-mongo-master/last_green?linux-64=1&windows-64=1
 	queryParams := r.URL.Query()
 
-	// Make sure all query params are valid variants and put them in an array
+	// Don't validate build variants are in project, since they may be generated
+	// variants that don't yet exist for the latest parser project.
 	var bvs []string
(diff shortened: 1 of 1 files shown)


service/rest_version.go
```diff
@@ -580,16 +580,11 @@ func (ra *restAPI) lastGreen(w http.ResponseWriter, r *http.Request) {
 	// GET /rest/v1/projects/mongodb-mongo-master/last_green?linux-64=1&windows-64=1
 	queryParams := r.URL.Query()
 
-	// Make sure all query params are valid variants and put them in an array
+	// Don't validate build variants are in project, since they may be generated
+	// variants that don't yet exist for the latest parser project.
```
malikchaya2, line 584: Are we opening it up for potential errors down the line if they don't end up being generated?

ablack12: The consequence would be that the query returns nothing, if the build variant actually doesn't match any existing version. I do think this is reasonable behavior.

malikchaya2 (approved):

bynn (approved):

PR #5415, opened 2022-02-23
Revert "Revert "EVG-16170: only consider host tasks for the distro scheduler""

Kimchelly:
Reverts evergreen-ci/evergreen#5369


hhoke (approved):
LGTM, proper indexes should be built now.

PR #5416, opened 2022-02-23
EVG-6496 Auto update CLI if it is set in client settings before each command

hadjri:

### Description 
New property has been added in client settings to optionally automatically update the evergreen CLI as a before function on a subset of CLI commands.

### Testing 
Deployed to staging and built CLI locally with new client settings to auto update and tested manually.


operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
 				return errors.Wrap(err, "problem loading configuration")
 			}
-
-			client := conf.getRestCommunicator(ctx)
-			defer client.Close()
-
-			update, err := checkUpdate(client, false, forceUpdate)
-			if err != nil {
-				return err
+			if !conf.AutoUpgradeCLI && !autoUpgrade {
```
malikchaya2, line 62: Do we want to do this always, or only when there is an update available?

operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
-			if err != nil {
-				return err
-			}
-			defer func() {
-				if doInstall {
-					grip.Error(os.Remove(updatedBin))
+			if !conf.AutoUpgradeCLI && autoUpgrade {
+				conf.SetAutoUpgradeCLI()
+				if err := conf.Write(""); err != nil {
+					return errors.Wrap(err, "error setting auto-upgrade CLI option")
```
malikchaya2, line 71: what does conf.Write("") do?

malikchaya2 (changes requested):

operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
 			}
-
-			client := conf.getRestCommunicator(ctx)
-			defer client.Close()
-
-			update, err := checkUpdate(client, false, forceUpdate)
-			if err != nil {
-				return err
+			if !conf.AutoUpgradeCLI && !autoUpgrade {
+				fmt.Println(fmt.Sprintf("Automatic CLI upgrades are not set up; specifying the -%s flag will enable automatic CLI upgrades before each command if the current CLI is out of date.", autoUpgradeFlagName))
```
ablack12, line 63: fmt.Printf

operations/update.go
ablack12: I think doing it always is reasonable since this is the update command.

operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
+	}
+}
 
-			if doInstall {
-				grip.Infoln("Upgraded binary successfully downloaded to temporary file:", updatedBin)
+// CheckAndUpdateVersion checks if the CLI is up to date. If there is no new CLI version it will simply notify that the current binary is up to date.
+// If there is a new version available, it will be downloaded, and if doInstall is set the new binary will automatically be replaced at the current path the old binary existed in.
+// Otherwise, it will simply be downloaded and a suggested 'mv' command will be printed so the user can replace the binary at their discretion.
+// Toggling forceUpdate will download a new CLI even if the current CLI does not have an out-of-date CLI version string.
+func CheckAndUpdateVersion(conf *ClientSettings, ctx context.Context, doInstall bool, forceUpdate bool, silent bool) error {
```
ablack12, line 84: This should be private since it's only used in operations

operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
-					grip.Error(os.Remove(updatedBin))
+			if !conf.AutoUpgradeCLI && autoUpgrade {
+				conf.SetAutoUpgradeCLI()
+				if err := conf.Write(""); err != nil {
+					return errors.Wrap(err, "error setting auto-upgrade CLI option")
 				}
-			}()
+				fmt.Println("Automatic CLI upgrades have successfully been setup.")
+			}
+			return CheckAndUpdateVersion(conf, ctx, doInstall, forceUpdate, false)
```
ablack12, line 75: We should do this regardless of auto upgrade things, right?

operations/update.go
ablack12: nit: --%s flag

operations/update.go
```diff
@@ -53,96 +59,116 @@ func Update() cli.Command {
...
-			update, err := checkUpdate(client, false, forceUpdate)
-			if err != nil {
-				return err
+			if !conf.AutoUpgradeCLI && !autoUpgrade {
+				fmt.Println(fmt.Sprintf("Automatic CLI upgrades are not set up; specifying the -%s flag will enable automatic CLI upgrades before each command if the current CLI is out of date.", autoUpgradeFlagName))
 			}
-			if !update.needsUpdate || len(update.binaries) == 0 {
-				return nil
+			if conf.AutoUpgradeCLI && autoUpgrade {
+				fmt.Println(fmt.Sprintf("Automatic CLI upgrades are already set up, specifying the %s flag is not necessary.", autoUpgradeFlagName))
```
ablack12, line 66: nit: --%s flag

ablack12 (changes requested):

operations/update.go
hadjri: When passing an empty string, the conf struct will be written to the default config file location stored in conf.LoadedFrom. If that is also not set, an error gets thrown.

operations/update.go
hadjri: I think so. Since this is explicitly the update command I think we should grant their wish since it wouldn't do any harm. I've made one change that will automatically install the new CLI if the --auto flag is passed or if the get-update command gets run and auto_upgrade_cli is set on the client settings.

hadjri:
Once this is ready for approval I'll update the wiki!

malikchaya2 (approved):

operations/before.go
```diff
@@ -95,6 +97,23 @@ var (
...
+		confPath := c.String("conf")
+		conf, err := NewClientSettings(confPath)
+		if err != nil {
+			return errors.Wrap(err, "problem loading configuration")
+		}
+		if conf.AutoUpgradeCLI {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if err = checkAndUpdateVersion(conf, ctx, true, false, true); err != nil {
+				fmt.Println("Automatic CLI update failed, continuing command execution")
```
ablack12, line 111: Will this actually continue command execution if we return an error? (Have we tested that?)

operations/commit_queue.go
```diff
@@ -57,6 +57,7 @@ func listQueue() cli.Command {
 		Usage: "list the contents of a project's commit queue",
 		Flags: addProjectFlag(),
 		Before: mergeBeforeFuncs(
+			autoUpdateCLI,
```
ablack12, line 60: Can we actually only do this for some functions? As opposed to all functions? Deleting a commit queue item for example, people might not want that to take a long time. And listing. Maybe we should only have autoUpdate be for some of the more common commands (evergreen commit-queue merge, evergreen patch, evergreen validate) Or maybe let's just remove it for some of the ones that should be fast (like this deleting item one) and drop a comment saying why it's not included (so we don't think it's an oversight)

operations/patch_cancel.go
```diff
@@ -13,7 +13,7 @@ func PatchCancel() cli.Command {
 		Name:   "cancel-patch",
 		Usage:  "cancel an existing patch",
 		Flags:  addPatchIDFlag(),
-		Before: requirePatchIDFlag,
+		Before: mergeBeforeFuncs(autoUpdateCLI, requirePatchIDFlag),
```
ablack12, line 16: This is another one that probably shouldn't auto-update

operations/commit_queue.go
hadjri: Sounds good, I've kept it for most patch and commit-queue functions that don't just delete something, and for quick to return functions like lists and version. Added a comment above autoUpdateCLI to explain the rationale.

operations/before.go
hadjri: It would return early and not execute the command if the download failed for some reason. I get why this would be problematic, better to allow the user to continue on with the old version rather than manually removing auto_upgrade_cli from their config yaml. I've removed the return err statement and now we'll just notify in the logs that the download failed in the off case that it does fail.

operations/before.go
```diff
@@ -95,6 +97,26 @@ var (
...
+		confPath := c.String("conf")
+		conf, err := NewClientSettings(confPath)
+		if err != nil {
+			return errors.Wrap(err, "problem loading configuration")
+		}
+		if conf.AutoUpgradeCLI {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if err = checkAndUpdateVersion(conf, ctx, true, false, true); err != nil {
+				fmt.Println("Automatic CLI update failed! Continuing with command execution.")
```
ablack12, line 115: Should we include the error in the print statement to help us debug this if the user asks about it? Also can this be grip.Error instead of fmt.Println

ablack12 (approved):

operations/before.go
hadjri: Sure, will add

PR #5417, opened 2022-02-23
EVG-16391 Add will run to taskStatus list

khelif96:
[EVG-16391](https://jira.mongodb.org/browse/EVG-16391)

### Description 
Will run and a few other task statuses were missing from the TaskStatuses list. They are used [here](https://github.com/evergreen-ci/evergreen/blob/73ba9f83725d2edbce41efa2995fad2317043b89/graphql/util.go#L1433-L1440) to determine what filters we can apply.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -292,6 +292,9 @@ var TaskStatuses = []string{
 	TaskSystemUnresponse,
 	TaskSystemTimedOut,
 	TaskTimedOut,
+	TaskWillRun,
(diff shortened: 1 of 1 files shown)


minnakt (approved):
Is this possibly missing `TaskUnstarted` and `TaskDispatched`? These are two possible filters that can be applied on the `/commits` page but not sure if they should also be in here.

malikchaya2 (approved):

PR #5418, opened 2022-02-23
EVG-16174: implement pod allocator job

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16174

### Description 
The pod allocator job has to create a new intent pod for a task which is waiting for a container. It also associates the task with the pod by creating/updating the pod dispatcher, which makes decisions on which task(s) get dispatched to which pod(s). Each pod dispatcher is responsible for dispatching a single task for now (task groups are future work, but it's designed to eventually be able to dispatch entire task groups). Once the allocation is done, the task transitions to the "container allocated" state, after which the task waits for the container to start and for the agent to pick up the task.

* Implement job to allocate a pod to a task and create the pod dispatcher.
    * I tried to avoid using a transaction, but there would be a lot of edge cases and require additional information in the data models to account for the possibility that the DB updates partly succeed. It's easier to make it correct with the transaction.
    * The new GroupID field still needs a unique index to enforce correctness for the dispatcher, which is going to be added in [EVG-16321](https://jira.mongodb.org/browse/EVG-16321).
* Add a bunch of DB helper methods for the pod dispatcher.

### Testing 
Add unit tests.

### DIFF
--- a/model/pod/dispatcher/db.go
+++ b/model/pod/dispatcher/db.go
@@ -1,11 +1,21 @@
 package dispatcher
 
 import (
+	"context"
+	"time"
+
+	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/db/mgo/bson"
+	"github.com/evergreen-ci/evergreen/model/event"
+	"github.com/evergreen-ci/evergreen/model/pod"
+	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/anser/bsonutil"
 	adb "github.com/mongodb/anser/db"
 	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 const Collection = "pod_dispatchers"
@@ -19,10 +29,9 @@ var (
 )
 
 // FindOne finds one dispatcher queue for the given query.
-// kim: TODO: test
-func FindOne(query bson.M) (*PodDispatcher, error) {
+func FindOne(q db.Q) (*PodDispatcher, error) {
 	var pd PodDispatcher
-	err := db.FindOneQ(Collection, db.Query(query), &pd)
+	err := db.FindOneQ(Collection, q, &pd)
 	if adb.ResultsNotFound(err) {
 		return nil, nil
 	}
@@ -30,18 +39,9 @@ func FindOne(query bson.M) (*PodDispatcher, error) {
 }
 
 // Find finds all dispatcher queues for the given query.
-// kim: TODO: test
-func Find(query bson.M) ([]PodDispatcher, error) {
+func Find(q db.Q) ([]PodDispatcher, error) {
 	pds := []PodDispatcher{}
-	return pds, errors.WithStack(db.FindAllQ(Collection, db.Query(query), &pds))
-}
-
-// FindOneByGroupID finds a pod dispatcher by its group ID.
-// kim: TODO: test
-func FindOneByGroupID(id string) (*PodDispatcher, error) {
-	return FindOne(bson.M{
-		GroupIDKey: id,
-	})
+	return pds, errors.WithStack(db.FindAllQ(Collection, q, &pds))
 }
 
 // UpdateOne updates one dispatcher queue.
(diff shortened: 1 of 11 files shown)


model/pod/dispatcher/dispatcher.go
```diff
@@ -1,15 +1,76 @@
...
+			ModificationCountKey: 1,
+		},
+	}
+}
+
+// UpsertAtomically inserts/updates the pod dispatcher depending on whether the
+// document already exists.
+func (pd *PodDispatcher) UpsertAtomically() (*adb.ChangeInfo, error) {
+	change, err := UpsertOne(pd.atomicUpsertQuery(), pd.atomicUpsertUpdate())
+	if err != nil {
```
julianedwards, line 71: Nit: this can be cleaned up a bit.
```
change, err := UpsertOne(pd.atomicUpsertQuery(), pd.atomicUpsertUpdate())
if err == nil {
        pd.ModificationCount++
}
return change, err
```

model/pod/dispatcher/db.go
```diff
@@ -0,0 +1,148 @@
...
+			pd = &newDispatcher
+		} else {
+			pd.PodIDs = append(pd.PodIDs, p.ID)
+
+			if !utility.StringSliceContains(pd.TaskIDs, t.Id) {
+				pd.TaskIDs = append(pd.TaskIDs, t.Id)
+			}
+		}
+
+		if _, err := env.DB().Collection(Collection).UpdateOne(sessCtx, pd.atomicUpsertQuery(), pd.atomicUpsertUpdate(), options.Update().SetUpsert(true)); err != nil {
```
julianedwards, line 102: Why not use `pd.UpsertAtomically` instead?

model/pod/dispatcher/db.go
```diff
@@ -0,0 +1,148 @@
...
+				task.StatusKey: evergreen.TaskContainerAllocated,
+			},
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "marking task as container allocated")
+		}
+		if update.ModifiedCount == 0 {
+			return nil, errors.New("task status was not updated")
+		}
+		t.Status = evergreen.TaskContainerAllocated
```
julianedwards, line 129: Is it ever possible that `t` will have `Activated` set to `false` and `PriorityKey` set to `evergreen.DisabledTaskPriority` in-mem but not in the DB? In which case, we should probably update those fields here as well.

units/pod_allocator.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := j.populate(); err != nil {
+		j.AddRetryableError(errors.Wrap(err, "populating job"))
+		return
+	}
+
+	if !j.task.ShouldAllocateContainer() {
+		return
+	}
+
+	intentPod, err := pod.NewTaskIntentPod(pod.TaskIntentPodOptions{
```
julianedwards, line 73: Does is make sense to do this in the `dispatcher.Allocate` transaction? Or is this already idempotent?

julianedwards (changes requested):

model/pod/dispatcher/db.go
Kimchelly: I had to write the entire transaction to use the mongo driver directly instead of Anser, because the Anser-based queries like `UpsertAtomically` don't accept contexts. If I used `UpsertAtomically` it would ignore the transaction session context.

model/pod/dispatcher/db.go
Kimchelly: They don't need to be updated because they're part of the query, not the update. In order for the transaction to succeed, the task has to have both of those fields be the same as they were when the task was first loaded from the DB to prove that the in-memory task state hasn't become stale.

model/pod/dispatcher/dispatcher.go
Kimchelly: I usually write my code to return from exception cases early. That way, the success case proceeds through the function linearly as if the exceptional if cases never happened.

units/pod_allocator.go
```diff
@@ -0,0 +1,112 @@
...
+		Base: job.Base{
+			JobType: amboy.JobType{
+				Name:    podAllocatorJobName,
+				Version: 0,
+			},
+		},
+	}
+}
+
+func NewPodAllocatorJob(taskID, ts string) amboy.Job {
```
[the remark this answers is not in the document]
Kimchelly: Done.

units/pod_allocator.go
Kimchelly: It shouldn't matter if this is in the transaction or not since this doesn't perform a DB operation. I tried keeping `Allocate` as simple as possible by only including the necessary transaction + the resulting `PodDispatcher`.

julianedwards (approved):

PR #5424, opened 2022-02-24

bynn:

### Description 
add hint to unschedule underwater tasks because using the wrong index was slowing it down 

### Testing 
ran the query on db 
unit tests

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -21,7 +21,7 @@ const (
 )
 
 var (
-	ActivatedTasksByDistroIndex = bson.D{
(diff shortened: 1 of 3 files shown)


model/task/db.go
```diff
@@ -20,6 +20,15 @@ const (
 	OldCollection = "old_tasks"
 )
 
+var (
+	ActivatedTasksByDistroIndex = bson.D{
```
ablack12, line 24: this can be a private variable since it's only used in model/task

ablack12 (approved):
LGTM with nit!

PR #5425, opened 2022-02-24
EVG-15928 Validate project config consistently and add version warnings

hadjri:
[EVG-15928](https://jira.mongodb.org/browse/EVG-15928)

### Description 
Not all places where project validation is performed were validating the project in the same way.  Now that ProjectConfigs are a part of a project's configuration file, places such as repotracker / CQ / patch intents should be validating it in the same way that the /validate POST request does so. Finally, I've re-added warnings to version creation (since the UI actually does have a built in way to display the version warnings and they were currently not being processed at all for a version).
### Testing 
Created new unit tests and modified existing ones to expect new errors / warnings (or modified the input to avoid these new errors in the tests).


validator/project_validator.go
```diff
@@ -246,9 +246,27 @@ func CheckProjectErrors(project *model.Project, includeLong bool) ValidationErro
...
 
+func CheckPatchedProjectConfigErrors(patchedProjectConfig string) ValidationErrors {
+	validationErrs := ValidationErrors{}
+	if len(patchedProjectConfig) <= 0 {
+		return validationErrs
+	}
+	projectConfig, err := model.CreateProjectConfig([]byte(patchedProjectConfig))
+	if err != nil {
+		validationErrs = append(validationErrs, ValidationError{
+			Message: fmt.Sprintf("can't retrieve aliases from patched config: %s", err.Error()),
```
ablack12, line 257: This error seems wrong

ablack12 (commented):
Can you verify in staging what happens with a version that has warnings? Does this actually get displayed in the UI somewhere (and where)?

hadjri:
> ntf("can't retrieve aliases from patched config: %

hadjri:
> Can you verify in staging what happens with a version that has warnings? Does this actually get displayed in the UI somewhere (and where)? I mostly want to make sure we don't balloon the UI by suddenly showing a ton of warnings.

Sure, I was able to find many old staging versions that have warnings, and they are displayed in the UI. Here is one example: https://evergreen-staging.corp.mongodb.com/version/mci_57fcf0eea21d2db781f3c40a0273111e8c4016ee

So it actually is a bit cluttering, and on the left hand side panel versions show a yellow triangle warning indicating they have warnings.  This PR is just re-instating this to work the way it used to (Chukwurado's project syntax / semantics reorganization into errors and warnings made it so that versions were briefly no longer checking warnings).

As a note it seems like this is only part of the legacy UI, haven't found anything on staging for these versions showing warnings.  I'm fine to make another ticket to de-clutter the legacy UI by making a dropdown in the case that there are a ton of warnings to remove clutter.

ablack12 (approved):

PR #5426, opened 2022-02-25
EVG-14450 add stepback to the commit queue

ablack12:
[EVG-14450](https://jira.mongodb.org/browse/EVG-14450)

### Description 
* Added stepback to the commit queue: this is described more in the ticket but if previous versions on the queue haven't run yet, we don't know if the failure is due to that commit or not, so dequeuing the current version may be the wrong thing to do, and then the user has to go through the process again.
* Lightly refactored some code to have less duplication and to possibly fix errors where we aren't restarting versions because we dequeue the item before looking for which versions to restart.

### Testing 
Added a test to cover all the scenarios I can think of, will also try to test at least one or two of them on staging.

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -13,14 +13,13 @@ import (
 	"time"
 
 	"github.com/99designs/gqlgen/graphql"
-	"github.com/evergreen-ci/evergreen/apimodels"
-	"github.com/evergreen-ci/evergreen/model/event"
-
 	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/apimodels"
 	"github.com/evergreen-ci/evergreen/cloud"
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/artifact"
 	"github.com/evergreen-ci/evergreen/model/commitqueue"
+	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/patch"
 	"github.com/evergreen-ci/evergreen/model/task"
@@ -598,7 +597,7 @@ type VersionModifications struct {
 	TaskIds           []string                  `json:"task_ids"` // deprecated
 }
 
-func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectRef, modifications VersionModifications) (int, error) {
+func ModifyVersion(version model.Version, user user.DBUser, modifications VersionModifications) (int, error) {
 	switch modifications.Action {
 	case Restart:
 		if modifications.VersionsToRestart == nil { // to maintain backwards compatibility with legacy Ui and support the deprecated restartPatch resolver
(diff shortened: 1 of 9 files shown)


graphql/util.go
```diff
@@ -598,7 +597,7 @@ type VersionModifications struct {
 	TaskIds           []string                  `json:"task_ids"` // deprecated
 }
 
-func ModifyVersion(version model.Version, user user.DBUser, proj *model.ProjectRef, modifications VersionModifications) (int, error) {
+func ModifyVersion(version model.Version, user user.DBUser, modifications VersionModifications) (int, error) {
```
hadjri, line 600: Does modifying this functions signature require you to run 'make gql' now?

model/task_lifecycle.go
```diff
@@ -727,21 +683,47 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
 	}
 	if p == nil {
-		return errors.Errorf("No patch for task")
+		return errors.New("patch not found")
 	}
-
-	if !p.IsCommitQueuePatch() {
-		return nil
+	if err := tryDequeueAndAbortCommitQueueVersion(p, *cq, t.Id, caller); err != nil {
+		return err
```
hadjri, line 712: How should we handle the situation if this fails after we've already successfully restarted CQ items after the current version? Maybe we should add a log explicitly stating that we tried to dequeue and abort the current version but it is still going to run (and fail).

hadjri (changes requested):
Can we test this on sandbox in staging?

graphql/util.go
ablack12: No because this is a helper function, and not specified in any schema.

model/task_lifecycle.go
ablack12: This would only happen in the event that there's a database error and isn't a new behavior, and typically returning an error assumes it will get propagated to something that will log it (either to the user or to splunk at the tope level) so I'm not super concerned with adding another log here.

graphql/util.go
malikchaya2: make gql should be run if either schema.graphql, resolvers.go, or, I believe gqlgen.yml are touched.

model/task_lifecycle.go
```diff
@@ -727,21 +676,47 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
+		if err != nil {
+			return errors.Wrapf(err, "can't get commit queue for id '%s'", t.Project)
+		}
+		if cq == nil {
+			return errors.Errorf("no commit queue found for '%s'", t.Project)
+		}
+	}
+	// this must be done before dequeuing so that we know which entries to restart
+	if err := RestartItemsAfterVersion(cq, t.Project, t.Version, caller); err != nil {
+		return errors.Wrap(err, "unable to restart versions")
```
malikchaya2, line 694: Should we maybe add more details to this error?

malikchaya2: I'm thinking the user may not even know why anything is being restarted, and this error is not wrapped.

malikchaya2 (commented):
I'd also feel more comfortable once I know that this passes on staging.

ablack12:
OKAY so I added three failing patches to the commit queue on staging:
[1st](https://spruce-staging.corp.mongodb.com/version/621e424097b1d31c21651530/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (fail immediately), [2nd](https://spruce-staging.corp.mongodb.com/version/621e42569dbe3229f8b3f6b4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (fail after some time), and [3rd](also fail immediately).

The 1st CQ task failed first and was dequeued immediately, since nothing else could've caused the failure.
The 3rd CQ task had finished before the 1st one, but hadn't been taken off the queue even though it failed, because the first one hadn't finished running. It was restarted accordingly when the 1st one failed.
This behavior repeated when the second task failed.

Overall testing showed that:
1) regular commit queue merges still go through,
2) commit queue patches do eventually get removed from the queue even if they're waiting on previous tasks, and
3) dequeueing previous commit queue patches do restart later patches that were waiting on that one to finish.

hadjri (approved):
Awesome!

PR #5430, opened 2022-02-25
EVG-16338 error saving a project if github conflicts exist

ablack12:
[EVG-16338](https://jira.mongodb.org/browse/EVG-16338)

### Description 
The UI handles this but we're adding to the backend in case things race and just to be careful. Enabling a project or a toggle, or changing owner/repo/branch are things that could affect this.
### Testing 
Added a test.

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/user"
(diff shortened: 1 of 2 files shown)


rest/data/project_settings.go
```diff
@@ -202,3 +229,30 @@ func (sc *DBConnector) SaveProjectSettingsForSection(ctx context.Context, projec
 	}
 	return &res, errors.Wrapf(catcher.Resolve(), "error saving section '%s'", section)
 }
+
+func handleGithubConflicts(pRef *model.ProjectRef, reason string) error {
+	if !pRef.IsPRTestingEnabled() && !pRef.CommitQueue.IsEnabled() && !pRef.IsGithubChecksEnabled() {
+		return nil // if nothing is toggled on, then there's no reason to look for conflicts
```
malikchaya2, line 235: good comment :)

malikchaya2 (approved):
🎉

malikchaya2:
It looks like the graphql tests are failing because the items in the list for projectSettings are not in the same order as the test is expecting. @khelif96 I feel like we've dealt with that before and it was an easy fix, but I don't remember what the fix was. Do you?

khelif96:
> It looks like the graphql tests are failing because the items in the list for projectSettings are not in the same order as the test is expecting. @khelif96 I feel like we've dealt with that before and it was an easy fix, but I don't remember what the fix was. Do you?

If this is different for every test run the fix would be to add a sort. It also will make things more consistent on the UI since we can avoid a different ordering every time someone refreshes the page.

ablack12:
Sigh -- this list actually just augments something else so it being sorted doesn't really matter. I'll either change the test to have one variable or add a sort but it's really not a necessary one.

ablack12:
Actually I just looked closer -- this error has nothing to do with my changes and is just flaky on the waterfall. @khelif96 I think it makes the most sense to open a different ticket to address this.

ablack12:
FYI Bynn is fixing in https://github.com/evergreen-ci/evergreen/pull/5435/files

PR #5434, opened 2022-02-25

bynn:

### Description 
add hint to unschedule underwater tasks because using the wrong index was slowing it down 

### Testing 
ran the query on db 
unit tests


ablack12 (approved):
Note: in the future it's best to update reverts by making a new branch, pulling main, and then reverting the reverted commit, so it's clear which commits are part of the review.

PR #5435, opened 2022-02-25
EVG-16048 Unhide admin only vars in project page

bynn:
[EVG-16048](https://jira.mongodb.org/browse/EVG-16048)

### Description 
Delete ng-show 
var resolver now returns sorted list

### Testing 

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -498,7 +498,6 @@ func (r *projectVarsResolver) PrivateVars(ctx context.Context, obj *restModel.AP
 			res = append(res, utility.ToStringPtr(privateAlias))
 		}
 	}
-	sort.Slice(res, func(i, j int) bool { return utility.FromStringPtr(res[i]) < utility.FromStringPtr(res[j]) })
 	return res, nil
 }
 
@@ -509,7 +508,6 @@ func (r *projectVarsResolver) AdminOnlyVars(ctx context.Context, obj *restModel.
 			res = append(res, utility.ToStringPtr(varAlias))
 		}
 	}
-	sort.Slice(res, func(i, j int) bool { return utility.FromStringPtr(res[i]) < utility.FromStringPtr(res[j]) })
 	return res, nil
 }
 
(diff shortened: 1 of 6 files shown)


graphql/tests/projectSettings/results.json
```diff
@@ -119,8 +119,8 @@
         "data": {
           "projectSettings": {
             "vars": {
-              "vars": {"hello": "", "foo":  "bar", "admintest": "woo", "both": ""},
-              "privateVars": ["hello", "both"],
+              "vars": {"admintest": "woo", "both": "", "hello": "", "foo":  "bar"},
+              "privateVars": ["both", "hello"],
```
ablack12, line 123: This isn't guaranteed to ever be ordered because of how graphql works, so just flipping them isn't going to help. If we add a sort to the privateVars and adminOnly resolvers, then the tests will be guaranteed to return in a consistent order.

bynn: returns sorted

graphql/resolvers.go
```diff
@@ -508,6 +509,7 @@ func (r *projectVarsResolver) AdminOnlyVars(ctx context.Context, obj *restModel.
 			res = append(res, utility.ToStringPtr(varAlias))
 		}
 	}
+	sort.Slice(res, func(i, j int) bool { return utility.FromStringPtr(res[i]) < utility.FromStringPtr(res[j]) })
```
ablack12, line 512: does sort.Sort work instead?

bynn: It doesn't like that they're pointers

graphql/tests/projectSettings/results.json
ablack12: Ths stilaThis still has two returned private_vars, which means that order is still going to matter. Should both not be part of these anymore?

bynn: Sorry I left it in there

ablack12 (approved):
LGTM if tests pass

PR #5438, opened 2022-02-28
EVG-16339 Conditionally enable buildVariantDisplayName in query

khelif96:
[EVG-16339](https://jira.mongodb.org/browse/EVG-16339)

### Description 
BuildVariantDisplayName will only get added to the GetTasksByVersion Pipeline if it is explicitly needed.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -1284,8 +1284,9 @@ func (r *patchResolver) TaskStatuses(ctx context.Context, obj *restModel.APIPatc
 		{Key: task.DisplayNameKey, Order: 1},
 	}
 	opts := data.TaskFilterOptions{
-		Sorts:            defaultSort,
-		IncludeBaseTasks: false,
+		Sorts:                          defaultSort,
+		IncludeBaseTasks:               false,
+		IncludeBuildVariantDisplayName: false,
 	}
 	tasks, _, err := r.sc.FindTasksByVersion(*obj.Id, opts)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


minnakt (commented):
I think some logic may be off because I can't see the build variant names anymore on this version  (running on `dev`). 🤔 I checked and they are visible on the main branch

Screen Shot 2022-03-01 at 1 04 50 PM

graphql/resolvers.go
```diff
@@ -3812,8 +3821,9 @@ func (r *versionResolver) BuildVariants(ctx context.Context, v *restModel.APIVer
 			{Key: task.DisplayNameKey, Order: 1},
 		}
 		opts := data.TaskFilterOptions{
-			Sorts:            defaultSort,
-			IncludeBaseTasks: true,
+			Sorts:                          defaultSort,
+			IncludeBaseTasks:               true,
+			IncludeBuildVariantDisplayName: false, // we don't need to include buildVariantDisplayName here because this is only used to determine if a task has been activated
```
minnakt, line 3826: This is more of an observation rather than something to fix, but I kind of wonder why we don't use the function `setVersionActivationStatus` from `graphql/util.go` here. If you put them side by side, they're almost identical 

Screen Shot 2022-03-01 at 3 12 16 PM

minnakt (approved):
Clicked through Spruce and everything looked okay! I also like that the conditional is gone now 🤣 

Great job!!! 🐛

graphql/resolvers.go
khelif96: @malikchaya2  and I observed this as well. tbh they should live in the same spot and it was an oversight.  I opened this ticket earlier which should fix this issue. https://jira.mongodb.org/browse/EVG-16376

minnakt: sounds good to me 😇

PR #5440, opened 2022-02-28
EVG-16415 Use project id instead of identifier for history queries

khelif96:
[EVG-16415](https://jira.mongodb.org/browse/EVG-16415)

### Description 
I made the dumb mistake and used project identifier to find tasks instead of the id and wrongfully blamed matrices. Crisis averted.

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3542,7 +3542,7 @@ func (r *queryResolver) TaskNamesForBuildVariant(ctx context.Context, projectId 
 	if repo == nil {
 		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("could not find repository '%s'", pid))
 	}
-	buildVariantTasks, err := task.FindTaskNamesByBuildVariant(projectId, buildVariant, repo.RevisionOrderNumber)
+	buildVariantTasks, err := task.FindTaskNamesByBuildVariant(pid, buildVariant, repo.RevisionOrderNumber)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


minnakt (approved):
LGTM, but I think it might be worth changing `projectId` param name to `projectIdentifier` just to be 100% clear and avoid confusion in the future 🐛

Looking at examples ([patches](https://evergreen-staging.corp.mongodb.com/rest/v2/patches/62047bd197b1d3461f3877c0/raw), [projects](https://evergreen-staging.corp.mongodb.com/rest/v2/projects)), it seems like we default to `identifier` when we mean the project "name" that appears in a URL. It might be good to be consistent with that 🤔

ablack12 (approved):

khelif96:
> LGTM, but I think it might be worth changing `projectId` param name to `projectIdentifier` just to be 100% clear and avoid confusion in the future 🐛
> 
> Looking at examples ([patches](https://evergreen-staging.corp.mongodb.com/rest/v2/patches/62047bd197b1d3461f3877c0/raw), [projects](https://evergreen-staging.corp.mongodb.com/rest/v2/projects)), it seems like we default to `identifier` when we mean the project "name" that appears in a URL. It might be good to be consistent with that 🤔

That SGTM i opened a [ticket](https://jira.mongodb.org/browse/EVG-16434) to make that change. Unfortunately there is no way to currently update the parameter name breaking backwards compatibility. It will need to happen in a series of prs and deploys.

PR #5442, opened 2022-03-01
EVG-6496 Do not let autoupdate ever return an error

hadjri:

### Description 
When retrieving client settings, if autoupdate fails an error will no longer be returned.

### DIFF
--- a/operations/before.go
+++ b/operations/before.go
@@ -1,6 +1,7 @@
 package operations
 
 import (
+	"context"
 	"os"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


operations/before.go
```diff
@@ -105,9 +105,9 @@ var (
 		confPath := c.String("conf")
```
ablack12, line 105: You mentioned that for the BF involved, the path given was the empty string. Does that mean confPath is empty, and can we return early in that case? (i.e. just return no error at all)

hadjri: Yep, that should be fine

ablack12 (approved):

PR #5443, opened 2022-03-01
EVG-16435: Add DefaultSectionToRepo mutation

sophstad:
[EVG-16435](https://jira.mongodb.org/browse/EVG-16435)

### Description 
- Add GraphQL mutation for calling `DefaultSectionToRepo`

### Testing 
- Added unit tests
- Tested manually in GraphQL playground

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -2444,6 +2444,15 @@ func (r *mutationResolver) SaveRepoSettingsForSection(ctx context.Context, obj *
 	return changes, nil
 }
 
+func (r *mutationResolver) DefaultSectionToRepo(ctx context.Context, projectId string, section ProjectSettingsSection) (*string, error) {
+	usr := MustHaveUser(ctx)
+	if err := model.DefaultSectionToRepo(projectId, model.ProjectPageSection(section), usr.Username()); err != nil {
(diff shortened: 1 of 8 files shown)


graphql/resolvers.go
```diff
@@ -2444,6 +2444,15 @@ func (r *mutationResolver) SaveRepoSettingsForSection(ctx context.Context, obj *
 	return changes, nil
 }
 
+func (r *mutationResolver) DefaultSectionToRepo(ctx context.Context, projectId string, section ProjectSettingsSection) (*string, error) {
+	usr := MustHaveUser(ctx)
+	if err := model.DefaultSectionToRepo(projectId, model.ProjectPageSection(section), usr.Username()); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error defaulting to repo for section: %s", err.Error()))
+	}
+
+	return &projectId, nil
```
ablack12, line 2453: What's the purpose of returning just the project ID? Would it be more useful to now return the updated project?

ablack12 (approved):
LGTM with question

graphql/resolvers.go
sophstad: Returning the updated project wouldn't really be helpful to the front end; after the mutation it will refetch project settings since managing a cache for this data has proven too complex. We could return a bool or something instead to indicate success, but there's no need to construct the new project when it won't be used by the UI.

PR #5445, opened 2022-03-01
EVG-16239: rename host task dispatch functions

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16239

### Description 
Rename some dispatch/undispatch task functions that only apply to tasks running in hosts. Instead of "undispatched", tasks that are waiting to run on a container use the "container-unallocated" status. Tasks that are sent to a container to run will still enter the "dispatched" state, but will require slightly different logic.

### Testing 
N/A

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -854,11 +854,12 @@ func (t *Task) cacheExpectedDuration() error {
 	)
 }
 
-// Mark that the task has been dispatched onto a particular host. Sets the
-// running task field on the host and the host id field on the task. If the host
-// is part of a display task, the display task is also marked as dispatched to a
-// host. Returns an error if any of the database updates fail.
-func (t *Task) MarkAsHostDispatched(hostId, distroId, agentRevision string, dispatchTime time.Time) error {
+// MarkAsHostDispatched marks that the task has been dispatched onto a
+// particular host. If the task is part of a display task, the display task is
+// also marked as dispatched to a host. Returns an error if any of the database
+// updates fail.
+func (t *Task) MarkAsHostDispatched(hostId, distroId, agentRevision string,
+	dispatchTime time.Time) error {
 	t.DispatchTime = dispatchTime
 	t.Status = evergreen.TaskDispatched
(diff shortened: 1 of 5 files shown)


model/task/task.go
```diff
@@ -855,9 +855,10 @@ func (t *Task) cacheExpectedDuration() error {
 }
 
 // Mark that the task has been dispatched onto a particular host. Sets the
-// running task field on the host and the host id field on the task.
-// Returns an error if any of the database updates fail.
-func (t *Task) MarkAsDispatched(hostId, distroId, agentRevision string, dispatchTime time.Time) error {
+// running task field on the host and the host id field on the task. If the host
```
julianedwards, line 858: `MarkAsHostDispatched marks...`

julianedwards (changes requested):

model/task/task.go
Kimchelly: Fixed. I also tweaked the comment slightly since the doc comment seems to be outdated/incorrect.

model/task_lifecycle.go
```diff
@@ -1185,9 +1185,12 @@ func MarkStart(t *task.Task, updates *StatusChanges) error {
 	return nil
 }
 
-func MarkTaskUndispatched(t *task.Task) error {
+// MarkHostTaskUndispatched marks a task as no longer dispatched to a host. If
+// it's part of a display task, update the display task as necessary as a result
```
julianedwards, line 1189: Does this mean that the caller should update the display task as necessary? Or does this function do take care of that? Wording is a bit confusing.

Kimchelly: It calls [this function](https://github.com/Kimchelly/evergreen/blob/33f5245e7f852c9b7c6dde6c85631025ada493db/model/task_lifecycle.go#L1462-L1569) to do a bunch of display task metadata updates that result from marking the task undispatched. I simplified the wording a bit.

julianedwards: Thanks!

PR #5447, opened 2022-03-02
EVG-16196 longer debounce

ybrill:
[EVG-16196](https://jira.mongodb.org/browse/EVG-16196)

### Description 
I noticed that as I was typing a bv filter on the legacy waterfall almost every keystroke made another request for waterfall data (and the results of all but the last are discarded). 
Screen Shot 2022-03-02 at 11 08 10 AM

It makes more sense to me to only request the data when the user is done typing. The debounce function is supposed to do this, but 100ms is too short. This PR extends it to a second.

### Testing 
When I deployed this to staging the wait after I finished typing seemed appropriate.


Kimchelly (approved):
Nice catch! 💯

ablack12 (approved):

PR #5448, opened 2022-03-02
EVG-16447 Update admin only variable access to mainline commits

bynn:
[EVG-16447](https://jira.mongodb.org/browse/EVG-16447)

### Description 
allow repotracker requesters to access admin only vars

### Testing 
took away minna's admin role and verified that her patches cant access admin vars but commits can 
https://evergreen-staging.corp.mongodb.com/task/sandbox_ubuntu1604_bynntask_14619adff2ad781b915772f1a1a5107e9f397ff9_22_03_03_17_02_44

### DIFF
--- a/model/project_vars.go
+++ b/model/project_vars.go
@@ -10,6 +10,7 @@ import (
 	"github.com/evergreen-ci/gimlet"
 	"github.com/mongodb/anser/bsonutil"
 	adb "github.com/mongodb/anser/db"
+	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 )
@@ -220,6 +221,7 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
 	return restrictedVars
 }
 
+// GetAdminOnlyVars will change to GetVars on EVG-16045 after removing restricted vars.
 func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) (map[string]string, error) {
 	adminOnlyVars := map[string]string{}
 	if evergreen.IsSystemActivator(t.ActivatedBy) {
@@ -228,15 +230,9 @@ func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) (map[string]strin
 				adminOnlyVars[k] = v
 			}
 		}
-	} else {
-		var u *user.DBUser
-		var err error
-		if t.ActivatedBy != "" {
-			u, err = user.FindOneById(t.ActivatedBy)
-		}
-		if err != nil {
-			return nil, err
-		}
+	} else if t.ActivatedBy != "" {
+		u, err := user.FindOneById(t.ActivatedBy)
+		grip.Error(err)
 		if u != nil {
 			isAdmin := u.HasPermission(gimlet.PermissionOpts{
 				Resource:      t.Project,
--- a/model/project_vars.go
+++ b/model/project_vars.go
@@ -3,7 +3,11 @@ package model
 import (
 	"fmt"
 
+	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/db"
+	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/evergreen/model/user"
+	"github.com/evergreen-ci/gimlet"
 	"github.com/mongodb/anser/bsonutil"
 	adb "github.com/mongodb/anser/db"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


service/api.go
```diff
@@ -345,21 +345,27 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
+			u, err = user.FindOneById(t.ActivatedBy)
+		}
+		if err != nil {
+			as.LoggedError(w, r, http.StatusInternalServerError, err)
+			return
+		}
+		if u != nil {
+			// check if user is an admin
+			if isAdmin(u, t.Project) {
+				for key, val := range projectVars.GetAdminOnlyVars() {
```
ablack12, line 366: Can we move all of this logic into GetAdminOnlyVars to make this function more modular? (i.e. it only returns vars from the function if the version/task is right)

ablack12 (commented):
You can write unit tests for GetAdminOnlyVars after moving the code to that function, but I would still suggest testing on staging :) you can ask someone without admin roles to commit something to commit-queue-sandbox when your changes are deployed if you don't want to remove your own roles.

service/api.go
bynn: done

malikchaya2 (approved):

service/api.go
```diff
@@ -384,6 +375,28 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
 	gimlet.WriteJSON(w, res)
 }
 
+func accessAdminVars(t *task.Task, v *model.Version) (bool, error) {
```
ablack12, line 378: I was thinking instead that GetAdminOnlyVars() could handle this. It looks like it would just need to take in the task, since task also has the requester field. (I think this should check IsSystemActivator instead of just repotracker). This would make it harder for future developers to add AdminOnlyVars accidentally.

bynn: got rid of the stepback logic because it should automatically haev access if its system activated 
isadmin function cant be imported so i replaced it

model/project_vars.go
```diff
@@ -216,15 +220,41 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
...
-	for k, v := range projectVars.Vars {
-		if projectVars.AdminOnlyVars[k] {
-			adminOnlyVars[k] = v
+	if evergreen.IsSystemActivator(t.ActivatedBy) {
+		for k, v := range projectVars.Vars {
+			if projectVars.AdminOnlyVars[k] {
+				adminOnlyVars[k] = v
+			}
+		}
+	} else {
```
ablack12, line 231: can this just be an else if t.ActivatedBy != ""? Nothing else in the function is touched if t.ActivatedBy is empty

model/project_vars.go
```diff
@@ -216,15 +220,41 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
...
+		var u *user.DBUser
+		var err error
+		if t.ActivatedBy != "" {
+			u, err = user.FindOneById(t.ActivatedBy)
+		}
+		if err != nil {
```
ablack12, line 237: Le's not error in this case. We can just not give them permissions and log the error.

service/api.go
```diff
@@ -345,25 +344,15 @@ func (as *APIServer) FetchExpansionsForTask(w http.ResponseWriter, r *http.Reque
...
 		return
 	}
-	var u *user.DBUser
-	if t.ActivatedBy == evergreen.StepbackTaskActivator {
-		u, err = user.FindOneById(v.Author)
-	} else if t.ActivatedBy != "" {
-		u, err = user.FindOneById(t.ActivatedBy)
-	}
+
+	adminVars, err := projectVars.GetAdminOnlyVars(t)
```
ablack12, line 348: There's actually not really a point to having this be separate from GetUnrestrictedVars, since we end up looping twice, and when we get rid of restricted vars it can then just be a general "GetVars" funtion.

model/project_vars.go

service/api.go
bynn: added a comment to GetAdminOnly and GetUnrestricted also has a comment to be changed on 16045

model/project_vars.go
bynn: gripped error because i wasnt super sure on how to log it

model/project_vars.go
```diff
@@ -216,15 +221,36 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
 	return restrictedVars
 }
 
-func (projectVars *ProjectVars) GetAdminOnlyVars() map[string]string {
+// GetAdminOnlyVars will change to GetVars on EVG-16045 after removing restricted vars.
+func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) (map[string]string, error) {
```
ablack12, line 225: This doesn't return any errors anymore so let's remove it from the function signature

model/project_vars.go
```diff
@@ -216,15 +221,36 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
...
-			adminOnlyVars[k] = v
+	if evergreen.IsSystemActivator(t.ActivatedBy) {
+		for k, v := range projectVars.Vars {
+			if projectVars.AdminOnlyVars[k] {
+				adminOnlyVars[k] = v
+			}
+		}
+	} else if t.ActivatedBy != "" {
+		u, err := user.FindOneById(t.ActivatedBy)
+		grip.Error(err)
```
ablack12, line 235: Let's log with message.Fields and include some additional information (taskId, activatedBy, etc)

model/project_vars.go
```diff
@@ -216,15 +221,36 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
...
 
-func (projectVars *ProjectVars) GetAdminOnlyVars() map[string]string {
+// GetAdminOnlyVars will change to GetVars on EVG-16045 after removing restricted vars.
+func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) (map[string]string, error) {
 	adminOnlyVars := map[string]string{}
-
-	for k, v := range projectVars.Vars {
-		if projectVars.AdminOnlyVars[k] {
-			adminOnlyVars[k] = v
+	if evergreen.IsSystemActivator(t.ActivatedBy) {
```
ablack12, line 227: Wait sorry this isn't right for mainline commits; this just checks if the person who activated the task was not a person but was Evergreen. You want to check that the requester is in this list. https://github.com/evergreen-ci/evergreen/blob/f7284bd30cef2f5f12d976983406ee9e718e47bc/globals.go#L485 Sorry if I said the wrong thing.

model/project_vars.go

model/project_vars.go
bynn: ahh okay

model/project_vars.go

model/project_vars.go
ablack12: grip.Error(message.WrapError(err, message.Fields{
"message": ...,
"task_id": ...
}))

model/project_vars.go
```diff
@@ -216,14 +223,41 @@ func (projectVars *ProjectVars) GetRestrictedVars() map[string]string {
 	return restrictedVars
 }
 
-func (projectVars *ProjectVars) GetAdminOnlyVars() map[string]string {
+// GetAdminOnlyVars will change to GetVars on EVG-16045 after removing restricted vars.
+func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) map[string]string {
```
ablack12, line 227: Can we also add a comment explaining who this returns variables for? (mainline commits + project admins)

bynn: done

model/project_vars.go
bynn: done

ablack12 (approved):

PR #5451, opened 2022-03-03
EVG-15593 untangle service and graphql packages

ablack12:
[EVG-15593](https://jira.mongodb.org/browse/EVG-15593 )

### Description 
This PR looks more painful than it is -- no functionality is changed so review should be limited to the refactor itself.
1) Any functions that are used by both the graphql and service packages shouldn't exist in graphql/service.go, since these functions aren't truly graphql specific.
2) For functions I moved, modified the function signatures to be more typical. (Many return status code and graphql code and error and success, but typically success can be inferred from error or status, and a function exists to map http status to graphl code.

STILL TODO: there's a graphql Handler that's _only_ used in service, but it uses Generated types which confuses me a bit. Will try to untangle in a future commit or PR. 

### Testing 
Checking that all existing tasks pass. Will also click buttons in staging to verify generally that things are good.


rest/data/volume.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/amboy"
+	"github.com/pkg/errors"
+)
+
+var errHostStatusChangeConflict = errors.New("conflicting host status modification is in progress")
+
+// To be moved to a better home when we restructure the resolvers.go file
+// TerminateSpawnHost is a shared utility function to terminate a spawn host
+func TerminateSpawnHost(ctx context.Context, env evergreen.Environment, u *user.DBUser, h *host.Host) (int, error) {
```
ablack12, line 21: I recognize that it's unusual to have functions in rest/data that aren't in the connector; however, since we intend to remove the connector, I think functions that are used between service/graphql/etc will still need a neutral home. (This can't live in the host package because of import cycles with unit)

units/schedule_patch.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/model"
+	"github.com/evergreen-ci/evergreen/model/patch"
+	"github.com/pkg/errors"
+)
+
+// SchedulePatch schedules a patch. It returns an error and an HTTP status code.
+func SchedulePatch(ctx context.Context, patchId string, version *model.Version, patchUpdateReq model.PatchUpdate) (int, error) {
```
ablack12, line 14: Tried to move a lot of this logic into model, but overall it couldn't be moved because of a dependency cycle with units. Not sure if this should exist in rest/data instead.

cloud/spawn.go
```diff
@@ -243,7 +228,25 @@ func CheckInstanceTypeValid(ctx context.Context, d distro.Distro, requestedType
 	return nil
 }
 
-func SetHostRDPPassword(ctx context.Context, env evergreen.Environment, host *host.Host, password string) error {
+// SetHostRDPPassword is a shared utility function to change the password on a windows host
+func SetHostRDPPassword(ctx context.Context, env evergreen.Environment, h *host.Host, pwd string) (int, error) {
```
malikchaya2, line 232: Why is this part of this refactor if it's not in a service or graphql package?

graphql/integration_atomic_test.go
```diff
@@ -21,9 +21,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/evergreen-ci/evergreen/graphql"
```
malikchaya2, line 24: The previous order seems more correct

rest/data/volume.go
malikchaya2: it feels strange to me to have spawnHost functions in volume.go since they are logically different. can we maybe create a rest/data/spawnHost.go?

graphql/integration_atomic_test.go
ablack12: Yeah I blame my editor

cloud/spawn.go
ablack12: All of this validation was previously in graphql but was called UpdateHostPassword. I liked this naming better so I just swapped which one was the public usable function name and which was private.

malikchaya2: ah nice!

ablack12:
Update: it does make sense for service to depend on graphql, since it's itself a route, but then graphql shouldn't be depending on service. I discussed with Brian; because it uses servers for integration tests, it really ultimately is a service test, so I moved the part of it that was problematic over to service, and moved some of the test functions around in graphql so that they could be called in both service and graphql test files. We no longer have the graphql_test package!

service/graphql_integration_test.go
```diff
@@ -0,0 +1,35 @@
+package service
```
malikchaya2, line 1: a graphql test in the service package hurts my eyes, but I understand the reasoning behind it 😆

malikchaya2 (approved):

ablack12:
@malikchaya2 I ended up needing to move some more stuff; if you want to re-review.

ablack12:
I made a new package service-graphql and it works except that it fails because no test results are attached? (This is true for lint and not lint). @khelif96 do you have any ideas for what this could be about? 🤔  Also [TestMakeTicket](https://evergreen.mongodb.com/test_log/evergreen_ubuntu1604_test_rest_data_patch_9f9905a8cd2ce35dccae58b354459fb2d7824369_622266a2c9ec442b7b000c77_22_03_04_19_21_07/0?test_name=output.rest-data.test&group_id=#L57) fails for a context reason which I've never seen before.

khelif96:
> 

I'm not sure but one of the tasks seems to be failing because of an unused variable. https://spruce.mongodb.com/task/evergreen_lint_lint_graphql_patch_9f9905a8cd2ce35dccae58b354459fb2d7824369_62262841a4cf4774480eec11_22_03_07_15_44_03/logs?execution=0&sortBy=STATUS&sortDir=ASC

ablack12:
That's actually a different task and was due to the merge conflict. I was referring to the new test-service-graphql. I'll ask  around to see how attaching test results works.

testutil/server.go
```diff
@@ -0,0 +1 @@
+package testutil
```
khelif96, line 1: Is this file supposed to be empty?

khelif96 (commented):
Great Refactor! just had 2 questions.
Could you also run the spruce e2e tests against this branch you can do so in a patch using the `evergreen set-module` command you showed me.  I just want to make sure everything still works.

ablack12:
@Kimchelly could you review the makefile change?

Kimchelly (approved):
LGTM the makefile change.

khelif96 (approved):

PR #5452, opened 2022-03-03
EVG-16291: check for context errors during host status check

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16291

### Description 
If the cloud host status check fails because the context errors, the host should not be marked as terminated in the DB. The job will be re-enqueued to retry terminating the host later.

### Testing 
There's no tests due to the difficulty of testing context cancellation behavior at a specific line in the job, but it's not difficult to verify correctness just by inspection. Testing with the mock provider is difficult since it ignores the context and doesn't make network requests, but using a real cloud provider would also be difficult to simulate.

### DIFF
--- a/units/host_termination.go
+++ b/units/host_termination.go
@@ -361,7 +361,7 @@ func (j *hostTerminationJob) checkAndTerminateCloudHost(ctx context.Context, old
 
 	cloudStatus, err := cloudHost.GetInstanceStatus(ctx)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


julianedwards (approved):

Kimchelly:
I removed the test since it's difficult to trigger this particular context error check, and it probably doesn't make senes to re-organize the code just to test this one error case.

PR #5453, opened 2022-03-03
EVG-16127: build Amboy indexes for local evergreen

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16127

### Description 
The current theory is that local evergreen is slow because of Amboy's background operations, which make unindexed queries to mongod. The longer that it runs, the more Amboy jobs (such as the cron populators) would be added to the remote queue and queue group collections, which would make the queries slower over time.

* Change the local evergreen data population script to build the Amboy indexes.
* Add necessary admin config docs to ensure that local evergreen uses the local Amboy DB/collections.
* Clean up tech debt in the data population script so that it no longer passes around a catcher to accumulate errors.

### Testing 
I ran evergreen locally and confirmed that it set up the indexes and used the correct Amboy DB/collections.


config_amboy.go
```diff
@@ -122,10 +122,13 @@ func (c *AmboyConfig) Set() error {
 }
 
 const (
+	// DefaultAmboyQueueName is the default name prefix for the Amboy remote
```
julianedwards, line 125: Should this just say `default prefix`?

cmd/load-smoke-data/load-smoke-data.go
```diff
@@ -81,76 +86,145 @@ func insertFileDocsToDb(ctx context.Context, fn string, catcher grip.Catcher, db
...
+		dbName      string
+		logsDBName  string
+		amboyDBName string
 	)
 
 	flag.StringVar(&path, "path", filepath.Join(wd, "testdata", "smoke"), "load data from json files from these paths")
 	flag.StringVar(&dbName, "dbName", "mci_smoke", "database name for directory")
-	flag.StringVar(&logsDbName, "logsDbName", "logs", "logs database name for directory")
+	flag.StringVar(&logsDBName, "logsDBName", "logs", "logs database name for directory")
+	flag.StringVar(&amboyDBName, "amboyDBName", "amboy_smoke", "")
```
julianedwards, line 219: Should this have documentation?

Kimchelly: Yes, you're right. Added.

config_amboy.go
Kimchelly: I clarified it to say "namespace". I think if I don't state what prefix it's referring to, it might just be more confusing about what prefixes have to do with Amboy for someone unfamiliar with it.

julianedwards (approved):

PR #5455, opened 2022-03-03
EVG-15171 Implement restart-failed

hadjri:
[EVG-15171](https://jira.mongodb.org/browse/EVG-15171)

### Description 
Introduce repeat-failed, a flag on the CLI patch command that will repeat only the failed tasks and build variants from the most recent patch (if any).
### Testing 
Tested in staging on sandbox project when the most recent patch was failed, and saw the previously failed tasks were automatically selected.

### DIFF
--- a/model/patch/db.go
+++ b/model/patch/db.go
@@ -305,7 +305,7 @@ func PatchesByProject(projectId string, ts time.Time, limit int) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(limit)
 }
 
-// FindFailedCommitQueuePatchesInTimeRange returns failed patches if they started within range,
+// FindFailedCommitQueuePatchesinTimeRange returns failed patches if they started within range,
 // or if they were never started but finished within time range. (i.e. timed out)
 func FindFailedCommitQueuePatchesinTimeRange(projectID string, startTime, endTime time.Time) ([]Patch, error) {
 	query := bson.M{
--- a/model/patch/db.go
+++ b/model/patch/db.go
@@ -305,9 +305,9 @@ func PatchesByProject(projectId string, ts time.Time, limit int) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(limit)
 }
 
-// FindFailedCommitQueuePatchesinTimeRange returns failed patches if they started within range,
+// FindFailedCommitQueuePatchesInTimeRange returns failed patches if they started within range,
 // or if they were never started but finished within time range. (i.e. timed out)
-func FindFailedCommitQueuePatchesinTimeRange(projectID string, startTime, endTime time.Time) ([]Patch, error) {
+func FindFailedCommitQueuePatchesInTimeRange(projectID string, startTime, endTime time.Time) ([]Patch, error) {
 	query := bson.M{
 		ProjectKey: projectID,
 		StatusKey:  evergreen.PatchFailed,
--- a/operations/patch.go
+++ b/operations/patch.go
@@ -62,7 +62,7 @@ func getPatchFlags(flags ...cli.Flag) []cli.Flag {
 				Usage: "use all of the same tasks/variants defined for the last patch you scheduled for this project",
 			},
 			cli.BoolFlag{
-				Name:  joinFlagNames(repeatFailedDefinitionFlag, "failed", "rf"),
+				Name:  joinFlagNames(repeatFailedDefinitionFlag, "rf"),
 				Usage: "use only the failed tasks/variants defined for the last patch you scheduled for this project",
 			},
 			cli.StringFlag{
--- a/operations/patch.go
+++ b/operations/patch.go
@@ -14,10 +14,11 @@ import (
 )
 
 const (
-	patchDescriptionFlagName = "description"
-	patchVerboseFlagName     = "verbose"
-	patchTriggerAliasFlag    = "trigger-alias"
-	reuseDefinitionFlag      = "reuse"
+	patchDescriptionFlagName   = "description"
+	patchVerboseFlagName       = "verbose"
+	patchTriggerAliasFlag      = "trigger-alias"
+	reuseDefinitionFlag        = "reuse"
+	repeatFailedDefinitionFlag = "repeat-failed"
 )
 
 func getPatchFlags(flags ...cli.Flag) []cli.Flag {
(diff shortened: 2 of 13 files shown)


units/patch_intent.go
```diff
@@ -483,6 +491,57 @@ func (j *patchIntentProcessor) getPreviousPatchDefinition(project *model.Project
 	return res, nil
 }
 
+func (j *patchIntentProcessor) getPreviousPatchFailures(project *model.Project) ([]patch.VariantTasks, error) {
+	previousPatch, err := patch.FindOne(patch.MostRecentPatchByUserAndProject(j.user.Username(), project.Identifier))
```
bynn, line 495: is there a chance that this patch might not be the patch we want to look at?

bynn (commented):
if a task keeps on failing would it keep rescheduling with this flag? 
and is it possible to link the patch that you tested on!

ablack12 (changes requested):

operations/patch.go
```diff
@@ -14,10 +14,11 @@ import (
...
 
 const (
-	patchDescriptionFlagName = "description"
-	patchVerboseFlagName     = "verbose"
-	patchTriggerAliasFlag    = "trigger-alias"
-	reuseDefinitionFlag      = "reuse"
+	patchDescriptionFlagName   = "description"
+	patchVerboseFlagName       = "verbose"
+	patchTriggerAliasFlag      = "trigger-alias"
+	reuseDefinitionFlag        = "reuse"
```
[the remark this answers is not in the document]
hadjri: I noticed that reuse_definition is a field set on patch intent objects in the intent collection which is why I did not change the reuse name to repeat.  However I can just rename the vairables and keep the BSON tag what it is, with a comment explaining the situation.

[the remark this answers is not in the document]
ablack12: You shouldn't have to change the name of anything besides the flag name; it's okay if it doesn't strictly match it as long as the user experience is clear. That or changing the flag to "reuse-failed".

hadjri:
Screen Shot 2022-03-10 at 2 48 53 PM

hadjri:
Screen Shot 2022-03-10 at 2 49 07 PM

hadjri:
Screen Shot 2022-03-10 at 2 48 43 PM

hadjri:
Attached 3 screenshots: 1st is a failed patch with 1 display task failed and 1 task failed.  2nd is CLI using -repeat-failed flag generating a new patch. 3rd is the newly created patch that has auto-selected the failed task and display task from screenshot 1.

hadjri:
> if a task keeps on failing would it keep rescheduling with this flag? and is it possible to link the patch that you tested on!

It's just a one time operation to retrieve the previous patch's failures and schedule them so it won't get stuck in an infinite failure scheduling loop or anything in the case that a task is just doomed to always fail due to a breaking bug.

operations/patch.go
```diff
@@ -58,11 +58,11 @@ func getPatchFlags(flags ...cli.Flag) []cli.Flag {
 				Usage: "patch trigger alias (set by project admin) specifying tasks from other projects",
 			},
 			cli.BoolFlag{
-				Name:  reuseDefinitionFlag,
+				Name:  joinFlagNames(repeatDefinitionFlag, "reuse"),
 				Usage: "use all of the same tasks/variants defined for the last patch you scheduled for this project",
 			},
 			cli.BoolFlag{
-				Name:  repeatFailedDefinitionFlag,
+				Name:  joinFlagNames(repeatFailedDefinitionFlag, "failed", "rf"),
```
ablack12, line 65: Let's just do "repeat-failed" and "rf" to keep this shorter.

units/patch_intent.go
```diff
@@ -491,55 +493,44 @@ func (j *patchIntentProcessor) getPreviousPatchDefinition(project *model.Project
...
-			for _, failedTask := range failedTasks {
-				if !failedTask.DisplayOnly && !failedTask.IsPartOfDisplay() {
-					failedTaskNames = append(failedTaskNames, failedTask.DisplayName)
-				}
-			}
+	failedTaskNames := make([]string, 0, len(failedTasks))
+	for _, failedTask := range failedTasks {
+		failedTaskNames = append(failedTaskNames, failedTask.DisplayName)
+	}
+	// I want the subset of vt.tasks that exist in tasksForVariant
```
ablack12, line 506: I think this was my comment so womp but can you change this to "We" and add a period?

model/patch/db.go
```diff
@@ -305,7 +305,7 @@ func PatchesByProject(projectId string, ts time.Time, limit int) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(limit)
 }
 
-// FindFailedCommitQueuePatchesInTimeRange returns failed patches if they started within range,
+// FindFailedCommitQueuePatchesinTimeRange returns failed patches if they started within range,
```
ablack12, line 308: I'm not sure how you found this haha but it's actually the function name, not the comment, that's incorrect if you wouldn't mind fixing that

units/patch_intent.go
```diff
@@ -483,6 +493,46 @@ func (j *patchIntentProcessor) getPreviousPatchDefinition(project *model.Project
 	return res, nil
 }
 
+func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, displayTasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, []patch.DisplayTask, error) {
+	failedTasks, err := task.FindAll(db.Query(task.FailedExecutionTasksByVersionAndBV(version, vt.Variant)))
```
ablack12, line 497: Do we need to query for tasks/display tasks separately? Can we query all of them together, and then when we loop through them, either add them to failedTaskNames or add them to a list of failedDisplayTasks?

units/patch_intent.go
```diff
@@ -483,6 +493,46 @@ func (j *patchIntentProcessor) getPreviousPatchDefinition(project *model.Project
...
+	}
+	// I want the subset of vt.tasks that exist in tasksForVariant
+	tasks := utility.StringSliceIntersection(tasksInProjectVariant, failedTaskNames)
+
+	var displayTasks []patch.DisplayTask
+	failedDisplayTasks, err := task.FindAll(db.Query(task.FailedDisplayTasksByVersionAndBV(version, vt.Variant)))
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "error querying for failed display tasks from previous patch")
+	}
+	for _, dt := range vt.DisplayTasks {
```
ablack12, line 514: I'm having trouble parsing this logic from the logic at 524. Is it possible to avoid this double for loop? If we convert failedDisplayTasks here into a string list instead of a display task list in the big for loop I suggested in the previous comment, could we just pass these things into getPreviousTasksAndDisplayTasks instead?

units/patch_intent.go
ablack12: Also, should these only find the tasks that we specifically defined in vt?

ablack12 (changes requested):

units/patch_intent.go
hadjri: Good point, I've converted failed display tasks and execution tasks to string arrays and now passing them into getPreviousTasksAndDisplayTasks

units/patch_intent.go
hadjri: I'm now using utility.StringSliceIntersection on both string arrays to ensure this although this could also be done in the DB operation

bynn (approved):

PR #5457, opened 2022-03-03
EVG-16448: atomically update host during termination

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16448

### Description 
Host termination does two updates - one to set the status, and another to set the termination time and unset the volumes. Since it's not atomic, it's possible for the host to be terminated in the DB, but no termination time is set yet. This temporary mismatch can cause [this](https://github.com/evergreen-ci/evergreen/blob/2406d863f1422c46b46ebd4abd2e9f19a12cad74/service/api_task.go#L500-L506) to log spuriously when the agent requests the next task.

* Atomically update the host to terminated using a single update.
* Refactor `SetStatus` to handle setting additional fields in the update.

### Testing 
Added unit tests for `SetStatusAndFields`.

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -435,8 +435,15 @@ func IsIntentHostId(id string) bool {
 	return strings.HasPrefix(id, "evg-")
 }
 
-func (h *Host) SetStatus(status, user string, logs string) error {
-	if h.Status == status {
+// SetStatus updates a host's status on behalf of the given user.
+func (h *Host) SetStatus(newStatus, user, logs string) error {
+	return h.SetStatusAndFields(newStatus, bson.M{}, user, logs)
+}
+
+// SetStatusAndFields is the same as SetStatus but also sets any of the other
+// given fields.
+func (h *Host) SetStatusAndFields(newStatus string, setFields bson.M, user, logs string) error {
+	if h.Status == newStatus {
 		return nil
 	}
 	if h.Status == evergreen.HostTerminated && h.Provider != evergreen.ProviderNameStatic {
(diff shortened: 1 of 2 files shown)


model/host/host.go
```diff
@@ -435,35 +435,41 @@ func IsIntentHostId(id string) bool {
 	return strings.HasPrefix(id, "evg-")
 }
 
-func (h *Host) SetStatus(status, user string, logs string) error {
-	if h.Status == status {
+// SetStatus updates a host's status on behalf of the given user.
+func (h *Host) SetStatus(newStatus, user, logs string) error {
```
julianedwards, line 439: Is this used anywhere else? If not, wouldn't it be simpler to have a single `SetFields` func that is called by `Terminate` with all of the fields we want to update?

ybrill:
I could see how this would cause the cases where this happens for a host once or twice right after it was terminated (until it actually dies), but what of the hosts where this goes on for days?

Kimchelly:
Those hit the context cancellation, which has already been addressed in [EVG-16291](https://jira.mongodb.org/browse/EVG-16291).

model/host/host.go
Kimchelly: Yes, `SetStatus` is used for most of the other status updates, which don't involve setting additional fields.

julianedwards (approved):

model/host/host.go
```diff
@@ -435,35 +435,41 @@ func IsIntentHostId(id string) bool {
...
 		grip.Warning(message.Fields{
 			"message": msg,
 			"host_id": h.Id,
-			"status":  status,
+			"status":  newStatus,
 		})
 		return errors.New(msg)
 	}
 
+	setFields[StatusKey] = newStatus
```
ybrill, line 459: would this panic if someone passed in nil for setFields? If so, should we check nil?

Kimchelly: It would panic, but I don't really think it should be necessary to always double-check non-nilness in function parameters (e.g. we assume people don't pass nil maps typically) unless there's good reason to believe that the nil value has some special case meaning.

PR #5459, opened 2022-03-03
EVG-16166: container task queue

hhoke:
[EVG-16166](https://jira.mongodb.org/browse/EVG-16166)

### Description 
Query for schedulable container tasks and some basic tests. This can be sorted with q.Sort when needed.

### Testing 
  Basic unit test to ensure we screen out inactivated tasks, tasks with the wrong Execution Platform, and tasks with negative priority.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -772,6 +772,25 @@ func schedulableHostTasksQuery() bson.M {
 	return q
 }
 
+// ScheduledContainerTasksQuery is a query that filters for scheduled tasks that are waiting for a pod
+// to be allocated to them, sorted by when they were scheduled.
+func ScheduledContainerTasksQuery() bson.M {
+	return bson.M{
+		ActivatedKey:         true,
+		StatusKey:            evergreen.TaskUndispatched,
+		ExecutionPlatformKey: ExecutionPlatformContainer,
+
+		// Filter out tasks disabled by negative priority
+		PriorityKey: bson.M{"$gt": evergreen.DisabledTaskPriority},
+
+		// Filter tasks containing unattainable dependencies
+		"$or": []bson.M{
+			{bsonutil.GetDottedKeyName(DependsOnKey, DependencyUnattainableKey): bson.M{"$ne": true}},
+			{OverrideDependenciesKey: true},
+		},
+	}
+}
+
 // TasksByProjectAndCommitPipeline fetches the pipeline to get the retrieve all tasks
 // associated with a given project and commit hash.
 func TasksByProjectAndCommitPipeline(projectId, commitHash, taskId, taskStatus string, limit int) []bson.M {
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -1855,6 +1855,56 @@ func TestUnattainableSchedulableHostTasksQuery(t *testing.T) {
 	assert.Len(schedulableTasks, 2)
 }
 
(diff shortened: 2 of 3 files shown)


model/task/db.go
```diff
@@ -772,6 +772,25 @@ func schedulableHostTasksQuery() bson.M {
...
 
+// ScheduledContainerTasksQuery is a query that filters for scheduled tasks that are waiting for a pod
+// to be allocated to them, sorted by when they were scheduled.
+func ScheduledContainerTasksQuery() bson.M {
+	return bson.M{
+		ActivatedKey:         true,
+		StatusKey:            evergreen.TaskUndispatched,
+		ExecutionPlatformKey: ExecutionPlatformContainer,
+
+		// Filter out tasks disabled by negative priority
```
Kimchelly, line 783: I don't think we need this comment. It should be clear just from the code itself.

model/task/db.go
```diff
@@ -772,6 +772,25 @@ func schedulableHostTasksQuery() bson.M {
...
+		PriorityKey: bson.M{"$gt": evergreen.DisabledTaskPriority},
+
+		// Filter tasks containing unattainable dependencies
```
Kimchelly, line 786: Same thing as above. I don't think this comment is necessary.

model/task/task_test.go
```diff
@@ -1855,6 +1855,56 @@ func TestUnattainableSchedulableHostTasksQuery(t *testing.T) {
...
+		{
+			Id:                "t4",
+			Activated:         false,
+			Status:            evergreen.TaskUndispatched,
+			ExecutionPlatform: ExecutionPlatformHost,
+			Priority:          0,
+		},
+	}
+	for _, task := range tasks {
+		assert.NoError(task.Insert())
```
Kimchelly, line 1899: require.NoError

Kimchelly (changes requested):

model/task/task_test.go
```diff
@@ -1855,6 +1855,56 @@ func TestUnattainableSchedulableHostTasksQuery(t *testing.T) {
...
+	}
+
+	q := db.Query(ScheduledContainerTasksQuery())
+	schedulableTasks, err := FindAll(q)
+	assert.NoError(err)
+	assert.Len(schedulableTasks, 2)
```
Kimchelly, line 1905: I would test this query more thoroughly since it has multiple checks. Usually when I write unit tests, I consider each clause in the query the equivalent of an "if" statment and to achieve good case coverage, it's important to verify that each individual "special case" in the definition of a "schedulable" task behaves as expected. For example, all the tasks tested in this have "undispatched" status, but it's worth verifying that a task that's lacking the undispatched status is filtered out of the query.

It's also important to test that the tasks you get back from each test are the ones you expect, not just the count (e.g. you could check that task IDs were for the tasks you thought this query would return).

When I have to test several special cases and expected behaviors, I usually do [something like this](https://github.com/evergreen-ci/evergreen/blob/2406d863f1422c46b46ebd4abd2e9f19a12cad74/model/pod/db_test.go#L14-L89).

hhoke:
https://github.com/hhoke/evergreen/commit/9e0857dcaecc850f31ed5147fd8c29f26d9f07f3
FYI, this was my last draft
@Kimchelly 
Sorry I couldn't finish it in time, thanks for all the help and guidance!

PR #5460, opened 2022-03-04
EVG-15555 Add aborted field to version

bynn:
[EVG-15555](https://jira.mongodb.org/browse/EVG-15555)

### Description 
version now keeps track of whether a task has been aborted or not. 
it updates in both the AbortTask function and updatebuildandversionstatus function that is is called in mark end of a task.
not sure if we need it in both places but that seems safer 

### Testing 
unit test

### DIFF
--- a/model/version.go
+++ b/model/version.go
@@ -37,6 +37,7 @@ type Version struct {
 	Branch              string               `bson:"branch_name" json:"branch_name,omitempty"`
 	BuildVariants       []VersionBuildStatus `bson:"build_variants_status,omitempty" json:"build_variants_status,omitempty"`
 	PeriodicBuildID     string               `bson:"periodic_build_id,omitempty" json:"periodic_build_id,omitempty"`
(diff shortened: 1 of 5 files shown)


model/version.go
```diff
@@ -155,6 +156,36 @@ func (self *Version) SetNotActivated() error {
 	)
 }
 
+func (self *Version) SetAborted() error {
```
ablack12, line 159: instead of having two very similar functions, we could instead pass in an aborted boolean and set it to that

model/version.go
```diff
@@ -155,6 +156,36 @@ func (self *Version) SetNotActivated() error {
...
+			"$set": bson.M{
+				VersionAbortedKey: false,
+			},
+		},
+	)
+
+}
+
+func (self *Version) IsAborted() bool {
+	return utility.FromBoolPtr(self.Aborted)
```
ablack12, line 186: This function doesn't quite work; if t.Aborted == nil, then would we need to query all tasks and then set it?

PR #5461, opened 2022-03-07
EVG-16353 add canCreateProject resolver

ablack12:
[EVG-16353](https://jira.mongodb.org/browse/EVG-16353)

### Description 
We'd discussed briefly having this be a resolver which returns an object, in case we want to return multiple permissions. However I'm actually waffling on that now; since each permission would require a separate call, it seems like it'd be more performant to split this up.

### Testing 
Adding a test seems a bit complicated since this is permission based. I think I can give the current user "testuser" the right permissions and then just verify that it does work? But since HasPermissions works and is tested more deeply it doesn't itself seem like a useful test.


graphql/schema.graphql
```diff
@@ -1131,6 +1132,10 @@ type GithubProjectConflicts {
   commitCheckIdentifiers: [String!]
 }
 
+type ProjectPermissions {
```
malikchaya2, line 1135: I don't think this is being used

graphql/resolvers.go
```diff
@@ -1658,6 +1658,15 @@ func (r *queryResolver) GithubProjectConflicts(ctx context.Context, projectID st
 	return &conflicts, nil
 }
 
+func (r *queryResolver) CanCreateProject(ctx context.Context) (bool, error) {
```
malikchaya2, line 1661: I remember now how difficult it is to write tests to test permissions. Can we maybe instead test this through the graphql playground, either on staging or local?

ablack12: Good idea I forgot about that! Did actually catch a bug. Verified that with the role we returned true and without it we returned false.

malikchaya2 (approved):
🎉

graphql/schema.graphql
```diff
@@ -9,6 +9,7 @@ type Query {
   projects: [GroupedProjects]!
   viewableProjectRefs: [GroupedProjects]!
   githubProjectConflicts(projectId: String!): GithubProjectConflicts!
+  canCreateProject: Boolean!
```
khelif96, line 12: Would this make sense under the user object? 
Since this determines if a user can create a project it would make it a little more flexible since we can query against any user instead of only your own user.

sophstad (commented):
> We'd discussed briefly having this be a resolver which returns an object, in case we want to return multiple permissions. However I'm actually waffling on that now; since each permission would require a separate call, it seems like it'd be more performant to split this up.

For more context, do we know what other permission resolvers we'll need to add for project settings?

graphql/schema.graphql
malikchaya2: To further elaborate: 

I think there would be benefit to setting up a "permissions" resolver that returns what permissions a user has, like canAnnotate, CanCreateProject, etc, and add to that as we keep needing more and more permissions for users. 

However, I think that if we want to set that up, we should keep this as is, and then create a ticket for that refactor since this is needed now, and that refactor would include also pulling out the permissions we have now in other resolvers (I can think of at least one).

ablack12:
> For more context, do we know what other permission resolvers we'll need to add for project settings?

The only other permission I can think of is Repo Admin for an eventual Create Branch button, but that's future work at this point.

graphql/schema.graphql
khelif96: A permissions resolver makes sense to me. I still think it should belong to a user object since permissions are something a user has. It also helps create a more complete graph and would allow for more flexibility in the future. (e.g) looking up what permissions a user has.

ablack12: Updated to have a permissions resolver! I'm not sure if having userId be part of the permissions actually makes sense but I'm not sure how else to support having userId be passed in.

graphql/resolvers.go
```diff
@@ -1070,6 +1088,25 @@ func (r *queryResolver) MyHosts(ctx context.Context) ([]*restModel.APIHost, erro
 	return apiHosts, nil
 }
 
+func (r *queryResolver) Permissions(ctx context.Context, userID string) (*Permissions, error) {
```
sophstad, line 1091: At this point, it seems like the definite majority of use cases for this resolver will relate to the current user accessed via context; it would be more cumbersome to get the user ID on the front end instead of looking it up in the query. My inclination is that we should at least offer a version of this resolver that does not require a user ID, but interested in other opinions!

graphql/schema.graphql
```diff
@@ -10,6 +10,7 @@ type Query {
   viewableProjectRefs: [GroupedProjects]!
   githubProjectConflicts(projectId: String!): GithubProjectConflicts!
   project(projectId: String!): Project!
+  permissions(userId: String!): Permissions!
```
khelif96, line 13: This should be a child of User 
https://github.com/evergreen-ci/evergreen/blob/00a4c8f3e8e4559cc23e04a019b6d1725c40c3e5/graphql/schema.graphql#L1447-L1452

It will have userId automatically passed in.

khelif96 (changes requested):

gqlgen.yml
```diff
@@ -336,4 +336,8 @@ models:
       agentLogs:
         resolver: true
       allLogs:
+        resolver: true
+  Permissions:
+    fields:
+      canCreateProject:
```
khelif96, line 342: This is not needed it is implied. And the default behavior of gqlgen.

graphql/resolvers.go
```diff
@@ -97,6 +97,10 @@ func (r *Resolver) ProjectSubscriber() ProjectSubscriberResolver {
 	return &projectSubscriberResolver{r}
 }
 
+func (r *Resolver) Permissions() PermissionsResolver {
+	return &permissionsResolver{r}
+}
```
khelif96, line 102: This to .

graphql/resolvers.go
```diff
@@ -120,6 +124,7 @@ type repoSettingsResolver struct{ *Resolver }
 type projectSubscriberResolver struct{ *Resolver }
 type projectVarsResolver struct{ *Resolver }
 type taskLogsResolver struct{ *Resolver }
+type permissionsResolver struct{ *Resolver }
```
khelif96, line 127: And this

khelif96 (changes requested):
Just one more thing.

khelif96 (approved):
Ignore my other comments

graphql/schema.graphql
```diff
@@ -1126,6 +1126,11 @@ type GroupedProjects {
   projects: [Project!]!
 }
 
+type Permissions {
+  userId: String!
+  canCreateProject: Boolean!
```
malikchaya2, line 1131: Now that we have this setup, we should also move canModifyAnnotation to here, either in this pr or in another ticket.

malikchaya2 (approved):
approved mod canModifyAnnotation being addressed in some way.

graphql/schema.graphql
malikchaya2: and probably also CanOverrideDependencies

ablack12: @malikchaya2 could you make a ticket to handle this in other ticket? (not sure if those are big enough to be separate tickets or if they should be handled in the same one)

malikchaya2: https://jira.mongodb.org/browse/EVG-16500

PR #5462, opened 2022-03-07
EVG-16410 modify PatchTriggerAliasInput to accurately reflect the input

ablack12:
[EVG-16410](https://jira.mongodb.org/browse/EVG-16410)

### Description 
Also changed TaskSpecifierInput items to be nullable, because on the legacy UI it's either patchAlias OR taskRegex and variantRegex, but if that seems wrong I'll change it back!

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -608,18 +608,16 @@ input WorkstationSetupCommandInput {
 
 input PatchTriggerAliasInput {
   alias: String!
-  childProjectId: String!
   childProjectIdentifier: String!
-  taskSpecifiers: [TaskSpecifierInput]
+  taskSpecifiers: [TaskSpecifierInput!]!
   status: String
   parentAsModule: String
-  variantsTasks: [VariantTaskInput]!
 }
 
 input TaskSpecifierInput {
-  patchAlias: String!
-  taskRegex: String!
-  variantRegex: String!
+  patchAlias: String
+  taskRegex: String
+  variantRegex: String
 }
 
 input ProjectVarsInput {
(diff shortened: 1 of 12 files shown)


graphql/schema.graphql
```diff
@@ -608,18 +608,16 @@ input WorkstationSetupCommandInput {
 
 input PatchTriggerAliasInput {
   alias: String!
-  childProjectId: String!
   childProjectIdentifier: String!
```
sophstad, line 611: I wasn't able to successfully call the save mutation in a GraphQL playground; is that perhaps because because we're never actually setting the `childProjectId` based on the identifier? I don't see anywhere where that occurs. Let me know if you were able to get this working though

ablack12: Good call to test -- there was an issue with not handling the non-deprecated fields in the service functions. This should work as expected now.

graphql/schema.graphql
```diff
@@ -608,18 +608,16 @@ input WorkstationSetupCommandInput {
...
-  variantsTasks: [VariantTaskInput]!
 }
 
 input TaskSpecifierInput {
-  patchAlias: String!
-  taskRegex: String!
-  variantRegex: String!
+  patchAlias: String
+  taskRegex: String
+  variantRegex: String
```
sophstad, line 620: I figured that these were non-nullable because empty strings would be provided for the "inactive" choice. Using empty strings would be more in line with the approach for [ProjectAliasInput](https://github.com/evergreen-ci/evergreen/blob/86d205927187656bc70273447b308fcd51b54f63/graphql/schema.graphql#L636-L645), but either works.

sophstad (approved):
🎉 great to have the deprecated field removed.

graphql/schema.graphql
ablack12: Okay that makes sense, i can change it back!

PR #5468, opened 2022-03-08
EVG-16150 Add option to fail or noop s3.put if file already exists

malikchaya2:
[EVG-16150](https://jira.mongodb.org/browse/EVG-16150)

### Description 
This adds a no_overwrite flag that can be added to the s3.put command to skip files that already exist. 

### Testing 
I [tested it on staging](https://spruce-staging.corp.mongodb.com/version/622775a19dbe323e09f7d4c2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) by adding the no_overwrite flag and restarting the task. The first run has the file, the second one doesn't and prints the noop message.

### DIFF
--- a/agent/command/s3_put.go
+++ b/agent/command/s3_put.go
@@ -10,12 +10,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/evergreen-ci/evergreen/agent/internal"
 	"github.com/evergreen-ci/evergreen/agent/internal/client"
 	agentutil "github.com/evergreen-ci/evergreen/agent/util"
 	"github.com/evergreen-ci/evergreen/model/artifact"
+	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/pail"
 	"github.com/evergreen-ci/utility"
@@ -85,10 +87,14 @@ type s3put struct {
 	// for missing files.
 	Optional string `mapstructure:"optional" plugin:"expand"`
 
+	// NoOverwrite, when set to true, will not upload files if they already exist in s3
+	NoOverwrite string `mapstructure:"no_overwrite" plugin:"expand"`
+
 	// workDir sets the working directory relative to which s3put should look for files to upload.
 	// workDir will be empty if an absolute path is provided to the file.
-	workDir     string
-	skipMissing bool
+	workDir         string
+	skipMissing     bool
+	noOverwriteBool bool
 
 	bucket pail.Bucket
 
@@ -190,6 +196,15 @@ func (s3pc *s3put) expandParams(conf *internal.TaskConfig) error {
 		}
 	} else {
 		s3pc.skipMissing = false
+	}
+
+	if s3pc.NoOverwrite != "" {
+		s3pc.noOverwriteBool, err = strconv.ParseBool(s3pc.NoOverwrite)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	} else {
+		s3pc.noOverwriteBool = false
 	}
 	return nil
 }
@@ -338,6 +353,18 @@ retryLoop:
 				}
 
 				fpath = filepath.Join(filepath.Join(s3pc.workDir, s3pc.LocalFilesIncludeFilterPrefix), fpath)
+
+				if s3pc.noOverwriteBool {
+					// check if the file exists
+					exists, err := s3pc.fileExists(remoteName)
+					if err != nil {
+						return errors.Wrapf(err, "error checking if file %s exists", remoteName)
+					}
+					if exists {
+						logger.Task().Infof("noop: file '%s' already exists. Continuing to upload other files.", fpath)
+						continue uploadLoop
+					}
+				}
 				err = s3pc.bucket.Upload(ctx, remoteName, fpath)
 				if err != nil {
 					// retry errors other than "file doesn't exist", which we handle differently based on what
@@ -463,4 +490,27 @@ func (s3pc *s3put) isPrivate(visibility string) bool {
 func (s3pc *s3put) isPublic() bool {
 	return (s3pc.Visibility == "" || s3pc.Visibility == artifact.Public) &&
 		(s3pc.Permissions == s3.BucketCannedACLPublicRead || s3pc.Permissions == s3.BucketCannedACLPublicReadWrite)
+}
+
+func (s3pc *s3put) fileExists(remoteName string) (bool, error) {
+	requestParams := thirdparty.RequestParams{
+		Bucket:    s3pc.Bucket,
+		FileKey:   remoteName,
+		AwsKey:    s3pc.AwsKey,
+		AwsSecret: s3pc.AwsSecret,
+	}
+	_, err := thirdparty.GetHeadObject(requestParams)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case thirdparty.NotFoundError:
+				return false, nil
+			default:
+				return false, errors.Wrapf(err, "error getting head object for: %s", remoteName)
+			}
+		} else {
+			return false, errors.Wrapf(err, "error reading error while getting head object '%s'", remoteName)
+		}
+	}
+	return true, nil
 }
--- a/thirdparty/s3.go
+++ b/thirdparty/s3.go
@@ -10,10 +10,13 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
-//PresignExpireTime sets the amount of time the link is live before expiring
-const PresignExpireTime = 24 * time.Hour
+const (
+	//PresignExpireTime sets the amount of time the link is live before expiring
+	PresignExpireTime = 24 * time.Hour
+	NotFoundError     = "NotFound"
+)
 
-//RequestParams holds all the parameters needed to sign a url
+//RequestParams holds all the parameters needed to sign a url or fetch headObject
 type RequestParams struct {
 	Bucket    string `json:"bucket"`
 	FileKey   string `json:"fileKey"`
@@ -42,4 +45,29 @@ func PreSign(r RequestParams) (string, error) {
 
 	urlStr, err := req.Presign(PresignExpireTime)
 	return urlStr, err
+}
+
+func GetHeadObject(r RequestParams) (*s3.HeadObjectOutput, error) {
+	session, err := session.NewSession(&aws.Config{
+		Region: aws.String(endpoints.UsEast1RegionID),
+		Credentials: credentials.NewStaticCredentialsFromCreds(credentials.Value{
+			AccessKeyID:     r.AwsKey,
+			SecretAccessKey: r.AwsSecret,
+		}),
+	})
+	if err != nil {
+		return nil, err
+	}
+	svc := s3.New(session)
+
+	headObject, err := svc.HeadObject(&s3.HeadObjectInput{
+		Bucket: aws.String(r.Bucket),
+		Key:    aws.String(r.FileKey),
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return headObject, err
 }
--- a/config.go
+++ b/config.go
@@ -35,7 +35,7 @@ var (
 	ClientVersion = "2022-03-01"
 
 	// Agent version to control agent rollover.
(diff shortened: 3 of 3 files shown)


thirdparty/s3.go
```diff
@@ -10,10 +10,13 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
-//PresignExpireTime sets the amount of time the link is live before expiring
-const PresignExpireTime = 24 * time.Hour
+const (
+	//PresignExpireTime sets the amount of time the link is live before expiring
```
Kimchelly, line 14: nit: space between `//` and start of comment. Also, period at the end of a full sentence.

thirdparty/s3.go
```diff
@@ -10,10 +10,13 @@ import (
...
+	PresignExpireTime = 24 * time.Hour
+	NotFoundError     = "NotFound"
+)
 
-//RequestParams holds all the parameters needed to sign a url
+//RequestParams holds all the parameters needed to sign a url or fetch headObject
```
Kimchelly, line 19: Same thing as above about comment style.

thirdparty/s3.go
```diff
@@ -43,3 +46,28 @@ func PreSign(r RequestParams) (string, error) {
 	urlStr, err := req.Presign(PresignExpireTime)
 	return urlStr, err
 }
+
+func GetHeadObject(r RequestParams) (*s3.HeadObjectOutput, error) {
```
Kimchelly, line 50: Documentation.

thirdparty/s3.go
```diff
@@ -43,3 +46,28 @@ func PreSign(r RequestParams) (string, error) {
...
+	session, err := session.NewSession(&aws.Config{
+		Region: aws.String(endpoints.UsEast1RegionID),
```
Kimchelly, line 52: Is it guaranteed to be in us-east-1? Seems like users are allowed to specify the region when calling s3.put.

thirdparty/s3.go
```diff
@@ -43,3 +46,28 @@ func PreSign(r RequestParams) (string, error) {
...
+	headObject, err := svc.HeadObject(&s3.HeadObjectInput{
+		Bucket: aws.String(r.Bucket),
+		Key:    aws.String(r.FileKey),
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return headObject, err
```
Kimchelly, line 72: This can just `return headObject, nil` since it's already known that the error is non-nil.

thirdparty/s3.go
Kimchelly, line 50: Is it worth moving this functionality into Pail? The repo exists to abstract away S3 operations anyways so that we don't have as much miscellaneous third-party code.

agent/command/s3_put.go
```diff
@@ -464,3 +491,26 @@ func (s3pc *s3put) isPublic() bool {
 	return (s3pc.Visibility == "" || s3pc.Visibility == artifact.Public) &&
 		(s3pc.Permissions == s3.BucketCannedACLPublicRead || s3pc.Permissions == s3.BucketCannedACLPublicReadWrite)
 }
+
+func (s3pc *s3put) fileExists(remoteName string) (bool, error) {
```
Kimchelly, line 495: I would call this something like `remoteFileExists` so that it doesn't cause ambiguity between the local files vs. remote files.

agent/command/s3_put.go
```diff
@@ -338,6 +353,18 @@ retryLoop:
 				}
 
 				fpath = filepath.Join(filepath.Join(s3pc.workDir, s3pc.LocalFilesIncludeFilterPrefix), fpath)
+
+				if s3pc.noOverwriteBool {
+					// check if the file exists
```
Kimchelly, line 358: I don't think this comment is necessary since it's clear from the code itself that it's checking for the existence of the remote file.

agent/command/s3_put.go
```diff
@@ -85,10 +87,14 @@ type s3put struct {
 	// for missing files.
 	Optional string `mapstructure:"optional" plugin:"expand"`
 
+	// NoOverwrite, when set to true, will not upload files if they already exist in s3
```
Kimchelly, line 90: Period at the end of the sentence.

Kimchelly (changes requested):

agent/command/s3_put.go
```diff
@@ -85,10 +87,14 @@ type s3put struct {
...
+	NoOverwrite string `mapstructure:"no_overwrite" plugin:"expand"`
```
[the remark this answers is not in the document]
malikchaya2: I don't think that's true, because I used ` no_overwrite: true` when I was testing and it works. I think it just automatically parses the yaml as strings. I might be wrong though.

thirdparty/s3.go
malikchaya2: What do you mean by this? I plan to document the new flag in our wiki, but what kind of documentation would you like to see here?

malikchaya2: do you mean adding a comment here?

agent/command/s3_put.go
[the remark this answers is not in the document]
malikchaya2: I modeled this after optional, which is also a boolean.

agent/command/s3_put.go
```diff
@@ -338,6 +353,18 @@ retryLoop:
...
 				fpath = filepath.Join(filepath.Join(s3pc.workDir, s3pc.LocalFilesIncludeFilterPrefix), fpath)
+
+				if s3pc.noOverwriteBool {
+					// check if the file exists
+					exists, err := s3pc.fileExists(remoteName)
+					if err != nil {
+						return errors.Wrapf(err, "error checking if file %s exists", remoteName)
+					}
+					if exists {
+						logger.Task().Infof("noop: file '%s' already exists. Continuing to upload other files.", fpath)
```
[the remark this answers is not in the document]
malikchaya2: good point, thank you! fixing.

thirdparty/s3.go
malikchaya2: thanks for spotting that! done.

thirdparty/s3.go
Kimchelly: Yes, it's considered good practice to document things that are exported, since they're exposing functionality that could be used anywhere. [This blog post](https://go.dev/blog/godoc) talks about good style for Go doc comments, but the gist is to treat writing doc comments like you're writing documentation for the standard library (or any other widely-used Go library).
* Start the comment with the name of the thing you're documenting (i.e. `GetHeadObject`).
* Full sentences and proper grammar are golden.
* Describe expected behavior.
* Describe the expected inputs if they're not obvious.

thirdparty/s3.go
```diff
@@ -11,20 +11,22 @@ import (
 )
 
 const (
-	//PresignExpireTime sets the amount of time the link is live before expiring
+	// PresignExpireTime sets the amount of time the link is live before expiring.
 	PresignExpireTime = 24 * time.Hour
-	NotFoundError     = "NotFound"
+	// NotFound is returned by s3 when headObject doesn't exist
```
Kimchelly, line 16: Period at the end of the sentence.

I would phrase this more generally as "returned by S3 when an object does not exist" rather than assuming it specifically only applies to the `GetHeadObject` operation.

thirdparty/s3.go
```diff
@@ -47,9 +49,10 @@ func PreSign(r RequestParams) (string, error) {
 	return urlStr, err
 }
 
+// GetHeadObject fetches the metadata of an s3 object
```
Kimchelly, line 52: Period at the end of the sentence.

thirdparty/s3.go
Kimchelly: What do you mean? We wrote [pail](https://github.com/evergreen-ci/pail), so we can add whatever S3-related functionality we want.

thirdparty/s3.go
```diff
@@ -47,9 +49,10 @@ func PreSign(r RequestParams) (string, error) {
...
 func GetHeadObject(r RequestParams) (*s3.HeadObjectOutput, error) {
 	session, err := session.NewSession(&aws.Config{
-		Region: aws.String(endpoints.UsEast1RegionID),
+		Region: &r.Region,
```
Kimchelly, line 55: nit: to make this consistent when using the AWS SDK, can we use `aws.String(r.Region)`?

thirdparty/s3.go
malikchaya2: thank you!

malikchaya2: My bad, I didn't realize that. In that case, we should probably also move over PreSign. I think it makes sense to create a ticket for it.

thirdparty/s3.go
Kimchelly: Bump the first comment about the period.

Kimchelly (approved):
LGTM mod ticket for moving things to pail and the one period nit.

thirdparty/s3.go
malikchaya2: https://jira.mongodb.org/browse/EVG-16477

agent/command/s3_put.go
```diff
@@ -85,10 +87,14 @@ type s3put struct {
 	// for missing files.
 	Optional string `mapstructure:"optional" plugin:"expand"`
 
+	// NoOverwrite, when set to true, will not upload files if they already exist in s3.
+	NoOverwrite string `mapstructure:"no_overwrite" plugin:"expand"`
```
ablack12, line 91: I think this naming convention is non-standard; can we call this SkipExisting instead?

agent/command/s3_put.go
```diff
@@ -191,6 +197,15 @@ func (s3pc *s3put) expandParams(conf *internal.TaskConfig) error {
...
 		s3pc.skipMissing = false
 	}
+
+	if s3pc.NoOverwrite != "" {
+		s3pc.noOverwriteBool, err = strconv.ParseBool(s3pc.NoOverwrite)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	} else {
+		s3pc.noOverwriteBool = false
```
ablack12, line 207: I know you copied the existing optional code but this is implied already so we don't need to specify it

agent/command/s3_put.go
```diff
@@ -338,6 +353,17 @@ retryLoop:
 				}
 
 				fpath = filepath.Join(filepath.Join(s3pc.workDir, s3pc.LocalFilesIncludeFilterPrefix), fpath)
+
+				if s3pc.noOverwriteBool {
+					exists, err := s3pc.remoteFileExists(remoteName)
+					if err != nil {
+						return errors.Wrapf(err, "error checking if file %s exists", remoteName)
```
ablack12, line 360: nit: file '%s'

thirdparty/s3.go
```diff
@@ -10,18 +10,23 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
-//PresignExpireTime sets the amount of time the link is live before expiring
-const PresignExpireTime = 24 * time.Hour
+const (
+	// PresignExpireTime sets the amount of time the link is live before expiring.
+	PresignExpireTime = 24 * time.Hour
+	// NotFound is returned by S3 when an object does not exist.
+	NotFoundError = "NotFound"
```
ablack12, line 17: Can we either call this S3NotFoundError (since thirdparty has more than s3 in it) or define it as a local constant in s3_put

ablack12 (changes requested):
No need for me to look again unless you'd like

thirdparty/s3.go
malikchaya2: I moved it to s3_put

agent/command/s3_put.go
malikchaya2: good point. removed.

agent/command/s3_put.go
```diff
@@ -87,21 +87,24 @@ type s3put struct {
...
+	skipExistingBool bool
 
 	bucket pail.Bucket
 
 	taskdata client.TaskData
 	base
 }
 
+// NotFound is returned by S3 when an object does not exist.
+const NotFoundError = "NotFound"
```
ablack12, line 106: this can be a private constant now

agent/command/s3_put.go
```diff
@@ -198,13 +201,11 @@ func (s3pc *s3put) expandParams(conf *internal.TaskConfig) error {
 		s3pc.skipMissing = false
```
ablack12, line 201: could we remove here too while we're here?

ablack12 (approved):

agent/command/s3_put.go
malikchaya2: done

agent/command/s3_put.go
malikchaya2: done

PR #5470, opened 2022-03-08
EVG-16395 Initial interface removal

hadjri:
[EVG-16395](https://jira.mongodb.org/browse/EVG-16395)

### Description 
For rest routes that no longer have a unit test mocking their connectors, I've removed the interface function for it and also removed the data connector from the route properties.
### Testing 
Confirmed tests remain green.


graphql/util.go
```diff
@@ -61,7 +61,7 @@ func GetGroupedFiles(ctx context.Context, name string, taskID string, execution
 	return &GroupedFiles{TaskName: &name, Files: apiFileList}, nil
 }
 
-func SetScheduled(ctx context.Context, sc data.Connector, taskID string, isActive bool) (*restModel.APITask, error) {
+func SetScheduled(ctx context.Context, url string, taskID string, isActive bool) (*restModel.APITask, error) {
```
ablack12, line 64: nit: url, taskID string

graphql/util.go
```diff
@@ -114,8 +114,9 @@ func GetFormattedDate(t *time.Time, timezone string) (*string, error) {
 	return &newTime, nil
 }
 
-func getVersionBaseTasks(d data.Connector, versionID string) ([]task.Task, error) {
-	version, err := d.FindVersionById(versionID)
+func getVersionBaseTasks(versionID string) ([]task.Task, error) {
+	dc := data.DBVersionConnector{}
+	version, err := dc.FindVersionById(versionID)
```
ablack12, line 119: Rather than still using the connector, we should just call the actual model.FindVersionById function

graphql/util.go
```diff
@@ -497,8 +498,9 @@ func generateBuildVariants(sc data.Connector, versionId string, searchVariants [
 		IncludeBaseTasks:               true,
 		IncludeBuildVariantDisplayName: true,
 	}
+	dc := data.DBTaskConnector{}
 	start := time.Now()
-	tasks, _, err := sc.FindTasksByVersion(versionId, opts)
+	tasks, _, err := dc.FindTasksByVersion(versionId, opts)
```
ablack12, line 503: Same feedback about not using the connector at all here.

ablack12 (changes requested):
This is a really surface review; I think there's a slight misunderstanding on this ticket. The issue isn't that we pass the connector around all the time but really just that we use it at all. We don't need these functions to be methods on a connector if they don't need to be mocked, so we should move the function to a model function or a rest/data function and just call it directly. I'll review in depth after that, lmk if you want to discuss it.

rest/data/task.go
```diff
@@ -16,13 +16,9 @@ import (
...
 )
 
-// DBTaskConnector is a struct that implements the Task related methods
-// from the Connector through interactions with he backing database.
-type DBTaskConnector struct{}
-
 // FindTaskById uses the service layer's task type to query the backing database for
 // the task with the given taskId.
-func (tc *DBTaskConnector) FindTaskById(taskId string) (*task.Task, error) {
+func FindTaskById(taskId string) (*task.Task, error) {
```
ablack12, line 21: These data functions that just wrap the model functions directly should just be removed, and we should call the model function directly.

ablack12 (commented):
I'm going to  do a thorough review of this today but wanted to mention that the merge conflicts are pretty large; this may be better to review after they're resolved. Did leave a comment about not having functions in data that are just wrappers of the model functions.

malikchaya2 (commented):
I love how much cleaner this makes things!

graphql/util.go
```diff
@@ -340,11 +340,14 @@ func getCedarFailedTestResultsSample(ctx context.Context, tasks []task.Task, tes
...
 // modifyVersionHandler handles the boilerplate code for performing a modify version action, i.e. schedule, unschedule, restart and set priority
-func modifyVersionHandler(ctx context.Context, dataConnector data.Connector, patchID string, modification model.VersionModification) error {
-	v, err := dataConnector.FindVersionById(patchID)
+func modifyVersionHandler(ctx context.Context, patchID string, modification model.VersionModification) error {
+	v, err := model.VersionFindOneId(patchID)
 	if err != nil {
 		return ResourceNotFound.Send(ctx, fmt.Sprintf("error finding version %s: %s", patchID, err.Error()))
 	}
+	if v == nil {
+		return ResourceNotFound.Send(ctx, fmt.Sprintf("Unable to find version with id: `%s`", patchID))
```
ablack12, line 349: nit: no : between id and '%s'

model/lifecycle.go
```diff
@@ -268,6 +268,9 @@ func SetVersionPriority(versionId string, priority int64, caller string) error {
 	return nil
 }
 
+// RestartTasksInVersion restarts completed tasks associated with a given versionId.
```
ablack12, line 271: Thanks for adding !!

rest/data/aliases.go
```diff
@@ -59,7 +55,7 @@ func (d *DBAliasConnector) FindProjectAliases(projectId, repoId string, aliasesT
 }
 
 // CopyProjectAliases finds the aliases for a given project and inserts them for the new project.
-func (d *DBAliasConnector) CopyProjectAliases(oldProjectId, newProjectId string) error {
+func CopyProjectAliases(oldProjectId, newProjectId string) error {
```
ablack12, line 58: Can this be moved to model since it doesn't need API models?

rest/data/aliases.go
```diff
@@ -152,7 +148,7 @@ func shouldSkipAliasForSection(section model.ProjectPageSection, alias string) b
 }
 
 //  GetMatchingGitTagAliasesForProject returns matching git tag aliases that match the given git tag
-func (d *DBAliasConnector) HasMatchingGitTagAliasAndRemotePath(projectId, tag string) (bool, string, error) {
+func HasMatchingGitTagAliasAndRemotePath(projectId, tag string) (bool, string, error) {
```
ablack12, line 151: Can this be moved to model?

rest/data/distro.go
```diff
@@ -4,21 +4,15 @@ import (
...
-	"github.com/pkg/errors"
 )
 
-// DBDistroConnector is a struct that implements the Distro related methods
-// from the Connector through interactions with the backing database.
-type DBDistroConnector struct{}
-
 // FindDistroById queries the database to find a given distros.
-func (dc *DBDistroConnector) FindDistroById(distroId string) (*distro.Distro, error) {
+func FindDistroById(distroId string) (*distro.Distro, error) {
```
ablack12, line 15: Since we're just wrapping one function, we should just call the FindOne function directly

rest/data/distro.go
```diff
@@ -73,7 +67,7 @@ func (dc *DBDistroConnector) UpdateDistro(old, new *distro.Distro) error {
 }
 
 // CreateDistro inserts the given distro.Distro.
-func (dc *DBDistroConnector) CreateDistro(distro *distro.Distro) error {
+func CreateDistro(distro *distro.Distro) error {
```
ablack12, line 70: Similarly I think this doesn't need to be it's own function since it's only wrapping one operation and this error is already going to be checked outside of the function

rest/data/host.go
```diff
@@ -57,34 +51,6 @@ func (hc *DBHostConnector) FindHostsInRange(apiParams restmodel.APIHostParams, u
...
-
-func (dbc *DBConnector) FindHostByIdWithOwner(hostID string, user gimlet.User) (*host.Host, error) {
-	return findHostByIdWithOwner(dbc, hostID, user)
-}
-
-func (hc *DBHostConnector) FindHostsByDistro(distro string) ([]host.Host, error) {
-	return host.Find(db.Query(host.ByDistroIDsOrAliasesRunning(distro)))
-}
-
 func (hc *DBConnector) GetPaginatedRunningHosts(hostID, distroID, currentTaskID string, statuses []string, startedBy string, sortBy string, sortDir, page, limit int) ([]host.Host, *int, int, error) {
```
ablack12, line 54: I don't see that the mock is used here; do we need to keep this as a connector function?

rest/data/interface.go
```diff
@@ -54,348 +35,10 @@ type Connector interface {
...
-	UpdatePodStatus(id string, current, updated restModel.APIPodStatus) error
-	// CheckPodSecret checks that the ID and secret match the server's
-	// stored credentials for the pod.
-	CheckPodSecret(id, secret string) error
-
-	// FindProjectAliases queries the database to find all aliases (including or excluding those specified).
-	// Includes repo aliases if repoId is provided.
-	FindProjectAliases(string, string, []restModel.APIProjectAlias) ([]restModel.APIProjectAlias, error)
-	// CopyProjectAliases copies aliases from the first project for the second project.
-	CopyProjectAliases(string, string) error
```
ablack12, line 321: I didn't review this closely but if these functions don't have comments, can we move their interface comment to be the function comment?

ablack12 (commented):
I only did a partial review again (one day I'll get through it haha) but I think there are some more functions that could be moved into model. We don't really have a rule around this but I think the rule could be:

1) If the rest/data function just wraps a model or etc function, we should just call that function directly. (This isn't true for any functions that use the units package, since moving it into model would create an import cycle.) 
2) If the rest/data function calls a series of functions but doesn't do anything with error codes then we should move that function into model (or etc) rather than having it in rest/data.
3) If the function uses API models or does many things and returns different error codes for AC, it _should_ be in rest/data.

If we want to audit these usages in another ticket we can do that.

rest/data/host.go
hadjri: Nope, this slipped through the cracks. Thanks!

graphql/resolvers.go
```diff
@@ -1306,15 +1305,15 @@ func (r *patchResolver) TaskStatuses(ctx context.Context, obj *restModel.APIPatc
 		IncludeBaseTasks:               false,
 		IncludeBuildVariantDisplayName: false,
 	}
-	tasks, _, err := r.sc.FindTasksByVersion(*obj.Id, opts)
+	tasks, _, err := data.FindTasksByVersion(*obj.Id, opts)
```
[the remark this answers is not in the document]
hadjri: It doesn't seem to be needed no - I've removed it and the entire data.FindTasksByVersion func along with it

graphql/resolvers.go
```diff
@@ -1869,11 +1872,11 @@ func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution
 
 		apiTestResults[i] = apiTest
 	}
-	totalTestCount, err := r.sc.GetTestCountByTaskIdAndFilters(taskID, "", []string{}, dbTask.Execution)
+	totalTestCount, err := data.GetTestCountByTaskIdAndFilters(taskID, "", []string{}, dbTask.Execution)
```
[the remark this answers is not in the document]
hadjri: Putting it in testresult causes an import cycle but I've moved it to model/task

model/project_ref.go
```diff
@@ -1022,6 +1023,28 @@ func getCommonProjectVariables(projectIds []string) (*ProjectVars, error) {
 	}, nil
 }
 
+func FindTaskWithinTimePeriod(startedAfter, finishedBefore time.Time,
```
ablack12, line 1026: I don't this really makes sense to be in projectRef, since it doesn't have anything to do with it besides getting the right ID. Since this is only used by projectTaskGetHandler, this can just be a helper method for that handler.

model/project_ref.go
```diff
@@ -1372,6 +1395,143 @@ func FindOneProjectRefWithCommitQueueByOwnerRepoAndBranch(owner, repo, branch st
...
+	}
+
+	if err = hook.Insert(); err != nil {
+		return false, errors.Wrapf(err, "error inserting new webhook for project '%s'", projectRef.Id)
+	}
+	projectRef.TracksPushEvents = utility.TruePtr()
+	return true, nil
+}
+
+func EnablePRTesting(projectRef *ProjectRef) error {
```
ablack12, line 1438: I think this should also be a helper function of the route that calls it, since it's only part of one route (same for EnableCommitQueue), or else this needs to be reworded as "canEnable" since it doesn't actually enable anything

model/project_ref.go
```diff
@@ -1372,6 +1395,143 @@ func FindOneProjectRefWithCommitQueueByOwnerRepoAndBranch(owner, repo, branch st
...
+
+func UpdateProjectRevision(projectID, revision string) error {
+	if err := UpdateLastRevision(projectID, revision); err != nil {
+		return errors.Wrapf(err, "error updating revision for project '%s'", projectID)
+	}
+
+	return nil
+}
+
+func UpdateProjectVarsByValue(toReplace, replacement, username string, dryRun bool) (map[string][]string, error) {
```
ablack12, line 1486: Can this be moved to model/project_vars ?

model/task/task.go
```diff
@@ -3099,7 +3120,7 @@ type GetTasksByVersionOptions struct {
 	Sorts                          []TasksSortOrder
 	IncludeExecutionTasks          bool
 	IncludeBaseTasks               bool
-	IncludeEmptyActivaton          bool
+	IncludeEmptyActivation         bool
```
ablack12, line 3123: Haha good catch

rest/data/patch.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func validatePatchID(patchId string) error {
+func ValidatePatchID(patchId string) error {
```
ablack12, line 24: Why was this function made public?

rest/data/project.go
```diff
@@ -344,14 +181,10 @@ func GetProjectEventLog(project string, before time.Time, n int) ([]restModel.AP
...
-	return getEventsById(id, before, n)
+	return GetEventsById(id, before, n)
 }
 
-func GetRepoEventLog(repoId string, before time.Time, n int) ([]restModel.APIProjectEvent, error) {
-	return getEventsById(repoId, before, n)
-}
-
-func getEventsById(id string, before time.Time, n int) ([]restModel.APIProjectEvent, error) {
+func GetEventsById(id string, before time.Time, n int) ([]restModel.APIProjectEvent, error) {
```
ablack12, line 187: Why was this made public?

rest/data/project_settings.go
```diff
@@ -64,14 +64,14 @@ func CopyProject(ctx context.Context, opts CopyProjectOpts) (*restModel.APIProje
 	if err := CopyProjectVars(oldId, projectToCopy.Id); err != nil {
```
ablack12, line 64: Should this be moved to model/project_vars?

rest/data/project_settings.go
```diff
@@ -64,14 +64,14 @@ func CopyProject(ctx context.Context, opts CopyProjectOpts) (*restModel.APIProje
...
 		return nil, errors.Wrapf(err, "error copying project vars from project '%s'", oldIdentifier)
 	}
-	if err := CopyProjectAliases(oldId, projectToCopy.Id); err != nil {
+	if err := model.CopyProjectAliases(oldId, projectToCopy.Id); err != nil {
 		return nil, errors.Wrapf(err, "error copying aliases from project '%s'", oldIdentifier)
 	}
 	if err := CopyProjectSubscriptions(oldId, projectToCopy.Id); err != nil {
```
ablack12, line 70: Should this be moved to model/event/subscriptions?

rest/route/build.go
```diff
@@ -53,9 +53,18 @@ func (b *buildGetHandler) Run(ctx context.Context) gimlet.Responder {
...
-	if err != nil {
-		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Database error"))
+	var tasks []task.Task
+	if len(taskIDs) > 0 {
+		tasks, err = task.Find(task.ByIds(taskIDs))
+		if err != nil {
+			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Database error"))
+		}
+		if len(tasks) == 0 {
+			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(gimlet.ErrorResponse{
```
ablack12, line 63: there's no reason to wrap it this way. We can just return the gimlet.ErrorResponse wrapped directly by MakeJSONErrorResponder (All "Internal" does is add the http.StatusInternalError status code, which we'd be overwriting anyway). 

However, since _we_ determine the task Ids and they aren't passed in, I think a internal server error is correct since it's our fault, so we can just return `gimlet.MakeJSONInternalErrorResponder(errors.New("no tasks found"))` or something

rest/route/distro.go
```diff
@@ -50,13 +51,16 @@ func (h *distroIDGetSetupHandler) Parse(ctx context.Context, r *http.Request) er
 
 // Run returns the given distro's setup script.
 func (h *distroIDGetSetupHandler) Run(ctx context.Context) gimlet.Responder {
-	d, err := data.FindDistroById(h.distroID)
+	d, err := distro.FindOne(distro.ById(h.distroID))
```
ablack12, line 54: can use FindOneID

rest/route/distro.go
```diff
@@ -50,13 +51,16 @@ func (h *distroIDGetSetupHandler) Parse(ctx context.Context, r *http.Request) er
...
 	if err != nil {
-		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Database error for find() by distro id '%s'", h.distroID))
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(gimlet.ErrorResponse{
```
ablack12, line 56: Similar to earlier thought, don't need to double wrap the error (the message error is sufficient)

rest/route/distro.go
```diff
@@ -150,10 +157,14 @@ func (h *distroIDPutHandler) Parse(ctx context.Context, r *http.Request) error {
 // (b) creates a new resource based on the Request-URI and JSON payload
 func (h *distroIDPutHandler) Run(ctx context.Context) gimlet.Responder {
 	user := MustHaveUser(ctx)
-	original, err := data.FindDistroById(h.distroID)
-	if err != nil && err.(gimlet.ErrorResponse).StatusCode != http.StatusNotFound {
-		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Database error for find() by distro id '%s'", h.distroID))
+	original, err := distro.FindOne(distro.ById(h.distroID))
+	if err != nil && !adb.ResultsNotFound(err) {
```
ablack12, line 161: We don't need to check ResultsNotFound anymore since we can just check if original == nil (should update distroExists for that as well)

rest/route/distro.go
```diff
@@ -238,7 +252,13 @@ func (h *distroIDDeleteHandler) Parse(ctx context.Context, r *http.Request) erro
...
-	_, err := data.FindDistroById(h.distroID)
+	_, err := distro.FindOne(distro.ById(h.distroID))
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("distro with id '%s' not found", h.distroID),
+		}, "Database error for find() by distro id '%s'", h.distroID))
+	}
 	if err != nil {
 		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "Database error for find() by distro id '%s'", h.distroID))
```
ablack12, line 263: Mistake to keep both error messages but should check if distro == nil now

rest/route/distro.go
```diff
@@ -285,13 +305,16 @@ func (h *distroIDPatchHandler) Parse(ctx context.Context, r *http.Request) error
 // Run updates a distro by id.
 func (h *distroIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
 	user := MustHaveUser(ctx)
-	old, err := data.FindDistroById(h.distroID)
+	old, err := distro.FindOne(distro.ById(h.distroID))
```
ablack12, line 308: distro.FindOneID in this file too

rest/route/distro.go
```diff
@@ -345,13 +368,19 @@ func (h *distroIDGetHandler) Parse(ctx context.Context, r *http.Request) error {
...
 func (h *distroIDGetHandler) Run(ctx context.Context) gimlet.Responder {
-	d, err := data.FindDistroById(h.distroID)
+	d, err := distro.FindOne(distro.ById(h.distroID))
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("distro with id '%s' not found", h.distroID),
+		}, "Database error for find() by distro id '%s'", h.distroID))
+	}
 	if err != nil {
```
ablack12, line 378: Also repeated error here

rest/route/distro.go
```diff
@@ -386,9 +415,12 @@ func (h *distroAMIHandler) Parse(ctx context.Context, r *http.Request) error {
 }
 
 func (h *distroAMIHandler) Run(ctx context.Context) gimlet.Responder {
-	d, err := data.FindDistroById(h.distroID)
+	d, err := distro.FindOne(distro.ById(h.distroID))
```
ablack12, line 418: Should check if distro is nil

rest/route/task.go
```diff
@@ -298,7 +298,10 @@ func (tep *taskExecutionPatchHandler) Run(ctx context.Context) gimlet.Responder
 				})
 			}
 		}
-		if err := data.SetTaskPriority(tep.task, tep.user.Username(), priority); err != nil {
+		if tep.task == nil {
```
ablack12, line 301: Did you encounter this? I would've thought it gets hit at line 268

rest/data/patch.go
hadjri: It is now being used in graphql package in ChildVersions since the data function that it called (that is now removed) used this function.

rest/data/project.go
hadjri: There was a public function wrapping this private function in 1 line, so it was essentially already public. I just removed the wrapper function and made this function public.

rest/route/distro.go
hadjri: Created and refactored all usages of this.

rest/route/task.go
hadjri: I was just carrying over the nil check that was previously in the SetTaskPriority method, but yes it looks like in this case we already do a nil check so I can remove this.

rest/route/distro.go
hadjri: distro.FindOne was a bit different from the other model FindOne functions in that it has no check for `if adb.ResultsNotFound(err)` in it, so a nil distro would also cause an error to return.  There are quite a few places where this function is used not touched by this PR, I was concerned that making a nil distro no longer an error in the DB function might cause unexpected panics/errors elsewhere where we rely on our distro DB result not being nil.

That being said I can change the distro DB functions and their usages to be more in line with how tasks/hosts/etc handle it but that may be out of scope for this PR since it addresses a different tech debt. Happy to create a ticket for it!

rest/data/project.go
ablack12: I think it was actually worth keeping that function, even though it just wraps that one, because it was clear then that that one should be used for Repos and the other should be used for Projects, and this is just a logical helper function.

ablack12: That being said I think we should move both into model/project_events.go (and then ProjectEventsBefore can also be made private)

rest/data/interface.go
```diff
@@ -54,348 +35,10 @@ type Connector interface {
...
-	GeneratePoll(context.Context, string, amboy.QueueGroup) (bool, []string, error)
-
-	// SaveSubscriptions, given an owner, a list of current subscriptions and an isProject boolean and
-	// upserts the given set of API subscriptions (if the owner is a project type, verify that all current subscriptions have
-	// the right owner and owner type).
-	SaveSubscriptions(string, []restModel.APISubscription, bool) error
-	// GetSubscriptions returns the subscriptions that belong to a user
-	GetSubscriptions(string, event.OwnerType) ([]restModel.APISubscription, error)
-	DeleteSubscriptions(string, []string) error
-	// CopyProjectSubscriptions copies subscriptions from the first project for the second project.
```
ablack12, line 351: Can you move this comment (and others if needed) to be function comments?

rest/route/distro.go
```diff
@@ -252,17 +250,13 @@ func (h *distroIDDeleteHandler) Parse(ctx context.Context, r *http.Request) erro
 
 // Run deletes a distro by id.
 func (h *distroIDDeleteHandler) Run(ctx context.Context) gimlet.Responder {
-	_, err := distro.FindOne(distro.ById(h.distroID))
+	_, err := distro.FindOneId(h.distroID)
```
ablack12, line 253: We should verify here that the distro isn't nil

model/distro/db.go
```diff
@@ -83,6 +83,11 @@ func FindOne(query db.Q) (Distro, error) {
 	return d, db.FindOneQ(Collection, query, &d)
 }
 
+// FindOneId returns one Distro by Id.
+func FindOneId(id string) (Distro, error) {
```
ablack12, line 87: Sorry I meant to use FindByID which already exists and correctly returns nil, nil if the item doesn't exist. (Unless returning a pointer causes a lot of issues but then we should make a ticket to get rid of this function bc it doesn't fit convention, and we should call it something like FindOneIdWithoutPointer)

rest/route/distro.go
ablack12: Sorry I actually addressed this in a different comment I can't find haha but yeah a different ticket seems reasonable!

rest/data/interface.go
hadjri: I thought I had caught all of these but I suppose not and missed a few - moved the remaining comments

rest/data/commit_queue.go
```diff
@@ -165,6 +166,7 @@ func writePatchInfo(patchDoc *patch.Patch, patchSummaries []thirdparty.Summary,
 	return nil
 }
 
+// if bool is true, move the commit queue item to be processed next.
```
ablack12, line 169: nit: comments should start with function name (can we just shore this comment up though, since that just explains enqueueNext)

rest/data/commit_queue.go
```diff
@@ -248,7 +250,7 @@ func (pc *DBCommitQueueConnector) CommitQueueRemoveItem(identifier, issue, user
 	return &apiRemovedItem, nil
 }
 
-func (pc *DBCommitQueueConnector) IsItemOnCommitQueue(id, item string) (bool, error) {
+func IsItemOnCommitQueue(id, item string) (bool, error) {
```
ablack12, line 253: This is only used by one route; can we just make this a helper function?

rest/data/commit_queue.go
```diff
@@ -264,7 +266,7 @@ func (pc *DBCommitQueueConnector) IsItemOnCommitQueue(id, item string) (bool, er
 	return false, nil
 }
 
-func (pc *DBCommitQueueConnector) CommitQueueClearAll() (int, error) {
+func CommitQueueClearAll() (int, error) {
```
ablack12, line 269: Can we get rid of this function and just call commitqueue.ClearAllCommitQueues directly? (We can probably get rid of the tests too since ClearAllCommitQueues has its own tests)

rest/data/commit_queue.go
```diff
@@ -331,7 +333,7 @@ func (pc *DBCommitQueueConnector) CreatePatchForMerge(ctx context.Context, exist
 	return apiPatch, nil
 }
 
-func (pc *DBCommitQueueConnector) GetMessageForPatch(patchID string) (string, error) {
+func GetMessageForPatch(patchID string) (string, error) {
```
ablack12, line 336: Can we move this into model and out of rest?

rest/data/task.go
```diff
@@ -177,18 +95,9 @@ func (tc *DBTaskConnector) FindTasksByProjectAndCommit(project, commitHash, task
...
-		return errors.New("task cannot be nil")
-	}
-	return model.SetTaskPriority(*t, priority, user)
-}
-
-// SetTaskPriority changes the priority value of a task using a call to the
+// SetTaskActivated changes the priority value of a task using a call to the
 // service layer function.
-func (tc *DBTaskConnector) SetTaskActivated(taskId, user string, activated bool) error {
+func SetTaskActivated(taskId, user string, activated bool) error {
```
ablack12, line 100: This  task is unnecessary; we can just call SetActiveStateById directly https://github.com/evergreen-ci/evergreen/blob/cfb0a7a16e8bf6b22d79f196b94eef3c37e1e1d5/model/task_lifecycle.go#L142

rest/data/task.go
```diff
@@ -203,7 +112,7 @@ func (tc *DBTaskConnector) SetTaskActivated(taskId, user string, activated bool)
 
 // ResetTask sets the task to be in an unexecuted state and prepares it to be run again.
 // If given an execution task, marks the display task for reset.
-func (tc *DBTaskConnector) ResetTask(taskId, username string) error {
+func ResetTask(taskId, username string) error {
```
ablack12, line 115: Can we get rid of this function and either move this logic directly into the route that uses it or make it a helper function for the route?

rest/data/task.go
```diff
@@ -228,7 +133,7 @@ func (tc *DBTaskConnector) CheckTaskSecret(taskID string, r *http.Request) (int,
 }
 
 // GetManifestByTask finds the manifest corresponding to the given task.
-func (tc *DBTaskConnector) GetManifestByTask(taskId string) (*manifest.Manifest, error) {
+func GetManifestByTask(taskId string) (*manifest.Manifest, error) {
```
ablack12, line 136: Is it possible to move this into model/manifest? Or move the logic into the getManifestHandler directly.

rest/data/task.go
```diff
@@ -263,7 +168,7 @@ type TaskFilterOptions struct {
 
 // FindTasksByVersion gets all tasks for a specific version
 // Results can be filtered by task name, variant name and status in addition to being paginated and limited
-func (tc *DBTaskConnector) FindTasksByVersion(versionID string, opts TaskFilterOptions) ([]task.Task, int, error) {
+func FindTasksByVersion(versionID string, opts TaskFilterOptions) ([]task.Task, int, error) {
```
ablack12, line 171: This isn't used anymore since we're calling GetTasksByVersion directly so we can just remove it.

rest/data/test_test.go
```diff
@@ -105,51 +103,51 @@ func TestGetTestCountByTaskIdAndFilter(t *testing.T) {
 
 	for i := 0; i < numTasks; i++ {
 		taskId := fmt.Sprintf("task_%d", i)
-		count, err := serviceContext.GetTestCountByTaskIdAndFilters(taskId, "", []string{}, 0)
+		count, err := task.GetTestCountByTaskIdAndFilters(taskId, "", []string{}, 0)
```
ablack12, line 106: All tests (like this one) that now no longer test functions in the rest/data package should be moved to the appropriate package. We can split that to a different ticket if you prefer.

ablack12 (approved):
Did a last audit of rest/data to see if anything else was redundant so that's this last batch of comments. I'm marking myself as approved so you're not blocked on me tomorrow but I'm happy to look again or discuss Monday if something isn't as straightforward as I thought. Great work on this! 🚀

rest/data/task.go
hadjri: Yup, that fixes the import cycle error I found when moving it into model/task

rest/data/test_test.go
hadjri: I took a crack at it and it looks like moving all the tests around caused quite a few context errors (and I've since reverted those changes) and might be slightly more work than anticipated. I created https://jira.mongodb.org/browse/EVG-16570 to bring in next sprint.

PR #5472, opened 2022-03-09
EVG-16466: mark when a dependency is finished

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16466

### Description
The new container pipeline needs to avoid allocating containers to tasks that have unmet dependencies, because it would be wasteful to allocate resources to a task that can't run yet. It's not possible to do that efficiently right now - the only way to check if a task's dependencies are met is to call `DependenciesMet`, which would be expensive to do every time we want to check if a task can be allocated a container. To get around this, I added a field to cache when a dependency finishes running. Checking if all dependencies are finished + attainable should be sufficient to know if a task's dependencies are all met.

* Add field to indicate when a dependency is finished running.
* Add method to mark dependency as finished/unfinished.
* Mark dependency as finished when a task finishes running.
* Mark dependency as unfinished when a task is restarted.

### Testing 
* Added unit tests. 
* Verified manually in staging with various types of tasks (with dependencies, without dependencies, single-host task group tasks, display tasks, execution tasks) that the dependency is marked as finished when the task ends and marked as unfinished when the task is restarted.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/simple"
 	"gonum.org/v1/gonum/graph/topo"
@@ -63,7 +64,7 @@ type Task struct {
 	// time information for task
 	// Create - the creation time for the task, derived from the commit time or the patch creation time.
 	// Dispatch - the time the task runner starts up the agent on the host.
-	// Scheduled - the time the commit is scheduled.
+	// Scheduled - the time the task is scheduled.
 	// Start - the time the agent starts the task on the host after spinning it up.
 	// Finish - the time the task was completed on the remote host.
 	// Activated - the time the task was marked as available to be scheduled, automatically or by a developer.
@@ -260,6 +261,8 @@ type Dependency struct {
 	TaskId       string `bson:"_id" json:"id"`
 	Status       string `bson:"status" json:"status"`
 	Unattainable bool   `bson:"unattainable" json:"unattainable"`
+	// Finished indicates if the dependency has finished running or not.
+	Finished bool `bson:"finished" json:"finished"`
 }
 
 // BaseTaskInfo is a subset of task fields that should be returned for patch tasks.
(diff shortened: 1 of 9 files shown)


model/task/task.go
```diff
@@ -775,6 +778,39 @@ func (t *Task) AllDependenciesSatisfied(cache map[string]Task) (bool, error) {
 	return true, nil
 }
 
+// MarkDependenciesFinished updates all direct dependencies on this task to
+// cache whether or not this task has finished running.
+func (t *Task) MarkDependenciesFinished(finished bool) error {
```
julianedwards, line 783: Is there ever a case where we would need to mark this as `false`?

Kimchelly: Yes, it gets marked false (i.e. task is not finished) when the task is restarted to indicate that the dependency is no longer finished running.

It would be preferable to derive the finished state from the task's in-memory status (e.g. "completed", "failed") rather than pass it in as a parameter, but I'm not confident that `RestartVersion`, `RestartBuild`, and `RestartAllBuildTasks` (which call `MarkTasksReset`) actually reset execution tasks to the "undispatched" status. It seems like [`MarkTasksReset` only resets the status of display tasks](https://github.com/evergreen-ci/evergreen/blob/d08da70677174ecc33f9b70e2d0e4d708269e129/model/task_lifecycle.go#L1253-L1261), so when I wrote a test for `RestartVersions` it failed to mark the task's dependencies as unfinished because the task was never reset. Hence, I had to manually tell it to unset the dependencies.

I'm not entirely confident that the restart logic is correct for builds/versions, but I don't really want to mess with the existing logic to find out. There's [a ticket](https://jira.mongodb.org/browse/EVG-16480) to investigate the confusion I mentioned above.

malikchaya2, line 783: Can we maybe call this function SetDependenciesFinished? I think that may make it clearer that we are setting a db field as opposed to marking finished as true.

model/task/task.go
```diff
@@ -775,6 +778,39 @@ func (t *Task) AllDependenciesSatisfied(cache map[string]Task) (bool, error) {
...
+		},
+		bson.M{
+			"$set": bson.M{bsonutil.GetDottedKeyName(DependsOnKey, "$[elem]", DependencyFinishedKey): finished},
+		},
+		options.Update().SetArrayFilters(options.ArrayFilters{Filters: []interface{}{
+			bson.M{bsonutil.GetDottedKeyName("elem", DependencyTaskIdKey): t.Id},
+		}}),
+	)
+	if err != nil {
+		return errors.Wrap(err, "marking finished dependencies")
```
malikchaya2, line 808: Can we change this error to `error setting finished dependencies to '%s', finished`?

model/task/task.go
Kimchelly: Evergreen is inconsistent about this because Set, Mark, and Update are all used interchangeably when updating something in the DB. For example, `MarkUnattainableDependencies` is extremely similar to this function. I would prefer to leave it alone since it's consistent with the other dependency-modifying functions.

model/task/task.go
Kimchelly: I intentionally don't write my error messages starting with "failed to do xyz" or "error doing xyz" or "cannot do xyz" because they tend to produce excessively verbose error messages. Instead, I write them in the style of the Go standard library (which is the golden standard for Go code), which, when wrapping errors with more context, says what operation was being done but doesn't produce ugly wrapped errors that look like "error doing abc: failed to do def: could not do xyz: ". My experience with this is that it's much easier to read error messages in logs without this noise.

This error message style isn't consistent across Evergreen teams, so I just use the preference of PLT, which is no noisy error messages.

malikchaya2: That makes sense. Do you think it would be worth creating a tech debt ticket to modify the error messages across our codebase to use this style?

malikchaya2 (approved):

julianedwards (approved):

model/task/task.go
Kimchelly: Sure, I made [EVG-16501](https://jira.mongodb.org/browse/EVG-16501) to clean it up in all the Evergreen-owned repos and [EVG-16504](https://jira.mongodb.org/browse/EVG-16504) to clean it up in Evergreen itself.

PR #5475, opened 2022-03-10
EVG-1822 Update set-module command to not require patch id

bynn:
[EVG-1822](https://jira.mongodb.org/browse/EVG-1822)

### Description 
set-module now defaults to the user's latest patch when not given a patch id

### Testing 
added evergreen as module patch to spruce patch

image
https://spruce.mongodb.com/version/6229a92261837d493fdce229/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/operations/http.go
+++ b/operations/http.go
@@ -256,27 +256,6 @@ func (ac *legacyClient) GetPatch(patchId string) (*patch.Patch, error) {
 	return &res, nil
 }
 
-// GetLatestPatchByUser gets the latest patch from the given user id and returns it as a Patch.
-func (ac *legacyClient) GetLatestPatchByUser(userId string) (*patch.Patch, error) {
-	resp, err := ac.get2(fmt.Sprintf("users/%v/patches?limit=1", userId), nil)
-	if err != nil {
-		return nil, err
-	}
-	defer resp.Body.Close()
-
-	if resp.StatusCode == http.StatusUnauthorized {
-		return nil, client.AuthError
-	}
-	if resp.StatusCode != http.StatusOK {
-		return nil, NewAPIError(resp)
-	}
-	patches := []patch.Patch{}
-	if err := utility.ReadJSON(resp.Body, &patches); err != nil {
-		return nil, err
-	}
-	return &patches[0], nil
-}
-
 // GetProjectRef requests project details from the API server for a given project ID.
 func (ac *legacyClient) GetProjectRef(projectId string) (*model.ProjectRef, error) {
 	resp, err := ac.get(fmt.Sprintf("/ref/%s", projectId), nil)
--- a/operations/http.go
+++ b/operations/http.go
@@ -256,6 +256,27 @@ func (ac *legacyClient) GetPatch(patchId string) (*patch.Patch, error) {
 	return &res, nil
 }
 
+// GetLatestPatchByUser gets the latest patch from the given user id and returns it as a Patch.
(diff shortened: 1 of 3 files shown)


operations/patch_module.go
```diff
@@ -57,14 +57,23 @@ func PatchSetModule() cli.Command {
...
 			}
 
-			existingPatch, err := ac.GetPatch(patchID)
+			var existingPatch *patch.Patch
+			if patchID == "" {
+				patchList, err := ac.GetPatches(1)
+				if err != nil {
+					return errors.Wrapf(err, "problem getting patches from user")
+				}
+				existingPatch = &patchList[0]
```
malikchaya2, line 66: should we make sure that there is at least one patch in the list before accessing the index, so that we can avoid out of bounds errors?

operations/http.go
```diff
@@ -256,6 +256,27 @@ func (ac *legacyClient) GetPatch(patchId string) (*patch.Patch, error) {
...
+		return nil, client.AuthError
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, NewAPIError(resp)
+	}
+	patches := []patch.Patch{}
+	if err := utility.ReadJSON(resp.Body, &patches); err != nil {
+		return nil, err
+	}
+	return &patches[0], nil
```
malikchaya2, line 277: same comment about checking the length first

malikchaya2 (changes requested):
This is a fantastic idea!

operations/patch_module.go
```diff
@@ -27,7 +28,6 @@ func PatchSetModule() cli.Command {
 		Before: mergeBeforeFuncs(
 			autoUpdateCLI,
 			setPlainLogger,
-			requirePatchIDFlag,
```
ablack12, line 30: Can we add text to the patch ID flag usage to say that we default to the latest submitted patch?

ablack12 (changes requested):
Also need to increment client version

operations/http.go
bynn: unused function oops

operations/patch_module.go
```diff
@@ -57,14 +57,23 @@ func PatchSetModule() cli.Command {
 				return errors.Wrap(err, "problem accessing evergreen service")
 			}
 
-			existingPatch, err := ac.GetPatch(patchID)
+			var existingPatch *patch.Patch
+			if patchID == "" {
+				patchList, err := ac.GetPatches(1)
```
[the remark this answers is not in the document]
bynn: got rid of the function

operations/patch_module.go
bynn: added check

operations/patch_module.go
bynn: added it manually but if there's a better way to do this that preserves the order let me know!

operations/patch_module.go
```diff
@@ -57,14 +61,23 @@ func PatchSetModule() cli.Command {
...
+				if err != nil || len(patchList) != 1 {
```
ablack12, line 67: in the case that the error is len(patchList) != 1 , then we'll return a nil error since we're wrapping a nil error. We also could return something more relevant in this case ("no patches for user" if len(patchList) == 0)

ablack12 (approved):
LGTM with fix

operations/patch_module.go
malikchaya2: +1

malikchaya2 (approved):
LGTM after Annie's comment is addressed :)

PR #5476, opened 2022-03-10
EVG-15086: Update WorkstationConfigInput

sophstad:
Relates to UI ticket [EVG-15086](https://jira.mongodb.org/browse/EVG-15086)

### Description 
- Make `gitClone` nullable to support projects that default to repo
- Disallow null entry in `setupCommands` array

### Testing 
- Tested manually


ablack12 (approved):
LGTM, sorry I got all of these so mixed up when I did this 😅

PR #5477, opened 2022-03-10
EVG-16045 Remove restricted variables

bynn:
[EVG-16045](https://jira.mongodb.org/browse/EVG-16045)

### Description 
deprecate restricted vars 

### Testing 
all existing tests pass

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -979,7 +979,6 @@ func getCommonProjectVariables(projectIds []string) (*ProjectVars, error) {
 	// add in project variables and aliases here
 	commonProjectVariables := map[string]string{}
 	commonPrivate := map[string]bool{}
-	commonRestricted := map[string]bool{}
 	commonAdminOnly := map[string]bool{}
 	for i, id := range projectIds {
 		vars, err := FindOneProjectVars(id)
@@ -996,22 +995,16 @@ func getCommonProjectVariables(projectIds []string) (*ProjectVars, error) {
 			if vars.PrivateVars != nil {
 				commonPrivate = vars.PrivateVars
 			}
-			if vars.RestrictedVars != nil {
-				commonRestricted = vars.RestrictedVars
-			}
 			if vars.AdminOnlyVars != nil {
 				commonAdminOnly = vars.AdminOnlyVars
 			}
 			continue
 		}
 		for key, val := range commonProjectVariables {
-			// if the key is private/restricted in any of the projects, make it private/restricted in the repo
+			// if the key is private/admin only in any of the projects, make it private/admin only in the repo
 			if vars.Vars[key] == val {
 				if vars.PrivateVars[key] {
 					commonPrivate[key] = true
-				}
-				if vars.RestrictedVars[key] {
-					commonRestricted[key] = true
 				}
 				if vars.AdminOnlyVars[key] {
 					commonAdminOnly[key] = true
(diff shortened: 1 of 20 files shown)


model/project_ref.go
```diff
@@ -996,23 +995,17 @@ func getCommonProjectVariables(projectIds []string) (*ProjectVars, error) {
...
-				commonRestricted = vars.RestrictedVars
-			}
 			if vars.AdminOnlyVars != nil {
 				commonAdminOnly = vars.AdminOnlyVars
 			}
 			continue
 		}
 		for key, val := range commonProjectVariables {
-			// if the key is private/restricted in any of the projects, make it private/restricted in the repo
+			// if the key is private/admin only in any of the projects, make it private/admin only in the repo
```
ablack12, line 1004: I know you didn't write this (and I probably did womp) but could you turn this comment into a full sentence?

model/project_vars.go
```diff
@@ -212,19 +203,19 @@ func (projectVars *ProjectVars) FindAndModify(varsToDelete []string) (*adb.Chang
...
 	}
-	return restrictedVars
+	for k, v := range projectVars.GetAdminOnlyVars(t) {
+		vars[k] = v
+	}
+	return vars
 }
 
-// GetAdminOnlyVars currently returns admin only variables for mainline commits and project admins.
-// Will change to GetVars on EVG-16045 after removing restricted vars.
```
ablack12, line 227: I think rather than GetVars calling this function and looping through all vars twice, this could be a helper function that just determines the "shouldUseAdminOnlyVars" part, and then GetVars can just loop through the vars once and include private vars if shouldUseAdminOnlyVars is true.

ablack12 (changes requested):

model/project_vars.go
bynn: done

model/project_ref.go
bynn: done!

model/project_vars.go
```diff
@@ -206,32 +206,24 @@ func (projectVars *ProjectVars) FindAndModify(varsToDelete []string) (*adb.Chang
 func (projectVars *ProjectVars) GetVars(t *task.Task) map[string]string {
 	vars := map[string]string{}
 	for k, v := range projectVars.Vars {
-		if !projectVars.AdminOnlyVars[k] {
+		if !projectVars.AdminOnlyVars[k] || projectVars.ShouldGetAdminOnlyVars(t) {
```
ablack12, line 209: We should store ShouldgetAdminOnlyVars as a variable so we aren't re-calculating it on each iteration of the if statement

malikchaya2 (approved):

model/project_vars.go

model/project_vars.go
```diff
@@ -250,36 +234,19 @@ func (projectVars *ProjectVars) GetAdminOnlyVars(t *task.Task) map[string]string
 				RequiredLevel: evergreen.ProjectSettingsEdit.Value,
 			})
 			if isAdmin {
```
ablack12, line 236: we can just return isAdmin here

model/project_vars.go
```diff
@@ -212,35 +203,28 @@ func (projectVars *ProjectVars) FindAndModify(varsToDelete []string) (*adb.Chang
...
-	adminOnlyVars := map[string]string{}
+func (projectVars *ProjectVars) ShouldGetAdminOnlyVars(t *task.Task) bool {
 	if utility.StringSliceContains(evergreen.SystemVersionRequesterTypes, t.Requester) {
-		for k, v := range projectVars.Vars {
-			if projectVars.AdminOnlyVars[k] {
-				adminOnlyVars[k] = v
-			}
-		}
+		return true
 	} else if t.ActivatedBy != "" {
```
ablack12, line 220: Now that we're returning, it's cleaner to return more frequently and get things out of if statements. So by checking `if t.ActivatedBy == "" { return false}` at line 220, everything after that can now be out of the else if statement, if that makes sense

model/project_vars.go

ablack12 (approved):

PR #5478, opened 2022-03-10
EVG-16397 remove regex selectors from filter

ybrill:
[EVG-16397](https://jira.mongodb.org/browse/EVG-16397)

### Description 
It was overly ambitious to think the regex selector matching could be pushed to the db. I got confused by the direction: we aren't matching a regex to a string in the document. Rather, we're matching a string to a regex in the document.
It's possible we can do something down the line with [$regexFind](https://docs.mongodb.com/manual/reference/operator/aggregation/regexFind/) but it's a 4.2 feature.

My plan is to 
1. deploy this
2. run `db.subscriptions.updateMany({}, {$unset: {filter:""}}` (maybe starting with a few to make sure I'm not dropping the db) to clear all the filter fields
3. rerun the migration script with the lines to set the regex field removed

### Testing 
  None.

### DIFF
--- a/graphql/tests/saveSubscription/results.json
+++ b/graphql/tests/saveSubscription/results.json
@@ -82,7 +82,7 @@
       "result": {
         "errors": [
           {
-            "message": "error saving subscription: 400 (Bad Request): Error parsing request body: setting filter from selectors: selector has an empty type\nselector '' has no data",
+            "message": "error saving subscription: 400 (Bad Request): Error validating subscription: selector has an empty type\nselector '' has no data",
             "path": ["saveSubscription"],
             "extensions": { "code": "INTERNAL_SERVER_ERROR" }
           }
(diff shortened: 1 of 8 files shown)


Kimchelly (commented):
Is it still possible to utilize the proposed index efficiently if regex selectors aren't migrated to the new filter field?

ybrill:
It will definitely help. How much we'll have to see.

At the moment there aren't any regex selectors that don't have at least two regular selectors. I think the UI won't create them. The ones I checked seemed like the regular selectors were pretty unique, at least as much as other subscriptions. In other words, a particular event would only match a small number of them. Like one specified the project and the requester type and there were only a handful of such subscriptions that were also looking for that project and requester type.

I put back the validation change that would make sure we don't create one without any selectors in the future through the API.

Kimchelly (approved):

Kimchelly:
LGTM mod the tests passing.

ybrill:
evergreen retry

ybrill:
I'm going to merge this. 
The remaining failure is the Spruce e2e_test that's broken in the same way on the spruce waterfall and doesn't seem connected.

PR #5479, opened 2022-03-10
EVG-15555 Add aborted field to version

bynn:
[EVG-15555](https://jira.mongodb.org/browse/EVG-15555)

### Description 
updatebuildstatus loops through tasks and update versionstatus looks through those builds and checks for any aborted tasks

### Testing 
unit test

### DIFF
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -47,6 +47,7 @@ type Build struct {
 	DisplayName         string        `bson:"display_name" json:"display_name,omitempty"`
 	PredictedMakespan   time.Duration `bson:"predicted_makespan" json:"predicted_makespan,omitempty"`
 	ActualMakespan      time.Duration `bson:"actual_makespan" json:"actual_makespan,omitempty"`
+	Aborted             bool          `bson:"aborted" json:"aborted,omitempty"`
 
 	// The status of the subset of the build that's used for github checks
 	GithubCheckStatus string `bson:"github_check_status,omitempty" json:"github_check_status,omitempty"`
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -919,6 +919,13 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
 		return false, errors.Wrapf(err, "getting tasks in build '%s'", b.Id)
 	}
 
+	b.Aborted = false
+	for _, t := range buildTasks {
+		if t.Aborted {
+			b.Aborted = true
+			break
+		}
+	}
 	buildStatus := getBuildStatus(buildTasks)
 
 	if buildStatus == b.Status {
@@ -1003,6 +1010,13 @@ func UpdateVersionStatus(v *Version) (string, error) {
 	builds, err := build.Find(build.ByVersion(v.Id).WithFields(build.ActivatedKey, build.StatusKey, build.IsGithubCheckKey))
 	if err != nil {
 		return "", errors.Wrapf(err, "getting builds for version '%s'", v.Id)
+	}
+
+	for _, b := range builds {
+		if b.Aborted {
+			v.Aborted = true
+			break
+		}
 	}
 	versionStatus := getVersionStatus(builds)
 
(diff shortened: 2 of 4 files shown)


model/task_lifecycle.go
```diff
@@ -919,6 +919,13 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
...
+	b.Aborted = false
+	for _, t := range buildTasks {
+		if t.Aborted {
+			b.Aborted = true
+			break
+		}
+	}
 	buildStatus := getBuildStatus(buildTasks)
 
 	if buildStatus == b.Status {
```
ablack12, line 931: If the build status hasn't changed, I think the build aborted status also couldn't have changed, so there's no reason to check for b.Aborted until after this.

model/task_lifecycle.go
```diff
@@ -919,6 +919,13 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
 		return false, errors.Wrapf(err, "getting tasks in build '%s'", b.Id)
 	}
 
+	b.Aborted = false
+	for _, t := range buildTasks {
+		if t.Aborted {
+			b.Aborted = true
```
ablack12, line 925: We're only updating this locally it looks like; need to update in the DB as well (but only if the aborted state has actually changed I think)

ablack12 (changes requested):

model/task_lifecycle.go
bynn: updating the db when changed

model/task_lifecycle.go
bynn: only updating if status changed

model/build/build.go
```diff
@@ -172,6 +173,17 @@ func (b *Build) UpdateStatus(status string) error {
 	)
 }
 
+func (b *Build) SetAbortedStatus(aborted bool) error {
```
ablack12, line 176: This should just say SetAborted; "Status" is confusing since this doesn't actually set status

model/task_lifecycle.go
```diff
@@ -925,6 +925,19 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
 		return false, nil
 	}
 
+	abortStatus := false
```
ablack12, line 928: similarly, isAborted seems more relevant than abortStatus

model/task_lifecycle.go
```diff
@@ -925,6 +925,19 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
...
+	for _, t := range buildTasks {
+		if t.Aborted {
+			abortStatus = true
+			break
+		}
+	}
+	if abortStatus != b.Aborted {
+		if err = b.SetAbortedStatus(abortStatus); err != nil {
+			return false, errors.Wrapf(err, "setting aborted status for build '%s'", b.Id)
```
ablack12, line 937: should also tweak this comment

ablack12 (commented):
This functionality looks right to me but it's up to Mohamed to determine if this will work for the UI use case.

model/build/build.go
malikchaya2: +1

model/version.go
```diff
@@ -155,6 +156,18 @@ func (self *Version) SetNotActivated() error {
 	)
 }
 
+func (self *Version) SetAbortedStatus(aborted bool) error {
```
malikchaya2, line 159: can we rename this to just SetAborted as well?

model/build/build.go
bynn: done

model/task_lifecycle.go

model/task_lifecycle.go
bynn: done for build and version

model/version.go
bynn: done

malikchaya2 (approved):

model/task/db.go
```diff
@@ -360,7 +360,7 @@ func ByAborted(aborted bool) bson.M {
 	}
 }
 
-// ByAborted creates a query to return tasks with an aborted state
+// ByActivation creates a query to return tasks with an activated state
```
ablack12, line 363: nice catch!

model/task_lifecycle.go
```diff
@@ -925,6 +925,19 @@ func UpdateBuildStatus(b *build.Build) (bool, error) {
 		return false, nil
 	}
 
+	isAborted := false
```
ablack12, line 928: Can we add a comment that's like "only need to check if aborted has changed if status has changed" or something so it's clear that this is intentional?

ablack12 (approved):
LGTM with nit

khelif96 (approved):

PR #5481, opened 2022-03-11
EVG-1498 add interface to cancel stepback

ablack12:
[EVG-1498](https://jira.mongodb.org/browse/EVG-1498)

### Description 
The first commit can be ignored because it's just moving functions that _should_ be in a db_test.go file out of task_test.go.

Users have been asking (for a long time it seems) for an interface to cancel stepback, so they don't have to hunt down currently running stepback tasks when they know a failure. This deactivates/aborts them but doesn't turn off stepback for the project.

### Testing 
Added a unit test and tested on staging that https://evergreen-staging.corp.mongodb.com/task/sandbox_ubuntu1604_unit_tests_05f962b68a4300dd01101ae3df578991829b9746_22_03_11_19_56_46 was aborted after saving with the checkbox.
image


model/task/task_test.go
```diff
@@ -1284,6 +1284,83 @@ func TestUnscheduleStaleUnderwaterHostTasksNoDistro(t *testing.T) {
 	assert.EqualValues(-1, dbTask.Priority)
 }
 
+func TestDeactivateStepbackTasksForProject(t *testing.T) {
```
ablack12, line 1287: This is the only actually new test so no other tests should be reviewed.

model/task/db.go
```diff
@@ -510,18 +504,6 @@ func ByBeforeRevision(revisionOrder int, buildVariant, displayName, project, req
 	}, []string{"-" + RevisionOrderNumberKey}
 }
 
-// ByBuildIdAfterTaskId provides a way to get an ordered list of tasks from a
-// build. Providing a taskId allows indexing into the list of tasks that
-// naturally exists when tasks are sorted by taskId.
-func ByBuildIdAfterTaskId(buildId, taskId string) (bson.M, []string) {
```
malikchaya2, line 516: nice tech debt cleanup!

model/task/db.go
```diff
@@ -1675,3 +1657,16 @@ func AddHostCreateDetails(taskId, hostId string, execution int, hostCreateError
 		}})
 	return errors.Wrapf(err, "error adding details of host creation failure to task")
 }
+
+func FindActivatedStepbackTasks(projectId string) ([]Task, error) {
+	tasks, err := Find(bson.M{
+		ProjectKey:     projectId,
+		ActivatedKey:   true,
+		ActivatedByKey: evergreen.StepbackTaskActivator,
+		StatusKey:      bson.M{"$in": evergreen.TaskUncompletedStatuses},
```
malikchaya2, line 1666: I happened to notice that TaskContainerUnallocated and TaskStatusPending are not part of TaskUncompletedStatuses. Should it be?

service/templates/projects.html
```diff
@@ -275,6 +275,14 @@ <h3> Admins </h3>
           <div class="muted small">When checked, tasks from previous revisions will be unscheduled when the equivalent task in a newer commit finishes successfully.</div>
         </div>
       </div>
+      <div class="form-group">
```
malikchaya2, line 278: should we create a ticket to add this to the new project page?

malikchaya2 (changes requested):

service/templates/projects.html
ablack12: Created https://jira.mongodb.org/browse/EVG-16360 already but I'm not sure if it actually belongs on the project page going forward or if something on mainline would be more intuitive.

model/task/db.go
ablack12: Pending is a UI state, not a real status. Unallocated does seem relevant.

malikchaya2 (approved):

PR #5484, opened 2022-03-14
EVG-15700 SetActiveState accepts a list of tasks

hadjri:
[EVG-15700](https://jira.mongodb.org/browse/EVG-15700)

### Description 
SetActiveState now accepts a list of tasks and performs a bulk DB operation on activating tasks and versions, likely improving performance. This is passed in through setScheduled in our graphql resolvers.
### Testing 
Modified unit tests.


graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+	if tasks == nil {
+		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task IDs %s not found", strings.Join(taskIDs, ",")).Error())
 	}
-	if t.Requester == evergreen.MergeTestRequester && isActive {
-		return nil, InputValidationError.Send(ctx, "commit queue tasks cannot be manually scheduled")
+	taskPtrs := []*task.Task{}
+	for _, t := range tasks {
+		taskPtrs = append(taskPtrs, &t)
+		if t.Requester == evergreen.MergeTestRequester && isActive {
+			return nil, InputValidationError.Send(ctx, "commit queue tasks cannot be manually scheduled")
```
malikchaya2, line 76: Do we want to use a catcher here so that the behavior mimics what the behavior was when we had a for loop in the resolver? Or is returning  when one errors and not attempting to schedule the rest the preferred setup?

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
+	tasksToActivate := []task.Task{}
+	versionsToActivate := []Version{}
+	for _, t := range tasks {
+		originalTasks := []task.Task{*t}
+		originalTasks = append(originalTasks, *t)
+		if t.DisplayOnly {
+			execTasks, err := task.Find(task.ByIds(t.ExecutionTasks))
 			if err != nil {
-				return errors.Wrapf(err, "error getting tasks '%s' depends on", t.Id)
+				return errors.Wrapf(err, "error getting execution tasks")
```
malikchaya2, line 43: same question about using a catcher

model/version_db.go
```diff
@@ -314,6 +314,27 @@ func VersionUpdateOne(query interface{}, update interface{}) error {
 	)
 }
 
+func ActivateVersions(versions []Version) error {
```
malikchaya2, line 317: since all we need from the Version object is the id, does it make sense to instead take in a []string of version ids?

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
 	usr := mustHaveUser(ctx)
-	t, err := task.FindOneId(taskID)
+	tasks, err := task.FindAll(db.Query(task.ByIds(taskIDs)))
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
-	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task %s not found", taskID).Error())
+	if tasks == nil {
+		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task IDs %s not found", strings.Join(taskIDs, ",")).Error())
```
khelif96, line 70: An array of taskIds would not make for a useful/readable error message Since now there could be thousands of taskIds. Is there something else we could return?

graphql/util.go
khelif96: I would expect all of the tasks in the same version to have the same requester?

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
 	}
-	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, err.Error())
-	}
-	apiTask := restModel.APITask{}
-	err = apiTask.BuildFromService(t)
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, err.Error())
+	if tasks == nil {
+		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task IDs %s not found after active state was set", strings.Join(taskIDs, ",")).Error())
```
khelif96, line 89: Same comment about taskIds in an error.

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
-		// if the task is being activated and it doesn't override its dependencies
-		// activate the task's dependencies as well
-		tasksToActivate := []task.Task{}
-		if !t.OverrideDependencies {
-			deps, err := task.GetRecursiveDependenciesUp(originalTasks, nil)
+func SetActiveState(caller string, active bool, tasks ...*task.Task) error {
+	tasksToActivate := []task.Task{}
+	versionsToActivate := []Version{}
+	for _, t := range tasks {
+		originalTasks := []task.Task{*t}
```
khelif96, line 38: curious about this syntax. Does this create an array with one task pointer in it?

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
+		originalTasks = append(originalTasks, *t)
```
khelif96, line 39: why are we appending the same task in to the array?
https://go.dev/play/p/amgkAOS4LnX

khelif96 (changes requested):

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
 	return &GroupedFiles{TaskName: &name, Files: apiFileList}, nil
 }
 
-func setScheduled(ctx context.Context, sc data.Connector, taskID string, isActive bool) (*restModel.APITask, error) {
+func setScheduled(ctx context.Context, sc data.Connector, isActive bool, taskIDs ...string) ([]*restModel.APITask, error) {
```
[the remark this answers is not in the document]
ablack12: Can we rename this to setManyTasksScheduled ?

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
 	usr := mustHaveUser(ctx)
-	t, err := task.FindOneId(taskID)
+	tasks, err := task.FindAll(db.Query(task.ByIds(taskIDs)))
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
-	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task %s not found", taskID).Error())
+	if tasks == nil {
```
ablack12, line 69: I think we should check if len(tasks) != len(taskIDs) to be logically equivalent

graphql/util.go
ablack12: Might be good in this case to determine which taskIds weren't found and just return those (or at least log them, since yeah there could still be too many to return to the user)

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+		}
 	}
-	if err = model.SetActiveState(t, usr.Username(), isActive); err != nil {
+	if err = model.SetActiveState(usr.Username(), isActive, taskPtrs...); err != nil {
 		return nil, InternalServerError.Send(ctx, err.Error())
 	}
 
 	// Get the modified task back out of the db
-	t, err = task.FindOneId(taskID)
+	tasks, err = task.FindAll(db.Query(task.ByIds(taskIDs)))
```
ablack12, line 84: Would it make sense to just take in a task list instead of pointers to tasks so we don't have to reconstruct the list?

graphql/util.go
```diff
@@ -59,40 +60,48 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+		if t.Requester == evergreen.MergeTestRequester && isActive {
+			return nil, InputValidationError.Send(ctx, "commit queue tasks cannot be manually scheduled")
+		}
 	}
-	if err = model.SetActiveState(t, usr.Username(), isActive); err != nil {
+	if err = model.SetActiveState(usr.Username(), isActive, taskPtrs...); err != nil {
 		return nil, InternalServerError.Send(ctx, err.Error())
 	}
 
 	// Get the modified task back out of the db
```
ablack12, line 83: nit: tasks

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
+		if t.DisplayOnly {
+			execTasks, err := task.Find(task.ByIds(t.ExecutionTasks))
```
ablack12, line 41: Can we compile a list of all exec tasks and just query them once?

model/task_lifecycle.go
ablack12: It dereferences the task pointer, so it's just the task object. it's the same as what's in line 35 in the original code

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
-				return errors.Wrapf(err, "error finding task group '%s'", t.TaskGroup)
+				return errors.Wrapf(err, "error find associated version")
+			}
+			if version == nil {
+				return errors.Errorf("could not find associated version : `%s`", t.Version)
 			}
-			for _, taskInGroup := range tasksInGroup {
-				if taskInGroup.TaskGroupOrder > t.TaskGroupOrder {
-					originalTasks = append(originalTasks, taskInGroup)
+			versionsToActivate = append(versionsToActivate, *version)
```
ablack12, line 94: Let's just keep a list of versionIdsToActivate so that we don't have to keep re-querying the version (especially since these will all commonly be for the same version) -- we can just use this list directly then in ActivateVersions

model/task_lifecycle.go
```diff
@@ -31,109 +31,114 @@ type StatusChanges struct {
...
 			}
 		}
-	} else {
-		return nil
-	}
 
-	if t.IsPartOfDisplay() {
-		if err := UpdateDisplayTaskForTask(t); err != nil {
-			return errors.Wrap(err, "problem updating display task")
+		if err := UpdateBuildAndVersionStatusForTask(t); err != nil {
```
ablack12, line 127: I think it doesn't work to call this before actually activating/deactivating the tasks. But also we really only need to call this once per build so we could keep a list of builds that tasksToActivate belongs to, and then iterate through that after the final if/else.

ablack12 (changes requested):

graphql/resolvers.go
```diff
@@ -2673,13 +2673,12 @@ func (r *mutationResolver) ScheduleUndispatchedBaseTasks(ctx context.Context, pa
...
 
+	taskIDs := make([]string, len(tasksToSchedule))
 	for taskId := range tasksToSchedule {
-		task, err := setScheduled(ctx, r.sc, taskId, true)
-		if err != nil {
-			return nil, err
-		}
-		scheduledTasks = append(scheduledTasks, task)
+		taskIDs = append(taskIDs, taskId)
 	}
```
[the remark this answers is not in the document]
hadjri: Thanks!

graphql/util.go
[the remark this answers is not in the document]
hadjri: There are a couple places that are going to pass in a single task, and others where a list is passed in.  I used a spread to handle both cases cleanly instead of converting the single task into an array.

graphql/resolvers.go
```diff
@@ -2673,13 +2673,12 @@ func (r *mutationResolver) ScheduleUndispatchedBaseTasks(ctx context.Context, pa
...
+	scheduled, err := setScheduled(ctx, r.sc, true, taskIDs...)
```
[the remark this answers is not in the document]
hadjri: I did this to support a single task string and a list of task strings as input.

graphql/resolvers.go
```diff
@@ -2788,31 +2787,33 @@ func (r *mutationResolver) EnqueuePatch(ctx context.Context, patchID string, com
...
 }
 
 func (r *mutationResolver) UnscheduleTask(ctx context.Context, taskID string) (*restModel.APITask, error) {
-	task, err := setScheduled(ctx, r.sc, taskID, false)
+	scheduled, err := setScheduled(ctx, r.sc, false, taskID)
 	if err != nil {
 		return nil, err
 	}
-	return task, nil
+	if len(scheduled) != 1 {
```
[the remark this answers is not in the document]
hadjri: In the cases where we are processing a single task in we'd only ever want 1 task to be returned, so yes that'd be the only valid state to receive an array of length 1 (now that setScheduled returns an array).  However it is a little unconventional if statement, I'll change it to if len(scheduled) == 0 to be less confusing.

model/task_lifecycle.go
hadjri: I believe it would interfere with this block: https://github.com/evergreen-ci/evergreen/blob/main/model/task_lifecycle.go#L49

Where we are computing the dependencies of a single display task + all its execution tasks in one batch.  If we retrieved all of the relevant execution tasks for all tasks at once, I think we could be recomputing the same dependencies over and over again and adding them to the list.

model/task_lifecycle.go
hadjri: Sgtm, since if it failed halfway through the loop there would be a subset of activated tasks anyway. Would likely be a bit more resilient.

graphql/util.go
```diff
@@ -58,40 +58,60 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+			"message":    "could not find all tasks",
+			"taskIds":    taskIDs,
+			"foundTasks": tasks,
+		})
+		return nil, ResourceNotFound.Send(ctx, errors.New("tasks not found").Error())
+	}
+	catcher := grip.NewBasicCatcher()
+	for _, t := range tasks {
+		if t.Requester == evergreen.MergeTestRequester && isActive {
+			catcher.Add(InputValidationError.Send(ctx, "commit queue tasks cannot be manually scheduled"))
```
malikchaya2, line 78: I may have been wrong suggesting a catcher in this case considering Mohamed's comment that we would expect all of the tasks in the same version to have the same requester. We may want to just return the error in this case.

model/version_db.go
```diff
@@ -315,6 +315,23 @@ func VersionUpdateOne(query interface{}, update interface{}) error {
 	)
 }
 
+func ActivateVersions(versions []string) error {
```
malikchaya2, line 318: Can we please call these versionIds instead of versions?

malikchaya2 (approved):

model/task_lifecycle.go
```diff
@@ -31,111 +31,123 @@ type StatusChanges struct {
...
+				deps, err := task.GetRecursiveDependenciesUp(originalTasks, nil)
+				if err != nil {
+					catcher.Add(errors.Wrapf(err, "error getting tasks '%s' depends on", t.Id))
+				}
+				if t.IsPartOfSingleHostTaskGroup() {
+					for _, dep := range deps {
+						// reset any already finished tasks in the same task group
+						if dep.TaskGroup == t.TaskGroup && t.TaskGroup != "" && dep.IsFinished() {
+							if err = resetTask(dep.Id, caller, false); err != nil {
+								catcher.Add(errors.Wrapf(err, "error resetting dependency '%s'", dep.Id))
```
ablack12, line 65: catcher.Wrapf

model/task_lifecycle.go
```diff
@@ -31,111 +31,123 @@ type StatusChanges struct {
...
+			}
+			if t.Requester == evergreen.MergeTestRequester {
+				if err = DequeueAndRestartForTask(nil, &t, message.GithubStateError, caller, fmt.Sprintf("deactivated by '%s'", caller)); err != nil {
+					catcher.Add(err)
+				}
 			}
-		} else {
 			tasksToActivate = append(tasksToActivate, originalTasks...)
+		} else {
+			return nil
```
ablack12, line 115: Shouldn't we continue to the next item in the loop?

model/task_lifecycle.go
```diff
@@ -31,111 +31,123 @@ type StatusChanges struct {
...
-		}
-	} else {
-		return nil
 	}
 
-	if t.IsPartOfDisplay() {
-		if err := UpdateDisplayTaskForTask(t); err != nil {
-			return errors.Wrap(err, "problem updating display task")
+	for b := range buildIdsSet {
+		t := buildToTaskMap[b]
```
ablack12, line 145: Do we need both buildIdsSet and buildToTaskMap? Seems like we could just iterate through buildToTaskMap instead

ablack12 (changes requested):

model/task_lifecycle.go
hadjri: I had assumed that the ActivatedBy field would not change for any of the tasks in the list, and since the caller / active parameters do not change it'd be best to return early here (same setup as before this implementation).

graphql/util.go
```diff
@@ -58,40 +58,77 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
-		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task %s not found", taskID).Error())
+	if len(tasks) != len(taskIDs) {
+		foundTaskIds := []string{}
+		for _, ft := range tasks {
+			foundTaskIds = append(foundTaskIds, ft.Id)
+		}
+		missingTaskIds := sliceDifference(taskIDs, foundTaskIds)
+		grip.Error(message.Fields{
+			"message":      "could not find all tasks",
+			"missingTasks": missingTaskIds,
```
ablack12, line 75: nit: we typically do snake case for grip statements so missing_tasks

graphql/util.go
```diff
@@ -58,40 +58,77 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
-	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task %s not found", taskID).Error())
+	if len(tasks) != len(taskIDs) {
+		foundTaskIds := []string{}
+		for _, ft := range tasks {
+			foundTaskIds = append(foundTaskIds, ft.Id)
+		}
+		missingTaskIds := sliceDifference(taskIDs, foundTaskIds)
+		grip.Error(message.Fields{
+			"message":      "could not find all tasks",
```
ablack12, line 74: This message is slightly generic, can we add a line "function":"setManyTasksScheduled" or something? (Or the new function name if we move it)

graphql/util.go
```diff
@@ -58,40 +58,77 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+func setManyTasksScheduled(ctx context.Context, url string, isActive bool, taskIDs ...string) ([]*restModel.APITask, error) {
 	usr := mustHaveUser(ctx)
-	t, err := task.FindOneId(taskID)
+	tasks, err := task.FindAll(db.Query(task.ByIds(taskIDs)))
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
-	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, errors.Errorf("task %s not found", taskID).Error())
+	if len(tasks) != len(taskIDs) {
```
ablack12, line 67: Should we make a FindAllByIds function that does this logging within it rather than duplicating the logic?

graphql/util.go
```diff
@@ -58,40 +58,77 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+		foundTaskIds := []string{}
+		for _, ft := range tasks {
+			foundTaskIds = append(foundTaskIds, ft.Id)
+		}
+		missingTaskIds := sliceDifference(taskIDs, foundTaskIds)
+		grip.Error(message.Fields{
+			"message":      "could not find all tasks",
+			"missingTasks": missingTaskIds,
+		})
+		return nil, ResourceNotFound.Send(ctx, errors.New("tasks not found").Error())
```
ablack12, line 77: Should we return here and still modify the tasks that we do have?

graphql/util.go
```diff
@@ -58,40 +58,77 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
...
+		}
+		apiTasks = append(apiTasks, &apiTask)
 	}
-	err = apiTask.BuildFromService(url)
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, err.Error())
+	return apiTasks, nil
+}
+
+func sliceDifference(a, b []string) (diff []string) {
```
ablack12, line 121: We actually already have utility.StringSliceSymmetricDifference

model/task_lifecycle.go
ablack12: But the "else if" is related specifically to the task ActivatedBy, so it's possible that the first task would skip the else if because it wasn't activated by the right person, but the second task was.

model/task_lifecycle.go
```diff
@@ -31,111 +31,114 @@ type StatusChanges struct {
...
-				"message": "task is missing distro id",
-				"task_id": t.Id,
-			})
+		if t.IsPartOfDisplay() {
+			if err := UpdateDisplayTaskForTask(&t); err != nil {
+				catcher.Wrap(err, "problem updating display task")
+			}
 		}
+	}
+	if catcher.HasErrors() {
```
ablack12, line 114: We've already handled some resetting and etc above, right? So I'm wondering if we should continue to the lower else/if and status updates, and just return the catcher last.

ablack12 (changes requested):

graphql/util.go
hadjri: Sure

graphql/util.go
hadjri: Was looking for something like this but didn't find it. Thanks!

graphql/util.go
hadjri: Sure, I suppose we can remove the return statement here and just let the grip.error remain.

model/task_lifecycle.go
hadjri: Ah, I had thought the tasks in the submitted batch would all have the same ActivatedBy and that wouldn't be a case to consider.  Ok, changing to continue in this case.

model/task_lifecycle.go
hadjri: True since we are already resetting and potentially dequeuing / restarting some tasks in this block it would probably be better if we allow all of our status updates run for the tasks that do get collected successfully instead of prematurely returning an err.

graphql/util.go
hadjri: Actually I might still error if we find no tasks at all because otherwise the output will be '"status": "started"' which isn't accurate.

ablack12 (approved):

PR #5488, opened 2022-03-15
EVG-15085: Fix PatchTriggerAlias type

sophstad:
Partially undo change introduced in #5483

### Description 
- PatchTriggerAlias is used outside of project settings, and these components do not use status or parentAsModule. They should remain nullable.

### Testing 
- Tested manually


khelif96 (approved):

sophstad:
This bug made me realize -- once downstream tasks are fixed, we could add another Spruce task to evergreen that runs `yarn codegen` and then `yarn check-types` to ensure GraphQL schema changes here don't cause a breakage

PR #5490, opened 2022-03-16
EVG-16166: implement container task queue

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16166

### Description 
* Add queue wrapper to get the next container task to be allocated a container.
    * Add query to get tasks that need allocation.
    * Check project ref to ensure the task is allowed to run. I mostly duplicated [the host scheduler's existing checks](https://github.com/evergreen-ci/evergreen/blob/1adeffd04b411b798ec9cb29c8c8b2b68cf8df0b/scheduler/task_finder.go#L68-L147).

### Testing 
Added unit tests.

### DIFF
--- a/model/container_task_queue.go
+++ b/model/container_task_queue.go
@@ -74,7 +74,7 @@ func (q *ContainerTaskQueue) filterByProjectRefSettings(tasks []task.Task) ([]ta
 			// GitHub PR tasks are still allowed to run for disabled hidden
 			// projects.
 			if t.Requester == evergreen.GithubPRRequester && ref.IsHidden() {
-				grip.Info(message.Fields{
+				grip.Debug(message.Fields{
 					"message": "queueing task because GitHub PRs are allowed to run tasks for projects that are both hidden and disabled",
 					"outcome": "not skipping",
 					"task":    t.Id,
@@ -82,9 +82,9 @@ func (q *ContainerTaskQueue) filterByProjectRefSettings(tasks []task.Task) ([]ta
 					"context": "container task queue",
 				})
 			} else {
-				grip.Info(message.Fields{
+				grip.Debug(message.Fields{
 					"message": "skipping task because project is disabled",
-					"outcome": "not skipping",
+					"outcome": "skipping",
 					"task":    t.Id,
 					"project": t.Project,
 					"context": "container task queue",
@@ -94,7 +94,7 @@ func (q *ContainerTaskQueue) filterByProjectRefSettings(tasks []task.Task) ([]ta
 		}
 
 		if ref.IsDispatchingDisabled() {
-			grip.Info(message.Fields{
+			grip.Debug(message.Fields{
 				"message": "skipping task because dispatching is disabled for its project",
 				"outcome": "skipping",
 				"task":    t.Id,
@@ -105,7 +105,7 @@ func (q *ContainerTaskQueue) filterByProjectRefSettings(tasks []task.Task) ([]ta
 		}
 
 		if t.IsPatchRequest() && ref.IsPatchingDisabled() {
-			grip.Info(message.Fields{
+			grip.Debug(message.Fields{
 				"message": "skipping task because patch testing is disabled for its project",
 				"outcome": "skipping",
 				"task":    t.Id,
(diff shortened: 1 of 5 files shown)


model/container_task_queue.go
```diff
@@ -0,0 +1,150 @@
...
+				"context": "container task queue",
+			})
+			continue
+		}
+
+		if !ref.IsEnabled() {
+			// GitHub PR tasks are still allowed to run for disabled hidden
+			// projects.
+			if t.Requester == evergreen.GithubPRRequester && ref.IsHidden() {
+				grip.Info(message.Fields{
```
julianedwards, line 77: Should this be a debug message?

model/container_task_queue.go
```diff
@@ -0,0 +1,150 @@
...
+					"message": "queueing task because GitHub PRs are allowed to run tasks for projects that are both hidden and disabled",
+					"outcome": "not skipping",
+					"task":    t.Id,
+					"project": t.Project,
+					"context": "container task queue",
+				})
+			} else {
+				grip.Info(message.Fields{
+					"message": "skipping task because project is disabled",
+					"outcome": "not skipping",
```
julianedwards, line 87: `"outcome": "skipping",`

model/container_task_queue.go
```diff
@@ -0,0 +1,150 @@
...
+			return nil, errors.Wrap(err, "initial population of container task queue")
+		}
+	}
+	if q.position >= len(q.queue) {
+		return nil, nil
+	}
+
+	next := q.queue[q.position]
+
+	q.position++
```
julianedwards, line 32: Do we not want to pop ever?

model/task/db.go
```diff
@@ -766,6 +766,32 @@ func schedulableHostTasksQuery() bson.M {
...
+				// are all met. All dependencies are met if they're all finished
+				// running and are still attainable (i.e. the dependency's
+				// required status matched the task's actual ending status).
+				bsonutil.GetDottedKeyName(DependsOnKey, DependencyFinishedKey):     true,
+				bsonutil.GetDottedKeyName(DependsOnKey, DependencyUnattainableKey): bson.M{"$ne": true},
+			},
+			{OverrideDependenciesKey: true},
+		},
+	}
+	return FindAll(db.Query(q).Sort([]string{ActivatedTimeKey}))
```
julianedwards, line 792: Since this queue is in-memory, how do we ensure that multiple app servers do not pick up the same tasks? Should this be a transaction to ensure that we update the `StatusKey`?

julianedwards (changes requested):

model/container_task_queue.go
Kimchelly: For now, there's no need to have any operation other than pop. This is only used to cron populate pod allocator jobs.

model/container_task_queue.go
Kimchelly: Probably yeah. I changed it and all the other logs to debug since they're mostly for tracing purposes.

model/container_task_queue.go
Kimchelly: Fixed.

model/task/db.go
Kimchelly: Amboy scopes are going to resolve any issues with duplicate pod allocator jobs being created from these tasks.

We don't need to update the status key since this is just the query. The status is only updated from unallocated to allocated by the individual pod allocator jobs. This task queue will just feed into the cron population of pod allocator jobs (in a later PR).

model/container_task_queue.go
julianedwards: I meant, is it worth it to delete the element from the slice?

model/task/db.go
julianedwards: Won't we be doing a lot of duplicate/extra work on the app servers? Or are we expecting these allocation jobs to happen so fast that two or more app servers picking up duplicate container tasks that need to be allocated will be rare?

model/container_task_queue.go
Kimchelly: I don't think so. There's no cost/benefit to deleting elements for now since it's in-memory. There might have to be some deletion/shuffling anyways once task groups are considered.

julianedwards (approved):

PR #5492, opened 2022-03-16
EVG-16368 fix project creation for new project page

ablack12:
[EVG-16368](https://jira.mongodb.org/browse/EVG-16368)

### Description 
Need to add blank project variables in order for this button to be useable. 

### Testing 
Tested that a created project can be visited on Spruce, and tested that the copy project route still works.

### DIFF
--- a/rest/data/project.go
+++ b/rest/data/project.go
@@ -64,9 +64,28 @@ func (pc *DBProjectConnector) CreateProject(projectRef *model.ProjectRef, u *use
 	if err != nil {
 		return gimlet.ErrorResponse{
 			StatusCode: http.StatusInternalServerError,
-			Message:    errors.Wrapf(err, "project with id '%s' was not inserted", projectRef.Id).Error(),
+			Message:    errors.Wrapf(err, "inserting project '%s'", projectRef.Identifier).Error(),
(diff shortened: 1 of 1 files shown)


rest/data/project.go
```diff
@@ -64,9 +64,28 @@ func (pc *DBProjectConnector) CreateProject(projectRef *model.ProjectRef, u *use
...
-			Message:    errors.Wrapf(err, "project with id '%s' was not inserted", projectRef.Id).Error(),
+			Message:    errors.Wrapf(err, "inserting project '%s'", projectRef.Identifier).Error(),
 		}
 	}
+
+	newProjectVars := model.ProjectVars{
+		Id: projectRef.Id,
+	}
+
+	err = newProjectVars.Insert()
```
hadjri, line 75: If we handle the UI issue from the resolver side wouldn't we be able to avoid these additional DB entries?

ablack12: This is the expected state from the backend's perspective, it's not actually a UI issue.

hadjri: Ok, SGTM

hadjri (approved):

PR #5497, opened 2022-03-17
EVG-16423 Define container models and integrate into parser project

hadjri:
[EVG-16423](https://jira.mongodb.org/browse/EVG-16423)

### Description 
Added containers struct into parser projects detailing the data model detailed in the design doc for this epic.
### Testing 
Updated existing unit tests and created new test asserting correctness of container parsing.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -1085,23 +1085,7 @@ const (
 	LogViewerLobster LogViewer = "lobster"
 )
 
-type ContainerSize string
-
-const (
-	SmallContainer  = "small"
-	MediumContainer = "medium"
-	LargeContainer  = "large"
-)
-
-func (c ContainerSize) Validate() error {
-	switch c {
-	case SmallContainer, MediumContainer, LargeContainer:
-		return nil
-	default:
-		return errors.Errorf("unrecognized container size '%s'", c)
-	}
-}
-
+// ContainerOS denotes the operating system of a running container. Only linux and windows are supported.
 type ContainerOS string
 
 const (
@@ -1118,14 +1102,15 @@ func (c ContainerOS) Validate() error {
 	}
 }
 
-type CpuArchitecture string
+// CPUArchitecture represents the architecture property of a container required to provide to Amazon ECS
+type CPUArchitecture string
 
 const (
 	ArchARM64 = "arm64"
 	ArchAMD64 = "x86_64"
 )
 
-func (c CpuArchitecture) Validate() error {
+func (c CPUArchitecture) Validate() error {
 	switch c {
 	case ArchARM64, ArchAMD64:
 		return nil
@@ -1134,6 +1119,7 @@ func (c CpuArchitecture) Validate() error {
 	}
 }
 
+// WindowsVersion specifies the Windows version of a container if running on Windows device (required if os is windows)
 type WindowsVersion string
 
 const (
--- a/globals.go
+++ b/globals.go
@@ -1085,14 +1085,15 @@ const (
 	LogViewerLobster LogViewer = "lobster"
 )
 
-// ContainerOS denotes the operating system of a running container. Only linux and windows are supported.
+// ContainerOS denotes the operating system of a running container.
 type ContainerOS string
 
 const (
-	LinuxOS   = "linux"
-	WindowsOS = "windows"
+	LinuxOS   ContainerOS = "linux"
+	WindowsOS ContainerOS = "windows"
 )
 
+// Validate checks that the container OS is recognized.
 func (c ContainerOS) Validate() error {
 	switch c {
 	case LinuxOS, WindowsOS:
@@ -1102,14 +1103,15 @@ func (c ContainerOS) Validate() error {
 	}
 }
 
-// CPUArchitecture represents the architecture property of a container required to provide to Amazon ECS
+// CPUArchitecture represents the architecture necessary to run the container.
 type CPUArchitecture string
 
 const (
-	ArchARM64 = "arm64"
-	ArchAMD64 = "x86_64"
+	ArchARM64 CPUArchitecture = "arm64"
+	ArchAMD64 CPUArchitecture = "x86_64"
 )
 
+// Validate checks that the container CPU architecture is recognized.
 func (c CPUArchitecture) Validate() error {
 	switch c {
 	case ArchARM64, ArchAMD64:
@@ -1119,15 +1121,16 @@ func (c CPUArchitecture) Validate() error {
 	}
 }
 
-// WindowsVersion specifies the Windows version of a container if running on Windows device (required if os is windows)
+// WindowsVersion specifies the compatability version of Windows that is required for the container to run.
 type WindowsVersion string
 
 const (
-	Windows2022 = "2022"
-	Windows2019 = "2019"
-	Windows2016 = "2016"
+	Windows2022 WindowsVersion = "2022"
+	Windows2019 WindowsVersion = "2019"
+	Windows2016 WindowsVersion = "2016"
 )
 
+// Validate checks that the container Windows version is recognized.
 func (w WindowsVersion) Validate() error {
 	switch w {
 	case Windows2022, Windows2019, Windows2016:
--- a/globals.go
+++ b/globals.go
@@ -1084,3 +1084,69 @@ const (
 	LogViewerHTML    LogViewer = "html"
 	LogViewerLobster LogViewer = "lobster"
 )
+
+type ContainerSize string
+
+const (
+	SmallContainer  = "small"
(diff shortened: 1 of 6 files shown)


Kimchelly (changes requested):

model/project.go
```diff
@@ -294,6 +295,27 @@ type ParameterInfo struct {
 	Description     string `yaml:"description" bson:"description"`
 }
 
+type Container struct {
+	Name        string                  `yaml:"name,omitempty" bson:"name"`
+	DisplayName string                  `yaml:"display_name,omitempty" bson:"display_name"`
```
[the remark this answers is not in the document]
hadjri: I had only thought this might be needed for readability in the UI but I'm fine to remove this.

model/project_parser_merge_functions.go
```diff
@@ -68,6 +68,19 @@ func (pp *ParserProject) mergeUnorderedUnique(toMerge *ParserProject) error {
...
 	}
 
+	containerExist := map[string]bool{}
+	for _, container := range pp.Containers {
+		containerExist[container.Name] = true
+	}
+	for _, container := range toMerge.Containers {
+		if _, ok := containerExist[container.Name]; ok {
+			catcher.Errorf("container '%s' has been declared already", container.Name)
+		} else {
```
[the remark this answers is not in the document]
hadjri: Thanks!

globals.go
```diff
@@ -1084,3 +1084,69 @@ const (
 	LogViewerHTML    LogViewer = "html"
 	LogViewerLobster LogViewer = "lobster"
 )
+
+type ContainerSize string
+
+const (
+	SmallContainer  = "small"
+	MediumContainer = "medium"
+	LargeContainer  = "large"
```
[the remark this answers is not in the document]
hadjri: SGTM, I suppose they could be any alias name. Going to change Size field to just a string.

model/project.go
```diff
@@ -294,6 +295,27 @@ type ParameterInfo struct {
 	Description     string `yaml:"description" bson:"description"`
 }
 
+type Container struct {
+	Name        string                  `yaml:"name,omitempty" bson:"name"`
```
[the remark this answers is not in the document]
hadjri: I'd just noticed all fields in project_parser.go or project.go had omitempty on them even if they were seemingly required fields, so I followed that pattern in case it was required. But I don't think it would cause a problem and is better to specify what fields are required and which aren't, so I'll update

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+
+// ContainerOS denotes the operating system of a running container. Only linux and windows are supported.
+type ContainerOS string
+
+const (
+	LinuxOS   = "linux"
+	WindowsOS = "windows"
+)
+
+func (c ContainerOS) Validate() error {
```
Kimchelly, line 1096: Documentation for exported methods.

model/project.go
```diff
@@ -294,6 +295,26 @@ type ParameterInfo struct {
 	Description     string `yaml:"description" bson:"description"`
 }
 
+type Container struct {
```
Kimchelly, line 298: Documentation for exported type.

model/project.go
```diff
@@ -294,6 +295,26 @@ type ParameterInfo struct {
...
+	Name       string             `yaml:"name" bson:"name"`
+	WorkingDir string             `yaml:"working_dir,omitempty" bson:"working_dir"`
+	Image      string             `yaml:"image" bson:"image"`
+	Size       string             `yaml:"size,omitempty" bson:"size"`
+	Resources  ContainerResources `yaml:"resources,omitempty" bson:"resources"`
+	System     ContainerSystem    `yaml:"system,omitempty" bson:"system"`
+}
+
+type ContainerSystem struct {
```
Kimchelly, line 307: Documentation for exported type.

model/project.go
```diff
@@ -294,6 +295,26 @@ type ParameterInfo struct {
...
+	CPUArchitecture evergreen.CPUArchitecture `yaml:"cpu_architecture,omitempty" bson:"cpu_architecture"`
+	OperatingSystem evergreen.ContainerOS     `yaml:"operating_system,omitempty" bson:"operating_system"`
+	WindowsVersion  evergreen.WindowsVersion  `yaml:"windows_version,omitempty" bson:"windows_version"`
+}
+
+type ContainerResources struct {
```
Kimchelly, line 313: Documentation for exported type.

model/project.go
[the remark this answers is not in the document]
Kimchelly: The way distros show up in the UI is always by their `_id` (which is the equivalent of the container name). I suppose there could be some desire for a display name eventually, but I think unless there's some specific need for that, it should be okay to just use the name for now.

model/project_parser_merge_functions.go
```diff
@@ -23,10 +23,10 @@ func (pp *ParserProject) mergeUnorderedUnique(toMerge *ParserProject) error {
 	for _, task := range toMerge.Tasks {
 		if _, ok := taskNameExist[task.Name]; ok {
 			catcher.Errorf("task '%s' has been declared already", task.Name)
-		} else {
-			pp.Tasks = append(pp.Tasks, task)
-			taskNameExist[task.Name] = true
+			continue
```
Kimchelly, line 26: Thanks for change the other if-else uses to be consistent as well! 💯 (no change)

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
 	LogViewerHTML    LogViewer = "html"
 	LogViewerLobster LogViewer = "lobster"
 )
+
+// ContainerOS denotes the operating system of a running container. Only linux and windows are supported.
```
Kimchelly, line 1088: I don't think we need to say that it only supports Windows and Linux since it should be obvious from the two consts.

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+type ContainerOS string
+
+const (
+	LinuxOS   = "linux"
```
Kimchelly, line 1092: In order for these to be of type `ContainerOS`, you have to include the type declaration or else these will be counted as just plain strings:
```
const (
LinuxOS ContainerOS = "linux"
WindowsOS ContainerOS = "windows"
)
```

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+	default:
+		return errors.Errorf("unrecognized container OS '%s'", c)
+	}
+}
+
+// CPUArchitecture represents the architecture property of a container required to provide to Amazon ECS
+type CPUArchitecture string
+
+const (
+	ArchARM64 = "arm64"
```
Kimchelly, line 1109: Same thing about `ContainerOS` type applies here for `CPUArchitecture`.

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+	default:
+		return errors.Errorf("unrecognized CPU architecture '%s'", c)
+	}
+}
+
+// WindowsVersion specifies the Windows version of a container if running on Windows device (required if os is windows)
+type WindowsVersion string
+
+const (
+	Windows2022 = "2022"
```
Kimchelly, line 1126: Same thing as above about type declaration.

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+func (c ContainerOS) Validate() error {
+	switch c {
+	case LinuxOS, WindowsOS:
+		return nil
+	default:
+		return errors.Errorf("unrecognized container OS '%s'", c)
+	}
+}
+
+// CPUArchitecture represents the architecture property of a container required to provide to Amazon ECS
```
Kimchelly, line 1105: I would avoid saying "Amazon ECS" here since that assumes a particular container implementation. Instead, I would say something more generic like "CPUArchitecture is the architecture necessary to run this container".

globals.go
```diff
@@ -1084,3 +1084,55 @@ const (
...
+func (c CPUArchitecture) Validate() error {
+	switch c {
+	case ArchARM64, ArchAMD64:
+		return nil
+	default:
+		return errors.Errorf("unrecognized CPU architecture '%s'", c)
+	}
+}
+
+// WindowsVersion specifies the Windows version of a container if running on Windows device (required if os is windows)
```
Kimchelly, line 1122: Period at the end of comment/documentation sentences. I don't think the comment in parentheses is necessary since it's already stated that it's running on a Windows machine. I would phrase the comment as something like: "WindowsVersion specifies the compatibility version of Windows required for the container to run."

Kimchelly (changes requested):

Kimchelly (approved):
LGTM mod the one comment about CPU units to vCPU conversion.

PR #5498, opened 2022-03-17
EVG-16494: reduce logging level for agents on already-terminated hosts

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16494

### Description 
There's two different issues that I've found as part of the investigation:
* Some agents stay alive for much longer than expected (on the order of days), well after the cloud provider has said the host is shutting down/terminated. This seems like an issue on the cloud provider side and not something we can do anything about, so it's not a bug on our end. At least we aren't being billed for these instances since they're in the terminated state.
* A single intent host somehow started two cloud hosts. We got one instance ID back from the API request (`RunInstances`), which we recorded as the cloud host created from the intent host, but a second doppelganger cloud host also started an agent. Since they share the same intent host ID, they're considered the "same" host in Evergreen despite being two distinct cloud hosts. This seems like a bug, but fixing it would require a better understanding of how it happened. Unfortunately, Splunk logs also seem to be missing some important messages about the problematic intent host (e.g. Splunk logs are missing the expected successful call to `RunInstances`), which precludes any further investigation.

I reduced the log level, which should remove both issues from the error rate alert. In the first case, there's not much we can actually do about it, so we might as well not include that in the error rate. In the second case, it seems like the only thing to do is wait for the bug to recur and re-investigate when it happens.

* Reduce log level noise.
* Slightly refactor parameters passed to `prepareHostForAgentExit`.

### Testing 
N/A

### DIFF
--- a/service/api_task.go
+++ b/service/api_task.go
@@ -506,7 +506,7 @@ func (as *APIServer) prepareHostForAgentExit(ctx context.Context, params agentEx
 		// fully cleaned up. If this message logs, that means there is a
 		// mismatch between either:
 		// 1. What Evergreen believes is this host's state (terminated) and the
-		//    cloud instance's actual state(not terminated). This is a bug.
+		//    cloud instance's actual state (not terminated). This is a bug.
 		// 2. The cloud instance's status and the actual liveliness of the host.
 		//    There are some cases where the agent is still checking in for a
 		//    long time after the cloud provider says the instance is already
--- a/service/api_task.go
+++ b/service/api_task.go
@@ -260,7 +260,10 @@ func (as *APIServer) EndTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if checkHostHealth(currentHost) {
-		if _, err := as.prepareHostForAgentExit(r.Context(), currentHost, r.RemoteAddr); err != nil {
+		if _, err := as.prepareHostForAgentExit(r.Context(), agentExitParams{
+			host:       currentHost,
+			remoteAddr: r.RemoteAddr,
+		}); err != nil {
 			grip.Error(message.WrapError(err, message.Fields{
 				"message":       "could not prepare host for agent to exit",
 				"host_id":       currentHost.Id,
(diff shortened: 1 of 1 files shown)


service/api_task.go
```diff
@@ -470,41 +473,56 @@ func (as *APIServer) transitionIntentHostToDecommissioned(ctx context.Context, h
...
-		// has clearly not been terminated in the cloud.
-		if time.Now().Sub(h.TerminationTime) > 10*time.Minute {
-			grip.Error(message.Fields{
+		// The host should already be terminated but somehow the agent is still
+		// alive in the cloud. If the host was terminated very recently, it may
+		// simply need time for the cloud instance to actually be terminated and
+		// fully cleaned up. If this message logs, that means there is a
+		// mismatch between either:
+		// 1. What Evergreen believes is this host's state (terminated) and the
+		//    cloud instance's actual state(not terminated). This is a bug.
```
julianedwards, line 509: Nit: add space between `state(not terminated)`.

service/api_task.go
```diff
@@ -470,41 +473,56 @@ func (as *APIServer) transitionIntentHostToDecommissioned(ctx context.Context, h
...
-		// mismatch where Evergreen has marked the host terminated but the host
-		// has clearly not been terminated in the cloud.
-		if time.Now().Sub(h.TerminationTime) > 10*time.Minute {
-			grip.Error(message.Fields{
+		// The host should already be terminated but somehow the agent is still
+		// alive in the cloud. If the host was terminated very recently, it may
+		// simply need time for the cloud instance to actually be terminated and
+		// fully cleaned up. If this message logs, that means there is a
+		// mismatch between either:
+		// 1. What Evergreen believes is this host's state (terminated) and the
```
julianedwards, line 508: It seems like there is a third case, which many not be a bug, and is described above as `If the host was terminated very recently, it may simply need time for the cloud instance to actually be terminated and fully cleaned up.` Should that be added to this list?

julianedwards (changes requested):

service/api_task.go
Kimchelly: Fixed.

service/api_task.go
Kimchelly: The two cases I mentioned are cases when it would be logged to Splunk. The "terminated recently" case would not cause it to log because we give the host the 10-minute grace period to actually go away.

julianedwards: Ah I missed that, thanks!

julianedwards (approved):

ybrill (approved):
would we consider opening a support case with AWS on the case where some agents stay alive for much longer than expected (to make certain we aren't being charged)?

Kimchelly:
I'm much more inclined to investigate our own code first before we ask AWS because of how battle-tested EC2 is as opposed to our own cloud provider code.

PR #5499, opened 2022-03-18
EVG-13488: Return recently terminated hosts in MyHosts resolver

minnakt:
[EVG-13488](https://jira.mongodb.org/browse/EVG-13488)

### Description 
This PR makes it so that recently terminated hosts are returned alongside non-terminated hosts in the `MyHosts` resolver. "Recently terminated" in this case means any hosts that were terminated in the last 5 minutes.

### Testing 
- Modified data & results for `graphql/tests/myHosts`
- Checked on staging that if I terminated a host, it would be returned in the `GET_MY_HOSTS` query for the next 5 minutes


graphql/tests/myHosts/data.json
```diff
@@ -67,6 +67,73 @@
...
+        "$date": "2020-07-20T14:48:05.256Z"
+      },
+      "start_time": {
+        "$date": "2020-07-20T14:48:27Z"
+      },
+      "agent_start_time": {
+        "$date": "2020-07-20T14:49:20.449Z"
+      },
+      "termination_time": {
+        "$date": "5000-01-01T00:00:00Z"
```
minnakt, line 102: Usually when we unit test time-dependent functions we can just set the field of the struct to `time.Now()`. But since the data for `graphql` tests comes from a JSON, I couldn't find a way to do that.

To get around this I set this `termination_time` to be in the year 5000, which means that technically it has been terminated in the last 5 minutes, so this host should be returned. I can remove this if you think it's okay to assume that the query works without actually testing it

graphql/resolvers.go
```diff
@@ -1070,14 +1070,30 @@ func (r *queryResolver) MyVolumes(ctx context.Context) ([]*restModel.APIVolume,
...
+		err = apiHost.BuildFromService(host)
+		if err != nil {
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error building APIHost from service: %s", err.Error()))
+		}
+		apiHosts = append(apiHosts, &apiHost)
+	}
+	for _, host := range recentlyTerminatedHosts {
 		apiHost := restModel.APIHost{}
 		err = apiHost.BuildFromService(host)
 		if err != nil {
```
khelif96, line 1099: You can combine runningHosts and recentlyTerminatedHosts before the for loop and only have 1 for loop instead of 2 which would reduce allot of duplicated code.

graphql/tests/myHosts/data.json
khelif96: But what happens when we get to the year 5000 our tests will break! JK

But you did reveal a flaw in the way we seed our data that makes it difficult to test features that rely on time. Could you open a ticket so we can look into this and maybe come up with a way to get around this.

khelif96 (approved):
Great job just had 1 suggestion

graphql/tests/myHosts/data.json
minnakt: Opened [EVG-16565](https://jira.mongodb.org/browse/EVG-16565) 🤓

PR #5500, opened 2022-03-18
EVG-16402: add method to create collection

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16402

### Description 
Replace usages of `CreateCollection` in tests to use the Anser helper.

### Testing 
Replaced its usages in tests and added an idempotency test.

### DIFF
--- a/db/db_utils.go
+++ b/db/db_utils.go
@@ -90,6 +90,32 @@ func InsertManyUnordered(c string, items ...interface{}) error {
 	return errors.WithStack(err)
 }
 
+// CreateCollections ensures that all the given collections are created,
+// returning an error immediately if creating any one of them fails.
+func CreateCollections(collections ...string) error {
+	session, db, err := GetGlobalSessionFactory().GetSession()
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	const namespaceExistsErrCode = 48
+	for _, collection := range collections {
+		_, err := db.CreateCollection(collection)
+		if err == nil {
+			continue
+		}
+		// If the collection already exists, this does not count as an error.
+		if mongoErr, ok := errors.Cause(err).(mongo.CommandError); ok && mongoErr.HasErrorCode(namespaceExistsErrCode) {
+			continue
+		}
+		if err != nil {
+			return errors.Wrapf(err, "creating collection '%s'", collection)
+		}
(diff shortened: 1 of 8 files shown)


graphql/directive_test.go
```diff
@@ -23,8 +24,7 @@ func setupPermissions(t *testing.T) {
 	require.NoError(t, env.DB().Drop(ctx))
 
 	// Create scope and role collection to avoid RoleManager from trying to create them in a collection https://jira.mongodb.org/browse/EVG-15499
```
ybrill, line 26: This (preexisting) comment threw me for a loop. I think he meant "from trying to create them in a transaction"?

db/db_utils.go
```diff
@@ -90,6 +90,32 @@ func InsertManyUnordered(c string, items ...interface{}) error {
...
+	defer session.Close()
+
+	const namespaceExistsErrCode = 48
+	for _, collection := range collections {
+		_, err := db.CreateCollection(collection)
+		if err == nil {
+			continue
+		}
+		// If the collection already exists, this does not count as an error.
+		if mongoErr, ok := errors.Cause(err).(mongo.CommandError); ok && mongoErr.HasErrorCode(namespaceExistsErrCode) {
```
ybrill, line 109: What are your thoughts on matching against the error code vs. [listing the existing collection names](https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo#Database.ListCollectionNames) and only creating the ones that don't already exist?

Kimchelly: I prefer not to do that since this is a stable error code that proves in an atomic way that at the moment of the creation attempt, it already existed. Also, list + create adds an extra DB operation and would require more Anser changes.

graphql/directive_test.go
Kimchelly: Yeah, the ticket says it's because it's trying to insert docs into nonexistent collections in a transaction. I fixed the comments.

db/db_utils.go
ybrill: sgtm

ybrill (approved):

graphql/directive_test.go
ybrill: ty!

PR #5501, opened 2022-03-18
EVG-16052 Add project command for assume role

bynn:
[EVG-16052](https://jira.mongodb.org/browse/EVG-16052)

### Description 
added project command that calls assume role given ARN and returns credentials as expansions

### Testing 
very basic unit test

image

image

https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu1604_bynntask_patch_33f13782b26be2bd33a48d378fb2293db7316034_6234f34bb237367d9ac27d7c_22_03_18_21_02_10/logs?execution=0

testing secrets manager 
image

image

### DIFF
--- a/agent/command/assume_ec2_role.go
+++ b/agent/command/assume_ec2_role.go
@@ -18,13 +18,19 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	AwsAccessKeyId     = "AWS_ACCESS_KEY_ID"
+	AwsSecretAccessKey = "AWS_SECRET_ACCESS_KEY"
+	AwsSessionToken    = "AWS_SESSION_TOKEN"
+	AwsRoleExpiration  = "AWS_ROLE_EXPIRATION"
+)
+
 type ec2AssumeRole struct {
 	// The Amazon Resource Name (ARN) of the role to assume.
 	// Required.
 	RoleARN string `mapstructure:"role_arn" plugin:"expand"`
 
 	// A unique identifier that might be required when you assume a role in another account.
-	// This variable will only be accessible by admins and mainline.
 	ExternalId string `mapstructure:"external_id" plugin:"expand"`
 
 	// An IAM policy in JSON format that you want to use as an inline session policy.
@@ -53,6 +59,10 @@ func (r *ec2AssumeRole) validate() error {
 
 	if r.RoleARN == "" {
 		catcher.New("must specify role ARN")
+	}
+
+	if r.DurationSeconds < 0 {
+		catcher.New("cannot specify a non-positive duration")
 	}
 
 	return catcher.Resolve()
@@ -85,11 +95,11 @@ func (r *ec2AssumeRole) Execute(ctx context.Context,
 		SecretAccessKey: secret,
 	})
 
-	session := session.Must(session.NewSession(&aws.Config{
+	session1 := session.Must(session.NewSession(&aws.Config{
 		Credentials: defaultCreds,
 	}))
 
-	creds := stscreds.NewCredentials(session, r.RoleARN, func(arp *stscreds.AssumeRoleProvider) {
+	creds := stscreds.NewCredentials(session1, r.RoleARN, func(arp *stscreds.AssumeRoleProvider) {
 		arp.RoleSessionName = fmt.Sprintf("evergreen_%s_%d", conf.Task.DisplayName, conf.Task.Execution)
 		if r.ExternalId != "" {
 			arp.ExternalID = utility.ToStringPtr(r.ExternalId)
@@ -112,9 +122,9 @@ func (r *ec2AssumeRole) Execute(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
-	conf.Expansions.Put("AWS_ACCESS_KEY_ID", credValues.AccessKeyID)
-	conf.Expansions.Put("AWS_SECRET_ACCESS_KEY", credValues.SecretAccessKey)
-	conf.Expansions.Put("AWS_SESSION_TOKEN", credValues.SessionToken)
-	conf.Expansions.Put("AWS_ROLE_EXPIRATION", expTime.String())
+	conf.Expansions.Put(AwsAccessKeyId, credValues.AccessKeyID)
+	conf.Expansions.Put(AwsSecretAccessKey, credValues.SecretAccessKey)
+	conf.Expansions.Put(AwsSessionToken, credValues.SessionToken)
+	conf.Expansions.Put(AwsRoleExpiration, expTime.String())
 	return nil
 }
--- a/agent/command/assume_ec2_role.go
+++ b/agent/command/assume_ec2_role.go
@@ -2,7 +2,6 @@ package command
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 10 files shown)


agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+)
+
+type ec2AssumeRole struct {
+	// The Amazon Resource Name (ARN) of the role to assume.
+	// Required.
+	RoleARN string `mapstructure:"role_arn" plugin:"expand"`
+
+	// A unique identifier that might be required when you assume a role in another account.
```
malikchaya2, line 26: I appreciate these detailed comments :)

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+func (r *ec2AssumeRole) ParseParams(params map[string]interface{}) error {
+	if err := mapstructure.Decode(params, r); err != nil {
+		return errors.Wrapf(err, "error parsing '%s' params", r.Name())
+	}
+
+	return r.validate()
+}
+
+func (r *ec2AssumeRole) validate() error {
+	catcher := grip.NewSimpleCatcher()
```
malikchaya2, line 52: It seems like we can just return the error right away, I don't know if we need this catcher.

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+	return catcher.Resolve()
+}
+
+func (r *ec2AssumeRole) Execute(ctx context.Context,
+	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
+	if err := util.ExpandValues(r, conf.Expansions); err != nil {
+		return errors.WithStack(err)
+	}
+	// Re-validate the command here, in case an expansion is not defined.
+	if err := r.validate(); err != nil {
```
malikchaya2, line 67: Is this necessary if parse already does it?

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+	if len(conf.EC2Keys) == 0 {
+		return errors.New("no EC2 keys in config")
+	}
+
+	key := conf.EC2Keys[0].Key
+	secret := conf.EC2Keys[0].Secret
+
+	// Error if key or secret are blank
+	if key == "" || secret == "" {
+		return errors.New("AWS ID and Secret must not be blank")
```
malikchaya2, line 80: Is this error actionable for users? If it is blank, is there something they can do about it? (their default thought would probably be to add it to the command itself.)

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	expTime, err := creds.ExpiresAt()
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	conf.Expansions.Put("AWS_ACCESS_KEY_ID", credValues.AccessKeyID)
```
malikchaya2, line 115: It still feels weird to me that these are put in easily printable expansions. Are we fully comfortable with them being temporary and okay being exposed? Are they not exploitable in any way? 
cc: @ablack12

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,120 @@
...
+		return errors.New("AWS ID and Secret must not be blank")
+	}
+
+	defaultCreds := credentials.NewStaticCredentialsFromCreds(credentials.Value{
+		AccessKeyID:     key,
+		SecretAccessKey: secret,
+	})
+
+	session := session.Must(session.NewSession(&aws.Config{
+		Credentials: defaultCreds,
```
malikchaya2, line 89: Is there any reason why we wouldn't let users specify their aws key and secret in the command and use it from there?

agent/command/assume_ec2_role.go
ablack12: This was discussed briefly in the design; I think as long as we're very clear in the documentation about it, it's not a huge concern. Maybe it would make sense though to explicitly remove these from expansions.write (similarly to [here](https://github.com/evergreen-ci/evergreen/blob/758c4464174ac9413e31fbe071e7740a72ad2e54/agent/command/expansion_write.go#L40)).

ablack12: @bynn feel free to either do in this ticket or create a new one; we shouldn't document this command until the work is done though.

bynn (commented):
added screenshot for e2e testing 
using the credential to access secrets manager

agent/command/assume_ec2_role.go
malikchaya2: ah, good point!

agent/command/assume_ec2_role.go
malikchaya2: In that case, should we maybe change the language of this error? "AWS ID and Secret must not be blank" would probably make me think "oops, I did something wrong". Should we maybe change it to something like "couldn't populate AWS credentials" or something like that?

bynn: done

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (r *ec2AssumeRole) validate() error {
+	catcher := grip.NewSimpleCatcher()
+
+	if r.RoleARN == "" {
+		catcher.New("must specify role ARN")
+	}
+
+	if r.DurationSeconds < 0 {
```
ablack12, line 64: I think 0 would also be an invalid duration

agent/command/assume_ec2_role.go
```diff
@@ -0,0 +1,130 @@
...
+		AccessKeyID:     key,
+		SecretAccessKey: secret,
+	})
+
+	session1 := session.Must(session.NewSession(&aws.Config{
+		Credentials: defaultCreds,
+	}))
+
+	creds := stscreds.NewCredentials(session1, r.RoleARN, func(arp *stscreds.AssumeRoleProvider) {
+		arp.RoleSessionName = fmt.Sprintf("evergreen_%s_%d", conf.Task.DisplayName, conf.Task.Execution)
```
ablack12, line 103: Will we be able to reuse this role session name or should it include a truncated timestamp or something? (Not sure what it's used for)

agent/task.go
```diff
@@ -382,6 +382,11 @@ func (a *Agent) makeTaskConfig(ctx context.Context, tc *taskContext) (*internal.
 		return nil, err
 	}
 	taskConfig.Redacted = tc.expVars.PrivateVars
+
+	grip.Info("Populating AWS credentials.")
```
ablack12, line 386: leftover debugging?

agent/command/expansion_write.go
```diff
@@ -7,11 +7,21 @@ import (
...
+	"github.com/evergreen-ci/utility"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v3"
 )
 
+var (
+	expansionsToRedact = []string{
+		evergreen.GlobalGitHubTokenExpansion,
+		AwsAccessKeyId,
```
ablack12, line 19: nit: can we call these AWSAccessKeyId and etc? It's convention to capitalize all the letters in an acronym.

ablack12 (approved):
Some comments but I don't need to review again 🚀

PR #5503, opened 2022-03-21
EVG-6963 return a client error instead of a database duplicate key error

malikchaya2:
[EVG-6963 ](https://jira.mongodb.org/browse/EVG-6963)

### Description 
When there is a race, users can see "duplicate key error collection: amboy.evg.service.jobs". This checks if it's a duplicate error and if it is, returns a more descriptive error. 

### Testing 
I was not able to duplicate this error.

### DIFF
--- a/rest/route/project.go
+++ b/rest/route/project.go
@@ -9,7 +9,6 @@ import (
 	"strconv"
 
 	"github.com/evergreen-ci/evergreen"
-	"github.com/evergreen-ci/evergreen/db"
 	dbModel "github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/artifact"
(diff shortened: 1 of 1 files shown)


rest/route/project.go
```diff
@@ -570,6 +571,9 @@ func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
 
 		queue := evergreen.GetEnvironment().RemoteQueue()
 		if err = queue.Put(ctx, j); err != nil {
+			if db.IsDuplicateKey(err) {
```
Kimchelly, line 574: Is it necessary to return this error, or should it just continue on duplicate key error? If the remote queue already has a pending job to run the repotracker on this revision, it seems like the route could just treat it a success since the work will be done eventually.

malikchaya2: Thanks! updated.

PR #5505, opened 2022-03-21
EVG-16452 ensure private variables aren't overridden

ablack12:
[EVG-16452](https://jira.mongodb.org/browse/EVG-16452)

### Description 
Wasn't considering before that the UI would pass values in as PrivateVarsList. Also added a clause to ensure that we don't overwrite the value in the case that the value is purposefully deleted / re-added.

### Testing 
Tried to break manually and also modified unit test (didn't work without my changes).

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -174,8 +174,8 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 	case model.ProjectPageVariablesSection:
 		for key, value := range before.Vars.Vars {
 			// Private variables are redacted in the UI, so re-set to the real value
-			// before updating (assuming the value isn't deleted).
-			if before.Vars.PrivateVars[key] && changes.Vars.PrivateVars[key] {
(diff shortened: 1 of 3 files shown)


sophstad (approved):
Tested from the UI and it works like a charm -- thanks Annie!

rest/data/project_settings.go
```diff
@@ -219,7 +219,6 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 		}
 		catcher.Wrapf(DeleteSubscriptions(projectId, toDelete), "Database error deleting subscriptions")
 	}
-	fmt.Println("GETTING TO SAVE FOR SECTION")
```
malikchaya2, line 222: did you mean to leave this here?

malikchaya2 (changes requested):

rest/data/project_settings.go
ablack12: It was definitely an accident that it was here to begin with haha

malikchaya2 (approved):

PR #5506, opened 2022-03-21
EVG-16159 add Fleet options

ybrill:
[EVG-16159](https://jira.mongodb.org/browse/EVG-16159)

### Description 
Add two options to distros using Fleet: 
1. use the capacity-optimized [allocation strategy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html). I'd like to experiment with this option for distros that are getting interrupted a bunch to see if it would be a cheaper alternative to going to on-demand. For example, I spawned three ubuntu1604-small hosts 1) on-demand 2) spot with the regular (lowest-cost) allocation strategy and 3) spot with capacity-optimized. The on-demand host cost $.17/hr, the regular spot cost $0.073/hr and the capacity-optimized was $0.0816/hr. It's not a huge spread with such a small instance (c5.xlarge), but with larger instance types it could net a bigger difference (like a bunch of the *-large distros are m4.4xlarge where it's less than half).
2. add an on-demand option for Fleet. I wanted to experiment with this anyway ([EVG-16306](https://jira.mongodb.org/browse/EVG-16306)) but it's particularly appealing in light of the issues with instances not having enough room in their subnet: with Fleet they'll go in whichever subnet/AZ has room for them.

### Testing 
I started hosts in staging with Fleet set to on-demand, spot capacity-optimized, spot lowest-cost, and spot on-demand fallback.

### DIFF
--- a/public/static/js/distros.js
+++ b/public/static/js/distros.js
@@ -108,7 +108,7 @@ mciModule.controller('DistrosCtrl', function ($scope, $window, $http, $location,
     'id': 'spot',
     'display': 'Spot'
   }, {
-    'id': 'on-demand-fallback',
+    'id': 'fallback',
     'display': 'Spot with on-demand fallback'
   }, {
     'id': 'on-demand',
@@ -156,7 +156,7 @@ mciModule.controller('DistrosCtrl', function ($scope, $window, $http, $location,
         $scope.activeDistro.settings.fleet_options.use_on_demand = false;
         $scope.activeDistro.settings.fallback = false;
         break;
-      case "on-demand-fallback":
+      case "fallback":
         $scope.activeDistro.settings.fleet_options.use_on_demand = false;
         $scope.activeDistro.settings.fallback = true;
         break;
@@ -169,7 +169,7 @@ mciModule.controller('DistrosCtrl', function ($scope, $window, $http, $location,
 
   $scope.getFleetInstanceType = function(settings) {
     if (!settings?.fleet_options?.use_on_demand && settings?.fallback) {
-      return "on-demand-fallback";
+      return "fallback";
     }
 
     if (settings?.fleet_options?.use_on_demand && !settings?.fallback) {
--- a/public/static/js/distros.js
+++ b/public/static/js/distros.js
@@ -104,6 +104,17 @@ mciModule.controller('DistrosCtrl', function ($scope, $window, $http, $location,
     'display': 'OAuth Token'
   }]
 
+  $scope.fleetInstanceTypes = [{
+    'id': 'spot',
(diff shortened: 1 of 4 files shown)


service/templates/distros.html
```diff
@@ -424,11 +424,25 @@ <h2 style="display:inline-block; padding-right:15px">Configure
                 <div class="icon fa fa-warning distro-error" ng-show="form.keyName.$dirty && form.keyName.$error.required || form.keyName.$invalid">EC2
                   key name is required</div>
               </div>
-              <div class="distro-label">
-                <div ng-show="activeDistro.provider == 'ec2-spot' || activeDistro.provider == 'ec2-fleet'">
-                  <label><input style="margin-right:10px;" ng-disabled="readOnly" type="checkbox"
-                    name="fallback" ng-model="activeDistro.settings.fallback">Fallback to EC2 On-Demand in the case of insufficient capacity </label> <br>
+              <div ng-show="activeDistro.provider === 'ec2-spot'">
```
malikchaya2, line 427: can you please include a screenshot of the visible changes?

ybrill:
This is what was added to the distro provider settings section
Screen Shot 2022-03-22 at 2 24 35 PM


This is the dropdown expanded
Screen Shot 2022-03-22 at 2 24 58 PM

service/templates/distros.html
ybrill: done

malikchaya2 (approved):

public/static/js/distros.js
```diff
@@ -104,6 +104,17 @@ mciModule.controller('DistrosCtrl', function ($scope, $window, $http, $location,
     'display': 'OAuth Token'
   }]
 
+  $scope.fleetInstanceTypes = [{
+    'id': 'spot',
+    'display': 'Spot'
+  }, {
+    'id': 'fallback',
+    'display': 'Spot with on-demand fallback'
```
ablack12, line 112: nit: can we capitalize display values? (Spot with On-Demand Fallback)

ablack12 (approved):

ybrill:
The failing js-test threw me for a loop. It passed on my machine but was consistently failing in the patch. It appears support for [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) was added in node v14 and karma (the thing that runs the js tests) relies on node to parse the javascript. Since the ubuntu1604 distros have an old version of node (8.11.3) the tests were failing (my laptop has node v14.7.1).

I changed out the use of optional chaining in favor of its logical equivalent. Hopefully that fixes it.

PR #5507, opened 2022-03-22
EVG-15786: use scopes in generate tasks and tune performance

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15786

### Description
This is similar to #4640 but with some extra queue tuning to try addressing performance issues.

Scopes in Amboy are a distributed lease mechanism (i.e. a lock with an expiration) to prevent multiple conflicting generate tasks jobs from running concurrently. Adding scopes also requires tuning the performance of queues running generate tasks or else the performance of the scoped jobs will be poor in some cases. Therefore, I moved generate tasks into a single queue in the queue group and tuned the queue's parameters in the admin settings.

* Add scope for generate tasks jobs.
* Move generate tasks jobs into a single queue so that its performance can be tuned.
* Add feature flag to use new experimental features or not. Toggling the flag will break all tasks that are currently being generated in the app server. The most it can do is avoid the need to revert a deploy.
* Slightly simplify two generate tasks queries that have an `$or` but only one condition in the `$or`.
* Fix an ng-model in the admin page for Amboy queue settings.

### Testing 
* Update the existing generate tasks job test.
* Tested basic correctness in staging with the tuned parameters with the generate-lint task.


units/generate_tasks.go
```diff
@@ -47,11 +47,17 @@ func makeGenerateTaskJob() *generateTasksJob {
...
+// configuration based on the given task's generate.tasks configuration.
+func NewGenerateTasksJob(versionID, taskID string, ts string, useScopes bool) amboy.Job {
 	j := makeGenerateTaskJob()
-	j.TaskID = id
+	j.TaskID = taskID
 
-	j.SetID(fmt.Sprintf("%s-%s-%s", generateTasksJobName, id, ts))
+	j.SetID(fmt.Sprintf("%s-%s-%s", generateTasksJobName, taskID, ts))
+	if useScopes {
+		j.SetScopes([]string{versionID, taskID})
```
ablack12, line 58: I'm curious what this means; I've noticed that in some other places we use scopes we use "%s.%s" rather than passing in two strings -- is that functionally the same thing or is passing in the list different? How does this interact with SetEnqueueScopes?

Kimchelly: Each string counts as a separate lock. These locks are functionally the same as the `%s.%s` format in this case because generate task jobs are in their own dedicated queue. But for the sake of consistency, I changed it to use the `%s.%s` format including the job name.

The two locks achieve slightly different goals. `SetEnqueueScope` tells Amboy when to apply the lock in the job lifecycle. By default, a normal scope applies the lock when the job is dispatched, which prevents multiple jobs with the same scope from running at the same time. If a scope is marked as an enqueue scope, it applies the lock when the job is enqueued rather than when it's dispatched. Applying the scope earlier disallows multiple jobs with the same enqueue scope from being put in the queue. This resolves an issue where jobs that have to do the same work might end up being enqueued more than once. For example, in the current way generate tasks works (without enqueue scopes), if task ABC is running generate tasks, there's nothing stopping the cron populator from enqueueing multiple jobs that are effectively supposed to handle the same task except by adding a rounded timestamp to the `_id`. This deduplication u sing timestamps is not really necessary now that enqueue scopes are available. Using the task-level enqueue scope means that there can be at most 1 generate tasks job for ABC in the Amboy queue at any given time, either waiting to run or running.

Using the version-level scope satisfies the correctness goal of preventing multiple tasks in the same version from running concurrently. Using the task-level enqueue scope is a performance improvement to prevent duplicate generate task jobs from being enqueued for the same task.

ablack12: Very cool, thank you!

ablack12 (approved):

hadjri (approved):

units/crons.go
```diff
@@ -680,31 +680,50 @@ func PopulateAgentMonitorDeployJobs(env evergreen.Environment) amboy.QueueOperat
...
+		if err != nil {
+			return errors.WithStack(err)
+		}
+
+		if flags.GenerateTasksExperimentDisabled {
+			versions := map[string]amboy.Queue{}
+			group := env.RemoteQueueGroup()
+
+			for _, t := range tasks {
+				var q amboy.Queue
```
julianedwards, line 701: Nit:
```
var (
        q amboy.Queue
        ok bool
        err error
)

File: units/crons.go [hunk 3]
```diff
@@ -680,31 +680,50 @@ func PopulateAgentMonitorDeployJobs(env evergreen.Environment) amboy.QueueOperat
...
+				var ok bool
+				var err error
+				if q, ok = versions[t.Version]; !ok {
+					q, err = group.Get(ctx, t.Version)
+					if err != nil {
+						return errors.Wrapf(err, "problem getting queue for version %s", t.Version)
```
julianedwards, line 707: Nit: `getting queue for...`

units/generate_tasks.go
```diff
@@ -47,11 +47,19 @@ func makeGenerateTaskJob() *generateTasksJob {
...
+// configuration based on the given task's generate.tasks configuration.
+func NewGenerateTasksJob(versionID, taskID string, ts string, useScopes bool) amboy.Job {
 	j := makeGenerateTaskJob()
-	j.TaskID = id
+	j.TaskID = taskID
 
-	j.SetID(fmt.Sprintf("%s-%s-%s", generateTasksJobName, id, ts))
+	j.SetID(fmt.Sprintf("%s-%s-%s", generateTasksJobName, taskID, ts))
+	if useScopes {
+		versionScope := fmt.Sprintf("%s.%s", generateTasksJobName, versionID)
```
julianedwards, line 58: Is the prefixing the scopes with `generateTasksJobName` necessary since these jobs are in a dedicated queue group already?

julianedwards (changes requested):

units/crons.go
Kimchelly: Fixed.

units/crons.go
Kimchelly: Fixed.

units/generate_tasks.go
Kimchelly: It's not necessary, but it's following the existing convention of using the job name, which is sometimes helpful to indicate the scope's purpose instead of just using a raw ID.

julianedwards (approved):

PR #5509, opened 2022-03-22
EVG-12407 Only allow patchable tasks in configure view

khelif96:
[EVG-12407](https://jira.mongodb.org/browse/EVG-12407)

### Description 
Only allow tasks that have the patchable flag set. I would assume this value is true by default for all patchable tasks could someone on app confirm?

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -181,7 +181,9 @@ func getPatchProjectVariantsAndTasksForUI(ctx context.Context, apiPatch *restMod
 		}
 		projTasks := []string{}
 		for _, taskUnit := range buildVariant.Tasks {
-			projTasks = append(projTasks, taskUnit.Name)
+			if utility.FromBoolPtr(taskUnit.Patchable) {
+				projTasks = append(projTasks, taskUnit.Name)
+			}
 		}
 		for _, displayTask := range buildVariant.DisplayTasks {
 			projTasks = append(projTasks, displayTask.Name)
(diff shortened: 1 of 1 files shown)


ablack12:
Patchable by default is true, so you'd want to use FromBoolTPtr. But patchable: false only means that it can't run on a patch; it can still run on a variant (then there are other considerations listed in the wiki below). It looks to me like GetVariantsAndTasksFromProject should already be doing this though, which use before that https://github.com/evergreen-ci/evergreen/blob/7d08b7af8de36133c90b0b67d785c1b14bb5c258/model/project.go#L1892. 
https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#limiting-when-a-task-will-run

ablack12 (commented):
Have you been able to test that this works? What I was saying in my previous comment was that GetVariantsAndTasksFromProject _should_ already do this https://github.com/evergreen-ci/evergreen/blob/7d08b7af8de36133c90b0b67d785c1b14bb5c258/model/project.go#L1892 so I'm not convinced that adding this new line is doing anything different. 

I wonder if this is actually more related to this ticket: https://jira.mongodb.org/browse/EVG-16503 since the task Kim mentions is upload-clis, which is a display task. That being said I don't actually see this as available for scheduling in your PR or the one that she linked so maybe this has just gone away?

khelif96:
@ablack12  it looks like you are right! I don't see any patchable:false tasks when i create a patch.

PR #5512, opened 2022-03-22
EVG-16496 reorder patch insertion and trigger processing

ybrill:
[EVG-16496](https://jira.mongodb.org/browse/EVG-16496)

### Description 
It looks to me like the parent's changes aren't being applied to spruce because the alias doesn't specify the module they should be applied as. I think we want it to look like this:
Screen Shot 2022-03-22 at 6 23 33 PM

This was broken, though, for the case of finalizing the patch from the CLI. When we went looking for the parent patch [here](https://github.com/evergreen-ci/evergreen/blob/f64f66e3635f1711d483ad65dac155c386281032/units/patch_intent.go#L857) it wasn't there because [it only got inserted after](https://github.com/evergreen-ci/evergreen/blob/f64f66e3635f1711d483ad65dac155c386281032/units/patch_intent.go#L359-L365). This PR switches the order of when the child intents get processed and when the parent patch is inserted. This necessitated updating the parent patch's child patch ids since the patch was already inserted.

### Testing 
On staging the patch was created with the parent module applied.

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -360,7 +360,7 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 		return err
 	}
 
-	if _, err = ProcessTriggerAliases(ctx, patchDoc, pref, j.env, patchDoc.Triggers.Aliases); err != nil {
+	if err = ProcessTriggerAliases(ctx, patchDoc, pref, j.env, patchDoc.Triggers.Aliases); err != nil {
 		return errors.Wrap(err, "problem processing trigger aliases")
 	}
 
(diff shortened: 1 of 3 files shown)


ybrill:
It occurred to me this fix is probably also necessary for PRs, since their patches follow the same code path as CLI patches that are immediately finalized.

units/patch_intent.go
```diff
@@ -356,14 +356,14 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 	}
```
malikchaya2, line 356: I'm not sure I understand why it is that the module needs to be specified as Evergreen in Spruce's case. Since the patch is on the evergreen project, shouldn't that happen automatically?

malikchaya2: Discussed over slack. I understand it better now and will think about how to add more help text to help the user understand as well. 
I'll review again once [this](https://github.com/evergreen-ci/evergreen/pull/5512#issuecomment-1076414623) is addressed.

ybrill: Sorry, [the comment you're referring to](https://github.com/evergreen-ci/evergreen/pull/5512#issuecomment-1076414623) was unclear. What I meant was that the fix already in this PR should also address the issue for PRs as well. I was coming to say that [my original comment](https://github.com/evergreen-ci/evergreen/pull/5512#issue-1177376627) implied the only issue was for CLI patches but I realized it was also an issue for PR patches.

Would you mind to review again at your earliest convenience?

malikchaya2: ah that makes sense, thanks!

malikchaya2 (approved):
🎉

PR #5513, opened 2022-03-23
EVG-16231: populate the pod allocator jobs

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16231

### Description 
* Add some helper methods to container task queue.
* Populate pod allocator jobs in remote queue group from container task queue.
* Add service flag for disabling pod allocation.
* Respect max parallel pod request limit in cron populator and in the pod allocator job itself. 
* I noticed an oddity where the mock.Environment uses the mock DB sessions but allows the usage of a real MongoDB client, so I just switched it to use the MongoDB client session.

### Testing 
Added unit tests. Also added an integration test between the container task queue, cron populator, and pod allocation.

### DIFF
--- a/model/container_task_queue.go
+++ b/model/container_task_queue.go
@@ -15,23 +15,44 @@ type ContainerTaskQueue struct {
 	position int
 }
 
+// NewContainerTaskQueue returns a populated iterator representing an ordered
+// queue of container tasks that are ready to be allocated a container.
+func NewContainerTaskQueue() (*ContainerTaskQueue, error) {
+	q := &ContainerTaskQueue{}
+	if err := q.populate(); err != nil {
+		return nil, errors.Wrap(err, "initial population of container task queue")
+	}
+	return q, nil
+}
+
 // Next returns the next task that's ready for container allocation. It will
 // return a nil task once there are no tasks remaining in the queue.
-func (q *ContainerTaskQueue) Next() (*task.Task, error) {
-	if q.queue == nil {
-		if err := q.populate(); err != nil {
-			return nil, errors.Wrap(err, "initial population of container task queue")
-		}
-	}
+func (q *ContainerTaskQueue) Next() *task.Task {
 	if q.position >= len(q.queue) {
-		return nil, nil
+		return nil
 	}
 
 	next := q.queue[q.position]
 
 	q.position++
 
-	return &next, nil
+	return &next
+}
+
+// HasNext returns whether or not there are more container tasks that have not
+// yet been returned.
+func (q *ContainerTaskQueue) HasNext() bool {
+	return q.position < len(q.queue)
+}
+
+// Len returns the total number of tasks in the queue.
+func (q *ContainerTaskQueue) Len() int {
+	return len(q.queue)
+}
+
+// Remaining returns the number of tasks that have not yet been returned.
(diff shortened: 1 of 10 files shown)


units/crons.go
```diff
@@ -1260,6 +1260,62 @@ func PopulateDataCleanupJobs(env evergreen.Environment) amboy.QueueOperation {
 	}
 }
 
+func PopulatePodAllocatorJobs(env evergreen.Environment) amboy.QueueOperation {
```
julianedwards, line 1263: Documentation.

julianedwards (changes requested):
Instead of relying on the pod status for getting the count of initializing pods, would it simply things (and maybe be more precise) to limit the size of the pod allocation amboy queue based on the `MaxParallelPodRequests`?

model/container_task_queue.go
```diff
@@ -15,23 +15,44 @@ type ContainerTaskQueue struct {
...
+}
+
+// HasNext returns whether or not there are more container tasks that have not
+// yet been returned.
+func (q *ContainerTaskQueue) HasNext() bool {
+	return q.position < len(q.queue)
+}
+
+// Len returns the total number of tasks in the queue.
+func (q *ContainerTaskQueue) Len() int {
```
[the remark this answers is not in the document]
Kimchelly: Yeah, I think I was originally going to use them for two separate things but ended up not doing that later on. I consolidated them so there's just one `Len` method returning the # tasks remaining.

units/crons.go
Kimchelly: Done.

Kimchelly (commented):
This populator accounts for the maximum potential rate of requests by counting the number of intent pods (i.e. initializing ones) and number of pod allocator jobs, each of which can create at most one more intent pod (for now). Counting just the number of pod allocator jobs would undercount the number of pods that could be making cloud requests in parallel. It would also pose an efficiency problem because queues in the queue group cannot efficiently query their queue stats.

The actual place where we have to throttle request rate with the max parallel pod request limit is in the [pod creation job](https://github.com/evergreen-ci/evergreen/blob/f01780508d4604e5f18358de8f5f5491c1837941/units/pod_creation.go#L108), not the pod allocator. All the pod allocator does is making an intent pod. The intent pod gets turned into a real pod running in the cloud by the pod creation job.

units/pod_allocator.go
```diff
@@ -93,6 +113,26 @@ func (j *podAllocatorJob) Run(ctx context.Context) {
...
+	}
+	numInitializing, err := pod.CountByInitializing()
+	if err != nil {
+		return false, errors.Wrap(err, "counting currently-initializing pods")
+	}
+	if numInitializing >= settings.PodInit.MaxParallelPodRequests {
+		return false, nil
+	}
+
+	flags, err := evergreen.GetServiceFlags()
```
julianedwards, line 129: Not really a big deal, but it feels like this check should come first since if the pod allocator is disabled, no point in doing the count query.

julianedwards (approved):
LGTM % nit

units/pod_allocator.go
Kimchelly: Good catch, I swapped them.

PR #5514, opened 2022-03-23
EVG-16609: Fix handleGithubConflicts

sophstad:
[EVG-16609](https://jira.mongodb.org/browse/EVG-16609)

### Description 

### Testing 

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -184,7 +184,11 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 		}
 		modified = true
 	case model.ProjectPageGithubAndCQSection:
-		if err = handleGithubConflicts(mergedProjectRef, "Toggling GitHub features"); err != nil {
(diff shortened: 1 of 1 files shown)


malikchaya2 (approved):

rest/data/project_settings.go
```diff
@@ -184,7 +190,11 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 		}
 		modified = true
 	case model.ProjectPageGithubAndCQSection:
-		if err = handleGithubConflicts(mergedProjectRef, "Toggling GitHub features"); err != nil {
+		conflictProjectRef := mergedProjectRef
+		conflictProjectRef.Owner = mergedBeforeRef.Owner
```
ablack12, line 194: Can we just add these directly to mergedProjectRef? The backend only updates fields that are on that page so it wouldn't cause a problem (since mergedProjectRef is a pointer anyway I'm not sure this isn't already modifying it)

ablack12 (approved):

PR #5515, opened 2022-03-23
EVG-16558 stop storing duplicate params in the DB

ablack12:
[EVG-16558](https://jira.mongodb.org/browse/EVG-16558)

### Description 
SO I realized that we've been storing both Params and ParamsYAML in the DB, and only storing one of these would reduce yaml sizes. 
The reason we store ParamsYAML is because Params is a `map[string]interface{}`, but mongoDB doesn't handle nested interfaces well and this data can become corrupted, so in order to save ParserProject in it's own collection we needed a different way to store it.

The reason we kept Params is just so that, while ParserProject was in development, reverting the work wouldn't leave the parser project string in a state that we wouldn't know how to interpret (i.e. ParamsYAML only). 

It's possible that a few old versions created during the time we were migrating to ParserProject will have issues with this, but since we're still reading Params in from the YAML and most of these read from the version ParserProject yaml string I don't foresee any issues.

### Testing 
Tested in staging that [new ](https://evergreen-staging.corp.mongodb.com/version/623b4d7a9dbe320a169bf4ba)and old patches work with this change, as well as validate, and verified in the DB that paramsYAML was no longer being stored.

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -396,10 +396,11 @@ type PluginCommandConf struct {
 	// TimeoutSecs indicates the maximum duration the command is allowed to run for.
 	TimeoutSecs int `yaml:"timeout_secs,omitempty" bson:"timeout_secs,omitempty"`
 
-	// Params are used to supply configuration specific information.
-	Params map[string]interface{} `yaml:"params,omitempty" bson:"params,omitempty"`
+	// Params is used to define params in the yaml and parser project,
+	// but is not stored in the DB (instead see ParamsYAML).
+	Params map[string]interface{} `yaml:"params,omitempty" bson:"-"`
 
-	// YAML string of Params to store in database
+	// ParamsYAML is the marshalled Params to store in the database, to preserve nested interfaces.
 	ParamsYAML string `yaml:"params_yaml,omitempty" bson:"params_yaml,omitempty"`
 
 	// Vars defines variables that can be used within commands.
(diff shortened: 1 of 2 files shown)


malikchaya2 (approved):
and NOW, the parser project is finally complete :)

model/project.go
```diff
@@ -396,10 +396,11 @@ type PluginCommandConf struct {
 	// TimeoutSecs indicates the maximum duration the command is allowed to run for.
 	TimeoutSecs int `yaml:"timeout_secs,omitempty" bson:"timeout_secs,omitempty"`
 
-	// Params are used to supply configuration specific information.
-	Params map[string]interface{} `yaml:"params,omitempty" bson:"params,omitempty"`
+	// Params is used to define params in the yaml and parser project,
+	// but is not stored in the DB (instead see ParamsYAML).
+	Params map[string]interface{} `yaml:"params,omitempty" bson:"-"`
```
bynn, line 401: Is this the only change needed to stop storing it in the dB?

bynn (approved):

model/project.go
ablack12: Yep! The struct tag "-" tells Go to ignore that field when parsing. Because we only change this for BSON, it ensures that we still read it in for yaml. This is also helpful for fields that we want to pass around within the code but don't want to actually store because it's repeated (for example, DisplayTask in the Task object). 
https://vsupalov.com/go-json-omitempty/#ignore-json-fields-when-parsing-it-even-though-your-struct-has-them

PR #5516, opened 2022-03-23
EVG-16577: do not limit max hosts for generate-lint

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16577

### Description 
Don't use a single-host task group for patches unless there's only 1 task. The single-host task group behavior is undesirable for patches containing multiple packages to lint, since it prevents other packages from being linted if any lint task fails.

### Testing 
N/A

### DIFF
--- a/cmd/generate-lint/generate-lint.go
+++ b/cmd/generate-lint/generate-lint.go
@@ -20,8 +20,6 @@ const (
 	lintPrefix        = "lint"
 	lintVariant       = "lint"
 	lintGroup         = "lint-group"
-	commitMaxHosts    = 4
-	patchMaxHosts     = 2
 	evergreenLintTask = "evergreen"
 	jsonFilename      = "bin/generate-lint.json"
 	scriptsDir        = "scripts"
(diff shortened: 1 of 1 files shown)


cmd/generate-lint/generate-lint.go
```diff
@@ -135,11 +135,15 @@ func generateTasks() (*shrub.Configuration, error) {
...
 	} else {
-		maxHosts = patchMaxHosts
 		targets, err = targetsFromChangedFiles(changes)
 		if err != nil {
 			return nil, err
 		}
+		if len(targets) == 1 {
+			maxHosts = 1
+		} else {
+			maxHosts = patchMaxHosts
```
ablack12, line 145: should we just set this to len(targets) since we don't really care if they run on different hosts?

Kimchelly: In that case, we should just set it to `len(targets)` for both patches and waterfall versions.

I removed the max host limit.

ablack12 (approved):

PR #5517, opened 2022-03-23
EVG-16428 Allow custom container resource aliases defined on project page and in yaml

hadjri:
[EVG-16428](https://jira.mongodb.org/browse/EVG-16428)

### Description 
Added a form on the projects page allowing users to define default container resource sizes that are referenced by an alias. These can be saved to the project ref. Then, when the 'size' property is set on a container in YAML, the resource specs from the same name will be retrieved from the project ref at runtime.
### Testing 
Updated existing unit tests.

### DIFF
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -530,12 +530,12 @@ mciModule.controller(
             disabled_stats_cache: data.ProjectRef.disabled_stats_cache,
             periodic_builds: data.ProjectRef.periodic_builds,
             container_sizes: data.ProjectRef.container_sizes || {},
-            container_size_names: [],
             use_repo_settings: $scope.projectRef.use_repo_settings,
             build_baron_settings: data.ProjectRef.build_baron_settings || {},
             task_annotation_settings: data.ProjectRef.task_annotation_settings || {},
             perf_enabled: data.ProjectRef.perf_enabled || false,
           };
+          $scope.settingsFormData.container_size_names = Object.keys($scope.settingsFormData.container_sizes)
           // Divide aliases into categories
           $scope.settingsFormData.github_pr_aliases = $scope.aliases.filter(
             function (d) {
@@ -634,6 +634,10 @@ mciModule.controller(
     };
 
     $scope.saveProject = function () {
+        _.each($scope.settingsFormData.container_sizes, function (size) {
+            size.cpu = parseInt(size.cpu)
+            size.memory_mb = parseInt(size.memory_mb)
+        });
       $scope.settingsFormData.batch_time = parseInt(
         $scope.settingsFormData.batch_time
       );
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -529,6 +529,8 @@ mciModule.controller(
             github_trigger_aliases: data.ProjectRef.github_trigger_aliases || [],
             disabled_stats_cache: data.ProjectRef.disabled_stats_cache,
             periodic_builds: data.ProjectRef.periodic_builds,
+            container_sizes: data.ProjectRef.container_sizes || {},
+            container_size_names: [],
             use_repo_settings: $scope.projectRef.use_repo_settings,
             build_baron_settings: data.ProjectRef.build_baron_settings || {},
             task_annotation_settings: data.ProjectRef.task_annotation_settings || {},
@@ -668,6 +670,10 @@ mciModule.controller(
 
       if ($scope.patch_alias) {
         $scope.addPatchAlias();
+      }
+
+      if ($scope.container_size) {
+        $scope.addContainerSize();
       }
 
       $scope.settingsFormData.subscriptions = _.filter(
@@ -812,6 +818,22 @@ mciModule.controller(
       $scope.invalidPatchAliasMessage = "";
     };
 
+    $scope.addContainerSize = function () {
+      if (!$scope.validContainerSize($scope.container_size)) {
+        $scope.invalidContainerSizeMessage =
+          "A valid container size must have a name, memory and cpu properties.";
+          return;
+      }
+      var item = Object.assign({}, $scope.container_size);
+      $scope.settingsFormData.container_sizes[item.name] = {
+          "memory_mb": item.memory_mb,
+          "cpu": item.cpu,
+      }
+      $scope.settingsFormData.container_size_names = Object.keys($scope.settingsFormData.container_sizes)
+      delete $scope.container_size;
+      $scope.invalidContainerSizeMessage = "";
+    };
+
     $scope.addWorkstationCommand = function () {
       if (!$scope.settingsFormData.workstation_config) {
         $scope.settingsFormData.workstation_config = {};
@@ -880,6 +902,14 @@ mciModule.controller(
       $scope.settingsFormData.patch_aliases.splice(i, 1);
       $scope.isDirty = true;
     };
+
+      $scope.removeContainerSize = function (name) {
+          if ($scope.settingsFormData.container_sizes[name]) {
+              delete $scope.settingsFormData.container_sizes[name]
+              $scope.settingsFormData.container_size_names = Object.keys($scope.settingsFormData.container_sizes)
+          }
+          $scope.isDirty = true;
+      };
 
     $scope.removeProjectTrigger = function (i) {
       if ($scope.project_triggers[i]) {
@@ -1108,6 +1138,11 @@ mciModule.controller(
     $scope.validPatchAlias = function (alias) {
       // Same as GitHub alias, but with alias required
       return $scope.validPatchDefinition(alias) && alias.alias;
+    };
+
+    $scope.validContainerSize = function (container_size) {
+      // Same as GitHub alias, but with alias required
+      return container_size && container_size.cpu && container_size.memory_mb;
     };
 
     $scope.validWorkstationCommand = function (obj) {
--- a/public/static/js/projects.js
+++ b/public/static/js/projects.js
@@ -381,9 +381,6 @@ mciModule.controller(
 
       // if copyProject is set, copy the current project
       if ($scope.newProject.copyProject) {
-        $scope.settingsFormData.batch_time = parseInt(
-          $scope.settingsFormData.batch_time
-        );
         $http
           .put("/project/" + $scope.newProject.identifier, $scope.newProject)
           .then(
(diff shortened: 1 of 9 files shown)


service/templates/projects.html
```diff
@@ -968,6 +968,54 @@ <h3>Periodic Builds</h3>
...
+          <div class="col-lg-2"> <label class="control-label"> Memory MB </label> </div>
+          <div class="col-lg-2"> <label class="control-label"> CPU </label> </div>
+          <div class="col-lg-2"></div>
+        </div>
+        <div id="container-sizes-list" class="form-group" ng-repeat="obj in settingsFormData.container_sizes track by $index">
+          <div class="col-lg-2">
+            <input class="form-control" ng-model="settingsFormData.container_size_names[$index]" type="text" placeholder="name">
+          </div>
+          <div class="col-lg-2">
+            <input class="form-control" ng-model="obj.memory_mb" type="text" placeholder="memory mb">
```
Kimchelly, line 990: Should this have `type="number"` to ensure it's an int?

service/templates/projects.html
```diff
@@ -968,6 +968,54 @@ <h3>Periodic Builds</h3>
...
+          </div>
+          <div class="col-lg-2">
+            <input class="form-control" ng-model="obj.cpu" type="text" placeholder="cpu">
```
Kimchelly, line 993: Same thing as above about the input type.

public/static/js/projects.js
```diff
@@ -1110,6 +1144,11 @@ mciModule.controller(
       return $scope.validPatchDefinition(alias) && alias.alias;
     };
 
+    $scope.validContainerSize = function (container_size) {
```
Kimchelly, line 1147: I know the save button already checks if they're integers, but should it instead check earlier when adding the container size that CPU/memory are non-negative ints?

service/project.go
```diff
@@ -630,6 +631,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 	projectRef.DisabledStatsCache = &responseRef.DisabledStatsCache
 	projectRef.PeriodicBuilds = []model.PeriodicBuildDefinition{}
 	projectRef.PerfEnabled = &responseRef.PerfEnabled
+	projectRef.ContainerSizes = responseRef.ContainerSizes
```
Kimchelly, line 634: Should there be validation above this like there is for the other structs to make sure the container sizes are valid?

service/project.go
```diff
@@ -312,6 +312,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 		PerfEnabled             bool                                `json:"perf_enabled"`
 		BuildBaronSettings      restModel.APIBuildBaronSettings     `json:"build_baron_settings"`
 		TaskAnnotationSettings  restModel.APITaskAnnotationSettings `json:"task_annotation_settings"`
+		ContainerSizes          map[string]model.ContainerResources `json:"container_sizes"`
```
Kimchelly, line 315: Can this use the rest model instead of the DB model? That way, it could use `ToService` to create the DB model.

service/templates/projects.html
```diff
@@ -968,6 +968,54 @@ <h3>Periodic Builds</h3>
...
+        </div>
+        <div id="container-sizes-list-header" class="form-group">
+          <div class="col-lg-2"> <label class="control-label"> Name </label> </div>
+          <div class="col-lg-2"> <label class="control-label"> Memory MB </label> </div>
+          <div class="col-lg-2"> <label class="control-label"> CPU </label> </div>
+          <div class="col-lg-2"></div>
+        </div>
+        <div id="container-sizes-list" class="form-group" ng-repeat="obj in settingsFormData.container_sizes track by $index">
+          <div class="col-lg-2">
+            <input class="form-control" ng-model="settingsFormData.container_size_names[$index]" type="text" placeholder="name">
```
Kimchelly, line 987: I'm not that good with JS, but since it seems like `container_size_names` is only needed to track the name here, is there no other way to just get the container size name from `ng-repeat` on  `container_sizes` keys? Can we do something like `ng-repeat="(name, obj) in settingsFormData.container_sizes`?

Kimchelly (changes requested):
Can you post a gif of the UI page functionality?

Is this supposed to be added to the rest model's `APIProjectRef`? It would make the REST/DB conversions cleaner.

public/static/js/projects.js
```diff
@@ -1110,6 +1144,11 @@ mciModule.controller(
...
+      // Same as GitHub alias, but with alias required
```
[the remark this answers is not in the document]
hadjri: Nothing - bad copy. Removing

hadjri:
> assert.Equal(t, 2, projectRef.ContainerSizes["large"].CPU)

container_size_demo

hadjri:
> Can you post a gif of the UI page functionality?
> 
> Is this supposed to be added to the REST model's `APIProjectRef`? It would make the REST/DB conversions cleaner and include container sizes in the project ref response.

container_size_demo

service/templates/projects.html
hadjri: That works! I tried something similar eariler but removed it because it was erroring out. This approach works though.

service/project.go
```diff
@@ -586,6 +601,14 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 	for i, buildDef := range responseRef.PeriodicBuilds {
 		catcher.Wrapf(buildDef.Validate(), "invalid periodic build definition on line %d", i+1)
 	}
+	for _, containerResource := range responseRef.ContainerSizes {
+		if *containerResource.CPU <= 0 {
```
Kimchelly, line 605: I would use `utility.ToIntPtr(containerResource.CPU) < 0` since the goroutine will panic if CPU is nil.

service/project.go
```diff
@@ -586,6 +601,14 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
...
+			catcher.New("container resource CPU must be a positive integer")
+		}
+		if *containerResource.MemoryMB <= 0 {
```
Kimchelly, line 608: Same two things as above.

service/project.go
```diff
@@ -586,6 +601,14 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 	for i, buildDef := range responseRef.PeriodicBuilds {
 		catcher.Wrapf(buildDef.Validate(), "invalid periodic build definition on line %d", i+1)
 	}
+	for _, containerResource := range responseRef.ContainerSizes {
+		if *containerResource.CPU <= 0 {
+			catcher.New("container resource CPU must be a positive integer")
```
Kimchelly, line 606: nit: L605-606 can be combined into `catcher.NewWhen(utility.ToIntPtr(containerResource.CPU) <= 0, "container resource CPU must be a positive integer")`.

rest/model/project.go
```diff
@@ -373,6 +373,33 @@ type APIWorkstationConfig struct {
...
 }
 
+type APIContainerResources struct {
+	MemoryMB *int `bson:"memory_mb" json:"memory_mb"`
+	CPU      *int `bson:"cpu" json:"cpu"`
+}
+
+func (cr *APIContainerResources) BuildFromService(h interface{}) error {
+	var def model.ContainerResources
+	switch h.(type) {
```
Kimchelly, line 383: Type asserting a second time in each switch case is a little awkward; you can assign a typed value on the switch line with something like this:
```go
switch dbContainerResources := h.(type) {
case model.ContainerResources:
    def = dbContainerResources
case *model.ContainerResources:
    def = *dbContainerResources
default:
    return errors.Errorf("...")
}
```

Alternatively, you could get rid of the `interface{}` input entirely and just pass in a `model.ContainerResources` type. The only reason a lot of the models pass in interface and return interfaces is to satisfy [this interface](https://github.com/evergreen-ci/evergreen/blob/b221ba413febb0c7d9d2010acac8f3d662edc09b/rest/model/model.go#L6-L9). However, that model interface is only used in a few specific situations (e.g. admin settings, event logs, some top-level REST models `APITask` which are used in user notifications) For all other settings models like this one, it's not necessary to satisfy that interface.

rest/model/project.go
```diff
@@ -373,6 +373,33 @@ type APIWorkstationConfig struct {
...
+		def = *h.(*model.ContainerResources)
+	default:
+		return errors.Errorf("Invalid container resources of type '%T'", h)
+	}
+	cr.MemoryMB = utility.ToIntPtr(def.MemoryMB)
+	cr.CPU = utility.ToIntPtr(def.CPU)
+	return nil
+}
+
+func (cr *APIContainerResources) ToService() (interface{}, error) {
```
Kimchelly, line 396: Same thing as above, you don't need to return an `interface{}` if you prefer to just pass back `model.ContainerResource`, since this isn't an admin setting or an event log.

repotracker/repotracker_test.go
```diff
@@ -266,6 +266,10 @@ func TestStoreRepositoryRevisions(t *testing.T) {
 }
 
 func TestBatchTimeForTasks(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env := testutil.NewEnvironment(ctx, t)
+	evergreen.SetEnvironment(env)
```
Kimchelly, line 272: I would unset this environment in a defer since it could affect later tests that need the environment.

Kimchelly (changes requested):

repotracker/repotracker_test.go
hadjri: I actually dont think I need to do this

rest/model/project.go
hadjri: Sure sounds much cleaner to me

repotracker/repotracker_test.go
```diff
@@ -328,7 +330,7 @@ tasks:
 	assert.NoError(t, d.Insert())
 
 	p := &model.Project{}
-	ctx := context.Background()
+	ctx = context.Background()
```
Kimchelly, line 333: This assignment shouldn't be needed since the wrapped context on L269 is already using the background context. The `context.WithCancel(context.Background())` wrapping at the top is only used to ensure that any context-dependent operations that were running in the test (e.g. goroutines) will stop when the main goroutine (i.e. the one running this test) reaches the last line of this test.

Kimchelly (approved):
LGTM mod lint/tests passing.

PR #5518, opened 2022-03-23
EVG-12839 Automatically call manifest.load whenever calling git.get_project

malikchaya2:
[EVG-12839 ](https://jira.mongodb.org/browse/EVG-12839 )

### Description 
Whenever the git.get_project command is run, it checks if manifest.load already ran and if it hasn't, it runs it first. 

### Testing 
Added a unit test


agent/command.go
```diff
@@ -114,6 +115,33 @@ func (a *Agent) runCommandSet(ctx context.Context, tc *taskContext, commandInfo
 			tc.setCurrentIdleTimeout(nil)
 		}
 
+		if cmd.Name() == "manifest.load" {
+			manifestLoadCalled = true
+		}
+
+		if cmd.Name() == "git.get_project" && !manifestLoadCalled {
```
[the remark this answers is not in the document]
malikchaya2: Hmmm, I'm not sure. I think the if the user does it again it may be intentional? 
cc: Ryan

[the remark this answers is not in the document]
rtimmons: The hypothesis of this change is that calling `get_project` without `manifest.load` is a bug in a project's config and that this bug prevents proper module pinning.

If the project is calling `manifest.load` *after* `get_project` then either it is a bug, or they are intentionally lying to later parts in the project about the `${*_rev}` expansions that were used during checkout or something. The workaround for them in that case is to simply not use the expansions and to use the original branch spec used in the module definition.

So TLDR: Reasonable question, but if we're going to consider this case, I'd advocate to error if the user attempts to call `manifest.load` after `get_project`: it's a sign that evergreen (the team)'s understanding of the module lifecycle is very different from the project's.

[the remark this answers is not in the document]
ablack12: Erroring seems strong since it could break existing behavior but it could make sense to no-op and/or provide a warning that this may not be correct (not sure how easy it would be to do this though, possibly a second ticket)

agent/command.go
```diff
@@ -114,6 +115,33 @@ func (a *Agent) runCommandSet(ctx context.Context, tc *taskContext, commandInfo
...
+			tc.logger.Task().Infof("calling manifest.load on variant %s ",
+				tc.taskConfig.BuildVariant.Name)
+			a.runCommands(ctx, tc, []model.PluginCommandConf{
```
ablack12, line 125: Ideally we would run this as _part_ of git.get_project as opposed to it's own command here. Can git.get_project have a new parameter "runManifestLoad" or something that can be added to the params? And the code in manifest.load could be moved to a helper function that can be called explicitly by git.get_project? I'm thinking a future outcome of this work would be to remove the manifest.load command.

agent/command.go
```diff
@@ -114,6 +115,33 @@ func (a *Agent) runCommandSet(ctx context.Context, tc *taskContext, commandInfo
 			tc.setCurrentIdleTimeout(nil)
 		}
 
+		if cmd.Name() == "manifest.load" {
```
ablack12, line 118: Rather than hardcoding command names, if we're going to be looking at them here can we add to globals? 
https://github.com/evergreen-ci/evergreen/blob/7d08b7af8de36133c90b0b67d785c1b14bb5c258/globals.go#L514

ablack12 (changes requested):

agent/command.go
malikchaya2: Based on what Ryan explained [here](https://jira.mongodb.org/browse/EVG-12839), it doesn't seem like it would be useful to take it in as a parameter and let users set it to true or false because git.get_project should never run without manifest.load.

malikchaya2: The problem with this approach would also be that we don't know if manifest.load was already called, and we may end up repeating work a lot of the time.

agent/command.go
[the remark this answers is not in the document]
rtimmons: Ya not erroring is fine.

agent/command.go
rtimmons: IMHO this change is 👍🏽 as it is. We could shuffle things around to make the manifest.load logic a true part of get_project and make manifest.load a true nop and remove/nop/deprecate the command but that feels like a future optimization.

:shipit::shipit::shipit:

ablack12: What I meant was that _we_ could add manifestLoadCalled to commandInfo.Params (rather than asking users to do it) and then at the top of git.get_project we could call the manifest load logic. I'm happy to pair on this if it isn't clear.

I'd really like to avoid putting this much command specific implementation logic in the agent, which is already pretty intricate. If we're trying to change the behavior of the git.get_project it seems like it belongs in the command itself.

malikchaya2: I deprecated the manifest.load command and moved the logic into git.get_project.

agent/command/git_test.go
```diff
@@ -962,3 +965,90 @@ index edc0c34..8e82862 100644
 	}
 	s.True(foundSuccessMessage, "did not see the following in task output: %s", successMessage)
 }
+
+func TestManifestLoad(t *testing.T) {
```
malikchaya2, line 969: this was only moved, I didn't add this.

agent/command/git.go
```diff
@@ -213,6 +213,31 @@ func (opts cloneOpts) buildSSHCloneCommand() ([]string, error) {
 	}, nil
 }
 
+func moduleRevExpansionName(name string) string { return fmt.Sprintf("%s_rev", name) }
+
+// Load performs a GET on /manifest/load
+func (c *gitFetchProject) manifestLoad(ctx context.Context,
```
malikchaya2, line 219: this was only moved.

agent/command/deprecated.go
```diff
@@ -56,3 +57,15 @@ func (cc *shellTrack) Execute(ctx context.Context,
...
+
+type manifestLoad struct{ base }
+
+func manifestLoadFactory() Command                                      { return &manifestLoad{} }
+func (c *manifestLoad) Name() string                                    { return evergreen.ManifestLoadCommandName }
+func (c *manifestLoad) ParseParams(params map[string]interface{}) error { return nil }
+func (c *manifestLoad) Execute(ctx context.Context,
+	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
+
+	logger.Execution().Warning("manifest.load is deprecated. Manifest load is now called automatically in git.get_project.")
```
ablack12, line 69: Can we use manifestLoadCommandName in this warning?

agent/command/git.go
ablack12, line 219: It looks to me like this can be a helper function that isn't a method on gitFetchProject, since it doesn't use anything within c.

ablack12 (commented):
This looks pretty good to me -- it's riskier than the before change of still allowing manifest.load to work (which I think we still could support checking if manifest load has already run by passing it through params) but would be the eventual preferred state, so it could be okay to deploy and if we have issues we can reconsider a more gradual approach.

agent/command/git.go
malikchaya2: I did that because otherwise when it's called above, it automatically references [this struct](https://github.com/evergreen-ci/evergreen/blob/main/agent/command/manifest.go#L15) and not the function. Making it part of c felt cleaner than messing with the name.

agent/command/deprecated.go
malikchaya2: done

ablack12 (approved):

PR #5519, opened 2022-03-23
EVG-16350 Add MergedFrom to commit queue merge patches

bynn:
[EVG-16350](https://jira.mongodb.org/browse/EVG-16350)

### Description 
Add MorgedFrom field so that when you enqueue patch from cli or ui, it will contain the patch id of the original patch it was based off

### Testing 
unit test

tested on spruce 
https://evergreen-staging.corp.mongodb.com/rest/v2/patches/623b949e97b1d310aeb0b06d

on cli 
https://evergreen-staging.corp.mongodb.com/rest/v2/patches/623cbaa19dbe3216e5c60c76

### DIFF
--- a/model/patch/patch.go
+++ b/model/patch/patch.go
@@ -172,6 +172,8 @@ type Patch struct {
 	// MergeStatus is only used in gitServePatch to send the status of this
 	// patch on the commit queue to the agent
 	MergeStatus string `json:"merge_status"`
+	// MergedFrom is only used when a merge patch is created from an existing patch.
+	MergedFrom string `bson:"merged_from,omitempty"`
 }
 
 func (p *Patch) MarshalBSON() ([]byte, error)  { return mgobson.Marshal(p) }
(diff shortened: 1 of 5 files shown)


model/patch/patch.go
```diff
@@ -172,6 +172,8 @@ type Patch struct {
 	// MergeStatus is only used in gitServePatch to send the status of this
 	// patch on the commit queue to the agent
 	MergeStatus string `json:"merge_status"`
+	// MergedFrom is only used when a merge patch is created from an existing patch.
```
ablack12, line 175: Can this specify that the value is set to the patch ID of the existing patch?

ablack12 (commented):
Were you able to test both the CLI and UI methods of merging?

model/patch/patch.go
bynn: done

model/patch/patch.go
```diff
@@ -172,6 +172,9 @@ type Patch struct {
 	// MergeStatus is only used in gitServePatch to send the status of this
 	// patch on the commit queue to the agent
 	MergeStatus string `json:"merge_status"`
+	// MergedFrom is populated with patch id of the existing patch
```
ablack12, line 175: nit: "with the"

ablack12: and maybe also an ", if applicable." at the end since this isn't typically populated

ablack12 (approved):

PR #5520, opened 2022-03-24
EVG-16605: Expose VersionControlEnabled via GraphQL

sophstad:
[EVG-16605](https://jira.mongodb.org/browse/EVG-16605)

### Description 
- Expose `VersionControlEnabled` and allow it to be saved
- Note that I assumed that projects can default to repo values for this setting, i.e. `VersionControlEnabled` can be null. Let me know if this is not the case!

### Testing 
- Tested manually in GraphQL playground


sophstad:
evergreen retry

ablack12 (approved):

PR #5522, opened 2022-03-24
EVG-15391 Allow non-admin users to override dependencies for patches

bynn:
[EVG-15391](https://jira.mongodb.org/browse/EVG-15391)

### Description 
non-admins can override dependencies in patches

### Testing 
tested on staging that non-admins can click the button (even on spruce) but cannot on waterfall tasks

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -435,8 +435,12 @@ func (r *taskResolver) CanOverrideDependencies(ctx context.Context, at *restMode
 		RequiredLevel: evergreen.TasksAdmin.Value,
 		Resource:      *at.ProjectId,
 	}
-	if len(at.DependsOn) > 0 && (currentUser.HasPermission(requiredPermission) ||
-		utility.StringSliceContains(evergreen.PatchRequesters, utility.FromStringPtr(at.Requester))) {
+	overrideRequesters := []string{
+		evergreen.PatchVersionRequester,
+		evergreen.GithubPRRequester,
+	}
+	if len(at.DependsOn) > 0 && (utility.StringSliceContains(overrideRequesters, utility.FromStringPtr(at.Requester)) ||
+		currentUser.HasPermission(requiredPermission)) {
 		return true, nil
 	}
 	return false, nil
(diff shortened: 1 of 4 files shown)


graphql/resolvers.go
```diff
@@ -435,7 +435,8 @@ func (r *taskResolver) CanOverrideDependencies(ctx context.Context, at *restMode
 		RequiredLevel: evergreen.TasksAdmin.Value,
 		Resource:      *at.ProjectId,
 	}
-	if len(at.DependsOn) > 0 && currentUser.HasPermission(requiredPermission) {
+	if len(at.DependsOn) > 0 && (currentUser.HasPermission(requiredPermission) ||
+		utility.StringSliceContains(evergreen.PatchRequesters, utility.FromStringPtr(at.Requester))) {
```
ablack12, line 439: The ticket does note that this shouldn't include commit queue patches, which PatchRequesters does include, so we may just want to explicitly check the correct patch requesters.

graphql/resolvers.go
```diff
@@ -435,7 +435,8 @@ func (r *taskResolver) CanOverrideDependencies(ctx context.Context, at *restMode
 		RequiredLevel: evergreen.TasksAdmin.Value,
 		Resource:      *at.ProjectId,
 	}
-	if len(at.DependsOn) > 0 && currentUser.HasPermission(requiredPermission) {
+	if len(at.DependsOn) > 0 && (currentUser.HasPermission(requiredPermission) ||
```
ablack12, line 438: Ors are evaluated in order. There's no reason to look up if the user has permission (which is a DB query) if it's a patch requester, so we should put this as the latter part of the clause.

ablack12 (changes requested):

graphql/resolvers.go
bynn: done

bynn (commented):
added legacy support

ablack12 (approved):

public/static/js/task_admin.js
```diff
@@ -25,6 +25,7 @@ mciModule.controller('AdminOptionsCtrl', ['$scope', '$window', '$rootScope', 'mc
         $scope.canSchedule = !$scope.task.activated && !$scope.canRestart && !$scope.isAborted;
         $scope.canUnschedule = $scope.task.activated && ($scope.task.status == "undispatched") ;
         $scope.canSetPriority = ($scope.task.status == "undispatched");
+        $scope.canOverrideDependencies = ($scope.task.r == "patch_request" || $scope.task.r == "github_pull_request" || $window.permissions.project_tasks>=30);
```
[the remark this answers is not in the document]
bynn: thats the admin level and it was >30 before as well i just moved it from html to here

malikchaya2 (approved):

PR #5524, opened 2022-03-25
EVG-16608 Fix Project/Repo Events & Settings collision

malikchaya2:
[EVG-16608](https://jira.mongodb.org/browse/EVG-16608)

### Description 
The resolvers were returning the wrong data for event logs because it was sharing a resolver with projectSettings and the data shape is different. For events, the vars and other variables are stored right in the struct and should be fetched from there. When it was using the projectSettings resolvers, certain fields had separate resolvers that added calculations or fetched those field from different collections which was causing the data to be overwritten. 

### Testing 
Tested in staging and fixed the graphql tests.


graphql/resolvers.go
```diff
@@ -1286,6 +1286,7 @@ func (r *patchResolver) CommitQueuePosition(ctx context.Context, apiPatch *restM
 	return commitQueuePosition, nil
 }
 
+//here
```
sophstad, line 1289: Delete

graphql/schema.graphql
```diff
@@ -1144,6 +1144,22 @@ type ProjectSettings {
...
 
+type ProjectEventsSettings{
+  githubWebhooksEnabled: Boolean!
+  projectRef: Project
+  vars: ProjectVars
+  aliases: [ProjectAlias!]
+  subscriptions: [ProjectSubscription!]
+}
+
+type RepoEventsSettings{
```
sophstad, line 1155: Same thought as above

sophstad (approved):
Yay this looks great!! Minimally invasive solution for what I feared would be very complicated 🎉 awesome work Chaya

graphql/schema.graphql
```diff
@@ -1144,6 +1144,22 @@ type ProjectSettings {
   subscriptions: [ProjectSubscription!]
 }
 
+type ProjectEventsSettings{
```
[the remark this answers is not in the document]
malikchaya2: good point, thanks!

PR #5526, opened 2022-03-28
EVG-16401 Lower permission requirements for copy projects API

malikchaya2:
[EVG-16401](https://jira.mongodb.org/browse/EVG-16401)

### Description 
This removes the createProject middleware which is what makes user require superUser access to use the copyProject route. We don't want that, we want anyone with admin access to be able to copy a project. 

### Testing 
none

### DIFF
--- a/rest/route/service.go
+++ b/rest/route/service.go
@@ -156,7 +156,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/projects/{project_id}/detach_from_repo").Version(2).Post().Wrap(requireUser, addProject, requireProjectAdmin, editProjectSettings).RouteHandler(makeDetachProjectFromRepoHandler())
 	app.AddRoute("/projects/{project_id}/repotracker").Version(2).Post().Wrap(requireUser, addProject).RouteHandler(makeRunRepotrackerForProject())
 	app.AddRoute("/projects/{project_id}").Version(2).Put().Wrap(createProject).RouteHandler(makePutProjectByID())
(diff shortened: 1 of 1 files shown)


Kimchelly (commented):
It seems like this change gives anyone permission to create a new project as long as they're admin of at least one project, since the user can do anything they want to the new cloned project. If we're allowing normal project admins to create any new projects now, do we even need `createProject` to require superuser permission?

Can we just make the `createProject` permission not require superuser?

malikchaya2:
> It seems like this change provides a loophole to give any project admin permission to create a new project as long as they're admin of at least one project, since the user can do anything they want to the new copied project. If we're allowing normal project admins to create any new projects now via copying, do we even need `createProject` to require superuser permission?
> 
> Is it okay to make project creation not require superuser? This seems like the actual change being proposed in this PR. In which case, it seems like we could reduce the `createProject` permission to be usable by any project admin.

cc @ablack12

ablack12:
Hmm that's an interesting point. The concept behind not allowing everybody to use the createProject permission was because we didn't want to go over our API limit because users who wouldn't consider that were creating too many projects. I think this is still a consideration so maybe we give this to Cloud for now to unblock them but we figure out how to add superuser permissions to Mana so that we can just grant this to the relevant cloud teams, which will handle updating when people join/leave, rather than us manually adding/removing users (although that would take some time). Let me sync with product / Geoff to see what to do here.

Kimchelly (approved):

malikchaya2:
We shouldn’t advertise this feature until we figure out how to limit tracking. 
https://jira.mongodb.org/browse/EVG-16634

PR #5529, opened 2022-03-29
EVG-16636: Fix periodic builds

sophstad:
[EVG-16636](https://jira.mongodb.org/browse/EVG-16636)

### Description 
- Validate periodic build definition (notably add ID if one is not provided)
- Fix bug where periodic builds were not accurately constructed in BuildFromService and subsequently not included in GraphQL responses

### Testing 
- Tested manually

### DIFF
--- a/rest/model/project.go
+++ b/rest/model/project.go
@@ -737,7 +737,7 @@ func (p *APIProjectRef) BuildFromService(v interface{}) error {
 			if err := periodicBuild.BuildFromService(pb); err != nil {
 				return err
 			}
-			periodicBuilds = append(periodicBuilds)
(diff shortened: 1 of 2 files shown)


rest/model/project.go
```diff
@@ -737,7 +737,7 @@ func (p *APIProjectRef) BuildFromService(v interface{}) error {
 			if err := periodicBuild.BuildFromService(pb); err != nil {
 				return err
 			}
-			periodicBuilds = append(periodicBuilds)
```
ablack12, line 740: oops good catch

ablack12 (approved):

PR #5530, opened 2022-03-30
EVG-16407 Update task_stats route to check last_run

bynn:
[EVG-16407](https://jira.mongodb.org/browse/EVG-16407)

### Description 
when task stats returns nothing, it now looks up when the last time they were generated and returns 503 if it has not been run in the last 24 hours 

### Testing 
https://evergreen-staging.corp.mongodb.com/rest/v2/projects/sandbox/task_stats?tasks=thistaskdoesn%27texistlol

image

### DIFF
--- a/rest/route/stats.go
+++ b/rest/route/stats.go
@@ -459,7 +459,7 @@ func (tsh *taskStatsHandler) Run(ctx context.Context) gimlet.Responder {
 		if err != nil {
 			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Failed to retrieve stats status"))
 		}
-		if statsStatus.ProcessedTasksUntil.After(tsh.filter.AfterDate) {
+		if statsStatus.ProcessedTasksUntil.Before(tsh.filter.AfterDate) {
 			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
 				Message:    "stats for this project have not been generated yet",
 				StatusCode: http.StatusServiceUnavailable,
--- a/rest/route/stats.go
+++ b/rest/route/stats.go
@@ -459,9 +459,9 @@ func (tsh *taskStatsHandler) Run(ctx context.Context) gimlet.Responder {
 		if err != nil {
 			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Failed to retrieve stats status"))
 		}
-		if time.Since(statsStatus.LastJobRun).Hours() > time.Duration.Hours(24) {
+		if statsStatus.ProcessedTasksUntil.After(tsh.filter.AfterDate) {
 			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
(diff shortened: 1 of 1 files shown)


rest/route/stats.go
```diff
@@ -453,6 +453,19 @@ func (tsh *taskStatsHandler) Run(ctx context.Context) gimlet.Responder {
 	if err != nil {
 		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Failed to retrieve the task stats"))
 	}
+	if len(taskStatsResult) == 0 {
+		// Lookup last sync date for project
+		statsStatus, err := stats.GetStatsStatus(tsh.filter.Project)
+		if err != nil {
+			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Failed to retrieve stats status"))
+		}
+		if time.Since(statsStatus.LastJobRun).Hours() > time.Duration.Hours(24) {
```
ablack12, line 462: It seems wrong that this should error if it hasn't been generated in the last 24 hours. What if the user is asking for task stats from a week ago? Those are generated and could be used without error. I think we could use ProcessedTasksUntil to just make sure that the time range the user is asking for is valid.

Kimchelly, line 462:
```go
if time.Since(statsStatus.LastJobRun) > 24 * time.Hours {
```

Kimchelly, line 462: It seems like this assumes that the user is asking for the latest stats and not a specific date range. How is it supposed to behave if they specify a date range in the past and it's not available?

Kimchelly (changes requested):

rest/route/stats.go
bynn: yes that makes more sense

bynn: removed

bynn: changed to checking the "afterdate" 
we always require before and after fields

Kimchelly (approved):
LGTM mod nit.

rest/route/stats.go
```diff
@@ -453,6 +453,19 @@ func (tsh *taskStatsHandler) Run(ctx context.Context) gimlet.Responder {
...
 	}
+	if len(taskStatsResult) == 0 {
+		// Lookup last sync date for project
+		statsStatus, err := stats.GetStatsStatus(tsh.filter.Project)
+		if err != nil {
+			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "Failed to retrieve stats status"))
+		}
+		if statsStatus.ProcessedTasksUntil.Before(tsh.filter.AfterDate) {
+			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+				Message:    "stats for this project have not been generated yet",
```
ablack12, line 464: Maybe "stats for this time range" since it's not the project itself that's the problem?

ablack12 (approved):

PR #5531, opened 2022-03-30
EVG-12806 skip assigning task to host if the AMI is outdated

ablack12:
[EVG-12806](https://jira.mongodb.org/browse/EVG-12806)

### Description 
When a distro has it's image updated, it's frustrating for engineers that new tasks they create may still be using the old image, if those hosts haven't been decommissioned yet. However, it's expensive to decommission all running hosts when AMI is updated. This change instead considers when a task has been created, and if it's been created since the AMI was updated, we wait for a new host (eventually old hosts will be decommissioned as idle when there are no more old tasks to run).

This may negatively impact new task wait times in these cases, however it would improve correctness. 

(Also, I chose to use events to determine if AMI has been updated because storing that in the Distro collection seemed a bit wrong, but I can still pivot to that)

### Testing 
Added unit tests and also tested in staging: verified that [this task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_test_db_patch_041c911277be79717a3b3842a5444e7119b9a391_6244a388b2373649f74aa7ea_22_03_30_18_38_07/logs?execution=0) (which was created after changing the AMI) [wasn't scheduled on hosts](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22ticket%22%3D%22EVG-12806%22%20message%3D%22skipping%20because%20AMI%20is%20outdated%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=1648664887&latest=1648665787&workload_pool=standard_perf&sid=1648666215.249727) that were using the old AM.

### DIFF
--- a/model/event/distro_event.go
+++ b/model/event/distro_event.go
@@ -12,9 +12,10 @@ const (
 	ResourceTypeDistro = "DISTRO"
 
 	// event types
-	EventDistroAdded    = "DISTRO_ADDED"
-	EventDistroModified = "DISTRO_MODIFIED"
-	EventDistroRemoved  = "DISTRO_REMOVED"
+	EventDistroAdded      = "DISTRO_ADDED"
+	EventDistroModified   = "DISTRO_MODIFIED"
+	EventDistroAMIModfied = "DISTRO_AMI_MODIFIED"
+	EventDistroRemoved    = "DISTRO_REMOVED"
 )
 
 // DistroEventData implements EventData.
@@ -55,4 +56,8 @@ func LogDistroModified(distroId, userId string, data interface{}) {
 // LogDistroRemoved should take in DistroData in order to preserve the ProviderSettingsList
 func LogDistroRemoved(distroId, userId string, data interface{}) {
 	LogDistroEvent(distroId, EventDistroRemoved, DistroEventData{UserId: userId, Data: data})
+}
+
+func LogDistroAMIModified(distroId, userId string) {
+	LogDistroEvent(distroId, EventDistroAMIModfied, DistroEventData{UserId: userId})
 }
--- a/model/event/distro_event.go
+++ b/model/event/distro_event.go
@@ -58,6 +58,7 @@ func LogDistroRemoved(distroId, userId string, data interface{}) {
 	LogDistroEvent(distroId, EventDistroRemoved, DistroEventData{UserId: userId, Data: data})
 }
 
+// LogDistroAMIModified logs when the default region's AMI is modified.
 func LogDistroAMIModified(distroId, userId string) {
 	LogDistroEvent(distroId, EventDistroAMIModfied, DistroEventData{UserId: userId})
 }
--- a/model/event/event_finder.go
+++ b/model/event/event_finder.go
@@ -39,6 +39,15 @@ func Find(coll string, query db.Q) ([]EventLogEntry, error) {
 	return events, nil
 }
 
+func FindOne(query db.Q) (*EventLogEntry, error) {
+	event := &EventLogEntry{}
+	err := db.FindOneQ(AllLogCollection, query, event)
+	if adb.ResultsNotFound(err) {
+		return nil, nil
+	}
+	return event, err
+}
+
 func FindPaginated(hostID, hostTag, coll string, limit, page int) ([]EventLogEntry, int, error) {
 	query := MostRecentHostEvents(hostID, hostTag, limit)
 	events := []EventLogEntry{}
@@ -159,11 +168,12 @@ func DistroEventsForId(id string) db.Q {
 	return db.Query(filter)
 }
 
+// DistroAMIModifiedForId returns the query for AMI modified distro events.
 func DistroAMIModifiedForId(id string) db.Q {
 	filter := ResourceTypeKeyIs(ResourceTypeDistro)
 	filter[ResourceIdKey] = id
 	filter[TypeKey] = EventDistroAMIModfied
-	return db.Query(filter).Sort([]string{"-" + TimestampKey}).Limit(1)
+	return db.Query(filter).Sort([]string{"-" + TimestampKey})
 }
 
 func MostRecentDistroEvents(id string, n int) db.Q {
--- a/model/event/event_finder.go
+++ b/model/event/event_finder.go
@@ -39,24 +39,6 @@ func Find(coll string, query db.Q) ([]EventLogEntry, error) {
 	return events, nil
 }
 
-func FindOne(query db.Q) (*EventLogEntry, error) {
-	event := &EventLogEntry{}
-	err := db.FindOneQ(AllLogCollection, query, event)
-	if adb.ResultsNotFound(err) {
-		return nil, nil
-	}
-	return event, err
-}
-
-func Aggregate(pipeline []bson.M) ([]EventLogEntry, error) {
-	events := []EventLogEntry{}
-	err := db.Aggregate(AllLogCollection, pipeline, &events)
-	if err != nil {
-		return nil, err
-	}
-	return events, nil
-}
-
 func FindPaginated(hostID, hostTag, coll string, limit, page int) ([]EventLogEntry, int, error) {
 	query := MostRecentHostEvents(hostID, hostTag, limit)
 	events := []EventLogEntry{}
@@ -173,20 +155,26 @@ func TaskEventsInOrder(id string) db.Q {
 
 // FindLatestPrimaryDistroEvents return the most recent non-AMI events for the distro.
 func FindLatestPrimaryDistroEvents(id string, n int) ([]EventLogEntry, error) {
-	return Aggregate(latestDistroEventsPipeline(id, n, false))
+	events := []EventLogEntry{}
+	err := db.Aggregate(AllLogCollection, latestDistroEventsPipeline(id, n, false), &events)
+	if err != nil {
+		return nil, err
+	}
+	return events, err
 }
 
 // FindLatestAMIModifiedDistroEvent returns the most recent AMI event. Returns an empty struct if nothing exists.
 func FindLatestAMIModifiedDistroEvent(id string) (EventLogEntry, error) {
-	event := EventLogEntry{}
-	events, err := Aggregate(latestDistroEventsPipeline(id, 1, true))
+	events := []EventLogEntry{}
+	res := EventLogEntry{}
+	err := db.Aggregate(AllLogCollection, latestDistroEventsPipeline(id, 1, true), &events)
 	if err != nil {
-		return event, err
+		return res, err
 	}
 	if len(events) > 0 {
-		event = events[0]
+		res = events[0]
 	}
-	return event, nil
+	return res, nil
 }
 
 func latestDistroEventsPipeline(id string, n int, amiOnly bool) []bson.M {
--- a/model/event/event_finder.go
+++ b/model/event/event_finder.go
@@ -92,16 +92,6 @@ func FindByID(eventID string) (*EventLogEntry, error) {
 	return &e, nil
 }
 
-// FindEventsByIDs finds all events that match the given event IDs.
-func FindEventsByIDs(events []string) ([]EventLogEntry, error) {
-	query := bson.M{
-		idKey: bson.M{
-			"$in": events,
-		},
-	}
-	return Find(AllLogCollection, db.Query(query))
-}
-
 func FindLastProcessedEvent() (*EventLogEntry, error) {
 	q := db.Query(bson.M{
 		processedAtKey: bson.M{
@@ -167,6 +157,13 @@ func DistroEventsForId(id string) db.Q {
 	filter[ResourceIdKey] = id
 
 	return db.Query(filter)
+}
+
+func DistroAMIModifiedForId(id string) db.Q {
+	filter := ResourceTypeKeyIs(ResourceTypeDistro)
+	filter[ResourceIdKey] = id
+	filter[TypeKey] = EventDistroAMIModfied
+	return db.Query(filter).Sort([]string{"-" + TimestampKey}).Limit(1)
 }
 
 func MostRecentDistroEvents(id string, n int) db.Q {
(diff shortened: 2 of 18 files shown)


model/event/event_finder.go
```diff
@@ -169,6 +159,13 @@ func DistroEventsForId(id string) db.Q {
 	return db.Query(filter)
 }
 
+func DistroAMIModifiedForId(id string) db.Q {
```
Kimchelly, line 162: Documentation for exported methods.

model/event/event_finder.go
```diff
@@ -169,6 +159,13 @@ func DistroEventsForId(id string) db.Q {
...
+	filter := ResourceTypeKeyIs(ResourceTypeDistro)
+	filter[ResourceIdKey] = id
+	filter[TypeKey] = EventDistroAMIModfied
+	return db.Query(filter).Sort([]string{"-" + TimestampKey}).Limit(1)
```
Kimchelly, line 166: Does this need an index? Or are distro events TTL'd?

model/task_queue_service_dependency.go
```diff
@@ -348,6 +348,20 @@ func (d *basicCachedDAGDispatcherImpl) FindNextTask(spec TaskSpec) *TaskQueueIte
...
+			// AMI Updated time is only provided if the host is running with an outdated AMI.
+			// If the task was created after the time that the AMI was updated, then we should wait for an updated host.
+			if !utility.IsZeroTime(amiUpdatedTime) && nextTaskFromDB.IngestTime.After(amiUpdatedTime) {
+				grip.Debug(message.Fields{
+					"dispatcher":       DAGDispatcher,
+					"function":         "FindNextTask",
+					"message":          "skipping because AMI is outdated",
+					"task_id":          nextTaskFromDB.Id,
+					"distro_id":        d.distroID,
+					"ami_updated_time": amiUpdatedTime,
```
Kimchelly, line 361: Can this also log the task ingestion time?

model/host/host.go
```diff
@@ -407,6 +407,14 @@ func (h *Host) IdleTime() time.Duration {
 	return time.Since(h.CreationTime)
 }
 
+func (h *Host) GetAMI() string {
+	if len(h.Distro.ProviderSettingsList) == 0 {
+		return ""
+	}
+	ami, _ := h.Distro.ProviderSettingsList[0].Lookup("ami").StringValueOK()
```
Kimchelly, line 414: I don't really get how the `ProviderSettingsList` works conceptually. How do we know that it must be the first in the list? Is it because hosts are always started in us-east-1?

model/event/distro_event.go
```diff
@@ -56,3 +57,7 @@ func LogDistroModified(distroId, userId string, data interface{}) {
 func LogDistroRemoved(distroId, userId string, data interface{}) {
 	LogDistroEvent(distroId, EventDistroRemoved, DistroEventData{UserId: userId, Data: data})
 }
+
+func LogDistroAMIModified(distroId, userId string) {
```
Kimchelly, line 61: Documentation for exported methods.

service/api_task.go
```diff
@@ -599,6 +599,19 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
 	})
 	stepStart = time.Now()
 
+	var amiUpdatedTime time.Time
+	if d.GetDefaultAMI() != currentHost.GetAMI() {
+		events, err := event.Find(event.AllLogCollection, event.DistroAMIModifiedForId(d.Id))
```
Kimchelly, line 604: Instead of doing a find all, should this just do a find one since it only cares about the first result anyways?

model/task_queue.go
```diff
@@ -118,6 +118,7 @@ type TaskQueueItem struct {
 	ExpectedDuration    time.Duration `bson:"exp_dur" json:"exp_dur"`
 	Priority            int64         `bson:"priority" json:"priority"`
 	Dependencies        []string      `bson:"dependencies" json:"dependencies"`
+	CreateTime          time.Time     `bson:"create_time" json:"create_time"`
```
Kimchelly, line 121: Maybe I'm blind and don't see where it's used, but what is this key for?

model/task_queue_service_dependency.go
```diff
@@ -348,6 +348,20 @@ func (d *basicCachedDAGDispatcherImpl) FindNextTask(spec TaskSpec) *TaskQueueIte
 			if !dependenciesMet {
 				continue
 			}
+
+			// AMI Updated time is only provided if the host is running with an outdated AMI.
+			// If the task was created after the time that the AMI was updated, then we should wait for an updated host.
+			if !utility.IsZeroTime(amiUpdatedTime) && nextTaskFromDB.IngestTime.After(amiUpdatedTime) {
```
Kimchelly, line 354: Since this is based on ingest time, are restarted tasks expected to use the latest AMI or is it okay for them to use an older one?

Kimchelly (changes requested):
It seems like this change will cause the host allocator to under-allocate because it doesn't account for when tasks were created when deciding how many new hosts it needs. It might also cause more turnover in the distro host pools. Is there a way to track how much extra time tasks spend waiting for a host with a newer AMI?

model/event/event_finder.go
ablack12: Good question; an index already exists for resource ID / resource type / timestamp but not including event type, which I believe would mean that the timestamp portion of the index wouldn't be used. I think a new index would be about 100 GB. That being said, there aren't very many distro events relatively, so I'm not sure it's worth an additional index when resource/resource type would narrow it down a lot. What do you think?

model/task_queue.go
ablack12: ah this is leftover from a different idea for how to implement

service/api_task.go
ablack12: I thought about that but we don't have a FindOne for events right now and I figured adding that should mean refactoring existing use cases. I can just add it to this though.

model/host/host.go
ablack12: For distros, providerSettingsList can have multiple items because there are definitions for multiple regions, however within the host document we only save the region that is relevant to that host, so it's guaranteed to have only one item.

model/task_queue_service_dependency.go
ablack12: Yeah good question, restarted tasks would use an older one. I couldn't quite decide which was better, scheduled time or ingest time. I was thinking that scheduled tasks are a superset of ingest tasks, so more tasks would be affected by this logic, whereas ingest time requires that it be a new item. I think it might be good to limit what tasks hit this constraint given the cons but it's not our typical way of handling tasks.

ablack12 (commented):
> It seems like this change will cause the host allocator to under-allocate because it doesn't account for when tasks were created when deciding how many new hosts it needs. It might also cause more turnover in the distro host pools. Is there a way to track how much extra time tasks spend waiting for a host with a newer AMI?

Yeah it's a reasonable thought. I'm not really sure how we would track how much extra time tasks spend due to a new AMI, without storing a lot more information. I think we could use the  timestamps of "skipping because AMI is outdated" to see how long that time period affects each task? Can discuss that splunk query with Brian.

ablack12:
I also just added a commit to terminate outdated idle hosts more aggressively (couldn't figure out how to not verbatim repeat the commit message there 😢 ) which should help in the case that all tasks are waiting for new hosts somehow.

model/event/event_finder.go
Kimchelly: Could we use a partial index to include just the events with distro AMI modified?

I don't think it's that big of a performance concern for now. If it becomes an issue, it would show up in the slow queries eventually.

Kimchelly (approved):
LGTM mod the one leftover comment about documentation.

I think a Splunk query just to verify that post-deploy tasks aren't waiting a significantly longer time than usual would be fine.

ablack12:
[This Splunk query](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22skipping%20because%20AMI%20is%20outdated%22%20%7C%20timechart%20distinct_count(task_id)%20span%3D10&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-7d%40h&latest=now&workload_pool=standard_perf&display.page.search.tab=visualizations&display.general.type=visualizations&sid=1648840883.453725) will tell us how many distinct tasks are getting hit by needing to wait for AMI. We do also have this [task start wait time dashboard](https://mongodb.splunkcloud.com/en-US/app/search/unblocked_task_metrics?form.unused=aws&form.timeselector.earliest=-60m%40m&form.timeselector.latest=now) but it may be hard to interpret for something this specific.

model/event/event_finder.go
```diff
@@ -39,6 +39,24 @@ func Find(coll string, query db.Q) ([]EventLogEntry, error) {
...
+func FindOne(query db.Q) (*EventLogEntry, error) {
+	event := &EventLogEntry{}
+	err := db.FindOneQ(AllLogCollection, query, event)
+	if adb.ResultsNotFound(err) {
+		return nil, nil
+	}
+	return event, err
+}
+
+func Aggregate(pipeline []bson.M) ([]EventLogEntry, error) {
```
malikchaya2, line 51: do we want to take in an  "out" or "result" interface to be more consistent with other Aggregate functions we have?

model/task_queue.go
```diff
@@ -144,21 +144,7 @@ var (
...
 	taskQueueInfoPlanCreatedAtKey = bsonutil.MustHaveTag(DistroQueueInfo{}, "PlanCreatedAt")
-	// taskQueueInfoCountDurationOverThresholdKey = bsonutil.MustHaveTag(DistroQueueInfo{}, "CountDurationOverThreshold")
-	// taskQueueInfoCountWaitOverThresholdKey = bsonutil.MustHaveTag(DistroQueueInfo{}, "CountWaitOverThreshold")
-	// taskQueueInfoTaskGroupInfosKey     = bsonutil.MustHaveTag(DistroQueueInfo{}, "TaskGroupInfos")
-	// taskQueueInfoAliasQueueKey         = bsonutil.MustHaveTag(DistroQueueInfo{}, "AliasQueue")
-
-	// taskQueueInfoGroupNameKey                  = bsonutil.MustHaveTag(TaskGroupInfo{}, "Name")
-	// taskQueueInfoGroupCountKey                 = bsonutil.MustHaveTag(TaskGroupInfo{}, "Count")
-	// taskQueueInfoGroupMaxHostsKey              = bsonutil.MustHaveTag(TaskGroupInfo{}, "MaxHosts")
-	// taskQueueInfoGroupExpectedDuratioKey       = bsonutil.MustHaveTag(TaskGroupInfo{}, "ExpectedDuration")
```
malikchaya2, line 159: haha what happened here?

units/host_monitoring_idle_termination.go
```diff
@@ -25,7 +25,9 @@ const (
 	idleHostJobName = "idle-host-termination"
 
 	// idleTimeCutoff is the amount of time we wait for an idle host to be marked as idle.
-	idleTimeCutoff            = 4 * time.Minute
+	idleTimeCutoff = 4 * time.Minute
+	// outdatedIdleTimeCutoff is the amount of time we wait for an outdated idle host to be marked idle.
+	outdatedIdleTimeCutoff    = 2 * time.Minute
```
malikchaya2, line 30: Why not make this even lower?

model/event/event_finder.go
```diff
@@ -161,29 +170,41 @@ func TaskEventsInOrder(id string) db.Q {
...
-	filter := ResourceTypeKeyIs(ResourceTypeDistro)
-	filter[ResourceIdKey] = id
-	filter[TypeKey] = EventDistroAMIModfied
-	return db.Query(filter).Sort([]string{"-" + TimestampKey})
-}
-
-func MostRecentDistroEvents(id string, n int) db.Q {
-	return DistroPrimaryEventsForId(id).Sort([]string{"-" + TimestampKey}).Limit(n)
+// FindLatestAMIModifiedDistroEvent returns the most recent AMI event. Returns an empty struct if nothing exists.
+func FindLatestAMIModifiedDistroEvent(id string) (EventLogEntry, error) {
```
Kimchelly, line 180: nit: this could just return a pointer if we don't want to return empty structs. I don't care that much, it just makes it consistent with other `FindOne`-style queries.

units/host_monitoring_idle_termination.go
ablack12: The host could run older tasks so I do think it's okay to wait a bit for an older task. That being said I halved it arbitrarily; I'd be fine making it a minute or something.

model/task_queue.go
ablack12: Sam 😂

model/event/event_finder.go
Kimchelly: It's generally less preferred to pass C-style in-out parameters if you can just pass back a typed value. The only reason some things accept in-out parameters like `db.Aggregate` is because it can't know the type that's being aggregated, but in this case, we know that it must be `[]EventLogEntry`.

Kimchelly (approved):
LGTM mod nit.

model/event/event_finder.go
ablack12: I guess we can just call db.Aggregate directly in the functions

model/task_queue.go
Kimchelly: I'm pretty sure these were just leftover BSON keys from the scheduler project that were left in just in case someone might use them one day (insce we usually enumerate every single BSON key in the DB model, even if we don't use it).

But since we're not using them and they're not exported, there's not much reason to keep them, and we can always re-introduce them if we ever need them.

model/event/event_finder.go
ablack12: I'm kind of okay with this since then we don't need to make a nil check for latest timestamp.

units/host_monitoring_idle_termination.go
malikchaya2: Ah. One minute feels better to me because the idea of keeping a host around that's not currently running something that can only run a subset of all our tasks feels less than ideal. But I'm easily swayed, either way is fine with me.

malikchaya2 (approved):

PR #5533, opened 2022-03-30
EVG-15622 Begin the time_taken calculation at activated_time

hadjri:
[EVG-15622](https://jira.mongodb.org/browse/EVG-15622)

### Description 
It was noticed that the time_taken_ms was not accurate for builds that had been restarted. For example if a build runs, and it is then restarted the next day and runs for 1 hour, the time taken field will say 25 hours since it is doing an (end time - start time) calculation.  

In the ticket it was proposed to subtract the time between activated and started time from the result, however in this PR I've set activated time as the new anchor for calculating the duration.  This is because it seems to me that activation time will always be after the start time, and (end - start - (activation - start)) simplifies to (end - activation). However if there's some context I'm missing here I'm happy to add some more logic.
### Testing 
Updated existing unit test.


model/build/build.go
```diff
@@ -252,7 +252,7 @@ func TryMarkStarted(buildId string, startTime time.Time) error {
 func (b *Build) MarkFinished(status string, finishTime time.Time) error {
 	b.Status = status
 	b.FinishTime = finishTime
-	b.TimeTaken = finishTime.Sub(b.StartTime)
+	b.TimeTaken = finishTime.Sub(b.ActivatedTime)
```
ablack12, line 255: Sorry for the late CR -- the issue here is that the task can be activated before it actually is started, since it has to wait for a host, and that could be a while.

ablack12 (changes requested):

model/build/build.go
hadjri: In these cases where we perform restarts to be completely accurate we'd need the start time of the most recent run which I don't think is readily available and might need some more doing - the change to begin with the activation time in this case though would still definitely be an improvement over what is currently happening.

model/task/task.go
```diff
@@ -1572,7 +1572,11 @@ func (t *Task) MarkEnd(finishTime time.Time, detail *apimodels.TaskEndDetail) er
 		}
```
ablack12, line 1572: This logic actually doesn't make much sense; it should probably be "set it to the activated time or etc" if TimeTaken is going to be accurate

ablack12 (commented):
Actually my earlier comment may have been wrong sorry; I didn't realize this was only calculated in MarkFinished and MarkEnd, i.e. StartTime is going to need to be populated. I'm not sure how careful we are at keeping Start updated but it does seem good enough then to just do finishTime.Sub(b.StartTime).

ablack12 (approved):

PR #5534, opened 2022-03-31
EVG-16645 shield implicit subscriptions

ybrill:
[EVG-16645](https://jira.mongodb.org/browse/EVG-16645)

### Description 
If a user clicks the big red button we should shield their implicit (patch finish, build break, etc.) subscriptions from getting deleted.
Screen Shot 2022-03-31 at 12 02 32 PM
The old UI's corresponding button didn't clear them because [they were filtered from the list of IDs](https://github.com/evergreen-ci/evergreen/blob/88d45f4e8d0de025d89c9d75683a2d99cde96d2d/public/static/js/notifications.js#L28-L39) sent to the application.

### Testing 
Added unit tests. 
Verified on staging that before these changes implicit subscriptions were removed and with the changes they aren't removed.

### DIFF
--- a/model/user/user.go
+++ b/model/user/user.go
@@ -419,7 +419,8 @@ func (u *DBUser) DeleteRoles(roles []string) error {
 	return nil
 }
 
-func (u *DBUser) ImplicitSubscriptionIDs() []string {
+// GeneralSubscriptionIDs returns a slice of the ids of the user's general subscriptions.
+func (u *DBUser) GeneralSubscriptionIDs() []string {
 	var ids []string
 	if id := u.Settings.Notifications.BuildBreakID; id != "" {
 		ids = append(ids, id)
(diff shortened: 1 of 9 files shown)


khelif96 (approved):

model/user/user.go
```diff
@@ -419,6 +419,30 @@ func (u *DBUser) DeleteRoles(roles []string) error {
 	return nil
 }
 
+func (u *DBUser) ImplicitSubscriptionIDs() []string {
```
ablack12, line 422: Can we add a function comment to this?

ablack12: Idk if I agree with the wording of "implicit" since users _do_ explicitly subscribe to these. It seems like they're more like General subscriptions or something.

ablack12 (approved):

PR #5535, opened 2022-03-31
EVG-16570 Move tests from rest/data to model if necessary

hadjri:
[EVG-16570](https://jira.mongodb.org/browse/EVG-16570)

### Description 
There are many tests not in the rest/data package which do not use any functions that live in rest/data.  I have moved them to packages that make more sense (typically /model)
### Testing 
Tests have not (hopefully) broken from moving around


malikchaya2 (approved):
It looks like the tests are broken, but LGTM once they are fixed.

PR #5536, opened 2022-03-31
EVG-16559 Refactor distro DB functions to fit convention

hadjri:
[EVG-16559](https://jira.mongodb.org/browse/EVG-16559)

### Description 
distro.FindOne works differently from other model FindOne functions in that it has no check for if adb.ResultsNotFound(err) in it, so a nil distro would also cause an error to return, and it returns a Distro object instead of a pointer which makes nil checks more difficult.

Changed to 1) no longer treat a nil distro as an error, and  2) return a distro pointer in distro DB functions.

### Testing 
Tests remain passing

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2131,7 +2131,7 @@ func getNumNewParentsAndHostsToSpawn(pool *evergreen.ContainerPool, newContainer
 	}
 
 	if !ignoreMaxHosts { // only want to spawn amount of parents allowed based on pool size
-		if numNewParentsToSpawn, err = parentCapacity(parentDistro, numNewParentsToSpawn, len(existingParents), pool); err != nil {
+		if numNewParentsToSpawn, err = parentCapacity(*parentDistro, numNewParentsToSpawn, len(existingParents), pool); err != nil {
 			return 0, 0, errors.Wrap(err, "could not calculate number of parents needed to spawn")
 		}
 	}
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2129,6 +2129,9 @@ func getNumNewParentsAndHostsToSpawn(pool *evergreen.ContainerPool, newContainer
 	if err != nil {
 		return 0, 0, errors.Wrap(err, "error find parent distro")
 	}
+	if parentDistro == nil {
+		return 0, 0, errors.Errorf("distro '%s' not found", pool.Distro)
+	}
 
 	if !ignoreMaxHosts { // only want to spawn amount of parents allowed based on pool size
 		if numNewParentsToSpawn, err = parentCapacity(*parentDistro, numNewParentsToSpawn, len(existingParents), pool); err != nil {
--- a/model/task_queue_service.go
+++ b/model/task_queue_service.go
@@ -85,12 +85,13 @@ func (s *taskDispatchService) Refresh(distroID string) error {
 }
 
 func (s *taskDispatchService) ensureQueue(distroID string) (CachedDispatcher, error) {
-	d, err := distro.FindOneId(distroID)
+	d := distro.Distro{}
+	foundDistro, err := distro.FindOneId(distroID)
 	if err != nil {
 		return nil, errors.Wrapf(err, "database error for find() by distro id '%s'", distroID)
 	}
-	if d == nil {
-		return nil, errors.Errorf("distro '%s' not found", distroID)
+	if foundDistro != nil {
+		d = *foundDistro
 	}
 	// If there is a "distro": *basicCachedDispatcherImpl in the cachedDispatchers map, return that.
 	// Otherwise, get the "distro"'s taskQueue from the database; seed its cachedDispatcher; put that in the map and return it.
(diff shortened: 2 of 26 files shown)


model/host/host.go
```diff
@@ -2131,7 +2131,7 @@ func getNumNewParentsAndHostsToSpawn(pool *evergreen.ContainerPool, newContainer
 	}
 
 	if !ignoreMaxHosts { // only want to spawn amount of parents allowed based on pool size
-		if numNewParentsToSpawn, err = parentCapacity(parentDistro, numNewParentsToSpawn, len(existingParents), pool); err != nil {
+		if numNewParentsToSpawn, err = parentCapacity(*parentDistro, numNewParentsToSpawn, len(existingParents), pool); err != nil {
```
ablack12, line 2134: Need to do a nil check before dereferencing the pointer

service/spawn.go
```diff
@@ -137,7 +137,7 @@ func (uis *UIServer) spawnPage(w http.ResponseWriter, r *http.Request) {
 		SetupScriptPath              string
 		NewUILink                    string
 		ViewData
-	}{spawnDistro, spawnTask, maxHosts, settings.Spawnhost.UnexpirableHostsPerUser, settings.Spawnhost.UnexpirableVolumesPerUser, settings.Providers.AWS.MaxVolumeSizePerUser,
+	}{*spawnDistro, spawnTask, maxHosts, settings.Spawnhost.UnexpirableHostsPerUser, settings.Spawnhost.UnexpirableVolumesPerUser, settings.Providers.AWS.MaxVolumeSizePerUser,
```
ablack12, line 140: need nil check before dereference (although it seems like this page shouldn't fail to load if distro doesn't exist)

ablack12 (changes requested):

service/spawn.go
hadjri: Changed it back to using a regular distro and using a dereferenced pointer if it is not nil.

model/task_queue_service.go
```diff
@@ -89,6 +89,9 @@ func (s *taskDispatchService) ensureQueue(distroID string) (CachedDispatcher, er
 	if err != nil {
 		return nil, errors.Wrapf(err, "database error for find() by distro id '%s'", distroID)
 	}
+	if d == nil {
+		return nil, errors.Errorf("distro '%s' not found", distroID)
+	}
 	// If there is a "distro": *basicCachedDispatcherImpl in the cachedDispatchers map, return that.
```
ablack12, line 95: This comment implies that it shouldn't be an error for distro not to exist, but rather that we'd just default to something else. Not super sure but might be worth keeping a non-pointer and just assigning *d to that if it's not nil

rest/route/distro.go
```diff
@@ -156,7 +156,7 @@ func (h *distroIDPutHandler) Parse(ctx context.Context, r *http.Request) error {
 // (b) creates a new resource based on the Request-URI and JSON payload
 func (h *distroIDPutHandler) Run(ctx context.Context) gimlet.Responder {
 	user := MustHaveUser(ctx)
-	original, err := distro.FindByID(h.distroID)
+	original, err := distro.FindOneId(h.distroID)
 	if err != nil {
```
ablack12, line 160: || original == nil

service/api_task.go
```diff
@@ -578,7 +578,7 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
 	}
 
 	d, err := distro.FindOneId(currentHost.Distro.Id)
-	if err != nil {
+	if err != nil || d == nil {
 		// Should we bailout if there is a database error leaving us unsure if the distro document actually exists?
 		m := "database error while retrieving distro document;"
 		if adb.ResultsNotFound(err) {
```
ablack12, line 584: This is the d == nil case now, since we return nil as the error in this case.

units/host_monitoring_parent_decommission.go
```diff
@@ -56,12 +57,18 @@ func (j *parentDecommissionJob) Run(ctx context.Context) {
...
 	if err != nil {
 		j.AddError(err)
 		return
 	}
-	minHosts := parentDistro.HostAllocatorSettings.MinimumHosts
+	if parentDistro == nil {
+		j.AddError(errors.Errorf("distro '%s' not found", j.DistroId))
+	}
+	minHosts := 0
+	if parentDistro != nil {
```
ablack12, line 69: Can we make this an else with 65?

ablack12 (approved):
LGTM with some small changes

ablack12:
I have some memory of this having unintended changes so can you just click through staging / run a task and make sure nothing crazy happens?

rest/route/distro.go
hadjri: There is already casing for the original being nil on line 196

hadjri: I don't think this route wants to treat a nil distro as an error

PR #5540, opened 2022-03-31
EVG-6274 Update github PR patches to support ignore options

bynn:
[EVG-6274](https://jira.mongodb.org/browse/EVG-6274)

### Description 
goes through all changed files and doesnt create patches when all the changes are in the ignore for github prs 
still works for normal patches 

### Testing 
https://github.com/evergreen-ci/commit-queue-sandbox/pull/346

image

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -267,6 +267,10 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 		j.gitHubError = ProjectFailsValidation
 		return errors.Wrapf(validationCatcher.Resolve(), "patched project config has errors")
 	}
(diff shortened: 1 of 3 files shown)


units/patch_intent.go
```diff
@@ -267,6 +267,10 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 		j.gitHubError = ProjectFailsValidation
 		return errors.Wrapf(validationCatcher.Resolve(), "patched project config has errors")
 	}
+	// Don't create patches for github PRs if the only changes are in ignore files.
```
ablack12, line 270: nit: can we say "ignored" instead rather than using the field name?

ablack12 (approved):
Love the cleanup!

PR #5547, opened 2022-04-05
EVG-16624 Fix ec2 role test

bynn:
[EVG-16624](https://jira.mongodb.org/browse/EVG-16624)

### Description 
test fails because the map is not ordered 
test now specifies what we want each time 

### Testing 
test passes in both orders


malikchaya2 (approved):

agent/command/assume_ec2_role_test.go
```diff
@@ -50,6 +50,8 @@ func TestEc2AssumeRoleExecute(t *testing.T) {
 	c := &ec2AssumeRole{}
 	for testName, testCase := range map[string]func(ctx context.Context, t *testing.T, c *ec2AssumeRole, comm *client.Mock, logger client.LoggerProducer, conf *internal.TaskConfig){
 		"FailsWithNoARN": func(ctx context.Context, t *testing.T, c *ec2AssumeRole, comm *client.Mock, logger client.LoggerProducer, conf *internal.TaskConfig) {
+			c.RoleARN = ""
```
ablack12, line 53: If we just move defining c into t.Run, then we guarantee that it's always new. (We also can move anything that isn't in that section out of the for loop func() definition, since anything defined outside of the for loop is accessible without needing to pass it in)

ablack12 (changes requested):

agent/command/assume_ec2_role_test.go
bynn: done

PR #5549, opened 2022-04-05
EVG-15358: skip cloud host termination if host is externally terminated

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15358

### Description 
If the host is externally terminated, skip the cloud host status check and termination. This assumes that the host really will terminate on its own without our intervention.

* Skip cloud host termination if host is externally terminated.
* Slightly refactor host termination job options.

### Testing 
Updated unit tests.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -105,7 +105,10 @@ func ModifyHostStatus(ctx context.Context, env evergreen.Environment, queue ambo
 			return "", http.StatusInternalServerError, errors.New(HostTerminationQueueingError)
 		}
 
-		if err := amboy.EnqueueUniqueJob(ctx, queue, units.NewHostTerminationJob(env, h, true, fmt.Sprintf("terminated by %s", u.Username()))); err != nil {
+		if err := amboy.EnqueueUniqueJob(ctx, queue, units.NewHostTerminationJob(env, h, units.HostTerminationOptions{
+			TerminateIfBusy:   true,
+			TerminationReason: fmt.Sprintf("terminated by %s", u.Username()),
+		})); err != nil {
 			return "", http.StatusInternalServerError, errors.New(HostTerminationQueueingError)
 		}
 		return fmt.Sprintf(HostTerminationQueueingSuccess, h.Id), http.StatusOK, nil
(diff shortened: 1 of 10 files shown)


ybrill:
You mentioned this assumes that the host really will terminate on its own without our intervention. 
Is it guaranteed once they give notice the instance will definitely be terminated?

Kimchelly:
It's kind of difficult to test for this, since their docs say that AWS will terminate the host but don't say whether they can change their mind about interrupting the host. I'm inclined to just believe what the docs say, which is that they will terminate the host. The original request in the ticket was to stop unnecessarily wasting AWS API calls, so it seems like the only way to reduce the API calls would be to believe that external termination really means the host will be taken away without us having to double-check using API calls.

If it's not the case, the [zombie host log](https://github.com/evergreen-ci/evergreen/blob/3f00dee2a0af3afd1d74de384be794dc815ae344/service/api_task.go#L502-L525) will catch it in Splunk. And if it does log for externally terminated hosts, we can conclude that external termination isn't guaranteed. In which case, there's no chance for optimization and we'll always have to check/terminate externally-terminated hosts (and simply suppress the "host is already terminated" error).

ybrill:
Got it. That sounds reasonable.

A second question: if we skip terminating the host when we get the interruption warning is it worthwhile to schedule a termination job when we get the interruption message? We'll terminate it anyway [when we get the termination message](https://github.com/evergreen-ci/evergreen/blob/3f00dee2a0af3afd1d74de384be794dc815ae344/rest/route/sns.go#L284).

In other words, if we're going to terminate the host when we get the interruption warning then it makes sense to enqueue the termination job: let's (potentially) save the money for the two minutes. But if we aren't going to terminate it, why do we need to react to the interruption warning?

Kimchelly:
I don't think we need to handle interruptions specially if we're assuming that external termination is supposed to deal with the termination for us. I removed the handler for instance interruption warnings.

ybrill:
I'm sorry to keep going back and forth. 
It occurred to me now that [the logging here](https://github.com/evergreen-ci/evergreen/blob/3f00dee2a0af3afd1d74de384be794dc815ae344/rest/route/sns.go#L225-L231) has been useful for monitoring spot interruption rates by distro/instance type. Would it work to have the interruption case but just have it do logging?

Kimchelly:
Changed it to just log the instance interruption message.

PR #5550, opened 2022-04-05
EVG-16302 Deprecate unused graphql schema fields.

khelif96:
[EVG-16302](https://jira.mongodb.org/browse/EVG-16302)

### Description 
Removes most of the fields that had the deprecated directive and are no longer used. 

### Testing 
Spruce patch with evergreen changes as a module.
https://spruce.mongodb.com/version/624cb4691e2d1702fdc5ed34/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


graphql/schema.graphql
```diff
@@ -871,7 +865,7 @@ type Patch {
   duration: PatchDuration
   time: PatchTime
   taskCount: Int
-  baseVersionID: String
+  baseVersionID: String @deprecated(reason: "Use versionFull.baseVersion.id instead")
```
minnakt, line 868: What's the reason we can't remove this one at the moment? :o

minnakt (approved):
It looks good to me...! I just had a question 🤓

graphql/schema.graphql
khelif96: it is being used here https://github.com/evergreen-ci/spruce/blob/ae1f4d36bb82735479366507f625761386dd46e2/src/gql/queries/get-patch.graphql#L7
I tried refactoring it but it ended up needing more work than i was willing to commit. (Lots of prop drilling).

I opened a follow up ticket to update it. https://jira.mongodb.org/browse/EVG-16674

I wanted to ensure this pr was backwards compatible. So this can be merged and deployed without blocking the front end pr and vice versa.

PR #5553, opened 2022-04-06
EVG-16554 Update find host by ip to only return running hosts

bynn:
[EVG-16554](https://jira.mongodb.org/browse/EVG-16554)

### Description 
change find host by ip query to only find running hosts 

### Testing 
unit test

finding terminated hosts returns nothing 

image

### DIFF
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -588,7 +588,7 @@ func ById(id string) db.Q {
 }
 
 // ByIP produces a query that returns a running host with the given ip address.
-func ByIP(ip string) db.Q {
+func ByIPAndRunning(ip string) db.Q {
 	return db.Query(bson.M{
 		"$or": []bson.M{
 			{IPKey: ip},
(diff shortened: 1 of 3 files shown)


model/host/db.go
```diff
@@ -587,13 +587,14 @@ func ById(id string) db.Q {
 	return db.Query(bson.D{{Key: IdKey, Value: id}})
 }
 
-// ByIP produces a query that returns a host with the given ip address.
+// ByIP produces a query that returns a running host with the given ip address.
```
ybrill, line 590: should the function name also reflect that it only looks for running hosts?

model/host/db.go
```diff
@@ -587,13 +587,14 @@ func ById(id string) db.Q {
 	return db.Query(bson.D{{Key: IdKey, Value: id}})
 }
 
-// ByIP produces a query that returns a host with the given ip address.
-func ByIP(ip string) db.Q {
+// ByIP produces a query that returns a running host with the given ip address.
```
ybrill, line 590: don't forget the function name in its comment!

bynn: ahh sorry! done

ybrill (approved):
🎉

PR #5554, opened 2022-04-06
EVG-14586: use version aborted field to determine status

SupaJoon:
[EVG-14586](https://jira.mongodb.org/browse/EVG-14586)

### Description 
These code changes help us avoid querying all tasks associated with a version to determine the version status

Merging is blocked on: https://jira.mongodb.org/browse/EVG-16731


rest/model/version.go
```diff
@@ -30,6 +30,7 @@ type APIVersion struct {
 	Requester          *string        `json:"requester"`
 	Errors             []*string      `json:"errors"`
 	Activated          *bool          `json:"activated"`
+	Aborted            *bool          `json:"aborted`
```
bynn, line 33:
```suggestion
	Aborted            *bool          `json:"aborted"
```

graphql/resolvers.go
```diff
@@ -4104,24 +4093,12 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
 			patchStatuses := []string{*p.Status}
 			for _, cp := range p.ChildPatches {
 				patchStatuses = append(patchStatuses, *cp.Status)
-				// add the child patch tasks to tasks so that we can consider their status
```
malikchaya2, line 4107: Can you please explain the motivation behind this change a bit more?

malikchaya2 (changes requested):

graphql/resolvers.go
SupaJoon: We want to fetch version status for up to 100 versions at a time and querying for version tasks makes this expensive. A version is "aborted" when at least one task is "aborted" and none are "failed." These code changes from [EVG-15555](https://github.com/evergreen-ci/evergreen/pull/5479) create an aborted field for versions. The hope is we can examine the status field and aborted field of a version to determine it's status as opposed to looping through tasks. 

I think if the child patch tasks don't set the parent version's aborted field, we will have to update the GetCollectiveStatus  to consider the aborted field of the child patches version. Let me know what you think and if this sounds right!

SupaJoon: I changed the logic to fetch the version associated with the child patch to whether it has been aborted. Child patches won't set the aborted field on their parent patch.

SupaJoon:
These code changes rely on https://jira.mongodb.org/browse/EVG-16731. The aborted field won't be set if task failures exist.

graphql/resolvers.go
```diff
@@ -3951,25 +3940,21 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
 		if len(p.ChildPatches) > 0 {
 			patchStatuses := []string{*p.Status}
 			for _, cp := range p.ChildPatches {
-				patchStatuses = append(patchStatuses, *cp.Status)
-				// add the child patch tasks to tasks so that we can consider their status
-				childPatchTasks, _, err := task.GetTasksByVersion(*cp.Id, opts)
+				v, err := model.VersionFindOneId(*cp.Id)
```
malikchaya2, line 3943: Can you please test this on staging to make sure that an aborted child patch updates the parent status correctly?

SupaJoon: I saw in staging that when task is aborted, the version that it belongs to will be set to failed and the aborted field will be set to true. The aborted field of the patch containing the child tasks will not be set. This resolver will calculate the status of the parent task (the one containing the child patches) based on all of the child patches statuses.

SupaJoon: @malikchaya2 Could you help me verify this condition: https://github.com/evergreen-ci/evergreen/pull/5554/files#diff-71a3e44f5234030321d7134c8dc226478f2d9f6f7349af9125cf5d595aedcf68R1081

malikchaya2: replying on slack

SupaJoon:
resolver changes are taken care of in  this PR: https://github.com/evergreen-ci/evergreen/pull/5599/files#diff-a3ce50f20b746b196f29c6d414e685b0110c24b768bcf83756de8d842c98f6dcR3956

PR #5555, opened 2022-04-06
EVG-16073: Add an end point to get recent running versions of a task

hadjri:
[EVG-16073](https://jira.mongodb.org/browse/EVG-16073)

### Description 
New route was created allowing users to get recent versions of a task for a specific project. Allows for filtering on build variant, starting revision number, and limit on the number of results returned.
### Testing 
Added new unit tests for the DB function and rest route.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1062,7 +1062,7 @@ func CountProjectRefsWithIdentifier(identifier string) (int, error) {
 	return db.CountQ(ProjectRefCollection, byId(identifier))
 }
 
-// GetTasksWithOptions will find the last number of tasks (denoted by NumVersions) that have run on a given project.
+// GetTasksWithOptions will find the last number of tasks (denoted by NumVersions) that exist for a given project.
 // This function may also filter on tasks running on a specific build variant, or tasks that come after a specific revision order number.
 func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTasksOpts) ([]task.Task, error) {
 	projectId, err := GetIdForProject(projectName)
@@ -1072,9 +1072,11 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 	if opts.NumVersions <= 0 {
 		opts.NumVersions = defaultVersionLimit
 	}
+	finishedStatuses := append(evergreen.TaskFailureStatuses, evergreen.TaskSucceeded)
 	match := bson.M{
 		task.ProjectKey:     projectId,
 		task.DisplayNameKey: taskName,
+		task.StatusKey:      bson.M{"$in": finishedStatuses},
 	}
 	if opts.StartAt > 0 {
 		match[task.RevisionOrderNumberKey] = bson.M{"$lte": opts.StartAt}
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1062,7 +1062,8 @@ func CountProjectRefsWithIdentifier(identifier string) (int, error) {
 	return db.CountQ(ProjectRefCollection, byId(identifier))
 }
 
-// GetTasksWithOptions
+// GetTasksWithOptions will find the last number of tasks (denoted by NumVersions) that have run on a given project.
+// This function may also filter on tasks running on a specific build variant, or tasks that come after a specific revision order number.
 func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTasksOpts) ([]task.Task, error) {
 	projectId, err := GetIdForProject(projectName)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


model/project_ref.go
```diff
@@ -1062,6 +1062,37 @@ func CountProjectRefsWithIdentifier(identifier string) (int, error) {
 	return db.CountQ(ProjectRefCollection, byId(identifier))
 }
 
+// GetTasksWithOptions
```
malikchaya2, line 1065: Can we elaborate on this comment, explaining what the function does?

model/version.go
```diff
@@ -599,9 +599,6 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 	if err := db.Aggregate(VersionCollection, pipeline, &res); err != nil {
 		return nil, errors.Wrapf(err, "error aggregating versions and builds")
 	}
-	if len(res) == 0 {
```
malikchaya2, line 602: why was this removed?

malikchaya2 (changes requested):

model/version.go
malikchaya2: waiting on this question

bynn: i think because we would be returning the same thing if we return in this if statement or out of it there's no point in doing the check

malikchaya2: oh, true! thanks!

rest/data/project.go
```diff
@@ -106,6 +106,23 @@ func VerifyUniqueProject(name string) error {
 	return nil
 }
 
+// GetProjectTasksWithOptions
```
ablack12, line 109: need a comment here

rest/route/project.go
```diff
@@ -978,6 +978,63 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+	if len(b) > 0 {
+		if err := json.Unmarshal(b, &h.opts); err != nil {
+			return errors.Wrap(err, "error parsing request body")
+		}
+	}
+
+	if h.opts.NumVersions == 0 {
+		h.opts.NumVersions = defaultVersionLimit
+	}
+	if h.opts.NumVersions < 1 {
```
ablack12, line 1014: Preference for this to just error if h.opts.NumVersion < 0, and keep the default handling in GetProjectTasksWithOptions.

rest/route/project.go
```diff
@@ -978,6 +978,63 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+	return nil
+}
+
+func (h *getProjectTasksHandler) Run(ctx context.Context) gimlet.Responder {
+	versions, err := data.GetProjectTasksWithOptions(h.projectName, h.taskName, h.opts)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "error getting versions"))
+	}
+
+	resp, err := gimlet.NewBasicResponder(http.StatusOK, gimlet.JSON, versions)
```
ablack12, line 1030: You can just return gimlet.NewJSONResponse(versions)

model/project_ref.go
```diff
@@ -1062,6 +1062,38 @@ func CountProjectRefsWithIdentifier(identifier string) (int, error) {
 	return db.CountQ(ProjectRefCollection, byId(identifier))
 }
 
+// GetTasksWithOptions will find the last number of tasks (denoted by NumVersions) that have run on a given project.
+// This function may also filter on tasks running on a specific build variant, or tasks that come after a specific revision order number.
+func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTasksOpts) ([]task.Task, error) {
```
ablack12, line 1067: This doesn't specifically filter tasks that have _run_. A big reason cloud wants this is to filter out non-activated / scheduled tasks so that they don't have to paginate. We should also reflect this in the test.

ablack12 (changes requested):

rest/route/project.go
```diff
@@ -1007,13 +1007,12 @@ func (h *getProjectTasksHandler) Parse(ctx context.Context, r *http.Request) err
 			return errors.Wrap(err, "error parsing request body")
 		}
 	}
-
+	if h.opts.NumVersions < 0 {
```
ablack12, line 1010: Can this be renamed to limit? I think that may be more in line with our existing routes.

ablack12 (approved):
LGTM with comment and assuming this is tested in staging before merging (and added to wiki).

hadjri:
> LGTM with comment and assuming this is tested in staging before merging (and added to wiki).

Documentation: https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-tasks-for-a-project

Attached screenshot testing in staging with opts:
Screen Shot 2022-04-13 at 1 10 04 PM

hadjri:
> LGTM with comment and assuming this is tested in staging before merging (and added to wiki).

Documentation: https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-tasks-for-a-project

Attached screenshot testing in staging with opts:
Screen Shot 2022-04-13 at 1 10 04 PM

PR #5556, opened 2022-04-07
EVG-16396  Replace ResultStatus where safe  

malikchaya2:
[EVG-16396](https://jira.mongodb.org/browse/EVG-16396)

### Description 
This replaces ResultStatus with GetDisplayStatus wherever it seems safe to do so. I left it where it's used in the UI server and in [FilterTasksOnStatus](https://github.com/evergreen-ci/evergreen/blob/main/model/task/results.go#L124). I wasn't sure if it's safe to replace in FilterTasksOnStatus because I don't know how the [recentTasksGetHandler](https://github.com/evergreen-ci/evergreen/blob/main/rest/route/status.go#L80) is used and what statuses it can handle. If anyone has more context and knows that it is safe to replace, please let me know. 

### Testing 
Existing tests only

### DIFF
--- a/model/task/results.go
+++ b/model/task/results.go
@@ -45,17 +45,19 @@ func (t *Task) ResultStatus() string {
 // This type implements the grip/message.Composer interface and may be
 // passed directly to grip loggers.
 type ResultCounts struct {
-	Total              int `json:"total"`
-	Inactive           int `json:"inactive"`
-	Unstarted          int `json:"unstarted"`
-	Started            int `json:"started"`
-	Succeeded          int `json:"succeeded"`
-	Failed             int `json:"failed"`
-	SetupFailed        int `json:"setup-failed"`
-	SystemFailed       int `json:"system-failed"`
-	SystemUnresponsive int `json:"system-unresponsive"`
-	SystemTimedOut     int `json:"system-timed-out"`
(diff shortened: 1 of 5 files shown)


model/task/results.go
```diff
@@ -69,7 +71,7 @@ func GetResultCounts(tasks []Task) *ResultCounts {
 
 	for _, t := range tasks {
 		out.Total++
```
malikchaya2, line 73: GetDisplayStatus() does not sum up inactive, and since this is used in a rest route, I will leave it as ResultStatus in case someone relies on that value.

hadjri (approved):
LGTM with build fixes !

model/task_lifecycle_test.go
```diff
@@ -3484,7 +3484,7 @@ func TestDisplayTaskUpdates(t *testing.T) {
 	dbTask, err := task.FindOne(db.Query(task.ById(dt.Id)))
 	assert.NoError(err)
 	assert.NotNil(dbTask)
-	assert.Equal(evergreen.TaskFailed, dbTask.Status)
+	assert.Equal(evergreen.TaskSystemUnresponse, dbTask.Status)
```
khelif96, line 3487: Why did this change?

malikchaya2: Discussed and fixed.

PR #5558, opened 2022-04-07
EVG-16430 Add validation to container configurations

hadjri:
[EVG-16430](https://jira.mongodb.org/browse/EVG-16430)

### Description 
Validations for containers, container systems, container resources and task/BV run_on properties have been added to the project validator and in the modifyProject route. These validations reflect what was discussed in the design doc: https://docs.google.com/document/d/1JIGq81Sg4wgegxOiru2g7J7kPCpapz0saUxtUBhnUFk/edit?pli=1#
### Testing 
Added unit tests.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -1133,7 +1133,7 @@ const (
 // Validate checks that the container Windows version is recognized.
 func (w WindowsVersion) Validate() error {
 	switch w {
-	case Windows2022, Windows2019, Windows2016:
+	case Windows2022, Windows2019, Windows2016, "":
 		return nil
 	default:
 		return errors.Errorf("unrecognized windows version '%s'", w)
--- a/globals.go
+++ b/globals.go
@@ -1096,7 +1096,7 @@ const (
 // Validate checks that the container OS is recognized.
 func (c ContainerOS) Validate() error {
 	switch c {
-	case LinuxOS, WindowsOS:
+	case LinuxOS, WindowsOS, "":
 		return nil
 	default:
 		return errors.Errorf("unrecognized container OS '%s'", c)
@@ -1114,7 +1114,7 @@ const (
 // Validate checks that the container CPU architecture is recognized.
 func (c CPUArchitecture) Validate() error {
 	switch c {
-	case ArchARM64, ArchAMD64:
+	case ArchARM64, ArchAMD64, "":
 		return nil
 	default:
 		return errors.Errorf("unrecognized CPU architecture '%s'", c)
--- a/globals.go
+++ b/globals.go
@@ -1096,7 +1096,7 @@ const (
 // Validate checks that the container OS is recognized.
 func (c ContainerOS) Validate() error {
 	switch c {
-	case LinuxOS, WindowsOS, "":
+	case LinuxOS, WindowsOS:
 		return nil
 	default:
 		return errors.Errorf("unrecognized container OS '%s'", c)
@@ -1114,7 +1114,7 @@ const (
 // Validate checks that the container CPU architecture is recognized.
 func (c CPUArchitecture) Validate() error {
 	switch c {
-	case ArchARM64, ArchAMD64, "":
+	case ArchARM64, ArchAMD64:
 		return nil
 	default:
 		return errors.Errorf("unrecognized CPU architecture '%s'", c)
@@ -1133,7 +1133,7 @@ const (
 // Validate checks that the container Windows version is recognized.
 func (w WindowsVersion) Validate() error {
 	switch w {
-	case Windows2022, Windows2019, Windows2016, "":
+	case Windows2022, Windows2019, Windows2016:
 		return nil
 	default:
 		return errors.Errorf("unrecognized windows version '%s'", w)
--- a/model/project_configs.go
+++ b/model/project_configs.go
@@ -36,7 +36,7 @@ type ProjectConfigFields struct {
 	TaskSync               *TaskSyncOptions               `yaml:"task_sync,omitempty" bson:"task_sync,omitempty"`
 	GithubTriggerAliases   []string                       `yaml:"github_trigger_aliases,omitempty" bson:"github_trigger_aliases,omitempty"`
 	PeriodicBuilds         []PeriodicBuildDefinition      `yaml:"periodic_builds,omitempty" bson:"periodic_builds,omitempty"`
-	ContainerSizes         map[string]ContainerResources  `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
+	ContainerSizes         map[string]*ContainerResources `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
 }
 
 // Comment above is used by the linter to detect the end of the struct.
--- a/model/project_configs.go
+++ b/model/project_configs.go
@@ -36,7 +36,7 @@ type ProjectConfigFields struct {
 	TaskSync               *TaskSyncOptions               `yaml:"task_sync,omitempty" bson:"task_sync,omitempty"`
 	GithubTriggerAliases   []string                       `yaml:"github_trigger_aliases,omitempty" bson:"github_trigger_aliases,omitempty"`
 	PeriodicBuilds         []PeriodicBuildDefinition      `yaml:"periodic_builds,omitempty" bson:"periodic_builds,omitempty"`
-	ContainerSizes         map[string]*ContainerResources `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
+	ContainerSizes         map[string]ContainerResources  `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
 }
 
 // Comment above is used by the linter to detect the end of the struct.
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2441,9 +2441,13 @@ func ValidateContainers(pRef *ProjectRef, containers []Container) error {
 	catcher := grip.NewSimpleCatcher()
 	for _, container := range containers {
 		catcher.Add(container.System.Validate())
-		catcher.Add(container.Resources.Validate())
 		size, ok := pRef.ContainerSizes[container.Size]
-		catcher.AddWhen(ok, size.Validate())
+		if ok {
+			catcher.Add(size.Validate())
+		}
+		if container.Resources != nil {
+			catcher.Add(container.Resources.Validate())
+		}
 		catcher.ErrorfWhen(container.Size != "" && !ok, "size '%s' is not defined anywhere", container.Size)
 		catcher.NewWhen(container.Size != "" && container.Resources != nil, "size and resources cannot both be defined")
 		catcher.NewWhen(container.Size == "" && container.Resources == nil, "either size or resources must be defined")
@@ -2453,10 +2457,15 @@ func ValidateContainers(pRef *ProjectRef, containers []Container) error {
 	return catcher.Resolve()
 }
 
+// Validate that essential ContainerSystem fields are properly defined and no data contradictions exist.
 func (c ContainerSystem) Validate() error {
 	catcher := grip.NewSimpleCatcher()
-	catcher.Add(c.OperatingSystem.Validate())
-	catcher.Add(c.CPUArchitecture.Validate())
+	if c.OperatingSystem != "" {
+		catcher.Add(c.OperatingSystem.Validate())
+	}
+	if c.CPUArchitecture != "" {
+		catcher.Add(c.CPUArchitecture.Validate())
+	}
 	if c.OperatingSystem == evergreen.WindowsOS {
 		catcher.Add(c.WindowsVersion.Validate())
 	}
@@ -2464,6 +2473,7 @@ func (c ContainerSystem) Validate() error {
 	return catcher.Resolve()
 }
 
+// Validate that essential ContainerResources fields are properly defined.
 func (c ContainerResources) Validate() error {
 	catcher := grip.NewSimpleCatcher()
 	catcher.NewWhen(c.CPU <= 0, "container resource CPU must be a positive integer")
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -108,7 +108,7 @@ type ProjectRef struct {
 	PerfEnabled        *bool                        `bson:"perf_enabled,omitempty" json:"perf_enabled,omitempty" yaml:"perf_enabled,omitempty"`
 
 	// Container settings
-	ContainerSizes map[string]ContainerResources `bson:"container_sizes,omitempty" json:"container_sizes,omitempty" yaml:"container_sizes,omitempty"`
+	ContainerSizes map[string]*ContainerResources `bson:"container_sizes,omitempty" json:"container_sizes,omitempty" yaml:"container_sizes,omitempty"`
 
 	RepoRefId string `bson:"repo_ref_id" json:"repo_ref_id" yaml:"repo_ref_id"`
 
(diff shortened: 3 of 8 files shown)


model/project_ref.go
```diff
@@ -2434,6 +2434,44 @@ func GetMessageForPatch(patchID string) (string, error) {
 	return project.CommitQueue.Message, nil
 }
 
+func ValidateContainers(project string, containers []Container) error {
```
Kimchelly, line 2437: Document exported functions.

model/project_ref.go
```diff
@@ -2434,6 +2434,44 @@ func GetMessageForPatch(patchID string) (string, error) {
...
+	pRef, err := FindMergedProjectRef(project, "", true)
```
Kimchelly, line 2438: I don't completely understand what this merging project ref logic does when passed a version, so this comment might be wrong.

It seems like `ValidateContainers` might need a specific version's config rather than assuming that it's not associated with a version. For example, [generate tasks](https://github.com/evergreen-ci/evergreen/blob/c2a10070a720bc20018b7a0bed8b40b44820834e/units/generate_tasks.go#L162-L172), passes in the project and its particular version's project ref. `CheckProjectErrors` does not take the project ref into account (although `CheckProjectSettings` respects the passed-in project ref). Will this still validate correctly for a generated task that might add more configuration dynamically?

If this relies on the project ref to do validation, it seems like `validateContainers` could be part of `projectSettingsValidators` instead of `projectErrorValidators` so that the correct project ref gets passed into this validation without needing to do a lookup and missing the version-specific config.

validator/project_validator.go
```diff
@@ -677,19 +716,55 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
...
 			}
 		}
+		runOnHasDistro := false
+		runOnHasContainer := false
 		for _, distro := range buildVariant.RunOn {
-			if !utility.StringSliceContains(distroIDs, distro) && !utility.StringSliceContains(distroAliases, distro) {
+			if !utility.StringSliceContains(distroIDs, distro) && !utility.StringSliceContains(distroAliases, distro) && !containerNameMap[distro] {
 				errs = append(errs,
 					ValidationError{
 						Message: fmt.Sprintf("buildvariant '%s' references a nonexistent distro '%s'",
```
Kimchelly, line 738: "buildvariant '%s' references a nonexistent distro or container named '%s'"

validator/project_validator.go
```diff
@@ -668,7 +707,7 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 			}
 
 			for _, distro := range task.RunOn {
```
Kimchelly, line 709: I would rename `distro` to something more generic like `name` since distros and containers are mutually exclusive concepts.

validator/project_validator.go
```diff
@@ -619,7 +658,7 @@ func checkProjectFields(project *model.Project) ValidationErrors {
 // 1. A referenced task within a buildvariant task object exists in the set of project
 // tasks and is not referenced in the task group of the buildvariant.
 // 2. Any referenced distro exists within the current setting's distro directory
```
Kimchelly, line 660: Update this comment. I might remove some of the excessive comment description since it seems outdated and just summarize that `ensureReferentialIntegrity` checks that fields that reference other named things (e.g. task names, distro names, container names) are referring to valid names.

validator/project_validator.go
```diff
@@ -668,7 +707,7 @@ func ensureReferentialIntegrity(project *model.Project, distroIDs []string, dist
 			}
 
 			for _, distro := range task.RunOn {
-				if !utility.StringSliceContains(distroIDs, distro) && !utility.StringSliceContains(distroAliases, distro) {
+				if !utility.StringSliceContains(distroIDs, distro) && !utility.StringSliceContains(distroAliases, distro) && !containerNameMap[distro] {
 					errs = append(errs,
 						ValidationError{
 							Message: fmt.Sprintf("task '%s' in buildvariant '%s' "+
```
Kimchelly, line 713: "task '%s' in buildvariant '%s' references a nonexistent distro or container named '%s'"

validator/project_validator.go
```diff
@@ -243,7 +245,14 @@ func CheckProjectErrors(project *model.Project, includeLong bool) ValidationErro
 	if err != nil {
 		validationErrs = append(validationErrs, ValidationError{Message: "can't get distros from database"})
 	}
-	validationErrs = append(validationErrs, ensureReferentialIntegrity(project, distroIDs, distroAliases)...)
+	containerNameMap := map[string]bool{}
+	for _, container := range project.Containers {
+		if containerNameMap[container.Name] {
+			validationErrs = append(validationErrs, ValidationError{Message: fmt.Sprintf("container '%s' is already defined", container.Name)})
```
Kimchelly, line 251: Maybe phrase this as something like "container '%s' is defined multiple times"?

model/project_ref.go
```diff
@@ -2434,6 +2434,44 @@ func GetMessageForPatch(patchID string) (string, error) {
...
+	if err != nil {
+		return errors.Errorf("Unable to find merged project ref for project %s", project)
+	}
+	catcher := grip.NewSimpleCatcher()
+	for _, container := range containers {
+		catcher.Add(ValidateContainerSystem(container.System))
+		catcher.Add(ValidateContainerSize(container.Resources))
+		catcher.Add(ValidateContainerSize(pRef.ContainerSizes[container.Size]))
```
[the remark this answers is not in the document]
Kimchelly: On an unrelated note, since this is checking the project ref for correctness of its defined container sizes, is it more appropriate to put this check as part of `projectSettingsValidators` instead of this (which is under `projectErrorValidators`)?

service/project.go
```diff
@@ -592,9 +592,8 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 	for i, buildDef := range responseRef.PeriodicBuilds {
 		catcher.Wrapf(buildDef.Validate(), "invalid periodic build definition on line %d", i+1)
 	}
-	for _, containerResource := range responseRef.ContainerSizes {
-		catcher.NewWhen(utility.FromIntPtr(containerResource.CPU) <= 0, "container resource CPU must be a positive integer")
-		catcher.NewWhen(utility.FromIntPtr(containerResource.MemoryMB) <= 0, "container resource Memory MB must be a positive integer")
+	for _, containerResource := range containerSizes {
+		catcher.Add(model.ValidateContainerSize(containerResource))
```
Kimchelly, line 596: For a more user-friendly error message, I would wrap the error to include the container size name that was invalid:
```go
catcher.Wrapf(model.ValidateContainerSize(containerResource), "invalid container size '%s'", size)
```

Kimchelly (changes requested):

model/project_ref.go
```diff
@@ -2434,6 +2434,44 @@ func GetMessageForPatch(patchID string) (string, error) {
 	return project.CommitQueue.Message, nil
 }
 
+func ValidateContainers(project string, containers []Container) error {
+	pRef, err := FindMergedProjectRef(project, "", true)
+	if err != nil {
+		return errors.Errorf("Unable to find merged project ref for project %s", project)
```
Kimchelly, line 2440: There's a new error style (which you can probably see in the rest of the files if you rebase/merge), so errors involving DB lookups should be consistent with the new style.

```go
return errors.Wrapf(err, "finding merged project ref '%s'", project)
```

This comment might not matter at all depending on what's done with [this comment](https://github.com/evergreen-ci/evergreen/pull/5558/files#r846445422), but the new error style is the way to go.

Kimchelly, line 2440: There also needs to be a nil pRef check after this since `FindOne`s typically `return nil, nil` when the document is not found.

model/project_ref.go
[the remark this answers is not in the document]
hadjri: Sure, that makes sense

model/project_configs.go
```diff
@@ -36,7 +36,7 @@ type ProjectConfigFields struct {
 	TaskSync               *TaskSyncOptions               `yaml:"task_sync,omitempty" bson:"task_sync,omitempty"`
 	GithubTriggerAliases   []string                       `yaml:"github_trigger_aliases,omitempty" bson:"github_trigger_aliases,omitempty"`
 	PeriodicBuilds         []PeriodicBuildDefinition      `yaml:"periodic_builds,omitempty" bson:"periodic_builds,omitempty"`
-	ContainerSizes         map[string]ContainerResources  `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
+	ContainerSizes         map[string]*ContainerResources `yaml:"container_sizes,omitempty" bson:"container_sizes,omitempty"`
```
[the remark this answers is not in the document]
hadjri: I wanted to be able to check in validation if a ContainerResources struct was empty without it being an error (since it should be ok for the resources property to be missing from a container if the size is defined in its place). Since memory and cpu at 0 are errors, I made this into a pointer to differentiate the two cases. But I think I can just change the resources field on Container to a pointer and not in this map.

globals.go
```diff
@@ -1096,7 +1096,7 @@ const (
 // Validate checks that the container OS is recognized.
 func (c ContainerOS) Validate() error {
 	switch c {
-	case LinuxOS, WindowsOS:
+	case LinuxOS, WindowsOS, "":
```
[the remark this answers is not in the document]
hadjri: I'm not sure I understand - these properties are only set on a container itself, not on a task or variant. There are other validations checking if a task / variant's run_on is referencing a valid list of containers or distros.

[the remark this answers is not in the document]
hadjri: It wouldn't be possible for a host task/variant to have this set.

model/project_ref.go
hadjri: Moving into `projectSettingsValidators ` sgtm so we can remove the need for this DB call to retrieve the project ref.  The function call has been removed but to explain, the empty string is used for merging with a version controlled project config, where an empty string will cause the ref to be merged with the most recent good version of the project config. It doesn't signify an empty version. In some cases such as specific tasks / revisions a version is passed in, but in this case I was just merging with the most recent project config.

globals.go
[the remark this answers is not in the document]
Kimchelly: Sorry, I can rephrase this - I meant that an empty OS is not really a valid OS for a container definition since we don't have a default OS. We should expect `(model.ContainerSystem).Validate` to error for an empty OS. 

In any case, it seems like the empty string validation went away in a newer commit.

model/project_ref.go
```diff
@@ -2434,17 +2434,17 @@ func GetMessageForPatch(patchID string) (string, error) {
...
 	catcher := grip.NewSimpleCatcher()
 	for _, container := range containers {
-		catcher.Add(ValidateContainerSystem(container.System))
-		catcher.Add(ValidateContainerSize(container.Resources))
-		catcher.Add(ValidateContainerSize(pRef.ContainerSizes[container.Size]))
-		catcher.NewWhen(container.Size != "" && pRef.ContainerSizes[container.Size] == nil, fmt.Sprintf("size '%s' is not defined anywhere", container.Size))
+		catcher.Add(container.System.Validate())
+		catcher.Add(container.Resources.Validate())
+		size, ok := pRef.ContainerSizes[container.Size]
+		catcher.AddWhen(ok, size.Validate())
```
Kimchelly, line 2446: I would check `ok` separately because when written like this, Go will execute `size.Validate()` regardless of whether `ok` is true or false:
```go
if ok {
    catcher.Add(size.Validate())
}
```

Small example might explain why better than my words:
```go
import (
	"fmt"

	"github.com/mongodb/grip"
)

func foo() string {
	fmt.Println("hi, don't print me")
	return "hi"
}

func main() {
	m := map[string]struct{}{}
	m["apple"] = struct{}{}
	_, ok := m["banana"]
	catcher.NewWhen(ok, foo()) // foo still executes, so it prints "hi, don't print me"
	fmt.Println("should not have added any errors:", catcher.Resolve())
}
````

File: validator/project_validator.go [hunk 14]
```diff
@@ -654,10 +654,7 @@ func checkProjectFields(project *model.Project) ValidationErrors {
 	return errs
 }
 
-// Ensures that:
-// 1. A referenced task within a buildvariant task object exists in the set of project
-// tasks and is not referenced in the task group of the buildvariant.
-// 2. Any referenced distro exists within the current setting's distro directory
+// ensureReferentialIntegrity checks all fields that reference other entities defined in the YAML and ensure that they are referring to valid names
```
Kimchelly, line 657: nit: period at the end of the sentence.

model/project_ref.go
```diff
@@ -2453,22 +2453,21 @@ func ValidateContainers(project string, containers []Container) error {
 	return catcher.Resolve()
 }
 
-func ValidateContainerSystem(containerSystem ContainerSystem) error {
+func (c ContainerSystem) Validate() error {
```
Kimchelly, line 2456: Documentation for exported methods.

model/project_ref.go
```diff
@@ -2453,22 +2453,21 @@ func ValidateContainers(project string, containers []Container) error {
...
+	}
+	catcher.NewWhen(c.OperatingSystem == evergreen.LinuxOS && c.WindowsVersion != "", "cannot specify windows version when OS is linux")
 	return catcher.Resolve()
 }
 
-func ValidateContainerSize(containerResource *ContainerResources) error {
-	if containerResource == nil {
-		return nil
-	}
+func (c ContainerResources) Validate() error {
```
Kimchelly, line 2467: Documentation for exported methods.

Kimchelly (changes requested):

validator/project_validator_test.go
```diff
@@ -3048,6 +3048,10 @@ func TestValidateVersionControl(t *testing.T) {
 }
 
 func TestValidateContainers(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	env := tu.NewEnvironment(ctx, t)
```
Kimchelly, line 3053: Is this necessary? Seems like nothing in `validateContainers` depends on the environment.

In general, I would avoid relying on `evergreen.SetEnvironment` because it changes global in-memory state for any test that runs in the same process as this one (i.e. any other test being tested in this `go test` invocation), which could affect other tests that depend on it (and for similar reasons, magicking up the environment in arbitrary places using `evergreen.GetEnvironment` in source code is also not preferable except in a few specific scenarios). The more preferred way to do this is to dependency inject an `evergreen.Environment` into functions that depend on it and unit test them by passing in mock environments. I usually do this by just using `mock.Environment`, calling `(*mock.Environment).Configure` on it, and then passing it into whatever function depends on the global environment.

Kimchelly (approved):
LGTM mod one comment.

validator/project_validator_test.go
hadjri: Ok, I'll try that out, I had added this because I was having a local failure in my IDE that was fixed when I manually set the environment, but it doesn't seem to be needed when running the patch.

Kimchelly: If I had to take a  guess, it's probably because there's no specific ordering of test execution, which is actually a nice property since it becomes obvious if your tests fail to clean up after themselves because they'll become flaky depending on the order that tests run. Unfortunately, all tests that involve the DB rely on Anser, which relies on the global env's mongo client to perform operations. Once the context is cancelled on the global env (i.e. the `defer cancel()` runs), the only way to continue doing DB operations is to reset the environment.

It should generally not be necessary to `SetEnvironment` in tests unless 1. the code you're testing calls `evergreen.GetEnvironment` (which should be done sparingly) or 2. another test in the package has either (often unnecessarily) set it and failed to reset it after the test finishes, which means the test that originally had to set it should probably either reset it to the original env. Unfortunately, a lot of unit tests in Evergreen have an issue with not tearing down and cleaning up the global state changes they produce (I'm trying to fix this right now).

PR #5559, opened 2022-04-08
EVG-16414: stop and start spawn hosts idempotently

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16414

### Description 
I made stop/start less prone to failing partway through the process. This code is mostly tricky because of the "stopping" intermediate UI state and the fact that it has to poll after making the request until the cloud host reports that it's truly stopped.

* The code no longer assumes that DB hosts perfectly reflect the true state of the cloud host, so it calls `StopInstance`/`StartInstance` to both modify the cloud host and get the current cloud host status. I tested it in AWS and both operations are idempotent, so calling them multiple times is fine and they'll just return the current status. Checking the current status response from `StopInstances`/`StartInstances` avoids the polling in the case that the cloud host has already been stopped or stops quickly.
* It's now possible for users to re-request stopping a host if the host stop job finished with errors (it's assumed that if the user wants to know what went wrong during stopping, they can check the event logs to see that the job failed partway through execution and press the stop button again).
    * If the job gets stranded because of a deploy, it'll restart the job after the lock timeout.
* The "stopping" state is just an intermediate state to show in the UI that the host is in the process of stopping. Since it seems more like a nice-to-have, I changed it so that failing to set the host to "stopping" does not exit from the stop attempt.

#### Tech Debt
* Standardize the way that spawn hosts are started/stopped from REST routes since it should be logically the same as starting/stopping from the UI.
* Move some REST data host methods from `rest/data/volume.go` into a more appropriate file.

### Testing 
Updated unit tests and tested stopping a host in staging from the UI works as expected.


cloud/ec2.go
```diff
@@ -1257,22 +1240,20 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
...
-	return errors.Wrap(h.SetStopped(user), "failed to mark instance as stopped in db")
+	return errors.Wrap(h.SetStopped(user), "marking DB host as stopped")
 }
 
 // StartInstance starts a stopped EC2 instance.
 func (m *ec2Manager) StartInstance(ctx context.Context, h *host.Host, user string) error {
-	// Check that target instance is stopped
 	if h.Status != evergreen.HostStopped {
-		return errors.Errorf("cannot start '%s' - host is not stopped", h.Id)
+		return errors.Errorf("cannot start '%s' - host status is '%s'", h.Id, h.Status)
```
ybrill, line 1249: would it be a benefit to allow HostStarting through here as well?

rest/data/host.go
```diff
@@ -165,3 +168,62 @@ func FindHostByIdWithOwner(hostID string, user gimlet.User) (*host.Host, error)
...
+		}
+		return http.StatusInternalServerError, err
+	}
+	return http.StatusOK, nil
+
+}
+
+// StartSpawnHost enqueues a job to start a stopped spawn host.
+func StartSpawnHost(ctx context.Context, env evergreen.Environment, u *user.DBUser, h *host.Host) (int, error) {
+	if h.Status != evergreen.HostStopped {
```
ybrill, line 215: do we want to let HostStarting through here too?

cloud/ec2.go
```diff
@@ -1167,66 +1167,57 @@ func (m *ec2Manager) TerminateInstance(ctx context.Context, h *host.Host, user,
...
+				"host_id": h.Id,
+				"user":    user,
+			}))
+		case StatusStopped:
+			return errors.Wrap(h.SetStopped(user), "marking DB host as stopped")
+		default:
+			return errors.Errorf("instance is in unexpected state '%s'", status)
+		}
+	}
+	grip.WarningWhen(len(out.StoppingInstances) != 1, message.Fields{
```
ybrill, line 1204: if this is the case is it worth going through with the rest of the function to poll the instance status waiting for it to stop?

cloud/ec2.go
```diff
@@ -1257,22 +1240,20 @@ func (m *ec2Manager) StopInstance(ctx context.Context, h *host.Host, user string
...
+		return errors.Errorf("cannot start '%s' - host status is '%s'", h.Id, h.Status)
 	}
 
 	if err := m.client.Create(m.credentials, m.region); err != nil {
 		return errors.Wrap(err, "error creating client")
 	}
 	defer m.client.Close()
 
-	// Make request to start the instance
 	_, err := m.client.StartInstances(ctx, &ec2.StartInstancesInput{
```
ybrill, line 1257: could it make sense to get the status from the response here like for stopping?

cloud/ec2.go
```diff
@@ -1167,66 +1167,57 @@ func (m *ec2Manager) TerminateInstance(ctx context.Context, h *host.Host, user,
...
 	}
 
-	// Check whether instance stopped
+	if len(out.StoppingInstances) == 1 {
+		instance := out.StoppingInstances[0]
+		status := ec2StatusToEvergreenStatus(utility.FromStringPtr(instance.CurrentState.Name))
+		switch status {
+		case StatusStopping:
+			grip.Error(message.WrapError(h.SetStopping(user), message.Fields{
+				"message": "could not mark host as stopping, continuing to poll instance status anyways",
```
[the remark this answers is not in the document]
Kimchelly: Probably. But if that happens, two things can happen to the job, both of which seem acceptable:
* The job also fails to update the Amboy DB, in which case it'll re-dispatch after the stale lock timeout and try doing this again.
* The job successfully updates the Amboy DB and logs an event to the host events saying the failure reason, at which point the user could see the error message in the host event page and press the stop button again if they wanted to try again.

cloud/ec2.go
Kimchelly: This shouldn't happen unless we write a bug in the code, AWS returns an unexpected result, or our understanding of the AWS API results is incorrect. This is more of a sanity check since AWS is allowed to return us a slice of any number of results. The API should return exactly 1 result if it succeeded, but we can't guarantee that on our end. Checking `out.StoppingInstances` is just a shortcut to avoid an unnecessary polling loop if AWS tells us it's already stopped.

In the case of unexpected number of results, polling should still probably be done (at least from a correctness perspective) so that `GetInstanceInfo` can tell us what the real cloud host status is after we've received an unexpected number of results to make it more likely that the cloud host and DB host statuses are in sync with each other.

cloud/ec2.go
Kimchelly: Allowing `HostStarting` would produce some ambiguity in the host lifecycle with not much benefit, since users can't start a host that's in the `HostStarting` state anyways. `HostStarting` is not analogous to `HostStopping`, and specifically refers to the initial host start, not subsequent reboots. `HostStopping` only exists as a convenience warning to the user to indicate that stopping is happening and they should probably stop whatever work they're doing on the host. In contrast, rebooting from stopped to running doesn't require the user to act on it since they won't be able to use the host until it's running anyways (and they get their IP address).

cloud/ec2.go
Kimchelly: The `GetInstanceInfo` is unavoidable here since we have to cache the instance data (e.g. to get the new IP address) in `cacheHostData`.

rest/data/host.go
```diff
@@ -165,3 +168,62 @@ func FindHostByIdWithOwner(hostID string, user gimlet.User) (*host.Host, error)
...
+
+	return http.StatusOK, nil
+}
+
+// StopSpawnHost enqueues a job to stop a running spawn host.
+func StopSpawnHost(ctx context.Context, env evergreen.Environment, u *user.DBUser, h *host.Host) (int, error) {
+	if h.Status == evergreen.HostStopped {
+		return http.StatusBadRequest, errors.Errorf("host '%s' is already stopped", h.Id)
+	}
+	if h.Status != evergreen.HostRunning && h.Status != evergreen.HostStopping {
```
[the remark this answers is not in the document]
Kimchelly: Hmm yeah, it seems like Spruce should also allow stopping hosts to be stopped again.

Alternatively, we could get rid of the "stopping" intermediate state altogether since it only serves as a convenience warning to the user.

rest/data/host.go
Kimchelly: See my reasoning above why not.

cloud/ec2.go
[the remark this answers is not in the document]
ybrill: but I mean to say: you mentioned it was an explicit decision not to return here. I was wondering why not return now if we can't set a state on the host.

cloud/ec2.go
ybrill: I get it now. Thank you for explaining.
As a user I appreciate that something changes on the UI when I click the stop button and the visibility it gives into the host state. I never realized I wasn't getting it for starting.

rest/data/host.go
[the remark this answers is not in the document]
ybrill: One way or another works for me. I mean, I like the stopping state as a user, but if it has to go 🤷 .

cloud/ec2.go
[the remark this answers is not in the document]
Kimchelly: Mostly because "stopping" is only a nice to have state for the user's benefit and is not necessary for `StopInstances` to be considered a success. It's still possible that there's something that prevents any DB writes, but if it's just a momentary blip (or something else that prevents `SetStopping` from updating), I don't want it to give up here just because the user doesn't see the "stopping" state. Once it's reached this point, the cloud host is really stopping, so it's more useful to at least have debug logs or more information by continuing on error.

cloud/ec2.go
Kimchelly: The difference is mostly in whether the user might be interrupted while doing something. For starting a stopped host, the user can't be interrupted while using the host since they don't have access to it anyways, so there's not as much benefit to propagating the restarting state except to remind the user that they pressed the start button. The "stopping" state is possibly more important since the user might still be actively using the host.

rest/data/host.go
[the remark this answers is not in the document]
Kimchelly: I filed a UI ticket ([EVG-16712](https://jira.mongodb.org/browse/EVG-16712)) to let users re-press the stop button if something goes wrong.

ybrill (approved):

PR #5561, opened 2022-04-08
EVG-16099 Add patch route for task annotations

bynn:
[EVG-16099](https://jira.mongodb.org/browse/EVG-16099)

### Description 
add patch route 
Note and Meta data field will be overwritten if it is specified in the body 
Issues will be added on 

### Testing 
unit test

### DIFF
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -207,6 +207,57 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 	return errors.Wrapf(err, "adding task annotation for task '%s'", a.TaskId)
 }
 
+func AddToAnnotation(a *TaskAnnotation, userDisplayName string) error {
+	existingAnnotation, err := FindOneByTaskIdAndExecution(a.TaskId, a.TaskExecution)
+	if err != nil {
+		return errors.Wrapf(err, "finding task annotation for task id '%s'", a.TaskId)
+	}
+	if existingAnnotation == nil {
+		return errors.Wrapf(err, "missing task annotation for task id '%s'", a.TaskId)
+	}
+	source := &Source{
+		Author:    userDisplayName,
+		Time:      time.Now(),
+		Requester: APIRequester,
+	}
+	update := bson.M{
+		TaskIdKey:        a.TaskId,
+		TaskExecutionKey: a.TaskExecution,
+	}
+
+	// Overwrite note and metadata.
+	if a.Note != nil {
+		a.Note.Source = source
+		update[NoteKey] = a.Note
+	}
+	if existingAnnotation.Metadata != nil {
+		update[MetadataKey] = existingAnnotation.Metadata
+	}
+
+	if a.Issues != nil {
+		for i := range a.Issues {
+			a.Issues[i].Source = source
+			existingAnnotation.Issues = append(existingAnnotation.Issues, a.Issues[i])
+		}
+	}
+	update[IssuesKey] = existingAnnotation.Issues
+
+	if a.SuspectedIssues != nil {
+		for i := range a.SuspectedIssues {
+			a.SuspectedIssues[i].Source = source
+			existingAnnotation.SuspectedIssues = append(existingAnnotation.SuspectedIssues, a.SuspectedIssues[i])
+		}
+	}
+	update[SuspectedIssuesKey] = existingAnnotation.SuspectedIssues
+
+	err = db.Update(
+		Collection,
+		ByTaskIdAndExecution(a.TaskId, a.TaskExecution),
+		update,
+	)
+	return errors.Wrapf(err, "updating task annotation for '%s'", a.TaskId)
+}
+
 func AddCreatedTicket(taskId string, execution int, ticket IssueLink, userDisplayName string) error {
 	source := &Source{
 		Author:    userDisplayName,
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -225,15 +225,6 @@ func AddToAnnotation(a *TaskAnnotation, userDisplayName string) error {
 		TaskExecutionKey: a.TaskExecution,
 	}
 
-	// Overwrite note and metadata.
-	if a.Note != nil {
-		a.Note.Source = source
-		update[NoteKey] = a.Note
-	}
-	if a.Metadata != nil {
-		update[MetadataKey] = a.Metadata
-	}
-
 	if a.Issues != nil {
 		for i := range a.Issues {
 			a.Issues[i].Source = source
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -105,7 +105,7 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	}
 
 	if annotation != nil && annotation.Note != nil && annotation.Note.Message != originalMessage {
-		return errors.New("note is out of sync, please try again")
+		return errors.New(", note is out of syncplease try again")
 	}
 	_, err = db.Upsert(
 		Collection,
@@ -211,15 +211,23 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 func PatchAnnotation(a *TaskAnnotation, userDisplayName string, upsert bool) error {
 	existingAnnotation, err := FindOneByTaskIdAndExecution(a.TaskId, a.TaskExecution)
 	if err != nil {
-		return errors.Wrapf(err, "finding task annotation for task id '%s'", a.TaskId)
+		return errors.Wrapf(err, "finding task annotation")
 	}
-	if existingAnnotation == nil {
+	if existingAnnotation == nil && upsert {
 		if !upsert {
-			return nil
+			return errors.Errorf("annotation for task '%s' and execution %d not found", a.TaskId, a.TaskExecution)
 		} else {
 			return UpdateAnnotation(a, userDisplayName)
 		}
 	}
+
+	if a.Note != nil {
+		return errors.New("note should be empty")
+	}
+	if a.Metadata != nil {
+		return errors.New("metadata should be empty")
+	}
+
 	source := &Source{
 		Author:    userDisplayName,
 		Time:      time.Now(),
@@ -230,28 +238,26 @@ func PatchAnnotation(a *TaskAnnotation, userDisplayName string, upsert bool) err
 	if a.Issues != nil {
 		for i := range a.Issues {
 			a.Issues[i].Source = source
-			existingAnnotation.Issues = append(existingAnnotation.Issues, a.Issues[i])
 		}
-		update[IssuesKey] = existingAnnotation.Issues
+		update[IssuesKey] = bson.M{"$each": a.Issues}
 	}
 
 	if a.SuspectedIssues != nil {
 		for i := range a.SuspectedIssues {
 			a.SuspectedIssues[i].Source = source
-			existingAnnotation.SuspectedIssues = append(existingAnnotation.SuspectedIssues, a.SuspectedIssues[i])
 		}
-		update[SuspectedIssuesKey] = existingAnnotation.SuspectedIssues
+		update[SuspectedIssuesKey] = bson.M{"$each": a.SuspectedIssues}
 	}
 
 	if len(update) == 0 {
 		return nil
 	}
 
-	_, err = db.Upsert(
+	err = db.Update(
 		Collection,
 		ByTaskIdAndExecution(a.TaskId, a.TaskExecution),
 		bson.M{
-			"$set": update,
+			"$push": update,
 		},
 	)
 	return errors.Wrapf(err, "updating task annotation for '%s'", a.TaskId)
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -230,8 +230,8 @@ func AddToAnnotation(a *TaskAnnotation, userDisplayName string) error {
 		a.Note.Source = source
 		update[NoteKey] = a.Note
 	}
-	if existingAnnotation.Metadata != nil {
-		update[MetadataKey] = existingAnnotation.Metadata
+	if a.Metadata != nil {
+		update[MetadataKey] = a.Metadata
 	}
 
 	if a.Issues != nil {
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -105,7 +105,7 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	}
 
 	if annotation != nil && annotation.Note != nil && annotation.Note.Message != originalMessage {
-		return errors.New(", note is out of syncplease try again")
+		return errors.New("note is out of sync, please try again")
 	}
 	_, err = db.Upsert(
 		Collection,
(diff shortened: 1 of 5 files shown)


model/annotations/task_annotations.go
```diff
@@ -204,6 +204,57 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 	return errors.Wrapf(err, "problem adding task annotation for '%s'", a.TaskId)
 }
 
+func AddToAnnotation(a *TaskAnnotation, userDisplayName string) error {
+	existingAnnotation, err := FindOneByTaskIdAndExecution(a.TaskId, a.TaskExecution)
+	if err != nil {
+		return errors.Wrapf(err, "finding task annotation for task id '%s'", a.TaskId)
+	}
+	if existingAnnotation == nil {
+		return errors.Wrapf(err, "missing task annotation for task id '%s'", a.TaskId)
```
malikchaya2, line 213: Why is this an error? If an annotation doesn't exist, do we not want to create one?

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,57 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
+		Time:      time.Now(),
+		Requester: APIRequester,
+	}
+	update := bson.M{
+		TaskIdKey:        a.TaskId,
+		TaskExecutionKey: a.TaskExecution,
+	}
+
+	// Overwrite note and metadata.
+	if a.Note != nil {
```
malikchaya2, line 226: We can technically add on to the note. Should we consider doing that?

rest/route/annotations.go
```diff
@@ -323,6 +323,125 @@ func (h *annotationByTaskPutHandler) Run(ctx context.Context) gimlet.Responder {
...
+
+type annotationByTaskPatchHandler struct {
+	taskId string
+
+	user       gimlet.User
+	annotation *restModel.APITaskAnnotation
+}
+
+func makePatchAnnotationsByTask() gimlet.RouteHandler {
+	return &annotationByTaskPutHandler{}
```
malikchaya2, line 338: *annotationByTaskPatchHandler

rest/route/annotations.go
```diff
@@ -323,6 +323,125 @@ func (h *annotationByTaskPutHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+func (h *annotationByTaskPatchHandler) Factory() gimlet.RouteHandler {
+	return &annotationByTaskPutHandler{}
```
malikchaya2, line 342: *annotationByTaskPatchHandler

rest/route/annotations.go
```diff
@@ -323,6 +323,125 @@ func (h *annotationByTaskPutHandler) Run(ctx context.Context) gimlet.Responder {
...
+		return gimlet.ErrorResponse{
+			Message:    "task execution must be specified in the url or request body",
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+
+	// check if the task exists
+	t, err := task.FindByIdExecution(h.taskId, h.annotation.TaskExecution)
+	if err != nil {
+		return errors.Wrap(err, "error finding task")
```
malikchaya2, line 393: let's remove "error" to be more consistent with the new style preference

malikchaya2 (changes requested):
Let's also update the docs before closing this, or create a ticket for it so that it doesn't get lost.

rest/route/annotations.go
```diff
@@ -323,6 +323,125 @@ func (h *annotationByTaskPutHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+func (h *annotationByTaskPatchHandler) Factory() gimlet.RouteHandler {
+	return &annotationByTaskPutHandler{}
+}
+
+func (h *annotationByTaskPatchHandler) Parse(ctx context.Context, r *http.Request) error {
+	var err error
+	h.taskId = gimlet.GetVars(r)["task_id"]
+	taskExecutionsAsString := r.URL.Query().Get("execution")
```
hadjri, line 348: This looks a lot like Parse function on annotationByTaskPutHandler - can we modularize these a bit?

hadjri (changes requested):

rest/route/annotations.go
bynn: created a helper fuction

model/annotations/task_annotations.go
malikchaya2: hmm, I'm not sure. It would depend on how they are using it and if DAG is still sending the metadata in the note. This would probably be a question for Maria or Sanjog.

bynn: @sanjogDB do we have a preference on what we want to do?

sanjogDB: let me understand how DAG uses it and I'll get back after that!

rest/route/annotations.go
```diff
@@ -203,41 +203,31 @@ func (h *annotationByTaskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
-	return &annotationByTaskPutHandler{}
-}
-
-func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
-	return &annotationByTaskPutHandler{}
-}
-
-func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request) error {
+func annotationByTaskPutOrPatchParser(ctx context.Context, r *http.Request) (*annotationByTaskPutOrPatchHandler, error) {
+	ret := &annotationByTaskPutOrPatchHandler{}
```
ablack12, line 215: Nit: res (for result) is more typical

rest/route/annotations.go
```diff
@@ -203,41 +203,31 @@ func (h *annotationByTaskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
-func makePutAnnotationsByTask() gimlet.RouteHandler {
-	return &annotationByTaskPutHandler{}
-}
-
-func (h *annotationByTaskPutHandler) Factory() gimlet.RouteHandler {
-	return &annotationByTaskPutHandler{}
-}
-
-func (h *annotationByTaskPutHandler) Parse(ctx context.Context, r *http.Request) error {
+func annotationByTaskPutOrPatchParser(ctx context.Context, r *http.Request) (*annotationByTaskPutOrPatchHandler, error) {
```
ablack12, line 214: Rather than this returning an entirely new kind of handler, ti would be preferable if this function just returned the things needed by the handlers (taskId and annotation), or returned a private struct of annotationRouteParams

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,48 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 	return errors.Wrapf(err, "problem adding task annotation for '%s'", a.TaskId)
 }
 
+func AddToAnnotation(a *TaskAnnotation, userDisplayName string) error {
```
ablack12, line 207: Should add function comment. Also preference for PatchAnnotation

model/annotations/task_annotations.go
ablack12: Error is nil here so wrapping it still returns nil

ablack12: I think it would make sense to add an Upsert boolean to the route, where we insert a new annotation if one already exists (as this behavior may confuse new users or require them to make multiple calls).

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,48 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
+		TaskExecutionKey: a.TaskExecution,
+	}
+
+	if a.Issues != nil {
+		for i := range a.Issues {
+			a.Issues[i].Source = source
+			existingAnnotation.Issues = append(existingAnnotation.Issues, a.Issues[i])
+		}
+	}
+	update[IssuesKey] = existingAnnotation.Issues
```
ablack12, line 231: This should be in the if statement at 225, i.e. we should only overwrite issues if the user intended to.

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,48 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
+	}
+	if existingAnnotation == nil {
+		return errors.Wrapf(err, "missing task annotation for task id '%s'", a.TaskId)
+	}
+	source := &Source{
+		Author:    userDisplayName,
+		Time:      time.Now(),
+		Requester: APIRequester,
+	}
+	update := bson.M{
```
ablack12, line 220: We shouldn't allow the user to update taskId or taskExecution.

ablack12 (changes requested):

model/annotations/task_annotations.go
bynn: done

model/annotations/task_annotations.go
bynn: done

model/annotations/task_annotations.go
bynn: done

rest/route/annotations.go
bynn: done

rest/route/annotations.go
bynn: done

model/annotations/task_annotations.go
```diff
@@ -207,6 +207,56 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 	return errors.Wrapf(err, "adding task annotation for task '%s'", a.TaskId)
 }
 
+// PatchAnnotation adds issues onto existing annotations.
+func PatchAnnotation(a *TaskAnnotation, userDisplayName string, upsert bool) error {
+	existingAnnotation, err := FindOneByTaskIdAndExecution(a.TaskId, a.TaskExecution)
+	if err != nil {
+		return errors.Wrapf(err, "finding task annotation for task id '%s'", a.TaskId)
+	}
+	if existingAnnotation == nil {
```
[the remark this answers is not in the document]
bynn: switched to using update and $push

model/annotations/task_annotations.go
```diff
@@ -105,7 +105,7 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	}
 
 	if annotation != nil && annotation.Note != nil && annotation.Note.Message != originalMessage {
-		return errors.New("note is out of sync, please try again")
+		return errors.New(", note is out of syncplease try again")
```
malikchaya2, line 108: this looks like an accident

model/annotations/task_annotations.go
```diff
@@ -211,15 +211,23 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
 func PatchAnnotation(a *TaskAnnotation, userDisplayName string, upsert bool) error {
 	existingAnnotation, err := FindOneByTaskIdAndExecution(a.TaskId, a.TaskExecution)
 	if err != nil {
-		return errors.Wrapf(err, "finding task annotation for task id '%s'", a.TaskId)
+		return errors.Wrapf(err, "finding task annotation")
```
malikchaya2, line 214: are we sure we want to remove the task id?

model/annotations/task_annotations.go
```diff
@@ -211,15 +211,23 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
 	}
-	if existingAnnotation == nil {
+	if existingAnnotation == nil && upsert {
 		if !upsert {
```
malikchaya2, line 217: at this point we know upsert is true because of the if on 216

model/annotations/task_annotations.go
bynn: yes dont know what happened

model/annotations/task_annotations_test.go
```diff
@@ -159,3 +159,42 @@ func TestMoveSuspectedIssueToIssue(t *testing.T) {
 	assert.Equal(t, "different user", annotationFromDB.Issues[0].Source.Author)
 	assert.Equal(t, "someone new", annotationFromDB.Issues[1].Source.Author)
 }
+
+func TestPatchIssue(t *testing.T) {
```
malikchaya2, line 163: do we maybe want to add a test that has a note or metadata?

model/annotations/task_annotations.go
bynn: it seems that that's what kim did in her fixes in other functions in this file so i wanted to stay consistent

model/annotations/task_annotations.go
```diff
@@ -207,6 +207,62 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
+	}
+	if existingAnnotation == nil && upsert {
+		if !upsert {
+			return errors.Errorf("annotation for task '%s' and execution %d not found", a.TaskId, a.TaskExecution)
+		} else {
+			return UpdateAnnotation(a, userDisplayName)
+		}
+	}
+
+	if a.Note != nil {
```
malikchaya2, line 224: how about if no annotation exists? do we want to allow a note or metadata then?

model/annotations/task_annotations.go
bynn: ah it wasnt supposed to check it in the first if

model/annotations/task_annotations.go
malikchaya2: her fix was more about voiding phrases like "failed to" to make things less noisy. You can read more here: https://jira.mongodb.org/browse/EVG-16501. 
I don't think we should be removing the task id specifically.

model/annotations/task_annotations_test.go
bynn: done

model/annotations/task_annotations.go
bynn: added taskid and execution

model/annotations/task_annotations.go
malikchaya2: that's for `existingAnnotation == nil && upsert`, but how about for `existingAnnotation == nil && !upsert`?

model/annotations/task_annotations_test.go
```diff
@@ -159,3 +160,47 @@ func TestMoveSuspectedIssueToIssue(t *testing.T) {
...
+	assert.NotNil(t, annotation.SuspectedIssues[0].Source)
+	assert.Equal(t, APIRequester, annotation.SuspectedIssues[0].Source.Requester)
+	assert.Equal(t, "insert", annotation.SuspectedIssues[0].Source.Author)
+	assert.Equal(t, "EVG-3456", annotation.SuspectedIssues[0].IssueKey)
+
+	badInsert := TaskAnnotation{TaskId: "t1", TaskExecution: 1, Note: &Note{Message: "shouldn't work"}}
+	assert.Error(t, PatchAnnotation(&badInsert, "error out ", true))
+
+	badInsert2 := TaskAnnotation{TaskId: "t1", TaskExecution: 1, Metadata: &birch.Document{}}
+	assert.Error(t, PatchAnnotation(&badInsert2, "error out ", false))
```
malikchaya2, line 205: can we add a test for the use case where upsert is true and there is a note, to make sure that the notes get set and to document the desired behavior?

malikchaya2 (approved):
Approved with one last test request. 
great work! 🎉

PR #5562, opened 2022-04-11
EVG-16706: convert container allocation statuses into separate field

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16706

### Description 
The container-unallocated and container-allocated states were used to represent different phases of container allocation for undispatched tasks, but the task lifecycle is logically simpler and requires fewer changes and less if-else checking if both host and container tasks use undispatched as the initial task state. Instead, there's a separate field for container tasks to indicate whether the container has been allocated to it yet or not for an undispatched task. The only downside is that some indexed queries that are specific to host/container tasks might be less than optimal, but it shouldn't be a major performance issue right now.

* Convert "container-unallocated" and "container-allocated" task statuses into a boolean field.
* Use "undispatched" as initial task status for both host and container tasks, which is used to indicate that the task should continue making progress towards dispatching to an agent.

### Testing 
Updated unit tests.

### DIFF
--- a/model/task/db_test.go
+++ b/model/task/db_test.go
@@ -622,23 +622,33 @@ func TestDisplayStatus(t *testing.T) {
 			},
 		},
 	}
+	// No CheckStatuses for t12 to avoid paradox
 	assert.NoError(t, t12.Insert())
 	checkStatuses(t, evergreen.TaskWillRun, t11)
 	t13 := Task{
-		Id:        "t13",
-		Status:    evergreen.TaskContainerUnallocated,
-		Activated: true,
+		Id:                 "t13",
+		Status:             evergreen.TaskUndispatched,
+		Activated:          true,
+		ContainerAllocated: false,
 	}
-	assert.NoError(t, t13.Insert())
-	// No CheckStatuses for t12 to avoid paradox
+	require.NoError(t, t13.Insert())
+	checkStatuses(t, evergreen.TaskWillRun, t13)
 	t14 := Task{
-		Id:        "t14",
-		Status:    evergreen.TaskContainerAllocated,
-		Activated: true,
+		Id:                 "t14",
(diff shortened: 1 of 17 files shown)


model/task/db_test.go
```diff
@@ -622,23 +622,33 @@ func TestDisplayStatus(t *testing.T) {
 			},
 		},
 	}
+	// No CheckStatuses for t12 to avoid paradox
```
julianedwards, line 625: What does this mean?

model/task_lifecycle.go
```diff
@@ -339,10 +339,9 @@ func AbortTask(taskId, caller string) error {
 // tasks for the same build variant + display name + project combination
 // as the task.
 func DeactivatePreviousTasks(t *task.Task, caller string) error {
-	statuses := []string{evergreen.TaskUndispatched, evergreen.TaskContainerUnallocated}
 	filter, sort := task.ByActivatedBeforeRevisionWithStatuses(
 		t.RevisionOrderNumber,
-		statuses,
+		[]string{evergreen.TaskUndispatched},
```
julianedwards, line 344: Does this function still need to take in a slice of strings?

julianedwards (changes requested):

model/task/db_test.go
Kimchelly: It doesn't seem like a paradox to me, so I don't really know. I just moved this comment because it was on the wrong line.

Kimchelly:
I slightly modified the dispatcher logic during deallocation because the dispatcher should only care about the `ContainerAllocated` flag when deallocating a task. The reasoning is that it doesn't really matter if the task's status is "undispatched" or not, the task only needs to know that it no longer has a container to run it.

julianedwards (approved):

PR #5563, opened 2022-04-11
EVG-12900: Support sorting tasks by TimeTaken field

minnakt:
[EVG-12900](https://jira.mongodb.org/browse/EVG-12900)

### Description 
This PR adds a `taskDuration` field to the `taskResolver`. 

Tasks in the database already have a `timeTaken` field, however it is meaningless if a task has not finished (i.e. the `timeTaken` will always be 0 if the task has not finished). The `taskDuration` field returns how long the task has been running (if it hasn’t finished) or how long it ran (if it has finished).

Additionally, this PR adds the ability to sort by `taskDuration` for `patchTasks` by modifying the aggregation pipeline in `GetTasksByVersion`.

### Testing 
- Added graphql test `patchTasks/sort-by-duration.graphql`
- Added graphql test `task/taskDuration.graphql` (important note: I did not test this for the case where a task is running because I would need to have a static `time.Now()`, which is related to [EVG-16565](https://jira.mongodb.org/browse/EVG-16565))
- Added test for sorting `TestGetTasksByVersionSorting` in `model/task/task_test.go` (possibly unnecessary, as this file didn't have any tests for sorting before)


graphql/resolvers.go
```diff
@@ -3350,6 +3352,21 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
 	return apiTask, nil
 }
 
+func (r *taskResolver) TaskDuration(ctx context.Context, obj *restModel.APITask) (*restModel.APIDuration, error) {
+	// TaskDuration is sometimes added via aggregation when Task is resolved via GetTasksByVersion.
+	if obj.TaskDuration != 0 {
+		return &obj.TaskDuration, nil
+	}
+	// If status is running / dispatched, calculate the time elapsed since it became active.
```
ablack12, line 3360: Activated doesn't mean started time, it means the time at which the user marked it to be scheduled. More accurate would be StartTime (not sure if this would be set for TaskDispatched so that may need to be 0; that makes sense to me though since that's Evergreen getting the task ready still, and not how long the task has been running for).

ablack12 (approved):

ablack12 (changes requested):
whoops 😅

model/task/task.go
```diff
@@ -3668,6 +3674,24 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
+	// Time taken for a task is in nanoseconds, but subtracting two dates yields milliseconds.
+	convertToNanoseconds := time.Millisecond
+	return bson.M{
+		"$addFields": bson.M{
+			TaskDurationKey: bson.M{
+				"$cond": bson.M{
+					"if": bson.M{
+						"$eq": []string{"$" + StatusKey, evergreen.TaskStarted},
+					},
+					"then": bson.M{"$multiply": []interface{}{convertToNanoseconds, bson.M{"$subtract": []interface{}{time.Now(), "$" + StartTimeKey}}}},
```
khelif96, line 3687: `$$NOW` or `$$CLUSTER_TIME` 
https://www.mongodb.com/docs/manual/reference/aggregation-variables/#mongodb-variable-variable.NOW

model/task/task_test.go
```diff
@@ -2678,6 +2678,127 @@ func TestGetTasksByVersionBaseTasks(t *testing.T) {
...
+	t1 := Task{
+		Id:           "t1",
+		Version:      "v1",
+		BuildVariant: "bv_foo",
+		DisplayName:  "displayName_foo",
+		Execution:    0,
+		Status:       evergreen.TaskSucceeded,
+		BaseTask:     BaseTaskInfo{Id: "t1_base", Status: evergreen.TaskSucceeded},
+		StartTime:    time.Date(2022, time.April, 7, 23, 0, 0, 0, time.UTC),
+		TimeTaken:    72839128901, // ~ 1 min
```
khelif96, line 2693: `time.Minute`

khelif96 (changes requested):

graphql/resolvers.go
```diff
@@ -3350,6 +3352,20 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
...
+	}
+	// If status is running, calculate the time elapsed since it started.
+	if utility.FromStringPtr(obj.DisplayStatus) == evergreen.TaskStarted {
+		timeSinceStart := restModel.NewAPIDuration(time.Since(utility.FromTimePtr(obj.StartTime)))
+		return &timeSinceStart, nil
+	}
+	// If any other status, return timeTaken.
+	return &obj.TimeTaken, nil
+}
+
```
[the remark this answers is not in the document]
ablack12: If we do that then we should add to the docs. I'm actually wondering if we should just change the behavior of TimeTaken to be like this though. It makes sense that TimeTaken would be incomplete for Started tasks and if that does break someone's existing use of the API I think they could be expected to case on Status.

graphql/resolvers.go
```diff
@@ -3350,6 +3352,20 @@ func (r *taskResolver) BaseTask(ctx context.Context, obj *restModel.APITask) (*r
...
+	if obj.TaskDuration != 0 {
+		return &obj.TaskDuration, nil
+	}
+	// If status is running, calculate the time elapsed since it started.
+	if utility.FromStringPtr(obj.DisplayStatus) == evergreen.TaskStarted {
+		timeSinceStart := restModel.NewAPIDuration(time.Since(utility.FromTimePtr(obj.StartTime)))
+		return &timeSinceStart, nil
+	}
+	// If any other status, return timeTaken.
+	return &obj.TimeTaken, nil
```
ablack12, line 3366: Would it make sense to return nil in the case that the task hasn't started / time taken is the zero time? (Not sure how this will be used but)

model/task/task.go
```diff
@@ -3668,6 +3674,24 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 	return pipeline
 }
 
+func addTaskDuration() bson.M {
+	// Time taken for a task is in nanoseconds, but subtracting two dates yields milliseconds.
+	convertToNanoseconds := time.Millisecond
+	return bson.M{
+		"$addFields": bson.M{
+			TaskDurationKey: bson.M{
+				"$cond": bson.M{
```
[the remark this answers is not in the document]
ablack12: I think it wouldn't be the worst idea to introduce the idea of a "loop through tasks to set things" for this function, especially as we continue to add things. One for loop may be less expensive than adding aggregation steps (depending on what's needed to be done).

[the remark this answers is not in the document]
ablack12: I also wonder if we start using TimeTaken instead, just on the UI we could check "if TimeTaken exists use that, else use Now - Start" since we'd be checking fields anyway but I'm not sure if there's some issue with that. On the legacy UI I know we have the timer that increments every second which was pretty cool.

ablack12 (changes requested):

graphql/resolvers.go
minnakt: update; removed `TaskDuration` resolver entirely

minnakt:
Commenting for update:

All references of `TaskDuration` have been removed. Instead, the already existing field `TimeTaken` is used to find and sort the running times of started & finished tasks.

The aggregation pipeline now overwrites `TimeTaken` if the task has been started (so that the value can be used for sorting). Additionally, `TimeTaken` is recalculated in `BuildFromService` if a task has been started.

rest/model/task.go
```diff
@@ -264,6 +263,16 @@ func (at *APITask) BuildFromService(t interface{}) error {
 			}
 		}
 
+		if v.TimeTaken != 0 {
+			at.TimeTaken = NewAPIDuration(v.TimeTaken)
+		} else {
+			if v.Status == evergreen.TaskStarted {
+				at.TimeTaken = NewAPIDuration(time.Duration(time.Since(v.StartTime).Nanoseconds()))
+			} else {
+				at.TimeTaken = 0
```
ablack12, line 272: We don't need to specify this, since that's the default value for an integer. (So then 269 can just be else if)

ablack12 (approved):
LGTM with some nits

rest/model/task.go
```diff
@@ -264,6 +263,12 @@ func (at *APITask) BuildFromService(t interface{}) error {
 			}
 		}
 
+		if v.TimeTaken != 0 {
+			at.TimeTaken = NewAPIDuration(v.TimeTaken)
+		} else if v.Status == evergreen.TaskStarted {
+			at.TimeTaken = NewAPIDuration(time.Duration(time.Since(v.StartTime).Nanoseconds()))
+		}
```
khelif96, line 270: Is this nanoseconds conversion needed? time.Since returns a Duration. it seems like you are converting a duration to nanoseconds then back to a duration.

khelif96 (approved):

PR #5564, opened 2022-04-11
EVG-16679 don't error for cycles/loops

ybrill:

### Description 
If someone manages to create a dependency cycle or a loop in the dependency graph we shouldn't let it put the dispatcher out of commission. 
This PR 
1.  replaces a simple graph with a multigraph so we can model loops in the dependency graph
2. logs an unorderable error instead of returning

I'd thought to disable to tasks when we see they're part of a cycle so they won't keep coming up in the distro's queue, but opted not to. My thinking is that today the dispatcher is exclusively concerned with dispatching tasks and reaching into the task lifecycle to disable a task seems like it violates separation of concerns. Some mitigating factors are that [eventually](https://github.com/evergreen-ci/evergreen/blob/c2a10070a720bc20018b7a0bed8b40b44820834e/model/task/task.go#L40), we'll decide they're underwater and disable them. Also, I plan to fix the thing that made the circular dependency, so this is kind of belt and suspenders anyway.

### Testing 
Added/adjusted unit tests.
Manually created a dependency cycle on staging. The dispatcher [logs the warning](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22tasks%20in%20the%20queue%20form%20dependency%20cycle(s)%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1649702400&latest=1649703000&sid=1649702813.542557) instead of erroring.

### DIFF
--- a/model/task_queue_service_dependency.go
+++ b/model/task_queue_service_dependency.go
@@ -227,7 +227,7 @@ func (d *basicCachedDAGDispatcherImpl) rebuild(items []TaskQueueItem) error {
 			return errors.Wrap(err, "topologically sorting the dependency graph")
 		}
 
-		var cycles [][]string
+		cycles := make([][]string, 0, len(unorderableNodes))
 		for _, cycle := range unorderableNodes {
 			cycleIDs := make([]string, 0, len(cycle))
 			for _, node := range cycle {
@@ -235,7 +235,7 @@ func (d *basicCachedDAGDispatcherImpl) rebuild(items []TaskQueueItem) error {
 			}
 			cycles = append(cycles, cycleIDs)
 		}
-		grip.Warning(message.Fields{
+		grip.Error(message.Fields{
 			"dispatcher": DAGDispatcher,
 			"function":   "rebuild",
 			"message":    "tasks in the queue form dependency cycle(s)",
--- a/model/task_queue_service_dependency.go
+++ b/model/task_queue_service_dependency.go
@@ -13,7 +13,7 @@ import (
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 	"gonum.org/v1/gonum/graph"
-	"gonum.org/v1/gonum/graph/simple"
+	"gonum.org/v1/gonum/graph/multi"
 	"gonum.org/v1/gonum/graph/topo"
 )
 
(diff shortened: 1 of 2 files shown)


model/task_queue_service_dependency.go
```diff
@@ -227,16 +213,35 @@ func (d *basicCachedDAGDispatcherImpl) rebuild(items []TaskQueueItem) error {
...
+
+		var cycles [][]string
+		for _, cycle := range unorderableNodes {
+			cycleIDs := make([]string, 0, len(cycle))
+			for _, node := range cycle {
+				cycleIDs = append(cycleIDs, d.nodeItemMap[node.ID()].Id)
+			}
+			cycles = append(cycles, cycleIDs)
+		}
+		grip.Warning(message.Fields{
```
Kimchelly, line 238: Should this be an error so that it shows up in the error rate? This should supposedly not happen since the task lifecycle should error for cyclical graphs.

model/task_queue_service_dependency.go
```diff
@@ -328,7 +337,7 @@ func (d *basicCachedDAGDispatcherImpl) FindNextTask(spec TaskSpec, amiUpdatedTim
 			// However, it won't actually be dispatched to a host if it doesn't satisfy all constraints.
 			item.IsDispatched = true // *TaskQueueItem
 
-			if nextTaskFromDB.StartTime != utility.ZeroTime {
+			if !utility.IsZeroTime(nextTaskFromDB.StartTime) {
```
Kimchelly, line 340: 💯 (no change)

model/task_queue_service_dependency.go
```diff
@@ -227,16 +213,35 @@ func (d *basicCachedDAGDispatcherImpl) rebuild(items []TaskQueueItem) error {
...
+				"message":                    "problem ordering the tasks and associated dependencies within the DirectedGraph",
+				"distro_id":                  d.distroID,
+				"initial_num_taskqueueitems": len(items),
+				"num_task_groups":            len(d.taskGroups),
+			}))
+
+			return errors.Wrap(err, "topologically sorting the dependency graph")
+		}
+
+		var cycles [][]string
```
Kimchelly, line 230: Should this be `make([][]string, 0, len(unorderableNodes))` for consistency with L232?

model/task_queue_service_dependency.go
ybrill: done

model/task_queue_service_dependency.go
ybrill: done

model/task_queue_service_dependency.go
ybrill: thanks!

PR #5566, opened 2022-04-12
EVG-16625 save un-generated tasks/variants in future iterations of generate.tasks

ablack12:
[EVG-16625](https://jira.mongodb.org/browse/EVG-16625)

### Description 
Intermediate step to using transactions (might only need to use them in a couple of places now). Store whether the parser project has already been updated by this generator and only update it if it hasn't been. The Save function actually already handles very well the cases where tasks/variants already exist; don't need to add any special casing to make this more specific.

Also refactored getting existing builds to make this more performant.

A risk I can see with this is: if generatorA runs, fails halfway (but the parser project is updated), and then generatorB runs, it would theoretically generate un-generated tasks/variants _if_ the variant is modified by both generators. That feels pretty specific though and maybe a worthwhile risk.

### Testing 
Added a unit test.

### DIFF
--- a/model/generate.go
+++ b/model/generate.go
@@ -122,8 +122,12 @@ func (g *GeneratedProject) NewVersion(p *Project, pp *ParserProject, v *Version)
 	// Cache project data in maps for quick lookup
 	cachedProject := cacheProjectData(p)
 
+	// We've updated the parser project in a previous iteration of the generator job, so we don't try to update.
+	if utility.StringSliceContains(pp.UpdatedByGenerators, g.TaskID) {
+		return p, pp, v, nil
+	}
 	// Validate generated project against original project.
-	if err := g.validateGeneratedProject(p, cachedProject); err != nil {
+	if err := g.validateGeneratedProject(cachedProject); err != nil {
 		// Return version in this error case for handleError, which checks for a race. We only need to do this in cases where there is a validation check.
 		return nil, pp, v, errors.Wrap(err, "generated project is invalid")
 	}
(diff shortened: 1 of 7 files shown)


ablack12:
The merge function is failing because I added an internally used field to ParserProject, so it doesn't need to be added to merge. I modified that task to only validate mergeable fields, but it will continue to fail until that change is merged in.

model/generate.go
```diff
@@ -168,14 +172,19 @@ func (g *GeneratedProject) Save(ctx context.Context, p *Project, pp *ParserProje
...
+		return nil
+	}
 	updateNum := pp.ConfigUpdateNumber + 1
-	// legacy: most likely a version for which no parser project exists
+	// Legacy: most likely a version for which no parser project exists.
 	if pp.ConfigUpdateNumber < v.ConfigUpdateNumber {
 		updateNum = v.ConfigUpdateNumber + 1
 	}
-
+	pp.UpdatedByGenerators = append(pp.UpdatedByGenerators, taskId)
```
malikchaya2, line 187: Why not do this after line 190 so that it doesn't set it if it errors?

model/project_parser.go
```diff
@@ -84,13 +90,10 @@ type ParserProject struct {
...
 	Loggers            *LoggerConfig              `yaml:"loggers,omitempty" bson:"loggers,omitempty"`
 	CreateTime         time.Time                  `yaml:"create_time,omitempty" bson:"create_time,omitempty"`
-	// List of yamls to merge
-	Include []Include `yaml:"include,omitempty" bson:"include,omitempty"`
 
 	// Matrix code
 	Axes []matrixAxis `yaml:"axes,omitempty" bson:"axes,omitempty"`
-} // End of ParserProject struct
-// Comment above is used by the linter to detect the end of the struct.
+} // End of ParserProject mergeable fields (used by the linter).
```
malikchaya2, line 96: I think this comment is less understandable to me now. Can we maybe change it to (this comment is used by the linter)?

model/generate.go
ablack12: This needs to be upserted along with the rest of the parser project at 188

malikchaya2 (approved):

PR #5567, opened 2022-04-12
EVG-16549: double-check dependency satisfaction during allocation and dispatch

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16549

### Description 
* Check that parent dependencies are met before allocating a container for a task and before dispatching. Since dependency children depend on the latest execution of a task to decide if dependencies are met, the parent dependency could become unsatisfied at any point in the container task pipeline (e.g. the dependency parent is restarted after a container was already allocated for the dependency child, so the dependency child no longer has fully met dependencies).
* Rename a few container task related functions.
* Fix a host termination job test that wasn't resetting the global mock cloud provider state.

### Testing 
Added unit tests.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -497,7 +497,7 @@ func (t *Task) IsFinished() bool {
 // IsHostDispatchable returns true if the task should run on a host and can be
 // dispatched.
 func (t *Task) IsHostDispatchable() bool {
-	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && len(t.ExecutionTasks) == 0 && t.Status == evergreen.TaskUndispatched && t.Activated
+	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && !t.DisplayOnly && t.Status == evergreen.TaskUndispatched && t.Activated
 }
 
 // IsContainerDispatchable returns true if the task should run in a container
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -497,19 +497,54 @@ func (t *Task) IsFinished() bool {
 // IsHostDispatchable returns true if the task should run on a host and can be
 // dispatched.
 func (t *Task) IsHostDispatchable() bool {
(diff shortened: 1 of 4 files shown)


model/task/task.go
```diff
@@ -497,19 +497,54 @@ func (t *Task) IsFinished() bool {
...
 func (t *Task) IsHostDispatchable() bool {
-	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && t.Status == evergreen.TaskUndispatched && t.Activated
+	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && len(t.ExecutionTasks) == 0 && t.Status == evergreen.TaskUndispatched && t.Activated
 }
 
 // IsContainerDispatchable returns true if the task should run in a container
 // and can be dispatched.
 func (t *Task) IsContainerDispatchable() bool {
-	return t.ExecutionPlatform == ExecutionPlatformContainer && t.Status == evergreen.TaskUndispatched && t.Activated && t.ContainerAllocated && t.Priority != evergreen.DisabledTaskPriority
+	if !t.ContainerAllocated {
```
julianedwards, line 506: Can this logic be part of `isContainerScheduled`?

model/task/task.go
```diff
@@ -497,19 +497,54 @@ func (t *Task) IsFinished() bool {
 // IsHostDispatchable returns true if the task should run on a host and can be
 // dispatched.
 func (t *Task) IsHostDispatchable() bool {
-	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && t.Status == evergreen.TaskUndispatched && t.Activated
+	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && len(t.ExecutionTasks) == 0 && t.Status == evergreen.TaskUndispatched && t.Activated
```
julianedwards, line 500: Why not use `!t.DisplayOnly`?

julianedwards (changes requested):

model/task/task.go
Kimchelly: I don't want to roll it into that one since `ShouldAllocateContainer` checks the opposite on L515.

julianedwards (approved):

PR #5569, opened 2022-04-12
EVG-16509, EVG-16718: avoid setting global environment in tests

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16718
https://jira.mongodb.org/browse/EVG-16509

### Description 
Lots of tests were setting the global environment state using `evergreen.SetEnvironment`. This is risky and inadvisable in testing because most tests assume that the Anser DB functions are always available in all tests, but the global Anser DB functions themselves depend on the global environment's mongo client. If `evergreen.SetEnvironment` sets a global environment whose contexts cancels at the end of the test, it'll close the mongo driver when the context cancels and break all later tests in that `go test` execution that rely on the DB. I removed all usages of `evergreen.SetEnvironment` in tests except where it's unavoidable (or `evergreen.SetEnvironment` is itself being tested) so that we can avoid fiddling with fragile global env state in every single test. Every test now should be able to assume that `testutil.Setup` is called in the `init()` block to set up the global environment so the DB is available, and then no other test will overwrite that global environment.

* Remove usages of `evergreen.SetEnvironment` in tests. (This coincidentally fixes [EVG-16509](https://jira.mongodb.org/browse/EVG-16509)).
* Update a bunch of old testing code to use the new `db.CreateCollections` method rather than issuing raw DB commands to create the collection.
* Fix some tests that are not re-runnable because they rely on enqueueing duplicate jobs in Amboy.
    * Add method to drop a DB (mostly for the Amboy DB).
* Do a small refactor for a method to avoid another easily-avoidable `evergreen.GetEnvironment`.

### Testing 
Updated all the unit tests. I ran the tests several times to verify that the REST tests no longer depend on order of test execution and don't unexpectedly modify the global state in a way that breaks other tests.

### DIFF
--- a/db/db_utils.go
+++ b/db/db_utils.go
@@ -199,6 +199,24 @@ func DropAllIndexes(collections ...string) error {
 	return nil
 }
 
+// DropDatabases drops all of the given databases, returning an error immediately
+// if dropping any of the databases fails.
+func DropDatabases(dbs ...string) error {
+	session, _, err := GetGlobalSessionFactory().GetSession()
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	for _, db := range dbs {
+		if err := session.DB(db).DropDatabase(); err != nil {
+			return errors.Wrapf(err, "dropping database '%s'", db)
+		}
+	}
+
+	return nil
+}
+
 // Remove removes one item matching the query from the specified collection.
 func Remove(collection string, query interface{}) error {
 	session, db, err := GetGlobalSessionFactory().GetSession()
--- a/rest/route/host_spawn_test.go
+++ b/rest/route/host_spawn_test.go
@@ -121,10 +121,12 @@ func TestHostStopHandler(t *testing.T) {
 	require.NoError(t, db.ClearCollections(host.Collection, event.SubscriptionsCollection))
 	testutil.DisablePermissionsForTests()
 	defer testutil.EnablePermissionsForTests()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 	h := &hostStopHandler{
-		env: evergreen.GetEnvironment(),
+		env: testutil.NewEnvironment(ctx, t),
 	}
-	ctx := gimlet.AttachUser(context.Background(), &user.DBUser{Id: "user"})
+	ctx = gimlet.AttachUser(ctx, &user.DBUser{Id: "user"})
 
 	hosts := []host.Host{
 		host.Host{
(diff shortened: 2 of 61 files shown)


db/db_utils.go
```diff
@@ -199,6 +199,24 @@ func DropAllIndexes(collections ...string) error {
...
+func DropDatabases(dbs ...string) error {
+	session, _, err := GetGlobalSessionFactory().GetSession()
+	if err != nil {
+		return err
+	}
+	defer session.Close()
+
+	for _, db := range dbs {
+		if err := session.DB(db).DropDatabase(); err != nil {
+			return errors.Wrapf(err, "dropping database '%s'", db)
```
julianedwards, line 213: `DB` ?

rest/route/host_spawn_test.go
```diff
@@ -169,10 +171,12 @@ func TestHostStartHandler(t *testing.T) {
...
 	testutil.DisablePermissionsForTests()
 	defer testutil.EnablePermissionsForTests()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 	h := &hostStartHandler{
-		env: evergreen.GetEnvironment(),
+		env: testutil.NewEnvironment(ctx, t),
 	}
-	ctx := gimlet.AttachUser(context.Background(), &user.DBUser{Id: "user"})
+	ctx = gimlet.AttachUser(context.Background(), &user.DBUser{Id: "user"})
```
julianedwards, line 179: `ctx`

rest/route/host_test.go
```diff
@@ -32,15 +32,23 @@ import (
 
 type HostsChangeStatusesSuite struct {
 	route *hostsChangeStatusesHandler
+	ctx   context.Context
```
julianedwards, line 35: Does this need a context? It doesn't seem like it's ever used.

julianedwards (changes requested):

db/db_utils.go
Kimchelly: Eh, either one works since DB is just an acronym for database. I'm inclined to just leave it.

rest/route/host_spawn_test.go
Kimchelly: Fixed.

rest/route/host_test.go
Kimchelly: Ah yeah, I forgot to change these tests to get rid of the excessive background context depedence in these tests. I changed them all to use cancelled contexts.

PR #5571, opened 2022-04-13
EVG-16100 Bulk creation of annotation

hadjri:
[EVG-16100](https://jira.mongodb.org/browse/EVG-16100)

### Description 
New endpoint allowing user to specify many annotations to create for an associated list of taskid/execution pairs and bulk insert them into the DB
### Testing 
Added unit test


rest/route/service.go
```diff
@@ -188,6 +188,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/tasks/{task_id}").Version(2).Patch().Wrap(requireUser, addProject, editTasks).RouteHandler(makeModifyTaskRoute())
 	app.AddRoute("/tasks/{task_id}/annotations").Version(2).Get().Wrap(requireUser, viewAnnotations).RouteHandler(makeFetchAnnotationsByTask())
 	app.AddRoute("/tasks/{task_id}/annotation").Version(2).Put().Wrap(requireUser, editAnnotations).RouteHandler(makePutAnnotationsByTask())
+	app.AddRoute("/tasks/annotations").Version(2).Patch().Wrap(requireUser, editAnnotations).RouteHandler(makeBulkCreateAnnotations())
```
malikchaya2, line 191: If this is a patch route, then the word "create" here doesn't feel fully accurate, since it can potentially be an update.

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,49 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
 
+// InsertManyAnnotations updates the source for a list of task annotations and their issues and then inserts them into the DB.
+func InsertManyAnnotations(annotations []TaskAnnotation, userDisplayName string) error {
+	source := &Source{
+		Author:    userDisplayName,
+		Time:      time.Now(),
+		Requester: APIRequester,
+	}
+	for i, _ := range annotations {
+		update := bson.M{}
```
malikchaya2, line 215: This seems to be reusing a lot of the work done [here](https://github.com/evergreen-ci/evergreen/pull/5561/files). Do we want to maybe extract that into it's own function so that we can reuse whatever code we can?

malikchaya2: Additionally, there is some conversation there about how the note and metadata should be handled. Please coordinate with that and make sure that the same is done for this route.

model/annotations/task_annotations.go
```diff
@@ -204,6 +204,49 @@ func UpdateAnnotation(a *TaskAnnotation, userDisplayName string) error {
...
+			update[IssuesKey] = annotations[i].Issues
+		}
+		if annotations[i].SuspectedIssues != nil {
+			for i := range annotations[i].SuspectedIssues {
+				annotations[i].SuspectedIssues[i].Source = source
+			}
+			update[SuspectedIssuesKey] = annotations[i].SuspectedIssues
+		}
+
+		_, err := db.Upsert(
```
malikchaya2, line 236: Do we maybe want to do this in one db call instead?

malikchaya2 (changes requested):

model/annotations/task_annotations.go
hadjri: Since the structure of the request [here](https://jira.mongodb.org/browse/EVG-16100?focusedCommentId=4286412&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4286412) is that each task/execution pair is associated with a single annotation, I think the best I can do is bulk upserting the task/execution pairs for each annotation. I've done so

rest/route/annotations.go
```diff
@@ -203,6 +203,81 @@ func (h *annotationByTaskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+func (h *bulkPatchAnnotationHandler) Parse(ctx context.Context, r *http.Request) error {
+	var err error
+	body := utility.NewRequestReader(r)
+	defer body.Close()
+	err = json.NewDecoder(body).Decode(&h.opts)
+	if err != nil {
+		return gimlet.ErrorResponse{
+			Message:    fmt.Sprintf("API error while unmarshalling JSON: '%s'", err.Error()),
```
malikchaya2, line 234: @Kimchelly should this be `unmarshalling JSON: '%s'`, or should it be `API error while unmarshalling JSON: '%s'` since it's specifically an API error?

rest/route/annotations.go
```diff
@@ -203,6 +203,81 @@ func (h *annotationByTaskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+	for _, update := range h.opts.TaskUpdates {
+		for _, t := range update.TaskData {
+			// check if the task exists
+			foundTask, err := task.FindOneIdAndExecution(t.TaskId, t.Execution)
+			if err != nil {
+				return errors.Wrap(err, "error finding task")
```
malikchaya2, line 243: this should be `finding task` to be more consistent with the new style guide.

malikchaya2 (approved):
Please also update the [docs](https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#task-annotations) before merging or create a ticket for it.

rest/route/annotations.go
bynn, line 243: 'finding task with id '%s' and execu...' to abide kim's style guide

rest/route/annotations.go
Kimchelly: This can be simplified down to just `return errors.Wrapf(err "unmarshalling options from JSON request body")` since [`Parse` already returns Bad Request in the Gimlet framework when the error is not of type `gimlet.ErrorResponse`](https://github.com/evergreen-ci/gimlet/blob/b882e01673e7fb37badc0d8f04cac8e308d2f9ec/framework.go#L43-L46).

This error message probably doesn't matter since I'm going to fix all the error messages in this PR after I rebase.

bynn (approved):

PR #5574, opened 2022-04-14
EVG-16732 Move host AMI lookup and project identifier lookup out of task build from service

khelif96:
[EVG-16732](https://jira.mongodb.org/browse/EVG-16732)

### Description 
Move host ami lookup out of build from service

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -4094,7 +4094,7 @@ func (r *taskResolver) Ami(ctx context.Context, obj *restModel.APITask) (*string
 	if err != nil {
 		return nil, InternalServerError.Send(ctx, err.Error())
 	}
-	return obj.Ami, nil
+	return obj.AMI, nil
 }
 
 func (r *taskResolver) Annotation(ctx context.Context, obj *restModel.APITask) (*restModel.APITaskAnnotation, error) {
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -250,6 +250,16 @@ func (r *taskResolver) Project(ctx context.Context, obj *restModel.APITask) (*re
 	return &apiProjectRef, nil
 }
 
+func (r *taskResolver) ProjectIdentifier(ctx context.Context, obj *restModel.APITask) (*string, error) {
+	if utility.FromStringPtr(obj.ProjectId) != "" {
+		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(obj.ProjectId))
+		if err == nil {
+			return utility.ToStringPtr(identifier), nil
+		}
+	}
+	return nil, nil
+}
+
 func (r *taskResolver) AbortInfo(ctx context.Context, at *restModel.APITask) (*AbortInfo, error) {
 	if !at.Aborted {
 		return nil, nil
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -4084,6 +4084,20 @@ func (r *ticketFieldsResolver) ResolutionName(ctx context.Context, obj *thirdpar
 
 func (r *Resolver) TicketFields() TicketFieldsResolver { return &ticketFieldsResolver{r} }
 
+func (r *taskResolver) Ami(ctx context.Context, obj *restModel.APITask) (*string, error) {
+	if obj.HostId != nil {
+		host, err := host.FindOneId(utility.FromStringPtr(obj.HostId))
+		if err != nil {
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding host %s: '%s'", *obj.HostId, err.Error()))
+		}
+		if host != nil {
+			ami := host.GetAMI()
+			return utility.ToStringPtr(ami), nil
+		}
+	}
+	return nil, nil
+}
+
 func (r *taskResolver) Annotation(ctx context.Context, obj *restModel.APITask) (*restModel.APITaskAnnotation, error) {
 	annotation, err := annotations.FindOneByTaskIdAndExecution(*obj.Id, obj.Execution)
 	if err != nil {
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -251,13 +251,8 @@ func (r *taskResolver) Project(ctx context.Context, obj *restModel.APITask) (*re
 }
 
 func (r *taskResolver) ProjectIdentifier(ctx context.Context, obj *restModel.APITask) (*string, error) {
-	if utility.FromStringPtr(obj.ProjectId) != "" {
-		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(obj.ProjectId))
-		if err == nil {
-			return utility.ToStringPtr(identifier), nil
-		}
-	}
-	return nil, nil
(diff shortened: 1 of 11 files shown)


graphql/resolvers.go
```diff
@@ -4084,6 +4084,20 @@ func (r *ticketFieldsResolver) ResolutionName(ctx context.Context, obj *thirdpar
 
 func (r *Resolver) TicketFields() TicketFieldsResolver { return &ticketFieldsResolver{r} }
 
+func (r *taskResolver) Ami(ctx context.Context, obj *restModel.APITask) (*string, error) {
+	if obj.HostId != nil {
```
ablack12, line 4098: This will actually always be not nil since we set it regardless of whether or not it's "", so we should check the empty string specifically.

graphql/resolvers.go
```diff
@@ -4084,6 +4084,20 @@ func (r *ticketFieldsResolver) ResolutionName(ctx context.Context, obj *thirdpar
...
+		host, err := host.FindOneId(utility.FromStringPtr(obj.HostId))
+		if err != nil {
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding host %s: '%s'", *obj.HostId, err.Error()))
```
ablack12, line 4101: FromStringPtr (could just set the hostId variable at the top of the function and use throughout)

graphql/resolvers.go
```diff
@@ -4084,6 +4084,20 @@ func (r *ticketFieldsResolver) ResolutionName(ctx context.Context, obj *thirdpar
...
+		}
+		if host != nil {
+			ami := host.GetAMI()
+			return utility.ToStringPtr(ami), nil
```
ablack12, line 4105: Do we care if ami is "" ? This won't be populated for all kinds of hosts which is why we check "" but if we don't care we can just return utility.ToStringPtr(host.GetAMI) directly

rest/route/task.go
```diff
@@ -94,11 +95,22 @@ func (tgh *taskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
 	}
+	if foundTask.HostId != "" {
+		host, err := host.FindOneId(foundTask.HostId)
+		if err != nil {
+			return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "Database error"))
+		}
+		if host != nil {
+			ami := host.GetAMI()
+			if ami != "" {
+				taskModel.Ami = utility.ToStringPtr(ami)
```
ablack12, line 110: Can we make a function on taskModel like taskModel.GetAMI() or something?

ablack12 (changes requested):

graphql/resolvers.go
```diff
@@ -4084,6 +4089,14 @@ func (r *ticketFieldsResolver) ResolutionName(ctx context.Context, obj *thirdpar
 
 func (r *Resolver) TicketFields() TicketFieldsResolver { return &ticketFieldsResolver{r} }
 
+func (r *taskResolver) Ami(ctx context.Context, obj *restModel.APITask) (*string, error) {
```
ablack12, line 4092: Nit: I kind of feel like I wrote this when I was young and foolish but now it burns my eyes? AMI definitely should be capitalized 😓  if that's an easy refactor to do now

rest/model/task.go
```diff
@@ -330,6 +310,31 @@ func (at *APITask) BuildFromService(t interface{}) error {
 	return nil
 }
 
+func (at *APITask) GetAMI() error {
+	if at.HostId != nil {
```
ablack12, line 314: Reminder that it's not if it isn't nil, if it's the string doesn't equal "" (since we always set HostId to a pointer)

rest/model/task.go
```diff
@@ -330,6 +310,31 @@ func (at *APITask) BuildFromService(t interface{}) error {
...
+		h, err := host.FindOneId(utility.FromStringPtr(at.HostId))
+		if err != nil {
+			return errors.Wrapf(err, "error finding host '%s' for task", utility.FromStringPtr(at.HostId))
```
ablack12, line 317: To follow new error conventions this should drop the word "error"

rest/route/task.go
```diff
@@ -94,11 +95,27 @@ func (tgh *taskGetHandler) Run(ctx context.Context) gimlet.Responder {
 	if err != nil {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "API model error"))
 	}
-
 	err = taskModel.BuildFromService(tgh.url)
 	if err != nil {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "API model error"))
 	}
+	if foundTask.HostId != "" {
```
ablack12, line 102: We can get rid of this if statement now that we're calling GetAMI below

ablack12 (changes requested):

graphql/resolvers.go
khelif96: It isn't the resolver definition is automatically defined by gqlgen. I didn't see any way to change it in the docs.

rest/model/task.go
khelif96: oops thanks!

graphql/resolvers.go
ablack12: Sorry what I meant was, the field _in_ APITask is Ami instead of AMI. if you change it there, it should also fix the resolver.

khelif96: I updated the field in APITask but the resolver field is actually a separate value that is generated by the [`schema.graphql`](https://github.com/evergreen-ci/evergreen/blob/d57ddbc00888b3e21356e65a8c9fde7d54075dbb/graphql/schema.graphql#L1044) file and `gqlgen` even though it maps to the APITask its not the same variable. I renamed it in APITask however.

rest/route/task.go
```diff
@@ -94,12 +94,17 @@ func (tgh *taskGetHandler) Run(ctx context.Context) gimlet.Responder {
...
 	err = taskModel.BuildFromService(tgh.url)
 	if err != nil {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "API model error"))
 	}
 
+	err = taskModel.GetAMI()
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "API model error"))
+	}
+	taskModel.GetProjectIdentifier()
```
ablack12, line 106: I think this is a bigger thing to remove actually 😓  we may need to add this to all the task routes.

I wonder if a prettier way to do this would be to have a BuildWithArgs() function with a variable "includeNonTaskFields", that calls both BuildFromService nad GetProjectIdentifier/GetAMI, and everywhere BuildFromService is used in the rest route we actually call BuildWithArgs (so that GetAMI and GetProjectIdentifier is also included).  That name isn't ideal but idk.

Alternatively you can just look at everywhere BuildFromService is used in rest/route for APITask and add GetProjectIdentifier.

khelif96: Done! I did the alternate way :)

ablack12 (approved):
That works; could you make a ticket for app to shore that up a bit?

PR #5575, opened 2022-04-14
EVG-16307: support activating/deactivating container tasks

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16307

### Description 
This is mostly to do with activation/deactivation but I fixed some tech debt and style issues that I spotted along the way.

* Ensure task activation/deactivation works for both container and host tasks.
* Fix how the container task allocated time so that it's set/unset when the container is allocated/deallocated.
* Make `(task.Task).HostId` field omitempty since it's only relevant to host tasks.
* Include task execution when logging that a task is dispatched to a pod. The task ID is not sufficient to uniquely identify which task execution a pod ran, so it must include the task execution for the event logs. Host tasks currently do a thing where they retroactively set the execution number, but I would prefer to just set the execution when it's dispatched rather than set it when the task execution is being archived.
* Fix a few doc comments.
* Remove outdated Python self receivers in version DB functions (and document those exported methods).
* Fix some random style issues (e.g. snake_case) in testing code.

### Testing 
Added previously-nonexistent tests for existing functions that I modified.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -50,12 +50,12 @@ const (
 	// pending a rerun
 	TaskUnstarted = "unstarted"
 
-	// TaskUndispatched indicates either
+	// TaskUndispatched indicates either:
 	//  1. a task is not scheduled to run (when Task.Activated == false)
 	//  2. a task is scheduled to run (when Task.Activated == true)
 	TaskUndispatched = "undispatched"
 	TaskUnscheduled  = "unscheduled"
-	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
+	// TaskWillRun is a subset of undispatched tasks and is only used in the UI
 	TaskWillRun = "will-run"
 
 	// TaskDispatched indicates that an agent has received the task, but
--- a/globals.go
+++ b/globals.go
@@ -55,7 +55,7 @@ const (
 	//  2. a task is scheduled to run (when Task.Activated == true)
 	TaskUndispatched = "undispatched"
 	TaskUnscheduled  = "unscheduled"
-	// TaskWillRun is a subset of undispatched tasks and is only used in the UI
+	// TaskWillRun is a subset of undispatched tasks and is only used in the UI.
 	TaskWillRun = "will-run"
 
 	// TaskDispatched indicates that an agent has received the task, but
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -128,8 +128,8 @@ type Task struct {
 	// Tags that describe the task
 	Tags []string `bson:"tags,omitempty" json:"tags,omitempty"`
 
-	// The host the task was run on. This value is empty for display tasks
-	HostId string `bson:"host_id" json:"host_id"`
+	// The host the task was run on. This value is only set for host tasks.
+	HostId string `bson:"host_id,omitempty" json:"host_id"`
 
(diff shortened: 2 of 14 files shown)


model/task/task.go
```diff
@@ -1824,41 +1846,38 @@ func resetTaskUpdate(t *Task) bson.M {
...
 		t.ScheduledTime = utility.ZeroTime
-		t.ContainerAllocatedTime = utility.ZeroTime
 		t.FinishTime = utility.ZeroTime
 		t.DependenciesMetTime = utility.ZeroTime
 		t.TimeTaken = 0
 		t.LastHeartbeat = utility.ZeroTime
 		t.Details = apimodels.TaskEndDetail{}
 		t.HasCedarResults = false
 		t.ResetWhenFinished = false
-		t.HostId = ""
```
julianedwards, line 1835: Should this stay here since you are still unsetting it in the update below?

model/task/task.go
```diff
@@ -2239,7 +2258,10 @@ func (t *Task) Insert() error {
 	return db.Insert(Collection, t)
 }
 
-// Inserts the task into the old_tasks collection
+// Archive modifies the currente execution of the task so that it is no longer
```
julianedwards, line 2261: `current`

model/task/task.go
```diff
@@ -1071,6 +1088,8 @@ func (t *Task) MarkAsContainerDeallocated(ctx context.Context, env evergreen.Env
 	t.ContainerAllocated = false
 	t.DispatchTime = utility.ZeroTime
 	t.LastHeartbeat = utility.ZeroTime
+	t.ContainerAllocatedTime = time.Time{}
```
julianedwards, line 1091: `utility.ZeroTime`?

julianedwards (changes requested):

globals.go
```diff
@@ -50,12 +50,12 @@ const (
...
 	TaskUnstarted = "unstarted"
 
-	// TaskUndispatched indicates either
+	// TaskUndispatched indicates either:
 	//  1. a task is not scheduled to run (when Task.Activated == false)
 	//  2. a task is scheduled to run (when Task.Activated == true)
 	TaskUndispatched = "undispatched"
 	TaskUnscheduled  = "unscheduled"
-	// TaskWillRun is a subset of TaskUndispatched and is only used in the UI
+	// TaskWillRun is a subset of undispatched tasks and is only used in the UI
```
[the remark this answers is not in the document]
Kimchelly: Fixed.

model/task/task.go
Kimchelly: This one is kind of weird because this is an optional time field (probaby all of them should be optional, but nothing I can do about them now). Go will only omitempty if it's the Go zero time, not Unix zero time.

I guess we could just use the Unix zero time for everything since the precedence was set, but I don't have strong feelings either way.

model/task/task.go
Kimchelly: Fixed.

model/task/task.go
Kimchelly: This is a duplicate unset, it's already unset on L1821.

julianedwards (approved):

PR #5578, opened 2022-04-14
EVG-16429 Add Container to tasks

hadjri:
[EVG-16429](https://jira.mongodb.org/browse/EVG-16429)

### Description 
Added container field to Task, checking if it should be set by inspecting the elements of its RunOn field


model/lifecycle.go
```diff
@@ -1297,6 +1267,18 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 		// TODO (EVG-16169): set ExecutionPlatform based on whether or not the task
 		// is running in a host or container.
 		ExecutionPlatform: task.ExecutionPlatformHost,
+		RunOnContainer:    utility.FromBoolPtr(buildVarTask.RunOnContainer),
```
Kimchelly, line 1270: It seems like this field is redundant since the `ExecutionPlatform` field above this has the same meaning. I would just set `ExecutionPlatform` based on whether RunOn is for a container or distros (which also fulfills EVG-16169).

model/lifecycle.go
```diff
@@ -1297,6 +1267,18 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+	}
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting service flags")
+	}
+	if !flags.ContainerConfigurationsDisabled && utility.FromBoolPtr(buildVarTask.RunOnContainer) {
```
Kimchelly, line 1276: I think if container configurations are disabled, this should error and fail to allow the task to be created. Otherwise, if they're trying to use containers and it's disabled by this flag, they'll end up with an invalid task associated with a nonexistent distro which can't run anyways.

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 	return t, nil
 }
 
+func generateDistros(id string, buildVarTask BuildVariantTaskUnit, buildVariant *BuildVariant, project *Project, v *Version) (string, []string) {
```
Kimchelly, line 1299: `getDistrosFromRunOn`?

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+		}
+
+	} else if len(buildVariant.RunOn) > 0 {
+		distroID = buildVariant.RunOn[0]
+
+		if len(buildVariant.RunOn) > 1 {
+			distroAliases = append(distroAliases, buildVariant.RunOn[1:]...)
+		}
+	} else {
+		grip.Warning(message.Fields{
```
Kimchelly, line 1319: This seems like a weird case that should also be an error (I know the code is originally written like this, but it probably shouldn't do this unless there's a special reason). Since a task like this won't be able to run anyways, can this just return an error rather than log a warning and continue?

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 	return t, nil
 }
 
+func generateDistros(id string, buildVarTask BuildVariantTaskUnit, buildVariant *BuildVariant, project *Project, v *Version) (string, []string) {
+	var (
+		distroID      string
+		distroAliases []string
+	)
+
+	if len(buildVarTask.RunOn) > 0 {
```
Kimchelly, line 1305: This if-elseif-else might be more easily readable if it went something like:
```go
if len(buildVarTask.RunOn) > 0 {
    // Set distro ID and aliases...
    return distroID, distroAliases
else if len(buildVariant.RunOn) > 0 {
    // Set distro ID and aliases...
    return distroID, distroAliases
}

return errors.New("task does not have a distro to run on")
```
It would also remove the need to declare `distroID` and `distroAliases` at the top of the function.

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+			"variant":   buildVariant.Name,
+			"project":   project.Identifier,
+			"version":   v.Revision,
+			"requester": v.Requester,
+		})
+	}
+	return distroID, distroAliases
+}
+
+func generateContainer(id string, buildVarTask BuildVariantTaskUnit, buildVariant *BuildVariant, project *Project, v *Version) string {
```
Kimchelly, line 1331: `getContainerFromRunOn`?

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+	var container string
+
+	if len(buildVarTask.RunOn) > 0 {
+		container = buildVarTask.RunOn[0]
+	} else if len(buildVariant.RunOn) > 0 {
+		container = buildVariant.RunOn[0]
```
Kimchelly, line 1337: It seems like this will be weird unless users are disallowed from passing multiple containers or a mix of containers/distros to `RunOn`. Is the project validation for RunOn going to be changed to make sure this is guaranteed to have 1 element in it?

model/project.go
```diff
@@ -136,7 +136,8 @@ type BuildVariantTaskUnit struct {
 	// with BatchTime and CronBatchTime being mutually exclusive.
 	CronBatchTime string `yaml:"cron,omitempty" bson:"cron,omitempty"`
 	// If Activate is set to false, then we don't initially activate the task.
-	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
+	Activate       *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
+	RunOnContainer *bool `yaml:"run_on_container,omitempty" bson:"run_on_container"`
```
Kimchelly, line 140: Rather than having a user manually set this field, would be it better to have it automatically detect the container vs. distro configuration based on what they've passed to `RunOn`? Otherwise, they'll have to mnnually set/unset this if they change `RunOn` between distros and containers.

For example, if they pass one name to `RunOn`, it checks if there's a matching container name and uses that if possible, but falls back to using a distro otherwise. If there are multiple names, they're assumed to refer to distros.

model/project.go
```diff
@@ -1165,6 +1170,26 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
 	return nil
 }
 
+func FindContainerConfigurationForTask(t task.Task) (*Container, error) {
```
Kimchelly, line 1173: * `FindContainerFromProject`?
* Documentation for exported methods.
* Is this used anywhere? Is it supposed to be used when resolving whether it's running on a container vs. distro?

model/project_parser.go
```diff
@@ -987,6 +990,19 @@ func evaluateTaskUnits(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluator, v
 	return tasks, groups, evalErrs
 }
 
+func shouldRunOnContainer(task ProjectTask, containers []Container) bool {
```
Kimchelly, line 993: Does this take into account containers from the project settings?

Instead of adding another config setting for users, can we instead resolve the running on distro vs. containers question in project validation + `createOneTask`?

service/rest_task.go
```diff
@@ -36,6 +36,7 @@ type RestTask struct {
 	Activated           bool                  `json:"activated"`
 	BuildId             string                `json:"build_id"`
 	DistroId            string                `json:"distro"`
+	Container           string                `json:"container"`
```
Kimchelly, line 39: Should this be omitempty for the JSON tag? It won't be set for regular host tasks.

service/stats.go
```diff
@@ -52,6 +52,7 @@ type UITask struct {
 	Status                 string    `json:"status"`
 	Host                   string    `json:"host"`
 	Distro                 string    `json:"distro"`
+	Container              string    `json:"container"`
```
Kimchelly, line 55: omitempty?

service/task.go
```diff
@@ -37,6 +37,7 @@ type uiTaskData struct {
 	Revision             string                  `json:"gitspec"`
 	BuildVariant         string                  `json:"build_variant"`
 	Distro               string                  `json:"distro"`
+	Container            string                  `json:"container"`
```
Kimchelly, line 40: omitempty?

model/task/task.go
```diff
@@ -108,11 +108,13 @@ type Task struct {
 
 	BuildId                 string       `bson:"build_id" json:"build_id"`
 	DistroId                string       `bson:"distro" json:"distro"`
+	Container               string       `bson:"container" json:"container"`
```
Kimchelly, line 111: Should the BSON/JSON tags have omitemtpy since it won't be relevant for non-container tasks?

model/lifecycle.go
```diff
@@ -1297,6 +1267,18 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting service flags")
+	}
+	if !flags.ContainerConfigurationsDisabled && utility.FromBoolPtr(buildVarTask.RunOnContainer) {
+		t.Container = generateContainer(id, buildVarTask, buildVariant, project, v)
+	} else {
+		distroID, distroAliases := generateDistros(id, buildVarTask, buildVariant, project, v)
+		t.DistroId = distroID
+		t.DistroAliases = distroAliases
```
Kimchelly, line 1281: Can we test the `createOneTask` changes?

model/project.go
```diff
@@ -1165,6 +1170,26 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
...
+	v, err := VersionFindOneId(t.Version)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error finding version %s", t.Version)
```
Kimchelly, line 1176: `errors.Wrapf(err, "finding version '%s'", t.Version)`

model/project.go
```diff
@@ -1165,6 +1170,26 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
...
+	}
+	if v == nil {
+		return nil, errors.Errorf("unable to find version %s", t.Version)
```
Kimchelly, line 1179: `errors.Errorf("version '%s' not found", t.Version)`

model/project.go
```diff
@@ -1165,6 +1170,26 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
...
+	}
+	projectInfo, err := LoadProjectForVersion(v, t.Project, false)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting project for version %s", t.Version)
```
Kimchelly, line 1183: `errors.Wrapf(err, "getting project for version '%s'", t.Version)`

model/project_parser.go
```diff
@@ -950,6 +952,7 @@ func evaluateTaskUnits(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluator, v
 		if strings.Contains(strings.TrimSpace(pt.Name), " ") {
 			evalErrs = append(evalErrs, errors.Errorf("spaces are not allowed in task names ('%s')", pt.Name))
 		}
+		t.RunOnContainer = utility.ToBoolPtr(shouldRunOnContainer(t, containers))
```
Kimchelly, line 955: Should there be tests for this change?

Kimchelly (changes requested):

model/project.go
hadjri: This wasnt intended to be a manually set field, but something that gets derived and set during the parsing process. However I have changed it to only setting ExecutionPlatform based off inspection of RunOn now, so this property has been removed.

model/project.go
hadjri: It is not currently used since there is no use right now for actually applying the container configs defined in YAML. I just added this since finding a container configuration from a task's Container string is something that will be required in the future.

model/project_parser.go
hadjri: Containers themselves arent defined on project settings, just the container resource aliases that are referenced by 'size' on containers in the YAML.  They will be picked up from the project ref when they are needed later.

model/lifecycle.go
hadjri: I have added it as a warning but unsure if it is necessary to error the validation on this case.

model/lifecycle.go
```diff
@@ -1294,9 +1263,26 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 		CommitQueueMerge:    buildVarTask.CommitQueueMerge,
 		IsGithubCheck:       isGithubCheck,
 		DisplayTaskId:       utility.ToStringPtr(""), // this will be overridden if the task is an execution task
-		// TODO (EVG-16169): set ExecutionPlatform based on whether or not the task
-		// is running in a host or container.
-		ExecutionPlatform: task.ExecutionPlatformHost,
+	}
+	// is running in a host or container.
+	t.ExecutionPlatform = shouldRunOnContainer(buildVarTask.RunOn, buildVariant.RunOn, project.Containers)
+	flags, err := evergreen.GetServiceFlags()
```
Kimchelly, line 1269: Move this into the `if` block since we only use the flag if it's a container task.

model/lifecycle.go
```diff
@@ -1294,9 +1263,26 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+	if err != nil {
+		return nil, errors.Wrap(err, "getting service flags")
+	}
+	if t.ExecutionPlatform == task.ExecutionPlatformContainer {
```
Kimchelly, line 1273: `if t.IsContainerTask() {`

model/lifecycle.go
```diff
@@ -1294,9 +1263,26 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+		if flags.ContainerConfigurationsDisabled {
+			return nil, errors.Errorf("container configurations are disabled; task '%s' cannot run", t.DisplayName)
+		} else {
```
Kimchelly, line 1276: It's more idiomatic in Go to omit the `else` when the `if` block already returns.

model/lifecycle.go
```diff
@@ -1314,6 +1300,71 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+		}
+		return distroID, distroAliases, nil
+	} else if len(buildVariant.RunOn) > 0 {
+		distroID := buildVariant.RunOn[0]
+		if len(buildVariant.RunOn) > 1 {
+			distroAliases = buildVariant.RunOn[1:]
+		}
+		return distroID, distroAliases, nil
+	}
+	grip.Error(message.Fields{
```
Kimchelly, line 1318: I don't think we need to log this since it's already returned in the error.

model/lifecycle.go
```diff
@@ -1314,6 +1300,71 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
 	return t, nil
 }
 
+func getDistrosFromRunOn(id string, buildVarTask BuildVariantTaskUnit, buildVariant *BuildVariant, project *Project, v *Version) (string, []string, error) {
+	distroAliases := []string{}
```
Kimchelly, line 1304: move this into the if/elseif.

validator/project_validator.go
```diff
@@ -754,18 +762,27 @@ func ensureReferentialIntegrity(project *model.Project, containerNameMap map[str
...
-					Level:   Error,
-				},
-			)
-		}
+		errs = append(errs, addRunOnError(runOnHasDistro, runOnHasContainer, buildVariant.RunOn)...)
 	}
 	return errs
 }
 
+func addRunOnError(runOnHasDistro, runOnHasContainer bool, runOn []string) []ValidationError {
```
Kimchelly, line 770: `checkRunOn`?

validator/project_validator.go
```diff
@@ -754,18 +762,27 @@ func ensureReferentialIntegrity(project *model.Project, containerNameMap map[str
...
+	if runOnHasContainer && runOnHasDistro {
+		return []ValidationError{{
+			Message: "run_on cannot contain a mixture of containers and distros",
+			Level:   Error,
+		}}
+
+	} else if runOnHasContainer && len(runOn) > 1 {
+		return []ValidationError{{
+			Message: fmt.Sprint("only one container can be used from run_on; the first container in the list will be used"),
```
Kimchelly, line 779: The `fmt.Sprint` shouldn't be necessary since it's just a string.

model/lifecycle.go
Kimchelly: Warning seems fine since the task could presumably still run even if they give extra containers.

model/lifecycle.go
```diff
@@ -1314,6 +1296,58 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+
+func generateContainer(id string, buildVarTask BuildVariantTaskUnit, buildVariant *BuildVariant, project *Project, v *Version) string {
+	var container string
+
+	if len(buildVarTask.RunOn) > 0 {
+		container = buildVarTask.RunOn[0]
+	} else if len(buildVariant.RunOn) > 0 {
+		container = buildVariant.RunOn[0]
+	} else {
+		grip.Warning(message.Fields{
```
Kimchelly, line 1356: Return error in this case, since the task will be unrunnable without any container.

model/project.go
Kimchelly: I'm fine with leave it in then. Can we still write unit tests for it?

model/project.go
```diff
@@ -1165,6 +1165,27 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
 	return nil
 }
 
+// FindContainerFromProject finds the container configuration associated with a given task's Container field.
+func FindContainerFromProject(t task.Task) (*Container, error) {
+	v, err := VersionFindOneId(t.Version)
+	if err != nil {
+		return nil, errors.Wrapf(err, "finding version %s", t.Version)
```
Kimchelly, line 1172: Need single quotes around the string. We do this specially for strings because they could potentially be empty, which makes the error message weird.

model/project.go
```diff
@@ -1165,6 +1165,27 @@ func (p *Project) FindTaskGroup(name string) *TaskGroup {
...
+	}
+	if v == nil {
+		return nil, errors.Errorf("version '%s' not found", t.Version)
+	}
+	projectInfo, err := LoadProjectForVersion(v, t.Project, false)
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting project for version %s", t.Version)
```
Kimchelly, line 1179: Single quotes around string.

model/lifecycle.go
```diff
@@ -1314,6 +1300,71 @@ func createOneTask(id string, buildVarTask BuildVariantTaskUnit, project *Projec
...
+	})
+	return "", nil, errors.Errorf("task '%s' is not runnable as there is no distro specified", id)
+}
+
+func shouldRunOnContainer(taskRunOn, buildVariantRunOn []string, containers []Container) task.ExecutionPlatform {
+	containerNameMap := map[string]bool{}
+	for _, container := range containers {
+		containerNameMap[container.Name] = true
+	}
+	runOn := []string{}
```
Kimchelly, line 1334: `var runOn []string`
Since `runOn` is assigned to in both the if and else block, Go will complain that `runOn` is initialized to a value but never used because it's unconditionally overwritten.

Kimchelly (approved):

PR #5580, opened 2022-04-15
EVG-16735 fetch github status field to calculate check status

ablack12:
[EVG-16735](https://jira.mongodb.org/browse/EVG-16735)

### Description 
The github check field got lost in a refactor, meaning that UpdateVersionGithubStatus wasn't using the right field.

### Testing 
Added unit test (which failed pre-changes).

### DIFF
--- a/model/task_lifecycle_test.go
+++ b/model/task_lifecycle_test.go
@@ -772,6 +772,42 @@ func TestUpdateBuildStatusForTaskReset(t *testing.T) {
 	assert.Equal(t, evergreen.VersionStarted, data.Status)
 }
 
+func TestUpdateVersionStatusForGithubChecks(t *testing.T) {
+	require.NoError(t, db.ClearCollections(task.Collection, build.Collection, VersionCollection, event.AllLogCollection))
+	b1 := build.Build{
+		Id:                "b1",
+		Status:            evergreen.BuildStarted,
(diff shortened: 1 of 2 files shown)


model/task_lifecycle_test.go
```diff
@@ -772,6 +772,42 @@ func TestUpdateBuildStatusForTaskReset(t *testing.T) {
...
+
+	assert.NoError(t, b1.Insert())
+	assert.NoError(t, b2.Insert())
+	v1 := Version{
+		Id:     "v1",
+		Status: evergreen.VersionStarted,
+	}
+	assert.NoError(t, v1.Insert())
+	versionStatus, err := UpdateVersionStatus(&v1)
+	assert.NoError(t, err)
```
Kimchelly, line 801: Should this `assert.NoError` (and the other `NoError` checks) be `require` since the test probably can't continue anyways if it fails to insert or `UpdateVersionStatus` fails?

Kimchelly (changes requested):

model/task_lifecycle_test.go
ablack12: I typically prefer only using require for things that would then result in a nil pointer dereference.

Kimchelly (approved):

PR #5586, opened 2022-04-18
EVG-16552 only check dependencies if not modifying the whole version

ablack12:
[EVG-16552](https://jira.mongodb.org/browse/EVG-16552)

### Description 
Since dependencies only exist within versions, we don't need to check for dependencies if the whole thing is being updated. This will hopefully be enough of an improvement that we don't need a job to do this. It's possible that we should start relying more on the job to unschedule blocked dependencies rather than doing them in these kinds of calls ever, since it's expensive.

### Testing 
Verified existing tests work.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -74,7 +74,7 @@ func SetVersionActivation(versionId string, active bool, caller string) error {
 		return errors.Wrapf(err, "setting activation for builds in version '%s'", versionId)
 	}
 
-	return errors.Wrapf(setTaskActivationForBuilds(buildIDs, active, nil, caller),
+	return errors.Wrapf(setTaskActivationForBuilds(buildIDs, active, false, nil, caller),
 		"setting activation for tasks in version '%s'", versionId)
 }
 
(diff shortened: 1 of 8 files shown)


model/task/task.go
```diff
@@ -1412,7 +1418,7 @@ func ActivateTasksByIdsWithDependencies(ids []string, caller string) error {
 		return errors.Wrap(err, "getting recursive dependencies")
 	}
 
-	if err = ActivateTasks(append(tasks, dependOn...), time.Now(), caller); err != nil {
+	if err = ActivateTasks(append(tasks, dependOn...), time.Now(), true, caller); err != nil {
```
hadjri, line 1421: Should we put a bool parameter in this function? In case we want to activate tasks by ids without dependencies at some point

model/task/task.go
```diff
@@ -1560,10 +1566,10 @@ func (t *Task) DeactivateTask(caller string) error {
 	t.ScheduledTime = utility.ZeroTime
 	t.ContainerAllocatedTime = utility.ZeroTime
 
-	return DeactivateTasks([]Task{*t}, caller)
+	return DeactivateTasks([]Task{*t}, true, caller)
```
hadjri, line 1569: Same as above

model/lifecycle.go
```diff
@@ -328,23 +333,16 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
-		if err != nil {
-			return errors.WithStack(err)
-		}
-		catcher := grip.NewBasicCatcher()
-		for _, t := range aborted {
-			catcher.Add(t.SetResetWhenFinished())
-		}
-		if catcher.HasErrors() {
-			return catcher.Resolve()
+	if abortInProgress && len(finishedTasks) < len(taskIds) {
```
hadjri, line 336: Can we add a comment above this line?

hadjri (changes requested):

hadjri (approved):

PR #5587, opened 2022-04-18
EVG-16737 deprecated APITask.BuildFromService in favor of BuildFromArgs.

khelif96:
[EVG-16737](https://jira.mongodb.org/browse/EVG-16737)

### Description 
Created an `APITask.BuildFromArgs` function which implements `APITask.BuildFromService` but allows for conditional db queries to populate certain fields.

Replaced BuildFromService in most places and added a deprecated comment to it.

This prevents some unnecessary db queries from happening despite some of their fields not being used. 

This has an impact on db queries since we were previous doing 2 extra queries for each APITask we build. And in some cases that could be 4000+ times.
As well as load times when returning multiple tasks.
eg. 
Loading 4000+ tasks 
before: 
10.51s
after:
3.7 s
### Testing 
Verified project_identifier field is populated in api.
image


graphql/resolvers.go
```diff
@@ -250,6 +250,11 @@ func (r *taskResolver) Project(ctx context.Context, obj *restModel.APITask) (*re
 	return &apiProjectRef, nil
 }
 
+func (r *taskResolver) ProjectIdentifier(ctx context.Context, obj *restModel.APITask) (*string, error) {
```
malikchaya2, line 253: Why don't we do this in buildFromService so that we don't need this resolver?

khelif96: This performs a db query to resolve. We do not always need this field. So it becomes a wasted db query.
BuildFromService should not really run any db queries anyways it is meant to convert one struct into another. By having a db query it slows that conversion down dramatically.

malikchaya2 (approved):

graphql/resolvers.go
malikchaya2: wouldn't buildFromService have it without needing a query?

malikchaya2: oh, never mind. this is the task resolver.

rest/route/build_task.go
```diff
@@ -99,15 +99,12 @@ func (tbh *tasksByBuildHandler) Run(ctx context.Context) gimlet.Responder {
 	for i := range tasks {
 		taskModel := &model.APITask{}
 
-		if err = taskModel.BuildFromService(&tasks[i]); err != nil {
+		if err = taskModel.BuildFromArgs(&tasks[i], &model.APITaskArgs{
+			IncludeAMI: true, IncludeArtifacts: true, IncludeProjectIdentifier: true,
```
ablack12, line 103: Style nit: can you break this into separate lines?

rest/route/task.go
```diff
@@ -90,12 +90,11 @@ func (tgh *taskGetHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 
 	taskModel := &model.APITask{}
-	err = taskModel.BuildFromService(foundTask)
+	err = taskModel.BuildFromArgs(foundTask, &model.APITaskArgs{IncludeProjectIdentifier: true, IncludeAMI: true, IncludeArtifacts: true})
```
ablack12, line 93: style nit separate lines

rest/route/build_task.go
```diff
@@ -99,15 +99,12 @@ func (tbh *tasksByBuildHandler) Run(ctx context.Context) gimlet.Responder {
...
+		}); err != nil {
 			return gimlet.MakeJSONErrorResponder(err)
 		}
-
-		if err = taskModel.GetArtifacts(); err != nil {
-			return gimlet.MakeJSONErrorResponder(err)
-		}
-
-		if err = taskModel.BuildFromService(tbh.url); err != nil {
+		if err = taskModel.BuildFromArgs(tbh.url, nil); err != nil {
```
ablack12, line 107: BuildFromService taking in a string at all was kind of hacky. I think now that we have BuildFromArgs it's cleaner to add a parameter for adding logs, and just move 
```
case string:
		ll := LogLinks{
			AllLogLink:    utility.ToStringPtr(fmt.Sprintf(TaskLogLinkFormat, v, utility.FromStringPtr(at.Id), at.Execution, "ALL")),
			TaskLogLink:   utility.ToStringPtr(fmt.Sprintf(TaskLogLinkFormat, v, utility.FromStringPtr(at.Id), at.Execution, "T")),
			AgentLogLink:  utility.ToStringPtr(fmt.Sprintf(TaskLogLinkFormat, v, utility.FromStringPtr(at.Id), at.Execution, "E")),
			SystemLogLink: utility.ToStringPtr(fmt.Sprintf(TaskLogLinkFormat, v, utility.FromStringPtr(at.Id), at.Execution, "S")),
			EventLogLink:  utility.ToStringPtr(fmt.Sprintf(EventLogLinkFormat, v, utility.FromStringPtr(at.Id))),
		}
		at.Logs = ll
```
into that (I think this is especially true if we're going to mark BuildFromService as "deprecated").

rest/route/task_project.go
```diff
@@ -100,11 +100,12 @@ func (tph *tasksByProjectHandler) Run(ctx context.Context) gimlet.Responder {
 
 	for _, t := range tasks {
 		taskModel := &model.APITask{}
-		err = taskModel.BuildFromService(&t)
+		err = taskModel.BuildFromArgs(&t, &model.APITaskArgs{IncludeAMI: true, IncludeProjectIdentifier: true})
 		if err != nil {
 			return gimlet.MakeJSONErrorResponder(err)
 		}
-		err = taskModel.BuildFromService(tph.url)
+		taskModel.GetProjectIdentifier()
```
ablack12, line 107: We shouldn't need this anymore

rest/model/task.go
```diff
@@ -173,20 +172,19 @@ func (ad *APIOomTrackerInfo) ToService() (interface{}, error) {
...
 		}
-		if err := at.PreviousExecutions[i].GetArtifacts(); err != nil {
-			return errors.Wrap(err, "failed to fetch artifacts for previous executions")
-		}
 	}
 
 	return nil
 }
 
+// Deprecated: use BuildFromArgs instead
```
ablack12, line 186: Can this be a full sentence like "BuildFromArgs should be used instead to add fields that aren't from the task collection" or something.

rest/model/task.go
```diff
@@ -173,20 +172,19 @@ func (ad *APIOomTrackerInfo) ToService() (interface{}, error) {
 func (at *APITask) BuildPreviousExecutions(tasks []task.Task, url string) error {
 	at.PreviousExecutions = make([]APITask, len(tasks))
 	for i := range at.PreviousExecutions {
-		if err := at.PreviousExecutions[i].BuildFromService(&tasks[i]); err != nil {
+		if err := at.PreviousExecutions[i].BuildFromArgs(&tasks[i], &APITaskArgs{IncludeProjectIdentifier: true, IncludeAMI: true, IncludeArtifacts: true}); err != nil {
 			return errors.Wrap(err, "error marshalling previous execution")
 		}
 		if err := at.PreviousExecutions[i].BuildFromService(url); err != nil {
```
ablack12, line 178: Shouldn't be using BuildFromService anymore

rest/model/task.go
```diff
@@ -330,6 +309,61 @@ func (at *APITask) BuildFromService(t interface{}) error {
...
+	IncludeAMI               bool
+	IncludeArtifacts         bool
+}
+
+func (at *APITask) BuildFromArgs(t interface{}, args *APITaskArgs) error {
+	err := at.BuildFromService(t)
+	if err != nil {
+		return err
+	}
+	if args != nil {
```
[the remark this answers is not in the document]
khelif96: the reason i did this was to avoid passing in an empty struct if we don't need to use any args.

rest/route/build_task.go
khelif96: I moved this over and move all usages to use the buildFromArgs version. There are still some tests that asserted on this value in buildFromService and seemed difficult to move move over. So i left this functionality in buildFromService for those tests.

PR #5591, opened 2022-04-19
EVG-13451 Support modern ssh key types

bynn:
[EVG-13451](https://jira.mongodb.org/browse/EVG-13451)

### Description 
add 2 extra types of ssh keys that we allow 

### Testing 
unit test 

image

created host on staging using new key type
image

spruce

https://github.com/evergreen-ci/spruce/pull/1226

### DIFF
--- a/cloud/spawn.go
+++ b/cloud/spawn.go
@@ -70,7 +70,7 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
 
 	// validate public key
 	if err = evergreen.ValidateSSHKey(so.PublicKey); err != nil {
-		return errors.Wrap(err, "Invalid spawn options: ")
+		return errors.Wrap(err, "Invalid spawn options")
 	}
 
 	sections := strings.Split(so.PublicKey, " ")
--- a/cloud/spawn.go
+++ b/cloud/spawn.go
@@ -23,6 +23,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Valid key types.
+const (
+	rsa     = "ssh-rsa"
+	dss     = "ssh-dss"
+	ed25519 = "ssh-ed25519"
+	ecdsa   = "ecdsa-sha2-nistp256"
+)
+
 // Options holds the required parameters for spawning a host.
 type SpawnOptions struct {
 	DistroId              string
@@ -69,14 +77,8 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
 	}
 
 	// validate public key
-	rsa := "ssh-rsa"
-	dss := "ssh-dss"
-	isRSA := strings.HasPrefix(so.PublicKey, rsa)
-	isDSS := strings.HasPrefix(so.PublicKey, dss)
-	if !isRSA && !isDSS {
-		return errors.Errorf("Invalid spawn options: "+
-			"either an invalid Evergreen-managed key name has been provided,"+
-			"or the key value does not start with %s or %s", rsa, dss)
+	if err = ValidateSSHKey(so.PublicKey); err != nil {
+		return errors.Wrap(err, "Invalid spawn options: ")
 	}
 
 	sections := strings.Split(so.PublicKey, " ")
@@ -94,6 +96,18 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
 	}
 
 	return nil
+}
+
+// ValidateSSHKey errors if the given key does not start with one of the allowed prefixes.
+func ValidateSSHKey(key string) error {
+	validKeys := []string{rsa, dss, ed25519, ecdsa}
+	for _, prefix := range validKeys {
+		if strings.HasPrefix(key, prefix) {
+			return nil
+		}
+	}
+	return errors.Errorf("either an invalid Evergreen-managed key name has been provided,"+
+		"or the key value does not start with one of these options: '%s'", validKeys)
 }
 
 func checkSpawnHostLimitExceeded(numCurrentHosts int, settings *evergreen.Settings) error {
--- a/cloud/spawn.go
+++ b/cloud/spawn.go
@@ -23,14 +23,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Valid key types.
-const (
-	rsa     = "ssh-rsa"
-	dss     = "ssh-dss"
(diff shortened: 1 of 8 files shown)


cloud/spawn.go
```diff
@@ -96,6 +98,18 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
...
 
+// ValidateSSHKey errors if the given key does not start with one of the allowed prefixes.
+func ValidateSSHKey(key string) error {
+	validKeys := []string{rsa, dss, ed25519, ecdsa}
+	for _, prefix := range validKeys {
+		if strings.HasPrefix(key, prefix) {
+			return nil
+		}
+	}
+	return errors.Errorf("either an invalid Evergreen-managed key name has been provided,"+
```
Kimchelly, line 109: nit: Need a space after the comma.

cloud/spawn.go
```diff
@@ -96,6 +98,18 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
...
+		"or the key value does not start with one of these options: '%s'", validKeys)
```
Kimchelly, line 110: Suggestion: `or the key value is not one of the valid types: %s`?

For slices, we don't usually need the single quotes around the string since empty/nil slices are formatted as `[]` with `%s`, so the concern about potentially empty strings doesn't apply here.

public/static/partials/user_host_options.html
```diff
@@ -70,7 +70,7 @@
       Please enter a unique key name for your new key<br />
     </span>
     <span class="semi-muted invalid" ng-show="spawnHostForm.pubkey.$dirty && (spawnHostForm.pubkey.$error.required || spawnHostForm.pubkey.$error.keyBaseValid)">
-      Key is invalid. It must begin with 'ssh-rsa' or 'ssh-dss'. Ensure you have copied the public half of the key.<br />
+      Key is invalid. It must begin with 'ssh-rsa' or 'ssh-dss' or 'ssh-ed25519' or 'ecdsa-sha2-nistp256'. Ensure you have copied the public half of the key.<br />
```
Kimchelly, line 73: nit: `'ssh-rsa', 'ssh-dss', 'ssh-ed25519', or 'ecdsa-sha2-nistp256'`

cloud/spawn.go
```diff
@@ -23,6 +23,14 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Valid key types.
+const (
+	rsa     = "ssh-rsa"
```
Kimchelly, line 28: nit: can these be called `pubKeyRSA`, `pubKeyDSS`, etc. since it's basically an enum of strings?

Kimchelly (changes requested):
I think [this check in the CLI](https://github.com/evergreen-ci/evergreen/blob/17f9e61403adf169d9a7a0ab3444746a8aa5b0a3/operations/keys.go#L78-L81) also has to be modified.

cloud/spawn.go
```diff
@@ -69,14 +77,8 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
...
 	// validate public key
-	rsa := "ssh-rsa"
-	dss := "ssh-dss"
-	isRSA := strings.HasPrefix(so.PublicKey, rsa)
-	isDSS := strings.HasPrefix(so.PublicKey, dss)
-	if !isRSA && !isDSS {
-		return errors.Errorf("Invalid spawn options: "+
-			"either an invalid Evergreen-managed key name has been provided,"+
-			"or the key value does not start with %s or %s", rsa, dss)
+	if err = ValidateSSHKey(so.PublicKey); err != nil {
```
Kimchelly, line 80: If we want this to be available to the CLI and to the server-side code, maybe move this into the top-level `evergreen` package? Otherwise, we would ahve to call this cloud provider validation code from the `operations` package.

cloud/spawn.go
bynn: done

cloud/spawn.go
bynn: done

public/static/partials/user_host_options.html
bynn: done

cloud/spawn.go
bynn: done

cloud/spawn.go
bynn: done

cloud/spawn.go
```diff
@@ -69,23 +69,13 @@ func (so *SpawnOptions) validate(settings *evergreen.Settings) error {
...
-	rsa := "ssh-rsa"
-	dss := "ssh-dss"
-	isRSA := strings.HasPrefix(so.PublicKey, rsa)
-	isDSS := strings.HasPrefix(so.PublicKey, dss)
-	if !isRSA && !isDSS {
-		return errors.Errorf("Invalid spawn options: "+
-			"either an invalid Evergreen-managed key name has been provided,"+
-			"or the key value does not start with %s or %s", rsa, dss)
+	if err = evergreen.ValidateSSHKey(so.PublicKey); err != nil {
+		return errors.Wrap(err, "Invalid spawn options: ")
```
Kimchelly, line 73: `errors.Wrap` already formats error strings with colons between each message, so you only need to put `"invalid spawn options"` as the wrapping context string and it'll look like `invalid spawn options: <wrapped error message>`.

public/static/partials/user_host_options.html
```diff
@@ -70,7 +70,7 @@
       Please enter a unique key name for your new key<br />
     </span>
     <span class="semi-muted invalid" ng-show="spawnHostForm.pubkey.$dirty && (spawnHostForm.pubkey.$error.required || spawnHostForm.pubkey.$error.keyBaseValid)">
-      Key is invalid. It must begin with 'ssh-rsa' or 'ssh-dss'. Ensure you have copied the public half of the key.<br />
+      Key is invalid. It must begin with 'ssh-rsa', or 'ssh-dss', or 'ssh-ed25519', or 'ecdsa-sha2-nistp256'. Ensure you have copied the public half of the key.<br />
```
Kimchelly, line 73: nit: remove 1st `or` since it's a list of 4 strings.

Kimchelly (approved):
LGTM mod two minor comments.

PR #5592, opened 2022-04-19
EVG-16076 disallow Evergreen from decomissioning hosts running task groups

ablack12:
[EVG-16076](https://jira.mongodb.org/browse/EVG-16076)

### Description 
Right now we sometimes have a race where Evergreen assigns a task to a running host, but then Evergreen decommissions it because our local host document doesn't show this running task yet. This is an issue when the task is part of a task group, since we shouldn't be decommissioning that host.

Since we check when assigning a task that the status is running in the query itself, it makes sense to also check the task group field when decommissioning, since that would eliminate the race.

### Testing 
Added a unit test.

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -580,10 +580,10 @@ func (h *Host) SetDecommissioned(user string, checkTaskGroup bool, logs string) 
 	err := h.setStatusAndFields(evergreen.HostDecommissioned, query, nil, user, logs)
 	// Shouldn't consider it an error if the host isn't found when checking task group,
 	// because a task group may have been set for the host.
-	if err != nil && checkTaskGroup && !adb.ResultsNotFound(err) {
-		return err
+	if err != nil && checkTaskGroup && adb.ResultsNotFound(err) {
+		return nil
 	}
-	return nil
+	return err
 }
 
 func (h *Host) SetRunning(user string) error {
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -457,22 +457,16 @@ func IsIntentHostId(id string) bool {
 
 // SetStatus updates a host's status on behalf of the given user.
 func (h *Host) SetStatus(newStatus, user, logs string) error {
-	return h.SetStatusAndFields(newStatus, bson.M{}, user, logs)
+	return h.setStatusAndFields(newStatus, nil, nil, user, logs)
 }
 
-// SetStatusAndFields sets the status as well as any of the other given fields.
-// It also validates running tasks if needed.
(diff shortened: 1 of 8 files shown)


Kimchelly (commented):
There's a bunch of legitimate reasons a host might be decommissioned which might not be just due to a desire to optimize host usage like in the host drawdown. It seems like the task lifecycle should be resilient to host interrupts which occur for reasons other than optimization of host usage. Can I suggest making a special function for the host drawdown job so that it respects when a host is running a task group rather than disallowing decommissioning for any reason?

If not, a bunch of places might need to be fixed:

Since [the host drawdown job stops on first error](https://github.com/evergreen-ci/evergreen/blob/17f9e61403adf169d9a7a0ab3444746a8aa5b0a3/units/host_drawdown.go#L161-L163), should it instead continue on error since it should continue decommissioning even if hits this case?

This also affects [host termination](https://github.com/evergreen-ci/evergreen/blob/17f9e61403adf169d9a7a0ab3444746a8aa5b0a3/units/host_termination.go#L235) since the host might be terminated for other reasons outside our control like external termination, which would require clearing the last group in the else block right above it.

[This function](https://github.com/evergreen-ci/evergreen/blob/17f9e61403adf169d9a7a0ab3444746a8aa5b0a3/model/host/host.go#L1551-L1561) also ignores when it's running a task group. Is it supposed to do that?

ablack12:
Yeah that's fair. I'm having some trouble deciding if all of these other circumstances _should_ be considering this as well, but let's start with this job and go from there. I'm going to include this as a parameter to the existing SetDecommissioned function though as opposed to having two very similar functions.

model/host/host.go
```diff
@@ -544,24 +550,40 @@ func (h *Host) SetProvisioning() error {
...
+		grip.Warning(message.WrapError(catcher.Resolve(), message.Fields{
+			"message":  "error decommissioning containers",
+			"host_ids": failedContainerIds,
+		}))
 	}
-	return h.SetStatus(evergreen.HostDecommissioned, user, logs)
+	err := h.setStatusAndFields(evergreen.HostDecommissioned, query, nil, user, logs)
+	// Shouldn't consider it an error if the host isn't found when checking task group,
+	// because a task group may have been set for the host.
+	if err != nil && checkTaskGroup && !adb.ResultsNotFound(err) {
```
Kimchelly, line 583: Is it supposed to return the error if `checkTaskGroup` is false?

ablack12: ah yes my bad

Kimchelly (approved):

PR #5599, opened 2022-04-20
EVG-16731 Update version aborted field

bynn:
[EVG-16731](https://jira.mongodb.org/browse/EVG-16731)

### Description 
version and build aborted field now dynamically reflects whether we should show the status as aborted or not 

### Testing 

normal patch with success and abort will show abort 
will show failure if tasks are abort and failure 
image

works with child patches 
https://spruce-staging.corp.mongodb.com/version/613b92639dbe323620f081ab/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
image
image

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3913,56 +3913,37 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
 	if err != nil {
 		return "", InternalServerError.Send(ctx, fmt.Sprintf("An error occurred when converting a version status: %s", err.Error()))
 	}
-	if utility.FromBoolPtr(obj.Aborted) && !utility.StringSliceContains(evergreen.TaskFailureStatuses, status) {
-		return evergreen.PatchAborted, nil
-	} else {
-		return status, nil
+	isAborted := utility.FromBoolPtr(obj.Aborted)
+	nonAbortedStatuses := []string{}
+	if !isAborted {
+		nonAbortedStatuses = append(nonAbortedStatuses, status)
+	}
+	if evergreen.IsPatchRequester(*obj.Requester) {
+		p, err := data.FindPatchById(*obj.Id)
+		if err != nil {
+			return status, InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch Patch %s: %s", *obj.Id, err.Error()))
+		}
+		if len(p.ChildPatches) > 0 {
+			for _, cp := range p.ChildPatches {
+				cpVersion, err := model.VersionFindOneId(*cp.Version)
+				if err != nil {
+					return "", InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch version for patch: %s ", err.Error()))
+				}
+				if cpVersion.Aborted {
+					isAborted = true
+				} else {
+					nonAbortedStatuses = append(nonAbortedStatuses, *cp.Status)
+				}
+			}
+			status = patch.GetCollectiveStatus(nonAbortedStatuses)
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


bynn:
should we file a UI ticket for downstream tasks to also show aborted instead of failure?

graphql/resolvers.go
```diff
@@ -3930,48 +3930,39 @@ func (r *versionResolver) PreviousVersion(ctx context.Context, obj *restModel.AP
...
 	}
 
-	taskStatuses := getAllTaskStatuses(tasks)
-
 	// If theres an aborted task we should set the patch status to aborted if there are no other failures
-	if utility.StringSliceContains(taskStatuses, evergreen.TaskAborted) {
-		if len(utility.StringSliceIntersection(taskStatuses, evergreen.TaskFailureStatuses)) == 0 {
-			status = evergreen.PatchAborted
-		}
+	if isAborted && len(utility.StringSliceIntersection(evergreen.TaskFailureStatuses, nonAbortedStatuses)) == 0 {
```
malikchaya2, line 3964: Since child patch statuses are already in nonAbortedStatuses, aren't child patches already covered? What would a `ticket for downstream tasks to also show aborted instead of failure` cover?

malikchaya2 (changes requested):

graphql/resolvers.go
```diff
@@ -3930,48 +3930,39 @@ func (r *versionResolver) PreviousVersion(ctx context.Context, obj *restModel.AP
...
 		if err != nil {
 			return status, InternalServerError.Send(ctx, fmt.Sprintf("Could not fetch Patch %s: %s", *obj.Id, err.Error()))
 		}
 		if len(p.ChildPatches) > 0 {
-			patchStatuses := []string{*p.Status}
 			for _, cp := range p.ChildPatches {
-				patchStatuses = append(patchStatuses, *cp.Status)
-				// add the child patch tasks to tasks so that we can consider their status
-				childPatchTasks, _, err := task.GetTasksByVersion(*cp.Id, opts)
+				cpVersion, err := model.VersionFindOneId(*cp.Version)
```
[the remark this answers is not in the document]
bynn: added to description

graphql/resolvers.go
bynn: not too sure by what you mean by >>Since child patch statuses are already in nonAbortedStatuses, aren't child patches already covered?
Do you mean that we can just check if `status` is failing and not all nonabortedstatusus? because i think that does make sense

malikchaya2: Ah I understand now. Yes, I think it's a good idea to create a ticket and include the above photo. We should probably prioritize that ticket because I think this can end up confusing users. 
cc: @ablack12

malikchaya2 (approved):

graphql/resolvers.go
ablack12: 👍  assuming this is a UI ticket I think this will be up to them to prioritize (I'm not sure where that status comes from). Since this isn't yet a heavily used feature I don't know that it needs to be done ASAP but you're right that we shouldn't let it slip.

SupaJoon, line 3964: this line is checking Patch statuses against task statuses. I think we should change TaskFailureStatuses to represent patch failure statuses.

SupaJoon (changes requested):

graphql/resolvers.go
SupaJoon: I will include the UI changes regarding the aborted status badge in this ticket: https://jira.mongodb.org/browse/EVG-14586

bynn: done

SupaJoon (approved):

PR #5607, opened 2022-04-25
EVG-16033: Expose number of tests belonging to a task 

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-16033

### Description 
Add new API endpoint `rest/v2/tasks/<task_id>/tests/count` to return the number of tests for a given task ID. Once this is merged and deployed, I will add documentation to the Evergreen wiki.

### Testing 
- Unit tests for fetching the test count from either Cedar or the database, depending on where the tests live.
- Tested this in staging with both test results in Cedar and in the database.

### DIFF
--- a/rest/data/test.go
+++ b/rest/data/test.go
@@ -1,14 +1,19 @@
 package data
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/apimodels"
 	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/db/mgo/bson"
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/model/testresult"
 	"github.com/evergreen-ci/gimlet"
+	"github.com/evergreen-ci/utility"
+	"github.com/pkg/errors"
 )
 
 // DBTestConnector is a struct that implements the Test related methods
@@ -75,4 +80,62 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
 	}
 
 	return res, nil
+}
+
+func (*DBTestConnector) TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, error) {
+	t, err := task.FindOneIdNewOrOld(taskID)
+	if err != nil {
+		return 0, gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    errors.Wrapf(err, "finding task '%s'", taskID).Error(),
+		}
+	}
+	if t == nil {
+		return 0, gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("task '%s' not found", taskID),
+		}
+	}
+
+	var count int
+	if t.HasCedarResults {
+		stats, status, err := apimodels.GetCedarTestResultsStats(ctx, apimodels.GetCedarTestResultsOptions{
+			BaseURL:     evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:      taskID,
+			Execution:   utility.ToIntPtr(execution),
+			DisplayTask: t.DisplayOnly,
+		})
+		if err != nil {
+			return 0, gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    errors.Wrapf(err, "getting test results stats from Cedar for task '%s'", taskID).Error(),
+			}
+		}
+		if status != http.StatusOK && status != http.StatusNotFound {
+			return 0, gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    errors.Wrapf(err, "getting test results stats from Cedar for task '%s' return status '%d'", taskID, status).Error(),
+			}
+		}
+
+		count = stats.TotalCount
+	} else {
+		var taskIDs []string
+		if t.DisplayOnly {
+			taskIDs = t.ExecutionTasks
+		} else {
+			taskIDs = []string{taskID}
+		}
+
+		var err error
+		count, err = testresult.TestResultCount(taskIDs, "", nil, execution)
+		if err != nil {
+			return 0, gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    errors.Wrapf(err, "getting test count for task '%s'", taskID).Error(),
+			}
+		}
+	}
+
+	return count, nil
 }
--- a/rest/data/test.go
+++ b/rest/data/test.go
@@ -82,7 +82,9 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
 	return res, nil
 }
 
-func TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, error) {
+// CountTestsByTaskID returns the number of tests for a given task ID and
+// execution.
+func CountTestsByTaskID(ctx context.Context, taskID string, execution int) (int, error) {
 	t, err := task.FindOneIdOldOrNew(taskID, execution)
 	if err != nil {
 		return 0, gimlet.ErrorResponse{
@@ -97,7 +99,6 @@ func TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, 
 		}
 	}
 
-	var count int
 	if t.HasCedarResults {
 		stats, status, err := apimodels.GetCedarTestResultsStats(ctx, apimodels.GetCedarTestResultsOptions{
 			BaseURL:     evergreen.GetEnvironment().Settings().Cedar.BaseURL,
@@ -118,7 +119,7 @@ func TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, 
 			}
 		}
 
-		count = stats.TotalCount
+		return stats.TotalCount, nil
 	} else {
 		var taskIDs []string
 		if t.DisplayOnly {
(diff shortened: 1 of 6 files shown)


rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
 
 	return res, nil
 }
+
+func TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, error) {
+	t, err := task.FindOneIdOldOrNew(taskID, execution)
```
Kimchelly, line 86: Will this guarantee that the task execution it gets back is the expected task execution? For example, if they pass execution=100, is it supposed to return the latest execution or error? Same for execution=0 - is it supposed to return the latest execution or the first execution?

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
...
+			}
+		}
+		if status != http.StatusOK && status != http.StatusNotFound {
+			return 0, gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    fmt.Sprintf("getting test results stats from Cedar for task '%s' returned status %d", taskID, status),
+			}
+		}
+
+		count = stats.TotalCount
```
Kimchelly, line 121: nit: rather than returning `count` at the very end of the entire if-else, would it be better to just return from this `if` block and unindent the else?

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
...
+	} else {
+		var taskIDs []string
+		if t.DisplayOnly {
+			taskIDs = t.ExecutionTasks
+		} else {
+			taskIDs = []string{taskID}
+		}
+
+		var err error
+		count, err = testresult.TestResultCount(taskIDs, "", nil, execution)
```
Kimchelly, line 131: Similar question as above about unspecified execution numbers. Is that valid, or should this function error when it gets a zero execution?

rest/data/test_test.go
```diff
@@ -311,3 +317,138 @@ func TestFindTestsByDisplayTaskId(t *testing.T) {
...
+				statusCode: http.StatusOK,
+			},
+		} {
+			t.Run(test.name, func(t *testing.T) {
+				responseData, err := json.Marshal(&test.response)
+				require.NoError(t, err)
+				handler.Response = responseData
+				handler.StatusCode = test.statusCode
+
+				count, err := TestCountByTaskID(ctx, test.task.Id, test.task.Execution)
```
Kimchelly, line 393: These seem to always assume that there's a single execution and it's the latest one. Should there be a test for other execution cases?

* Non-latest archived task executions
* Zero execution with multiple task executions
* Positive execution numbers that are higher than the latest execution

rest/route/test.go
```diff
@@ -16,7 +16,10 @@ import (
 	"github.com/pkg/errors"
 )
 
-// testGetHandler is the MethodHandler for the GET /tasks/{task_id}/tests route.
+///////////////////////////////////////////////////////////////////////////////
+//
+// GET  /tasks/{task_id}/tests
```
Kimchelly, line 21: nit: remove double whitespace between `GET` and `/tasks`.

Kimchelly, line 21: nit: remove double spaces between `GET` and `/tasks`.

rest/route/test.go
```diff
@@ -217,3 +217,56 @@ func (tgh *testGetHandler) addDataToResponse(resp gimlet.Responder, testResult i
 
 	return nil
 }
+
+///////////////////////////////////////////////////////////////////////////////
+//
+// GET  /tasks/{task_id}/tests/count
```
Kimchelly, line 223: nit: remove double spaces.

rest/route/test.go
```diff
@@ -217,3 +217,56 @@ func (tgh *testGetHandler) addDataToResponse(resp gimlet.Responder, testResult i
...
+	h.taskID = projCtx.Task.Id
+
+	var err error
+	vals := r.URL.Query()
+	execution := vals.Get("execution")
+
+	if execution != "" {
+		h.execution, err = strconv.Atoi(execution)
+		if err != nil {
+			return gimlet.ErrorResponse{
```
Kimchelly, line 255: `Parse` already returns `http.StatusBadRequest` by default so there's no need to pass back a `gimlet.ErrorResponse`:
```go
return errors.Wrap(err, "invalid execution number")
```

rest/route/test.go
```diff
@@ -217,3 +217,56 @@ func (tgh *testGetHandler) addDataToResponse(resp gimlet.Responder, testResult i
...
+	return nil
+}
+
+func (h *testCountGetHandler) Run(ctx context.Context) gimlet.Responder {
+	count, err := data.TestCountByTaskID(ctx, h.taskID, h.execution)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "database error"))
+	}
+
+	return gimlet.NewTextResponse(count)
```
Kimchelly, line 271: Should this be a text response? It's not necessarily a problem, but most of the rest of the API returns JSON-structured responses, like it does in the error case on L268.

rest/route/test.go
```diff
@@ -217,3 +217,56 @@ func (tgh *testGetHandler) addDataToResponse(resp gimlet.Responder, testResult i
...
+		}
+	}
+
+	return nil
+}
+
+func (h *testCountGetHandler) Run(ctx context.Context) gimlet.Responder {
+	count, err := data.TestCountByTaskID(ctx, h.taskID, h.execution)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "database error"))
```
Kimchelly, line 268: This could be a DB or Cedar issue:
```go
errors.Wrap(err, "counting tests for task '%s' and execution %d", h.taskID, h.execution)
```

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
...
+			Message:    fmt.Sprintf("task '%s' not found", taskID),
+		}
+	}
+
+	var count int
+	if t.HasCedarResults {
+		stats, status, err := apimodels.GetCedarTestResultsStats(ctx, apimodels.GetCedarTestResultsOptions{
+			BaseURL:     evergreen.GetEnvironment().Settings().Cedar.BaseURL,
+			TaskID:      taskID,
+			Execution:   utility.ToIntPtr(execution),
```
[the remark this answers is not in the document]
julianedwards: Yeah, it will just be set to zero here in that case since the route passes in the zero value of an `int`.

rest/route/test.go
julianedwards: I am not sure, I guess it could be a structured response but it is literally just an int. I am not super opinionated.

rest/data/test_test.go
julianedwards: I think that is testing too much of the Cedar specific behavior, which this test can assume just works. We really just want to make sure we are passing in the right parameters to Cedar's API endpoint and that we handle the Cedar response accordingly. I can add a test that checks that this function fails if the execution does not exist (that would come from the database check before hitting the API endpoint).

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
 
 	return res, nil
 }
+
+func TestCountByTaskID(ctx context.Context, taskID string, execution int) (int, error) {
```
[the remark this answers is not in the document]
julianedwards: Done and done.

rest/data/test.go
julianedwards: Done.

rest/data/test.go
julianedwards: See above.

rest/route/test.go
julianedwards: Done.

julianedwards: Done.

rest/route/test.go
julianedwards: Done.

rest/route/test.go
julianedwards: Done.

rest/route/test.go
julianedwards: I just return the error since the `rest/data` code takes care of setting the error message.

rest/data/test.go
julianedwards: Ah I see what you are saying now, I am going to add a check that the returned task has the same execution otherwise return an error. I think it makes more sense for users to return an error if the execution does not exist rather than silently return the latest execution.

rest/route/test.go
julianedwards: SGTM!

rest/data/test_test.go
julianedwards: I added a test to make sure that it will fail when the execution differs from the one passed in to the function.

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
 
 	return res, nil
 }
+
+// CountTestsByTaskID returns the number of tests for a given task ID and
+// execution.
```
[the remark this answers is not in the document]
julianedwards: I think that this is fine since it is not taking in a pointer and we fail if the execution does not exist now.

rest/data/test.go
```diff
@@ -76,3 +81,61 @@ func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]test
...
+		}
+		if status != http.StatusOK && status != http.StatusNotFound {
+			return 0, gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    fmt.Sprintf("getting test results stats from Cedar for task '%s' returned status %d", taskID, status),
+			}
+		}
+
+		return stats.TotalCount, nil
+	} else {
```
Kimchelly, line 123: The else can be unindented now since if already returns.

julianedwards: Done.

PR #5608, opened 2022-04-26
EVG-16557: Disable GQL queries

SupaJoon:
[EVG-16557](https://jira.mongodb.org/browse/EVG-16557)

### Description 
These code changes include gql middleware and legacy ui components to disable queries specified in the admin collection disabled_gql_queries field. The interface requires the user to click the "Save" button at the top of the page. 

### Testing 
unit and integration
Screen Shot 2022-04-26 at 2 03 22 PM


graphql/disablequery.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+)
+
+// SplunkTracing is a graphql extension that adds splunk logging to graphql.
+// It is used to log the duration of a query and the user that made the request.
+// It does this by hooking into lifecycle events that gqlgen uses.
```
khelif96, line 16: copypasta

graphql/disablequery.go
```diff
@@ -0,0 +1,41 @@
...
+func (DisableQuery) Validate(graphql.ExecutableSchema) error {
+	return nil
+}
+
+func (DisableQuery) MutateOperationContext(ctx context.Context, rc *graphql.OperationContext) *gqlerror.Error {
+	settings, err := evergreen.GetConfig()
+	if err != nil {
+		grip.Error(errors.Wrap(err, "getting Evergreen admin settings"))
+	}
+	if utility.StringSliceContains(settings.DisabledGQLQueries, rc.Operation.Name) {
```
khelif96, line 32: Should we nil check `settings`

graphql/disablequery.go
```diff
@@ -0,0 +1,41 @@
...
+		return &gqlerror.Error{
```
khelif96, line 33: Should we use `ServiceUnavailable.Send()`?

graphql/tests/serviceDegradation/data.json
```diff
@@ -0,0 +1,77 @@
...
+                "token": "",
+                "channel": ""
+            },
+            "ssh_key_directory": "",
+            "ssh_key_pairs": []
+        },
+        {
+            "_id": "jira",
+            "basic_auth": {
+                "username": "mohamed.khelif",
```
khelif96, line 64: 😎

graphql/disablequery.go
```diff
@@ -0,0 +1,41 @@
+package graphql
```
khelif96, line 1: How do you feel about `QueryValidation` or something along those lines?

khelif96:
Great Job works great! Left a few comments. We should also get App buy-in for this.

graphql/disablequery.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (DisableQuery) ExtensionName() string {
+	return "DisableQuery"
+}
+
+func (DisableQuery) Validate(graphql.ExecutableSchema) error {
+	return nil
+}
+
+func (DisableQuery) MutateOperationContext(ctx context.Context, rc *graphql.OperationContext) *gqlerror.Error {
```
ablack12, line 26: Can we change the filename to disable_query.go since i t's two words?

graphql/disablequery.go
SupaJoon: I think our middleware could belong in a graphql-middleware package but I don't see a great benefit since the middleware is only used in one file so far, http_handler.go

service/templates/admin.html
```diff
@@ -458,6 +458,32 @@ <h2 class="modal-title">[[modalTitle]]</h2>
...
+										</md-input-container>
+										<md-button ng-click="addDisabledGQLQuery()">
+											<i class="fa fa-plus"></i>
+										</md-button>
+										<div id="ec2-list" class="form-group" ng-repeat="(index, queryName) in Settings.disabled_gql_queries">
+											<section layout="row" flex>
+												<div>
+													<span>[[queryName]]</span>
+													<button class="btn btn-default btn-danger" type="button" ng-click="removeDisabledGQLQuery(queryName)">
+														<i class="fa fa-trash"></i>
```
ablack12, line 480: Can we move the trashcan to be in a consistent column (rather than right next to the value)? At least that's how it's looking in the screenshot.

service/templates/admin.html
```diff
@@ -458,6 +458,32 @@ <h2 class="modal-title">[[modalTitle]]</h2>
...
+									</md-card-title-text>
+								</md-card-title>
+								<md-card-content>
+									<div>
+										<md-input-container class="control" style="width: 200px;">
+											<label>Operation name to disable</label>
+											<input type="text" default="ABC" ng-model="queryMapping.newQuery" />
+										</md-input-container>
+										<md-button ng-click="addDisabledGQLQuery()">
+											<i class="fa fa-plus"></i>
```
ablack12, line 473: This plus button isn't lined up / consistent with the rest of the page (we don't take legacy consistency too seriously but it does make it easier to understand).
image

I'd prefer if we went with the lined up green plus sign or the list format (example is CORS origin).

image

image

graphql/disablequery.go
khelif96: Sorry I should have clarified I was referring to file name lol

service/templates/admin.html
SupaJoon: I changed the styles to align the buttons
Screen Shot 2022-04-28 at 2 03 19 PM

graphql/disablequery.go
SupaJoon: I don't prefer QueryValidation because there is already an error code called "GRAPHQL_VALIDATION_FAILED" that is returned when the query doesn't validate against the gql schema. This check executes before our custom middleware.

service/templates/admin.html
```diff
@@ -458,6 +458,35 @@ <h2 class="modal-title">[[modalTitle]]</h2>
...
+								<md-card-title>
+									<md-card-title-text>
+										<span>Disabled GQL Queries</span>
+									</md-card-title-text>
+								</md-card-title>
+								<md-card-content>
+									<div>
+										<section style="height: 70px;" layout="row" flex>
+											<md-input-container class="control"  style="width: 200px;">
+												<label>Operation name to disable</label>
```
ablack12, line 470: To clarify is this the resolver name? Will pages fail to load if we've disabled operations? Do we intend to remove this code once mainline is baked in or do we want to keep this as a degradable option?

graphql/disable_query.go
```diff
@@ -0,0 +1,34 @@
...
+func (DisableQuery) ExtensionName() string {
+	return "DisableQuery"
+}
+
+func (DisableQuery) Validate(graphql.ExecutableSchema) error {
+	return nil
+}
+
+func (DisableQuery) MutateOperationContext(ctx context.Context, rc *graphql.OperationContext) *gqlerror.Error {
+	settings, err := evergreen.GetConfig()
```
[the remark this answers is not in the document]
SupaJoon: Disabled queries are still logged with SplunkTracing

PR #5614, opened 2022-04-27
EVG-16763 include task timed out in trigger failures

ablack12:
[EVG-16763](https://jira.mongodb.org/browse/EVG-16763)

### Description 
Need to include task timed out for triggers.

### DIFF
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -396,7 +396,7 @@ func (t *taskTriggers) taskOutcome(sub *event.Subscription) (*notification.Notif
 		return nil, nil
 	}
 
-	if t.data.Status != evergreen.TaskSucceeded && !isFailedTaskStatus(t.data.Status) {
+	if t.data.Status != evergreen.TaskSucceeded && !isValidFailedTaskStatus(t.data.Status) {
 		return nil, nil
 	}
 
@@ -412,7 +412,7 @@ func (t *taskTriggers) taskFailure(sub *event.Subscription) (*notification.Notif
 		return nil, nil
 	}
 
-	if !isFailedTaskStatus(t.data.Status) {
+	if !isValidFailedTaskStatus(t.data.Status) {
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


trigger/task.go
```diff
@@ -703,8 +703,11 @@ func (t *taskTriggers) taskRuntimeChange(sub *event.Subscription) (*notification
 	return t.generate(sub, fmt.Sprintf("changed in runtime by %.1f%% (over threshold of %s%%)", percentChange, percentString), "")
 }
 
-func isFailedTaskStatus(status string) bool {
-	return status == evergreen.TaskFailed || status == evergreen.TaskSystemFailed || status == evergreen.TaskTestTimedOut
+// isValidFailedTaskStatus only matches task statuses that should be triggered for failure.
+// For example, it excludes  setup failures.
+func isValidFailedTaskStatus(status string) bool {
```
malikchaya2, line 708: I feel like the addition of the word `valid` here doesn't really help me understand better what the function does. I personally feel like isFailedTaskStatus with the comment you added feels sufficient to me. 
But I don't feel strongly either way so you decide :)

malikchaya2 (approved):

trigger/task.go
```diff
@@ -703,8 +703,11 @@ func (t *taskTriggers) taskRuntimeChange(sub *event.Subscription) (*notification
...
+	return status == evergreen.TaskFailed || status == evergreen.TaskSystemFailed ||
```
bynn, line 709: would it make sense to use [TaskNonGenericFailureStatuses](https://github.com/evergreen-ci/evergreen/blob/7fa53940bcca04122f7c23e7fffcdfa37a1139df/globals.go#L314)?

bynn (approved):
if you see a notification from me it's because i asked a question and deleted it!

trigger/task.go
ablack12: I kind of just want to distinguish it from the non-private-trigger function "IsFailedTaskStatus" but yeah i recognize it doesn't add much

malikchaya2: Oh, that makes sense

PR #5619, opened 2022-04-29
EVG-16103 Abort tasks encountering 409 conflicts when performing heartbeat

hadjri:

### Description 
Spikes of errors saying "invalid task: Wrong secret sent for task" had been noticed in production consistently. Each of these tasks that these errors arose for had a common trait where they all became unresponsive and were eventually assigned the "system-unresponsive" display status.  Although MarkEnd for the task had run for these tasks updating fields like finish_time and status, the task did not appear to be actually terminated and persisted past its supposed finish time.  During these time windows up until the task is eventually actually terminated we see the "wrong secret" error spikes.  Thus, there can be tons of error messages for a single task if it takes a long time to terminate. Other spawn hosts likely are pulling the task from its respective task queue during these times and since the task still exists on the original host, "wrong secret" errors come up on these new spawn hosts.

2 changes were made:

-Additional logging saying what the spawn host id is when sending in a request to one of evergreen's app servers so we can check the hypothesis that the tasks are being picked up by other spawn hosts during this limbo period
-Modifying the logic in the agent to hopefully effectively terminate a task when heartbeat fails with 409 conflict, as well as adding a couple more lines to the log output. Ideally this will fix the issue itself.


agent/background.go
```diff
@@ -49,6 +51,7 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 				return
 			}
 		case <-ctx.Done():
+			tc.logger.Task().Error("Heartbeat ticker canceled")
 			grip.Info("Heartbeat ticker canceled")
```
ablack12, line 55: Honestly I think we can just remove this, does no good.

agent/background.go
```diff
@@ -49,6 +51,7 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 				return
 			}
 		case <-ctx.Done():
+			tc.logger.Task().Error("Heartbeat ticker canceled")
```
ablack12, line 54: Should this say "aborting task" the way the others do?

ablack12 (approved):

PR #5620, opened 2022-04-29
EVG-16535Fix build status updates

bynn:
[EVG-16535](https://jira.mongodb.org/browse/EVG-16535)

### Description 
currently, we return build created if there are no running tasks (even if they are blocked) so i added a check for that 

### Testing 
restarted a task in these versions and they now say failed instead of running 
https://spruce-staging.corp.mongodb.com/version/curator_9801fc7d910fe9fc76d09a92649676efa4ad9fa7/tasks?statuses=undispatched-umbrella,unscheduled,aborted,blocked&variant=%5Earchlinux%24
https://spruce-staging.corp.mongodb.com/version/curator_3b114c007611cafa8a179d6a6c14955c75cc3f25/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


model/task_lifecycle.go
```diff
@@ -886,7 +886,7 @@ func getBuildStatus(buildTasks []task.Task) string {
 
 	// finished but failed
 	for _, t := range buildTasks {
-		if evergreen.IsFailedTaskStatus(t.Status) || t.Aborted {
+		if evergreen.IsFailedTaskStatus(t.Status) || t.Aborted || t.Blocked() {
```
ablack12, line 889: This is a pretty big change; if all tasks are successful, and one task is blocked, does it make sense to return failed when nothing in the build itself actually failed? I think not -- for example, you _can_ create tasks that only run if the dependency _failed_ . It seems expected in that case then that being blocked doesn't mean it's a failure.

bynn: that makes a lot of sense
removed

model/task_lifecycle.go
```diff
@@ -865,7 +865,7 @@ func getBuildStatus(buildTasks []task.Task) string {
 	// not started
 	noStartedTasks := true
 	for _, t := range buildTasks {
-		if !evergreen.IsUnstartedTaskStatus(t.Status) {
+		if !evergreen.IsUnstartedTaskStatus(t.Status) && !t.Blocked() {
```
ablack12, line 868: So I've been trying to parse this logic for a while and it's pretty confusing to think about, so I  think we need to change the variable names (and also write a unit test to confirm that this change fixes the case described in the ticket). 

Originally we set "noStartedTasks" to false only if ANY task wasn't in the "undispatched" state (i.e. any task starts or fails). Now we're saying "if any task isn't in undispatched AND it's not blocked", but wouldn't a task that's started also be unblocked? So I'm not sure this is the right change.

ablack12 (changes requested):
Can you modify the PR title to have a space after the ticket number ?

model/task_lifecycle.go
bynn: updated

model/task_lifecycle.go
```diff
@@ -865,7 +865,7 @@ func getBuildStatus(buildTasks []task.Task) string {
 	// not started
 	noStartedTasks := true
 	for _, t := range buildTasks {
-		if !evergreen.IsUnstartedTaskStatus(t.Status) {
+		if !evergreen.IsUnstartedTaskStatus(t.Status) || t.Blocked() {
```
ablack12, line 868: I think this is only part of the fix -- what if we have only unstarted and blocked tasks? Do we already say that the build is failed? Or do we wait until the unstarted ones are finished? I think we only want to tweak the status if all tasks are finished or blocked.

model/task_lifecycle_test.go
```diff
@@ -920,6 +920,22 @@ func TestGetBuildStatus(t *testing.T) {
 		{Status: evergreen.TaskFailed},
 	}
 	assert.Equal(t, evergreen.BuildFailed, getBuildStatus(buildTasks))
+
+	// blocked tasks don't prevent the build from completing
```
ablack12, line 924: Should use a full sentence per style guide

model/task_lifecycle.go
```diff
@@ -865,7 +865,7 @@ func getBuildStatus(buildTasks []task.Task) string {
 	// not started
```
ablack12, line 865: Can we tweak this comment since it's not clear cut anymore? Maybe a full sentence

ablack12 (changes requested):

model/task_lifecycle.go
ablack12: It seems from the ticket description that the bug isn't in this case but in the lower "do we mark it as started or failed" case.

bynn: fixed

model/task_lifecycle_test.go
bynn: fixed

model/task_lifecycle.go
bynn: fixed

model/build/build.go
```diff
@@ -187,6 +189,17 @@ func (b *Build) SetAborted(aborted bool) error {
 	)
 }
 
+func (b *Build) SetBlocked(blocked bool) error {
```
ablack12, line 192: Similarly can we say SetAllTasksBlocked

ablack12: Should put a function comment also

model/build/db.go
```diff
@@ -43,6 +43,8 @@ var (
 	AbortedKey             = bsonutil.MustHaveTag(Build{}, "Aborted")
 
 	TaskCacheIdKey = bsonutil.MustHaveTag(TaskCache{}, "Id")
+
+	BlockedKey = "Blocked"
```
ablack12, line 47: We should use bsonutil.MustHaveTag like above (also should be up with the other Build tags)

ablack12 (changes requested):

model/build/build.go
```diff
@@ -72,6 +72,10 @@ type Build struct {
 	TriggerID    string `bson:"trigger_id,omitempty" json:"trigger_id,omitempty"`
 	TriggerType  string `bson:"trigger_type,omitempty" json:"trigger_type,omitempty"`
 	TriggerEvent string `bson:"trigger_event,omitempty" json:"trigger_event,omitempty"`
+
+	// Set to true if all tasks in the build are blocked.
+	// Should not be exposed, only for internal use.
+	AllTasksBlocked bool `bson:"blocked"`
```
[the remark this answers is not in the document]
bynn: done

model/build/db.go
bynn: done

model/task_lifecycle.go
```diff
@@ -849,37 +849,47 @@ func updateMakespans(b *build.Build, buildTasks []task.Task) error {
...
 			break
 		}
 	}
-	if noStartedTasks {
-		return evergreen.BuildCreated
+	// Check if all tasks are blocked in a build.
+	var allTasksBlocked bool
+	for _, t := range buildTasks {
+		if !t.Blocked() {
+			allTasksBlocked = false
```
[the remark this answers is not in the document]
bynn: done

model/task_lifecycle.go
```diff
@@ -849,37 +849,47 @@ func updateMakespans(b *build.Build, buildTasks []task.Task) error {
...
+			break
+		}
+		allTasksBlocked = true
```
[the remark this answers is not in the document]

model/task_lifecycle.go
```diff
@@ -972,14 +988,14 @@ func getVersionStatus(builds []build.Build) string {
 		return evergreen.VersionCreated
 	}
 
-	// started but not finished
+	// Check if builds are started but not finished.
```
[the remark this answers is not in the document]
bynn: image

there's nothing going on in my head. I think this should fix everything

model/task_lifecycle.go
```diff
@@ -906,21 +916,27 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
...
+
+	if allTasksBlocked != b.AllTasksBlocked {
+		if err = b.SetBlocked(allTasksBlocked); err != nil {
+			return false, errors.Wrapf(err, "setting build '%s' as blocked", b.Id)
+		}
+	}
 
 	if buildStatus == b.Status {
-		return false, nil
+		return allTasksBlocked != b.AllTasksBlocked, nil
```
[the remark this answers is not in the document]
bynn: ah youre right, i still added a boolean to check if there has been a change because we would still want to update the version status if the build status goes from "created, not blocked" to "created, blocked"

model/build/db.go
```diff
@@ -41,6 +41,7 @@ var (
 	ActualMakespanKey      = bsonutil.MustHaveTag(Build{}, "ActualMakespan")
 	IsGithubCheckKey       = bsonutil.MustHaveTag(Build{}, "IsGithubCheck")
 	AbortedKey             = bsonutil.MustHaveTag(Build{}, "Aborted")
+	BlockedKey             = bsonutil.MustHaveTag(Build{}, "AllTasksBlocked")
```
ablack12, line 44: Should be AllTasksBlockedKey

model/task_lifecycle.go
```diff
@@ -849,37 +849,45 @@ func updateMakespans(b *build.Build, buildTasks []task.Task) error {
 	return errors.WithStack(b.UpdateMakespans(depPath.TotalTime, CalculateActualMakespan(buildTasks)))
 }
 
-func getBuildStatus(buildTasks []task.Task) string {
-	// not started
+// getBuildStatus returns a string denoting the status of the build and
+// a boolean denoting if all tasks in the build are blocked.
+func getBuildStatus(buildTasks []task.Task) (string, bool) {
+	// Check if no tasks have started in the build.
```
ablack12, line 855: should modify comment to also mention blocked

model/task_lifecycle.go
```diff
@@ -906,21 +914,26 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
 	return b.UpdateGithubCheckStatus(githubBuildStatus)
 }
 
-// UpdateBuildStatus updates the status of the build based on its tasks' statuses.
+// updateBuildStatus updates the status of the build based on its tasks' statuses.
 // Returns true if the build's status has changed.
```
ablack12, line 918: Should update to include all blocked

model/task_lifecycle.go
```diff
@@ -972,14 +985,17 @@ func getVersionStatus(builds []build.Build) string {
 		return evergreen.VersionCreated
 	}
 
-	// started but not finished
+	// Check if builds are started but not finished.
 	for _, b := range builds {
+		if b.AllTasksBlocked {
```
ablack12, line 990: I think instead of this continue, line 993 could just also have`&& !b.AllTasksBlocked`

ablack12 (changes requested):

model/build/db.go
bynn: done

model/task_lifecycle.go

model/task_lifecycle.go

model/task_lifecycle.go
bynn: sorry i feel like it makes so much sense when i look back on it :(

ablack12: No worries! It's easier as a reviewer, and this is more a style thing than a correctness thing

ablack12 (approved):

PR #5623, opened 2022-05-03
EVG-15168: Fetch base statuses based on version's activated tasks

minnakt:
[EVG-15168](https://jira.mongodb.org/browse/EVG-15168)

### Description 
In the video below, you can see that my patch has some extra base statuses filters that shouldn't be there. The reason they appear is because we currently don't limit the base statuses to tasks that actually ran on the version; instead, we get every single status of every task that ran on the base version.

https://user-images.githubusercontent.com/47064971/166401940-e5260ba1-71e5-4c7d-bf9c-8ef6c36315fe.mov

This PR makes it so that we restrict the base statuses to tasks that actually ran on the version. This way, we don't see other base statuses that seem irrelevant to the patch.

### Note
Note that this base status functionality was not working correctly for mainline commits at all because we were only looking up the base commits but not previous commits. This should now be fixed.

### Testing 
- Modified graphql tests in `TestAtomicGQLQueries/version`
- Add unit test `TestGetMatchingBaseTasks`


malikchaya2 (approved):
approved with nit

bynn (approved):

model/task/task.go
```diff
@@ -3669,6 +3669,58 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
...
+// If a task ran on the base version but was not scheduled on the version, then it won't be returned.
+func GetMatchingBaseTasks(versionID string, baseVersionID string) ([]Task, error) {
+	pipeline := []bson.M{}
+
+	pipeline = append(pipeline,
+		bson.M{"$match": bson.M{VersionKey: baseVersionID}},
+	)
+
+	// Look up and filter out tasks that don't exist on version.
+	VersionTasksKey := "VersionTasks"
```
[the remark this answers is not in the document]
ablack12: Since it's only used in this function it could be okay to leave here but shouldn't be an uppercase constant

model/task/task.go
```diff
@@ -3669,6 +3669,58 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
...
+	taskMatch := []bson.M{
+		{"$eq": []string{"$" + BuildVariantKey, "$$" + BuildVariantKey}},
+		{"$eq": []string{"$" + DisplayNameKey, "$$" + DisplayNameKey}},
+	}
+	pipeline = append(pipeline, []bson.M{
+		{"$lookup": bson.M{
```
khelif96, line 3688: `$lookups` are inherently slower and we will notice a perf hit on larger versions. Since we already know the both of the versions we can can accomplish the same thing much more performantly by using a `$group` instead.
1. `$match` on both version ids
2. `$match` on `ActivatedTimeKey: {"$ne": utility.ZeroTime}`
3. `$group` tasks together by buildvariant and displayname
4. `$unwind` the tasks and get their status
5. `$profit`

model/task/task.go
```diff
@@ -3669,6 +3669,58 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
 
 }
 
+// GetMatchingBaseTasks returns the corresponding base tasks for tasks on a version.
+// If a task ran on the base version but was not scheduled on the version, then it won't be returned.
+func GetMatchingBaseTasks(versionID string, baseVersionID string) ([]Task, error) {
```
khelif96, line 3674: We technically don't need to return all of the tasks here we only need their statuses. It would be prefferable if we filtered out the statuses in the db and just return the already unique statuses instead of a full task doc.

khelif96 (changes requested):

model/task/task.go
minnakt: Hmm so I'm not sure if this can be straightforward-ly accomplished without a `$lookup`, but I did push up a version of the pipeline which is what I'd imagine it would look like without a lookup (heavily annotated to explain what's going on)

model/task/task.go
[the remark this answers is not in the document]
malikchaya2: If we want to keep it here, we can rewrite it as 
`const versionTasksKey = "VersionTasks"`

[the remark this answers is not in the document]
malikchaya2: however, we do already have `dependencyKey = "dependencies"` up top, so adding it there would keep them together.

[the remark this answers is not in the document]
ablack12: Fair; I also don't really think that should be a top-of-file constant either though since it's just a variable to make sure the string is used consistently within a query, rather than a constant that defines some behavior. I'm not invested either way though as long as it's private.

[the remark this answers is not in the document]
malikchaya2: same

khelif96 (changes requested):
For some reason this doesn't work on spruce main for me
image

Here is the patch i tried (This will likely break the e2e tests)
http://localhost:3000/version/5e4ff3abe3c3317e352062e4/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

khelif96:
I ended up rewriting the pipeline like so. I tested this out in compass and it seems to work. You would just need to convert this to go and add the display status parts and the activation part for the current version id. But i think this should be faster and simpler than your approach.
```js
[

// First fetch all of the tasks 
{$match: {
 version: {
  $in: [
   <version_id>,
   <baseversion_id>
  ]
 }
}},
// Group them by display name and variant and store the status and version ids (we need version id for later)
 {$group: {
 _id: {
  display_name: '$display_name',
  build_variant: '$build_variant'
 },
 tasks: {
  $push: {
   status: '$status',
   version: '$version'
  }
 }
}}, 
// Find all of the grouped elements that have 2 tasks (Aka tasks that existed in both versions)
{$match: {
 tasks: {
  $size: 2
 }
}}, 
// unwind the tasks so they are in their own documents.
{$unwind: {
 path: '$tasks'
}},
// Dispose of all the current version and only keep the base version tasks.
 {$match: {
 'tasks.version': <baseversion_id>
}}, 
// group all of the statuses so we only have the unique statuses.
{$group: {
 _id: {
  status: '$tasks.status'
 }
}}]

```

minnakt:
#### Pipeline
Used the pipeline with some tweaks

1. Since we only want the _activated_ tasks on the version, I also matched on that field
2. Since we need the display status as opposed to the regular status for the UI, I added that into the pipeline

#### Missing and Incorrect Data
The reason why the base filter is now empty on the version you were looking at is because the existing logic was incorrect for mainline commits. It seems like we don't have the data for the previous commit's tasks in the local database, so I added it. 

I also changed some task data in `patch-2` (I honestly don't know what this patch is) because for some reason the database contained two copies of `dist` for the version `evergreen_33016573166a36bd5f46b4111151899d5c4e95b1`. This caused my pipeline to fail since it expects at most one copy on the version and one copy on the base version.

#### e2e tests
This PR will break the [e2e tests](https://spruce.mongodb.com/version/629a7c40a4cf471d68a4e1d8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) since the contents of the base status filter will change. I will put up a Spruce PR to fix it.

model/task/task.go
```diff
@@ -3709,6 +3709,67 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
...
+	pipeline := []bson.M{}
+	taskField := "tasks"
+
+	// Fetch all ACTIVATED tasks from version, and all tasks from base version.
+	pipeline = append(pipeline, bson.M{
+		"$match": bson.M{
+			"$or": []bson.M{
+				{VersionKey: baseVersionID},
+				{VersionKey: versionID, ActivatedTimeKey: bson.M{"$ne": utility.ZeroTime}},
+			},
```
khelif96, line 3724: Nice :)

model/task/task.go
```diff
@@ -3709,6 +3709,67 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
...
+		}})
+	// Add display status, then group by display name and build variant
+	pipeline = append(pipeline, addDisplayStatus)
```
khelif96, line 3727: fyi known issues won't be surfaced here because the known-issue status is never added before this step.

khelif96 (approved):
Great job with this ! One comment. Can you also run a e2e test patch against this to make sure it don't fail once this is merged.

model/task/task.go
minnakt: Oh yeah, I see what you mean, because it doesn't run the annotation facet 😓

But I think it might actually be impossible for any base status in the Tasks table to show as `Known Issue`, since we never run the [annotation facet](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L3837) directly on the base tasks. I tried it out in Compass to see if the facet + setting ROOT affects the Collection in `$lookup` but it seems like it doesn't.

Since it seems like there'd be no real effect of adding the annotations, I'll merge this as is.

graphql/resolvers.go
```diff
@@ -3708,24 +3708,22 @@ func (r *versionResolver) TaskStatuses(ctx context.Context, v *restModel.APIVers
 }
 
 func (r *versionResolver) BaseTaskStatuses(ctx context.Context, v *restModel.APIVersion) ([]string, error) {
-	baseVersion, err := model.VersionFindOne(model.BaseVersionByProjectIdAndRevision(*v.Project, *v.Revision).WithFields(model.VersionIdentifierKey))
+	var baseVersion *model.Version
+	var err error
+
+	if !evergreen.IsPatchRequester(utility.FromStringPtr(v.Requester)) { // Get previous commit if mainline commit.
```
ablack12, line 3714: Small note that this won't work for Periodic builds, since they aren't technically patches but do use the patch revision order number for whatever reason (https://github.com/evergreen-ci/evergreen/blob/698add52308336caf4e8b56f6e0e9c3e118be296/repotracker/repotracker.go#L842). I think these should actually show the base commit.

model/task/task.go
```diff
@@ -3709,6 +3709,70 @@ func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOption
...
+
+	// Fetch all activated tasks from version, and all tasks from base version
+	pipeline = append(pipeline, bson.M{
+		"$match": bson.M{
+			"$or": []bson.M{
+				{VersionKey: baseVersionID},
+				{VersionKey: versionID, ActivatedTimeKey: bson.M{"$ne": utility.ZeroTime}},
+			},
+		}})
+	// Add display status
```
ablack12, line 3725: Thank you for all of these comments!

PR #5624, opened 2022-05-03
EVG-15548 Expose confidence score field 

khelif96:
[EVG-15548](https://jira.mongodb.org/browse/EVG-15548)

### Description 
Exposes the confidence score field for IssueLinks.
Changes its type to Float64 instead of Float32 so it can be used more easily with graphql. (gqlgen expects floats to be Float64)


minnakt (approved):
it looks good to me once the `float32` call on `L303` is fixed in `annotations_test.go`! :]

hadjri (approved):

PR #5625, opened 2022-05-03
EVG-16788 add HasUsedMainlineCommitsBefore field and fix setting the values

khelif96:
[EVG-16788](https://jira.mongodb.org/browse/EVG-16788)

### Description 
Adds a new field call `hasUsedMainlineCommitsBefore` to the UseSpruceOptions struct this will be used to trigger loading the MainlineCommits modal. 

Fixed a bug where updating a field in `UseSpruceOptions` would reset the rest of the fields to their default value. 

We were [working around it before](https://github.com/evergreen-ci/spruce/blob/main/src/components/WelcomeModal/WelcomeModal.tsx#L69) but this should handle fixing it permanently.


util/changes.go
```diff
@@ -0,0 +1,35 @@
+package util
```
khelif96, line 1: Not sure if this is the right location for this function. Glad to move it if necessary.

util/changes.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"reflect"
+)
+
+func RecursivelySetUndefinedFields(structToSet, structToDefaultFrom reflect.Value) {
```
khelif96, line 7: Needed to make some changes from the original implementation as this was panicking when used with the `UserSettings` Struct.

minnakt, line 7: I think this function needs a documentation comment!

util/reflect.go
```diff
@@ -46,3 +46,7 @@ func IsFieldUndefined(v reflect.Value) bool {
 		return false
 	}
 }
+
+func IsFieldPtr(v reflect.Value) bool {
```
minnakt, line 50: I would guess this needs a comment as well?

util/changes.go
```diff
@@ -0,0 +1,35 @@
...
+	if structToSet.Kind() == reflect.Ptr {
+		structToSet = structToSet.Elem()
+	}
+	if structToDefaultFrom.Kind() == reflect.Ptr {
+		structToDefaultFrom = structToDefaultFrom.Elem()
+	}
```
minnakt, line 13: I ended up removing this block of code and running the tests for this function and for functions that call this function, and they all still passed. It might be worth it to add a test case that specifically addresses this block of code

khelif96: Done. It was used to guard against deeply nested nil values in defaulted from structs.

util/changes.go
```diff
@@ -0,0 +1,42 @@
+package util
+
+import (
+	"reflect"
+)
+
+// RecursivelySetUndefinedFields sets all fields that are not set in structToSet to the value of the corresponding field in structToDefaultFrom.
+// It takes two reflect.Values, the first is the struct to set the fields on, and the second is the struct to get the default values from.
+func RecursivelySetUndefinedFields(structToSet, structToDefaultFrom reflect.Value) {
+	if structToSet.Kind() == reflect.Ptr {
```
ablack12, line 10: Can you explain this logic? It looks to me like the issue would be fixed if you just didn't pass in pointers, but I'm not sure.

ablack12: I think this should also have a comment, since reflect is confusing and we won't know looking back later why you've done this.

util/changes.go
```diff
@@ -0,0 +1,42 @@
...
+		// and we don't check for subfields. This allows us to group some settings together as defined or undefined.
+		if IsFieldUndefined(branchField) {
+			reflectedField := structToDefaultFrom.Field(i)
+			branchField.Set(reflectedField)
+
+			// If the field is a struct and isn't undefined, then we check each subfield recursively.
+		} else if branchField.Kind() == reflect.Struct {
+			RecursivelySetUndefinedFields(branchField, structToDefaultFrom.Field(i))
+		} else if IsFieldPtr(branchField) {
+			// If the field is a struct and isn't undefined, then we check each subfield recursively.
```
ablack12, line 36: Here are you saying, "if the nested structure is a pointer, but is undefined, still set fields recursively?" I'm pretty sure this will break ParserProject logic, which assumes that defined pointers are themselves a defined field, even if not all inner fields are set (although I'm not sure[ if the test](https://github.com/evergreen-ci/evergreen/blob/9ebdda2fa9564930e86737e2cde6408c1ce94382/model/project_ref_test.go#L110) covers that case, may need to add to it to verify). 

If this is the case, we may need to add a boolean like "recurseOnPointers".

(Also nit that this comment shouldn't just be copied over).

ablack12 (changes requested):

util/changes.go
ablack12: This belongs better in reflect.go

util/changes_test.go
```diff
@@ -0,0 +1,93 @@
...
+			Inner: &inner{
+				Foo: "bar",
+			},
+		},
+	}
+
+	aPtr := &a
+	bPtr := &b
+
+	reflectedA := reflect.ValueOf(aPtr).Elem()
```
ablack12, line 51: This test would benefit from some comments explaining what case each RecursivelySetUndefinedFields call is testing.

util/changes.go
khelif96: Since this function is recursive we don't have any say if a pointer gets passed in as one of the nested fields may have a pointer. This casing handles that case.

util/changes.go
ablack12: This also is reflected (no pun intended but lol) in the comment:
```
		// and we don't check for subfields. This allows us to group some settings together as defined or undefined.
```

SupaJoon (approved):

bynn (approved):

util/reflect.go
```diff
@@ -46,3 +46,44 @@ func IsFieldUndefined(v reflect.Value) bool {
...
 	}
 }
+
+// IsFieldPtr returns a boolean indicating whether the reflected value is a pointer.
+func IsFieldPtr(v reflect.Value) bool {
+	return v.Kind() == reflect.Ptr
+}
+
+// RecursivelySetUndefinedFields sets all fields that are not set in structToSet to the value of the corresponding field in structToDefaultFrom.
+// It takes two reflect.Values, the first is the struct to set the fields on, and the second is the struct to get the default values from.
```
ablack12, line 56: Thank you for the comment! This second sentence is probably not necessary since the variable names are explanatory but I'm not against it

util/changes.go
ablack12: @khelif96 did you see this comment? I'm still concerned about this.

khelif96: I think that is guarded against on line 76 by this case in `IsFieldUndefined`
https://github.com/evergreen-ci/evergreen/blob/44ce47e9fb36162d512f29f0fd890874b699221c/util/reflect.go#L27-L28
if the branchField is a pointer and is undefined then we just defaultFrom the whole field. We don't recurse on it in that case and we set the whole struct. (Thus setting defined and undefined fields.)

We only recurse on it if the field is a struct or is a pointer to a defined struct. 

I think this test should capture that case but feel free to correct me if i'm wrong.
https://github.com/evergreen-ci/evergreen/pull/5625/files#diff-c2b5cccbb3c39a7a30e5046a24c02a28a7fe993a16dbbcfd283c80390491fad1R71-R92

ablack12: Oh yeah that actually looks good; it seems we only don't recurse if it's a struct (and not a pointer) and that's maintained.

util/reflect.go
```diff
@@ -46,3 +46,41 @@ func IsFieldUndefined(v reflect.Value) bool {
...
 }
+
+// IsFieldPtr returns a boolean indicating whether the reflected value is a pointer.
+func IsFieldPtr(v reflect.Value) bool {
+	return v.Kind() == reflect.Ptr
+}
+
+// RecursivelySetUndefinedFields sets all fields that are not set in structToSet to the value of the corresponding field in structToDefaultFrom.
+func RecursivelySetUndefinedFields(structToSet, structToDefaultFrom reflect.Value) {
+
```
ablack12, line 57: nit: we don't need this space

ablack12 (approved):
🚀

PR #5626, opened 2022-05-03
EVG-16750: add error logs to event when host modification fails

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16750

### Description 
The host event logs currently only say whether a spawn host modification succeeded or failed. This adds the error reason to the host event logs so that users can see why the modification failed.

### Testing 
Updated existing unit tests for spawn host modification jobs to check event logs.

### DIFF
--- a/model/event/host_event.go
+++ b/model/event/host_event.go
@@ -110,20 +110,46 @@ func LogHostCreated(hostId string) {
 	LogHostEvent(hostId, EventHostCreated, HostEventData{})
 }
 
-func LogHostStartFinished(hostId string, successful bool) {
-	LogHostEvent(hostId, EventHostStarted, HostEventData{Successful: successful})
+// LogHostStartSucceeded logs an event indicating that the host was successfully
+// started.
+func LogHostStartSucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostStarted, HostEventData{Successful: true})
 }
 
+// LogHostStartSucceeded logs an event indicating that the host errored while
+// starting.
+func LogHostStartError(hostID, logs string) {
+	LogHostEvent(hostID, EventHostStarted, HostEventData{Successful: false, Logs: logs})
+}
+
+// LogHostStopSucceeded logs an event indicating that the host was successfully
+// stopped.
+func LogHostStopSucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostStopped, HostEventData{Successful: true})
+}
+
+// LogHostStopError logs an event indicating that the host errored while
+// stopping.
+func LogHostStopError(hostID, logs string) {
+	LogHostEvent(hostID, EventHostStopped, HostEventData{Successful: false, Logs: logs})
+}
+
+// LogHostModifySucceeded logs an event indicating that the host was
+// successfully modified.
+func LogHostModifySucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostModified, HostEventData{Successful: true})
+}
+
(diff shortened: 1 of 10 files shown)


model/event/host_event.go
```diff
@@ -110,20 +110,46 @@ func LogHostCreated(hostId string) {
...
+// started.
+func LogHostStartSucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostStarted, HostEventData{Successful: true})
 }
 
-func LogHostFallback(hostId string) {
-	LogHostEvent(hostId, EventHostFallback, HostEventData{})
+// LogHostStartSucceeded logs an event indicating that the host errored while
+// starting.
+func LogHostStartError(hostID, logs string) {
```
ablack12, line 121: I think it would also be fine to just keep LogHostStartFinished and also take in a log string, rather than having many versions of all these functions, since settings logs to the empty string isn't itself a problem, but I don't feel strongly

model/event/host_event.go
```diff
@@ -110,20 +110,46 @@ func LogHostCreated(hostId string) {
...
-	LogHostEvent(hostId, EventHostStarted, HostEventData{Successful: successful})
+// LogHostStartSucceeded logs an event indicating that the host was successfully
+// started.
+func LogHostStartSucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostStarted, HostEventData{Successful: true})
 }
 
-func LogHostFallback(hostId string) {
-	LogHostEvent(hostId, EventHostFallback, HostEventData{})
+// LogHostStartSucceeded logs an event indicating that the host errored while
```
ablack12, line 119: Wrong function name

model/event/host_event.go
```diff
@@ -110,20 +110,46 @@ func LogHostCreated(hostId string) {
...
+
+// LogHostModifySucceeded logs an event indicating that the host was
+// successfully modified.
+func LogHostModifySucceeded(hostID string) {
+	LogHostEvent(hostID, EventHostModified, HostEventData{Successful: true})
 }
 
-func LogHostStopFinished(hostId string, successful bool) {
-	LogHostEvent(hostId, EventHostStopped, HostEventData{Successful: successful})
+// LogHostModifySucceeded logs an event indicating that the host errored while
```
ablack12, line 143: Wrong function name

units/spawnhost_start.go
```diff
@@ -59,11 +59,12 @@ func (j *spawnhostStartJob) Run(ctx context.Context) {
 
 	startCloudHost := func(mgr cloud.Manager, h *host.Host, user string) error {
 		if err := mgr.StartInstance(ctx, h, user); err != nil {
-			event.LogHostStartFinished(h.Id, false)
-			return errors.Wrapf(err, "starting spawn host '%s'", h.Id)
+			err = errors.Wrap(err, "starting spawn host")
+			event.LogHostStartError(h.Id, err.Error())
```
ablack12, line 63: Do we need the wrapped error since it'll always start with the same thing, and the event log itself has the started event?

model/event/host_event.go
Kimchelly: I do this way too often 🤦 

Fixed.

model/event/host_event.go
Kimchelly: Fixed.

units/spawnhost_start.go
Kimchelly: Probably not. I moved the wrap back to the return value.

ablack12 (approved):

PR #5627, opened 2022-05-03
EVG-12923 only log if we unblock more tasks than we block

ablack12:

### Description 
The origin of this change _seems_ to be that we sometimes have more blocked tasks than we really should, so we first unblock everything possible, and then re-run Block to make sure we re-block everything that was actually supposed to be blocked. It was hard for me to discern from logs if these always modify the same tasks, so updated the logging to be 
A) just for single-host task groups, since the problem seems limited to this interaction (if we run into errors doing this then we know the problem is larger than we expected), and 
B) just to log the difference between blocked/unblocked (if there is any)

### Testing 
Modified tests to ensure compiling IDs recursively works.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -518,13 +518,28 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		return errors.Wrap(err, "marking task finished")
 	}
 
-	if err = UpdateUnblockedDependencies(t, true, "MarkEnd"); err != nil {
-		return errors.Wrap(err, "updating unblocked dependencies")
+	unblockedIds := []string{}
+	if t.IsPartOfSingleHostTaskGroup() {
+		unblockedIds, err = UpdateUnblockedDependencies(t)
+		if err != nil {
+			return errors.Wrap(err, "updating unblocked dependencies")
+		}
 	}
 
-	if err = UpdateBlockedDependencies(t); err != nil {
+	blockedIds, err := UpdateBlockedDependencies(t)
+	if err != nil {
 		return errors.Wrap(err, "updating blocked dependencies")
 	}
+
+	unblockedAndNotBlockedIds, _ := utility.StringSliceSymmetricDifference(unblockedIds, blockedIds)
+	// The question to determine here is: do we ever unblock more dependencies than we block?
+	grip.DebugWhen(len(unblockedAndNotBlockedIds) > 0, message.Fields{
+		"message":                       "unblocked task group dependent tasks",
+		"ticket":                        "EVG-12923",
+		"unblocked_and_not_blocked_ids": unblockedIds,
+		"task":                          t.Id,
+		"caller":                        caller,
+	})
 
 	if err = t.MarkDependenciesFinished(true); err != nil {
 		return errors.Wrap(err, "updating dependency met status")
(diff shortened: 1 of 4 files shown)


units/check_blocked_tasks.go
```diff
@@ -139,7 +139,8 @@ func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.T
 	if err == nil {
 		for _, blockingTask := range blockingTasks {
 			blockingTaskIds = append(blockingTaskIds, blockingTask.Id)
-			catcher.Wrapf(model.UpdateBlockedDependencies(&blockingTask), "can't update blocked dependencies for '%s'", blockingTask.Id)
+			_, err = model.UpdateBlockedDependencies(&blockingTask)
+			catcher.Wrapf(err, "can't update blocked dependencies for '%s'", blockingTask.Id)
```
Kimchelly, line 143: To follow the style guide:
`catcher.Wrapf(err, "updating blocked dependencies for task '%s'", blockingTask.Id)`

model/task_lifecycle.go
```diff
@@ -518,14 +518,29 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
...
+	if err != nil {
 		return errors.Wrap(err, "updating blocked dependencies")
 	}
 
+	unblockedAndNotBlockedIds, _ := utility.StringSliceSymmetricDifference(unblockedIds, blockedIds)
+	// The question to determine here is: do we ever unblock more dependencies than we block?
+	grip.DebugWhen(len(unblockedAndNotBlockedIds) > 0, message.Fields{
+		"message":                       "unblocked task group dependent tasks",
+		"ticket":                        "EVG-12923",
+		"unblocked_and_not_blocked_ids": unblockedIds,
```
Kimchelly, line 539: Is this meant to log `unblockedAndNotBlockedIds`, or all the unblocked ones (including the momentarily-unblocked ones)?

model/task_lifecycle.go
```diff
@@ -581,74 +596,57 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 
 // UpdateBlockedDependencies traverses the dependency graph and recursively sets each
 // parent dependency as unattainable in depending tasks.
-func UpdateBlockedDependencies(t *task.Task) error {
+// REturns a list of modified task IDs.
```
Kimchelly, line 599: Returns

model/task_lifecycle.go
```diff
@@ -518,14 +518,29 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
...
 	}
 
-	if err = UpdateBlockedDependencies(t); err != nil {
+	blockedIds, err := UpdateBlockedDependencies(t)
+	if err != nil {
 		return errors.Wrap(err, "updating blocked dependencies")
 	}
 
+	unblockedAndNotBlockedIds, _ := utility.StringSliceSymmetricDifference(unblockedIds, blockedIds)
+	// The question to determine here is: do we ever unblock more dependencies than we block?
```
Kimchelly, line 535: I don't quite understand this comment, can you clarify it a bit? The log seems to trigger when we unblock dependencies that remain unblocked after updating blocked dependencies, which is not dependent on the relative number of dependencies we block/unblock.

Should this be phrased as something like "do we ever unblock dependencies that don't immediately get re-blocked"?

Maybe I don't understand this log, since it seems correct to me that a previously-unattainable dependency might be unblocked because it's now attainable.

model/task_lifecycle.go
```diff
@@ -581,74 +596,57 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 
 // UpdateBlockedDependencies traverses the dependency graph and recursively sets each
 // parent dependency as unattainable in depending tasks.
-func UpdateBlockedDependencies(t *task.Task) error {
+// REturns a list of modified task IDs.
+func UpdateBlockedDependencies(t *task.Task) ([]string, error) {
 	const maxStackTraceSize = 5
```
Kimchelly, line 601: Can this be removed? It's not used anymore.

model/task_lifecycle.go
```diff
@@ -518,14 +518,29 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		return errors.Wrap(err, "marking task finished")
 	}
 
-	if err = UpdateUnblockedDependencies(t, true, "MarkEnd"); err != nil {
-		return errors.Wrap(err, "updating unblocked dependencies")
+	unblockedIds := []string{}
+	if t.IsPartOfSingleHostTaskGroup() {
+		unblockedIds, err = UpdateUnblockedDependencies(t)
```
Kimchelly, line 523: I don't really follow the unblocking/blocking logic in the task lifecycle, so this might be wrong - Don't tasks outside of single-host task groups still need to unblock attainable dependencies? Otherwise, how do they mark tasks that were previously unattainable as now attainable?

model/task_lifecycle.go
```diff
@@ -581,74 +596,57 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
...
-		grip.Debug(message.Fields{
-			"message":          "unblocked task group dependent tasks",
-			"ticket":           "EVG-12923",
-			"blocked_task_ids": blockedTaskIds,
-			"task":             t.Id,
-			"caller":           caller,
-		})
+	blockedTaskIds := []string{}
+	for _, blockedTask := range blockedTasks {
+		blockedTaskIds = append(blockedTaskIds, blockedTask.Id)
```
Kimchelly, line 635: Same question as in `UpdateBlockedDependencies` about adding the task IDs in the for loop below this vs. adding it in a separate loop.

model/task_lifecycle.go
```diff
@@ -581,74 +596,57 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
...
-			"num_dependent_tasks_to_update": len(dependentTasks),
-			"dependent_task_ids_to_update":  depTaskIds,
-			"stack":                         stack,
-		})
+		return nil, errors.Wrapf(err, "getting tasks depending on task '%s'", t.Id)
+	}
+
+	depTaskIds := []string{}
+	for _, depTask := range dependentTasks {
+		depTaskIds = append(depTaskIds, depTask.Id)
```
Kimchelly, line 609: Could this just be put in the for loop below this? Is the order of task IDs important?

Kimchelly (changes requested):

model/task_lifecycle.go
ablack12: The idea here is that UnblockedDependencies was added as a hot fix (that was only affecting single-host task groups) because of a bug with restarting that we didn't understand (when restarting a task, all dependencies should be marked as unattainable, so we shouldn't get into the state here that we're unblocking anything that won't be immediately re-blocked at 52).

I'm trying to understand if this is still a bug or not, since I'm not seeing any errors from our other logging that's supposed to catch this, and there've been lots of logic refactors since this was introduced. If I don't see anything suspicious, I'm going to attempt removing Unblocked altogether.

If preferred, we could make removing this under a feature flag and just handle it all together as the next piece.

model/task_lifecycle.go
ablack12: Yes the former 🤦‍♀️

model/task_lifecycle.go
ablack12: I'm fine to re-phrase the comment.

model/task_lifecycle.go
```diff
@@ -518,14 +518,29 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
...
 	}
 
-	if err = UpdateBlockedDependencies(t); err != nil {
+	blockedIds, err := UpdateBlockedDependencies(t)
+	if err != nil {
 		return errors.Wrap(err, "updating blocked dependencies")
 	}
 
+	permanentlyUnblockedIds, _ := utility.StringSliceSymmetricDifference(unblockedIds, blockedIds)
+	// The question to determine here is: do we ever unblock dependencies that don't immediately get re-blockeds?
```
Kimchelly, line 535: re-blocked

model/task_lifecycle.go
Kimchelly: Makes sense. I guess I have a follow-on question - if unblocking is only meant for single-host task groups but the blocking is done for all tasks, where do tasks that are not in single host task groups unblock their attainable dependencies? I did some code reading, but didn't find a definitive place where unattainable would be unset when a non-single-host-task-group task finishes.

model/task_lifecycle.go
ablack12: 🤦‍♀️

model/task_lifecycle.go
ablack12: Tasks are only set as "unattainable" when they finish, so they only become "attainable" again if they're restarted, which is something we handle in both MarkOneTaskReset and MarkTasksReset. I believe these should cover all restart cases for both single-host tasks and regular tasks -- if this logging doesn't uncover any edge cases for single-host task groups then I think we can get rid of this block completely, although possibly single-host task groups have some race with how they block old tasks.

Since this is a theory though I'll change this to still check and log everything for now, and include in the log whether or not it's a single-host task group.

Kimchelly: Ah yeah, it makes sense now. Thanks for the explanation!

Kimchelly (approved):

PR #5628, opened 2022-05-03
EVG-16679 dependency graph

ybrill:

### Description 
When I was looking to validate that generate tasks doesn't create dependency cycles I wrote a thing to abstract away the graph computer sciencey stuff and used it to simulate the new dependency graph before actually adding the dependencies. The case to use the DependencyGraph in other places seemed compelling so I began refactoring other places to use it too. 

This PR includes the DependencyGraph itself and applying it to two problems in the project validator. Future PRs will use it to validate adding dependencies on generated tasks, getRecursiveDependencies (down and up), and for the DAG dispatcher.

### Testing 
Unit testing

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -785,6 +785,10 @@ func IsPatchRequester(requester string) bool {
 	return requester == PatchVersionRequester || IsGitHubPatchRequester(requester)
 }
 
+func IsNonPatchRequester(requester string) bool {
+	return utility.StringSliceContains([]string{RepotrackerVersionRequester, TriggerRequester, MergeTestRequester, AdHocRequester}, requester)
+}
+
 func IsGitHubPatchRequester(requester string) bool {
 	return requester == GithubPRRequester || requester == MergeTestRequester
 }
--- a/globals.go
+++ b/globals.go
@@ -512,6 +512,16 @@ const (
 	AdHocRequester              = "ad_hoc"
 )
 
+var AllRequesterTypes = []string{
+	PatchVersionRequester,
+	GithubPRRequester,
+	GitTagRequester,
+	RepotrackerVersionRequester,
+	TriggerRequester,
+	MergeTestRequester,
+	AdHocRequester,
+}
+
 // Constants related to requester types.
 var (
 	SystemVersionRequesterTypes = []string{
@@ -783,10 +793,6 @@ func IsSystemActivator(caller string) bool {
 
 func IsPatchRequester(requester string) bool {
 	return requester == PatchVersionRequester || IsGitHubPatchRequester(requester)
-}
-
-func IsNonPatchRequester(requester string) bool {
-	return utility.StringSliceContains([]string{RepotrackerVersionRequester, TriggerRequester, MergeTestRequester, AdHocRequester}, requester)
 }
 
 func IsGitHubPatchRequester(requester string) bool {
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -746,44 +746,19 @@ func CreateBuildFromVersionNoInsert(args BuildCreateArgs) (*build.Build, task.Ta
 }
 
 func CreateTasksFromGroup(in BuildVariantTaskUnit, proj *Project, requester string) []BuildVariantTaskUnit {
-	tasks := []BuildVariantTaskUnit{}
 	tg := proj.FindTaskGroup(in.Name)
 	if tg == nil {
-		return tasks
+		return nil
 	}
+	tasks := proj.TasksFromGroup(in, *tg)
 
-	taskMap := map[string]ProjectTask{}
-	for _, projTask := range proj.Tasks {
-		taskMap[projTask.Name] = projTask
-	}
-
-	for _, t := range tg.Tasks {
-		bvt := BuildVariantTaskUnit{
-			Name: t,
-			// IsGroup is not persisted, and indicates here that the
-			// task is a member of a task group.
-			IsGroup:          true,
-			GroupName:        in.Name,
-			Patchable:        in.Patchable,
-			PatchOnly:        in.PatchOnly,
-			Disable:          in.Disable,
-			AllowForGitTag:   in.AllowForGitTag,
-			GitTagOnly:       in.GitTagOnly,
-			Priority:         in.Priority,
-			DependsOn:        in.DependsOn,
-			RunOn:            in.RunOn,
-			ExecTimeoutSecs:  in.ExecTimeoutSecs,
-			Stepback:         in.Stepback,
-			Activate:         in.Activate,
-			CommitQueueMerge: in.CommitQueueMerge,
-		}
-		// Default to project task settings when unspecified
-		bvt.Populate(taskMap[t])
+	var willRun []BuildVariantTaskUnit
+	for _, bvt := range tasks {
 		if !bvt.IsDisabled() && !bvt.SkipOnRequester(requester) {
-			tasks = append(tasks, bvt)
+			willRun = append(willRun, bvt)
 		}
 	}
-	return tasks
+	return willRun
 }
 
 // createTasksForBuild creates all of the necessary tasks for the build.  Returns a
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -746,14 +746,8 @@ func CreateBuildFromVersionNoInsert(args BuildCreateArgs) (*build.Build, task.Ta
 }
 
 func CreateTasksFromGroup(in BuildVariantTaskUnit, proj *Project, requester string) []BuildVariantTaskUnit {
-	tg := proj.FindTaskGroup(in.Name)
-	if tg == nil {
-		return nil
-	}
-	tasks := proj.TasksFromGroup(in, *tg)
-
 	var willRun []BuildVariantTaskUnit
-	for _, bvt := range tasks {
+	for _, bvt := range proj.tasksFromGroup(in) {
 		if !bvt.IsDisabled() && !bvt.SkipOnRequester(requester) {
 			willRun = append(willRun, bvt)
 		}
--- a/model/project.go
+++ b/model/project.go
@@ -271,7 +271,7 @@ func (bvt *BuildVariantTaskUnit) IsDisabled() bool {
 	return utility.FromBoolPtr(bvt.Disable)
 }
 
-func (bvt *BuildVariantTaskUnit) ToTaskNode() task.TaskNode {
+func (bvt *BuildVariantTaskUnit) toTaskNode() task.TaskNode {
 	return task.TaskNode{
 		Name:    bvt.Name,
 		Variant: bvt.Variant,
@@ -1457,23 +1457,15 @@ func (p *Project) FindAllVariants() []string {
 // for all variants of a project.
 func (p *Project) FindAllBuildVariantTasks() []BuildVariantTaskUnit {
 	tasksByName := map[string]ProjectTask{}
-	taskGroups := map[string]TaskGroup{}
 	for _, t := range p.Tasks {
 		tasksByName[t.Name] = t
-	}
-	for _, tg := range p.TaskGroups {
-		taskGroups[tg.Name] = tg
 	}
 	allBVTs := []BuildVariantTaskUnit{}
 	for _, b := range p.BuildVariants {
 		for _, t := range b.Tasks {
 			t.Variant = b.Name
 			if t.IsGroup {
-				group, exists := taskGroups[t.Name]
-				if !exists {
-					continue
-				}
-				allBVTs = append(allBVTs, p.TasksFromGroup(t, group)...)
+				allBVTs = append(allBVTs, p.tasksFromGroup(t)...)
 			} else {
 				t.Populate(tasksByName[t.Name])
 				allBVTs = append(allBVTs, t)
@@ -1483,8 +1475,14 @@ func (p *Project) FindAllBuildVariantTasks() []BuildVariantTaskUnit {
 	return allBVTs
 }
 
-// TasksFromGroup returns a slice of the task group's tasks. groupTask is merged with each of the tasks in the group.
-func (p *Project) TasksFromGroup(groupTask BuildVariantTaskUnit, tg TaskGroup) []BuildVariantTaskUnit {
+// tasksFromGroup returns a slice of the task group's tasks.
+// Settings missing from the group task are populated from the task definition.
+func (p *Project) tasksFromGroup(bvTaskGroup BuildVariantTaskUnit) []BuildVariantTaskUnit {
+	tg := p.FindTaskGroup(bvTaskGroup.Name)
+	if tg == nil {
+		return nil
+	}
+
 	tasks := []BuildVariantTaskUnit{}
 	taskMap := map[string]ProjectTask{}
 	for _, projTask := range p.Tasks {
@@ -1493,23 +1491,26 @@ func (p *Project) TasksFromGroup(groupTask BuildVariantTaskUnit, tg TaskGroup) [
 
 	for _, t := range tg.Tasks {
 		bvt := BuildVariantTaskUnit{
-			Name:             t,
+			Name: t,
+			// IsGroup is not persisted, and indicates here that the
+			// task is a member of a task group.
 			IsGroup:          true,
-			Variant:          groupTask.Variant,
-			GroupName:        groupTask.Name,
-			Patchable:        groupTask.Patchable,
-			PatchOnly:        groupTask.PatchOnly,
-			Disable:          groupTask.Disable,
-			AllowForGitTag:   groupTask.AllowForGitTag,
-			GitTagOnly:       groupTask.GitTagOnly,
-			Priority:         groupTask.Priority,
-			DependsOn:        groupTask.DependsOn,
-			RunOn:            groupTask.RunOn,
-			ExecTimeoutSecs:  groupTask.ExecTimeoutSecs,
-			Stepback:         groupTask.Stepback,
-			Activate:         groupTask.Activate,
-			CommitQueueMerge: groupTask.CommitQueueMerge,
+			Variant:          bvTaskGroup.Variant,
+			GroupName:        bvTaskGroup.Name,
+			Patchable:        bvTaskGroup.Patchable,
+			PatchOnly:        bvTaskGroup.PatchOnly,
+			Disable:          bvTaskGroup.Disable,
+			AllowForGitTag:   bvTaskGroup.AllowForGitTag,
+			GitTagOnly:       bvTaskGroup.GitTagOnly,
+			Priority:         bvTaskGroup.Priority,
+			DependsOn:        bvTaskGroup.DependsOn,
+			RunOn:            bvTaskGroup.RunOn,
+			ExecTimeoutSecs:  bvTaskGroup.ExecTimeoutSecs,
+			Stepback:         bvTaskGroup.Stepback,
+			Activate:         bvTaskGroup.Activate,
+			CommitQueueMerge: bvTaskGroup.CommitQueueMerge,
 		}
+		// Default to project task settings when unspecified
 		bvt.Populate(taskMap[t])
 		tasks = append(tasks, bvt)
 	}
@@ -1958,7 +1959,7 @@ func (p *Project) DependencyGraph() task.DependencyGraph {
 	g := task.NewDependencyGraph(false)
 
 	for _, t := range tasks {
-		g.AddTaskNode(t.ToTaskNode())
+		g.AddTaskNode(t.toTaskNode())
 	}
 
 	for _, dependencyEdge := range dependenciesForTaskUnit(tasks) {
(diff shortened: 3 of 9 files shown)


ybrill:
One behavioral change worth calling out is that the validator currently errors when a dependency is specified that doesn't exist. It will no longer do this because it seemed inconsistent: we don't validate that * dependencies match anything.

validator/project_validator.go
```diff
@@ -1850,178 +1804,68 @@ func validateTaskSyncCommands(p *model.Project, runLong bool) ValidationErrors {
...
 
-type dependencyRequirements struct {
-	lastDepNeedsSuccess bool
-	requireOnPatches    bool
-	requireOnNonPatches bool
-	requireOnGitTag     bool
-}
+		var dep model.TaskUnitDependency
+		for _, dependency := range dependentTaskUnit.DependsOn {
+			if dependency.Name == dependedOn.Name && dependency.Variant == dependedOn.Variant {
```
ablack12, line 1825: It's a bit confusing for this to be called only "dep" when we have both dependent and dependedOn being used; can we call this something like dependentInfo? (just a suggestion, we can also leave it but a comment would be helpful)

validator/project_validator.go
```diff
@@ -1850,178 +1804,68 @@ func validateTaskSyncCommands(p *model.Project, runLong bool) ValidationErrors {
...
-	if depReqs.requireOnPatches && (taskUnit.SkipOnPatchBuild() || taskUnit.SkipOnNonGitTagBuild()) {
-		return false, nil
-	}
-	if depReqs.requireOnNonPatches && (taskUnit.SkipOnNonPatchBuild() || taskUnit.SkipOnNonGitTagBuild()) {
-		return false, nil
-	}
-	if depReqs.requireOnGitTag && (taskUnit.SkipOnNonPatchBuild() || taskUnit.SkipOnGitTagBuild()) {
-		return false, nil
-	}
+		if statuses != nil && dependedOn == dependedOnNode {
```
ablack12, line 1840: Can you explain the dependedOn / dependedOnNode bit?
In general I think we should try explanation comments throughout, not just at the top of the function, so this is easier to read as we go in the future.

validator/project_validator.go
```diff
@@ -1850,178 +1804,68 @@ func validateTaskSyncCommands(p *model.Project, runLong bool) ValidationErrors {
...
+	tvTaskUnitMap := tvToTaskUnit(project)
 
-	sourceTask, ok := tvToTaskUnit[source]
-	if !ok {
-		return errors.Errorf("could not find task '%s' in build variant '%s'",
-			source.TaskName, source.Variant)
-	}
+	dependentNode := task.TaskNode{Name: dependentTask.TaskName, Variant: dependentTask.Variant}
+	dependedOnNode := task.TaskNode{Name: dependedOnTask.TaskName, Variant: dependedOnTask.Variant}
+	traversal := func(edge task.DependencyEdge) bool {
```
ablack12, line 1816: It would be nice to modularize this by defining the  traversal function outside of this function.

validator/project_validator.go
```diff
@@ -1850,178 +1804,68 @@ func validateTaskSyncCommands(p *model.Project, runLong bool) ValidationErrors {
...
-		for _, depNode := range depNodes {
-			reachable, err := dependencyMustRun(target, depNode, depReqs, allNodes, visited, tvToTaskUnit)
-			if err != nil {
-				return false, errors.Wrap(err, "dependency graph has problems")
-			}
-			if reachable {
-				return true, nil
-			}
+		if statuses != nil {
+			errMsg = fmt.Sprintf(errMsg+" with status in [%s]", strings.Join(statuses, ", "))
```
ablack12, line 1863: Could we instead do `fmt.Sprintf("%s with status in [%s]", errMsg, strings.Join...)` ?

ablack12 (changes requested):
Cool refactor! My main request is for many more comments since graphs can be hard to follow. 

Would it make sense to put this behind a feature flag so we could test it out with less risk? (Or run it through staging a bit if that seems unnecessary)

model/project.go
```diff
@@ -255,10 +255,29 @@ func (bvt *BuildVariantTaskUnit) SkipOnNonGitTagBuild() bool {
 	return utility.FromBoolPtr(bvt.GitTagOnly)
 }
 
+func (bvt *BuildVariantTaskUnit) RunsOnPatches() bool {
+	return !(bvt.SkipOnPatchBuild() || bvt.SkipOnNonGitTagBuild())
```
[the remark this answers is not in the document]
ybrill: I went the second route to write that to use these functions in SkipOnRequester

model/project.go
```diff
@@ -1899,6 +1918,49 @@ func (p *Project) GetDisplayTask(variant, name string) *patch.DisplayTask {
 	return nil
 }
 
+func (p *Project) DependencyGraph() task.DependencyGraph {
```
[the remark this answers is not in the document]
ybrill: done

model/project.go
```diff
@@ -1899,6 +1918,49 @@ func (p *Project) GetDisplayTask(variant, name string) *patch.DisplayTask {
...
+	var taskNodes []task.TaskNode
+
+	for _, t := range tasks {
+		g.AddTaskNode(t.ToTaskNode())
+		taskNodes = append(taskNodes, t.ToTaskNode())
+	}
+
+	for _, t := range tasks {
+		for dep, depTasks := range dependenciesForTaskUnit(t, taskNodes) {
+			for _, depNode := range depTasks {
```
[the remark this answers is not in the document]
ybrill: I rewrote it to return a slice of `DependencyEdge`s

model/project.go
```diff
@@ -1899,6 +1918,49 @@ func (p *Project) GetDisplayTask(variant, name string) *patch.DisplayTask {
...
+	g := task.NewDependencyGraph()
+	var taskNodes []task.TaskNode
+
+	for _, t := range tasks {
+		g.AddTaskNode(t.ToTaskNode())
+		taskNodes = append(taskNodes, t.ToTaskNode())
+	}
+
+	for _, t := range tasks {
+		for dep, depTasks := range dependenciesForTaskUnit(t, taskNodes) {
```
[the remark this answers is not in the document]
ybrill: this was addressed by https://github.com/evergreen-ci/evergreen/pull/5628#discussion_r866230431

model/task/dependency_graph.go
```diff
@@ -0,0 +1,280 @@
...
+
+// String represents DependencyCycles as a string.
+func (dc DependencyCycles) String() string {
+	cycles := make([]string, 0, len(dc))
+	for _, cycle := range dc {
+		cycleStrings := make([]string, 0, len(cycle))
+		for _, node := range cycle {
+			cycleStrings = append(cycleStrings, node.String())
+		}
+		cycles = append(cycles, "["+strings.Join(cycleStrings, ", ")+"]")
```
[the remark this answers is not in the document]
ybrill: done

model/task/dependency_graph.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return strings.Join(cycles, ", ")
+}
+
+// Cycles returns cycles in the graph, if any.
+// Self-loops are not included as cycles.
+func (g *DependencyGraph) Cycles() DependencyCycles {
+	var cycles DependencyCycles
+	stronglyConnectedComponenets := topo.TarjanSCC(g.graph)
```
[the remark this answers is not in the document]
ybrill: 😓 
fixed

validator/project_validator.go
ybrill: I renamed it to `edgeInfo`

validator/project_validator.go
ybrill: I added a comment
```go
// If statuses is specified we need to check the edge's status when the edge points to the target node.
```

validator/project_validator.go
ybrill: done

validator/project_validator.go
ybrill: My aesthetic sense likes it inside since it's so specific to the logic of this function, but if you feel strongly I can move it out.

model/task/dependency_graph.go
```diff
@@ -0,0 +1,280 @@
...
+	g.buildFromTasks(tasks, false)
+	return g
+}
+
+// reversedTaskDependencyGraph constructs a DependencyGraph from tasks.
+// Edges in the graph are reversed so they point from depended on tasks to dependent tasks.
+func reversedTaskDependencyGraph(tasks []Task) DependencyGraph {
+	g := NewDependencyGraph()
+	g.buildFromTasks(tasks, true)
+	return g
```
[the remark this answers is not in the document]
ybrill: added reversed as a parameter

model/task/dependency_graph.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func taskDependencyGraph(tasks []Task) DependencyGraph {
+	g := NewDependencyGraph()
+	g.buildFromTasks(tasks, false)
+	return g
+}
+
+// reversedTaskDependencyGraph constructs a DependencyGraph from tasks.
+// Edges in the graph are reversed so they point from depended on tasks to dependent tasks.
```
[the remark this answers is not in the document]
ybrill: It will come in handy in future PRs. 

I hope to use it in the dispatcher which uses a topological sort to put depended on tasks before the tasks that depend on them. Since a topological sort orders the nodes such that for every edge `u -> v ` u comes before v we want u to be the depended on tasks and v to be the dependent tasks. I puzzled over whether it would be equivalent to use the regular graph and just reverse the result. I found [this](https://www.quora.com/Is-a-topological-order-on-the-reverse-graph-the-same-as-a-topological-order-reversed) which proves (?) that it is (and reminded me not to attempt an MS in CS), but it would be an additional effort to get it to work for the case where the topological ordering is ambiguous and we want it to default to the queue order.

It's also applicable to [getRecursiveDependenciesDown](https://github.com/evergreen-ci/evergreen/blob/31a6c66aca73f769b8b003a89c4bffaa529b3b0a/model/task/task.go#L2110-L2113). With a reversed graph we can just call `reachableFromNode` to find the recursive dependencies.

ybrill:
> Would it make sense to put this behind a feature flag so we could test it out with less risk? (Or run it through staging a bit if that seems unnecessary)

idk. I think it would depend on what criteria we use to decide when to use a feature flag. I think of them as most useful for cases where the scaffolding will be up for a while and we don't want anyone to get hurt in the interim. Another case could be where something can break everything and we want a way to turn it off faster than it would take to revert.

It seems to me we reduce risk with the strategy of using the DependencyGraph for the project validator first, before the future PRs that will use it for generate tasks, the task lifecycle, and the dispatcher.

globals.go
```diff
@@ -785,6 +785,10 @@ func IsPatchRequester(requester string) bool {
 	return requester == PatchVersionRequester || IsGitHubPatchRequester(requester)
 }
 
+func IsNonPatchRequester(requester string) bool {
+	return utility.StringSliceContains([]string{RepotrackerVersionRequester, TriggerRequester, MergeTestRequester, AdHocRequester}, requester)
```
ablack12, line 789: Could we instead just check !IsPatchRequester since it should be one or the other?

model/lifecycle.go
```diff
@@ -746,44 +746,19 @@ func CreateBuildFromVersionNoInsert(args BuildCreateArgs) (*build.Build, task.Ta
...
-
-	taskMap := map[string]ProjectTask{}
-	for _, projTask := range proj.Tasks {
-		taskMap[projTask.Name] = projTask
-	}
-
-	for _, t := range tg.Tasks {
-		bvt := BuildVariantTaskUnit{
-			Name: t,
-			// IsGroup is not persisted, and indicates here that the
```
ablack12, line 763: Why did we remove this comment?

model/lifecycle.go
```diff
@@ -746,44 +746,19 @@ func CreateBuildFromVersionNoInsert(args BuildCreateArgs) (*build.Build, task.Ta
...
-			GitTagOnly:       in.GitTagOnly,
-			Priority:         in.Priority,
-			DependsOn:        in.DependsOn,
-			RunOn:            in.RunOn,
-			ExecTimeoutSecs:  in.ExecTimeoutSecs,
-			Stepback:         in.Stepback,
-			Activate:         in.Activate,
-			CommitQueueMerge: in.CommitQueueMerge,
-		}
-		// Default to project task settings when unspecified
```
ablack12, line 780: I also think this comment was helpful.

model/project.go
```diff
@@ -1441,28 +1467,55 @@ func (p *Project) FindAllBuildVariantTasks() []BuildVariantTaskUnit {
...
-				t.Variant = b.Name
+				t.Populate(tasksByName[t.Name])
 				allBVTs = append(allBVTs, t)
 			}
 		}
 	}
 	return allBVTs
 }
 
+// TasksFromGroup returns a slice of the task group's tasks. groupTask is merged with each of the tasks in the group.
```
ablack12, line 1486: Can this comment specify that anything not specified in the task group definition groupTask is populated from the task definition (as opposed to "merge" which feels ambiguous)

model/project.go
```diff
@@ -1441,28 +1467,55 @@ func (p *Project) FindAllBuildVariantTasks() []BuildVariantTaskUnit {
...
+func (p *Project) TasksFromGroup(groupTask BuildVariantTaskUnit, tg TaskGroup) []BuildVariantTaskUnit {
```
ablack12, line 1487: nit: Can "groupTask" be renamed to taskGroupTask or bvTaskGroup? I think we don't typically abbreviate to group (except in this function name I guess haha) so this did confuse me

model/project.go
```diff
@@ -255,10 +255,36 @@ func (bvt *BuildVariantTaskUnit) SkipOnNonGitTagBuild() bool {
...
+
+func (bvt *BuildVariantTaskUnit) RunsOnGitTag() bool {
+	return !(bvt.SkipOnNonPatchBuild() || bvt.SkipOnGitTagBuild())
+}
+
 func (bvt *BuildVariantTaskUnit) IsDisabled() bool {
 	return utility.FromBoolPtr(bvt.Disable)
 }
 
+func (bvt *BuildVariantTaskUnit) ToTaskNode() task.TaskNode {
```
ablack12, line 274: Can these be private functions?

model/project.go
ablack12: Could we also  potentially only pass in tg.Tasks as taskGroupTaskNames or something instead of the whole task group?

model/task/dependency_graph.go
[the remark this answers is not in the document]
ablack12: I understood just enough of that to let this go :joy:

validator/project_validator.go
ablack12: Yeah fair; I think a function comment here would just be helpful saying when the function returns true.

validator/project_validator.go
```diff
@@ -1850,178 +1804,69 @@ func validateTaskSyncCommands(p *model.Project, runLong bool) ValidationErrors {
...
-	visited := map[model.TVPair]bool{}
-	var allTVs []model.TVPair
-	for tv := range tvToTaskUnit {
-		visited[tv] = false
-		allTVs = append(allTVs, tv)
-	}
+// validateTVDependsOnTV checks that the dependent task always has a dependency on the depended on task.
+// The dependedOnTask and every other task along the path must run on all the same requester types as the dependentTask
+// and the dependency on the dependedOnTask must be with a status in statuses, if provided.
+func validateTVDependsOnTV(dependentTask, dependedOnTask model.TVPair, statuses []string, project *model.Project) error {
```
ablack12, line 1810: I definitely prefer the original source/target naming for what it's worth -- since the variables were more different, it was easier to parse the code when reading it, whereas now you have to look closely for the "On" which separates them from each other, but that's just my 2 cents. (This also maybe separates these variables from "dependent / dependedOn" in traversal; since that's a function that iterates over everything it's hard to remember that those variable names aren't strictly related)

model/project.go
ablack12: Last comment here sorry 😅  but this function should also be private.

model/task/dependency_graph.go
```diff
@@ -0,0 +1,272 @@
...
+func (t TaskNode) String() string {
+	if t.ID != "" {
+		return t.ID
+	}
+
+	return fmt.Sprintf("%s/%s", t.Variant, t.Name)
+}
+
+// VersionDependencyGraph finds all the tasks from the version given by versionID and constructs a DependencyGraph from them.
+func VersionDependencyGraph(versionID string, transposed bool) (DependencyGraph, error) {
```
ablack12, line 73: Do we anticipate using this? (Will we remember to remove this if we don't?)

model/task/dependency_graph.go
```diff
@@ -0,0 +1,272 @@
...
+
+	line := g.graph.NewLine(fromNode, toNode)
+	g.graph.SetLine(line)
+	g.edgesToDependencies[edgeKey{from: edge.From, to: edge.To}] = edge
+}
+
+// EdgesIntoTask returns all the edges that point to t.
+// For a regular graph these edges are tasks that directly depend on t.
+// If the graph is transposed these edges are tasks t directly depends on.
+func (g *DependencyGraph) EdgesIntoTask(t TaskNode) []DependencyEdge {
```
ablack12, line 143: Can be private

globals.go
ybrill: I think because `!IsPatchRequester` includes git tag requesters SkipOnRequester would skip a git tag requester when the bvt is git tag only.

model/lifecycle.go
ybrill: restored

model/lifecycle.go
ybrill: restored

model/project.go
ybrill: done

model/project.go
ybrill: I rearranged it to just take the `bvTaskGroup`

model/project.go
ybrill: done

validator/project_validator.go
ybrill: I left the variable names in the function signature as dependent/depended on but renamed the variables within the function.

model/task/task.go
```diff
@@ -3301,6 +3301,14 @@ func (t *Task) FindAllMarkedUnattainableDependencies() ([]Task, error) {
 	return FindAll(query)
 }
 
+func (t *Task) ToTaskNode() TaskNode {
```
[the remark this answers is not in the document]
ybrill: done (for now)

model/task/dependency_graph.go
ybrill: removed. We'll use it eventually, but I can put it back then.
Though `buildFromTasks` is extra now it's used by a bunch of tests and rewriting them would be a lot.

model/task/dependency_graph.go
ybrill: done

globals.go
ablack12: Git tag versions aren't patches though, so it seems incorrect to exclude them from this case, logically, so we may need to either rename to be really explicit (although I don't see when this case would be relevant globally so possibly a private function in model), use the logic directly in the model function, or use the original logic.

model/project.go
```diff
@@ -1431,38 +1457,66 @@ func (p *Project) FindAllVariants() []string {
...
+			GitTagOnly:       bvTaskGroup.GitTagOnly,
+			Priority:         bvTaskGroup.Priority,
+			DependsOn:        bvTaskGroup.DependsOn,
+			RunOn:            bvTaskGroup.RunOn,
+			ExecTimeoutSecs:  bvTaskGroup.ExecTimeoutSecs,
+			Stepback:         bvTaskGroup.Stepback,
+			Activate:         bvTaskGroup.Activate,
+			CommitQueueMerge: bvTaskGroup.CommitQueueMerge,
+		}
+		// Default to project task settings when unspecified
```
ablack12, line 1513: nit: can we punctuate to match new style guidelines?

ybrill: I had a back and forth with Kim in the comments on the style guidelines where the conclusion was that proper punctuation was encouraged, but not required.

globals.go
ybrill: Got it. This logic is a headache 🤯 .
I went back to the original logic and removed the new stuff.

validator/project_validator.go
```diff
@@ -1829,14 +1829,16 @@ func validateTVDependsOnTV(dependentTask, dependedOnTask model.TVPair, statuses
 			}
 		}
 
-		if fromTaskUnit.RunsOnPatches() && (!toTaskUnit.RunsOnPatches() || edgeInfo.PatchOptional) {
+		// PatchOptional dependencies are skipped when the edge.From task is running on a patch.
```
ablack12, line 1832: preference for using the variable name (fromTaskUnit) rather than the struct name (edge.From) in comments, since that requires another translation step (and it's easy to read the . as the end of a sentence when skimming)

ablack12 (approved):
LGTM, left a comment suggestion

validator/project_validator.go
ybrill: done.
I'd actually meant the variable from lines 1822-1823 but I think that further underscores your point 😃

PR #5630, opened 2022-05-04
EVG-16809 check if user has patch submission permission

ablack12:
[EVG-16809](https://jira.mongodb.org/browse/EVG-16809)

### Description 
We have a middleware that checks that commit queue submissions are being submitted by the author, which isn't true if the `--author` flag is used. However, that flag has restricted access so we should check if the user has that permission.

(PR also looks big bc I changed how we handle "Unauthorized" since it swallows real errors and can be confusing.)

### Testing 
Tested on staging using the same credentials that the code review bot uses, and added a unit test.

### DIFF
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -405,25 +405,7 @@ func (m *MockCommitQueueItemOwnerMiddleware) ServeHTTP(rw http.ResponseWriter, r
 		return
 	}
 
-	// A superuser or project admin is authorized.
-	isAdmin := user.HasPermission(gimlet.PermissionOpts{
-		Resource:      opCtx.ProjectRef.Id,
-		ResourceType:  evergreen.ProjectResourceType,
-		Permission:    evergreen.PermissionProjectSettings,
-		RequiredLevel: evergreen.ProjectSettingsEdit.Value,
-	})
-	if isAdmin {
-		next(rw, r)
-		return
-	}
-	// User may be authorized specifically to patch on behalf of other users.
-	isAuthorizedPatchAdmin := user.HasPermission(gimlet.PermissionOpts{
-		Resource:      opCtx.ProjectRef.Id,
-		ResourceType:  evergreen.ProjectResourceType,
-		Permission:    evergreen.PermissionPatches,
-		RequiredLevel: evergreen.PatchSubmitAdmin.Value,
-	})
-	if isAuthorizedPatchAdmin {
+	if canSubmitAlwaysPatchesForProject() {
 		next(rw, r)
 		return
 	}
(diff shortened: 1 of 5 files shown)


rest/route/middleware.go
```diff
@@ -531,7 +553,7 @@ func (m *CommitQueueItemOwnerMiddleware) ServeHTTP(rw http.ResponseWriter, r *ht
 		if user.Id != *patch.Author {
 			gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
 				StatusCode: http.StatusUnauthorized,
-				Message:    "Not authorized",
+				Message:    "Not authorized  commit queue author",
```
hadjri, line 556: Double space here

rest/route/middleware.go
```diff
@@ -416,6 +416,17 @@ func (m *MockCommitQueueItemOwnerMiddleware) ServeHTTP(rw http.ResponseWriter, r
 		next(rw, r)
 		return
 	}
+	// User may be authorized specifically to patch on behalf of other users.
+	isAuthorizedPatchAdmin := user.HasPermission(gimlet.PermissionOpts{
```
hadjri, line 420: This is duplicated at L522 can we modularize it a bit?

hadjri (changes requested):

rest/route/middleware.go
ablack12: oops i actually made these changes for testing so I'll get rid of it

rest/route/middleware.go
```diff
@@ -531,7 +438,7 @@ func (m *CommitQueueItemOwnerMiddleware) ServeHTTP(rw http.ResponseWriter, r *ht
 		if user.Id != *patch.Author {
 			gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
 				StatusCode: http.StatusUnauthorized,
-				Message:    "Not authorized",
+				Message:    "Not authorized patch author",
```
malikchaya2, line 441: This wording confuses me a bit, because it's not the patch author that's not authorized (whoever is hitting this is specifically not the patch author). So it's technically `not authorized because not patch author or person with the proper permissions`.

rest/route/middleware.go
```diff
@@ -552,7 +459,7 @@ func (m *CommitQueueItemOwnerMiddleware) ServeHTTP(rw http.ResponseWriter, r *ht
 		if githubUID == 0 || user.Settings.GithubUser.UID != githubUID {
 			gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
 				StatusCode: http.StatusUnauthorized,
-				Message:    "Not authorized",
+				Message:    "Not authorized Github user",
```
malikchaya2, line 462: Same comment as above.

malikchaya2 (changes requested):

hadjri (approved):

malikchaya2 (approved):

PR #5631, opened 2022-05-04
EVG-16784: Return partial error from copyProject

sophstad:
[EVG-16784](https://jira.mongodb.org/browse/EVG-16784)

### Description
- When a project has successfully copied but some fields were unable to be copied, return both data _and_ error from the `copyProject` resolver. This enables us to redirect the user to the new project, but still warn them that some errors were encountered.
- Add new `PartialError` GraphQL error type

### Testing 
- Add unit test, tested in GraphQL playground

### DIFF
--- a/model/project_vars.go
+++ b/model/project_vars.go
@@ -157,6 +157,9 @@ func CopyProjectVars(oldProjectId, newProjectId string) error {
 	if err != nil {
 		return errors.Wrapf(err, "finding variables for project '%s'", oldProjectId)
 	}
+	if vars == nil {
+		vars = &ProjectVars{}
+	}
 	vars.Id = newProjectId
 	_, err = vars.Upsert()
 	return errors.Wrapf(err, "inserting variables for project '%s", newProjectId)
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -77,6 +77,7 @@ func CopyProject(ctx context.Context, opts CopyProjectOpts) (*restModel.APIProje
 	if err := model.UpdateAdminRoles(projectToCopy, projectToCopy.Admins, nil); err != nil {
 		catcher.Wrapf(err, "updating admin DB for project '%s'", opts.NewProjectIdentifier)
 	}
+	// Since the errors above are nonfatal and still permit copying the project, return both the new project and any errors that were encountered.
 	return apiProjectRef, catcher.Resolve()
 }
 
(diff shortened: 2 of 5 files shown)


model/project_vars.go
```diff
@@ -157,6 +157,9 @@ func CopyProjectVars(oldProjectId, newProjectId string) error {
 	if err != nil {
 		return errors.Wrapf(err, "finding variables for project '%s'", oldProjectId)
 	}
+	if vars == nil {
+		vars = &ProjectVars{}
```
ablack12, line 161: Did you run into this case? No project should exist without vars so if no vars exists that seems like an error (although I guess we should be returning one here then).

rest/data/project_settings.go
```diff
@@ -63,20 +63,21 @@ func CopyProject(ctx context.Context, opts CopyProjectOpts) (*restModel.APIProje
...
 	if err := event.CopyProjectSubscriptions(oldId, projectToCopy.Id); err != nil {
-		return nil, errors.Wrapf(err, "error copying subscriptions from project '%s'", oldIdentifier)
+		catcher.Wrapf(err, "copying subscriptions from project '%s'", oldIdentifier)
 	}
 	// set the same admin roles from the old project on the newly copied project.
 	if err := model.UpdateAdminRoles(projectToCopy, projectToCopy.Admins, nil); err != nil {
-		return nil, errors.Wrapf(err, "Database error updating admins for project '%s'", opts.NewProjectIdentifier)
+		catcher.Wrapf(err, "updating admin DB for project '%s'", opts.NewProjectIdentifier)
 	}
-	return apiProjectRef, nil
```
ablack12, line 79: Can we add a comment explaining this since it's not the typical pattern?

rest/data/project_settings_test.go
```diff
@@ -553,3 +553,129 @@ func TestSaveProjectSettingsForSection(t *testing.T) {
 		})
 	}
 }
+
+func TestCopyProject(t *testing.T) {
```
ablack12, line 557: 💯

model/project_vars.go
sophstad: I ran into this nil pointer dereference while writing tests so I suppose that's just a matter of improving the test data -- but I noticed we do the same thing [here](https://github.com/evergreen-ci/evergreen/blob/6026a8f1f2aa5a4f24aaf89ff00c7dc124a71575/model/project_vars.go#L170-L172) so perhaps it has been encountered before?

ablack12: Ah interesting. I'm fine with just leaving it, just wanted to make sure it wasn't something you ran into in production.

malikchaya2 (approved):

ablack12 (approved):

PR #5634, opened 2022-05-04
EVG-16752 Remove mention of evergreen faq

bynn:
[EVG-16752](https://jira.mongodb.org/browse/EVG-16752)

### Description 
remove faq and check for warning is fixed

### Testing 
https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu1604_bynntask_patch_b14644c4918e0a8f632efe00e4d4e8ccc344b327_625de82f97b1d30d902fc0e3_22_05_04_20_33_35/logs?execution=2

### DIFF
--- a/agent/command/shell.go
+++ b/agent/command/shell.go
@@ -117,13 +117,13 @@ func (c *shellExec) Execute(ctx context.Context, _ client.Communicator, logger c
 	logger.Execution().Debug("Preparing script...")
 
 	var err error
-	var originalScript string
-	if err = c.doExpansions(conf.Expansions); err != nil {
+	originalScript := c.Script
+	if err := c.doExpansions(conf.Expansions); err != nil {
 		logger.Execution().Warning(err.Error())
 		return errors.WithStack(err)
 	}
 	if originalScript != c.Script {
-		logger.Task().Info("`${SHELL_EXPANSION}` syntax will only work for Evergreen expansions. See Evergreen FAQ for details.")
+		logger.Task().Info("`${SHELL_EXPANSION}` syntax will only work for Evergreen expansions. For bash expansions, try using 'sed' instead.")
 	}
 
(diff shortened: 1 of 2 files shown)


agent/command/shell.go
```diff
@@ -117,13 +117,13 @@ func (c *shellExec) Execute(ctx context.Context, _ client.Communicator, logger c
...
 
 	var err error
-	var originalScript string
-	if err = c.doExpansions(conf.Expansions); err != nil {
+	originalScript := c.Script
+	if err := c.doExpansions(conf.Expansions); err != nil {
 		logger.Execution().Warning(err.Error())
 		return errors.WithStack(err)
 	}
 	if originalScript != c.Script {
```
ablack12, line 125: oh LOL good catch

ablack12 (approved):
Increment agent version

agent/command/shell.go
```diff
@@ -117,13 +117,13 @@ func (c *shellExec) Execute(ctx context.Context, _ client.Communicator, logger c
...
-		logger.Task().Info("`${SHELL_EXPANSION}` syntax will only work for Evergreen expansions. See Evergreen FAQ for details.")
+		logger.Task().Info("`${SHELL_EXPANSION}` syntax will only work for Evergreen expansions. For bash expansions, try using 'sed' instead.")
```
bsamek, line 126: I think the `sed` trick only work if they're using backticks to do string substitution, and is likely to confuse people.

bsamek (changes requested):

agent/command/shell.go
ablack12: Should we just drop the suggestion on how to fix it?

bsamek: I think so. We could say something like

> ${SHELL_EXPANSION} will not get passed to the shell, since Evergreen uses this syntax for its own expansions.

bynn: i think it makes sense to just remove

bsamek: LGTM

bsamek (approved):

PR #5635, opened 2022-05-04
EVG-16371: propagate container options from task to pod

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16371

### Description 
* Resolve the task's container config when creating the DB task. The container config cannot be changed once the task is created.
    * This is necessary to avoid some extra work later on. The container definitions only exist in the `model.Project` (the YAML model) and not in a DB model. Therefore, resolving the container options when creating the DB task avoids the need to re-read the whole YAML into a `model.Project` to get the container definitions in the pod lifecycle.
    * Resolving the container options at task creation time also has an implication for container sizes - since container sizes can be defined in the project ref, they can change dynamically. I decided that it's better to snapshot the container size at task creation time and disallow dynamically changing the task's container size after the task is already created (e.g. when restarting a task), since that would make the container parameters unstable across different executions. For example, if a user defined the task's container size as "small" with memoryMB=128 and cpu=128 in the project settings page, then ran a task with the "small" size, then updated the container size to be memoryMB=256 and cpu=256 on the project settings page, then restarted the task, the newer execution of the "small" task will still use the original "small" parameters (memoryMB=128 and cpu=128).
    * Plumb the project ref down to `createOneTask` to minimize the number of times we fetch the project ref. Doing this also removes some unnecessary re-fetching of the project ref in some of the task creation helper functions.
* Populate the intent pods using the task's container options.
* Add a small validation for the container size name to ensure it's non-empty.

### Testing 
Added unit tests for container task creation, the pod allocator, and project validation.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -1111,6 +1111,10 @@ const (
 	WindowsOS ContainerOS = "windows"
 )
 
+// ValidContainerOperatingSystems contains all recognized container operating
+// systems.
+var ValidContainerOperatingSystems = []ContainerOS{LinuxOS, WindowsOS}
+
 // Validate checks that the container OS is recognized.
 func (c ContainerOS) Validate() error {
 	switch c {
@@ -1121,16 +1125,20 @@ func (c ContainerOS) Validate() error {
 	}
 }
 
-// CPUArchitecture represents the architecture necessary to run a container.
-type CPUArchitecture string
+// ContainerArch represents the CPU architecture necessary to run a container.
+type ContainerArch string
 
 const (
-	ArchARM64 CPUArchitecture = "arm64"
-	ArchAMD64 CPUArchitecture = "x86_64"
+	ArchARM64 ContainerArch = "arm64"
+	ArchAMD64 ContainerArch = "x86_64"
 )
 
+// ValidContainerArchitectures contains all recognized container CPU
+// architectures.
+var ValidContainerArchitectures = []ContainerArch{ArchARM64, ArchAMD64}
+
 // Validate checks that the container CPU architecture is recognized.
-func (c CPUArchitecture) Validate() error {
+func (c ContainerArch) Validate() error {
 	switch c {
 	case ArchARM64, ArchAMD64:
 		return nil
@@ -1147,6 +1155,9 @@ const (
 	Windows2019 WindowsVersion = "2019"
 	Windows2016 WindowsVersion = "2016"
 )
+
+// ValidWindowsVersions contains all recognized container Windows versions.
+var ValidWindowsVersions = []WindowsVersion{Windows2016, Windows2019, Windows2022}
 
 // Validate checks that the container Windows version is recognized.
 func (w WindowsVersion) Validate() error {
--- a/model/generate.go
+++ b/model/generate.go
@@ -274,8 +274,8 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		return errors.Errorf("project '%s' not found", p.Identifier)
 	}
 
-	activatedTasksInExistingBuilds, err := addNewTasks(ctx, activationInfo, v, p, newTVPairsForExistingVariants,
-		existingBuilds, syncAtEndOpts, projectRef.Identifier, g.TaskID)
+	activatedTasksInExistingBuilds, err := addNewTasks(ctx, activationInfo, v, p, projectRef, newTVPairsForExistingVariants,
+		existingBuilds, syncAtEndOpts, g.TaskID)
 	if err != nil {
 		return errors.Wrap(err, "adding new tasks")
 	}
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -525,7 +525,7 @@ func RefreshTasksCache(buildId string) error {
 }
 
 // addTasksToBuild creates/activates the tasks for the given build of a project
-func addTasksToBuild(ctx context.Context, b *build.Build, project *Project, v *Version, taskNames []string,
+func addTasksToBuild(ctx context.Context, b *build.Build, project *Project, pRef *ProjectRef, v *Version, taskNames []string,
 	displayNames []string, activationInfo specificActivationInfo, generatedBy string, tasksInBuild []task.Task,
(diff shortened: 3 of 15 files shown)


model/generate.go
```diff
@@ -274,8 +274,8 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		return errors.Errorf("project '%s' not found", p.Identifier)
 	}
 
-	activatedTasksInExistingBuilds, err := addNewTasks(ctx, activationInfo, v, p, newTVPairsForExistingVariants,
-		existingBuilds, syncAtEndOpts, projectRef.Identifier, g.TaskID)
+	activatedTasksInExistingBuilds, err := addNewTasks(ctx, activationInfo, v, p, projectRef, newTVPairsForExistingVariants,
```
hadjri, line 277: We can move the project ref lookup at L269 here into addNewTasks

model/lifecycle.go
```diff
@@ -1355,6 +1345,42 @@ func getContainerFromRunOn(id string, buildVarTask BuildVariantTaskUnit, buildVa
...
 
+// getContainerOptions resolves the task's container configuration based on the
+// task's container name and the container definitions available to the project.
+func getContainerOptions(project *Project, pRef *ProjectRef, container string) (*task.ContainerOptions, error) {
+	for _, c := range project.Containers {
+		if c.Name != container {
+			continue
+		}
+
+		opts := task.ContainerOptions{
```
hadjri, line 1356: How about instead of doing this we change the model.Container struct to contain just Name and ContainerOpts as its two properties. Then we can just do `opts := c.ContainerOpts` and not need to duplicate similar looking structs. 

We'd still loop through the container sizes and overwrite the resources if we find a matching size.

hadjri (changes requested):

globals.go
```diff
@@ -1148,6 +1156,9 @@ const (
 	Windows2016 WindowsVersion = "2016"
 )
 
+// ValidWindowsVersions contains all recognized container Windows versions.
+var ValidWindowsVersions = []WindowsVersion{Windows2016, Windows2019, Windows2022}
```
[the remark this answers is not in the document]
Kimchelly: This is only used in the unit tests for now to test all valid enum values, but they're useful sometimes to conveniently get a list of all valid enum values since Go doesn't have a built-in way to do this.

I didn't change the validation to a slice containment check because `utility.StringSliceContains` doesn't allow passing a `[]ContainerOS` (even though `ContainerOS` is a `string` enum) due to Go typing rules. If you still prefer the slice way though, I can modify the validation to iterate through the slice. I think it's not too important though, because it would be pretty uncommon for us to add/remove enum values OS/arch/Windows version.

model/lifecycle.go
Kimchelly: I prefer to avoid doing this for future-proofing reasons - using the same model for both would mixes YAML models and DB models together, so modifying the YAML structure will also affect the DB model (and vice-versa). Even if they're the same now, I usually write things to keep data models separate to avoid tight coupling.

I'm also assuming that that the container options for the DB model will become more complicated over time in ways that aren't related to YAML configuration.

model/task/task.go
```diff
@@ -107,14 +107,19 @@ type Task struct {
...
-	BuildId                 string       `bson:"build_id" json:"build_id"`
-	DistroId                string       `bson:"distro" json:"distro"`
-	Container               string       `bson:"container,omitempty" json:"container,omitempty"`
-	BuildVariant            string       `bson:"build_variant" json:"build_variant"`
-	BuildVariantDisplayName string       `bson:"build_variant_display_name" json:"-"`
-	DependsOn               []Dependency `bson:"depends_on" json:"depends_on"`
-	NumDependents           int          `bson:"num_dependents,omitempty" json:"num_dependents,omitempty"`
-	OverrideDependencies    bool         `bson:"override_dependencies,omitempty" json:"override_dependencies,omitempty"`
+	BuildId  string `bson:"build_id" json:"build_id"`
+	DistroId string `bson:"distro" json:"distro"`
```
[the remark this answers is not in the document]
Kimchelly: Done. I didn't do this earlier because I was concerned that some existing queries would break if this wasn't always populated. I _think_ I've modified all the queries that rely on distro-specific logic to check specifically host tasks at this point, so it should be safe to make this omitempty.

model/generate.go
Kimchelly: I think the project ref lookup is still necessary for [`addNewBuilds`](https://github.com/evergreen-ci/evergreen/pull/5635/files#diff-96655072263d498b6dc1571c45fc69e886abcc4bc9994343a80af2f54375c7bcR283).

hadjri: Whoops. Missed that

globals.go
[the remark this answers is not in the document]
hadjri: Ok SGTM, it's fine we can keep the validation switch statement. It's probably easier to read anyway.

model/lifecycle.go
```diff
@@ -1355,6 +1345,42 @@ func getContainerFromRunOn(id string, buildVarTask BuildVariantTaskUnit, buildVa
...
+			WindowsVersion: c.System.WindowsVersion,
+		}
+
+		if c.Resources != nil {
+			opts.CPU = c.Resources.CPU
+			opts.MemoryMB = c.Resources.MemoryMB
+			return &opts, nil
+		}
+
+		for name, size := range pRef.ContainerSizes {
```
hadjri, line 1370: ContainerSizes is a map so I think we can just do a typical `containerSize, ok := pRef.ContainerSizes[c.Size]` instead

model/lifecycle.go
hadjri: Gotcha. Sure if it'll give future flexibility SGTM

hadjri (changes requested):

model/lifecycle.go
Kimchelly: Ah yeah, good catch. Fixed.

hadjri (approved):

PR #5636, opened 2022-05-04
EVG-16713 refactor handling of nonexistent containers

ablack12:
[EVG-16713](https://jira.mongodb.org/browse/EVG-16713)

### Description 
Discovered two things in this ticket:
1) We double log everything that we both grip and add to the job's errors, so I reduced this so we only log once.
2) We have special handling for EC2 instances not being found, but not for docker. It's weird to do this for an interface also, since docker has errors formatted differently, so I thought a new status could make sense, and I looked at usages of GetInstanceStatus to verify terminations would still happen where needed. 

If this seems like an issue, I could change it to case on EC2 errors _or_ docker errors but I think it makes sense to move that logic out.

### Testing 
Not sure how to test besides putting in staging and doing some things and verifying no breakage.

### DIFF
--- a/cloud/docker_client.go
+++ b/cloud/docker_client.go
@@ -427,8 +427,7 @@ func (c *dockerClientImpl) GetDockerStatus(ctx context.Context, containerID stri
 	}
 	container, err := c.GetContainer(ctx, parent, containerID)
 	if err != nil {
-		if strings.Contains(err.Error(), "No such container") ||
-			strings.Contains(err.Error(), "Is the docker daemon running?") {
+		if docker.IsErrNotFound(err) || docker.IsErrConnectionFailed(err) {
 			return &ContainerStatus{HasStarted: false}, nil
 		}
 		return nil, errors.Wrapf(err, "Error getting container %s", containerID)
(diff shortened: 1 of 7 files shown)


cloud/docker_client.go
```diff
@@ -427,8 +427,7 @@ func (c *dockerClientImpl) GetDockerStatus(ctx context.Context, containerID stri
 	}
 	container, err := c.GetContainer(ctx, parent, containerID)
 	if err != nil {
-		if strings.Contains(err.Error(), "No such container") ||
-			strings.Contains(err.Error(), "Is the docker daemon running?") {
+		if docker.IsErrNotFound(err) || docker.IsErrConnectionFailed(err) {
```
malikchaya2, line 430: In 2 functions in docker.go we return StatusTerminated and "error making connection" for client.IsErrConnectionFailed. I don't think `HasStarted: false` would be a certainty is the container is terminated or if we simply can't connect to it.

malikchaya2 (changes requested):
Good catch!

malikchaya2 (approved):

PR #5638, opened 2022-05-05
EVG-16040 Update test endpoint to return latest execution

bynn:
[EVG-16040](https://jira.mongodb.org/browse/EVG-16040)

### Description 
adding latest=true will give the most recent execution and default behavior is not modified
only works if execution is not specified 

### Testing 
staging
image

https://evergreen-staging.corp.mongodb.com/rest/v2/tasks/evg_race_detector_test_model_pod_4a1ed094bee959b235b34f3a2cffd71a43586e02_22_05_04_14_10_29/tests?latest=true

### DIFF
--- a/rest/route/test.go
+++ b/rest/route/test.go
@@ -69,6 +69,8 @@ func (tgh *testGetHandler) Parse(ctx context.Context, r *http.Request) error {
 				StatusCode: http.StatusBadRequest,
 			}
 		}
+	} else {
+		tgh.testExecution = projCtx.Task.Execution
 	}
 
 	if status := vals.Get("status"); status != "" {
--- a/rest/route/test.go
+++ b/rest/route/test.go
@@ -31,6 +31,7 @@ type testGetHandler struct {
 	key           string
 	limit         int
 	sc            data.Connector
+	latest        bool
 }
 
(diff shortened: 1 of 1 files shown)


rest/route/test.go
```diff
@@ -69,6 +69,8 @@ func (tgh *testGetHandler) Parse(ctx context.Context, r *http.Request) error {
 				StatusCode: http.StatusBadRequest,
 			}
 		}
+	} else {
+		tgh.testExecution = projCtx.Task.Execution
```
malikchaya2, line 73: is projCtx.Task always the latest task?

bynn: yeah i think all the older ones get archived so the one we do get should always be the latest execution

malikchaya2: true

malikchaya2 (approved):

bynn:
default option is now the same but you can specify latest 
cannot specify latest and also an execution

ablack12 (approved):
Remember to add to documentation and the ticket

bynn:
https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-tests-from-a-task

PR #5639, opened 2022-05-06
EVG-16749: Fix incorrect logic for fetching old execution tasks

minnakt:
[EVG-16749](https://jira.mongodb.org/browse/EVG-16749)

### Description 
This PR fixes an issue where the execution tasks were being returned incorrectly. You can see in the video below that when I go through different executions in this [version](https://spruce.mongodb.com/task/mongodb_mongo_v4.4_windows_display_multiversion_319d8d8145fb5893ab3bf5d077730226f3f2af7c_22_04_06_18_31_56/execution-tasks?execution=0&sortBy=STATUS&sortDir=ASC&sorts=STATUS%3AASC), I start getting multiple copies of the execution tasks (there should only ever be four). 

https://user-images.githubusercontent.com/47064971/167049541-b0c120ed-b52e-442d-b8c5-dd1cd2fcd966.mov

The reason this happens is that there is an error in the query for fetching execution tasks. Specifically, the problem comes from the fact that we find old tasks by matching on `{ execution: { $lte: specified_execution } }`.

In the video above there are three executions of the task, which means there is one task in the `Tasks` collection, and two older tasks in the `OldTasks` collection. If we want execution 2, then we'll need to query the `OldTasks` collection, but since we get all executions _less than or equal to_ execution 2, we'll end up also getting the tasks from execution 1. That's why the duplicates appear.

Since we only want one copy of each execution task, I modified the pipeline so that we return the task with the most recent execution.

### Testing 
- Added test for untested case in `TestByExecutionTasksAndMaxExecution`

### DIFF
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -3858,8 +3858,10 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
 		assertTasksAreEqual(t, t1, tasks[0], 2)
 		assertTasksAreEqual(t, t2, tasks[1], 1)
 	})
-	t.Run("Fetching older executions with same execution", func(t *testing.T) {
+	t.Run("Fetching old executions when there are even older executions", func(t *testing.T) {
 		require.NoError(t, db.ClearCollections(Collection, OldCollection))
+
+		// Both tasks have 2 previous executions.
 		t1 := Task{
 			Id:        "t1",
 			Version:   "v1",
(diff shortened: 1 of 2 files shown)


khelif96 (approved):
LGTM pending app approval

model/task/task_test.go
```diff
@@ -3901,6 +3901,53 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
...
+		// Both tasks have two previous executions.
+		t1 := Task{
+			Id:        "t1",
+			Version:   "v1",
+			Execution: 2,
+			Status:    evergreen.TaskSucceeded,
+		}
+		assert.NoError(t, db.Insert(Collection, t1))
+
+		ot1_execution1 := t1
```
ablack12, line 3916: We don't use snake case. Can all of these variables be modified to use camelcase?

ablack12 (commented):
LGTM besides test style issue

PR #5640, opened 2022-05-06
EVG-12923 block task group tasks before clearing host's running task

ablack12:

### Description 
Found this bug through investigating [these logs](https://mongodb.splunkcloud.com/en-US/app/search/search?s=%2FservicesNS%2Fannie.black%2Fsearch%2Fsaved%2Fsearches%2Funblocked%2520dependencies%2520that%2520aren%2527t%2520re-blocked&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&q=search%20index%3Devergreen%20%22message%22%3D%22unblocked%20dependent%20tasks%22%20%22ticket%22%3D%20%22EVG-12923%22&earliest=1651723200&latest=1651896000&sid=1651846628.1336415). If a single host task group task fails, we block all subsequent tasks in the group. However, we clear the host's running task first, so it sometimes picks up the later task group task before it's actually marked as blocked. We should add these dependencies before allowing the host to pick up a new task.

### DIFF
--- a/service/api_task.go
+++ b/service/api_task.go
@@ -180,8 +180,8 @@ func (as *APIServer) EndTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// For a single-host task group, if a task fails, block and dequeue later tasks in that group.
-	// Call before MarkEnd so the version is marked finished when this is the last task in the version
-	// to finish, and before clearing the running host so later tasks in the group aren't picked up by the host.
+	// Call before MarkEnd so the version is marked finished when this is the last task in the version to finish,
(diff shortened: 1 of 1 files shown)


ablack12:
We actually only add these dependencies in the case that one of them is blocked (code: https://github.com/evergreen-ci/evergreen/blob/5e8f49dfb8a290050cdc226571c2a02922c7d367/model/task_queue.go#L281-L280). I'm not sure if there's a reason why we don't just add these dependencies to begin with (I think we used to not require earlier tasks to run if they existed but now we do so it's effectively the same). I've created https://jira.mongodb.org/browse/EVG-16826 to investigate it as part of the improvements epic.

PR #5641, opened 2022-05-06
EVG-16803: log statistics for container task pipeline performance

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16803

### Description 
I had to plug in the dispatcher to the pod next task route since it was necessary for logging. It doesn't actually make the agent run the task for now, but it does log that the necessary DB changes were made to indicate the task is considered dispatched.

We're most interested in timing how long each step takes from the moment the task starts waiting to run (when it's activated) to when it actually run. I changed it to log timing statistics when:
* Container task queue is generated.
* Pod allocation jobs are being enqueued (so we know how many parallel pod requests there are currently).
* Task is allocated a pod.
* Task's pod is actually requested from the cloud provider.
* Agent makes first contact with the app server.
* Task is dispatched to the agent.

### Testing 
Tested the new DB functions and updated pod next task tests.

### DIFF
--- a/rest/route/pod_agent.go
+++ b/rest/route/pod_agent.go
@@ -2,10 +2,13 @@ package route
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/apimodels"
+	"github.com/evergreen-ci/evergreen/model/pod"
+	"github.com/evergreen-ci/evergreen/model/pod/dispatcher"
 	"github.com/evergreen-ci/evergreen/units"
 	"github.com/evergreen-ci/gimlet"
 	"github.com/evergreen-ci/utility"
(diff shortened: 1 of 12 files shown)


rest/route/pod_agent.go
```diff
@@ -115,13 +121,91 @@ func (h *podAgentNextTask) Parse(ctx context.Context, r *http.Request) error {
...
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("pod '%s' not found", h.podID),
+		}
+	}
+
+	return p, nil
+}
+
+func (h *podAgentNextTask) setAgentFirstContactTime(p *pod.Pod) {
+	if !p.TimeInfo.Initializing.IsZero() {
```
hadjri, line 175: .Initializing was a slightly confusing name to me at first. Can we renamed to something like InitializedTime (to keep consistent with past tense timestamp names like LastCommunicated).

hadjri (changes requested):

rest/route/pod_agent.go
```diff
@@ -115,13 +121,91 @@ func (h *podAgentNextTask) Parse(ctx context.Context, r *http.Request) error {
...
 	}
+	if nextTask == nil {
+		j := units.NewPodTerminationJob(h.podID, "reached end of pod lifecycle", utility.RoundPartOfMinute(0))
+		if err := amboy.EnqueueUniqueJob(ctx, h.env.RemoteQueue(), j); err != nil {
+			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    errors.Wrap(err, "enqueueing job to terminate pod").Error(),
+			})
+		}
+	}
```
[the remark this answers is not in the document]
Kimchelly: The empty struct is equivalent to returning 200 OK and telling the agent that there's no next task to run. Since the agent doesn't actually run tasks, this is supposedly going to be filled in with real logic in PM-2616.

rest/route/pod_agent.go
Kimchelly: It's already in a `TimeInfo` struct whose sole purpose is to hold timing information about when certain things happened to the pod, so it seems pretty clear. I don't generally use [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) since it's redundant to the `TimeInfo` struct and modern IDEs include type information that it's a `time.Time` anyways.

units/pod_creation.go
```diff
@@ -178,3 +185,43 @@ func (j *podCreationJob) populateIfUnset(ctx context.Context) error {
...
+		"dispatcher_group": disp.GroupID,
+	}
+
+	tsk, err := task.FindOneId(disp.GroupID)
+	if err != nil {
+		return errors.Wrapf(err, "finding tasks associated with dispatcher group '%s'", disp.GroupID)
+	}
+
+	if tsk != nil {
+		msg["is_for_task_group"] = false
```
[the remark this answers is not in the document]
Kimchelly: I want to keep the if-else because of the comment on the else case, but it would look more awkward to add an if-else block whose sole purpose is to no-op and make a comment.

rest/route/pod_agent.go
hadjri: Ok, I'm fine with keeping it as is.

hadjri (approved):

PR #5646, opened 2022-05-09
EVG-12923 consider in nextTask if a task is currently blocked

ablack12:

### Description 
Gave myself a headache reading through scheduler code to think through this:
I think blocking task group tasks isn't quite enough because we depend on the task queue to handle removing unattainable dependencies. However, we only refresh the task queue on hosts [every one minute](https://github.com/evergreen-ci/evergreen/blob/bbef382d237202a8074f0113bdf5214abb21c386/service/api.go#L60). So even later task group tasks are blocked and dequeued, this will only take affect for hosts [once they refresh their task queues](https://github.com/evergreen-ci/evergreen/blob/5e8f49dfb8a290050cdc226571c2a02922c7d367/model/task_queue.go#L753-L756). 

PS: I believe the reason regular tasks don't hit this is because they already _have_ the dependencies and so the dependencies not being finished is considered by the scheduler already, so this might again be fixed by https://jira.mongodb.org/browse/EVG-16826, although it isn't certain still if this can be done.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1418,20 +1418,8 @@ func ClearAndResetStrandedTask(h *host.Host) error {
 		return nil
 	}
 
-	if err = h.ClearRunningTask(); err != nil {
-		return errors.Wrapf(err, "clearing running task from host '%s'", h.Id)
-	}
-
-	if t.IsFinished() {
-		return nil
-	}
-
-	if err = t.MarkSystemFailed(evergreen.TaskDescriptionStranded); err != nil {
-		return errors.Wrap(err, "marking task failed")
-	}
-
 	// For a single-host task group, block and dequeue later tasks in that group.
-	if t.IsPartOfSingleHostTaskGroup() {
+	if t.IsPartOfSingleHostTaskGroup() && t.Status != evergreen.TaskSucceeded {
 		if err = BlockTaskGroupTasks(t.Id); err != nil {
 			grip.Error(message.WrapError(err, message.Fields{
 				"message": "problem blocking task group tasks",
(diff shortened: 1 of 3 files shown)


service/api_task.go
```diff
@@ -749,6 +749,29 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
...
+			grip.Debug(message.Fields{
+				"message":            "top task queue task is blocked, dequeuing task",
+				"host_id":            currentHost.Id,
+				"distro_id":          nextTask.DistroId,
+				"task_id":            nextTask.Id,
+				"task_group":         nextTask.TaskGroup,
+				"project":            projectRef.Id,
+				"project_identifier": projectRef.Enabled,
+			})
+			grip.Warning(message.WrapError(taskQueue.DequeueTask(nextTask.Id), message.Fields{
```
ablack12, line 763: This dequeue needs to happen per this comment:
https://github.com/evergreen-ci/evergreen/blob/b850ce711b1eacc39cad7ec329f7969ea9c860b1/service/api_task.go#L624-L627
However, it seems like this is only relevant to the legacy dispatcher which isn't in use. Created [EVG-16841](https://jira.mongodb.org/browse/EVG-16841) to address.

model/task_lifecycle.go
```diff
@@ -1418,20 +1418,8 @@ func ClearAndResetStrandedTask(h *host.Host) error {
 		return nil
 	}
 
-	if err = h.ClearRunningTask(); err != nil {
```
ablack12, line 1421: This being out of order isn't relevant to the bug I found but does still appear to be a bug.

Kimchelly: Just to check - the ordering is for the same reasoning about blocking single-host task groups as in #5640, right?

Kimchelly (approved):
LGTM mod clarification question.

model/task_lifecycle.go
ablack12: Yes exactly 👍

PR #5647, opened 2022-05-09
EVG-16304 authorize dependabot patches

ablack12:
[EVG-16304 ](https://jira.mongodb.org/browse/EVG-16304 )

### Description 
Dependabot patches should be automatically authorized. I was torn about whether to make dependabot a constant, or have this be definable via admin settings. However, dependabot is something of a special case, and we aren't intending to allow any other non-members to be authorized, so I think making this an editable admin setting field would be misleading.

### Testing 
Don't have a good way to test this; may make an alert for myself on the new log message and will close when it triggers (i.e. the code is validated).

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	patchIntentJobName = "patch-intent-processor"
+	patchIntentJobName   = "patch-intent-processor"
+	githubDependabotUser = "dependabot[bot]"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):

units/patch_intent.go
```diff
@@ -947,20 +948,34 @@ func (j *patchIntentProcessor) authAndFetchPRMergeBase(ctx context.Context, patc
...
+			"head_repo": fmt.Sprintf("%s/%s", patchDoc.GithubPatchData.HeadOwner, patchDoc.GithubPatchData.HeadRepo),
+			"pr_number": patchDoc.GithubPatchData.PRNumber,
+		})
+		isMember = true
+	} else {
+		isMember, err = thirdparty.GithubUserInOrganization(ctx, githubOauthToken, requiredOrganization, githubUser)
+		if err != nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"job":          j.ID(),
+				"message":      "Failed to authenticate github PR",
```
bynn, line 969: should we update the error message here too?

bynn (approved):

units/patch_intent.go
ablack12: Update how?

bynn: Ah i wasn't sure if we should follow kim's error style here saying something like "authenticating github pr"

ablack12: Oh gotcha -- this is for splunk so it's different; it's okay for messages to be full sentences.

PR #5648, opened 2022-05-10
EVG-16103 Treat heartbeat abort signal and task conflict error equivalently

hadjri:

### Description 
Tasks that timeout without providing a heartbeat response for a configured time limit get picked up by a "task-execution-timeout" job which will (amongst other things) wipe it's host of the running task, and reset the job including its secret.  There is another job "agent-monitor-deploy" job which attempts to redeploy the agent monitor to hosts it hasn't heard from for 15+ minutes, which happens here.  

What seems to be happening is the agent monitor redeploys to these hosts in these cases and the old agent resumes trying to ping heartbeat for the same task although it has already timed out / been cleaned up, and now since the secret has since been updated each of these requests gets a 409 Conflict error over and over.  We had code to abort a task once it gets this error - this PR fixes a bug that was preventing this 409 response from being properly recognized in the agent's heartbeat loop.

### Testing 
Agent heartbeat unit tests still pass, this requires simulating a non-responsive task in staging / local which doesn't occur naturally (though it happens often in prod).

### DIFF
--- a/agent/background.go
+++ b/agent/background.go
@@ -27,7 +27,7 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 		case <-ticker.C:
 			signalBeat, err = a.doHeartbeat(ctx, tc)
 			if signalBeat == evergreen.TaskFailed {
-				tc.logger.Task().Error("Received termination signal from app server, aborting task")
+				tc.logger.Task().Error("Heartbeat received signal to abort task")
 				if err != nil {
 					tc.logger.Task().Error(err.Error())
 				}
@@ -54,10 +54,10 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 }
 
 func (a *Agent) doHeartbeat(ctx context.Context, tc *taskContext) (string, error) {
-	taskSignal, err := a.comm.Heartbeat(ctx, tc.task)
-	if taskSignal == evergreen.TaskFailed {
+	abort, err := a.comm.Heartbeat(ctx, tc.task)
+	if abort {
 		grip.Info("Task aborted")
-		return taskSignal, nil
+		return evergreen.TaskFailed, err
 	}
 	if err != nil {
 		return "", err
--- a/agent/background.go
+++ b/agent/background.go
@@ -5,10 +5,8 @@ import (
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
-	"github.com/evergreen-ci/evergreen/agent/internal/client"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/recovery"
-	"github.com/pkg/errors"
 )
 
 func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, tc *taskContext, heartbeat chan<- string) {
@@ -28,14 +26,13 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 		select {
 		case <-ticker.C:
 			signalBeat, err = a.doHeartbeat(ctx, tc)
-			if signalBeat == evergreen.TaskConflict {
-				tc.logger.Task().Error("Encountered task conflict while checking heartbeat, aborting task")
+			if signalBeat == evergreen.TaskFailed {
+				tc.logger.Task().Error("Received termination signal from app server, aborting task")
+				if err != nil {
+					tc.logger.Task().Error(err.Error())
+				}
 				heartbeat <- evergreen.TaskFailed
 				cancel()
-				return
-			}
-			if signalBeat != "" {
-				heartbeat <- signalBeat
 				return
 			}
 			if err != nil {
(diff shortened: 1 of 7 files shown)


agent/internal/client/mock.go
```diff
@@ -200,22 +200,22 @@ func (c *Mock) GetExpansions(ctx context.Context, taskData TaskData) (util.Expan
 }
 
 // Heartbeat returns false, which indicates the heartbeat has succeeded.
```
Kimchelly, line 202: Update this comment.

agent/internal/client/interface.go
```diff
@@ -49,7 +49,7 @@ type Communicator interface {
 	GetExpansions(context.Context, TaskData) (util.Expansions, error)
 	// Heartbeat sends a heartbeat to the API server. The server can respond with
 	// an "abort" response. This function returns true if the agent should abort.
```
Kimchelly, line 51: Update this comment.

agent/internal/client/host_methods.go
```diff
@@ -279,7 +279,7 @@ func (c *hostCommunicator) GetExpansions(ctx context.Context, taskData TaskData)
 
 // Heartbeat sends a heartbeat to the API server. The server can respond with
 // an "abort" response. This function returns true if the agent should abort.
```
Kimchelly, line 281: Update this comment.

agent/background.go
```diff
@@ -28,16 +26,15 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 		select {
 		case <-ticker.C:
 			signalBeat, err = a.doHeartbeat(ctx, tc)
-			if signalBeat == evergreen.TaskConflict {
-				tc.logger.Task().Error("Encountered task conflict while checking heartbeat, aborting task")
+			if signalBeat == evergreen.TaskFailed {
+				tc.logger.Task().Error("Received termination signal from app server, aborting task")
```
Kimchelly, line 30: This might be confusing since it doesn't seem like the agent is actually terminating, it's just being told to abort the task. Can this be rephrased as `"Heartbeat received signal to abort task"`?

agent/internal/client/host_methods.go
```diff
@@ -279,7 +279,7 @@ func (c *hostCommunicator) GetExpansions(ctx context.Context, taskData TaskData)
...
-func (c *hostCommunicator) Heartbeat(ctx context.Context, taskData TaskData) (bool, error) {
+func (c *hostCommunicator) Heartbeat(ctx context.Context, taskData TaskData) (string, error) {
```
Kimchelly, line 282: Does this have to return a string instead of a boolean? It seems like `true` is equivalent to `TaskFailed`, so returning an untyped string here for something that still only returns two possible values is more confusing.

agent/internal/client/host_methods.go
```diff
@@ -292,23 +292,26 @@ func (c *hostCommunicator) Heartbeat(ctx context.Context, taskData TaskData) (bo
...
 
 	heartbeatResponse := &apimodels.HeartbeatResponse{}
 	if err = utility.ReadJSON(resp.Body, heartbeatResponse); err != nil {
 		err = errors.Wrapf(err, "Error unmarshaling heartbeat response for task %s", taskData.ID)
-		return false, err
+		return "", err
 	}
-	return heartbeatResponse.Abort, nil
+	if heartbeatResponse.Abort {
+		return evergreen.TaskFailed, nil
```
Kimchelly, line 312: Is there a unit test for abort/conflict?

Kimchelly (changes requested):
If the task was just aborted, why would the agent monitor attempt to deploy? It seems like that should happen since resetting the task shouldn't reset the host's last communicated time.

agent/internal/client/host_methods.go
hadjri: I did this since doHeartbeat returns a string, so we can directly return the result from a.comm.Heartbeat. But if the only two values are "" and evergreen.TaskFailed I'm fine with changing it back to bool

agent/internal/client/host_methods.go
hadjri: Created

agent/background_test.go
```diff
@@ -118,6 +120,24 @@ func (s *BackgroundSuite) TestHeartbeatSometimesFailsDoesNotFailTask() {
...
+	s.a.opts.HeartbeatInterval = time.Millisecond
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	heartbeat := make(chan string)
+	start := time.Now()
+	go s.a.startHeartbeat(ctx, cancel, s.tc, heartbeat)
+	beat := <-heartbeat
+	end := time.Now()
+	sender := s.sender.GetMessage()
+	s.Equal("Heartbeat received signal to abort task", sender.Message.String())
```
Kimchelly, line 134: I don't feel that strongly about this, but still worth saying - I don't think we need to check for exact log message strings. Checking for exact error message strings and messages isn't wrong, but I avoid doing it unless the message has some really important significance. The downside of unit testing exact log message strings is that it becomes more difficult to change source code log strings without a string matching test failing.

agent/background_test.go
```diff
@@ -118,6 +120,24 @@ func (s *BackgroundSuite) TestHeartbeatSometimesFailsDoesNotFailTask() {
...
 
+func (s *BackgroundSuite) TestHeartbeatFailsOnTaskConflict() {
+	s.mockCommunicator.HeartbeatShouldConflict = true
+	s.a.opts.HeartbeatInterval = time.Millisecond
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	heartbeat := make(chan string)
+	start := time.Now()
+	go s.a.startHeartbeat(ctx, cancel, s.tc, heartbeat)
+	beat := <-heartbeat
```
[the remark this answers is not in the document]
hadjri: Thanks!

Kimchelly (approved):

PR #5649, opened 2022-05-10
EVG-16589 reenable code signing

ybrill:
[EVG-16589](https://jira.mongodb.org/browse/EVG-16589)

### Description 
This reverts the changes from https://github.com/evergreen-ci/evergreen/pull/5502

I think we want to not fail the build if signing fails. We should prioritize being able to deploy a build over a guarantee to users that the binary will always be signed. I can add a section to the FAQ explaining how to get it to run when it's not signed.

If we aren't going to fail the build we should at least log that it happened for traceability. 

### Testing 
None


ablack12:
Where would this get logged? From the original cloud request it seems like the workarounds for it not being signed aren't ideal; does the flakiness seem out of our control?

ybrill:
The log would end up in dist's task log.
I'm with you the best case-scenario would be for signing to always work. The outages we've seen have been Apple's fault.

cmd/sign-executable/sign-executable.go
```diff
@@ -176,7 +177,8 @@ func signExecutable(ctx context.Context, opts signOpts) error {
 
 	signedZipPath := path.Join(tempDir, "evergreen_signed.zip")
 	if err = signWithNotaryClient(ctx, toSignPath, signedZipPath, opts); err != nil {
-		return errors.Wrap(err, "signing with client")
+		fmt.Printf("code signing failed: %s", err.Error())
```
Kimchelly, line 180: Should this print to stderr instead of stdout so it shows up in the logs more obviously when it fails?

Kimchelly (approved):
LGTM mod suggestion and question.

Do other teams that rely on signing for onboarding have to be notified of the need to fallback handle unsigned Evergreen CLIs?

ablack12:
Does signing take a lot of time / would there be some benefit in adding retry logic to attempt to get a signed one?

cmd/sign-executable/sign-executable.go
ybrill: done

ybrill:
@Kimchelly 
> Do other teams that rely on signing for onboarding have to be notified of the need to fallback handle unsigned Evergreen CLIs?

That's a good point. No one relies on it yet, but I imagine if we have a good run of signed executables the onboarding docs will probably start to assume the executable is signed. I can mention it on the original ticket and let someone from cloud-dev-prod know. 
For everyone else I'm not sure who to talk to, though hopefully if I include a good link in the FAQ (and people actually look at the FAQ 😅) users can figure it out.

ybrill:
@ablack12 I looked back and failed attempts were on the order of seconds so I added retries.

I considered adding a timeout on the context, but I figured the overall amount of time we spend on trying to sign should be determined by the task calling it, so if we _really_ want to add a timeout we could add an execution timeout to the task, but 🤷‍♂️

ablack12 (approved):

PR #5653, opened 2022-05-11
EVG-16449 Return project info for versions. 

khelif96:
[EVG- 16649](https://jira.mongodb.org/browse/EVG-16649)

### Description 
Supports surfacing Github commit link for versions

### DIFF
--- a/graphql/resolvers.go
+++ b/graphql/resolvers.go
@@ -3976,6 +3976,20 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
 	}
 	return status, nil
 }
+func (*versionResolver) ProjectMetadata(ctx context.Context, obj *restModel.APIVersion) (*restModel.APIProjectRef, error) {
+	projectRef, err := model.FindBranchProjectRef(*obj.Project)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding project ref for project `%s`: %s", *obj.Project, err.Error()))
+	}
+	if projectRef == nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding project ref for project `%s`: %s", *obj.Project, "Project not found"))
+	}
+	apiProjectRef := restModel.APIProjectRef{}
+	if err = apiProjectRef.BuildFromService(projectRef); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error building APIProjectRef from service for `%s`: %s", projectRef.Id, err.Error()))
+	}
+	return &apiProjectRef, nil
+}
 
 func (*versionResolver) UpstreamProject(ctx context.Context, obj *restModel.APIVersion) (*UpstreamProject, error) {
 	v, err := model.VersionFindOneId(utility.FromStringPtr(obj.Id))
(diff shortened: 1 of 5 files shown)


graphql/resolvers.go
```diff
@@ -3976,6 +3976,20 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
 	}
 	return status, nil
 }
+func (*versionResolver) ProjectMetadata(ctx context.Context, obj *restModel.APIVersion) (*restModel.APIProjectRef, error) {
+	projectRef, err := model.FindBranchProjectRef(*obj.Project)
```
ablack12, line 3980: What parts of the project ref do you need? You probably need to use FindMergedProjectRef but I'm not sure what the use case is here.

khelif96: Is FindMergedProjectRef preferable?

graphql/resolvers.go
```diff
@@ -3976,6 +3976,20 @@ func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion)
...
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding project ref for project `%s`: %s", *obj.Project, err.Error()))
+	}
+	if projectRef == nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding project ref for project `%s`: %s", *obj.Project, "Project not found"))
+	}
+	apiProjectRef := restModel.APIProjectRef{}
+	if err = apiProjectRef.BuildFromService(projectRef); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error building APIProjectRef from service for `%s`: %s", projectRef.Id, err.Error()))
```
SupaJoon, line 3989: can we reword the error message to exclude "error"?

SupaJoon (approved):

graphql/resolvers.go
ablack12: Well if you want all fields to be accurate then yes, FindMergedProjectRef is preferable. However, it is more expensive, and since most fields in the project ref are private to admins (and such they shouldn't be exposed to the general user) I'm not sure it's worth adding this complexity. Returning all fields from FindBranchProjectRef however makes it seem like those fields are populated correctly, and are allowed to be used by the version page, which isn't true, so I'm hesitant about returning all fields.

khelif96: We already return a project ref under the[ projects() resolver](https://github.com/khelif96/evergreen/blob/88c92b4ca2a08c658d760297b20881413756e0e1/graphql/resolvers.go#L1584) which is used by the project selector. Maybe we should come up with another way to block access to sensitive fields?

ablack12: Ah, that's unfortunate. That one definitely needs merging because of how it groups projects though, which I assumed was also used for the project page sidebar (although I'm not entirely sure).

Regardless then, this either should use FindMerged or just return owner/branch.

khelif96: https://jira.mongodb.org/browse/EVG-16859 I opened this to address the sensitive fields.

This field is resolved in parallel with other resolvers so i didn't notice an overall hit when using FindMergedProjectRef vs FindBranchProjectRef

ablack12 (approved):

PR #5655, opened 2022-05-11
EVG-16848 handle Github conflicts in more places

ablack12:
[EVG-16848](https://jira.mongodb.org/browse/EVG-16848)

### Description 
We want to try to make sure that a project doesn't get attached or change owner/repo and suddenly introduce project conflicts on the UI that prevent the page from being usable, and makes it ambiguous in the backend which project is being used. 

I had this silently update conflicts. For AttachToRepo this definitely makes sense, since any new conflicts would be due to defaulting to repo, so explicitly setting to false is pretty much the same as what they had before. For AttachToNewRepo this may not be clear, however I don't anticipate this being used often (if at all) except to rename existing repos, and returning an error in the conflict case would require some refactoring and also may not be clear anyway. (I can change this if there are objections though.)

Also added comments and did some refactor tweaking while I was poking around.

### Testing 
Added to unit test.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -411,6 +411,8 @@ func (p *ProjectRef) Add(creator *user.DBUser) error {
 	if p.Id == "" {
 		p.Id = mgobson.NewObjectId().Hex()
 	}
+	// Ensure that any new project is originally explicitly disabled.
+	p.Enabled = utility.FalsePtr()
 
 	// if a hidden project exists for this configuration, use that ID
 	if p.Owner != "" && p.Repo != "" && p.Branch != "" {
@@ -436,7 +438,7 @@ func (p *ProjectRef) Add(creator *user.DBUser) error {
 	if err != nil {
 		return errors.Wrap(err, "inserting project ref")
 	}
-	return p.AddPermissions(creator)
+	return p.addPermissions(creator)
 }
 
 func (p *ProjectRef) GetPatchTriggerAlias(aliasName string) (patch.PatchTriggerDefinition, bool) {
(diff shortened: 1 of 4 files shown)


model/project_ref.go
```diff
@@ -486,30 +488,9 @@ func (p *ProjectRef) MergeWithProjectConfig(version string) (err error) {
 	return err
 }
 
-// AttachToRepo adds the branch to the relevant repo scopes, and updates the project to point to the repo.
-// Any values that previously were unset will now use the repo value.
-// If no repo ref currently exists, the user attaching it will be added as the repo ref admin.
-func (p *ProjectRef) AttachToRepo(u *user.DBUser) error {
```
ablack12, line 492: Moved this function down to live with AttachToNewRepo

model/project_ref.go
```diff
@@ -654,19 +663,60 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
...
 	return GetAndLogProjectModified(p.Id, u.Id, false, before)
 }
 
+// addGithubConflictsToUpdate turns off any settings that may introduce conflicts by
+// adding fields to the given update and returning them.
+func (p *ProjectRef) addGithubConflictsToUpdate(update bson.M) bson.M {
+	// If the project ref doesn't default to repo, will just return the original project.
+	mergedProject, err := GetProjectRefMergedWithRepo(*p)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
```
ablack12, line 688: Chose not to error here since errors are unlikely, and erroring would require us to revert changes that could also error (similar to the case I described in the comments, I'm not sure it's worth all that). Open to discussion on it though.

model/project_ref.go
```diff
@@ -800,30 +852,19 @@ func FindMergedProjectRef(identifier string, version string, includeProjectConfi
...
 }
 
-// GetProjectRefMergedWithRepo merges the project with the repo that matches it, if one exists.
+// GetProjectRefMergedWithRepo merges the project with the repo, if one exists.
 // Otherwise, it will return the project as given.
 func GetProjectRefMergedWithRepo(pRef ProjectRef) (*ProjectRef, error) {
 	if !pRef.UseRepoSettings() {
 		return &pRef, nil
 	}
-	if pRef.RepoRefId != "" {
```
ablack12, line 809: If useResoSettings() is true then RepoRefId is also true. I couldn't actually find a case where we relied on the FindRepoRefByOwnerAndRepo case being used.

sophstad (approved):
💯

model/project_ref.go
```diff
@@ -654,19 +663,60 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
...
 }
 
+// addGithubConflictsToUpdate turns off any settings that may introduce conflicts by
+// adding fields to the given update and returning them.
+func (p *ProjectRef) addGithubConflictsToUpdate(update bson.M) bson.M {
+	// If the project ref doesn't default to repo, will just return the original project.
+	mergedProject, err := GetProjectRefMergedWithRepo(*p)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
+			"message":            "unable to merge project with attached repo",
```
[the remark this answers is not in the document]
ablack12: Theses style guide suggestions are for wrapped errors, whereas these messages are logged directly to splunk, so it's okay for the messages to be a bit more readable.

model/project_ref.go
```diff
@@ -654,19 +663,60 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
...
+			grip.Debug(message.WrapError(err, message.Fields{
+				"message":            "unable to get github project conflicts",
+				"project_id":         p.Id,
+				"project_identifier": p.Identifier,
+				"repo_id":            mergedProject.RepoRefId,
+			}))
+			return update
+		}
+		if len(conflicts.CommitQueueIdentifiers) > 0 {
+			update[bsonutil.GetDottedKeyName(projectRefCommitQueueKey, commitQueueEnabledKey)] = false
```
malikchaya2, line 708: maybe I'm not fully understanding what's happening here, but how do we know that the user doesn't want to explicitly have this set to true?

ablack12: Yes exactly, and that branch should take priority.

Alternatively, we could suppress moving repo for enabled projects overall, and that would remove the conflicts.

malikchaya2: Should we add some help text to the modal letting users know that these values will be automatically set?

malikchaya2 (approved):

model/project_ref.go
ablack12: Sure I can just add that to the tooltip PR

PR #5656, opened 2022-05-12
EVG-16103 Fix unreachable task conflict handling in agent

hadjri:

### Description 
Some of the logic to handle a task encountering a conflict error was unreachable in the agent heartbeat loop. Fixed so that the loop context is properly cancelled when this occurs (and from the issue yesterday, ensuring that context is not cancelled when the heartbeat receives an abort request but simply sending failed status to the heartbeat channel in this case).
### Testing 
Added unit test


agent/internal/client/host_methods.go
```diff
@@ -278,8 +278,8 @@ func (c *hostCommunicator) GetExpansions(ctx context.Context, taskData TaskData)
 }
 
 // Heartbeat sends a heartbeat to the API server. The server can respond with
-// an "abort" response. This function returns true if the agent should abort.
-func (c *hostCommunicator) Heartbeat(ctx context.Context, taskData TaskData) (bool, error) {
+// an "abort" response. Returning evergreen.TaskFailed signals the agent to abort the task.
```
Kimchelly, line 281: What about `evergreen.TaskConflict`?

agent/internal/client/interface.go
```diff
@@ -48,8 +48,8 @@ type Communicator interface {
 	// GetExpansions returns all expansions for the task known by the app server
 	GetExpansions(context.Context, TaskData) (util.Expansions, error)
 	// Heartbeat sends a heartbeat to the API server. The server can respond with
-	// an "abort" response. This function returns true if the agent should abort.
-	Heartbeat(context.Context, TaskData) (bool, error)
+	// an "abort" response. Returning evergreen.TaskFailed signals the agent to abort the task.
```
Kimchelly, line 51: Same question as above.

agent/internal/client/pod_methods.go
```diff
@@ -86,9 +86,9 @@ func (c *podCommunicator) GetExpansions(ctx context.Context, taskData TaskData)
 }
 
 // Heartbeat sends a heartbeat to the API server. The server can respond with
-// an "abort" response. This function returns true if the agent should abort.
-func (c *podCommunicator) Heartbeat(ctx context.Context, taskData TaskData) (bool, error) {
-	return false, errors.New("TODO: implement")
+// an "abort" response. Returning evergreen.TaskFailed signals the agent to abort the task.
```
Kimchelly, line 89: Same question as above.

Kimchelly (changes requested):

agent/internal/client/host_methods.go
hadjri: Fixed

Kimchelly: Usually, when I write a comment, I try to communicate high-level intent and behavior. I feel like the comment is overly specifying the implementation of the agent's abort process (which is external to this method) by saying that it must cancel the context and must continue "the agent loop". Instead, it would be more useful/maintainable to say what the heartbeat's return value actually signifies and let the agent deal with how that is handled. I would instead say how the agent should interpret a task conflict at a high level.

I'm assuming `evergreen.TaskConflict` means that the agent is no longer authorized to run this task (due to the task secret changing) - maybe phrase as something along these lines:
"Heartbeat will return a non-empty task status if the agent should no longer run the task. Returning TaskConflict means the agent is no longer authorized to run this task and should move on to the next available one"

agent/internal/client/interface.go
Kimchelly: If the comment is going to be the same across the interface and the 2 implementations (+ mock), would it be easier to just stick the comment above the Heartbeat interface method and remove all the duplicate comments? I'm assuming the implementations will all have the same behavior.

hadjri: Yes they will - removed comments on the implementations to make the comment easier to maintain

agent/internal/client/host_methods.go
hadjri: Sounds good - updated comment to be more high-level

agent/background.go
```diff
@@ -30,11 +28,13 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 			signalBeat, err = a.doHeartbeat(ctx, tc)
 			if signalBeat == evergreen.TaskConflict {
 				tc.logger.Task().Error("Encountered task conflict while checking heartbeat, aborting task")
-				heartbeat <- evergreen.TaskFailed
+				if err != nil {
```
ablack12, line 31: So this was the unreachable code path, correct? Is there a reason why we no longer want to assign something to heartbeat?

Kimchelly: The [select that waits for the heartbeat](https://github.com/evergreen-ci/evergreen/blob/a4cd5cd31a93e48486fd275876723b565008a653/agent/agent.go#L520-L527) can accept either context cancellation or a heartbeat status, so from a correctness/consistency perspective, the heartbeat must either cancel the context _or_ send a task status, but not both. In this case, I think `evergreen.StatusConflict` is not a real task status, so it seems like cancelling is preferable to sending a fake task status and hoping that it correctly handles it.

Kimchelly (approved):

agent/background.go
hadjri: It isn't that we are no longer sending a status to the heartbeat channel - it was actually never set up this way, here is an older version of the same function: https://github.com/evergreen-ci/evergreen/blob/42e23e63cb85a0f303e7e7dbfa36f37b3b85c556/agent/background.go#L32

In hindsight this cancel() was never getting reached though.  In that regard we are making it such that the old code is reachable.  

A child context is created from the parent context used in the heartbeat loop [here](https://github.com/evergreen-ci/evergreen/blob/main/agent/agent.go#L498): When the parent heartbeat context gets cancelled, this innerContext  Done channel is closed and it looks like the two startTask and handleTaskResponse functions will not complete, causing the agent to return back to its high-level loop looking for new tasks which I think is the intended scenario when we run into a task conflict.

Conversely when we receive an abort signal we have historically just been sending a failed signal through the heartbeat channel which propagates down through the agent to [finishTask](https://github.com/evergreen-ci/evergreen/blob/098d5c237fde79ce9cf7a6884b8362f332c512c7/agent/agent.go#L577), and will cause post task commands etc to run. I think it makes sense that we wouldn't want to cancel the context when this happens which is why we saw some problems with the change deployed yesterday that cancelled context on abort signals.

PR #5658, opened 2022-05-12
EVG-16863: populate host termination reason from note

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16863

### Description 
If the user terminates the host with a note from the host event page, the note should eventually show up in the event logs.

### Testing 
Didn't seem that critical enough to test, and I didn't feel like unit testing all this existing untested code.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -105,9 +105,13 @@ func ModifyHostStatus(ctx context.Context, env evergreen.Environment, queue ambo
 			return "", http.StatusInternalServerError, errors.New(HostTerminationQueueingError)
 		}
 
+		reason := notes
+		if reason == "" {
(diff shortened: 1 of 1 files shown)


ablack12 (approved):

api/host.go
```diff
@@ -105,9 +105,13 @@ func ModifyHostStatus(ctx context.Context, env evergreen.Environment, queue ambo
 			return "", http.StatusInternalServerError, errors.New(HostTerminationQueueingError)
 		}
 
+		reason := notes
```
ablack12, line 108: Should this actually be handled also outside of HostTerminated so it gets hit at 117?

ablack12 (approved):

PR #5661, opened 2022-05-12
EVG-5756 Add local_aliases field to client settings

bynn:
[EVG-5756](https://jira.mongodb.org/browse/EVG-5756)

### Description 
added local_aliases that can be set in personal evergreen.ymls 

**update 
the added verification doesnt allow users to combine task tags and names 

### Testing 
unittest
image

can define for each project 
accepts tags and names 
can provide a comma separated list of regexs for tasks and variants 

image


operations/model.go
```diff
@@ -77,18 +78,19 @@ func isValidPath(path string) bool {
...
+	User                  string               `json:"user" yaml:"user,omitempty"`
+	UncommittedChanges    bool                 `json:"patch_uncommitted_changes" yaml:"patch_uncommitted_changes,omitempty"`
+	AutoUpgradeCLI        bool                 `json:"auto_upgrade_cli" yaml:"auto_upgrade_cli,omitempty"`
+	PreserveCommits       bool                 `json:"preserve_commits" yaml:"preserve_commits,omitempty"`
+	Projects              []ClientProjectConf  `json:"projects" yaml:"projects,omitempty"`
+	Admin                 ClientAdminConf      `json:"admin" yaml:"admin,omitempty"`
+	LoadedFrom            string               `json:"-" yaml:"-"`
+	DisableAutoDefaulting bool                 `json:"disable_auto_defaulting" yaml:"disable_auto_defaulting"`
+	ProjectsForDirectory  map[string]string    `json:"projects_for_directory,omitempty" yaml:"projects_for_directory,omitempty"`
+	LocalAliases          []model.ProjectAlias `json:"local_aliases,omitempty" yaml:"local_aliases,omitempty"`
```
ablack12, line 93: Should this be moved under ClientProjectConf? It seems possible for users to have the `test` alias defined for each of their different repos, for example.

operations/patch.go
```diff
@@ -148,6 +148,20 @@ func Patch() cli.Command {
 				return errors.Errorf("can't define tasks/variants when reusing previous patch's tasks and variants")
 			}
 
+			if params.Alias != "" {
+				for _, alias := range conf.LocalAliases {
+					if alias.Alias == params.Alias {
```
ablack12, line 153: We should validate the alias at this point; if both variant and task both aren't populated, we should error. I'm also thinking we can support tags if we add these to the regular Tasks/Variants logic.

operations/patch.go
```diff
@@ -148,6 +148,20 @@ func Patch() cli.Command {
 				return errors.Errorf("can't define tasks/variants when reusing previous patch's tasks and variants")
 			}
 
+			if params.Alias != "" {
```
ablack12, line 151: We should move this above the validation at 147.

ablack12 (changes requested):
Did you try testing this with regex tags/variants also included, and verifying that both regexes are used? (Same with tags if we can implement that)

rest/route/build.go
```diff
@@ -76,7 +76,7 @@ func (b *buildGetHandler) Run(ctx context.Context) gimlet.Responder {
 
 ////////////////////////////////////////////////////////////////////////
 //
-// PATH /builds/{build_id}
+// PATCH /builds/{build_id}
```
ablack12, line 79: lol

malikchaya2 (approved):
lgtm after Annie's comments :)

operations/patch.go
bynn: done

bynn (commented):
i changed it so that you can specify your own alias for each project and allows tags and multiple regex's 

example update in description

operations/patch_util.go
```diff
@@ -289,20 +289,60 @@ func (p *patchParams) loadProject(conf *ClientSettings) error {
...
+	if project == "" {
 		conf.SetDefaultProject(cwd, p.Project)
 		if err := conf.Write(""); err != nil {
 			grip.Warning(message.WrapError(err, message.Fields{
 				"message": "failed to set default project",
 				"project": p.Project,
 			}))
 		}
+	} else {
+		p.Project = project
```
ablack12, line 345: Can you explain this change?

operations/patch_util.go
```diff
@@ -289,20 +289,60 @@ func (p *patchParams) loadProject(conf *ClientSettings) error {
 	return nil
 }
 
+func (p *patchParams) setLocalAliases(conf *ClientSettings) error {
+	for i := range conf.Projects {
+		if conf.Projects[i].Name == p.Project {
+			for _, alias := range conf.Projects[i].LocalAliases {
+				if alias.Alias == p.Alias {
+					if (alias.Variant != "" || alias.VariantTags != nil) && (alias.Task != "" || alias.TaskTags != nil) {
```
ablack12, line 297: I understand what's going on here after staring for a minute but it's hard to read; I would add a function comment or a comment to this large if block, and/or refactor to a new function called "addAliasToPatchParams" or something.

ablack12: Additionally, before adding an alias to the patch params, you should validate that the alias is valid:
https://github.com/evergreen-ci/evergreen/blob/e535f590fa4e06a7bd0502255e27de21b39c3a10/model/project_aliases.go#L490

operations/patch_util.go
ablack12: I'm a little confused about how that's possible, since it's possible to type `evergreen patch` from a directory that is specified in your yaml without the `-p` flag, and it still works, so we must set it somewhere.
I'm guessing it's [this line](https://github.com/evergreen-ci/evergreen/blob/4fb33a038c3091f7699757b04931774abd94a557/operations/patch_util.go#L220)? In which case I'm wondering if this function already does what you want. (I didn't deep read to understand how these functions are different though)

operations/patch.go
```diff
@@ -144,6 +145,20 @@ func Patch() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "problem loading configuration")
 			}
+			params.setDefaultProject(conf)
+
+			for _, project := range conf.Projects {
+				if errStrs := model.ValidateProjectAliases(project.LocalAliases, "Local Aliases"); errStrs != nil {
```
ablack12, line 151: I don't think when a user wants to create a patch we _always_ need to validate all aliases in their project. We should be able to instead do this in setLocalAliases just for any aliases that are used I'd think.

operations/patch_util.go
bynn: ah this is where they were setting it 
i thought it was the setDefaultProject but was also confused on why it wasnt already being set

malikchaya2 (approved):

operations/patch_util.go
```diff
@@ -289,6 +293,53 @@ func (p *patchParams) loadProject(conf *ClientSettings) error {
...
+	if p.Alias != "" {
+		for i := range conf.Projects {
+			if conf.Projects[i].Name == p.Project {
+				if errStrs := model.ValidateProjectAliases(conf.Projects[i].LocalAliases, "Local Aliases"); len(errStrs) != 0 {
+					return errors.Errorf("validating local aliases: '%s'", errStrs)
+				}
+				for _, alias := range conf.Projects[i].LocalAliases {
+					if alias.Alias == p.Alias {
+						p.addAliasToPatchParams(alias)
+					} else {
```
ablack12, line 306: This if/else statement doesn't make sense; you're saying "if the first alias in the list doesn't match, return an error. Besides, we should only error if the alias has an invalid definition (like at 300) since aliases could be project-level.

operations/patch_util.go
```diff
@@ -289,6 +293,53 @@ func (p *patchParams) loadProject(conf *ClientSettings) error {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// addAliasToPatchParams add the matching local alias to patch params.
+func (p *patchParams) addAliasToPatchParams(alias model.ProjectAlias) {
+	if (alias.Variant != "" || alias.VariantTags != nil) && (alias.Task != "" || alias.TaskTags != nil) {
```
ablack12, line 318: We don't need this overall if statement now that we validate I think.

ablack12 (changes requested):

operations/patch.go
```diff
@@ -177,6 +174,16 @@ func Patch() cli.Command {
 			}
 			params.Description = params.getDescription()
 
+			if params.Alias != "" {
```
ablack12, line 177: Do we need this since we already do this in validatePatchCommand?

operations/patch_util.go
bynn: yeah we dont

operations/patch.go
```diff
@@ -177,6 +174,14 @@ func Patch() cli.Command {
 			}
 			params.Description = params.getDescription()
 
+			if err = params.setLocalAliases(conf); err != nil {
```
ablack12, line 177: I can't put a comment on the actual line SOS but we should change the patch flag description now to say something about how aliases can be defined locally

ablack12 (changes requested):
LGTM with incrementing client version string and the help text update; we should also add documentation somewhere for this so that people can reference it for the formatting

ablack12 (approved):

PR #5662, opened 2022-05-12
EVG-16158 Add host create details to task

bynn:
[EVG-16158](https://jira.mongodb.org/browse/EVG-16158)

### Description 
add error 

### Testing 
not sure how to test


units/provisioning_create_host.go
```diff
@@ -116,13 +116,17 @@ func (j *createHostJob) Run(ctx context.Context) {
...
+			notFoundError := errors.New("could not find host")
 			grip.Warning(message.Fields{
 				"host_id": j.HostID,
 				"task_id": j.TaskID,
 				"attempt": j.RetryInfo().CurrentAttempt,
 				"job":     j.ID(),
-				"message": "could not find host",
+				"message": notFoundError.Error(),
 			})
+			if err := task.AddHostCreateDetails(j.host.StartedBy, j.host.Id, j.host.SpawnOptions.TaskExecutionNumber, notFoundError); err != nil {
```
ybrill, line 127: iirc `host.list` relies on the number of errors in host create details in the task corresponding to the number of failed hosts the task tried to spawn. If so:
If it's in fact the case that the host document doesn't exist here is it definitely going to be because of an error spawning it? And is that error definitely not going to have added an entry in host create details?

units/provisioning_create_host.go
```diff
@@ -116,13 +116,17 @@ func (j *createHostJob) Run(ctx context.Context) {
 		}
 		if j.host == nil {
 			//host intent document has been removed by another evergreen process
+			notFoundError := errors.New("could not find host")
 			grip.Warning(message.Fields{
 				"host_id": j.HostID,
 				"task_id": j.TaskID,
 				"attempt": j.RetryInfo().CurrentAttempt,
 				"job":     j.ID(),
-				"message": "could not find host",
```
ablack12, line 124: We usually try to keep messages something readable; I think the only benefit to including the error in this grip statement is if we want the stack trace, otherwise wrapping it seems like clutter.
I do also think "could not find host" isn't a helpful user error to add to host create details at 127; do we know why the host isn't found at this point? Were we unable to create it?

units/provisioning_create_host.go
bynn: i added an extra condition to the current condition of maxing out the attempts

units/provisioning_create_host.go
```diff
@@ -214,9 +214,9 @@ func (j *createHostJob) Run(ctx context.Context) {
 	}
 
 	defer func() {
-		if j.RetryInfo().GetRemainingAttempts() == 0 && j.HasErrors() && (j.host.Status == evergreen.HostUninitialized || j.host.Status == evergreen.HostBuilding) && j.host.SpawnOptions.SpawnedByTask {
+		if ((j.RetryInfo().GetRemainingAttempts() == 0 && j.HasErrors()) || j.host == nil) && (j.host.Status == evergreen.HostUninitialized || j.host.Status == evergreen.HostBuilding) && j.host.SpawnOptions.SpawnedByTask {
```
ybrill, line 217: it looks to me like the condition in its entirety will never be true for a nil host (because `j.host.Status` will be "").

Separately, I'm not sure this addresses the concern I raised earlier: if we didn't hit an error in the job maybe the host was really spawned successfully by another parallel job. Or really who knows why the host document isn't there. Does it necessarily imply an error?

cloud/ec2.go
```diff
@@ -426,6 +427,13 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 			"host_provider": h.Distro.Provider,
 			"distro":        h.Distro.Id,
 		}))
+		detailErr := task.AddHostCreateDetails(h.StartedBy, h.Id, h.SpawnOptions.TaskExecutionNumber, err)
+		grip.ErrorWhen(detailErr != nil, message.WrapError(detailErr, message.Fields{
```
ablack12, line 431: ErrorWhen(detailErr != nil) is actually the same as grip.Error(message.WrapError(detailErr, ...) because grip will noop if detailErr is nil.

cloud/ec2.go
```diff
@@ -426,6 +427,13 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
...
+			"message":       "adding host create error details",
```
ablack12, line 432: For splunk messages, I think it makes sense for these to be full sentences since they aren't wrapped errors (added a suggestion to the convention doc)

ablack12 (approved):

cloud/ec2.go
```diff
@@ -426,6 +427,13 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 			"host_provider": h.Distro.Provider,
 			"distro":        h.Distro.Id,
 		}))
+		detailErr := task.AddHostCreateDetails(h.StartedBy, h.Id, h.SpawnOptions.TaskExecutionNumber, err)
```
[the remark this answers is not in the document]
bynn: done

cloud/ec2.go
bynn: done

cloud/ec2.go
bynn: ah thats so smart

ybrill (approved):

PR #5668, opened 2022-05-17
EVG-16782 use correct structures for event log

ablack12:
[EVG-16782](https://jira.mongodb.org/browse/EVG-16782)

### Description 
1. Correctly populate alias ID when building event log.
2. Use the same event log structure for project and repo so that the strong typing doesn't break event logs (since not everything necessarily exists)

### Testing 
Note that I tested before removing the repo event, so the fixes explained above have been tested but won't currently work with spruce -- Sophie is going to fix in a UI PR that will be merged first #buildtogether
Added test data so that the evergreen project and its repo would have a legit project settings object.
Also tested on staging.
No visible change except that things no longer error that used to error.


graphql/schema.graphql
```diff
@@ -1184,17 +1181,6 @@ type ProjectEventLogEntry {
   after: ProjectEventSettings
 }
 
-type RepoEvents {
```
malikchaya2, line 1187: the repo schemas don't seem to be typed more strongly than the project ones, so why does this switch help?

malikchaya2 (changes requested):

graphql/schema.graphql
sophstad: They actually _are_ more strongly typed. The [RepoRef](https://github.com/evergreen-ci/evergreen/blob/9aff211951ec3109842c8abc67d889bd8538070b/graphql/schema.graphql#L1329) struct has a lot of non-nullable fields that are nullable in [Project](https://github.com/evergreen-ci/evergreen/blob/9aff211951ec3109842c8abc67d889bd8538070b/graphql/schema.graphql#L1276). This could cause errors if, for example, a new field is added; previous events would have this field as `null`, but the schema would indicate that the field is non-nullable since this is the case for current projects (as opposed to historic project states).

malikchaya2: thank you for explaining it, I understand better now.

malikchaya2 (approved):

sophstad (approved):

khelif96 (approved):

PR #5669, opened 2022-05-18
EVG-16871 add backend alias validation

ablack12:
[EVG-16871](https://jira.mongodb.org/browse/EVG-16871)

### Description 
Even though the front end does validation to ensure that aliases exist on the project or repo, we should also check this in teh backend.

### Testing 
Unit test.


rest/data/aliases.go
```diff
@@ -131,6 +132,47 @@ func UpdateAliasesForSection(projectId string, updatedAliases []restModel.APIPro
...
+		if err != nil {
+			return err
+		}
+		for _, a := range repoAliases {
+			aliasesMap[a.Alias] = true
+		}
+
+	}
+
+	msg := "%s cannot be enabled without aliases"
```
bynn, line 158: '%s'? even though it'll never be an empty string

rest/data/aliases.go
```diff
@@ -131,6 +132,47 @@ func UpdateAliasesForSection(projectId string, updatedAliases []restModel.APIPro
 	return modified, catcher.Resolve()
 }
 
+// validateFeaturesHaveAliases returns an error if project/repo aliases are not defined for a Github/CQ feature.
+// Does not error if version control is enabled.
+func validateFeaturesHaveAliases(pRef *model.ProjectRef, aliases []restModel.APIProjectAlias) error {
+	if pRef.IsVersionControlEnabled() {
```
bynn, line 138: can i ask for clarification on why we dont error

rest/data/aliases.go
ablack12: I'd rather not actually, since this is purely for string formatting purposes and not wrapping an ID or something where it isn't clear where it starts.

rest/data/aliases.go
malikchaya2: >  require that projects choose either to use repo-level settings OR yaml settings (or neither, and stick with the existing format where all settings are defined in only one place)

I like this thing you mentioned in the ticket, at least as a first step until we start gathering more user feedback.

rest/data/aliases.go
```diff
@@ -131,6 +132,47 @@ func UpdateAliasesForSection(projectId string, updatedAliases []restModel.APIPro
...
+		return nil
+	}
+
+	aliasesMap := map[string]bool{}
+	for _, a := range aliases {
+		aliasesMap[utility.FromStringPtr(a.Alias)] = true
+	}
+
+	if pRef.RepoRefId != "" {
```
malikchaya2, line 147: we should probably use pRef.UseRepoSettings() here instead

malikchaya2 (approved):
LGTM with nit

sophstad (approved):

PR #5673, opened 2022-05-20
EVG-16870 Prevent dead / unused code from passing lint

hadjri:
[EVG-16870](https://jira.mongodb.org/browse/EVG-16870)

### Description 
There are quite a few chunks of unused and dead code in the codebase, as well as a few instances where the logic can be simplified for readability. The Go linter can check for these things moving forward so that we can maintain the project more easily.  These changes reflect what was a necessary 'cleanup' before we can incorporate these new checks in our linter configuration.
### Testing 
Tests should remain passing.


.golangci.yml
```diff
@@ -8,6 +8,11 @@ linters:
     - ineffassign
     - misspell
     - unconvert
+    - structcheck
+    - asciicheck
+    - unused
+    - gosimple
+    - deadcode
```
Kimchelly, line 15: nit: can we keep this list in alphabetical order?

cloud/ec2_fleet_test.go
```diff
@@ -13,10 +13,15 @@ import (
 	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/model/distro"
 	"github.com/evergreen-ci/evergreen/model/host"
+	"github.com/evergreen-ci/evergreen/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
+	testutil.Setup()
```
Kimchelly, line 22: This doesn't seem related to linting. What is this change for?

config_podinit.go
```diff
@@ -1,16 +1,13 @@
...
 import (
-	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-var podInitConfigKey = bsonutil.MustHaveTag(Settings{}, "PodInit")
```
Kimchelly, line 12: I would put a [`//nolint: deadcode, unused`](https://github.com/evergreen-ci/evergreen/blob/168aae8a383eb8a76041e89597bace57d02902ac/config_db.go#L15) above this rather than remove it because we generally keep BSON tag keys even if we're not using them.

db/mgo/bson/bson.go
```diff
@@ -301,12 +301,12 @@ func (id *ObjectId) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 	if len(data) != 26 || data[0] != '"' || data[25] != '"' {
-		return errors.New(fmt.Sprintf("invalid ObjectId in JSON: %s", string(data)))
+		return fmt.Errorf("invalid ObjectId in JSON: %s", string(data))
```
Kimchelly, line 304: `errors.Errorf`

db/mgo/bson/bson.go
```diff
@@ -301,12 +301,12 @@ func (id *ObjectId) UnmarshalJSON(data []byte) error {
...
 	}
 	var buf [12]byte
 	_, err := hex.Decode(buf[:], data[1:25])
 	if err != nil {
-		return errors.New(fmt.Sprintf("invalid ObjectId in JSON: %s (%s)", string(data), err))
+		return fmt.Errorf("invalid ObjectId in JSON: %s (%s)", string(data), err)
```
Kimchelly, line 309: `errors.Errorf`

units/commit_queue.go
```diff
@@ -296,26 +295,6 @@ func (j *commitQueueJob) TryUnstick(ctx context.Context, cq *commitqueue.CommitQ
 	}
 
 	// patch is done
-	if !utility.IsZeroTime(patchDoc.FinishTime) {
```
Kimchelly, line 299: Why is this removed? Is it impossible for `patchDoc.FinishTime` to be zero?

units/provisioning_convert_host_to_new.go
```diff
@@ -19,7 +19,6 @@ import (
 
 const (
 	convertHostToNewProvisioningJobName = "convert-host-to-new-provisioning"
-	maxProvisioningConversionAttempts   = 15
```
Kimchelly, line 22: Rather than delete this, it's supposed to be used on [L65](https://github.com/evergreen-ci/evergreen/pull/5673/files#diff-1dff55e2600df65798f1b093187554eef68fb91209d4efbf17911b06537d69ceL65) - can we re-add this and use it on L65?

units/ssh_keys_update_static.go
```diff
@@ -61,41 +56,4 @@ func (j *staticUpdateSSHKeysJob) Run(ctx context.Context) {
 		}
 		j.host = h
 	}
-	if j.host.Status != evergreen.HostRunning {
```
Kimchelly, line 64: Why was this job's logic removed? Without it, this job doesn't do anything.

Kimchelly (changes requested):

.golangci.yml
hadjri: Fixed!

units/commit_queue.go
hadjri: I'm not sure how this got removed.. possibly residual local changes from a different ticket or a bad merge. Re-added

hadjri: I see what happened. When theres a redundant return the linter underlines it in my IntelliJ IDE and offers a suggestion to remove it automatically for you. Sometimes if you click "yes" the return statement and the above logic block gets removed along with it. I must have just clicked for auto-removal and not noticed this logic went away with it. Moral of the story do not use that feature in IntelliJ and just delete the statement with your keyboard haha.

units/ssh_keys_update_static.go
hadjri: Same issue as I posted above. Fixed now

units/provisioning_convert_host_to_new.go
hadjri: Done

cloud/ec2_fleet_test.go
hadjri: I added this while troubleshooting tests in this file - turns out the issue was caused by a change to a regex the linter suggested to use raw string (`...`) notation. This breaks things so I suppressed the linter on that line. Anyways yes, removing this line.

Kimchelly (approved):
Thank you for fixing these things (especially the `== true`, that was bothering me a lot).

PR #5675, opened 2022-05-20
EVG-15769 add patchable and patch_only to s3.put

ablack12:
[EVG-15769 ](https://jira.mongodb.org/browse/EVG-15769)

### Description 
Often teams use s3.put differently for patches than for non-patches. This can make for completely identical tasks except for the s3.put stage, or the commands have to be marked as optional and the is_patch expansion has to be used to figure things out. This is a tricky enough workaround and a common enough use case that it's worth adding patchable and patch_only directly to s3.put.

I'll add this to the docs after the PR is approved.

### Testing 
Only tested patch_only bc I don't want to have to make a commit and I'm not sure that sandbox has s3 permissions but, verified that a patch was able to run a patch_only s3.put command but skipped the patchable: false one: https://spruce-staging.corp.mongodb.com/task/evg_lint_generate_lint_patch_a9e7ac5298111ab45396c18f9b9aef1a9460058d_6287ded7b237364964ab81c9_22_05_20_18_32_59/files?execution=0


agent/command/s3_put.go
```diff
@@ -87,6 +87,12 @@ type s3put struct {
 	// for missing files.
 	Optional string `mapstructure:"optional" plugin:"expand"`
 
+	// Patchable defaults to true. If set to false, this command will noop without error for patch tasks.
+	Patchable string `mapstructure:"patchable" plugin:"patchable"`
+
+	// PatchOnly defaults to false. If set to true, this command will noop without error for non-patch tasks.
+	PatchOnly string `mapstructure:"patch_only" plugin:"patch_only"`
```
malikchaya2, line 94: since patchable and patchOnly seem like inverses of each other, what is the benefit of having both?

malikchaya2: wait no, never mind, lol

agent/command/s3_put.go
```diff
@@ -186,29 +194,43 @@ func (s3pc *s3put) expandParams(conf *internal.TaskConfig) error {
 		return errors.WithStack(err)
 	}
 
+	s3pc.workDir = conf.WorkDir
```
malikchaya2, line 197: nice refactor!

malikchaya2 (approved):

PR #5679, opened 2022-05-23
EVG-16777 Add option to git.get_project

bynn:
[EVG-16777](https://jira.mongodb.org/browse/EVG-16777)

### Description 
allow users to write in options for git clone using 
clone_params: "--filter=something"

because the git version has only updated for rhel80 (and may be upgraded for specific distros in the future) 
this change first calls the git command with the new cloneParams but will delete them if the fetch commands errors 
It does not affect the number of times that fetch retries 

### Testing 
unit tests
staging
image

https://spruce-staging.corp.mongodb.com/version/633dec7e9dbe320e4a7c81f0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


agent/command/git.go
```diff
@@ -154,20 +157,20 @@ func parseToken(token string) (string, error) {
 	return splitToken[1], nil
 }
 
-func (opts cloneOpts) getCloneCommand() ([]string, error) {
+func (opts cloneOpts) getCloneCommand(cloneParams string) ([]string, error) {
```
ablack12, line 160: Why don't we just also add clone params to opts in the opts function?

ablack12 (changes requested):

agent/command/git.go
bynn: done

bynn:
git version has only been updated for rhel80 hosts due to complications for upgrading git on older os's but build may be willing to do so if this change makes things better?

agent/command/git.go
```diff
@@ -208,6 +215,9 @@ func (opts cloneOpts) buildSSHCloneCommand() ([]string, error) {
 	if opts.branch != "" {
 		cloneCmd = fmt.Sprintf("%s --branch '%s'", cloneCmd, opts.branch)
 	}
+	if opts.cloneParams != "" {
+		cloneCmd = fmt.Sprintf("%s %s", cloneCmd, opts.cloneParams)
```
malikchaya2, line 219: If both branch and cloneParams are defined, do we want to combine both in the cloneCmd?

bynn: yes! i think one of the use cases is --single-branch and branch which work together to only clone the specified branch

malikchaya2: But can a user specify both --branch and --cloneParams? Right now, it seems like we overwrite the cloneCmd if cloneParams are defined, but we don't seem to be blocking the user from specifying both. If overwriting it is the desired behavior, then maybe we should error if the user specifies both?

bynn: ah the code is currently adding on cloneParams to the end of cloneCmd if that's what you're suggesting (unless I'm reading it wrong?)

ablack12 (commented):
@bynn are we able to test in staging that this now works for the relevant distro?

bynn:
> @bynn are we able to test in staging that this now works for the relevant distro?

@ablack12 yes! the link in the description has it 
(you have to go to the first execution to see it for the rhel distro because i think the later ones ran when i wasn't on staging)

agent/command/git.go
malikchaya2: Nope, I was reading it wrong :)

agent/command/git.go
```diff
@@ -454,6 +465,10 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
 		func() (bool, error) {
 			err := c.fetch(ctx, comm, logger, conf, opts)
 			if err != nil {
+				if opts.cloneParams != "" && strings.Contains(err.Error(), "exit status 129") {
```
ablack12, line 468: what is this exist status 129? can we make this a function constant as opposed to hardcoding here, with a comment?

agent/command/git.go
```diff
@@ -454,6 +465,10 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
...
+					logger.Execution().Error("Error cloning with clone params, retrying without")
```
ablack12, line 469: Do we do something similar here for any other params? I would think we wouldn't want to do this since it'd falsely make it look like some clone params were successful.

ablack12 (changes requested):

agent/command/git.go
bynn: removed

agent/command/git.go
bynn: removed

ablack12 (approved):

PR #5680, opened 2022-05-24
EVG-16958: Add webhook documents

sophstad:
[EVG-16958](https://jira.mongodb.org/browse/EVG-16958)

### Description 
- Add github_hooks collection so that local projects have webhooks enabled for testing

### Testing 
- Verified in local dev environment that `githubWebhooksEnabled` field is returned as true for evergreen-ci/evergreen and evergreen-ci/spruce projects


ablack12:
Should we now introduce not allowing Github things to be modifiable or saved if webhooks aren't enabled? (I imagine that's on the spruce side but)

sophstad:
> Should we now introduce not allowing Github things to be modifiable or saved if webhooks aren't enabled? (I imagine that's on the spruce side but)

Yup, will do after this!

ablack12 (approved):

PR #5681, opened 2022-05-24
EVG-16954 default nil values

ablack12:
[EVG-16954](https://jira.mongodb.org/browse/EVG-16954)

### Description 
Default nil values for unattached projects.

### Testing 
Verified through testing that unset booleans before the change had no radio box highlighted, and after the change the Disabled one was highlighted.


sophstad (approved):
🎉

PR #5684, opened 2022-05-25
EVG-16650 send task scheduled events to Splunk instead of the db

ybrill:
[EVG-16650](https://jira.mongodb.org/browse/EVG-16650)

### Description 
In order to whittle down the event_log collection I'm going to open a few PRs to reduce the daily write volume to the collection. [We've set a 90 day TTL on the events with a TASK r_type](https://jira.mongodb.org/browse/EVG-15227), but even with the TTL task events are still a significant part of the bulk of the collection. The bulk of these are TASK_SCHEDULED events which don't really need to be in the db. Logging these to Splunk instead will reduce the size of the collection over time.

Once we're already logging to Splunk, include additional information about the parameters that determined each task's position in the queue. This way we'll finally have visibility into why a particular task ranked before another task in the queue. Given two task ids you can answer why one had a higher priority in the queue (see [EVG-15554](https://jira.mongodb.org/browse/EVG-15554))

There's a risk that this will result in too much logging to Splunk. I can't think of a good way to know apriori. We'll get alerted by the Splunk alert if that's the case.

### Testing 
When this was deployed to staging I got the logging [here](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22prioritizated%20distro%20unit%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1653496500&latest=1653496800&sid=1653497588.1802084).

### DIFF
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -30,12 +30,14 @@ const (
 	TaskActivated              = "TASK_ACTIVATED"
 	TaskDeactivated            = "TASK_DEACTIVATED"
 	TaskAbortRequest           = "TASK_ABORT_REQUEST"
-	TaskScheduled              = "TASK_SCHEDULED"
 	ContainerAllocated         = "CONTAINER_ALLOCATED"
 	TaskPriorityChanged        = "TASK_PRIORITY_CHANGED"
 	TaskJiraAlertCreated       = "TASK_JIRA_ALERT_CREATED"
 	TaskDependenciesOverridden = "TASK_DEPENDENCIES_OVERRIDDEN"
 	MergeTaskUnscheduled       = "MERGE_TASK_UNSCHEDULED"
+
+	// TODO (EVG-16969) remove once TaskScheduled events TTL
+	TaskScheduled = "TASK_SCHEDULED"
 )
 
 // implements Data
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -173,11 +173,6 @@ func LogManyTaskAbortRequests(taskIds []string, userId string) {
 		TaskEventData{UserId: userId})
 }
 
-func LogTaskScheduled(taskId string, execution int, scheduledTime time.Time) {
-	logTaskEvent(taskId, TaskScheduled,
-		TaskEventData{Execution: execution, Timestamp: scheduledTime})
-}
-
 func LogTaskContainerAllocated(taskId string, execution int, containerAllocatedTime time.Time) {
 	logTaskEvent(taskId, ContainerAllocated,
 		TaskEventData{Execution: execution, Timestamp: containerAllocatedTime})
--- a/scheduler/planner.go
+++ b/scheduler/planner.go
@@ -186,15 +186,15 @@ type unitInfo struct {
 	// NumDeps is the total number of tasks depending on tasks in the unit.
 	NumDeps int64 `json:"num_deps"`
 	// ContainsInCommitQueue indicates if the unit contains any tasks that are part of a commit queue version.
-	ContainsInCommitQueue bool `json:"in_commit_queue"`
+	ContainsInCommitQueue bool `json:"contains_in_commit_queue"`
 	// ContainsInPatch indicates if the unit contains any tasks that are part of a patch.
-	ContainsInPatch bool `json:"in_patch"`
+	ContainsInPatch bool `json:"contains_in_patch"`
 	// ContainsNonGroupTasks indicates if the unit contains any tasks that are not part of a task group.
-	ContainsNonGroupTasks bool `json:"any_non_group_tasks"`
+	ContainsNonGroupTasks bool `json:"contains_non_group_tasks"`
 	// ContainsGenerateTask indicates if the unit contains generator task.
-	ContainsGenerateTask bool `json:"generate_task"`
+	ContainsGenerateTask bool `json:"contains_generate_task"`
 	// ContainsStepbackTask indicates if the unit contains task activated by stepback.
-	ContainsStepbackTask bool `json:"stepback_task"`
+	ContainsStepbackTask bool `json:"contains_stepback_task"`
 }
 
 func (u *unitInfo) value() int64 {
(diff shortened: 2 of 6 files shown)


scheduler/planner.go
```diff
@@ -281,13 +260,74 @@ func (unit *Unit) RankValue() int64 {
...
 	// that most of our workloads have different runtimes, and we
 	// don't want to have longer makespans if longer running tasks
 	// have to execute after shorter running tasks.
-	unit.cachedValue += priority * unit.distro.GetExpectedRuntimeFactor() * int64(math.Floor(expectedRuntime.Minutes()/float64(length)))
+	value += priority * u.Settings.GetExpectedRuntimeFactor() * int64(math.Floor(u.ExpectedRuntime.Minutes()/float64(length)))
+
+	return value
+}
+
+func (unit *Unit) info() unitInfo {
```
Kimchelly, line 274: No change, but I like the way this was rewritten.

scheduler/planner.go
```diff
@@ -281,13 +260,74 @@ func (unit *Unit) RankValue() int64 {
...
+			info.InCommitQueue = true
+		} else if evergreen.IsPatchRequester(t.Requester) {
+			info.InPatch = true
+		}
+
+		if t.TaskGroup == "" {
+			info.AnyNonGroupTasks = true
+		}
+		if t.GenerateTask {
+			info.GenerateTask = true
```
Kimchelly, line 290: nit: some of these if-and-then-set-true blocks could be simplified down (e.g. `info.GenerateTask = t.GenerateTask`).

Kimchelly (approved):
LGTM mod unimportant nits.

model/event/task_event.go
```diff
@@ -173,11 +173,6 @@ func LogManyTaskAbortRequests(taskIds []string, userId string) {
 		TaskEventData{UserId: userId})
 }
 
-func LogTaskScheduled(taskId string, execution int, scheduledTime time.Time) {
```
ablack12, line 176: Right now the UI resolver looks for scheduled events to avoid cluttering up the UI:
https://github.com/evergreen-ci/evergreen/blob/a1df2bd607681cc3c5f60de215abce9c855a82dd/graphql/resolvers.go#L2048
Should we make a note to remove that logic after the TTL?

scheduler/planner.go
```diff
@@ -170,105 +172,82 @@ func (unit *Unit) ID() string {
...
+	TimeInQueue time.Duration `json:"time_in_queue_ns"`
+	// TotalPriority is the sum of the priority values of all the tasks in the unit.
+	TotalPriority int64 `json:"total_priority"`
+	// NumDeps is the total number of tasks depending on tasks in the unit.
+	NumDeps int64 `json:"num_deps"`
+	// InCommitQueue indicates if the unit contains any tasks that are part of a commit queue version.
+	InCommitQueue bool `json:"in_commit_queue"`
+	// InPatch indicates if the unit contains any tasks that are part of a patch.
+	InPatch bool `json:"in_patch"`
+	// GenerateTask indicates if the unit contains any tasks that are not part of a task group.
```
ablack12, line 192: Wrong variable for comment

scheduler/planner.go
```diff
@@ -281,13 +260,74 @@ func (unit *Unit) RankValue() int64 {
...
+func (unit *Unit) RankValue() int64 {
+	if unit.cachedValue > 0 {
+		return unit.cachedValue
+	}
+
+	info := unit.info()
+	unit.cachedValue = info.value()
+
+	grip.Info(message.Fields{
+		"message":   "prioritizated distro unit",
```
ablack12, line 326: prioritized I'm guessing, unless we're coining an new word 🤪

ablack12 (approved):
LGTM with some nits

model/event/task_event.go
ybrill: I added TODOs and opened [EVG-16969](https://jira.mongodb.org/browse/EVG-16969)

scheduler/planner.go
ybrill: fixed

scheduler/planner.go
ybrill: 😅 
fixed

scheduler/planner.go
ybrill: I refactored to use the value of t.GenerateTask

scheduler/planner.go
ybrill: thank you!

scheduler/planner.go
```diff
@@ -170,105 +172,82 @@ func (unit *Unit) ID() string {
...
+	// TotalPriority is the sum of the priority values of all the tasks in the unit.
+	TotalPriority int64 `json:"total_priority"`
+	// NumDeps is the total number of tasks depending on tasks in the unit.
+	NumDeps int64 `json:"num_deps"`
+	// InCommitQueue indicates if the unit contains any tasks that are part of a commit queue version.
+	InCommitQueue bool `json:"in_commit_queue"`
+	// InPatch indicates if the unit contains any tasks that are part of a patch.
+	InPatch bool `json:"in_patch"`
+	// GenerateTask indicates if the unit contains any tasks that are not part of a task group.
+	AnyNonGroupTasks bool `json:"any_non_group_tasks"`
```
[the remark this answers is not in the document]
Kimchelly: My question is basically in what case would a unit contain multiple tasks when it's not a task group.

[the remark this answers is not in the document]
ybrill: ah, I understand now.
It appear from [here](https://github.com/evergreen-ci/evergreen/blob/e1cc000830857b640b606cb6e36723f6b91cc8d6/scheduler/planner.go#L363) one reason is like you said, when the distro groups versions. Another is that we group tasks with their dependencies.

ybrill:
We decided this was in fact too much logging 😞 . It will be removed by [EVG-17060](https://jira.mongodb.org/browse/EVG-17060)

PR #5685, opened 2022-05-25
EVG-16872: clean up unrecognized pods

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16872

### Description
When we get an SNS notification for a pod that we aren't tracking in the DB (e.g. due to multiple cloud pods being created for a single DB pod), terminate it to clean up the resources.

* Clean up unrecognized pods that are within Evergreen-owned clusters.
* Upgrade Cocoa so that Evergreen has the necessary features for this change.
* Add a service flag to disable unrecognized pod cleanup. We need the flag because Cocoa and Evergreen staging share the same underlying ECS infrastructure, and we don't want tests in Cocoa triggering pod cleanup in staging. This can be removed once [BUILD-13643](https://jira.mongodb.org/browse/BUILD-13643) is complete.

### Testing
* Added unit tests.
* Tested the integration in staging by manually starting some pods with the AWS CLI. The SNS notification did trigger and attempted to make the API call to stop the pod. Unfortunately, due to a change in the staging hosts' IAM roles, the staging apps no longer has the permissions to make ECS API calls (even though it used to work). Build is working on fixing this.


rest/route/sns.go
```diff
@@ -357,23 +366,31 @@ func (sns *ecsSNS) handleNotification(ctx context.Context, notification ecsEvent
...
-		switch notification.Detail.LastStatus {
-		case ecs.TaskStatusProvisioning, ecs.TaskStatusPending, ecs.TaskStatusActivating, ecs.TaskStatusRunning:
+		status := ecs.TaskStatus(notification.Detail.LastStatus).ToCocoaStatus()
+		switch status {
+		case cocoa.StatusStarting, cocoa.StatusRunning:
 			// No-op because the pod is not considered running until the agent
 			// contacts the app server.
 			return nil
-		case ecs.TaskStatusDeactivating, ecs.TaskStatusStopping, ecs.TaskStatusDeprovisioning:
+		case cocoa.StatusStopping:
```
hadjri, line 389: Can this be consolidated to line 385?

rest/route/sns.go
```diff
@@ -425,3 +442,69 @@ func (sns *ecsSNS) handleStoppedPod(ctx context.Context, p *model.APIPod, reason
...
+		SetTaskID(details.TaskARN)
+
+	statusInfo := cocoa.NewECSPodStatusInfo().SetStatus(status)
+
+	podOpts := ecs.NewBasicECSPodOptions().
+		SetClient(c).
+		SetStatusInfo(*statusInfo).
+		SetResources(*resources)
+
+	p, err := ecs.NewBasicECSPod(podOpts)
```
hadjri, line 491: This is probably me just misunderstanding but - the documentation for this function says it "initializes a new pod that is backed by ECS".  Later we delete it and its owned resources, but are we just creating a pod here to delete it later? I'm assuming that this pod that we need to clean up should already exist.

hadjri: Gotcha, makes sense. Thanks!

hadjri (approved):

PR #5686, opened 2022-05-25
EVG-16858 include display tasks when checking existing generated tasks

ablack12:
[EVG-16858](https://jira.mongodb.org/browse/EVG-16858)

### Description 
Alas, Find omits display tasks, which means the check for existing display tasks at 1705 - 1707 never works. 

I also noticed we still have "sanity check" used throughout the codebase, which is not super PC these days so I went ahead and changed those.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -746,15 +746,15 @@ func createTasksForBuild(project *Project, pRef *ProjectRef, buildVariant *Build
 	tasksInBuild []task.Task, syncAtEndOpts patch.SyncAtEndOptions, distroAliases map[string][]string, createTime time.Time,
 	githubChecksAliases ProjectAliases) (task.Tasks, error) {
 
-	// the list of tasks we should create.  if tasks are passed in, then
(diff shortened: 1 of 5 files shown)


model/lifecycle.go
```diff
@@ -1678,8 +1678,8 @@ func addNewTasks(ctx context.Context, activationInfo specificActivationInfo, v *
 	activatedTasks := []task.Task{}
 	for _, b := range existingBuilds {
 		wasActivated := b.Activated
-		// Find the set of task names that already exist for the given build
-		tasksInBuild, err := task.FindWithFields(task.ByBuildId(b.Id), task.DisplayNameKey, task.ActivatedKey)
+		// Find the set of task names that already exist for the given build, including display tasks.
+		tasksInBuild, err := task.FindAll(db.Query(task.ByBuildId(b.Id)).WithFields(task.DisplayNameKey, task.ActivatedKey))
```
ablack12, line 1682: This is the only real change

hadjri (approved):

PR #5690, opened 2022-05-26
EVG-12923 remove re-computing dependencies and reset last running task

ablack12:

### Description 

The couple of times we run into this now seem valid; we don't 100% guarantee that task group tasks will run on the same host (for example, if something happens to host1, then everything else can try to run on host2). That means sometimes funny races will happen. One I've found is when host1 is running a task group but is quarantined, host2 will start running the task group, but then the quarantined host starts running again and thinks it can continue with the task group since that was its last running task, even though host2 has now taken over (headache yet?).

Whereas before improvements were made we hit this error many times a day, we now only hit it a couple of times a week, and always for cases like the above where it makes sense for later task group tasks to be blocked since the task wasn't supposed to run in the first place. I think it's time we removed this re-computation.

### DIFF
--- a/rest/route/host.go
+++ b/rest/route/host.go
@@ -97,10 +97,7 @@ func (h *hostsChangeStatusesHandler) Run(ctx context.Context) gimlet.Responder {
 			err = foundHost.SetStatus(status.Status, user.Id, fmt.Sprintf("changed by %s from API", user.Id))
 		}
 		if err != nil {
-			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
-				StatusCode: http.StatusInternalServerError,
-				Message:    err.Error(),
-			})
+			return gimlet.MakeJSONInternalErrorResponder(err)
 		}
 
 		host := &model.APIHost{}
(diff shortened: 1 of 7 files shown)


rest/route/host.go
```diff
@@ -90,20 +90,17 @@ func (h *hostsChangeStatusesHandler) Run(ctx context.Context) gimlet.Responder {
...
-				})
-			}
+		switch status.Status {
+		case evergreen.HostTerminated:
+			err = errors.WithStack(cloud.TerminateSpawnHost(ctx, evergreen.GetEnvironment(), foundHost, user.Id, "terminated via REST API"))
+		default:
+			err = foundHost.SetStatus(status.Status, user.Id, fmt.Sprintf("changed by %s from API", user.Id))
+		}
+		if err != nil {
+			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
```
Kimchelly, line 100: nit: this could be simplified down to just `gimlet.MakeJSONInternalErrorResponder(err)`.

Kimchelly (approved):
LGTM mod nit.

PR #5691, opened 2022-05-26
EVG-15698: Improve util functions used in canRestart and canSchedule resolvers

minnakt:
[EVG-15698](https://jira.mongodb.org/browse/EVG-15698)

### Description 
This PR makes it so that the util functions `isTaskBlocked`, `isExecutionTask`, and `canRestartTask` return booleans instead of pointers to booleans. 

Additionally, it corrects some faulty logic used in the `canSchedule` resolver by introducing a new util function `canScheduleTask`. 

#### Note about incorrect logic
The logic in the `canSchedule` resolver was not correct. In particular, any execution task in prod ([example](https://spruce.mongodb.com/task/mongodb_mongo_master_enterprise_rhel_80_64_bit_dynamic_all_feature_flags_required_aggregation_timeseries_fuzzer_1_enterprise_956195f40885c1ab7930fcbe975e17e52eddc2d5_22_05_26_11_52_00?execution=0&sortBy=STATUS&sortDir=ASC)) has the Schedule button enabled, which should not be the case. 

Previously this resolver relied on the result from `canRestartTask`, but that didn't make that much sense given that the criteria for restarting and scheduling is different. For example, finished or aborted tasks can be restarted, but neither can be scheduled. I added the new util function `canScheduleTask` so that the returned value from the resolver would be correct.

### Testing 
- Checked on staging that:
  - Blocked tasks, running tasks, undispatched tasks, and execution tasks cannot be restarted. Finished and aborted tasks can be restarted.
  - Blocked tasks, running tasks, finished tasks, aborted tasks, and execution tasks cannot be scheduled. Unscheduled tasks can be scheduled.
- Ran e2e tests on spruce through set-module command (see [here](https://spruce.mongodb.com/version/628fb807850e612a7d51a22b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))
- Wrote tests for `canRestartTask` and `canScheduleTask` in `gql/util_test.go`
- Updated gql tests in `graphql/tests/task`


graphql/util.go
```diff
@@ -427,46 +427,58 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
...
-func canRestartTask(ctx context.Context, at *restModel.APITask) (*bool, error) {
+func canRestartTask(ctx context.Context, at *restModel.APITask) (bool, error) {
 	taskBlocked, err := isTaskBlocked(ctx, at)
 	if err != nil {
-		return nil, err
+		return false, err
 	}
-	canRestart := !utility.StringSliceContains(evergreen.TaskUncompletedStatuses, *at.Status) || at.Aborted || (at.DisplayOnly && *taskBlocked)
-	isExecTask, err := isExecutionTask(ctx, at) // Cant restart execution tasks.
+	canRestart := !utility.StringSliceContains(evergreen.TaskUncompletedStatuses, *at.Status) || at.Aborted || (at.DisplayOnly && taskBlocked)
```
minnakt, line 461: Wanted to point out this existing code because it seems strange; apparently you are allowed to restart a task if it’s a Display Task and it’s blocked? Does that sound right? 🤔

rest/model/task.go
```diff
@@ -470,6 +471,7 @@ func (ad *APITask) ToService() (interface{}, error) {
 	for i, dep := range ad.DependsOn {
 		dependsOn[i].TaskId = dep.TaskId
 		dependsOn[i].Status = dep.Status
+		dependsOn[i].Unattainable = dep.Unattainable
```
ablack12, line 474: Since this is business logic, I hesitate slightly to expose this as a field to rest routes. Any chance of moving this to a resovler?

graphql/util.go
```diff
@@ -427,46 +427,58 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
 	}
 }
 
-func isTaskBlocked(ctx context.Context, at *restModel.APITask) (*bool, error) {
-	t, err := task.FindOneIdNewOrOld(*at.Id)
+func isTaskBlocked(ctx context.Context, at *restModel.APITask) (bool, error) {
```
ablack12, line 430: Both of these helper functions have to convert the API task document, and then just call one function. Is it possible to instead do the conversion in canRestartTask so we only need to do it once? Then we can just call t.Blocked() and t.IsPartOfDisplay() directly.

graphql/util.go
ablack12: I think this makes sense since later tasks in a display task could be blocked on early tasks in the display task, in which case restarting the entire display task may unblock these.

However, I would recommend that you break this line into pieces for readability. Really what this function could be is:
if case1 { return false }
if case2 { return false }
if case3 { return false}
return true 

There's no reason to compute canRestart in the executionTask case for example; it makes sense to just check that first and return early.

ablack12 (changes requested):

rest/model/task.go
minnakt: Ahh sorry I didn’t realize that, I removed it. Because I lose access to the `Unattainable` field but don't want to do an additional db lookup, I now use the `DisplayStatus` to check if the task is blocked instead

graphql/util.go
```diff
@@ -427,46 +427,37 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
 	}
 }
 
-func isTaskBlocked(ctx context.Context, at *restModel.APITask) (*bool, error) {
-	t, err := task.FindOneIdNewOrOld(*at.Id)
-	if err != nil {
-		return nil, ResourceNotFound.Send(ctx, err.Error())
+func canRestartTask(t *task.Task) bool {
+	// Cannot restart execution tasks.
+	isExecTask := t.IsPartOfDisplay()
```
ablack12, line 432: no reason to assign this to a variable since we don't actually use it except in the if

graphql/util.go
```diff
@@ -427,46 +427,37 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
...
-	i, err := at.ToService()
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while converting task %s to service", *at.Id))
+	if t.Aborted {
+		return true
 	}
-	t, ok := i.(*task.Task)
-	if !ok {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Unable to convert APITask %s to Task", *at.Id))
+	if t.DisplayStatus == evergreen.TaskStatusBlocked && t.DisplayOnly {
```
ablack12, line 442: Might help future UI engineers to put a comment here to explain this case

graphql/util.go
```diff
@@ -427,46 +427,37 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
...
+	return false
 }
 
-func canRestartTask(ctx context.Context, at *restModel.APITask) (*bool, error) {
-	taskBlocked, err := isTaskBlocked(ctx, at)
-	if err != nil {
-		return nil, err
+func canScheduleTask(t *task.Task) bool {
+	// Cannot schedule execution tasks.
+	isExecTask := t.IsPartOfDisplay()
```
ablack12, line 450: same comments here  -- or in this case, `if t.IsPartOfDisplay() || t.Aborted` is pretty readable.

There's no rule around this but I think generally boolean logic is readable if it's not more complicated than a "this and/or this"

graphql/util.go
```diff
@@ -427,46 +427,37 @@ func mapHTTPStatusToGqlError(ctx context.Context, httpStatus int, err error) *gq
...
 	}
-	isBlocked := t.Blocked()
-	return &isBlocked, nil
-}
-
-func isExecutionTask(ctx context.Context, at *restModel.APITask) (*bool, error) {
-	i, err := at.ToService()
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error while converting task %s to service", *at.Id))
+	if t.Aborted {
```
ablack12, line 439: I would suggest moving this case to the bottom and just returning !t.Aborted (as opposed to return false) to free up an if statement

ablack12 (changes requested):
Just some nits

graphql/tests/task/queries/canRestart.graphql
```diff
@@ -1,5 +1,5 @@
 {
-  yes: task(taskId: "taskity_task") {
```
ablack12, line 2: I will miss taskity task :')

ablack12 (approved):

PR #5693, opened 2022-05-27
EVG-16679 fix generate tasks retry loop

ybrill:

### Description 
Cherry pick the changes from #5677 that were reverted and fix the problem.
The fix is just the second commit (https://github.com/evergreen-ci/evergreen/commit/78270ed7918252620792cd480953cd29b2cb768c)

### Testing 
Generate tasks [works in staging](https://spruce-staging.corp.mongodb.com/version/6290ff22b237364300b7b73f/tasks). Without the fix this would have also been broken.

### DIFF
--- a/rest/route/task_generate.go
+++ b/rest/route/task_generate.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/evergreen-ci/evergreen/apimodels"
 	"github.com/evergreen-ci/evergreen/db"
+	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/rest/data"
 	"github.com/evergreen-ci/gimlet"
 	"github.com/evergreen-ci/utility"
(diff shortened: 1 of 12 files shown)


rest/route/task_generate.go
```diff
@@ -92,22 +93,28 @@ func (h *generatePollHandler) Parse(ctx context.Context, r *http.Request) error
...
+	var errors []string
+	if len(jobErr) > 0 {
+		errors = append(errors, jobErr)
 	}
+
 	return gimlet.NewJSONResponse(&apimodels.GeneratePollResponse{
 		Finished:   finished,
 		ShouldExit: shouldExit,
-		Errors:     jobErrs,
+		Errors:     errors,
```
ablack12, line 118: It seems worth maybe adding a new field to GeneratePollResponse for Error (as opposed to Errors) and sending/processing both for now and then removing Errors, since it's not really necessary to have an errors list, but could just ticket this

ablack12 (approved):

rest/route/task_generate.go
ybrill: I started us off with the extra field and opened [EVG-16977](https://jira.mongodb.org/browse/EVG-16977) to remove the deprecated one.

PR #5694, opened 2022-05-27
EVG-16147 Add patch route for versions

bynn:
[EVG-16147](https://jira.mongodb.org/browse/EVG-16147)

### Description 
a couple of things: 
added new version route 
modified setversionactivation to not call expensive version status update function 
i think version activated was not being properly set so i also updated that 

### Testing 
unit test


model/build/build.go
```diff
@@ -166,6 +166,31 @@ func UpdateActivation(buildIds []string, active bool, caller string) error {
 	return errors.Wrapf(err, "setting build activation to %t", active)
 }
 
+// UpdateActivationAndStatus updates builds with the given ids
+// to the given activation setting and marks builds as created if active.
+func UpdateActivationAndStatus(buildIds []string, active bool, caller string) error {
```
ablack12, line 171: We aren't actually using this function anywhere it looks like? But also if we intend to only use this in the case that all buildIds are being updated in a version, we could change this to take in the version ID and query on that.

(We could also maybe create a setTaskActivationForVersion and refactor setTaskActivationForBuilds so they can share a lot of code). I would note that SetTaskActivationForBuilds does call  UpdateVersionAndPatchStatusForBuilds, so I'm not sure where the refactor to improve this is that you mentioned in your comment

rest/route/version.go
```diff
@@ -63,6 +64,70 @@ func (vh *versionHandler) Run(ctx context.Context) gimlet.Responder {
...
+	vh.versionId = gimlet.GetVars(r)["version_id"]
+	if vh.versionId == "" {
+		return errors.New("request data incomplete")
+	}
+	return nil
+}
+
+// Run calls the data model.VersionFindOneId function
+func (vh *versionPatchHandler) Run(ctx context.Context) gimlet.Responder {
+	foundVersion, err := dbModel.VersionFindOneId(vh.versionId)
```
ablack12, line 106: Seems like no reason to fetch the version here since SetVersionActivation does this (except for the error code but perhaps that function could return it then, or the function could be refactored to just update using the ID rather than querying and _then_ updating)

ablack12 (changes requested):

model/build/build.go
bynn: i tried to refactor it and i think i try to minimize database queries as much as possible

graphql/util.go
```diff
@@ -684,11 +684,7 @@ func setVersionActivationStatus(version *model.Version) error {
 	if err != nil {
 		return errors.Wrapf(err, "error getting tasks for version %s", version.Id)
 	}
-	if !task.AnyActiveTasks(tasks) {
-		return errors.Wrapf(version.SetNotActivated(), "Error updating version activated status for `%s`", version.Id)
-	} else {
-		return errors.Wrapf(version.SetActivated(), "Error updating version activated status for `%s`", version.Id)
-	}
+	return errors.Wrapf(version.SetActivated(task.AnyActiveTasks(tasks)), "Error updating version activated status for `%s`", version.Id)
```
malikchaya2, line 687: let's instead use just `updating version activated status for `%s`"` bo be consistent with the new error styles :)

rest/route/version.go
```diff
@@ -63,6 +64,59 @@ func (vh *versionHandler) Run(ctx context.Context) gimlet.Responder {
...
+	if vh.Activated == nil {
+		return gimlet.ErrorResponse{
+			Message:    "Must set 'activated'",
+			StatusCode: http.StatusBadRequest,
+		}
+	}
+
+	vh.versionId = gimlet.GetVars(r)["version_id"]
+	if vh.versionId == "" {
+		return errors.New("request data incomplete")
```
malikchaya2, line 99: should we specify that it is missing a versionId?

malikchaya2 (commented):
I will be out on Monday so please consider my review non-blocking.

model/lifecycle.go
```diff
@@ -57,25 +57,85 @@ type VersionToRestart struct {
...
+			return false, errors.Wrap(err, "getting tasks to activate")
+		}
+		if withDependencies {
+			dependOn, err := task.GetRecursiveDependenciesUp(tasksToActivate, nil)
+			if err != nil {
+				return false, errors.Wrap(err, "getting recursive dependencies")
+			}
+			tasksToActivate = append(tasksToActivate, dependOn...)
+		}
+		if len(tasksToActivate) > 0 {
```
ablack12, line 103: I'm wondering if in the !withDependencies case we could not query tasks at all, but just update tasks/builds that match that query.

We could refactor ActivateTasks to take in a query, and for other usages  of ActivateTasks (that _do_ use a task list) make some helper function like `byTaskIds(taskList)` that returns `IdKey: bson.M{"$in": taskIDs}` if that makes sense

model/lifecycle.go
```diff
@@ -57,25 +57,85 @@ type VersionToRestart struct {
...
+				return false, errors.Wrap(err, "updating tasks for activation")
+			}
+			return true, nil
+		}
+	} else {
+		query := bson.M{
+			task.VersionKey: versionId,
+			task.StatusKey:  evergreen.TaskUndispatched,
+		}
+		// if the caller is the default task activator only deactivate tasks that have not been activated by a user
```
ablack12, line 123: nit: comments should be complete sentences

model/lifecycle.go
```diff
@@ -57,25 +57,85 @@ type VersionToRestart struct {
...
+		if evergreen.IsSystemActivator(caller) {
+			query[task.ActivatedByKey] = bson.M{"$in": evergreen.SystemActivators}
+		}
+
+		tasks, err := task.FindAll(db.Query(query).WithFields(task.IdKey, task.ExecutionKey))
```
ablack12, line 128: Similarly, I don't think we need to query the tasks to do this update, if we refactor deactivateTasks to take in a query

rest/route/version.go
malikchaya2: do we want to elaborate on this error message?

model/build/build.go
```diff
@@ -166,6 +166,29 @@ func UpdateActivation(buildIds []string, active bool, caller string) error {
...
+	if !active && evergreen.IsSystemActivator(caller) {
+		query[ActivatedByKey] = bson.M{"$in": evergreen.SystemActivators}
+	}
+	update := bson.M{
+		ActivatedKey:     active,
+		ActivatedTimeKey: time.Now(),
+		ActivatedByKey:   caller,
+	}
+	if active {
+		update[StatusKey] = evergreen.BuildCreated
```
ablack12, line 182: So, this isn't actually true in the case that we're only passing in _some_ buildIds (what if buildA is being activated but buildB is already running)? I recognize that isn't actually a concern since we're only using this function to update builds for an _entire_ version. That being said, should this function instead take in a versionId (rather than buildIds) and use that in the query, to indicate that this should only be used for full versions?

model/lifecycle.go
```diff
@@ -57,25 +57,81 @@ type VersionToRestart struct {
...
+		}
 	}
+	return nil
+}
 
-	// Update activation for all builds before updating their tasks so the version won't spend
-	// time in an intermediate state where only some builds are updated
-	if err = build.UpdateActivation(buildIDs, active, caller); err != nil {
-		return errors.Wrapf(err, "setting activation for builds in version '%s'", versionId)
+func setTaskActivationForVersion(versionId string, active, withDependencies bool, ignoreTasks []string, caller string) (bool, error) {
```
ablack12, line 82: Can we addd a comment to be clear about the boolean being returned?

model/version.go
```diff
@@ -127,34 +127,17 @@ func (v *Version) UpdateBuildVariants() error {
 	)
 }
 
-// SetActivated sets this version to activated if it's not already activated.
-func (v *Version) SetActivated() error {
-	if utility.FromBoolPtr(v.Activated) {
+// SetActivated sets version activated field to specified boolean.
+func (v *Version) SetActivated(activated bool) error {
+	if v.Activated != nil && *v.Activated == activated {
```
ablack12, line 132: Can just use utility.FromBoolPtr

rest/route/version.go
```diff
@@ -63,6 +64,59 @@ func (vh *versionHandler) Run(ctx context.Context) gimlet.Responder {
...
+	}
+	return nil
+}
+
+// Run calls the data model.VersionFindOneId function
+func (vh *versionPatchHandler) Run(ctx context.Context) gimlet.Responder {
+	u := MustHaveUser(ctx)
+	if err := dbModel.SetVersionActivation(vh.versionId, utility.FromBoolPtr(vh.Activated), u.Id); err != nil {
+		state := "inactive"
```
ablack12, line 108: I think this messaging is pretty outdated; I think the error at 112 should just actually return the error

ablack12 (changes requested):

model/lifecycle.go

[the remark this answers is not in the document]
bynn: no it's always false and nil

model/version.go
bynn: done

rest/route/version.go
bynn: sorry i didn't see this before!

rest/route/version.go
```diff
@@ -63,6 +64,59 @@ func (vh *versionHandler) Run(ctx context.Context) gimlet.Responder {
...
+	}
+
+	vh.versionId = gimlet.GetVars(r)["version_id"]
+	if vh.versionId == "" {
+		return errors.New("request data incomplete")
+	}
+	return nil
+}
+
+// Run calls the data model.VersionFindOneId function
```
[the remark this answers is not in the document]
bynn: fixed

rest/route/version.go
bynn: done

malikchaya2 (approved):

model/lifecycle.go
```diff
@@ -57,25 +57,73 @@ type VersionToRestart struct {
...
+			}
+			if err = build.UpdateActivationAndStatus(buildIds, active, caller); err != nil {
+				return false, errors.Wrapf(err, "setting activation for builds in version '%s'", versionId)
+			}
+			if err = task.ActivateTasks(tasksToActivate, time.Now(), false, caller); err != nil {
+				return false, errors.Wrap(err, "updating tasks for activation")
+			}
+			return true, nil
+		}
+	} else {
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
bynn: done 2

bynn (commented):
i did a bit of a refactor so the build status will be accurate

model/lifecycle.go
```diff
@@ -55,25 +55,79 @@ type VersionToRestart struct {
...
+	}
+	changed, err := setTaskActivationForVersion(versionId, active, caller)
 	if err != nil {
-		return errors.Wrapf(err, "getting builds for version '%s'", versionId)
+		return errors.Wrapf(err, "setting activation for tasks in version '%s'", versionId)
 	}
-	buildIDs := make([]string, 0, len(builds))
-	for _, build := range builds {
-		buildIDs = append(buildIDs, build.Id)
+	if active && changed && evergreen.IsFinishedVersionStatus(version.Status) {
```
ablack12, line 72: So given that we're calling updateBuildStatus anyway, I think it does also maybe make reasonable sense to call UpdateVersionStatus and UpdatePatchStatus like we do for updating all builds: https://github.com/evergreen-ci/evergreen/blob/0aa9ce4746a3a31eb1d6cb9280c49e90d39b3efc/model/task_lifecycle.go#L1159 (really, I wonder if we should just call this function rather than calling updateBuildStatus explicitly at 108).

model/lifecycle.go
```diff
@@ -55,25 +55,79 @@ type VersionToRestart struct {
...
+				return false, errors.Wrap(err, "fetching builds")
+			}
+			for _, b := range builds {
+				_, err := updateBuildStatus(&b)
+				if err != nil {
+					return false, errors.Wrapf(err, "updating status for build '%s'", b.Id)
+				}
+			}
+			return true, nil
+		}
```
ablack12, line 114: We should return false, nil here right? Otherwise if len(tasksToActivate) is 0, we go on to deactivate all tasks.

model/lifecycle.go
```diff
@@ -55,25 +55,79 @@ type VersionToRestart struct {
...
 	}
 
-	return errors.Wrapf(setTaskActivationForBuilds(buildIDs, active, false, nil, caller),
-		"setting activation for tasks in version '%s'", versionId)
+	tasks, err := task.FindAll(db.Query(q).WithFields(task.IdKey, task.ExecutionKey))
+	if err != nil {
+		return false, errors.Wrap(err, "getting tasks to deactivate")
+	}
+	if len(tasks) > 0 {
+		if err = task.DeactivateTasks(tasks, false, caller); err != nil {
```
ablack12, line 126: We should also be updating the build and version status in the deactivate case, right?

ablack12 (commented):
Sorry for another comment; just trying to pare down the additional work that was added. I think existing functions already do what's trying to be done here, but more succinctly.

bynn (commented):
the existing v1 route actually calls the SetVersionActivation function which we modified so this also changes the original behavior.

The biggest change between the routes is that the v2 route does not return the info which I think was the problem with the timeout.

model/lifecycle.go
bynn: yes

model/lifecycle.go
bynn: calls UpdateVersionAndPatchStatusForBuilds

model/lifecycle.go
```diff
@@ -55,25 +55,61 @@ type VersionToRestart struct {
...
+// setTaskActivationForVersion activates all tasks in a version.
+func setTaskActivationForVersion(versionId string, active bool, caller string) error {
+	q := bson.M{
+		task.VersionKey: versionId,
+		task.StatusKey:  evergreen.TaskUndispatched,
 	}
+	tasksToModify := []task.Task{}
+	// If activating a task, set the ActivatedBy field to be the caller.
+	if active {
+		tasksToModify, err := task.FindAll(db.Query(q).WithFields(task.IdKey, task.DependsOnKey, task.ExecutionKey, task.BuildIdKey))
```
ablack12, line 76: You need to use = instead of := to set to tasksToModify, otherwise this variable will only exist within the if statement.
https://go.dev/play/p/1F7QWjlBGDC

model/lifecycle.go
```diff
@@ -135,11 +171,6 @@ func setTaskActivationForBuilds(buildIds []string, active, withDependencies bool
 			return errors.Wrap(err, "deactivating tasks")
 		}
 	}
-
-	if err := UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
```
ablack12, line 139: Can you explain why we remove this? Won't this break places that use this?

model/lifecycle.go
bynn: ah yes i didnt realize what it would do with 2 vars 
defined err outside of the if and fixed

model/lifecycle.go
bynn: i meant to copy it and i think i cut it sorry

ablack12 (approved):

PR #5697, opened 2022-05-31
EVG-16761 Remove $lookup from FindUniqueBuildVariantNamesByTask

hadjri:
[EVG-16761](https://jira.mongodb.org/browse/EVG-16761)

### Description 
Added FindUniqueBuildVariantNamesByTask to the task struct, allowing us to directly reference the build variant display name on tasks in the future without having to do a $lookup function for the display name on the build collection. Since this a forward moving change and we're not retroactively assigning variant display name to old tasks, there still needs to be a check for the variant display names of tasks that do not have the new property set.
### Testing 
Updated existing unit tests to retrieve tasks with and without a set BV display name.


model/task/db.go
```diff
@@ -160,6 +160,32 @@ var (
 		},
 	}
 
+	sortDisplayName = bson.M{
```
ablack12, line 163: Can we add comments to explain this whole section?

model/task/db.go
```diff
@@ -985,6 +1015,29 @@ const VersionLimit = 50
 // It attempts to return the most recent display name for each build variant to avoid returning duplicates caused by display names changing.
 // It only checks the last 50 versions that ran for a given task name.
 func FindUniqueBuildVariantNamesByTask(projectId string, taskName string, repoOrderNumber int) ([]*BuildVariantTuple, error) {
+	noLookupPipeline := variantByTaskPipeline(projectId, taskName, repoOrderNumber)
+	lookupPipeline := variantByTaskPipelineWithLookup(projectId, taskName, repoOrderNumber)
+	facet := bson.M{
+		"$facet": bson.M{
```
ablack12, line 1021: I think similar to the other use of facet, we would want to only split for the necessary section (looking up build variant) and then rejoining, similar to the other linked use of facet. This would simplify the code (not having to have these with/without functions) I would think, since the only thing that should be different is the lookup itself.

model/task/db.go
```diff
@@ -160,6 +160,32 @@ var (
...
+		"$sort": bson.M{
+			"display_name": 1,
+		},
+	}
+
+	project = bson.M{
+		"$project": bson.M{
+			"_id":           0,
+			"build_variant": "$" + BuildVariantKey,
+			"display_name":  "$" + BuildVariantDisplayNameKey,
```
ablack12, line 173: I hesitate to call this display_name since it's for the build variant; the naming should reflect that

ablack12 (commented):
Didn't review closely bc time crunch but wanted to put some thoughts in. cc @khelif96 should also review

model/task/db.go
hadjri: I had done this because the two pipelines are performing their operations on different subsets of the task collection: those with a BV display name set, and those without it set. Therefore the $match stages are different as well. However if there is a way to make this work with two different $match stages where one receives a $lookup and the other does not I'm happy to discuss.

model/task/db.go
```diff
@@ -160,6 +160,32 @@ var (
 		},
 	}
 
+	sortDisplayName = bson.M{
+		"$sort": bson.M{
+			"display_name": 1,
```
khelif96, line 165: `DisplayNameKey` assuming this refers to a task display name?

model/task/db.go
```diff
@@ -985,6 +1015,29 @@ const VersionLimit = 50
...
+	facet := bson.M{
+		"$facet": bson.M{
+			"noLookup": noLookupPipeline,
+			"lookup":   lookupPipeline,
+		},
+	}
+	result := []*buildVariantTupleResult{}
+	if err := Aggregate([]bson.M{facet, {"$project": bson.M{
+		"build_variants": bson.M{
+			"$setUnion": []string{"$lookup", "$noLookup"},
```
khelif96, line 1029: can we rename these variables. Its confusing seeing `$lookup` as a parameter since it represents a mongodb aggregation step.

model/task/db.go
```diff
@@ -160,6 +160,32 @@ var (
 		},
 	}
 
+	sortDisplayName = bson.M{
+		"$sort": bson.M{
+			"display_name": 1,
+		},
+	}
+
+	project = bson.M{
```
khelif96, line 169: I think this needs a better name  maybe `projectBuildVariant`? When i read the usage below i was confused about what this did. Especially since this is so far away from its usage.

model/task/db.go
khelif96: `sortByDisplayName`?

model/task/db.go
khelif96: Possibly even move this down so it is colocated with where it's being used.

model/task/db.go
khelif96: I think this setup make sense although it would likely only be needed temporarily. Eventually most tasks will have the build variant display name field and this part of the aggregation wont be used anymore. (We only look back 50 mainline commits for this specific query)

khelif96 (changes requested):
Thanks for making these changes! I left a few comments but it makes sense to me so far.

model/task/db.go
hadjri: This refers to BV display name. Renamed (changed it before requesting review but forgot to actually push it!)

rest/model/task.go
```diff
@@ -388,33 +388,34 @@ func (at *APITask) GetProjectIdentifier() {
...
+		Version:                 utility.FromStringPtr(ad.Version),
+		Revision:                utility.FromStringPtr(ad.Revision),
+		Priority:                ad.Priority,
+		Activated:               ad.Activated,
+		ActivatedBy:             utility.FromStringPtr(ad.ActivatedBy),
+		BuildId:                 utility.FromStringPtr(ad.BuildId),
+		DistroId:                utility.FromStringPtr(ad.DistroId),
+		Container:               utility.FromStringPtr(ad.Container),
+		BuildVariant:            utility.FromStringPtr(ad.BuildVariant),
+		BuildVariantDisplayName: utility.FromStringPtr(ad.BuildVariantDisplayName),
```
ablack12, line 402: Should also probably add to BuildFromService

model/task/db.go
ablack12: Yep that sounds good to me; can we add that ticket as a TODO here as well?

ablack12 (approved):

model/task/db.go
hadjri: Sure, created ticket here as well: [EVG-17040](https://jira.mongodb.org/browse/EVG-17040)

rest/model/task.go
hadjri: It's actually already there

PR #5700, opened 2022-06-01
EVG-17014 use correct task to determine test results type

ablack12:
[EVG-17014](https://jira.mongodb.org/browse/EVG-17014)

### Description 
Right now, we check if we're using cedarResults using the task from the project context, rather than the correct task for that task/execution pair. If we're querying an old execution and the newest execution is running, then the projCtx.Task.CedarResults returns false, which means we query using the legacy method for old tasks (which correctly would've had CedarResults set to true if we used the right document).

Refactored so we'd only ever need to query the task document once.

### Testing 
Tested on staging; https://evergreen-staging.corp.mongodb.com/rest/v2/tasks/evg_ubuntu1604_test_model_manifest_78708dfdbd1f5c4e517a61ebc5734f9b54de1323_22_05_31_21_59_52/tests?execution=0 didn't work after restarting before changes, and did work after restarting after changes.

### DIFF
--- a/rest/data/test.go
+++ b/rest/data/test.go
@@ -39,26 +39,10 @@ func FindTestById(id string) ([]testresult.TestResult, error) {
 }
 
 func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]testresult.TestResult, error) {
-	t, err := task.FindOneIdNewOrOld(opts.TaskID)
-	if err != nil {
-		return []testresult.TestResult{}, gimlet.ErrorResponse{
-			StatusCode: http.StatusNotFound,
-			Message:    errors.Wrapf(err, "finding task '%s'", opts.TaskID).Error(),
-		}
+	taskIDs := []string{opts.TaskID}
+	if len(opts.ExecutionTasks) > 0 {
+		taskIDs = opts.ExecutionTasks
 	}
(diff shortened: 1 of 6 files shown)


rest/data/test.go
```diff
@@ -39,26 +39,10 @@ func FindTestById(id string) ([]testresult.TestResult, error) {
...
 
 func (tc *DBTestConnector) FindTestsByTaskId(opts FindTestsByTaskIdOpts) ([]testresult.TestResult, error) {
-	t, err := task.FindOneIdNewOrOld(opts.TaskID)
-	if err != nil {
-		return []testresult.TestResult{}, gimlet.ErrorResponse{
-			StatusCode: http.StatusNotFound,
-			Message:    errors.Wrapf(err, "finding task '%s'", opts.TaskID).Error(),
-		}
-	}
-	if t == nil {
```
bynn, line 49: tests seem to be failing because we dont check the task exists i think?

ablack12: I'll take a look! FWIW there's no need for reviewers to debug why tests might be failing, that falls on the author. LGTMs always have the caveat that it's for when tests pass. If enough things are failing that it doesn't seem worth reviewing yet you can always wait until the author has fixed them to review (and/or comment something like that)

PR #5703, opened 2022-06-02
EVG-16975 Add revision number cap to getProjectTasksHandler as well as max query time

hadjri:
[EVG-16975](https://jira.mongodb.org/browse/EVG-16975)

### Description 
This route was prone to timeouts for projects with an extensive task history (i.e. server).  Since it is looking for tasks with a given name rather than ID, it walks back to every task in existence for a project. If a name is nonexistent, it won't return an empty list until it checks all tasks which in server's case gets to a timeout before that happens.  

3 changes:

- introduce a max query time on the route so that the request doesn’t continue holding a read ticket on the DB after the route returns a 504 timeout
- add a max number of revisions to look back as an optional parameter so that large projects can return more quickly for nonexistent task names
- updated variable naming from task_id -> task_name because it was not correct. Updated docs as well.
### Testing 
Updated existing unit test.


db/db_utils.go
```diff
@@ -411,3 +412,18 @@ func AggregateWithHint(collection string, pipeline interface{}, hint interface{}
 
 	return errors.WithStack(pipe.All(out))
 }
+
+// AggregateWithMaxTime runs aggregate and specifies a max quert time which ensures the query won't go on indefinitely when the request is cancelled.
+func AggregateWithMaxTime(collection string, pipeline interface{}, out interface{}, maxTime time.Duration) error {
+	session, db, err := GetGlobalSessionFactory().GetSession()
+	if err != nil {
+		err = errors.Wrap(err, "error establishing db connection")
```
malikchaya2, line 420: let's change this to `establishing db connection` to be consistent with the new naming.

rest/route/service.go
```diff
@@ -167,7 +167,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/projects/{project_id}/task_stats").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTaskStats(opts.URL))
 	app.AddRoute("/projects/{project_id}/test_stats").Version(2).Get().Wrap(requireUser, viewTasks, cedarTestStats).RouteHandler(makeGetProjectTestStats(opts.URL))
 	app.AddRoute("/projects/{project_id}/versions").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectVersionsHandler(opts.URL))
-	app.AddRoute("/projects/{project_id}/tasks/{task_id}").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTasksHandler(opts.URL))
+	app.AddRoute("/projects/{project_id}/tasks/{task_name}").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTasksHandler(opts.URL))
```
malikchaya2, line 170: Who is currently using this route? Is it heavily used by users? If yes, we may want to make this change in a backwards compatible way (maybe still set h.taskName with both gimlet.GetVars(r)["task_name"] and gimlet.GetVars(r)["task_id"]

ablack12: I don't think this should effect users, since this is just related to how we parse the URL and it's not something that users set (for example, "/projects/evg/tasks/dist" should work either way )

db/db_utils.go
```diff
@@ -411,3 +412,18 @@ func AggregateWithHint(collection string, pipeline interface{}, hint interface{}
 
 	return errors.WithStack(pipe.All(out))
 }
+
+// AggregateWithMaxTime runs aggregate and specifies a max quert time which ensures the query won't go on indefinitely when the request is cancelled.
+func AggregateWithMaxTime(collection string, pipeline interface{}, out interface{}, maxTime time.Duration) error {
```
ablack12, line 417: Should we refactor the other usages of MaxTime() to use this function?

model/project.go
```diff
@@ -37,7 +37,8 @@ const (
 )
 
 type GetProjectTasksOpts struct {
-	Limit        int    `json:"num_versions"`
+	NumVersions  int    `json:"num_versions"`
```
ablack12, line 40: We can't change limit's json; this will break existing usages.

model/project_ref.go
```diff
@@ -1106,19 +1110,38 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 		task.DisplayNameKey: taskName,
 		task.StatusKey:      bson.M{"$in": finishedStatuses},
 	}
+	var repo *Repository
+	var revisionMatch []bson.M
+	if opts.NumVersions > 0 {
```
ablack12, line 1115: Perhaps rather than allowing this to be user preferenced, we can just start enforcing this with a large number (1000 versions or something).

ablack12 (changes requested):

db/db_utils.go
hadjri: Sounds good!

model/project_ref.go
hadjri: SGTM

model/project.go
hadjri: Reverted and removed NumVersions per the other comment

model/project_ref.go
```diff
@@ -1106,19 +1110,38 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
...
+		repo, err = FindRepository(projectId)
+		if err != nil {
+			return nil, err
+		}
+		if repo == nil {
+			return nil, errors.Errorf("finding repository '%s'", projectId)
+		}
+		revisionMatch = []bson.M{
+			{task.RevisionOrderNumberKey: bson.M{"$gte": repo.RevisionOrderNumber - opts.NumVersions}},
```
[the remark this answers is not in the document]
hadjri: NumVersions has since been removed and after the most recent refactor the revision numbers will be bound starting either at the repo's revision number or `StartAt` if the user defined it. In both cases the query will not look back more than 1000 revisions

db/db_utils.go
```diff
@@ -388,16 +389,23 @@ func Aggregate(collection string, pipeline interface{}, out interface{}) error {
 	// socket timeout, which isn't really an option here. (other
 	// operations had a 90s timeout, which is no longer specified)
 
-	pipe := db.C(collection).Pipe(pipeline)
+	pipe := ConstructPipe(db, collection, pipeline)
+	return unmarshalPipelineResults(pipe, out)
+}
+
+func ConstructPipe(db db.Database, collection string, pipeline interface{}) db.Aggregation {
```
ablack12, line 396: I don't know that I agree with this being a helper function, since it's just one line and it's easily readable, and is only used in one place

model/task_history.go
```diff
@@ -336,7 +336,7 @@ func (iter *taskHistoryIterator) GetChunk(v *Version, numBefore, numAfter int, i
 		{"$group": groupStage},
 		{"$sort": bson.M{task.RevisionOrderNumberKey: -1}},
 	}
-	agg := database.C(task.Collection).Pipe(pipeline).MaxTime(taskHistoryMaxTime)
+	agg := db.ConstructPipeWithMaxTime(database, task.Collection, pipeline, taskHistoryMaxTime)
 	var aggregatedTasks []bson.M
 	if err = agg.All(&aggregatedTasks); err != nil {
```
ablack12, line 341: I meant rather that here we should call AggregateWithMaxTime (which prevents us from needing ConstructPipeWithMaxTime)

model/task_history.go
```diff
@@ -359,46 +359,44 @@ func (self *taskHistoryIterator) GetDistinctTestNames(ctx context.Context, numCo
...
+				},
+			},
+			task.IdKey:                 1,
+			task.TestResultTestFileKey: 1,
+		}},
+		{"$unwind": fmt.Sprintf("$%v", testResultsKey)},
+		{"$group": bson.M{"_id": fmt.Sprintf("$%v.%v", testResultsKey, task.TestResultTestFileKey)}},
+	}
 
+	agg := db.ConstructPipeWithMaxTime(mdb, task.Collection, pipeline, taskHistoryMaxTime)
```
ablack12, line 396: I think here it's okay to not change the function since we use .Iter instead of .All (I'm not sure how long this will exist anyway)

model/task_history.go
ablack12: I'm maybe confused, but we could just replace 339 through 341 with AggregateWithMaxTime, right? Basically I'm advocating not to have ConstructPipeWithMaxTime since I don't think it's really a helpful helper function.

hadjri: Ok, if we go with that route then before line 346 we will need to reconstruct the same aggregation `agg := database.C(task.Collection).Pipe(pipeline).MaxTime(taskHistoryMaxTime)`. That's why I thought it might be redundant to call `AggregateWithMaxTime` in this function but I can just do that and revert all the helper creations.

hadjri: This solution above broke some unit tests and had me scratching my head for a bit but eventually I got this to work by having `AggregateWithMaxTime ` return both an aggregation as well as an error.  Tests broke because `GetFailedTests ` expects the `Cursor` on the aggregation to be non-nil when it is passed in, which is not the case if we just define a whole new aggregation. It needs to receive the Aggregation in a state after `.All()` has been called on it.

db/db_utils.go
```diff
@@ -411,3 +412,19 @@ func AggregateWithHint(collection string, pipeline interface{}, hint interface{}
 
 	return errors.WithStack(pipe.All(out))
 }
+
+// AggregateWithMaxTime runs aggregate and specifies a max query time which ensures the query won't go on indefinitely when the request is cancelled.
```
ablack12, line 416: Can we add something to explain why we return the agg?

ablack12 (approved):
LGTM! Thanks for the back and forth

PR #5705, opened 2022-06-06
EVG-16877: handle container instance interrupt

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16877

### Description 
This adds support . The general behavior should be something like this:
* AWS tells Evergreen via SNS that the container instance is going to be spot instance interrupted (i.e. drained).
* Find the pods that will are about to be stopped due to the interrupt and enqueue the job to terminate them.
* When terminating them, fix any task lifecycle issues caused by abrupt termination. If the task is already running in the pod, mark it as system failed and restart it. If there are any tasks still waiting to run on this pod and it's the only one left that can run them, mark those tasks as needing another container allocated to them.

### Testing 
Added unit tests.

I couldn't test this in staging with AWS because we can't actually run container tasks yet and we can't simulate draining containers until [BUILD-14978](https://jira.mongodb.org/browse/BUILD-14978) is done.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -126,8 +126,8 @@ github.com/aws/aws-sdk-go v1.15.11/go.mod h1:mFuSZ37Z9YOHbQEwBWztmVzqXrEkub65tZo
 github.com/aws/aws-sdk-go v1.41.11/go.mod h1:585smgzpB/KqRA+K3y/NL/oYRqQvpNJYvLm+LY1U59Q=
 github.com/aws/aws-sdk-go v1.41.12/go.mod h1:585smgzpB/KqRA+K3y/NL/oYRqQvpNJYvLm+LY1U59Q=
 github.com/aws/aws-sdk-go v1.42.44/go.mod h1:OGr6lGMAKGlG9CVrYnWYDKIyb829c6EVBRjxqjmPepc=
-github.com/aws/aws-sdk-go v1.44.4 h1:ePN0CVJMdiz2vYUcJH96eyxRrtKGSDMgyhP6rah2OgE=
-github.com/aws/aws-sdk-go v1.44.4/go.mod h1:y4AeaBuwd2Lk+GepC1E9v0qOiTws0MIWAX4oIKwKHZo=
+github.com/aws/aws-sdk-go v1.44.25 h1:cJZ4gtEpWAD/StO9GGOAyv6AaAoZ9OJUhu96gF9qaio=
+github.com/aws/aws-sdk-go v1.44.25/go.mod h1:y4AeaBuwd2Lk+GepC1E9v0qOiTws0MIWAX4oIKwKHZo=
 github.com/aybabtme/iocontrol v0.0.0-20150809002002-ad15bcfc95a0 h1:0NmehRCgyk5rljDQLKUO+cRJCnduDyn11+zGZIc9Z48=
 github.com/aybabtme/iocontrol v0.0.0-20150809002002-ad15bcfc95a0/go.mod h1:6L7zgvqo0idzI7IO8de6ZC051AfXb5ipkIJ7bIA2tGA=
 github.com/benbjohnson/clock v1.1.0 h1:Q92kusRqC1XV2MjkWETPvjJVqKetz1OzxZB7mHJLju8=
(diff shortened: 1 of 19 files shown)


go.sum
```diff
@@ -126,8 +126,8 @@ github.com/aws/aws-sdk-go v1.15.11/go.mod h1:mFuSZ37Z9YOHbQEwBWztmVzqXrEkub65tZo
 github.com/aws/aws-sdk-go v1.41.11/go.mod h1:585smgzpB/KqRA+K3y/NL/oYRqQvpNJYvLm+LY1U59Q=
 github.com/aws/aws-sdk-go v1.41.12/go.mod h1:585smgzpB/KqRA+K3y/NL/oYRqQvpNJYvLm+LY1U59Q=
 github.com/aws/aws-sdk-go v1.42.44/go.mod h1:OGr6lGMAKGlG9CVrYnWYDKIyb829c6EVBRjxqjmPepc=
-github.com/aws/aws-sdk-go v1.44.4 h1:ePN0CVJMdiz2vYUcJH96eyxRrtKGSDMgyhP6rah2OgE=
```
hadjri, line 129: Looks like a few merge conflicts exist for these updates

model/task/task.go
```diff
@@ -1109,10 +1111,24 @@ func (t *Task) MarkAsContainerDeallocated(ctx context.Context, env evergreen.Env
 	}
 
 	t.ContainerAllocated = false
-	t.DispatchTime = utility.ZeroTime
```
hadjri, line 1112: What's the reason we don't need to set these props on the task anymore?

model/task_lifecycle.go
```diff
@@ -1393,7 +1396,40 @@ func doRestartFailedTasks(tasks []string, user string, results RestartResults) R
...
 }
 
-func ClearAndResetStrandedTask(h *host.Host) error {
+// ClearAndResetStrandedContainerTask clears the container task dispatched to a
+// pod. It also resets the task so that the current task execution is marked as
+// finished and, if necessary, a new execution is created to restart the task.
+// TODO (PM-2618): should probably block single host task groups once they're
+// supported.
+func ClearAndResetStrandedContainerTask(p *pod.Pod) error {
+	runningTaskID := p.RunningTask
```
hadjri, line 1405: Since we never reassign it maybe it's best to just directly use `p.RunningTask` instead of declaring a new variable

model/task_lifecycle.go
```diff
@@ -1393,7 +1396,40 @@ func doRestartFailedTasks(tasks []string, user string, results RestartResults) R
...
+	if runningTaskID == "" {
```
hadjri, line 1406: I see some of the same patterns here as `ClearAndResetStrandedHostTask`. Might be worth some consolidation of the two functions and their logic. I'm not particularly opinionated on that though, up to you

model/task_lifecycle.go
```diff
@@ -1424,26 +1460,36 @@ func ClearAndResetStrandedTask(h *host.Host) error {
...
 	}
 
-	if err = t.MarkSystemFailed(evergreen.TaskDescriptionStranded); err != nil {
-		return errors.Wrap(err, "marking task failed")
+	if err := t.MarkSystemFailed(evergreen.TaskDescriptionStranded); err != nil {
+		return errors.Wrap(err, "marking task as system failed")
 	}
 
 	if time.Since(t.ActivatedTime) > task.UnschedulableThreshold {
+		// If the task has already exceeded the unschedulable threshold, we
```
hadjri, line 1480: Appreciate this comment

rest/route/sns.go
```diff
@@ -391,17 +404,63 @@ func (sns *ecsSNS) handleNotification(ctx context.Context, notification ecsEvent
...
+			return errors.Wrapf(err, "listing ECS tasks associated with container instance '%s'", detail.ContainerInstanceARN)
+		}
+		for _, taskARN := range taskARNs {
+			p, err := pod.FindOneByExternalID(taskARN)
+			if err != nil {
+				return errors.Wrapf(err, "finding pod with external ID '%s'", taskARN)
+			}
+			if p == nil {
+				grip.Error(message.Fields{
+					"message": "Found an ECS task running in a draining container instance, but it's not associated with any known pod." +
```
hadjri, line 445: Is it worth having an alert for this case?

hadjri (changes requested):

model/task/task.go
Kimchelly: I originally modeled some of this logic based on [MarkAsHostUndispatched](https://github.com/evergreen-ci/evergreen/blob/5cd7c47bca7225acdc715a514f67a03a911b3621/model/task/task.go#L1028-L1051) but changed my mind about the necessity of some of the fields which are specific to dispatching the task. Allocating/deallocating a container is not really equivalent to dispatching/undispatching a task because dispatch is the moment that the task is assigned to run on a host/container, whereas container allocation is when the task is given a container that will eventually run it. Since they're separate parts of the task lifecycle, it doesn't make sense to unset them here.

model/task_lifecycle.go
Kimchelly: `(*pod.Pod).ClearRunningTask` unsets the running task but we still need it to find the task itself.

model/task_lifecycle.go
Kimchelly: I'm not sure how to consolidate this without doing some weird muxing of host/pod docs together into a helper function. I think the loss of clarity by making a small empty-string-check helper that works for both struct types wouldn't be worth the consolidation.

rest/route/sns.go
Kimchelly: I think it's worth being aware if/when it happens, because it's a case that breaks the assumption that Evergreen is properly tracking is cloud resources. Whenever we have a state mismatch between Evergreen and the cloud provider, we have to keep an eye on such things because they generally represent bugs, races, or leaked resources.

My best guess is that this can hypothetically be triggered in a race condition when first creating the pod. I don't have a good idea of how often it will actually occur in practice, but if it's common, we'll have to add more complex logic to deal with the race.

rest/route/sns.go
```diff
@@ -499,12 +557,64 @@ func (sns *ecsSNS) cleanupUnrecognizedPod(ctx context.Context, details ecsEventD
...
+		}
+
+		for _, arn := range resp.TaskArns {
+			if arn == nil {
+				continue
+			}
+			taskARNs = append(taskARNs, utility.FromStringPtr(arn))
+		}
+
+		token = resp.NextToken
```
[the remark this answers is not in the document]
Kimchelly: Nah, this is a completely reasonable question, and I did think about this when first writing this loop. The supposed behavior of the ECS API is that it will only return you a next token when there are actually more result page beyond the one you just asked for, so there's no hypothetical limit to how many times we could receive a next token. We won't actually infinitely loop here because the `context.Context` (which comes from the request) will eventually a timeout and fail the `ListTasks` request because the app server has a hard 60 second limit on request duration (to protect itself from unreasonably expensive requests).

To be extra safe, we can cap the infinite loop to a fixed number of since there shouldn't be an infinite number of containers running on a single container instance. I changed the loop to cap the max iterations and avoid the potential for infinite looping.

go.sum
Kimchelly: Yeah, they're merge conflicts against my own commits. I merged main.

model/task_lifecycle.go
hadjri: Ok, SGTM

hadjri (approved):
Thanks for the responses!

PR #5706, opened 2022-06-07
EVG-16810 Fix dependency task Ids being empty 16810

malikchaya2:

### Description 
We encountered some tasks with dependencies that don't exist, including tasks with empty taskIds. I was unable to reproduce in staging. I switched it to log an error and continue instead of returning. I also added some logging to try to capture more information on when empty dependency task ids are found. I'm hoping that by understanding when the empty ids occur, it'll also shed some light on the problem in general the next time it occurs. 

### Testing 
 existing unit tests


malikchaya2:
@ablack12 This PR is ready for re-review. I started from a fresh branch so the changes are new. I also updated the PR description.

model/lifecycle.go
```diff
@@ -969,6 +969,14 @@ func makeDeps(t BuildVariantTaskUnit, thisTask *task.Task, taskIds TaskIdTable)
 			if id == thisTask.Id {
 				continue
 			}
+			if id == "" {
```
ablack12, line 972: I'm actually thinking that this can't be the place where blank ids show up, since all of the above functions check "". Is there anywhere else we add dependencies? (In the DB, do we have tasks that actually have blank IDs? or else i wonder if there's a logic bug that makes it seem like there are tasks that do)

model/scheduler_stats.go
```diff
@@ -139,6 +140,14 @@ func hasTaskId(taskId string, dependsOn []task.Dependency) bool {
 		if d.TaskId == taskId {
 			return true
 		}
+		if d.TaskId == "" {
+			grip.Debug(message.Fields{
```
ablack12, line 144: This log message wouldn't really be helpful since it doesn't tell us the task that has the bad state.

model/task/task.go
```diff
@@ -709,7 +718,11 @@ func (t *Task) DependenciesMet(depCaches map[string]Task) (bool, error) {
 				return false, errors.Wrap(err, "finding dependency")
 			}
 			if foundTask == nil {
-				return false, errors.Errorf("dependency '%s' not found", dependency.TaskId)
+				grip.Error(message.Fields{
+					"message": "dependency not found",
+					"task_id": dependency.TaskId,
```
ablack12, line 723: Can we also log t.Id? And maybe the function name to distinguish this

units/commit_queue.go
```diff
@@ -220,6 +220,14 @@ func (j *commitQueueJob) addMergeTaskDependencies(cq commitqueue.CommitQueue) er
 		}
 		err = mergeTask.AddDependency(dependency)
 		if err != nil {
+			if dependency.TaskId == "" {
+				grip.Debug(message.Fields{
```
ablack12, line 224: same comment here

model/task_lifecycle.go
```diff
@@ -768,6 +768,15 @@ func removeNextMergeTaskDependency(cq commitqueue.CommitQueue, currentIssue stri
 			Status: AllStatuses,
 		}
 		if err = nextMerge.AddDependency(d); err != nil {
+			if d.TaskId == "" {
```
ablack12, line 771: I'm not sure it really makes sense to log this here; it doesn't seem like AddDependency errors specifically for the empty ID case

model/task/task.go
```diff
@@ -650,6 +650,15 @@ func (t *Task) AddDependency(d Dependency) error {
 		}
 	}
 	t.DependsOn = append(t.DependsOn, d)
+
+	if d.TaskId == "" {
+		grip.Debug(message.Fields{
```
ablack12, line 655: we can use grip.DebugWhen(d.TaskId == "")

model/task/task.go
```diff
@@ -650,6 +650,15 @@ func (t *Task) AddDependency(d Dependency) error {
...
+			"ticket":     "EVG-16810",
```
ablack12, line 656: Unless we intend to keep this ticket open, it's actually possibly not helpful to include this here

ablack12 (changes requested):

model/lifecycle.go
malikchaya2: that's a good point: the example task we are looking at doesn't have any empty taskIds in the DB 
image

PR #5710, opened 2022-06-10
EVG-17026: set capacity provider based on container OS/arch

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17026

### Description 
Instead of setting a placement constraint for the architecture, specify the capacity provider that will run the pod in an instance with the desired OS/arch.

### Testing 
Updated the existing unit test. We can't verify that this in an e2e way until [BUILD-15127](https://jira.mongodb.org/browse/BUILD-15127) fixes the staging AWS IAM role structure (because it prevents us from making ECS/Secrets Manager API calls), but setting the capacity provider worked as expected when I did the load test.

### DIFF
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -247,21 +247,7 @@ func exportECSPodContainerDef(settings *evergreen.Settings, p *pod.Pod) (*cocoa.
 	return def, nil
 }
 
-// podArchToECSArch exports a pod container CPU architecture to an ECS
-// CPU architecture.
-func exportECSPodArch(arch pod.Arch) string {
-	switch arch {
-	case pod.ArchAMD64:
-		return "x86_64"
-	case pod.ArchARM64:
-		return "arm64"
-	default:
-		return ""
-	}
-}
-
 const (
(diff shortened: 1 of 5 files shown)


cloud/pod_util.go
```diff
@@ -281,19 +267,27 @@ func exportECSPodExecutionOptions(ecsConfig evergreen.ECSConfig, containerOpts p
...
 	}
-	if arch := exportECSPodArch(containerOpts.Arch); arch != "" {
-		archConstraint := fmt.Sprintf("%s == %s", ecsCPUArchConstraint, arch)
-		placementOpts.AddInstanceFilters(archConstraint)
-	}
 	opts.SetPlacementOptions(*placementOpts)
 
+	var foundCapacityProvider bool
+	for _, cp := range ecsConfig.CapacityProviders {
+		if containerOpts.OS.Matches(cp.OS) && containerOpts.Arch.Matches(cp.Arch) {
```
hadjri, line 274: If you only plan to use these two `Matches` in conjunction with one another you could consider merging them into one function and using a bool function on this line.

hadjri (approved):

PR #5711, opened 2022-06-10
EVG-16009 Remove BuildNumber and global build number tracking

ZackarySantana:
[EVG-16009](https://jira.mongodb.org/browse/EVG-16009)

### Description 
BuildNumber was used to populate a legacy REST v1 route that the new API does not use. It has been removed from the REST v1 route, and should be removed from our models and DB build model.

Remove the global folder and buildnumber.

### Testing 
Removed tests and ran evergreen patch with affected tests


bynn (approved):

cmd/generate-lint/generate-lint.go
```diff
@@ -34,7 +34,7 @@ func whatChanged() ([]string, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "problem getting merge-base")
 	}
-	diffCmd := exec.Command("git", "diff", strings.TrimSpace(string(base)), "--name-only")
+	diffCmd := exec.Command("git", "diff", strings.TrimSpace(string(base)), "--name-only", "--diff-filter=d")
```
ablack12, line 37: what is this change for?

ablack12 (changes requested):

cmd/generate-lint/generate-lint.go
ablack12: jk did a google, good find

ablack12 (approved):

PR #5716, opened 2022-06-13
EVG-15622 Take activated as starting time for builds that have restarted once or more

hadjri:
[EVG-15622](https://jira.mongodb.org/browse/EVG-15622)

### Description 
Builds have an issue where the time taken is derived as (finish time -  start time).  However start time does not update with each restart of a build, hence some builds that have multiple restarts over a few days have far too high `time_taken_ms`.

Change has been made such that if a build has been restarted, then calculate time taken as (finish time - activated time) since activated time keeps up with each restart. Otherwise we still calculate time taken as (finish time - start time).
### Testing 
Updated existing unit test. Added cases


Kimchelly (commented):
Is there a reason start time can't be reset? Seems like we would want a separate start time each time the build is restarted.

PR #5717, opened 2022-06-13
EVG-16736: Split up GraphQL resolvers file

minnakt:
[EVG-16736](https://jira.mongodb.org/browse/EVG-16736)

### Description 
The `resolvers.go` file is over 4000 lines long and as a result it can be hard to parse or find things within it. This PR makes it so that the resolver definitions and functions are distributed among smaller, more digestible files by (1) grouping schema files and (2) modifying `gqlgen.yml`.

### Notes
- I didn't end up using the `follow-schema` command for `gqlgen.yml` because I didn't really like how it placed some limitations on the file structure. Here's my [branch](https://github.com/minnakt/evergreen/tree/EVG-16736_state2) where I tried out this format.  I put the root `Query` and `Mutation` definition in `base.graphql` (because I didn't want to put it in some arbitrary graphql file) but this would always generate a `base.resolvers.go` file which felt unnecessary given that a `resolver.go` file is always generated too.
- I moved a bunch of stuff into folders to prevent import cycles. The main reason I had to do this is because I wanted to keep all of the resolver files in a `/resolvers` folder instead of jumbled up in the general `/graphql` folder.
- The file grouping decisions were made according to my own interpretation. I'm happy to move/adjust things if it makes more sense to other people.

### Testing 
- Existing GQL tests pass
- On staging


gqlgen.yml
```diff
@@ -4,59 +4,227 @@
 # for detailed .gqlgen.yml documentation.
 
 schema:
-  - graphql/schema.graphql
```
malikchaya2, line 7: (the placement of this comment is random, I just wanted threading to be available) 

These changes are great! I love the organization! 

I am wondering a bit about how this changes the workflow. Before these changes I would add something to the schema, delete the resolvers file, run `make gqlgen` and then copy the function headers that were generated. How does that change now? Would it be possible to document the new flow somewhere, maybe in the dev guide?

ablack12 (changes requested):

graphql/model/models_gen.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by github.com/99designs/gqlgen, DO NOT EDIT.
```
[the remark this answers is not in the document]
minnakt: The [config](https://gqlgen.com/config/) docs seem to recommend that they go in separate packages but I will investigate!

gqlgen.yml
minnakt: yes good point! 😅 I'm not familiar with that workflow (I typically just write the function header from scratch) but you could do the same thing I think!

It would be something like:
- You add whatever field / query / mutation to the corresponding graphql file in schema
- Delete the `resolvers.go` file
- Run make `gqlgen`
- It will generate a `resolvers.go` file with quite literally EVERYTHING (it's not smart enough to partition into correct files, so you'll get a bunch of duplicate definition errors)
- Grab the function header you want
- Restore the original `resolvers.go` file
- Paste the function header into the file you want it to go (so if its a project-related function, put it in `resolvers/project.go`)

graphql/resolvers/util/ui_plugin_build_baron.go
```diff
@@ -1,12 +1,12 @@
-package graphql
+package util
```
khelif96, line 1: I don't think we need this package. By defining the util package we collide with the existing `util` package. 
This function is only used by Graphql we don't need to place it in a separate package and expose it as a public function.

graphql/schema.graphql
```diff
@@ -1,3 +1,4 @@
+# THIS FILE IS BEING KEPT FOR BACKWARDS COMPATABILITY.
```
khelif96, line 1: I'm confused by what this means?

graphql/http_handler.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/graphql/handler"
 	"github.com/99designs/gqlgen/graphql/handler/apollotracing"
+	generated "github.com/evergreen-ci/evergreen/graphql/generated"
```
khelif96, line 12: I don't think we need a generated package it is only used by graphql why do we need a separate package to store these values?

graphql/errors/errors.go
```diff
@@ -1,4 +1,4 @@
-package graphql
+package errors
```
khelif96, line 1: I don't think we need a separate errors package. These functions are only used by graphql and should just share the same package name.

khelif96 (changes requested):
Thanks for doing this. 
Although the functions are grouped by the "logical value they belong to" it disconnects from the graph aspect of graphql. 
It is difficult to track down what file a sub resolver is in because it requires context on what "group" the resolver belongs to.
Why is this `projectSettings` resolver in the `projects` file why not a `projectSettings` folder? 
Why is the `UpdatePublicKey` resolver in the user file; why not in the hosts directory? 
This structure requires an implementer to consider the use case of the individual resolver to locate where it belongs or decide where to place it. 

When i proposed this ticket i envisioned maintaining the "graphql graph grouping". Everything that is a `queryResolver` lives in a `queryResolver` file, a `mutationResolver` lives in the `mutationResolver` File, `taskResolver` in `taskResolver` etc. This way you maintain the native grouping we have defined in our `schema`

gqlgen.yml
malikchaya2: ah, that makes sense. thanks!

graphql/schema.graphql
minnakt: I deleted this file, I thought it would break Spruce frontend but it seems fine on staging

graphql/http_handler.go
minnakt: if the `/resolvers` folder is kept (which I'm not sure it is based on your previous comment), then moving this out will cause an import cycle (because the files outside of `/resolvers` import `/resolvers` but then `/resolvers` needs to import the files outside of itself)

graphql/host_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/evergreen-ci/evergreen/model/host"
+	restModel "github.com/evergreen-ci/evergreen/rest/model"
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	werrors "github.com/pkg/errors"
```
ablack12, line 16: Does this need to be werrors if we aren't usin ga different errors package?

graphql/mutation_resolver.go
```diff
@@ -0,0 +1,1133 @@
+package graphql
```
ablack12, line 1: @khelif96 do we foresee ever wanting to split mututations up into more files? (Could be a future ticket to think about the best way to do it but does seem like this file is also quite large)

graphql/patch_resolver.go
```diff
@@ -0,0 +1,294 @@
...
+	"github.com/evergreen-ci/evergreen/model"
+	"github.com/evergreen-ci/evergreen/model/build"
+	"github.com/evergreen-ci/evergreen/model/commitqueue"
+	"github.com/evergreen-ci/evergreen/model/patch"
+	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/evergreen/model/user"
+	"github.com/evergreen-ci/evergreen/rest/data"
+	restModel "github.com/evergreen-ci/evergreen/rest/model"
+	"github.com/evergreen-ci/utility"
+	werrors "github.com/pkg/errors"
```
ablack12, line 23: same question

graphql/mutation_resolver.go
khelif96: mutations are tougher to split up since they all stem from the mutation root. and are harder to associate with parent resolvers. [Federated Graphql ftw](https://www.apollographql.com/docs/federation/) But for now this might be the best ew can do.

graphql/host_resolver.go
minnakt: I think it's an issue with `gqlgen`. 😭 If I don't give it a specific name, it'll auto-import the `"errors"` package instead of `"github.com/pkg/errors"` package every time I run `make gqlgen`. 

There's actually an open issue about this in `99designs/gqlgen` repo and I left a comment in `resolver.go` linking to it. (I don't think I can leave a comment in the more specific resolver files because they get overwritten every time I generate)

graphql/schema/types/version.graphql
```diff
@@ -0,0 +1,170 @@
...
+type BuildVariantTuple {
+  buildVariant: String!
+  displayName: String!
+}
+
+"""
+CommitQueue is returned by the commitQueue query.
+It contains information about the patches on the commit queue (e.g. author, code changes) for a given project.
+"""
+type CommitQueue {
```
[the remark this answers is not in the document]
minnakt: you got it boss 👍

graphql/host_resolver.go
ablack12: Another option is we could switch to using fmt since we just use .Error anyway (like instead of `werrors.Wrapf(err, "Problem getting last known project for '%s'", alias.ChildProject).Error())` we could do `fmt.Sprintf("Problem getting last known project for '%s': %s", alias.ChildProject, err.Error())`
But whatever works!

khelif96 (approved):
Great job on this! Just had a few open questions.

Does the graphql playground still work? Will it still autocomplete queries now that the schema is still split up.

How does spruce dev work now that the schemas are split up? 
Do we need to symlink all of the files? 
Do the utilities around graphql query linting still work with the split up files?

minnakt:
> Does the graphql playground still work? Will it still autocomplete queries now that the schema is still split up.

Yes! I tested on staging below. You can see the autocomplete and also one of the doc comments I added in the Docs side panel.
Screen Shot 2022-06-29 at 1 40 58 PM

> Do we need to symlink all of the files? 

Yes, the instructions are in my Spruce PR! Rather than symlinking each of the files we can just symlink the entire folder (achieving the same effect).
* evergreen-ci/spruce/pull/1321

> Do the utilities around graphql query linting still work with the split up files?

I assume you mean linting in Spruce (I've never noticed GQL linting in the Evergreen repo, though I could be wrong). If so then yes, that's also handled by my Spruce PR. 

Screen Shot 2022-06-29 at 2 48 25 PM

PR #5719, opened 2022-06-14
EVG-6005 Set is_patch to explicitly an empty string for mainline commits

ZackarySantana:
[EVG-6005](https://jira.mongodb.org/browse/EVG-6005)

### Description 
When running expansions.write, users should be able to see is_patch as an empty string. Before is_patch is set to true, when it is a patch, but now it will be set to an empty string explicitly in all other cases. That way users can see it in expansions.write.

### Testing 
There was existing expansions tests that expected is_patch to not exist, and for the total amount of expansions default-ly available to be a certain amount. Updating the tests to expect an empty is_patch and +1 to the amount of expansions properly tests the new implementation.


malikchaya2 (approved):
Congratulations on getting your first (second?) ticket to code review 🎉  

Since the work in this PR has diverged from what the user initially had in mind, I think it would be a good idea to set the title to something closer to "Set is_patch to the empty string for mainline commits" or something like that. (Also feel free to update the new hire guide to communicate this concept as  well if you feel called to it :)

ablack12:
Agreed -- PR titles are meant to describe the change made, which doesn't always align with the ticket title

ZackarySantana:
Got it! Thank you for the feedback, I will set the title to something closer like that

ZackarySantana:
The lastest commit was intended for another branch, so I pushed it back one commit. Also thank you guys again for the feedback on the PR details

PR #5720, opened 2022-06-14
EVG-16879: allow task container allocation to retry up to a limit

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16879

### Description 
Each task gets its own counter that tracks the number of times Evergreen has attempted to allocate a container to run it. If the pod fails for whatever reason (e.g. the container cannot run the agent), it'll try re-allocating containers until it hits the limit. Once it hits the limit, Evergreen should system-fail the task, but I'm going to save that for a later PR ([EVG-16880](https://jira.mongodb.org/browse/EVG-16880)).

### Testing 
Added unit tests.


model/task_lifecycle.go
```diff
@@ -1641,6 +1645,9 @@ func checkResetSingleHostTaskGroup(t *task.Task, caller string) error {
 	return nil
 }
 
+// checkResetSingleHostTaskGroup attempts to reset all tasks that are under the
```
hadjri, line 1648: should be `checkResetDisplayTask`

hadjri (approved):

model/task_lifecycle.go
Kimchelly: Good catch, fixed.

PR #5721, opened 2022-06-15
EVG-16453 filter superuser results from mana routes

ablack12:
[EVG-16453 ](https://jira.mongodb.org/browse/EVG-16453 )

### Description 
Right now Evergreen users have permissions on every project and we expose these permissions to Mana (for example,  https://mana.corp.mongodbgov.com/resources/61a3e469f29b988eeefcf846), but really these are internally designated and shouldn't be modifiable by project admins, and may confuse them, so we should filter them out. 
(Have confirmed this with the MANA team)

Added a "all" option for Evergreen engineers who want to use this route instead of querying the DB (i.e. me).

### Testing 
Tested locally.


globals.go
```diff
@@ -1072,13 +1072,20 @@ var SuperuserPermissions = []string{
 }
 
 const (
-	BasicProjectAccessRole = "basic_project_access"
-	BasicDistroAccessRole  = "basic_distro_access"
+	BasicProjectAccessRole     = "basic_project_access"
+	BasicDistroAccessRole      = "basic_distro_access"
+	SuperUserRole              = "superuser"
+	SuperUserProjectAccessRole = "admin_project_access"
```
malikchaya2, line 1078: why is this ""admin_project_access" instead of "superuser_project_access"?

malikchaya2 (changes requested):

globals.go
ablack12: That's just what it is in the backend already and it would require a migration to change it. I wanted to make the variable name more clear since project_admins are something else.

malikchaya2 (approved):

PR #5722, opened 2022-06-15
EVG-17073 Redirect to new waterfall if hitting evergreen.mongodb.com

khelif96:
[EVG-17073](https://jira.mongodb.org/browse/EVG-17073)

### Description 
Visiting evergreen.mongodb.com will now redirect you to the spruce commits page if you have spruce enabled. Any other routes will still take you to the legacy pages

https://user-images.githubusercontent.com/4605522/173864988-eda65544-860b-42f0-ad9a-46112889aabf.mov

.


service/waterfall.go
```diff
@@ -637,6 +637,20 @@ func (uis *UIServer) waterfallPage(w http.ResponseWriter, r *http.Request) {
 	}{newUILink, uis.Settings.Jira.Host, uis.GetCommonViewData(w, r, false, true)}, "base", "waterfall.html", "base_angular.html", "menu.html")
 }
 
+// Create and return a redirect to the spruce waterfall page. If the user is opted in to the new UI,
+func (uis *UIServer) waterfallRedirect(w http.ResponseWriter, r *http.Request) {
```
ablack12, line 641: Should we call this mainlineCommitsRedirect or something since we don't call the new page the waterfall?

service/waterfall.go
```diff
@@ -637,6 +637,20 @@ func (uis *UIServer) waterfallPage(w http.ResponseWriter, r *http.Request) {
 	}{newUILink, uis.Settings.Jira.Host, uis.GetCommonViewData(w, r, false, true)}, "base", "waterfall.html", "base_angular.html", "menu.html")
 }
 
+// Create and return a redirect to the spruce waterfall page. If the user is opted in to the new UI,
```
ablack12, line 640: this comment ends in a comma?

bsamek:
@khelif96 One thing I hadn't thought of when we discussed this is that this potentially makes it confusing for external users of Evergreen to see it. I'll add a comment in the ticket.

service/waterfall.go
```diff
@@ -637,6 +638,23 @@ func (uis *UIServer) waterfallPage(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// Create and return a redirect to the spruce waterfall page if the user is opted in to the new UI.
+func (uis *UIServer) mainlineCommitsRedirect(w http.ResponseWriter, r *http.Request) {
+	u := gimlet.GetUser(r.Context())
+	if u != nil {
+		user, ok := u.(*user.DBUser)
+		if ok {
+			if user.Settings.UseSpruceOptions.SpruceV1 {
+				evergreen.GetEnvironment().Settings().Ui.UIv2Url = ""
```
bsamek, line 648: What's the reason for setting this?

service/waterfall.go
```diff
@@ -637,6 +638,23 @@ func (uis *UIServer) waterfallPage(w http.ResponseWriter, r *http.Request) {
...
+				http.Redirect(w, r, fmt.Sprintf("%s/commits/", uis.Settings.Ui.UIv2Url), http.StatusSeeOther)
+			}
+		}
+	}
+
+	http.Redirect(w, r, "/waterfall", http.StatusSeeOther)
```
bsamek, line 654: This redirects `/` to `/waterfall`. How does `/waterfall` redirect to spruce?

bsamek (changes requested):

service/waterfall.go
khelif96: I don't remember setting this it must have been a bad copy:pasta when i was cleaning up the code to make the pr. It didn't have an effect which is why i didn't notice it.

service/waterfall.go
bsamek: Oh sorry, I misread, resolving.

khelif96: I added a comment to make it more clear. This is essentially a fallback. If there is no user in the request/ or they do not have spruce enabled. We fallback and redirect to "/waterfall"

bsamek (approved):
lgtm with a question

ablack12 (approved):

service/waterfall.go
```diff
@@ -637,6 +638,23 @@ func (uis *UIServer) waterfallPage(w http.ResponseWriter, r *http.Request) {
...
 }
 
+// Create and return a redirect to the spruce waterfall page if the user is opted in to the new UI.
+func (uis *UIServer) mainlineCommitsRedirect(w http.ResponseWriter, r *http.Request) {
+	u := gimlet.GetUser(r.Context())
+	if u != nil {
+		user, ok := u.(*user.DBUser)
+		if ok {
+			// If the user is opted in to the new UI, redirect to the new UI.
+			if user.Settings.UseSpruceOptions.SpruceV1 {
```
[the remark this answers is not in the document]
khelif96: i think the 2  conditionals above should guard against this value being nil.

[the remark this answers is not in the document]
bsamek: Ah, I worried that some field would be a pointer. None is. LGTM.

PR #5723, opened 2022-06-15
EVG-15871 Added description output to the CLI and description to the backend API

ZackarySantana:
[EVG-15871](https://jira.mongodb.org/browse/EVG-15871)

### Description 
Users have external descriptions of their aliases and instead they should be able to view descriptions set to aliases. The backend was modified to accept descriptions in to the struct (defaulting to an empty string) and in the CLI, to output this description.

### Testing 
As of now, a user cannot create a patch alias with a description. To test, a description was added to the database inside the alias record in the staging database. Then running the command using the staging configuration, you are able to see the description put in. As well, the description shows up nicely on the rest API.

(Please excuse the commits, I had to cherry-pick them from my branches and I'm still getting used to handling them!)

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -22,6 +22,7 @@ var (
 	gitTagKey      = bsonutil.MustHaveTag(ProjectAlias{}, "GitTag")
 	remotePathKey  = bsonutil.MustHaveTag(ProjectAlias{}, "RemotePath")
 	variantKey     = bsonutil.MustHaveTag(ProjectAlias{}, "Variant")
+	descriptionKey = bsonutil.MustHaveTag(ProjectAlias{}, "Description")
 	taskKey        = bsonutil.MustHaveTag(ProjectAlias{}, "Task")
 	variantTagsKey = bsonutil.MustHaveTag(ProjectAlias{}, "VariantTags")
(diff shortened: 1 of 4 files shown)


model/project_aliases.go
```diff
@@ -22,6 +22,7 @@ var (
 	gitTagKey      = bsonutil.MustHaveTag(ProjectAlias{}, "GitTag")
 	remotePathKey  = bsonutil.MustHaveTag(ProjectAlias{}, "RemotePath")
 	variantKey     = bsonutil.MustHaveTag(ProjectAlias{}, "Variant")
+	descriptionKey = bsonutil.MustHaveTag(ProjectAlias{}, "Description")
```
malikchaya2, line 25: Can you please add a description of what it looks like when printing it in the cli?

malikchaya2: sorry, I meant screenshot

ZackarySantana: "bynn was here" is the description added to the output of the patch command. The above is the change using my own tarball, the below is using native evergreen cli (showing un-updated versions won't be affected by the description change).

malikchaya2: awesome, thank you

malikchaya2 (approved):

PR #5724, opened 2022-06-15
EVG-16783 remove reliance on model interface and update task functions

ablack12:
[EVG-16783](https://jira.mongodb.org/browse/EVG-16783)

### Description 
We only need the model interface for places that rely on not needing to know what the type is to call ToService/BuildFromService. As far as I can tell, the only place we do this is [here,](https://github.com/evergreen-ci/evergreen/blob/572396deda7b77b9f46d0d24553a164889cc0173/rest/model/admin.go#L230) i.e. only the admin configs need to implement this interface. Otherwise, it's far simpler to just return the struct correctly, as noted in this new comment: https://github.com/evergreen-ci/evergreen/blob/cfe0acc1c2b1f0d55e903ea8c0bbba7dc875b534/rest/model/model.go#L6-L9

Breaking this PR up so this is a POC to remove dependence on model.Model and fix the task struct, and a future PR will be made to fix other structs if this is gucci. 

### Testing 
Making sure tests pass and testing notifications in staging to make sure  they aren't broken by this change.


malikchaya2 (approved):
this is great, thank you for doing this!

PR #5725, opened 2022-06-15
EVG-16635: Update EC2 spot instance creation to handle UnfulfillableCapacity Errors

johndaniels:
[EVG-16635](https://jira.mongodb.org/browse/EVG-16635)

### Description 
This change updates our spot instance creation creation fallback logic in order to handle UnfulfillableCapacity errors that we sometimes get back from spot instance creation requests. It also refactors the code to pull the fallback logic into the EC2 cloud manager, so that we don't mix levels of abstraction in provisioning_create_host.go. 

### Testing 
Primarily I added the ability to return errors to our awsClientMock, so that when I pulled the code into our ec2 cloud manager, we could verify that we made the appropriate calls when an error was returned from our AWS client.

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -690,8 +690,8 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 				"distro":        h.Distro.Id,
 			}))
 			if h.ShouldFallbackToOnDemand() && (strings.Contains(err.Error(), EC2InsufficientCapacity) || strings.Contains(err.Error(), EC2UnfulfillableCapacity)) {
-				// Spot creation failed due to an InsufficientCapacity error.
-				// Try to spawn the host as OnDemand instead of Spot
+				// Spot creation fails relatively frequently due to InsufficientCapacity or UnfulfillableCapacity,
+				// so we fall back to OnDemand if requested since that is more consistently reliable.
 				event.LogHostFallback(h.Id)
 				grip.Info(message.Fields{
 					"message":  "could not spawn spot instance, falling back to on-demand",
@@ -709,7 +709,7 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 						"host_provider": h.Distro.Provider,
 						"distro":        h.Distro.Id,
 					}))
-					return nil, errors.Wrapf(err, "fallback to on-demand failed for host '%s'", h.Id)
+					return nil, errors.Wrapf(err, "spawning on-demand host '%s'", h.Id)
 				}
 				grip.Debug(message.Fields{
 					"message":       "spawned on-demand host",
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -13,6 +14,7 @@ import (
 	"github.com/evergreen-ci/birch"
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/model/distro"
+	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/rest/model"
@@ -687,6 +689,36 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 				"host_provider": h.Distro.Provider,
 				"distro":        h.Distro.Id,
 			}))
+			if strings.Contains(err.Error(), EC2InsufficientCapacity) && h.ShouldFallbackToOnDemand() {
+				// Spot creation failed due to an InsufficientCapacity error.
+				// Try to spawn the host as OnDemand instead of Spot
+				event.LogHostFallback(h.Id)
+				grip.Info(message.Fields{
+					"message":  "fallback to on-demand",
+					"host_id":  h.Id,
+					"host_tag": h.Tag,
+					"distro":   h.Distro.Id,
+					"provider": h.Provider,
+				})
+				h.Provider = evergreen.ProviderNameEc2OnDemand
+				h.Distro.Provider = evergreen.ProviderNameEc2OnDemand
+				if err = m.spawnOnDemandHost(ctx, h, ec2Settings, blockDevices); err != nil {
+					msg := "error spawning on-demand host"
+					grip.Error(message.WrapError(err, message.Fields{
+						"message":       msg,
+						"host_id":       h.Id,
+						"host_provider": h.Distro.Provider,
+						"distro":        h.Distro.Id,
+					}))
+					return nil, errors.Wrapf(err, "error falling back to on-demand for host '%s'", h.Id)
+				}
+				grip.Debug(message.Fields{
+					"message":       "spawned on-demand host",
+					"host_id":       h.Id,
+					"host_provider": h.Distro.Provider,
+					"distro":        h.Distro.Id,
+				})
+			}
 			return nil, errors.Wrap(err, msg)
 		}
 		grip.Debug(message.Fields{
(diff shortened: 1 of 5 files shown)


cloud/ec2.go
```diff
@@ -687,6 +689,37 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 				"host_provider": h.Distro.Provider,
 				"distro":        h.Distro.Id,
 			}))
+			if h.ShouldFallbackToOnDemand() && (strings.Contains(err.Error(), EC2InsufficientCapacity) || strings.Contains(err.Error(), EC2UnfulfillableCapacity)) {
+				// Spot creation failed due to an InsufficientCapacity error.
+				// Try to spawn the host as OnDemand instead of Spot
+				event.LogHostFallback(h.Id)
+				grip.Info(message.Fields{
+					"message":  "fallback to on-demand",
```
julianedwards, line 697: Nit: `could not spawn spot instance, falling back to on-demand`

cloud/ec2.go
```diff
@@ -687,6 +689,37 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 				"host_provider": h.Distro.Provider,
 				"distro":        h.Distro.Id,
 			}))
+			if h.ShouldFallbackToOnDemand() && (strings.Contains(err.Error(), EC2InsufficientCapacity) || strings.Contains(err.Error(), EC2UnfulfillableCapacity)) {
```
julianedwards, line 692: I would just double check that there isn't a case where we want to spawn a host but ignore `ShouldFallbackToOnDemand()` regardless if it returns true. @ybrill would know better than I do.

julianedwards (changes requested):
Looks great, nice refactor!

cloud/ec2.go
ybrill: I don't think so.

cloud/ec2.go
```diff
@@ -687,6 +689,37 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
...
+				h.Distro.Provider = evergreen.ProviderNameEc2OnDemand
+				if err = m.spawnOnDemandHost(ctx, h, ec2Settings, blockDevices); err != nil {
+					msg := "error spawning on-demand host"
+					grip.Error(message.WrapError(err, message.Fields{
+						"message":       msg,
+						"host_id":       h.Id,
+						"host_provider": h.Distro.Provider,
+						"distro":        h.Distro.Id,
+					}))
+					return nil, errors.Wrapf(err, "error falling back to on-demand for host '%s'", h.Id)
```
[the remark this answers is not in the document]
johndaniels: Actually, question here: Would a log line like:

`ERROR: falling back to on demand host`

not be confusing?  One might think that there was an error that caused us to fall back to the on demand host. I'll see if I can just make it clearer wording.

cloud/ec2_test.go
```diff
@@ -481,6 +482,140 @@ func (s *EC2Suite) TestSpawnHostClassicSpot() {
...
+	s.Require().Len(runInput.BlockDeviceMappings, 1)
+	s.Equal("virtual", *runInput.BlockDeviceMappings[0].VirtualName)
+	s.Equal("device", *runInput.BlockDeviceMappings[0].DeviceName)
+	s.Equal("sg-123456", *runInput.SecurityGroups[0])
+	s.Nil(runInput.SecurityGroupIds)
+	s.Nil(runInput.SubnetId)
+	s.Equal(base64OfSomeUserData, *runInput.UserData)
+}
+
+// Test that the ec2 spot instance cloud manager falls back to on demand when it receives
```
julianedwards, line 551: Remove this documentation as well.

cloud/ec2.go
```diff
@@ -687,6 +689,36 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
 				"host_provider": h.Distro.Provider,
 				"distro":        h.Distro.Id,
 			}))
+			if h.ShouldFallbackToOnDemand() && (strings.Contains(err.Error(), EC2InsufficientCapacity) || strings.Contains(err.Error(), EC2UnfulfillableCapacity)) {
+				// Spot creation failed due to an InsufficientCapacity error.
+				// Try to spawn the host as OnDemand instead of Spot
```
julianedwards, line 694: I would remove this comment. It is unnecessary since the code is quite clear in what it is doing.

cloud/ec2.go
```diff
@@ -687,6 +689,36 @@ func (m *ec2Manager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host, e
...
+				h.Provider = evergreen.ProviderNameEc2OnDemand
+				h.Distro.Provider = evergreen.ProviderNameEc2OnDemand
+				if err = m.spawnOnDemandHost(ctx, h, ec2Settings, blockDevices); err != nil {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message":       "spawning on-demand host",
+						"host_id":       h.Id,
+						"host_provider": h.Distro.Provider,
+						"distro":        h.Distro.Id,
+					}))
+					return nil, errors.Wrapf(err, "fallback to on-demand failed for host '%s'", h.Id)
```
julianedwards, line 712: Starting a new thread since the last one is now out-dated. 

On second thought, it is more appropriate for this message to just say `spawning on-demand host '%s'` since we have already logged that we are falling back to on-demand for this specific host. The error message convention is to simply use the gerund of the failed action; previously we would say `problem spawning host...` or `error spawning host..` but found that since these are already error messages it was redundant to say `problem` or `error`. This redundancy is exacerbated since we often wrap errors which leads to a chain of error messages starting from the bottom of the call stack. You will notice some inconsistency in error messages and overall style, especially in older code. Kim recently made a huge effort to make our error messages uniform across all of our code bases but there still may be some inconsistency in Evergreens since it is so big.

julianedwards (changes requested):

cloud/ec2.go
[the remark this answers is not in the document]
julianedwards: Updated comments in a newer thread.

cloud/ec2.go
johndaniels: That's helpful context about the wrapping.

cloud/ec2.go
johndaniels: Hmm... maybe it'd be helpful if I rephrased it? I feel like it would help future developers to know that we are doing this because Spot requests fail with these errors and that falling back to OnDemand fixes it, which I think is less clear from the code (and the current comment as is). I could also leave that to the ticket trail on git blame though. Not sure what the standard is.

cloud/ec2_test.go
johndaniels: Sorry about that.

cloud/ec2.go
johndaniels: Ok I gave it a stab that rephrases it more along the "why" dimension, but happy to remove it if we prefer to keep that as a rule to tickets and git blame.

cloud/ec2.go
johndaniels: Fixed

cloud/ec2_test.go
johndaniels: Fixed.

cloud/ec2_test.go
```diff
@@ -481,6 +482,138 @@ func (s *EC2Suite) TestSpawnHostClassicSpot() {
...
+
+	// Ensure that the mock for requesting a spot instance returns an InsufficientCapacity error.
+	mock.requestSpotInstancesError = awserr.New(EC2UnfulfillableCapacity, "Test error for Unfulfillable Capacity", nil)
+
+	returnedHost, err := s.spotManager.SpawnHost(ctx, h)
+	s.NoError(err)
+	s.Equal(evergreen.ProviderNameEc2OnDemand, returnedHost.Provider, "Provider should have been changed to OnDemand on fallback")
+	s.Equal(evergreen.ProviderNameEc2OnDemand, returnedHost.Distro.Provider, "Provider should have been changed to OnDemand on fallback")
+
+	// Check that we have made a request for a spot instance
```
julianedwards, line 589: Nit: just add periods to these comments.

julianedwards (approved):
LGTM % nit

ablack12 (approved):

PR #5726, opened 2022-06-16
EVG-16979 Add restarted failed functionality to display tasks

bynn:
[EVG-16979](https://jira.mongodb.org/browse/EVG-16979)

### Description 
simply adds a restartFailed field to tasks and only restarts failed tasks if it is set to true 
slight refactor on resetTask to minimize querying

### Testing 
unit test


model/task_lifecycle.go
```diff
@@ -1451,28 +1451,23 @@ func ClearAndResetStrandedTask(h *host.Host) error {
 		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &t.Details, false))
 	}
 
+	// Only restart failed tasks for stranded tasks.
+	t.SetRestartFailed()
```
bynn, line 1455: should only restart stranded but this will restart other tasks that have failed 
I think this behavior is preferable to the current one and can be amended in the future ticket to make stranded tasks work but I can remove this if need be

model/task_lifecycle.go
```diff
@@ -1451,28 +1451,23 @@ func ClearAndResetStrandedTask(h *host.Host) error {
...
-	if taskToReset.IsPartOfDisplay() { // if given an execution task, attempt to restart the full display task
-		dt, err := taskToReset.GetDisplayTask()
-		if err != nil {
-			return errors.Wrap(err, "getting display task")
-		}
-		if dt != nil {
-			taskToReset = *dt
-		}
+	if t.IsPartOfDisplay() {
+		return errors.New("cannot restart individual execution task")
```
bynn, line 1464: im not sure if this is a non-backwards compatible change but I don't think we should allow anyone to call this function on a execution task

model/task_lifecycle.go
```diff
@@ -303,7 +278,28 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 		return errors.Wrap(checkResetSingleHostTaskGroup(t, caller), "resetting single host task group")
 	}
 
-	return errors.WithStack(resetTask(t.Id, caller, false))
+	return errors.WithStack(resetTask(t, caller, false))
+}
+
+// resetTask attempts to archive the task, resets it, and resets the TaskCache in the build as well.
+func resetTask(t *task.Task, caller string, logIDs bool) error {
```
bynn, line 285: moved this function down for readability because it's called right above this

model/task/task.go
```diff
@@ -1934,6 +1935,7 @@ func resetTaskUpdate(t *Task) bson.M {
 		t.Details = apimodels.TaskEndDetail{}
 		t.HasCedarResults = false
 		t.ResetWhenFinished = false
+		t.RestartFailed = nil
```
hadjri, line 1938: Can we init to false instead of nil? Similar to `ResetWhenFinished`

model/task/task.go
```diff
@@ -198,6 +198,7 @@ type Task struct {
 	DisplayOnly       bool     `bson:"display_only,omitempty" json:"display_only,omitempty"`
 	ExecutionTasks    []string `bson:"execution_tasks,omitempty" json:"execution_tasks,omitempty"`
 	ResetWhenFinished bool     `bson:"reset_when_finished,omitempty" json:"reset_when_finished,omitempty"`
+	RestartFailed     *bool    `bson:"restart_failed,omitempty" json:"restart_failed,omitempty"`
```
hadjri, line 201: Does this need to be a pointer?

model/task/db.go
```diff
@@ -95,6 +95,7 @@ var (
 	GenerateTasksErrorKey       = bsonutil.MustHaveTag(Task{}, "GenerateTasksError")
 	GeneratedTasksToActivateKey = bsonutil.MustHaveTag(Task{}, "GeneratedTasksToActivate")
 	ResetWhenFinishedKey        = bsonutil.MustHaveTag(Task{}, "ResetWhenFinished")
+	RestartFailedKey            = bsonutil.MustHaveTag(Task{}, "RestartFailed")
```
ablack12, line 98: For consistency would it make sense to call this ResetFailedWhenFinished so it better matches ResetWhenFinished?

model/task/task.go
ablack12: same question^

model/task_lifecycle.go
ablack12, line 285: I think the reason this function gets the task again is to make sure it's getting the full task document, whereas the task we might be passing in could have only partial fields, meaning we'd do a partial archive (for example, with dependencies in SetActiveState)

model/task_lifecycle.go
```diff
@@ -1245,6 +1241,10 @@ func MarkOneTaskReset(t *task.Task, logIDs bool) error {
 			if err != nil {
 				return errors.Wrap(err, "retrieving execution task")
 			}
+			// Ignore succeeded execution tasks if we only want to restart failed.
+			if utility.FromBoolPtr(t.RestartFailed) && !evergreen.IsFailedTaskStatus(execTask.Status) {
```
ablack12, line 1245: Should also check ResetWhenFinished isn't set though; if the user did ask to restart the whole thing then it would make sense to do it all at once

model/task_lifecycle.go
```diff
@@ -1451,28 +1451,25 @@ func ClearAndResetStrandedTask(h *host.Host) error {
 		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &t.Details, false))
 	}
 
+	// Only restart failed tasks for stranded tasks.
+	if err := t.SetRestartFailed(); err != nil {
```
ablack12, line 1455: This isn't right since in ResetTaskOrDisplayTask we still call SetResetWhenFinished (that's the case where we should be setting failed i think)

model/task_lifecycle.go
ablack12: You'll need to trace where this function is called to verify an execution task can't be passed in

ablack12 (changes requested):

model/task_lifecycle.go
bynn: i think you were talking about the RestartFailed field? (which has been renamed to ResetFailedWhenFinished) so I started marking them groups

model/task_lifecycle.go
bynn: ResetTaskOrDisplayTask is only called in the reset route and here so I created a new ticket to add it to the route

model/task_lifecycle.go
bynn: the only place it's called is stranded tasks and the route. im worried about is the reset task route and I'm wondering if any users will complain if we disallow it now

but I'm convinced that it was only written like this to start with to handle stranded tasks and removing it should be fine

hadjri (approved):
LGTM mod merge conflict resolution and unit test fix.

model/task_lifecycle.go
```diff
@@ -1240,13 +1240,21 @@ func MarkHostTaskDispatched(t *task.Task, h *host.Host) error {
...
 func MarkOneTaskReset(t *task.Task, logIDs bool) error {
 	if t.DisplayOnly {
-		for _, et := range t.ExecutionTasks {
-			execTask, err := task.FindOneId(et)
+		if !t.ResetFailedWhenFinished {
+			if err := MarkTasksReset(t.ExecutionTasks); err != nil {
+				return errors.Wrap(err, "resetting execution tasks")
+			}
+		} else {
+			failedExecTasks, err := task.Find(task.FailedTasksByIds(t.ExecutionTasks))
```
ablack12, line 1248: Can this use withFields(Ids) since that's all we need?

model/task_lifecycle.go
```diff
@@ -1451,28 +1459,27 @@ func ClearAndResetStrandedTask(h *host.Host) error {
...
+			if err := t.SetResetFailedWhenFinished(); err != nil {
+				return errors.Wrap(err, "marking display task for reset")
+			}
+		} else {
+			if err := t.SetResetWhenFinished(); err != nil {
+				return errors.Wrap(err, "marking display task for reset")
+			}
 		}
-		return errors.Wrap(checkResetDisplayTask(&taskToReset), "checking display task reset")
+		return errors.Wrap(checkResetDisplayTask(t), "checking display task reset")
```
ablack12, line 1482: Is modifying checkResetDisplayTask also future work?

ablack12 (approved):

PR #5728, opened 2022-06-17
EVG-16832 Make JIRA BF ticket endpoint grab ticket by taskID and execution

ZackarySantana:
[EVG-16832](https://jira.mongodb.org/browse/EVG-16832)

### Description 
When JIRA BF tickets are generated through the UI, an endpoint is called that goes through the process of generating the JIRA ticket and posting it. Then the corresponding ticket would always be generated with information and links that are for the newest execution. This was because when finding the ticket, evergreen would search for tasks based on only taskID- thus getting the newest entry by default in MongoDB. Replacing the ticket query with one that finds one by taskID and execution solves the issue of incorrectly generated BF tickets.

One thing to note is that links are still generated to the old UI, but have a redirection in the URL causing it to go to the spruce UI. When transferring from the old UI to the spruce UI, the execution number is not preserved.

### Testing 
Generated fake BF and ensured they had information relating to the corresponding execution (host, build status, links- by coping them and removing the redirection parameter in the URL).

### DIFF
--- a/rest/data/build_baron.go
+++ b/rest/data/build_baron.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 
 	"github.com/evergreen-ci/evergreen"
-	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/notification"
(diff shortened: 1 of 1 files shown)


rest/data/build_baron.go
```diff
@@ -29,7 +28,7 @@ type FailingTaskData struct {
 // BbFileTicket creates a JIRA ticket for a task with the given test failures.
```
malikchaya2, line 28: Can you please add a link to the fake BF that you generated when testing this?

malikchaya2: thanks! and did you generate it by clicking the file ticket button on the first execution of the task?

ZackarySantana: Yes, I generated one with the latest execution, one with the first execution
I did this with a couple of different tests as well and checked the URL to make sure the execution is the same

malikchaya2 (approved):

PR #5714, opened 2022-06-13
EVG-16650 add grip logging for host events

ybrill:
[EVG-16650](https://jira.mongodb.org/browse/EVG-16650)

### Description 
@ablack12 and I discussed logging each of the host events to Splunk to aid debugging. I plan to TTL almost all events in the event_log collection, so it will be good to take advantage of Splunk's longer (6 month) persistence. I also think/hope having the events in Splunk, together with all the rest of the data there, will be better for debugging than mentally joining the information.

I thought about removing many of the host events altogether, but I'm unsure if/how users use host events now and wanted to preserve their experience as much as possible. I did remove one, `HOST_TASK_PID_SET` which we no longer log (as of https://github.com/evergreen-ci/evergreen/commit/9c9ad23589805c8a7c8c6b327f014b341156d4a3). I'll open a parallel PR in Spruce to remove it there too.

PR #5730, opened 2022-06-17
EVG-16650 new collection

ybrill:
[EVG-16650](https://jira.mongodb.org/browse/EVG-16650)

### Description 
This PR will _duplicate_ new events so they're stored in both the existing `event_logs` collection and a new `events` collection. New events are saved with an `expirable` field if they're supposed to be subject to the TTL. The TTL index I will eventually build on the new collection will be a partial index with events that aren't expirable filtered out. The plan is to leave this in place until the TTL (90 days?) has elapsed and then to drop the old collection.

Also attempt to keep the old and new collections in sync on subsequent writes, but it's possible a write would succeed on one and not the other. I don't think it's worth adding a transaction, though.

I won't merge this PR until I've created the new collection in prod with the appropriate indexes.

I included some refactoring of references to `AllLogsCollection`. Once upon a time we had another collection for event logs specifically for tasks (feadbe527). Now that that's been removed (c03ac242a) we can consolidate to always assume the same collection name. Canonicalizing the collection name makes it easier to switch the collection we're writing to (this PR) and reading from (later PR after the TTL elapses).

I also fixed a hack where the init functions in the event package relied on [the inits being called in lexical order](https://stackoverflow.com/a/32829593). So the events from files with names before `event_registry.go` (e.g. `admin_event.go`) had their types added within the `admin_event.go`'s init function and everything that came after it was in the correct place. I had the registry initialize itself before the first writes.

Finally, I hope to remove [this host event update](https://github.com/evergreen-ci/evergreen/blob/e9df1cd071ee76f791172ff1dfb3ae83a5c24129/model/event/host_event.go#L261) in a future PR by storing the task execution on the host document. Once all hosts running tasks have an execution I'll be able to remove it. In the meantime, the strategy is unchanged: they're originally set to 0 and updated when the task is archived.

There's some risk duplicating writes will be too much write pressure on the db. We should deploy this carefully.

### Testing 
Events in staging appear as expected. A project event does not have the expirable field while a host event does. Events are duplicated to the new collection.

```javascript
{
    _id: '62ac8dfe97b1d36b1abf0e02',
    r_type: 'PROJECT',
    processed_at: ISODate("2015-10-21T23:29:01.000Z"),
    ts: ISODate("2022-06-17T14:21:50.170Z"),
    expirable: false,
    ...
}
```

and 

```javascript
{
    _id: '62ac8e1397b1d36b1abf0e04',
    r_type: 'HOST',
    processed_at: ISODate("2015-10-21T23:29:01.000Z"),
    ts: ISODate("2022-06-17T14:22:11.995Z"),
    expirable: true,
    r_id: 'evg-ubuntu1804-small-20220617142211-5324909175015625306',
    e_type: 'HOST_CREATED',
    data: { prov_method: '', usr: '' }
},
```


ybrill:
Some back-of-the-envelope math to 1) make sure we don't run out of space by duplicating the collection, 2) estimate the eventual size of the TTLed collection:

The current collection contains 1,735,960,258 documents. Over the last 21 days we wrote 106,432,880 documents to the collection. So the last 21 days are 6.13% of the collection. Extrapolating a bit, 90 days would be ~25% of the collection. 
Given the collection takes up roughly a TB between the data and indexes the new collection will probably grow to use ~250GB  on disk over the next 90 days before the TTL kicks in. Since we have another 2.5TB available this should be safe.

model/event/event_registry.go
```diff
@@ -87,38 +77,36 @@ func (r *eventRegistry) IsSubscribable(resourceType, eventType string) bool {
...
 
-	f, ok := registry.types[resourceType]
-	if !ok {
-		return nil
+	if r.neverExpires == nil {
+		r.neverExpires = make(map[EventLogEntry]bool)
 	}
-
-	return f()
+	r.neverExpires[EventLogEntry{ResourceType: resourceType, EventType: eventType}] = true
```
ablack12, line 87: Do things with neverExpire set not get TTLed?

model/event/event_registry.go
```diff
@@ -87,38 +77,36 @@ func (r *eventRegistry) IsSubscribable(resourceType, eventType string) bool {
 	return r.isSubscribable[e]
 }
 
-func NewEventFromType(resourceType string) interface{} {
-	registry.lock.RLock()
-	defer registry.lock.RUnlock()
+func (r *eventRegistry) setNeverExpire(resourceType, eventType string) {
```
ablack12, line 80: nit: should be neverExpires to match variable name, although I'm wondering if it would be reasonable to switch to `unexpirable` to match the language we use for hosts elsewhere in the codebase.

model/event/event.go
```diff
@@ -74,6 +82,7 @@ func (e *EventLogEntry) SetBSON(raw mgobson.Raw) error {
 	}
 
 	// IDs for events were ObjectIDs previously, so we need to do this
+	// TODO (EVG-16650): Remove once old events are TTLed and/or migrated.
```
ablack12, line 85: Might make sense to split a new ticket to handle this work instead of keeping EVG-16650 open, since i imagine it'll be a few months

model/event/event_registry.go
```diff
@@ -87,38 +77,36 @@ func (r *eventRegistry) IsSubscribable(resourceType, eventType string) bool {
...
+	r.lock.Lock()
```
ablack12, line 81: What's the thought behind changing this lock type?

ablack12 (changes requested):

ablack12 (commented):
> The plan is to leave this in place until the TTL (90 days?) has elapsed and then to drop the old collection.

I would ask that we have a longer TTL, maybe 6 months, at least for hosts/tasks maybe (if it's possible to do it conditionally), since these are helpful for debugging.

model/event/event_registry.go
ybrill: I changed it to unexpirable

model/event/event.go
ybrill: done ([EVG-17214](https://jira.mongodb.org/browse/EVG-17214))

ybrill:
Thank you for reviewing this really long PR.

wrt the TTL as it pertains to debugging with those specific `r_types`: 
- task events: I'm pretty sure task events already have a 90 day TTL now. I didn't confirm that every event type makes it to Splunk, but it seems like it's been working.
- host events: I think debugging host events should be covered up to six months back since #5714 duplicated every host event to Splunk. 

Another consideration is that it would be good to have a TTL on the shorter side, at least initially, to unblock dropping the new collection since it's one of the prerequisites for moving to Atlas.

PR #5733, opened 2022-06-17
EVG-8072 Added computed field time_taken_ms generated field

ZackarySantana:
[EVG-8072](https://jira.mongodb.org/browse/EVG-8072)

### Description 
When querying a build#, the task cache will return a field labeled "time_taken" in nanoseconds when sent as a resp. To not break possible code that depends on this, a new entry "time_taken_ms" that is in milliseconds was added in the response (API struct for task cache).

### Testing 
Querying a build# via the rest API and seeing a new entry "time_taken_ms" with proper numbers.

### DIFF
--- a/rest/model/build.go
+++ b/rest/model/build.go
@@ -120,7 +120,7 @@ func (apiBuild *APIBuild) SetTaskCache(tasks []task.Task) {
 			StatusDetails: t.Details,
 			StartTime:     ToTimePtr(t.StartTime),
 			TimeTaken:     t.TimeTaken,
-			TimeTakenMs:   t.TimeTaken / 1000,
+			TimeTakenMs:   APIDuration(t.TimeTaken.Nanoseconds() / 1000),
 			Activated:     t.Activated,
 		})
 		apiBuild.StatusCounts.IncrementStatus(t.Status, t.Details)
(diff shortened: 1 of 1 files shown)


malikchaya2 (approved):

rest/model/build.go
```diff
@@ -120,6 +120,7 @@ func (apiBuild *APIBuild) SetTaskCache(tasks []task.Task) {
 			StatusDetails: t.Details,
 			StartTime:     ToTimePtr(t.StartTime),
 			TimeTaken:     t.TimeTaken,
+			TimeTakenMs:   APIDuration(t.TimeTaken.Nanoseconds() / 1000),
```
malikchaya2, line 123: Are we sure that this is the right conversion?  
We also might want to use a function to do this. If we don't yet have one, we should probably create a nanoSecondes to Ms function.

malikchaya2 (changes requested):

rest/model/build.go
malikchaya2: I think the conversion from nanoseconds to milliseconds would be nanoSeconds/ 1000000 but I may be misunderstanding something.

ZackarySantana:
I changed it to use native golang capabilities and get the milliseconds, I tested it and it gave the correct number

malikchaya2 (approved):

rest/model/build.go
```diff
@@ -138,6 +139,7 @@ type APITaskCache struct {
 	StatusDetails   apimodels.TaskEndDetail `json:"task_end_details"`
 	StartTime       *time.Time              `json:"start_time"`
 	TimeTaken       time.Duration           `json:"time_taken"`
+	TimeTakenMs     APIDuration             `json:"time_taken_ms"`
```
ablack12, line 142: nit: TimeTakenMS; we should capitalize all of an acronym

ablack12 (dismissed):

bynn (approved):
change PR title from 'Added' to 'Add'

rest/model/build.go
ZackarySantana: Fixed!

PR #5734, opened 2022-06-21
EVG-16293 Add return so that loading a project doesn't panic 

malikchaya2:
[EVG-16293](https://jira.mongodb.org/browse/EVG-16293)

### Description 
We currently panic  [here](https://github.com/evergreen-ci/evergreen/blob/fa0e2d13b1a7065724e7ebfe36e8d36c298ce382/service/patch.go#L81) if we've hit an error [here](https://github.com/evergreen-ci/evergreen/blob/fa0e2d13b1a7065724e7ebfe36e8d36c298ce382/graphql/util.go#L361-L363). This fixes it.


service/patch.go
```diff
@@ -61,6 +62,7 @@ func (uis *UIServer) patchPage(w http.ResponseWriter, r *http.Request) {
 		// still display patch page if problem finding commit queue
 		if err != nil {
 			uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "error finding commit queue"))
+			return
```
ablack12, line 65: I think we shouldn't add this return per the comment at 62

bynn (approved):

PR #5735, opened 2022-06-21
EVG-16880: system-fail tasks that try to exceed the container allocation limit

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16880

### Description 
This is a follow-on to #5720, which added a counter to track each task's number of allocation attempts. When a pod is being terminated and cleaned up, we have to re-allocate new pods for any task that it was supposed to run but could not. However, we don't want to try allocating a container for a task past a certain number of failed attempts, so mark tasks that try to exceed that limit as system-failed. Since we don't support task groups yet, each task gets its own dedicated pod to run it (i.e. there's a 1:1 mapping between container tasks and pods). The implication of this assumption is that if a pod gets terminated (for whatever reason) before it finishes running the task, the one task it was allocated for needs to have at least one more container allocation attempt in order to retry.

### Testing 
Added unit tests.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1729,3 +1729,44 @@ func checkResetDisplayTask(t *task.Task) error {
 	}
 	return errors.Wrap(TryResetTask(t.Id, evergreen.User, evergreen.User, details), "resetting display task")
 }
+
+// MarkUnallocatableContainerTasksSystemFailed marks any container task within
+// the candidate task IDs that needs to re-allocate a container but has used up
+// all of its container allocation attempts as finished due to system failure.
+func MarkUnallocatableContainerTasksSystemFailed(candidateTaskIDs []string) error {
+	var unallocatableTasks []task.Task
+	for _, taskID := range candidateTaskIDs {
+		tsk, err := task.FindOneId(taskID)
+		if err != nil {
+			return errors.Wrapf(err, "finding task '%s'", taskID)
+		}
+		if tsk == nil {
+			continue
+		}
+		if !tsk.IsContainerTask() {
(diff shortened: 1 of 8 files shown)


model/task_lifecycle.go
```diff
@@ -1729,3 +1729,44 @@ func checkResetDisplayTask(t *task.Task) error {
...
+// all of its container allocation attempts as finished due to system failure.
+func MarkUnallocatableContainerTasksSystemFailed(candidateTaskIDs []string) error {
+	var unallocatableTasks []task.Task
+	for _, taskID := range candidateTaskIDs {
+		tsk, err := task.FindOneId(taskID)
+		if err != nil {
+			return errors.Wrapf(err, "finding task '%s'", taskID)
+		}
+		if tsk == nil {
+			continue
```
hadjri, line 1744: Should this also be an error? I'd suspect if we got here from the pod allocator invocaction that this would be unexpected.

model/task_lifecycle.go
```diff
@@ -1729,3 +1729,44 @@ func checkResetDisplayTask(t *task.Task) error {
...
+		}
+		if !tsk.IsContainerTask() {
+			continue
+		}
+		if !tsk.ContainerAllocated {
+			continue
+		}
+		if tsk.RemainingContainerAllocationAttempts() > 0 {
+			continue
```
hadjri, line 1753: Would it make sense to increment the task's attempts if we get here?

hadjri: Ah I see, ok sgtm

hadjri (approved):

PR #5736, opened 2022-06-21
EVG-16810 Add error with activated_by to check_blocked_tasks

malikchaya2:

### Description 
There are tasks with dependencies that don't exist that are hard to identify or reproduce. It looks like they all belong to one user. This PR adds the user to the logging so that we can filter on it. 

Will add a splunk alert once this is merged.

### DIFF
--- a/units/check_blocked_tasks.go
+++ b/units/check_blocked_tasks.go
@@ -126,6 +126,12 @@ func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.T
 
 	dependenciesMet, err := t.DependenciesMet(dependencyCaches)
 	if err != nil {
+		grip.Error(message.Fields{
+			"message":      "checking if dependencies met for task",
+			"task_id":      t.Id,
+			"activated_by": t.ActivatedBy,
+			"depends_on":   t.DependsOn,
+		})
 		return 0, errors.Wrapf(err, "checking if dependencies met for task '%s'", t.Id)
 	}
 	if dependenciesMet {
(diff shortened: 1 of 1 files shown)


units/check_blocked_tasks.go
```diff
@@ -126,6 +126,12 @@ func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.T
 
 	dependenciesMet, err := t.DependenciesMet(dependencyCaches)
 	if err != nil {
+		grip.Error(message.Fields{
+			"message":      "checking if dependencies met for task",
```
bynn, line 130: if dependencies _are_ met for task 
might sound better

bynn (approved):

units/check_blocked_tasks.go
ablack12: this is meant to match the logging below i think, which we probably shouldn't change so that we can search consistently since this is still a problem

units/check_blocked_tasks.go
```diff
@@ -126,6 +126,12 @@ func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.T
 
 	dependenciesMet, err := t.DependenciesMet(dependencyCaches)
 	if err != nil {
+		grip.Error(message.Fields{
```
ablack12, line 129: I would preference grip.Debug, since we're only putting this here to help us understand what's going on, and we already log this error separately

ablack12 (approved):
I do think the splunk alert shouldn't _filter on_ activated_by: regardless of who activated it (even if it isn't the user we've regularly been seeing), we can ask them for clarification.

malikchaya2:
Note: the "failing" task was restarted and passed. The new status doesn't show up here.

PR #5737, opened 2022-06-21
EVG-16985 Rewrite EndTask to support containers

hadjri:
[EVG-16985](https://jira.mongodb.org/browse/EVG-16985)

### Description 
Implemented a separate EndTask specifically on REST V2 to handle tasks running on containers.
### Testing 
Created new unit tests for the endpoint and multiple happy/unhappy paths.


rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+			"build_agent":             evergreen.AgentVersion,
+			"task_details_from_agent": h.details,
+			"pod_id":                  p.ID,
+		})
+		endTaskResp.ShouldExit = true
+		return gimlet.NewJSONResponse(endTaskResp)
+	}
+
+	// Clear the running task on the pod now that the task has finished.
+	if err := p.ClearRunningTask(); err != nil {
```
Kimchelly, line 428: Is this safe to do before `model.MarkEnd` has succeeded? I'm concerned that if something happens in between this line and when `model.MarkEnd` actually runs and succeeds, we might tell the pod that it's done running this task and it'll be allowed to move on before the task, abandoning this task in a stuck and incomplete state.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "clearing running task %s for pod %s", t.Id, p.ID))
+	}
+
+	projectRef, err := model.FindMergedProjectRef(t.Project, t.Version, true)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
```
Kimchelly, line 434: nit: this can be simplified down to `gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "finding project")`.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		})
+	}
+	if projectRef == nil {
+		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    errors.Errorf("project '%s' not found", t.Project).Error(),
+		})
+	}
+
+	// mark task as finished
```
Kimchelly, line 446: Do we need this comment?

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		if err = model.HandleEndTaskForCommitQueueTask(t, h.details.Status); err != nil {
+			return gimlet.MakeJSONErrorResponder(err)
+		}
+	}
+
+	// the task was aborted if it is still in undispatched.
+	// the active state should be inactive.
+	if h.details.Status == evergreen.TaskUndispatched {
+		if t.Activated {
+			grip.Warningf("task %v is active and undispatched after being marked as finished", t.Id)
```
Kimchelly, line 463: We generally don't use `%v` formatting unless we need to specially log something complex like a struct: `task '%s' is active...`

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+			return gimlet.NewJSONResponse(&apimodels.EndTaskResponse{})
+		}
+		grip.Infof(fmt.Sprintf("task %v has been aborted and will not run", t.Id))
```
Kimchelly, line 466: Can we use `grip.Info(message.Fields` and include at least the task ID + route for structured logging?

Kimchelly: `task '%s' has been aborted`

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+			"task_status_from_db":     t.Status,
+			"task_details_from_db":    t.Details,
+			"current_agent":           p.AgentVersion == evergreen.AgentVersion,
+			"agent_version":           p.AgentVersion,
+			"build_revision":          evergreen.BuildRevision,
+			"build_agent":             evergreen.AgentVersion,
+			"task_details_from_agent": h.details,
+			"pod_id":                  p.ID,
+		})
+		endTaskResp.ShouldExit = true
```
Kimchelly, line 423: For pods, there's probably no case in which we would want to set `ShouldExit` to true. Asking the agent to exit is undesirable, because unlike for EC2 hosts, the agent won't re-download and restart. If the pod isn't running a task, the correct move here seems like it would be for the agent to not shut down and simply have it move on to requesting the next task route again.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+	return nil
+}
+
+// Run finishes the given task and generates a job to collect task end stats.
+// It then marks the task as finished or inactive if aborted.
+func (h *podAgentEndTask) Run(ctx context.Context) gimlet.Responder {
+	finishTime := time.Now()
+	p, err := findPod(h.podID)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(err)
```
Kimchelly, line 401: `gimlet.MakeJSONInternalErrorResponder` - if DB find methods return non-nil error, it would be due to an error other than the doc-not-found error, which is considered a server-side issue. `gimlet.MakeJSONErrorResponder` will return 400 Bad Request, which is supposed to be used for client-side errors, so using `gimlet.MakeJSONInternalErrorResponder` is generally preferred since it indicates a generic server-side error.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+	// GetDisplayTask will set the DisplayTask on t if applicable
+	// we set this before the collect task end job is run to prevent data race
+	dt, err := t.GetDisplayTask()
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "getting display task"))
+	}
+	queue := h.env.RemoteQueue()
+	job := units.NewCollectTaskEndDataJob(t, nil, p)
+	if err = queue.Put(ctx, job); err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
```
Kimchelly, line 479: `gimlet.MakeJSONInternalErrorResponder`

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+	}
+	details := apimodels.TaskEndDetail{}
+	if err := utility.ReadJSON(utility.NewRequestReader(r), &details); err != nil {
+		return errors.Wrap(err, "reading task end details from JSON request body")
+	}
+	h.details = details
+
+	// Check that status is either finished or aborted (i.e. undispatched)
+	if !evergreen.IsValidTaskEndStatus(details.Status) && details.Status != evergreen.TaskUndispatched {
+		return errors.Errorf("invalid end status '%s' for task %s", details.Status, h.taskID)
```
Kimchelly, line 390: `for task '%s'`

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+			"task_details_from_agent": h.details,
+			"pod_id":                  p.ID,
+		})
+		endTaskResp.ShouldExit = true
+		return gimlet.NewJSONResponse(endTaskResp)
+	}
+
+	// Clear the running task on the pod now that the task has finished.
+	if err := p.ClearRunningTask(); err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "clearing running task %s for pod %s", t.Id, p.ID))
```
Kimchelly, line 429: `clearing running task '%s' for pod '%s'`

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		if t.Activated {
+			grip.Warningf("task %v is active and undispatched after being marked as finished", t.Id)
+			return gimlet.NewJSONResponse(&apimodels.EndTaskResponse{})
+		}
+		grip.Infof(fmt.Sprintf("task %v has been aborted and will not run", t.Id))
+		return gimlet.NewJSONResponse(&apimodels.EndTaskResponse{})
+	}
+
+	// GetDisplayTask will set the DisplayTask on t if applicable
+	// we set this before the collect task end job is run to prevent data race
```
Kimchelly, line 471: What data race does this prevent? I don't see any place where this would be relevant in the job since the job only has BSON tags for the task ID and doesn't have anything for the display task.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "getting display task"))
+	}
+	queue := h.env.RemoteQueue()
+	job := units.NewCollectTaskEndDataJob(t, nil, p)
+	if err = queue.Put(ctx, job); err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
+	}
+
+	if p.Status != pod.StatusRunning {
+		endTaskResp.ShouldExit = true
```
Kimchelly, line 483: Same thing as above - if it's not running, the move should be to enqueue the termination job to stop the pod.

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+		"task_id":     t.Id,
+		"execution":   t.Execution,
+		"operation":   "mark end",
+		"duration":    time.Since(finishTime),
+		"should_exit": endTaskResp.ShouldExit,
+		"status":      h.details.Status,
+	}
+
+	if dt != nil {
+		msg["display_task_id"] = t.DisplayTask.Id
```
Kimchelly, line 497: Would this be different from just `t.DisplayTaskId`?

units/stats_task_end.go
```diff
@@ -52,51 +55,30 @@ func newTaskEndJob() *collectTaskEndDataJob {
...
-	if j.host == nil {
-		j.host, err = host.FindOneId(j.HostID)
-		j.AddError(err)
-		if err != nil {
-			return
-		}
-	}
-	if j.host == nil {
-		j.AddError(errors.Errorf("Could not find host '%s'", j.HostID))
+	isHostMode := j.host != nil
```
Kimchelly, line 79: This will always be nil because Amboy jobs in the remote queue are round-tripped through the DB, which zeroes out unexported fields such as `j.host`.

units/stats_task_end.go
```diff
@@ -52,51 +55,30 @@ func newTaskEndJob() *collectTaskEndDataJob {
...
+	if !isHostMode && j.pod == nil {
+		j.AddError(errors.New("Could not find host or pod"))
```
Kimchelly, line 81: Can the error include the expected host ID or pod ID?

units/stats_task_end.go
```diff
@@ -52,51 +55,30 @@ func newTaskEndJob() *collectTaskEndDataJob {
...
 	j := newTaskEndJob()
 	j.TaskID = t.Id
 	j.Execution = t.Execution
-	j.HostID = h.Id
+	var id string
+	if h != nil {
+		id = h.Id
+	} else {
+		id = p.ID
+	}
```
Kimchelly, line 67: Rather than storing both host ID and pod ID, would it be preferable to just combine the two and store one runtime env ID? Whether or not it ran in a host or pod can be checked by using the task's execution mode field.

Kimchelly (changes requested):

rest/route/pod_agent_test.go
```diff
@@ -301,3 +305,129 @@ func TestPodAgentNextTask(t *testing.T) {
 		})
 	}
 }
+
+func TestPodAgentEndTask(t *testing.T) {
```
Kimchelly, line 309: Can we test other valid cases like aborting a task or handling commit queue merge tasks?

rest/route/pod_agent.go
```diff
@@ -328,3 +348,155 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+	// mark task as finished
+	deactivatePrevious := utility.FromBoolPtr(projectRef.DeactivatePrevious)
+	err = model.MarkEnd(t, evergreen.APIServerTaskActivator, finishTime, &h.details, deactivatePrevious)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "calling mark finish on task %v", t.Id))
+	}
+
+	if t.Requester == evergreen.MergeTestRequester {
+		if err = model.HandleEndTaskForCommitQueueTask(t, h.details.Status); err != nil {
+			return gimlet.MakeJSONErrorResponder(err)
```
Kimchelly, line 455: `gimlet.MakeJSONInternalErrorResponder`

rest/route/pod_agent.go
hadjri: This was an older comment but looking at the PR from the commit it came in, the "race" it's referring to is a race that was causing a flaky unit test, not an actual race in evergreen. I'm going to remove this since it's not helpful.

rest/route/pod_agent.go
```diff
@@ -443,44 +434,57 @@ func (h *podAgentEndTask) Run(ctx context.Context) gimlet.Responder {
...
 	// the active state should be inactive.
 	if h.details.Status == evergreen.TaskUndispatched {
 		if t.Activated {
-			grip.Warningf("task %v is active and undispatched after being marked as finished", t.Id)
+			grip.Warningf("task %s is active and undispatched after being marked as finished", t.Id)
 			return gimlet.NewJSONResponse(&apimodels.EndTaskResponse{})
 		}
-		grip.Infof(fmt.Sprintf("task %v has been aborted and will not run", t.Id))
+		grip.Info(message.Fields{
+			"message": fmt.Sprintf("task %s has been aborted", t.Id),
```
Kimchelly, line 462: Usually for structured logging, we don't need to include the task ID in the message because the task ID is already included as one of the fields. The message is generally reserved for additional information about why it logged, whereas other fields like `task_id` are for specific resource identifiers that we want to include with the message.

rest/route/pod_agent.go
```diff
@@ -443,44 +434,57 @@ func (h *podAgentEndTask) Run(ctx context.Context) gimlet.Responder {
...
+			return gimlet.MakeJSONInternalErrorResponder(err)
 		}
 	}
 
 	// the task was aborted if it is still in undispatched.
 	// the active state should be inactive.
 	if h.details.Status == evergreen.TaskUndispatched {
 		if t.Activated {
-			grip.Warningf("task %v is active and undispatched after being marked as finished", t.Id)
+			grip.Warningf("task %s is active and undispatched after being marked as finished", t.Id)
```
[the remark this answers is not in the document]
Kimchelly: Can this also use `message.Fields` like the one below this?

rest/route/pod_agent.go
```diff
@@ -443,44 +434,57 @@ func (h *podAgentEndTask) Run(ctx context.Context) gimlet.Responder {
...
 	if err = queue.Put(ctx, job); err != nil {
-		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
 	}
 
 	if p.Status != pod.StatusRunning {
-		endTaskResp.ShouldExit = true
+		j := units.NewPodTerminationJob(h.podID, "pod is no longer running", utility.RoundPartOfMinute(0))
+		if err := amboy.EnqueueUniqueJob(ctx, h.env.RemoteQueue(), j); err != nil {
+			return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
```
Kimchelly, line 483: nit: `gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "enqueueing job to terminate pod"))`

rest/route/pod_agent.go
```diff
@@ -443,44 +434,57 @@ func (h *podAgentEndTask) Run(ctx context.Context) gimlet.Responder {
...
 	dt, err := t.GetDisplayTask()
 	if err != nil {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "getting display task"))
 	}
 	queue := h.env.RemoteQueue()
-	job := units.NewCollectTaskEndDataJob(t, nil, p)
+	job := units.NewCollectTaskEndDataJob(t, nil, p, p.ID)
 	if err = queue.Put(ctx, job); err != nil {
-		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "couldn't queue job to update task stats accounting"))
```
Kimchelly, line 477: `errors.Wrap(err, "enqueueing job to update task stats accounting")`

units/stats_task_end.go
```diff
@@ -27,11 +27,10 @@ func init() {
 }
 
 type collectTaskEndDataJob struct {
-	TaskID    string `bson:"task_id" json:"task_id" yaml:"task_id"`
-	Execution int    `bson:"execution" json:"execution" yaml:"execution"`
-	HostID    string `bson:"host_id" json:"host_id" yaml:"host_id"`
```
Kimchelly, line 31: We have to keep the `HostID` field (and the `j.host` logic in `Run`) temporarily until the app servers roll over and finish running the task end data jobs to maintain temporary backward compatibility with the existing task stats jobs already in the remote queue. We can attach a TODO to remove and then actually remove HostID after this commit is deployed.

rest/route/pod_agent.go
Kimchelly: I still have this question. It seems like we don't need `t.GetDisplayTask` at all because the parent display task is never used, and we could instead use `t.DisplayTaskId` here.

rest/route/pod_agent.go
Kimchelly: It seems like the race could be fixed by passing `task.Task` instead of `*task.Task` into `units.NewCollectTaskEndDataJob` so that the job in the unit test uses a copy rather a pointer to the same underlying Task struct. This would avoid doing this seemingly-unnecessary DB lookup entirely

units/stats_task_end.go
```diff
@@ -84,19 +87,7 @@ func (j *collectTaskEndDataJob) Run(ctx context.Context) {
...
-	if j.host == nil {
-		j.host, err = host.FindOneId(j.HostID)
-		j.AddError(err)
-		if err != nil {
-			return
-		}
-	}
-	if j.host == nil {
-		j.AddError(errors.Errorf("Could not find host '%s'", j.HostID))
+		j.AddError(errors.Errorf("finding task '%s'", j.TaskID))
```
Kimchelly, line 90: The `finding task` convention only applies to `errors.Wrap`/`errors.Wrapf` because they only wrap an existing error with more messages. For base errors (i.e. new errors created by `errors.New`/`errors.Errorf`), they should be written like an error message, like `errors.Errorf("task '%s' not found", j.TaskID)`

units/stats_task_end.go
```diff
@@ -141,11 +129,37 @@ func (j *collectTaskEndDataJob) Run(ctx context.Context) {
...
-			msg["instance_type"] = instanceType
+	isHostMode := j.task.IsHostTask()
+	if isHostMode {
+		j.host, err = host.FindOneId(j.RuntimeEnvID)
+		j.AddError(err)
+		if err != nil {
+			return
+		}
+		if j.host == nil {
+			j.AddError(errors.Errorf("finding host '%s'", j.RuntimeEnvID))
```
Kimchelly, line 140: `errors.Errorf("host '%s' not found", j.RuntimeEnvID)`

units/stats_task_end.go
```diff
@@ -141,11 +129,37 @@ func (j *collectTaskEndDataJob) Run(ctx context.Context) {
...
+			}
+		}
+	} else {
+		j.pod, err = pod.FindOneByID(j.RuntimeEnvID)
+		j.AddError(err)
+		if err != nil {
+			return
+		}
+		if j.pod == nil {
+			j.AddError(errors.Errorf("finding pod '%s'", j.RuntimeEnvID))
```
Kimchelly, line 159: `errors.Errorf("pod '%s' not found", j.RuntimeEnvID)`

units/stats_task_end.go
```diff
@@ -141,11 +129,37 @@ func (j *collectTaskEndDataJob) Run(ctx context.Context) {
...
+			return
 		}
+		msg["pod_id"] = j.pod.ID
```
Kimchelly, line 162: It seems like if we just log the pod ID, we don't need to look up the pod in the DB. Can we add other potentially useful information like the OS, arch, Windows version (if set), CPU, and memory?

rest/route/pod_agent.go
```diff
@@ -328,3 +348,159 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
...
+	}
+
+	msg := message.Fields{
+		"message":     "Successfully marked task as finished",
+		"task_id":     t.Id,
+		"execution":   t.Execution,
+		"operation":   "mark end",
+		"duration":    time.Since(finishTime),
+		"should_exit": endTaskResp.ShouldExit,
+		"status":      h.details.Status,
```
Kimchelly, line 497: Since this is the same log as in the legacy host route, should this (and the corresponding log in the legacy one) include the route so we can distinguish which route it came from?

rest/route/pod_agent.go
```diff
@@ -272,8 +274,8 @@ func (h *podAgentNextTask) Run(ctx context.Context) gimlet.Responder {
 	return gimlet.NewJSONResponse(apimodels.NextTaskResponse{})
 }
 
-func (h *podAgentNextTask) findPod() (*pod.Pod, error) {
-	p, err := pod.FindOneByID(h.podID)
+func findPod(podID string) (*pod.Pod, error) {
```
Kimchelly, line 277: Should `findPod` and `findTask` be moved to the REST data package?

Kimchelly (changes requested):

rest/route/pod_agent.go
hadjri: We need to call `GetDisplayTask` because it actually sets the display task. If we just call `t.DisplayTaskId` it may be nil

hadjri: I've changed this to now check `if t.IsPartOfDisplay()` at the end of the function which will call `GetDisplayTask` behind the scenes

Kimchelly (approved):

PR #5740, opened 2022-06-22
EVG-17062 Remove legacy cedar config route

hadjri:
[EVG-17062](https://jira.mongodb.org/browse/EVG-17062)

### Description 
Waited some time for agents to roll over before removing this route after the default cedar config route was migrated to v2. Removing the v1 route would not have been backwards compatible with agents on hosts that were started before [this commit](https://github.com/evergreen-ci/evergreen/pull/5707) was deployed.

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,7 +36,7 @@ var (
 	ClientVersion = "2022-06-07"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2022-06-22"
+	AgentVersion = "2022-06-17"
 )
 
 // ConfigSection defines a sub-document in the evergreen config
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -36,7 +36,7 @@ var (
 	ClientVersion = "2022-06-07"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2022-06-17"
```
Kimchelly, line 39: The agent revision doesn't need to be bumped because the agent executable should already be using the newer route.

Kimchelly (approved):
LGTM mod the one comment.

PR #5742, opened 2022-06-22
EVG-17061 Hide container configs until container tasks are available

hadjri:
[EVG-17061](https://jira.mongodb.org/browse/EVG-17061)

### Description 
It is currently misleading that container resource / secret configs have UI components on the project settings page despite the fact that they are not in use at this time.

I've created https://jira.mongodb.org/browse/EVG-17134 to track uncommenting this section from the HTML once we'd like to expose containerized tasks to users.

### DIFF
--- a/service/templates/projects.html
+++ b/service/templates/projects.html
@@ -976,103 +976,103 @@ Evergreen Projects
           </div>
         </div>
       </div>
-<!--      TODO: EVG-17134 Uncomment once container tasks are available-->
-<!--      &lt;!&ndash; Custom container resource aliases &ndash;&gt;-->
-<!--      <div class="variables">-->
-<!--        <div class="form-group">-->
-<!--          <div class="col-header col-lg-6 form-control-static">-->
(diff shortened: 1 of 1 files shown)


service/templates/projects.html
```diff
@@ -976,102 +976,103 @@ <h3>Periodic Builds</h3>
...
-          </div>
-          <div class="col-lg-2">
-            <button class="plus-button btn btn-primary " ng-disabled="!validContainerCredential(container_credential)" type="button" ng-click="addContainerCredential()">
-              <i class="fa fa-plus"></i>
-            </button>
-            <label class="distro-error">[[invalidContainerCredentialMessage]]</label>
-          </div>
-        </div>
-      </div>
+<!--      TODO: EVG-17134 Uncomment once container tasks are available-->
```
ablack12, line 979: An alternative way to do this would be to wrap the whole thing in a div that has hide=true, which may be cleaner

PR #5743, opened 2022-06-22
EVG-16888 Refactor addBuildVariantDisplayName to use cached field if it exists

khelif96:
[EVG-16888](https://jira.mongodb.org/browse/EVG-16888)

### Description 
Refactors AddBuildVariantDisplayName to use the cached value in the task struct that was introduced in #5715 if it exists. If the field does not exist (On older task documents) We will use the `$lookup` method to get the build variant display name.
### Testing 
Tested locally and staging and as a stand alone db query. Old unit tests should be sufficient as well.

### DIFF
--- a/graphql/tests/mainlineCommits/data.json
+++ b/graphql/tests/mainlineCommits/data.json
@@ -118,7 +118,7 @@
             "build_variant": "enterprise-ubuntu1604-64",
             "display_name": "Ubuntu 16.04"
         },
-                        {
+        {
             "_id": "evergreen_version5_build",
             "build_variant": "enterprise-ubuntu1604-64",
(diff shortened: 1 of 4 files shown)


hadjri (commented):
LGTM although can we add a few more cases to `model/task_build_variant_names_test.go` where we set `BuildVariantDisplayName` on a subset of the tasks?

khelif96:
> LGTM although can we add a few more cases to `model/task_build_variant_names_test.go` where we set `BuildVariantDisplayName` on a subset of the tasks?

Great catch! I wasn't even using the field for that query since the `build_variant_display_name` was projected out in the pipeline. Just updated it.

hadjri (approved):
LGTM once build passes!

PR #5744, opened 2022-06-22
EVG-14878: Remove checkbox to enable Cedar test results from project settings page

julianedwards:
EVG-14878: https://jira.mongodb.org/browse/EVG-14878

We will now be forcing all projects  except for the mongodb-mongo-* projects to use Cedar test results.
I checked in staging and the Test Results section is no longer on the page.


Kimchelly (commented):
Since we're just removing the ability to set it from the UI and not removing the flag itself, are all projects aside from mongodb-mongo-* already opted in?

julianedwards:
Nope, once this is deployed I am going to run an update command against the project ref collection to set the field to true in all projects that are not mongodb-mongo-*.

Kimchelly (approved):

PR #5751, opened 2022-06-24
EVG-17086 Add Support for Periodic Build requester to MainlineCommits Query

malikchaya2:
[EVG-17086](https://jira.mongodb.org/browse/EVG-17086)

### Description 
This adds the Periodic Build requester to MainlineCommits query so that it can be viewed on the UI. 

### Testing 
[Tested on staging](http://localhost:3000/commits/mci?requester=all,gitter_request,git_tag_request,trigger_request,ad_hoc&skipOrderNumber=7828) on the mci project. 

### Spruce PR 
[1340](https://github.com/evergreen-ci/spruce/pull/1340)

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -509,7 +509,7 @@ const (
 	RepotrackerVersionRequester = "gitter_request"
 	TriggerRequester            = "trigger_request"
 	MergeTestRequester          = "merge_test" // commit queue
-	AdHocRequester              = "ad_hoc"
+	AdHocRequester              = "ad_hoc"     // periodic build
 )
 
 var AllRequesterTypes = []string{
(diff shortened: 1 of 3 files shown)


khelif96 (approved):

malikchaya2:
> I think we should remove this too:
> 
> https://github.com/evergreen-ci/evergreen/blob/698add52308336caf4e8b56f6e0e9c3e118be296/repotracker/repotracker.go#L842
> 
> Right now, we use the patch number if someone creates an adhoc version using a rest route. (I'm not actually sure if anyone does this but). Since these also use the "adhoc" requester type, it seems bad that they could be wrongly ordered in the mainline commits view, based solely on the user's numbers. We should probably also order these right, and if it becomes a problem that these are viewable on the mainline commits page we could always look into a different requester for them.

I didn't know that, thanks for spotting it! I removed it.

ablack12 (approved):

PR #5752, opened 2022-06-27
EVG-17047 Redirect users to spruce UI in BF ticket links

ZackarySantana:
[EVG-17047](https://jira.mongodb.org/browse/EVG-17047)

### Description 
BF tickets don't redirect with every link to the spruce UI. This adds the parameter to the BF ticket links to redirect to the spruce UI and adds the redirection logic to the pages associated that did not have it. As well, updated redirection logic to be consistent with new methods.

### Testing 
Created a BF ticket and tested all the links that should have changed (host, project- waterfall/project health, task history). [This is the BF created in testing.](https://jira.mongodb.org/browse/EVG-17188)

### DIFF
--- a/service/host.go
+++ b/service/host.go
@@ -11,6 +11,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/event"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/evergreen/model/user"
 	"github.com/evergreen-ci/gimlet"
 	"github.com/evergreen-ci/gimlet/rolemanager"
 	"github.com/evergreen-ci/utility"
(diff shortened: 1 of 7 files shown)


ZackarySantana:
I updated the redirection logic but I don't know if I updated them to the correct version (if we prefer getting the user through gimlet or using the wrapper function that gets it for us.

bynn (changes requested):
could you test that all these new pages redirects to spruce when the users have the new UI setting on and stays on legacy when it doesn't?

ZackarySantana:
I tested turning off my setting, all the links go to the original UI. They keep the param of redirect in the URL but it doesn't redirect. The only exception is when I clicked the waterfall link, which has the param until it refreshes because it goes from MCI to the evergreen project. So it works as intended for all links.

bynn (approved):

PR #5756, opened 2022-06-27
EVG-17195: deallocate stranded container task during pod termination

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17195

### Description 
When a pod is about to terminate, tasks that are waiting to be dispatched to that pod are de-allocated to indicate that they need a new pod to run them. However, in the case that a pod is about to be terminated but it has already been dispatched to the pod, that stranded container task also has to be de-allocated so that when it restarts with a new execution, it gets allocated a new container to run it. I added some logic to ensure that the stranded task is re-allocated when the pod is abruptly terminated.

### Testing 
Added unit test.

### DIFF
--- a/units/pod_termination.go
+++ b/units/pod_termination.go
@@ -11,6 +11,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/pod"
 	"github.com/evergreen-ci/evergreen/model/pod/dispatcher"
+	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/mongodb/amboy"
 	"github.com/mongodb/amboy/job"
 	"github.com/mongodb/amboy/registry"
@@ -202,8 +203,14 @@ func (j *podTerminationJob) populateIfUnset(ctx context.Context) error {
 // tasks in the dispatcher as needing re-allocation.
 func (j *podTerminationJob) fixStrandedTasks(ctx context.Context) error {
 	if j.pod.RunningTask != "" {
+		// A stranded task will need to be re-allocated to ensure that it
+		// dispatches to a new pod after this one is terminated.
+		if err := task.MarkManyContainerDeallocated([]string{j.pod.RunningTask}); err != nil {
+			return errors.Wrapf(err, "marking stranded container task '%s' running on pod as deallocated", j.pod.RunningTask)
+		}
+
 		if err := model.ClearAndResetStrandedContainerTask(j.pod); err != nil {
-			return errors.Wrap(err, "resetting stranded container task running on pod")
+			return errors.Wrapf(err, "resetting stranded container task '%s' running on pod", j.pod.RunningTask)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


units/pod_termination.go
```diff
@@ -202,8 +203,14 @@ func (j *podTerminationJob) populateIfUnset(ctx context.Context) error {
 // tasks in the dispatcher as needing re-allocation.
 func (j *podTerminationJob) fixStrandedTasks(ctx context.Context) error {
 	if j.pod.RunningTask != "" {
+		// A stranded task will need to be re-allocated to ensure that it
+		// dispatches to a new pod after this one is terminated.
+		if err := task.MarkManyContainerDeallocated([]string{j.pod.RunningTask}); err != nil {
+			return errors.Wrapf(err, "marking stranded container task '%s' running on pod as deallocated", j.pod.RunningTask)
```
hadjri, line 209: can we add `pod '%s'` to this message?

units/pod_termination.go
```diff
@@ -202,8 +203,14 @@ func (j *podTerminationJob) populateIfUnset(ctx context.Context) error {
 // tasks in the dispatcher as needing re-allocation.
 func (j *podTerminationJob) fixStrandedTasks(ctx context.Context) error {
 	if j.pod.RunningTask != "" {
+		// A stranded task will need to be re-allocated to ensure that it
+		// dispatches to a new pod after this one is terminated.
+		if err := task.MarkManyContainerDeallocated([]string{j.pod.RunningTask}); err != nil {
```
hadjri, line 208: `MarkManyContainerDeallocated` is a bit confusing since I'm assuming there are going to be additional cases potentially where we'd just want to de-allocate a single container task.  Can we rename to something like `MarkTasksAsContainerDeallocated`?

units/pod_termination_test.go
```diff
@@ -156,7 +158,7 @@ func TestPodTerminationJob(t *testing.T) {
 			dbTask, err := task.FindOneId(tsk.Id)
 			require.NoError(t, err)
 			require.NotZero(t, dbTask)
-			assert.Equal(t, evergreen.TaskUndispatched, dbTask.Status, "stranded task should have been restarted")
+			assert.True(t, dbTask.ShouldAllocateContainer(), "stranded task should have been restarted to re-attempt allocation")
```
hadjri, line 161: can we check `ContainerAllocated` field on the task as well?

hadjri (changes requested):

units/pod_termination.go
Kimchelly: Sounds good, changed.

units/pod_termination.go
Kimchelly: Done. Also added it to the other `errors.Wrap`s in this functions.

units/pod_termination_test.go
Kimchelly: `ShouldAllocateContainer` checks `ContainerAllocated` because it'll return false if it's already been allocated a container.

hadjri (approved):

PR #5757, opened 2022-06-28
EVG-16866 Allow finalize flag with regex tasks/variants

ZackarySantana:
[EVG-16866](https://jira.mongodb.org/browse/EVG-16866)

### Description 
Enables users to use the finalize flag with regex tasks/variants and mixing regex tasks with a list of variants or regex variants with a list of tasks.

### Testing 
Running `evergreen patch --rv '.*' --rt '.*' -f` results in an error saying that no tasks/variants are listed. After these commits, running the same command will give a patch link that has all the tasks on all variants to be ran (the regex provided is all). If you run `evergreen patch -t generate-lint --rv lint -f` works and schedules the task 'generate-lint' on the 'lint' variant.

### DIFF
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -271,15 +271,10 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
 		}
 	}
 
-	if !p.Finalize || p.Alias != "" {
-		return ref, nil
-	}
-
-	if len(p.Tasks) == 0 && len(p.RegexTasks) == 0 {
-		return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing (No task specified).")
-	}
-	if len(p.Variants) == 0 && len(p.RegexVariants) == 0 {
-		return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing (No variants specified).")
+	if p.Finalize && p.Alias != "" {
+		if len(p.Variants)+len(p.RegexVariants) == 0 || len(p.Tasks)+len(p.RegexTasks) == 0 {
+			return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing")
+		}
 	}
 
 	return ref, nil
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -271,7 +271,7 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
 		}
 	}
 
-	if p.Finalize && p.Alias != "" {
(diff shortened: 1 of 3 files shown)


operations/patch_util.go
```diff
@@ -271,8 +271,15 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
 		}
 	}
 
-	if (len(p.Tasks) == 0 || len(p.Variants) == 0) && p.Alias == "" && p.Finalize {
-		return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing.")
+	if !p.Finalize || p.Alias != "" {
+		return ref, nil
```
bynn, line 275: instead of return here, you could just move the if statements below into the block so we dont have to return the same thing on 278

operations/patch_util.go
```diff
@@ -271,8 +271,15 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
...
+	}
+
+	if len(p.Tasks) == 0 && len(p.RegexTasks) == 0 {
+		return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing (No task specified).")
```
bynn, line 279: changing the message to say task or variant depending on the case is probably clearer instead of adding something on at the end

operations/patch_util.go
```diff
@@ -271,8 +271,15 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
...
+	}
+	if len(p.Variants) == 0 && len(p.RegexVariants) == 0 {
+		return ref, errors.Errorf("Need to specify at least one task/variant or alias when finalizing (No variants specified).")
```
bynn, line 282: same here

bynn (changes requested):

operations/patch_util.go
ZackarySantana: Fixed!

operations/patch_util.go
ZackarySantana: Modified error messages

ablack12 (commented):
Can you verify what happens if a patch _would_ be created with no tasks/variants? (For example, repeat failed on a successful patch)

ZackarySantana:
@ablack12 When doing repeat failed after a successful patch, it will create a patch with no tasks selected. When using the finalize flag, it will error saying "Need to specify at least one task/variant or alias when finalizing". I think both behaviors are expected, although users might have a better experience if the repeat failed flag warned them (or asked for confirmation with) that their patch will have no tasks because the last patch passed completely

PR #5760, opened 2022-06-28
EVG-17047 Update redirect to not require user

ZackarySantana:
[EVG-17047](https://jira.mongodb.org/browse/EVG-17047)

### Description 
In ticket EVG-17047, the logic was updated for redirect to be consistent but it was updated to a version that required a logged in user or it would through an error. This reverts those changes and makes it so the redirect doesn't require a logged in user.

### Testing 
Logging out and opening evergreen


khelif96 (approved):

ZackarySantana:
I extracted it to a helper function and I am testing now!

ZackarySantana:
I tested with my settings on redirect and off, and both gave the expected behavior

PR #5764, opened 2022-06-29
EVG-16810 Fix task dependencies broken for --reuse 

malikchaya2:

### Description 
We were running into patches that have tasks with dependencies that don't exist. Based on our sleuthing, it appears that they were created with the --reuse flag. This PR refactors how we attach variant tasks in that case. Instead of grabbing it from the previous tasks, we instead simply set the necessary things from the previous patch without attaching variant tasks. By doing that, it then hits BuildProjectTVPairs which will also include the dependencies. 

### Testing 
While I can't be certain that this solved the bug until we merge it and see if it logs in splunk again. However, I did test the --reuse and repeat-failed flags on staging to make sure they still work. 

Once this PR is merged, I will reset the splunk alert to its original form.


units/patch_intent.go
```diff
@@ -302,8 +302,16 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 	}
 
 	if j.intent.ReusePreviousPatchDefinition() {
-		patchDoc.VariantsTasks, err = j.getPreviousPatchDefinition(project, false)
+		err = j.setToPreviousPatchDefinition(patchDoc, project)
 		if err != nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message":     "setting patch definition to the previous patch",
```
ablack12, line 308: Do we return this error after finish patch anyway, since we add it to the job? Not sure we need to double log.

ablack12: Do we also need to modify 320? (only failed?)

ablack12 (changes requested):

units/patch_intent.go
malikchaya2: So we probably should. However, I'm hesitant to do it right now because if it doesn't work well, I won't know, because we don't have any instances of it to monitor. We also haven't seen any errors associated with it so I don't think it's bothering anyone right now.

malikchaya2: removed logging

self-tests.yml
```diff
@@ -329,11 +329,22 @@ tasks:
     name: dist-staging
     patch_only: true
     run_on: archlinux-new-small
+    depends_on:
```
malikchaya2, line 332: will remove. I'm still testing stuff, this branch is not ready for re-review

malikchaya2: it's ready for re-review now

units/patch_intent.go
```diff
@@ -339,7 +335,9 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 		}
 	}
 
-	if len(patchDoc.VariantsTasks) == 0 {
+	skipForFailed := failedOnly && !(previousPatchStatus == evergreen.PatchFailed)
+
+	if len(patchDoc.VariantsTasks) == 0 && !skipForFailed {
```
ablack12, line 340: Why do you want to skip this for failedOnly? I think there's actually a good argument to build pairs for failed, to ensure dependencies are added (since dependencies theoretically were successful and not failed).

units/patch_intent.go
```diff
@@ -339,7 +335,9 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 		}
 	}
 
-	if len(patchDoc.VariantsTasks) == 0 {
+	skipForFailed := failedOnly && !(previousPatchStatus == evergreen.PatchFailed)
```
ablack12, line 338: more succinct would be previousPatchStatus != evergreen.PatchFailed, although I don't really understand the purpose of this.

units/patch_intent.go
```diff
@@ -488,40 +486,43 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
-		} else {
-			tasks, displayTasks = getPreviousTasksAndDisplayTasks(tasksInProjectVariant, displayTasksInProjectVariant, vt)
-		}
-		if len(tasks)+len(displayTasks) > 0 {
-			res = append(res, patch.VariantTasks{
-				Variant:      vt.Variant,
-				Tasks:        tasks,
-				DisplayTasks: displayTasks,
-			})
+		tasks, _, err := getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant, displayTasksInProjectVariant, vt, previousPatch.Version)
```
ablack12, line 495: Do we not need display tasks anymore? shouldn't these be added somewhere still?

units/patch_intent.go
```diff
@@ -488,40 +486,43 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
-		return nil, errors.Wrap(err, "error querying for most recent patch")
-	}
-	if previousPatch == nil {
-		return nil, errors.Errorf("no previous patch available")
-	}
-	var res []patch.VariantTasks
-	if failedOnly && !(previousPatch.Status == evergreen.PatchFailed) {
-		return res, nil
-	}
+func setFailedTasksToPrevious(patchDoc, previousPatch *patch.Patch, project *model.Project) error {
```
ablack12, line 489: I think this function name isn't quite right. It's really more like setTasksToPreviousFailed or setTopreviousFailedPatchDefinition, right?

units/patch_intent.go
```diff
@@ -488,40 +486,43 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+	if err != nil {
+		return "", errors.Wrap(err, "querying for most recent patch")
+	}
+	if previousPatch == nil {
+		return "", errors.Errorf("no previous patch available")
+	}
+
+	patchDoc.BuildVariants = previousPatch.BuildVariants
+	patchDoc.RegexBuildVariants = previousPatch.RegexBuildVariants
+	patchDoc.RegexTasks = previousPatch.RegexTasks
```
ablack12, line 517: These should also be set based on failed right?

ablack12 (changes requested):
It's a little hard to read what the test is testing. Could we add some print statements? Also I realize FinishPatch doesn't have any tests :( but I think it'd be good if we could write a unit test for the specific thing we're trying to test (i.e. if dependencies don't exist or aren't failed in the previous patch, they should in the new patch). To make this easier I'll bet parts of https://github.com/ablack12/evergreen/blob/8627b302371241b613a480703e6da7eb8f831e6d/units/patch_intent.go#L304-L344 can be refactored to a different function to be tested for this.

units/patch_intent.go
malikchaya2: this exits if it's  using the failedOnly flag but the previous task didn't fail, meaning there's nothing to do since there are no tasks to add because nothing failed.

units/patch_intent.go
malikchaya2: yes :)

units/patch_intent.go
ablack12: This test was with display tasks too? In that case we should be able to remove it as a return then since we don't need it

malikchaya2: I tested it on staging, not not with a unit test. If there is a failed display task, it'll rebuild the entire display task including the non failed tasks. I think that is the current behavior, but let me double check.

malikchaya2: okay so I tested the current behavior again. Turns out that currently, if a failed display task has one passing test and one failing task, it'll restart the display task with only the failed tasks, and leave out the passing execution tasks. Is that the behavior we want? If yes, I'll have to modify the code.

units/patch_intent.go
malikchaya2: done

units/patch_intent.go
malikchaya2: removed the unused code.

units/patch_intent.go
```diff
@@ -488,73 +486,58 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+		setTasksToPreviousFailed(patchDoc, previousPatch, project)
+	} else {
+		patchDoc.Tasks = previousPatch.Tasks
+	}
+
+	return previousPatch.Status, nil
 }
 
-func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, displayTasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, []patch.DisplayTask, error) {
+func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, displayTasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, error) {
```
bynn, line 526: do we need the displayTasksInProjectVariant?

units/patch_intent.go
```diff
@@ -488,73 +486,58 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
-		return nil, nil, errors.Wrap(err, "error querying for failed tasks from previous patch")
+		return nil, errors.Wrap(err, "error querying for failed tasks from previous patch")
 	}
 	failedExecutionTasks := []string{}
-	failedDisplayTasks := []string{}
 	for _, failedTask := range failedTasks {
-		if failedTask.DisplayOnly {
-			failedDisplayTasks = append(failedDisplayTasks, failedTask.DisplayName)
-		} else {
+		if !failedTask.DisplayOnly {
```
bynn, line 533: Im a little confused 
are we assuming all failed tasks that arent a display task to be an execution task?

units/patch_intent.go
malikchaya2: thanks! removing.

units/patch_intent.go
```diff
@@ -488,73 +486,58 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+
+	return previousPatch.Status, nil
 }
 
-func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, displayTasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, []patch.DisplayTask, error) {
+func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, displayTasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, error) {
 	failedTasks, err := task.FindAll(db.Query(task.FailedTasksByVersionAndBV(version, vt.Variant)))
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "error querying for failed tasks from previous patch")
+		return nil, errors.Wrap(err, "error querying for failed tasks from previous patch")
```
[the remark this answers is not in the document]
malikchaya2: It still feels appropriate since the error would be coming from the task.FindAll query. Do you have a different error message in mind?

malikchaya2:
> I realize FinishPatch doesn't have any tests :( but I think it'd be good if we could write a unit test for the specific thing we're trying to test (i.e. if dependencies don't exist or aren't failed in the previous patch, they should in the new patch). To make this easier I'll bet parts of https://github.com/ablack12/evergreen/blob/8627b302371241b613a480703e6da7eb8f831e6d/units/patch_intent.go#L304-L344 can be refactored to a different function to be tested for this.

done

units/patch_intent.go
```diff
@@ -488,73 +450,103 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+		_, err := regexp.Compile(t)
+		if err != nil {
+			return errors.Wrapf(err, "compiling task regex '%s'", t)
+		}
+	}
+
+	// if the user only wants failed tasks but the previous patch has no failed tasks, there is nothing to build
+	skipForFailed := failedOnly && previousPatchStatus != evergreen.PatchFailed
+
+	if len(patchDoc.VariantsTasks) == 0 && !skipForFailed {
```
ablack12, line 493: Is it possible to define reuse with other things? (regex etc?) Because then skipping this for reuse-failed likely isn't worth it / could introduce more weirdness

units/patch_intent.go
```diff
@@ -488,73 +450,103 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
+		return "", errors.Wrap(err, "querying for most recent patch")
+	}
+	if previousPatch == nil {
+		return "", errors.Errorf("no previous patch available")
+	}
+
+	patchDoc.BuildVariants = previousPatch.BuildVariants
+
+	if failedOnly {
+		setTasksToPreviousFailed(patchDoc, previousPatch, project)
```
ablack12, line 527: Should check this error

units/patch_intent_test.go
```diff
@@ -27,6 +28,10 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 )
 
+const (
+	VersionCollection = "versions"
```
ablack12, line 32: It looks like the units collection already imports model so I'm not sure why we can't just use model.VersionCollection

units/patch_intent_test.go
```diff
@@ -328,22 +338,243 @@ func (s *PatchIntentUnitsSuite) TestGetPreviousPatchDefinition() {
...
+	previousPatchStatus, err = j.setToPreviousPatchDefinition(currentPatchDoc, &project, true)
+	s.NoError(err)
+	s.Equal(previousPatchStatus, "failed")
+	s.Equal(currentPatchDoc.BuildVariants, previousPatchDoc.BuildVariants)
+	s.Equal(currentPatchDoc.RegexBuildVariants, previousPatchDoc.RegexBuildVariants)
+	s.Equal(currentPatchDoc.RegexTasks, previousPatchDoc.RegexTasks)
+	s.Equal(currentPatchDoc.Tasks, []string{"t1"})
+}
+
+func (s *PatchIntentUnitsSuite) TestBuildTasksandVariantsWithRepeat() {
```
ablack12, line 361: The "reuse" flag is also called the "repeat" flag, whereas "reuse-failed" is only called that, so I think this test should be RepeatFailed for clarity.

units/patch_intent_test.go
```diff
@@ -328,22 +338,243 @@ func (s *PatchIntentUnitsSuite) TestGetPreviousPatchDefinition() {
...
+		},
+	}
+	s.NoError((previousPatchDoc).Insert())
+
+	intent, err := patch.NewCliIntent(patch.CLIIntentParams{
+		User:        s.user,
+		Project:     s.project,
+		BaseGitHash: s.hash,
+		Description: s.desc,
+		// --reuse flag
```
ablack12, line 417: repeat-failed

units/patch_intent.go
malikchaya2: I may be misunderstanding your comment, but if someone wants to restart only failed tasks, then no matter what else they specify, if there are only passing tests, then we always want to run nothing.

units/patch_intent.go
malikchaya2: thanks for catching that!

units/patch_intent_test.go
```diff
@@ -328,22 +338,243 @@ func (s *PatchIntentUnitsSuite) TestGetPreviousPatchDefinition() {
...
+	project := model.Project{
+		Identifier: s.project,
+		BuildVariants: model.BuildVariants{
+			{
+				Name: "bv1",
+				Tasks: []model.BuildVariantTaskUnit{
+					{
+						Name: "t1",
+						DependsOn: []model.TaskUnitDependency{
+							{Name: "t3", Status: evergreen.TaskFailed},
```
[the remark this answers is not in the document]
malikchaya2: good idea

units/patch_intent.go
```diff
@@ -488,73 +450,103 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
...
-	// We want the subset of vt.tasks that exist in tasksForVariant.
-	tasks := utility.StringSliceIntersection(tasksInProjectVariant, vt.Tasks)
-	var displayTasks []patch.DisplayTask
-	for _, dt := range vt.DisplayTasks {
-		if utility.StringSliceContains(displayTasksInProjectVariant, dt.Name) {
-			displayTasks = append(displayTasks, patch.DisplayTask{Name: dt.Name})
-		}
-	}
-	return tasks, displayTasks
+	tasks := utility.StringSliceIntersection(failedExecutionTasks, vt.Tasks)
```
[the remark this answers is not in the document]
malikchaya2: We probably could, but I don't think it's enough of an improvement to be worth re-writing it (this is from previously written code that's just moved around). I'll add a comment explaining.

units/patch_intent.go
ablack12: Gotcha, yeah ticketing makes sense then 👍

ablack12 (approved):

PR #5765, opened 2022-06-29
EVG-17167 simplify service fns for cq, distro, and notifications

ablack12:

### Description 
Decided to do this in chunks to keep code reviews lil and arbitrarily chose these chunks because they interacted with each other. Considered making BuildFromService work differently entirely (i.e. return the API model rather than destructively modify it) but I do think this is what we're now used to and that would be a bigger change.

Also renamed Icecream to IceCream because my new computer has more aggressive green syntax squiggles and I must tame them. (Although this doesn't match the bson tags as well so if you want me to switch it back I'll get it)

### Testing 
Existing unit tests.

### DIFF
--- a/rest/model/notification.go
+++ b/rest/model/notification.go
@@ -3,7 +3,6 @@ package model
 import (
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip/message"
-	"github.com/pkg/errors"
 )
 
 type APIJiraComment struct {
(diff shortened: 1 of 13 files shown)


rest/model/notification.go
```diff
@@ -104,37 +84,6 @@ type APISlack struct {
 	Attachments []APISlackAttachment `json:"attachments"`
 }
 
-// BuildFromService converts from service level message.Slack to APISlack.
-func (n *APISlack) BuildFromService(h interface{}) error {
```
ablack12, line 108: Removed this function because it actually doesn't work (at 123 we pass in a slack.Attachment but we error if it's not a message.SlackAttachment?) and we aren't using it anywhere.

malikchaya2 (approved):
🎉

PR #5767, opened 2022-06-30
EVG-17208 Remove macOS agent before downloading

bsamek:
[EVG-17208](https://jira.mongodb.org/browse/EVG-17208)

### Description 
macOS caches signatures, which means that a newly downloaded evergreen agent can fail to start. Removing the agent before downloading prevents this from occurring. See https://apple.stackexchange.com/questions/258623/how-to-fix-killed-9-error-in-mac-os/428388#428388.

This PR deliberately does this only on macOS to reduce the amount of work to deploy this, since otherwise every static distro would have to be manually redeployed.

### Testing 
This will require manual testing.

### DIFF
--- a/operations/agent_monitor.go
+++ b/operations/agent_monitor.go
@@ -285,9 +285,7 @@ func (m *monitor) fetchClient(ctx context.Context, urls []string, retry utility.
 	if runtime.GOOS == "darwin" {
 		_, err := os.Stat(m.clientPath)
 		if err == nil {
-			if err := os.Remove(m.clientPath); err != nil {
-				grip.Error(errors.Wrap(err, "deleting client"))
-			}
+			grip.Error(errors.Wrap(os.Remove(m.clientPath), "deleting client"))
 		} else {
 			grip.ErrorWhen(!errors.Is(err, os.ErrNotExist), errors.Wrap(err, "checking if client exists"))
 		}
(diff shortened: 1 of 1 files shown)


operations/agent_monitor.go
```diff
@@ -276,6 +277,20 @@ func handleMonitorSignals(ctx context.Context, serviceCancel context.CancelFunc)
 func (m *monitor) fetchClient(ctx context.Context, urls []string, retry utility.RetryOptions) error {
 	var downloaded bool
 	catcher := grip.NewBasicCatcher()
+
+	// macOS caches signatures, which means that a newly downloaded
+	// evergreen agent can fail to start. Removing the agent before
+	// downloading prevents this from occurring. See
+	// https://apple.stackexchange.com/questions/258623/how-to-fix-killed-9-error-in-mac-os/428388#428388
```
Kimchelly, line 284: Bless this comment (no change)

operations/agent_monitor.go
```diff
@@ -276,6 +277,20 @@ func handleMonitorSignals(ctx context.Context, serviceCancel context.CancelFunc)
...
+	if runtime.GOOS == "darwin" {
+		_, err := os.Stat(m.clientPath)
```
Kimchelly, line 286: Instead of this stat -> remove, can we just use `os.RemoveAll`? It already squashes the error when the file does not exist.

Kimchelly (approved):

bsamek:
Closing, and will re-open if necessary.

PR #5781, opened 2022-07-06
EVG-17167 modify more instances of buildFromService

ablack12:

### Description 
Sorry I tried to keep these PRs small but this got out of hand 🙃

### DIFF
--- a/rest/model/github_hook.go
+++ b/rest/model/github_hook.go
@@ -12,28 +12,19 @@ type APIGithubHook struct {
 	Repo   *string `json:"repo"`
 }
 
-func (a *APIGithubHook) BuildFromService(h interface{}) error {
-	v, ok := h.(model.GithubHook)
-	if !ok {
-		return errors.Errorf("programmatic error: expected GitHub hook but got type %T", h)
-	}
-	if v.HookID == 0 {
+func (a *APIGithubHook) BuildFromService(hook model.GithubHook) error {
+	if hook.HookID == 0 {
 		return errors.New("hook ID cannot be 0")
 	}
-	if v.Owner == "" {
+	if hook.Owner == "" {
 		return errors.New("owner cannot be empty")
 	}
-	if v.Repo == "" {
+	if hook.Repo == "" {
 		return errors.New("repo cannot be empty")
 	}
 
-	a.HookID = v.HookID
-	a.Owner = utility.ToStringPtr(v.Owner)
-	a.Repo = utility.ToStringPtr(v.Repo)
+	a.HookID = hook.HookID
+	a.Owner = utility.ToStringPtr(hook.Owner)
+	a.Repo = utility.ToStringPtr(hook.Repo)
 	return nil
-}
-
-// ToService returns a service layer build using the data from the APIBuild.
-func (*APIGithubHook) ToService() (interface{}, error) {
-	return nil, errors.New("not implemented for read-only route")
 }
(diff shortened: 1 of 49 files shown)


rest/model/github_hook.go
```diff
@@ -12,28 +12,19 @@ type APIGithubHook struct {
 	Repo   *string `json:"repo"`
 }
 
-func (a *APIGithubHook) BuildFromService(h interface{}) error {
-	v, ok := h.(model.GithubHook)
-	if !ok {
-		return errors.Errorf("programmatic error: expected GitHub hook but got type %T", h)
-	}
-	if v.HookID == 0 {
+func (a *APIGithubHook) BuildFromService(hook model.GithubHook) error {
```
hadjri, line 15: Should we remove the error returning here and just ensure it's caught in validation? So we can adhere to the pattern of not returning an err in `BuildFromService`

rest/model/notifications.go
```diff
@@ -13,17 +12,8 @@ type APIEventStats struct {
...
-		return errors.Errorf("programmatic error: expected notification stats but got type %T", h)
-	}
-
-	return n.PendingNotificationsByType.BuildFromService(stats)
-}
-
-func (n *APIEventStats) ToService() (interface{}, error) {
-	return nil, errors.New("(*APIEventStats) ToService not implemented")
+func (n *APIEventStats) BuildFromService(stats notification.NotificationStats) {
+	n.PendingNotificationsByType.BuildFromService(stats)
```
hadjri, line 16: Should we implement `ToService` ?

rest/model/user.go
```diff
@@ -204,22 +153,28 @@ func (n *APINotificationPreferences) ToService() (interface{}, error) {
 	spawnHostOutcome := utility.FromStringPtr(n.SpawnHostOutcome)
 	commitQueue := utility.FromStringPtr(n.CommitQueue)
 	if !user.IsValidSubscriptionPreference(buildBreak) {
-		return nil, errors.Errorf("invalid build break subscription preference '%s'", buildBreak)
+		return user.NotificationPreferences{},
```
hadjri, line 156: Can this be validated separately as well?

hadjri (commented):
Looks great! Thanks for doing this :rocket:

rest/model/notifications.go
ablack12: I think there's no point in implementing functions that aren't used anywhere, since we aren't trying to implement the interface.

rest/model/user.go
ablack12: I started removing this but I think it actually makes sense here. If these values don't match the values we have defined for UserSubscriptionPreference, then we would have an issue. I'm okay with some BuildFromService/ToService functions still returning errors, since I think the nice thing about not having these functions implement an interface is that they can be what we need them to be. I'll refactor this to make it cleaner though.

rest/model/notifications.go
hadjri: Fair, sgtm

rest/model/user.go
hadjri: Perhaps we'd want to name them something like `BuildFromServiceAndValidate` in these cases where we can return an error. Although I'm not extremely opinionated about it

ablack12: I think I'm fine with the error since it's more that we just _can't_ buildFromService if the types aren't right, in this case, not that the input isn't "correct" in terms of what we want.

hadjri (approved):

PR #5785, opened 2022-07-07
maybe

bynn:

### Description 

### Testing 

### DIFF
--- a/graphql/version_resolver.go
+++ b/graphql/version_resolver.go
@@ -69,9 +69,6 @@ func (r *versionResolver) BuildVariants(ctx context.Context, obj *restModel.APIV
 		obj.Activated = version.Activated
 	}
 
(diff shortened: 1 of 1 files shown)


khelif96:
Probably

ZackarySantana:
sometimes

graphql/version_resolver.go
```diff
@@ -69,9 +69,6 @@ func (r *versionResolver) BuildVariants(ctx context.Context, obj *restModel.APIV
 		obj.Activated = version.Activated
 	}
 
-	if !utility.FromBoolPtr(obj.Activated) {
```
khelif96, line 72: I don't think this is the right approach. This is an optimization to avoid over fetching tasks when there shouldn't be any.

The reason the tasks are failing is because the activated flag for the version is being set to false at some point even though there are tasks.

PR #5788, opened 2022-07-07
EVG-16147 Add patch route for versions

bynn:
[EVG-16147](https://jira.mongodb.org/browse/EVG-16147)

### Description 
Problem was that unschedule all tasks was calling the set version activation function with active=false. 
That caused the version activation to be set to false.
The active field in version is supposed to be a cached value of if the version has run ANY tasks at all so it should have never been possible to set that field from true to false as it was doing in this patch. 
I removed that change and the field can now only change if you set it to active.  

### Testing 


khelif96 (approved):
Graphql part lgtm

model/lifecycle.go
```diff
@@ -55,9 +55,6 @@ type VersionToRestart struct {
 // SetVersionActivation updates the "active" state of all builds and tasks associated with a
 // version to the given setting. It also updates the task cache for all builds affected.
 func SetVersionActivation(versionId string, active bool, caller string) error {
```
ablack12, line 57: Doesn't seem like we need this function at all since it just calls setTaskActivationForVersion. Should we just remove that private function and move the logic in here?

ablack12 (changes requested):

ablack12 (approved):

PR #5789, opened 2022-07-07
EVG-17010 Fix "host not assigned to run task" errors

malikchaya2:
[EVG-17010 ](https://jira.mongodb.org/browse/EVG-17010 )

### Description 
The fact that the “host not assigned to run tasks” errors are coming from the /files endpoint is a red herring. Based on both Splunk logs for the host and host event logs, it’s task timeouts that are the Issue: the host ran the older task, timed out, then ran the newer task. The /files endpoint is simply the last thing to run on the task so it’s what was likely running when they timed out.

If the agent has already moved on to a new task we don't want it to keep doing things for a previous task, because that can cause issues for later tasks. We therefore want the agent s3 commands to respect the context passed to them rather than use context.Background(). That way, if it's making requests to the S3 API, the app server won't know that it's doing that, and the agent would have to stop doing it on its own.

### DIFF
--- a/service/api.go
+++ b/service/api.go
@@ -159,27 +159,38 @@ func (as *APIServer) requireProject(next http.HandlerFunc) http.HandlerFunc {
 
 func (as *APIServer) requireHost(next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		h, code, err := model.ValidateHost(gimlet.GetVars(r)["hostId"], r)
+		h, _, err := model.ValidateHost(gimlet.GetVars(r)["hostId"], r)
+		UpdateHost(h)
 		if err != nil {
-			as.LoggedError(w, r, code, errors.Wrap(err, "host not assigned to run task"))
-			return
-		}
-		// update host access time
-		if err := h.UpdateLastCommunicated(); err != nil {
-			grip.Warningf("Could not update host last communication time for %s: %+v", h.Id, err)
-		}
-		// Since the host has contacted the app server, we should prevent the
-		// app server from attempting to deploy agents or agent monitors.
-		// Deciding whether or not we should redeploy agents or agent monitors
-		// is handled within the REST route handler.
-		if h.NeedsNewAgent {
-			grip.Warning(message.WrapError(h.SetNeedsNewAgent(false), "problem clearing host needs new agent"))
-		}
-		if h.NeedsNewAgentMonitor {
-			grip.Warning(message.WrapError(h.SetNeedsNewAgentMonitor(false), "problem clearing host needs new agent monitor"))
+			// validate the host again after update in case the agent and app server were out of sync,
+			// and the update fixed it
+			var code int
+			h, code, err = model.ValidateHost(gimlet.GetVars(r)["hostId"], r)
+			if err != nil {
+				as.LoggedError(w, r, code, errors.Wrap(err, "host not assigned to run task"))
+				return
+			}
+
 		}
 		r = setAPIHostContext(r, h)
 		next(w, r)
+	}
+}
+
+func UpdateHost(h *host.Host) {
+	// update host access time
+	if err := h.UpdateLastCommunicated(); err != nil {
+		grip.Warningf("Could not update host last communication time for %s: %+v", h.Id, err)
+	}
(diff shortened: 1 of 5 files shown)


service/api.go
```diff
@@ -159,30 +159,41 @@ func (as *APIServer) requireProject(next http.HandlerFunc) http.HandlerFunc {
...
-		// app server from attempting to deploy agents or agent monitors.
-		// Deciding whether or not we should redeploy agents or agent monitors
-		// is handled within the REST route handler.
-		if h.NeedsNewAgent {
-			grip.Warning(message.WrapError(h.SetNeedsNewAgent(false), "problem clearing host needs new agent"))
-		}
-		if h.NeedsNewAgentMonitor {
-			grip.Warning(message.WrapError(h.SetNeedsNewAgentMonitor(false), "problem clearing host needs new agent monitor"))
+			// validate the host again after update in case the agent and app server were out of sync,
+			// and the update fixed it
```
ybrill, line 166: I don't understand: how does `UpdateHost` have a bearing on the task the host is assigned?

ybrill (commented):
I  don't know if it's significant, but it's interesting to me that we hit this error [precisely 51 times for each task](https://mongodb.splunkcloud.com/en-US/app/search/search?s=%2FservicesNS%2Fnobody%2Fsearch%2Fsaved%2Fsearches%2FEvergreen%253A%2520Error%2520rate%2520in%2520production&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=&q=search%20index%3Devergreen%20metadata.level%20%3E%3D%2070%20code%3D409%20%22host%20not%20assigned%20to%20run%20task%22%20%0A%7C%20stats%20count%20by%20url%0A%7C%20sort%20-count&earliest=1654099200&latest=1654102800&display.page.search.tab=statistics&display.general.type=statistics&display.prefs.statistics.offset=0&sid=1657289830.2475242). Does it have to do with the number of times we retry or something?

service/api.go
malikchaya2: I don't know that it would. I was going off of the comment that `This may mean that "an agent has a different belief about what task it should be running than the app server does."` but it may not actually help at all.

malikchaya2: the 51 is indeed interesting. I also see 41 though which makes it even more confusing.

ablack12:
I'm not sure retrying validation is going to help, since we already attempt to validate many times. It could be interesting to look into some of the tasks that ran into this error, and the hosts they tried to run on -- what kind of task did the host run before? Was it terminated after? Were they all the same distro? Were they during a deploy?

malikchaya2:
> I'm not sure retrying validation is going to help, since we already attempt to validate many times. It could be interesting to look into some of the tasks that ran into this error, and the hosts they tried to run on -- what kind of task did the host run before? Was it terminated after? Were they all the same distro? Were they during a deploy? We could also trace the API calls themselves of both the host and the task during that time frame to see if that surfaces anything.

My initial investigation didn't yield anything fruitful, but I'll try again with focus on the questions you posed.

service/api.go
malikchaya2: Thanks to Kim's digging, we have an hypothesis for why it's 51. 
[the agent's client tries 10 times per ](https://github.com/evergreen-ci/evergreen/blob/cda1c49a668203306b3d09a27b1ffc15c62857ce/agent/internal/client/base_client.go#L55-L59)AttachFiles[ call](https://github.com/evergreen-ci/evergreen/blob/cda1c49a668203306b3d09a27b1ffc15c62857ce/agent/internal/client/base_client.go#L55-L59), and maxS3OpAttempts[ can call ](https://github.com/evergreen-ci/evergreen/blob/cda1c49a668203306b3d09a27b1ffc15c62857ce/agent/command/s3_util.go#L16)AttachFiles[ 5 times](https://github.com/evergreen-ci/evergreen/blob/cda1c49a668203306b3d09a27b1ffc15c62857ce/agent/command/s3_util.go#L16) in the outer retry loop for the s3 commands, so 5 * 10 = 50. 
The one extra one is a mystery.

ybrill: nice 🎉

malikchaya2:
This is ready for re-review. I have updated the PR description and title as well.

ablack12 (approved):

PR #5793, opened 2022-07-08
EVG-14787 Support inline task groups for variants

hadjri:
[EVG-14787](https://jira.mongodb.org/browse/EVG-14787)

### Description 
Task groups can now be directly defined either fully or by implementing a template within the build variants in a project config YAML.  Build variant tasks now have an optional `Group` field to hold task group data. This should improve modularity of large config files that currently have to define many different task groups that all follow a similar template. Will update the wiki with details on how to configure inline task groups once this receives approval.

### Testing 
Added unit test cases confirming inline task groups are parsed correctly. Configured an inline task group within build variants in the sandbox project and tested the changes in staging. Relevant screenshots attached.


hadjri:
Screen Shot 2022-07-12 at 5 33 51 PM

Patch was submitted on staging scheduling all tasks for the variant and succeeded. The two inline task group tasks (validate_commit_message and validate_commit_message2 both ran on the same host and have their task group fields set): https://spruce-staging.corp.mongodb.com/version/62cde6c597b1d3346726fe3f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

malikchaya2 (approved):

model/project.go
```diff
@@ -1173,7 +1179,10 @@ func (m *Module) GetRepoOwnerAndName() (string, string) {
 }
 
 // FindTaskGroup returns a specific task group from a project
-func (p *Project) FindTaskGroup(name string) *TaskGroup {
+func (p *Project) FindTaskGroup(name string, group *TaskGroup) *TaskGroup {
+	if group != nil {
```
ablack12, line 1183: Why have this be part of it? It seems like we just shouldn't call FindTaskGroup if group != nil

ablack12: It seems like FindTaskGroup should do the extra work of finding the task group in the variant list if it's not in the task group list, rather than doing it outside of this.

model/project.go
```diff
@@ -137,6 +137,8 @@ type BuildVariantTaskUnit struct {
 	CronBatchTime string `yaml:"cron,omitempty" bson:"cron,omitempty"`
 	// If Activate is set to false, then we don't initially activate the task.
 	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
+	// Group is set if an inline task group is defined on the build variant.
+	Group *TaskGroup `yaml:"group,omitempty" bson:"group,omitempty"`
```
ablack12, line 141: Can we call this task_group just to be consistent with what users define at the top-level?

validator/project_validator.go
```diff
@@ -637,6 +637,9 @@ func ensureReferentialIntegrity(project *model.Project, containerNameMap map[str
 	for _, buildVariant := range project.BuildVariants {
 		buildVariantTasks := map[string]bool{}
 		for _, task := range buildVariant.Tasks {
+			if task.Group != nil {
```
ablack12, line 640: We should check all the tasks in the task group like we do at 630 - 633 i think.

ablack12 (changes requested):
Hype about this!

model/project.go
```diff
@@ -138,7 +138,7 @@ type BuildVariantTaskUnit struct {
 	// If Activate is set to false, then we don't initially activate the task.
 	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
 	// Group is set if an inline task group is defined on the build variant.
-	Group *TaskGroup `yaml:"group,omitempty" bson:"group,omitempty"`
+	Group *TaskGroup `yaml:"task_group,omitempty" bson:"task_group,omitempty"`
```
ablack12, line 141: Sorry I wasn't clear but preferably the variable name would match the tags, i.e. TaskGroup.

validator/project_validator.go
```diff
@@ -618,6 +618,35 @@ func checkProjectFields(project *model.Project) ValidationErrors {
...
+				Message: fmt.Sprintf("buildvariant '%s' references a non-existent task '%s'",
+					variant, task),
+				Level: Error,
+			})
+		}
+	}
+	if _, ok := taskGroupTaskSet[task]; ok {
+		errs = append(errs,
+			ValidationError{
+				Message: fmt.Sprintf("task '%s' in build variant '%s' is already referenced in task group '%s'",
```
ablack12, line 642: Given the rational for this error in the original ticket EVG-6846, I don't think we'd want to give this warning for inline task group tasks passed in, since they _are _correctly being used in a task group.

ablack12 (changes requested):

model/project.go
hadjri: Ahh I see, apologies fixed now

validator/project_validator.go
hadjri: Good catch! Fixing.

model/project.go
```diff
@@ -866,12 +872,12 @@ func NewPatchTaskIdTable(proj *Project, v *Version, tasks TaskVariantPairs, proj
 	}
 	execTasksWithTaskGroupTasks := TVPairSet{}
 	for _, vt := range tasks.ExecTasks {
-		if _, ok := tgMap[vt.TaskName]; ok {
-			if tg := proj.FindTaskGroup(vt.TaskName); tg != nil {
-				for _, t := range tg.Tasks {
-					execTasksWithTaskGroupTasks = append(execTasksWithTaskGroupTasks, TVPair{vt.Variant, t})
-				}
+		bvt := proj.FindTaskForVariant(vt.TaskName, vt.Variant)
+		if tg := proj.FindTaskGroup(vt.TaskName); tg != nil {
```
ablack12, line 876: Sorry I'm noticing this so late. I'm slightly concerned about this; I think the purpose of having tgMap in the first place was to limit how often we need to loop through all of the task groups, and. this change would mean that for _every_ task we would loop through all task groups AND all tasks in all variants. Can we check tgMap and bvt.TaskGroup instead to avoid calling FindTaskGroup unless necessary?

model/project_parser.go
```diff
@@ -1130,7 +1132,11 @@ func evaluateBVTasks(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluator, vse
 				isGroup = true
 			}
 		}
-		if err1 != nil && err2 != nil {
+		if pbvt.TaskGroup != nil {
+			names = append(names, pbvt.Name)
```
ablack12, line 1136: Can you explain why we add name specifically here? (Would this not be covered by 1125 or 1129)?

model/project_parser.go
```diff
@@ -1130,7 +1132,11 @@ func evaluateBVTasks(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluator, vse
...
+			isGroup = true
+		}
+		if err1 != nil && err2 != nil && !isGroup {
```
ablack12, line 1139: Is the thought that err1 and err2 are going to be not nil if pbvt.TaskGroup != nil? Because in that case we just shouldn't check those cases.

ablack12 (approved):

ablack12 (changes requested):
Sorry for the false alarm wrong button click 😓 but I think this is close!

model/project_parser.go
hadjri: The above checks on 1125 and 1129 are checking the defined `tasks` and `task_groups` on the config.  Since now we can define a whole new task group inline in the build variants, the name of that group won't be found in either the `tasks` or `task_groups`. So to ensure these inline groups get added to the names list I've added this if statement.

ablack12 (approved):

PR #5794, opened 2022-07-11
EVG-16980 Update Archive/ArchiveMany to account for display tasks with failed only

ZackarySantana:
[EVG-16980](https://jira.mongodb.org/browse/EVG-16980)

### Description 
Archive (and ArchiveMany) do not account for ResetFailedWhenFinished. When a display task is Archived with 'ResetFailedWhenFinished' true, only the failed execution tasks should be restarted. This PR adds that functionality, adds tests to ensure that a 'gap' is left like described in the ticket.

### Testing 
Added unit testing to ensure Archive/ArchiveMany works as intended when 'ResetFailedWhenFinished' is true, and then a subsequent call when it is false. Testing on staging will be done soon (TBA).


model/task/task.go
```diff
@@ -2573,20 +2587,37 @@ func ArchiveMany(tasks []Task) error {
...
-			},
-		},
-			bson.M{
-				"$unset": bson.M{
-					AbortedKey:   "",
-					AbortInfoKey: "",
+		if execution == -1 {
+			_, err = taskColl.UpdateMany(ctx, bson.M{
+				IdKey: bson.M{
+					"$inc": taskIds,
```
bynn, line 2593: is the `inc` here and `in` on line 2601 supposed to be like that?

model/task/task.go
```diff
@@ -2522,7 +2534,9 @@ func (t *Task) Archive() error {
 	return nil
 }
 
-func ArchiveMany(tasks []Task) error {
+// Execution argument is for display tasks to override the natural incrementing
+// when archiving. To incremenent all executions, put execution = -1
+func ArchiveMany(tasks []Task, execution int) error {
```
bynn, line 2539: if we want to specify that the execution tasks have the same execution as their display task, but be incremented for all other tasks, would make sense to make a second function like ArchiveManyExecutionTasks() instead of passing in a -1 every other time?

model/task/task_test.go
```diff
@@ -4139,6 +4289,7 @@ type TaskConnectorFetchByIdSuite struct {
 	suite.Suite
 }
 
+// TODO RUN THIS TEST TO TEST
```
bynn, line 4292: delete todo!

model/task/task_test.go
```diff
@@ -4251,6 +4402,9 @@ func (s *TaskConnectorFetchByIdSuite) TestFindByVersion() {
 	s.Equal(evergreen.TaskFailed, t[0].DisplayStatus)
 }
 
+// TODO
```
bynn, line 4405: same here!

model/task/task_test.go
```diff
@@ -3669,6 +3669,156 @@ func TestArchive(t *testing.T) {
...
+		require.Equal(t, 1, dt.Execution)
+
+		checkTaskIsArchived(t, archivedT1)
+		checkTaskIsNotArchived(t, t2.Id, 0)
+		checkTaskIsArchived(t, archivedDisplayTaskID)
+
+		checkEventLogHostTaskExecutions(t, hostID, archivedT1, archivedExecution)
+	})
+
+	// TODO
```
bynn, line 3776: here!

bynn (changes requested):

model/task/task.go
ZackarySantana: Would it be good to have AchiveManyOverrideExecution(tasks []Tasks, execution int) as a good description for how that function differs from the last one? I was also thinking having ArchiveMany call this with -1 and keep the logic that it has? Or separating the logic (pro more clear, con more copy/paste code)?

model/task/task.go
```diff
@@ -2472,12 +2472,24 @@ func (t *Task) Insert() error {
...
-		}
-		if err = ArchiveMany(execTasks); err != nil {
-			return errors.Wrap(err, "archiving execution tasks")
+		if !t.ResetFailedWhenFinished {
+			execTasks, err := FindAll(db.Query(ByIds(t.ExecutionTasks)))
+
+			if err != nil {
+				return errors.Wrap(err, "retrieving execution tasks")
+			}
+			if err = ArchiveMany(execTasks, t.Execution); err != nil {
```
ablack12: Would be nice to call this outside of the if/else cases instead and just assign Find() to a neutral execTasksToArchive or something (will need to create the variable outside of the if/else statements to use the variable outside also)

model/task/task.go
```diff
@@ -2573,20 +2587,37 @@ func ArchiveMany(tasks []Task) error {
...
 				},
-				"$inc": bson.M{
-					ExecutionKey: 1,
+			},
+				bson.M{
+					"$unset": bson.M{
+						AbortedKey:   "",
+						AbortInfoKey: "",
+					},
+					"$in": bson.M{
```
ablack12: This looks like a typo

model/task/task.go
```diff
@@ -2573,20 +2587,37 @@ func ArchiveMany(tasks []Task) error {
...
-		_, err = taskColl.UpdateMany(ctx, bson.M{
-			IdKey: bson.M{
-				"$in": taskIds,
-			},
-		},
-			bson.M{
-				"$unset": bson.M{
-					AbortedKey:   "",
-					AbortInfoKey: "",
+		if execution == -1 {
```
ablack12: I'm not sure we need this if/else statement. In the if case we have inc, in the else statement we have set "execution + 1" which is what $inc means.

model/task/task.go
ZackarySantana: The error messages are slightly different to different, would you recommend merging the error messages to the same to do this?

model/task/task.go
```diff
@@ -2522,7 +2534,9 @@ func (t *Task) Archive() error {
 	return nil
 }
 
-func ArchiveMany(tasks []Task) error {
+// Execution argument is for display tasks to override the natural incrementing
```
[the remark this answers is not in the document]
ZackarySantana: Fixed!

model/task/task.go
ZackarySantana: Definitely was

model/task/task.go
ZackarySantana: The logic was largely rewritten to not depend on any magic numbers, booleans, or anything. Using Archive and ArchiveMany should now work as intended for different use cases that we use in the codebase (restart version, etc).

model/task/task.go
```diff
@@ -2472,75 +2474,114 @@ func (t *Task) Insert() error {
...
-	}
-	t.Aborted = false
-	return nil
+	return ArchiveTask(*t)
 }
 
+// ArchiveMany accepts tasks and display tasks (execution tasks are ignored) and calls archive on all of them.
+// Returning if any error. For tasks, it will bundle them together to execute one query. It also archives the display tasks.
+// It expects all tasks inside are NOT execution tasks.
 func ArchiveMany(tasks []Task) error {
```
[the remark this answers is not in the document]
ZackarySantana: I've just implemented the helper function that will reduce the queries to the optimized version I mentioned at the end of the above comment!

ZackarySantana:
@ablack12 Just to update you, this ticket is going to move in the following direction now: Adding a new variable to the task struct / collection (default being current execution) called something like "LatestParentExecution". And updating ArchiveMany to use one transaction, two queries (one archiving to old tasks collection, one updating tasks collection using LatestParentExecution)

ZackarySantana:
evergreen retry

model/task/task.go
```diff
@@ -2558,35 +2577,67 @@ func ArchiveMany(tasks []Task) error {
...
-				},
-				"$inc": bson.M{
-					ExecutionKey: 1,
+		if len(tasksIds) > 0 {
+			// TODO (EVG-17322): Remove RunCommand call and replace with aggregate + merge
+			result := evergreen.GetEnvironment().DB().RunCommand(sessCtx,
+				bson.D{
+					bson.E{Key: "update", Value: Collection},
+					bson.E{Key: "updates", Value: bson.A{bson.M{
+						"q": bson.D{{Key: "_id", Value: bson.D{{Key: "$in", Value: tasksIds}}}},
```
bynn, line 2594: would it be possible to build the aggregation like this and/or have more comments on it? 
https://github.com/evergreen-ci/evergreen/blob/1b36641d314cbf165cb1e36bb917a1a8d627520b/model/task_history.go#L559

model/task_lifecycle_test.go
```diff
@@ -37,6 +37,8 @@ var (
 func TestSetActiveState(t *testing.T) {
 	Convey("With one task with no dependencies", t, func() {
 		require.NoError(t, db.ClearCollections(task.Collection, build.Collection, task.OldCollection, VersionCollection))
+		require.NoError(t, db.CreateCollections(task.Collection, task.OldCollection))
```
bynn, line 40: no need to explicitly create task collections if we insert a task 
but it might be good to clear old collection

model/task/task_test.go
```diff
@@ -3669,6 +3669,156 @@ func TestArchive(t *testing.T) {
 	}
 }
 
+func TestArchiveFailedOnly(t *testing.T) {
+	defer func() {
+		assert.NoError(t, db.ClearCollections(Collection, OldCollection, event.LegacyEventLogCollection))
+	}()
+
+	resetDatabase := func() Task {
```
bynn, line 3677: would it be possible to split the test into multiple unit tests instead of calling reset database?

model/task/task_test.go
```diff
@@ -3669,6 +3669,156 @@ func TestArchive(t *testing.T) {
...
+		require.Equal(t, 1, dt.Execution)
+
+		checkTaskIsArchived(t, archivedT1)
+		checkTaskIsNotArchived(t, t2.Id, 0)
+		checkTaskIsArchived(t, archivedDisplayTaskID)
+
+		checkEventLogHostTaskExecutions(t, hostID, archivedT1, archivedExecution)
+	})
+
+	// This test is for the edge case of a archiving with only failed execution tasks, then archiving all execution tasks
```
bynn, line 3777: can you test this behavior with other tasks in here as well as the display task?

bynn (changes requested):

model/task/task.go
```diff
@@ -2471,83 +2472,101 @@ func (t *Task) Insert() error {
...
-			return errors.Wrap(err, "archiving execution tasks")
+		err = UpdateOne(
+			bson.M{IdKey: t.Id},
+			bson.M{
+				"$unset": bson.M{
+					AbortedKey:              "",
+					AbortInfoKey:            "",
+					OverrideDependenciesKey: "",
+				},
+				"$set": bson.M{ExecutionKey: t.Execution + 1, LatestParentExecutionKey: t.Execution + 1},
```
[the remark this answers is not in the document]
ZackarySantana: For tasks, their execution and parent execution should be the same. Since this isn't using the pipeline code, it doesn't have the set of setting latest parent execution to the conditional: if latest parent execution exists, then latest parent execution += 1 else, latest parent execution = execution + 1.

Essentially, it is using execution for backwards compatability and using the fact that they should be equal always for tasks.

model/task/task.go
ZackarySantana: Yes, I've added some comments and I think it helps explain the three steps in the pipeline very well!

model/task/task_test.go
ZackarySantana: The reset database call only happens once- I used it to help differentiate between what code is doing what. The tests are connected to each other and are to test the compounding of running archive over a couple iterations.

model/task/task_test.go
ZackarySantana: Yes, let me add some more tasks!

model/task_lifecycle_test.go
ZackarySantana: Fixed!

model/task/task_test.go
ZackarySantana: (to clarify, I'll make a new test for this because this is specifically testing calling archive on a display task

model/task/task_test.go
bynn: i think its a little strange that a function called resetDatabases arbitrarily returns the display task so would it be possible to not return the display task in that case?

model/task/task_test.go
bynn: im not sure where the newly added test is that looks at t3 and t4

model/task_lifecycle_test.go
```diff
@@ -822,6 +824,8 @@ func TestUpdateVersionStatusForGithubChecks(t *testing.T) {
 
 func TestUpdateBuildAndVersionStatusForTaskAbort(t *testing.T) {
 	require.NoError(t, db.ClearCollections(task.Collection, task.OldCollection, build.Collection, VersionCollection, event.LegacyEventLogCollection))
+	require.NoError(t, db.CreateCollections(task.Collection, task.OldCollection))
```
bynn, line 827: same comment about creating task collection for all the other create collections function you added

bynn (changes requested):

model/task/task.go
```diff
@@ -2470,83 +2471,103 @@ func (t *Task) Insert() error {
...
 	for _, t := range tasks {
-		// For any display tasks here, make sure that we also archive all their execution tasks.
-		// Consider each execution tasks individually, so we don't query a task we already passed in.
-		for _, et := range t.ExecutionTasks {
-			if !existingTasksMap[et] {
-				additionalTasks = append(additionalTasks, et)
-				existingTasksMap[et] = true
+		bundledTasks = append(bundledTasks, t.Id)
+
+		if !t.ResetFailedWhenFinished || evergreen.IsFailedTaskStatus(t.Status) {
```
ablack12, line 2523: Same question here and below.

model/task/task.go
```diff
@@ -2557,35 +2578,67 @@ func ArchiveMany(tasks []Task) error {
...
+					}}},
 				},
-			})
-		if err != nil {
-			return nil, err
+			)
+			return nil, errors.Wrap(result.Err(), "Error updating documents")
 		}
-		return nil, err
+		return nil, errors.Wrap(err, "updating tasks.")
```
ablack12, line 2635: nit, no periods in errors

ablack12 (changes requested):

model/task/task.go
```diff
@@ -2470,83 +2471,103 @@ func (t *Task) Insert() error {
...
-		if err = ArchiveMany(execTasks); err != nil {
-			return errors.Wrap(err, "archiving execution tasks")
+	if t.DisplayOnly {
+		var task []Task
+		task = append(task, *t)
+		err := ArchiveMany(task)
+		return errors.Wrapf(err, "archiving and updating display task with ID '%s'.", t.Id)
+	} else {
+		// Archiving a single task.
+		if !t.ResetFailedWhenFinished || evergreen.IsFailedTaskStatus(t.Status) {
```
ablack12, line 2481: I think when we consider ResetFailed we should also consider ResetWhenFinished, maybe making a helper function or etc (although I still think this isn't the right place to consider it I'm not sure where the right place is for this comment). it's possible that we set ResetFailedWhenFinished bc an execution task was stranded for example, but a user chose to restart the whole thing, i.e. ResetWhenFinished was set. This should take precedence.

model/task/task.go
```diff
@@ -2557,35 +2578,67 @@ func ArchiveMany(tasks []Task) error {
...
-			bson.M{
-				"$unset": bson.M{
-					AbortedKey:   "",
-					AbortInfoKey: "",
-				},
-				"$inc": bson.M{
-					ExecutionKey: 1,
+		if len(tasksIds) > 0 {
+			// TODO (EVG-17322): Remove RunCommand call and replace with aggregate + merge
+			result := evergreen.GetEnvironment().DB().RunCommand(sessCtx,
```
ablack12, line 2591: I discussed with Bynn but I'm concerned about introducing this because A) it introduces tech debt since it’s a bit harder to read and is very specific, and B) it’s likely much more expensive than what we have now, which is already expensive.

I'm not convinced we need to do any of the backwards compatible calls since this change will be baked in for a few months before users can start. Really what we need to do is:
1) unset anything that needs to be unset
2) Update execution to execution+1 for non-exec tasks, and LPE+1 for exec tasks (could probably use aggregation stage $project or something), 
3) Update LPE to be LPE +1 for exec tasks.

Seems like we could have 1 call for non-exec and 1 call for exec, or do a $split aggregation stage. It's not clear to me from your ticket what $merge is needed for but bynn  and/or i will sync with you.

model/context_test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 func TestLoadContext(t *testing.T) {
 	require.NoError(t, db.ClearCollections(task.Collection, task.OldCollection, ProjectRefCollection))
+	require.NoError(t, db.CreateCollections(task.Collection, task.OldCollection))
```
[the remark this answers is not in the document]
ZackarySantana: The run command function runs in a transaction. Transactions cannot create collections in our version of MongoDB.

https://www.mongodb.com/docs/upcoming/core/transactions/#:~:text=Starting%20in%20MongoDB%204.4%2C%20you%20can%20create%20collections%20and%20indexes,read%20concerns%20available%20for%20transactions.

model/task/task.go
```diff
@@ -2470,83 +2471,103 @@ func (t *Task) Insert() error {
...
 func (t *Task) Archive() error {
-	if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
-		execTasks, err := FindAll(db.Query(ByIds(t.ExecutionTasks)))
-		if err != nil {
-			return errors.Wrap(err, "retrieving execution tasks")
-		}
-		if err = ArchiveMany(execTasks); err != nil {
-			return errors.Wrap(err, "archiving execution tasks")
+	if t.DisplayOnly {
+		var task []Task
```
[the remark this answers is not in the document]
ZackarySantana: Thank you, oversight will change!

model/task/task.go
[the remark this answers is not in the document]
ZackarySantana: This else block covers non-display tasks, non-execution tasks. This was so a restart when failed at a version level could be done with proper archiving of successful results.

If ResetFailedWhenFInished is false (reset all, including successful), the "!" will make it true and it will update it. If it is true, it will only update it if the task is a failed status.

model/task/task.go
ZackarySantana: Got it!

model/context_test.go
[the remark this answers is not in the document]
ablack12: Right that's reasonable but I'm confused about what's changed. The previous transaction code _also_ handled both the task collection and the old task collection but didn't need this.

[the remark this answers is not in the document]
ZackarySantana: Hm, I'll do more testing without it to verify if it was a fix required only for an older commit or newer (after our sync)

model/task/task.go
```diff
@@ -2470,83 +2471,103 @@ func (t *Task) Insert() error {
...
+			}
+			err = UpdateOne(
+				bson.M{IdKey: t.Id},
+				bson.M{
+					"$unset": bson.M{
+						AbortedKey:              "",
+						AbortInfoKey:            "",
+						OverrideDependenciesKey: "",
+					},
+					"$set": bson.M{ExecutionKey: t.Execution + 1, LatestParentExecutionKey: t.Execution + 1},
```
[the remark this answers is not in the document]
ZackarySantana: Yes, fixed!

model/task/task.go
```diff
@@ -2486,82 +2487,90 @@ func (t *Task) Insert() error {
...
+		bundledTasks = append(bundledTasks, t.Id)
+		toArchive = append(toArchive, t.makeArchivedTask())
+		if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
+			eTasks, err := FindAll(db.Query(ByIds(t.ExecutionTasks)))
+			if err != nil {
+				return errors.Wrapf(err, "finding execution tasks for display task '%s'.", t.Id)
+			}
+			for _, et := range eTasks {
+				execTasks = append(execTasks, et.Id)
+				if shouldRestartTask(t.ResetFailedWhenFinished, t.ResetWhenFinished, et.Status) {
```
bynn, line 2548: were already at a stage where the reset was called 
so we dont really have to check if a task should be restarted at this point 

i would recommend querying for all failed tasks on line 2542 if resetwhenfailed is true and leaving it be otherwise and appending all of them at once

model/task/task.go
```diff
@@ -2572,35 +2581,90 @@ func ArchiveMany(tasks []Task) error {
...
+					bson.D{{Key: "$set", Value: bson.D{ // Execution = LPE
+						{Key: ExecutionKey, Value: "$" + LatestParentExecutionKey}}},
+					},
+					bson.D{{Key: "$unset", Value: bson.A{
+						AbortedKey,
+						AbortInfoKey,
+						OverrideDependenciesKey,
+					}}}})
+
+			return nil, errors.Wrap(err, "Error updating documents")
```
bynn, line 2660: "Updating documents" to follow error convention

bynn (changes requested):

model/task/task.go
```diff
@@ -2486,82 +2487,90 @@ func (t *Task) Insert() error {
...
+			bson.M{IdKey: t.Id},
+			bson.M{
+				"$unset": bson.M{
+					AbortedKey:              "",
+					AbortInfoKey:            "",
+					OverrideDependenciesKey: "",
+				},
+				"$set": bson.M{
+					ExecutionKey:             t.Execution + 1,
+					LatestParentExecutionKey: t.Execution + 1, // Will this matter for regular tasks?
```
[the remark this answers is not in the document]
ZackarySantana: Yeah, I added this a while back as a reminder to remove it, gonna remove it!

model/task/task.go
```diff
@@ -2486,82 +2487,90 @@ func (t *Task) Insert() error {
...
-			},
-			"$inc": bson.M{ExecutionKey: 1},
-		})
-	if err != nil {
-		return errors.Wrap(err, "updating task")
-	}
-	t.Aborted = false
-	return nil
+func shouldRestartTask(resetFailedWhenFinished bool, resetWhenFinished bool, status string) bool {
+	return resetWhenFinished || !resetFailedWhenFinished || evergreen.IsFailedTaskStatus(status)
```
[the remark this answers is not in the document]
ZackarySantana: If resetFailedWhenFinished is false, that means it should always be reset. If it is true, it should only be reset if evergreen.IsFailedTaskStatus(status) is true

model/task/task.go
ZackarySantana: Does the db query for failed exec tasks check for the same status as the evergreen.IsfailedTaskStatus call (going to follow up on this thread once I research)? If they're different, I wonder which set of exec tasks we want

model/task/task.go
ZackarySantana: Updated! I also updated other areas where it started with error

model/task/task.go
```diff
@@ -2486,82 +2487,90 @@ func (t *Task) Insert() error {
...
-				OverrideDependenciesKey: "",
-			},
-			"$inc": bson.M{ExecutionKey: 1},
-		})
-	if err != nil {
-		return errors.Wrap(err, "updating task")
-	}
-	t.Aborted = false
-	return nil
+func shouldRestartTask(resetFailedWhenFinished bool, resetWhenFinished bool, status string) bool {
```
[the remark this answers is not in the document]
ZackarySantana: Hm, but this is used for execution tasks. The two first bools are taken from the display task. So it would still be task.shouldRestartTask(bool, bool). If the query for exec tasks changes directly to failed, this function would be deleted- so first I'll check in the the query

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
+				"old_task_id":     archiveTask.OldTaskId,
+				"execution":       t.Execution,
+				"display_only":    t.DisplayOnly,
+			}))
+			return errors.Wrap(err, "inserting archived task into old tasks")
 		}
-		if err = ArchiveMany(execTasks); err != nil {
-			return errors.Wrap(err, "archiving execution tasks")
+		err = UpdateOne(
+			bson.M{IdKey: t.Id},
```
ablack12, line 2506: Since we aren't using $inc anymore, should we include ExecutionKey in the update to ensure that we're updating what we think we are?

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
-				OverrideDependenciesKey: "",
-			},
-			"$inc": bson.M{ExecutionKey: 1},
-		})
-	if err != nil {
-		return errors.Wrap(err, "updating task")
-	}
-	t.Aborted = false
-	return nil
+func shouldRestartTask(resetFailedWhenFinished bool, resetWhenFinished bool, status string) bool {
```
ablack12, line 2525: I know this is a private function but let's still add a comment to explain the inputs since the use case is niche.

Additionally, since this is it's own function, it would be cleaner to break these into logically distinct pieces that can be commented to explain meaning. 

```
if resetWhenFinished { return true }
if resetFailed && isFailed {return true}
else return false 
```
is a clearer way of breaking this up

model/task/task.go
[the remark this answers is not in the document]
ablack12: If resetFailedWhenFinished is false, then it should only be reset if resetWhenFinished is true. Possibly we just don't want to restart this display task at all, right?

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
+	return resetWhenFinished || !resetFailedWhenFinished || evergreen.IsFailedTaskStatus(status)
 }
 
+// ArchiveMany accepts tasks and display tasks (no execution tasks), bundles them all up, archives and updates all
+// that need to be (accounting for ResetFailedWhenFinished) in two queries
```
ablack12, line 2530: accounting for tasks that should only be reset if failed

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
-	}
-	existingTasksMap := map[string]bool{}
-	for _, t := range tasks {
-		existingTasksMap[t.Id] = true
-	}
-	additionalTasks := []string{}
+	bundledTasks := []string{}        // Contains all tasks and display tasks
+	execTasks := []string{}           // Contains all exec tasks
+	toUpdateExecTaskIds := []string{} // Contains all exec tasks that should update and have new execution
+	toArchive := []interface{}{}      // Contains all archived tasks that are being updated / have new execution
```
ablack12, line 2535: why interface?

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
 func ArchiveMany(tasks []Task) error {
-	if len(tasks) == 0 {
-		return nil
-	}
-	existingTasksMap := map[string]bool{}
-	for _, t := range tasks {
-		existingTasksMap[t.Id] = true
-	}
-	additionalTasks := []string{}
+	bundledTasks := []string{}        // Contains all tasks and display tasks
```
ablack12, line 2532: allTasks?

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
-		for _, et := range t.ExecutionTasks {
-			if !existingTasksMap[et] {
-				additionalTasks = append(additionalTasks, et)
-				existingTasksMap[et] = true
+		bundledTasks = append(bundledTasks, t.Id)
+		toArchive = append(toArchive, t.makeArchivedTask())
+		if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
+			eTasks, err := FindAll(db.Query(ByIds(t.ExecutionTasks)))
+			if err != nil {
+				return errors.Wrapf(err, "finding execution tasks for display task '%s'.", t.Id)
```
ablack12, line 2543: no periods in errors

model/task/task.go
ablack12: Agreed that we could just be querying for the wanted exec tasks based on the display task fields (in which case maybe this helper function isn't needed)

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
+	grip.DebugWhen(len(utility.UniqueStrings(bundledTasks)) != len(bundledTasks), message.Fields{
 		"ticket":           "EVG-17261",
 		"message":          "archiving same task multiple times",
-		"tasks_to_archive": taskIds,
+		"tasks_to_archive": bundledTasks,
 	})
 
+	err := archiveAll(bundledTasks, execTasks, toUpdateExecTaskIds, toArchive)
+
+	return err
```
ablack12, line 2563: return archiveAll(...)

model/task/task.go
ablack12: Or is this toArchiveIds ? I think it's helpful for variable names to match each other better if they contain similar information, and for the comment / variable name to say Id if it's just Ids (like at 2534)

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
+}
+
+// archiveAll takes in:
+// - taskIds         : All tasks id's of display tasks, execution tasks, and tasks
```
ablack12, line 2567: nit: ids, not id's (same at 2571)

model/task/task.go
```diff
@@ -2572,35 +2580,90 @@ func ArchiveMany(tasks []Task) error {
...
+				return nil, errors.Wrap(err, "archiving tasks")
+			}
 		}
+		if len(execTasksIds) > 0 {
+			// Backwards compatibility call + LPE setting for all tasks
+			_, err = evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(sessCtx,
+				bson.D{{Key: "_id", Value: bson.D{{Key: "$in", Value: execTasksIds}}}}, // Query all execution tasks
+				bson.A{ // Pipeline
+					bson.D{{Key: "$set", Value: bson.D{ // Sets LatestParentExecution (LPE) = !exists(LPE) ? execution + 1 : LPE + 1
+						{Key: LatestParentExecutionKey, Value: bson.D{
```
ablack12, line 2617: We shouldn't need to use this bson.D type / bson.A formatting for UpdateMany; the deleted code below is an example of this.

ablack12: Since old things can be reset we probably should keep the backwards compatible call though actually.

model/task/task.go
```diff
@@ -2572,35 +2580,90 @@ func ArchiveMany(tasks []Task) error {
...
-				},
-			})
-		if err != nil {
-			return nil, err
+			if err != nil {
+				return nil, errors.Wrap(err, "updating documents")
+			}
+
+			// Call to update all tasks that are actually restarting
+			_, err = evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(sessCtx,
```
ablack12, line 2647: should be in an if toUpdateExecTaskIds > 0 right?

model/task/task.go
```diff
@@ -2572,35 +2580,90 @@ func ArchiveMany(tasks []Task) error {
...
+			_, err = oldTaskColl.InsertMany(sessCtx, toArchive)
+			if err != nil {
+				return nil, errors.Wrap(err, "archiving tasks")
+			}
+		}
+		if len(taskIds) > 0 {
+			_, err = evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(sessCtx,
+				bson.D{{Key: "_id", Value: bson.D{{Key: "$in", Value: taskIds}}}}, // Query all display tasks and tasks
+				bson.A{ // Pipeline
+					bson.D{{Key: "$set", Value: bson.D{ // execution = execution + 1
```
ablack12, line 2595: This should just be inc

model/task/task.go
```diff
@@ -2486,82 +2487,89 @@ func (t *Task) Insert() error {
...
+	return err
+}
+
+// archiveAll takes in:
+// - taskIds         : All tasks id's of display tasks, execution tasks, and tasks
+// - toUpdateTaskIds : All tasks that should be progressing to the latest execution
+// - toArchive       : Compiled archived tasks to put in old tasks collection
+//
+// Each array *should* include execution tasks id's
+func archiveAll(taskIds, execTasksIds, toUpdateExecTaskIds []string, toArchive []interface{}) error {
```
ablack12, line 2572: Can we make these a bit clearer? It seems to me like taskIds is all tasks that are being reset, since we update execution. toUpdateExecTaskIds isn't very clear since we also run an update on execTaskIds. Seems more like those are the exec tasks being reset. Also I wonder if, in the case that toUpdateExecTaskIds == execTaskIds, we could have a combined update instead of split updates.

cc @bynn for that last idea we could split a new ticket, up to you

ablack12 (changes requested):

model/task/task.go
ZackarySantana: I looked in to the queries and it does take in to account the different failure status like my helper-- migrating to do a conditional to correctly query all or only exec!

model/task/task.go
ZackarySantana: Deleting the help function as mentioned in another thread!

model/task/task.go
ZackarySantana: Because tasks and display tasks are always highest execution, this can go back to using $inc!

model/task/task.go
ZackarySantana: Redid comments

model/task/task.go
ZackarySantana: I clarified the comments a lot and refactored each variable

allTaskIds := []string{}          // Contains all tasks and display tasks IDs
execTaskIds := []string{}         // Contains all exec tasks IDs
toUpdateExecTaskIds := []string{} // Contains all exec tasks IDs that should update and have new execution
archivedTasks := []interface{}{}  // Contains all archived tasks (task, display, and execution). Created by Task#makeArchivedTask()

The interface is because it is a going to be transfered into bson and put in as a document

model/task/task.go
ZackarySantana: I clarrified it and it should be more apparent (please let me know!!) but the bundledTasks is only tasks and display tasks Ids.

All tasks and display tasks that are put in to archivemany will always be archived

model/task/task.go
ZackarySantana: Fixed!

model/task/task.go
ZackarySantana: Fixed!

model/task/task.go
ZackarySantana: I didn't include this, because if there are any execTasks, there must be at least one updating

ZackarySantana: Users shouldn't be able to reset a display task that only has succeeding exec tasks-- but should I include it to cover all angles?

model/task/task.go
ZackarySantana: Pipelines don't support the inc operator (at least our version maybe? I test it with the same version as prod and it said unsupported). I did update the updateone call to use inc instead though!

model/task/task.go
ZackarySantana: I forgot to update the comments... I just did and I think they're really clear now and consistent with other comments (please lmk if it is cleaner)

The combined update would be possible with some checks- my only concern with it is that it will be a large query. Should I implement the grand query?

model/task/task.go
ZackarySantana: Fixed!

model/task/task.go
ZackarySantana: I updated to use bson.M instead of bson.D, but for bson.A (which is bson representation of array), I think there isn't any alternative and looks nice (easier to see with the bson.M!). The alternative which we had before was a bson.M with key and value with value being "". Which seems more verbose. What do you think?

ablack12: I'm not sure I understand what you mean by there isn't an alternative. This also can just be bson.M, as visible in the old way this was written: https://github.com/evergreen-ci/evergreen/blob/c8ac8688404d6dd1ac0e420e206238ef5895b98e/model/task/task.go#L2512

model/task/task.go
ablack12: Yes they can; this is the purpose of the general "restart when all is completed". It's similar to how you can restart a regular task that is successful.

model/task/task.go
ablack12: This is the old way that this is written. https://github.com/evergreen-ci/evergreen/blob/c8ac8688404d6dd1ac0e420e206238ef5895b98e/model/task/task.go#L2582 You should be able to do something similar.

model/task/task.go
ZackarySantana: There isn't an alternative to bson.A as in an unordered representation of a bson array. bson.M is an unordered key value list (as bson.D is an ordered key value list). The unset can be changed to a bson.M but I think it wouldn't be as semantic and it would add the quotes-- but I think it has the benefit of matching what we originally did and matches more what we do in the codebase. I think the tradeoffs lean towards using bson.A to be more semantic but which do you recommend?

model/task/task.go
ZackarySantana: Will there be a 'restart failed when all is completed" (in 'restart when all is completed', execTasks == toUpdateExecTask)? And if there is a restarted failed when all is completed, if all execution tasks succeed, the display task succeeds, and would not be counted as failed and would not be restarted. If the display task is failed, one execution has failed at least, and execTask >= toUpdateExecTask > 0.

I think the use case in which execTask > 0 but toUpdateExecTask == 0 (which is what the if statement would prevent), is when a user goes to a display task, and clicks restart failed when finished (if it's a button) or restart failed. And in this case, I discussed a little bit with Bynn and we thought there should be something blocking the user form doing so because there are none failed (or in the case of the when finished button, it should abort restarting if all failed?).

I think it makes sense to add the if statement to protect from a DB error, invalidating the transaction, but I'm not sure if it would ever be used

model/task/task.go
ZackarySantana: Oh! This is brilliant, let me change from the pipeline version to using this updatemany call. It definitely will be more concise and can use inc

ZackarySantana: I keep getting a memory error, I've been debugging but I am having trouble, could you look over these images when you review (no problem if you can't find the error)
Old:
Screen Shot 2022-08-08 at 3 38 14 PM
New:
Screen Shot 2022-08-08 at 3 37 22 PM

model/task/task.go
ZackarySantana: I tried setting the pipeline $unset to bson.M and I get

write exception: write errors: [$unset specification must be a string or an array]

model/task/task.go
ZackarySantana: I tried setting the pipeline $unset to bson.M and I get

write exception: write errors: [$unset specification must be a string or an array]

ZackarySantana: But for the old UpdateMany call that you linked in the other thread, when using that, $unset MUST be bson.M

So it seems that:
- UpdateMany old call (not pipeline): $unset must be bson.M
- UpdateMany new call (uses pipeline): $unset must be bson.A

ZackarySantana:
evergreen retry

model/task/task.go
ablack12: So what I don't understand is why we're using pipeline at all. Why can't this just be the following? Is it strictly needed to use conditional? 
```
_, err = evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(sessCtx,
				bson.M{"_id": bson.M{"$in": execTaskIds}}, // Query all execution tasks
				bson.M{"$set": bson.M{ // Sets LatestParentExecution (LPE) = !exists(LPE) ? execution + 1 : LPE + 1
						LatestParentExecutionKey: bson.M{
							"$cond": bson.A{
								bson.M{"$not": bson.A{ // !exists(LPE)
									"$" + LatestParentExecutionKey,
								}},
								bson.M{"$add": bson.A{ // execution + 1
									"$" + ExecutionKey, 1,
								}},
								bson.M{"$add": bson.A{ // LPE + 1
									"$" + LatestParentExecutionKey, 1,
								}},
						},
					}},
				})
```

model/task/task.go
ablack12: The query would be the same and only the update would be different right? I think a split ticket is better since this is a large code review

model/task/task.go
ablack12: This is the current behavior of restarting a display task so we will not be adding a new button; you can play around with this in staging if you want. Right now if a user tries to restart a display task (using just the regular Restart button), we set ResetWhenFinished, which is checked during EndTask (i.e. we check if the last execution task in the display task is finished, and if we want to reset it when finished, and then we call TryResetTask). 

I'm not sure I understand the next part of your comment but it's important that we don't break this functionality. The thing we're adding in this project is to allow users to only restart failed if they want, as opposed to restarting all tasks, which is currently the only option.

model/task/task.go
ZackarySantana: Yeah, $cond is a pipeline restricted feature: https://www.mongodb.com/docs/v4.2/reference/operator/aggregation/cond/

ZackarySantana: I tried without a pipeline and I get `The dollar ($) prefixed field '$cond' in 'latest_parent_execution.$cond' is not valid for storage.]` for additional info

model/task/task.go
ZackarySantana: In the current behavior, execTask == toUpdateExecTask since all exec tasks always restart. I don't think there is a use case in which results in execTack > 0 but toUpdateExecTask == 0. If the display task is being archived, it will have at least one exec task that it will have to update (aka advance to next execution)

So the if toUpdateExecTask > 0 I _think_ is not needed but I might be overseeing the case in which it is, but adding the if statement might save a random use case that was unforeseeable. I'm not sure which action is the best, with or without the if statement

model/task/task.go
ZackarySantana: Updated

model/task/task.go
ZackarySantana: I envision it as

if execTask == toUpdateExecTask {
// Combined query
} else {
// Existing query
}

But the combined query will be similar to the first PR version I had with a large query (but with bson.M instead of bson.D). The combined query would look like
$in: ids
$set: LPE = !LPE ? Execution + 1 : LPE + 1 
$set: Execution = LPE (so we get the newest LPE)
$unset --

The combined query could replace both by adding one pipeline
$in: execTaskIds
$set: LPE = !LPE ? Execution + 1 : LPE + 1 
$in: toUpdateExecTaskIds
$set: Execution = LPE (so we get the newest LPE)
$unset --

and we only have one query

model/task/task.go
```diff
@@ -2500,89 +2501,92 @@ func (t *Task) Insert() error {
...
+				"archive_task_id": archiveTask.Id,
+				"old_task_id":     archiveTask.OldTaskId,
+				"execution":       t.Execution,
+				"display_only":    t.DisplayOnly,
+			}))
+			return errors.Wrap(err, "inserting archived task into old tasks")
 		}
-		if err = ArchiveMany(execTasks); err != nil {
-			return errors.Wrap(err, "archiving execution tasks")
+		err = UpdateOne(
```
ablack12, line 2520: This update stage looks the same as the one at the top of ArchiveAll. So should we just always call ArchiveMany? (is there any reason to not call ArchiveMany here)? Or if that can't be done, can this part be refactored from each so there isn't duplicate code?

ZackarySantana: The logic for this (archiving in to old collection and updateone) are functionally the same as the ArchiveAll. I can simply this function just to always call 
err := ArchiveMany([]Task{*t})
return errors.Wrapf(err, "archiving task '%s'", t.Id)

but @bynn was concerned about performance (doing a transaction for every task) since it might be more expensive just restarting any task than before. I think a refactor can look really nice and is semantic, let me try implementing one!

ablack12: Ah Bynn brings up a good point about the transaction. Refactor seems like a good inbetween then 👍

model/task/task.go
```diff
@@ -2513,82 +2515,84 @@ func (t *Task) Insert() error {
...
-		// For any display tasks here, make sure that we also archive all their execution tasks.
-		// Consider each execution tasks individually, so we don't query a task we already passed in.
-		for _, et := range t.ExecutionTasks {
-			if !existingTasksMap[et] {
-				additionalTasks = append(additionalTasks, et)
-				existingTasksMap[et] = true
+		allTaskIds = append(allTaskIds, t.Id)
+		archivedTasks = append(archivedTasks, t.makeArchivedTask())
+		if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
+			var eTasks []Task
```
ablack12, line 2561: nit: execTasks is more consistent with execTaskIds

model/task/task.go
```diff
@@ -2513,82 +2515,84 @@ func (t *Task) Insert() error {
...
+		"tasks_to_archive": allTaskIds,
 	})
 
+	return archiveAll(allTaskIds, execTaskIds, toUpdateExecTaskIds, archivedTasks)
+}
+
+// archiveAll takes in:
+// - taskIds               : All tasks and display tasks IDs
+// - execTaskIds           : All execution task IDs
+// - toUpdateExecTaskIds   : All execution task IDs for execution tasks that will be archived/restarted
```
ablack12, line 2593: Sorry I think this is a repeat comment but I can't find it if you answered it -- can this be called toRestartExecTaskIds instead? Or something like that.  ToUpdate is confusing since we also update the execTaskIds.

model/task/task.go
```diff
@@ -2599,35 +2603,89 @@ func ArchiveMany(tasks []Task) error {
...
+								}},
+								bson.M{"$add": bson.A{ // LPE + 1
+									"$" + LatestParentExecutionKey, 1,
+								}},
+							},
+						},
+					}},
+				})
+
+			// Same call without backwards compatibility
```
ablack12, line 2653: Reminder to remove this, and update the follow up ticket as necessary

model/task/task.go
```diff
@@ -2599,35 +2603,89 @@ func ArchiveMany(tasks []Task) error {
...
-					AbortInfoKey: "",
-				},
-				"$inc": bson.M{
-					ExecutionKey: 1,
-				},
-			})
-		if err != nil {
-			return nil, err
+			if err != nil {
+				return nil, errors.Wrap(err, "updating documents")
```
ablack12, line 2665: Can we be more specific here? Updating latest parent executions or something

model/task/task.go
```diff
@@ -2599,35 +2603,89 @@ func ArchiveMany(tasks []Task) error {
...
+					bson.M{"$set": bson.M{ // Execution = LPE
+						ExecutionKey: "$" + LatestParentExecutionKey,
+					}},
+					bson.M{"$unset": bson.A{
+						AbortedKey,
+						AbortInfoKey,
+						OverrideDependenciesKey,
+					}}})
+
+			return nil, errors.Wrap(err, "updating documents")
```
ablack12, line 2681: also more specific here

model/task/task.go
```diff
@@ -3651,7 +3709,7 @@ func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) (*Ta
 	}
 	pipeline, err := getTasksByVersionPipeline(versionID, opts)
 	if err != nil {
-		return nil, errors.Wrap(err, "error getting tasks by version pipeline")
+		return nil, errors.Wrap(err, " etting tasks by version pipeline")
```
ablack12, line 3712: nit: getting

model/task/task_test.go
```diff
@@ -3669,6 +3788,154 @@ func TestArchive(t *testing.T) {
...
+
+		// Cross checks the collections to ensure that each task was or was not archived
+		checkTaskIsArchived(t, archivedT1)
+		checkTaskIsNotArchived(t, t2.Id, 0)
+		checkTaskIsArchived(t, archivedDisplayTaskID)
+
+		checkEventLogHostTaskExecutions(t, hostID, archivedT1, archivedExecution)
+	})
+
+	// This test is for the edge case of a archiving with only failed execution tasks, then archiving all execution tasks
```
ablack12, line 3893: nit: of archiving

model/task/task.go
```diff
@@ -2513,82 +2515,84 @@ func (t *Task) Insert() error {
...
-			if !existingTasksMap[et] {
-				additionalTasks = append(additionalTasks, et)
-				existingTasksMap[et] = true
+		allTaskIds = append(allTaskIds, t.Id)
+		archivedTasks = append(archivedTasks, t.makeArchivedTask())
+		if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
+			var eTasks []Task
+			var err error
+
+			if t.ResetFailedWhenFinished {
```
ablack12, line 2564: We should check ResetWhenFinished _before_ we check ResetFailedWhenFinished; if both are set, All would take precedence over just failed.

model/task/task_test.go
```diff
@@ -3140,6 +3141,124 @@ func TestArchiveMany(t *testing.T) {
 	}
 }
 
+func TestArchiveManyFailedOnly(t *testing.T) {
+	require.NoError(t, db.ClearCollections(Collection, OldCollection))
+	t1_et1 := Task{
```
ablack12, line 3146: nit: can these be re-written to use camelCase instead of snake_case? (just et1 / et2 would be reasonable)

model/task/task_test.go
```diff
@@ -3140,6 +3141,124 @@ func TestArchiveMany(t *testing.T) {
...
+		Status:                  evergreen.TaskFailed,
+		ExecutionTasks:          []string{t3_et1.Id, t3_et2.Id, t3_et3.Id},
+		Execution:               0,
+		Aborted:                 true,
+		DisplayOnly:             true,
+		ResetFailedWhenFinished: true,
+		Version:                 "v",
+	}
+	assert.NoError(t, t3.Insert())
+	assert.NoError(t, t3.Archive()) // Failed only is true
```
ablack12, line 3215: Should we check the results of having archived t3 with failed only true? (i.e. that just some exec tasks are restarted, and that LPE is what we expect?)

model/task/task_test.go
```diff
@@ -3140,6 +3141,124 @@ func TestArchiveMany(t *testing.T) {
...
+	// t3: t1 (execution 2), t2 (execution 2), t3 (execution 2)
+	// t4 (execution 2)
+
+	currentTasks, err := FindAll(db.Query(ByVersion("v")))
+	assert.NoError(t, err)
+	assert.Len(t, currentTasks, 9)
+
+	for _, task := range currentTasks {
+		switch task.Id {
+		case t1.Id, t1_et1.Id, t1_et2.Id:
```
ablack12, line 3248: Can we also check LPE?

model/task/task_test.go
```diff
@@ -3669,6 +3788,154 @@ func TestArchive(t *testing.T) {
...
+		hostID := "hostID"
+		event.LogHostRunningTaskSet(hostID, t1.Id, 0)
+		event.LogHostRunningTaskCleared(hostID, t1.Id, 0)
+
+		// Verifies the execution before and after calling Archive
+		archivedDisplayTaskID := MakeOldID(dt.Id, dt.Execution)
+		require.Equal(t, 0, dt.Execution)
+		require.NoError(t, dt.Archive())
+		dtPointer, err := FindOneId(dt.Id)
+		dt = *dtPointer
```
ablack12, line 3881: There's no reason to do this; execution can be checked directly from what's returned at 3880

model/task/task_test.go
```diff
@@ -3140,6 +3141,124 @@ func TestArchiveMany(t *testing.T) {
...
+		Execution:               0,
+		Aborted:                 true,
+		DisplayOnly:             true,
+		ResetFailedWhenFinished: true,
+		Version:                 "v",
+	}
+	assert.NoError(t, t3.Insert())
+	assert.NoError(t, t3.Archive()) // Failed only is true
+	t3_pointer, err := FindByIdExecution(t3.Id, nil)
+	t3 = *t3_pointer
```
ablack12, line 3217: Should check error / and require.NotNil(t3) before doing this (also not certain this is necessary; at 3228 you can just do *t3)

model/task/task_test.go
```diff
@@ -3669,6 +3788,154 @@ func TestArchive(t *testing.T) {
...
+		// Cross checks the collections to ensure that each task was or was not archived
+		checkTaskIsArchived(t, archivedT1)
+		checkTaskIsNotArchived(t, t2.Id, 0)
+		checkTaskIsArchived(t, archivedDisplayTaskID)
+
+		checkEventLogHostTaskExecutions(t, hostID, archivedT1, archivedExecution)
+	})
+
+	// This test is for the edge case of a archiving with only failed execution tasks, then archiving all execution tasks
+	t.Run("ArchivesExecutionTasksAfterFailedOnly", func(t *testing.T) {
```
ablack12, line 3894: Can we add LPE checks to this?

ablack12 (commented):
mostly small comments!

model/task/task.go
ZackarySantana: Will refactor!

model/task/task.go
ZackarySantana: Yes! That sounds much cleaner, I will rename

bynn (approved):
i think that this will have my seal of approval after addressing annies comments 
image

model/task/task_test.go
```diff
@@ -3140,6 +3141,148 @@ func TestArchiveMany(t *testing.T) {
...
+	// t3: t1 (execution 2), t2 (execution 2), t3 (execution 2)
+	// t4 (execution 2)
+
+	currentTasks, err = FindAll(db.Query(ByVersion("v")))
+	assert.NoError(t, err)
+	assert.Len(t, currentTasks, 9)
+
+	for _, task := range currentTasks {
+		switch task.Id {
+		case t1.Id:
```
ablack12, line 3266: Can we check that LPE isn't set for non display tasks?

model/task/task_test.go
```diff
@@ -3140,6 +3141,148 @@ func TestArchiveMany(t *testing.T) {
...
+		LatestParentExecution: 2,
+		Aborted:               true,
+		Version:               "v",
+	}
+	assert.NoError(t, et1.Insert())
+	et2 := Task{
+		Id:                    "et2",
+		Status:                evergreen.TaskSucceeded,
+		Execution:             2,
+		LatestParentExecution: 2,
```
ablack12, line 3159: Sorry for the late thought -- can we remove LPE here and ensure that the tests still work? This is the existing backwards incompatible case so I want to make sure this doesn't break restarts.

ablack12 (commented):
Some nits!

model/task/task_test.go
```diff
@@ -3140,6 +3141,154 @@ func TestArchiveMany(t *testing.T) {
...
+		ExecutionTasks:          []string{et3.Id, et4.Id, et5.Id},
+		Execution:               0,
+		Aborted:                 true,
+		DisplayOnly:             true,
+		ResetFailedWhenFinished: true,
+		Version:                 "v",
+	}
+	assert.NoError(t, t3.Insert())
+	assert.NoError(t, t3.Archive()) // Failed only is true
+	t3Pointer, err := FindByIdExecution(t3.Id, nil)
```
ablack12, line 3217: Should also move this down to 3245 since it's not used until then

ablack12 (approved):
LGTM with nits!

PR #5795, opened 2022-07-11
EVG-16730 assorted restart patch improvements

ablack12:

### Description 
Combined a couple of updates:
1) Marking aborted tasks as needing reset when aborting (need to test this though; may require changes to archive?).
2) Updating build activation

Refactored child patch queries to improve performance (although this isn't actually relevant to restarts). 

### Testing 


graphql/util.go
```diff
@@ -378,31 +378,18 @@ func modifyVersionHandler(ctx context.Context, patchID string, modification mode
...
 	if evergreen.IsPatchRequester(v.Requester) {
-		// restart is handled through graphql because we need the user to specify
-		// which downstream tasks they want to restart
+		// Restart is handled through graphql because we need the user to specify
+		// which downstream tasks they want to restart.
 		if modification.Action != evergreen.RestartAction {
-			//do the same for child patches
-			p, err := patch.FindOneId(patchID)
+			// Only modify the child patch if it is finalized.
+			childPatches, err := patch.GetFinalizedChildPatchIdsForPatch(patchID)
```
malikchaya2, line 385: When testing restarts in staging, can you please include scenarios with child patches?

graphql/util.go
```diff
@@ -378,31 +378,18 @@ func modifyVersionHandler(ctx context.Context, patchID string, modification mode
...
-				for _, childPatch := range childPatches {
-					// only modify the child patch if it is finalized
-					if childPatch.Version != "" {
-						err = modifyVersionHandler(ctx, childPatch.Id.Hex(), modification)
-						if err != nil {
-							return errors.Wrap(mapHTTPStatusToGqlError(ctx, httpStatus, err), fmt.Sprintf("error modifying child patch '%s'", patchID))
-						}
-					}
-
+					return errors.Wrap(mapHTTPStatusToGqlError(ctx, httpStatus, err), fmt.Sprintf("error modifying child patch '%s'", patchID))
```
malikchaya2, line 392: nit: `modifying child patch '%s`

model/lifecycle.go
```diff
@@ -341,8 +341,8 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	// remove execution tasks in case the caller passed both display and execution tasks
-	// the functions below are expected to work if just the display task is passed
+	// Remove execution tasks in case the caller passed both display and execution tasks.
```
malikchaya2, line 344: Not actually PR related, but should we add capitalization of comments to our style guide?

model/lifecycle.go
```diff
@@ -365,14 +365,8 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 		Build     string
 		TaskGroup string
 	}
-	// Mark aborted tasks to reset when finished if not all tasks are finished.
-	if abortInProgress && len(finishedTasks) < len(taskIds) {
```
malikchaya2, line 369: why do we no longer need this?

model/patch/db.go
```diff
@@ -361,3 +361,29 @@ func FindProjectForPatch(patchID mgobson.ObjectId) (string, error) {
...
+	if p == nil {
+		return nil, errors.Wrapf(err, "patch '%s' not found", patchID)
+	}
+	if !p.IsParent() {
+		return nil, nil
+	}
+
+	childPatches, err := Find(ByStringIds(p.Triggers.ChildPatches).WithFields(VersionKey))
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting child patches")
```
malikchaya2, line 382: nit: `getting child patches`

model/task/db.go
```diff
@@ -1649,16 +1649,17 @@ func AbortTasksForBuild(buildId string, taskIds []string, caller string) error {
...
+			VersionKey: versionId, // Include to improve query.
 			IdKey:      bson.M{"$in": taskIds},
 			StatusKey:  bson.M{"$in": evergreen.TaskAbortableStatuses},
 		},
 		bson.M{"$set": bson.M{
-			AbortedKey:   true,
-			AbortInfoKey: AbortInfo{User: caller},
+			AbortedKey:           true,
+			AbortInfoKey:         AbortInfo{User: caller},
+			ResetWhenFinishedKey: true,
```
malikchaya2, line 1662: what does ResetWhenFinishedKey do?

model/task/task.go
```diff
@@ -2542,6 +2542,11 @@ func ArchiveMany(tasks []Task) error {
 		archived = append(archived, *t.makeArchivedTask())
 		taskIds = append(taskIds, t.Id)
 	}
+	grip.DebugWhen(len(utility.UniqueStrings(taskIds)) != len(taskIds), message.Fields{
+		"ticket":           "EVG-17261",
```
malikchaya2, line 2546: sneaky addition, I like.

malikchaya2 (changes requested):

graphql/util.go
ablack12: This actually doesn't affect restarts since we specifically only do this if it's not a restart. I can add a unit test to verify this new function though.

model/lifecycle.go
ablack12: It is already in the style guide but is listed as preferable where I actually thought it was required https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit# ; I think this was done to avoid PRs from needing to be nit picky (oops) / to avoid us needing to go through the codebase explicitly to fix everything (like with error messages)

model/lifecycle.go
ablack12: Mentioned this in the comment but I modified the function at the top AbortTasksForVersion to be AbortAndMarkResetTasksForVersion. Potentially there's a reason we don't do this together already, but Aborted tasks aren't archived since we only archive finished tasks so I don't see why it needs to be done separately.

model/task/db.go
ablack12: This relates to your earlier comment about why I removed SetAbortedTasksResetWhenFinished, that's all that function did. I believe when the task officially aborts we check ResetWhenFinished to determine whether to restart it or just fail it (like when a PR is aborted).

graphql/util.go
ablack12: I'd prefer to rely on unit testing in this case, since the change that was made was small and query related so it can be verified in a repeatable way. I'll push this shortly!

malikchaya2 (approved):

PR #5797, opened 2022-07-11
EVG-17170: handle already-stopped pods during termination and work around incompatible yaml.v3 versions

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17170

### Description 
Code-wise, this is the same change as #5780 which was already approved, so the code change related to Cocoa/pods doesn't need to be re-reviewed. The difference in this PR is that the `gopkg.in/yaml.v3` package has to be worked around so that Evergreen's Go modules support both the old version of it from [when the dependnecy was first introduced](https://github.com/evergreen-ci/evergreen/pull/4784) (i.e. some commit after v3.0.0) and the newer minor version (v3.0.1) because upgrading it somehow breaks project validation (even though minor package upgrades are _supposed_ to be backward compatible and therefore safe to upgrade). We can't just downgrade to the old version either because other modules imported by Evergreen already have dependencies on the new version (and transitively downgrading everything and forcing them to follow Evergreen's backward compatibility workaround seems like too much work).

To work with both the old and new version, I had to specially pin Evergreen's version of `gopkg.in/yaml.v3` to the old version, so all code within the Evergreen repo that uses `gopkg.in/yaml.v3` will use the old version of `gopkg.in/yaml.v3`. Other dependencies (such as other modules) are still allowed to use the newer version of `gopkg.in/yaml.v3`.

This is probably not a great long-term solution. Going forward, anyone using yaml.v3 in their code will have to remember to use the old version (`gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b`) and not the new version (`gopkg.in/yaml.v3 v3.0.1`). I made [a ticket](https://jira.mongodb.org/browse/EVG-17291) to actually fix whatever is the underlying issue and allow us to upgrade above the old version.

### Testing 
Not sure how to test this. I'm inclined to say we should trust that Go modules work as they're documented (i.e. that Evergreen will use the old yaml.v3 version for code in the Evergreen repo) and existing tests pass.

`verify-agent-version-update` and `verify-client-version-update` won't pass because the files were modified, but they're not actual source code changes, so I won't bump them.


service/rest_version.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
```
ablack12, line 22: Was this purposeful or should this also be `yaml "gopkg.in/20210107192922/yaml.v3"`?

bynn (approved):
🚀

ablack12 (approved):

PR #5798, opened 2022-07-11
EVG-17726: specify Windows version for capacity providers

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17226

### Description 
Originally, we used placement constraints to specify that the container had to run on a host that ran a particular Windows version because [Windows containers have to be compatible with their host OS version](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility). However, the ECS infrastructure was changed so instead, we now have to specify a capacity provider that maps to a particular Windows version. For example, all Windows server 2022 containers have to run in the one capacity provider that runs Windows server 2022 hosts.

* Add admin setting for setting the Windows version for a capacity provider.
* Remove placement constraint for Windows version and replace with capacity provider.

### Testing 
Added unit tests and tested in local-evergreen that the admin settings work correctly.

### DIFF
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -247,10 +247,6 @@ func exportECSPodContainerDef(settings *evergreen.Settings, p *pod.Pod) (*cocoa.
 	return def, nil
 }
 
-const (
-	ecsWindowsVersionTagConstraint = "attribute:WindowsVersion"
-)
-
 // exportECSPodExecutionOptions exports the ECS configuration into
 // cocoa.ECSPodExecutionOptions.
 func exportECSPodExecutionOptions(ecsConfig evergreen.ECSConfig, containerOpts pod.TaskContainerCreationOptions) (*cocoa.ECSPodExecutionOptions, error) {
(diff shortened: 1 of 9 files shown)


cloud/pod_util.go
```diff
@@ -262,22 +258,21 @@ func exportECSPodExecutionOptions(ecsConfig evergreen.ECSConfig, containerOpts p
...
+		}
 		if containerOpts.OS.Matches(cp.OS) && containerOpts.Arch.Matches(cp.Arch) {
 			opts.SetCapacityProvider(cp.Name)
 			foundCapacityProvider = true
 			break
 		}
 	}
 	if !foundCapacityProvider {
+		if containerOpts.OS == pod.OSWindows {
+			return nil, errors.Errorf("container OS '%s' with Windows version '%s' and arch '%s' did not match any recognized capacity provider", containerOpts.OS, containerOpts.WindowsVersion, containerOpts.Arch)
```
hadjri, line 274: Since we know OS is `OSWindows` here it may be redundant to say "with Windows version"

config_cloud.go
```diff
@@ -213,24 +258,16 @@ func (p *ECSCapacityProvider) Validate() error {
...
-
-func (a ECSArch) Validate() error {
-	switch a {
-	case ECSArchAMD64, ECSArchARM64:
-		return nil
-	default:
-		return errors.Errorf("unrecognized ECS capacity provider arch '%s'", a)
+	if p.OS == ECSOSWindows {
+		if p.WindowsVersion == "" {
+			catcher.New("must specify a particular Windows version when using Windows OS")
```
hadjri, line 263: Can we add this case for the empty string to `Validate` and remove the if/else structure here?

cloud/pod_util.go
```diff
@@ -262,22 +258,21 @@ func exportECSPodExecutionOptions(ecsConfig evergreen.ECSConfig, containerOpts p
...
 
-	placementOpts := cocoa.NewECSPodPlacementOptions()
-	if containerOpts.WindowsVersion != "" {
-		windowsVersionConstraint := fmt.Sprintf("%s == %s", ecsWindowsVersionTagConstraint, containerOpts.WindowsVersion)
-		placementOpts.AddInstanceFilters(windowsVersionConstraint)
-	}
-	opts.SetPlacementOptions(*placementOpts)
-
 	var foundCapacityProvider bool
 	for _, cp := range ecsConfig.CapacityProviders {
```
hadjri, line 262: Can we add a comment right before the for loop here explaining the rationale?

config_cloud.go
Kimchelly: We don't want empty string validation to pass if the OS is Windows, so I'm inclined to not allow the empty string unconditionally. This is not particularly complex logic anyways.

hadjri (approved):

PR #5799, opened 2022-07-11
EVG-16882: modify container command to bootstrap using provisioning REST route

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16882

### Description 
In order to provision the container, we have to download the agent and start it. Instead of directly downloading the agent from Evergreen using the container command, the container command now asks for the script to run which contains the command to actually download the agent from the REST route (from #5689). This indirection is necessary because as part of the design, we have to cache pod definitions and want them to change as little as possible. This future-proofs the pod definition against future changes so if we ever want to change how the agent is provisioned on the host, we can do so by modifying the REST route's response rather than the container command itself (and therefore prevent the pod definition from changing).

There's also some small but important details about the REST route that I had to change for this to work in Windows, but I'm omitting them because I don't think they'll make sense without a longer explanation of PowerShell and Windows.

* Change container command to download the provisioning script from the REST API instead of directly provisioning the agent.
* Update REST route to work with PowerShell script.
* Fix small bug where a gimlet response isn't returned.

### Testing 
Added unit tests and checked that it works for Windows/Linux pods in staging.

### DIFF
--- a/cloud/pod_agent_util.go
+++ b/cloud/pod_agent_util.go
@@ -9,27 +9,44 @@ import (
 	"github.com/evergreen-ci/evergreen/model/pod"
 )
 
-// agentScript returns the script to provision and run the agent in the pod's
-// container.
-func agentScript(settings *evergreen.Settings, p *pod.Pod) []string {
-	scriptCmds := []string{downloadAgentCommands(settings, p)}
-	if p.TaskContainerCreationOpts.OS == pod.OSLinux {
-		scriptCmds = append(scriptCmds, fmt.Sprintf("chmod +x %s", clientName(p)))
-	}
-	agentCmd := strings.Join(agentCommand(settings, p), " ")
-	scriptCmds = append(scriptCmds, agentCmd)
+// bootstrapContainerCommand returns the script to bootstrap the pod's primary container
+// to run the agent.
+func bootstrapContainerCommand(settings *evergreen.Settings, p *pod.Pod) []string {
+	scriptCmds := downloadPodProvisioningScriptCommand(settings, p)
 
-	return append(invokeShellScriptCommand(p), strings.Join(scriptCmds, " && "))
+	return append(invokeShellScriptCommand(p), scriptCmds)
 }
 
 // invokeShellScriptCommand returns the arguments to invoke an in-line shell
 // script in the pod's container.
 func invokeShellScriptCommand(p *pod.Pod) []string {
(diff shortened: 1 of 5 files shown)


rest/route/pod_agent.go
```diff
@@ -120,18 +130,32 @@ func (h *podProvisioningScript) downloadAgentCommands(p *pod.Pod, downloadFromS3
 	)
 	retryArgs := h.curlRetryArgs(curlDefaultNumRetries, curlDefaultMaxSecs)
 
-	var curlCmd string
+	curlExecutable := "curl"
+	if p.TaskContainerCreationOpts.OS == pod.OSWindows {
+		curlExecutable = curlExecutable + ".exe"
```
hadjri, line 135: We use this .exe string in several places now, can we make it a global constant?

cloud/pod_agent_util.go
```diff
@@ -2,113 +2,50 @@ package cloud
...
-		"--mode=pod",
-		fmt.Sprintf("--log_prefix=%s", filepath.Join(p.TaskContainerCreationOpts.WorkingDir, "agent")),
-		fmt.Sprintf("--working_directory=%s", p.TaskContainerCreationOpts.WorkingDir),
-	}
-}
-
-// downloadAgentCommands returns the commands to download the agent in the pod's
-// container.
-func downloadAgentCommands(settings *evergreen.Settings, p *pod.Pod) string {
+func downloadPodProvisioningScriptCommand(settings *evergreen.Settings, p *pod.Pod) string {
```
hadjri, line 29: It is relatively straightforward but a comment would be nice to be consistent with the other util functions

hadjri (approved):
LGTM with nits

cloud/pod_agent_util.go
Kimchelly: Added a comment.

rest/route/pod_agent.go
Kimchelly: Eh, I don't have strong feelings about this because hosts/pods usually have a lot of plain string manipulation logic when writing shell scripts. I don't think there's a lot of value in making constants out of `.exe` since we don't do it for other common shell script strings (like `bash` and `ssh`) and it's generally understood that `.exe` is a thing for Windows executables.

hadjri: Alright sgtm

PR #5800, opened 2022-07-12
EVG-17252: remove EC2 auto provider and price cacher

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17252

### Description 
aws-sdk-go bumped their module minor version but broke backward compatibility (they're not supposed to do this). However, the only code that broke in Evergreen was for the EC2 price caching logic, which is only depended on by the currently-unused EC2 auto provider. Rather than try to fix the price caching logic, I instead opted to just remove the auto provider since we don't use it and don't have plans to use it.

* Remove all code that depends on unused EC2 auto provider, including the broken EC2 price cacher.
* Upgrade aws-sdk-go.

### Testing 
Updated existing tests to not depend on the auto provider and remove auto-specific tests. I also verified that:
* No distros in staging/prod still use the auto provider.
* host.create and spawn hosts don't use the auto provider (they use the on-demand provider).


cloud/ec2_test.go
```diff
@@ -389,9 +377,7 @@ func (s *EC2Suite) TestSpawnHostClassicOnDemand() {
 }
 
 func (s *EC2Suite) TestSpawnHostVPCOnDemand() {
-	pkgCachingPriceFetcher.ec2Prices = map[odInfo]float64{
-		odInfo{"Linux", "instanceType", "US East (N. Virginia)"}: .1,
-	}
+	// kim: TODO: check this test
```
malikchaya2, line 380: What are these todos for? are they meant for a different ticket? If yes, do we want to add the ticket number?

malikchaya2 (changes requested):

malikchaya2 (approved):

PR #5801, opened 2022-07-13
EVG-17040 Remove pipeline stage looking for tasks without BV display name set

hadjri:
[EVG-17040](https://jira.mongodb.org/browse/EVG-17040)

### Description 
Given that `FindUniqueBuildVariantNamesByTask` looks back no more than 50 versions and build variant display names have been cached in structs for the past 30 days, it seems appropriate now to directly read the `BuildVariantDisplayName` on a task when reading variant data for the task history page rather than performing a facet/lookup combination.
### Testing 
Augmented existing unit tests given new assumptions when retrieving BV display name, and tested in staging that task history page remains accurate.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -988,49 +988,12 @@ type BuildVariantTuple struct {
 	DisplayName  string `bson:"build_variant_display_name"`
 }
 
-const VersionLimit = 50
-
-// Reusable pipeline stages when finding build variant display names
 var (
-	// sort build variant display names alphabetically
-	sortByVariantDisplayName = bson.M{
-		"$sort": bson.M{
-			"build_variant_display_name": 1,
-		},
-	}
-
-	// cleanup the results
-	projectBvResults = bson.M{
-		"$project": bson.M{
-			"_id":                        0,
-			"build_variant":              "$" + BuildVariantKey,
-			"build_variant_display_name": "$" + BuildVariantDisplayNameKey,
-		},
-	}
-
-	// reorganize the results to get the build variant names and a corresponding build id
-	projectBuildIdAndVariant = bson.M{
-		"$project": bson.M{
-			"_id":                      0,
-			BuildVariantKey:            bsonutil.GetDottedKeyName("$_id", BuildVariantKey),
-			BuildVariantDisplayNameKey: bsonutil.GetDottedKeyName("$_id", BuildVariantDisplayNameKey),
-			BuildIdKey:                 "$" + BuildIdKey,
-		},
-	}
-
-	// group the build variants by unique build variant names and get a build id for each
-	groupByBuildVariant = bson.M{
-		"$group": bson.M{
-			"_id": bson.M{
-				BuildVariantKey:            "$" + BuildVariantKey,
-				BuildVariantDisplayNameKey: "$" + BuildVariantDisplayNameKey,
-			},
-			BuildIdKey: bson.M{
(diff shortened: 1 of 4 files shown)


model/task/db.go
```diff
@@ -985,72 +985,75 @@ func FindByExecutionTasksAndMaxExecution(taskIds []*string, execution int) ([]Ta
 
 type BuildVariantTuple struct {
 	BuildVariant string `bson:"build_variant"`
```
ablack12, line 987: Does this even need JSON tags? If we aren't returning this to users or reading JSON into it then I'd suggest removing it altogether.

model/task/db.go
```diff
@@ -985,72 +985,75 @@ func FindByExecutionTasksAndMaxExecution(taskIds []*string, execution int) ([]Ta
...
+			"_id":                      0,
+			BuildVariantKey:            bsonutil.GetDottedKeyName("$_id", BuildVariantKey),
+			BuildVariantDisplayNameKey: bsonutil.GetDottedKeyName("$_id", BuildVariantDisplayNameKey),
+			BuildIdKey:                 "$" + BuildIdKey,
 		},
 	}
 
 	// group the build variants by unique build variant names and get a build id for each
-	groupByBuildVariant := bson.M{
+	groupByBuildVariant = bson.M{
```
ablack12, line 1022: If these are actually only being used in FindUniqueBuildVariantNamesByTask, I'm not sure it's worth pulling out. It's a bit easier to read pipelines when you can just read them sequentially rather than navigating between variables and the function itself.

model/task/db.go
```diff
@@ -985,72 +985,75 @@ func FindByExecutionTasksAndMaxExecution(taskIds []*string, execution int) ([]Ta
...
-			"$and": []bson.M{
-				{RevisionOrderNumberKey: bson.M{"$gte": repoOrderNumber - VersionLimit}},
-				{RevisionOrderNumberKey: bson.M{"$lte": repoOrderNumber}},
-			},
+// Reusable pipeline stages when finding build variant display names
+var (
+	// sort build variant display names alphabetically
+	sortByVariantDisplayName = bson.M{
+		"$sort": bson.M{
+			"build_variant_display_name": 1,
```
ablack12, line 998: It would be better to make bsonutil keys for BuildVariantTuple's fields rather than hardcoding here.

ablack12 (changes requested):

model/task/db.go
hadjri: The bson tags are necessary for pipeline stages - there are no JSON tags currently

ablack12: Is this still necessary with the bsonutil tags?

hadjri: Np! Yes we need them because the bsonutil tags resolve to the bson tags defined on the struct and `MustHaveTag` will err if they do not exist.

hadjri:
> Can you verify that this is safe to deploy? You should be able to perform a db query to verify if any projects have < 50 versions without this field? I worry if there are any projects that aren't committed to as often.

I checked and there are a significant amount of smaller projects that haven't committed anything in several months - If we want the task history page to remain correct for all these projects we'd need to maintain this legacy case for a while.  I've added a bit more logic to run the old lookup stage in the pipeline only if we find no results when we query the task doc directly (which would not apply for any regularly used project).

khelif96 (approved):

PR #5804, opened 2022-07-13
EVG-17167 add final BuildFromService changes

ablack12:

### Description 
Final build from service changes!

### Testing 
Will verify existing tests pass.


config.go
```diff
@@ -25,15 +25,15 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 	"go.mongodb.org/mongo-driver/mongo/readconcern"
 	"go.mongodb.org/mongo-driver/mongo/writeconcern"
-	yaml "gopkg.in/20210107192922/yaml.v3"
+	"gopkg.in/20210107192922/yaml.v3"
```
ablack12, line 28: My editor won't seem to allow these names haha but they are using the right package.

rest/model/user.go
```diff
@@ -91,9 +91,6 @@ func (g *APIGithubUser) BuildFromService(usr user.GithubUser) {
 }
 
 func (g *APIGithubUser) ToService() user.GithubUser {
-	if g == nil {
```
malikchaya2, line 94: Why do we no longer need the nil check here?

ablack12: It's more that we just don't make that check anywhere. Conventionally we don't call ToService on anything uninitialized, and I checked all usages of this function and it's the same way.

malikchaya2 (approved):

PR #5805, opened 2022-07-14
EVG-17926: Sort private and admin only vars

sophstad:
[EVG-17926](https://jira.mongodb.org/browse/EVG-17926)

### Description 
The `PrivateVars` and `AdminOnlyVars` resolvers construct lists from [maps](https://github.com/evergreen-ci/evergreen/blob/1c0d68ff52d836fd7fcca4df84beb745afd1db10/rest/model/project_event.go#L39-L40) to return to the front end. This results in no guaranteed order, with a differently ordered list sent in each `ProjectEventLogs` request. This change in ordering caused privateVars updates to appear in every event log entry.

To fix, return alphabetically sorted private and admin only vars so that we do not inaccurately report an event.

Before:
image

After:
image

### Testing 
- Tested updated backend with current `main` branch of Spruce in staging

### DIFF
--- a/graphql/generated.go
+++ b/graphql/generated.go
@@ -1300,8 +1300,8 @@ type ProjectSubscriberResolver interface {
 	Subscriber(ctx context.Context, obj *model.APISubscriber) (*Subscriber, error)
 }
 type ProjectVarsResolver interface {
-	AdminOnlyVars(ctx context.Context, obj *model.APIProjectVars) ([]*string, error)
-	PrivateVars(ctx context.Context, obj *model.APIProjectVars) ([]*string, error)
+	AdminOnlyVars(ctx context.Context, obj *model.APIProjectVars) ([]string, error)
+	PrivateVars(ctx context.Context, obj *model.APIProjectVars) ([]string, error)
 }
 type QueryResolver interface {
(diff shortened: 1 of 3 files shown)


malikchaya2 (approved):

ablack12:
We should definitely be redacting the values of private variables even in the logs. Does your screenshot include private variables? (Not sure what project this is for).

Also, not sure if this is your change or not but the event logs show more variables than were actually changed for some reason.
image
(In this case, I changed the display name and that's it, using https://spruce-staging.corp.mongodb.com/project/annie-copy2/settings/event-log). This seems like something that could be ticketed for a future improvement though.

sophstad:
> We should definitely be redacting the values of private variables even in the logs. Does your screenshot include private variables? (Not sure what project this is for).
> 
> Also, not sure if this is your change or not but the event logs show more variables than were actually changed for some reason. image (In this case, I changed the display name and that's it, using https://spruce-staging.corp.mongodb.com/project/annie-copy2/settings/event-log). This seems like something that could be ticketed for a future improvement though.

All private values are redacted! That screenshot just indicates that the order of the keys in the array was changing. All of those strings are private variable names, not values. And your screenshot also shows the problem before my fix since Julian is now on staging, ha. This ticket is what prevents those variables from appearing.

ablack12:
Ah perfect 👍

PR #5811, opened 2022-07-15
EVG-16989 Upgrade legacy agent routes to v2

hadjri:
[EVG-16989](https://jira.mongodb.org/browse/EVG-16989)

### Description 
Upgrading host agent routes to REST v2 to address the fact that the newly written pod agent routes are using REST v2. After hosts have had time to roll over we should be able to remove the existing REST v1 agent routes.

Much of this PR is copying over existing functionality and tweaking request parsing / error handling to be in line with REST v2 patterns.
### Testing 
Scheduled new versions in staging and checked splunk that the newly created agent routes are in use and succeeding, and that tasks are getting scheduled and completed properly. Updated unit tests.


Kimchelly (commented):
The migrated routes look good, just have some questions:
* Are [these routes](https://github.com/evergreen-ci/evergreen/pull/5811/files#diff-60d29311566609ebc0eb9df80ffb4b4d7ed91f4ebcb2fc46c56ba8570e79f51fR805) supposed to be part of the migration? They seem to be agent command routes.
* Is there a plan to migrate/copy-paste the legacy tests into the REST v2 routes?
* This isn't directly related to your changes, but a question about whether the migrated routes might behave slightly differently now. For task authentication, it seems like the REST v2 logic is slightly different from the legacy REST logic - if the task secret is invalid, [`TaskAuthMiddleware` converts `http.StatusConflict` into `http.StatusUnauthorized`](https://github.com/evergreen-ci/evergreen/blob/b1ff40cf5a5ebaa3e23bc290c2cc69bdf630546b/rest/data/task.go#L99-L104), whereas in the legacy middleware, [it returns `http.StatusConflict` as-is](https://github.com/evergreen-ci/evergreen/blob/b1ff40cf5a5ebaa3e23bc290c2cc69bdf630546b/service/api.go#L113-L117). I know the agent has handling for `http.StatusConflict` in the heartbeat when a task is aborted, but does the agent also handle `http.StatusUnauthorized`? Would this affect the behavior of the routes that are being ported over when a task is aborted, or does the agent correctly handle `http.StatusUnauthorized`?

hadjri:
> The migrated routes look good, just have some questions:
> 
> * Are [these routes](https://github.com/evergreen-ci/evergreen/pull/5811/files#diff-60d29311566609ebc0eb9df80ffb4b4d7ed91f4ebcb2fc46c56ba8570e79f51fR805) supposed to be part of the migration? They seem to be agent command routes. If we migrated those, I believe that would be all of the `apiVersion1` usages in the agent package, so we could remove the dependency on legacy v1 REST routes permanently.
> * Is there a plan to migrate/copy-paste the legacy tests into the REST v2 routes?
> * This isn't directly related to your changes, but a question about whether the migrated routes might behave slightly differently now. For task authentication, it seems like the REST v2 logic is slightly different from the legacy REST logic - if the task secret is invalid, [`TaskAuthMiddleware` converts `http.StatusConflict` into `http.StatusUnauthorized`](https://github.com/evergreen-ci/evergreen/blob/b1ff40cf5a5ebaa3e23bc290c2cc69bdf630546b/rest/data/task.go#L99-L104), whereas in the legacy middleware, [it returns `http.StatusConflict` as-is](https://github.com/evergreen-ci/evergreen/blob/b1ff40cf5a5ebaa3e23bc290c2cc69bdf630546b/service/api.go#L113-L117). I know the agent has handling for `http.StatusConflict` in the heartbeat when a task is aborted, but does the agent also handle `http.StatusUnauthorized`? Would this affect the behavior of the routes that are being ported over when a task is aborted, or does the agent correctly handle `http.StatusUnauthorized`?

- I hadn't initially planned to migrate these routes as a part of this ticket, though I've gone ahead and done that as well since it would be nice to remove the agent's v1 dependency entirely.
- I've now migrated the existing legacy REST v1 tests into rest/route package using v2, and have refactored them accordingly to no longer use the `Convey` testing syntax.
- Thank you for the catch, I've changed the `TaskAuthMiddleware` to now return `http.StatusConflict` as the status code instead of converting to `StatusUnauthorized`

rest/route/agent_test.go
```diff
@@ -84,3 +96,151 @@ func TestAgentCedarConfig(t *testing.T) {
...
+			tCase(ctx, t, r, s)
+		})
+	}
+}
+
+func TestDownstreamParams(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	if err := db.ClearCollections(patch.Collection, task.Collection, host.Collection); err != nil {
```
Kimchelly, line 180: nit: this could be simplified down to `require.NoError(db.ClearCollections(...))`.

rest/route/host_agent_test.go
```diff
@@ -0,0 +1,1698 @@
...
+				assert.NoError(t, db.DropCollections(colls...))
+			}()
+
+			env := &mock.Environment{}
+			require.NoError(t, env.Configure(ctx))
+			q, err := queue.NewLocalLimitedSizeSerializable(1, 1)
+			require.NoError(t, err)
+			env.Remote = q
+
+			if err := modelUtil.AddTestIndexes(host.Collection, true, true, host.RunningTaskKey); err != nil {
```
Kimchelly, line 747: nit: this could also be simplified down to `require.NoError`.

Kimchelly (commented):
Thanks for all the tech debt cleanup! 🚀

Kimchelly (approved):

Kimchelly:
LGTM mod tests passing.

PR #5812, opened 2022-07-20
EVG-14471 Fix admin events page

bynn:
[EVG-14471](https://jira.mongodb.org/browse/EVG-14471)

### Description 
1. the admin events route was broken because of presto so i do a manual check now 
2. created a new splunkconfig struct to make it more similar to the other configs
3. it now saves splunk config as a separate document in the admin collection 

### Testing 
unit tests 
image

### DIFF
--- a/config_splunk.go
+++ b/config_splunk.go
@@ -21,22 +21,20 @@ func (c *SplunkConfig) Get(env Environment) error {
 
 	res := coll.FindOne(ctx, byId(c.SectionId()))
 	if err := res.Err(); err != nil {
-		if err == mongo.ErrNoDocuments {
-			// Get the splunk config from global if the document doesn't exist.
-			globalConfig := coll.FindOne(ctx, byId(ConfigDocID))
-			if err := globalConfig.Err(); err != nil {
-				*c = SplunkConfig{}
-				return nil
-			} else {
-				s := Settings{}
-				if err := globalConfig.Decode(&s); err != nil {
-					return errors.Wrap(err, "decoding global config")
-				}
-				c.SplunkConnectionInfo = s.Splunk.SplunkConnectionInfo
-				return nil
-			}
+		if err != mongo.ErrNoDocuments {
+			return errors.Wrapf(err, "retrieving section '%s'", c.SectionId())
 		}
-		return errors.Wrapf(err, "retrieving section '%s'", c.SectionId())
+		// Get the splunk config from global if the document doesn't exist.
+		globalConfig := coll.FindOne(ctx, byId(ConfigDocID))
+		if err := globalConfig.Err(); err != nil {
+			return errors.Wrap(err, "retrieving global settings")
+		}
+		s := Settings{}
+		if err := globalConfig.Decode(&s); err != nil {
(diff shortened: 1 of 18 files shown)


malikchaya2 (approved):

config_splunk.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (c *SplunkConfig) Get(env Environment) error {
+	ctx, cancel := env.Context()
+	defer cancel()
+	coll := env.DB().Collection(ConfigCollection)
+
+	res := coll.FindOne(ctx, byId(c.SectionId()))
+	if err := res.Err(); err != nil {
+		if err == mongo.ErrNoDocuments {
+			// Get the splunk config from global if the document doesn't exist.
```
ablack12, line 25: Is this temporary code? Should we Set it in this case so that we can remove this code after staging/prod have been updated to set it in a separate document?

config_splunk.go
```diff
@@ -0,0 +1,66 @@
...
+			globalConfig := coll.FindOne(ctx, byId(ConfigDocID))
+			if err := globalConfig.Err(); err != nil {
+				*c = SplunkConfig{}
```
ablack12, line 28: Why not c = &SplunkConfig?

ablack12: I'm allo not sure why in this situation we wouldn't just return an error

config_splunk.go
```diff
@@ -0,0 +1,66 @@
...
+func (c *SplunkConfig) SectionId() string { return "splunk" }
+
+func (c *SplunkConfig) Get(env Environment) error {
+	ctx, cancel := env.Context()
+	defer cancel()
+	coll := env.DB().Collection(ConfigCollection)
+
+	res := coll.FindOne(ctx, byId(c.SectionId()))
+	if err := res.Err(); err != nil {
+		if err == mongo.ErrNoDocuments {
```
ablack12, line 24: It might read nicer to say `if err != mongo.ErrNoDocuments { return err }` so that the rest of the code can not be so nested.

config_splunk.go
```diff
@@ -0,0 +1,66 @@
...
+			// Get the splunk config from global if the document doesn't exist.
+			globalConfig := coll.FindOne(ctx, byId(ConfigDocID))
+			if err := globalConfig.Err(); err != nil {
+				*c = SplunkConfig{}
+				return nil
+			} else {
```
ablack12, line 30: no need for an else since the previous statement returns nil

ablack12 (changes requested):

config_splunk.go
bynn: this would only be called after a change to the admin page so i think this might need to be semi permanent

ablack12: Yeah I am thinking we should do something to the admin page to update it though so that we can remove this code. If it doesn't need to be something in the Splunk section we could just add / remove a dummy username to one of the host configs (like GCE) or we could add/remove something from the Splunk channel section.

bynn: https://jira.mongodb.org/browse/EVG-17353

PR #5813, opened 2022-07-21
EVG-16729 Allow different host type timeouts to be configurable on admin and distro pages

hadjri:
[EVG-16729](https://jira.mongodb.org/browse/EVG-16729)

### Description 
Timeouts for different types of host timeouts are currently hardcoded constants and the admin settings to set idle host timeout times are not in use. Change made removes those constants and reflects the host timeout values set on the admin or distro page in the idle host termination job for the distro.
### Testing 
Tested setting values on both pages in local-evergreen. Updated existing unit tests.


ablack12 (commented):
Sorry we weren't able to discuss this before but I think we should be doing the opposite -- it's better to have these constraints definable in the admin settings so we can tweak them if necessary, so if a constant is being used instead then we should remove that usage of the constant and make sure it's set in admin instead.

hadjri:
> Sorry we weren't able to discuss this before but I think we should be doing the opposite -- it's better to have these constraints definable in the admin settings so we can tweak them if necessary, so if a constant is being used instead then we should remove that usage of the constant and make sure it's set in admin instead.

Ah okay, that sounds good. Changed to remove the constants and since there were 3 different types of timeouts defined in the idle host termination job I've allowed all 3 of them to be set separately.

model/distro/distro.go
```diff
@@ -649,14 +654,16 @@ func (d *Distro) GetResolvedHostAllocatorSettings(s *evergreen.Settings) (HostAl
 	config := s.Scheduler
 	has := d.HostAllocatorSettings
 	resolved := HostAllocatorSettings{
-		Version:                has.Version,
-		MinimumHosts:           has.MinimumHosts,
-		MaximumHosts:           has.MaximumHosts,
-		AcceptableHostIdleTime: has.AcceptableHostIdleTime,
```
ablack12, line 655: This actually makes it look like the ticket description "AcceptableHostIdleTime is unused" is untrue.  Do we use this anywhere?

ablack12 (commented):
Sorry I didn't think about this deeper before but I didn't realize there'd be three values here. I'm a little worried this will clutter the distro page. Do we attempt to configure Acceptable host idle time in any distros right now? I wonder if these should only be definable on the admin page rather than the distro page. Alternatively maybe the task group / outdated times should just be factors of the regular time. We definitely can't have all constants in the codebase on the admin/distro page so we need to make sure there's a good case for the things we put there.

Sorry for all the questions but I may have advocated for the wrong approach before 😅

model/distro/distro.go
hadjri: This was also unused prior to this PR

ablack12: Resolving this thread bc I don't remember what I meant haha

hadjri:
> Sorry I didn't think about this deeper before but I didn't realize there'd be three values here. I'm a little worried this will clutter the distro page. Do we attempt to configure Acceptable host idle time in any distros right now? I wonder if these should only be definable on the admin page rather than the distro page. Alternatively maybe the task group / outdated times should just be factors of the regular time. We definitely can't have all constants in the codebase on the admin/distro page so we need to make sure there's a good case for the things we put there.
> 
> Sorry for all the questions but I may have advocated for the wrong approach before 😅 We can also discuss in DM.

I see what you mean. I think a decent solution is to put a single host acceptable idle time on the admin/distro page with the distro page taking precedence if set.  Then, task group and outdated host idle times can be factors of the value set in the UI.  I've made it such the task group time is twice as long, and outdated is half as long, since that roughly reflects what was already the case with the constants.

units/host_monitoring_idle_termination.go
```diff
@@ -165,11 +159,11 @@ func (j *idleHostJob) checkAndTerminateHost(ctx context.Context, h *host.Host, d
...
 	communicationTime := h.GetElapsedCommunicationTime()
 
-	idleThreshold := idleTimeCutoff
+	idleThreshold := d.HostAllocatorSettings.AcceptableHostIdleTime
 	if h.RunningTaskGroup != "" {
-		idleThreshold = idleTaskGroupHostCutoff
+		idleThreshold = idleThreshold * 2
 	} else if hostHasOutdatedAMI(*h, d) {
-		idleThreshold = outdatedIdleTimeCutoff
+		idleThreshold = idleTime / 2
```
ablack12, line 166: This change does increase the idleThreshhold from 1 minute to 2 minutes, right? I think one minute is a standard enough duration that we don't need it to be based on the host allocator setting.

ablack12 (approved):
LGTM with change

hadjri:
> LGTM with change; we'll also need to add this to the admin settings before deploy I think

We already have host acceptable idle seconds on the admin page in prod so I don't think any action is necessary. We can bring it from 5 minutes down to 4 though.
image

PR #5814, opened 2022-07-21
EVG-17331: Use proper sort key in FindFirstProjectRef.

johndaniels:
[EVG-17331](https://jira.mongodb.org/browse/EVG-17331)

### Description 
I discovered this during local setup, since local setup's development fixtures don't set a default project. This occurs for anonymous users without any cookies hitting the waterfall endpoint

### Testing 
Added unit tests to verify that the function doesn't fail. Also verified locally by clearing cookies and hitting localhost:9090

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1141,27 +1141,27 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 }
 
 func FindFirstProjectRef() (*ProjectRef, error) {
-	projectRefSlice := &[]ProjectRef{}
+	projectRefSlice := []ProjectRef{}
 	pipeline := projectRefPipelineForValueIsBool(ProjectRefPrivateKey, RepoRefPrivateKey, false)
 	pipeline = append(pipeline, bson.M{"$sort": bson.M{ProjectRefPrivateKey: -1}}, bson.M{"$limit": 1})
 	err := db.Aggregate(
 		ProjectRefCollection,
 		pipeline,
-		projectRefSlice,
+		&projectRefSlice,
 	)
 
 	if err != nil {
 		return nil, errors.Wrap(err, "aggregating project ref")
 	}
 
-	if len(*projectRefSlice) == 0 {
+	if len(projectRefSlice) == 0 {
 		return nil, errors.New("No project found in FindFirstProjectRef")
 	}
-	projectRef := &(*projectRefSlice)[0]
+	projectRef := projectRefSlice[0]
 
 	projectRef.checkDefaultLogger()
 
-	return projectRef, nil
+	return &projectRef, nil
 }
 
 // FindAllMergedTrackedProjectRefs returns all project refs in the db
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1143,7 +1143,7 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 func FindFirstProjectRef() (*ProjectRef, error) {
 	projectRefSlice := []ProjectRef{}
 	pipeline := projectRefPipelineForValueIsBool(ProjectRefPrivateKey, RepoRefPrivateKey, false)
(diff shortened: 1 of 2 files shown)


model/project_ref_test.go
```diff
@@ -1847,6 +1847,25 @@ func TestGetProjectSetupCommands(t *testing.T) {
 	assert.Contains(t, cmds[1].String(), "c1")
 }
 
+func TestFindFirstProjectRef(t *testing.T) {
+	assert.NoError(t, db.ClearCollections(ProjectRefCollection))
+
+	_, err := FindFirstProjectRef()
+	assert.Error(t, err, "Should return error and not panic if there are no matching projects")
```
ybrill, line 1854: I think assert has `NotPanics` if we want to test that it doesn't panic. I think we don't usually concern ourselves to test for it explicitly, though, but idk why. Maybe because a panic would also abort the test and cause the test to return non zero?

model/project_ref.go
```diff
@@ -1141,18 +1141,24 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
...
-	projectRef := &ProjectRef{}
+	projectRefSlice := &[]ProjectRef{}
 	pipeline := projectRefPipelineForValueIsBool(ProjectRefPrivateKey, RepoRefPrivateKey, false)
-	pipeline = append(pipeline, bson.M{"$sort": "-" + ProjectRefDisplayNameKey}, bson.M{"$limit": 1})
+	pipeline = append(pipeline, bson.M{"$sort": bson.M{ProjectRefPrivateKey: -1}}, bson.M{"$limit": 1})
 	err := db.Aggregate(
 		ProjectRefCollection,
 		pipeline,
-		projectRef,
+		projectRefSlice,
```
ybrill, line 1150: would it work for `projectRefSlice` to not be a pointer and the pass in a pointer to it here?

model/project_ref_test.go
johndaniels: Yeah I assume that panicking automatically errors out the test. I can explicitly test that though.

johndaniels: I added a NotPanics check... I mainly wanted to assert that it didn't panic because panicking would have been the result of me forgetting to do a bounds check on the resultant array, but that might just be kind of pointless?

model/project_ref.go
```diff
@@ -1141,21 +1141,27 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 }
 
 func FindFirstProjectRef() (*ProjectRef, error) {
-	projectRef := &ProjectRef{}
+	projectRefSlice := []ProjectRef{}
 	pipeline := projectRefPipelineForValueIsBool(ProjectRefPrivateKey, RepoRefPrivateKey, false)
-	pipeline = append(pipeline, bson.M{"$sort": "-" + ProjectRefDisplayNameKey}, bson.M{"$limit": 1})
+	pipeline = append(pipeline, bson.M{"$sort": bson.M{ProjectRefPrivateKey: -1}}, bson.M{"$limit": 1})
```
ablack12, line 1146: Why did you switch what this is sorting by? It seems to me like the issue in the ticket is the nil pointer reference that you fixed by checking the length of the slice, but we still want to get the first alphabetical project.

model/project_ref.go
johndaniels: Yeah I got rid of a lot of the pointer usage. Go's similarity to C made me weary of things like taking the address of a "stack variable" but the Go docs tell me that is ok.

model/project_ref.go
johndaniels: 🤦 This is fixed. I did not mean to change the key. Good catch!

To be clear, the original issue wasn't the nil pointer reference, the issue was that "-DisplayName" is not valid mongodb bson, but the old driver used to allow it.

Fixing things for the new driver, we also had to switch to using an array since the new driver requires an array for the "out parameter" of the "Aggregate" function, which is also why I added the bounds checking.

ablack12: Ah gotcha, cool!

ablack12 (approved):

PR #5815, opened 2022-07-22
EVG-17076 Refactor addBuildVariantDisplayName to use cached `BuildVariantDisplayName` field if it exists.

khelif96:
[EVG-17076](https://jira.mongodb.org/browse/EVG-17076)

### Description 
Reverts the revert of EVG-16888. 
Refactor `AddAnnotation` into its own variable to clean up this aggregation.
ensure graphql uses `buildVariantDisplayName` from the task doc if it exists instead of trying to query for it.

<del>If after merging this and we discover we are still hitting the 16 mb doc limit on prod we can explore replacing the `$facet` with an exploratory query that determines which code path we use.</del>

jk. Gonna go with the exploratory query instead. The aggregation failed at the `$facet` stage when trying it on the prod db. Now we no longer do the `AddBuildVariantDisplayName`  `$facet` and instead have an additional query to determine if we should perform the `$lookup` before the aggregation

e.g
```
task := get1TaskByVersion(versionID) 

if hasBuildVariantDisplayNameField(task) {
       doQueryWithOutLegacyFallback(versionID)
}else{
       doQueryWithLegacyMethod(versionID)
}
```
### Testing 
Attempted this query on a [patch](https://spruce-staging.corp.mongodb.com/version/62db074797b1d369e3e28ded/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with 10000+ tasks and could not get it to hit the 16mb limit.


model/task/task.go
```diff
@@ -3869,18 +3902,13 @@ func HasMatchingTasks(versionID string, opts HasMatchingTasksOptions) (bool, err
...
-
-		match = bson.M{
-			"$or": []bson.M{
-				{BuildVariantDisplayNameKey: bson.M{"$regex": variantsAsRegex, "$options": "i"}},
-				{BuildVariantKey: bson.M{"$regex": variantsAsRegex, "$options": "i"}},
-			},
-		}
+	if !opts.IncludeBuildVariantDisplayName && opts.UseLegacyAddBuildVariantDisplayName {
+		grip.Alert(message.Fields{
+			"message": "Implementer error: should not use UseLegacyAddBuildVariantDisplayName with !IncludeBuildVariantDisplayName",
```
malikchaya2, line 3907: we should probably call this programmer error to be consistent with similar errors throughout the codebase.

malikchaya2 (changes requested):

model/task/task.go
khelif96: Thanks i knew there was another term for it but my grep skills weren't up to the task.

model/task/task.go
```diff
@@ -3469,25 +3468,29 @@ type GetTasksByProjectAndCommitOptions struct {
...
+	IsMainlineCommit                    bool
+	UseLegacyAddBuildVariantDisplayName bool
 }
 
 // GetTasksByVersion gets all tasks for a specific version
 // Query results can be filtered by task name, variant name and status in addition to being paginated and limited
 func GetTasksByVersion(versionID string, opts GetTasksByVersionOptions) ([]Task, int, error) {
-
+	if opts.IncludeBuildVariantDisplayName {
+		shouldUseLegacyAddBuildVariantDisplayName := ShouldUseLegacyAddBuildVariantDisplayName(versionID)
```
hadjri, line 3491: I think this can be simplified to one line

model/task/task.go
```diff
@@ -3600,6 +3603,10 @@ type TaskStats struct {
 }
 
 func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) (*TaskStats, error) {
+	if opts.IncludeBuildVariantDisplayName {
+		shouldUseLegacyAddBuildVariantDisplayName := ShouldUseLegacyAddBuildVariantDisplayName(versionID)
```
hadjri, line 3607: Same comment here

model/task/task.go
```diff
@@ -3671,13 +3678,33 @@ func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) (*Ta
...
+func ShouldUseLegacyAddBuildVariantDisplayName(versionID string) bool {
+	query := db.Query(ByVersion(versionID))
+	task, err := FindOne(query)
+	if err != nil {
+		return false
+	}
+	if task == nil {
+		return false
+	}
+	if task.BuildVariantDisplayName != "" {
```
hadjri, line 3692: How about `return task.BuildVariantDisplayName == ""`

model/task/task.go
```diff
@@ -3844,6 +3871,10 @@ func HasMatchingTasks(versionID string, opts HasMatchingTasksOptions) (bool, err
 		Statuses:                       opts.Statuses,
 		IncludeBuildVariantDisplayName: true,
 	}
+	if len(opts.Variants) > 0 {
+		shouldUseLegacyAddBuildVariantDisplayName := ShouldUseLegacyAddBuildVariantDisplayName(versionID)
```
hadjri, line 3875: Same here

model/task/task.go
```diff
@@ -3671,13 +3678,33 @@ func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) (*Ta
...
+
 type GroupedTaskStatusCount struct {
 	Variant      string         `bson:"variant"`
 	DisplayName  string         `bson:"display_name"`
 	StatusCounts []*StatusCount `bson:"status_counts"`
 }
 
 func GetGroupedTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) ([]*GroupedTaskStatusCount, error) {
+	opts.IncludeBuildVariantDisplayName = true
+	shouldUseLegacyAddBuildVariantDisplayName := ShouldUseLegacyAddBuildVariantDisplayName(versionID)
```
hadjri, line 3706: Same here

hadjri (approved):
LGTM with just a few small comments

malikchaya2 (approved):

model/task/task.go
```diff
@@ -3869,18 +3894,15 @@ func HasMatchingTasks(versionID string, opts HasMatchingTasksOptions) (bool, err
...
-		variantsAsRegex := strings.Join(opts.Variants, "|")
-
-		match = bson.M{
-			"$or": []bson.M{
-				{BuildVariantDisplayNameKey: bson.M{"$regex": variantsAsRegex, "$options": "i"}},
-				{BuildVariantKey: bson.M{"$regex": variantsAsRegex, "$options": "i"}},
-			},
-		}
+	if !opts.IncludeBuildVariantDisplayName && opts.UseLegacyAddBuildVariantDisplayName {
+		grip.Alert(message.Fields{
```
ablack12, line 3898: Do we expect this to get hit often in the case that the programmer error exists? This will log to ops-alerts and spam that requires a revert can be excessive. If we do want to log to ops-alerts it may be preferable to make this a regular error and create a splunk alert on it so we can throttle / turn it off if needed, or else just code read carefully to ensure we aren't in this position / maybe have an opts validation stage.

khelif96: This doesn't seem to be really consistent through out the codebase. https://github.com/evergreen-ci/evergreen/blob/2a2b5f433c464af9361753b5624bdcb8f572a779/model/patch/patch.go#L1068 
https://github.com/evergreen-ci/evergreen/blob/2a2b5f433c464af9361753b5624bdcb8f572a779/trigger/registry.go#L39
https://github.com/evergreen-ci/evergreen/blob/2a2b5f433c464af9361753b5624bdcb8f572a779/units/event_notifier.go#L250
https://github.com/evergreen-ci/evergreen/blob/2a2b5f433c464af9361753b5624bdcb8f572a779/units/host_monitoring_idle_termination.go#L196
 . Since either of these values can be dynamic and we use this function in so many places it will be difficult to notice this in code reviews.

khelif96: Done replaced this with an error and some validation.

PR #5816, opened 2022-07-25
EVG-17126 Do not allow resetting non-complete tasks

hadjri:
[EVG-17126](https://jira.mongodb.org/browse/EVG-17126)

### Description 
_(This is a boat load of investigation context and I apologize for that)_

We have been seeing occasional [radical spikes](https://mongodb.splunkcloud.com/en-US/app/search/search?display.page.search.mode=verbose&dispatch.sample_ratio=1&q=search%20index%3Devergreen%20metadata.level%20%3E%3D%2070%20%22wrong%20secret%22%0A%7C%20timechart%20count&earliest=1654117800&latest=1657833600&workload_pool=standard_perf&display.page.search.tab=visualizations&display.general.type=visualizations&sid=1658259124.1997534) of 'wrong task secret' error rates on specific days, where the occurrences can reach 100,000+ errors rapidly whereas most other days have far fewer occurrences around 1,000 total.

Looking at the bulk of problematic tasks during these spikes they all show a strange behavior where there are many task executions with the exact same timespan and a final execution that is of a different time, such as [this task](https://spruce.mongodb.com/task/mongo_node_driver_next_macos_1014_fermium_test_6.0_server_3545698bae6ac07dd9edadf0b66afba696eb83fb_22_07_01_17_18_23/logs?execution=5&logtype=task).  Specifically here we see 6 executions where only the first and last execution actually ran, the other clones of the first execution were created as documents but did not execute.

Looking at this task's specific version we can see a [user rapidly restarted the version over and over many times](https://mongodb.splunkcloud.com/en-US/app/search/search?display.page.search.mode=verbose&dispatch.sample_ratio=1&q=search%20index%3Devergreen%20mongo_node_driver_next_3545698bae6ac07dd9edadf0b66afba696eb83fb%20%7C%20spath%20user%20%7C%20search%20user%3D%22neal.beeken%22%7C%20spath%20method%20%7C%20search%20method!%3DGET&earliest=1656705000&latest=1656706200&workload_pool=standard_perf&display.page.search.tab=events&display.general.type=events&sid=1658764665.3058584). We wouldn't normally expect someone spamming the restart button to create a bunch of tasks that never run because logic is currently set up to only restart finished tasks.  However taking a closer look at those user manual requests you can see they took much longer than typical to complete most of them 5-15 minutes to complete the request.

What looks like happened is that due to the combination of user clicking restart rapidly in sequence and the requests taking long to complete, multiple version restarts were accepted as database modifications at the same time since multiple requests were sent in before the first request had updated the version task's DB state, thus evergreen accepted these restarts on the same state of the same version.

The task's event logs corroborate this:
Screen Shot 2022-07-25 at 11 51 53 AM

Multiple consecutive restart event logs should not be possible as far as I've been able to see.  It seems that resetting the tasks is what was bottlenecking the slow requests, so what can happen is many `RestartVersion` invocations begin, the same task is archived many times in a row on each request (which would explain why these tasks have multiple executions with identical timestamps), one of the request successfully resets all of the version's tasks which causes an agent elsewhere to pick up the tasks, but then the other queued version restart requests reset the tasks again after the tasks have already been re-dispatched. You can see this at the bottom of the event logs screenshot where a task gets restarted after it was just dispatched and started.  The timestamp of that last event log lines up right at the time where 'task conflict' errors begin flooding in for this task.  Also, looking at the full event logs in the DB, the conflict errors _stop_ erroring once the last version restart request completes which probably means the task was reset and had its secret was changed for the last time, allowing it to finally complete.

For all of the 'task conflict' spikes shown in the first link, they all line up with versions that have been restarted several times in a row and were all slow to complete which probably caused these race conditions and therefore repeating conflict errors for every task in the restarted version.  If they're big versions, it'd explain the large number of errors we see on the graph.

One way to remedy these issues since it is not obvious why these requests occasionally get slow is to do more to ensure that tasks are only reset if they are in a completed state since that is what these race conditions boil down to. Change I've made here in this PR is taking that approach.

I've removed some unused parameters that I noticed along the way as well.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1954,23 +1954,27 @@ func (t *Task) Reset() error {
 
 // ResetTasks performs the same DB updates as (*Task).Reset, but resets many
 // tasks instead of a single one.
-func ResetTasks(tasks []Task) error {
+func ResetTasks(tasks []Task) (*adb.ChangeInfo, error) {
 	if len(tasks) == 0 {
-		return nil
+		return nil, nil
 	}
 	var taskIDs []string
 	for _, t := range tasks {
 		taskIDs = append(taskIDs, t.Id)
 	}
 
-	if _, err := UpdateAll(
-		bson.M{IdKey: bson.M{"$in": taskIDs}},
+	info, err := UpdateAll(
+		bson.M{
+			IdKey:     bson.M{"$in": taskIDs},
+			StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
+		},
 		resetTaskUpdate(nil),
-	); err != nil {
-		return err
+	)
+	if err != nil {
+		return nil, err
 	}
 
-	return nil
+	return info, nil
 }
 
 func resetTaskUpdate(t *Task) bson.M {
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2511,7 +2511,7 @@ func (t *Task) Archive() error {
 			},
 			"$inc": bson.M{ExecutionKey: 1},
 		})
-	if err != nil {
+	if err != nil && !adb.ResultsNotFound(err) {
 		return errors.Wrap(err, "updating task")
 	}
 
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1955,27 +1955,26 @@ func (t *Task) Reset() error {
 
 // ResetTasks performs the same DB updates as (*Task).Reset, but resets many
 // tasks instead of a single one.
-func ResetTasks(tasks []Task) (*adb.ChangeInfo, error) {
+func ResetTasks(tasks []Task) error {
 	if len(tasks) == 0 {
-		return nil, nil
+		return nil
 	}
 	var taskIDs []string
 	for _, t := range tasks {
 		taskIDs = append(taskIDs, t.Id)
 	}
 
-	info, err := UpdateAll(
+	if _, err := UpdateAll(
 		bson.M{
 			IdKey:     bson.M{"$in": taskIDs},
 			StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
 		},
 		resetTaskUpdate(nil),
-	)
-	if err != nil {
-		return nil, err
+	); err != nil {
+		return err
 	}
 
-	return info, nil
+	return nil
 }
 
 func resetTaskUpdate(t *Task) bson.M {
@@ -2002,6 +2001,7 @@ func resetTaskUpdate(t *Task) bson.M {
 		t.HostCreateDetails = []HostCreateDetail{}
 		t.OverrideDependencies = false
 		t.ContainerAllocationAttempts = 0
+		t.Archived = false
 	}
 	update := bson.M{
 		"$set": bson.M{
@@ -2017,6 +2017,7 @@ func resetTaskUpdate(t *Task) bson.M {
 			TimeTakenKey:                   0,
 			LastHeartbeatKey:               utility.ZeroTime,
 			ContainerAllocationAttemptsKey: 0,
+			ArchivedKey:                    false,
 		},
 		"$unset": bson.M{
 			DetailsKey:                 "",
@@ -2475,7 +2476,7 @@ func (t *Task) Insert() error {
 // considered the latest execution. This task execution is inserted
 // into the old_tasks collection. If this is a display task, its execution tasks
 // are also archived.
-func (t *Task) Archive() (bool, error) {
+func (t *Task) Archive() error {
 	archiveTask := t.makeArchivedTask()
 	err := db.Insert(OldCollection, archiveTask)
 	if err != nil {
@@ -2485,14 +2486,17 @@ func (t *Task) Archive() (bool, error) {
 			"execution":       t.Execution,
 			"display_only":    t.DisplayOnly,
 		}))
-		return false, errors.Wrap(err, "inserting archived task into old tasks")
+		return errors.Wrap(err, "inserting archived task into old tasks")
 	}
-	info, err := UpdateAll(
+	err = UpdateOne(
 		bson.M{
 			IdKey:     t.Id,
 			StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
 		},
 		bson.M{
+			"$set": bson.M{
+				ArchivedKey: true,
+			},
 			"$unset": bson.M{
 				AbortedKey:              "",
 				AbortInfoKey:            "",
@@ -2501,23 +2505,20 @@ func (t *Task) Archive() (bool, error) {
 			"$inc": bson.M{ExecutionKey: 1},
 		})
 	if err != nil {
-		return false, errors.Wrap(err, "updating task")
-	}
-	if info.Updated <= 0 {
-		return true, nil
+		return errors.Wrap(err, "updating task")
 	}
 	// only archive execution tasks after we confirm that the task was in a completed state
 	if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
 		execTasks, err := FindAll(db.Query(ByIds(t.ExecutionTasks)))
 		if err != nil {
-			return false, errors.Wrap(err, "retrieving execution tasks")
+			return errors.Wrap(err, "retrieving execution tasks")
 		}
 		if err = ArchiveMany(execTasks); err != nil {
-			return false, errors.Wrap(err, "archiving execution tasks")
+			return errors.Wrap(err, "archiving execution tasks")
 		}
 	}
 	t.Aborted = false
-	return false, nil
+	return nil
 }
 
 func ArchiveMany(tasks []Task) error {
@@ -2581,6 +2582,9 @@ func ArchiveMany(tasks []Task) error {
 			StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
 		},
 			bson.M{
+				"$set": bson.M{
+					ArchivedKey: true,
+				},
 				"$unset": bson.M{
 					AbortedKey:   "",
 					AbortInfoKey: "",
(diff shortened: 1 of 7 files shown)


Kimchelly (commented):
Two questions:
1. If the problem is that we're archiving + resetting tasks that aren't in a completed state, should the archive queries also check the status before attempting to archive? Otherwise, an incomplete task might be archived without being reset to run again, which is unexpected behavior for restarting a task.
2. Since resetting many tasks only happens if the status is complete, is [resetting a single task](https://github.com/evergreen-ci/evergreen/blob/3f1c4bf60d2d1ee79b5d9720291d8e0d9c9ad697/model/task/task.go#L1948-L1950) supposed to do the same? It seems like the change on [L294-296](https://github.com/evergreen-ci/evergreen/pull/5816/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R294) is still prone to races if there are several restart requests going on at the same time because it doesn't check the DB state at the moment that the task is reset.

hadjri:
> Two questions:
> 
> 1. If the problem is that we're archiving + resetting tasks that aren't in a completed state, should the archive queries also check the status before attempting to archive? Otherwise, an incomplete task might be archived without being reset to run again, which is unexpected behavior for restarting a task.
> 2. Since resetting many tasks only happens if the status is complete, is [resetting a single task](https://github.com/evergreen-ci/evergreen/blob/3f1c4bf60d2d1ee79b5d9720291d8e0d9c9ad697/model/task/task.go#L1948-L1950) supposed to do the same? It seems like the change on [L294-296](https://github.com/evergreen-ci/evergreen/pull/5816/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R294) is still prone to races if there are several restart requests going on at the same time because it doesn't check the completion state of the task in the DB when doing the reset update.

1. I've changed archiving operations to only perform DB operations on completed tasks
2. I've modified the single task restart logic to keep track of whether the DB was updated when archiving is attempted. If no update is made, we'll return early and skip resetting which should address the issue. I've added a debug log to track this as well.

model/task_lifecycle.go
```diff
@@ -1368,9 +1377,13 @@ func MarkTasksReset(taskIds []string) error {
 		return errors.WithStack(err)
 	}
 
-	if err = task.ResetTasks(tasks); err != nil {
+	info, err := task.ResetTasks(tasks)
+	if err != nil {
 		return errors.Wrap(err, "resetting tasks in database")
 	}
+	if info.Updated == 0 {
```
Kimchelly, line 1384: Same qusetion as above about idempotency of operations like marking a task reset. If we exit early here, we might be unable to retry updating dependencies.

model/task/task.go
```diff
@@ -2503,10 +2501,23 @@ func (t *Task) Archive() error {
 			"$inc": bson.M{ExecutionKey: 1},
 		})
 	if err != nil {
-		return errors.Wrap(err, "updating task")
+		return false, errors.Wrap(err, "updating task")
+	}
+	if info.Updated <= 0 {
```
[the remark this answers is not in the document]
hadjri: No, it was just a stylistic choice. Now changed to check for `adb.ResultsNotFound(err)`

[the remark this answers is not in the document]
hadjri: I see what you mean. I may be misinterpreting but `Archived = true` does not currently happen on the task during archived, it just gets set on the archived task that is sent to the old tasks collection. However I've now made the change to set `Archived` in the DB on the task doc during archiving operation and adding that to the query done before we reset the tasks. So that should make it so that if the reset task functions fail between archiving and resetting, we can retry the operation and it'll succeed.

model/task_lifecycle.go
hadjri: Understood, following same pattern as resetting a single task now.

model/task/task.go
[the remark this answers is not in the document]
Kimchelly: Ah yeah, I didn't notice that archive is only set on the old task. Your change looks good though!

model/task/task.go
```diff
@@ -1946,7 +1946,9 @@ func (t *Task) displayTaskPriority() int {
 func (t *Task) Reset() error {
 	return UpdateOne(
 		bson.M{
-			IdKey: t.Id,
+			IdKey:       t.Id,
+			StatusKey:   bson.M{"$in": evergreen.TaskCompletedStatuses},
+			ArchivedKey: true,
```
ablack12, line 1951: I'm a bit uncertain why Archived is included in this reset. Wouldn't we only want to reset if it _hasn't_ been archived yet?

ablack12: Archived is only set on the old task collection I thought. I'm not certain how this would ever hit something since this is for the current collection so I'm assuming I'm missing some context.

model/task/task.go
```diff
@@ -2012,6 +2019,7 @@ func resetTaskUpdate(t *Task) bson.M {
 			TimeTakenKey:                   0,
 			LastHeartbeatKey:               utility.ZeroTime,
 			ContainerAllocationAttemptsKey: 0,
+			ArchivedKey:                    false,
```
ablack12, line 2022: It would make sense to just put this at unset.

model/task/task.go
```diff
@@ -2493,8 +2491,14 @@ func (t *Task) Archive() error {
...
 	}
 	err = UpdateOne(
-		bson.M{IdKey: t.Id},
 		bson.M{
+			IdKey:     t.Id,
+			StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
+		},
+		bson.M{
+			"$set": bson.M{
+				ArchivedKey: true,
```
ablack12, line 2500: This seems to be misusing the usage of Archive. Archive is supposed to be set only on tasks that are being inserted into the old task collection (i.e. makeArchivedTask).

ablack12 (changes requested):
We actually know from investigation during that outage that users were _not_ spamming the restart button; it seemed instead like an evergreen internal problem. Have we seen this recur since the outage? Otherwise I'm not sure a whole new field for archive is right.

model/task/task.go
Kimchelly: This is necessary for idempotency since we only want to reset the current execution if it was preceded by an archive: https://github.com/evergreen-ci/evergreen/pull/5816#discussion_r933348376

model/task/task.go
hadjri: Yes currently the Archived field is unused in the active tasks collection but this change is extending the use of the Archived field to be used as an indicator that the task is in a state where it is safe to reset and doing so will not result in task conflict errors.

hadjri:
> We actually know from investigation during that outage that users were _not_ spamming the restart button; it seemed instead like an evergreen internal problem. Have we seen this recur since the outage? Otherwise I'm not sure a whole big change is needed.

I see, I think regardless of whether the requests came in manually or not I think we need to be resilient to this situation where many restart requests rapidly come in quick succession (for all evergreen types not just versions), or the case where a task somehow gets reset in an unfinished state due to another bug we're unaware of. 

Resetting an unfinished task is not something that should be able to happen (it will cause a burst of task conflict errors) and there aren't strong guards against it happening prior to this change.  I think once this is resolved it's possible we'll see a further drop in task conflict errors generally.

model/task/task.go
ablack12: This logic all makes sense -- my concern with reusing Archived is that right now we use that field to understand if something is something is in the old_tasks collection or not (for example, here https://github.com/evergreen-ci/evergreen/blob/c8ac8688404d6dd1ac0e420e206238ef5895b98e/model/task/task.go#L2651 we use OldTaskId if Archived is set). 

Should we start using OldTaskId to indicate if something is in the old_tasks collection instead? i.e. change existing usage of `Archived == true` to `OldTaskId != ""` ? This may also limit confusion with this field meaning multiple things.

ablack12 (changes requested):

model/task/task.go
Kimchelly: I agree about not wanting multiple meaning, but I think `OldTaskId` is only set for the task execution in the old task collection and not for the latest task execution in the regular task collection (i.e. the task doc we want to reset). We want a flag on the latest task indicating [this update](https://github.com/evergreen-ci/evergreen/blob/c8ac8688404d6dd1ac0e420e206238ef5895b98e/model/task/task.go#L2509-L2521) succeeded so that the latest task is in a "waiting to reset" state.

hadjri: Although that solution makes sense I feel that we can keep this solution simpler without introducing cross collection lookups and aggregations. We'd also need to ensure that the execution incrementation and variable unsetting on the task in the second update linked above is handled properly if we remove the DB update altogether.  

Since the multiple meaning for `Archived` was the initial concern I've added a new single bool `CanReset` on the task doc to indicate if it's in a safe state to reset. This is now replacing all of the usages of Archived we were using in this PR.

ablack12: Sorry yeah I wasn't suggesting we use OldTaskId to handle the stuff in this PR but rather to replace the _old_ use of Archived to check whether a document was in the tasks or old_tasks collection.

model/task/task.go
```diff
@@ -151,6 +151,9 @@ type Task struct {
 	// Set to true if the task should be considered for mainline github checks
 	IsGithubCheck bool `bson:"is_github_check,omitempty" json:"is_github_check,omitempty"`
 
+	// CanReset indicates if the task is in a valid state to be reset.
+	CanReset bool `bson:"can_reset,omitempty" json:"can_reset,omitempty"`
```
ablack12, line 155: Would you mind adding a comment here or to the reset functions elaborating on this a bit? (i.e. that a valid state to be reset is that it's been archived successfully)

ablack12 (approved):

PR #5818, opened 2022-07-25
EVG-13661 Limit the create project route 

malikchaya2:
[EVG-13661](https://jira.mongodb.org/browse/EVG-13661)

### Description 
This limits the create project route to only set the fields that the create project button on the UI allows. 

### Testing 
Existing tests pass. 

Will update the documentation once this is merged.


Kimchelly (approved):

rest/route/project.go
```diff
@@ -607,8 +607,12 @@ func (h *projectIDPutHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.MakeJSONErrorResponder(errors.Wrap(err, "unmarshalling JSON request body into project ref"))
 	}
 
-	dbProjectRef := apiProjectRef.ToService()
-	dbProjectRef.Identifier = h.projectName
+	dbProjectRef := dbModel.ProjectRef{
+		Identifier: h.projectName,
+		Id:         utility.FromStringPtr(apiProjectRef.Id),
```
ablack12, line 612: Can we add validation for the owner? (i.e. that it's within the allowed organizations)

ablack12 (changes requested):

rest/route/project.go
malikchaya2: Done, and I added it to the graphql resolver as well.

graphql/mutation_resolver.go
```diff
@@ -371,6 +371,16 @@ func (r *mutationResolver) AttachProjectToRepo(ctx context.Context, projectID st
 func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.APIProjectRef) (*restModel.APIProjectRef, error) {
 	dbProjectRef := project.ToService()
 	u := gimlet.GetUser(ctx).(*user.DBUser)
+
+	config, err := evergreen.GetConfig()
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("getting evergreen config: %s", err.Error()))
+	}
+
+	if err := dbProjectRef.ValidateOwnerAndRepo(config.GithubOrgs); err != nil {
```
ablack12, line 380: Can we just add both of these statements to data.CreateProject since both places need to check it? (I think it's fine to re-get the config in the UI case rather than using the passed in one)

ablack12: No need to block on another review from me when this is done

malikchaya2: Done. I'm assuming you meant that it's okay to get the config in the rest case?

PR #5819, opened 2022-07-25
EVG-17209: Avoid directly updating expiration time in EditSpawnHost

minnakt:
[EVG-17209](https://jira.mongodb.org/browse/EVG-17209)

### Description 
Currently we update the `ExpirationTime` db field in [`EditSpawnHost`](https://github.com/evergreen-ci/evergreen/blob/c00b71f7826af0ef870848b6011a8bb542826891/graphql/mutation_resolver.go#L545-L549) without checking that it is valid first. This sometimes causes us to incorrectly update the expiration time in the database despite encountering an error. To see this in action, you can try to extend any host's expiration date past 14 days. You will get an error saying that the new expiration date is not valid, but the UI will have been updated to the new date.

This PR makes it so that [`ModifyHost`](https://github.com/evergreen-ci/evergreen/blob/c00b71f7826af0ef870848b6011a8bb542826891/cloud/ec2.go#L896-L902), which checks if the new expiration time is valid, handles the `ExpirationTime` db updates instead.

### Testing 
- Modified `TestModifyHost` in `ec2_test.go`
- On staging


cloud/ec2_test.go
```diff
@@ -757,26 +757,51 @@ func (s *EC2Suite) TestModifyHost() {
...
 
-	intent := host.Host{
-		Id:           "evg-1234",
-		NoExpiration: false,
+	// intent := host.Host{
+	// 	Id:             "evg-1234",
+	// 	NoExpiration:   false,
+	// 	CreationTime:   time.Now(),
+	// }
+	// s.NoError(intent.Insert())
```
minnakt, line 778: I commented this out because It didn't seem like it was being used 🤔 Can I delete it?

minnakt, line 778: I commented this out because it didn't seem like it was being used 🤔 Can I delete it?

Kimchelly: Yeah, this should be safe to delete since the host being modified is `s.h`.

Kimchelly (approved):

PR #5820, opened 2022-07-26
EVG-17140  Remove user settings as part of offboarding job.

malikchaya2:
[EVG-17140](https://jira.mongodb.org/browse/EVG-17140)

### Description 
This was causing issues for returning users, specifically the github user Id. 

### Testing 
Added a unit test for the db function.

### DIFF
--- a/model/user/db.go
+++ b/model/user/db.go
@@ -414,6 +414,25 @@ func ClearLoginCache(user gimlet.User) error {
 	return nil
 }
 
+// ClearUserSettings clears one user's settings
+func ClearUserSettings(userId string) error {
+	update := bson.M{"$unset": bson.M{SettingsKey: 1}}
+
+	u, err := FindOneById(userId)
+	if err != nil {
+		return errors.Wrapf(err, "finding user '%s' by ID", userId)
+	}
+	if u == nil {
+		return errors.Errorf("user '%s' not found", userId)
+	}
+	query := bson.M{IdKey: userId}
+	if err := UpdateOne(query, update); err != nil {
+		return errors.Wrap(err, "unsetting user settings")
+	}
+
+	return nil
+}
+
 // ClearAllLoginCaches clears all users' login caches, forcibly logging them
 // out.
 func ClearAllLoginCaches() error {
--- a/model/user/db.go
+++ b/model/user/db.go
@@ -416,15 +416,7 @@ func ClearLoginCache(user gimlet.User) error {
 
 // ClearUserSettings clears one user's settings
 func ClearUserSettings(userId string) error {
-	update := bson.M{"$unset": bson.M{SettingsKey: 1}}
-
-	u, err := FindOneById(userId)
-	if err != nil {
-		return errors.Wrapf(err, "finding user '%s' by ID", userId)
-	}
-	if u == nil {
-		return errors.Errorf("user '%s' not found", userId)
-	}
+	update := bson.M{"$set": bson.M{SettingsKey: bson.M{}}}
 	query := bson.M{IdKey: userId}
 	if err := UpdateOne(query, update); err != nil {
 		return errors.Wrap(err, "unsetting user settings")
--- a/rest/route/host.go
+++ b/rest/route/host.go
@@ -413,6 +413,11 @@ func (ch *offboardUserHandler) Run(ctx context.Context) gimlet.Responder {
 			"context": "user offboarding",
 			"user":    ch.user,
 		}))
+		grip.Error(message.WrapError(user.ClearUserSettings(ch.user), message.Fields{
+			"message": "could not clear user settings",
+			"context": "user offboarding",
+			"user":    ch.user,
+		}))
 	}
 
 	if catcher.HasErrors() {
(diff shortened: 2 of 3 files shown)


model/user/db.go
```diff
@@ -414,6 +414,25 @@ func ClearLoginCache(user gimlet.User) error {
 	return nil
 }
 
+// ClearUserSettings clears one user's settings
+func ClearUserSettings(userId string) error {
+	update := bson.M{"$unset": bson.M{SettingsKey: 1}}
+
+	u, err := FindOneById(userId)
```
ablack12, line 421: I'd rather not run this query since we just use userId for the update anyway.

model/user/db.go
```diff
@@ -414,6 +414,25 @@ func ClearLoginCache(user gimlet.User) error {
 	return nil
 }
 
+// ClearUserSettings clears one user's settings
+func ClearUserSettings(userId string) error {
+	update := bson.M{"$unset": bson.M{SettingsKey: 1}}
```
Kimchelly, line 419: Rather than unset, can we `$set` it to an empty doc? I worry that we've set the BSON tag to not specify `omitempty`, so we should be able to assume for DB model consistency that the settings subdoc is either populated or empty (but not null/unset).

rest/route/host.go
```diff
@@ -413,6 +413,11 @@ func (ch *offboardUserHandler) Run(ctx context.Context) gimlet.Responder {
 			"context": "user offboarding",
 			"user":    ch.user,
 		}))
+		grip.Error(message.WrapError(user.ClearUserSettings(ch.user), message.Fields{
```
Kimchelly, line 416: Should this be added to the catcher rather than just logged?

model/user/db.go
malikchaya2: done

model/user/db.go
malikchaya2: done. but since it doesn't have `omitempty` set, the end result is still the same, the db will still hold the struct with the values unset.

rest/route/host.go
malikchaya2: done

rest/route/host.go
```diff
@@ -413,6 +413,13 @@ func (ch *offboardUserHandler) Run(ctx context.Context) gimlet.Responder {
 			"context": "user offboarding",
 			"user":    ch.user,
 		}))
+		err = user.ClearUserSettings(ch.user)
+		catcher.Wrapf(err, "clearing user settings for user '%s'", ch.user)
+		grip.Error(message.WrapError(err, message.Fields{
```
Kimchelly, line 418: I can't comment on the line because it's not modififed, but rather than logging here, can we just depend on the log on L427 (and change the error message about "not all unexpirable hosts/volumes terminated")?

model/user/db.go
Kimchelly: The distinction I was making is that `$set`'ing it to the empty doc will result in the doc looking like:
```
{
    "settings": {}
}
```
which is consistent with the way the data model is defined. In contrast, `$unset` will actually delete the field completely, so that field will no longer exist in the doc at all.

malikchaya2: However, I had a unit test set up for this function and in both cases (before and after the change) it went form a settings struct with fields, to 

` UserSettings{
		Timezone: "", Region: "", GithubUser: GithubUser{UID: 0, LastKnownAs: ""},
		SlackUsername: "", Notifications: NotificationPreferences{BuildBreak: "",
			BuildBreakID: "", PatchFinish: "", PatchFinishID: "", PatchFirstFailure: "",
			PatchFirstFailureID: "", SpawnHostExpiration: "", SpawnHostExpirationID: "",
			SpawnHostOutcome: "", SpawnHostOutcomeID: "", CommitQueue: "", CommitQueueID: ""},
		UseSpruceOptions: UseSpruceOptions{PatchPage: false, SpruceV1: false, HasUsedSpruceBefore: false, HasUsedMainlineCommitsBefore: false}`

rest/route/host.go
malikchaya2: done

model/user/db.go
Kimchelly: Yeah, the Golang struct will look like this because this is the default zero value for the struct, but the actual DB doc won't have the field set at all.

Kimchelly (approved):
LGTM mod changing the error message string on [L423](https://github.com/evergreen-ci/evergreen/pull/5820/commits/2fefef22c21d3964ff34dd2fd3791347f2c69da4#diff-d3ff34a4c5e4308b36ed70a61b69676caec5ca99594eff8c800491065be4f594R423).

rest/route/host.go
malikchaya2: Re 
`LGTM mod changing the error message string on [L423](https://github.com/evergreen-ci/evergreen/pull/5820/commits/2fefef22c21d3964ff34dd2fd3791347f2c69da4#diff-d3ff34a4c5e4308b36ed70a61b69676caec5ca99594eff8c800491065be4f594R423).`

I thought that was already done. Is this still outstanding?

Kimchelly: The "not all unexpirable hosts/volumes terminated" message is misleading because it could be due to terminating hosts/volumes or due to clearing user settings.

Kimchelly: GitHub doesn't let me link to a line that's not part of the diff: https://github.com/evergreen-ci/evergreen/blob/a4c509ab4daeddd20465bcebf57aebe6a355a7f4/rest/route/host.go#L421

malikchaya2: oh! of course! my bad, thank you.

PR #5822, opened 2022-07-26
EVG-17343 add stats for mainline commits query

khelif96:
[EVG-17343](https://jira.mongodb.org/browse/EVG-17343)

### Description 
Add some logic to help us better understand how much searching we are doing and why some queries are slow to return.


khelif96:
Closing the queries load now

PR #5825, opened 2022-07-29
EVG-16503: Hide execution tasks when configuring patch

minnakt:
[EVG-16503](https://jira.mongodb.org/browse/EVG-16503)

### Description 
On the legacy UI, execution tasks are not shown when configuring a patch. Conversely, on Spruce, execution tasks _are_ shown when configuring a patch. Users can actually schedule these execution tasks on Spruce which often leads to strange errors.

This PR makes it so that execution tasks are not shown when configuring a patch on Spruce. Only regular tasks and display tasks will be shown on the configure page.

### Testing 
- I renamed our test files to follow our standard convention and alphabetized the `results.json` file for readability. You can ignore these changes; the only relevant test is [project-variants-and-tasks.graphql](https://github.com/evergreen-ci/evergreen/compare/main...minnakt:EVG-16503?expand=1#diff-3dca5fe77ba25c2ac2a7cee4ea81528ca619613c1ed2d640e98ad808a9f00583) which I modified to test that execution tasks are not returned.
- On staging


malikchaya2 (approved):

graphql/util.go
```diff
@@ -200,12 +200,18 @@ func getPatchProjectVariantsAndTasksForUI(ctx context.Context, apiPatch *restMod
...
-			projTasks = append(projTasks, taskUnit.Name)
-		}
+		executionTasks := []string{}
 		for _, displayTask := range buildVariant.DisplayTasks {
 			projTasks = append(projTasks, displayTask.Name)
+			executionTasks = append(executionTasks, displayTask.ExecTasks...)
+		}
+		for _, taskUnit := range buildVariant.Tasks {
+			// Only add task if it is not an execution task.
+			if !utility.StringSliceContains(executionTasks, taskUnit.Name) {
```
ablack12, line 210: This is a bit expensive since we'll need to loop through all execution tasks worst case. Using a map[string]bool for executionTasks would be more performant.

graphql/util.go
```diff
@@ -200,12 +200,18 @@ func getPatchProjectVariantsAndTasksForUI(ctx context.Context, apiPatch *restMod
...
+				projTasks = append(projTasks, taskUnit.Name)
+			}
 		}
+		// sort tasks alphanumerically by display name
```
ablack12, line 214: nit: capitalize comment per style guides

ablack12 (changes requested):

ablack12 (approved):

PR #5827, opened 2022-07-29
EVG-16894: use pod definition cache

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16894

### Description 
The way that pod creation used to work is we would create a new definition in ECS each time we created a pod. This is not a scalable solution since we'll quickly hit the AWS API rate limit for creating definitions. To reduce the number of API calls, I created an intermediate step where there's another job queue that creates pod definitions that are then used by the pod creation job to actually start the pod from the template. The benefit of this is that we can reuse pod definitions across tasks that use the same configuration. For example, all tasks in a YAML that use the same container config will run in their own pods and each pod will use the same, shared pod definition. The general procedure to create a pod is now:
1. The pod is allocated to a task in preparation to actually create it in ECS (same as before).
2. The pod definition job is enqueued when there's a new container configuration to be created. It creates the pod definition in ECS and sticks the necessary information in the pod definition collection so that pods that depend on that definition can use it. If it fails to be created, it'll decommission all the pods that were depending on that definition.
3. The pod creation job polls repeatedly until the pod definition exists. Once it exists, it uses that pre-created definition to start the pod.
4. Most of the rest of the pod lifecycle afterwards is same as before until it's time to terminate the pod.
5. When cleaning up the pod during the pod termination job, we used to clean up the pod definition in ECS since only one pod would be using it. Now, the pod definition will _not_ be cleaned up since it could potentially be reused again in the future (there will be separate work to manage cleanup of long-unused pod definitions).

The actual changes to accomplish this:
* Create a hash function for the task container creation options. This is pretty much the same as [the Cocoa PR](https://github.com/evergreen-ci/cocoa/pull/78). We need this so that we can easily figure out if two tasks use identical container configuration parameters, meaning they will use the same pod definition. The hash goes into the pod's family name such that if two pods have the same family name, they will both use the exact same pod definition.
* Add new job and Amboy queue for pod definition creation. The pod definition job creates the pod definition in ECS and then stores it in the pod definition collection, including the ECS identifier for the definition and the family, which is used to determine which pods depend on that particular definition.
* Update the pod creation job to use the pre-created pod definition.
* Refactor a bunch of things as a result of the additional pod definition creation step.

### Testing 
* Added unit tests.
* Updated existing tests for new behavior + necessary refactoring.
* Verified in staging that the pod definition job work as expected and pods can still be created via the REST route.


units/pod_definition_creation.go
```diff
@@ -0,0 +1,206 @@
...
+		WaitUntil:   utility.ToTimeDurationPtr(10 * time.Second),
+	})
+
+	return j
+}
+
+func (j *podDefinitionCreationJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	defer func() {
```
hadjri, line 79: Can we combine these two deferred functions?

units/pod_definition_creation.go
```diff
@@ -0,0 +1,206 @@
...
+		return errors.Wrap(err, "finding intent pods to decommission")
+	}
+	catcher := grip.NewBasicCatcher()
+	var podIDs []string
+	for _, p := range podsToDecommission {
+		catcher.Wrapf(p.UpdateStatus(pod.StatusDecommissioned), "pod '%s'", p.ID)
+		podIDs = append(podIDs, p.ID)
+	}
+
+	grip.Info(message.Fields{
```
hadjri, line 198: Do we want to skip this logging if podIDs is empty?

units/pod_creation.go
```diff
@@ -108,13 +115,26 @@ func (j *podCreationJob) Run(ctx context.Context) {
...
+		// definition is not ready yet, retry again later.
+		podDef, err := j.waitForPodDefinition(j.pod.Family)
 		if err != nil {
-			j.AddError(errors.Wrap(err, "exporting pod creation options"))
+			j.AddRetryableError(errors.Wrap(err, "waiting for pod definition to be created"))
+			return
+		}
+
+		if err := podDef.UpdateLastAccessed(); err != nil {
+			j.AddRetryableError(errors.Wrapf(err, "updating last access time for pod definition '%s'", podDef.ID))
```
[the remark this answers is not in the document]
Kimchelly: It's possible for any DB operation to fail for various reasons that aren't the fault of the job, such as transient networking issues between the app and the DB.

units/pod_creation.go
```diff
@@ -191,6 +215,18 @@ func (j *podCreationJob) populateIfUnset(ctx context.Context) error {
 	return nil
 }
 
+func (j *podCreationJob) waitForPodDefinition(family string) (*definition.PodDefinition, error) {
```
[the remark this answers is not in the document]
Kimchelly: I named it this way because the job is intended to poll repeatedly on each job attempt until the pod definition is eventually created by the pod definition job. I don't have that strong of feelings about this though, so I renamed it.

units/pod_definition_creation.go
Kimchelly: Done.

units/pod_definition_creation.go
Kimchelly: Done.

units/pod_creation.go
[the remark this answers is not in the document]
hadjri: Ok thanks!

hadjri (approved):

PR #5829, opened 2022-08-01
EVG-17353 Remove getting splunk config from global

bynn:
[EVG-17353](https://jira.mongodb.org/browse/EVG-17353)

### Description 
remove getting splunk settings from global now that the splunk section exists in the admin database 

### Testing 
passes all unit tests
checked both staging and prod db


config_splunk.go
```diff
@@ -24,20 +24,7 @@ func (c *SplunkConfig) Get(env Environment) error {
...
-			}
-			c = &SplunkConfig{}
-			return nil
-		}
-		s := Settings{}
-		if err := globalConfig.Decode(&s); err != nil {
-			return errors.Wrap(err, "decoding global config")
-		}
-		c.SplunkConnectionInfo = s.Splunk.SplunkConnectionInfo
+		c = &SplunkConfig{}
```
malikchaya2, line 27: Do we need this considering that c is passed in? I'm assuming c is already defined since it's being accessed above?

bynn: i think this is to make sure that the config is empty if we do find an empty document like in here? 
https://github.com/evergreen-ci/evergreen/blob/264ae58b71ecdef7a5f1c9e9bc5020288033f8cd/config_slack.go#L28https://github.com/evergreen-ci/evergreen/blob/264ae58b71ecdef7a5f1c9e9bc5020288033f8cd/config_slack.go#L28

malikchaya2: ah that makes sense. Do we want to maybe flip the if just to be consistent with the function you linked?

bynn: i think @ablack12 preferred it written this way but i wasnt sure why

ablack12: I wanted it that way in the original PR so that the bulk of the code wouldn't have to be nested, since that isn't as readable (I did mention this in the original PR: https://github.com/evergreen-ci/evergreen/pull/5812#discussion_r929227347) . Now that most of the code is being deleted, this isn't relevant.

bynn: updated!

ablack12:
Removing my review since i was answering a question but have not reviewed.

malikchaya2 (approved):

PR #5830, opened 2022-08-01
EVG-17362 Fix project disappearing when defaulting to repo 

malikchaya2:
[EVG-17362](https://jira.mongodb.org/browse/EVG-17362)

### Description 
Because the function set the project to an empty struct when defaulting to repo, it ended up resetting the identifier, display name, and disabled field as well. This adds some casing to avoid that. 

### Testing 
Added to unit test and tested on staging.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1762,17 +1762,17 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			return false, errors.New("can't default project ref for a repo")
 		}
 	}
+	defaultToRepo := false
 	if p == nil {
+		defaultToRepo = true
 		p = &ProjectRef{} // use a blank project ref to default the section to repo
 	}
+
 	var err error
 	switch section {
 	case ProjectPageGeneralSection:
 		setUpdate := bson.M{
-			ProjectRefEnabledKey:                 p.Enabled,
 			ProjectRefBranchKey:                  p.Branch,
-			ProjectRefDisplayNameKey:             p.DisplayName,
-			ProjectRefIdentifierKey:              p.Identifier,
 			ProjectRefBatchTimeKey:               p.BatchTime,
 			ProjectRefRemotePathKey:              p.RemotePath,
 			projectRefSpawnHostScriptPathKey:     p.SpawnHostScriptPath,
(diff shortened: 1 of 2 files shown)


model/project_ref.go
```diff
@@ -1787,10 +1787,15 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			ProjectRefDisabledStatsCacheKey:      p.DisabledStatsCache,
 			ProjectRefFilesIgnoredFromCacheKey:   p.FilesIgnoredFromCache,
 		}
-		if !isRepo && !p.UseRepoSettings() {
+		if !isRepo && !p.UseRepoSettings() && !defaultToRepo {
```
ablack12, line 1790: If !p.UseRepoSettings then !defaultToRepo is true right, since we aren't using repo settings anyway?

model/project_ref.go
```diff
@@ -1787,10 +1787,15 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
 			setUpdate[ProjectRefOwnerKey] = p.Owner
 			setUpdate[ProjectRefRepoKey] = p.Repo
 		}
+		if !defaultToRepo {
+			setUpdate[ProjectRefEnabledKey] = p.Enabled
```
ablack12, line 1795: Enabled actually is something you should be able to default to repo so I don't think this should be included here.

ablack12 (changes requested):
Good find! Can we add a test for this case?

model/project_ref.go
malikchaya2: good, point, they won't have access to the button until they attach to a repo.

model/project_ref.go
malikchaya2: So if a repo is disabled, and a project is defaulted to the repo, then the project will become disabled. Is that what we want?

malikchaya2: (done)

model/project_ref.go
malikchaya2: (done)

model/project_ref.go
```diff
@@ -1762,17 +1762,18 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+		defaultToRepo = true
 		p = &ProjectRef{} // use a blank project ref to default the section to repo
 	}
+
 	var err error
 	switch section {
 	case ProjectPageGeneralSection:
 		setUpdate := bson.M{
 			ProjectRefEnabledKey:                 p.Enabled,
 			ProjectRefBranchKey:                  p.Branch,
```
sophstad, line 1776: Should branch default to repo? This seems like it could introduce conflicts (although I'm slightly fuzzy on the meaning of a branch when it comes to repos) cc @ablack12

malikchaya2: good point, especially considering that people probably specifically want different projects under one repo to correspond do different branches off of that repo.

ablack12: Discussed off the PR -- still makes sense to default since we don't know users use case

model/project_ref.go
```diff
@@ -1791,6 +1792,10 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			setUpdate[ProjectRefOwnerKey] = p.Owner
 			setUpdate[ProjectRefRepoKey] = p.Repo
 		}
+		if !defaultToRepo {
```
ablack12, line 1795: Can we add a comment here to explain why this section is different? (honestly I should've done above too)

ablack12 (approved):
LGTM!

PR #5831, opened 2022-08-01
EVG-16730 add debugging to understand long unschedule times

ablack12:
[EVG-16730](https://jira.mongodb.org/browse/EVG-16730 )

### Description 
UnschedulePatchTasks follows a pretty simple/reused path so it's not clear to me why this is reported to occasionally take 15 minutes for large patches. It looks like this only seems to happen for aborted tasks also. I've added some logging to try and pinpoint what part is causing an issue.


bynn (approved):
looks good!

malikchaya2 (approved):

PR #5836, opened 2022-08-02
EVG-17365: Default to main branch for new projects and repos

minnakt:
[EVG-17365](https://jira.mongodb.org/browse/EVG-17365)

### Description 
Branch name is a required field for projects and should not be empty. 

On legacy we default to `main` if no branch name is provided. This is done by using a [fallback value](https://github.com/evergreen-ci/evergreen/blob/5674df70257675405f0620298817f1e395a7b2db/public/static/js/projects.js#L489) on the UI. This only works because all of the project settings are on one page, so any saved change will also update the branch value.

On Spruce the project settings are broken up into different pages, so we can't use the same solution. Instead we can set the branch to `main` if none is provided when a project is created.

### Note
I set the branch to `main` at this part  of the code because we check for hidden projects right before ([link](https://github.com/evergreen-ci/evergreen/blob/main/model/project_ref.go#L431-L448)), and usually this step would be skipped if the branch was empty.

### Testing 
- Added test `TestAddEmptyBranch` in `project_ref_test.go`
- On staging, created a new project and verified the branch was automatically set to `main`


model/project_ref.go
```diff
@@ -447,6 +447,11 @@ func (p *ProjectRef) Add(creator *user.DBUser) error {
 		}
 	}
 
+	// Default to main for branch
+	if p.Branch == "" {
```
malikchaya2, line 451: considering this change, should we add "branch" to the create project modal so we give users a chance to indicate a branch before defaulting to main?
image

minnakt: That sounds reasonable to me, I opened [EVG-17391](https://jira.mongodb.org/browse/EVG-17391).

I wrote optional in the ticket but if we go this route, the branch could be a required field that prevents submission if it's empty. We could also pre-fill it to say main, which would kind of duplicate the work in this ticket. I'm not sure which approach is better 🤔

malikchaya2: Either approach seems reasonable to me. I'll defer to Annie and Sophie.

ablack12: I would say no personally. I'd prefer to just remove the default than add it to the modal.

ablack12: Oh I see now that we _had_ to add a default because spruce needs it. This now seems like a design question to me about what would be least confusing to users. I personally think it's fine to just always default to main though; it's simpler, unclutters the create project modal, and if users know they need a certain branch then they'll change it.

malikchaya2: In that case, instead of defaulting, should we maybe just make the text box yell at users so they can see why they can't save the page, similar to what we do here?
image

malikchaya2: or that. I typed my response before seeing the above.

minnakt:
Will mark ready for review again once I add the `buildlogger` default.

model/project_ref.go
```diff
@@ -447,6 +447,11 @@ func (p *ProjectRef) Add(creator *user.DBUser) error {
 		}
 	}
 
+	// Default to main if branch is empty.
+	if p.Branch == "" {
+		p.Branch = "main"
```
ablack12, line 452: Can we add a constant for this (maybe at the top of this file) just so we aren't hardcoding?

ablack12 (approved):
Discussed offline adding https://jira.mongodb.org/browse/EVG-17412 as a TODO comment.

malikchaya2 (approved):

model/project_ref.go
malikchaya2: :this:

PR #5838, opened 2022-08-02
EVG-17376: Validate owner and repo

sophstad:
[EVG-17376](https://jira.mongodb.org/browse/EVG-17376)

### Description 
- Verify GitHub owner appears in list of valid orgs when saving a project

### Testing 
- Add unit test, test in GraphQL playground

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1763,8 +1763,8 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 		}
 	}
 	if p == nil {
-                p = &ProjectRef{} // use a blank project ref to default the section to repo
-        }
+		p = &ProjectRef{} // use a blank project ref to default the section to repo
+	}
 	var err error
 	switch section {
 	case ProjectPageGeneralSection:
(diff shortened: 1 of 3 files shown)


malikchaya2 (approved):

model/project_ref.go
```diff
@@ -1788,6 +1788,14 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			ProjectRefFilesIgnoredFromCacheKey:   p.FilesIgnoredFromCache,
 		}
 		if !isRepo && !p.UseRepoSettings() {
+			// Don't validate owner if user is defaulting page to repo
+			if p.Owner != "" {
+				allowedOrgs := evergreen.GetEnvironment().Settings().GithubOrgs
+				if err := validateOwner(p.Owner, allowedOrgs); err != nil {
+					return false, errors.Wrap(err, "validating new owner/repo")
+				}
```
ablack12, line 1796: This should maybe just say validating new owner since we don't actually validate repo.

model/project_ref.go
```diff
@@ -1788,6 +1788,14 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			ProjectRefFilesIgnoredFromCacheKey:   p.FilesIgnoredFromCache,
 		}
 		if !isRepo && !p.UseRepoSettings() {
+			// Don't validate owner if user is defaulting page to repo
+			if p.Owner != "" {
+				allowedOrgs := evergreen.GetEnvironment().Settings().GithubOrgs
+				if err := validateOwner(p.Owner, allowedOrgs); err != nil {
```
ablack12, line 1794: FYI https://github.com/evergreen-ci/evergreen/blob/d0385e39aff6c24036dc2d7d203e48bf4361eb0c/model/project_ref.go#L2107 already exists. Possibly we could drop this "no owner/repo specified" logic (or only run that line if the project is enabled).

ablack12 (changes requested):

model/project_ref.go
sophstad: Yes, I updated that function to call `validateOwner` in order to split out the logic that checks if owner and repo are empty strings. This approach more cleanly handled the "default to repo" case where owner and repo are empty strings, which is independent of enabled status.

ablack12: Oops sorry my speed review was no good.

ablack12 (approved):

PR #5840, opened 2022-08-03
EVG-17390: Add Presto admin settings to the UI

julianedwards:
EVG-17390: https://jira.mongodb.org/browse/EVG-17390

### Description 
This fixes a bug that was causing the presto config section to be wiped in the DB every time the admin settings were updated. 
- Add the Presto settings to the UI and the API settings model.
- Remove the field `SessionProperties` since it was not being used / not needed and just made the UI more complicated.
- Remove[ Bynn's workaround](https://github.com/bynn/evergreen/blob/766a2deb8469c1f6b627991f2dad615dd0a2a16a/rest/model/admin.go#L2393-L2394) to skip the Presto config section when updating the admin events page.

### Testing 
I tested that the Presto settings both load correctly and are updated in the UI in staging.

### DIFF
--- a/config_presto.go
+++ b/config_presto.go
@@ -18,27 +18,25 @@ import (
 // PrestoConfig represents configuration information for the application level
 // Presto DB connection.
 type PrestoConfig struct {
-	BaseURI           string            `bson:"base_uri" json:"base_uri" yaml:"base_uri"`
-	Port              int               `bson:"port" json:"port" yaml:"port"`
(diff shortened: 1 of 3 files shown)


bynn (approved):
it looks good but if this is permanent, could we add a section to admin_test to test it? 
if it's temporary, i think it's fine

julianedwards:
It is temporary!

PR #5845, opened 2022-08-03
EVG-16490: Don't return 500s when getting test results fails.

johndaniels:
Currently we don't return the error, but we also send an error
response globally, causing this to 500 in the event of a Cedar
outage. This changes the logic so that we will return no test
results if Cedar is down but otherwise continue to process the
response.

[EVG-16490](https://jira.mongodb.org/browse/EVG-16490)

### Description 
This updates getTestResults to not send a 500 response if for some reason Cedar is down. It now silently logs the error and the request continues to process normally.

### Testing 
  I added testing to make sure that we do in fact return a nil list of results if Cedar is unavailable. Also, the function no longer has access to the response and request objects so it is impossible for it to send a 500 response code header.

### DIFF
--- a/service/task.go
+++ b/service/task.go
@@ -333,7 +333,7 @@ func (uis *UIServer) taskPage(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	testResults := uis.getTestResults(w, r, projCtx, &uiTask)
+	testResults := uis.getTestResults(projCtx, &uiTask)
 	if projCtx.Patch != nil {
 		var taskOnBaseCommit *task.Task
 		var testResultsOnBaseCommit []task.TestResult
@@ -891,9 +891,12 @@ func (uis *UIServer) testLog(w http.ResponseWriter, r *http.Request) {
 	uis.render.Stream(w, http.StatusOK, data, "base", "task_log.html")
 }
 
-func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, projCtx projectContext, uiTask *uiTaskData) []task.TestResult {
+func (uis *UIServer) getTestResults(projCtx projectContext, uiTask *uiTaskData) []task.TestResult {
 	if err := projCtx.Task.PopulateTestResults(); err != nil {
-		uis.LoggedError(w, r, http.StatusInternalServerError, err)
+		grip.Error(message.WrapError(err, message.Fields{
+			"task_id": projCtx.Task.Id,
+			"message": "fetching test results for task",
+		}))
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


service/task.go
```diff
@@ -911,7 +914,11 @@ func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, proj
 				et, err = task.FindOneId(t)
 			}
 			if err != nil {
-				uis.LoggedError(w, r, http.StatusInternalServerError, err)
+				grip.Error(message.Fields{
+					"message": "fetching execution task",
```
julianedwards, line 918: `fetching test results for execution task`

service/task.go
```diff
@@ -911,7 +914,11 @@ func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, proj
...
+					"task":    t,
```
julianedwards, line 919: `task_id`

service/task.go
```diff
@@ -911,7 +914,11 @@ func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, proj
...
+					"parent":  projCtx.Task.Id,
```
julianedwards, line 920: `parent_task_id`

julianedwards (changes requested):

service/task_test.go
```diff
@@ -0,0 +1,38 @@
...
+	uis := UIServer{hostCache: make(map[string]hostCacheItem)}
+	projectContext := projectContext{
+		Context: model.Context{
+			Task: &newTask,
+		},
+	}
+	uiTask := uiTaskData{}
+	results := uis.getTestResults(projectContext, &uiTask)
+	var expected []task.TestResult = nil
+	assert.Equal(t, expected, results)
```
ablack12, line 37: assert.Nil

service/task.go
```diff
@@ -891,9 +891,12 @@ func (uis *UIServer) testLog(w http.ResponseWriter, r *http.Request) {
...
 
-func (uis *UIServer) getTestResults(w http.ResponseWriter, r *http.Request, projCtx projectContext, uiTask *uiTaskData) []task.TestResult {
+func (uis *UIServer) getTestResults(projCtx projectContext, uiTask *uiTaskData) []task.TestResult {
 	if err := projCtx.Task.PopulateTestResults(); err != nil {
-		uis.LoggedError(w, r, http.StatusInternalServerError, err)
+		grip.Error(message.WrapError(err, message.Fields{
+			"task_id": projCtx.Task.Id,
+			"message": "fetching test results for task",
+		}))
 		return nil
```
ablack12, line 900: Should we initialize uis.TestResults above? (Is there concern around nil pointer errors if we aren't sending an error? Didn't read back where this is used though.

ablack12 (approved):

service/task.go
johndaniels: The only place it gets used in Go code is as part of append inside that function. After that I _think_ it goes to JS code, which could have an issue with null maybe? 

I don't see any JS errors in the console but all I'm doing is loading the task page and that seems to work.

johndaniels: Happy to change it though if it seems like a good idea.

service/task.go
johndaniels: fixed

service/task.go
johndaniels: fixed

service/task.go
johndaniels: fixed

julianedwards (approved):

PR #5847, opened 2022-08-04
EVG-16908: create model for container secrets

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16908

### Description 
This updates the model for container secrets so that they're suitable for Secrets Manager. They don't actually store the secrets in Secrets Manager until [EVG-16913](https://jira.mongodb.org/browse/EVG-16913).

The gist of how it's going to work is that, through the REST model, the user can either send Evergreen a secret to store or ask Evergreen to give them their secret in plaintext. If they ask us to store a repo cred, this is a special case because it must be stored in Secrets Manager as a JSON-encoded string. The JSON-encoded credentials will then be stored (in [EVG-16913](https://jira.mongodb.org/browse/EVG-16913)) in Secrets Manager. After it's stored, if Evergreen ever needs to retrieve the secret value in plaintext to give back to the user, it queries Secrets Manager to get the JSON-encoded string, which gets decoded into a more nicely formatted REST response containing the username/password.

### Testing 
Updated existing unit tests and added some more validation unit tests.

### DIFF
--- a/graphql/mutation_resolver.go
+++ b/graphql/mutation_resolver.go
@@ -369,10 +369,13 @@ func (r *mutationResolver) AttachProjectToRepo(ctx context.Context, projectID st
 }
 
 func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.APIProjectRef) (*restModel.APIProjectRef, error) {
-	dbProjectRef := project.ToService()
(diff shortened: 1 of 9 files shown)


hadjri (approved):
Looks good, I believe the linked ticket in the description is wrong though.

PR #5849, opened 2022-08-05
EVG-17395 Fix verify owner/repo 

malikchaya2:
[EVG-17395](https://jira.mongodb.org/browse/EVG-17395)

### Description 
Duplicate project didn't work because verifyOwner would make sure that the owner and repo weren't nil, which wasn't true if a project was defaulted to the repo. I updated the function to also check the repo. 

### Testing 
Tested on staging


ablack12 (approved):
LGTM but can you add a test? (Generally I think anything that introduces a bug that can be tested easily should be tested, since it verifies we don't accidentally break this behavior and re-introduce the bug)

sophstad (approved):
thank you! 🙇‍♀️

PR #5850, opened 2022-08-05
EVG-17425: Use new logkeeper service and build lobster

SupaJoon:
[EVG-17425](https://jira.mongodb.org/browse/EVG-17425)

### Description 
Builds this Lobster commit  https://github.com/evergreen-ci/lobster/pull/130.
Running `make update-lobster` in the evergreen repo will pull the latest lobster main branch and build it.
These two links render logkeeper logs from each service once you run `make local-evergreen`:
http://localhost:9090/lobster/build/321662d0fb6c60eec7166c1591dca54e/test/62ed1b1fbe07c435a39a1bd6#bookmarks=0%2C425
http://localhost:9090/lobster/build/8be6cac51d7e8bdf7b5c120c91a1bf6e/test/62ed1707c5e2faef36be7bc7#bookmarks=0%2C2613


model/task/task.go
```diff
@@ -397,13 +397,15 @@ func (tr TestResult) GetDisplayTestName() string {
 // generation for other log viewers.
 func (tr TestResult) GetLogURL(viewer evergreen.LogViewer) string {
 	root := evergreen.GetEnvironment().Settings().ApiUrl
-	deprecatedLobsterURL := "https://logkeeper.mongodb.org"
+	deprecatedLobsterURLs := []string{"https://logkeeper.mongodb.org", "https://logkeeper2.build.10gen.cc"}
```
SupaJoon, line 400: make sure that spruce only renders lobster urls on the evergreen domain

ybrill:
Does [this](https://github.com/evergreen-ci/lobster/blob/1f0ced043ff5c5fcb91ca71a3cc741b62f229735/src/api/logkeeper.js#L5-L8) mean for every request lobster makes from logkeeper that we'll make the request twice? If so we could totally add a route to logkeeper to just answer the question of whether the data exists.

SupaJoon:
> Does [this](https://github.com/evergreen-ci/lobster/blob/1f0ced043ff5c5fcb91ca71a3cc741b62f229735/src/api/logkeeper.js#L5-L8) mean for every request lobster makes from logkeeper that we'll make the request twice? If so we could totally add a route to logkeeper to just answer the question of whether the data exists.

The logs will be requested once if it's an old logkeepr log and 3 times if it's a new logkeeper log. The route you mention can make things better. If this is possible I can edit the Lobster code to use this endpoint for when it's available.

PR #5851, opened 2022-08-05
EVG-17025: Create new agent rest route to fetch the Data-Pipes config

julianedwards:
EVG-17025: https://jira.mongodb.org/browse/EVG-17025

### Description 
This is the second of three PRs for this ticket (in the spirit of smaller PRs). We are going to start sending perf results to both Cedar and the TIPS' new [Data-Pipes](https://github.com/10gen/data-pipes) service. In order to do that we will have to pipe through to the agent information and credentials for the service.

- Add a new REST route `/rest/v2/agent/data_pipes_config` to fetch the Data-Pipes information in the agent.
- Fix up the Cedar config agent route.

### Testing 
New unit tests.


testutil/config.go
```diff
@@ -189,7 +195,14 @@ func MockConfig() *evergreen.Settings {
 				},
 			},
 		},
-		Credentials:        map[string]string{"k1": "v1"},
+		Credentials: map[string]string{"k1": "v1"},
+		DataPipes: evergreen.DataPipesConfig{
+			Host:         "https://url.com",
```
johndaniels, line 200: I generally prefer to use "*.example.com" for fake URLs because that is reserved. Alternatively a mongodb domain would work as well. (url.com is actually owned by someone)

rest/route/agent.go
```diff
@@ -12,31 +12,58 @@ import (
...
+}
+
+func (*agentDataPipesConfig) Parse(_ context.Context, _ *http.Request) error { return nil }
+
+func (h *agentDataPipesConfig) Run(ctx context.Context) gimlet.Responder {
+	return gimlet.NewJSONResponse(apimodels.DataPipesConfig{
+		Host:         h.config.Host,
+		Region:       h.config.Region,
+		AWSAccessKey: h.config.AWSAccessKey,
+		AWSSecretKey: h.config.AWSSecretKey,
```
johndaniels, line 66: This is probably beyond this change, but it'd be nice if there was some way we could use the machine IAM roles for this rather than passing around keys.

johndaniels (changes requested):

johndaniels:
Just the one change for "url.com"

rest/route/agent.go
julianedwards: We discussed this a bunch, it isn't really feasible since we have static hosts outside of AWS.

testutil/config.go
julianedwards: Yeah these aren't actually calling anything, I just changed it to `url`.

rest/route/agent.go
johndaniels: Cool, thanks for the info!

testutil/config.go
johndaniels: That works. Just don't want to set up a situation where it would be easy to accidentally call it and leak data.

PR #5853, opened 2022-08-08
EVG-17367 Remove distro dependencies of agent's internal task config

hadjri:
[EVG-17367](https://jira.mongodb.org/browse/EVG-17367)

### Description 
Removed distro dependence within the agent's internal task config which was previously required in commands such as `git push`, `git get project`, as well as setting the task directory and populating expansions.
### Testing 
Running containerized tasks in staging was not possible until this change. Example succeeded containerized task: https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_test_util_patch_f2059199a605e918e1b0c89c570acf53f360e167_62f0959e97b1d36ece0869c9_22_08_08_04_48_44/logs?execution=0&logtype=task&sortBy=STATUS&sortDir=ASC


agent/command/git.go
```diff
@@ -384,14 +384,15 @@ func (c *gitFetchProject) buildModuleCloneCommand(conf *internal.TaskConfig, opt
 }
 
 func (c *gitFetchProject) opts(projectMethod, projectToken string, conf *internal.TaskConfig) (cloneOpts, error) {
+	shallowCloneEnabled := conf.Distro != nil && !conf.Distro.DisableShallowClone
```
Kimchelly, line 387: I don't have any context on why some distros would want to have shallow clone enabled/disabled. It seems like we should respect the distro-level shallow clone setting but for container tasks, we should always allow shallow cloning and assume that if users ask for a shallow clone, they'll supply an image that supports shallow cloning.

agent/command/git.go
```diff
@@ -421,7 +422,11 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
 		return errors.Wrap(err, "error expanding github parameters")
 	}
 
-	projectMethod, projectToken, err := getProjectMethodAndToken(c.Token, conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion), conf.Distro.CloneMethod)
+	cloneMethod := distro.CloneMethodOAuth
```
Kimchelly, line 425: Since it's no longer just a property of distros, can we move `distro.CloneMethodOAuth` and `distro.CloneMethodSSH` into the global constants in the top-level `evergreen` package?

agent/command/git_push.go
```diff
@@ -81,8 +82,12 @@ func (c *gitPush) Execute(ctx context.Context, comm client.Communicator, logger
 	}
 
 	// get commit information
+	cloneMethod := distro.CloneMethodOAuth
+	if conf.Distro != nil {
+		cloneMethod = conf.Distro.CloneMethod
+	}
```
Kimchelly, line 88: Can we put L85-88 in a method on `internal.TaskConfig` so that we always consistently return the proper OAuth default across git commands?

Kimchelly: Can we also add a test to verify that the git commands no longer depend on the distro clone method?

Kimchelly (changes requested):

agent/command/git_push.go
hadjri: Sure, moved to a method and updated existing unit tests.

agent/command/git.go
hadjri: Done

agent/command/git.go
hadjri: Gotcha, updated the bool so that `shallowCloneEnabled` will be true if Distro is nil

agent/command/git.go
```diff
@@ -384,14 +384,15 @@ func (c *gitFetchProject) buildModuleCloneCommand(conf *internal.TaskConfig, opt
 }
 
 func (c *gitFetchProject) opts(projectMethod, projectToken string, conf *internal.TaskConfig) (cloneOpts, error) {
+	shallowCloneEnabled := conf.Distro == nil || (conf.Distro != nil && !conf.Distro.DisableShallowClone)
```
Kimchelly, line 387: nit: since logical operators short-circuit in Go, it's not necessary to check if `conf.Distro != nil`. If `conf.Distro == nil`, it will skip evaluating the second clause in the boolean anyways (so it won't panic).

agent/internal/task_config.go
```diff
@@ -118,6 +120,14 @@ func (c *TaskConfig) GetWorkingDirectory(dir string) (string, error) {
 	return dir, nil
 }
 
+func (c *TaskConfig) GetCloneMethod() string {
+	cloneMethod := evergreen.CloneMethodOAuth
+	if c.Distro != nil {
+		cloneMethod = c.Distro.CloneMethod
+	}
+	return cloneMethod
```
Kimchelly, line 128: nit: this can be simplified down to:
```go
if c.Distro != nil {
    return c.Distro.CloneMethod
}

return evergreen.CloneMethodOAuth
```

model/distro/distro.go
```diff
@@ -574,7 +565,7 @@ func ValidateArch(arch string) error {
 // ValidateCloneMethod checks that the clone mechanism is one of the supported
 // methods.
 func ValidateCloneMethod(method string) error {
```
Kimchelly, line 567: Can this also be moved into the top-level package with the clone methods?

agent/internal/task_config.go
```diff
@@ -87,7 +87,9 @@ func NewTaskConfig(d *apimodels.DistroView, p *model.Project, t *task.Task, r *m
 		Task:         t,
 		BuildVariant: bv,
 		Expansions:   &e,
-		WorkDir:      d.WorkDir,
+	}
+	if d != nil {
+		taskConfig.WorkDir = d.WorkDir
```
Kimchelly, line 92: I think this could just be set to the agent working directory. In doing so, we could also remove `(DistroView).WorkDir` field entirely because it's the same as the agent working directory in all cases anyways.

agent/internal/task_config.go
hadjri: Fixed

model/distro/distro.go
hadjri: Sure, done

agent/internal/task_config.go
hadjri: Sounds good

agent/command/git.go
hadjri: Oh thanks!

Kimchelly (approved):
We can also delete the `(DistroView).WorkDir` since it's not used in any place other than tests. We can put that in a separate ticket though since the agents will have to roll over first anyways.

hadjri:
> WorkDir

Created: https://jira.mongodb.org/browse/EVG-17505

PR #5855, opened 2022-08-08
EVG-17025: Send perf report to Data-Pipes service in perf.send

julianedwards:
EVG-17025: https://jira.mongodb.org/browse/EVG-17025

### Description 
This is the last of three PRs for this ticket (in the spirit of smaller PRs). We are going to start sending perf results to both Cedar and the TIPS' new [Data-Pipes](https://github.com/10gen/data-pipes) service. In order to do that we will have to pipe through to the agent information and credentials for the service.

- Update poplar.
- Add new function to the agent client interface `GetDataPipesConfig` that calls the agent rest route to fetch the Data Pipes admin config from the app servers.
- Send perf data to Data-Pipes as well as Cedar via poplar in the perf.send command code.
- Update the agent version.

### Testing 
New unit tests.


agent/command/perf_send.go
```diff
@@ -53,30 +54,41 @@ func (c *perfSend) ParseParams(params map[string]interface{}) error {
...
+		AWSAccessKey:        dataPipes.AWSAccessKey,
+		AWSSecretKey:        dataPipes.AWSSecretKey,
+		AWSToken:            dataPipes.AWSToken,
+		DataPipesHTTPClient: httpClient,
 	}
-	return errors.Wrap(rpc.UploadReport(ctx, opts), "failed to upload report to cedar")
+	return errors.Wrap(rpc.UploadReport(ctx, opts), "uploading report to Cedar")
 }
 
 func (c *perfSend) addEvgData(report *poplar.Report, conf *internal.TaskConfig) {
```
aggrand, line 94: Hey just stalking the PR here, but by the end of the poplar-side PR we also added a `Requester` field to the report. If that could be added to the evg data and also if we could set the `Mainline` flag when the requester == RepoTrackerVersionRequester, that would be great!

Super cool if that should require a separate ticket though, just lmk.

julianedwards: Good catch! Done.

testutil/config.go
```diff
@@ -169,9 +169,15 @@ func MockConfig() *evergreen.Settings {
...
 			BackgroundReauthMinutes: 60,
 		},
-		AWSInstanceRole:   "role",
-		Banner:            "banner",
-		BannerTheme:       "important",
+		AWSInstanceRole: "role",
+		Banner:          "banner",
+		BannerTheme:     "important",
+		Cedar: evergreen.CedarConfig{
+			BaseURL: "url.com",
```
johndaniels: And here to non-routable or example.com

apimodels/cedar_models.go
```diff
@@ -5,6 +5,7 @@ type CedarConfig struct {
 	RPCPort  string `json:"rpc_port"`
 	Username string `json:"username"`
 	APIKey   string `json:"api_key,omitempty"`
+	Insecure bool   `json:"tls"`
```
johndaniels, line 8: This json name likely to be confusing, because tls: true sounds like the opposite of insecure: true

johndaniels: This is "insecure" everywhere else.

agent/internal/client/base_client.go
```diff
@@ -132,7 +132,11 @@ func (c *baseCommunicator) createCedarGRPCConn(ctx context.Context) error {
 			RPCPort:     cc.RPCPort,
 			Username:    cc.Username,
 			APIKey:      cc.APIKey,
-			Retries:     10,
+			// Insecure is should always be set to false except
```
johndaniels, line 135: Fix the comment here.

johndaniels (changes requested):

johndaniels:
I'm fine with you merging after those fixes

apimodels/cedar_models.go
julianedwards: Nice catch!

PR #5859, opened 2022-08-10
EVG-16730 improve version activation

ablack12:

### Description 
Used some logging to understand where things are slow. Realized that we shouldn't be using SetBuildActivation because it repeats updating tasks and updating the version/patch status. 

Also changed logging here to use InsertMany instead of an InsertOne loop.

### Testing 
Ensure existing tests pass.


model/event/task_event.go
```diff
@@ -160,10 +161,17 @@ func LogTaskActivated(taskId string, execution int, userId string) {
...
+	return getTaskEvent(taskId, TaskActivated, TaskEventData{Execution: execution, UserId: userId})
+}
+
 func LogTaskDeactivated(taskId string, execution int, userId string) {
 	logTaskEvent(taskId, TaskDeactivated, TaskEventData{Execution: execution, UserId: userId})
 }
 
+func GetTaskDeactivatedEvent(taskId string, execution int, userId string) EventLogEntry {
+	return getTaskEvent(taskId, TaskDeactivated, TaskEventData{Execution: execution, UserId: userId})
+}
```
Kimchelly, line 174: nit: newline between L174-175.

Kimchelly (commented):
I think the logic looks good, going to hold on LGTM for the tests to be updated.

model/task/task.go
```diff
@@ -1814,9 +1832,16 @@ func DeactivateDependencies(tasks []string, caller string) error {
 	if err != nil {
 		return errors.Wrap(err, "deactivating dependencies")
 	}
+
+	logs := []event.EventLogEntry{}
 	for _, t := range tasksToUpdate {
-		event.LogTaskDeactivated(t.Id, t.Execution, caller)
+		logs = append(logs, event.GetTaskDeactivatedEvent(t.Id, t.Execution, caller))
 	}
+	grip.Error(message.WrapError(event.LogManyEvents(logs), message.Fields{
```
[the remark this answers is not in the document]
ablack12: From googling it does seems more performant which would make it preferable for large operations, but probably not enough to rewrite everything without cause.

model/task/task.go
```diff
@@ -1699,9 +1705,15 @@ func ActivateDeactivatedDependencies(tasks []string, caller string) error {
 		return errors.Wrap(err, "updating activation for dependencies")
 	}
 
+	logs := []event.EventLogEntry{}
 	for _, t := range tasksToActivate {
-		event.LogTaskActivated(t.Id, t.Execution, caller)
+		logs = append(logs, event.GetTaskActivatedEvent(t.Id, t.Execution, caller))
 	}
+	grip.Error(message.WrapError(event.LogManyEvents(logs), message.Fields{
+		"message":  "problem logging task activated events",
```
malikchaya2, line 1713: nit: should this one and the one on 1791 be just "logging task..."?

malikchaya2: (and 1842)

malikchaya2 (approved):

model/task/task.go
Kimchelly: I think it's actually okay like this. The style guide says that the `"problem"` rule would only apply to wrapped error messages (i.e. strings used for `errors.Wrap` or `errors.Wrapf`) since they tend to wrap several error messages up the call stack. In this case, this just logs as a single error message, so it's not an issue.

model/lifecycle.go
```diff
@@ -92,19 +92,21 @@ func SetVersionActivation(versionId string, active bool, caller string) error {
 		}
 	}
 	if len(tasksToModify) > 0 {
```
bynn, line 94: did we want to switch this?

bynn (changes requested):

model/lifecycle.go
ablack12: yes 🤦‍♀️

bynn (approved):
:shipit:

PR #5861, opened 2022-08-11
EVG-16825 Rename self in functions attached to structs

ZackarySantana:
[EVG-16825](https://jira.mongodb.org/browse/EVG-16825)

### Description 
Golang conventions specify that functions attached to structs not label with 'this'/'self'. This is a refactor for functions that currently use self to not.

### Testing 
Unit tests

### DIFF
--- a/agent/util/subtree_windows.go
+++ b/agent/util/subtree_windows.go
@@ -229,20 +229,20 @@ func NewJob(name string) (*Job, error) {
 	return &Job{handle: hJob}, nil
 }
 
-func (self *Job) AssignProcess(pid uint) error {
+func (j *Job) AssignProcess(pid uint) error {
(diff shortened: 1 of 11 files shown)


Kimchelly (approved):
This change seems fine. One note (which I don't feel that strongly about) is that usually, the method receiver name is a shorthand few-characters-long variable name rather than just the entire type spelled out as a variable. I'm alright with this as it is though.

ZackarySantana:
I renamed with one letter / shorthands, does that look better?

ZackarySantana:
Screen Shot 2022-08-11 at 1 14 30 PM

I didn't do it for any js file (or that one xml file) because js has a concept/convention of using self

Kimchelly (approved):
Yes, the change looks good. Thanks for the tech debt cleanup!

PR #5866, opened 2022-08-16
EVG-17387: Fix project dropdown naming

sophstad:
[EVG-17387](https://jira.mongodb.org/browse/EVG-17387)

### Description 
Attached projects in the project dropdown were displaying their repo's display name when they did not have one set (i.e. empty string in the db). This is because the function to fetch the projects (`FindProjectRefsByIds`) was merging the projects with their repos, resulting in the displayName field being overwritten. To fix:
- Rename `FindProjectRefsByIds` to `FindMergedProjectRefsByIds`
- Added a new `FindProjectRefsByIds` that does not merge the projects and repos
- Replace all existing uses of `FindProjectRefsByIds` with `FindMergedProjectRefsByIds` to maintain previous functionality

### Testing 
Front end now renders attached project names correctly. If a project lacks a display name, its identifier is used.

### DIFF
--- a/model/pod/dispatcher/dispatcher.go
+++ b/model/pod/dispatcher/dispatcher.go
@@ -251,7 +251,7 @@ func (pd *PodDispatcher) checkTaskIsDispatchable(ctx context.Context, env evergr
 		return false, nil
 	}
 
-	refs, err := model.FindProjectRefsByIds(t.Project)
+	refs, err := model.FindMergedProjectRefsByIds(t.Project)
 	if err != nil {
 		return false, errors.Wrapf(err, "finding project ref '%s' for task '%s'", t.Project, t.Id)
 	}
--- a/model/pod/dispatcher/dispatcher.go
+++ b/model/pod/dispatcher/dispatcher.go
@@ -251,7 +251,7 @@ func (pd *PodDispatcher) checkTaskIsDispatchable(ctx context.Context, env evergr
 		return false, nil
 	}
 
-	refs, err := model.FindMergedProjectRefsByIds(t.Project)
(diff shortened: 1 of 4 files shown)


model/project_ref.go
```diff
@@ -1266,6 +1277,17 @@ func FindProjectRefsQ(filter bson.M) ([]ProjectRef, error) {
 	return addLoggerAndRepoSettingsToProjects(projectRefs)
 }
 
+func FindProjectRefsQ(filter bson.M) ([]ProjectRef, error) {
```
ablack12, line 1280: Could  this could take in a boolean `merged` to determine whether to call addLoggerANdRepoSettings as opposed to having both functions do similar things? We could also make this private to make it more obvious this is just a helper function.

ablack12 (changes requested):

model/pod/dispatcher/dispatcher.go
```diff
@@ -251,7 +251,7 @@ func (pd *PodDispatcher) checkTaskIsDispatchable(ctx context.Context, env evergr
 		return false, nil
 	}
 
-	refs, err := model.FindProjectRefsByIds(t.Project)
+	refs, err := model.FindMergedProjectRefsByIds(t.Project)
```
ablack12, line 254: Side thought is that this should've just called the regular FindMergedProjectRef since it's just for one project? Could ticket that though if you don't want to change it here

ablack12 (approved):

model/pod/dispatcher/dispatcher.go
sophstad: Let's get 'er done 🤠

PR #5868, opened 2022-08-16
EVG-16837 Remove generate tasks retry logic

hadjri:
[EVG-16837](https://jira.mongodb.org/browse/EVG-16837)

### Description 
Since the merging of [this PR](https://github.com/evergreen-ci/evergreen/pull/5507) the generate tasks job uses scopes and has been moved to a single queue within the queue group. Some of the retry logic within the agent was structured to retry on error in case of a race between jobs. Now, if an error is encountered in generate.tasks the command will fail without retry.  

I did not remove the entire retry block, and we are still checking for the successful completion of a spawned generate tasks job for the given task. 
### Testing 
Tested generate.tasks in staging successfully. Added new unit test and mock property to test a failure response from polling generate.tasks.

### DIFF
--- a/agent/command/generate.go
+++ b/agent/command/generate.go
@@ -44,7 +44,7 @@ func (c *generateTask) ParseParams(params map[string]interface{}) error {
 func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
 	var err error
 	if err = util.ExpandValues(c, conf.Expansions); err != nil {
-		return errors.Wrap(err, "error expanding params")
+		return errors.Wrap(err, "expanding params")
 	}
 
 	include := utility.NewGitIgnoreFileMatcher(conf.WorkDir, c.Files...)
@@ -53,15 +53,15 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
 		Include:    include,
 	}
 	if c.Files, err = b.Build(); err != nil {
-		return errors.Wrap(err, "problem building wildcard paths")
+		return errors.Wrap(err, "building wildcard paths")
 	}
 
 	if len(c.Files) == 0 {
 		if c.Optional {
-			logger.Task().Info("No files found and optional is true, skipping generate.tasks")
+			logger.Task().Info("no files found and optional is true, skipping generate.tasks")
 			return nil
 		}
-		return errors.New("No files found for generate.tasks")
+		return errors.New("no files found for generate.tasks")
 	}
 
 	catcher := grip.NewBasicCatcher()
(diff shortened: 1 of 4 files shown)


agent/command/generate.go
```diff
@@ -87,14 +87,14 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
...
-		return errors.Wrap(err, "problem parsing JSON")
+		return errors.Wrap(err, "parsing JSON")
 	}
 	if err = comm.GenerateTasks(ctx, td, post); err != nil {
 		if strings.Contains(err.Error(), evergreen.TasksAlreadyGeneratedError) {
 			logger.Task().Info("Tasks have already been generated, nooping.")
 			return nil
 		}
-		return errors.Wrap(err, "Problem posting task data")
+		return errors.Wrap(err, "posting task data")
```
malikchaya2, line 97: nice work updating all the errors, thank you :)

malikchaya2 (approved):

agent/command/generate.go
```diff
@@ -120,11 +120,10 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
 				return false, generateErr
 			}
 			if generateErr != nil {
-				// if the error isn't related to saving the generated task, log it but still retry in case of race condition
 				if !strings.Contains(generateErr.Error(), evergreen.SaveGenerateTasksError) {
-					logger.Task().Infof("Problem polling for generate tasks job, retrying (%s)", generateErr.Error())
+					logger.Task().Infof("Problem polling for generate tasks job", generateErr.Error())
```
ablack12, line 124: Now that we won't potentially poll forever when something is misconfigured we shouldn't need this special casing to make sure we tell the user what happened (or this SaveGenerateTasksError at all), it should log it at the end (you could test this by making a generate task with an invalid config and see if it logs)

agent/command/generate.go
```diff
@@ -108,7 +108,7 @@ func (c *generateTask) Execute(ctx context.Context, comm client.Communicator, lo
 		func() (bool, error) {
 			generateStatus, err := comm.GenerateTasksPoll(ctx, td)
 			if err != nil {
-				return true, err
+				return false, err
```
ablack12, line 111: This won't let me comment below but it looks like we should remove the ShouldExit special casing since we're always now skipping the retry, and should remove from the struct:
https://github.com/evergreen-ci/evergreen/blob/bdc2f5e9a42213d2202aa8f4bbf38e4a9b1fe6cf/rest/route/task_generate.go#L99
(I think for backwards compatibility we probably want to do that in a different PR though)

hadjri: Sounds good! Made a new-eng ticket: https://jira.mongodb.org/browse/EVG-17561

agent/command/generate.go
hadjri: Sounds good, did this and tested in staging to confirm this log always hits on error.

ablack12: Well actually what i meant was -- do we need to have this log here or do we already log the error returned from the loop somewhere? (I would think we hit https://github.com/evergreen-ci/evergreen/blob/68c43797d61a0c21c338dfc7f08352b42a5ddff0/agent/command.go#L131 already so we're duplicating logging then)

bynn (approved):

ablack12 (commented):
Added a comment to a thread but that didn't mark me as having reviewed hence this dummy comment 🤦‍♀️

agent/command/generate.go
hadjri: Ah yes - that line is duplicative. Removed.

ablack12 (approved):

PR #5869, opened 2022-08-16
EVG-17518: Don't clear owner or repo when defaulting project

sophstad:
[EVG-17518](https://jira.mongodb.org/browse/EVG-17518)

### Description 
- Because some queries depend on a project having an owner/repo, we shouldn't clear these fields when "Default to Repo" is clicked on the General settings page
- Allow owner/repo to be empty only if project is disabled

### Testing 
- Updated unit tests, tested on UI


ablack12 (approved):
Good find! 🚀

PR #5870, opened 2022-08-16
EVG-16730 improve updates on multiple builds

ablack12:

### Description 
Well i forgot to merge main so this didn't actually impact much the function I was trying to improve 🦙 😞  but should still be a worthwhile improvement for the others so keeping it. Should take advantage of these functions being written for multiple builds by compiling updates and performing them at the end.

Also removes logging because this seems much improved per splunk data.

### Testing 
Ensure existing tests pass.


model/lifecycle.go
```diff
@@ -1756,19 +1760,22 @@ func addNewTasks(ctx context.Context, activationInfo specificActivationInfo, v *
...
 		}
 	}
-	if activationInfo.hasActivationTasks() {
-		grip.Error(message.WrapError(v.UpdateBuildVariants(), message.Fields{
-			"message": "unable to add batchtime tasks",
-			"version": v.Id,
-		}))
+	if len(buildIdsToActivate) > 0 {
+		if err := build.UpdateActivation(buildIdsToActivate, true, evergreen.DefaultTaskActivator); err != nil {
+			return nil, err
```
malikchaya2, line 1768: Why do we want to switch from logging the error and continuing to returning an error?

ablack12: Are you referring to the log I removed "unable to add batchtime tasks" ? (I'm not sure why github isn't showing your comment in context on the file tab, alas) 

That was when the update _only_ handled adding batchtime info, but now with the refactor it also updates activation, which was an error before and should still be an error.

malikchaya2: ah that makes perfect sense, thank you!

malikchaya2 (approved):

PR #5874, opened 2022-08-17
EVG-16981 Update query resolver to account for skipped execution levels

ZackarySantana:
[EVG-16981](https://jira.mongodb.org/browse/EVG-16981)

### Description 
Exposes two new fields for api tasks that hold the information if the task information was pulled from a previous task, as well as the execution of that task that the information was pulled from. Adds a couple db methods to pull tasks from previous executions, and updates the query resolver to use them. The query resolver also sets the fields if the task was taken from a previous one. This is so spruce can include these fields in its graphql request and then display information like (cached from x execution).

### Testing 
Deployed to spruce and tested on patch https://spruce-staging.corp.mongodb.com/version/62fbf4bf9dbe32458bd4dff7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

This patch has the following history / state:
display_task
- always_fail
- validate_commit_message (always succeeds)

It was restarted like:
- Only failed
- Restart all
- Restart all

Leaving the complete state as such:
display_task execution on the fourth (execution = 3)
- always_fail
  - execution 0: failed
  - execution 1: failed
  - execution 2: failed
  - execution 3: failed
- validate_commit_message
  - execution 0: success
  - execution 1: N/A (cache from execution 0)
  - execution 2: success
  - execution 3: success

Viewing validate_commit_message without this PR will show a broken execution dropdown, and you can manually adjust the execution number at the end of the link and you can successfully view 0, 2, and 3. You cannot view the non-existent 1.

With this PR, you can view execution = 1 (the 2nd execution) and it will display the same information as execution = 0 (the 1st execution). As well as a proper dropdown that has access to all tasks (the frontend also recieves if it was cached here too)

It also returns the IsCachedResult and FromExecution inside the APITask as well, but the frontend doesn't read those values.

Viewing dropdown
Viewing execution = 0 (1st execution)
Viewing execution = 1 (2nd execution)

With PR:
Screen Shot 2022-08-18 at 9 29 45 AM
Screen Shot 2022-08-18 at 9 20 55 AM
Screen Shot 2022-08-18 at 9 29 34 AM

Without PR:
Screen Shot 2022-08-18 at 9 37 21 AM
Screen Shot 2022-08-18 at 9 37 14 AM
Screen Shot 2022-08-18 at 9 36 57 AM

### DIFF
--- a/graphql/query_resolver.go
+++ b/graphql/query_resolver.go
@@ -499,7 +499,9 @@ func (r *queryResolver) MyVolumes(ctx context.Context) ([]*restModel.APIVolume, 
 }
 
 func (r *queryResolver) Task(ctx context.Context, taskID string, execution *int) (*restModel.APITask, error) {
-	dbTask, err := task.FindOneIdAndExecutionWithDisplayStatus(taskID, execution)
+	dbTask, err := task.FindOneIdWithHighestExecutionWithDisplayStatus(taskID, *execution)
+	isCached := dbTask.Execution != *execution
+	from_execution := execution
 	if err != nil {
 		return nil, ResourceNotFound.Send(ctx, err.Error())
 	}
@@ -509,6 +511,10 @@ func (r *queryResolver) Task(ctx context.Context, taskID string, execution *int)
 	apiTask, err := getAPITaskFromTask(ctx, r.sc.GetURL(), *dbTask)
 	if err != nil {
 		return nil, InternalServerError.Send(ctx, "error converting task")
+	}
+	if isCached {
+		apiTask.IsCachedResult = isCached
+		apiTask.FromExecution = *from_execution
 	}
 	return apiTask, err
 }
@@ -529,7 +535,13 @@ func (r *queryResolver) TaskAllExecutions(ctx context.Context, taskID string) ([
 			return nil, ResourceNotFound.Send(ctx, err.Error())
 		}
 		if dbTask == nil {
-			return nil, werrors.Errorf("unable to find task %s", taskID)
+			// If not found, grab the latest execution and display those results
+			apiTask := *allTasks[i-1]
+			apiTask.IsCachedResult = true
+			apiTask.FromExecution = apiTask.Execution
+			apiTask.Execution = i
+			allTasks = append(allTasks, &apiTask)
+			continue
 		}
 		var apiTask *restModel.APITask
 		apiTask, err = getAPITaskFromTask(ctx, r.sc.GetURL(), *dbTask)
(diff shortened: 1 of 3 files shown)


model/task/db.go
```diff
@@ -1230,6 +1230,23 @@ func FindByIdExecution(id string, execution *int) (*Task, error) {
...
+	}
+	var task *Task
+	var err error
+	for i := *execution; i >= 0; i-- {
+		task, err = FindByIdExecution(id, execution)
+		if err == nil && task != nil {
+			break
+		}
+	}
+	return task, errors.Wrap(err, "finding task with highest execution")
```
bynn, line 1247: lets return the task id

rest/model/task.go
```diff
@@ -76,6 +76,11 @@ type APITask struct {
 	AMI                     *string             `json:"ami"`
 	MustHaveResults         bool                `json:"must_have_test_results"`
 	BaseTask                APIBaseTaskInfo     `json:"base_task"`
+
+	// These fields are only for cached tasks (tasks displaying info from a previous task)
+	IsCachedResult bool `json:"is_cached"`
+	FromExecution  int  `json:"from_execution"`
```
bynn, line 82: do we want  to differentiate between 0 and nil?

model/task/db.go
```diff
@@ -1274,6 +1291,24 @@ func FindOneIdAndExecutionWithDisplayStatus(id string, execution *int) (*Task, e
...
+	}
+	var task *Task
+	var err error
+	for i := *execution; i >= 0; i-- {
+		task, err = FindOneIdAndExecutionWithDisplayStatus(id, execution)
+		if err == nil && task != nil {
+			break
+		}
+	}
+	return task, errors.Wrap(err, "finding task with highest execution")
```
bynn, line 1309: same with task id

bynn (changes requested):

graphql/query_resolver.go
```diff
@@ -510,6 +510,11 @@ func (r *queryResolver) Task(ctx context.Context, taskID string, execution *int)
 	if err != nil {
 		return nil, InternalServerError.Send(ctx, "error converting task")
 	}
+	// If the executions don't match, the dbTask is a previous task (this task was cached)
+	if dbTask.Execution != *execution {
```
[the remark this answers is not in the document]
ZackarySantana: Which utility function could help with this? For now I added a manual check to see if the execution is nil for now

graphql/query_resolver.go
```diff
@@ -529,7 +534,13 @@ func (r *queryResolver) TaskAllExecutions(ctx context.Context, taskID string) ([
 			return nil, ResourceNotFound.Send(ctx, err.Error())
 		}
 		if dbTask == nil {
-			return nil, werrors.Errorf("unable to find task %s", taskID)
+			// If not found, grab the latest execution and display those results
```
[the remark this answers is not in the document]
ZackarySantana: I added some functionality in case of an error, but this function is called for the dropdown info for an execution task/display task/task. So my thinking was, when the latest execution wasn't found, or a specific execution wasn't found, its result is cached from the last execution. This accounts for if the latest also grabbed their results from their previous.

rest/model/task.go
ZackarySantana: I changed to be an int pointer with omitempty so it doesn't show up with clients using the api routes (this is in APITask), but it will be accessible via spruce so data displaying it was cached is there.

PR #5876, opened 2022-08-18
EVG-16983 Update stranded restarts to account for ResetFailedWhenFinished

ZackarySantana:
[EVG-16983](https://jira.mongodb.org/browse/EVG-16983)

### Description 
Adds checks to test if a display task has ResetFailedWhenFinished set to true (overriden by RestartWhenFinished)

### Testing 
Existing unit tests and TBA testing

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1570,36 +1570,25 @@ func resetStrandedTask(t *task.Task) error {
 		// If the task has already exceeded the unschedulable threshold, we
 		// don't want to restart it, so just mark it as finished.
 		if t.DisplayOnly {
-			var execTasks []task.Task
-			var err error
-
-			if t.ResetFailedWhenFinished && !t.ResetWhenFinished {
-				execTasks, err = task.Find(task.FailedTasksByIds(t.ExecutionTasks))
-			} else {
-				execTasks, err = task.FindAll(db.Query(task.ByIds(t.ExecutionTasks)))
-			}
-
+			execTasks, err := task.FindAll(db.Query(task.ByIds(t.ExecutionTasks)))
 			if err != nil {
 				return errors.Wrap(err, "finding execution tasks")
 			}
-
 			if len(execTasks) == 0 {
 				return errors.Wrap(err, "no execution tasks found")
 			}
-
 			for _, execTask := range execTasks {
-				if err != nil { // Keeping existing error check logic?
-					return errors.Wrap(err, "finding execution task")
-				}
-				if err := MarkEnd(&execTask, evergreen.MonitorPackage, time.Now(), &t.Details, false); err != nil {
-					return errors.Wrap(err, "marking execution task as ended")
+				if evergreen.IsFinishedTaskStatus(execTask.Status) {
+					if err := MarkEnd(&execTask, evergreen.MonitorPackage, time.Now(), &t.Details, false); err != nil {
+						return errors.Wrap(err, "marking execution task as ended")
+					}
 				}
 			}
 		}
 		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &t.Details, false))
 	}
 
-	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, t.ResetFailedWhenFinished && !t.ResetWhenFinished, &t.Details), "resetting task")
+	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, true, &t.Details), "resetting task")
 }
 
 // ResetTaskOrDisplayTask is a wrapper for TryResetTask that handles execution and display tasks that are restarted
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1583,6 +1583,10 @@ func resetStrandedTask(t *task.Task) error {
 				return errors.Wrap(err, "finding execution tasks")
 			}
 
+			if len(execTasks) == 0 {
+				return errors.Wrap(err, "no execution tasks found")
+			}
+
 			for _, execTask := range execTasks {
 				if err != nil { // Keeping existing error check logic?
 					return errors.Wrap(err, "finding execution task")
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1574,11 +1574,8 @@ func resetStrandedTask(t *task.Task) error {
 			if err != nil {
 				return errors.Wrap(err, "finding execution tasks")
 			}
-			if len(execTasks) == 0 {
-				return errors.Wrap(err, "no execution tasks found")
-			}
 			for _, execTask := range execTasks {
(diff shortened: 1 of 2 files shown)


model/task_lifecycle.go
```diff
@@ -1570,24 +1570,32 @@ func resetStrandedTask(t *task.Task) error {
...
+
+			if err != nil {
+				return errors.Wrap(err, "finding execution tasks")
+			}
+
+			for _, execTask := range execTasks {
+				if err != nil { // Keeping existing error check logic?
 					return errors.Wrap(err, "finding execution task")
 				}
-				if execTask == nil {
```
bynn, line 1579: should we still check for empty exec tasks? 
like len(execTasks) = 0?

bynn (changes requested):

model/task_lifecycle.go
ZackarySantana: I added a test to see if it's 0 and to error out. But i realized that this method is often given the exec task as the running host and not the display task so I don't know if this is the right logic to add here

model/task_lifecycle.go
```diff
@@ -1570,24 +1570,36 @@ func resetStrandedTask(t *task.Task) error {
...
 		// don't want to restart it, so just mark it as finished.
 		if t.DisplayOnly {
-			for _, etID := range t.ExecutionTasks {
-				var execTask *task.Task
-				execTask, err := task.FindOneId(etID)
-				if err != nil {
+			var execTasks []task.Task
+			var err error
+
+			if t.ResetFailedWhenFinished && !t.ResetWhenFinished {
```
ablack12, line 1576: Are you only querying failed tasks in case not all tasks were restarted? We actually should be clearing this field when we do restarts (cc @bynn do we do that?), otherwise we'll be trying to restart on a loop. I think we should just query all execution tasks and mark end any that's not already finished.

model/task_lifecycle.go
```diff
@@ -1570,24 +1570,36 @@ func resetStrandedTask(t *task.Task) error {
...
+				if err := MarkEnd(&execTask, evergreen.MonitorPackage, time.Now(), &t.Details, false); err != nil {
 					return errors.Wrap(err, "marking execution task as ended")
 				}
 			}
 		}
 		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &t.Details, false))
 	}
 
-	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, false, &t.Details), "resetting task")
+	return errors.Wrap(ResetTaskOrDisplayTask(t, evergreen.User, evergreen.MonitorPackage, t.ResetFailedWhenFinished && !t.ResetWhenFinished, &t.Details), "resetting task")
```
ablack12, line 1602: If you look at what ResetTaskOrDisplayTask does, it uses this boolean to determine whether to _set_ ResetFailedWhenFinished, or whether to set ResetWhenFinished. So we should just be setting this to true so that we toggle ResetFailedWhenFinished to true. If a user has already requested to restart the whole task, then ResetWhenFinished will also already be true and that'll be handled later.

ablack12 (changes requested):

model/task_lifecycle.go
ZackarySantana: I changed to query all tasks and then only mark end for those completed.

model/task_lifecycle.go
ZackarySantana: I forced it to true instead of this logic! I also tested for completed status in the loop above

model/task_lifecycle.go
```diff
@@ -1570,24 +1570,25 @@ func resetStrandedTask(t *task.Task) error {
...
-				if execTask == nil {
-					return errors.New("execution task not found")
-				}
-				if err := MarkEnd(execTask, evergreen.MonitorPackage, time.Now(), &t.Details, false); err != nil {
-					return errors.Wrap(err, "marking execution task as ended")
+			execTasks, err := task.FindAll(db.Query(task.ByIds(t.ExecutionTasks)))
+			if err != nil {
+				return errors.Wrap(err, "finding execution tasks")
+			}
+			if len(execTasks) == 0 {
```
ablack12, line 1577: We basically check with the old code, since if t.ExecutionTasks is length 0 we just skip the loop, so I think we can drop this check (cc @bynn)

model/task_lifecycle.go
```diff
@@ -1570,24 +1570,25 @@ func resetStrandedTask(t *task.Task) error {
...
+				return errors.Wrap(err, "no execution tasks found")
+			}
+			for _, execTask := range execTasks {
+				if evergreen.IsFinishedTaskStatus(execTask.Status) {
```
ablack12, line 1581: Should be the opposite; we should only mark end if it's _not_ finished

model/task_lifecycle_test.go
```diff
@@ -3367,6 +3367,73 @@ func TestClearAndResetStaleStrandedHostTask(t *testing.T) {
...
+
+	dispTask := &task.Task{
+		Id:                      "dt",
+		Status:                  evergreen.TaskStarted,
+		Version:                 "version",
+		Activated:               true,
+		ActivatedTime:           time.Now(),
+		BuildId:                 "b",
+		ExecutionTasks:          []string{"et1", "et2"},
+		ResetFailedWhenFinished: true,
```
ablack12, line 3381: I think you shouldn't set this here since the point of testing ClearAndResetStrandedHostTask is that it's supposed to do the setting.

model/task_lifecycle_test.go
```diff
@@ -3367,6 +3367,73 @@ func TestClearAndResetStaleStrandedHostTask(t *testing.T) {
...
+		Id: "version",
+	}
+	assert.NoError(t, v.Insert())
+
+	assert.NoError(t, ClearAndResetStrandedHostTask(h))
+	restartedDisplayTask, err := task.FindOne(db.Query(task.ById("dt")))
+	assert.NoError(t, err)
+	assert.Equal(t, evergreen.TaskUndispatched, restartedDisplayTask.Status)
+	assert.Equal(t, 1, restartedDisplayTask.Execution)
+	restartedExecutionTask, err := task.FindOne(db.Query(task.ById("et1")))
```
ablack12, line 3427: Can we check that the old version of this task is marked as failed?

model/task_lifecycle_test.go
```diff
@@ -3367,6 +3367,73 @@ func TestClearAndResetStaleStrandedHostTask(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, 1, restartedExecutionTask.Execution)
```
ablack12, line 3429: Can we also check LPE for this and the below?

model/task_lifecycle.go
bynn: ah makes sense

bynn (changes requested):

ablack12 (approved):
🚀

bynn (approved):
i think the tests all look good now!

PR #5877, opened 2022-08-19
EVG-17556 check webhooks at more places

ablack12:
[EVG-17556](https://jira.mongodb.org/browse/EVG-17556)

### Description 
Check webhooks during project creation and when owner/repo changes.
Created https://github.com/evergreen-ci/spruce/pull/1407 to fix spruce text.

### Testing 
Tested with a local project and verified that we attempted to setup webhooks with a new project.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -687,7 +687,7 @@ func (p *ProjectRef) AttachToRepo(u *user.DBUser) error {
 // AttachToNewRepo modifies the project's owner/repo, updates the old and new repo scopes (if relevant), and
 // updates the project to point to the new repo. Any Github project conflicts are disabled.
 // If no repo ref currently exists for the new repo, the user attaching it will be added as the repo ref admin.
-func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
+func (p *ProjectRef) AttachToNewRepo(ctx context.Context, u *user.DBUser) error {
 	before, err := GetProjectSettingsById(p.Id, false)
 	if err != nil {
 		return errors.Wrap(err, "getting before project settings event")
@@ -718,6 +718,16 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
 	})
 	if err != nil {
 		return errors.Wrap(err, "updating owner/repo in the DB")
+	}
+	_, err = EnableWebhooks(ctx, p)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
+			"message":            "error enabling webhooks",
+			"project_id":         p.Id,
+			"project_identifier": p.Identifier,
+			"owner":              p.Owner,
+			"repo":               p.Repo,
+		}))
 	}
 	return GetAndLogProjectModified(p.Id, u.Id, false, before)
 }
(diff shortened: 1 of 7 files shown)


model/project_ref.go
```diff
@@ -719,6 +719,16 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
 	if err != nil {
 		return errors.Wrap(err, "updating owner/repo in the DB")
 	}
+	_, err = EnableWebhooks(ctx, p)
```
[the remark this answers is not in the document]
ablack12: validateOwnerAndRepo ensures that these are set 👍 (same thing below)

model/project_ref.go
```diff
@@ -719,6 +719,16 @@ func (p *ProjectRef) AttachToNewRepo(u *user.DBUser) error {
...
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
+			"message":            "error enabling webhooks",
```
malikchaya2, line 725: `enabling webhooks` :)

rest/data/project.go
```diff
@@ -101,6 +101,16 @@ func CreateProject(projectRef *model.ProjectRef, u *user.DBUser) error {
 		"project_identifier": projectRef.Identifier,
 		"user":               u.DisplayName(),
 	}))
+	_, err = model.EnableWebhooks(ctx, projectRef)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
```
malikchaya2, line 106: same

rest/data/project_settings.go
```diff
@@ -134,12 +134,11 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 			if err = handleGithubConflicts(mergedBeforeRef, "Changing owner/repo"); err != nil {
 				return nil, err
 			}
-			// check if webhook is enabled if the owner/repo has changed
+			// Check if webhook is enabled if the owner/repo has changed
 			_, err = model.EnableWebhooks(ctx, mergedProjectRef)
 			if err != nil {
 				return nil, errors.Wrapf(err, "enabling webhooks for project '%s'", projectId)
 			}
-			modified = true
```
malikchaya2, line 142: why is this no longer needed?

model/project_ref.go
Kimchelly: This only applies to `errors.Wrap` and `errors.Wrapf`. For `message.Fields`, it doesn't apply because it's a log message, not a wrapped error.

rest/data/project_settings.go
ablack12: I was thinking there wouldn't be anything to log here but actually I think that's inaccurate since we log if webhooks have been enabled, so putting back.

ablack12:
> I don't know what the state of maintenance for the legacy project settings page is, but if we still plan on maintaining it at all, can we have [the legacy modify project route](https://github.com/evergreen-ci/evergreen/blob/8a1a9e31be3646bdc46bec67eaef6b655caa0377/service/project.go#L381-L429) also just call model.EnableWebhooks? 

I took a quick glance; this doesn't super work since the legacy UI constructs the before event log, so it needs to look up the hook separately to decide whether or not to add it to the before event. Doesn't seem worth untangling since we aren't adding any new things to this page now.

Kimchelly (approved):

PR #5880, opened 2022-08-19
EVG-16900: clean up stale pod definitions

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16900

### Description 
Create job to clean up stale unused pod definitions after they pass a TTL. I arbitrarily set it to 30 days, but we could change it if it's too low/high.

### Testing 
Added unit tests.

### DIFF
--- a/units/pod_creation.go
+++ b/units/pod_creation.go
@@ -94,8 +94,7 @@ func (j *podCreationJob) Run(ctx context.Context) {
 			})
 		}
 	}()
-
-	if err := j.populateIfUnset(ctx); err != nil {
+	if err := j.populateIfUnset(); err != nil {
 		j.AddRetryableError(err)
 		return
 	}
@@ -125,11 +124,6 @@ func (j *podCreationJob) Run(ctx context.Context) {
 			return
 		}
 
-		if err := podDef.UpdateLastAccessed(); err != nil {
-			j.AddRetryableError(errors.Wrapf(err, "updating last access time for pod definition '%s'", podDef.ID))
-			return
-		}
-
 		p, err := j.ecsPodCreator.CreatePodFromExistingDefinition(ctx, cloud.ExportECSPodDefinition(*podDef), *execOpts)
 		if err != nil {
 			j.AddRetryableError(errors.Wrap(err, "starting pod"))
@@ -155,7 +149,7 @@ func (j *podCreationJob) Run(ctx context.Context) {
 	}
 }
 
-func (j *podCreationJob) populateIfUnset(ctx context.Context) error {
+func (j *podCreationJob) populateIfUnset() error {
 	if j.env == nil {
 		j.env = evergreen.GetEnvironment()
 	}
@@ -203,6 +197,17 @@ func (j *podCreationJob) checkForPodDefinition(family string) (*definition.PodDe
 	}
 	if podDef == nil {
 		return nil, errors.Errorf("pod definition with family '%s' not found", family)
+	}
+
+	grip.WarningWhen(podDef.LastAccessed.Add(podDefinitionTTL).Before(time.Now()), message.Fields{
+		"message":        "Using a pod definition whose TTL has already elapsed, so it's at risk of being cleaned up. Starting this pod may fail if the pod definition gets cleaned up.",
+		"pod":            j.pod.ID,
+		"pod_definition": podDef.ID,
+		"job":            j.ID(),
+	})
+
+	if err := podDef.UpdateLastAccessed(); err != nil {
+		return nil, errors.Wrapf(err, "updating last access time for pod definition '%s'", podDef.ID)
 	}
 
 	return podDef, nil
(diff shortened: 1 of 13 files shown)


units/pod_creation.go
```diff
@@ -125,11 +124,6 @@ func (j *podCreationJob) Run(ctx context.Context) {
 			return
 		}
 
-		if err := podDef.UpdateLastAccessed(); err != nil {
```
hadjri, line 128: Why are we now unconditionally updating `LastAccessed` rather than only when the pod is initializing?

units/pod_definition_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+
+	podDefs, err := definition.FindByLastAccessedBefore(podDefinitionTTL, -1)
+	if err != nil {
+		j.AddError(err)
+		return
+	}
+
+	for _, podDef := range podDefs {
+		if podDef.ExternalID == "" {
+			if err := podDef.Remove(); err != nil {
```
hadjri, line 81: I may be missing something - wouldn't we want to remove all pod definitions that have passed TTL?

units/pod_definition_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Use the latest service flags instead of those cached in the environment.
+	settings := *j.env.Settings()
+	if err := settings.ServiceFlags.Get(j.env); err != nil {
+		return errors.Wrap(err, "getting service flags")
+	}
+	j.settings = settings
+
+	if j.ecsClient == nil {
+		client, err := cloud.MakeECSClient(&settings)
```
hadjri, line 123: We can directly set `j.ecsClient` on this line

units/pod_definition_cleanup.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return errors.Wrap(err, "initializing ECS client")
+		}
+		j.ecsClient = client
+	}
+
+	if j.podDefMgr == nil {
+		podDefMgr, err := cloud.MakeECSPodDefinitionManager(j.ecsClient, nil)
```
hadjri, line 131: same here

units/pod_definition_cleanup.go
Kimchelly: We do. This loop removes all the pod definitions that pass their TTL regardless of whether they're tied to an external ID or not. The first if on L80 handles a future-proofing case where a pod definition exists but is not associated with an external pod definition. That can't supposedly happen now, but I want to make sure that if a pod definition like that could exist, it gets deleted by the cleanup job as well since it serves no purpose.

units/pod_definition_cleanup.go
Kimchelly: I think it's okay to leave it as is. I have a small personal preference for not declaring `err` vars just for the sake of assigning a variable in one line.

units/pod_definition_cleanup.go
Kimchelly: Same as above.

units/pod_creation.go
Kimchelly: The GitHub PR view is buggy and isn't showing this comment when I click on the files changed... It's also having some difficulty letting me submit this comment. 😭 

This logic is equivalent to what it was before. I just moved updating the access time into `checkForPodDefinition`.

units/pod_definition_cleanup.go
hadjri: It looks like we're only removing the pod definitions from our own collection in this future proofing case - shouldn't we call `Remove` no matter what then?

units/pod_creation.go
hadjri: Ah, I see `checkForPodDefinition` returns early if the pod isnt initializing. Sorry! Never mind

units/pod_definition_cleanup.go
Kimchelly: Since the pod definition manager accepts an external storage cache, it's implicitly deleting the DB doc [as part of the manager functionality](https://github.com/evergreen-ci/cocoa/blob/df9129525b0266148e1191aa3dd2a1d410fb208f/ecs/pod_definition_manager.go#L159-L161), so [this](https://github.com/kimchelly/evergreen/blob/c2a6d2600c5146ea1254d35cae5281152c825c8a/model/pod/definition/definition.go#L85-L95) is called by the pod definition manager. I tested that this is working correctly [here](https://github.com/Kimchelly/evergreen/blob/c2a6d2600c5146ea1254d35cae5281152c825c8a/units/pod_definition_cleanup_test.go#L46-L48).

hadjri: Gotcha, my bad

hadjri (approved):

units/pod_definition_cleanup.go
Kimchelly: No problem, I realize that it might not be that clear just from a glance.

PR #5883, opened 2022-08-22
EVG-16990 Add container smoke tests to self-tests

hadjri:
[EVG-16990](https://jira.mongodb.org/browse/EVG-16990)

### Description 
Split up task smoke testing into two components: host and container tasks. Augmented the Makefile to have these as two separate commands and pass in the agent mode into each one.

Upgraded a few agent routes to v2 because a containerized task relies on there being pod middleware in place for those routes (which isn't available on v1).

Updated the mock `agent.yml` and relevant smoke mock testdata files to include a new containerized task to be used in the smoke test.

### Testing 
[New smoke test passes](https://spruce.mongodb.com/task/evergreen_ubuntu1604_smoke_test_container_task_patch_f0573cc00ff1af555ca4c7a2d342239ac062b3d6_631018d8e3c3315a7b364bd0_22_09_01_02_28_42/logs?execution=0)


malikchaya2:
Do we know for sure that the smoke test test patch picked up the new changes? Are there evidence of it in the test logs?

malikchaya2:
Can we please also run this in the new smoke test branch project? In addition to making sure that these run, I want to make sure that these changes don't require any changes to that enviroment.

hadjri:
> Can we please also run this in the new smoke test branch project? In addition to making sure that these run, It would also help us know if this introduces any changes that would necessitate a change to that project.

This smoke test is slightly different from the existing host-based smoke test in that it doesn't rely on or use the repotracker, so this test can be debugged / ran without the use of the smoke testing branch that's been set up for the existing smoke tests.  This is because the pod task dispatcher needs to talk to AWS ECS during runtime to add a task to its queue to get picked up later by the agent, which is difficult to mock here in the smoke test (compared to the existing smoke task test which doesn't rely on external services).

So, this test starts with a pod defined in `testdata` polling for next tasks from a pod dispatcher which has a predefined queue that isn't dynamically generated from the repotracker, but an existing task in `testdata`.  We then test that:

- The agent spins up in pod mode and initializes itself with the proper configuration needed to run container tasks
- The agent enters a loop to continuously poll for the next task in the dispatcher queue and receives the task that we put in there
- The task is retrieved properly and completes successfully by checking for succeeded status and log substrings

testdata/local/pods.json
```diff
@@ -0,0 +1,2 @@
+{"_id":"localhost", "secret":"de249183582947721fdfb2ea1796574b","agent_revision":"foo", "type": "agent", "status": "running", "task_creation_opts": {"env_secrets": {"POD_SECRET": "de249183582947721fdfb2ea1796574b"}}}
```
Kimchelly, line 1: * Is `secret` necessary here? Seems like this just needs the `env_secrets`. Also [`env_secrets`](https://github.com/evergreen-ci/evergreen/blob/a822913f1b3cb5e2977e7d00866b8f289febdecf/model/pod/pod.go#L276-L279) seems to have a different structure than what's given here.
* * `agent_revision` seems like it should be `agent_version`?

testdata/local/pods.json
```diff
@@ -0,0 +1,2 @@
...
+{"_id":"localhost2", "secret":"de249183582947721fdfb2ea1796574b","agent_revision":"foo", "type": "agent", "status": "running"}
```
Kimchelly, line 2: I see the `localhost` doc is used for the agent running the smoke test, but what's the second pod document for?

testdata/local/pod_dispatchers.json
```diff
@@ -0,0 +1 @@
+{"_id":"dispatcher0", "pod_ids":  ["localhost"]}
```
Kimchelly, line 1: Do we need this in the local data? Local Evergreen typically doesn't have data to run tasks.

self-tests.yml
```diff
@@ -344,8 +344,11 @@ tasks:
     name: dist-unsigned
     activate: false
   - <<: *run-smoke-test
-    name: smoke-test-task
+    name: smoke-test-host-task
     tags: ["smoke"]
+  - <<: *run-smoke-test
+    name: smoke-test-container-task
+    tags: [ "smoke" ]
```
Kimchelly, line 351: nit: remove whitespace inside square brackets.

scripts/agent.yml
```diff
@@ -211,3 +219,20 @@ buildvariants:
...
+    display_name: container_bv
+    run_on:
+      - container
+    tasks:
+      - name: container_task
+
+containers:
+  - name: container
+    working_dir: /
+    image: "hadjri/evg-e2e-test-ubuntu"
```
Kimchelly, line 232: Can we make a TODO to eventually use a more official image repo, like one in our AWS account's ECR repository? I'm not sure if this configuration value even matters since the pod and pod document are manually populated in the smoke test data anyways.

rest/route/agent.go
```diff
@@ -922,37 +929,42 @@ func (h *startTaskHandler) Run(ctx context.Context) gimlet.Responder {
...
-			return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "queuing host idle stats for %s", msg))
+		if host.Distro.IsEphemeral() {
+			queue := h.env.RemoteQueue()
+			job := units.NewCollectHostIdleDataJob(host, t, idleTimeStartAt, t.StartTime)
+			if err = queue.Put(ctx, job); err != nil {
+				return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "queuing host idle stats for %s", msg))
+			}
 		}
+	} else {
+		msg = fmt.Sprintf("task %s started on container %s", t.Id, t.Container)
```
Kimchelly, line 967: Should this log the pod ID instead of the container name?

Kimchelly, line 967: Since there's no equivalent to the host idle data job, can we add a TODO here + ticket to eventually collect these stats? Would be good to know how long the pod is idle before running the task.

operations/service_deploy_smoke_utils.go
```diff
@@ -216,43 +267,49 @@ func checkTaskLog(body []byte) error {
...
 	return nil
 }
 
-func checkTask(client *http.Client, username, key string, builds []apimodels.APIBuild, taskIndex int) (apimodels.APITask, error) {
+func checkTask(client *http.Client, username, key string, taskId string) (apimodels.APITask, error) {
 	task := apimodels.APITask{}
-	grip.Infof("checking for task %s", builds[0].Tasks[taskIndex])
-	r, err := http.NewRequest("GET", smokeUrlPrefix+smokeUiPort+"/rest/v2/tasks/"+builds[0].Tasks[taskIndex], nil)
+	grip.Infof("checking for task %s", taskId)
+	r, err := http.NewRequest("GET", smokeUrlPrefix+smokeUiPort+"/rest/v2/tasks/"+taskId, nil)
```
Kimchelly, line 312: nit: `http.MethodGet` instead of `"GET"`.

operations/service_deploy_smoke_utils.go
```diff
@@ -93,14 +95,57 @@ func getLatestGithubCommit() (string, error) {
...
+		if err != nil {
+			return errors.WithStack(err)
+		}
+
+		if task.Status == evergreen.TaskFailed {
+			return errors.Errorf("task status is %s (expected %s)", task.Status, evergreen.TaskSucceeded)
+		}
+		if task.Status != evergreen.TaskSucceeded {
+			grip.Infof("found task is status %s", task.Status)
+			task = apimodels.APITask{
```
Kimchelly, line 123: This isn't wrong, but it's written kind of strangely because the only reason the loop declares `task` outside the for block is to interpolate the status on L107. Can we simplify this down a bit to just set a string for the task status after `checkTask` rather than recycling the entire task between iterations?

operations/service_deploy_smoke_utils.go
```diff
@@ -93,14 +95,57 @@ func getLatestGithubCommit() (string, error) {
...
+		}
+
+		// retry for *slightly* delayed logger closing
+		err = checkLog(task, client, agent.PodMode, username, key)
+		if err != nil {
+			return err
+		}
+
+		grip.Info("Successfully checked container task")
+		return nil
```
Kimchelly, line 136: Seems like some of the copy-pasted for loop logic around task checking could be consolidated between `checkContainerTask` and `checkHostTaskByCommit` by creating a helper function that accepts the task ID + agent mode and returns the current task status.

operations/service_deploy_smoke_utils.go
```diff
@@ -93,14 +95,57 @@ func getLatestGithubCommit() (string, error) {
 	return "", errors.New("could not find latest commit in response")
 }
 
-func checkTaskByCommit(username, key string) error {
+func checkContainerTask(username, key string) error {
+	client := utility.GetHTTPClient()
+	defer utility.PutHTTPClient(client)
+
+	var task apimodels.APITask
```
Kimchelly, line 102: This smoke test is somewhat different from the host one because it tests the system in a less e2e way. This one solely tests that the agent can run the task assuming the smoke DB is already pre-populated with a perfectly consistent state ready to run the task. In comparison, the host one is testing a much larger breadth of the system because it triggers the repotracker (which creates the task in the DB) + host task scheduler (which generates the task queue).

Can we make a note of the difference between the host and container smoke test somewhere so we can retain knowledge of their differences?

operations/service_deploy_smoke_utils.go
```diff
@@ -216,43 +267,49 @@ func checkTaskLog(body []byte) error {
...
+		} else {
+			if strings.Contains(page, "teardown_group") {
+				return errors.New("teardown_group should only run in last (fourth) task")
+			}
 		}
-	} else {
-		if strings.Contains(page, "teardown_group") {
-			return errors.New("teardown_group should only run in last (fourth) task")
+	} else if mode == agent.PodMode {
+		if !strings.Contains(page, "container task") {
```
Kimchelly, line 301: Can we add a TODO here for PM-2617 to add task groups to the smoke test once they're supported?

operations/service_deploy_smoke_utils.go
```diff
@@ -20,7 +21,8 @@ const (
 	// uiPort is the local port the UI will listen on.
 	smokeUiPort = ":9090"
 	// urlPrefix is the localhost prefix for accessing local Evergreen.
-	smokeUrlPrefix = "http://localhost"
+	smokeUrlPrefix  = "http://localhost"
+	containerTaskId = "evergreen_container_bv_container_task_a71e20e60918bb97d41422e94d04822be2a22e8e_22_08_22_13_44_49"
```
Kimchelly, line 25: nit: can we call this `smokeContainerTaskID` or something like that?

Kimchelly (changes requested):

operations/service_deploy_smoke.go
```diff
@@ -118,14 +124,15 @@ func smokeStartEvergreen() cli.Command {
...
+			mode := c.String(modeFlagName)
 
 			exit := make(chan error, 3)
 
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
 			if startWeb {
-				if err := smokeRunBinary(exit, "web.service", wd, binary, "service", "web", "--conf", confPath); err != nil {
+				if err := smokeRunBinary(exit, "web.service", wd, binary, "service", "web", "--conf", confPath, "--db", "mci_smoke"); err != nil {
```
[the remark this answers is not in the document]
hadjri: Ah yes I believe this is leftover from debugging. Removing.

testdata/local/pods.json
hadjri: It's not used for anything I just felt the need to insert multiple pod documents for some reason. Removing.

testdata/local/pod_dispatchers.json
hadjri: I was just putting a dispatcher in the local DB in case we'd need it for some debugging purpose in the future but since it's not needed today I can omit it.

operations/service_deploy_smoke_utils.go
hadjri: Sure, done

rest/route/agent.go
hadjri: Yes, good catch

operations/service_deploy_smoke_utils.go
hadjri: Done, refactored the task checking loop to a helper function.

testdata/local/pods.json
hadjri: Yeah, the local `pods.json` deviated from the smoke `pods.json` as time / debugging progress went on, and didn't notice because I never needed to load this data locally. I'll make this file look like the one in the `smoke` folder, which is formatted correctly.

rest/route/agent.go
hadjri: Sounds good, created: https://jira.mongodb.org/browse/EVG-17647

Shall we add this to a specific epic or leave it to triage?

operations/service_deploy_smoke_utils.go
hadjri: Yes - my initial intention here was to accomplish the former, although the main difficulties I encountered were:

- The pod allocator job dynamically creates pods, whereas the `start-evergreen` command used in the smoke test uses the agent start command which requires an explicit pod ID to be passed as a parameter, which can't be one of the dynamic pod IDs generated by the allocator job
- Mocking the cocoa vaults seemed to require a good bit more work since I wasn't able to explicitly set the `vault` field on the allocator job to a `cocoaMock.Vault` in the context of the running smoke test (although you might have a better idea on how to accomplish this in a better way)

In any case, I'll update the [developer guide](https://docs.google.com/document/d/1KxK2-JhKiHg7ydoEJN6rAf63k94KE_5-DqlYBj48pig/edit#heading=h.osfhv6fi7598) with a section parallel to "smoke test debugging" explaining this situation.  Happy to make an additional ticket to work on the two points above as well if we can discuss a plan for it.

makefile
```diff
@@ -152,16 +152,21 @@ $(buildDir)/.load-local-data:$(buildDir)/load-smoke-data
 smoke-test-agent-monitor:$(localClientBinary) load-smoke-data
 	./$< service deploy start-evergreen --web --binary ./$< &
 	./$< service deploy start-evergreen --monitor --binary ./$< --distro localhost &
-	./$< service deploy test-endpoints --check-build --username admin --key abb623665fdbf368a1db980dde6ee0f0 $(smokeFile) || (pkill -f $<; exit 1)
+	./$< service deploy test-endpoints --check-build --username admin --key abb623665fdbf368a1db980dde6ee0f0
```
malikchaya2, line 155: Your explanation makes a lot of sense, and I see why it makes sense that we don't need to test the new smoke test in the sandbox. However, because there are changes to it, I think we should test the existing smoke test in the sandbox so that if something no longer passes, we can know before the merge.

hadjri: I understand, I was able to run the tasks successfully using a patch of this PR on the smoke project in staging: https://evergreen-staging.corp.mongodb.com/version/smoke_bd8d71bf821482f166fa75b3fc10f4441fde098a

scripts/agent.yml
```diff
@@ -219,13 +219,14 @@ buildvariants:
...
-  - name: container_bv
-    display_name: container_bv
+  - name: pod_bv
+    display_name: pod_bv
     run_on:
       - container
     tasks:
       - name: container_task
 
+# TODO: Use more official image repo such as an AWS ECR image repo
```
Kimchelly, line 229: Is there a ticket number for this?

Also, based on the way the smoke test is set up, I don't think the particular image even matters here since the configuration does not affect the smoke test. Maybe we can just plug in a dummy value (e.g. `not_a_real_image`) and add a comment above the `containers` section saying that the container configuration specified is irrelevant since it's not actually used in the smoke test to start a container?

operations/service_deploy_smoke_utils.go
```diff
@@ -197,21 +161,23 @@ func checkHostTaskByCommit(username, key string) error {
...
 		}
 		time.Sleep(10 * time.Second)
-		grip.Infof("checking for %d tasks (%d/30)", len(builds[0].Tasks), i+1)
+		grip.Infof("checking for %d tasks (%d/30)", len(tasks), i+1)
 
-		var err error
-		for t := 0; t < len(builds[0].Tasks); t++ {
-			taskId := builds[0].Tasks[t]
-			task, err = checkTask(client, username, key, taskId)
+		for t := 0; t < len(tasks); t++ {
```
Kimchelly, line 178: nit: this could be re-written to a `for _, taskId := range tasks` since it doesn't use the index for anything else.

operations/service_deploy_smoke_utils.go
```diff
@@ -221,17 +187,15 @@ OUTER:
...
 				continue OUTER
 			}
-			err = checkLog(task, client, agent.HostMode, username, key)
+			err = checkLog(task, client, mode, username, key)
 			if err != nil {
 				return err
 			}
 		}
-		grip.Info("Successfully checked tasks")
+		grip.Infof("Successfully checked %s tasks", string(mode))
```
Kimchelly, line 198: Would it help to include the number of tasks that were checked here for debugging?

operations/service_deploy_smoke_utils.go
Kimchelly: Yeah, the dynamic allocation poses an issue for the smoke test since we can't know the pod credentials necessary to run the task currently. I believe some improvements we could make is to fake out the agent-creation step:
* Update the `testdata/smoke/admin.json` to [use the mock client for ECS and Secrets Manager](https://github.com/kimchelly/evergreen/blob/60f71d56df8ac9c69609db28f62a70684c73afcf/config_test/evg_settings.yml#L31-L37).
* Load the smoke test data's project ref from `testdata/smoke/project/project_ref.json` into the smoke DB and then send a REST request to set the project's pod secret value (so it's populated in the fake Secrets Manager).
* Trigger the repotracker to start the process of creating the task(s) for the version. If everything is working properly, it should use a project pod secret that we set, so the smoke test will know the plaintext secret being shared between the agent and app server.
* Continuously poll the fake app server until it tells us what pod is allocated for the task. We'll probably end up exposing this functionality in the REST routes anyways as part of the container UI project since the patch page will have to show which pod is running each task anyways.
* Use the pod ID and the project's pod secret to start an agent that has the correct auth and proceed with the test.

We can definitely leave this for future work though. I think this smoke test is sufficient for now.

rest/route/agent.go
Kimchelly: I think we can leave it for triage and do it whenever there's time. There's already a couple of small one-off container tickets which aren't that urgent or relevant to any epic.

operations/service_deploy_smoke_utils.go
Kimchelly: Bumping this nit again.

operations/service_deploy_smoke_utils.go
hadjri: Sure, added `len(tasks)` to the formatted print statement

scripts/agent.yml
hadjri: You're right it doesn't currently matter - I just included it in anticipation of future improvements to this test that will need the repotracker.  I'll change the image name and update the comment saying it's not used for now.

hadjri: I'll include the work of the ticket with what you mentioned [here](https://github.com/evergreen-ci/evergreen/pull/5883#discussion_r965312592).

Ticket created here and included in comment: https://jira.mongodb.org/browse/EVG-17658

operations/service_deploy_smoke_utils.go
hadjri: Sounds good. I've created this ticket for future triage based on these ideas https://jira.mongodb.org/browse/EVG-17658

hadjri: Sounds good. I've created this ticket for future triage based on these ideas https://jira.mongodb.org/browse/EVG-17658

Kimchelly (approved):
LGTM mod the [one leftover comment](https://github.com/evergreen-ci/evergreen/pull/5883#discussion_r964094613).

makefile
malikchaya2: that one tests smoke-test-container-task and smoke-test-host-task. I think it would also be a good idea to test smoke-test-endpoints and smoke-test-task.

PR #5884, opened 2022-08-22
EVG-17456: log remote queue group stats

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17456

### Description 
Log job stats for queues in the queue group. I'm going to make a Splunk alert for queues that are unusually long.

### Testing
Tested in staging that it works as expected.

### DIFF
--- a/units/stats_amboy.go
+++ b/units/stats_amboy.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/logging"
 	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
 )
 
 const amboyStatsCollectorJobName = "amboy-stats-collector"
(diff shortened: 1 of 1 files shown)


units/stats_amboy.go
```diff
@@ -90,4 +91,24 @@ func (j *amboyStatsCollector) Run(ctx context.Context) {
...
+		// to process or has only finished processing all of its jobs very
+		// recently.
+		for _, queueName := range remoteQueueGroup.Queues(ctx) {
+			queue, err := remoteQueueGroup.Get(ctx, queueName)
+			if err != nil {
+				j.AddError(errors.Wrapf(err, "getting queue stats for queue '%s' in queue group", queueName))
+				continue
+			}
+
+			j.logger.Info(message.Fields{
```
malikchaya2, line 107: How often will this log? Should we be concerned about it potentially exhausting our splunk limit?
I wonder if there is maybe a way the logging to long queues here instead of limiting the splunk alert itself?

malikchaya2 (changes requested):

units/stats_amboy.go
Kimchelly: This logs once a minute, once per queue. My expectation is that this will create a total of 5-10 times per minute, so this doesn't pose a risk to our Splunk rate limit.

malikchaya2 (approved):

ablack12 (approved):

PR #5886, opened 2022-08-22
EVG-17424: Spoof nil arrays in project settings events

sophstad:
[EVG-17424](https://jira.mongodb.org/browse/EVG-17424)

### Description 
- To overcome limitations that come from casting project settings events to generic interfaces, add flags to indicate to the following fields to indicate that they represent `null`:
  - `FilesIgnoredFromCache`
  - `GitTagAuthorizedTeams`
  - `GitTagAuthorizedUsers`
  - `PatchTriggerAliases`
  - `PeriodicBuilds`
  - `Triggers`
  - `WorkstationConfig.SetupCommands`
- When project settings event logs are queried, check these flags in order to return `null` value

### Testing 
- Updated `TestGetProjectEvents` to ensure that PeriodicBuilds is returned as `nil`; this test currently fails on `main` ([patch](https://spruce.mongodb.com/version/6303e842a4cf470c7d8f3469/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))
- Added test verifying that existing event documents are properly read
- Tested on Spruce and verified these changes can be deployed with no UI updates required


rest/data/project_test.go
```diff
@@ -141,9 +142,14 @@ func TestProjectConnectorGetSuite(t *testing.T) {
 				EventType:    event.EventTypeProjectModified,
 				ResourceId:   projectId,
 				Data: &model.ProjectChangeEvent{
-					User:   username,
-					Before: before,
-					After:  after,
+					User: username,
+					Before: model.ProjectSettingsEvent{
```
ablack12, line 146: One thing I would also like to check in a test before going forwards is if we have a ProjectChangeEvent that is in the old format (i.e. Before is a ProjectSettings struct) does it still unmarshal correctly when retrieving from the database?

ablack12 (commented):
Looks like the right track to me! As clean as we can get it, alas, Will want to add many comments to explain why we're doing such a sad thing.

model/project_event.go
```diff
@@ -22,14 +22,36 @@ type ProjectSettings struct {
...
 
+func (p *ProjectChangeEvents) ApplyDefaults() {
+	for _, event := range *p {
+		changeEvent, isChangeEvent := event.Data.(*ProjectChangeEvent)
+		if !isChangeEvent {
+			continue
+		}
+
+		if changeEvent.Before.PeriodicBuildsDefault {
+			changeEvent.Before.ProjectRef.PeriodicBuilds = nil
```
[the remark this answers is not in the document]
sophstad: Editing `changeEvent` seems to work correctly for [RedactPrivateVars](https://github.com/evergreen-ci/evergreen/blob/de6a1e461490704f23ddd969ee4eebe0645a647a/model/project_event.go#L33-L43), and anyway `p` needs to be cast from a `ProjectChangeEventEntry` to a `ProjectChangeEvent` (they are different, it's confusing, I know). I'll revisit this if it seems like the tests reveal anything unexpected.

model/project_event.go
```diff
@@ -158,8 +190,8 @@ func LogProjectModified(projectId, username string, before, after *ProjectSettin
 
 	eventData := ProjectChangeEvent{
 		User:   username,
-		Before: *before,
-		After:  *after,
+		Before: *before.resolveDefaults(),
```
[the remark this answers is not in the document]
sophstad: LogProjectEvent also handles project added events which lack `before` and `after`, so resolving defaults here does seem to be the most straightforward approach.

model/project_event.go
```diff
@@ -22,14 +22,86 @@ type ProjectSettings struct {
...
-	Before ProjectSettings `bson:"before" json:"before"`
-	After  ProjectSettings `bson:"after" json:"after"`
+	User   string               `bson:"user" json:"user"`
+	Before ProjectSettingsEvent `bson:"before" json:"before"`
+	After  ProjectSettingsEvent `bson:"after" json:"after"`
 }
 
 type ProjectChangeEvents []ProjectChangeEventEntry
 
+// Look at the flags set in the ProjectSettingsEvent so that fields that were converted to empty arrays when casting to an interface{} can be correctly set to nil
```
malikchaya2, line 44: I'm probably missing some context, but couldn't we just iterate and check for empty arrays, and convert all empty arrays to nil?

malikchaya2: Is there ever a time when we specifically want there to be an empty array? I would assume that an empty array could always be safely changed to nil?

sophstad: The use case is a setting like this, where "override" means empty array and "default" means null
image

All of the settings I added flags for use this pattern, which is why assuming null on empty array is invalid.

malikchaya2 (approved):
Approved with nit to add comment.

model/project_event.go
```diff
@@ -22,14 +22,87 @@ type ProjectSettings struct {
 	Subscriptions      []event.Subscription `bson:"subscriptions" json:"subscriptions"`
 }
 
+type ProjectSettingsEvent struct {
+	ProjectSettings              `bson:",inline"`
+	FilesIgnoredFromCacheDefault bool `bson:"files_ignored_from_cache_default,omitempty" json:"files_ignored_from_cache_default,omitempty"`
```
ablack12, line 27: re chaya's comment thought, I think there should also be a comment here to explain the bools

model/project_event.go
ablack12: nit: function comments start with the function name and explain what the function does before going into details

model/project_event.go
```diff
@@ -147,6 +220,37 @@ func GetAndLogProjectModified(id, userId string, isRepo bool, before *ProjectSet
 	return errors.Wrap(LogProjectModified(id, userId, before, after), "logging project modified")
 }
 
+// ProjectChangeEvents must be cast to a generic interface to utilize event logging, which casts all nil objects of array types to empty arrays.
```
ablack12, line 223: same here

model/project_event.go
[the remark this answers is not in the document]
ablack12: I think the methods then can just return nil if the ProjectSettings is nil so I still find doing it in LogProjectEvent to be preferable but I don't feel too strongly

rest/data/project_test.go
```diff
@@ -347,3 +362,42 @@ func TestGetProjectAliasResults(t *testing.T) {
...
 }
+
+func TestGetLegacyProjectEvents(t *testing.T) {
+	project := &model.ProjectRef{Id: projectId}
+	require.NoError(t, project.Insert())
+
+	before := getMockProjectSettings()
+	after := getMockProjectSettings()
+
+	// Use an interface{} to mimic legacy data that was not inserted as a ProjectSettingsEvent
```
ablack12, line 373: Awesome! 🎉 🌮

rest/data/project_test.go
```diff
@@ -347,3 +362,42 @@ func TestGetProjectAliasResults(t *testing.T) {
...
+	h :=
+		event.EventLogEntry{
```
ablack12, line 375: nit: `h := event.EventLogEntry{ ` on one line is cleaner / more conventional

rest/data/project_test.go
```diff
@@ -347,3 +362,42 @@ func TestGetProjectAliasResults(t *testing.T) {
...
+			EventType:    event.EventTypeProjectModified,
+			ResourceId:   projectId,
+			Data: map[string]interface{}{
+				"user":   username,
+				"before": before,
+				"after":  after,
+			},
+		}
+
+	require.NoError(t, db.ClearCollections(event.LegacyEventLogCollection))
```
ablack12, line 387: nit: can you move this to the top of the test?

rest/data/project_test.go
```diff
@@ -347,3 +362,42 @@ func TestGetProjectAliasResults(t *testing.T) {
 	assert.Len(t, variantTasks, 1)
 	assert.Len(t, variantTasks[0].Tasks, 2)
 }
+
+func TestGetLegacyProjectEvents(t *testing.T) {
```
ablack12, line 366: Just because I'm a little nervous about schema changes, should we write a test for the backwards compatible version? (i.e. can we read the new structure into the old structure?) If you've tested that yourself though that's also fine, I feel like I have a memory of that

ablack12 (changes requested):
Great work this is tricky stuff!

model/project_ref.go
```diff
@@ -101,7 +101,7 @@ type ProjectRef struct {
 	DisabledStatsCache    *bool    `bson:"disabled_stats_cache,omitempty" json:"disabled_stats_cache,omitempty"`
 
 	// List of commands
-	WorkstationConfig WorkstationConfig `bson:"workstation_config,omitempty" json:"workstation_config,omitempty"`
+	WorkstationConfig WorkstationConfig `bson:"workstation_config" json:"workstation_config"`
```
ablack12, line 104: I'm just curious about this change; did you remove this because it wasn't doing anything or because it was actively causing a problem or? I notice we do still have it for the others.

ablack12 (approved):
LGTM with question

model/project_ref.go
ablack12: Isn't that okay though? If WorkstationConfig is omitted entirely then GitClone / SetupCommands would implicitly be nil as well right? Or did that behavior also occur when setupCommands was explicitly empty array?

ablack12: I _think_ we included omitempty on all non-nil arrays to ensure that defaulting worked correctly, although that memory might not be quite right 🤔

sophstad: Yeah, the problem is that `omitempty` is being applied on both `nil` and `[]` :(

ablack12: Okay rip. Do other things also need to not be omitempty then? (TaskAnnotationSettings has a list inside it for example)

sophstad: No, SetupCommands is the only setting that is a nested, nullable list (i.e. one of the fields we flag).

ablack12: I'm afraid someone will break this on accident since it breaks convention; maybe add a comment to say why it doesn't have omitempty?

sophstad:
Looks like Evergreen is not properly reporting the patch status, but all tasks and downstream tasks have passed 👍

PR #5888, opened 2022-08-22
EVG-17126 Do not allow resetting non-complete tasks

hadjri:
[EVG-17126](https://jira.mongodb.org/browse/EVG-17126)

### Description 
Follow up from https://github.com/evergreen-ci/evergreen/pull/5816

Previous change was to protect against resetting incomplete tasks with explicit guards when performing DB updates.  Post-change, when a request to reset an incomplete task did occur, the task was archived, and reset was skipped due to the new DB guarding.  When the task tried to reset on the same execution that was incorrectly attempted previously, duplicate old task ID errors prevented the execution from proceeding past the archival step, causing the task execution to get stuck, like [this one](https://spruce.mongodb.com/task/mongodb_mongo_master_commit_queue_compile_dbtest_patch_576ed6000ddbb266004e93e391c4955faa9a3061_62f405d461837d540617065c_22_08_10_19_24_17/logs?execution=0) which was unable to proceed past execution 0 while its other task group tasks moved forward.

This PR includes all previous changes plus additional steps to prevent duplicate key errors from preventing task resetting from proceeding.

This whole change was necessary because resetting in-progress tasks causes bursts of "wrong secret" errors to occur since the task is assigned a new secret while the original task execution persists with the old secret.  I was able to reproduce this error in staging by manually running the restart build handler (which prior to this change would reset in-progress tasks) before a build fully completed. This resulted in a [short flurry of wrong secret](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen-staging%22%20%22wrong%20secret%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1662091200&latest=1662177600&sid=1662448126.565439) errors for the affected tasks in the build that were still running when the restart request was received.  I redeployed to staging with the changes in this PR, and manually restarting a build halfway through no longer resulted in these errors.
### Testing 
Created new unit tests and added to existing tests. Deployed to staging and ran multiple versions, builds, display tasks and task groups and restarted them multiple times to check for errors doing so.


model/task/task.go
```diff
@@ -2514,13 +2515,22 @@ func (t *Task) Insert() error {
 // into the old_tasks collection. If this is a display task, its execution tasks
 // are also archived.
 func (t *Task) Archive() error {
+	if !utility.StringSliceContains(evergreen.TaskCompletedStatuses, t.Status) {
+		grip.Debug(message.Fields{
```
ablack12, line 2519: Is this logging meant to be temporary or permanent?

model/task/task.go
```diff
@@ -2514,13 +2515,22 @@ func (t *Task) Insert() error {
...
+		return nil
+	}
 	if t.DisplayOnly && len(t.ExecutionTasks) > 0 {
 		return errors.Wrapf(ArchiveMany([]Task{*t}), "archiving display task '%s'", t.Id)
 	} else {
 		// Archiving a single task.
 		archiveTask := t.makeArchivedTask()
-		err := db.Insert(OldCollection, archiveTask)
-		if err != nil && !db.IsDuplicateKey(err) {
+		_, err := db.Upsert(OldCollection, bson.M{IdKey: MakeOldID(t.Id, t.Execution)}, archiveTask)
```
ablack12, line 2532: Is the idea that we won't need this to be upsert after the deploy, or we will always want this? If the task was already archived for this execution then we wouldn't want to update right?

model/task/task.go
hadjri: I put this here initially for my staging testing but I think it'd be good to monitor post deploy, since it should allow us to better identify future attempts of the system to reset unfinished tasks and we can address that cause. 

I can create a ticket for removing once we determine it's not helpful anymore

model/task/task.go
hadjri: I was thinking of keeping it since I reasoned that a dup key error here will block the possibility of completing the reset for that execution forever, and it seems that it [happens occasionally in prod currently](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen%22%20old_tasks%20dup&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-7d%40h&latest=now&sid=1662673545.1073158).  But this seems like a bug specific to the `task-execution-timeout` job possibly picking up the same execution twice, which I don't think is a big problem as long as it succeeds one time.  I'll create another ticket to investigate / address this because I actually don't think it's related to this ticket change.

I think you're right - now that L2518 makes us return early if the task is unfinished, we don't run the risk of archiving an unfinished task and then getting stuck at the archive `Insert` step on the next attempt.  So it should be fine change this back to `Insert`

hadjri: For the task exec timeout job bug: https://jira.mongodb.org/browse/EVG-17676

model/task/task.go
ablack12: Yeah that sounds reasonable -- a lot of times as part of the debug statement we include the ticket for removing/tracking it  ("ticket": <ticket_num>) if that's helpful (since this also helps with checking on the debug statements)

ablack12 (approved):
I think this looks good to me but @Kimchelly can you also review since you helped review the initial PR?

model/lifecycle_test.go
```diff
@@ -195,6 +195,13 @@ func TestBuildSetPriority(t *testing.T) {
 }
 
 func TestBuildRestart(t *testing.T) {
+	defer func() {
+		assert.NoError(t, db.ClearCollections(build.Collection, task.Collection, task.OldCollection))
+	}()
+
+	// Running a multi-document transaction requires the collections to exist
+	// first before any documents can be inserted.
+	require.NoError(t, db.CreateCollections(build.Collection, task.Collection, task.OldCollection))
```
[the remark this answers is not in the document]
hadjri: True, removing the build collection from this line

model/task/db.go
```diff
@@ -65,6 +65,7 @@ var (
 	LatestParentExecutionKey       = bsonutil.MustHaveTag(Task{}, "LatestParentExecution")
 	OldTaskIdKey                   = bsonutil.MustHaveTag(Task{}, "OldTaskId")
 	ArchivedKey                    = bsonutil.MustHaveTag(Task{}, "Archived")
+	CanResetKey                    = bsonutil.MustHaveTag(Task{}, "CanReset")
```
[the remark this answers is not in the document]
hadjri: `CanReset` is not a proxy for having a completed status but rather is meant to be a protection mechanism against a situation where we succeed archiving but fail to reset the task, so we can still have a way to retry the "restart" operation to fix the state.  I'm realizing now that setting `CanReset` to true on `MarkEnd` and `MarkSystemFailed` defeats the benefit of having it - we should only set it directly after a successful archival. I've removed those instances.

Since it can now only ever get toggled true upon successful archival, it shouldn't be possible to have `CanReset` true for a task in an unfinished state.  Likewise the converse shouldn't be possible. `CanReset` should just be ephemerally set to true for a split second between archiving and resetting, and nothing else.

model/task/task.go
```diff
@@ -150,6 +150,9 @@ type Task struct {
 	// Set to true if the task should be considered for mainline github checks
 	IsGithubCheck bool `bson:"is_github_check,omitempty" json:"is_github_check,omitempty"`
 
+	// CanReset indicates if the task is in a valid state to be reset.
```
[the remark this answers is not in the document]
hadjri: Sure, updated the comment

model/lifecycle_test.go
```diff
@@ -195,6 +195,13 @@ func TestBuildSetPriority(t *testing.T) {
 }
 
 func TestBuildRestart(t *testing.T) {
+	defer func() {
+		assert.NoError(t, db.ClearCollections(build.Collection, task.Collection, task.OldCollection))
```
Kimchelly, line 199: Also remove `build.Collection` here.

model/task/task.go
```diff
@@ -2501,18 +2513,31 @@ func (t *Task) Insert() error {
...
 			return errors.Wrap(err, "inserting archived task into old tasks")
 		}
 		t.Aborted = false
 		err = UpdateOne(
-			bson.M{IdKey: t.Id},
+			bson.M{
+				IdKey:       t.Id,
+				StatusKey:   bson.M{"$in": evergreen.TaskCompletedStatuses},
+				CanResetKey: false,
+			},
```
Kimchelly, line 2540: For `UpdateOne`, Anser will [return a not found error if no docs were modified](https://github.com/mongodb/anser/blob/99dd61768f4a246b31e8e3ff8b2d0e51760f2194/db/wrapper.go#L207-L209), so we would have to return nil error for `adb.ResultsNotFound` (i.e. the task has already been archived) just in case the task has already been archived.

model/task/db.go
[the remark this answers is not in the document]
Kimchelly: Ah yeah, this usage of `CanReset` makes more sense now. Thanks!

model/task/task.go
```diff
@@ -2582,13 +2625,16 @@ func archiveAll(taskIds, execTaskIds, toRestartExecTaskIds []string, archivedTas
...
+			if err != nil && !db.IsDuplicateKey(err) {
 				return nil, errors.Wrap(err, "archiving tasks")
 			}
 		}
 		if len(taskIds) > 0 {
 			_, err = evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(sessCtx,
-				bson.M{IdKey: bson.M{"$in": taskIds}},
+				bson.M{
+					IdKey:     bson.M{"$in": taskIds},
+					StatusKey: bson.M{"$in": evergreen.TaskCompletedStatuses},
```
Kimchelly, line 2636: Should this also check if `CanReset` is false?

bynn (approved):
this looks really solid!

model/task/task.go
hadjri: Ah, thank you

model/task/task.go
hadjri: Yes... my bad

Kimchelly (approved):

Kimchelly:
LGTM mod fixing up all the tests.

PR #5892, opened 2022-08-23
EVG-17314 return commit queue position for patch routes

ablack12:
[EVG-17314](https://jira.mongodb.org/browse/EVG-17314)

### Description 
Modify BuildFromArgs to optionally add commit queue position.
Adding UI review to verify that not always getting project identifier isn't going to cause issues, in which case I can refactor.


rest/model/patch.go
```diff
@@ -97,8 +99,9 @@ func (p *APIParameter) ToService() patch.Parameter {
 	return res
 }
 
-// BuildFromService converts from service level structs to an APIPatch
-func (apiPatch *APIPatch) BuildFromService(p patch.Patch) error {
+// BuildFromService converts from service level structs to an APIPatch.
+// If withArgs is set, includes identifier and commit queue position from the DB, if applicable.
+func (apiPatch *APIPatch) BuildFromService(p patch.Patch, withArgs bool) error {
```
malikchaya2, line 104: Since withArgs is seems to be used as a "isThisACommitQueuePatch" bool, do we maybe want to rename it to something more closely related to that?

rest/model/patch.go
```diff
@@ -150,13 +153,24 @@ func (apiPatch *APIPatch) BuildFromService(p patch.Patch) error {
...
+		if p.IsCommitQueuePatch() {
+			cq, err := commitqueue.FindOneId(p.Project)
+			if err != nil {
+				return errors.Wrap(err, "error getting commit queue position")
+			}
+			apiPatch.CommitQueuePosition = utility.ToIntPtr(-1)
+			if cq != nil {
+				apiPatch.CommitQueuePosition = utility.ToIntPtr(cq.FindItem(p.Id.Hex()))
+			}
+		}
```
malikchaya2, line 173: It seems to me that all this is doing is setting another field on the patch. Can you please explain more about where this risk that you mentioned in the description is? 

> verify that not always getting project identifier isn't going to cause issues, in which case I can refactor

malikchaya2 (changes requested):

rest/model/patch.go
ablack12: Well it does also include the identifier thing. I think it's also not for if it's a commit queue but rather if we want to check that it is.

rest/model/patch.go
ablack12: This isn't referencing CommitQueuePosition, it's referencing identifier. As a part of this ticket I'm changing the behavior so if withArgs isn't specified, we just translate Patch -> APIPatch, and if withArgs is specified, then we get things from the DB to add to APIPatch (we're using this format slightly more for BuildFromService things).

Right now I'm setting withArgs to true for rest route functions only, i.e. for resolvers identifier won't be set. I'd like to verify that this isn't an issue for any existing UI patch usage, or if I should refactor this to consider identifier / commit queue position separately.

malikchaya2: I understand better now, thank you!

graphql/project_resolver.go
```diff
@@ -42,7 +42,7 @@ func (r *projectResolver) Patches(ctx context.Context, obj *restModel.APIProject
 	apiPatches := []*restModel.APIPatch{}
 	for _, p := range patches {
 		apiPatch := restModel.APIPatch{}
-		err = apiPatch.BuildFromService(p)
+		err = apiPatch.BuildFromService(p, false)
```
khelif96, line 45: So we do reference `projectIdentifier` in a few places. 
So it would need to be defined here. 
https://github.com/evergreen-ci/spruce/blob/6849093c411ac34c05b4e70d5b7e9b8f449a0863/src/gql/fragments/patchesPage.graphql#L6

graphql/user_resolver.go
```diff
@@ -29,7 +29,7 @@ func (r *userResolver) Patches(ctx context.Context, obj *restModel.APIDBUser, pa
 	apiPatches := []*restModel.APIPatch{}
 	for _, p := range patches {
 		apiPatch := restModel.APIPatch{}
-		if err = apiPatch.BuildFromService(p); err != nil {
+		if err = apiPatch.BuildFromService(p, false); err != nil {
```
khelif96, line 32: and here

graphql/project_resolver.go
malikchaya2, line 45: Right now the project patches query being fetched on spruce fetches the project identifier. Someone from UI may be able to say if it can be removed or not.

malikchaya2: [the query](https://github.com/evergreen-ci/spruce/blob/main/src/gql/queries/project-patches.graphql#L8)
[the fragment](https://github.com/evergreen-ci/spruce/blob/main/src/gql/fragments/patchesPage.graphql#L6)

khelif96 (changes requested):
and [here](https://github.com/evergreen-ci/spruce/blob/6849093c411ac34c05b4e70d5b7e9b8f449a0863/src/gql/queries/get-patch-configure.graphql#L18) this originates from the childPatches resolver.

It looks like we depend on `projectIdentifier` for a patch in a few places so this would likely break the way it is currently implemented.

khelif96:
The e2e tests should be a good indication of where we need these values. https://spruce.mongodb.com/task/spruce_ubuntu1604_e2e_test_patch_906333d53ed8ba72e94f209d70b6d650e5792621_6304eb7732f41722eed82d74_22_08_23_15_00_08/tests?execution=0&sortBy=STATUS&sortDir=ASC

graphql/user_resolver.go
malikchaya2, line 32: This also uses the same fragment on spruce as the query above, so spruce is currently fetching projectIdentifier here as well. 
[code](https://github.com/evergreen-ci/spruce/blob/main/src/gql/queries/user-patches.graphql#L7)

malikchaya2 (changes requested):

graphql/project_resolver.go
malikchaya2: haha jinx

ablack12:
@malikchaya2 and @khelif96 I've tweaked to have this take in individual parameters (I realized also that this is used from the CLI so certain commands are currently broken for patches with child patches, so moved that into args also). Tried to audit where things were needed but will see if the tests complain.

graphql/project_resolver.go
```diff
@@ -42,7 +42,7 @@ func (r *projectResolver) Patches(ctx context.Context, obj *restModel.APIProject
 	apiPatches := []*restModel.APIPatch{}
 	for _, p := range patches {
 		apiPatch := restModel.APIPatch{}
-		err = apiPatch.BuildFromService(p)
+		err = apiPatch.BuildFromService(p, &restModel.APIPatchArgs{IncludeProjectIdentifier: true})
```
malikchaya2, line 45: I think we would need to include the [childPatches](https://github.com/evergreen-ci/evergreen/blob/main/graphql/schema/types/patch.graphql#L64) in the two graphql resolvers (and tests) as well.

rest/model/patch.go
```diff
@@ -97,8 +99,47 @@ func (p *APIParameter) ToService() patch.Parameter {
...
+	}
+	apiPatch.buildBasePatch(p)
+	apiPatch.buildModuleChanges(p, projectIdentifier)
+
+	fmt.Println(len(apiPatch.ModuleCodeChanges))
```
bynn, line 135: do we want to print this?

bynn: Might be a dumb question but why are we using println if we do?

ablack12: no good question, this is outdated debugging!

graphql/project_resolver.go
khelif96: Yup thats correct! We use it on the version page, and on the configure page but we don't use it on the general "My Patches" page.

For ProjectIdentifier i think a hybrid solution would work best. Like we do for the taskResolver projectIdentifier.
https://github.com/evergreen-ci/evergreen/blob/34ccd3df43c009ba56a7fcf657a6c2b8516f0716/graphql/task_resolver.go#L455-L458

A solution where we would have a fallback to the  other source would be ideal. 

Depending on how the patch model is built the projectIdentifier could be included from the `BuildFromService` but if not having the fallback to its own resolver would ensure that we never accidentally omit this field without introducing the risk of doing an unnecessary additional query.

khelif96 (approved):
LGTM mod the comment about how to handle `projectIdentifier`

bynn (approved):

graphql/project_resolver.go
malikchaya2: What did we end up deciding with regards to child patches? It's currently being fetched on the main patch query so I think we need to add it in somehow? The ideal set up would be to put it in its own resolver, but at the very least we need it in the main query (and then maybe move it as an optimization in a separate ticket).

ablack12: @malikchaya2 the actual FindPatchById that is used by the Patch resolver _does_ fetch it since it uses FindPatchById which loads everything so your request is already handled. https://github.com/evergreen-ci/evergreen/blob/25f010ae0c90c76ddcb40b036735a8ee9fabcb16/rest/data/patch.go#L82

if we do want a separate resolver for it, that'll be more complicated than I know how to do since the child patches function modifies many variables, not just one, so UI could take that optimization ticket if it seems worth doing. cc @khelif96

khelif96 (approved):

graphql/project_resolver.go
malikchaya2: Looking at the [project patches](https://github.com/evergreen-ci/spruce/blob/main/src/gql/queries/project-patches.graphql#L1), the child patches are actually not being fetched there, so that makes sense to me.

malikchaya2 (approved):

graphql/project_resolver.go
ablack12: Ah that was also discussed above 👍 
> We use it on the version page, and on the configure page but we don't use it on the general "My Patches" page.

PR #5895, opened 2022-08-24
EVG-16902: clean up stranded container secrets

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16902

### Description 
* Add a job to clean up container secrets that were created but never stored in Evergreen's DB.
* Fix a flaky pod definition cleanup test.

### Testing
Added unit tests. I also spot checked that running the job cleaned up stranded secrets in Secrets Manager.

### DIFF
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -378,13 +378,22 @@ func GetFilteredResourceIDs(ctx context.Context, c cocoa.TagClient, resources []
 
 	var allIDs []string
 	remaining := limit
-	for ids, nextToken, err := getResourcesPage(ctx, c, resourceFilters, tagFilters, nil, limit); ; ids, nextToken, err = getResourcesPage(ctx, c, resourceFilters, tagFilters, nextToken, remaining) {
+	var nextToken *string
+	for {
+		var (
+			ids []string
+			err error
+		)
+		ids, nextToken, err = getResourcesPage(ctx, c, resourceFilters, tagFilters, nextToken, limit)
 		if err != nil {
 			return nil, errors.Wrap(err, "getting resources matching filters")
 		}
 		allIDs = append(allIDs, ids...)
 		remaining = remaining - len(ids)
 		if remaining <= 0 {
+			break
+		}
+		if len(ids) == 0 {
 			break
 		}
 		if nextToken == nil {
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -1,10 +1,16 @@
 package cloud
 
 import (
+	"context"
+	"math"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
 	"github.com/evergreen-ci/cocoa"
 	"github.com/evergreen-ci/cocoa/awsutil"
 	"github.com/evergreen-ci/cocoa/ecs"
 	"github.com/evergreen-ci/cocoa/secret"
+	"github.com/evergreen-ci/cocoa/tag"
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/pod"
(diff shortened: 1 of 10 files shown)


Kimchelly:
evergreen retry

cloud/pod_util.go
```diff
@@ -335,3 +356,102 @@ func exportPodEnvVars(smConf evergreen.SecretsManagerConfig, opts pod.TaskContai
 
 	return allEnvVars
 }
+
+// GetFilteredResourceIDs gets at most n resources that match the given resource
```
hadjri, line 360: Can we refer to n as limit since that is how it's described in the function definition?

cloud/pod_util.go
```diff
@@ -335,3 +356,102 @@ func exportPodEnvVars(smConf evergreen.SecretsManagerConfig, opts pod.TaskContai
...
+		tagFilters = append(tagFilters, &resourcegroupstaggingapi.TagFilter{
+			Key:    aws.String(key),
+			Values: utility.ToStringPtrSlice(vals),
+		})
+	}
+	resourceFilters := utility.ToStringPtrSlice(resources)
+
+	var allIDs []string
+	remaining := limit
+	for ids, nextToken, err := getResourcesPage(ctx, c, resourceFilters, tagFilters, nil, limit); ; ids, nextToken, err = getResourcesPage(ctx, c, resourceFilters, tagFilters, nextToken, remaining) {
```
hadjri, line 381: Can we break this up into two lines for readability?

units/container_secret_cleanup.go
```diff
@@ -0,0 +1,119 @@
...
+
+	j.AddError(errors.Wrap(catcher.Resolve(), "deleting secrets"))
+}
+
+func (j *containerSecretCleanupJob) populate() error {
+	if j.env == nil {
+		j.env = evergreen.GetEnvironment()
+	}
+
+	if j.smClient == nil {
```
hadjri, line 94: Is there a reason the secrets manager and tag clients are getting initialized independently of the vault (unlike [here](https://github.com/evergreen-ci/evergreen/blob/main/units/pod_allocator.go#L209)) in the pod allocator job?

cloud/pod_util.go
```diff
@@ -335,3 +356,102 @@ func exportPodEnvVars(smConf evergreen.SecretsManagerConfig, opts pod.TaskContai
...
+	}
+
+	return allIDs, nil
+}
+
+func getResourcesPage(ctx context.Context, c cocoa.TagClient, resourceFilters []*string, tagFilters []*resourcegroupstaggingapi.TagFilter, nextToken *string, limit int) ([]string, *string, error) {
+	var ids []string
+
+	resp, err := c.GetResources(ctx, &resourcegroupstaggingapi.GetResourcesInput{
+		PaginationToken:     nextToken,
```
hadjri, line 402: There is a `ResourcesPerPage` property you can set here from 1 to 100, which might be relevant if limit is < 100, and might reduce the amount of calls we need to make. 

The docs do say that "A query can return fewer than this maximum, even if there are more results still to return. You should always check the PaginationToken response value to see if there are more results", so it isn't an airtight thing.  Just letting you know it's an option in case you feel like it'd be helpful.

cloud/pod_util.go
Kimchelly: Hmm, it's hard to break up into two separate lines because the initialization condition is different from the repeated iteration condition. I tried to modify it to be more readable.

cloud/pod_util.go
Kimchelly: Hmm I don't have that strong of feelings one way or the other on using vs. not using the page size. Ideally, we just want it to return as many as possible per page (which is what it does by default), so the number of calls needed is the same whether we set page size or not. We don't care too deeply about AWS returning too many results in a single page either, since AWS is supposedly structured to be efficient at getting tagged resources.

units/container_secret_cleanup.go
Kimchelly: Hmm yeah, would be good to initialize them the same way. I'm going to change the pod allocator one to follow this because it's equivalent and slightly less if logic to think through.

cloud/pod_util.go
hadjri: Thanks!

cloud/pod_util.go
hadjri: Yeah, makes sense. Looking closer it doesn't really seem like there's much reason to use it especially if there isn't a guarantee returning less than the limit means we are done getting results.

hadjri (approved):

PR #5896, opened 2022-08-24
EVG-17582: fix flaky Windows tests

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17582

### Description 
In Windows, deleting a directory will fail if there's a process that's actively using it or there's an unclosed file handle (which is a resource leak). I had to tweak the tests where this was relevant.

* One failing agent test is checking the expected behavior of the agent that if a command times out, the agent should continue on without waiting for the command to finish. Since the command exits asynchronously, there will be a race between the asynchronous command exiting and the main testing thread attempting to clean up that command's directory when the command may not be exited yet. I just reverted this to its previous temporary directory usage since there's no easy way to wait until the command finishes. As to why removing the directory succeeds if it's in the teardown block, I don't know, but it's probably due to something internal to the testify suite implementation.
* Another failing agent test was failing to close a file handle.
* The S3 test was failing because it wasn't closing file handles.

### Testing 
Re-ran unit tests.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -3,6 +3,9 @@ package agent
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


Kimchelly:
evergreen retry

malikchaya2 (approved):

PR #5897, opened 2022-08-24
EVG-17510 Allow emptying Github PR aliases when PR testing is enabled

hadjri:
[EVG-17510](https://jira.mongodb.org/browse/EVG-17510)

### Description 
The modify project route returns an error when PR testing is enabled and the PR patch definition list is empty.  Since PR patch definitions can now be defined in YAML, this change ignores that error if "version control" is enabled on the project page.

Also fixed a few things in project validation including a missing nil check for build baron plugins and automatically infer alias names to YAML defined aliases so that things like `alias: github_pr` aren't necessary to write in YAML.
### Testing 
Updated existing unit tests. 

Enabled version control on the sandbox project in staging and emptied the PR patch definitions, and replaced them with YAML definitions. Staged a few PRs using this YAML to confirm it was working.

[Example configuration](https://github.com/evergreen-ci/commit-queue-sandbox/blob/df8f7fc3529a11809802164f1d3716023a2ef6a9/evergreen.yml#L205), and [subsequent created patch](https://spruce-staging.corp.mongodb.com/version/6306afc497b1d31c2bb47ccc/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -54,6 +54,11 @@ const (
 // "linux"; and to run all tasks beginning with the string "compile" to run on all
 // variants beginning with the string "ubuntu1604".
 
+// For regular patch aliases, the Alias field is required to be a custom string defined by the user.
+// For all other special alias types (commit queue, github PR, etc) the Alias field must match its associated
+// constant in globals.go, i.e. evergreen.GithubPRAlias. For aliases defined within a project's config YAML
+// the Alias field for non-patch aliases is not-required since it will be inferred and assigned at runtime.
+
 // Git tags use a special alias "__git_tag" and create a new version for the matching
 // variants/tasks, assuming the tag matches the defined git_tag regex.
 // In this way, users can define different behavior for different kind of tags.
--- a/service/project.go
+++ b/service/project.go
@@ -461,7 +461,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 				uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "can't check if GitHub aliases are set"))
 				return
 			}
-			if !aliasesDefined {
+			if !aliasesDefined && !responseRef.VersionControlEnabled {
 				uis.LoggedError(w, r, http.StatusBadRequest, errors.New("cannot enable PR testing without patch definitions"))
 				return
 			}
--- a/service/project.go
+++ b/service/project.go
@@ -484,7 +484,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 				uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "can't check if GitHub check aliases are set"))
 				return
 			}
-			if !aliasesDefined {
+			if !aliasesDefined && !responseRef.VersionControlEnabled {
 				uis.LoggedError(w, r, http.StatusBadRequest, errors.New("cannot enable Github checks without definitions"))
 				return
 			}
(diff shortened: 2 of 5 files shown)


service/project.go
```diff
@@ -461,7 +461,7 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 				uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "can't check if GitHub aliases are set"))
 				return
 			}
-			if !aliasesDefined {
+			if !aliasesDefined && !responseRef.VersionControlEnabled {
```
malikchaya2, line 464: Aren't we still running the risk of there being no aliases defined? While they can defined in YAML, how do we know that it is?

hadjri: I don't think it would be necessary to add any more checks because a PR patch will already fail validation if there is no PR alias defined: https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go#L910

The `FindAliasInProjectRepoOrPatchedConfig ` checks both the project ref and the project YAML for alias definitions.

malikchaya2: Do we want to allow enabling PR testing though without doing both checks?

hadjri: Yes, because since PRs are their own patches they will need to be validated at the time they are submitted.  If project page PR aliases are empty then Evergreen will next check if PR aliases are defined in the YAML for the patch. If not, the validation error I linked above will hit and the user can just fix it easily by adding PR aliases to their config.

It wouldn't be feasible to check anything else here since we don't know what PRs look like until they are created in the future.  The patch validation will need to happen separately on an individual basis.

validator/project_validator.go
```diff
@@ -421,6 +421,18 @@ func validateProjectConfigPeriodicBuilds(pc *model.ProjectConfig) ValidationErro
...
+	}
+	for i := range pc.GitHubChecksAliases {
+		pc.GitHubChecksAliases[i].Alias = evergreen.GithubChecksAlias
+	}
+	for i := range pc.CommitQueueAliases {
+		pc.CommitQueueAliases[i].Alias = evergreen.CommitQueueAlias
+	}
+	for i := range pc.GitHubPRAliases {
+		pc.GitHubPRAliases[i].Alias = evergreen.GithubPRAlias
+	}
```
malikchaya2, line 435: Can you please explain what this does?

validator/project_validator.go
```diff
@@ -466,6 +478,9 @@ func validateProjectConfigPlugins(pc *model.ProjectConfig) ValidationErrors {
 	if annotationSettings != nil {
 		webhook = &annotationSettings.FileTicketWebhook
 	}
+	if pc.BuildBaronSettings == nil {
+		return errs
```
malikchaya2, line 482: We should probably append an error to errs explaining what's wrong.

validator/project_validator.go
malikchaya2, line 435: Can you please explain what this does? Why was it not needed before, but needed now?

malikchaya2: Can you please also expand on the [explanation here](https://github.com/evergreen-ci/evergreen/blob/main/model/project_aliases.go#L59) so that it's easier for future engineers to grok?

validator/project_validator.go
hadjri: `errs` at this stage will be empty so this is actually meant to be a no-op. Added a comment

validator/project_validator.go
hadjri: Alias validation expectes the `Alias` field to be [populated](https://github.com/evergreen-ci/evergreen/blob/main/model/project_aliases.go#L495). When defining aliases in YAML it shouldn't be necessary to define stuff like `alias: _github` manually since we can instead infer the alias name depending on what alias list it is in. I.e. for all `GitHubPRAliases` we'll add the relevant Github PR alias to it.

hadjri: Added some more documentation

validator/project_validator.go
malikchaya2: Can we return nil instead of errs if that's the case?

model/project_aliases.go
```diff
@@ -54,6 +54,11 @@ const (
 // "linux"; and to run all tasks beginning with the string "compile" to run on all
 // variants beginning with the string "ubuntu1604".
 
+// For regular patch aliases, the Alias field is required to be a custom string defined by the user.
```
malikchaya2, line 57: thank you!

validator/project_validator.go
hadjri: Sure

malikchaya2 (approved):

PR #5899, opened 2022-08-26
EVG-16229 Add downstream_expansions.set and timeout.update commands to smoke test

malikchaya2:
[EVG-16229](https://jira.mongodb.org/browse/EVG-16229)

### Description 
This adds the project commands  downstream_expansions.set and timeout.update to the smoke test. The first commit just applies text formatting to the file, so I would suggest just looking at the second command when reviewing so that it's easier to parse. 

### Testing 
[Tested the changes ](https://spruce-staging.corp.mongodb.com/version/smoke_9173e2fbe49cbe5820a7650ce289f322a62a63a1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)on the smoke project branch on staging


scripts/agent.yml
```diff
@@ -84,7 +83,7 @@ tasks:
           script: |
             set -o verbose
             set -o errexit
-            echo "hi"
+            echo "hello "
```
Kimchelly, line 86: Does this need to change?

scripts/agent.yml
```diff
@@ -192,12 +195,25 @@ tasks:
           aws_key: ${aws_key}
           aws_secret: ${aws_secret}
           s3_copy_files:
-              - {'source': {'path': 'evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}', 'bucket': 'mciuploads'},
-                'destination': {'path': 'evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}-copy', 'bucket': 'mciuploads'}}
+            - {
+                "source":
+                  {
```
Kimchelly, line 200: If we're going to unroll these lines, can we just use YAML syntax here instead of JSON?

scripts/agent.yml
```diff
@@ -159,6 +158,10 @@ tasks:
         params:
           key: "test"
           destination: "test_num"
+      - command: downstream_expansions.set
```
Kimchelly, line 161: It seems like `downstream_expansions.set` requires a child patch in order to verify that this command actually set the expansions. Do we think this is enough coverage of its expected behavior?

scripts/agent.yml
```diff
@@ -159,6 +158,10 @@ tasks:
...
+        params:
+          file: ../testdata/downstream_expansions.yml
```
Kimchelly, line 163: Can we put the downstream expansions file in the same directory as this file?

Kimchelly (changes requested):

scripts/agent.yml
malikchaya2: It doesn't anymore, I can remove it.

scripts/agent.yml
malikchaya2: Why would that be helpful?
I don't particularly want to unroll it, the editor does it automatically.

scripts/agent.yml
malikchaya2: We did role out this command because the child patch needs this to be set so that it can use it, but all the command really does is set the parameters on the triggers of the parent patch. So this does test the entire command (albeit it's not that useful unless you have a child patch that accesses it). Additionally, setting up a child patch in the smoke test would be too involved to be worth the effort, in my opinion.

scripts/agent.yml
malikchaya2: done

scripts/agent.yml
malikchaya2: removed

scripts/agent.yml
malikchaya2: I'm actually not totally right about that because it uses parameterized builds to write to the child patch's patch.Parameter field. However, I don't think that that is the most risky part of the command, an I still don't think it's worth doing.

Kimchelly: SGTM

scripts/agent.yml
malikchaya2: that makes a lot of sense, thanks! 
done

scripts/agent.yml
```diff
@@ -195,18 +195,12 @@ tasks:
...
-                "destination":
-                  {
-                    "path": "evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}-copy",
-                    "bucket": "mciuploads",
-                  },
-              }
+            - source:
+              path: "evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}"
+              bucket: mciuploads
+            - destination:
```
Kimchelly, line 201: `source` and `destination` have to be under the same YAML list item since `s3_copy_files` is a slice of source/destination parameters.

scripts/agent.yml
```diff
@@ -195,18 +195,12 @@ tasks:
...
-                    "bucket": "mciuploads",
-                  },
-                "destination":
-                  {
-                    "path": "evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}-copy",
-                    "bucket": "mciuploads",
-                  },
-              }
+            - source:
+              path: "evergreen/smoke/${build_id}-${build_variant}/evergreen-${task_name}-${revision}"
```
Kimchelly, line 199: `source` is a YAML dictionary, so `path` and `bucket` have to be indented one more. Same thing with `destination`'s `path` and `bucket` parameters.

scripts/agent.yml
Kimchelly: Looks good, I think there's just a couple YAML syntax errors.

malikchaya2: thanks! fixed

scripts/agent.yml
malikchaya2: done

scripts/agent.yml
malikchaya2: done

Kimchelly (approved):

malikchaya2:
evergreen retry

malikchaya2:
patch because pr testing is not working. 
https://spruce.mongodb.com/version/630deabf61837d20ee1373df/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

malikchaya2:
The tests that are failing pass locally and also fail on main.

PR #5904, opened 2022-08-29
EVG-15530 Remove json.send

malikchaya2:
[EVG-15530](https://jira.mongodb.org/browse/EVG-15530)

### Description 
Json.send needs to be removed because we want users to relying on the perf.send command to generate new performance data.


agent/command/taskdata_send.go
```diff
@@ -1,75 +0,0 @@
...
-		}
-
-		jsonData := map[string]interface{}{}
-		err = utility.ReadJSON(jsonFile, &jsonData)
-		if err != nil {
-			errChan <- errors.Wrap(err, "File contained invalid json")
-			return
-		}
-
-		errChan <- errors.Wrapf(comm.PostJSONData(ctx, td, c.DataName, jsonData),
```
Kimchelly, line 61: Seems like `PostJSONData`, [the legacy REST v1 route](https://github.com/evergreen-ci/evergreen/blob/976819115af360a38710e7ba65ec9dedd21ac7dd/service/api.go#L814), and [the newer REST v2 route](https://github.com/evergreen-ci/evergreen/blob/976819115af360a38710e7ba65ec9dedd21ac7dd/rest/route/service.go#L94) could be removed since they're now unused.

Kimchelly (commented):
I did a bit of digging because I could have sworn that old projects were still using this and [Splunk](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen%20%2Fjson%2Fdata&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1661832000&latest=1661918400&sid=1661874364.9903763) confirms that the `json.send` command still being used, such as in [the branch for mongodb v4.4](https://github.com/mongodb/mongo/blob/bc17cf2c788c5dda2801a090ea79da5ff7d5fac9/etc/evergreen.yml#L1213). I also think it's unlikely that the server will backport their more recent perf.send migration to older branches so I don't think it's feasible to completely remove `json.send` until these projects are no longer in active use.

malikchaya2:
> I did a bit of digging because I could have sworn that old projects were still using this and [Splunk](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen%20%2Fjson%2Fdata&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1661832000&latest=1661918400&sid=1661874364.9903763) confirms that the `json.send` command still being used, such as in [the branch for mongodb v4.4](https://github.com/mongodb/mongo/blob/bc17cf2c788c5dda2801a090ea79da5ff7d5fac9/etc/evergreen.yml#L1213). I also think it's unlikely that the server will backport their more recent perf.send migration to older branches so I don't think it's feasible to completely remove `json.send` until these projects are no longer in active use.

That makes sense. I'll close this PR and put the ticket back on the backlog so that we can check again if we are ready in 6 months or so.

PR #5906, opened 2022-08-29
fix(EVG-17606): assume role session name

jasonwalsh:
The [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) API has the following pattern constraint:

```
[\w+=,.@-]*
```

`time.Now().String()` returns a string like the following:

```
2009-11-10 23:00:00 +0000 UTC m=+0.000000001
```

Which does not match the pattern constraint.

Using `time.Now().Unix()` matches the pattern constraint and should mitigate the `ValidationError` returned by the AWS API.

See [EVG-17606](https://jira.mongodb.org/browse/EVG-17606)

### DIFF
--- a/agent/command/assume_ec2_role.go
+++ b/agent/command/assume_ec2_role.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


Kimchelly (approved):

Kimchelly:
Since this changes the agent's command execution logic, you have to also bump [the agent version](https://github.com/evergreen-ci/evergreen/blob/976819115af360a38710e7ba65ec9dedd21ac7dd/config.go#L39) to trigger new agents to roll out.

PR #5910, opened 2022-09-01
EVG-17272 add configurable clone depth

ablack12:
[EVG-17272](https://jira.mongodb.org/browse/EVG-17272)

### Description 
Allow clone depth to be configurable, as opposed to 0 or full. Will add to documentation after approval.

### Testing 
Added a unit test and [tested on staging](https://evergreen-staging.corp.mongodb.com/task_log_raw/evg_ubuntu1604_verify_agent_version_update_patch_82568ad5bc48fd1d677094b74938aa0c0f6c2f2f_6310bea0b237366ddb177aa5_22_09_01_14_16_13/0?type=T#L6) that the value passed into self-tests is used.


agent/command/git.go
```diff
@@ -48,7 +49,9 @@ type gitFetchProject struct {
 
 	Token string `plugin:"expand" mapstructure:"token"`
 
+	// ShallowClone sets CloneDepth to 100, and is kept for backwards compatibility.
 	ShallowClone bool `mapstructure:"shallow_clone"`
+	CloneDepth   int  `mapstructure:"clone_depth"`
```
hadjri, line 54: If `ShallowClone` is set to true, and `CloneDepth` is also set, it seems like `ShallowClone` will have no effect. It might be worth it to print out a warning if this happens saying that it's not necessary to set shallow clone. I'll leave this up to you though.

hadjri (approved):

agent/command/git.go
ablack12: Good idea 👍

PR #5911, opened 2022-09-01
EVG-12490: Adjust commit queue position to be one-indexed on patch page

minnakt:
[EVG-12490](https://jira.mongodb.org/browse/EVG-12490)

### Description 
On the commit-queue page ([example](https://spruce.mongodb.com/commit-queue/mongodb-mongo-master)), the list of patches is 1-indexed. However on the patches page, the commit queue position is 0-indexed. If a patch has a commit queue position of 0, that means it's actually the first patch on the commit queue.

To be consistent with the commit queue list, I adjusted the commit queue position to be 1-indexed in `patch_resolver.go`. This is also consistent with how we report task queue positions, which are 1-indexed.

### Note
This is only a UI facing change. Internally we rely on the 0-indexed commit queue in a few different places so I did not make any changes around that.

I'm not sure if this means that I should also update the commit queue position for the patch route once [EVG-17314](https://jira.mongodb.org/browse/EVG-17314) is complete. It could be confusing to see different commit queue positions between the UI and the API. 

### Testing 
- On staging Sandbox
- Adjusted GraphQL test

### DIFF
--- a/graphql/patch_resolver.go
+++ b/graphql/patch_resolver.go
@@ -59,10 +59,14 @@ func (r *patchResolver) CommitQueuePosition(ctx context.Context, obj *restModel.
 	if *obj.Alias == evergreen.CommitQueueAlias {
 		cq, err := commitqueue.FindOneId(*obj.ProjectId)
 		if err != nil {
-			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting commit queue position for patch %s: %s", *obj.Id, err.Error()))
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("getting commit queue position for patch '%s': %s", *obj.Id, err.Error()))
 		}
 		if cq != nil {
 			position := cq.FindItem(*obj.Id)
+			// Adjust to be one-indexed if found on commit queue.
+			if position >= 0 {
+				position += 1
+			}
 			commitQueuePosition = &position
 		}
 	}
--- a/graphql/tests/patch/results.json
+++ b/graphql/tests/patch/results.json
@@ -134,7 +134,7 @@
       "result": {
         "data": {
           "patch": {
-            "commitQueuePosition": 0
+            "commitQueuePosition": 1
           }
         }
       }


graphql/patch_resolver.go
```diff
@@ -59,10 +59,14 @@ func (r *patchResolver) CommitQueuePosition(ctx context.Context, obj *restModel.
...
 		cq, err := commitqueue.FindOneId(*obj.ProjectId)
 		if err != nil {
-			return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting commit queue position for patch %s: %s", *obj.Id, err.Error()))
+			return nil, InternalServerError.Send(ctx, fmt.Sprintf("getting commit queue position for patch '%s': %s", *obj.Id, err.Error()))
 		}
 		if cq != nil {
 			position := cq.FindItem(*obj.Id)
+			// Adjust to be one-indexed if found on commit queue.
+			if position >= 0 {
+				position += 1
```
khelif96, line 68: Would it make sense to just make this change on the front end instead? 
On the commit queue page we increment the `commitQueuePosition` by 1 before rendering it i think we should be consistent with how we handle/return this value.
https://github.com/evergreen-ci/spruce/blob/71b8a6253f06da15d08562866aae25c5854f9608/src/pages/CommitQueue.tsx#L60

khelif96 (changes requested):

graphql/patch_resolver.go
minnakt: I don't necessarily think it would be more consistent because the code linked technically doesn't increment the `commitQueuePosition` by 1 but instead renders a list index and increments that index by 1.

I think I'm kind of being swayed by the fact that both the task resolver and the legacy tasks route use a single function [`FindMinimumQueuePositionForTask`](https://github.com/evergreen-ci/evergreen/blob/main/model/task_queue.go#L539) to get the task queue position for a task. Maybe it could be good if I made a function like `FindCommitQueuePosition` that could be reused by the patch resolver ([here](https://github.com/evergreen-ci/evergreen/blob/main/graphql/patch_resolver.go#L60-L68)) and API ([here](https://github.com/ablack12/evergreen/blob/EVG-17314/rest/model/patch.go#L122-L129), which is Annie's code for [EVG-17314](https://jira.mongodb.org/browse/EVG-17314))? I guess it would only be useful if we want them to match 😥 but I think that could make sense because it might be confusing for users to see two different values... 

sorry for rambling 🧍 I'm okay with doing it on the UI side if you feel strongly about it!

ablack12:
Preference to make everything 0 indexed instead of 1-indexed if that isn't too great a change; this _is_ a product for engineers so zero-indexing is well understood and would provide more consistency with routes and with our own internals.

minnakt:
> Preference to make everything 0 indexed instead of 1-indexed if that isn't too great a change; this _is_ a product for engineers so zero-indexing is well understood and would provide more consistency with routes and with our own internals.

Okay, in that case it's just a minimal change in how we render the list on the frontend so I will close this PR.

PR #5913, opened 2022-09-06
EVG-17137 Refactor container sizes to be GQL friendly

hadjri:
[EVG-17137](https://jira.mongodb.org/browse/EVG-17137)

### Description 
Added new container entities to our GraphQL schema. Since GraphQL is a strongly-typed language, and does not provide any kind of map type out of the box, it'd be a little complex to integrate the resolvers that read/write project settings data (and make Spruce frontend changes) with the current setup of having `ContainerSizes` being a map of string names to resource definitions. 

I've refactored `ContainerSizes` to `ContainerResourceDefinitions` which is now a list rather than a map, and I've updated the bson tags accordingly so we don't get any DB collision.
### Testing 
Added graphql test cases for the new schema.


model/project_configs.go
```diff
@@ -25,18 +25,18 @@ type ProjectConfigFields struct {
...
+	GitHubPRAliases          []ProjectAlias                 `yaml:"github_pr_aliases,omitempty" bson:"github_pr_aliases,omitempty"`
+	GitTagAliases            []ProjectAlias                 `yaml:"git_tag_aliases,omitempty" bson:"git_tag_aliases,omitempty"`
+	GitHubChecksAliases      []ProjectAlias                 `yaml:"github_checks_aliases,omitempty" bson:"github_checks_aliases,omitempty"`
+	PatchAliases             []ProjectAlias                 `yaml:"patch_aliases,omitempty" bson:"patch_aliases,omitempty"`
+	WorkstationConfig        *WorkstationConfig             `yaml:"workstation_config,omitempty" bson:"workstation_config,omitempty"`
+	TaskSync                 *TaskSyncOptions               `yaml:"task_sync,omitempty" bson:"task_sync,omitempty"`
+	GithubTriggerAliases     []string                       `yaml:"github_trigger_aliases,omitempty" bson:"github_trigger_aliases,omitempty"`
+	PeriodicBuilds           []PeriodicBuildDefinition      `yaml:"periodic_builds,omitempty" bson:"periodic_builds,omitempty"`
+	ContainerSizeDefinitions []ContainerResources           `yaml:"container_size_definitions,omitempty" bson:"container_size_definitions,omitempty"`
```
Kimchelly, line 39: I don't have opinions one way or the other about this field name, but was there a particular reason to rename this? It's not being used by end users yet, so we can keep the old name if you prefer it.

public/static/js/projects.js
```diff
@@ -902,8 +902,8 @@ mciModule.controller(
     };
 
       $scope.removeContainerSize = function (name) {
-          if ($scope.settingsFormData.container_sizes[name]) {
-              delete $scope.settingsFormData.container_sizes[name]
+          if ($scope.settingsFormData.container_size_definitions[name]) {
```
Kimchelly, line 905: I think I mentioned this in the Jira ticket, but since we don't plan on supporting container settings in the legacy project settings UI, we can remove the legacy container UI logic entirely.

rest/model/project.go
```diff
@@ -531,12 +534,12 @@ func (p *APIProjectRef) ToService() (*model.ProjectRef, error) {
 		projectRef.PatchTriggerAliases = patchTriggers
 	}
 
-	if p.ContainerSizes != nil {
-		containerSizes := map[string]model.ContainerResources{}
-		for name, size := range p.ContainerSizes {
-			containerSizes[name] = size.ToService()
+	if p.ContainerSizeDefinitions != nil {
```
Kimchelly, line 537: Since this field is `omitempty` in both the DB and the API models, it cannot be an empty slice, so this could be simplified down to remove the if statement:
```go
for _, size := range p.ContainerSizeDefinitions {
    projectRef.ContainerSizeDefinitions = append(projectRef.ContainerSizeDefinitions, size.ToService())
}
```

rest/model/project.go
```diff
@@ -644,14 +647,14 @@ func (p *APIProjectRef) BuildFromService(projectRef model.ProjectRef) error {
 		p.PatchTriggerAliases = patchTriggers
 	}
 
-	if projectRef.ContainerSizes != nil {
-		containerSizes := map[string]APIContainerResources{}
-		for name, resources := range projectRef.ContainerSizes {
+	if projectRef.ContainerSizeDefinitions != nil {
```
Kimchelly, line 650: Same comment as above about `omitempty` slices.

rest/model/project.go
```diff
@@ -644,14 +647,14 @@ func (p *APIProjectRef) BuildFromService(projectRef model.ProjectRef) error {
...
+		containerSizes := []APIContainerResources{}
+		for _, resources := range projectRef.ContainerSizeDefinitions {
```
Kimchelly, line 652: nit: should this be called `size` instead of `resources` for consistency?

Kimchelly (changes requested):
I only reviewed the non-GQL parts of the PR.

model/project_ref.go
```diff
@@ -1950,6 +1953,12 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			bson.M{
 				"$set": bson.M{projectRefPeriodicBuildsKey: p.PeriodicBuilds},
 			})
+	case ProjectPageContainerSection:
+		err = db.Update(coll,
```
Kimchelly, line 1957: Does this need to validate the container sizes before updating them?

model/lifecycle.go
```diff
@@ -1364,13 +1364,19 @@ func getContainerOptions(project *Project, pRef *ProjectRef, container string) (
...
+				break
+			}
+		}
+		if containerSize == nil {
 			return nil, errors.Errorf("container size '%s' not found", c.Size)
 		}
 
-		opts.CPU = size.CPU
-		opts.MemoryMB = size.MemoryMB
+		opts.CPU = containerSize.CPU
```
malikchaya2, line 1378: should these be set in the if statement on like 1371?

model/lifecycle_test.go
```diff
@@ -732,15 +732,15 @@ func TestCreateBuildFromVersion(t *testing.T) {
...
-		smallContainerSize := ContainerResources{
-			MemoryMB: 128,
-			CPU:      256,
-		}
 		pref := &ProjectRef{
 			Id:         "projectId",
 			Identifier: "projectName",
-			ContainerSizes: map[string]ContainerResources{
-				"small": smallContainerSize,
+			ContainerSizeDefinitions: []ContainerResources{
```
malikchaya2, line 738: If there may be multiple, can we please add more to the array?

graphql/tests/projectSettings/data.json
```diff
@@ -63,6 +63,13 @@
           "field" : "sample_field"
         }]
       },
+      "container_size_definitions": [
+        {
```
malikchaya2, line 67: Can we please test it with an array that contains a few?

malikchaya2 (changes requested):

model/project_configs.go
hadjri: It was set on a few projects in the staging DB so I just renamed it to avoid any trouble

public/static/js/projects.js
hadjri: Sounds good, removing from JS/HTML and the `modifyProject` function

rest/model/project.go
hadjri: Good catch, done

model/project_ref.go
hadjri: There should be form validation on Spruce's side to ensure that memory and CPU are properly defined, but we can add validation here as well for additional assuredness.

hadjri: I'll add it for some other sections too while I'm at it since they don't have validation here either.

graphql/tests/projectSettings/data.json
```diff
@@ -63,6 +63,13 @@
           "field" : "sample_field"
         }]
       },
+      "container_size_definitions": [
```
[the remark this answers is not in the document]
hadjri: Sure, done

model/lifecycle.go
```diff
@@ -1364,13 +1364,19 @@ func getContainerOptions(project *Project, pRef *ProjectRef, container string) (
 			return &opts, nil
 		}
 
-		size, ok := pRef.ContainerSizes[c.Size]
-		if !ok {
+		var containerSize *ContainerResources
+		for _, size := range pRef.ContainerSizeDefinitions {
+			if size.Name == c.Size {
```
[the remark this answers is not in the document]
hadjri: Users will be able to define many container size definitions on their project.  When configuring the container itself, they can refer to one of these existing size definitions by putting its name as its `Size` field. So a container configuration can only have one size, but the list of container sizes (like aliases) to choose from can be arbitrarily long

model/lifecycle_test.go
hadjri: Sure

model/lifecycle.go
hadjri: These two lines need to be performed after we ensure the container size that we retrieve is not nil, so that's why I chose to put these at the end of the function.

graphql/tests/projectSettings/data.json
hadjri: Sure thing, added one more size to project and repo data files

graphql/schema/types/project_settings.graphql
```diff
@@ -70,6 +71,7 @@ input ProjectInput {
   triggers: [TriggerAliasInput!]
   versionControlEnabled: Boolean
   workstationConfig: WorkstationConfigInput
+  containerSizeDefinitions: [ContainerResourcesInput!]
```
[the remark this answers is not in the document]
hadjri: Ah yes, they are defined alongside other project types in `schema/types/project.graphql`.

model/project_ref.go
```diff
@@ -1950,6 +1953,21 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
 			})
+	case ProjectPageContainerSection:
+		catcher := grip.NewSimpleCatcher()
+		for _, size := range p.ContainerSizeDefinitions {
+			if err = size.Validate(evergreen.GetEnvironment().Settings().Providers.AWS.Pod.ECS); err != nil {
+				catcher.Add(err)
+			}
+		}
+		if catcher.HasErrors() {
+			return false, errors.Wrapf(catcher.Resolve(), "validating container size deginitions")
```
Kimchelly, line 1964: nit: deginitions -> definitions
Can we include the name of the invalid container size?

Kimchelly (approved):
LGTM mod nit.

malikchaya2 (approved):

khelif96 (approved):
LGTM

PR #5917, opened 2022-09-08
EVG-17243 remove grip senders for patched config

ablack12:
[EVG-17243 ](https://jira.mongodb.org/browse/EVG-17243 )

### Description 
Taking a mild shot in the dark with this one. I'm not convinced that the output from this command is ever particularly useful, since we only ever log it to grip and haven't known about it for some time, so I'm wondering if just removing the sender is better than figuring a better way to send it.


ablack12:
@Kimchelly if we do want to log this I'm honestly not sure the best way to write to a different sender and then send that to grip, but I'm not certain it's worth the effort.

Kimchelly:
I don't have particular opinions about getting the output from this, but one common way of capturing the output without sending it straight to the global logger is write the command stdout/stderr to a temporary in-memory buffer ([example](https://github.com/evergreen-ci/evergreen/blob/773eea085ef862e841baac462340b21d58d0a51f/model/host/hostutil.go#L260)).

julianedwards:
Would setting `--quiet` do the trick? [Documentation here](https://git-scm.com/docs/git-apply#Documentation/git-apply.txt---quiet)

Kimchelly:
I think using `--quiet` would be equivalent to what this change already does. I think this is more a question of whether we find the stderr from this command important enough to log at all.

julianedwards:
I am not opposed to removing it if its not important! Or maybe setting it to debug?

ablack12:
Sorry I'm getting to this so late 😅 I think writing everything to a buffer and only logging if there's an error maybe be good? Theoretically in an error case we may want to know what happened.

Kimchelly (approved):

julianedwards (approved):

PR #5920, opened 2022-09-09
EVG-17637 Add the slack memberId  to the user struct 

malikchaya2:
[EVG-17637](https://jira.mongodb.org/browse/EVG-17637)

### Description 
This adds the slack memberId to the user struct and adds a method to find it based on the userId. This will be needed when switching over to using the slack app. 

Todo: add an index. 

### Testing 
Added a unit test and modified existing tests.


graphql/schema/types/user.graphql
```diff
@@ -111,6 +112,7 @@ type UserSettings {
   notifications: Notifications
   region: String
   slackUsername: String
+  slackMemberId: String
```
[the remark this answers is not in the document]
malikchaya2: Yes, we want them to be able to. I will add it here in another PR. 
image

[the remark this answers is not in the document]
khelif96: How would a user find their memberId? it isn't immediately obvious to me and i would imagine it would be confusing for users as well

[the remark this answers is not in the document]
malikchaya2: I added a tooltip with information in the legacy UI, I can do the same for the UI PR.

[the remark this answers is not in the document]
malikchaya2: image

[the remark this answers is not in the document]
malikchaya2: (Copy Member ID)

model/user/user.go
```diff
@@ -454,3 +455,12 @@ func IsValidSubscriptionPreference(in string) bool {
 		return false
 	}
 }
+
+func GetSlackMemberId(username string) (string, error) {
+	u, err := FindBySlackUsername(username)
+	if err != nil {
+		return "", errors.Wrapf(err, "finding user by slack username '%s'", username)
+	}
+
```
bynn, line 464: should we nil check for u here so we dont get a nil dereference

bynn (changes requested):

model/user/user.go
malikchaya2: done

bynn (approved):
it looks good to me!

agent/internal/client/request.go
```diff
@@ -42,20 +42,21 @@ func (c *baseCommunicator) newRequest(method, path, taskID, taskSecret, version
...
 			out, err = json.Marshal(data)
 			if err != nil {
 				return nil, err
 			}
-			r.Header.Add(evergreen.ContentLengthHeader, strconv.Itoa(len(out)))
 			r.Body = ioutil.NopCloser(bytes.NewReader(out))
 		}
 	}
 
+	r.Header.Add(evergreen.ContentLengthHeader, strconv.Itoa(len(out)))
```
khelif96, line 58: I think you accidentally committed this.

config.go
```diff
@@ -36,7 +36,7 @@ var (
 	ClientVersion = "2022-08-31"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2022-09-09"
+	AgentVersion = "2022-09-13"
```
khelif96, line 39: Was this intendeded?

khelif96 (changes requested):

khelif96 (approved):

agent/internal/client/request.go
malikchaya2: removed

config.go
malikchaya2: removed

PR #5921, opened 2022-09-09
EVG-16883: refactor existing stranded task job logic

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16883

### Description 
This is some pre-work to actually handle stranded container tasks because of some tech debt in the stranded task jobs. There will be a follow-up PR to actually handle stranded container tasks in these jobs.

I still think these jobs have some issues in term of having overlapping (and not quite correct) functionality, but I didn't want to address all those issues in this cleanup PR. I'll include some notes in the follow-up PR on how we might clean up these jobs further, because I believe the task heartbeat timeout job might subsume all of the stranded task job's functionality.

* Use Amboy retryable job features rather than manually re-enqueueing task heartbeat timeout jobs.
* Remove unused fields (such as task execution number) from task heartbeat timeout job.
* Replace a strange choice in the stranded task job to unschedule vs. restart a task based on task creation time. Instead, it checks when the task was activated and uses the more standard "task unscheduling threshold" instead of the magic 2-week threshold originally set in the job.
* Fix a bug where the stranded task job would not check for stale dispatched tasks if there were no already-terminated hosts with running tasks.
* Clean up job logic to be more readable.
* Improve doc comments for jobs and some tricky stranded task cases.
* Improve logging so we can better track of stranded tasks cleaned up by these jobs.

### Testing 
The existing tests only cover portions of the jobs and it will require some work to test them more thoroughly, so I'm inclined to say these are simple enough changes to say it's logically equivalent by inspection for now. The follow-up PR will improve the test coverage.

### DIFF
--- a/units/task_monitor_execution_timeout.go
+++ b/units/task_monitor_execution_timeout.go
@@ -24,7 +24,7 @@ const (
 	heartbeatTimeoutThreshold             = 7 * time.Minute
 	taskExecutionTimeoutJobName           = "task-execution-timeout"
 	taskExecutionTimeoutPopulationJobName = "task-execution-timeout-populate"
-	maxAttempts                           = 10
+	maxTaskExecutionTimeoutAttempts       = 10
 )
 
 func init() {
@@ -37,12 +37,9 @@ func init() {
 }
 
 type taskExecutionTimeoutJob struct {
-	Task      string `bson:"task_id"`
-	Execution int    `bson:"execution"`
-	Attempt   int    `bson:"attempt"`
+	Task string `bson:"task_id"`
 
-	successful bool
-	job.Base   `bson:"metadata" json:"metadata" yaml:"metadata"`
+	job.Base `bson:"metadata" json:"metadata" yaml:"metadata"`
 }
 
 func makeTaskExecutionTimeoutMonitorJob() *taskExecutionTimeoutJob {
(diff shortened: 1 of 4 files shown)


units/task_monitor_execution_timeout.go
```diff
@@ -57,12 +54,18 @@ func makeTaskExecutionTimeoutMonitorJob() *taskExecutionTimeoutJob {
 	return j
 }
 
-func NewTaskExecutionMonitorJob(taskID string, execution int, attempt int, ts string) amboy.Job {
+// NewTaskExecutionMonitorJob returns a job to check if a running task has to
```
hadjri, line 57: I think you may have meant something like "if a running task has failed to send a heartbeat"

units/task_stranded_cleanup.go
```diff
@@ -69,39 +74,54 @@ func (j *taskStrandedCleanupJob) Run(ctx context.Context) {
 		taskIDs = append(taskIDs, h.RunningTask)
 		hostIDs = append(hostIDs, h.Id)
 
-		j.AddError(model.ClearAndResetStrandedHostTask(&h))
+		catcher.Wrapf(model.ClearAndResetStrandedHostTask(&h), "fixing stranded host task '%s' on host '%s'", h.RunningTask, h.Id)
 	}
 
+	grip.InfoWhen(len(taskIDs) > 0 || len(hostIDs) > 0, message.Fields{
```
hadjri, line 80: Wouldn't we want `&&` here? Seems like we would only want to log out when we have found tasks to fix.

units/task_stranded_cleanup.go
```diff
@@ -69,39 +74,54 @@ func (j *taskStrandedCleanupJob) Run(ctx context.Context) {
...
-	grip.InfoWhen(!j.HasErrors(),
-		message.Fields{
-			"job":       j.ID(),
-			"op":        j.Type().Name,
-			"num_hosts": len(hosts),
-			"num_tasks": len(taskIDs),
-			"tasks":     taskIDs,
-			"hosts":     hostIDs,
-		})
+	grip.InfoWhen(len(tasks) > 0, message.Fields{
```
hadjri, line 118: Can we split this into two grip statements? One that tells us which tasks were in `tasksToDeactivate` and which ones were not?

hadjri (changes requested):

units/task_stranded_cleanup.go
Kimchelly: Hmm, I re-read this and checking either `taskIDs` or `hostIDs` being non-empty is redundant. It seems like it's easier to just return early if the query returns no hosts, so we don't need to conditionally log this.

units/task_stranded_cleanup.go
Kimchelly: Done.

hadjri (approved):

PR #5923, opened 2022-09-09
EVG-16907: fix temporary nil interface check

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16907

### Description 
I forgot about [the nil interface checking rule](https://mangatmodi.medium.com/go-check-nil-interface-the-right-way-d142776edef1), so checking for a non-nil Secrets Manager client is insufficient because an interface (`cocoa.SecretsManagerClient`) wrapping a nil pointer (`*BasicSecretsManagerClient`) is not nil. It seems like setting up container secrets _should_ happen (or else the project can't run container tasks), but we have to wait until Secrets Manager is available to production.

### Testing 
Not really testable, so I just did it by inspection.

### DIFF
--- a/rest/data/project.go
+++ b/rest/data/project.go
@@ -93,9 +93,11 @@ func CreateProject(ctx context.Context, env evergreen.Environment, projectRef *m
 	}
 
 	cachedSettings := env.Settings()
-	smClient, err := cloud.MakeSecretsManagerClient(cachedSettings)
-	if err != nil {
-		grip.Warning(message.WrapError(err, message.Fields{
+	// TODO (PM-2950): remove this temporary error-checking once the AWS
+	// infrastructure is productionized and AWS admin settings are set.
+	smClient, clientErr := cloud.MakeSecretsManagerClient(cachedSettings)
+	if clientErr != nil {
+		grip.Warning(message.WrapError(clientErr, message.Fields{
 			"message":            "cannot set up Secrets Manager client to store newly-created project's container secrets",
 			"op":                 "CreateProject",
 			"project_id":         projectRef.Id,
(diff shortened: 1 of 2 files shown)


rest/data/project.go
```diff
@@ -169,6 +129,51 @@ func CreateProject(ctx context.Context, env evergreen.Environment, projectRef *m
...
+	smClient, err := cloud.MakeSecretsManagerClient(settings)
+	if err != nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    errors.Wrap(err, "setting up Secrets Manager client to store newly-created project's container secrets").Error(),
+		}
+	}
+	defer smClient.Close(ctx)
+
+	vault, err := cloud.MakeSecretsManagerVault(smClient)
```
ablack12, line 144: Previously we only did this if smClient wasn't nil; do we still need to do that?

rest/data/project.go
```diff
@@ -169,6 +129,51 @@ func CreateProject(ctx context.Context, env evergreen.Environment, projectRef *m
...
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    errors.Wrapf(err, "updating project ref's container secrets").Error(),
+		}
+	}
+	// This updates the container secrets in the DB project ref only, not
+	// the in-memory copy.
+	if err := UpsertContainerSecrets(ctx, vault, pRef.ContainerSecrets); err != nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
```
ablack12, line 169: If this function is always going to return a StatuInternalServerError and we're only logging it then there's not really a reason to just return the error.Wraps directly.

rest/data/project.go
Kimchelly: `cloud.MakeSecretsManagerClient` will return either an error or a non-nil object, but not both, so we don't need to check for nil here.

rest/data/project.go
Kimchelly: True, I can remove all the error responses.

ablack12 (approved):

bynn (approved):

hadjri (approved):

PR #5924, opened 2022-09-09
EVG-17335 validate commit queue aliases

ablack12:
[EVG-17335](https://jira.mongodb.org/browse/EVG-17335)

### Description 
User made a case that commit queue aliases are important enough that they should be validated if they don't match anything. Wrote a change to do this, and it included some refactors.

Open question @hadjri : I think it would be cool to validate all aliases but its tricky thinking of how that would work, in terms of joining all the aliases together. Should we just validate all aliases since it's only warnings anyway? But then we'd probably need to show the source (maybe that's okay though).
### Testing 
Added unit test and tested locally:
```
commit-queue-sandbox [new-branch] $ ../evergreen/clients/darwin_amd64/evergreen -c ~/.evergreen-local.yml validate -p spruce evergreen.yml 
WARNING: Commit queue alias matching variant regexp 'nothing' has no matching variants
WARNING: Commit queue alias matching variant regexp 'ubuntu1604' and matching task tags '[something specific]' has no matching tasks
```

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -74,7 +74,16 @@ type ProjectAlias struct {
 	VariantTags []string         `bson:"variant_tags,omitempty" json:"variant_tags" yaml:"variant_tags"`
 	Task        string           `bson:"task,omitempty" json:"task" yaml:"task"`
 	TaskTags    []string         `bson:"tags,omitempty" json:"tags" yaml:"task_tags"`
+
+	// Source is not stored; indicates where the alias is stored for the project.
+	Source string `bson:"-" json:"-" yaml:"-"`
 }
+
+const (
+	AliasSourceProject = "project"
+	AliasSourceConfig  = "config"
+	AliasSourceRepo    = "repo"
+)
 
 type ProjectAliases []ProjectAlias
 
@@ -91,15 +100,23 @@ func FindAliasesForProjectFromDb(projectID string) ([]ProjectAlias, error) {
 	return out, nil
 }
 
-// MergeAliasesWithProjectConfig returns a merged list of project aliases that includes the merged result of aliases defined
+// GetAliasesMergedWithProjectConfig returns a merged list of project aliases that includes the merged result of aliases defined
 // on the project ref and aliases defined in the project YAML.  Aliases defined on the project ref will take precedence over the
 // project YAML in the case that both are defined.
-func MergeAliasesWithProjectConfig(projectID string, dbAliases []ProjectAlias) ([]ProjectAlias, error) {
-	dbAliasMap := aliasesToMap(dbAliases)
+func GetAliasesMergedWithProjectConfig(projectID string, dbAliases []ProjectAlias) ([]ProjectAlias, error) {
 	projectConfig, err := FindProjectConfigForProjectOrVersion(projectID, "")
 	if err != nil {
 		return nil, errors.Wrap(err, "finding project config")
 	}
+	return mergeProjectConfigAndAliases(projectConfig, dbAliases), nil
+}
+
+func mergeProjectConfigAndAliases(projectConfig *ProjectConfig, dbAliases []ProjectAlias) []ProjectAlias {
+	if projectConfig == nil {
+		return dbAliases
+	}
+	dbAliasMap := aliasesToMap(dbAliases)
+
 	patchAliases := []ProjectAlias{}
 	for alias, aliases := range dbAliasMap {
 		if IsPatchAlias(alias) {
@@ -129,7 +146,7 @@ func MergeAliasesWithProjectConfig(projectID string, dbAliases []ProjectAlias) (
 	mergedAliases = append(mergedAliases, dbAliasMap[evergreen.GitTagAlias]...)
 	mergedAliases = append(mergedAliases, dbAliasMap[evergreen.GithubPRAlias]...)
 	mergedAliases = append(mergedAliases, patchAliases...)
-	return mergedAliases, nil
+	return mergedAliases
 }
 
 // FindAliasesForRepo fetches all aliases for a given project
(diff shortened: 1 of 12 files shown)


validator/project_validator.go
```diff
@@ -449,6 +441,113 @@ func validateProjectConfigAliases(pc *model.ProjectConfig) ValidationErrors {
...
+			// If we still need a task to match the variant, still check if the alias matches, so we know if checking tasks is needed.
+			matchesThisVariant, err := alias.HasMatchingVariant(bv.Name, bv.Tags)
+			if err != nil {
+				return nil, nil, err
+			}
+			if !matchesThisVariant { // If the variant doesn't match, then there's no reason to keep checking tasks.
+				continue
+			}
+			aliasNeedsVariant[aliasID] = false
+			for _, t := range bv.Tasks {
```
hadjri, line 524: Can we refactor this third loop to another helper function that takes in the `bvtCache`, `bv`, and `alias` and returning a bool/error?

hadjri (changes requested):

hadjri:
> [EVG-17335](https://jira.mongodb.org/browse/EVG-17335)
> 
> ### Description
> User made a case that commit queue aliases are important enough that they should be validated if they don't match anything. Wrote a change to do this, and it included some refactors.
> 
> Open question @hadjri : I think it would be cool to validate all aliases but its tricky thinking of how that would work, in terms of joining all the aliases together. Should we just validate all aliases since it's only warnings anyway? But then we'd probably need to show the source (maybe that's okay though).
> 
> ### Testing
> Added unit test and tested locally:
> 
> ```
> commit-queue-sandbox [new-branch] $ ../evergreen/clients/darwin_amd64/evergreen -c ~/.evergreen-local.yml validate -p spruce evergreen.yml 
> WARNING: Commit queue alias matching variant regexp 'nothing' has no matching variants
> WARNING: Commit queue alias matching variant regexp 'ubuntu1604' and matching task tags '[something specific]' has no matching tasks
> ```

I think doing this for all aliases would be helpful since we've done all the core work to detect if an alias's task/variant regex references anything or not. Especially since as you said it's just a warning and I don't think this would be too computationally expensive to add for the rest of the aliases.  We can refactor it a bit by passing all aliases into `CheckAliasWarnings` and checking for the alias type when generating a validation error message. (We also have `IsPatchAlias` to detect if an alias is a patch alias since they don't have hard defined alias names).

ablack12:
Updated for all aliases, combined into one warning but could be convinced not to do that:
```
WARNING: Commit queue alias matching variant regexp '.*' and matching task tags '[no_tag]' has no matching tasks
Git tag alias matching variant regexp '.*' and matching task tags '[nonexistent]' has no matching tasks
Github PR alias matching variant regexp '.*' and matching task tags '[nothing]' has no matching tasks
Github check alias matching variant regexp 'nothing' has no matching variants
Patch alias matching variant regexp 'variant_regex' has no matching variants
Patch alias matching variant tags '[tag]' has no matching variants
self-tests.yml is valid with warnings
```

validator/project_validator.go
```diff
@@ -229,6 +228,10 @@ func CheckProjectWarnings(project *model.Project) ValidationErrors {
 	return validationErrs
 }
 
+func CheckAliasWarnings(project *model.Project, aliases model.ProjectAliases) ValidationErrors {
```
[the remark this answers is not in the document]
hadjri: Ok, sure that makes sense, I'm fine to keep it

rest/data/aliases.go
```diff
@@ -26,6 +26,7 @@ func FindProjectAliases(projectId, repoId string, aliasesToAdd []restModel.APIPr
 			aliasModels = append(aliasModels, a)
 		}
 	}
+	// TODO <ticket_num>: modify to correctly merge aliases based on type
```
hadjri, line 29: Can we create a ticket and link it here?

service/api.go
```diff
@@ -688,19 +688,14 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
...
-				}
-			}
+		aliases, err := model.FindMergedAliasesFromProjectRepoOrProjectConfig(projectRef, projectConfig)
+		if err != nil {
+			errs = append(errs, validator.ValidationError{
+				Message: "problem finding aliases; validation will proceed without checking alias coverage",
+				Level:   validator.Warning,
+			})
+		} else {
+			fmt.Println("len aliases: ", len(aliases))
```
hadjri, line 698: Is this leftover debugging?

model/project_aliases.go
```diff
@@ -236,6 +261,71 @@ func FindAliasInProjectRepoOrConfig(projectID, alias string) ([]ProjectAlias, er
...
+	for _, alias := range projectAliases {
+		aliasName := alias.Alias
+		if IsPatchAlias(aliasName) {
+			aliasName = patchAliasKey
+		}
+		alias.Source = AliasSourceProject
+		aliasesToReturn[aliasName] = append(aliasesToReturn[aliasName], alias)
+	}
+	// All aliases are covered in the project, so there's no reason to look at other sources
+	aliasesCovered := allAliasesCovered(aliasesToReturn)
```
hadjri, line 286: Can we instead return a list of alias names that are not covered instead of checking if we have all 5 alias names in the map? Then we can just check for these missing alias names in the loop below and we can remove the `if len(aliasesToReturn[aliasName]) == 0 || ...` checking

hadjri (changes requested):

rest/data/aliases.go
ablack12: yeah wanted to hold on to until this change was approved 👍  will make now

hadjri (approved):

hadjri:
LGTM mod lint fix

ablack12:
evergreen merge

ablack12:
evergreen retry

ablack12:
evergreen retry

PR #5926, opened 2022-09-12
Evg 16883 handle stale heartbeat container tasks

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-16883

### Description
This is largely a rewrite of the stale heartbeat task checking logic to ensure it's up-to-date.

* Use the common helper functions for handling stale host tasks. This was for correctness (a stale execution task should end up resetting only the failed tasks in the display task, just like for stranded tasks) and maintainability (the logic for stale heartbeat tasks should be very similar to the logic for stranded tasks).
* Handle container tasks that have a stale heartbeat.
* Add helper function for blocking single host task groups to reduce some copy-pasting in various places related to ending tasks.
* Improve stale heartbeat checking job to be more unit testable.

### Testing 
Added unit tests and migrated some old tests off of go convey.

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -499,7 +499,7 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
 		return ProjectInfo{}, errors.Wrap(err, "finding parser project")
 	}
 	var pc *ProjectConfig
-	if pRef.IsVersionControlEnabled() {
+	if pRef != nil && pRef.IsVersionControlEnabled() {
 		pc, err = FindProjectConfigForProjectOrVersion(v.Identifier, v.Id)
 		if err != nil {
 			return ProjectInfo{}, errors.Wrap(err, "finding project config")
--- a/units/task_monitor_execution_timeout.go
+++ b/units/task_monitor_execution_timeout.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
-	"github.com/evergreen-ci/evergreen/apimodels"
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/task"
@@ -129,6 +128,13 @@ func (j *taskExecutionTimeoutJob) Run(ctx context.Context) {
 // cleanUpTimedOutTask cleans up a single host task that has exceeded the
 // task heartbeat timeout.
 func cleanUpTimedOutTask(ctx context.Context, env evergreen.Environment, id string, t *task.Task) error {
+	if t.IsContainerTask() {
+		// kim: TODO: fix up this logic to handle container tasks.
+		return nil
+	} else {
+
+	}
+
 	host, err := host.FindOne(host.ById(t.HostId))
 	if err != nil {
 		return errors.Wrapf(err, "finding host '%s' for task '%s'", t.HostId, t.Id)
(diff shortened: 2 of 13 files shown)


Kimchelly:
@bynn I made a change to make stale tasks that hit their heartbeat timeout reset only the failed tasks to follow the behavior of stranded tasks. Can you confirm that this is how it's supposed to behave?

bynn (approved):
from what i can see the logic seems good!

units/task_monitor_execution_timeout.go
```diff
@@ -126,87 +137,64 @@ func (j *taskExecutionTimeoutJob) Run(ctx context.Context) {
...
+		return errors.Wrapf(err, "resetting stale task '%s'", j.task.Id)
 	}
 
-	// try to reset the task
-	if t.IsPartOfDisplay() {
-		dt, err := t.GetDisplayTask()
-		if err != nil {
-			return errors.Wrapf(err, "getting display task")
-		}
-		if err = dt.SetResetWhenFinished(); err != nil {
```
hadjri, line 196: Did we lose the functionality of this line in the refactoring?

units/task_monitor_execution_timeout.go
```diff
@@ -89,32 +93,39 @@ func (j *taskExecutionTimeoutJob) Run(ctx context.Context) {
...
 	}
 
-	t, err := task.FindOneId(j.Task)
-	if err != nil {
-		j.AddRetryableError(errors.Wrapf(err, "finding task '%s'", j.Task))
-		return
+	if j.task == nil {
+		t, err := task.FindOneId(j.Task)
+		if err != nil {
+			j.AddRetryableError(errors.Wrapf(err, "finding task '%s'", j.Task))
```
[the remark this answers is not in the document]
Kimchelly: The document not found error is covered by L102 - this query (and most queries in Evergreen) will only return a non-nil error for exceptional things like being unable to connect to the DB.

units/task_monitor_execution_timeout.go
Kimchelly: This is part of the change I mentioned in the description with display tasks. We should ideally treat tasks failing due to Evergreen-internal system failures the same, so I standardized the stale heartbeat task case to have the same effect as stranded tasks of [resetting only the failed tasks instead of all the tasks under the display task](https://github.com/evergreen-ci/evergreen/blob/70098687f379d7cc73be2d659ce04221ac1bfdcc/model/task_lifecycle.go#L1639-L1642). This is the desired behavior outlined in the [scope doc](https://docs.google.com/document/d/1xmFAyJUvksEYZgqmbiWo990mX5ZlXydcuWeTZTO-iMA/edit).

model/project_parser.go
```diff
@@ -499,7 +499,7 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
 		return ProjectInfo{}, errors.Wrap(err, "finding parser project")
 	}
 	var pc *ProjectConfig
-	if pRef.IsVersionControlEnabled() {
+	if pRef != nil && pRef.IsVersionControlEnabled() {
```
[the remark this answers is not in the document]
Kimchelly: Got it. I didn't know if this was a continue or error case, so I originally changed it to just avoid a panic. It should now error if there's no project ref found.

units/task_monitor_execution_timeout.go
hadjri: Ah I see, thanks for the clarification.

hadjri (approved):

PR #5927, opened 2022-09-13
EVG-17685 Redirect legacy task queue page to spruce

bynn:
[EVG-17658](https://jira.mongodb.org/browse/EVG-17658)

### Description 
task queue was broken on legacy because it couldnt handle that many tasks 
so i am redirecting task queue to spruce now 

### Testing 
manually on staging

### DIFF
--- a/service/task_queue.go
+++ b/service/task_queue.go
@@ -12,6 +12,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/patch"
 	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/gimlet"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


khelif96 (approved):

malikchaya2 (approved):

bynn:
https://jira.mongodb.org/browse/EVG-17689 
ticket created for adding switch to legacy button on spruce

khelif96:
> https://jira.mongodb.org/browse/EVG-17689 ticket created for adding switch to legacy button on spruce

The switch to legacy might not work with the way this is implemented since this now just automatically redirects you. fyi

bynn:
> > https://jira.mongodb.org/browse/EVG-17689 ticket created for adding switch to legacy button on spruce
> 
> The switch to legacy might not work with the way this is implemented since this now just automatically redirects you. fyi

i created a new legacy route for task_queue/distro/task_id so i think it should be fine to redirect to the old task_queue/task_id route. I might be able to take the ticket next sprint

ablack12:
Even if it automatically redirects you there'd still be a button option for switching right? Or else I'd like to hold off on this automatic redirect and only add the button until the priority work is done.

ablack12:
@sanjogDB discussed with Bynn in-person; since the legacy task queue page does seem more broken than less broken it seems like EVG-16828 is more accessible than fixing the legacy page until it's done. Does this seem like something we can prioritize higher?

ablack12:
Also discussed offline -- @sanjogDB is going to work on the priority of this so Bynn I think you're good to just let the legacy page die, and maybe open a ticket to remove it once we're sure

bynn:
because the priority was added to spruce, i updated this pr to delete the legacy task queue page

PR #5929, opened 2022-09-13
EVG-16828 Reduce the amount of parameters passed into task/build creation functions

hadjri:
[EVG-16828](https://jira.mongodb.org/browse/EVG-16828)

### Description 
Removed the `BuildCreateArgs` struct (which was previously only used for providing parameters to the `CreateBuildFromVersionNoInsert` function) and repurposed it into `TaskCreationInfo`, which is now a struct that contains all of the common parameters that are passed back and forth between task & build creation functions.

Change was done to reduce complexity and improve readability since a lot of the task/build creation functions were passing in many (in some cases ~10) parameters.

After the reviewal process I plan to move these new modified functions to a dedicated task_creation.go file since task creation logic is complex enough for its own file. I haven't done that yet because it would make it harder to view the diff in this PR.
### Testing 
Modified existing unit tests to fit the new paradigm. Deployed to staging and ran a few patches to create & run new tasks and builds.

### DIFF
--- a/model/task_creation.go
+++ b/model/task_creation.go
@@ -1,11 +1,12 @@
 package model
 
 import (
+	"time"
+
 	"github.com/evergreen-ci/evergreen/model/build"
 	"github.com/evergreen-ci/evergreen/model/distro"
 	"github.com/evergreen-ci/evergreen/model/patch"
 	"github.com/evergreen-ci/evergreen/model/task"
-	"time"
 )
 
 // TaskCreationInfo contains the needed parameters to construct new builds and tasks for a given version.


malikchaya2 (approved):

model/task_creation.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+)
+
+// TaskCreationInfo contains the needed parameters to construct new builds and tasks for a given version.
+type TaskCreationInfo struct {
+	Version             *Version
+	Project             *Project
+	ProjectRef          *ProjectRef
+	buildVariant        *BuildVariant
+	build               *build.Build
```
ablack12, line 17: is there a reason some of these are uppercase versus lowercase? (If the lowercase ones have certain in-package use, can they be moved below and commented?)

model/task_creation.go
```diff
@@ -0,0 +1,34 @@
...
+	pairs               TaskVariantPairs
+	TaskIDs             TaskIdConfig            // pre-generated IDs for the tasks to be created
+	BuildName           string                  // name of the buildvariant
+	ActivateBuild       bool                    // true if the build should be scheduled
```
ablack12, line 21: nit that comments should be capitalized

ablack12 (commented):
More of a skim than a review so no need to block on me

model/task_creation.go
hadjri: No - this is a leftover from me merging the new task creation struct (which was initially private) with the old `BuildCreateArgs` struct which is/was public. Fixed now, they are all public.

ablack12 (approved):

PR #5931, opened 2022-09-14
EVG-17450 Move host agent routes to requirePodOrHost middleware

hadjri:
[EVG-17450](https://jira.mongodb.org/browse/EVG-17450)

### Description 
Context: https://mongodb.slack.com/archives/C866SR2LR/p1660676410998359

We have some agent-side v2 REST upgrades and in the past when we deploy, legacy Windows SSH hosts start failing, constantly retrying to deploy the agent, never succeeding, and then ultimately terminating the EC2 hosts.  After some digging, what seems to be the culprit is a difference between how `requireHost` works for v1 vs v2.  On v1, we [update the last communicated time](https://github.com/evergreen-ci/evergreen/blob/main/service/api.go#L168) for the host whereas on v2 we do not (probably because host based agent routes were not on v2 until recently).  So, when a host finishes running tasks and enters an idle polling period for `next_task`, it wasn't updating the last communicated time each time as it should.  This in turn causes our cron job for agent deployment to legacy SSH hosts [think that these hosts are idle](MaxLCTInterval) during this period since they don't communicate to the app server in our predefined cutoff time (5 minutes).

The only new change in this PR is to change the middleware on the pod / host next task and end task endpoints to use `requirePodOrHost` middleware, which actually will update the last communication time from the host/pod to the app server.
### Testing 
Deployed to staging without the new middleware change on a legacy SSH host `ubuntu1604-packer`, and sure enough 5 minutes after the host cleared its task the host started attempting new agent deployments (screenshot attached, ignore the setup failure it isn't relevant to the change here). Furthermore in splunk, this reproduced the ["Failed to create the file evergreen.exe: Text file busy"](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen-staging%22%20ubuntu1604-packer%20%22text%20file%20busy%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1663189680&latest=1663193334&sid=1663193358.2227890) errors that occurred in prod last time we deployed these agent side upgrades.

Deployed again with the middleware update and the above splunk errors did not occur nor did the host try to redeploy the agent 5 minutes after the task ended.

### DIFF
--- a/rest/route/service.go
+++ b/rest/route/service.go
@@ -61,10 +61,10 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddWrapper(gimlet.WrapperMiddleware(allowCORS))
 
 	// Agent routes
-	app.AddRoute("/pods/{pod_id}/agent/next_task").Version(2).Get().Wrap(requirePodOrHost).RouteHandler(makePodAgentNextTask(env))
-	app.AddRoute("/pods/{pod_id}/task/{task_id}/end").Version(2).Post().Wrap(requirePodOrHost, requireTask).RouteHandler(makePodAgentEndTask(env))
-	app.AddRoute("/hosts/{host_id}/agent/next_task").Version(2).Get().Wrap(requirePodOrHost).RouteHandler(makeHostAgentNextTask(env, opts.TaskDispatcher, opts.TaskAliasDispatcher))
-	app.AddRoute("/hosts/{host_id}/task/{task_id}/end").Version(2).Post().Wrap(requirePodOrHost, requireTask).RouteHandler(makeHostAgentEndTask(env))
+	app.AddRoute("/pods/{pod_id}/agent/next_task").Version(2).Get().Wrap(requirePod).RouteHandler(makePodAgentNextTask(env))
+	app.AddRoute("/pods/{pod_id}/task/{task_id}/end").Version(2).Post().Wrap(requirePod, requireTask).RouteHandler(makePodAgentEndTask(env))
+	app.AddRoute("/hosts/{host_id}/agent/next_task").Version(2).Get().Wrap(requireHost).RouteHandler(makeHostAgentNextTask(env, opts.TaskDispatcher, opts.TaskAliasDispatcher))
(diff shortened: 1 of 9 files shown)


rest/route/service.go
```diff
@@ -61,10 +61,10 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddWrapper(gimlet.WrapperMiddleware(allowCORS))
 
 	// Agent routes
-	app.AddRoute("/pods/{pod_id}/agent/next_task").Version(2).Get().Wrap(requirePod).RouteHandler(makePodAgentNextTask(env))
-	app.AddRoute("/pods/{pod_id}/task/{task_id}/end").Version(2).Post().Wrap(requirePod, requireTask).RouteHandler(makePodAgentEndTask(env))
-	app.AddRoute("/hosts/{host_id}/agent/next_task").Version(2).Get().Wrap(requireHost).RouteHandler(makeHostAgentNextTask(env, opts.TaskDispatcher, opts.TaskAliasDispatcher))
-	app.AddRoute("/hosts/{host_id}/task/{task_id}/end").Version(2).Post().Wrap(requireHost, requireTask).RouteHandler(makeHostAgentEndTask(env))
+	app.AddRoute("/pods/{pod_id}/agent/next_task").Version(2).Get().Wrap(requirePodOrHost).RouteHandler(makePodAgentNextTask(env))
```
Kimchelly, line 64: Rather than change all the pod or host specific ones to accept both auth methods, it should be safe to change `requireHost` to [do the same logic as pod or host auth for hosts](https://github.com/evergreen-ci/evergreen/blob/b28150c48556bf161354f9b80990f5d9eb36af82/rest/route/middleware.go#L345-L358). I checked the two routes that use it and both of them are agent routes where it's okay to update the last communicated time.

Kimchelly (changes requested):

rest/route/service.go
hadjri: Sounds good, I've moved this to a helper function

Kimchelly (approved):

PR #5934, opened 2022-09-14
EVG-17338 Remove error when starting task is not found

bynn:
[EVG-17338](https://jira.mongodb.org/browse/EVG-17338)

### Description 
We currently error when the start_at task is not found for the projects/{project_id}/revisions/{hash}/tasks route 
I'm not super sure why we error at this point but I think we can log something at this point instead of returning a 404 and investigate if it keeps coming up 
(I was also not able to recreate this bug myself)

### DIFF
--- a/rest/data/task.go
+++ b/rest/data/task.go
@@ -9,6 +9,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/gimlet"
 	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 )
 
@@ -85,8 +86,13 @@ func FindTasksByProjectAndCommit(opts task.GetTasksByProjectAndCommitOptions) ([
 				break
 			}
 		}
-		// TODO: EVG-17897 remove the debug statement
-		grip.DebugWhen(!found, fmt.Sprintf("starting task '%s' not found", opts.StartingTaskId))
+		grip.DebugWhen(!found, message.Fields{
+			"message":       "starting task not found",
+			"ticket":        "EVG-17338",
+			"starting_task": opts.StartingTaskId,
+			"project":       projectId,
+			"commit":        opts.CommitHash,
+		})
 	}
 	return res, nil
 }
--- a/rest/data/task.go
+++ b/rest/data/task.go
@@ -50,7 +50,7 @@ func FindTasksByProjectAndCommit(opts task.GetTasksByProjectAndCommitOptions) ([
 	if err != nil {
 		return nil, gimlet.ErrorResponse{
 			StatusCode: http.StatusNotFound,
-			Message:    errors.Wrapf(err, "project '%s' not found").Error(),
+			Message:    errors.Wrapf(err, "project '%s' not found", projectId).Error(),
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rest/data/task.go
```diff
@@ -84,12 +85,7 @@ func FindTasksByProjectAndCommit(opts task.GetTasksByProjectAndCommitOptions) ([
 				break
 			}
 		}
-		if !found {
-			return []task.Task{}, gimlet.ErrorResponse{
-				StatusCode: http.StatusNotFound,
-				Message:    fmt.Sprintf("task '%s' not found", opts.StartingTaskId),
-			}
-		}
+		grip.DebugWhen(!found, fmt.Sprintf("starting task '%s' not found", opts.StartingTaskId))
```
ablack12, line 88: We should avoid logging debug statements we don't have a plan for -- without a ticket number and/or planned investigation for it, we put logging in splunk that we don't actually need. 

If you think this will be helpful for future debugging, then I would include the current ticket number, and then not close the current ticket until you feel confident this change fixed the problem, and then remove the logging. Otherwise, removing the whole for loop is more useful.

ablack12 (commented):
Do we have reason to believe that this is where the loop is getting hit, or are we thinking let's try it and see if it fixes the problem? (If the latter then it may be helpful to add some debug statements to the other places we have 404s, and/or ask the reporter if they have information about the error message they've run into in addition to the status code).

rest/data/task.go
bynn: Yeah, I'm pretty sure this is the part where the route returns 404 in the way that Brooke described so not returning an error should "solve" the issue. 
The problem is that I can't figure out why the result of the loop would suddenly change from not being able to find the task to the route working normally. 

I created a ticket to follow up 
https://jira.mongodb.org/browse/EVG-17897

ablack12: Yeah I'm not sure either but this loop does seem a bit unnecessary so seems worth retrying. 
But by "include the ticket number" I meant to format grip statements in a meaningful way using message.Fields.

I would also ask that you comment on the ticket to let Brooke know that you're trying something but it may not fix the issue and she should let you know if she sees it again, in which case you'll re-open the existing ticket (or don't close the existing ticket for a few days).

ablack12 (changes requested):

rest/data/task.go
bynn: Sorry, I misunderstood. 
I updated brook on the ticket and changed the debug message!

ablack12 (approved):

rest/data/task.go
```diff
@@ -84,12 +86,13 @@ func FindTasksByProjectAndCommit(opts task.GetTasksByProjectAndCommitOptions) ([
 				break
 			}
 		}
-		if !found {
-			return []task.Task{}, gimlet.ErrorResponse{
-				StatusCode: http.StatusNotFound,
-				Message:    fmt.Sprintf("task '%s' not found", opts.StartingTaskId),
-			}
-		}
+		grip.DebugWhen(!found, message.Fields{
```
ablack12, line 89: Can we include any other opts that we may be filtering on?

ablack12 (changes requested):

rest/data/task.go
bynn: printing opts

ablack12 (approved):

PR #5941, opened 2022-09-16
EVG-17631 Lookup full project ref when saving Github & CQ section

hadjri:
[EVG-17631](https://jira.mongodb.org/browse/EVG-17631)

### Description 
Since only a partial project ref is submitted when saving project settings (only containing relevant data per section), the `VersionControlEnabled` property of the project ref is lost when saving for sections outside of the "general" tab and the `IsVersionControlEnabled` doesn't return the correct value.   Change has been made so that we check the full project ref document for `VersionControlEnabled` before we check the flag.
### Testing 
On local evergreen, confirmed that after the change I was able to successfully enable version control for a project ref and then successfully empty the github alias list on the project page and save without an error, indicating that `VersionControlEnabled` was picked up correctly.


rest/data/aliases.go
```diff
@@ -127,7 +127,11 @@ func updateAliasesForSection(projectId string, updatedAliases []restModel.APIPro
 // validateFeaturesHaveAliases returns an error if project/repo aliases are not defined for a Github/CQ feature.
 // Does not error if version control is enabled.
 func validateFeaturesHaveAliases(pRef *model.ProjectRef, aliases []restModel.APIProjectAlias) error {
-	if pRef.IsVersionControlEnabled() {
+	fullProjectRef, err := model.FindBranchProjectRef(pRef.Id)
```
ablack12, line 130: This doesn't take into account the repo ref so this isn't quite right. I think this should instead take in the original project ref as well as the new project ref (which is what it currently takes in) which functions calling it should have available (example: https://github.com/evergreen-ci/evergreen/blob/358b4af89e357a1db8d44b91603b3817e6796e73/rest/data/project_settings.go#L118)

ablack12 (changes requested):

rest/data/aliases.go
hadjri: Sounds good, did this and tested again locally to confirm

rest/data/aliases.go
```diff
@@ -126,8 +126,8 @@ func updateAliasesForSection(projectId string, updatedAliases []restModel.APIPro
 
 // validateFeaturesHaveAliases returns an error if project/repo aliases are not defined for a Github/CQ feature.
 // Does not error if version control is enabled.
-func validateFeaturesHaveAliases(pRef *model.ProjectRef, aliases []restModel.APIProjectAlias) error {
-	if pRef.IsVersionControlEnabled() {
+func validateFeaturesHaveAliases(originalProjectRef *model.ProjectRef, newProjectRef *model.ProjectRef, aliases []restModel.APIProjectAlias) error {
+	if originalProjectRef.IsVersionControlEnabled() {
```
ablack12, line 130: Should we add a comment here or at 128 about why we use this other project ref?

rest/data/project_settings.go
```diff
@@ -220,7 +220,7 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 				return nil, err
 			}
 		}
-		if err = validateFeaturesHaveAliases(mergedProjectRef, changes.Aliases); err != nil {
+		if err = validateFeaturesHaveAliases(mergedBeforeRef, mergedProjectRef, changes.Aliases); err != nil {
```
ablack12, line 223: Would it be crazy to rename mergedProjectRef to mergedSection so we remember in the future that this doesn't contain all of the sections?

ablack12 (approved):
With some thoughts

rest/data/aliases.go
hadjri: Sure, adding

PR #5943, opened 2022-09-19
EVG-16796 Allow patch aliases to define patch parameters

hadjri:
[EVG-16796](https://jira.mongodb.org/browse/EVG-16796)

### Description 
Added `Parameters` as a property on a `ProjectAlias`, which can currently be defined in YAML and will take precedence over the parameters set in the `parameters` field.  Params set directly on the patch configuration page or the CLI will override the alias parameters (although open to discuss this).  This is the backend change necessary for adding parameters UI fields to Spruce allowing for alias parameters to be configured in the UI.  This will be a follow-up ticket.
### Testing 
Added unit test checking that parameters defined on a patch alias are properly returned when the agent requests expansion vars for a given task.

Deployed to staging and defined some parameters on a patch alias and confirmed that they were properly picked up by the tasks that they generated.


model/project.go
```diff
@@ -1769,14 +1769,14 @@ func (p *Project) IsGenerateTask(taskName string) bool {
 	return ok
 }
 
-func (p *Project) findAliasesForPatch(alias string, patchDoc *patch.Patch) ([]ProjectAlias, error) {
-	vars, shouldExit, err := FindAliasInProjectOrRepoFromDb(p.Identifier, alias)
+func FindAliasesForPatch(identifier, alias string, patchDoc *patch.Patch) ([]ProjectAlias, error) {
```
ablack12, line 1772: nit: can we change identifier to projectId ? (Unfortunately patch identifier points to project ID instead of project identifier since those didn't use to be distinct but using identifier here makes it unclear that this isn't project identifier)

model/project_aliases.go
```diff
@@ -343,6 +346,32 @@ func CopyProjectAliases(oldProjectId, newProjectId string) error {
 	return nil
 }
 
+func FindParametersForPatchAlias(identifier string, v *Version) ([]patch.Parameter, error) {
```
ablack12, line 349: Rather than getting aliases from the DB and config each time we call FetchExpansions, could we instead cache them with the patch at the time that we finalize it? (This might allow us to just look at what aliases are being used at that time as well, although I'm not totally sure)

ablack12 (changes requested):

model/project_aliases.go
hadjri: Sure, I can move this logic into `FinalizePatch`

model/patch_lifecycle.go
```diff
@@ -625,6 +630,36 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
 }
 
+// getFullPatchParams will retrieve a merged list of parameters defined on the patch alias (if any)
+// with the parameters that were explicitly user-specified, with the latter taking precedence.
+func getFullPatchParams(p *patch.Patch) ([]patch.Parameter, error) {
+	paramsMap := map[string]string{}
+	if p.Alias != "" && IsPatchAlias(p.Alias) {
+		aliases, err := findAliasesForPatch(p.Project, p.Alias, p)
+		if err != nil {
+			return nil, errors.Wrapf(err, "retrieving alias '%s' for patched project config '%s'", p.Alias, p.Id.Hex())
```
ablack12, line 640: for patched project config seems like it shouldn't be part of the alias message, since aliases can also be defined on the project page, and parameters will be added to this eventually

model/patch_lifecycle.go
```diff
@@ -625,6 +630,36 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 	return patchVersion, nil
 }
 
+// getFullPatchParams will retrieve a merged list of parameters defined on the patch alias (if any)
+// with the parameters that were explicitly user-specified, with the latter taking precedence.
+func getFullPatchParams(p *patch.Patch) ([]patch.Parameter, error) {
+	paramsMap := map[string]string{}
+	if p.Alias != "" && IsPatchAlias(p.Alias) {
```
ablack12, line 637: can we just return p.Parameters in this case rather than re-iterating through at 650?

rest/route/agent_test.go
```diff
@@ -25,6 +26,69 @@ var (
...
+			pRef := model.ProjectRef{
+				Id: "p1",
+			}
+			vars := &model.ProjectVars{
+				Id:          "p1",
+				Vars:        map[string]string{"a": "1", "b": "3"},
+				PrivateVars: map[string]bool{"b": true},
+			}
+			v2 := &model.Version{
+				Id:         "aaaaaaaaaaff001122334456",
```
ablack12, line 73: Shoudl this test be for (or include) the alias parameter case?

hadjri: After responding to the first CR the parameters now get cached on the `Version` at the time of finalizing, so there isn't any new change `FetchExpansionsForTask` as part of this PR. I did keep this test from the initial implementation since we might as well up our code coverage where we can.  I've added another case here where the `Version` has params and checks if they are picked up correctly.

model/patch_lifecycle.go
hadjri: I suppose it is true that we can skip re-iterating and just return `p.Parameters` at L650 if this boolean evaluates false. I've added an else block to check for this.

rest/route/agent_test.go
ablack12: Ah yeah that's what I was looking for, thanks!

model/patch_lifecycle.go
ablack12: I would rather we just did `if p.Alias == "" || !IsPatchAlias(p.Alias) { return p.Parameters, nil } ` at the top, since then we can have the rest of the work not in an if statement, which is cleaner to read.

ablack12 (approved):
LGTM with change

PR #5944, opened 2022-09-19
EVG-17640 data cleanup efficiency

ybrill:
[EVG-17640](https://jira.mongodb.org/browse/EVG-17640)

### Description 
As the TSE explained, the db is suffering from these cleanup jobs since the deletes are each racing for the same document and the losers have an expensive backoff algorithm. Looking over the timing in Splunk, it doesn't appear that there are multiple instances of the job running simultaneously. Rather, each job fires off 100k parallel deletes in a bulk write. Still, I added scopes to make sure there wouldn't be more than one job running at once.

I considered parallelizing the id reads with the deletions. When I tried it out locally it took roughly the same time either way (i.e. most of the time was for the deletions) so the additional complexity seemed unjustified.

We'll have to keep an eye on how this affects the deletion rate. The job is intended to run ~50s and gets enqueued once a minute. Right now it's not using the full time since we never get past the first or second batch. I think the reason is that we break [here](https://github.com/evergreen-ci/evergreen/blob/c5abc064bc679a5063ab4cbe5ce14af5cad1287c/units/data_cleanup_testresults.go#L105) when we haven't deleted the full number of documents and an unintended consequence is that if some of the deletes race with each other we will also break. This should be fixed now, so we will actually be deleting several batches. In staging it was usually 4 batches in 50s.

We'd been breaking after [50s](https://github.com/evergreen-ci/evergreen/blob/c5abc064bc679a5063ab4cbe5ce14af5cad1287c/units/data_cleanup_testresults.go#L94). I changed this to 45s since, at least in staging, each batch was ~12s. This way we'll only start on a batch if we have 15 seconds to go and we won't have to cancel db operations by cancelling the ctx.

### Testing 
I deployed to staging and we appear to be deleting documents apace because we're no longer breaking when the deletes race.

We'll see how things go in prod. If someone advocates for it we could add an admin setting for the batch size and/or number of batches since it's possible that speeding up deletes will put extra load on the db.

### DIFF
--- a/model/test_log.go
+++ b/model/test_log.go
@@ -16,7 +16,11 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-const TestLogCollection = "test_logs"
+const (
+	TestLogCollection = "test_logs"
+
+	maxDeleteCount = 100000
+)
 
 type TestLog struct {
 	Id            string   `bson:"_id" json:"_id"`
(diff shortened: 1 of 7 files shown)


model/test_log.go
```diff
@@ -60,14 +64,31 @@ func FindOneTestLog(name, task string, execution int) (*TestLog, error) {
...
+}
+
 func DeleteTestLogsWithLimit(ctx context.Context, env evergreen.Environment, ts time.Time, limit int) (int, error) {
-	if limit > 100*1000 {
-		panic("cannot delete more than 100k documents in a single operation")
+	if limit > maxDeleteCount {
+		return 0, errors.Errorf("cannot delete more than %d documents in a single operation", maxDeleteCount)
+	}
+
+	docsToDelete, err := findAllTestLogs(db.Query(bson.M{TestLogIdKey: bson.M{"$lt": primitive.NewObjectIDFromTimestamp(ts).Hex()}}).WithFields(TestLogIdKey).Limit(limit))
```
johndaniels, line 80: To confirm: This will read up to 100,000 test docs into memory. That could easily be 10s of megabytes. I don't _know_ if that's a problem.

model/test_log.go
```diff
@@ -60,14 +64,31 @@ func FindOneTestLog(name, task string, execution int) (*TestLog, error) {
...
+		return 0, errors.Wrap(err, "getting docs to delete")
+	}
+
+	if len(docsToDelete) == 0 {
+		return 0, nil
 	}
 
-	ops := make([]mongo.WriteModel, limit)
-	for idx := 0; idx < limit; idx++ {
-		ops[idx] = mongo.NewDeleteOneModel().SetFilter(bson.M{TestLogIdKey: bson.M{"$lt": primitive.NewObjectIDFromTimestamp(ts).Hex()}})
```
johndaniels, line 70: This old code seems super weird... it was repeatedly "deleting one" with a range condition?

johndaniels (approved):

model/test_log.go
ybrill: That's a good point. Maybe around [37MB](https://go.dev/play/p/6Kqq-nDFv0G) for the results docs? 
It seems okay? We should only have one of these at a time.

model/test_log.go
ybrill: I think the problem it was solving for was that we wanted to delete a bunch of matching docs in parallel, but with a limit. Since deleteMany doesn't take a limit parameter we used an unordered bulk write which (ostensibly) executes in parallel.

PR #5947, opened 2022-09-19
EVG-17608 add some logging to alias copying

ablack12:
[EVG-17608 ](https://jira.mongodb.org/browse/EVG-17608 )

### Description 
I've been unable to replicate this bug, and I haven't been able to reason over how we would've copied some but not all aliases, given that we use catchers everywhere. I've added some logging to try to make this a bit more traceable.

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -11,6 +11,7 @@ import (
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 )
@@ -405,6 +406,11 @@ func UpsertAliasesForProject(aliases []ProjectAlias, projectId string) error {
 		}
 		catcher.Add(aliases[i].Upsert())
 	}
+	grip.Debug(message.WrapError(catcher.Resolve(), message.Fields{
+		"ticket":     "EVG-17608",
+		"message":    "problem getting aliases",
+		"project_id": projectId,
+	}))
 	return catcher.Resolve()
 }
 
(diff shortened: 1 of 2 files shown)


rest/data/project_settings.go
```diff
@@ -75,10 +75,15 @@ func CopyProject(ctx context.Context, env evergreen.Environment, opts CopyProjec
 	if err := event.CopyProjectSubscriptions(oldId, projectToCopy.Id); err != nil {
 		catcher.Wrapf(err, "copying subscriptions from project '%s'", oldIdentifier)
 	}
-	// set the same admin roles from the old project on the newly copied project.
+	// Set the same admin roles from the old project on the newly copied project.
 	if err := model.UpdateAdminRoles(projectToCopy, projectToCopy.Admins, nil); err != nil {
 		catcher.Wrapf(err, "updating admins for project '%s'", opts.NewProjectIdentifier)
 	}
+
+	// We can pass in an empty before here since the project is new
```
malikchaya2, line 83: Can you please elaborate on this comment (an empty what)?

model/project_aliases.go
```diff
@@ -405,6 +406,11 @@ func UpsertAliasesForProject(aliases []ProjectAlias, projectId string) error {
 		}
 		catcher.Add(aliases[i].Upsert())
 	}
+	grip.Debug(message.WrapError(catcher.Resolve(), message.Fields{
```
malikchaya2, line 409: Can we maybe save the results of catcher.Resolve instead of resolving it twice?

malikchaya2 (changes requested):

rest/data/project_settings.go
ablack12: Before is the name of the project settings variable. I can add that.

model/project_aliases.go
ablack12: I don't think it's particularly important since the idea will be to remove this, and this reads cleanly

rest/data/project_settings.go
malikchaya2: thanks!

malikchaya2 (approved):
approved with nit about comment

PR #5952, opened 2022-09-21
EVG-17928: Add ability to unset home volume

sophstad:
[EVG-17928](https://jira.mongodb.org/browse/EVG-17928)

### Description 
- Add ability to unset home volume on stopped host (for internal use only). This will facilitate the migration of volumes between hosts by allowing a volume to be detached without requiring termination of the host.

### Testing 
- Add unit test
- Spruce failure appears to be a flaky test

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2663,6 +2663,22 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
 	)
 }
 
+// UnsetHomeVolume disassociates a home volume from a (stopped) host.
+// This is for internal use, and should only be used on hosts that
+// will be terminated imminently; otherwise, the host will fail to boot.
+func (h *Host) UnsetHomeVolume() error {
+	err := UpdateOne(
+		bson.M{IdKey: h.Id},
+		bson.M{"$unset": bson.M{HomeVolumeIDKey: true}},
+	)
+	if err != nil {
+		return err
+	}
+
+	h.HomeVolumeID = ""
+	return nil
+}
+
 func (h *Host) SetHomeVolumeID(volumeID string) error {
 	h.HomeVolumeID = volumeID
 	return UpdateOne(bson.M{
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2669,7 +2669,7 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
 func (h *Host) UnsetHomeVolume() error {
 	err := UpdateOne(
 		bson.M{IdKey: h.Id},
-		bson.M{"$unset": bson.M{HomeVolumeIDKey: true}},
+		bson.M{"$set": bson.M{HomeVolumeIDKey: ""}},
 	)
 	if err != nil {
 		return err
--- a/model/host/host_test.go
+++ b/model/host/host_test.go
@@ -4516,6 +4516,38 @@ func TestAddVolumeToHost(t *testing.T) {
 	}, foundHost.Volumes)
 }
 
+func TestUnsetHomeVolume(t *testing.T) {
+	require.NoError(t, db.ClearCollections(Collection))
+	h := &Host{
+		Id:           "host-1",
+		HomeVolumeID: "volume-1",
+		Volumes: []VolumeAttachment{
+			{
+				VolumeID:   "volume-1",
+				DeviceName: "device-1",
+			},
+		},
+	}
+	assert.NoError(t, h.Insert())
+	assert.NoError(t, h.UnsetHomeVolume())
+	assert.Equal(t, "", h.HomeVolumeID)
+	assert.Equal(t, []VolumeAttachment{
+		{
+			VolumeID:   "volume-1",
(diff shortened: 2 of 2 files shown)


model/host/host.go
```diff
@@ -2663,6 +2663,22 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
 	)
 }
 
+// UnsetHomeVolume disassociates a home volume from a (stopped) host.
+// This is for internal use, and should only be used on hosts that
+// will be terminated imminently; otherwise, the host will fail to boot.
+func (h *Host) UnsetHomeVolume() error {
```
malikchaya2, line 2669: It looks like the function is not currently being called anywhere. Do we plan to add a button on the UI or an api route? 

Whatever we end up doing with it, I think we should log it somewhere so we can trace it and keep track of it.

sophstad: It'll be used [EVG-17485](https://jira.mongodb.org/browse/EVG-17485) — that's a huge ticket so figured I'd take this opportunity to break it up a tiny bit. The job that unsets the home volume will definitely have logging if that seems sufficient to you!

model/host/host.go
```diff
@@ -2663,6 +2663,22 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
 	)
 }
 
+// UnsetHomeVolume disassociates a home volume from a (stopped) host.
```
ablack12, line 2666: <3 this comment

model/host/host.go
```diff
@@ -2663,6 +2663,22 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
...
+func (h *Host) UnsetHomeVolume() error {
+	err := UpdateOne(
+		bson.M{IdKey: h.Id},
+		bson.M{"$unset": bson.M{HomeVolumeIDKey: true}},
+	)
+	if err != nil {
+		return err
+	}
+
+	h.HomeVolumeID = ""
```
ablack12, line 2678: I think it's generally considered okay to set this before the UpdateOne (and just return that error directly) since we return an error if it doesn't work and theoretically don't do stuff with the host anymore; we can still keep it this way if you prefer, just wanted to mention

ablack12 (approved):
Do we know if the existing AWS call to unset a volume already works with a home volume or if that still needs to be understood as part of the next ticket? (either is fine, just curious!)

model/host/host.go
Kimchelly: +1 for 💯 comment

model/host/host.go
```diff
@@ -2663,6 +2663,22 @@ func (h *Host) MarkShouldExpire(expireOnValue string) error {
 	)
 }
 
+// UnsetHomeVolume disassociates a home volume from a (stopped) host.
+// This is for internal use, and should only be used on hosts that
+// will be terminated imminently; otherwise, the host will fail to boot.
+func (h *Host) UnsetHomeVolume() error {
+	err := UpdateOne(
+		bson.M{IdKey: h.Id},
+		bson.M{"$unset": bson.M{HomeVolumeIDKey: true}},
```
Kimchelly, line 2672: Since the BSON tags for `HomeVolumeID` is not `omitempty`, the data model implies that the field should always be set, even if it's empty. Should this field instead be set to the empty string?

model/host/host.go
Kimchelly: I usually prefer to do it this way because there are some cases where there's some continue-on-error after a DB operation fails, so you know the in-memory copy is in agreement with what's in the DB.

model/host/host.go
sophstad: Based on the unit test it seems like these have the same effect? But `$set` does seem more correct, updated!

sophstad:
> Do we know if the existing AWS call to unset a volume already works with a home volume or if that still needs to be understood as part of the next ticket? (either is fine, just curious!)

@ablack12 It does work! AWS just [says](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.DetachVolume) "If an EBS volume is the root device of an instance, it can't be detached while the instance is running," so with a stopped instance we're good to go 👍

Kimchelly (approved):

model/host/host.go
ablack12: I think unset is a bit clearer personally, and is more correct for moving to an empty state; i think omitempty just causes it to default.

ablack12: (I see this was already merged so never mind 😅 but a future thought)

Kimchelly: I guess most of the difference here is how you query for a zero vlaue field later. Unset will delete the field from the doc (so you would query for it with `{$exists: false}`). Omitempty will unset if it's the Go zero value (so you would query for it with {"home_volume_id": ""}`. Since the field isn't `omitempty`, the Go driver will always set it to `""` if unspecified, unless you write a query that explicitly `$unset`s it. It seems easier to just have it be consistent with the data model and not have to query both for existence and for the empty string.

PR #5953, opened 2022-09-21
EVG-17946: fix auto update CLI context

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17946

### Description 
It seems like the auto upgrade CLI option for the user Evergreen YAMLs isn't working because it's checking the sub-command's CLI context flags (e.g. the flags immediately following `evergreen patch` or `evergreen commit-queue list`), but it should be checking the root CLI context flags (i.e. the flags immediately after `evergreen` but before any subcommand).

After this is deployed, we'll have to announce that any person using the `auto_upgrade_cli` option in their YAML with an `evergreen --version` older than 2022-09-21 (containing this fix) has to manually run `evergreen update` to get the correctly auto-updating CLI.

### Testing 
* Enabled `auto_upgrade_cli` in my YAML.
* Compiled the Evergreen CLI with an older CLI date date (2022-09-10) using what's currently in main, then ran `evergreen patch`. No auto update.
* Compiled the Evergreen CLI with an older CLI date (2022-09-10) and this change, then ran `evergreen patch`. Auto-update succeeded.

### DIFF
--- a/operations/before.go
+++ b/operations/before.go
@@ -102,7 +102,7 @@ var (
 	// Some functions that one would expect to return quickly have been omitted from having this as a 'before' function since downloading and installing
 	// takes time that would be cumbersome to the user (e.g. list functions, delete functions).
 	autoUpdateCLI = func(c *cli.Context) error {
-		confPath := c.String("conf")
+		confPath := c.Parent().String(confFlagName)
 		// we do not return an error in case of failure to find a valid config path because we do not want to block the underlying CLI operation.
 		if confPath == "" {
 			return nil
--- a/operations/before.go
+++ b/operations/before.go
@@ -121,8 +121,9 @@ var (
 		}
 
 		confPath := rootCtx.String(confFlagName)
-		// we do not return an error in case of failure to find a valid config path because we do not want to block the underlying CLI operation.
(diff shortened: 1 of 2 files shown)


operations/before.go
```diff
@@ -102,11 +102,30 @@ var (
...
+				break
+			}
+			rootCtx = parentCtx
+			i++
+		}
+
+		confPath := rootCtx.String(confFlagName)
 		// we do not return an error in case of failure to find a valid config path because we do not want to block the underlying CLI operation.
 		if confPath == "" {
 			return nil
```
ablack12, line 126: Should we log a warning in this case just to make sure we don't silently have some sort of problem with this again?

ablack12 (approved):
Oof this looks good to me. We may need to take a few avenues to unblock this since if users don't see this they might never notice what's happened. A message in users, an email to eng, maybe a temporary Evergreen banner?

operations/before.go
Kimchelly: Good idea, I added a warning if it fails to find the config path flag.

Kimchelly:
I think a message in the users Slack channel + email would be sufficient (assuming all Evergreen users get Evergreen announcement emails).

PR #5955, opened 2022-09-21
EVG-17698: consistently check execution number for tasks assigned to pods

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17698

### Description 
I made the task/pod lifecycle interaction a bit safer by ensuring that if the pod lifecycle tries to make a decision based on its currently running task, it also accounts for the execution number. Before, it would assume that the pod was definitely running the latest execution of a task, but there are some cases (such as restarting a task and aborting currently running tasks) where the pod might end up having an archived task execution assigned to it, in which case it should act based on its own task execution and not assume it's the latest. I'd like to avoid that entirely if possible by always acting on the exact task execution that the pod was assigned to run.

* Add new struct to hold information about current/previous tasks run by a pod (previous tasks are only relevant to task groups).
* Update REST API model for pods.
* Check both task ID and execution when looking for tasks assigned to pods.
* No-op in cases where the task execution running on the pod is already archived. Since this is an edge case, I added logging for cases where an archived task is still assigned to a pod for traceability.
* Remove a couple unneeded YAML tags on DB models.

Once this is merged, I'm going to drop the unique index on `running_task` in the DB. I don't think we need it anyways since assigning the next task is done in a transaction.

### Testing 
Updated existing unit tests for refactor and added a couple extra tests.

### DIFF
--- a/model/pod/pod.go
+++ b/model/pod/pod.go
@@ -37,11 +37,12 @@ type Pod struct {
 	TimeInfo TimeInfo `bson:"time_info,omitempty" json:"time_info,omitempty"`
 	// Resources are external resources that are owned and managed by this pod.
 	Resources ResourceInfo `bson:"resource_info,omitempty" json:"resource_info,omitempty"`
+	// TaskRuntimeInfo contains information about the tasks that a pod is
+	// assigned.
+	TaskRuntimeInfo TaskRuntimeInfo `bson:"task_runtime_info,omitempty" json:"task_runtime_info,omitempty"`
 	// AgentVersion is the version of the agent running on this pod if it's a
 	// pod that runs tasks.
 	AgentVersion string `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
-	// RunningTask is the ID of the task currently running on the pod.
-	RunningTask string `bson:"running_task,omitempty" json:"running_task,omitempty"`
 }
 
 // TaskIntentPodOptions represents options to create an intent pod that runs
@@ -568,10 +569,10 @@ func (o TaskContainerCreationOptions) IsZero() bool {
 type Secret struct {
 	// ExternalID is the unique external resource identifier for a secret that
 	// already exists in the secrets storage service.
-	ExternalID string `bson:"external_id,omitempty" json:"external_id,omitempty" yaml:"external_id,omitempty"`
+	ExternalID string `bson:"external_id,omitempty" json:"external_id,omitempty"`
 	// Value is the value of the secret. This is a cached copy of the actual
 	// secret value stored in the secrets storage service.
-	Value string `bson:"value,omitempty" json:"value,omitempty" yaml:"value,omitempty"`
+	Value string `bson:"value,omitempty" json:"value,omitempty"`
 }
 
 func (s Secret) hash() string {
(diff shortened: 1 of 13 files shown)


model/pod/pod.go
```diff
@@ -660,22 +677,26 @@ func (p *Pod) GetSecret() (*Secret, error) {
...
+	if p.TaskRuntimeInfo.RunningTaskID == taskID && p.TaskRuntimeInfo.RunningTaskExecution == taskExecution {
 		return nil
 	}
-	if p.RunningTask != "" {
-		return errors.Errorf("cannot set running task to '%s' when it is already set to '%s'", taskID, p.RunningTask)
+	if p.TaskRuntimeInfo.RunningTaskID != "" {
+		return errors.Errorf("cannot set running task to '%s' execution %d when it is already set to '%s' execution %d", taskID, taskExecution, p.TaskRuntimeInfo.RunningTaskID, p.TaskRuntimeInfo.RunningTaskExecution)
 	}
 
+	runningTaskIDKey := bsonutil.GetDottedKeyName(TaskRuntimeInfoKey, TaskRuntimeInfoRunningTaskIDKey)
```
hadjri, line 688: Can we make these constants in `pod/db.go`?

model/pod/pod.go
```diff
@@ -691,7 +712,8 @@ func (p *Pod) SetRunningTask(ctx context.Context, env evergreen.Environment, tas
 		return errors.New("pod was not updated")
 	}
 
-	p.RunningTask = taskID
+	p.TaskRuntimeInfo.RunningTaskID = taskID
```
hadjri, line 715: We can assign the whole `TaskRuntimeInfo` here for clarity rather than setting both fields individually since the whole struct is just a task Id and execution.

model/pod/pod.go
```diff
@@ -701,22 +723,36 @@ func (p *Pod) SetRunningTask(ctx context.Context, env evergreen.Environment, tas
...
-	if p.RunningTask == "" {
+	if p.TaskRuntimeInfo.RunningTaskID == "" {
 		return nil
 	}
 
-	if err := UpdateOne(bson.M{
-		IDKey:          p.ID,
-		RunningTaskKey: p.RunningTask,
-	}, bson.M{
+	runningTaskIDKey := bsonutil.GetDottedKeyName(TaskRuntimeInfoKey, TaskRuntimeInfoRunningTaskIDKey)
```
hadjri, line 730: Same comment

model/pod/pod.go
```diff
@@ -701,22 +723,36 @@ func (p *Pod) SetRunningTask(ctx context.Context, env evergreen.Environment, tas
...
-			RunningTaskKey: 1,
+			runningTaskIDKey:        1,
+			runningTaskExecutionKey: 1,
 		},
 	}); err != nil {
 		return errors.Wrap(err, "clearing running task")
 	}
 
-	p.RunningTask = ""
+	p.TaskRuntimeInfo.RunningTaskID = ""
```
hadjri, line 754: Same comment here

units/pod_termination.go
```diff
@@ -214,3 +206,39 @@ func (j *podTerminationJob) fixStrandedTasks(ctx context.Context) error {
 
 	return nil
 }
+
+func (j *podTerminationJob) fixStrandedRunningTask(ctx context.Context) error {
+	if j.pod.TaskRuntimeInfo.RunningTaskID == "" {
+		return nil
```
hadjri, line 212: A grip statement in this case may be good for monitoring

rest/route/pod_agent.go
```diff
@@ -372,6 +355,60 @@ func (h *podAgentNextTask) findDispatcher() (*dispatcher.PodDispatcher, error) {
 	return pd, nil
 }
 
+func (h *podAgentNextTask) checkAndRedispatchRunningTask(ctx context.Context, p *pod.Pod) gimlet.Responder {
+	t, err := task.FindOneIdAndExecution(p.TaskRuntimeInfo.RunningTaskID, p.TaskRuntimeInfo.RunningTaskExecution)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "getting running task '%s' execution %d", p.TaskRuntimeInfo.RunningTaskID, p.TaskRuntimeInfo.RunningTaskExecution))
+	}
+	if t == nil {
```
hadjri, line 363: Should we validate that the task's status indicates that it should be running as well?

units/pod_termination.go
Kimchelly: This is the normal case since the pod would only be terminated while still running a task in a few edge cases. I think it's still worth tracking when we clear and reset stranded tasks though, so I added a log for cases where we clear and reset tasks that are stale or stranded.

model/pod/pod.go
Kimchelly: I think we normally put `bsonutil.GetDottedKeyName` for nested key names. I'm not that inclined to add it since people are used to combining dotted key names as they need it for their query.

model/pod/pod.go
hadjri: Ok, sgtm

hadjri (approved):

PR #5960, opened 2022-09-22
EVG-17612 add stepback improvements and logging

ablack12:
[EVG-17612](https://jira.mongodb.org/browse/EVG-17612)

### Description 
(New PR bc wrong branch)

Added some stepback improvements that will be built on in future tickets:
1) Add stepback depth to log how far back we stepback, to consider adding a limit to this ([EVG-17949](https://jira.mongodb.org/browse/EVG-17949))
2) Add disable stepback flag (this will need to be added to the new UI: [EVG-17951](https://jira.mongodb.org/browse/EVG-17951))
3) Add variant/task to "cancel stepback" option, with the assumption that there's usually one thing that's going crazy, and if there are more they can use the new "turn off stepback" option (this will need to be added to the new UI: [EVG-17947](https://jira.mongodb.org/browse/EVG-17947))

### Testing 
Added tests / verifying existing tests pass. Will test stepback depth on staging.


model/task/task.go
```diff
@@ -102,6 +102,11 @@ type Task struct {
 	Activated                bool   `bson:"activated" json:"activated"`
 	ActivatedBy              string `bson:"activated_by" json:"activated_by"`
 	DeactivatedForDependency bool   `bson:"deactivated_for_dependency" json:"deactivated_for_dependency"`
+
+	// TaskDepth indicates how far into stepback this task was activated.
```
Kimchelly, line 106: StepbackDepth indicates how far...

Should we make a note that a task with no stepback has depth=0, and the first stepback has depth=1 to clear possible ambiguity over what is the initial depth for stepback?

model/generate.go
```diff
@@ -495,7 +514,8 @@ func (g *GeneratedProject) findTasksAndVariantsWithSpecificActivations(requester
 		batchTimeTasks := []string{}
 		for _, bvt := range bv.Tasks {
 			if isStepbackTask(g.Task, bv.Name, bvt.Name) {
-				res.stepbackTasks[bv.Name] = append(res.stepbackTasks[bv.Name], bvt.Name)
+				stepbackInfo := specificStepbackInfo{task: bvt.Name, depth: g.Task.StepbackDepth + 1}
```
Kimchelly, line 517: Is it worth adding a note that depth + 1 is intended only for new tasks being created by stepback? Or can we move the `+ 1` part to [here](https://github.com/evergreen-ci/evergreen/blob/211ebdf33de10de18951e2ed3adcfb1a3ad870a6/model/lifecycle.go#L1247) where it's more immediately relevant?

model/task/db.go
```diff
@@ -1785,7 +1707,39 @@ func FindActivatedStepbackTasks(projectId string) ([]Task, error) {
...
-
+		return nil, err
 	}
 	return tasks, nil
 }
+
+// FindActivatedStepbackTaskByName queries for running/scheduled stepback tasks with
+// matching build variant and task name.
+func FindActivatedStepbackTaskByName(projectId string, variantName string, taskName string) (*Task, error) {
+	t, err := FindOne(db.Query(bson.M{
```
Kimchelly, line 1718: Can this be simplified down to `return FindOne(...`?

model/task/task.go
```diff
@@ -1419,6 +1425,29 @@ func DeactivateStepbackTasksForProject(projectId, caller string) error {
...
 
+// DeactivateStepbackTaskForProject deactivates and aborts the matching stepback task.
+// Will be used instead of LegacyDeactivateStepbackTasksForProject as part of EVG-17947
+func DeactivateStepbackTaskForProject(projectId, buildVariantName, taskName, caller string) error {
+	t, err := FindActivatedStepbackTaskByName(projectId, buildVariantName, taskName)
+	if err != nil {
+		return err
+	}
+	if t == nil {
+		return errors.New("no stepback task found")
```
Kimchelly, line 1436: Should this error message include some info like the task name or whatnot so we can trace when this happens?

model/task/task.go
```diff
@@ -1419,6 +1425,29 @@ func DeactivateStepbackTasksForProject(projectId, caller string) error {
 	return nil
 }
 
+// DeactivateStepbackTaskForProject deactivates and aborts the matching stepback task.
+// Will be used instead of LegacyDeactivateStepbackTasksForProject as part of EVG-17947
+func DeactivateStepbackTaskForProject(projectId, buildVariantName, taskName, caller string) error {
```
Kimchelly, line 1430: Since this is no longer for an entire project, can we call this something else, like `DeactivateStepbackTask`?

model/task/task.go
```diff
@@ -4060,7 +4089,7 @@ func HasMatchingTasks(versionID string, opts HasMatchingTasksOptions) (bool, err
 }
 
 func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions) ([]bson.M, error) {
-	var match bson.M = bson.M{}
+	var match = bson.M{}
```
Kimchelly, line 4092: nit: `match := bson.M{}`

model/task_lifecycle.go
```diff
@@ -407,6 +415,17 @@ func getStepback(taskId string) (bool, error) {
 	if t == nil {
 		return false, errors.Errorf("task '%s' not found", taskId)
 	}
+	projectRef, err := FindMergedProjectRef(t.Project, "", false)
+	if err != nil {
+		return false, errors.Wrapf(err, "finding merged project ref for task '%s'", taskId)
+	}
+	if projectRef == nil {
+		return false, errors.Errorf("project 'for task '%s' not found", taskId)
```
Kimchelly, line 423: nit: remove extra apostrophe before "for".

model/task_lifecycle.go
```diff
@@ -869,6 +887,7 @@ func evalStepback(t *task.Task, caller, status string, deactivatePrevious bool)
 		}
 
 		if t.IsPartOfSingleHostTaskGroup() {
+			fmt.Println("is this part of a task group?")
```
Kimchelly, line 890: Remove print.

Kimchelly (changes requested):
Are users going to have to enter each build variant/task they want to step back one by one in the UI when it's updated, or will we support some kind of batch selection of build variants/tasks?

model/task/task.go
ablack12: Sure

model/task/task.go
ablack12: blame my editor for this change :fistshake:

ablack12 (commented):
> Are users going to have to enter each build variant/task they want to step back one by one in the UI when it's updated, or will we support some kind of batch selection of build variants/tasks?

Yes users will have to enter each variant/task, but I think this is usually used when one particular task is going crazy. Otherwise they can just use the project page to temporarily disable stepback at the project level until the tasks shh which would be much less expensive than us trying to batch query.

model/task/task.go
Kimchelly: nit: update TaskDepth -> StepbackDepth

Kimchelly (commented):
I think a couple comments got missed? I left them unresolved because I don't know if you saw them.

ablack12:
>I think a couple comments got missed? I left them unresolved because I don't know if you saw them.

Ah thanks idk why I didn't see those in the file view. Will re-request later.

model/generate.go
ablack12: I think it's clear enough here since we have the line `isStepbackTask` right above and I'd rather what's passed in via this info is the depth to be used, but I can add a comment?

ablack12: I think it's clear enough here since we have the line `isStepbackTask` right above that this is only intended for stepback; however, I think it's actually a mistake to use +1 here since generated tasks should have the same stepback depth as their generator (since this isn't a deeper stepback).

model/task/db.go
ablack12: No because FindOne takes in a query; that being said this is the DB file so the lack of abstraction is I think more okay here

model/generate.go
ablack12: Tested this behavior ^ and confirmed shouldn't add +1

ablack12:
@Kimchelly thanks for waiting, this is ready for re-review now that staging testing is done.

Kimchelly (approved):

PR #5963, opened 2022-09-23
EVG-17963: remove auto-upgrade CLI check from 'evergreen evaluate'

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17963

### Description
Some people use the Evergreen CLI in CI testing to do automated stuff. `evergreen evaluate` doesn't require the `~/.evergreen.yml` file to exist to evaluate the YAML, so it should be fine to remove the auto-upgrading functionality temporarily. We can evaluate re-adding it and possibly improving the way it logs to interfere with scripting in a follow-up.

I bumped both the CLI and agent versions because this command is being used for CI tests, so we have to force the agents to roll over to the latest version.

### Testing 
N/A

### DIFF
--- a/config.go
+++ b/config.go
@@ -33,10 +33,10 @@ var (
 	BuildRevision = ""
 
 	// Commandline Version String; used to control auto-updating.
-	ClientVersion = "2022-09-21"
+	ClientVersion = "2022-09-23"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2022-09-14"
+	AgentVersion = "2022-09-23"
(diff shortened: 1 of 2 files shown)


ablack12 (approved):
fwiw i'm fine with just not adding it back to evaluate. We may want to audit what it _is_ added to in case there are other commands that don't require this path that are erroring on it.

Kimchelly:
Sure, I can do a follow-up audit for other commands that check for auto-upgrades.

PR #5964, opened 2022-09-26
EVG-17515 Add id to task log

khelif96:
[EVG-17515](https://jira.mongodb.org/browse/EVG-17515)

### Description 
Make some changes to test data to accommodate parsley e2e tests


service/task.go
```diff
@@ -876,6 +876,38 @@ func (uis *UIServer) testLog(w http.ResponseWriter, r *http.Request) {
 				"message":   "problem getting buildlogger logs",
 			}))
 		}
+	} else {
+		// Fallback to legacy test results
+		// Direct link to a log document in the database.
+		testLog, err = model.FindOneTestLog(testName, taskID, taskExec)
```
julianedwards, line 882: In your case of running this locally, will the test results also live in the Evg DB?

khelif96: yes they will! The testresults query also has a similar fallback.

ablack12:
Will this be related to https://jira.mongodb.org/browse/EVG-17974 / should we link a TODO here with that ticket since this may be temporary?

service/task.go
julianedwards: Is it possible then to pass in the `logId` to the rest route so you do not need this third clause in the `if/else` statement?

khelif96: No this change was meant to replicate the same behavior as the cedar route. A user will only have a task id, execution and a test name.

khelif96:
> Will this be related to https://jira.mongodb.org/browse/EVG-17974 / should we link a TODO here with that ticket since this may be temporary?

Yeah i think so just linked it!

service/task.go
julianedwards: Ah ok, makes sense then!

julianedwards (approved):
LGTM % nit

service/task.go
```diff
@@ -876,6 +876,39 @@ func (uis *UIServer) testLog(w http.ResponseWriter, r *http.Request) {
 				"message":   "problem getting buildlogger logs",
 			}))
 		}
+	} else {
+		// TODO: EVG-17974 Replace this with a more permanent option
+		// Fallback to legacy test results
```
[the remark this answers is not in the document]
ablack12: FYI this is only recommended by the style guide, not required https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit#heading=h.vi2qpfwvlklv

PR #5966, opened 2022-09-27
EVG-17485: Add volume migration job

sophstad:
[EVG-17485](https://jira.mongodb.org/browse/EVG-17485)

### Description 
Adds a job to migrate a volume from one host to another. A GraphQL mutation will be added in a subsequent PR to expose this functionality to the UI. Based on a discussion with Mohamed, I slightly changed the order of operations in the job from the [spec](https://docs.google.com/document/d/1je3-CHD0qL-B0u0VdvPFKYrwi7d4woBnEdsSG1PFtMU/edit?usp=sharing) to simplify the process. It's now as follows:

1. Stop initial host
2. Detach volume from initial host (as enabled by https://github.com/evergreen-ci/evergreen/pull/5952)
3. Start new host with volume attached
4. Terminate initial host

The job will attempt to restart the initial host with the volume attached if any of the subsequent operations fail.

This is also my first time working with Amboy so any style/best practices notes are very welcome! Thank you in advance 😸 

### Testing 
- Write unit tests for Amboy job
- I've run successful migrations on staging


units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (j *volumeMigrationJob) Run(ctx context.Context) {
+	defer j.finishJob(ctx)
+
+	if err := j.populateIfUnset(); err != nil {
+		j.AddRetryableError(err)
+		return
+	}
+
+	if err := j.volume.SetMigrating(true); err != nil {
```
ablack12, line 83: What if migrating is already set to true? (I imagine this could happen if the job got stopped unexpectedly.) I don't think this would error but would the rest of the job work as expected?

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+			"host_id": j.InitialHostID,
+		})
+		j.InitialHostStopped = true
+		return true, false
+	}
+
+	// If host is not stopped, verify whether the job is still in progress and error if it completed without stopping
+	stopJob, foundJob := j.env.RemoteQueue().Get(ctx, j.SpawnhostStopJobID)
+	if !foundJob {
+		j.AddRetryableError(errors.Errorf("spawnhost-stop job '%s' could not be found", j.SpawnhostStopJobID))
```
ablack12, line 173: We don't remove the SpawnhostStopJobID so theortically we would just loop forever when we retry here right?

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+		return
+	}
+
+	if !j.InitialHostStopped {
+		initialHostStopped, hasErr := j.stopInitialHost(ctx)
+		if hasErr {
+			return
+		}
+		if !initialHostStopped {
+			j.AddRetryableError(errors.Errorf("initial host '%s' not yet stopped", j.InitialHostID))
```
ablack12, line 94: Rather than adding a retryableError here if we return false from stopInitialHost, does it make sense to just do it within the stopInitialHost function itself? It looks like we only return false, false in two places

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func NewVolumeMigrationJob(env evergreen.Environment, volumeID string, modifyOptions cloud.SpawnOptions, ts string) amboy.Job {
+	j := makeVolumeMigrationJob()
+	j.env = env
+	j.SetID(fmt.Sprintf("%s.%s.%s", volumeMigrationJobName, volumeID, ts))
+	j.SetScopes([]string{fmt.Sprintf("%s.%s", volumeMigrationJobName, volumeID)})
+	j.SetEnqueueAllScopes(true)
+	j.VolumeID = volumeID
+	j.ModifyOptions = modifyOptions
```
ablack12, line 65: Where do these ModifyOptions come from? The new or old host? I think this is a lot to store; it'd be preferable if we could just look it up when setting up the job. (Although I'm just thinking of convention; I'm not sure if there's a reason or it's just happenstance that we don't usually store much in the job @Kimchelly )

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+	registry.AddJobType(volumeMigrationJobName,
+		func() amboy.Job { return makeVolumeMigrationJob() })
+}
+
+type volumeMigrationJob struct {
+	job.Base      `bson:"job_base" json:"job_base" yaml:"job_base"`
+	VolumeID      string             `bson:"volume_id" yaml:"volume_id"`
+	ModifyOptions cloud.SpawnOptions `bson:"modify_options" json:"modify_options" yaml:"modify_options"`
+
+	VolumeDetached        bool   `bson:"volume_detached"`
```
ablack12, line 34: Since these aren't stored with the job do they need to be public? (I might be misunderstanding what's going on here though)

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+		j.AddRetryableError(err)
+		return
+	}
+
+	if err := j.volume.SetMigrating(true); err != nil {
+		j.AddRetryableError(err)
+		return
+	}
+
+	if !j.InitialHostStopped {
```
ablack12, line 88: Do we prevent users from stopping/starting an instance if migrating is true? If not we may want to check this from the host document directly since things could've changed

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			j.AddError(err)
+			return
+		}
+		j.volume = volume
+	}
+
+	// Avoid recreating new host on retry
+	if j.NewHostID == "" {
+		if hasErr := j.startNewHost(ctx); hasErr {
```
ablack12, line 123: I think it would be preferable to check j.HasErrors() after helper functions instead of returning booleans at each point

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+			return
+		}
+	}
+
+	// Terminate initial host
+	if !j.InitialHostTerminated {
+		ts := utility.RoundPartOfMinute(1).Format(TSFormat)
+		terminateJob := NewSpawnHostTerminationJob(j.initialHost, evergreen.User, ts)
```
ablack12, line 131: I'm a little nervous about actually terminating the initial host. Is there a world in which the migration could be problematic or the image not suitable for the user or etc and they may want to go back to their old instance? (I have a fuzzy memory of discussing this on one of the docs but not sure if that was real haha). I wonder if we might instead want to just set the host to expirable in a few days or something to give them a chance to go back.

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Avoid recreating new host on retry
+	if j.NewHostID == "" {
+		if hasErr := j.startNewHost(ctx); hasErr {
+			return
+		}
+	}
+
+	// Terminate initial host
+	if !j.InitialHostTerminated {
```
ablack12, line 129: Do we need hostTerminated? It seems like if we'd successfully terminated the host we wouldn't even get this far because the job would be finished.

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+
+	if stopJob.Status().Completed {
+		j.AddError(errors.Errorf("host '%s' failed to stop", j.SpawnhostStopJobID))
+		return false, true
+	}
+
+	return false, false
+}
+
+// startNewHost attempts to start a new host with the volume attached
```
ablack12, line 185: nit: complete sentence comments should end with punctuation for readability

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+	// Ensure volume has been detached
+	if j.volume.Host == j.InitialHostID {
+		j.AddRetryableError(errors.New("volume still attached to host"))
+		return true
+	}
+
+	j.ModifyOptions.HomeVolumeID = j.VolumeID
+	intentHost, err := cloud.CreateSpawnHost(ctx, j.ModifyOptions, j.env.Settings())
+	if err != nil {
+		j.AddError(errors.Wrapf(err, "creating new intent host"))
```
ablack12, line 197: This might be a good one to retry on because of AWS capacity issues or etc

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+		return true
+	}
+
+	if err := intentHost.Insert(); err != nil {
+		j.AddError(errors.Wrap(err, "inserting new intent host"))
+		return true
+	}
+	grip.Info(message.Fields{
+		"message":  "new intent host created",
+		"attempts": j.RetryInfo().CurrentAttempt,
```
ablack12, line 207: Can we add host IDs / job IDs etc to these logs? Otherwise they're not going to be super useful and we should just remove them.

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+	})
+	j.NewHostID = intentHost.Id
+	return false
+}
+
+// finishJob marks the job as completed and attempts some additional cleanup if this is the job's final attempt
+func (j *volumeMigrationJob) finishJob(ctx context.Context) {
+	// Amboy's ShouldRetry() method does not account for reaching MaxAttempts, so manually determine if the current attempt is the last
+	retryInfo := j.RetryInfo()
+	isFinalAttempt := retryInfo.CurrentAttempt == (retryInfo.MaxAttempts - 1)
```
ablack12, line 217: Can we use > instead of == just to be safe? (for example if 227 fails it seems like we could restart again although I don't know the amboy order of operations well)

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+			j.AddRetryableError(err)
+			return
+		}
+
+	}
+
+	j.MarkComplete()
+}
+
+// restartInitialHost restarts the stopped host with the volume attached
```
ablack12, line 236: Also punctuation here for complete sentences

units/migrate_volume_job.go
```diff
@@ -0,0 +1,298 @@
...
+	ts := utility.RoundPartOfMinute(1).Format(TSFormat)
+	if err := j.env.RemoteQueue().Put(ctx, NewSpawnhostStartJob(j.initialHost, evergreen.User, ts)); err != nil {
+		j.AddError(errors.Wrap(err, "enqueuing job to restart initial host"))
+		return true
+	}
+	return false
+
+}
+
+func (j *volumeMigrationJob) getVolume() (*host.Volume, error) {
```
ablack12, line 258: This helper function ultimately only moves one if statement out of the main loop so I'm not certain it's really needed

units/migrate_volume_job.go
sophstad: These ModifyOptions are the options that a user has configured via a modal for their new host. I'm adding a GraphQL mutation in [EVG-17486](https://jira.mongodb.org/browse/EVG-17486) that will have a similar workflow to the [SpawnHost](https://github.com/evergreen-ci/evergreen/blob/a10726bcd4549772aa18985e2d43cb27d90b87d2/graphql/mutation_resolver.go#L618) mutation; the new volume migration mutation will accept an [APIHost](https://github.com/evergreen-ci/evergreen/blob/f0573cc00ff1af555ca4c7a2d342239ac062b3d6/rest/model/host.go#L15-L38) struct that is converted to the SpawnOptions type.

Since the job is kicked off from a mutation that the user calls I'm not sure if there's anywhere that makes sense to store this information, but let me know if I'm missing anything!

units/migrate_volume_job.go
sophstad: To the best of my understanding, we need to make these structs public so that they are preserved on subsequent attempts of the job (this is a characteristic of using a remote queue). See "A Note on Idempotency" in the [Amboy doc](https://docs.google.com/document/d/1FJ1cj60H6gQ3Cw0M-Wx149YSvRTXbWgSLZePhpiVlII/edit?usp=sharing). If the `VolumeDetached` field isn't preserved in the job document, we won't be able to pick up the job in its current state.

units/migrate_volume_job.go
sophstad: Yep it would work! The main purpose of the flag is so that we can label the job as migrating on the UI.

I imagine if a job was randomly stopped, that would be a one-off error that we could look into. But not sure about what our normal approach is to random scenarios like that, so let me know if something else would make more sense!

units/migrate_volume_job.go
sophstad: This is a great point 👍

units/migrate_volume_job.go
sophstad: Sure—assuming you meant >= so that we do catch the last attempt!

units/migrate_volume_job.go
sophstad: True, the only scenarios where we'd end up here again would be a failure to put the job on the queue or an error unsetting the migrating flag. But I think the scoping of `NewSpawnHostTerminationJob` would make this operation idempotent anyway so I'll remove.

units/migrate_volume_job.go
sophstad: Yes – changed to `AddError`!

units/migrate_volume_job.go
sophstad: That's a good point—I guess we have to assume anything is possible. I like the idea of setting the expiry date, or we could make it an option for users to select whether or not the host terminates?

units/migrate_volume_job.go
```diff
@@ -0,0 +1,286 @@
...
+	if err := intentHost.Insert(); err != nil {
+		j.AddError(errors.Wrap(err, "inserting new intent host"))
+		return
+	}
+	grip.Info(message.Fields{
+		"message":        "new intent host created",
+		"job_id":         j.ID(),
+		"intent_host_id": intentHost.Id,
+	})
+	j.NewHostID = intentHost.Id
```
Kimchelly, line 206: Can this be replaced by a `FindOne` to check if an (intent) host has the volume ID rather than setting this field?

units/migrate_volume_job.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// finishJob marks the job as completed and attempts some additional cleanup if this is the job's final attempt.
+func (j *volumeMigrationJob) finishJob(ctx context.Context) {
```
Kimchelly, line 210: Should this job log either to Splunk or to the event log when it fails to migrate the volume?

units/migrate_volume_job.go
```diff
@@ -0,0 +1,286 @@
...
+	// Amboy's ShouldRetry() method does not account for reaching MaxAttempts, so manually determine if the current attempt is the last.
+	retryInfo := j.RetryInfo()
+	isFinalAttempt := retryInfo.CurrentAttempt >= (retryInfo.MaxAttempts - 1)
+	if !retryInfo.ShouldRetry() || isFinalAttempt {
+		// If a new host was never created, restart initial host with volume attached.
+		if j.NewHostID == "" {
+			if hasErr := j.restartInitialHost(ctx); hasErr {
```
Kimchelly, line 217: Do we want to restart the initial host in case the migration fails? I imagine that the user might just want to retry the migration, so re-stopping the just-restarted host might be redundant.

Kimchelly (changes requested):

units/migrate_volume_job.go
sophstad: I had this thought too, but some spec reviewers thought it'd be more useful to leave users with a usable host in case of failure, so I went with this.

Kimchelly: I agree that a usable host would be nice, but I think it might be hard to resuscitate the old host at this point. Even if it can restart, by the time it's reached this step, it's already detached the home volume which probably has most of the host's files, so the user would have to manually re-attach the volume as well.

ablack12: This job does reattach the home volume in this step though right?

Kimchelly: Hmm yeah, looking at it, I guess it does try to attach the volume. I'm trying to avoid this sort of expensive end-of-job fixup operations because it's not a reliable mechanism to fix the old host.

ablack12: Would there be a way for users to reattach the home volume manually in error @sophstad ? Or is it something we'd consider adding a button for?

sophstad: > I'm trying to avoid this sort of expensive end-of-job fixup operation though because it's not a reliable mechanism to fix the old host and it's probably happening in a case that can't be fixed

This makes a lot of sense! And @ablack12 if we don't go ahead with the restarting step, we could allow users to remount their volume with the normal UI mount button. The initial host would be unusable without the volume attached at this point (see [EVG-7879](https://jira.mongodb.org/browse/EVG-7879)), so it would maybe make sense to just terminate it?

ablack12: I'd rather we not terminate it, if they are still able to remount their volume. Maybe we could set the expiration date back to expirable though so it doesn't stick around if they don't want it, similar to if the job does actually work? Also I wonder if we could add a UI warning label thing like "no home volume attached" (that might be too far though; it would also be sufficient to just have trying to start it again return an error)

ablack12: People do get really testy about us deleting their hosts when they don't want them deleted so I think we'd better err on the side of caution with it.

sophstad: Annie is out now, but to sum up the takeaway here, I'm going to:

- Remove the recovery portion of the job that attempts to restart the initial host in case of failure. The job will finish with a stopped initial host and unmounted volume that can then be attached and restarted. I don't think we need to touch the expiry date as it seems inconsequential/maybe more confusing.
- Remove the final termination-upon-success step, and instead set the initial host to expire in 24 hours (as discussed in [this comment](https://github.com/evergreen-ci/evergreen/pull/5966#discussion_r982824998)

@Kimchelly let me know if you have any objections!

units/migrate_volume_job.go
```diff
@@ -0,0 +1,286 @@
...
+
+func (j *volumeMigrationJob) Run(ctx context.Context) {
+	defer j.finishJob(ctx)
+
+	if err := j.populateIfUnset(); err != nil {
+		j.AddRetryableError(err)
+		return
+	}
+
+	if err := j.volume.SetMigrating(true); err != nil {
```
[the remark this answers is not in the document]
sophstad: Its main purpose is to be used by the UI! Since most of the operations are idempotent and we're checking the host/volume documents throughout the job there is nothing that truly needs to be locked via this field.

units/migrate_volume_job.go
Kimchelly: This sounds good to me! I have two questions about the expiration:
* If it's set to 24 hours, does the user get a notification that their host will be expired soon? It would be nice to make sure that the user knows it's going to go away soon.
* Is it worth logging an event for the host to tell the user that it failed to migrate and will be expired soon?

units/migrate_volume_job.go
Kimchelly: I don't think there's an issue with storing large-ish options structs in the Amboy job. This is a normal way to set up an Amboy job when all you have are options to configure the new resource.

units/migrate_volume_job.go
```diff
@@ -0,0 +1,257 @@
...
+			return
+		}
+		if newHost == nil {
+			event.LogVolumeMigrationFailed(j.InitialHostID)
+			grip.Error(message.Fields{
+				"message": "volume failed to migrate",
+				"job_id":  j.ID(),
+				"host_id": j.InitialHostID,
+			})
+		}
```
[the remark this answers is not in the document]
sophstad: Good point, added checking for either scenario!

Kimchelly (approved):

PR #5967, opened 2022-09-28
EVG-17655: add logging for container task dashboard

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17655

### Description 
* Add metadata information for logs that are relevant to the [container task health dashboard](https://mongodb.splunkcloud.com/en-US/app/search/evergreen_container_task_health). I added a `usage` field as a signal that these logs are depended on to populate the dashboard.
* Add job to periodically collect stats on pods and container tasks. This has the equivalent purpose to the [host stats collector](https://github.com/evergreen-ci/evergreen/blob/c5abc064bc679a5063ab4cbe5ce14af5cad1287c/units/stats_host.go).

### Testing 
Added unit tests.

### DIFF
--- a/units/stats_pod.go
+++ b/units/stats_pod.go
@@ -1,0 +1,104 @@
+package units
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/model/pod"
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+)
+
+const podStatsCollectorJobName = "pod-stats-collector"
+
+func init() {
+	registry.AddJobType(podStatsCollectorJobName, func() amboy.Job {
+		return makePodStatsCollector()
+	})
+}
+
+type podStatsCollector struct {
+	job.Base `bson:"job_base" json:"job_base" yaml:"job_base"`
+}
+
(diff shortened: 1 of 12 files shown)


units/stats_pod.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return j
+}
+
+func (j *podStatsCollector) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		j.AddError(err)
```
hadjri, line 55: Can we instead use `errors.Wrap(err, "getting service flags")`?

units/stats_pod.go
```diff
@@ -0,0 +1,104 @@
...
+			"job_type": j.Type().Name,
+		})
+		return
+	}
+
+	j.AddError(j.logStats())
+}
+
+func (j *podStatsCollector) logStats() error {
+	statuses := []pod.Status{
```
hadjri, line 73: If we are hardcoding the list of statuses to check should we just pass no parameters to `GetStatsByStatus` and define the list there instead?

hadjri: Or do we plan on potentially using this with different inputs in the future? Not a CR, just a suggestion.

hadjri (approved):
LGTM with nit / suggestion. Apols for the wait

units/stats_pod.go
Kimchelly: Done.

units/stats_pod.go
Kimchelly: I mostly did this so I could zero out tha values below this in the log message and use the same statuses between the find and the log messages. We might change the inputs, but not that frequently.

PR #5969, opened 2022-09-28
EVG-17678 Require auth for all routes

hadjri:
[EVG-17678](https://jira.mongodb.org/browse/EVG-17678)

### Description 
Introduced a new auth middleware that wraps `requireUser` and `requireLogin` for UI pages and added them to all routes that can currently be accessed by any 3rd party without any error.  An admin service flag can toggle whether or not these auth checks are active or not for the updated routes.
### Testing 
Deployed to staging and confirmed that I can no longer visit key UI pages and call REST routes from my personal laptop.  Did not see any concerning errors when running new tasks and versions with the change deployed.


config_serviceflags.go
```diff
@@ -35,6 +35,7 @@ type ServiceFlags struct {
 	CloudCleanupDisabled            bool `bson:"cloud_cleanup_disabled" json:"cloud_cleanup_disabled"`
 	ContainerConfigurationsDisabled bool `bson:"container_configurations_disabled" json:"container_configurations_disabled"`
 	SlackAppDisabled                bool `bson:"slack_app_disabled" json:"slack_app_disabled"`
+	RequireAuthAllRoutesDisabled    bool `bson:"require_auth_all_routes_disabled" json:"require_auth_all_routes_disabled"`
```
ablack12, line 38: Can we flip the wording of this route so that the false state is the default that we want? (since we do want this to be disabled to begin with) Like lightAuthDisabled or something. That or we can insert true into prod before we deploy so it'll start in the disabled state.

rest/route/service.go
```diff
@@ -33,6 +33,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	sc.SetURL(opts.URL)
 
 	// Middleware
+	requireUserToggleable := NewRequireAuthHandler()
```
ablack12, line 36: Can we link a TODO for a ticket to eventually replace these with require user?

rest/route/service.go
```diff
@@ -130,19 +131,19 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
...
+	app.AddRoute("/distros/{distro_id}/client_urls").Version(2).Get().Wrap(requireUserToggleable).RouteHandler(makeGetDistroClientURLs(env))
 	app.AddRoute("/distros/{distro_id}/execute").Version(2).Patch().Wrap(editHosts).RouteHandler(makeDistroExecute(env))
 	app.AddRoute("/distros/{distro_id}/icecream_config").Version(2).Patch().Wrap(editHosts).RouteHandler(makeDistroIcecreamConfig(env))
 	app.AddRoute("/distros/{distro_id}/setup").Version(2).Get().Wrap(editDistroSettings).RouteHandler(makeGetDistroSetup())
 	app.AddRoute("/distros/{distro_id}/setup").Version(2).Patch().Wrap(editDistroSettings).RouteHandler(makeChangeDistroSetup())
 
-	app.AddRoute("/hooks/github").Version(2).Post().RouteHandler(makeGithubHooksRoute(sc, opts.APIQueue, opts.GithubSecret, settings))
-	app.AddRoute("/hooks/aws").Version(2).Post().RouteHandler(makeEC2SNS(env, opts.APIQueue))
-	app.AddRoute("/hooks/aws/ecs").Version(2).Post().RouteHandler(makeECSSNS(env, opts.APIQueue))
+	app.AddRoute("/hooks/github").Version(2).Post().Wrap(requireUserToggleable).RouteHandler(makeGithubHooksRoute(sc, opts.APIQueue, opts.GithubSecret, settings))
```
ablack12, line 140: Hmm I'm actually concerned this will break our integrations 😬 Right now we validate in Parse for example https://github.com/evergreen-ci/evergreen/blob/e37538a7fd96246c2241148634adf2b4beabc438/rest/route/github.go#L80 but maybe that should be its own middleware so it doesn't look like these routes are unauthed? (Could be a split ticket though) Regardless though the average user isn't able to use these.

rest/route/service.go
```diff
@@ -204,14 +205,14 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
...
 	app.AddRoute("/repos/{repo_id}").Version(2).Patch().Wrap(requireUser, requireRepoAdmin, editProjectSettings).RouteHandler(makePatchRepoByID(env.Settings()))
 	app.AddRoute("/roles").Version(2).Get().Wrap(requireUser).RouteHandler(acl.NewGetAllRolesHandler(env.RoleManager()))
 	app.AddRoute("/roles").Version(2).Post().Wrap(requireUser).RouteHandler(acl.NewUpdateRoleHandler(env.RoleManager()))
 	app.AddRoute("/roles/{role_id}/users").Version(2).Get().Wrap(requireUser).RouteHandler(makeGetUsersWithRole())
 	app.AddRoute("/scheduler/compare_tasks").Version(2).Post().Wrap(requireUser).RouteHandler(makeCompareTasksRoute())
-	app.AddRoute("/status/cli_version").Version(2).Get().RouteHandler(makeFetchCLIVersionRoute())
+	app.AddRoute("/status/cli_version").Version(2).Get().Wrap(requireUserToggleable).RouteHandler(makeFetchCLIVersionRoute())
 	app.AddRoute("/status/hosts/distros").Version(2).Get().Wrap(requireUser).RouteHandler(makeHostStatusByDistroRoute())
 	app.AddRoute("/status/notifications").Version(2).Get().Wrap(requireUser).RouteHandler(makeFetchNotifcationStatusRoute())
 	app.AddRoute("/status/recent_tasks").Version(2).Get().RouteHandler(makeRecentTaskStatusHandler())
```
ablack12, line 218: Should this one have auth?

service/api.go
```diff
@@ -94,6 +94,22 @@ func MustHaveProject(r *http.Request) *model.Project {
 	return p
 }
 
+// requireUserToggleable
```
ablack12, line 97: can we add to this comment or remove it?

service/api.go
```diff
@@ -765,15 +782,15 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
...
-	app.AddRoute("/status/consistent_task_assignment").Handler(as.consistentTaskAssignment).Get()
-	app.AddRoute("/status/stuck_hosts").Handler(as.getStuckHosts).Get()
-	app.AddRoute("/status/info").Handler(as.serviceStatusSimple).Get()
-	app.AddRoute("/task_queue").Handler(as.getTaskQueueSizes).Get()
-	app.AddRoute("/task_queue/limit").Handler(as.checkTaskQueueSize).Get()
+	app.AddRoute("/status/consistent_task_assignment").Wrap(requireUserToggleable).Handler(as.consistentTaskAssignment).Get()
+	app.AddRoute("/status/stuck_hosts").Wrap(requireUserToggleable).Handler(as.getStuckHosts).Get()
+	app.AddRoute("/status/info").Wrap(requireUserToggleable).Handler(as.serviceStatusSimple).Get()
+	app.AddRoute("/task_queue").Wrap(requireUserToggleable).Handler(as.getTaskQueueSizes).Get()
+	app.AddRoute("/task_queue/limit").Wrap(requireUserToggleable).Handler(as.checkTaskQueueSize).Get()
```
ablack12, line 793: Can you check where these routes are used (since we use them as opposed to users) and verify that we pass in user information?

ablack12: Also should /dockerfile also have this flag?

service/rest.go
```diff
@@ -84,7 +84,7 @@ func GetRESTv1App(evgService restAPIService) *gimlet.APIApp {
 	app.AddRoute("/tasks/{task_id}").Version(1).Get().Handler(rest.getTaskInfo).Wrap(middleware)
 	app.AddRoute("/tasks/{task_id}/status").Version(1).Get().Handler(rest.getTaskStatus).Wrap(middleware)
 	app.AddRoute("/versions/{version_id}").Version(1).Get().Handler(rest.getVersionInfo).Wrap(middleware)
-	app.AddRoute("/versions/{version_id}").Version(1).Patch().Handler(requireUser(rest.modifyVersionInfo, nil)).Wrap(middleware)
+	app.AddRoute("/versions/{version_id}").Version(1).Patch().Handler(requireUser(false, rest.modifyVersionInfo, nil)).Wrap(middleware)
```
ablack12, line 87: A little confused; do all the other routes already have auth as part of their middleware or handlers?  From what I remember for example the waterfall is supposed to be unauthed.

ablack12 (changes requested):
We may have to split a ticket to extend this to the new UI but @khelif96 may know more about that.

khelif96:
> We may have to split a ticket to extend this to the new UI but @khelif96 may know more about that.

All spruce pages already require auth by default.

khelif96:
Can you please validate legacy lobster links still work.

rest/route/service.go
hadjri: Added: https://jira.mongodb.org/browse/EVG-17989

rest/route/service.go
hadjri: Missed it - thank you

rest/route/service.go
hadjri: Hmm ok - I'll remove the middleware for now.  I actually don't have a strong preference on whether we should validate via middleware or in `Parse`, since they are ultimately very similar concepts.  I'll add a ticket we can discuss in triage.

hadjri: https://jira.mongodb.org/browse/EVG-18002

service/api.go
hadjri: I've checked our project and others in our org and haven't found any invocation of any of these internal routes with the exception of `status/info`. They also don't show up in splunk history so I believe they should be unused and safe to remove.  

The `status/info` endpoint is invoked by the app servers setup scripts which don't pass user information. I'll remove the middleware for now and add a comment mentioning that we're currently leaving this route without middleware. Since this is a simple / low latency endpoint it worth to keep it this way or do we want to introduce a new ticket to work on passing user info from the python setup scripts that the app servers use?

service/rest.go
hadjri: To be honest I'm not sure of the context why this route in particular had this handler function wrapper on it... I just refactored the function when I was making changes to the way `requireUser` was used for the UI routes.

On a related note, the middleware currently used in this file is not doing what I initially thought and is only checking for a logged in user when the project context's project ref is private, or if the project context's patch is defined.

I've imported `NewRequireAuthHandler` to this file and applied it to all routes so now they should all actually be only available to logged in users.

model/project_ref.go
```diff
@@ -313,8 +313,12 @@ func (p *ProjectRef) IsEnabled() bool {
 	return utility.FromBoolPtr(p.Enabled)
 }
 
+// IsPrivate checks if the project ref should be accessed by non-logged in users.
+// If PartialRouteAuthDisabled is set, all project routes require users to be logged in
+// so this function will return false.
 func (p *ProjectRef) IsPrivate() bool {
-	return utility.FromBoolPtr(p.Private)
+	flags, err := evergreen.GetServiceFlags()
+	return err == nil && !flags.PartialRouteAuthDisabled && utility.FromBoolPtr(p.Private)
```
ablack12, line 321: This is pedantic but if err != ni, we should just use utility.FromBoolPtr(p.Private) I. think, so we don't return false for actually private projects

service/middleware.go
```diff
@@ -119,8 +119,21 @@ func (uis *UIServer) GetSettings() evergreen.Settings {
...
+// routes will affect users.
+func requireUser(skipWithToggle bool, onSuccess, onFail http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		if skipWithToggle {
+			flags, err := evergreen.GetServiceFlags()
+			if err != nil {
+				gimlet.WriteResponse(w, gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "retrieving admin settings")))
+			}
+			if flags.PartialRouteAuthDisabled {
+				onFail(w, r)
```
ablack12, line 133: Do we need to do the `if onFail != nil` check like below?

ablack12: I'm also not certain why this isn't just returning onSuccess if flags.PartialRouteAuthDisabled is false

ablack12 (approved):
A couple thoughts but generally looks good to me! We can discuss how/when to turn this on with Tara and maybe the security working group in a couple of weeks 🎉

service/middleware.go
hadjri: Yes, I had it as `onSuccess` this morning but somehow convinced myself it should be the other way around 🤦.

Changed it back to 
`if !flags.PartialRouteAuthDisabled {
				onSuccess(w, r)
				return
			}`

PR #5970, opened 2022-09-28
EVG-17667: use container task pod ID for notifications

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17667

### Description 
This is a mishmash of different usages of the task's pod ID after it's been dispatched, which is pretty similar to usages of the task's `HostId` when it's a host task:

* Use pod ID for container tasks in notifications. Instead of giving a link to the host ID, container task notifications should have a link to the pod ID.
* Log events for task completion and task being cleared from a pod.
* Check pod health when a task hits the stale execution timeout.

### Testing
Added unit tests.

For the UI route, we only want to support it in Spruce, so I tested it in local Evergreen to ensure that it returns 501 if auto-redirection to Spruce is disabled by the user. Spruce itself has no UI for pods so far, but I verified that it did redirect to the expected Spruce URL when auto-redirection to Spruce was enabled.

### DIFF
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -151,7 +151,7 @@ func LogTaskFinished(taskId string, execution int, status string) {
 // it was assigned to run on a host, it logs an additional host event indicating
 // that its assigned task has finished.
 func LogHostTaskFinished(taskId string, execution int, hostId, status string) {
-	logTaskEvent(taskId, TaskFinished, TaskEventData{Execution: execution, Status: status})
+	LogTaskFinished(taskId, execution, status)
 	if hostId != "" {
 		LogHostEvent(hostId, EventHostTaskFinished, HostEventData{Execution: strconv.Itoa(execution), TaskStatus: status, TaskId: taskId})
 	}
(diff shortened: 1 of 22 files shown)


model/event/task_event.go
```diff
@@ -142,10 +142,28 @@ func LogTaskStarted(taskId string, execution int) {
...
+		LogHostEvent(hostId, EventHostTaskFinished, HostEventData{Execution: strconv.Itoa(execution), TaskStatus: status, TaskId: taskId})
+	}
+}
+
+// LogContainerTaskFinished logs an event for a container task being marked
+// finished. If it was assigned to run on a pod, it logs an additional pod event
+// indicating that its assigned task has finished.
+func LogContainerTaskFinished(taskID string, execution int, podID, status string) {
+	LogTaskFinished(taskID, execution, status)
+	if podID != "" {
```
hadjri, line 165: It seems we'd want to alert / catch if we get a container task finishing event with a nonexistent pod

service/pod.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/evergreen-ci/evergreen/model/pod"
+	"github.com/evergreen-ci/gimlet"
+)
+
+func (uis *UIServer) podPage(w http.ResponseWriter, r *http.Request) {
+	id := gimlet.GetVars(r)["pod_id"]
+
+	p, err := pod.FindOneByID(id)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, err)
```
hadjri, line 16: can we wrap this error with "finding pod"?

Kimchelly: Done.

hadjri (approved):

PR #5973, opened 2022-09-30
EVG-18000 Get and Store slack information when creating a user

malikchaya2:
[EVG-18000](https://jira.mongodb.org/browse/EVG-18000)

### Description 
Get and Store slack information when creating a user

### Testing 
I recreated my own user on staging and verified that it fetches the slack information correctly. 
image


model/user/db.go
```diff
@@ -306,16 +307,34 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
...
 	}
 	setOnInsertFields := bson.M{
 		APIKeyKey: utility.RandomString(),
 		bsonutil.GetDottedKeyName(SettingsKey, UseSpruceOptionsKey, SpruceV1Key): true,
 	}
+
+	if accessToken != "" {
+		setFields[bsonutil.GetDottedKeyName(LoginCacheKey, LoginCacheAccessTokenKey)] = accessToken
+	} else {
+		// only fetch the slack user if the evergreen user doesn't exist (which is the case when there is no access token)
```
Kimchelly, line 321: The access token is an Okta-specific field, but I don't think it can be relied on to indicate that it's the first login ever because it could be cleared if the user is logged out. I imagine what we actually want is to check if the user we're about to update has `SlackMemberId`/`SlackUsername` set, regardless of whether it's their first or Nth login. Would it be okay to just look up the user, see if their Slack settings are populated (or if the user doesn't exist yet), and fill them in here if they're empty?

Kimchelly (changes requested):

model/user/db.go
malikchaya2: wait, ignore me, lol

malikchaya2: updated.

model/user/db.go
```diff
@@ -336,9 +341,55 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
 		return nil, errors.Wrapf(err, "decoding user '%s'", userId)
 
 	}
+
+	if err := setSlackInformation(env, u); err != nil {
+		return nil, errors.Wrapf(err, "setting slack information for user '%s'", userId)
```
Kimchelly, line 346: Can we log + continue on error here? The Okta user manager requires only the DB modification above to succeed, so it seems like setting the Slack auth would be a nice-to-have.

model/user/db.go
```diff
@@ -336,9 +341,55 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
...
+	// get the slack userID and memberId from the user's email
+	slackEnv := env.Settings().Slack
+	slackUser, err := slackEnv.Options.GetSlackUser(slackEnv.Token, u.EmailAddress)
+	if err != nil {
+		return errors.Wrapf(err, "getting Slack user with email address '%s'", u.EmailAddress)
+	}
+	if slackUser == nil {
+		grip.Error(message.Fields{
+			"message": fmt.Sprintf("Couldn't find slack user by email address", u.EmailAddress),
+			"userId":  u.Id,
```
Kimchelly, line 371: nit: `user_id`

model/user/db.go
```diff
@@ -336,9 +341,55 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
...
+
+	// get the slack userID and memberId from the user's email
+	slackEnv := env.Settings().Slack
+	slackUser, err := slackEnv.Options.GetSlackUser(slackEnv.Token, u.EmailAddress)
+	if err != nil {
+		return errors.Wrapf(err, "getting Slack user with email address '%s'", u.EmailAddress)
+	}
+	if slackUser == nil {
+		grip.Error(message.Fields{
+			"message": fmt.Sprintf("Couldn't find slack user by email address", u.EmailAddress),
```
Kimchelly, line 370: Can we include the email address as a separate field?

model/user/db.go
```diff
@@ -336,9 +341,55 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
...
+	slackFields := bson.M{
+		bsonutil.GetDottedKeyName(SettingsKey, userSettingsSlackMemberIdKey): slackUser.ID,
+	}
+	if slackUser.Name != "" {
+		slackFields[bsonutil.GetDottedKeyName(SettingsKey, userSettingsSlackUsernameKey)] = slackUser.Name
+	}
+
+	update := bson.M{"$set": slackFields}
+
+	if err := UpdateOne(bson.M{IdKey: u.Id}, update); err != nil {
```
Kimchelly, line 385: Can this update be combined with the one update above this? Seems like the Slack fields could just be set in L327 (contingent on `GetSlackUser` succeeding).

malikchaya2: done

model/user/db.go
malikchaya2: done

model/user/db.go
malikchaya2: done

model/user/db.go
malikchaya2: I moved it out because GetOrCreateUser doesn't do a find on the userID, so we can't know if the slackMemberId is set until it returns the user at the end.

malikchaya2: (so we would have to ping slack every time GetOrCreateUser is called, which I'm trying to avoid)

graphql/directive_test.go
```diff
@@ -79,6 +79,15 @@ func setupPermissions(t *testing.T) {
 
 func TestRequireSuperUser(t *testing.T) {
 	setupPermissions(t)
+	dbUser := &user.DBUser{
```
Kimchelly, line 82: Seems like this should clear the user collection before attempting to insert.

model/user/db.go
```diff
@@ -336,9 +340,71 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
 		return nil, errors.Wrapf(err, "decoding user '%s'", userId)
 
 	}
+
+	if err := setSlackInformation(env, u); err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
+			"message":       "setting slack information for user",
```
Kimchelly, line 346: "could not set Slack information for user"

model/user/db.go
```diff
@@ -336,9 +340,71 @@ func GetOrCreateUser(userId, displayName, email, accessToken, refreshToken strin
...
+	if u.Settings.SlackMemberId != "" {
+		// user already has a slack member id set
+		return nil
+	}
+	if u.EmailAddress == "" {
+		// we can't fetch the slack information without the user's email address
+		return errors.New("user has no email address")
+	}
+
+	// get the slack userID and memberId from the user's email
```
Kimchelly, line 376: Do we need the comment?

bynn (approved):
i think everything looks good after kims comment about clearing the db for testing!

model/user/db.go
malikchaya2: done

graphql/directive_test.go
malikchaya2: done

PR #5975, opened 2022-09-30
EVG-17123 Allow variable lookup in expansion secondary value

hadjri:
[EVG-17123](https://jira.mongodb.org/browse/EVG-17123)

### Description 
To improve the maintainability of config files, the secondary value that can be specified in an expansion (i.e. `${val1|val2}`) is no longer a hard coded string but can itself act as a backup variable lookup.  To distinguish between what should be a hardcoded string vs a variable lookup secondary value, I'm proposing that users should specify that the secondary value should be looked up by prepending a `*` to it. Example: `${val1|*val2}`
### Testing 
Updated unit tests, and checked in staging that the secondary value in `ExpandString` has its value looked up properly if it is prepended with `*`

### DIFF
--- a/util/expansion.go
+++ b/util/expansion.go
@@ -90,10 +90,10 @@ func (exp *Expansions) ExpandString(toExpand string) (string, error) {
 				malformedFound = true
 			}
 
-			// parse into the name and default value
-			var defaultVal string
+			// parse into the name and secondary value
+			var secondaryValue string
 			if idx := strings.Index(match, "|"); idx != -1 {
-				defaultVal = match[idx+1:]
+				secondaryValue = match[idx+1:]
 				match = match[0:idx]
 			}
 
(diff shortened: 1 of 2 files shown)


util/expansion.go
```diff
@@ -102,7 +102,15 @@ func (exp *Expansions) ExpandString(toExpand string) (string, error) {
 				return []byte(exp.Get(match))
 			}
 
-			return []byte(defaultVal)
+			// look for an expansion in the secondary value
+			if strings.HasPrefix(secondaryValue, "*") {
+				// trim off *
+				secondaryValue = secondaryValue[1:]
```
malikchaya2, line 108: do we have any precedent for this pattern of appending a * to something to indicate that it should be looked up?

hadjri: There isn't any precedent - the user who submitted the ticket suggested this syntax (among others) and I thought it was cleaner than nesting ${val1|${val2}} while still making it clear that the second value is a variable and not a hard coded string

bynn (approved):
looks good but is there a reason we're changing the name of the var from defaultVal to secondaryVal? 

Wouldn't default be more accurate (unless you can add another tertiary val?)

hadjri:
> This looks good but if we cannot find the expansion, we default to an empty string right? Do we still want users to be able to add ANOTHER default value if we can't find either one? Or just let them know that this is the risk that they are taking

I'm inclined for the latter - this feature request was specifically to provide a programmable secondary default value. If we wanted to do N number of defaults i.e. `${val1|val2|val3|val4}` I'd propose we wait for demand for that feature first / discuss it in triage.

PR #5984, opened 2022-10-04
EVG-17638 Use new index for UnscheduleStaleUnderwaterHostTasks

hadjri:
[EVG-17638](https://jira.mongodb.org/browse/EVG-17638)

### Description 
This query is currently slow because the current `distro_1_status_1_activated_1_priority_1` index doesn't have any sorting on the `depends_on.unattainable` key which causes us to return many index keys for some queries which we then have to filter down. Now that the `distro_1_status_1_activated_1_priority_1_override_dependencies_1_depends_on.unattainable_1` index has been created in prod we'll need to update our hint for this function to use the new index.
### Testing 
In Compass I constructed manually the aggregation used by this query and ran it with and without the new index, and noticed a significant reduction in returned index keys. (screenshots attached).

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -27,6 +27,8 @@ var (
 		{Key: StatusKey, Value: 1},
 		{Key: ActivatedKey, Value: 1},
 		{Key: PriorityKey, Value: 1},
+		{Key: OverrideDependenciesKey, Value: 1},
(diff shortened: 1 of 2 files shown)


hadjri:
image

image

Kimchelly (approved):

PR #5985, opened 2022-10-04
EVG-17213 find and replace collection name

ybrill:
[EVG-17213](https://jira.mongodb.org/browse/EVG-17213)

### Description 
This is mostly just a find and replace on the old collection name.
For now let's continue to write to both collections. This way if we need to roll it back we'll be okay. Once this PR is deployed for a bit (a week?) we can stop writing to the old collection.

### Testing 
It's hard to test everything this touches. I did a small test by restarting a task in staging and signing up for a notification.

One risk is that there's an index we need that I missed when I made the new collection. The db/queries seem fine in staging, but we'll only know for sure in prod 😅


malikchaya2 (commented):
I probably missed it, but where is the writing to two collections happening?

PR #5981, opened 2022-10-03
EVG-17930: Remove legacy agent routes

hadjri:
[EVG-17930](https://jira.mongodb.org/browse/EVG-17930)

### Description 
Since the merging of [this commit](https://github.com/evergreen-ci/evergreen/commit/9705ec3175f6584129e13886a2c0d14bc2ce930e) agents have had sufficient time to roll over to begin using v2 for all agent routes. It should be safe to remove v1 agent routes and the functions that depended solely on them.
### Testing 
Deployed to staging, ran new versions and tasks without error.

PR #5987, opened 2022-10-04
EVG-17644, EVG-17899: refactor host task dispatch

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-17899
https://jira.mongodb.org/browse/EVG-17644

### Description 
I split this ticket in two because the current logic is too messy to easily add changes to the dispatching logic. This PR has no functional change and is solely a refactor of the existing logic to clean it up and reduce the potential for accidental bugs in the actual logic change. I'm going to make a follow-up PR to actually change the dispatch logic.

* Refactor and make equivalent methods for the host/task dispatch DB methods. This is pre-work for the actual change. I kept the old methods as-is, even though they're almost identical, because we have to feature flag the follow-up change to fall back to the old logic.
* Refactor the task group ordering and max host check into a helper function. Also clean up some of the comments so it's easier to understand.
* Handle an unlikely nil task case when sending back a running task to the host that's already assigned to run it.
* Remove unused logs for distro monitoring in dispatch.
* ~Remove the legacy REST v1 next task route because I didn't feel like porting the logic to a route that's going to be imminently deleted. This is mostly to make compilation succeed and I'm going to merge main once #5981 is merged.~ Main is merged, so this should be irrelevant.

### Testing
Existing tests for dispatch behavior still pass.


model/host/host.go
```diff
@@ -1394,16 +1394,45 @@ func (h *Host) ClearRunningTask() error {
...
+		"task_id":  t.Id,
+		"distro":   h.Distro.Id,
+	})
+
+	return nil
+}
+
+// UpdateRunningTaskWithContext updates the running task for the host. It does
+// not log an event for task assignment.
+func (h *Host) UpdateRunningTaskWithContext(ctx context.Context, env evergreen.Environment, t *task.Task) error {
```
hadjri, line 1422: Was this function created to be used in the future PR?

model/task/sorter.go
```diff
@@ -44,6 +44,8 @@ func (t Tasks) InsertUnordered(ctx context.Context) error {
 	return err
 }
 
+// ByPriority sorts tasks according to their display statuses (and has nothing
+// to do with its scheduling priority).
```
hadjri, line 48: We should add that this is exclusively used for display tasks

model/task/task.go
```diff
@@ -2039,6 +2055,7 @@ func (t *Task) displayTaskPriority() int {
 	case evergreen.TaskSucceeded:
 		return 110
 	}
+	// Note that this include evergreen.TaskDispatched.
```
hadjri, line 2058: nit: includes

model/task/task.go
```diff
@@ -1093,6 +1069,46 @@ func (t *Task) MarkAsHostDispatched(hostId, distroId, agentRevision string,
 	return nil
 }
 
+// MarkAsHostDispatchedWithContext marks that the task has been dispatched onto
+// a particular host. Unlike MarkAsHostDispatched, this does not update the
+// parent display task.
+func (t *Task) MarkAsHostDispatchedWithContext(ctx context.Context, env evergreen.Environment, hostID, distroID, agentRevision string, dispatchTime time.Time) error {
```
hadjri, line 1075: Same question here

model/task/task.go
```diff
@@ -1093,6 +1069,46 @@ func (t *Task) MarkAsHostDispatched(hostId, distroId, agentRevision string,
...
+			DistroIdKey:      distroID,
+			AgentVersionKey:  agentRevision,
+		},
+		"$unset": bson.M{
+			AbortedKey:   "",
+			AbortInfoKey: "",
+			DetailsKey:   "",
+		},
+	}); err != nil {
+		return err
```
hadjri, line 1099: Can we wrap the error with a message here?

rest/route/host_agent.go
```diff
@@ -778,28 +683,77 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
...
+// still respected after the task has already been assigned to the host and that
+// for single-host task groups, it is dispatching the next task in the group.
+//
+// It's possible for dispatchers on different app servers to race, assigning
+// different tasks in a task group to more hosts than the task group's max
+// hosts. Therefore, we must check that the number of hosts running this task
+// group does not exceed the max after assigning the task to the host. If it
+// exceeds the max host limit, this will return true to indicate that the host
+// should not run this task.
+func checkHostTaskGroupAfterDispatch(h *host.Host, t *task.Task) (isDispatchable bool, err error) {
```
hadjri, line 707: It seems `isDispatchable` is a proxy for `err != nil`, can we remove it?

rest/route/host_agent.go
```diff
@@ -778,28 +683,77 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
...
+	var minTaskGroupOrderNum int
+	if t.TaskGroupMaxHosts == 1 {
```
hadjri, line 709: We can use `IsPartOfSingleHostTaskGroup` here

rest/route/host_agent.go
```diff
@@ -778,28 +683,77 @@ func assignNextAvailableTask(ctx context.Context, taskQueue *model.TaskQueue, di
...
+	}
+
+	// For multiple-host task groups and single-host task groups without order
+	// cached in the host, check that max hosts is respected.
+	if minTaskGroupOrderNum == 0 {
+		numHosts, err := host.NumHostsByTaskSpec(t.TaskGroup, t.BuildVariant, t.Project, t.Version)
+		if err != nil {
+			return false, errors.Wrap(err, "getting number of hosts running task group")
+		}
+		if numHosts > t.TaskGroupMaxHosts {
```
hadjri, line 749: Since we only run this function for hosts that are new to the group, wouldn't we want to check if `numHosts >= t.TaskGroupMaxHosts` since adding one more host would breach the limit?

hadjri (changes requested):

rest/route/host_agent.go
Kimchelly: The max hosts check is done after the task is already dispatched because we can't be sure when doing the dispatch updates whether or not max hosts is exceeded yet.

hadjri: Ah I see, thanks for clarifying

hadjri (approved):

PR #5988, opened 2022-10-04
EVG-17922 Introduce container variant in staging config

hadjri:
[EVG-17922](https://jira.mongodb.org/browse/EVG-17922)

### Description 
Pre-productionizing containers, it's useful to regularly run tasks on pods in the staging waterfall to monitor containerized tasks' overall behavior.  We can introduce a separate config YAML just for staging that imports our main `self-tests.yml`, and adds some container configurations and a containerized BV.
### Testing 
Since self-tests.staging.yml isn't in the main branch yet, I did a workaround - I ran a patch where the contents of `self-tests-staging.yml` was copied and pasted into `self-tests.yml` and vice versa. (Which is the opposite direction of how the importing will work but it's the same principle). Then ran a patch and confirmed that the config files merged correctly and all variants (including the new one) were able to run. 

Post-merging I'll change the remote path for Evergreen in staging to point to this new file.


self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      goos: linux
+      goarch: amd64
+      IS_DOCKER: true
+      GOROOT: /usr/local/go
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-4.2.18.tgz
+      decompress: tar zxvf
+      notary_client_url: https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.1.2/linux_amd64.zip
+      sign_macos: true
+    tasks:
+      - name: "dist-and-upload-clis"
```
Kimchelly, line 30: Can we omit this one or maybe just add the `dist` task to ensure it compiles? Running just `dist` should be okay, but `upload-clis` will conflict with staging's `upload-clis` because they'll both upload to the same location in S3. This also relies on single-host task groups, which we don't support yet.

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-4.2.18.tgz
+      decompress: tar zxvf
+      notary_client_url: https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.1.2/linux_amd64.zip
+      sign_macos: true
+    tasks:
+      - name: dist
+      - name: ".smoke"
+      - name: ".test"
+      - name: "js-test"
+      - name: ".linter"
```
Kimchelly, line 34: Now that I think of it, [the linter also uses a task group](https://github.com/evergreen-ci/evergreen/blob/cfe34fb8acc102cad213d1b550a24e3ef1a3f6d4/cmd/generate-lint/generate-lint.go#L151). I think it's alright to not test the linting variant.

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: "docker-cleanup"
```
Kimchelly, line 35: I think this may fail because it assumes the Docker daemon is running in the runtime environment, which doesn't work inside a container (since we don't have Docker-in-Docker support).

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+    run_on:
+      - evg-container
+    expansions:
+      goos: linux
+      goarch: amd64
+      IS_DOCKER: true
+      GOROOT: /usr/local/go
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-4.2.18.tgz
+      decompress: tar zxvf
+      notary_client_url: https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.1.2/linux_amd64.zip
```
Kimchelly, line 27: I think we can remove `notary_client_url` and `sign_macos` because the container doesn't have the MacOS signing service installed on it.

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      sign_macos: true
+    tasks:
+      - name: dist
```
Kimchelly, line 30: I thought about this some more, and I think what we actually want is `dist-unsigned` because we don't support MacOS signing (for the reason above this).

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: ".smoke"
+      - name: ".test"
+      - name: "js-test"
+      - name: ".linter"
+      - name: "docker-cleanup"
+      - name: test-db-auth
+      - name: test-cloud
```
Kimchelly, line 37: I believe [this](https://github.com/evergreen-ci/evergreen/blob/cfe34fb8acc102cad213d1b550a24e3ef1a3f6d4/self-tests.yml#L407-L409) will also fail due to the Docker integration test. I imagine it should be okay to skip.

self-tests-staging.yml
```diff
@@ -0,0 +1,37 @@
...
+      GOROOT: /usr/local/go
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-4.2.18.tgz
+      decompress: tar zxvf
+      notary_client_url: https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.1.2/linux_amd64.zip
+      sign_macos: true
+    tasks:
+      - name: dist
+      - name: ".smoke"
+      - name: ".test"
+      - name: "js-test"
```
Kimchelly, line 33: I think `js-test` requires node and such (as well as expansions like `NODE_BIN_PATH` and `KARMA_REPORTRER`). Does the container have that installed, or should we skip this?

Kimchelly (commented):
Sorry, I did a more in-depth think about all the tasks and realized some of the other ones may have to be skipped as well.

self-tests-staging.yml
hadjri: The container does have node installed, this test should be fine to keep.

self-tests-staging.yml
Kimchelly: Did you mean to remove docker-cleanup and keep js-test?

Kimchelly (approved):
LGTM mod one question.

hadjri:
> LGTM mod one question.

I did.... thanks!

PR #5989, opened 2022-10-04
EVG-17457: check project ref for dispatchability during pod allocation

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17457

### Description 
Before allocating a pod, check whether a task will be allowed to dispatch based on project ref settings. If it's not allowed, don't allocate it.

* Add check for project ref against the task to ensure it can dispatch before allocating a pod.
* Reduce some copy-pasted identical project ref checks between host and container task schedulers.

### Testing 
Updated existing unit tests.

### DIFF
--- a/units/pod_allocator.go
+++ b/units/pod_allocator.go
@@ -128,11 +128,8 @@ func (j *podAllocatorJob) Run(ctx context.Context) {
 
 	canDispatch, reason := model.ProjectCanDispatchTask(j.pRef, j.task)
 	if !canDispatch {
-		if reason == "" {
-			reason = "of unknown reason"
-		}
 		grip.Info(message.Fields{
-			"message": "refusing to allocate task because it is not dispatchable",
+			"message": "refusing to allocate task due to project ref settings",
 			"reason":  reason,
 			"task":    j.task.Id,
 			"project": j.pRef.Identifier,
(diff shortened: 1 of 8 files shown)


units/pod_allocator.go
```diff
@@ -107,9 +107,37 @@ func (j *podAllocatorJob) Run(ctx context.Context) {
...
+		return
+	}
+
+	canDispatch, reason := model.ProjectCanDispatchTask(j.pRef, j.task)
+	if !canDispatch {
+		if reason == "" {
+			reason = "of unknown reason"
+		}
+		grip.Info(message.Fields{
+			"message": "refusing to allocate task because it is not dispatchable",
```
hadjri, line 135: This implies something is amiss with the task itself. Can we change it to say that it can't be allocated due to the project ref's settings?

units/pod_allocator.go
```diff
@@ -107,9 +107,37 @@ func (j *podAllocatorJob) Run(ctx context.Context) {
...
+			"task":    j.task.Id,
+			"project": j.pRef.Identifier,
+			"job":     j.ID(),
+		})
+		return
+	}
+
+	canDispatch, reason := model.ProjectCanDispatchTask(j.pRef, j.task)
+	if !canDispatch {
+		if reason == "" {
```
hadjri, line 131: We can also default `reason` to be "of unknown reason" on L1 of `ProjectCanDispatchTask` if you think that's cleaner

units/pod_allocator.go
Kimchelly: Done.

units/pod_allocator.go
Kimchelly: Hmm, I looked back at this logic and I think we don't need to deal with the empty string case anymore since it'll just show up in the log as an empty field. I removed this.

PR #5993, opened 2022-10-07
EVG-18034 Add taskgroup tasks to get tasks by variant

bynn:
[EVG-18034](https://jira.mongodb.org/browse/EVG-18034)

### Description 
taskgroups werent being taken into account for --repeat-failed 

### Testing 
unit testing 
staging 
https://spruce-staging.corp.mongodb.com/patch/633f44e397b1d3368cc303be/configure/tasks

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -1439,11 +1439,7 @@ func (p *Project) FindTasksForVariant(build string) []string {
 		if b.Name == build {
 			tasks := make([]string, 0, len(b.Tasks))
 			for _, task := range b.Tasks {
-				if task.IsGroup {
-					tasks = append(tasks, p.FindTaskGroup(task.Name).Tasks...)
-				} else {
-					tasks = append(tasks, task.Name)
-				}
+				tasks = append(tasks, task.Name)
 			}
 			return tasks
 		}
--- a/model/project.go
+++ b/model/project.go
@@ -1439,7 +1439,11 @@ func (p *Project) FindTasksForVariant(build string) []string {
 		if b.Name == build {
 			tasks := make([]string, 0, len(b.Tasks))
 			for _, task := range b.Tasks {
-				tasks = append(tasks, task.Name)
+				if task.IsGroup {
+					tasks = append(tasks, p.FindTaskGroup(task.Name).Tasks...)
+				} else {
+					tasks = append(tasks, task.Name)
+				}
 			}
 			return tasks
 		}
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -501,7 +501,7 @@ func setTasksToPreviousFailed(patchDoc, previousPatch *patch.Patch, project *mod
 	for _, vt := range previousPatch.VariantsTasks {
 		tasksInProjectVariant := project.FindTasksForVariant(vt.Variant)
 		var tasks []string
-		tasks, err := getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant, vt, previousPatch.Version)
+		tasks, err := getPreviousFailedTasksAndDisplayTasks(project, tasksInProjectVariant, vt, previousPatch.Version)
 		if err != nil {
 			return err
 		}
@@ -527,7 +527,6 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
 		if err = setTasksToPreviousFailed(patchDoc, previousPatch, project); err != nil {
 			return "", errors.Wrap(err, "settings tasks to previous failed")
 		}
-
 	} else {
 		patchDoc.Tasks = previousPatch.Tasks
 	}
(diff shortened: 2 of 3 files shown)


units/patch_intent_test.go
```diff
@@ -333,7 +349,14 @@ func (s *PatchIntentUnitsSuite) TestSetToPreviousPatchDefinition() {
...
 		},
-	}}
+	},
+		TaskGroups: []model.TaskGroup{
+			{
+				Name:  "tg",
+				Tasks: []string{"tgt1", "tgt2"},
+			},
+		},
+	}
```
Kimchelly, line 359: If a single-host task group task fails, it should supposedly include all the tasks preceding the failed one as well. Does this account for the previous successful tasks?

Can we test the single host task group case?

bynn: It will only work if you have a dependency on it. 
Would the current behavior be preferable? Or do we just assume that single host tasks are dependent on each other?

model/project.go
```diff
@@ -1439,7 +1439,11 @@ func (p *Project) FindTasksForVariant(build string) []string {
 		if b.Name == build {
 			tasks := make([]string, 0, len(b.Tasks))
 			for _, task := range b.Tasks {
-				tasks = append(tasks, task.Name)
+				if task.IsGroup {
+					tasks = append(tasks, p.FindTaskGroup(task.Name).Tasks...)
```
ablack12, line 1443: Can tasks be across different variants? This definitely doesn't seem like the right place for this change since the function does imply it will _only_ return tasks that belong in that variant and might cause weird behavior. More likely you'll want to find where tasks are added for --repeat-failed and then add earlier task group tasks at that time.

ablack12 (changes requested):

units/patch_intent_test.go
Kimchelly: I think single-host task group dependencies aren't added until the task is about to end ([here](https://github.com/evergreen-ci/evergreen/blob/9e6ebecbc17cef8663460d75703fd3c05af64e63/model/task_queue.go#L279-L283), which is called in the agent end task route), so single-host task groups aren't guaranteed to have the dependencies when the user reuses the patch.

bynn: updated

model/project.go
bynn: updated

units/patch_intent.go
```diff
@@ -527,29 +527,40 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 }
 
-func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, error) {
+func getPreviousFailedTasksAndDisplayTasks(project *model.Project, tasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, error) {
 	failedTasks, err := task.FindAll(db.Query(task.FailedTasksByVersionAndBV(version, vt.Variant)))
 	if err != nil {
 		return nil, errors.Wrapf(err, "finding failed tasks in build variant '%s' from previous patch '%s'", vt.Variant, version)
 	}
 	failedExecutionTasks := []string{}
+	taskGroupFailedTasks := map[string][]string{}
```
ablack12, line 543: Why do we store these in a separate place rather than just adding them to failedExecutionTasks? They would hit the same StringSliceIntersection which would remove any deduplication (I'm thinking it might be to check that g is in tasksInProjectVariant but I think now that you're passing in the current project you're already handling that just by using project.FindTaskGroup)

units/patch_intent.go
```diff
@@ -501,7 +501,7 @@ func setTasksToPreviousFailed(patchDoc, previousPatch *patch.Patch, project *mod
 	for _, vt := range previousPatch.VariantsTasks {
 		tasksInProjectVariant := project.FindTasksForVariant(vt.Variant)
```
ablack12, line 502: Now that we're passing project into getPreviousFailedTasksAndDisplayTasks, we can just move this call into that function rather than pass it in

Kimchelly (approved):

units/patch_intent.go
bynn: moved

units/patch_intent.go
bynn: updated as we spoke about in meeting!

units/patch_intent.go
```diff
@@ -499,9 +499,8 @@ func (j *patchIntentProcessor) buildTasksandVariants(patchDoc *patch.Patch, proj
 func setTasksToPreviousFailed(patchDoc, previousPatch *patch.Patch, project *model.Project) error {
 	var failedTasks []string
 	for _, vt := range previousPatch.VariantsTasks {
-		tasksInProjectVariant := project.FindTasksForVariant(vt.Variant)
 		var tasks []string
```
ablack12, line 502: Unrelated nit but can we remove this declaration? Seems like the := at 503 should handle it so I'm guessing this is outdated

units/patch_intent.go
```diff
@@ -527,30 +526,38 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 
-func getPreviousFailedTasksAndDisplayTasks(tasksInProjectVariant []string, vt patch.VariantTasks, version string) ([]string, error) {
+func getPreviousFailedTasksAndDisplayTasks(project *model.Project, vt patch.VariantTasks, version string) ([]string, error) {
+	tasksInProjectVariant := project.FindTasksForVariant(vt.Variant)
 	failedTasks, err := task.FindAll(db.Query(task.FailedTasksByVersionAndBV(version, vt.Variant)))
 	if err != nil {
 		return nil, errors.Wrapf(err, "finding failed tasks in build variant '%s' from previous patch '%s'", vt.Variant, version)
 	}
-	failedExecutionTasks := []string{}
+	// Verify that the taskgroup/task is in the current project definition and in the previous run.
```
ablack12, line 542: nit: task group or task

units/patch_intent.go
```diff
@@ -527,30 +526,38 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+	allFailedTasks := []string{}
 	for _, failedTask := range failedTasks {
-		if !failedTask.DisplayOnly {
-			failedExecutionTasks = append(failedExecutionTasks, failedTask.DisplayName)
+		if failedTask.TaskGroup != "" &&
+			utility.StringSliceContains(tasksInProjectVariant, failedTask.TaskGroup) &&
+			utility.StringSliceContains(vt.Tasks, failedTask.DisplayName) {
```
ablack12, line 547: Can we check this above and just "continue" since we check it in both the if and else case? And then we can add a comment like "skip failed tasks that aren't in the patch definition"  (Although is that a new thing? I don't really understand why we need it; theoretically anything returned from the failed tasks query is part of the patch definition or it wouldn't be a failed task)

bynn: Done, I wrapped it in an extra if 

I'm also not super sure why it's necessary to verify this at this point 
I think @hadjri worked on this last so I wanted to ask if this was necessary for a particular use case?

units/patch_intent.go
```diff
@@ -527,30 +525,38 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
-	failedExecutionTasks := []string{}
+	// Verify that the task group or task is in the current project definition and in the previous run.
+	allFailedTasks := []string{}
 	for _, failedTask := range failedTasks {
-		if !failedTask.DisplayOnly {
-			failedExecutionTasks = append(failedExecutionTasks, failedTask.DisplayName)
+		if utility.StringSliceContains(vt.Tasks, failedTask.DisplayName) {
+			if failedTask.TaskGroup != "" &&
+				utility.StringSliceContains(tasksInProjectVariant, failedTask.TaskGroup) {
+				if failedTask.IsPartOfSingleHostTaskGroup() {
```
ablack12, line 547: Can we add a comment to this to indicate why we're adding earlier task group tasks in the single host case?

ablack12: In general since these clauses are long it would be helpful to have a comment for each of them; comments are free!

ablack12 (approved):
Approving with nits

PR #5997, opened 2022-10-10
EVG-17054 Show activated tasks for mainline commits

khelif96:
[EVG-17054](https://jira.mongodb.org/browse/EVG-17054)

### Description 
Now surfaces unschedule tasks for mainline commits as long as they were not generated. I couldn't test this with the specific server use case we were trying to solve but i think this should cover it. @ablack12  would know more about their use case. 

I think i covered all the areas where setting this conditional would matter but its possible i could have missed something.

Before:
image
After:
image
image


graphql/version_resolver.go
```diff
@@ -83,6 +83,7 @@ func (r *versionResolver) BuildVariantStats(ctx context.Context, obj *restModel.
 		Variants:                       options.Variants,
 		Statuses:                       options.Statuses,
 		IncludeBuildVariantDisplayName: true,
+		IsMainlineCommit:               !evergreen.IsPatchRequester(utility.FromStringPtr(obj.Requester)),
```
ablack12, line 86: Could we say "IsPatch" instead and case on !IsPatch? MainlineCommit is difficult wording since sometimes we use it to mean a commit version (i.e. repotracker version) and sometimes we use it for anything that isn't a patch (i.e. including git tag and etc).

model/task/task.go
```diff
@@ -4129,9 +4131,16 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
-	// Activated Time is needed to filter out generated tasks that have been generated but not yet activated
-	if !opts.IncludeEmptyActivation {
-		match[ActivatedTimeKey] = bson.M{"$ne": utility.ZeroTime}
+	if !opts.IsMainlineCommit {
+		// Activated Time is needed to filter out generated tasks that have been generated but not yet activated
+		if !opts.IncludeEmptyActivation {
+			match[ActivatedTimeKey] = bson.M{"$ne": utility.ZeroTime}
+		}
+	}
+	if opts.IsMainlineCommit {
```
ablack12, line 4140: else

model/task/task.go
```diff
@@ -4129,9 +4131,16 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 		taskNamesAsRegex := strings.Join(opts.TaskNames, "|")
 		match[DisplayNameKey] = bson.M{"$regex": taskNamesAsRegex, "$options": "i"}
 	}
-	// Activated Time is needed to filter out generated tasks that have been generated but not yet activated
-	if !opts.IncludeEmptyActivation {
-		match[ActivatedTimeKey] = bson.M{"$ne": utility.ZeroTime}
+	if !opts.IsMainlineCommit {
```
ablack12, line 4134: We should combine these if statements and update the comment to make it clearer why we handle these cases differently.

model/task/task.go
```diff
@@ -4129,9 +4131,16 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
+		// Activated Time is needed to filter out generated tasks that have been generated but not yet activated
+		if !opts.IncludeEmptyActivation {
+			match[ActivatedTimeKey] = bson.M{"$ne": utility.ZeroTime}
+		}
+	}
+	if opts.IsMainlineCommit {
+		if opts.IncludeEmptyActivation {
```
ablack12, line 4141: Is there a way on mainline for users to toggle this on?

ablack12 (changes requested):
You can test this by pushing an empty sandbox commit (and put a high batchtime on the project so that the commit version doesn't activate), and then verify that all the tasks except the generated ones are visible. (The generated tasks staying invisible is server specific, but not being able to see unactivated tasks affects all projects when they don't activate versions right away).

PR #5998, opened 2022-10-11
EVG-17656: add TTL field for pod dispatchers

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17656

### Description 
Pod dispatchers are only supposed to be used for the brief period of time that they're dispatching tasks and afterwards serve no purpose to keep around (even if they are used again after a long period of time to dispatch a task, they can simply be re-created). Therefore, we can apply a TTL to them once they haven't been used for a while.

* Add a field indicating the last time the pod dispatcher was updated. This will be the key for a TTL index for the collection.

### Testing 
Added unit tests.

### DIFF
--- a/model/pod/definition/definition.go
+++ b/model/pod/definition/definition.go
@@ -73,7 +73,7 @@ func (pdc PodDefinitionCache) Put(_ context.Context, item cocoa.ECSPodDefinition
 			LastAccessedKey: time.Now(),
 		},
 		"$setOnInsert": bson.M{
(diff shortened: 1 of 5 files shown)


hadjri (approved):
I assume the cleanup approach in a follow-up Pr will introduce a new amboy job to check for dispatchers that have surpassed TTL?

Kimchelly:
I'm going to build [a TTL index](https://www.mongodb.com/docs/manual/core/index-ttl/) on this field since MongoDB natively supports them.

PR #5999, opened 2022-10-11
EVG-17693: Promote variables from project to repo

sophstad:
[EVG-17693](https://jira.mongodb.org/browse/EVG-17693)

### Description 
- Add function + resolver for promoting variables from a project to the repo to which it is attached. Variables are removed from projects as part of this operation. This solves the problem where users could not move secret variables (where the value is no longer known to them) to their repo settings.
  - Note that the ticket wasn't specific about whether or not variables should be removed from the project or simply copied to the repo. I opted for the former, but am open to the other approach too if it seems preferable!

### Testing 
- Add unit tests
- Test resolver via GraphQL playground

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -97,6 +97,9 @@ func disableStartingSettings(p *model.ProjectRef) {
 	p.CommitQueue.Enabled = utility.FalsePtr()
 }
 
+// PromoteVarsToRepo moves variables from an attached project to its repo.
+// Promoted vars are removed from the project as part of this operation.
+// Variables whose names already appear in the repo settings will be overwritten.
 func PromoteVarsToRepo(projectId string, varNames []string, userId string) error {
 	project, err := model.GetProjectSettingsById(projectId, false)
 	if err != nil {
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -153,13 +153,13 @@ func PromoteVarsToRepo(projectId string, varNames []string, userId string) error
 		}
 	}
 
-	for key, _ := range projectVars.PrivateVars {
+	for key := range projectVars.PrivateVars {
 		if _, contains := apiProjectVars.Vars[key]; contains {
 			apiProjectVars.PrivateVars[key] = true
 		}
 	}
 
-	for key, _ := range projectVars.AdminOnlyVars {
+	for key := range projectVars.AdminOnlyVars {
 		if _, contains := apiProjectVars.Vars[key]; contains {
 			apiProjectVars.AdminOnlyVars[key] = true
 		}
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -97,6 +97,90 @@ func disableStartingSettings(p *model.ProjectRef) {
 	p.CommitQueue.Enabled = utility.FalsePtr()
 }
 
+func PromoteVarsToRepo(projectId string, varNames []string, userId string) error {
+	project, err := model.GetProjectSettingsById(projectId, false)
+	if err != nil {
+		return errors.Wrapf(err, "getting project settings for project '%s'", projectId)
+	}
+	projectVars, err := model.FindOneProjectVars(projectId)
+	if err != nil {
+		return errors.Wrapf(err, "getting project variables for project '%s'", projectId)
+	}
+
+	repo, err := model.GetProjectSettingsById(project.ProjectRef.RepoRefId, true)
+	if err != nil {
+		return errors.Wrapf(err, "getting repo settings for repo '%s'", project.ProjectRef.RepoRefId)
+	}
+	repoVars, err := model.FindOneProjectVars(project.ProjectRef.RepoRefId)
+	if err != nil {
+		return errors.Wrapf(err, "getting repo variables for repo '%s'", project.ProjectRef.RepoRefId)
+	}
+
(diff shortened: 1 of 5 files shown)


rest/data/project_settings.go
```diff
@@ -97,6 +97,97 @@ func disableStartingSettings(p *model.ProjectRef) {
...
+	if err != nil {
+		return errors.Wrapf(err, "getting repo variables for repo '%s'", project.ProjectRef.RepoRefId)
+	}
+
+	// Add each promoted variable to existing repo vars
+	apiRepoVars := &restModel.APIProjectVars{}
+	apiRepoVars.BuildFromService(*repoVars)
+	for _, varName := range varNames {
+		// Ignore nonexistent variables
+		if _, contains := projectVars.Vars[varName]; !contains {
```
Kimchelly, line 127: nit: usually, the second variable from a map lookup just called `ok` (the ["comma-ok" idiom](https://go.dev/doc/effective_go#maps)) just like how error types are `err` unless you have a special need for a different variable name.

rest/data/project_settings.go
```diff
@@ -97,6 +97,97 @@ func disableStartingSettings(p *model.ProjectRef) {
...
+		if _, contains := projectVars.PrivateVars[varName]; contains {
+			apiRepoVars.PrivateVars[varName] = true
+		}
+		if _, contains := projectVars.AdminOnlyVars[varName]; contains {
+			apiRepoVars.AdminOnlyVars[varName] = true
+		}
+	}
+
+	if err = UpdateProjectVars(project.ProjectRef.RepoRefId, apiRepoVars, true); err != nil {
+		return errors.Wrapf(err, "updating project variables for project '%s'", projectId)
```
Kimchelly, line 141: `errors.Wrapf(... "updating repo ref variables for repo '%s'", project.ProjectRef.RepoRefId)`

rest/data/project_settings_test.go
```diff
@@ -561,6 +561,206 @@ func TestSaveProjectSettingsForSection(t *testing.T) {
 	}
 }
 
+func TestPromoteVarsToRepo(t *testing.T) {
+	for name, test := range map[string]func(t *testing.T, ref model.ProjectRef){
+		"SuccessfullyPromotesAllVariables": func(t *testing.T, ref model.ProjectRef) {
+			projectId := "pId"
+			repoId := "rId"
```
Kimchelly, line 568: Since the project ref is already passed into the test, it might be cleaner to just use `ref.Id` and `ref.RepoRefId` rather than re-define the IDs here.

rest/data/project_settings_test.go
```diff
@@ -561,6 +561,206 @@ func TestSaveProjectSettingsForSection(t *testing.T) {
...
+
+			varsToPromote := []string{"a", "b", "c"}
+			err := PromoteVarsToRepo(projectId, varsToPromote, "u")
+			assert.NoError(t, err)
+
+			projectVarsFromDb, err := model.FindOneProjectVars(projectId)
```
Kimchelly, line 574: nit: from [the style guide](https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit#heading=h.3rcuk9z7pv53), prefer to case this variable as `DB` instead of `Db` to follow acronym capitalization.

Kimchelly (approved):

rest/data/project_settings.go
```diff
@@ -97,6 +97,101 @@ func disableStartingSettings(p *model.ProjectRef) {
...
+		if _, contains := projectVars.AdminOnlyVars[varName]; contains {
+			apiRepoVars.AdminOnlyVars[varName] = true
+		}
+	}
+
+	if err = UpdateProjectVars(repoId, apiRepoVars, true); err != nil {
+		return errors.Wrapf(err, "adding variables from project '%s' to repo", projectId)
+	}
+
+	// Log repo update
```
ablack12, line 147: good catch!

rest/data/project_settings.go
```diff
@@ -97,6 +97,101 @@ func disableStartingSettings(p *model.ProjectRef) {
...
+	repoAfter, err := model.GetProjectSettingsById(repoId, true)
+	if err != nil {
+		return errors.Wrapf(err, "getting settings for repo '%s' after adding promoted variables", repoId)
+	}
+	if err = model.LogProjectModified(repoId, userId, repo, repoAfter); err != nil {
+		return errors.Wrapf(err, "logging repo '%s' modified", repoId)
+	}
+
+	// Remove promoted variables from project
+	apiProjectVars := &restModel.APIProjectVars{
```
ablack12, line 157: Rather than doing this, can the vars list just be passed into the VarsToDelete APIProjectVars field?

ablack12 (changes requested):

rest/data/project_settings.go
sophstad: I tried this but `VarsToDelete` does not work in the way we both probably thought! It requires the ProjectVars struct to already be updated with the values removed, or else a `ConflictingUpdateOperators` is thrown due to the `$set` and `$unset` operations; this [test](https://github.com/evergreen-ci/evergreen/blob/b6792a6094208e2b4e522ad2ad170401d1180ab7/model/project_vars_test.go#L156-L177) is a good example of how it works.

ablack12 (approved):

PR #6002, opened 2022-10-13
EVG-17291: bump gopkg.in/yaml.v3 to v3.0.1

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17291

### Description 
This bumps the YAML v3 package to the latest tagged version. The reason the upgrade caused the issue described in the ticket is because the minor version upgrade (from v3.0.0 to v3.0.1) fixed a bug in YAML parsing that was relied upon by a user. The bug is described in https://github.com/go-yaml/yaml/issues/81, but the gist is that the YAML library behaved in a way that is counter to [the official YAML spec with regards to nested map fields](https://yaml.org/type/merge.html) (such as command `params`). When the bug was present, it would resolve conflicting nested map fields by merging the maps together, but the actual spec behavior is that for conflicting fields, it should do a shallow merge rather than deep merge. For example, with a YAML like this:
```yaml
variables:
- &something
  params:
    foo: 1
    bar: 2

commands:
  - <<: *something
    params:
      quux: 3
```
Before, YAML would parse this as:
```yaml
commands:
    # merge the nested params fields together
  - params:
        foo: 1
        bar: 2
        quux: 3
```
However, the spec-compliant behavior now produces this:
```yaml
commands:
    # first write wins, ignore other params
  - params:
        quux: 3
```

This bug was fixed by https://github.com/go-yaml/yaml/commit/539c8e751b99281a79dc659ba8dc44d91e444723, which is included in the minor version upgrade, so the nested map merge is not possible anymore.

Some pre-work before I merge this:
* Filed [a ticket](https://jira.mongodb.org/browse/COMPASS-6200) with the affected project that depended on the YAML parser bug to update their YAML.
* Will do some comms at least a week in advance of deploying this change since it's possible that other projects were also relying on this behavior.

### Testing 
Tested this in a local playground to verify that the behavior described in the spec is what the upgraded YAML does.


ablack12 (approved):

go.mod
```diff
@@ -70,5 +66,5 @@ require (
 	github.com/shirou/gopsutil/v3 v3.22.9
 	github.com/trinodb/trino-go-client v0.300.0
 	google.golang.org/genproto v0.0.0-20211129164237-f09f9a12af12 // indirect
-	gopkg.in/20210107192922/yaml.v3 v3.0.0-00010101000000-000000000000
+	gopkg.in/yaml.v3 v3.0.1
```
bynn, line 69: is there a reason we specify only this version to be .1

Kimchelly: v3.0.1 is the latest stable version available. The previous thing was a workaround to pin it to one specific commit within v3.0.0, which is generally not something we'd want to do since individual commits aren't considered stable enough for release until they have a git tag release.

Kimchelly: `v3.0.1` is just stating the particular minimum version of `gopkg.in/yaml.v3` that's required. The version after the module name in the `go.mod` tells Go which version of the YAML module to use when `import gopkg.in/yaml.v3` is used in a source file.

bynn: ahh okay
thank you! i hope i wasnt holding this up from being merged

bynn (approved):

malikchaya2 (approved):

Kimchelly:
This upgrade is blocked until the deadline, so I'm going to close this PR For now and re-open with the updated changes once we're closer to the upgrade day.

PR #6004, opened 2022-10-13
EVG-16728: Achieve consistency across displayTaskPriority and addStat…

hadjri:
…usColorSort

### Description 

### Testing 

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/model/annotations"
 	"net/url"
 	"regexp"
 	"strings"
@@ -2009,8 +2010,26 @@ func (t *Task) GetDisplayStatus() string {
 	if t.DisplayStatus != "" {
 		return t.DisplayStatus
 	}
-	if t.Aborted && t.IsFinished() {
+	if t.Aborted {
 		return evergreen.TaskAborted
+	}
+	if t.Status == evergreen.TaskSucceeded {
+		return evergreen.TaskSucceeded
+	}
+	if t.Details.Type == evergreen.CommandTypeSetup {
+		return evergreen.TaskSetupFailed
+	}
+	if t.Details.Type == evergreen.CommandTypeSystem {
+		if t.Details.TimedOut && t.Details.Description == evergreen.TaskDescriptionHeartbeat {
+			return evergreen.TaskSystemUnresponse
+		}
+		if t.Details.TimedOut {
+			return evergreen.TaskSystemTimedOut
+		}
+		return evergreen.TaskSystemFailed
+	}
+	if t.Details.TimedOut {
+		return evergreen.TaskTimedOut
 	}
 	if t.Status == evergreen.TaskUndispatched {
 		if !t.Activated {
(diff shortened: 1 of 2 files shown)


model/task/task.go
```diff
@@ -2009,24 +2010,15 @@ func (t *Task) GetDisplayStatus() string {
 	if t.DisplayStatus != "" {
 		return t.DisplayStatus
 	}
-	if t.Aborted && t.IsFinished() {
+	if t.Aborted {
```
ablack12, line 2013: Should this function now also set t.DisplayStatus if it works the same way?

model/task/task.go
```diff
@@ -2036,12 +2028,25 @@ func (t *Task) GetDisplayStatus() string {
...
+		}
+		if t.Blocked() {
+			return evergreen.TaskStatusBlocked
+		}
+		return evergreen.TaskWillRun
+	}
+	if !t.IsFinished() {
+		return t.Status
+	}
+	annotation, err := annotations.FindOneByTaskIdAndExecution(t.Id, t.Execution)
```
ablack12, line 2046: Do we call getDisplayStatus in a lot of places? I'm a little wary about having a DB call here

model/task/task.go
```diff
@@ -2036,12 +2028,25 @@ func (t *Task) GetDisplayStatus() string {
...
+	if err == nil && annotation != nil {
+		return evergreen.TaskKnownIssue
```
ablack12, line 2048: I think it should only be a known issue if the annotation known issues list isn't empty

model/task/task.go
```diff
@@ -2050,31 +2055,21 @@ func (t *Task) GetDisplayStatus() string {
 // for example, if there are both successful and failed tasks, one would expect to see "failed"
 func (t *Task) displayTaskPriority() int {
```
ablack12, line 2056: I don't think these changes are right; the purpose of the order of case here is that if there are multiple execution tasks, it determines which status we should see (for example if any execution tasks are in progress, then the overall display task status would be in progress). 

Additionally each status has its own number to ensure that a list of statuses would always result in the same display task status. Display task priority is used to sort the execution tasks and then we use the "highest priority" execution task status as the display task status, so by giving multiple statuses the same priority it means that display task statuses can be inconsistent.

ablack12 (changes requested):

model/task/task.go
hadjri: Ok - I had initially done this since these shared statuses hold the same rank in `addStatusColorSort` but I see what you mean.  I've separated them out now so that the sorting can be consistent, and I've added a comment that makes it clear that there is a notion of display statuses that have the same badge color.  One way to organize this is to differentiate these priorities by the singles digits, not the tens digit.

model/task/task.go
hadjri: Sure - it might be that this should've always needed to been the case.

model/task/task.go
hadjri: We do call it often but this is a little bit tricky since looking at the task annotations collection is necessary to determine if the display status for a failed task is a "known issue".  Couple things that come to my mind: 

1) A large set of tasks will be caught in the above case and won't make it to this call, and since we now set `DisplayStatus` here it could also decrease the number of invocations we make here since we return early if `DisplayStatus` is set. 

2) We can remove this entirely and accept that in the case of known issue failures, there is a slight deviation between `displayStatusExpression` and `GetDisplayStatus`

ablack12 (changes requested):
Added a comment but forgot to hit the review button 🙃

model/task/task.go
hadjri: Ah - my understanding was that we follow the thought process of "If at least one exec task failed, there's no way the display task can succeed so we mark it as preemptively failed".

hadjri: Hmm, ok, thanks for the clarity. I'd say that given this some of the work here is still relevant because of [this comment](https://github.com/evergreen-ci/evergreen/blob/main/model/task/db.go#L176) saying `displayStatusExpression ` should reflect `GetDisplayStatus` which isn't currently exactly the case. I'll revert the `displayTaskPriority` and `addStatusColorSort` if they are fine the way they are.

model/task/task.go
khelif96: This concerns me as well. Especially if called from a graphql resolver this can easily spiral into thousands of calls. Could you ensure we always perform a db query that returns a displayStatus there?

hadjri:
I've updated the description and title of this PR to focus only on the changes made to `GetDisplayStatus`

model/task/task.go
ablack12: Love this refactor!

ablack12 (approved):

khelif96 (approved):
LGTM Great Refactor!

PR #6005, opened 2022-10-14
EVG-16860: Set task doc as failed due to abort immediately upon aborting

hadjri:
[EVG-16860](https://jira.mongodb.org/browse/EVG-16860)

### Description 
This is a follow up to this [previous PR](https://github.com/evergreen-ci/evergreen/pull/5996).  There are some instances in the code when resetting a version / build where the version/build's running tasks are marked as aborted and the finished tasks are reset.  With the previous PR's changes to set the failed status at the same time as setting the aborted field, an issue arose where Evergreen now thought that these tasks are in a state where we can reset them - which is not the case. This is because when we reset these tasks their secrets get reset, which prevents the agent from performing its normal function of calling `"/end"` on a task once it sees it has been aborted - which is essential task ending logic.

As it stands aborted tasks that have been stranded by their hosts will eventually get cleaned up by an existing job, although the job will incorrectly mark the task as failed due to "heartbeat". This change adds the same new aborted failure description as before, as well as a check in the `cleanupTimedOutTask` function to check if the task was marked as aborted or not and if so, fail it with the appropriate aborted description instead.
### Testing 
Added unit tests.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1643,13 +1643,11 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
 	return nil
 }
 
-// ResetStaleTask fixes a task that has either exceeded the heartbeat timeout
-// or has been marked as aborted but was never ended by the agent.
-// The current task execution is marked as finished and, if the task is not
-// aborted, a new execution is created to restart the task.
-// If the task is aborted, the task is marked as failed alongside necessary
-// logic that would typically be run by the agent.
-func ResetStaleTask(t *task.Task) error {
+// FixStaleTask fixes a task that has exceeded the heartbeat timeout.
+// The current task execution is marked as finished and, if the task was not
+// aborted, the task is reset. If the task was aborted, we do not reset the task
+// and it is just marked as failed alongside other necessary updates to finish the task.
+func FixStaleTask(t *task.Task) error {
 	CheckAndBlockSingleHostTaskGroup(t, t.Status)
 
 	failureDesc := evergreen.TaskDescriptionHeartbeat
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -497,7 +497,7 @@ func MarkEnd(t *task.Task, caller string, finishTime time.Time, detail *apimodel
 		detailsCopy.Status = evergreen.TaskFailed
 	}
 
-	if t.Status == detailsCopy.Status {
+	if detailsCopy.Description != evergreen.TaskDescriptionAborted && t.Status == detailsCopy.Status {
 		grip.Warning(message.Fields{
 			"message": "tried to mark task as finished twice",
 			"task":    t.Id,
@@ -1648,6 +1648,17 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
 // execution is created to restart the task.
 func ResetStaleTask(t *task.Task) error {
 	CheckAndBlockSingleHostTaskGroup(t, t.Status)
+
+	// Skip resetting the task if it was marked as aborted by the user
+	if t.Aborted {
+		grip.Info(message.Fields{
+			"message":            "timed out task was aborted, skipping reset",
+			"task":               t.Id,
+			"execution":          t.Execution,
+			"execution_platform": t.ExecutionPlatform,
+		})
+		return nil
+	}
 
 	if err := resetSystemFailedTask(t, evergreen.TaskDescriptionHeartbeat); err != nil {
 		return errors.Wrap(err, "resetting task")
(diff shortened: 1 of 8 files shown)


model/task_lifecycle.go
```diff
@@ -1643,18 +1643,27 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
...
+		}
+	} else {
+		if err := resetSystemFailedTask(t, failureDesc); err != nil {
+			return errors.Wrap(err, "resetting heartbeat task")
+		}
 	}
 
 	grip.Info(message.Fields{
-		"message":            "successfully fixed stale heartbeat task",
+		"message":            fmt.Sprintf("successfully fixed stale %s task", failureDesc),
```
Kimchelly, line 1666: Can we include the failure description a separate field to make Splunk searching easier?

model/task_lifecycle.go
```diff
@@ -1643,18 +1643,27 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
...
 	CheckAndBlockSingleHostTaskGroup(t, t.Status)
 
-	if err := resetSystemFailedTask(t, evergreen.TaskDescriptionHeartbeat); err != nil {
-		return errors.Wrap(err, "resetting task")
+	failureDesc := evergreen.TaskDescriptionHeartbeat
+	if t.Aborted {
+		failureDesc = evergreen.TaskDescriptionAborted
+		if err := t.MarkSystemFailed(failureDesc); err != nil {
+			return errors.Wrap(err, "marking aborted task as system failed")
+		}
```
Kimchelly, line 1658: I'm slightly confused how this will work if we're assuming that the agent end task route must be called for this abort to work. My question here is, if the task is not reset, we're requiring that the end task logic for this task will happen sometime. is it guaranteed that all the end task logic will actually happen? If the stale heartbeat job was triggered, it seems like the host/agent might already be in a bad state, so it might never hit the end task route to do the end task updates.

hadjri: This is meant to capture the case where a task is marked as `Aborted` in the task doc, but no agent ever hits the end task route to do the end task updates within the defined heartbeat timeout threshold. 

It's meant to perform cleanup on the bad state of these tasks and the logic is meant to be analagous to how the heartbeat tasks are cleaned up on L1660.  The only difference is that after the heartbeat tasks are marked system failed, they then get automatically reset and create a new execution, whereas if a task is marked as aborted we should just mark it as ended due to abort and stop there.

Kimchelly: I guess my question was, since `(*Task).MarkSystemFailed` just updates this one task and will not reset it to try again, should we expect this to do all the other task-ending logic as well (like update the commit queue, update dependencies, reset when finished, etc.)?

hadjri: Gotcha, I see.  Ok in this case the work you described that the agent would normally do that is not happening here is in `MarkEnd`.  

The execution timeout job already performs `CheckAndBlockSingleHostTaskGroup` in the same way the agent's end task route does, and `HandleEndTaskForCommitQueueTask` is a no-op for failed and aborted tasks so I've omitted that. The only change that I've made is to call `MarkEnd` rather than `(*Task).MarkSystemFailed`, which handles all of the dependency updating, reset when finished, etc logic.

model/task_lifecycle.go
```diff
@@ -1643,26 +1643,57 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
 	return nil
 }
 
-// ResetStaleTask fixes a task that has exceeded the heartbeat timeout by
-// marking the current task execution as finished and, if possible, a new
-// execution is created to restart the task.
+// ResetStaleTask fixes a task that has either exceeded the heartbeat timeout
```
Kimchelly, line 1646: Can we call this function something else, like `FixStaleTask` or something, since it no longer resets?

model/task_lifecycle.go
```diff
@@ -1643,26 +1643,57 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
...
+// or has been marked as aborted but was never ended by the agent.
```
Kimchelly, line 1647: I think this second comment is a subset of the first one, which is that the heartbeat timed out because the agent has supposedly stopped running this task. For clarity of reading this comment, can we remove this mention and substantiate the later part (on L1651) with a comment on why an aborting task is treated as a special case for the heartbeat timeout?

model/task_lifecycle.go
```diff
@@ -1643,26 +1643,57 @@ func ClearAndResetStrandedHostTask(h *host.Host) error {
...
+// The current task execution is marked as finished and, if the task is not
+// aborted, a new execution is created to restart the task.
+// If the task is aborted, the task is marked as failed alongside necessary
+// logic that would typically be run by the agent.
```
Kimchelly, line 1651: Instead of saying this is specifically on the agent, can we phrase this as something along the lines of "the task is marked as failed and any other updates to finish the task are done".

Kimchelly (commented):
Changes look good, just have some suggestions on doc comment phrasing so it's clear to readers why aborting is special.

Kimchelly (approved):

PR #6007, opened 2022-10-14
EVG-18064 Add help text for re-inviting the slack bot

malikchaya2:
[EVG-18064](https://jira.mongodb.org/browse/EVG-18064)

### Description
This adds help text to the notifications page to help users debug why they are not receiving notifications. 

### Screenshots
image

### Testing
Tested on staging 

### Spruce PR 
[1466](https://github.com/evergreen-ci/spruce/pull/1466)

### DIFF
--- a/public/static/partials/subscription_list.html
+++ b/public/static/partials/subscription_list.html
@@ -14,6 +14,10 @@
             <button class="pull-right plus-button btn btn-primary" type="button" ng-click="addSubscription()">
                 <i class="fa fa-plus"></i>
             </button>
+            <div class="muted small">
+                The evergreen bot needs to be re-invited to private slack channels for notifications to go through. You can do so by
+                running 'invite @evergreen-bot' in the channel.
+            </div>
         </h4>
         <div ng-show="subscriptions.length === 0">
             No event subscriptions. Click the plus to add some.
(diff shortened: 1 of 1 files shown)


bynn (approved):

ablack12:
I think help text was initially my idea but actually this wording is really specific to our use case to have in the modal text. Perhaps this should instead be something like "Private slack channels may require further Slack configuration" or something like that? Or maybe the name of the bot should be something in the admin channel that we can add, and this help text only displays if that is included.

malikchaya2:
@ablack12 considering that we have a longer term plan for this (having IT invite everyone automatically) and this is meant to carry us over till then, I think this is better. Would adding a note about removing this to https://jira.mongodb.org/browse/EVG-17483 make this feel better?

public/static/partials/subscription_list.html
```diff
@@ -14,6 +14,10 @@ <h4>Subscriptions
             <button class="pull-right plus-button btn btn-primary" type="button" ng-click="addSubscription()">
                 <i class="fa fa-plus"></i>
             </button>
+            <div class="muted small">
```
ablack12, line 17: Can we add a comment with the ticket number that we're going to fix this in? (also to the spruce PR)

ablack12 (approved):
Discussed offline; we're going to need this message long-term so will put up a fix to make it not hardcoded but will merge this for now to get some messaging out there for folks.

public/static/partials/subscription_list.html
malikchaya2: for sure

PR #6011, opened 2022-10-19
EVG-17569: ensure repotracker behaves correctly for branch changes

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17569

### Description 
The issue that the user experienced (that their periodic builds still use a revision from the old branch after updating to a new branch) is because of the fact that they changed their branch without making any more commits and the fact that periodic builds are always based on the last mainline commit. Periodic builds use the most recent mainline version as the basis for the periodic build version - they effectively `git clone` the project ref's current branch, then `git reset` to the last mainline commit's revision. What happened in this case is that the branch name in the project ref was updated, but no commit was made after the branch change. Therefore, the last mainline commit was a commit from the old branch, so `git reset` kept reusing that commit from the old branch.

I considered a bunch of alternative solutions (like trying to search for the last mainline commit that was on that particular branch, or trying to make the repotracker smart enough to figure out when the branches diverged), but all those solutions had issues with either edge cases or a bad user experience in terms of understanding what revision their own periodic build is based off of. Ultimately, this seems like an atypical use case for a project ref to change its branch (compared to the usual workflow of just creating different projects for different branches), so I don't think it's worth putting a ton of effort into supporting it.

The actual change basically doesn't directly fix the use case described in the ticket, but it makes Evergreen slightly more graceful in providing options for a project ref that changes its tracking branch. The current fix for this project will be to try catching the project up to the latest revision on `main` by either 1. manually re-running the repotracker or 2. pushing a commit to the new branch (`main`) (which basically just triggers the repotracker). 

* Ensure that revisions collected by repotracker respects admin setting for revision search limit (to avoid adding more revisions to the waterfall than it's configured to add).
* Do some cleanup here and there for repotracker code style. I didn't clean up the repotracker package along with the other packages because I thought "oh, we rarely touch the repotracker anyways, why bother", but here I am.
* Add a couple comments for some non-obvious logic in the repotracker.
* Fix some minor issues with how repotracker jobs are enqueued in Amboy.

### Testing 
Added test coverage to verify that the GitHub poller respects the revision search limit.

I tested a bunch in staging to verify that changing a branch and then force re-running the repotracker works as expected. The current behavior is that (assuming the old and new branches haven't diverged too greatly for this project), it will attempt to add new mainline versions for all the commits made to `main` since the branches diverged, up to the revision search limit.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -198,6 +198,9 @@ func (opts cloneOpts) buildHTTPCloneCommand() ([]string, error) {
 }
 
 func (opts cloneOpts) buildSSHCloneCommand() ([]string, error) {
+	// kim: NOTE: clones with commands:
+	// git clone git@github.com:wiredtiger/vendoring-tools.git vendoring-tools --branch main
+	// cd vendoring-tools
 	cloneCmd := fmt.Sprintf("git clone '%s' '%s'", opts.location, opts.dir)
 	if opts.recurseSubmodules {
 		cloneCmd = fmt.Sprintf("%s --recurse-submodules", cloneCmd)
@@ -265,6 +268,7 @@ func (c *gitFetchProject) buildCloneCommand(ctx context.Context, conf *internal.
 		fmt.Sprintf("rm -rf %s", c.Directory),
 	}
 
+	// kim: NOTE: get the git clone invocation for SSH clone.
 	cloneCmd, err := opts.getCloneCommand()
 	if err != nil {
 		return nil, errors.Wrap(err, "getting command to clone repo")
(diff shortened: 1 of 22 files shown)


hadjri (commented):
Code LGTM. One question - can we add another change to run a repotracker job when a project ref's branch changes to avoid needing to manually run or commit anything?

Kimchelly:
I did consider this and tried it out. I think manually triggering instead of automagically triggering the repotracker is still preferable in this case since this does break the continuity of the waterfall and this is solely for an unusual case of updating a branch which has no new commits pushed to it. It would also lead to some unexpected cases where we would have to manually trigger the repotracker every time the owner/repo/branch changes for the project. For example, we would also have to handle when an existing branch project is attached to a different repo.

hadjri:
> I did consider this and tried it out. I think manually triggering instead of automagically triggering the repotracker is still preferable in this case since it would lead to user questions around this magic behavior where their waterfall now looks substantially different and they didn't want it to do that. It would also lead to some unexpected cases where we would have to manually trigger the repotracker every time the owner/repo/branch changes for the project. For example, we would also have to handle when an existing branch project is attached to a different repo.
> 
> We could always revisit running the repotracker automatically if this sort of use case becomes more common.

Sounds fair, it does seem like a good bit more complexity. SGTM

hadjri (approved):

PR #6017, opened 2022-10-20
EVG-17291: add CLI option to validate against upgraded YAML version

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17291

### Description 
This is related to #6002 and adds a more user-friendly means to check if a project YAML will be invalidated by the upcoming YAML upgrade. Since we're not sure how many projects are affected by the change to how the YAML module treats nested maps (especially `params`) and I'm not sure people will know what to look for, I added a temporary flag to `evergreen validate` so users can conveniently check if their YAML is still valid using the upgraded YAML version. If their project YAML passes with `--use_upgraded_yaml`, their project config is still valid after the YAML upgrade; otherwise, they'll have to fix their YAML. We'll provide a grace period for them to fix the nested map merging dependency.

### Testing 
Tested `evergreen validate` against staging using [the project YAML with a known dependency on nested map merging](https://github.com/mongodb-js/compass/blob/d65d51613adab4a3e86cf062eac64d06ad9d861f/.evergreen.yml), with and without `--use_upgraded_yaml`. Without the upgraded YAML flag, it passes validation. With the upgraded YAML flag, it has errors.

I also tested against a YAML that I know doesn't have this nested map issue, and it passed validation both with and without the flag.

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -568,12 +568,12 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 // opts is used to look up files on github if the main parser project has an Include.
 func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, identifier string, project *Project) (*ParserProject, error) {
 	unmarshalStrict := false
-	checkUpgradedYAML := false
+	useUpgradedYAML := false
 	if opts != nil {
 		unmarshalStrict = opts.UnmarshalStrict
-		checkUpgradedYAML = opts.CheckUpgradedYAML
+		useUpgradedYAML = opts.UseUpgradedYAML
 	}
-	intermediateProject, err := createIntermediateProject(data, unmarshalStrict, checkUpgradedYAML)
+	intermediateProject, err := createIntermediateProject(data, unmarshalStrict, useUpgradedYAML)
 	if err != nil {
 		return nil, errors.Wrapf(err, LoadProjectError)
 	}
@@ -603,7 +603,7 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 		if err != nil {
 			return intermediateProject, errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, path.FileName)
 		}
-		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict, opts.CheckUpgradedYAML)
+		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict, opts.UseUpgradedYAML)
 		if err != nil {
 			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
 		}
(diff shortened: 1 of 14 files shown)


model/project_parser.go
```diff
@@ -797,9 +800,21 @@ func GetProjectFromFile(ctx context.Context, opts GetProjectOpts) (ProjectInfo,
 // intermediate project representation (i.e. before selectors or
 // matrix logic has been evaluated).
 // If unmarshalStrict is true, use the strict version of unmarshalling.
-func createIntermediateProject(yml []byte, unmarshalStrict bool) (*ParserProject, error) {
+func createIntermediateProject(yml []byte, unmarshalStrict, useUpgradedYAML bool) (*ParserProject, error) {
 	p := ParserProject{}
-	if unmarshalStrict {
+	if useUpgradedYAML && unmarshalStrict {
```
hadjri, line 805: Should we add a TODO here + JIRA ticket for after the grace period?

operations/validate.go
```diff
@@ -35,13 +34,17 @@ func Validate() cli.Command {
 		}, cli.StringFlag{
 			Name:  joinFlagNames(projectFlagName, "p"),
 			Usage: "specify project identifier in order to run validation requiring project settings",
+		}, cli.BoolFlag{
```
hadjri, line 37: One thing we can do to make users more aware of this pending upgrade is to include some text output when running `validate` saying that you'll need to ensure your validation still passes with the upgraded flag.

util/yaml.go
```diff
@@ -37,3 +38,33 @@ func UnmarshalYAMLStrictWithFallback(in []byte, out interface{}) error {
...
 }
+
+// UnmarshalUpgradedYAMLWithFallback is identical to UnmarshalYAMLWithFallback but uses the upgraded YAML version
+// (v3.0.1) instead of the current YAML version.
+// This should only be used for user validation to see if their project's YAML hypothetically produces a valid config
+// should the YAML version be upgraded.
+func UnmarshalUpgradedYAMLWithFallback(in []byte, out interface{}) error {
+	if err := upgradedYAML.Unmarshal(in, out); err != nil {
+		// try the older version of yaml before erroring, in case it's just an outdated yaml
+		if err2 := yaml2.Unmarshal(in, out); err2 != nil {
```
ablack12, line 49: I'm thinking we shouldn't use fallback here at all; the whole purpose of fallback was to avoid having to force users to update things to use yaml3 since lots of projects had issues with it, but if we allow falling back to yaml2 in the upgraded case then we _will_ end up hiding the errors people are explicitly looking for with the flag, since the nested issue probably won't fail for yaml2.

bynn:
would it be more visible to users to set --use_upgraded to true by default so that they know there's a problem and can set it to false manually if they want to validate?

operations/validate.go
Kimchelly: I hesitate to do this because a bunch of users script using `evergreen validate` and may break if they use `--quiet` and expect no output. My expectation is that if they don't know about the YAML upgrade already, they're unlikely to happenstance use `evergreen validate` manually (aside from in a script, where they won't check the output).

util/yaml.go
Kimchelly: The fallback is only necessary if their YAML is not already v3 compatible, which is not affected by the YAML upgrade. If it's not v3 compatible, it will continue being incompatible and parse with v2 fallback logic (i.e. the v3 YAML upgrade is completely irrelevant to them). If it _is_ v3 compatible but is subject to the nested map bug, it will not error, but will instead produce a different (but valid) parsed result, which can only be checked by later logic.

Kimchelly (commented):
We can't use `--check_upgraded_yaml` by default because people write scripts and workflows using `evergreen validate`. If their YAML is incompatible, these will break. Giving them a grace period to fix their YAML (a code fix) means we probably shouldn't break their existing critical scripts either (also a code fix).

hadjri (approved):

hadjri:
LGTM mod merge conflict resolution

ablack12 (approved):

model/project_parser.go
Kimchelly: Forgot to respond to this earlier, but I'm gonna keep this ticket open until this temporary logic is removed, so I don't think we need another follow-up ticket.

PR #6021, opened 2022-10-24
EVG-17402: Correct base statuses for task tests on commits

minnakt:
[EVG-17402](https://jira.mongodb.org/browse/EVG-17402)

### Description 
This PR makes it so that base statuses for task tests are correct on mainline commits.

Currently the base statuses for task tests are correct on personal patches, but not on commits. This happens because we call the function [`FindTaskOnBaseCommit`](https://github.com/evergreen-ci/evergreen/blob/f2059199a605e918e1b0c89c570acf53f360e167/graphql/query_resolver.go#L709). The base commit of a commit is actually itself, which means this function does not work for commits.

In the case we are looking at a commit we actually need to call the function [`FindTaskOnPreviousCommit`](https://github.com/evergreen-ci/evergreen/blob/f2059199a605e918e1b0c89c570acf53f360e167/model/task/task.go#L943).

### Screenshots
Below is a screenshot of my changes applied on staging. Here is a [link](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1804_arm64_test_agent_command_850e9262aea2193372d50475a63addda79ba1e39_22_10_17_19_48_59/tests?execution=0&sortBy=STATUS&sortDir=ASC) showing what this page looks like without my changes applied (notice that the base statuses of the first two tests are different).

Screen Shot 2022-10-24 at 3 44 44 PM

### Testing 
- On staging
- Added [taskTests/queries/base-status-commit.graphql](https://github.com/evergreen-ci/evergreen/compare/main...minnakt:EVG-17402?expand=1#diff-155b97b3d60b0d89e8a4b6eaa5d5ba6ea364468a57da52cd83753eecde2d9bd4) to test for base statuses on commits
- Added [taskTests/queries/base-status-patch.graphql](https://github.com/evergreen-ci/evergreen/compare/main...minnakt:EVG-17402?expand=1#diff-607955765cc684255f166ab95567a637cd0c5d0843d3d6187695366225b02f7d) to test for base statuses on patches


khelif96 (approved):
LGTM

graphql/query_resolver.go
```diff
@@ -618,11 +618,17 @@ func (r *queryResolver) TaskLogs(ctx context.Context, taskID string, execution *
 func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution *int, sortCategory *TestSortCategory, sortDirection *SortDirection, page *int, limit *int, testName *string, statuses []string, groupID *string) (*TaskTestResult, error) {
 	dbTask, err := task.FindByIdExecution(taskID, execution)
 	if dbTask == nil || err != nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("finding task with id %s", taskID))
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("finding task with id '%s'", taskID))
```
bynn, line 621: can we add execution is the message

bynn (approved):

graphql/query_resolver.go
```diff
@@ -618,11 +618,17 @@ func (r *queryResolver) TaskLogs(ctx context.Context, taskID string, execution *
...
+	}
+
+	var baseTask *task.Task
+	if evergreen.IsPatchRequester(dbTask.Requester) {
+		baseTask, err = dbTask.FindTaskOnBaseCommit()
+	} else {
+		baseTask, err = dbTask.FindTaskOnPreviousCommit()
```
Kimchelly, line 628: I don't know what is the current/desired behavior here, so I'm just asking to make sure - Will this do the correct thing for non-commit tasks on the waterfall, like periodic builds and downstream builds? For example, periodic builds are usually a re-run of the latest commit.

minnakt: I ended up swapping the conditional to only look for previous commit if the requester is `"gitter_request"`!

From my understanding, patch requesters, periodic build requesters, git tag requesters, and trigger requesters should look back at their base commit based on the `revision` rather than the `order` number. I added more tests to verify this works as intended.

Let me know if that sounds right!

Kimchelly: This sounds good to me! 👍

graphql/query_resolver.go
```diff
@@ -618,14 +618,15 @@ func (r *queryResolver) TaskLogs(ctx context.Context, taskID string, execution *
 func (r *queryResolver) TaskTests(ctx context.Context, taskID string, execution *int, sortCategory *TestSortCategory, sortDirection *SortDirection, page *int, limit *int, testName *string, statuses []string, groupID *string) (*TaskTestResult, error) {
 	dbTask, err := task.FindByIdExecution(taskID, execution)
 	if dbTask == nil || err != nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("finding task with id '%s'", taskID))
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("finding task with id '%s' and execution %v", taskID, utility.FromIntPtr(execution)))
```
Kimchelly, line 621: nit: you can use `%d` formatting instead of `%v` here.

Kimchelly (approved):
LGTM mod nit.

PR #6027, opened 2022-10-25
EVG-17477 Update subscriber target and placeholder text 

malikchaya2:
[EVG-17477](https://jira.mongodb.org/browse/EVG-17477)

### Description 
I updated everywhere I can find that uses user.Settings.SlackUsername as the target automatically. I feel like there are more but I couldn't find them, please let me know if you know what I'm missing. 

I also updated some placeholder text. 

### Testing 
None because the changes are very basic.


public/static/partials/subscription_modal.html
```diff
@@ -60,7 +60,7 @@ <h4>[[ c.verb ]] Subscription</h4>
                     </div>
                     <div ng-show="method.value === 'slack'">
                         <label for="slack">Slack Username or Channel</label>
-                        <input id="slack" ng-model="targets['slack']" placeholder="@user"></input>
+                        <input id="slack" ng-model="targets['slack']" placeholder="MEMBERID, @user or #channel"></input>
```
ablack12, line 63: Do we expect users to input memberID? I don't think they're going to know how to find that; we may want to keep this simpler.

ablack12 (changes requested):
I searched just SlackUsername and I see a bunch of usages in data/user.go to build subscribers. I am wondering if old notifications are still going to work that use username or if we're going to have to do some sort of update on subscriptions.

repotracker/repotracker.go
```diff
@@ -626,7 +626,11 @@ func makeBuildBreakSubscriber(userID string) (*event.Subscriber, error) {
 		if preference == user.PreferenceEmail {
 			subscriber.Target = u.Email()
 		} else if preference == user.PreferenceSlack {
-			subscriber.Target = u.Settings.SlackUsername
+			slackTarget := u.Settings.SlackUsername
```
bynn, line 629: style(?) would it be more readable if we used a switch to check all the different preference options?

rest/data/user.go
```diff
@@ -20,16 +20,23 @@ func UpdateSettings(dbUser *user.DBUser, settings user.UserSettings) error {
...
 		}
 	}
+
 	settings.SlackUsername = strings.TrimPrefix(settings.SlackUsername, "@")
 	settings.Notifications.PatchFinishID = dbUser.Settings.Notifications.PatchFinishID
 	settings.Notifications.SpawnHostOutcomeID = dbUser.Settings.Notifications.SpawnHostOutcomeID
 	settings.Notifications.SpawnHostExpirationID = dbUser.Settings.Notifications.SpawnHostExpirationID
 	settings.Notifications.CommitQueueID = dbUser.Settings.Notifications.CommitQueueID
 
+	slackTarget := fmt.Sprintf("@%s", settings.SlackUsername)
```
bynn, line 30: can we move it under the if check so that we don't do this if settings.SlackMembeId != ""

rest/route/host_spawn.go
```diff
@@ -1296,9 +1296,15 @@ func validateID(id string) (string, error) {
 func makeSpawnHostSubscription(hostID, subscriberType string, user *user.DBUser) (model.APISubscription, error) {
 	var subscriber model.APISubscriber
 	if subscriberType == event.SlackSubscriberType {
+
+		target := fmt.Sprintf("@%s", user.Settings.SlackUsername)
```
bynn, line 1300: only do this if the if check below fails

bynn (changes requested):

repotracker/repotracker.go
malikchaya2: I agree, but it's a bit out of scope, and I think it's small enough to be worth keeping like this.

rest/data/user.go
malikchaya2: It's easier if I initialize it to one thing to start with, and since this is just a print and not a db check, I think it's okay. If I move it under the if, I'd need to have it in an else which feels less clean to me.

rest/route/host_spawn.go
malikchaya2: same as above, having it overwrite it if it's available feels more intuitive to me.

malikchaya2: It saves a blank initialization and else statement

bynn (approved):
bump so it's not block on me!

public/static/partials/subscription_modal.html
ablack12: Does that mean the script will need to run through old notifications and update them to use memberIDs or anything like that?

malikchaya2: The script will simply add the memberID to all users. Old ones that have the slack userName will still work, but new ones that are created will have the memberID. I think at this point updating all old notifications seems overkill since it still works as expected.

ablack12: If we expect it to stop working as expected then it is something we'll need to do, at least for project level subscriptions, even though it sounds like we don't have a timeline on that. 

Anyway this all makes sense to me -- can we just change the text to say Member ID instead of MEMBERID since that's a bit more user-friendly as text?

ablack12 (approved):
LGTM with text change

public/static/partials/subscription_modal.html
ablack12: (We also may want to update in spruce; not sure if that's already been done)

malikchaya2: We had a similar discussion on the Spruce PR. The reason it is "MEMBERID" is to provide a hint to the user about the format of the memberID, which will look like this: ASDLY25TL.

PR #6028, opened 2022-10-25
EVG-18131: exclude hosts running tasks from self-throttle

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18131

### Description 
Hosts that are already running a task should be excluded from the self-throttle count of provisioning hosts. If they're already running a task, they're already provisioned.

### Testing 
There's no unit tests for this query or for the self-throttling behavior, so I didn't think it was worth improving this test coverage issue here for such a tiny change.

### DIFF
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -268,8 +268,8 @@ func CountAllRunningDynamicHosts() (int, error) {
 	return num, errors.Wrap(err, "counting running dynamic hosts")
 }
 
-// CountIdleStartedTaskHosts returns the count of task hosts that are not
-// currently running a task.
+// CountIdleStartedTaskHosts returns the count of task hosts that are starting
+// and not currently running a task.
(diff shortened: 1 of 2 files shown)


model/host/db.go
```diff
@@ -267,16 +268,13 @@ func CountAllRunningDynamicHosts() (int, error) {
 	return num, errors.Wrap(err, "counting running dynamic hosts")
 }
 
-func CountStartedTaskHosts() (int, error) {
-	num, err := Count(db.Query(startedTaskHostsQuery("")))
+// CountIdleStartedTaskHosts returns the count of task hosts that are not
+// currently running a task.
```
ybrill, line 272: It seems worth mentioning that the count is also limited to hosts in a starting state.

ybrill (approved):

model/host/db.go
Kimchelly: Good idea, added.

PR #6029, opened 2022-10-26
EVG-18031 Update setPatchToPreviousPatchDefinition to only schedule activated tasks

bynn:
[EVG-18031](https://jira.mongodb.org/browse/EVG-18031)

### Description 
Server has a workflow where they generate inactive tasks and this was being scheduled when using the --repeat flag for patches

### Testing 
unit test


ablack12 (changes requested):
Can you add a new or ore specific test for the restart case, since it seems there are a couple of bugs we should add tests to make sure we feel confident about the implementation.

units/patch_intent.go
```diff
@@ -521,15 +522,33 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 			return "", errors.Wrap(err, "settings tasks to previous failed")
 		}
 
 	} else {
-		patchDoc.Tasks = previousPatch.Tasks
+		// only add activated tasks from previous patch
+		query := db.Query(bson.M{
+			"$and": []bson.M{
+				{task.ProjectKey: project.Identifier},
+				{task.DisplayNameKey: bson.M{"$in": previousPatch.Tasks}},
```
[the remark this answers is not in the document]
bynn: added version id

units/patch_intent.go
```diff
@@ -522,14 +523,27 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 	patchDoc.BuildVariants = previousPatch.BuildVariants
-
 	if failedOnly {
 		if err = setTasksToPreviousFailed(patchDoc, previousPatch, project); err != nil {
 			return "", errors.Wrap(err, "settings tasks to previous failed")
 		}
-
 	} else {
-		patchDoc.Tasks = previousPatch.Tasks
+		// only add activated tasks from previous patch
```
ablack12, line 531: nit: capitalize

units/patch_intent.go
```diff
@@ -522,14 +523,27 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+		query := db.Query(bson.M{
+			task.ProjectKey:     project.Identifier,
+			task.VersionKey:     previousPatch.Version,
+			task.DisplayNameKey: bson.M{"$in": previousPatch.Tasks},
+			task.ActivatedKey:   true,
+		}).WithFields(task.DisplayNameKey, task.BuildVariantKey)
```
ablack12, line 537: Doesn't look like we're using build variant key

units/patch_intent.go
```diff
@@ -522,14 +523,27 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 			return "", errors.Wrap(err, "settings tasks to previous failed")
 		}
-
 	} else {
-		patchDoc.Tasks = previousPatch.Tasks
+		// only add activated tasks from previous patch
+		query := db.Query(bson.M{
+			task.ProjectKey:     project.Identifier,
+			task.VersionKey:     previousPatch.Version,
+			task.DisplayNameKey: bson.M{"$in": previousPatch.Tasks},
```
ablack12, line 535: Want to check on the index question; does using display name get us to a better index? Otherwise it seems redundant since any tasks matching the version should be considered

ablack12 (commented):
Want to make sure you saw my comment about adding more tests.

units/patch_intent.go
ablack12: @bynn approval actually pending this; can you confirm that you've verified there's a good index fit for this?

bynn (commented):
There is a bug that I found while testing this ticket and it's that tasks with the same name will be scheduled in other variants if you use --repeat (confirmed that it exists is prod) 

Created a ticket for it https://jira.mongodb.org/browse/EVG-18202

units/patch_intent.go
ablack12: Idk what you're talking about 👀

ablack12 (approved):

PR #6030, opened 2022-10-26
EVG-18129: upgrade Amboy and improve global env testing

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18129

### Description 
* Upgrade Amboy to provide option to configure desired index and remove hard-coded hotfix.
* Add a setting to make the preferred index usage configurable based on the execution environment. For staging/production, it will always use the preferred index using `mod_ts`. For testing environments, we don't want to depend on the existence of that particular index, so I added field to skip using the preferred index when using `testutil.NewEnvironment`, `mock.Environment`, or loading the admin settings from a YAML/JSON file.
    * This will only skip the preferred index if it's explicitly set in the DB or if it's set in the test config YAML/JSON, so there's not really a chance of it accidentally skipping the preferred index in staging/production.
* Fix a flaky test scenario where running `make test-evergreen`, and then running `make test-units` would fail the cache historical stats tests because one of the admin settings tests disabled all the service flags. The admin settings test suite now resets the flags to their original state before the test.
* Fix a bug where the agent would try to unmarshal the agent setup data from the app server into a non-pointer type. Supposedly, this means that the agent has been failing to load the agent setup data for some time, but seemingly had no noticeable ill effects.

### Testing 
* Existing tests pass to verify that preferred index is disabled for testing.
* Ran it in staging to verify that the queues still run and use the preferred index.


ybrill:
re the agent setup data, if we aren't using it can it just be removed?

Kimchelly (commented):
I think the agent setup data is still used for purposes that don't alert us like [uploading agent logs to S3](https://github.com/evergreen-ci/evergreen/blob/552c5e1de0125ff62e84979404fe12595e5d3bf3/agent/logging.go#L200-L215) in case we want to look them over after the host is already terminated.

ybrill:
Are there agent logs that are supposed to go to S3? 
afaict system/task/agent logs are sent to cedar and the agent's grip.<whatever> logs are not included in the S3 upload.

Kimchelly:
As far as I'm aware, the agent is supposed to send `grip` logs to Splunk if they're a high enough threshold, and otherwise log to the file unconditionally, which then gets uploaded to S3. From a glance at [the actual bucket prefix](https://s3.console.aws.amazon.com/s3/buckets/mciuploads?prefix=logs/&region=us-east-1), it seems like it was uploading agent log files into at least 2021. If it's not doing this, it's probably a regression.

I can open a ticket to investigate if we still find the agent grip logs useful enough to keep putting them in S3. Regardless of the status of the S3 log files, I believe the agent setup data is still necessary because it uses the Splunk information to log from the agent directly to Splunk.

ybrill:
Ah I see [what you're saying about Splunk](https://github.com/evergreen-ci/evergreen/blob/c34370f61dab2cd2b31d31ecaf6e0713e677bc9d/agent/logging.go#L59-L63). Thanks so much for finding this! It's amazing we never realized we were missing it.

I'm definitely with you that it was uploading _something_ it just seems like it's duplicating what we already have in cedar. I could be wrong though. The ones I clicked through were deep glaciered. I'm cool with deferring further investigation 😎 .

ybrill (approved):
It would be nice if for [EVG-18141](https://jira.mongodb.org/browse/EVG-18141) we discover we could do without the hint.

Kimchelly:
I agree it would be nice to not have the hint, but I suspect that we'll have to keep it since the query planner has shown it may continue to pick the wrong index.

PR #6032, opened 2022-10-26
EVG-16337 move some queries to unclutter tasks

ablack12:
[EVG-16337](https://jira.mongodb.org/browse/EVG-16337)

### Description 
This wasn't worth thinking too much about since it's a relatively low-priority tech debt ticket but I wanted to move some of the more intense queries out of task and into db.go (I think task should be more operational update logic generally).

**Did not change any existing function so no need to review the logic**

### Testing 
Moved relevant testing to db_test.go


model/task/task.go
```diff
@@ -2645,7 +2659,7 @@ func (t *Task) Archive() error {
 					},
 				},
 			},
-			updateDisplayTasksAndTasksBson(),
+			updateDisplayTasksAndTasksExpression,
```
malikchaya2, line 2662: Where is this defined? My grepping is letting me down here.

malikchaya2: The file wasn't expanded and it of course wasn't in the code base yet. Found it!

malikchaya2 (approved):

PR #6033, opened 2022-10-27
EVG-7086: Remove support for Version Config string

hadjri:
[EVG-7086](https://jira.mongodb.org/browse/EVG-7086)

### Description 
Versions without parser projects will no longer work since the config string is no longer supported.  The parser project collection has been in use for quite a while so it should be safe to remove without impacting users.

### Testing 
Refactored existing unit tests that relied on version config strings


model/project_parser.go
```diff
@@ -508,50 +508,17 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
...
-	if v.Config == "" {
-		return ProjectInfo{}, errors.New("version has no config")
-	}
-	p := &Project{}
-	// opts empty because project yaml with `include` will not hit this case
-	ctx := context.Background()
-	pp, err = LoadProjectInto(ctx, []byte(v.Config), nil, id, p)
-	if err != nil {
-		return ProjectInfo{}, errors.Wrap(err, "loading project")
+	if pp.Functions == nil {
```
ablack12, line 511: I'm not sure why this isn't just part of TranslateProject; maybe because we wanted to upsert it correctly? does it seem possible to move it in? (or into findOne)

model/project_parser.go
```diff
@@ -486,7 +486,7 @@ func (pss *parserStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) e
 
 // LoadProjectForVersion returns the project for a version, either from the parser project or the config string.
 // If read from the config string and shouldSave is set, the resulting parser project will be saved.
-func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo, error) {
+func LoadProjectForVersion(v *Version, id string) (ProjectInfo, error) {
```
ablack12, line 489: Can we audit usages of this function? I'm wondering if (now that we don't need version config) we can just use a smaller function to get and translate the parserProject (and not the projectRef and config) in some places (and not have to look up version since we could just use the id)

model/version.go
```diff
@@ -31,7 +31,6 @@ type Version struct {
 	Message             string               `bson:"message" json:"message,omitempty"`
 	Status              string               `bson:"status" json:"status,omitempty"`
 	RevisionOrderNumber int                  `bson:"order,omitempty" json:"order,omitempty"`
-	Config              string               `bson:"config" json:"config,omitempty"`
 	ConfigUpdateNumber  int                  `bson:"config_number" json:"config_number,omitempty"`
```
ablack12, line 34: We can also get rid of this I think

operations/cli_integration_test.go
```diff
@@ -3,6 +3,7 @@ package operations
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/util"
```
ablack12, line 6: Should move down

units/task_monitor_execution_timeout_test.go
```diff
@@ -2,6 +2,7 @@ package units
 
 import (
 	"context"
+	"github.com/evergreen-ci/evergreen/util"
```
ablack12, line 5: should move down

ablack12 (changes requested):

ablack12:
FYI spruce e2e is broken rn bc of build; Mohamed is fixing but we should probably make sure that works before merging

model/project_parser.go
hadjri: Looks like `createVersionForTag` is the only invocation that needs the `Ref` and `Config` fields to be set on the `ProjectInfo`.  I've refactored most usages to use a new function `FindAndTranslateProjectForVersion` that will as you suggest, only get and translate the parser proj without checking for project ref or project config (reducing DB calls). I'm keeping this function here but it now uses `FindAndTranslateProjectForVersion` as a helper.

model/project_parser.go
hadjri: Not sure why it was there either - there are two instances where we do that - one here and one in the agent route that retrieves the parser project. Since both of them don't use TranslateProject I think it's ok to move it into `ParserProjectFindOneById`

model/project_parser.go
ablack12: Awesome! Can we rename this function to LoadProjectInfoForVersion to distinguish it more?

model/project_parser.go
```diff
@@ -484,10 +484,20 @@ func (pss *parserStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) e
...
-	var pp *ParserProject
+// FindAndTranslateProjectForVersion translates a parser project for a version into a Project
+func FindAndTranslateProjectForVersion(v *Version, id string) (*Project, *ParserProject, error) {
+	pp, err := ParserProjectFindOneById(v.Id)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "finding parser project")
+	}
+	pp.Identifier = utility.ToStringPtr(id)
+	var p *Project
+	p, err = TranslateProject(pp)
```
ablack12, line 495: I think it's more correct to return nil for the project/parser project if there's an error

model/project_parser.go
```diff
@@ -497,61 +507,22 @@ func LoadProjectForVersion(v *Version, id string, shouldSave bool) (ProjectInfo,
...
-			return ProjectInfo{}, errors.Wrap(err, "updating version with project")
-		}
+		return ProjectInfo{}, errors.Wrap(err, "translating project")
 	}
 	return ProjectInfo{
 		Project:             p,
 		IntermediateProject: pp,
-	}, nil
+		Config:              pc,
+	}, err
```
ablack12, line 525: same here

ablack12 (approved):
LGTM with a couple of nits!

PR #6035, opened 2022-10-28
EVG-17375 Fetch updated patch module commit sha upon patch creation

hadjri:

### Description 
Patches that refer to other projects as modules do not get their own dedicated manifest associated with the patch version, rather they have to defer back to the last repotracker version that they were created off of.  This means that all patches running of a given mainline commit will not respond to changes in the HEAD of the branches of the modules that they track if new commits come in for those projects.

If commit hashes are present in the `Patches` field of a patch however, the `manifest/load` endpoint will add these commit hashes to a module override list and use these newer hashes instead.  Currently Github and CLI patches don't add updated module hashes to this `Patches` field - change has been made to upon patch creation run through the modules list for a project and either update the commit hash for it if it already exists in `Patches`, or to add a new entry in `Patches` specifying the latest commit hash for the module.

### Testing 
Added CQ sandbox to the self-tests.yml module list and submitted a patch in staging. I checked the `manifest/load` response before and after I added a trivial commit to the CQ sanbox.  The commit SHA for sandbox was correctly updated.

This did not occur if this change was not deployed to staging.

Added unit test.


model/manifest/db.go
```diff
@@ -77,22 +77,32 @@ func FindFromVersion(versionID, project, revision, requester string) (*Manifest,
...
+	moduleOverrides, err := GetManifestModuleOverrides(requester, versionID)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting module overrides")
 	}
+	manifest.ModuleOverrides = moduleOverrides
 
 	return manifest, err
 }
+
+func GetManifestModuleOverrides(requester, versionID string) (map[string]string, error) {
```
ablack12, line 89: We should comment this since it's a public function

model/manifest/db.go
```diff
@@ -77,22 +77,32 @@ func FindFromVersion(versionID, project, revision, requester string) (*Manifest,
...
+	if !evergreen.IsPatchRequester(requester) {
+		return nil, nil
+	}
+	var p *patch.Patch
```
ablack12, line 93: no need to explicitly declare this

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 
+// updatePatchDocPatches updates the commit sha for existing patch modules and adds new modules to the patch for modules
+// that exist in the project but not in the patch.  This is necessary so that the manifest load endpoint will include the
+// updated commit sha in the response's module overrides field.
+func updatePatchDocPatches(ctx context.Context, patchDoc *patch.Patch, project *model.Project, token string) error {
+	var err error
+	for i := range patchDoc.Patches {
+		if patchDoc.Patches[i].ModuleName != "" {
+			// validate the module exists
+			var module *model.Module
```
ablack12, line 537: no need to explicitly declar this or the err above

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
 }
 
+// updatePatchDocPatches updates the commit sha for existing patch modules and adds new modules to the patch for modules
+// that exist in the project but not in the patch.  This is necessary so that the manifest load endpoint will include the
+// updated commit sha in the response's module overrides field.
+func updatePatchDocPatches(ctx context.Context, patchDoc *patch.Patch, project *model.Project, token string) error {
+	var err error
+	for i := range patchDoc.Patches {
+		if patchDoc.Patches[i].ModuleName != "" {
+			// validate the module exists
```
ablack12, line 536: nit : capitlize

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+			var module *model.Module
+			module, err = project.GetModuleByName(patchDoc.Patches[i].ModuleName)
```
ablack12, line 538: can we set patchDoc.Patches[i].Modulename above to shorten these expressions?

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+				return errors.Wrapf(err, "finding module '%s'", patchDoc.Patches[i].ModuleName)
+			}
+			if module == nil {
+				return errors.Errorf("module '%s' not found", patchDoc.Patches[i].ModuleName)
+			}
+			branch, err := getBranch(ctx, module.Repo, patchDoc.Patches[i].ModuleName, module.Branch, token)
+			if err != nil {
+				return errors.Wrapf(err, "getting branch for module '%s'", patchDoc.Patches[i].ModuleName)
+			}
+			patchDoc.Patches[i].Githash = *branch.Commit.SHA
```
ablack12, line 549: utility.FromStringPtr

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+			Githash:    *branch.Commit.SHA,
+			ModuleName: mod.Name,
+		})
+
+	}
+	return nil
+}
+
+// getBranch retrieves the branch for the given repo, module, and branch name.
+func getBranch(ctx context.Context, repo, name, moduleBranch, token string) (*github.Branch, error) {
```
ablack12, line 577: Should this be moved somewhere else (maybe to the manifest file?) and it could just return the githash, and then also be used by the commit queue function https://github.com/ablack12/evergreen/blob/566384cbfed96d6b28eaa6678d3c087de227e625/units/commit_queue.go#L703

units/patch_intent_test.go
```diff
@@ -177,7 +177,7 @@ func (s *PatchIntentUnitsSuite) TearDownTest() {
 	s.cancel()
 }
 
-func (s *PatchIntentUnitsSuite) makeJobAndPatch(intent patch.Intent) *patchIntentProcessor {
+func (s *PatchIntentUnitsSuite) makeJobAndPatch(intent patch.Intent, patcherParserProject string) *patchIntentProcessor {
```
ablack12, line 180: patchedParserProject* I assume

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+				return errors.Wrapf(err, "getting branch for module '%s'", patchDoc.Patches[i].ModuleName)
+			}
+			patchDoc.Patches[i].Githash = *branch.Commit.SHA
+		}
+	}
+	for _, mod := range project.Modules {
+		patchFound := false
+		for _, p := range patchDoc.Patches {
+			if p.ModuleName == mod.Name {
+				patchFound = true
```
ablack12, line 556: we could just continue here; we don't need the variable

units/patch_intent.go
```diff
@@ -535,6 +526,69 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+				return errors.Errorf("module '%s' not found", patchDoc.Patches[i].ModuleName)
+			}
+			branch, err := getBranch(ctx, module.Repo, patchDoc.Patches[i].ModuleName, module.Branch, token)
+			if err != nil {
+				return errors.Wrapf(err, "getting branch for module '%s'", patchDoc.Patches[i].ModuleName)
+			}
+			patchDoc.Patches[i].Githash = *branch.Commit.SHA
+		}
+	}
+	for _, mod := range project.Modules {
```
ablack12, line 552: Could this just be the main loop, and in the if case at 556 we do the work above, otherwise we do 563? (Seems like  we're doing duplicate work, since above we only continue to getBranch if module exists in the project)

ablack12 (changes requested):

units/patch_intent.go
hadjri: Done

units/patch_intent.go
```diff
@@ -535,6 +525,30 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+		if err != nil {
+			return errors.Wrapf(err, "getting branch for module '%s'", mod.Branch)
+		}
+		for i := range patchDoc.Patches {
+			patchName := patchDoc.Patches[i].ModuleName
+			if patchName == mod.Name {
+				patchDoc.Patches[i].Githash = sha
+			}
+		}
+		patchDoc.Patches = append(patchDoc.Patches, patch.ModulePatch{
```
ablack12, line 543: This should only happen if we don't find the module name at 539, right? Should we add a continue up there?

units/patch_intent.go
```diff
@@ -535,6 +525,30 @@ func (j *patchIntentProcessor) setToPreviousPatchDefinition(patchDoc *patch.Patc
...
+// that exist in the project but not in the patch. This is necessary so that the manifest load endpoint will include the
+// updated commit sha in the response's module overrides field.
+func updatePatchDocPatches(ctx context.Context, patchDoc *patch.Patch, project *model.Project, token string) error {
+	for _, mod := range project.Modules {
+		sha, err := thirdparty.GetBranchCommitHash(ctx, mod.Repo, mod.Branch, token)
+		if err != nil {
+			return errors.Wrapf(err, "getting branch for module '%s'", mod.Branch)
+		}
+		for i := range patchDoc.Patches {
+			patchName := patchDoc.Patches[i].ModuleName
```
ablack12, line 538: patchName might be misleading; patchModuleName maybe?

ablack12 (changes requested):

units/patch_intent.go
hadjri: Yes 🤦‍♂️ I had meant to add some logic to enforce that but evidently not! Fixed now. I think a `found` variable is necessary instead of adding a continue though.

units/patch_intent_test.go
```diff
@@ -656,6 +659,79 @@ func (s *PatchIntentUnitsSuite) TestProcessCliPatchIntent() {
...
+		GithubPRTestingDisabled: true,
+	}
+	s.NoError(evergreen.SetServiceFlags(flags))
+
+	patchContent, _, err := thirdparty.GetGithubPullRequestDiff(context.Background(), githubOauthToken, s.githubPatchData)
+	s.Require().NoError(err)
+
+	patchedParserProject := `
+modules:
+  - name: sandbox
```
ablack12, line 676: Can we add another module so we hit both the "exists in the patch already" and the "doesn't exist in the patch" case?

ablack12 (approved):
LGTM with test improvement

PR #6036, opened 2022-10-28
EVG-17305 Fix  Patch status still  being "started" after all tasks are done running

malikchaya2:

### Description 
In situations where all tasks were unscheduled in a build variant, the patch would remain in a "started" status even when done. 

### Testing 
Tested on staging. 
[before](https://spruce-staging.corp.mongodb.com/version/635ac2db97b1d37aea36782b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[after fix](https://spruce-staging.corp.mongodb.com/version/635c12ca9dbe32774c967245/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


model/build/build.go
```diff
@@ -194,6 +194,18 @@ func (b *Build) SetAborted(aborted bool) error {
 	)
 }
 
+// SetAborted sets the build aborted field to the given boolean.
+func (b *Build) SetActivated(activated bool) error {
```
hadjri, line 198: Update comment

model/task_lifecycle.go
```diff
@@ -1017,6 +1022,14 @@ func updateBuildStatus(b *build.Build) (bool, error) {
 	}
 
 	buildStatus, allTasksBlocked := getBuildStatus(buildTasks)
+	if buildStatus == evergreen.BuildUnscheduled {
+		if err = b.SetActivated(false); err != nil {
+			return true, errors.Wrapf(err, "setting build '%s' as inactive", b.Id)
+		}
+
+		return true, nil
```
ablack12, line 1030: Should this instead be a return value like AllTasksBlocked (could return a small struct to make this neater)? Having it be a status seems misleading since we don't actually use Unscheduled as a build status.

ablack12 (changes requested):

model/task_lifecycle.go
```diff
@@ -949,7 +944,12 @@ func getBuildStatus(buildTasks []task.Task) (string, bool) {
 	// Check if no tasks have started and if all tasks are blocked.
 	noStartedTasks := true
 	allTasksBlocked := true
+	allTasksUnscheduled := true
 	for _, t := range buildTasks {
+		taskUnscheduled := t.Status == evergreen.TaskUndispatched && !t.Activated
+		if !taskUnscheduled {
```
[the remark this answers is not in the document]
malikchaya2: done

model/task_lifecycle.go
malikchaya2: Are you referring to getBuildStatus returning evergreen.BuildUnscheduled as a status on line 963?

hadjri (approved):

model/task_lifecycle.go
ablack12: Right; should have commented on 1024 but intsead of checking buildStatus it may make more sense to check something like allTasksUnscheduled, since unscheduled isn't truly a status

malikchaya2: done

globals.go
```diff
@@ -130,10 +130,11 @@ const (
 	TestSkippedStatus        = "skip"
 	TestSucceededStatus      = "pass"
 
-	BuildStarted   = "started"
-	BuildCreated   = "created"
-	BuildFailed    = "failed"
-	BuildSucceeded = "success"
+	BuildStarted     = "started"
+	BuildCreated     = "created"
+	BuildUnscheduled = "unscheduled"
```
ablack12, line 135: Should remove this now

model/build/build.go
```diff
@@ -194,6 +194,18 @@ func (b *Build) SetAborted(aborted bool) error {
 	)
 }
 
+// SetActivated sets the build activated field to the given boolean.
+func (b *Build) SetActivated(activated bool) error {
```
ablack12, line 198: Should we be calling SetActivated to true now for when tasks are scheduled, since the builds might not be activated?

model/task/task.go
```diff
@@ -3320,6 +3320,11 @@ func (t *Task) Blocked() bool {
 	return false
 }
 
+// isScheduled returns if a task is not scheduled to run
```
ablack12, line 3323: This function is called IsScheduled but seems to return if it's not scheduled (comment and code-wise); can we have this return if it is scheduled?

model/task_lifecycle.go
```diff
@@ -943,13 +938,23 @@ func updateMakespans(b *build.Build, buildTasks []task.Task) error {
 	return errors.WithStack(b.UpdateMakespans(depPath.TotalTime, CalculateActualMakespan(buildTasks)))
 }
 
+type BuildStatus struct {
```
ablack12, line 941: can this be a private struct since it's only used by a private function?

model/task_lifecycle.go
```diff
@@ -1016,14 +1028,23 @@ func updateBuildStatus(b *build.Build) (bool, error) {
 		return false, errors.Wrapf(err, "getting tasks in build '%s'", b.Id)
 	}
 
-	buildStatus, allTasksBlocked := getBuildStatus(buildTasks)
-	blockedChanged := allTasksBlocked != b.AllTasksBlocked
+	buildStatus := getBuildStatus(buildTasks)
+	// if all the tasks are unscheduled, set active to false
```
ablack12, line 1032: nit: capitalize comment

ablack12 (changes requested):

model/build/build.go
ablack12: I think this still needs to be addressed re the todo haha

model/task_lifecycle.go
```diff
@@ -943,13 +945,23 @@ func updateMakespans(b *build.Build, buildTasks []task.Task) error {
...
 // getBuildStatus returns a string denoting the status of the build and
 // a boolean denoting if all tasks in the build are blocked.
-func getBuildStatus(buildTasks []task.Task) (string, bool) {
+func getBuildStatus(buildTasks []task.Task) buildStatus {
 	// Check if no tasks have started and if all tasks are blocked.
 	noStartedTasks := true
 	allTasksBlocked := true
+	allTasksUnscheduled := true
 	for _, t := range buildTasks {
+		if !t.IsUnscheduled() {
```
ablack12, line 962: Preference for just having this function be IsScheduled -- more intuitive to check on the positive; and then the function comment makes a bit more. sense, and this line can just check if t.IsScheduled()

ablack12 (changes requested):

model/build/build.go
malikchaya2: It's done here: https://github.com/evergreen-ci/evergreen/pull/6036/commits/2df525b1032374d19d92137c6880c33f44aaafb7

And I tested it on staging and it's working as expected

model/task_lifecycle.go
malikchaya2: I think I prefer isUnscheduled because it mimics blocked. It also echos the UI "unscheduled" state. 
image

I see how the comment left something to be desired. I updated to what is hopefully more intuitive. Does it feel better with the comment update?

ablack12 (approved):

model/build/build.go
ablack12: Ah sounds good; we need to remove the todo then

ablack12 (approved):

model/build/build.go
```diff
@@ -194,6 +194,19 @@ func (b *Build) SetAborted(aborted bool) error {
 	)
 }
 
+// todo: remove. Should we be calling SetActivated to true now for when tasks are scheduled, since the builds might not be activated?
```
ablack12, line 197: LGTM with comment removal

PR #6038, opened 2022-10-28
EVG-16723: Send pending message to GH when a PR status updates

hadjri:
[EVG-16723](https://jira.mongodb.org/browse/EVG-16723)

### Description 
Added a function to send a GitHubStatePending message back to GitHub when a version's status updates and the requester is a pull request.

### Testing 
Deployed to staging, reset a task on the finished patch and confirmed that the PR status went back to "pending".


model/task_lifecycle.go
```diff
@@ -1198,8 +1201,74 @@ func UpdatePatchStatus(p *patch.Patch, versionStatus string) error {
...
+			if err = SendVersionStatusToGithub(p.Id.Hex(), p.GithubPatchData.BaseOwner, p.GithubPatchData.BaseRepo, p.GithubPatchData.HeadHash, "patch status change", "pr-task-reset"); err != nil {
+				return errors.Wrapf(err, "sending patch '%s' status to Github", p.Id.Hex())
+			}
+		}
+	}
+
+	return nil
+}
+
+func SendVersionStatusToGithub(id, owner, repo, ref, desc, job string) error {
```
ablack12, line 1214: This may belong better in this file https://github.com/evergreen-ci/evergreen/blob/580ebc277164a4d5442b14d6c14f20b4bb239fa6/thirdparty/github.go#L910

model/task_lifecycle.go
```diff
@@ -1198,8 +1201,74 @@ func UpdatePatchStatus(p *patch.Patch, versionStatus string) error {
...
+	catcher := grip.NewBasicCatcher()
```
ablack12, line 1215: there's no point in this catcher since we always just return if there's an error

model/task_lifecycle.go
ablack12: Also this takes in enough inputs that a struct would be better. We also need to comment this function since it's public

model/task_lifecycle.go
```diff
@@ -1198,8 +1201,74 @@ func UpdatePatchStatus(p *patch.Patch, versionStatus string) error {
 		if err = p.UpdateStatus(patchStatus); err != nil {
 			return errors.Wrapf(err, "updating patch '%s' with status '%s'", p.Id.Hex(), patchStatus)
 		}
+		if p.IsGithubPRPatch() {
+			if err = SendVersionStatusToGithub(p.Id.Hex(), p.GithubPatchData.BaseOwner, p.GithubPatchData.BaseRepo, p.GithubPatchData.HeadHash, "patch status change", "pr-task-reset"); err != nil {
```
ablack12, line 1205: I didn't look into this too much but I'm wondering if instead of this function we could repurpose makeGithubStatusUpdateJob 
https://github.com/evergreen-ci/evergreen/blob/e37538a7fd96246c2241148634adf2b4beabc438/units/github_status_api.go#L85

ablack12 (changes requested):

model/task_lifecycle.go
hadjri: Alas units cannot be imported into the model package, that is what my initial solution was.  It also seems like running amboy jobs from the model package is a pattern we avoid.  Although I'm not necessarily opposed to refactoring the packages a bit to make it possible.

model/task_lifecycle.go
```diff
@@ -1198,6 +1199,11 @@ func UpdatePatchStatus(p *patch.Patch, versionStatus string) error {
 		if err = p.UpdateStatus(patchStatus); err != nil {
 			return errors.Wrapf(err, "updating patch '%s' with status '%s'", p.Id.Hex(), patchStatus)
 		}
+		if p.IsGithubPRPatch() {
+			if err = thirdparty.SendVersionStatusToGithub(p.Id.Hex(), p.GithubPatchData.BaseOwner, p.GithubPatchData.BaseRepo, p.GithubPatchData.HeadHash, "patch status change", "pr-task-reset"); err != nil {
```
ablack12, line 1203: Tests failing probably bc this isn't using the new struct

thirdparty/github.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/mongodb/grip/sometimes"
```
ablack12, line 8: should move down

thirdparty/github.go
```diff
@@ -51,6 +52,15 @@ type GithubPatch struct {
 	CommitMessage  string `bson:"commit_message"`
 }
 
+type SendGithubStatusInput struct {
```
ablack12, line 55: nit: should comment since it's public

thirdparty/github.go
```diff
@@ -51,6 +52,15 @@ type GithubPatch struct {
...
+	Id    string
+	Owner string
+	Repo  string
+	Ref   string
+	Desc  string
+	Job   string
```
ablack12, line 61: Can this be caller instead, since it's not necessarily going to be an amboy job name

thirdparty/github.go
```diff
@@ -51,6 +52,15 @@ type GithubPatch struct {
 	CommitMessage  string `bson:"commit_message"`
 }
 
+type SendGithubStatusInput struct {
+	Id    string
```
ablack12, line 56: VersionId I think?

ablack12 (approved):
LGTM with nits

PR #6040, opened 2022-11-01
EVG-17330 script from stdin

ybrill:
[EVG-17330](https://jira.mongodb.org/browse/EVG-17330)

### Description 
If a user-provided provisioning script is greater than a certain length it won't run on spawn hosts of Windows userdata distros.
idk if there's a specific reason we avoided this until now, but this PR gets around the issue by passing the script to the shell through stdin. This is consistent with [how the script is executed through ssh for non-userdata distros](https://github.com/ybrill/evergreen/blob/d4eab8faf353d473bfec8478c3572ab85450d039/model/host/hostutil.go#L234).

### Testing 
With these changes deployed to staging [the example problem script from the ticket](https://github.com/visemet/mongo/blob/5af9ad0e1ec467b9848636ff426a8fe9d8c49d63/buildscripts/setup_spawnhost_coredump) ran on a Windows spawn host.

### DIFF
--- a/units/host_execute.go
+++ b/units/host_execute.go
@@ -89,10 +89,11 @@ func (j *hostExecuteJob) Run(ctx context.Context) {
 				args = append(args, fmt.Sprintf("--user=%s", j.SudoUser))
 			}
 		}
-		args = append(args, j.host.Distro.ShellBinary(), "-l", "-c", j.Script)
+		args = append(args, j.host.Distro.ShellBinary(), "-s")
 		var output []string
 		output, err := j.host.RunJasperProcess(ctx, j.env, &options.Create{
-			Args: args,
+			Args:               args,
+			StandardInputBytes: []byte(j.Script),
 		})
 		if err != nil {
 			event.LogHostScriptExecuteFailed(j.host.Id, err)


units/host_execute.go
```diff
@@ -89,10 +89,11 @@ func (j *hostExecuteJob) Run(ctx context.Context) {
 				args = append(args, fmt.Sprintf("--user=%s", j.SudoUser))
 			}
 		}
-		args = append(args, j.host.Distro.ShellBinary(), "-l", "-c", j.Script)
+		args = append(args, j.host.Distro.ShellBinary(), "-s")
```
Kimchelly, line 92: I think we still need `-l` here. `RunSSHShellScript` is written for provisioning and the provisioning SSH commands are written in such a way to not depend on the particular shell context (except where it's necessary to do so). I imagine some scripts may already assume that they're running in a login session (i.e. a shell environment similar to what SSH will give them). However, the way this is implemented, it is running the command via exec (i.e. Jasper), so it will not have any login session unless it's explicitly set.

Kimchelly: I imagine the `-s` is a way to overcome [this issue on Windows](https://github.com/dotnet/msbuild/issues/2530). Can we also add a comment that `-s` is necessary due to exec string length limitations?

ybrill: done and done

ybrill: I think that was Max's conclusion too.

Kimchelly (approved):

PR #6041, opened 2022-11-01
EVG-18110: remove fallback for task dispatch logic

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18110

### Description
Using a transaction to dispatch host tasks has not caused any noticeable performance or correctness issues so far, and it's been on for several weeks now. This removes the temporary feature flag to fall back to the old non-transaction dispatch so that it always uses the transaction.

### Testing 
Existing tests were already using the transaction and they still pass.


ablack12 (approved):
🎉

PR #6043, opened 2022-11-03
EVG-18186 Fix changes caused by gqlgen upgrade

khelif96:
[EVG-18186](https://jira.mongodb.org/browse/EVG-18186)

### Description 
The gqlgen update introduced some changes to how docs are generated and how PRs are created. This introduces those changes as well as addresses a resolver type mismatch bug for the `Subscription.target` field. 

So while working on [EVG-18186](https://jira.mongodb.org/browse/EVG-18186) I came across a rather important change in the behavior of our gqlgen library that was introduced after we upgraded it. The behavior of falsy nil-able values ("", false, 0)  was changed in [#1623 ](https://github.com/99designs/gqlgen/pull/1623). Now nil-able values will inherit a default value instead of being nil. Which is what they were supposed to do according to the [graphql spec](https://graphql.org/learn/schema/#arguments). The upgrade fixed the bug in the gqlgen library but introduced some test failures that relied on the bug being present in our code. I'm working on a fix but I thought it would be worthwhile to share.

This will affect how we do nil checks in the future for values that originate from gqlgen input fields. If we want to do a nil check in the future the input should not have a default value.

This also Fixes the panics that were being investigated in EVG-18123 and reenables the introspection query.


graphql/user_resolver.go
```diff
@@ -37,11 +38,22 @@ func (r *userResolver) Patches(ctx context.Context, obj *restModel.APIDBUser, pa
...
 
+// Permissions is the resolver for the permissions field.
 func (r *userResolver) Permissions(ctx context.Context, obj *restModel.APIDBUser) (*Permissions, error) {
 	return &Permissions{UserID: utility.FromStringPtr(obj.UserID)}, nil
 }
 
+// Target is the resolver for the target field.
+func (r *subscriberInputResolver) Target(ctx context.Context, obj *restModel.APISubscriber, data string) error {
+	obj.Target = data
+	return nil
```
khelif96, line 49: This is the only change i made manually

malikchaya2 (approved):

PR #6045, opened 2022-11-03
EVG-17358 Return display tasks in get build by id route

bynn:
[EVG-17358](https://jira.mongodb.org/browse/EVG-17358)

### Description 
The build route was erroring out because it only had 1 display task. 
The task.Find method automatically filters out display tasks.
We now use findall which will return display only tasks as well but still gets rid of the extra check

### Testing 
unit

### DIFF
--- a/rest/route/build_test.go
+++ b/rest/route/build_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/evergreen-ci/evergreen/db"
 	serviceModel "github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/build"
+	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/model/user"
 	"github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/evergreen-ci/gimlet"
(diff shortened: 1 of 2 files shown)


malikchaya2 (approved):

ablack12 (changes requested):
Marking myself as request changes for re-review of what we discussed in person (adding display task info to task cache, test)

rest/route/build_test.go
```diff
@@ -56,9 +76,21 @@ func (s *BuildByIdSuite) TestFindByIdProjFound() {
...
+	s.rm.(*buildGetHandler).buildId = "build2"
+	resp = s.rm.Run(context.TODO())
+	s.Equal(resp.Status(), http.StatusOK)
+	s.NotNil(resp.Data())
+
+	b, ok = (resp.Data()).(*model.APIBuild)
+	s.True(ok)
+	s.Equal(utility.ToStringPtr("build2"), b.Id)
+	s.Equal(utility.ToStringPtr("notbranch"), b.ProjectId)
+	s.Equal("task2", b.TaskCache[0].Id)
```
ablack12, line 90: Can we also validate that something else that would be added via SetFromTaskCache is added here? (since this would be true even if we didn't find the task from the find query)

ablack12 (approved):
LGTM with change but i don't need to review again

PR #6046, opened 2022-11-03
EVG-18139: gracefully handle nonexistent hosts when checking statuses

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18139

### Description
Looking through the logs, AWS did an unexpectedly new thing with `DescribeInstances` that we haven't seen before. In the past, when we requested at least 1 instance ID that AWS has no knowledge of, it would return an `InvalidInstanceID.NotFound` error (e.g. because the instance never existed or it was terminated a long time ago), which we handle already by terminating it. This time, it behaved in a different way - it had no status for some of the requested IDs (and those instances did not exist in AWS), but instead of producing the `InvalidInstanceID.NotFound` error, it simply returned no instance reservation for them (i.e. no status). I updated the logic to be more resilient to the possibility that either it will return `InvalidInstanceID.NotFound` or it simply will return no error and no status for that instance. The working assumption I'm making here is that if it doesn't have any status in AWS, it's safe to terminate (which appeared to be the case). That being said, I've never seen it return no error and give fewer statuses than there are hosts except in this one case.

* Change `GetInstanceStatuses` to set the host status as nonexistent the response is successful but the result is still missing that host.
* Refactor `GetInstanceStatuses` to return a map instead of a slice since it's a little weird to return statuses that depend on the order of the input slice.
* Handle nonexistent (i.e. those resulting in `InvalidInstanceID.NotFound` or those with no status) hosts the same as terminated ones for `set-cloud-hosts-ready`.
* Remove some old debug logs that are no longer used.

### Testing
Updated existing tests for EC2 providers and `set-cloud-hosts-ready` job.

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -113,8 +113,11 @@ type ContainerManager interface {
 
 // BatchManager is an interface for cloud providers that support batch operations.
 type BatchManager interface {
-	// GetInstanceStatuses gets the status of a slice of instances.
-	GetInstanceStatuses(context.Context, []host.Host) ([]CloudStatus, error)
+	// GetInstanceStatuses gets the statuses of a slice of instances. It returns
+	// a map of the instance IDs to their current status. If some of the
+	// instance statuses cannot be retrieved, implementations are allowed to
+	// either error or return StatusNonExistent.
+	GetInstanceStatuses(context.Context, []host.Host) (map[string]CloudStatus, error)
 }
 
 // ManagerOpts is a struct containing the fields needed to get a new cloud manager
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -116,7 +116,9 @@ type BatchManager interface {
 	// GetInstanceStatuses gets the statuses of a slice of instances. It returns
 	// a map of the instance IDs to their current status. If some of the
 	// instance statuses cannot be retrieved, implementations are allowed to
-	// either error or return StatusNonExistent.
+	// either return an error or return StatusNonExistent for those hosts.
+	// If there is no error, implementations should return the same number of
+	// results in the map as there are hosts.
 	GetInstanceStatuses(context.Context, []host.Host) (map[string]CloudStatus, error)
 }
 
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -937,7 +937,7 @@ func addPublicKey(ctx context.Context, h *host.Host, key string) error {
 }
 
 // GetInstanceStatuses returns the current status of a slice of EC2 instances.
-func (m *ec2Manager) GetInstanceStatuses(ctx context.Context, hosts []host.Host) ([]CloudStatus, error) {
+func (m *ec2Manager) GetInstanceStatuses(ctx context.Context, hosts []host.Host) (map[string]CloudStatus, error) {
 	if err := m.client.Create(m.credentials, m.region); err != nil {
 		return nil, errors.Wrap(err, "creating client")
 	}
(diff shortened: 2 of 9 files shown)


cloud/ec2_fleet.go
```diff
@@ -174,7 +174,8 @@ func (m *ec2FleetManager) ModifyHost(context.Context, *host.Host, host.HostModif
 	return errors.New("can't modify instances for EC2 fleet provider")
 }
 
-func (m *ec2FleetManager) GetInstanceStatuses(ctx context.Context, hosts []host.Host) ([]CloudStatus, error) {
+// kim; TODO: add test for instance not found.
```
ybrill, line 177: Is this still TODO?

units/host_status.go
```diff
@@ -176,6 +164,13 @@ func (j *cloudHostReadyJob) terminateUnknownHosts(ctx context.Context, awsErr st
 		if h == nil {
 			continue
 		}
+		// Decommission the host to prevent this job from checking it again.
+		grip.Error(message.WrapError(h.SetDecommissioned(evergreen.User, false, "cloud host has no status"), message.Fields{
```
ybrill, line 168: why do we want to log this here instead of adding to the catcher?

cloud/cloud.go
```diff
@@ -113,8 +113,13 @@ type ContainerManager interface {
...
 // BatchManager is an interface for cloud providers that support batch operations.
 type BatchManager interface {
-	// GetInstanceStatuses gets the status of a slice of instances.
-	GetInstanceStatuses(context.Context, []host.Host) ([]CloudStatus, error)
+	// GetInstanceStatuses gets the statuses of a slice of instances. It returns
+	// a map of the instance IDs to their current status. If some of the
+	// instance statuses cannot be retrieved, implementations are allowed to
+	// either return an error or return StatusNonExistent for those hosts.
+	// If there is no error, implementations should return the same number of
+	// results in the map as there are hosts.
```
ybrill, line 121: Why is it desirable for implementations to have this flexibility?

Kimchelly: This is just how the AWS response behaves. In the past, it has returned an error if any of the instance IDs does not exist, but in the case of this outage, it returned no error but failed to return some statuses.

In the end, past experience has proven that it's unreliable to trust that AWS will return us what we think will be their response, so might as well make it as resilient as possible to strange behavior.

cloud/ec2.go
```diff
@@ -970,20 +970,24 @@ func (m *ec2Manager) GetInstanceStatuses(ctx context.Context, hosts []host.Host)
...
 		for i := range spotHosts {
-			if spotInstanceRequestsMap[spotHosts[i].Id].InstanceId == nil || *spotInstanceRequestsMap[spotHosts[i].Id].InstanceId == "" {
-				hostToStatusMap[spotHosts[i].Id] = cloudStatusFromSpotStatus(*spotInstanceRequestsMap[spotHosts[i].Id].State)
+			spotHostID := spotHosts[i].Id
+			spotRequest, ok := spotInstanceRequestsMap[spotHostID]
+			if !ok {
+				hostToStatusMap[spotHostID] = StatusNonExistent
+				continue
+			}
+			instanceID := utility.FromStringPtr(spotRequest.InstanceId)
```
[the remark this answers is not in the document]
Kimchelly: Done.

cloud/ec2_fleet.go
Kimchelly: Nope, I added a test. Removed this comment.

units/host_status.go
```diff
@@ -117,45 +110,39 @@ clientsLoop:
...
+				hostID := hosts[i].Id
+				status, ok := statuses[hostID]
+				if !ok {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message": "could not get any cloud instance status for host, defaulting to nonexistent status",
+						"host_id": hostID,
+						"job":     j.ID(),
+					}))
+					statuses[hostID] = cloud.StatusNonExistent
+				}
```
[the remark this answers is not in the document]
Kimchelly: This is a case where it's not supposed to happen (i.e. there's a problem with the `GetInstanceStatuses` implementation), but for resiliency against outages caused by `DescribeInstances` failing, it should still handle in some way. I opted to decide that if there's no status, that means the instance doesn't exist so that it will not keep trying to pick up the same instance (which has no status) over and over again.

I changed this to an alert instead of error so that we know if `GetInstanceStatuses` is violating the expected interface behavior.

units/host_status.go
Kimchelly: I guess it doesn't really matter, since both end up just being logged as errors anyway. I changed it to catch the error.

cloud/cloud.go
ybrill: ah I get it now.
Could it make sense for `GetInstanceStatuses` to never error for non-existent hosts and instead always return the non-existent hosts with the `StatusNonExistent` status? It anyways seems to me that the AWS error string parsing we do to get the offending instance ids ([this thing](https://github.com/evergreen-ci/evergreen/blob/f486e5cef70c4425dedb6b974d35597d13f14341/units/host_status.go#L147-L151)) belongs somewhere in ec2.go rather than host_status.go. 
It would have an additional advantage that we could consolidate `terminateUnknownHosts` into `setCloudHostStatus`.

Kimchelly: We could, but it would make the logic more complicated we'll have to push the error-parsing logic into `GetInstanceStatuses` and re-issue the `DescribeInstances` request with the smaller set of instance IDs (and repeatedly do it if we keep getting nonexistent hosts). I think it would be nice to have, but it seems more like a small improvement of the abstraction layers that I'd rather leave for a later time. My primary goal with this PR is to ensure that `GetInstanceStatuses` is more resilient this new response.

Kimchelly: https://jira.mongodb.org/browse/EVG-18280

ybrill (approved):

PR #6048, opened 2022-11-04
EVG-18206: prevent potential deadlock in EC2 SNS test

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18206

### Description 
Alright, apparently, I didn't re-run my previous PR (#6044) enough times because the test timed out for the same reason when committed to the waterfall... I suspect that the `TimestampMonitor` cleaning up dropped collections from the previous PR is a symptom and not the cause - the DB might only doing that because it's detected that it's idle (due to the test hanging for another reason).

My second best guess (which I verified by running it even more times) is that the test is flaky because `queue.NewLocalLimitedSize` has an undiagnosed deadlock somewhere in it, resulting in the call to `queue.Stats(ctx)` to hang until Go times out the test. I don't really see _how_ it could deadlock, but I don't totally trust that the implementation is bug-free either, so I changed it to another (more verified) implementation and put a timed context on the test so it doesn't block the other tests for 10 straight minutes.

* Swap out queue implementations for test and reduce chance of test deadlock.
* Fix a couple unrelated tests which were failing because they are relying on global Amboy remote queue DB state (i.e. `testutil.NewEnvironment` uses a DB-backed queue) and were making my re-runs of `test-rest-route` fail.

### Testing 
Ran `test-rest-route` in patches a bunch more times.

### DIFF
--- a/mock/environment.go
+++ b/mock/environment.go
@@ -96,9 +96,17 @@ func (e *Environment) Configure(ctx context.Context) error {
 		ScopeCollection: evergreen.ScopeCollection,
 	})
 
+	catcher := grip.NewBasicCatcher()
+	catcher.Add(e.roleManager.RegisterPermissions(evergreen.ProjectPermissions))
+	catcher.Add(e.roleManager.RegisterPermissions(evergreen.DistroPermissions))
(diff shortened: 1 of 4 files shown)


malikchaya2 (approved):
🤞🏻

PR #6049, opened 2022-11-04
EVG-17375 Create manifest upon patch finalization

hadjri:

### Description 
Currently, the git hashes set on all manifest documents reflect the state of the module's target branch at the time of the manifest's version's commit revision, which in the case of many patch build can be many days or weeks before the patch was created, making it not possible to run a patch using a different module hash than what exists on the base mainline commit version, which makes for difficulty in testing.  

Change to address this is to create a manifest upon finalization of a patch (i.e. when a patch officially becomes a version according to Evergreen).  This required a refactor of `CreateManifest` to check if the version's requester is that of a patch and if so, look up the most recent module commits rather than looking up the module commits at the time of the version's underlying revision creation.

Doing this caused cyclical import errors so `CreateManifest` was moved to the model package to remedy this.

### Testing 
Added unit testing for manifest creation of patch versions. Deployed to staging and confirmed that patch versions create a manifest with git hashes on modules equivalent to what existed in the module repos at the time of patch creation rather than what was the state of the module repo at the time of the version's underlying git revision.


malikchaya2 (approved):

model/version.go
```diff
@@ -601,6 +605,89 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 	return res, nil
 }
 
+func CreateManifest(v Version, proj *Project, projectRef *ProjectRef, settings *evergreen.Settings) (*manifest.Manifest, error) {
```
ablack12, line 608: Can this take in a version pointer instead?

ablack12: Also need a function comment for this since it's public

model/version.go
```diff
@@ -601,6 +605,89 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+	newManifest := &manifest.Manifest{
+		Id:          v.Id,
+		Revision:    v.Revision,
+		ProjectName: v.Identifier,
+		Branch:      projectRef.Branch,
+		IsBase:      v.Requester == evergreen.RepotrackerVersionRequester,
+	}
+	token, err := settings.GetGithubOauthToken()
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting github oauth token")
```
ablack12, line 621: remove error

model/version.go
```diff
@@ -601,6 +605,89 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+			// If this is a mainline commit, retrieve the module's commit from the time of the mainline commit.
+			// Otherwise, retrieve the module's commit from the time of the patch creation.
+			revisionTime := time.Unix(0, 0)
+			if !evergreen.IsPatchRequester(v.Requester) {
+				revisionTime = commit.Commit.Committer.GetDate()
+			}
+			var branchCommits []*github.RepositoryCommit
+			branchCommits, _, err = thirdparty.GetGithubCommits(ctx, token, owner, repo, module.Branch, revisionTime, 0)
+			if err != nil {
+				return nil, errors.Wrapf(err, "problem retrieving getting git branch for module %s", module.Name)
```
ablack12, line 650: remove problem (and below)

model/patch_lifecycle.go
```diff
@@ -575,6 +576,10 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		if err != nil {
 			return nil, errors.Wrapf(err, "inserting version '%s'", patchVersion.Id)
 		}
+		_, err = CreateManifest(*patchVersion, project, projectRef, settings)
```
ablack12, line 579: this is being done within a transaction; maybe we should pull out the "insert" part from the "create" part so that it can be created outside the transaction and inserted using the sessCtx?

thirdparty/github.go
```diff
@@ -383,6 +383,35 @@ func GetCommitDiff(ctx context.Context, oauthToken, repoOwner, repo, sha string)
 	return commit, nil
 }
 
+// GetBranchCommitHash retrieves the most recent commit hash for branch for the given repo, module, and branch name.
+func GetBranchCommitHash(ctx context.Context, repo, moduleBranch, token string) (string, error) {
```
ablack12, line 387: I'm slightly confused why we needed to move this; if this is only used in the units package should it just be a private function there?

ablack12 (changes requested):

ablack12:
I slightly didn't follow the description; what is the workflow when we don't have this manifest for patches? Is there a way to just use the manifest for the version created closest to the patch, or whatever its base revision is (i.e. enforce that users need to merge main to have an updated module, which is maybe reasonable) as opposed to inserting a new manifest? (Wondering if this will essentially be duplicate information in the manifest collection that we can push back on)

hadjri:
> Is there a way to just use the manifest for the version created closest to the patch, or whatever its base revision is (i.e. enforce that users need to merge main to have an updated module, which is maybe reasonable) as opposed to inserting a new manifest? (Wondering if this will essentially be duplicate information in the manifest collection that we can push back on)

That is the [current flow](https://github.com/evergreen-ci/evergreen/blob/main/model/manifest/db.go#L72) that is followed when a patch version requests a manifest.  It'll get the manifest for the version created closest to the patch, and the manifest modules will have git hashes that represent the state of those modules at the time of the base version's commit.

Enforcing users to merge main in order to have updated modules would help, but there is still the case where module repos have commits that came in after the most recent commit for the project in question.  If this were to happen then patch versions off of main would still have outdated modules since they'd just be a snapshot of what the world looked like at the most recent main commit for the project.

thirdparty/github.go
hadjri: Moved back to units

model/patch_lifecycle.go
hadjri: Sure!

model/patch_lifecycle.go
```diff
@@ -585,6 +591,12 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 				return nil, errors.Wrapf(err, "inserting project config for version '%s'", patchVersion.Id)
 			}
 		}
+		if manifest != nil {
+			_, err = manifest.TryInsert()
```
ablack12, line 595: This does still need to use the sessCtx

model/version.go
```diff
@@ -601,6 +605,98 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 	return res, nil
 }
 
+// ConstructManifest will construct a manifest from the given project and version.
+func ConstructManifest(v *Version, proj *Project, projectRef *ProjectRef, settings *evergreen.Settings) (*manifest.Manifest, error) {
```
ablack12, line 609: It looks to me like this is only used in model packages now in which case it can be private

model/patch_lifecycle.go
```diff
@@ -585,6 +591,12 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "creating manifest for version '%s'", patchVersion.Id)
```
ablack12, line 597: inserting manifest

model/version.go
```diff
@@ -601,6 +605,98 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+			Branch:   module.Branch,
+			Revision: sha,
+			Repo:     repo,
+			Owner:    owner,
+			URL:      url,
+		}
+
+	}
+	newManifest.Modules = modules
+	grip.Debug(message.Fields{
```
ablack12, line 678: I'm inclined to remove this (we have a lot of debugging statements and it's unclear what they're useful for; if we've created the manifest we theoretically should be able to find it in the db to debug with)

ablack12 (changes requested):

model/version.go
```diff
@@ -601,6 +605,88 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+			// If this is a mainline commit, retrieve the module's commit from the time of the mainline commit.
+			// Otherwise, retrieve the module's commit from the time of the patch creation.
+			revisionTime := time.Unix(0, 0)
+			if !evergreen.IsPatchRequester(v.Requester) {
+				revisionTime = commit.Commit.Committer.GetDate()
+			}
+			var branchCommits []*github.RepositoryCommit
+			branchCommits, _, err = thirdparty.GetGithubCommits(ctx, token, owner, repo, module.Branch, revisionTime, 0)
+			if err != nil {
+				return nil, errors.Wrapf(err, "retrieving getting git branch for module %s", module.Name)
```
ablack12, line 650: fix error message

model/version.go
```diff
@@ -601,6 +605,88 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+				sha = branchCommits[0].GetSHA()
+				url = branchCommits[0].GetURL()
+			}
+		} else {
+			sha = module.Ref
+			gitCommit, err = thirdparty.GetCommitEvent(ctx, token, owner, repo, module.Ref)
+			if err != nil {
+				return nil, errors.Wrapf(err, "retrieving getting git commit for module %s with hash %s", module.Name, module.Ref)
+			}
+			if gitCommit != nil {
```
ablack12, line 662: Sorry I know you didn't write this but we can skip this if statement and just do `url = gitCommit.GetURL()` directly

ablack12 (approved):
LGTM with nits / test fixes

PR #6051, opened 2022-11-08
EVG-18217: Allow migration of volumes with previously terminated hosts

sophstad:
[EVG-18217](https://jira.mongodb.org/browse/EVG-18217)

### Description 
- Fixes a bug where home volumes that were attached to a workstation that was terminated would fail to migrate because the initial host could not be found.
  - At present, these volumes can still be attached to a new workstation by adding them during the spawn host stage. However, migration should be supported too as it is in some ways a more obvious workflow.

### Testing 
- Tested on staging
- Add unit test for this case

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2770,9 +2770,9 @@ func FindHostWithVolume(volumeID string) (*Host, error) {
 	return FindOne(q)
 }
 
-// FindHostWithVolume finds the host associated with the
+// FindUpHostWithHomeVolume finds the up host associated with the
 // specified home volume ID.
-func FindHostWithHomeVolume(homeVolumeID string) (*Host, error) {
+func FindUpHostWithHomeVolume(homeVolumeID string) (*Host, error) {
 	q := db.Query(
 		bson.M{
 			StatusKey:       bson.M{"$in": evergreen.UpHostStatus},
@@ -2780,6 +2780,19 @@ func FindHostWithHomeVolume(homeVolumeID string) (*Host, error) {
 			HomeVolumeIDKey: homeVolumeID,
 		},
 	)
+	return FindOne(q)
+}
+
+// FindLatestTerminatedHostWithHomeVolume finds the most recently terminated host
+// associated with the specified home volume ID.
+func FindLatestTerminatedHostWithHomeVolume(homeVolumeID string) (*Host, error) {
+	q := db.Query(
+		bson.M{
+			StatusKey:       evergreen.HostTerminated,
+			UserHostKey:     true,
+			HomeVolumeIDKey: homeVolumeID,
+		},
+	).Sort([]string{TerminationTimeKey})
 	return FindOne(q)
 }
 
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2783,12 +2783,13 @@ func FindUpHostWithHomeVolume(homeVolumeID string) (*Host, error) {
 	return FindOne(q)
 }
 
-// FindLatestTerminatedHostWithHomeVolume finds the most recently terminated host
+// FindLatestTerminatedHostWithHomeVolume finds the user's most recently terminated host
 // associated with the specified home volume ID.
-func FindLatestTerminatedHostWithHomeVolume(homeVolumeID string) (*Host, error) {
+func FindLatestTerminatedHostWithHomeVolume(homeVolumeID string, startedBy string) (*Host, error) {
 	q := db.Query(
 		bson.M{
 			StatusKey:       evergreen.HostTerminated,
+			StartedByKey:    startedBy,
 			UserHostKey:     true,
 			HomeVolumeIDKey: homeVolumeID,
 		},
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -2792,7 +2792,7 @@ func FindLatestTerminatedHostWithHomeVolume(homeVolumeID string) (*Host, error) 
 			UserHostKey:     true,
 			HomeVolumeIDKey: homeVolumeID,
 		},
-	).Sort([]string{TerminationTimeKey})
+	).Sort([]string{"-" + TerminationTimeKey})
 	return FindOne(q)
 }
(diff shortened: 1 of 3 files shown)


model/host/host.go
```diff
@@ -2783,6 +2783,19 @@ func FindHostWithHomeVolume(homeVolumeID string) (*Host, error) {
 	return FindOne(q)
 }
 
+// FindLatestTerminatedHostWithHomeVolume finds the most recently terminated host
+// associated with the specified home volume ID.
+func FindLatestTerminatedHostWithHomeVolume(homeVolumeID string) (*Host, error) {
+	q := db.Query(
+		bson.M{
+			StatusKey:       evergreen.HostTerminated,
```
Kimchelly, line 2791: We have to also include the `StartedBy` key here to ensure we use the `{"started_by": 1, "status": 1}` index since just the status is not selective enough.

units/migrate_volume.go
```diff
@@ -121,11 +121,12 @@ func (j *volumeMigrationJob) Run(ctx context.Context) {
 			j.AddError(errors.Errorf("volume '%s' not found", j.VolumeID))
 			return
 		}
+		fmt.Println("UPDATING VOLUME")
```
Kimchelly, line 124: Delete print statement.

units/migrate_volume.go
```diff
@@ -137,10 +138,14 @@ func (j *volumeMigrationJob) Run(ctx context.Context) {
 		}
 	}
 
-	// Set initial host to have expiration in 24 hours
-	if err := j.initialHost.SetExpirationTime(time.Now().Add(evergreen.DefaultSpawnHostExpiration)); err != nil {
-		j.AddError(errors.Wrapf(err, "setting expiration for host '%s'", j.InitialHostID))
-		return
+	// If not terminated, set initial host to have expiration in 24 hours
```
Kimchelly, line 141: If not stopped?

units/migrate_volume.go
```diff
@@ -99,7 +99,7 @@ func (j *volumeMigrationJob) Run(ctx context.Context) {
 		return
 	}
 
-	if j.initialHost.HomeVolumeID != "" {
+	if j.volume.Host != "" {
```
Kimchelly, line 102: This way seems less idempotent than before if we're assuming the job may re-run because L103 will unset the volume's host, but then could fail to unset the host's home volume on L109, resulting in a broken one-way link between the initial host and the volume. Is it a problem if this happens?

units/migrate_volume.go
```diff
@@ -164,7 +169,7 @@ func (j *volumeMigrationJob) stopInitialHost(ctx context.Context) {
 // startNewHost attempts to start a new host with the volume attached.
 func (j *volumeMigrationJob) startNewHost(ctx context.Context) {
 	// Ensure volume has been detached
-	if j.volume.Host == j.InitialHostID {
+	if j.volume.Host != "" {
```
Kimchelly, line 172: Same comment as above.

units/migrate_volume.go
```diff
@@ -231,7 +236,20 @@ func (j *volumeMigrationJob) populateIfUnset() error {
 	}
 
 	if j.InitialHostID == "" {
-		j.InitialHostID = j.volume.Host
+		// If volume was attached to terminated host, query this host by its home volume field.
```
Kimchelly, line 239: Phrasing question - is this implying that we attached the volume to a host that got terminated, or that the initial host ended up terminated (e.g. because it expired) while it still had its home volume set?

units/migrate_volume.go
```diff
@@ -137,10 +138,14 @@ func (j *volumeMigrationJob) Run(ctx context.Context) {
 		}
 	}
 
-	// Set initial host to have expiration in 24 hours
-	if err := j.initialHost.SetExpirationTime(time.Now().Add(evergreen.DefaultSpawnHostExpiration)); err != nil {
-		j.AddError(errors.Wrapf(err, "setting expiration for host '%s'", j.InitialHostID))
-		return
+	// If not terminated, set initial host to have expiration in 24 hours
+	if j.initialHost.Status == evergreen.HostStopped {
```
Kimchelly, line 142: Why is this specific to just stopped hosts?

Kimchelly (changes requested):

units/migrate_volume.go
sophstad: If the initial host is already terminated, setting the expiration is useless afaik (we've already ensured the initial host is either stopped or terminated on L84).

units/migrate_volume.go
sophstad: The latter (this is currently the approach for users to transfer a home volume to a new host — terminate the workstation and spin up a new one with the volume attached). Will make this more clear!

units/migrate_volume.go
sophstad: It's true, but this pattern is with precent in the evergreen codebase; if a workstation is terminated, it still points to the volume, but the volume is no longer associated with the host. It's still not ideal though.

One other approach I considered was setting a flag if we find initially that the volume is unattached to a host (i.e. the host was terminated). Then we could more easily skip the "start new host" and "detach old host" stages since they're already done. Would be interested in any thoughts you have on that! 🙏

Kimchelly: I think this change looks good. It should be okay to re-run the start new host/detach old host checks since they're idempotent anyways, so even if there is a bit of unnecessary repeated work, it's not a big deal.

Kimchelly (commented):
GitHub doesn't let me comment on unchanged lines, but is [this line supposed to try stopping a host that's already terminated](https://github.com/evergreen-ci/evergreen/blob/02aac9e65fe5a5e07646fac68eed6d767b920b7b/units/migrate_volume.go#L84)? The spawn host stop job shouldn't be able to succeed if the host is already terminated.

Kimchelly (approved):

PR #6052, opened 2022-11-08
EVG-15530 Remove json.send 

malikchaya2:
[EVG-15530](https://jira.mongodb.org/browse/EVG-15530)

### Description 
Json.send needs to be removed because we want users to relying on the perf.send command to generate new performance data.


rest/route/agent.go
```diff
@@ -1328,49 +1328,3 @@ func (h *setDownstreamParamsHandler) Run(ctx context.Context) gimlet.Responder {
...
-		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "finding task '%s'", h.taskID))
-	}
-	if t == nil {
-		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
-			StatusCode: http.StatusNotFound,
-			Message:    fmt.Sprintf("task '%s' not found", h.taskID),
-		})
-	}
-
-	if err = model.InsertTaskJSON(t, h.name, h.rawData); err != nil {
```
ablack12, line 1371: I think we can get rid of this function and also maybe this collection (in which case we can drop the collection after it's deployed)

ablack12 (changes requested):

rest/route/agent.go
malikchaya2: done

ablack12 (approved):
LGTM! Last time we dropped a collection there was no downtime right @ybrill ? I wonder if that means we can drop this collection the day after this deploys without scheduling maintenance time

ybrill:
Yes, it seemed like we didn't miss a beat.

malikchaya2:
The smoke tests [pass](https://spruce-staging.corp.mongodb.com/version/smoked_edf1a8dbbba31ad579ee39de08383d4cab772db3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) as a mainline commit.

PR #6053, opened 2022-11-08
EVG-18231 Include author_email in patch created versions

khelif96:
[EVG-18231](https://jira.mongodb.org/browse/EVG-18231)

### Description 
author_email was not previously included in versions that were created from a patch. This now includes the email from their git account by default if not it falls back to the email included in gitinfo. 

### Testing 
Example doc of a patch I created on staging
image

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -453,8 +453,8 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 	}
 
 	authorEmail := p.GitInfo.Email
-	if p.GithubPatchData.AuthorUID != 0 {
-		u, _ := user.FindByGithubUID(p.GithubPatchData.AuthorUID)
+	if p.Author != "" {
+		u, _ := user.FindOneById(p.Author)
 		if u != nil {
 			authorEmail = u.Email()
 		}
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -454,10 +454,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 
 	authorEmail := p.GitInfo.Email
 	if p.GithubPatchData.AuthorUID != 0 {
-		u, err := user.FindByGithubUID(p.GithubPatchData.AuthorUID)
-		grip.Error(message.WrapError(err, message.Fields{
-			"message": fmt.Sprintf("failed to fetch evergeen user with Github UID %d", p.GithubPatchData.AuthorUID),
-		}))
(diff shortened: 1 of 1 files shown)


malikchaya2 (changes requested):
The code looks good, but I think we should test this on staging before merging

model/patch_lifecycle.go
```diff
@@ -452,6 +452,17 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		return nil, errors.Wrap(err, "fetching patch parameters")
 	}
 
+	authorEmail := p.GitInfo.Email
+	if p.GithubPatchData.AuthorUID != 0 {
+		u, err := user.FindByGithubUID(p.GithubPatchData.AuthorUID)
+		grip.Error(message.WrapError(err, message.Fields{
```
ablack12, line 458: I'd rather not log this if we aren't going to do anything with it; it's not really an error i think that a user might not have included this (although I see we use this same format elsewhere; we may just want to remove them both, or at least change them to info)

khelif96: Done just removed the error here.

khelif96:
> The code looks good, but I think we should test this on staging before merging

Done just tested on staging and attached the document that resulted.

model/patch_lifecycle.go
```diff
@@ -452,6 +452,14 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		return nil, errors.Wrap(err, "fetching patch parameters")
 	}
 
+	authorEmail := p.GitInfo.Email
```
ybrill, line 455: The patch's Author field and the corresponding version's Author/AuthorID correspond to the id of the user who created the patch. 
Could it work to fetch the user with that instead?

ybrill (commented):
feel free to ignore me 😅 
I was already thinking about this so I figured I'd look at the PR

malikchaya2 (approved):

model/patch_lifecycle.go
ablack12: That's a good idea since it may work for more kinds of patches

khelif96: Good point sure!

model/patch_lifecycle.go
```diff
@@ -452,6 +452,14 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+	if p.Author != "" {
+		u, _ := user.FindOneById(p.Author)
```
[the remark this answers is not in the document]
khelif96: Sure added!

ybrill (approved):
lgtm too! 😄

PR #6056, opened 2022-11-09
EVG-18203 remove erroneous command error logging

ablack12:
[EVG-18203 ](https://jira.mongodb.org/browse/EVG-18203 )

### Description 
Only log if there's actually an error.

### DIFF
--- a/agent/command/exec.go
+++ b/agent/command/exec.go
@@ -348,14 +348,7 @@ func (c *subprocessExec) runCommand(ctx context.Context, taskID string, cmd *jas
 		}
 	}
 
-	if c.ContinueOnError {
-		logger.Execution().Notice(message.WrapError(err, message.Fields{
-			"task":       taskID,
-			"binary":     c.Binary,
-			"background": c.Background,
-			"silent":     c.Silent,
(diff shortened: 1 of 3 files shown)


agent/command/exec.go
```diff
@@ -348,14 +348,7 @@ func (c *subprocessExec) runCommand(ctx context.Context, taskID string, cmd *jas
...
-	if c.ContinueOnError {
-		logger.Execution().Notice(message.WrapError(err, message.Fields{
-			"task":       taskID,
-			"binary":     c.Binary,
-			"background": c.Background,
-			"silent":     c.Silent,
-			"continue":   c.ContinueOnError,
-		}))
+	if c.ContinueOnError && err != nil {
 		logger.Execution().Noticef("Script errored, but continue on error is set - continuing task execution. Error: %s.", err)
```
malikchaya2, line 352: Do we no longer want the message.Fields here?

ablack12: I'm thinking it's kind of redundant, and we don't have it at the other place we do this. I'd prefer to remove it.

malikchaya2 (approved):

PR #6057, opened 2022-11-09
EVG-17497 Reset priority when activating task

hadjri:
[EVG-17497](https://jira.mongodb.org/browse/EVG-17497)

### Description 
Currently in both legacy and spruce UI, you can disable a task, and then subsequently manually hit "Schedule".  This sets the status of the task to "Will Run" as expected but the priority remains at -1 and the task will actually never get scheduled (screenshot attached).  Change has been made such that when we activate a task, if it is set as disabled, reset its priority to zero.  

### Testing 
Updated unit test and deployed to staging to confirm that manually scheduling a disabled task in the UI resets its priority and lets the task actually get scheduled.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1679,6 +1679,7 @@ func ActivateTasks(tasks []Task, activationTime time.Time, updateDependencies bo
 				ActivatedKey:     true,
 				ActivatedByKey:   caller,
 				ActivatedTimeKey: activationTime,
+				PriorityKey:      evergreen.DefaultPriority,
 			},
 		})
 	if err != nil {
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1673,10 +1673,6 @@ func ActivateTasks(tasks []Task, activationTime time.Time, updateDependencies bo
 	if err != nil {
 		return errors.Wrap(err, "activating tasks")
 	}
-	err = enableDisabledTasks(taskIDs)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


hadjri:
image

model/task/task.go
```diff
@@ -1700,6 +1693,41 @@ func ActivateTasks(tasks []Task, activationTime time.Time, updateDependencies bo
...
 
+func enableDisabledTasks(taskIDs []string) error {
+	_, err := UpdateAll(
+		bson.M{
+			IdKey:       bson.M{"$in": taskIDs},
+			PriorityKey: evergreen.DisabledTaskPriority,
+		},
+		bson.M{
+			"$set": bson.M{
+				PriorityKey: evergreen.DefaultPriority,
```
ablack12, line 1704: Instead of this we can just use "$unset": which will put it back to 0

model/task/task.go
```diff
@@ -1700,6 +1693,41 @@ func ActivateTasks(tasks []Task, activationTime time.Time, updateDependencies bo
 	return nil
 }
 
+func enableDisabledTasks(taskIDs []string) error {
```
ablack12, line 1696: Can we call this in activateTasks itself (and also move to db.go?)

ablack12 (changes requested):

ablack12 (approved):
LGTM with nits

PR #6058, opened 2022-11-09
EVG-17275 check parser project collection if patched parser project is empty

ablack12:

### Description 
Stop setting patched parser project after a patch is finalized, since we store the thing in the parser project collection anyway.

### Testing 
Tested by [creating a patch](https://evergreen-staging.corp.mongodb.com/rest/v2/patches/636c137997b1d36b5c582ee3), verifying it could be scheduled / reconfigured / run. 

This is a bit risky since random operations do seem to depend on patched parser project. I audited every instance I could, First deploying the change that checks for version, _then_ the one that unsets the parser project, then we can deploy the unset parser project line, so that if we need to revert the code will know how to handle patches with an empty string.

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -449,7 +449,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		parentPatchNumber = parentPatch.PatchNumber
 	}
 
-	params, err := getFullPatchParams(p)
+	params, err := getFullPatchParams(p, config)
 	if err != nil {
 		return nil, errors.Wrap(err, "fetching patch parameters")
 	}
@@ -649,7 +649,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 
 // getFullPatchParams will retrieve a merged list of parameters defined on the patch alias (if any)
 // with the parameters that were explicitly user-specified, with the latter taking precedence.
-func getFullPatchParams(p *patch.Patch) ([]patch.Parameter, error) {
+func getFullPatchParams(p *patch.Patch, config *ProjectConfig) ([]patch.Parameter, error) {
 	paramsMap := map[string]string{}
 	if p.Alias == "" || !IsPatchAlias(p.Alias) {
 		return p.Parameters, nil
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -27,6 +27,7 @@ import (
 	"github.com/mongodb/grip/level"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"gopkg.in/yaml.v2"
 )
@@ -116,8 +117,9 @@ type PatchUpdate struct {
 // Returns an http status code and error.
 func ConfigurePatch(ctx context.Context, p *patch.Patch, version *Version, proj *ProjectRef, patchUpdateReq PatchUpdate) (int, error) {
 	var err error
-	project := &Project{}
-	if _, err := LoadProjectInto(ctx, []byte(p.PatchedParserProject), nil, p.Project, project); err != nil {
+	project, _, err := FindAndTranslateProjectForPatch(ctx, p)
+	if err != nil {
+		grip.Debug("issue here")
 		return http.StatusInternalServerError, errors.Wrap(err, "unmarshalling project config")
 	}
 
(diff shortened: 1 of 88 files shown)


hadjri:
Now that we only need `PatchedParserProject` to make it to the patch finalization stage, can we update `FinalizePatch` to unset the field at the same time we add to the parser projects collection?

service/rest_patch.go
```diff
@@ -62,6 +64,20 @@ func (restapi restAPI) getPatchConfig(w http.ResponseWriter, r *http.Request) {
 	}
 	w.Header().Set("Content-Type", "application/x-yaml; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
-	_, err := w.Write([]byte(projCtx.Patch.PatchedParserProject))
+	var err error
+	if projCtx.Patch.PatchedParserProject != "" {
+		_, err = w.Write([]byte(projCtx.Patch.PatchedParserProject))
+	} else {
+		var pp *model.ParserProject
```
hadjri, line 71: This line can be removed

service/rest_patch.go
```diff
@@ -62,6 +64,20 @@ func (restapi restAPI) getPatchConfig(w http.ResponseWriter, r *http.Request) {
...
+		pp, err = model.ParserProjectFindOneById(projCtx.Patch.Version)
+		if pp != nil {
+			var projBytes []byte
```
hadjri, line 74: Same comment

hadjri:
I also thing `PatchedProjectConfig` also follows this pattern and can be removed after patch finalization. Can we create a ticket for that as well?

hadjri (changes requested):

model/patch_lifecycle.go
```diff
@@ -116,8 +117,9 @@ type PatchUpdate struct {
 // Returns an http status code and error.
 func ConfigurePatch(ctx context.Context, p *patch.Patch, version *Version, proj *ProjectRef, patchUpdateReq PatchUpdate) (int, error) {
 	var err error
-	project := &Project{}
-	if _, err := LoadProjectInto(ctx, []byte(p.PatchedParserProject), nil, p.Project, project); err != nil {
+	project, _, err := FindAndTranslateProjectForPatch(ctx, p)
+	if err != nil {
+		grip.Debug("issue here")
```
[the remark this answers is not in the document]
ablack12: 🤦‍♀️

service/rest_patch.go
ablack12: We need it to then return the error correctly at 82 (although i agree it's not clear so i'll restructure)

ablack12:
Ah I think this bit from the description answers your first question:
> This is a bit risky since random operations do seem to depend on patched parser project. I audited every instance I could, First deploying the change that checks for version, then the one that unsets the parser project, then we can deploy the unset parser project line, so that if we need to revert the code will know how to handle patches with an empty string.

I forgot about patched config also having a collection. I'll add that into this.

hadjri (approved):
LGTM!

ablack12:
Added a good amount of stuff if you can re-review @hadjri  (ended up being some refactors just from reading patched config code; it does look like generally this will already work though since find looks it up from the version): https://github.com/evergreen-ci/evergreen/blob/776ed1d14d3ca0835f20fe7d5ffaee38047f4300/model/project_ref.go#L533

model/patch_lifecycle.go
```diff
@@ -649,7 +649,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 
 // getFullPatchParams will retrieve a merged list of parameters defined on the patch alias (if any)
 // with the parameters that were explicitly user-specified, with the latter taking precedence.
-func getFullPatchParams(p *patch.Patch) ([]patch.Parameter, error) {
+func getFullPatchParams(p *patch.Patch, config *ProjectConfig) ([]patch.Parameter, error) {
```
hadjri, line 663: config seems to be unused?

units/patch_intent.go
```diff
@@ -931,10 +931,8 @@ func (j *patchIntentProcessor) verifyValidAlias(projectId string, patchDoc *patc
 	if err != nil {
 		return errors.Wrapf(err, "retrieving aliases for project '%s'", projectId)
 	}
-	for _, a := range aliases {
-		if a.Alias == alias {
-			return nil
-		}
+	if len(aliases) > 0 {
+
```
hadjri, line 935: This block is empty

units/patch_intent.go
```diff
@@ -914,15 +914,15 @@ func (j *patchIntentProcessor) buildTriggerPatchDoc(patchDoc *patch.Patch) error
...
-func (j *patchIntentProcessor) verifyValidAlias(projectId string, patchDoc *patch.Patch) error {
+func (j *patchIntentProcessor) verifyValidAlias(projectId string, configStr string) error {
 	alias := j.intent.GetAlias()
 	if alias == "" {
 		return nil
 	}
 	var projectConfig *model.ProjectConfig
-	if patchDoc.PatchedProjectConfig != "" {
+	if configStr != "" {
 		var err error
```
hadjri, line 924: This line can be removed

model/patch_lifecycle.go
ablack12: ah thank you; i was considering a refactor but decided against it

units/patch_intent.go
ablack12: It can't be because projectConfig needs to be used outside of the if statement

hadjri (approved):

PR #6060, opened 2022-11-10
EVG-15148 rename distro alias queue to secondary queue

ablack12:
[EVG-15148](https://jira.mongodb.org/browse/EVG-15148 )

### Description 
Longer rational included in ticket; TLDR s that "alias" is already used to give distros multiple names so shouldn't be used for "secondary queue" as well.

Didn't change bson tags to avoid migrations.

### DIFF
--- a/model/task_queue.go
+++ b/model/task_queue.go
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	TaskQueuesCollection      = "task_queues"
-	TaskAliasQueuesCollection = "task_alias_queues"
+	TaskQueuesCollection          = "task_queues"
+	TaskSecondaryQueuesCollection = "task_alias_queues"
 )
 
 type TaskGroupInfo struct {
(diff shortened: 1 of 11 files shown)


model/task_queue.go
```diff
@@ -399,7 +401,7 @@ func ClearTaskQueue(distroId string) error {
 	aliasQuery := bson.M{
 		taskQueueDistroKey: distroId,
 	}
-	aliasCount, err := db.Count(TaskAliasQueuesCollection, aliasQuery)
+	aliasCount, err := db.Count(TaskSecondaryQueuesCollection, aliasQuery)
```
Kimchelly, line 404: Change the "alias" variables in `ClearTaskQueue` into "secondary" (+ fix the catcher errors).

Kimchelly (commented):
Bless this change 🙏

Kimchelly (approved):

PR #6063, opened 2022-11-10
EVG-17968 suppress messages from CLI when relevant

ablack12:
[EVG-17968](https://jira.mongodb.org/browse/EVG-17968)

### Description 
Removed getSetupCommunicator since it's not immediately obvious when to use get vs setup (since a lot of CLI setup ends up being copy paste, having a boolean makes it more likely that engineers will consider its value). 

### Testing 
Ensured client still works, and used list-patches to audit the JSON flag (with it, we don't print banner).

### DIFF
--- a/operations/admin.go
+++ b/operations/admin.go
@@ -17,7 +17,7 @@ import (
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
-	yaml "gopkg.in/20210107192922/yaml.v3"
+	"gopkg.in/20210107192922/yaml.v3"
 )
 
 func Admin() cli.Command {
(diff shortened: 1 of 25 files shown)


operations/admin.go
```diff
@@ -103,7 +103,7 @@ func adminSetBanner() cli.Command {
 				return errors.Wrap(err, "loading configuration")
 			}
 
-			client, err := conf.setupRestCommunicator(ctx)
+			client, err := conf.setupRestCommunicator(ctx, false)
```
Kimchelly, line 106: Should this be false for the admin commands?

Kimchelly (approved):
LGTM mod trivial question.

operations/admin.go
ablack12: Yeah I kind of feel like it's weird to print the banner when we're specifically updating it or getting settings or etc. Kind of arbitrary but since admin commands are mostly for us and banners are mostly for users it seems reasonable.

PR #6064, opened 2022-11-14
EVG-17054 Conditionally surface Inactive tasks for mainline commits and patches 

khelif96:
[EVG-17054](https://jira.mongodb.org/browse/EVG-17054)

### Description 
This change now enables surfacing unactivated tasks on the commits page and version page if the correct param is passed in. 
Rename `includeEmptyActivation` param to `includeInactiveTasks` so its more descriptive of what it does. I preserved the old field to maintain backward compatibility. I will open a ticket to deprecate it once this has been deployed. 
Before:
image
image

After:
image


model/project_parser_test.go
```diff
@@ -893,7 +893,6 @@ buildvariants:
 	_, err = LoadProjectInto(ctx, []byte(yamlWithVariables), opts, "example_project", &proj)
 	require.NoError(t, err)
 
-	fmt.Println("MY TESTS ARE STARTING")
```
malikchaya2, line 896: 👀  lol

graphql/version_resolver.go
```diff
@@ -305,7 +306,7 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
 		Limit:                          limitParam,
 		Sorts:                          taskSorts,
 		IncludeBaseTasks:               true,
-		IncludeEmptyActivation:         utility.FromBoolPtr(options.IncludeEmptyActivation),
+		IncludeInactiveTasks:           utility.FromBoolPtr(options.IncludeEmptyActivation) || utility.FromBoolPtr(options.IncludeInactiveTasks),
```
malikchaya2, line 309: At what point would we be removing options.IncludeEmptyActivation?

khelif96: Sometime after both the frontend and backend are deployed. I'll open a follow-up ticket for it

ablack12:
I'm concerned about the design of having "Include Inactive tasks" in the dropdown but I also don't remember a lot of the discussion leading up to this. Is that off by default? It doesn't really seem like the right place to have a filter.  Should it instead be added to the Patch Status filter? That's where I would go intuitively to change what statues I wanted to see. 

Also wondering if this is by default added to the waterfall since this should be consistent.
Also do undispatched/unscheduled use the same color now?
(Important note: did not yet read code but had these questions from the description)

graphql/version_resolver.go
malikchaya2: Can we also add that ticket number in a comment here?

khelif96:
> I'm concerned about the design of having "Include Inactive tasks" in the dropdown but I also don't remember a lot of the discussion leading up to this. Is that off by default? It doesn't really seem like the right place to have a filter. Should it instead be added to the Patch Status filter? That's where I would go intuitively to change what statues I wanted to see.
> 
> Also wondering if this is by default added to the waterfall since this should be consistent. Also do undispatched/unscheduled use the same color now? (Important note: did not yet read code but had these questions from the description)

As a follow up I have now removed the option to toggle inactive tasks for mainline commits. They are always enabled by default for mainline commits. For patches, we will hide them. This should be identical to legacy UI functionality and is what we agreed on during the design sync.

malikchaya2 (approved):

ablack12 (changes requested):

graphql/version_resolver.go
```diff
@@ -297,15 +298,16 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
...
+		Statuses:         getValidTaskStatusesFilter(options.Statuses),
+		BaseStatuses:     getValidTaskStatusesFilter(options.BaseStatuses),
+		Variants:         []string{variantParam},
+		TaskNames:        []string{taskNameParam},
+		Page:             pageParam,
+		Limit:            limitParam,
+		Sorts:            taskSorts,
+		IncludeBaseTasks: true,
+		// If the version is a patch, we want to include inactive tasks. Otherwise return it only if requested.
+		IncludeInactiveTasks:           !evergreen.IsPatchRequester(v.Requester) || utility.FromBoolPtr(options.IncludeEmptyActivation) || utility.FromBoolPtr(options.IncludeInactiveTasks),
```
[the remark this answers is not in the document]
khelif96: Sorry i put an incorrect comment. The task statuses should show up automatically if they exist.

model/task/db.go
```diff
@@ -2453,7 +2461,7 @@ type GetTasksByVersionOptions struct {
 	Sorts                               []TasksSortOrder
 	IncludeExecutionTasks               bool
 	IncludeBaseTasks                    bool
-	IncludeEmptyActivation              bool
+	IncludeInactiveTasks                bool // InactiveTasks are defined as tasks that have been generated but lack an activation time
```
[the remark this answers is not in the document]
khelif96: I updated the comment.

graphql/version_resolver.go
```diff
@@ -70,7 +70,7 @@ func (r *versionResolver) BuildVariants(ctx context.Context, obj *restModel.APIV
 		obj.Activated = version.Activated
 	}
 
-	if !utility.FromBoolPtr(obj.Activated) {
+	if evergreen.IsPatchRequester(utility.FromStringPtr(obj.Requester)) && !utility.FromBoolPtr(obj.Activated) {
```
ablack12, line 73: If we're using this in a lot of places, it seems like it could be easier to make a function on APIVersion called IsPatchRequester that just returns this

model/project_parser_test.go
ablack12: yeah who in the world would commit test code like this *hyperventilates*

graphql/util.go
```diff
@@ -280,14 +280,25 @@ func generateBuildVariants(versionId string, buildVariantOpts BuildVariantOption
 	if buildVariantOpts.IncludeBaseTasks == nil {
 		buildVariantOpts.IncludeBaseTasks = utility.ToBoolPtr(true)
 	}
+	v, err := model.VersionFindOneId(versionId)
```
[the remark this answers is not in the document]
khelif96: I opted to just pass in the requester we don't always query for the version object since it's wrapped in a conditional.

graphql/version_resolver.go
khelif96: Done!

model/task/db.go
```diff
@@ -2471,7 +2478,7 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 		match[DisplayNameKey] = bson.M{"$regex": taskNamesAsRegex, "$options": "i"}
 	}
 	// Activated Time is needed to filter out generated tasks that have been generated but not yet activated
-	if !opts.IncludeEmptyActivation {
+	if !opts.IncludeInactiveTasks {
```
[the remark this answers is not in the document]
khelif96: Follow up discussed online renamed it to `IncludeNeverActivatedTasks`

graphql/version_resolver.go
```diff
@@ -297,15 +298,16 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
...
+		Statuses:         getValidTaskStatusesFilter(options.Statuses),
+		BaseStatuses:     getValidTaskStatusesFilter(options.BaseStatuses),
+		Variants:         []string{variantParam},
+		TaskNames:        []string{taskNameParam},
+		Page:             pageParam,
+		Limit:            limitParam,
+		Sorts:            taskSorts,
+		IncludeBaseTasks: true,
+		// If the version is a patch, we want to exclude inactive tasks by default.
+		IncludeNeverActivatedTasks:     !evergreen.IsPatchRequester(v.Requester) || utility.FromBoolPtr(options.IncludeEmptyActivation) || utility.FromBoolPtr(options.IncludeNeverActivatedTasks),
```
ablack12, line 310: Out of curiosity, why do we have both IncludeEmptyActivation AND IncludeNeverActivatedTasks? Couldn't we just combine these?

ablack12: Oh jk i see it's deprecated

rest/model/version.go
```diff
@@ -84,3 +85,7 @@ func (apiVersion *APIVersion) BuildFromService(v model.Version) {
 		}
 	}
 }
+
+func (apiVersion *APIVersion) IsPatchRequester() bool {
+	return evergreen.IsPatchRequester(*apiVersion.Requester)
```
ablack12, line 90: Still preference to use utility.FromBoolPtr

ablack12 (approved):
LGTM with nit!!

graphql/version_resolver.go
khelif96: Yup, we tend to keep around deprecated endpoints for a little bit of time until we ensure we remove all references to it. This just maps the deprecated field to the new field.

PR #6065, opened 2022-11-14
EVG-18222: add one host creation attempt for initial try

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18222

### Description 
There's an option exposed in host.create to specify the number of times that a host.create host is allowed to retry. All I did was make it so that the number of retries they specify means number of times to try again after an initial failure, rather than total number of attempts, which seems more intuitive based on the option name.

### Testing 
Didn't seem that applicable since host creation retries are a minor thing.


ablack12 (approved):

malikchaya2 (approved):
🎉

PR #6070, opened 2022-11-16
EVG-17275 unset parser project / config after finalizing patch

ablack12:

### Description 
Already deployed a change to ensure that patches will use the parser project in the collection if the patch field is unset, so if we need to revert this change that code should be able to work with unset patches in most places (although if we need to revert it'll be because we found somewhere that _didn't_ work for unset patches, so those will continue to not work without a hot fix I think).


ablack12:
(Asking for review but not merging until next week since it's a bit risky)

malikchaya2 (approved):
I am scared, lol

PR #6072, opened 2022-11-17
EVG-18050 add iops to io1 volumes

ablack12:
[EVG-18050](https://jira.mongodb.org/browse/EVG-18050)

### Description 
Iops is required for io1 so adding that; using 3000 because this is the default (if changing types, according to AWS) and I don't think anyone is going to want/need more than that.

### Testing 
Tested that creating io1 works now on staging.

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -232,6 +232,8 @@ const (
 	SpotStatusFailed   = "failed"
 
 	EC2ErrorSpotRequestNotFound = "InvalidSpotInstanceRequestID.NotFound"
+
+	defaultIops = 3000
 )
 
 const (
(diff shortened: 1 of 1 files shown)


malikchaya2 (approved):

ybrill:
@ablack12 can you explain more what you mean that 3000 is the default? 
(I'm asking because it could be relevant to what I'm doing for [EVG-16702](https://jira.mongodb.org/browse/EVG-16702))

ablack12:
Sure, I'm referencing [this line from the SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#ModifyVolumeInput :) specifically, for modifying volume type:

>  // Default: The existing value is retained if you keep the same volume type.
    // If you change the volume type to io1, io2, or gp3, the default is 3,000.

ybrill:
ah, got it.
thanks!

PR #6073, opened 2022-11-17
EVG-16649 Update disabling task to use same logic as deactivating

bynn:
[EVG-16649](https://jira.mongodb.org/browse/EVG-16649)

### Description 
disable uses deactivate logic now (which means it will dequeue and stop task and update display task)
we dont need to check for single host tasks because it deactivating dependencies will take care of it 

### Testing 
lot more tests


model/task_lifecycle.go
```diff
@@ -77,16 +77,6 @@ func SetActiveState(caller string, active bool, tasks ...task.Task) error {
 			// If the task was not activated by step back, and either the caller is not evergreen
 			// or the task was originally activated by evergreen, deactivate the task
 		} else if !evergreen.IsSystemActivator(caller) || evergreen.IsSystemActivator(t.ActivatedBy) {
-			// deactivate later tasks in the group as well, since they won't succeed without this one
```
Kimchelly, line 80: I read through the reasoning in the ticket, but I think the hypothesis about `DeactivateDependencies` deactivating single-host task groups is incorrect. I believe (@ablack12 double-check me that this is still true?) that the dependencies are absent when the tasks are first created, and the dependency chain between single-host task group tasks is only retroactively created _after_ the task runs and fails (using [BlockTaskGroupTasks](https://github.com/evergreen-ci/evergreen/blob/033f687de1f50c9c50f22a4e4090ee0e9661a1be/model/task_queue.go#L240-L293) to make the dependencies and mark them unattainable). Therefore, I would imagine that just blocking dependencies isn't guaranteed to block a single-host task group if it hasn't run yet. If we can't guarantee that single-host task groups have the dependencies set up when deactivating/disabling, I think we have to keep this logic.

Kimchelly (changes requested):

model/lifecycle.go
```diff
@@ -252,7 +252,7 @@ func SetTaskPriority(t task.Task, priority int64, caller string) error {
 
 	// negative priority - deactivate the task
 	if priority <= evergreen.DisabledTaskPriority {
-		if err = t.DeactivateTask(caller); err != nil {
+		if err = SetActiveState(caller, false, t); err != nil {
```
hadjri, line 255: There are a couple other key places where we currently skip the logic from `SetActiveState` when dealing with disabled tasks.  I think [SetDisabledPriority](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L2200), [DisableTasks](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L2227), [SetVersionPriority](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L288), and [SetBuildPriority](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L264) are instances that will need this change as well.

model/task_lifecycle_test.go
```diff
@@ -38,38 +38,81 @@ var (
 
 func TestSetActiveState(t *testing.T) {
 	Convey("With one task with no dependencies", t, func() {
-		require.NoError(t, db.ClearCollections(task.Collection, build.Collection, task.OldCollection, VersionCollection))
+		require.NoError(t, db.ClearCollections(task.Collection, build.Collection, task.OldCollection, VersionCollection, commitqueue.Collection))
```
hadjri, line 41: Can we also add cases for 1) checking for appropriate updates of associated build / version statuses and 2) an execution task updating its display task?

hadjri (changes requested):

model/task_lifecycle_test.go
bynn: added

model/task_lifecycle.go
```diff
@@ -123,6 +121,11 @@ func SetActiveState(caller string, active bool, tasks ...task.Task) error {
 		}
 	}
 
+	for _, t := range tasksToActivate {
```
bynn, line 124: display tasks werent correctly updated before because we never deactivated the execution tasks before we updated them

model/task_lifecycle.go
ablack12: cc @hadjri you may want to decide if this is out of scope of the project or not (I didn't read closely into it but can discuss if it's not clear)

Kimchelly (approved):

model/task_lifecycle.go
hadjri: I suppose it's reasonable to include this in the project since it has spawned off of this PR and is still relevant to the original ticket.

model/lifecycle.go
hadjri: Re our conversation `SetDisabledPriority` will be removed since it's functionality is covered by `DisableTasks` and `DisableTasks` will be moved up to the model package.

hadjri (changes requested):
Re our conversation SetDisabledPriority will be removed since it's functionality is covered by DisableTasks and DisableTasks will be moved up to the model package.

hadjri (approved):
LGTM mod test fixes!

PR #6074, opened 2022-11-17
EVG-16702 gp3 default volume type

ybrill:
[EVG-16702](https://jira.mongodb.org/browse/EVG-16702)

### Description 
Use gp3 volumes when we provision a home volume for a workstation. Other cases too, but they seem pretty rare (does anyone ever provision a version through the API?)

Specify throughput and IOPS that's either equivalent to or better than the current levels with gp2 (based on the AWS doc [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/general-purpose.html)).

I'd originally planned to let users select the IOPS/throughput they wanted so they could migrate their existing volumes, but maybe we're okay to leave them alone. If we really want to we can do a migration on the backend since volume changes are supposed to work in situ.

### Testing 
I created a workstation in staging with a 1500GiB home volume. It got created as a gp3 volume with 4500 IOPS and 250 MiB/s.
Screen Shot 2022-11-17 at 4 42 54 PM


cloud/ec2.go
```diff
@@ -1489,10 +1489,17 @@ func (m *ec2Manager) CreateVolume(ctx context.Context, volume *host.Volume) (*ho
 			{ResourceType: aws.String(ec2.ResourceTypeVolume), Tags: volumeTags},
 		},
 	}
-	// Iops is required for io1.
-	if volume.Type == VolumeTypeIo1 {
```
ablack12, line 1493: so short lived 😂  can we still keep the comment though

ablack12 (approved):
LGTM with nit

cloud/ec2.go
ybrill: ah I meant to keep it and then forgot about it when I couldn't figure out where to put it.
An inline comment? Or inside the else if?

ablack12:
Oh also can you tweak the title to follow style guidelines?

cloud/ec2.go
ablack12: both work, this looks good to me!

ybrill:
😃 done.

PR #6076, opened 2022-11-17
EVG-18115 Validate host status before adding ssh key

bynn:
[EVG-18115](https://jira.mongodb.org/browse/EVG-18115)

### Description 
only allow key adding if host status is running

### Testing 
graphql test

### DIFF
--- a/graphql/mutation_resolver.go
+++ b/graphql/mutation_resolver.go
@@ -627,7 +627,7 @@ func (r *mutationResolver) EditSpawnHost(ctx context.Context, spawnHost *EditSpa
 	}
 	if spawnHost.PublicKey != nil {
 		if h.Status != evergreen.HostRunning {
-			return nil, InputValidationError.Send(ctx, "Host must be running to add a public key")
+			return nil, InputValidationError.Send(ctx, fmt.Sprintf("Host must be running to add a public key but is '%s'", h.Status))
 		}
 		if utility.FromBoolPtr(spawnHost.SavePublicKey) {
 			if err = savePublicKey(ctx, *spawnHost.PublicKey); err != nil {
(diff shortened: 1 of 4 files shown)


graphql/mutation_resolver.go
```diff
@@ -626,6 +626,9 @@ func (r *mutationResolver) EditSpawnHost(ctx context.Context, spawnHost *EditSpa
 		opts.AttachVolume = *spawnHost.Volume
 	}
 	if spawnHost.PublicKey != nil {
+		if h.Status != evergreen.HostRunning {
+			return nil, InputValidationError.Send(ctx, "Host must be running to add a public key")
```
Kimchelly, line 630: nit: can we mention the host's current status in the error message?

Kimchelly (approved):
LGTM mod nit.

khelif96 (approved):

graphql/mutation_resolver.go
khelif96, line 630: You should also add some front-end validation for this. Otherwise, the request will just fail and it will be confusing as to why. We should prevent the interaction from happening altogether.

ablack12: Won't users see this error?

khelif96: We should disable the option to change the ssh key if the host is not running allowing a user to perform an operation that will always fail is a bad experience.

ablack12: Yeah that sounds good. We should also make sure this mouseover (that exists for instance type) shows up. 
image

If that seems too large @bynn feel free to cut a new ticket; I think making sure the users know about the error is priority #1.

bynn: https://jira.mongodb.org/browse/EVG-18003

I created a ticket for the frontend part of the validation!

PR #6079, opened 2022-11-18
EVG-17196 avoid setting defaults when reusing definitions

ablack12:
[EVG-17196](https://jira.mongodb.org/browse/EVG-17196 )

### Description 
Reuse doesn't work if any defaults are set for the user. 
### Testing 
Verified that this happens in prod and not with the changes.
```
evergreen [EVG-17196] $ eg patch -u --reuse     
Enter a description for this patch (optional): test
can't define tasks/variants when reusing previous patch's tasks and variants

evergreen [EVG-17196] $ egss patch --reuse          
<success>
```

### DIFF
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -221,28 +221,31 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
 		grip.Warningf("warning - failed to set default project: %v\n", err)
 	}
 
-	if err := p.setLocalAliases(conf); err != nil {
-		grip.Warningf("warning - setting local aliases")
-	}
+	// If reusing a previous definition, ignore defaults.
+	if !p.RepeatFailed && !p.RepeatDefinition {
+		if err := p.setLocalAliases(conf); err != nil {
+			grip.Warningf("warning - setting local aliases")
+		}
 
-	if err := p.loadAlias(conf); err != nil {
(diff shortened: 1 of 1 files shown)


operations/patch_util.go
```diff
@@ -221,30 +221,33 @@ func (p *patchParams) validatePatchCommand(ctx context.Context, conf *ClientSett
 		grip.Warningf("warning - failed to set default project: %v\n", err)
 	}
 
-	if err := p.setLocalAliases(conf); err != nil {
-		grip.Warningf("warning - setting local aliases")
-	}
+	// If reusing a previous definition, ignore defaults.
+	if !p.RepeatFailed && !p.RepeatDefinition {
+		if err := p.setLocalAliases(conf); err != nil {
+			grip.Warningf("warning - setting local aliases")
```
malikchaya2, line 227: I know this isn't new, but I'm curious why this is a warning as opposed to informational.

malikchaya2 (approved):
Approved with minor nit/question.

operations/patch_util.go
ablack12: Technically users can set the logging level they want for the CLI; theoretically this is a thing they'd expect (but not a thing we'd want to not create patches for) which is why I could imagine warning being the right level (whereas info tends to just be general "submitted /created /etc" logging) but also it could've just been not thought out haha

PR #6082, opened 2022-11-22
EVG-16857 add flags to repeat specific patch IDs

ablack12:
[EVG-16857](https://jira.mongodb.org/browse/EVG-16857)

### Description 
Allows users to specify what patch they want to repeat, rather than using the most recent one. I wanted there to be a way to just use the --repeat and --repeat-failed flags but I can't figure a way to have it be a string flag and not expect a value; CLI defaults exist but that default is used when they _don't_ include the flag.

SO I added two more flags. Output from the patch help page now looks like:
```
  --repeat, --reuse                          use the same tasks/variants defined for the last patch you scheduled for this project
   --repeat-failed, --rf                      use the failed tasks/variants defined for the last patch you scheduled for this project
   --repeat-patch value, --reuse-patch value  use the same tasks/variants defined for the given patch
   --repeat-failed-patch value                use the failed tasks/variants defined for the given patch
```
This is a bit clunky but verified with Sanjog that it's still probably worth doing (he's going to do a bit of investigation with users to verify though).

### Testing 
Tested using [--repeat-patch](https://spruce-staging.corp.mongodb.com/patch/637bea9cb237362349ad08c1/configure/tasks) and [--repeat-failed-patch](https://spruce-staging.corp.mongodb.com/patch/637beace9dbe320be63e55a2/configure/tasks).


operations/flags.go
```diff
@@ -298,6 +299,28 @@ func addUncommittedChangesFlag(flags ...cli.Flag) []cli.Flag {
 	})
 }
 
+func addReuseFlags(flags ...cli.Flag) []cli.Flag {
```
malikchaya2, line 302: I think the help output might be easier to read quickly if it looked something like this: 
image

malikchaya2 (changes requested):

operations/flags.go
ablack12: This isn't something we can control easily since it's handled by the library: https://github.com/urfave/cli/blob/main/help.go#L243

malikchaya2: I'm referring to appending either "repeat latest: " or "repeat specific: " to `message`

ablack12: ohh i see, thanks. sure

malikchaya2 (approved):

ablack12:
Discussed with Sanjog; we're going to remove the --repeat-failed-patch flag since there's not as much use for it.

units/patch_intent.go
```diff
@@ -510,25 +513,41 @@ func setTasksToPreviousFailed(patchDoc, previousPatch *patch.Patch, project *mod
...
+		if err != nil {
+			return "", errors.Wrap(err, "querying for most recent patch")
+		}
+		if reusePatch == nil {
+			return "", errors.Errorf("no previous patch available")
+		}
+	} else {
+		reusePatch, err = patch.FindOneId(patchId)
+		if err != nil {
+			return "", errors.Wrap(err, "querying for most recent patch")
```
bynn, line 534: error message should be something like "querying for patch '%s'" instead of most recent i think

bynn (changes requested):

ablack12:
Verified that this still works for repeat-patch: https://spruce-staging.corp.mongodb.com/patch/6387a3bc97b1d3634f4aa3c5/configure/tasks

bynn (approved):
whooo looks good! :shipit:

PR #6085, opened 2022-11-22
EVG-18199 Return links to Parsley for test results

khelif96:
[EVG-18199](https://jira.mongodb.org/browse/EVG-18199)

### Description 
We now resolve a `urlParsley` field for test results which should return a link to parsley instead of lobster for both logkeeper hosted test results and cedar hosted ones. Updates links on both new UI and legacy UI for test results.

image
image


model/task/task.go
```diff
@@ -412,17 +412,18 @@ func (tr TestResult) GetDisplayTestName() string {
...
 	deprecatedLobsterURLs := []string{"https://logkeeper.mongodb.org", "https://logkeeper2.build.10gen.cc"}
 
 	switch viewer {
 	case evergreen.LogViewerHTML:
+		// Return an empty string for logkeeper urls
 		if tr.URL != "" {
 			for _, url := range deprecatedLobsterURLs {
 				if strings.Contains(tr.URL, url) {
-					return strings.Replace(tr.URL, url, root+"/lobster", 1)
+					return ""
```
khelif96, line 424: This is the most notable change in behavior we no longer return a legacy lobster URL for the HTML field for resmoke logs. I made this change in an effort to consolidate the logic between resmoke and non resmoke logs.

ablack12 (approved):
Contextless LGTM

model/task/task.go
```diff
@@ -412,17 +412,18 @@ func (tr TestResult) GetDisplayTestName() string {
 // generation for other log viewers.
 func (tr TestResult) GetLogURL(viewer evergreen.LogViewer) string {
 	root := evergreen.GetEnvironment().Settings().ApiUrl
+	parsleyURL := evergreen.GetEnvironment().Settings().Ui.ParsleyUrl
 	deprecatedLobsterURLs := []string{"https://logkeeper.mongodb.org", "https://logkeeper2.build.10gen.cc"}
 
 	switch viewer {
 	case evergreen.LogViewerHTML:
+		// Return an empty string for logkeeper urls
```
julianedwards, line 420: Nit: `URLs.`

model/task/task.go
julianedwards: For resmoke logs, will there just be a `Parlsey` and `Raw` link then?

model/task/task.go
```diff
@@ -463,6 +469,18 @@ func (tr TestResult) GetLogURL(viewer evergreen.LogViewer) string {
 			url.QueryEscape(tr.GroupID),
 			tr.LineNum,
 		)
+	case evergreen.LogViewerParsley:
+		if parsleyURL == "" {
+			return ""
+		}
+		for _, url := range deprecatedLobsterURLs {
+			if strings.Contains(tr.URL, url) {
+				updatedResmokeParsleyURL := strings.Replace(tr.URL, fmt.Sprintf("%s/build", url), parsleyURL+"/resmoke", 1)
```
julianedwards, line 478: Would it make more sense to link to Parsley from Logkeeper instead, in the same way that we currently redirect to Lobster? I think it would avoid the URL surgery here and simplify the code.

julianedwards (changes requested):

model/task/task.go
khelif96: Yes, i think so! But we would still need to keep this URL surgery, for now, to account for older links.

julianedwards (approved):

PR #6086, opened 2022-11-23
EVG-17033 End max execution and unschedulable stranded tasks before marking system failed

hadjri:
[EVG-17033](https://jira.mongodb.org/browse/EVG-17033)

### Description 
There is currently a bug where ending stranded tasks rather than resetting them (either because they have breached the unschedulable threshold or they have breached the max execution) causes `MarkEnd` to return early and not perform the necessary steps of updating dependencies, updating build/version statuses, updating display task statuses, etc.  

Change has been made to not call `MarkSystemFailed` before calling `MarkEnd` in these cases because it would result in a premature return [here](https://github.com/evergreen-ci/evergreen/blob/c2844d0184381b161f61402ddc599253c2b9a248/model/task_lifecycle.go#L477-L483). Passing stranded system failure details to `MarkEnd` will allow it to perform the same function as `MarkSystemFailed` in these cases.  

I also did some light refactoring on how we construct system failed task details based off an input description since I noticed the same logic was used in a couple places.
### Testing 
TDD approach, wrote some extra tests in TestClearAndResetStrandedHostTask, and certified that they failed before the change and succeeded afterwards.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1503,15 +1503,7 @@ func (t *Task) MarkFailed() error {
 func (t *Task) MarkSystemFailed(description string) error {
 	t.Status = evergreen.TaskFailed
 	t.FinishTime = time.Now()
-
-	t.Details = apimodels.TaskEndDetail{
-		Status:      evergreen.TaskFailed,
-		Type:        evergreen.CommandTypeSystem,
-		Description: description,
-	}
-	if description == evergreen.TaskDescriptionHeartbeat {
-		t.Details.TimedOut = true
-	}
+	t.Details = GetSystemFailureDetails(description)
 
 	switch t.ExecutionPlatform {
 	case ExecutionPlatformHost:
@@ -1550,6 +1542,18 @@ func (t *Task) MarkSystemFailed(description string) error {
 			},
 		},
 	)
+}
+
+func GetSystemFailureDetails(description string) apimodels.TaskEndDetail {
+	details := apimodels.TaskEndDetail{
+		Status:      evergreen.TaskFailed,
+		Type:        evergreen.CommandTypeSystem,
+		Description: description,
+	}
+	if description == evergreen.TaskDescriptionHeartbeat {
+		details.TimedOut = true
+	}
+	return details
 }
 
 func SetManyAborted(taskIds []string, reason AbortInfo) error {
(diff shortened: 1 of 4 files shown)


model/task/task.go
```diff
@@ -1552,6 +1544,18 @@ func (t *Task) MarkSystemFailed(description string) error {
 	)
 }
 
+func GetSystemFailureDetails(description string) apimodels.TaskEndDetail {
```
ablack12, line 1547: should have function comment since public

model/task_lifecycle.go
```diff
@@ -1725,27 +1725,38 @@ func resetSystemFailedTask(t *task.Task, description string) error {
...
+	maxExecutionTask := t.Execution >= evergreen.MaxTaskExecution
+	if unschedulableTask || maxExecutionTask {
+		failureDetails := task.GetSystemFailureDetails(description)
 		// If the task has already exceeded the unschedulable threshold, we
 		// don't want to restart it, so just mark it as finished.
+		msg := "task exceeded unschedulable threshold"
+		if maxExecutionTask {
+			msg = "task exceeded max execution threshold"
+		}
+		grip.Warning(message.Fields{
```
ablack12, line 1738: Can we remove this? MarkEnd already does some logging so this seems redundant.

model/task_lifecycle.go
```diff
@@ -1725,27 +1725,38 @@ func resetSystemFailedTask(t *task.Task, description string) error {
...
 				if !evergreen.IsFinishedTaskStatus(execTask.Status) {
-					if err := MarkEnd(&execTask, evergreen.MonitorPackage, time.Now(), &t.Details, false); err != nil {
+					if err = MarkEnd(&execTask, evergreen.MonitorPackage, time.Now(), &failureDetails, false); err != nil {
 						return errors.Wrap(err, "marking execution task as ended")
 					}
 				}
 			}
 		}
-		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &t.Details, false))
+		return errors.WithStack(MarkEnd(t, evergreen.MonitorPackage, time.Now(), &failureDetails, false))
```
ablack12, line 1755: Could we also just pass in systemFailureDetails here (or some sort of system failure) and then not have to hit 1758 at all?

ablack12 (changes requested):

model/task_lifecycle.go
ablack12: Sorry I misread; it looks like we need that for the reset

ablack12 (approved):

PR #6087, opened 2022-11-23
EVG-16982 respect disabled for cron

ablack12:
[EVG-16982](https://jira.mongodb.org/browse/EVG-16982)

### Description 
Activation right now checks activate but not disabled. Additionally, there are a few places where we short circuit if cron/batchtime aren't defined, but we don't check either activate or disabled before we do that.

### Testing 
Added a test.

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -1293,7 +1293,7 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, er
 		"last good version for project '%s'", lastGoodVersion.Identifier)
 }
 
-// HasSpecificActivation returns if the build variant specifies an overriding activation,
+// HasSpecificActivation returns if the build variant task specifies an overriding activation,
 // such as cron/batchtime, disabling the task, or explicitly activating it.
 func (bvt *BuildVariantTaskUnit) HasSpecificActivation() bool {
 	return bvt.CronBatchTime != "" || bvt.BatchTime != nil || bvt.Activate != nil || bvt.IsDisabled()
--- a/model/project.go
+++ b/model/project.go
@@ -1293,8 +1293,8 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, er
 		"last good version for project '%s'", lastGoodVersion.Identifier)
 }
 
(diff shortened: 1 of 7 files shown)


model/project.go
```diff
@@ -1293,8 +1293,10 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, er
 		"last good version for project '%s'", lastGoodVersion.Identifier)
 }
 
-func (bvt *BuildVariantTaskUnit) HasBatchTime() bool {
-	return bvt.CronBatchTime != "" || bvt.BatchTime != nil || bvt.Activate != nil
+// HasSpecificActivation returns if the build variant task specifies an overriding activation,
```
Kimchelly, line 1296: The wording here was a little confusing for me to understand, can this be phrased to mention that the default is to unconditionally activate the task? Something like "returns if the build variant task specifies an activation condition that overrides the default to activate the task"?

model/project_parser.go
```diff
@@ -484,6 +484,20 @@ func (pss *parserStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) e
 	return nil
 }
 
+// hasSpecificActivation returns if the build variant task unit specifies an overriding activation,
```
Kimchelly, line 487: Same question as above.

model/project_parser.go
```diff
@@ -484,6 +484,20 @@ func (pss *parserStringSlice) UnmarshalYAML(unmarshal func(interface{}) error) e
...
+// such as cron/batchtime, disabling the task, or explicitly activating it.
+func (bvt *parserBVTaskUnit) hasSpecificActivation() bool {
+	return bvt.BatchTime != nil || bvt.CronBatchTime != "" ||
+		bvt.Activate != nil || utility.FromBoolPtr(bvt.Disable)
+}
+
+// hasSpecificActivation returns if the build variant specifies an overriding activation,
```
Kimchelly, line 494: Same question as above.

model/project.go
ablack12: Tweaked, how is that?

Kimchelly: Perfect, thanks!

Kimchelly (approved):

PR #6088, opened 2022-11-23
EVG-18056 Add option to not have tasks depending on the generator depend on generated tasks

hadjri:
[EVG-18056](https://jira.mongodb.org/browse/EVG-18056)

### Description 
The current default behavior for a task that depends on a generator task is to add all of the generator's activated generated tasks as dependencies.  Change here introduces an optional flag on the `generate.tasks` command to disable this behavior.

GeneratedTasksAreNotDependencies is not a very pretty variable name so if there are any suggestions to make it more succinct I'm all ears 😅 
### Testing 
Added new unit test certifying that the default behavior for tasks depending on the generator task is to also depend on all of the generated tasks, and that setting `GeneratedTasksAreNotDependencies` disables this behavior


ablack12 (changes requested):
Sorry this throws a wrench in things -- I kind of think it makes more sense to make this an optional parameter of DependsOn, right? Theoretically taskA could want to depend on task_gen and all created tasks, whereas taskB just wants to depend on task_gen. And then we could add that field to https://github.com/evergreen-ci/evergreen/blob/6ad3636179afc34af53c42b7ea479ef4d61e2e25/model/task/task.go#L3489 

Also, if it makes the defaulting easier we could assume this defaults to true and make it a boolean pointer, so we only skip adding children if the variable is actively set to false.

hadjri:
> Sorry this throws a wrench in things -- I kind of think it makes more sense to make this an optional parameter of DependsOn, right? Theoretically taskA could want to depend on task_gen and all created tasks, whereas taskB just wants to depend on task_gen. And then we could add that field to
> 
> https://github.com/evergreen-ci/evergreen/blob/6ad3636179afc34af53c42b7ea479ef4d61e2e25/model/task/task.go#L3489
> 
> Also, if it makes the defaulting easier we could assume this defaults to true and make it a boolean pointer, so we only skip adding children if the variable is actively set to false.

Ok I think that's reasonable. We'd discussed adding it as an option to generate.tasks previously which is why I implemented it this way but I agree moving the setting to `DependsOn` is more flexible.

model/project.go
```diff
@@ -535,10 +535,11 @@ func (c *YAMLCommandSet) UnmarshalYAML(unmarshal func(interface{}) error) error
...
 type TaskUnitDependency struct {
-	Name          string `yaml:"name,omitempty" bson:"name"`
-	Variant       string `yaml:"variant,omitempty" bson:"variant,omitempty"`
-	Status        string `yaml:"status,omitempty" bson:"status,omitempty"`
-	PatchOptional bool   `yaml:"patch_optional,omitempty" bson:"patch_optional,omitempty"`
+	Name                           string `yaml:"name,omitempty" bson:"name"`
+	Variant                        string `yaml:"variant,omitempty" bson:"variant,omitempty"`
+	Status                         string `yaml:"status,omitempty" bson:"status,omitempty"`
+	PatchOptional                  bool   `yaml:"patch_optional,omitempty" bson:"patch_optional,omitempty"`
+	DependOnGeneratedTasksDisabled bool   `yaml:"depend_on_generated_tasks_disabled" bson:"depend_on_generated_tasks_disabled"`
```
ablack12, line 542: maybe just "include_generated_tasks" and we can have this be a bool pointer so the default is true? Or `omit_generated_tasks` so that false can still be the default

units/generate_tasks_test.go
```diff
@@ -266,6 +431,128 @@ func TestGenerateTasks(t *testing.T) {
...
+
+	j := NewGenerateTasksJob(generateTask.Version, generateTask.Id, "1")
+	j.Run(context.Background())
+	assert.NoError(j.Error())
+	tasks := []task.Task{}
+	assert.NoError(db.FindAllQ(task.Collection, db.Query(task.ByVersion("sample_version")), &tasks))
+	assert.Len(tasks, 4)
+	for _, foundTask := range tasks {
+		switch foundTask.DisplayName {
+		case "version_gen":
```
ablack12, line 499: nit: I think you can do `case "version_gen", "dependency_task":` since checking the same thing

model/task/task.go
```diff
@@ -3487,8 +3490,9 @@ func (t *Task) UpdateDependsOn(status string, newDependencyIDs []string) error {
 	_, err := UpdateAll(
 		bson.M{
 			DependsOnKey: bson.M{"$elemMatch": bson.M{
-				DependencyTaskIdKey: t.Id,
-				DependencyStatusKey: status,
+				DependencyTaskIdKey:                         t.Id,
+				DependencyStatusKey:                         status,
+				DependencyDependOnGeneratedTasksDisabledKey: false,
```
ablack12, line 3495: Can you add a sentence to the function comment to explain this field?

model/project.go
hadjri: Ok! `omit_generated_tasks` sounds elegant

ablack12 (approved):
LGTM after verifying query will work with documents that don't have omit_generated_tasks set to anything

model/task/task.go
```diff
@@ -3490,9 +3491,9 @@ func (t *Task) UpdateDependsOn(status string, newDependencyIDs []string) error {
 	_, err := UpdateAll(
 		bson.M{
 			DependsOnKey: bson.M{"$elemMatch": bson.M{
-				DependencyTaskIdKey:                         t.Id,
-				DependencyStatusKey:                         status,
-				DependencyDependOnGeneratedTasksDisabledKey: false,
+				DependencyTaskIdKey:             t.Id,
+				DependencyStatusKey:             status,
+				DependencyOmitGeneratedTasksKey: false,
```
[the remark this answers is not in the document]
hadjri: It will work as is because the `Dependency` struct that gets computed from the parser dependencies (which do have omitempty set) doesn't use omitempty for any fields: [link](https://github.com/evergreen-ci/evergreen/pull/6088/files#diff-4b29d204064065e99ec7762e466f51f0ac8461fcc5f8fc2eeda53d87a5044060R312).  I did this for `OmitGeneratedTasks` to be consistent with `Unattainable` and `Finished`, but I suppose it makes sense to add omitempty here as well. Now we do need $ne: true, I've updated it.

PR #6089, opened 2022-11-23
EVG-17448 configurable webhook retries

ybrill:
[EVG-17448](https://jira.mongodb.org/browse/EVG-17448)

### Description 
Allow users to configure retries and timeout for a webhook notification.
We thought to just make it retry by default, but that would have caused minor issues with the pr bot.

My plan is to open a UI ticket to surface the new fields in the project settings. In the meantime, I'll set them in the db for the subscriptions referenced in the ticket.

I skipped adding the new fields to the old UI. I think this is our current posture? If not I can add them there.

### Testing 
Unit tests for the new fields in the subscriber/API subscriber and for retries.
I'll try it out in staging when I see an opening there.

### DIFF
--- a/model/event/subscribers.go
+++ b/model/event/subscribers.go
@@ -21,6 +21,10 @@ const (
 	EnqueuePatchSubscriberType      = "enqueue-patch"
 	SubscriberTypeNone              = "none"
 	RunChildPatchSubscriberType     = "run-child-patch"
+
+	webhookRetryLimit    = 10
+	webhookMinDelayLimit = 10000
+	webhookTimeoutLimit  = 30000
 )
 
 var SubscriberTypes = []string{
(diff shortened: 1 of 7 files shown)


ablack12 (commented):
In the UI ticket would we specify what we'd want the limits for these to be or should we add some validation to these in this ticket? 

Is it possible to have backend retries for only some kinds of webhooks if it causes a problem for others? (note: i don't have a lot of context on how this works)

ybrill:
I added backend validation.
The UI ticket should do retries on the front end too, I think.

wrt having retries for some and not others, as long as a webhook isn't configured to use retries it shouldn't be affected.

model/event/subscribers.go
```diff
@@ -135,6 +150,28 @@ func (s *WebhookSubscriber) String() string {
 	return s.URL
 }
 
+func (s *WebhookSubscriber) validate() error {
+	catcher := grip.NewBasicCatcher()
+	catcher.AddWhen(s.URL == "", errors.New("url cannot be empty"))
+	catcher.AddWhen(len(s.Secret) == 0, errors.New("secret cannot be empty"))
+
+	catcher.AddWhen(s.Retries < 0, errors.New("retries cannot be negative"))
+	catcher.AddWhen(s.Retries > webhookRetryLimit, errors.Errorf("retries cannot be greater than %d", webhookRetryLimit))
```
ablack12, line 159: suggestion: cannot retry more than %d times

ablack12 (approved):

model/event/subscribers.go
ybrill: done!

PR #6090, opened 2022-11-28
EVG-14528 Add TriggerFamilyOutcome type to events

malikchaya2:

### Description 
Before this change, when a patch or version was done, we would check in the trigger outcome to see if the children are done, and if they were not, we would exit.

What this does instead, is it creates an event when the children and parent are both done and subscribes on that event in situations where we only want to do something once everything is done. 

### Note 1 
Now that the family trigger is set up, the way we process Github child patch triggers can be optimized as well. I created [EVG-18381](https://jira.mongodb.org/browse/EVG-18381) to track that. 

### Note 2
In order for patch outcome notifications (the ones set up [here](https://spruce-staging.corp.mongodb.com/preferences/notifications)) to work after this change, the existing subscription db entries need to be updated. I will track that work [here](https://jira.mongodb.org/browse/EVG-18382). 

### Testing 
Tested on staging

### DIFF
--- a/model/event/version_event.go
+++ b/model/event/version_event.go
@@ -11,6 +11,7 @@ func init() {
 	registry.AddType(ResourceTypeVersion, versionEventDataFactory)
 	registry.AllowSubscription(ResourceTypeVersion, VersionStateChange)
 	registry.AllowSubscription(ResourceTypeVersion, VersionGithubCheckFinished)
+	registry.AllowSubscription(ResourceTypeVersion, VersionChildrenCompletion)
 }
 
 func versionEventDataFactory() interface{} {
@@ -21,11 +22,13 @@ const (
 	ResourceTypeVersion        = "VERSION"
 	VersionStateChange         = "STATE_CHANGE"
 	VersionGithubCheckFinished = "GITHUB_CHECK_FINISHED"
+	VersionChildrenCompletion  = "CHILDREN_FINISHED"
 )
 
 type VersionEventData struct {
 	Status            string `bson:"status,omitempty" json:"status,omitempty"`
 	GithubCheckStatus string `bson:"github_check_status,omitempty" json:"github_check_status,omitempty"`
+	Author            string `bson:"author,omitempty" json:"author,omitempty"`
 }
 
 func LogVersionStateChangeEvent(id, newStatus string) {
@@ -56,6 +59,27 @@ func LogVersionGithubCheckFinishedEvent(id, newStatus string) {
 		EventType:    VersionGithubCheckFinished,
 		Data: &VersionEventData{
 			GithubCheckStatus: newStatus,
+		},
+	}
+
+	if err := event.Log(); err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
+			"resource_type": ResourceTypeVersion,
+			"message":       "error logging event",
+			"source":        "event-log-fail",
+		}))
+	}
+}
+
+func LogVersionChildrenCompletionEvent(id, status, author string) {
+	event := EventLogEntry{
+		Timestamp:    time.Now().Truncate(0).Round(time.Millisecond),
+		ResourceId:   id,
+		ResourceType: ResourceTypeVersion,
+		EventType:    VersionChildrenCompletion,
+		Data: &VersionEventData{
+			Status: status,
+			Author: author,
 		},
 	}
 
--- a/model/patch/patch.go
+++ b/model/patch/patch.go
@@ -725,6 +725,32 @@ func (p *Patch) IsChild() bool {
 	return p.Triggers.ParentPatch != ""
 }
 
+func (p *Patch) CollectiveStatus() (string, error) {
+	parentPatch := p
+	if p.IsChild() {
+		parentPatch, err := FindOneId(p.Triggers.ParentPatch)
+		if err != nil {
+			return "", errors.Wrap(err, "getting parent patch")
+		}
+		if parentPatch == nil {
+			return "", errors.Errorf(fmt.Sprintf("parent patch '%s' does not exist", p.Triggers.ParentPatch))
+		}
+	}
+	allStatuses := []string{parentPatch.Status}
+	for _, childPatchId := range parentPatch.Triggers.ChildPatches {
+		cp, err := FindOneId(childPatchId)
+		if err != nil {
+			return "", errors.Wrapf(err, "getting child patch '%s' ", childPatchId)
+		}
+		if cp == nil {
+			return "", errors.Wrapf(err, "child patch '%s' not found", childPatchId)
+		}
+		allStatuses = append(allStatuses, cp.Status)
+	}
+
+	return GetCollectiveStatus(allStatuses), nil
+}
+
 func (p *Patch) IsParent() bool {
 	return len(p.Triggers.ChildPatches) > 0
 }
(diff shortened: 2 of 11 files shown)


model/event/version_event.go
```diff
@@ -67,3 +70,24 @@ func LogVersionGithubCheckFinishedEvent(id, newStatus string) {
 		}))
 	}
 }
+
+func LogVersionChildrenCompletionEvent(id, status, author string) {
```
hadjri, line 74: Seems as though this and `LogPatchChildrenCompletionEvent` could share a helper function that passes in ResourceType, EventType as parameters

model/patch/patch.go
```diff
@@ -752,6 +778,65 @@ func (p *Patch) GetPatchIndex(parentPatch *Patch) (int, error) {
 	return -1, nil
 }
 
+func (p *Patch) GetFamilyInformation() (bool, *Patch, error) {
+	if !p.IsChild() && !p.IsParent() {
+		if evergreen.IsFinishedPatchStatus(p.Status) {
```
hadjri, line 783: Can just `return evergreen.IsFinishedPatchStatus(p.Status), nil, nil`

model/patch/patch.go
```diff
@@ -752,6 +778,65 @@ func (p *Patch) GetPatchIndex(parentPatch *Patch) (int, error) {
...
+	}
+
+	isDone := false
+	childrenOrSiblings, parentPatch, err := p.GetPatchFamily()
+	if err != nil {
+		return isDone, parentPatch, errors.Wrap(err, "getting child or sibling patches")
+	}
+
+	// make sure the parent is done, if not, wait for the parent
+	if p.IsChild() {
```
hadjri, line 797: nit: can use && instead of nested if

model/task_lifecycle.go
```diff
@@ -1230,6 +1231,22 @@ func UpdatePatchStatus(p *patch.Patch, versionStatus, buildVariant string) error
 		}
 	}
 
+	isDone, parentPatch, err := p.GetFamilyInformation()
```
hadjri, line 1234: can we rename to something like patchIsDone to be more explicit?

rest/data/subscription.go
```diff
@@ -105,18 +75,20 @@ func SaveSubscriptions(owner string, subscriptions []restModel.APISubscription,
...
 
-func getVersionChildren(versionId string) ([]string, error) {
-	patchDoc, err := patch.FindOne(patch.ByVersion(versionId))
-	if err != nil {
-		return nil, errors.Wrapf(err, "finding patch for version '%s'", versionId)
-	}
-	if patchDoc == nil {
-		return nil, errors.Wrapf(err, "patch for version '%s' not found", versionId)
+func triggerToFamilyTrigger(t string) string {
+	var trigger string
```
hadjri, line 82: We can use `default` keyword after all of the cases and have a return statement on each case here instead of declaring a var and returning after the switch is complete

trigger/patch.go
```diff
@@ -103,79 +111,44 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
...
-		// now that the children are done, the parent can be considered done.
-		t.patch = parentPatch
-	}
-	return isReady, nil
-}
+		if t.patch.IsParent() || (t.patch.IsChild() && subType == event.WaitOnChild) {
+			// get the children or siblings to wait on
+			childrenOrSiblings, parentPatch, err := t.patch.GetPatchFamily()
+			if err != nil {
+				return nil, errors.Wrap(err, "error getting child or sibling patches")
```
hadjri, line 125: nit: "getting child or sibling patches"

trigger/patch.go
```diff
@@ -103,79 +111,44 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
...
+			}
 
-func checkPatchStatus(p *patch.Patch) (bool, *patch.Patch, bool, error) {
-	isReady := false
-	childrenOrSiblings, parentPatch, err := p.GetPatchFamily()
-	if err != nil {
-		return isReady, parentPatch, false, errors.Wrap(err, "getting child or sibling patches")
-	}
+			// make sure the parent is done, if not, wait for the parent
+			if t.patch.IsChild() {
```
hadjri, line 129: nit: && instead of nested if

trigger/patch.go
```diff
@@ -103,79 +111,44 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
...
+				if !evergreen.IsFinishedPatchStatus(parentPatch.Status) {
+					return nil, nil
+				}
+			}
+			childrenStatus, err := getChildrenOrSiblingsReadiness(childrenOrSiblings)
+			if err != nil {
+				return nil, errors.Wrap(err, "error getting child or sibling information")
```
hadjri, line 136: same comment

trigger/patch.go
```diff
@@ -389,3 +365,36 @@ func (t *patchTriggers) getGithubContext() (string, error) {
...
+		return nil, nil
+	}
+	if t.event.EventType != event.VersionChildrenCompletion {
+		return nil, nil
+	}
+	return t.generate(sub)
+}
+
+func (t *patchTriggers) patchFamilySuccess(sub *event.Subscription) (*notification.Notification, error) {
+	if t.data.Status != evergreen.PatchSucceeded {
```
hadjri, line 380: We can consolidate these 2 cases since they'll both have us return nil, nil

trigger/version.go
```diff
@@ -247,6 +235,54 @@ func (t *versionTriggers) versionExceedsDuration(sub *event.Subscription) (*noti
...
+		return nil, nil
+	}
+	if t.event.EventType != event.VersionChildrenCompletion {
+		return nil, nil
+	}
+	return t.generate(sub, "")
+}
+
+func (t *versionTriggers) versionFamilyFailure(sub *event.Subscription) (*notification.Notification, error) {
+	if t.data.Status != evergreen.VersionFailed {
```
hadjri, line 249: Same comment

trigger/version.go
```diff
@@ -247,6 +235,54 @@ func (t *versionTriggers) versionExceedsDuration(sub *event.Subscription) (*noti
...
+	}
+	if skipNotification {
+		return nil, nil
+	}
+
+	return t.generate(sub, "")
+}
+
+func (t *versionTriggers) versionFamilySuccess(sub *event.Subscription) (*notification.Notification, error) {
+	if t.data.Status != evergreen.VersionSucceeded {
```
hadjri, line 276: same comment

hadjri (changes requested):

rest/data/subscription.go
malikchaya2: done

trigger/patch.go
malikchaya2: done

trigger/patch.go
malikchaya2: done

trigger/version.go
malikchaya2: done

trigger/version.go
malikchaya2: done

hadjri (approved):

malikchaya2:
The Spruce tests pass when they work in conjunction with this: https://github.com/evergreen-ci/spruce/pull/1536

PR #6091, opened 2022-11-28
EVG-18391 use latest commit version for periodic build creation

ablack12:
[EVG-18391](https://jira.mongodb.org/browse/EVG-18391)

### Description 
This periodic build was defined with an error, so we created a stub version.
https://evergreen.mongodb.com/version/637d2a40a4cf472bcc36d583
However, the stub version ends up having no revision since we don't add it to the metadata. (This is also why https://jira.mongodb.org/browse/EVG-18390 occurred, because we use the revision for the link.) 

Because we just get the latest version (instead of the latest version with a valid config), we end up getting the broken periodic build, which has an empty revision, and so when we call GetGithubFile we get the master branch yaml instead of the v6.1 yaml. 

### Testing 
Not tested but reasoned about.

### DIFF
--- a/units/periodic_builds.go
+++ b/units/periodic_builds.go
@@ -154,7 +154,7 @@ func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.Peri
 		return "", errors.Wrap(err, "getting GitHub OAuth token")
 	}
 
-	mostRecentVersion, err := model.VersionFindOne(model.VersionByMostRecentSystemRequester(j.ProjectID))
+	mostRecentVersion, err := model.FindVersionByLastKnownGoodConfig(j.ProjectID, -1)
 	if err != nil {
 		return "", errors.Wrapf(err, "finding most recent version for project '%s'", j.ProjectID)
 	}
--- a/units/periodic_builds.go
+++ b/units/periodic_builds.go
@@ -96,7 +96,15 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
 		}))
 	}()
 
-	versionID, versionError := j.addVersion(ctx, *definition)
+	mostRecentVersion, err := model.FindVersionByLastKnownGoodConfig(j.ProjectID, -1)
+	if err != nil {
+		j.AddError(errors.Wrapf(err, "finding most recent version for project '%s'", j.ProjectID))
+	}
+	if mostRecentVersion == nil {
+		j.AddError(errors.Errorf("no recent version found for project '%s'", j.ProjectID))
+	}
+
+	versionID, versionError := j.addVersion(ctx, *definition, mostRecentVersion.Revision)
 	if versionError != nil {
 		// if the version fails to be added, create a stub version and
 		// log an event so users can get notified when notifications are configured
@@ -105,6 +113,9 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
 			Message:         definition.Message,
 			PeriodicBuildID: definition.ID,
 			Alias:           definition.Alias,
+			Revision: model.Revision{
+				Revision: mostRecentVersion.Revision,
+			},
 		}
 		stubVersion, dbErr := repotracker.ShellVersionFromRevision(ctx, j.project, metadata)
 		if dbErr != nil {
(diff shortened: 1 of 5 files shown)


units/periodic_builds.go
```diff
@@ -148,20 +159,13 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
...
 }
 
-func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.PeriodicBuildDefinition) (string, error) {
+func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.PeriodicBuildDefinition, mostRecentRevision string) (string, error) {
 	token, err := j.env.Settings().GetGithubOauthToken()
 	if err != nil {
 		return "", errors.Wrap(err, "getting GitHub OAuth token")
 	}
 
-	mostRecentVersion, err := model.VersionFindOne(model.VersionByMostRecentSystemRequester(j.ProjectID))
```
malikchaya2, line 157: VersionByMostRecentSystemRequester filters VersionRequester on 

 `VersionRequesterKey: bson.M{
        "$in": evergreen.SystemVersionRequesterTypes,
    }`
  and FindVersionByLastKnownGoodConfig switches it to  `evergreen.RepotrackerVersionRequester`. 
  
  Is that maybe part of why it was broken before?

units/periodic_builds.go
```diff
@@ -96,7 +96,15 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
 		}))
 	}()
 
-	versionID, versionError := j.addVersion(ctx, *definition)
```
malikchaya2, line 99: (comment placement is random)

Since this is hard to test in staging, would it be possible to come up with a plan for monitoring this after the deploy to make sure that it's working as expected?

malikchaya2 (changes requested):

units/periodic_builds.go
ablack12: We'll know if this fix works if Luke's next periodic build comes up correctly.

units/periodic_builds.go
ablack12: It sort of is! Really the issue is that we VersionByMostRecentSystemRequester doesn't consider if there's a stub version (i.e. a broken version), and at least for periodic builds, stub versions didn't have revisions, which is what caused the bug. There's also really no reason to consider all system requester types, since they use the configs from the commit versions (and also because that's really what we say the periodic build is testing, just the most recent commit).

I'm actually also realizing though that using FindVersionByLastKnownGoodConfig means we'd skip the most recent commit if it's got a broken config, which isn't quite what we promise, so I'll tweak this to just do a FindOne on the latest commit. (Verified that [commit stub versions will always have a revision](https://github.com/evergreen-ci/evergreen/blob/6c5e326eb294959ed5d9c99d11b3cb871d799a8c/repotracker/repotracker.go#L328)).

units/periodic_builds.go
```diff
@@ -96,25 +96,22 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
 		}))
 	}()
 
-	mostRecentVersion, err := model.FindVersionByLastKnownGoodConfig(j.ProjectID, -1)
+	mostRecentRevision, err := model.FindLatestRevisionForProject(j.ProjectID)
```
malikchaya2, line 99: nit: Should we error if mostRecentRevision = "" (since we expect it to always have one)?

malikchaya2 (approved):
approved with nit

units/periodic_builds.go
```diff
@@ -96,15 +96,23 @@ func (j *periodicBuildJob) Run(ctx context.Context) {
...
+
+	if versionErr != nil {
+		// If the version fails to be added, create a stub version and
 		// log an event so users can get notified when notifications are configured
 		metadata := model.VersionMetadata{
 			IsAdHoc:         true,
 			Message:         definition.Message,
 			PeriodicBuildID: definition.ID,
 			Alias:           definition.Alias,
+			Revision: model.Revision{
```
hadjri, line 113: If `mostRecentRevision` is empty, the Version created will end up having an empty revision as well. I assume that's fine for stub versions?

units/periodic_builds.go
ablack12: This happens in FindLatestRevisionForProject

units/periodic_builds.go
```diff
@@ -193,7 +194,7 @@ func (j *periodicBuildJob) addVersion(ctx context.Context, definition model.Peri
 		PeriodicBuildID: definition.ID,
 		Alias:           definition.Alias,
 		Revision: model.Revision{
-			Revision: mostRecentVersion.Revision,
+			Revision: mostRecentRevision,
```
hadjri, line 197: Since it's possible for `mostRecentRevision` to be empty here, should we retrieve the commit hash from the master branch since that is what `GetGithubFile` would do? Otherwise I think the created version will have an empty revision

hadjri (changes requested):

units/periodic_builds.go
ablack12: Shouldn't be relevant now that we return if mostRecentRevision is empty

hadjri (approved):

PR #6094, opened 2022-11-29
EVG-18359 always set IN_EVERGREEN for subprocesses

ybrill:
[EVG-18359](https://jira.mongodb.org/browse/EVG-18359)

### Description 
Set an environment variable to help processes run with shell.exec and subprocess.exec differentiate between when they're running under evergreen vs. another CI system.

### Testing 
I created [a task in staging](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu1604_echo_env_patch_91d8ccfe1f069a97bbe9ffd11f38742b4053b894_63867f7cb23736629a824c7e_22_11_29_21_54_13/logs?execution=0) that echos the contents of the variable. It was set.

### DIFF
--- a/agent/command/exec.go
+++ b/agent/command/exec.go
@@ -199,6 +199,9 @@ func defaultAndApplyExpansionsToEnv(env map[string]string, opts modifyEnvOptions
 	if _, ok := env["CI"]; !ok {
 		env["CI"] = "true"
 	}
+	if _, ok := env[agentutil.MarkerInEvergreen]; !ok {
+		env[agentutil.MarkerInEvergreen] = "true"
+	}
 
 	return env
 }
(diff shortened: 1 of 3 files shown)


Kimchelly (commented):
It seems like checking for `IN_EVERGREEN` would be equivalent to checking for `EVR_TASK_ID`. Is this just a convenience to check a boolean instead of a string? As far as I'm aware, `IN_EVERGREEN` was originally meant to be set specifically for the purpose of [this one self-test](https://github.com/evergreen-ci/evergreen/blob/9be67559962ca25bc1af3ab00c4bc9c7f3ffe666/agent/util/subtree_unix_test.go#L34) and wasn't actually meaningful for real calls to subprocess.exec/shell.exec.

If it's for convenience, do we have to update the one test?

ybrill:
Ah that's a good point.
Let me follow up with them to see if `EVR_TASK_ID` will work for them.

My original rationale for `IN_EVERGREEN` was to give cloud something to set to flag to us that a process should be cleaned up (for [EVG-12948](https://jira.mongodb.org/browse/EVG-12948))

ybrill:
Confirmed `EVR_TASK_ID` will work for them.
Thanks for pointing it out!

PR #6095, opened 2022-11-29
EVG-17375 Create manifest upon patch finalization

hadjri:

### Description 
Dynamically fetching the latest revision for every module upon patch creation has the potential to break the strict reproducability of patches since a change in a module branch can affect the outcome of a patch even if the patch itself contains no changes from the previous patch.

However at the same time there have been instances where users have run patches expecting a module to be running off of the most recent revision in the tracked branch but instead was stuck at the revision that existed at the time of the base version of the patch.

Introduced `auto_update_modules` flag which is specifiable in YAML and if set, will make created patches fetch the latest revisions for each module by creating their own manifest when they are created. Default value is false.
### Testing 
Updated the FinalizePatch test to check that only setting `auto_update_modules` will result in a manifest getting created upon patch creation. Will update docs once an agreed implementation is approved.

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -498,20 +498,9 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 	if err != nil {
 		return nil, errors.Wrap(err, "constructing manifest")
 	}
-	if mfst != nil {
-		baseManifest, err := manifest.FindFromVersion(patchVersion.Id, patchVersion.Identifier, patchVersion.Revision, patchVersion.Requester)
-		if err != nil {
-			return nil, errors.Wrap(err, "constructing base manifest")
-		}
-		if baseManifest != nil {
-			modulesToCopy := baseManifest.Modules
-			for key := range modulesToCopy {
-				if _, ok := mfst.Modules[key]; ok {
-					modulesToCopy[key] = mfst.Modules[key]
-				}
-			}
-			mfst.Modules = modulesToCopy
-		}
+	err = mergeWithBaseManifest(mfst, patchVersion)
+	if err != nil {
+		return nil, errors.Wrap(err, "merging with base manifest")
 	}
 
 	tasks := TaskVariantPairs{}
@@ -674,6 +663,27 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 	}
 
 	return patchVersion, nil
+}
+
+// mergeWithBaseManifest merges the manifest from the patch with the manifest from the
+// most recent base commit.
+func mergeWithBaseManifest(mfst *manifest.Manifest, patchVersion *Version) error {
+	if mfst != nil {
+		baseManifest, err := manifest.FindFromVersion(patchVersion.Id, patchVersion.Identifier, patchVersion.Revision, patchVersion.Requester)
+		if err != nil {
+			return errors.Wrap(err, "constructing base manifest")
+		}
+		if baseManifest != nil {
+			modulesToCopy := baseManifest.Modules
+			for key := range modulesToCopy {
+				if _, ok := mfst.Modules[key]; ok {
+					modulesToCopy[key] = mfst.Modules[key]
+				}
+			}
+			mfst.Modules = modulesToCopy
+		}
+	}
+	return nil
 }
 
 // getFullPatchParams will retrieve a merged list of parameters defined on the patch alias (if any)
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -3,7 +3,6 @@ package model
 import (
 	"context"
 	"fmt"
-	"github.com/evergreen-ci/evergreen/model/manifest"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -18,6 +17,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/commitqueue"
 	"github.com/evergreen-ci/evergreen/model/distro"
 	"github.com/evergreen-ci/evergreen/model/event"
+	"github.com/evergreen-ci/evergreen/model/manifest"
 	"github.com/evergreen-ci/evergreen/model/patch"
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/model/user"
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -492,9 +492,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 	}
 	intermediateProject.CreateTime = patchVersion.CreateTime
 
-	autoUpdateModules := project.GetAutoUpdateModules()
-
-	mfst, err := constructManifest(patchVersion, projectRef, autoUpdateModules, settings)
+	mfst, err := constructManifest(patchVersion, projectRef, project.Modules, settings)
 	if err != nil {
 		return nil, errors.Wrap(err, "constructing manifest")
 	}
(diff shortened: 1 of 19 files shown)


model/project_parser.go
```diff
@@ -82,6 +82,7 @@ type ParserProject struct {
 	EarlyTermination   *YAMLCommandSet            `yaml:"early_termination,omitempty" bson:"early_termination,omitempty"`
 	CallbackTimeout    *int                       `yaml:"callback_timeout_secs,omitempty" bson:"callback_timeout_secs,omitempty"`
 	Modules            []Module                   `yaml:"modules,omitempty" bson:"modules,omitempty"`
+	AutoUpdateModules  *bool                      `yaml:"auto_update_modules,omitempty" bson:"auto_update_modules,omitempty"`
```
ablack12, line 85: This is slightly complicated because really this does seem like it should only be relevant to PR patches 😅 otherwise to test a module at a specific revision you'd need to go and remove this from the yaml to do that (although maybe that's fine since there's a workaround). An alternative idea could be to have this be part of the PR testing section on the project page, configurable that way? (It would require a spruce PR but they could probably do it pretty fast, especially since this impacts spruce in particular) That could make it very clear that this just impacts PR testing.

ablack12 (changes requested):

model/project_parser.go
hadjri: I actually think it'd be useful to have this configurable for `evergreen patch` as well (@khelif96 mentions it in the JIRA comments) - if a module branch has commits ahead of even the most recent project commit then the currently the only way to run a patch with the updated module if the project is behind is to add an empty commit to the project.

Maybe we can allow CLI and PR patches to set this value individually? If we're looking to over time move configurations from the UI to YAML it might be preferable to put new configurations moving forward on the YAML side since this might be something a non-admin developer might want to toggle back and forth. Maybe we can move these flags into the `modules` YAML structure for more simplicity?

ablack12: Hmmm okay I see. I think maybe making this configurable at the module yaml level does make sense then.

ablack12: (Adding a comment with the review button so it shows me as having responded)

ablack12: (Adding a comment with the review button so it shows me as having responded)

ablack12 (changes requested):

model/project_parser.go
hadjri: Sounds good, per our offline talk I've moved `auto_update` to the module level, and now for patch finalization we fetch the base revision manifest, update it for the modules with `auto_update` , and then re-insert the manifest as opposed to constructing an entirely new manifest for a patch.

model/patch_lifecycle.go
```diff
@@ -3,6 +3,7 @@ package model
 import (
 	"context"
 	"fmt"
+	"github.com/evergreen-ci/evergreen/model/manifest"
```
ablack12, line 6: move down

model/patch_lifecycle_test.go
```diff
@@ -298,11 +311,14 @@ modules:
 			version, err := FinalizePatch(ctx, configPatch, evergreen.PatchVersionRequester, token)
 			require.NoError(t, err)
 			assert.NotNil(t, version)
-			// ensure that the manifest was created
+			// ensure that the manifest was created and that auto_update worked for
```
ablack12, line 314: nit: capitalize

model/project.go
```diff
@@ -1430,6 +1430,16 @@ func (p *Project) GetModuleByName(name string) (*Module, error) {
 	return nil, errors.New("no such module on this project")
 }
 
+func (p *Project) GetAutoUpdateModules() ModuleList {
```
ablack12, line 1433: nit: function comment (I know we don't have these for all exported functions but we should try to add them to keep up with the style guide

model/patch_lifecycle.go
```diff
@@ -491,10 +492,27 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
 	intermediateProject.CreateTime = patchVersion.CreateTime
 
-	manifest, err := constructManifest(patchVersion, project, projectRef, settings)
+	autoUpdateModules := project.GetAutoUpdateModules()
+
+	mfst, err := constructManifest(patchVersion, projectRef, autoUpdateModules, settings)
 	if err != nil {
 		return nil, errors.Wrap(err, "constructing manifest")
 	}
+	if mfst != nil {
```
ablack12, line 501: Should we also check that len(autoUpdateModules) != len(project.Modules) ?

model/patch_lifecycle.go
```diff
@@ -491,10 +492,27 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+		baseManifest, err := manifest.FindFromVersion(patchVersion.Id, patchVersion.Identifier, patchVersion.Revision, patchVersion.Requester)
```
ablack12, line 502: Can this logic be moved into constructManifest?

ablack12 (changes requested):

model/patch_lifecycle.go
ablack12: oh I see that constructManifest is also used by CreateManifest but I think it could still be a paramter for constructManifest or it's own helper function.

model/project.go
hadjri: 🤦‍♂️ sure thing!

model/patch_lifecycle.go
hadjri: I don't think so because we should be able to specify a subset of the modules to auto-update

model/patch_lifecycle.go
```diff
@@ -676,6 +665,27 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+		if err != nil {
+			return errors.Wrap(err, "constructing base manifest")
+		}
+		if baseManifest != nil {
+			modulesToCopy := baseManifest.Modules
+			for key := range modulesToCopy {
+				if _, ok := mfst.Modules[key]; ok {
+					modulesToCopy[key] = mfst.Modules[key]
```
ablack12, line 680: I might be forgetting what's going on now but I thought that we'd want to use the base manifest only for things that aren't auto-update (i.e. exists in the manifest), so !ok

model/version.go
```diff
@@ -593,6 +597,90 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 	return res, nil
 }
 
+// constructManifest will construct a manifest from the given project and version.
+func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList, settings *evergreen.Settings) (*manifest.Manifest, error) {
```
ablack12, line 601: Sorry to harp on the construction of this function too much but I think constructManifest and mergeWithBaseManifest can just be combined.
Instead of passing in just the auto-update modules to constructManifest, could we just pass in all of them? And then at 622 if we're looking at a patch and it's not auto-update, we use the base manifest (if I'm remembering the behavior correctly), otherwise we do the steps that exist here. 
Having the two helper functions that require very specific input (and rely on each other's specific output) seems like it'll make it bug prone in the future so it seems worth refactoring now if possible.

ablack12 (changes requested):

model/patch_lifecycle.go
hadjri: That's what should be happening here - we initialize `modulesToCopy` as the base manifest modules and for anything we find in `mfst` (which contains only auto-update modules) we replace the value in `modulesToCopy`.

hadjri: Not relevant anymore since I'm deleting this function though

ablack12 (approved):

PR #6098, opened 2022-11-30
EVG-17305 Deactivate BVs with no active tasks when generating tasks 

malikchaya2:

### Description 
This solves the bug we were seeing where patches remained in the "running" status despite having finished running. 

### Testing 
A patch with the same setup [before](https://spruce-staging.corp.mongodb.com/version/6386871297b1d36bb9fbf1e0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [after](https://spruce-staging.corp.mongodb.com/version/6387b174b237365ae87d5f9d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) the changes.  The problem is more visible in the legacy UI where you can see the build variants with inactive tasks.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -550,7 +550,7 @@ func RefreshTasksCache(buildId string) error {
 
 // addTasksToBuild creates/activates the tasks for the given build of a project
 func addTasksToBuild(ctx context.Context, creationInfo TaskCreationInfo) (*build.Build, task.Tasks, error) {
-	// find the build variant for this project/build
+	// Find the build variant for this project/build
 	creationInfo.BuildVariant = creationInfo.Project.FindBuildVariant(creationInfo.Build.BuildVariant)
 	if creationInfo.BuildVariant == nil {
 		return nil, nil, errors.Errorf("finding build '%s' in project file '%s'",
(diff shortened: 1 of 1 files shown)


model/lifecycle.go
```diff
@@ -550,7 +550,7 @@ func RefreshTasksCache(buildId string) error {
 
 // addTasksToBuild creates/activates the tasks for the given build of a project
 func addTasksToBuild(ctx context.Context, creationInfo TaskCreationInfo) (*build.Build, task.Tasks, error) {
-	// Find the build variant for this project/build
+	// find the build variant for this project/build
```
ablack12, line 553: can we re-capitalize? (this is the style guide preference)

model/lifecycle.go
```diff
@@ -709,14 +709,19 @@ func CreateBuildFromVersionNoInsert(creationInfo TaskCreationInfo) (*build.Build
 
 	// create task caches for all of the tasks, and place them into the build
 	tasks := []task.Task{}
+	containsActivatedTask := false
+
 	for _, taskP := range tasksForBuild {
 		if taskP.IsPartOfDisplay() {
 			continue // don't add execution parts of display tasks to the UI cache
 		}
+		if taskP.Activated {
```
ablack12, line 718: small thing: can we move this above IsPartOfDisplay (I can't think of a reason we'd need to but just in case)

Also I wonder if it'd make sense to also move the IsGithubCheck into this loop (we can't break but that doesn't matter since we're already not breaking because of this loop)

ablack12 (approved):
LGTM with nit

model/lifecycle.go
malikchaya2: I moved it up. 

And you're right, I don't see why IsGithubCheck can't be part of this. I moved it in.

PR #6099, opened 2022-11-30
EVG-17275 use yaml when yaml is expected

ablack12:

### Description 
use yaml because the header is yaml

### Testing 
will run the set-module command on something


khelif96 (approved):

ablack12:
Verified that this process works on staging and not on prod 🤞

PR #6106, opened 2022-12-02
EVG-17483 Remove deprecated code

malikchaya2:
[EVG-17483](https://jira.mongodb.org/browse/EVG-17483)

### Description 
Remove code that is no longer needed. 

### Testing 
Made sure that the app can still start up with the icon_url removed.


Kimchelly (commented):
I think [this](https://github.com/evergreen-ci/evergreen/blob/1fc3954c0c1e28d90790c9b020157d9af03b85aa/rest/route/admin_test.go#L199) need to be deleted.

~I don't think this is exactly relevant to this PR, but does the Slack app not need an icon? It seems like staging notifications have a lovely tree icon, while prod notifications don't.~ Ignore this, I forgot that there's a different Evergreen bot for prod now.

image
vs.
image

Kimchelly (approved):
LGTM mod remove the last usage of `SlackAppDisabled` from the REST route test.

bynn (approved):

PR #6107, opened 2022-12-02
EVG-16746 Disable execution tasks with display task

bynn:
[EVG-16746](https://jira.mongodb.org/browse/EVG-16746)

### Description 
execution tasks are now updated with display task

### Testing 
staging https://spruce-staging.corp.mongodb.com/version/638a617097b1d375e8784289/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

unit tests

### DIFF
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -2185,18 +2185,19 @@ func TestDeactivateTasks(t *testing.T) {
 		{Id: "t4", DependsOn: []Dependency{{TaskId: "t2"}}, Activated: true},
 		{Id: "t5", DependsOn: []Dependency{{TaskId: "t4"}}, Activated: true},
 		{Id: "t6", Activated: true},
+		{Id: "t7", DependsOn: []Dependency{{TaskId: "t6"}}, Activated: true},
 	}
 	for _, task := range tasks {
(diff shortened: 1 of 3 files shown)


model/task/task_test.go
```diff
@@ -2178,24 +2178,25 @@ func TestDeactivateTasks(t *testing.T) {
 	require.NoError(t, db.ClearCollections(Collection, event.EventCollection))
 
 	tasks := []Task{
-		{Id: "t0"},
+		{Id: "t0", DisplayOnly: true, ExecutionTasks: []string{"t6"}},
```
hadjri, line 2181: Can we add a case where we check that a task that depends on the execution task `t6` gets deactivated?

hadjri (changes requested):
Can we also add display task cases to `[TestTaskSetPriority](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle_test.go#L29)`?

model/task/task_test.go
bynn: added

bynn:
added tests for set task priority

hadjri (approved):

PR #6109, opened 2022-12-02
EVG-17046: fix race in TestMergeMultiplePatches

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17046

### Description 
The comment explains most of the rationale, but basically, this test was failing because git clone could sometimes be slow and the existing logic didn't handle possibility that the context could error (e.g. due to a timeout). If the context errors, `(jasper.Command).Run` exits early while the concurrently-running git command is still potentially writing to the buffer. Reading a `bytes.Buffer` via `String` is not safe when it's still being written by the git command.

* Use thread-safe wrapper for bytes.Buffer.
* Increase context timeout on test to reduce the possibility of the git commands timing out.

### Testing 
I updated the existing test slightly and ran this test a bunch of times on my computer with the race detector enabled.

### DIFF
--- a/agent/command/git_test.go
+++ b/agent/command/git_test.go
@@ -26,10 +26,11 @@ import (
 	"github.com/evergreen-ci/evergreen/testutil"
 	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/evergreen/util"
-	"github.com/mongodb/grip"
+	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip/level"
 	"github.com/mongodb/grip/send"
 	"github.com/mongodb/jasper"
+	"github.com/pkg/errors"
 	. "github.com/smartystreets/goconvey/convey"
 	"github.com/smartystreets/goconvey/convey/reporting"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


agent/command/git_test.go
```diff
@@ -989,9 +990,13 @@ index edc0c34..8e82862 100644
 			s.NotNil(pluginCmds)
 			pluginCmds[0].SetJasperManager(s.jasper)
 			err = pluginCmds[0].Execute(ctx, s.comm, logger, conf)
-			// this command will error because it'll apply the same patch twice. We are just testing that
-			// there was an attempt to apply the patch the second time
-			grip.Debug(err)
+			// Running the git commands takes time, so it could hit the test's
+			// context timeout if it's slow. Make sure that the error isn't due
+			// to a timeout.
+			s.False(utility.IsContextError(errors.Cause(err)))
```
ablack12, line 996: Does this mean that the ubuntu test could become flaky?

Kimchelly: Yes, but practically, it shouldn't happen that often, unless GitHub is unavailable when the test runs. This seems to primarily affect race detector because enabling the race detector slows down the test. Increasing the timeout should mitigate the possibility of failures due to slow cloning.

ablack12 (approved):

Kimchelly:
evergreen retry

Kimchelly:
I dunno what's going on with the webhook bot, but Evergreen tried creating [a patch with a different bot user](https://spruce.mongodb.com/version/638e1042d6d80a2f599636ae/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). I opened [a ticket](https://jira.mongodb.org/browse/EVG-18448) to figure out why it did this.

PR #6110, opened 2022-12-02
EVG-16853 Don't allow restarting non-finished aborted tasks

hadjri:
[EVG-16853](https://jira.mongodb.org/browse/EVG-16853)

### Description 
When "Abort" is hit in the UI, it marks the task as `abort: true`, but does not fail the task immediately. It is the agent that fails the task once it sees it was aborted, and this usually takes a few seconds. Hence, if you click "Abort" and then immediately click "Restart", the agent might not have had time to fail the task that had its aborted field set and we get the error that is screenshotted on the ticket.

Change introduced will cause the "Restart" button to be greyed out for aborted tasks until the agent actually finishes the task (which you'd be able to tell from the logs as well).
### Testing 
Tested on local evergreen

### DIFF
--- a/graphql/tests/task/results.json
+++ b/graphql/tests/task/results.json
@@ -268,7 +268,7 @@
             "canRestart": false
           },
           "abortedTask": {
-            "canRestart": true
+            "canRestart": false
           }
         }
       }
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -431,13 +431,8 @@ func canRestartTask(t *task.Task) bool {
 	}
 	// It is possible to restart blocked display tasks. Later tasks in a display task could be blocked on
 	// earlier tasks in the display task, in which case restarting the entire display task may unblock them.
-	if t.DisplayStatus == evergreen.TaskStatusBlocked && t.DisplayOnly {
-		return true
-	}
-	if !utility.StringSliceContains(evergreen.TaskUncompletedStatuses, t.Status) {
-		return true
-	}
-	return t.Aborted
+	return (t.DisplayStatus == evergreen.TaskStatusBlocked && t.DisplayOnly) ||
+		!utility.StringSliceContains(evergreen.TaskUncompletedStatuses, t.Status)
 }
 
 func canScheduleTask(t *task.Task) bool {
--- a/graphql/util_test.go
+++ b/graphql/util_test.go
@@ -135,7 +135,7 @@ func TestCanRestartTask(t *testing.T) {
 		Aborted:       true,
 	}
 	canRestart = canRestartTask(abortedTask)
-	assert.Equal(t, canRestart, true)
+	assert.Equal(t, canRestart, false)
 }
 
 func TestCanScheduleTask(t *testing.T) {


Kimchelly (commented):
I don't have strong opinions here, but is the workflow of abort button -> restart button a meaningfully different workflow from the restart modal with the checkbox to "abort in progress tasks"? It seems like we would want to keep the behavior the same between those two methods of abort + restart since they're logically the same operations.

hadjri:
> I don't have strong opinions here, but is the workflow of abort button -> restart button a meaningfully different workflow from the restart modal with the checkbox to "abort in progress tasks"? It seems like we would want to keep the behavior the same between those two methods of abort + restart since they're logically the same operations.

The difference is that hitting "abort in progress tasks" when restarting a version will set [ResetWhenFinished](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L337) for the running tasks in the version, so when the agent eventually fails these tasks it will reset them automatically. This would be analogous to having the restart button available for a task while it is still running (which isn't currently possible in the UI - it's greyed out). There also isn't an abort button for a version in the UI, you can only abort/restart in tandem.

hadjri:
> > I don't have strong opinions here, but is the workflow of abort button -> restart button a meaningfully different workflow from the restart modal with the checkbox to "abort in progress tasks"? It seems like we would want to keep the behavior the same between those two methods of abort + restart since they're logically the same operations.
> 
> The difference is that hitting "abort in progress tasks" when restarting a version will set [ResetWhenFinished](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L337) for the running tasks in the version, so when the agent eventually fails these tasks it will reset them automatically. This would be analogous to having the restart button available for a task while it is still running (which isn't currently possible in the UI - it's greyed out). There also isn't an abort button for a version in the UI, you can only abort/restart in tandem.

Maybe a transitory "Aborting" UI badge would be helpful for tasks in a state where `abort: true status: [not finished]` is true? Where it'd be set for the few seconds between marking abort and getting picked up by the agent.

khelif96 (approved):
This behavior lgtm

ablack12:
I'd prefer not to set a status because that affects a lot of things in the system. If we want to display aborted on the UI we could do it by adding to the version metadata (which I think we already do for aborted tasks that fail but i'm not sure?). 

I'm wondering if we try restarting an aborted task if we can just set ResetWhenFinished if it's not finished aborting (we may then need to make sure that the agent refreshes its task document before checking that field in case there's a race).

ablack12:
> I don't have strong opinions here, but is the workflow of abort button -> restart button a meaningfully different workflow from the restart modal with the checkbox to "abort in progress tasks"

Actually it looks like this button isn't allowed to be clicked on individual tasks which I think would be the issue here.

Kimchelly (approved):
LGTM the change mod fixing the GraphQL tests. I think displaying that a task is currently aborting would also be more user-friendly, but it's not necessary for this.

PR #6111, opened 2022-12-06
EVG-18413 Do not activate generated tasks that depend on explicitly inactive variants

hadjri:
[EVG-18413](https://jira.mongodb.org/browse/EVG-18413)

### Description 
Currently, generated dependencies of generated tasks that don't have their variant defined in the generator JSON get activated by default, because the way Evergreen determines the activation of a generated task is from the generated JSON, i.e. setting `"activate": false` on a variant.  If Evergreen can't find the variant, it assumes the variant will be active.

Change has been made to determine which generated dependency tasks do not have their variant defined in the generated JSON, and if the generated tasks that depend on them are all deactivated, then it is safe / necessary to also deactivate this newly generated variant.
### Testing 
Tested in staging and confirmed that a generated task that isn't set anywhere in the original generated JSON does not get activated by default.

### DIFF
--- a/model/dependencies.go
+++ b/model/dependencies.go
@@ -6,10 +6,10 @@ import (
 )
 
 type dependencyIncluder struct {
-	Project           *Project
-	requester         string
-	included          map[TVPair]bool
-	generatedVariants map[string]bool
+	Project                         *Project
+	requester                       string
+	included                        map[TVPair]bool
+	deactivateGeneratedVariantTasks map[string]bool
 }
 
 // IncludeDependencies takes a project and a slice of variant/task pairs names
@@ -28,7 +28,7 @@ func IncludeDependencies(project *Project, tvpairs []TVPair, requester string, a
 // of variants and tasks are returned.
 func (di *dependencyIncluder) include(initialDeps []TVPair, activationInfo *specificActivationInfo) ([]TVPair, error) {
 	di.included = map[TVPair]bool{}
-	di.generatedVariants = map[string]bool{}
+	di.deactivateGeneratedVariantTasks = map[string]bool{}
 	warnings := grip.NewBasicCatcher()
 	// handle each pairing, recursively adding and pruning based
 	// on the task's dependencies
@@ -43,8 +43,10 @@ func (di *dependencyIncluder) include(initialDeps []TVPair, activationInfo *spec
 			outPairs = append(outPairs, pair)
 		}
 	}
-	// some comment
-	for variant, addToActivationInfo := range di.generatedVariants {
+	// we deactivate variants that do not have specific activation set in the
+	// generated project but have been generated as dependencies of other
+	// explicitly inactive variants
+	for variant, addToActivationInfo := range di.deactivateGeneratedVariantTasks {
 		if addToActivationInfo {
 			activationInfo.activationVariants = append(activationInfo.activationVariants, variant)
 		}
@@ -96,17 +98,22 @@ func (di *dependencyIncluder) handle(pair TVPair, activationInfo *specificActiva
 	di.included[pair] = true
 
 	// queue up all dependencies for recursive inclusion
-	// if the pair is inactive, include the deps in the inactive list
-	// if a generated dependency exists in a build variant that is not
-	// defined in the generated project, it will be added to the inactive list.
 	deps := di.expandDependencies(pair, bvt.DependsOn)
-	setActivation := activationInfo != nil && activationInfo.taskOrVariantHasSpecificActivation(pair.Variant, pair.TaskName)
+
+	// If this function is invoked from generate.tasks, calculate all variants
+	// that need to be included as dependencies, but are not in the generated project.
+	// If all the task / variant pairs that spawn these dependencies are inactive, we
+	// also mark this newly generated variant as inactive.
+	pairSpecifiesActivation := activationInfo != nil && activationInfo.taskOrVariantHasSpecificActivation(pair.Variant, pair.TaskName)
 	for _, dep := range deps {
 		if activationInfo != nil && !activationInfo.variantExistsInGeneratedProject(dep.Variant) {
-			if _, ok := di.generatedVariants[dep.Variant]; !ok {
-				di.generatedVariants[dep.Variant] = setActivation
-			} else if !setActivation {
-				di.generatedVariants[dep.Variant] = setActivation
+			// If the new variant has not yet been added to deactivateGeneratedVariantTasks, or if the
+			// original pair needs to be active, we update deactivateGeneratedVariantTasks.
+			// We ultimately will only deactivate new variants where deactivateGeneratedVariantTasks[variant] = true.
+			// If deactivateGeneratedVariantTasks[variant] = false it signifies that there was at least
+			// one pair that depends on this new variant being active - so we cannot deactivate it.
+			if _, ok := di.deactivateGeneratedVariantTasks[dep.Variant]; !ok || !pairSpecifiesActivation {
+				di.deactivateGeneratedVariantTasks[dep.Variant] = pairSpecifiesActivation
 			}
 		}
 		ok, err := di.handle(dep, activationInfo)
--- a/model/dependencies.go
+++ b/model/dependencies.go
@@ -6,9 +6,10 @@ import (
 )
 
 type dependencyIncluder struct {
-	Project   *Project
-	requester string
-	included  map[TVPair]bool
+	Project           *Project
+	requester         string
+	included          map[TVPair]bool
+	generatedVariants map[string]bool
 }
 
 // IncludeDependencies takes a project and a slice of variant/task pairs names
@@ -16,23 +17,23 @@ type dependencyIncluder struct {
 // for the given set of tasks.
 // If any dependency is cross-variant, it will include the variant and task for that dependency.
 // This function can return an error, but it should be treated as an informational warning
-func IncludeDependencies(project *Project, tvpairs []TVPair, requester string) ([]TVPair, error) {
+func IncludeDependencies(project *Project, tvpairs []TVPair, requester string, activationInfo *specificActivationInfo) ([]TVPair, error) {
 	di := &dependencyIncluder{Project: project, requester: requester}
-	return di.include(tvpairs)
+	return di.include(tvpairs, activationInfo)
 }
 
 // include crawls the tasks represented by the combination of variants and tasks and
 // add or removes tasks based on the dependency graph. Dependent tasks
 // are added; tasks that depend on unreachable tasks are pruned. New slices
 // of variants and tasks are returned.
-func (di *dependencyIncluder) include(initialDeps []TVPair) ([]TVPair, error) {
+func (di *dependencyIncluder) include(initialDeps []TVPair, activationInfo *specificActivationInfo) ([]TVPair, error) {
 	di.included = map[TVPair]bool{}
-
+	di.generatedVariants = map[string]bool{}
 	warnings := grip.NewBasicCatcher()
 	// handle each pairing, recursively adding and pruning based
 	// on the task's dependencies
 	for _, d := range initialDeps {
-		_, err := di.handle(d)
+		_, err := di.handle(d, activationInfo)
 		warnings.Add(err)
 	}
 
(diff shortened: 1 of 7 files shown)


model/dependencies.go
```diff
@@ -90,8 +99,24 @@ func (di *dependencyIncluder) handle(pair TVPair) (bool, error) {
...
 	deps := di.expandDependencies(pair, bvt.DependsOn)
+
+	// If this function is invoked from generate.tasks, calculate all variants
+	// that need to be included as dependencies, but are not in the generated project.
+	// If all the task / variant pairs that spawn these dependencies are inactive, we
+	// also mark this newly generated variant as inactive.
+	pairSpecifiesActivation := activationInfo != nil && activationInfo.taskOrVariantHasSpecificActivation(pair.Variant, pair.TaskName)
 	for _, dep := range deps {
-		ok, err := di.handle(dep)
+		if activationInfo != nil && !activationInfo.variantExistsInGeneratedProject(dep.Variant) {
```
Kimchelly, line 109: It seems like `variantExistsInGeneratedProject` is the only usage of `generatedProjectVariants`, and that field is not actually needed by `specificActivationInfo` once it's exited from `IncludeDependencies`. Can we just pass `g.BuildVariants` as a separate parameter into `IncludeDependencies` instead of passing it in with the `specificActivationInfo`? Might save some confusion on the purpose of that field, since it's only used in one specific function.

model/dependencies.go
```diff
@@ -42,14 +43,22 @@ func (di *dependencyIncluder) include(initialDeps []TVPair) ([]TVPair, error) {
 			outPairs = append(outPairs, pair)
 		}
 	}
+	// we deactivate variants that do not have specific activation set in the
+	// generated project but have been generated as dependencies of other
+	// explicitly inactive variants
+	for variant, addToActivationInfo := range di.deactivateGeneratedVariantTasks {
```
Kimchelly, line 49: I'm slightly confused by this, but I'd like to understand. The comment on the `activationVariants` field says that it's all variants that have some specific activation, but the comment around `pairSpecifiesActivation` implies that we should deactivate those variants where `di.deactivateGeneratedVariantTasks = true`. Why does this append variants that are meant to be deactivated?

units/generate_tasks_test.go
```diff
@@ -547,6 +638,36 @@ func TestGeneratedTasksAreNotDependencies(t *testing.T) {
...
+		Status:                evergreen.TaskStarted,
+	}
+	require.NoError(generateTask.Insert())
+	j = NewGenerateTasksJob(generateTask.Version, generateTask.Id, "1")
+	j.Run(context.Background())
+	assert.NoError(j.Error())
+	tasks = []task.Task{}
+	assert.NoError(db.FindAllQ(task.Collection, db.Query(task.ByVersion("sample_version")), &tasks))
+	assert.Len(tasks, 3)
+	for _, foundTask := range tasks {
```
Kimchelly, line 666: Do we have to check whether the other tasks are activated? Or do they not matter?

Kimchelly (commented):
Mostly questions for my own understanding.

model/dependencies.go
```diff
@@ -90,8 +99,24 @@ func (di *dependencyIncluder) handle(pair TVPair) (bool, error) {
...
+
+	// If this function is invoked from generate.tasks, calculate all variants
+	// that need to be included as dependencies, but are not in the generated project.
+	// If all the task / variant pairs that spawn these dependencies are inactive, we
+	// also mark this newly generated variant as inactive.
+	pairSpecifiesActivation := activationInfo != nil && activationInfo.taskOrVariantHasSpecificActivation(pair.Variant, pair.TaskName)
 	for _, dep := range deps {
-		ok, err := di.handle(dep)
+		if activationInfo != nil && !activationInfo.variantExistsInGeneratedProject(dep.Variant) {
+			// If the new variant has not yet been added to deactivateGeneratedVariantTasks, or if the
```
ablack12, line 110: This double if statement is pretty hard to read; i think it might be good to break to a helper function that has a comment per line instead. 
variantExistsInGenerated can check if activationInfo == nil also.

I'm confused about why at 109 we just look at variant but I'm slightly confused about what's going on here. If a task depends on a variant, do we add all of those variants tasks to the depends on? (In which case I would think we'd want to keep track of the tasks instead)

model/dependencies.go
ablack12, line 49: Why add these here as opposed to activationVariants directly? Is it because we could be setting it back to false in handle?

model/dependencies.go
```diff
@@ -90,8 +99,24 @@ func (di *dependencyIncluder) handle(pair TVPair) (bool, error) {
...
+			// original pair needs to be active, we update deactivateGeneratedVariantTasks.
+			// We ultimately will only deactivate new variants where deactivateGeneratedVariantTasks[variant] = true.
+			// If deactivateGeneratedVariantTasks[variant] = false it signifies that there was at least
+			// one pair that depends on this new variant being active - so we cannot deactivate it.
+			if _, ok := di.deactivateGeneratedVariantTasks[dep.Variant]; !ok || !pairSpecifiesActivation {
+				di.deactivateGeneratedVariantTasks[dep.Variant] = pairSpecifiesActivation
```
ablack12, line 116: The name deactivateGeneratedVariantTasks is a bit misleading, since the "tasks" implies that we're storing task names, not variants. We also ultimately wrap this into activationVariants it looks like, so maybe instead this should be generatedActivationVariants?

model/generate.go
```diff
@@ -443,24 +443,35 @@ func (g *GeneratedProject) filterInactiveTasks(tasks TVPairSet, v *Version, p *P
 }
 
 type specificActivationInfo struct {
-	stepbackTasks      map[string][]specificStepbackInfo // tasks by variant that are being stepped back, along with the stepback depth to use
-	activationTasks    map[string][]string               // tasks by variant that have batchtime or activate specified
-	activationVariants []string                          // variants that have batchtime or activate specified
+	stepbackTasks            map[string][]specificStepbackInfo // tasks by variant that are being stepped back, along with the stepback depth to use
+	activationTasks          map[string][]string               // tasks by variant that have batchtime or activate specified
+	activationVariants       []string                          // variants that have batchtime or activate specified
+	generatedProjectVariants []parserBV                        // list of all variants that are specified in the generated project
```
ablack12, line 449: If this is only used by the dependency includer can it be added to that instead (same as deactivateGeneratedVariantTasks)?

ablack12 (changes requested):

model/dependencies.go
hadjri: Moved to helper function.

True, there could be a case where some generated dependencies in a newly generated variant need to be activated, while others remain deactivated (added this case to unit tests). I suppose I missed that since server's original issue was fixed by deactivating the entire new variant. I've updated the logic here to track individual task/variant pairs rather than entire variants for better granularity.

model/dependencies.go
hadjri: Yes - we need to wait for the recursion of `handle` to complete since even a single active task that depends on the newly generated dependency should make cause the generated dependency to also be active

model/generate.go
hadjri: I've since moved this to be an additional [parameter](https://github.com/evergreen-ci/evergreen/pull/6111#discussion_r1042650009) to `IncludeDependencies`

model/dependencies.go
hadjri: I've renamed to `deactivateGeneratedDep` since really the variable represents if a dependency should be added to the activation info's list of tasks with specific activation (i.e. the list to deactivate)

units/generate_tasks_test.go
hadjri: They're not relevant in this case since before this fix only `testBV4` would be erroneously activated.

model/dependencies.go
hadjri: That is the way that specific activation is computed (checking if activation is non nil amongst other things) but the way it is used is looking like variants / tasks with "specific activation" are marked as inactive no matter what the value actually is. This is a bug since setting `activate: true` in the generated project will actually cause the task/variant to be deactivated. I was thinking we should start referring to it explicitly as "deactivation" rather than "specific activation" in the variable function naming, but open to other suggestions.

model/dependencies.go
hadjri: I considered this but went with the current approach for fear of overloading `IncludeDependencies` and its helpers with parameters, but it does seem a little needless to add a new field to `specificActivationInfo` that is sort of unrelated to its purpose. Done.

model/dependencies.go
```diff
@@ -6,33 +6,34 @@ import (
 )
 
 type dependencyIncluder struct {
-	Project   *Project
-	requester string
-	included  map[TVPair]bool
+	Project                         *Project
+	requester                       string
+	included                        map[TVPair]bool
+	deactivateGeneratedVariantTasks map[string]bool
```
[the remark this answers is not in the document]
hadjri: One further problem is that a task or variant having a "specific activation" at this point actually means that it is meant to be inactive ([here](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L1157) and [here](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L1565) among others). My best guess is that the name meant that we are "specifically" setting a task/variant to `active: false` hence its name.  

`deactivateGeneratedVariantTasks` represents if a dependency should be added to the activation info's list of tasks (i.e. list to deactivate).  It's a `map[string]bool` because we only want to deactivate a generated dependency everything that depends on it is also marked inactive. If even one task that depends on the generated dependency is active, we'll need to activate the generated dependency as well.  If we encounter this case then the value in the map is set to `false` and we skip it later when adding tasks to the activation list.

Perhaps a name like `deactivateGeneratedDep` is more accurate, I'll rename for now but we can discuss.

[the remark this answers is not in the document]
Kimchelly: Got it. I think I was confused because `activate` is really exposing an option to deactivate, so `activationVariants` are those that explicitly set `active: false`. Thanks for the clear explanation!

model/dependencies.go
```diff
@@ -6,33 +6,37 @@ import (
...
+	requester              string
+	included               map[TVPair]bool
+	deactivateGeneratedDep map[TVPair]bool
 }
 
 // IncludeDependencies takes a project and a slice of variant/task pairs names
 // and returns the expanded set of variant/task pairs to include all the dependencies/requirements
 // for the given set of tasks.
 // If any dependency is cross-variant, it will include the variant and task for that dependency.
 // This function can return an error, but it should be treated as an informational warning
```
Kimchelly, line 19: nit: Period at the end of this sentence so we know that `activationInfo` is not a continuation.

Kimchelly (approved):
LGTM mod nit + follow-up question about `activate: true`.

model/dependencies.go
```diff
@@ -6,33 +6,37 @@ import (
...
-func IncludeDependencies(project *Project, tvpairs []TVPair, requester string) ([]TVPair, error) {
+// activationInfo and generatedVariants are required in the case for generate tasks to detect if
+// new generated dependency's task/variant pairs are depended on by inactive tasks. If so,
+// we also set these new dependencies to inactive.
+func IncludeDependencies(project *Project, tvpairs []TVPair, requester string, activationInfo *specificActivationInfo, generatedVariants []parserBV) ([]TVPair, error) {
```
ablack12, line 23: Organizational nit: can we have two functions IncludeDependencies and IncludeDependenciesWithGenerated (or etc)?
IncludeDependenciesWithGenerated can have this function signature ^^^ and IncludeDependencies can just wrap IncludeDependencesWithGenerated but pass in nil, nil for the last two arguments. That would clean up the usage of this in everywhere but generated tasks.

ablack12 (approved):
LGTM with some nits and re-testing in staging

bynn (approved):
i dont think you need an extra one but i think all the changes make sense after reading your explanation thank you!

model/dependencies.go
hadjri: For the purpose of not expanding this PR I'll ticket it out: https://jira.mongodb.org/browse/EVG-18543

model/dependencies.go
```diff
@@ -6,33 +6,37 @@ import (
 )
 
 type dependencyIncluder struct {
-	Project   *Project
-	requester string
-	included  map[TVPair]bool
+	Project                *Project
+	requester              string
+	included               map[TVPair]bool
+	deactivateGeneratedDep map[TVPair]bool
```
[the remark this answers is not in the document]
hadjri: Updated to deps since I think it should be clear that we're only doing this for generated dependencies.

model/generate.go
```diff
@@ -461,6 +461,18 @@ func newSpecificActivationInfo() specificActivationInfo {
 	}
 }
 
+func (b *specificActivationInfo) variantExistsInGeneratedProject(variant string, variants []parserBV) bool {
```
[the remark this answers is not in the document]
hadjri: Good point. This is a relic from when I initially added the generated BVs as a field on `specificActivationInfo `

PR #6112, opened 2022-12-06
EVG-17305 Don't activate builds during addNewTasks if no tasks are active

malikchaya2:

### Description 
When generating tasks, if the build already exists, the addNewTasks function activates the build even if it shouldn't be activated. 

### Testing 
Tested on staging. [Before](https://spruce-staging.corp.mongodb.com/version/638eb4e79dbe327e7b1b0be1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [After](https://spruce-staging.corp.mongodb.com/version/638eb79197b1d331945c27e4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/model/lifecycle_test.go
+++ b/model/lifecycle_test.go
@@ -2609,6 +2609,7 @@ func TestAddNewTasks(t *testing.T) {
 			assert.NoError(t, err)
 			build, err := build.FindOneId("b0")
 			assert.NoError(t, err)
+			assert.NotNil(t, build)
 			assert.Equal(t, len(testCase.activatedTasks), len(activatedTasks))
 			for _, task := range activatedTasks {
 				assert.Contains(t, testCase.activatedTasks, task.DisplayName)


Kimchelly (commented):
Is there a test case for this case when all the tasks in the build variant are inactive? If not, can we add one? Is it also worth testing the explicitly setting `activate: false` for the build variant keeps it deactivated (or is that already tested)?

Kimchelly:
I still have the same question about the tests.

malikchaya2:
> I still have the same question about the tests.

I added a case to test this explicitly, and added a check for the bv activated state to each case.

model/lifecycle_test.go
```diff
@@ -2587,10 +2607,13 @@ func TestAddNewTasks(t *testing.T) {
 			assert.NoError(t, err)
 			activatedTasks, err := task.FindAll(db.Query(bson.M{task.ActivatedKey: true}))
 			assert.NoError(t, err)
+			build, err := build.FindOneId("b0")
+			assert.NoError(t, err)
```
Kimchelly, line 2611: Is this needed? Seems like we don't need to double-check for the existence of `b0`.

Kimchelly (approved):
The test looks good! LGTM mod one question about the test.

model/lifecycle_test.go
malikchaya2: We need to pull it from the db to check if addNewTasks activated the build.

Kimchelly: Ah I see. Does it need to also check if `build` is non-nil? Just checking the error won't check for the doc not found case.

Kimchelly: The test will panic on that line if `build` is nil, but I kind of prefer that tests don't panic if they don't have to (because they stop the entire test runtime, rather than just this test). I don't think this is a hard requirement given how often the existing tests _can_ panic, but it is part of the style guide and it would be nice if we could aim for less panicking in tests.

malikchaya2: good point, fixed

PR #6113, opened 2022-12-06
EVG-18436 set if projects track push events more carefully

ablack12:
[EVG-18436](https://jira.mongodb.org/browse/EVG-18436)

### Description 
This ticket highlighted a couple of projects that didn't get TracksPushEvents set for the project or repo, because we'd create the hook but not update the project with it (the new UI neglects to set this field entirely). Made modifications to both update the repo structure when a project is attached/detached and to update the project ref when we check webhooks.

Will need to manually update repos that don't have this field set to ensure that testing works correctly.

### Testing 
Added to unit tests; verified on staging that modifying the general page updates the field correctly (i.e. doesn't update it unless checking the changed owner/repo field -- used https://spruce-staging.corp.mongodb.com/project/annie-copy2/settings/general)

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -281,6 +281,7 @@ var (
 	ProjectRefAdminsKey                   = bsonutil.MustHaveTag(ProjectRef{}, "Admins")
 	ProjectRefGitTagAuthorizedUsersKey    = bsonutil.MustHaveTag(ProjectRef{}, "GitTagAuthorizedUsers")
 	ProjectRefGitTagAuthorizedTeamsKey    = bsonutil.MustHaveTag(ProjectRef{}, "GitTagAuthorizedTeams")
+	ProjectRefTracksPushEventsKey         = bsonutil.MustHaveTag(ProjectRef{}, "TracksPushEvents")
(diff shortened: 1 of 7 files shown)


model/project_ref.go
```diff
@@ -745,22 +747,15 @@ func (p *ProjectRef) AttachToNewRepo(ctx context.Context, u *user.DBUser) error
...
+
 	err = db.UpdateId(ProjectRefCollection, p.Id, bson.M{
-		"$set": update,
+		"$set":   update,
+		"$unset": bson.M{ProjectRefTracksPushEventsKey: 1},
 	})
 	if err != nil {
 		return errors.Wrap(err, "updating owner/repo in the DB")
 	}
-	_, err = EnableWebhooks(ctx, p)
```
bynn, line 754: i might be confused but was there a need to get rid of enabling webhooks here?

bynn (approved):
i had a question but it looks good

model/project_ref.go
ablack12: I switched to unsetting it because we can rely on the repo's value.

PR #6120, opened 2022-12-08
EVG-17118 add route to get updated view of num executions

ablack12:
[EVG-17118](https://jira.mongodb.org/browse/EVG-17118)

### Description 
Right now the TPMs use "failure frequency" for build baron prioritization, and they use task_stats to understand how often a task has run. However, task_stats is cached so the data might not be up-to-date, so they may get notifications about more failures than they see executions for (when they check task_stats).

A solution to this would be to just count the number of executions. Task stats is more expensive because of the aggregation which separates failure types but really count queries are fast when they use proper indexes. I ran an explain on this query in the shell and it looks like for the task collection we use the index `branch_1_build_variant_1_display_name_1_status_1_finish_time_1_start_time_1` and for old_tasks we use `branch_1_r_1_display_name_1_create_time_1` which isn't quite as specific so there could be an argument to add a new index to that collection, but I think it's still granular enough.

I also returned a struct because there's a world in which we extend this (for example, Brooke's team uses task_stats to understand num_success / num_failed; considering these queries shouldn't be too expensive, having four to split things up probably also wouldn't be too expensive, but it's hard to say then when this just becomes a non-cached task_stats).

### Testing 
Added many unit tests.


model/task/db.go
```diff
@@ -2692,3 +2692,43 @@ func enableDisabledTasks(taskIDs []string) error {
...
+	StartTime    time.Time
+	EndTime      time.Time
+}
+
+func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, error) {
+	query := bson.M{
+		ProjectKey:      input.ProjectId,
+		BuildVariantKey: input.BuildVarName,
+		DisplayNameKey:  input.TaskName,
+		StatusKey:       bson.M{"$in": evergreen.TaskCompletedStatuses},
```
malikchaya2, line 2710: Why do we want to filter this on TaskCompletedStatuses? If a task is in it's third execution, and the third one is still running, don't we still want to see 3 as the number of executions?

model/task/db_test.go
```diff
@@ -1632,3 +1632,142 @@ func TestFindAllMarkedUnattainableDependencies(t *testing.T) {
...
+	}
+
+	for testName, test := range map[string]func(*testing.T){
+		"nothingInRange": func(t *testing.T) {
+			input := NumExecutionsForIntervalInput{
+				ProjectId:    "myProject",
+				BuildVarName: "bv1",
+				TaskName:     "task1",
+				StartTime:    time.Now().Add(-20 * time.Hour),
+				EndTime:      time.Now().Add(-18 * time.Hour),
```
malikchaya2, line 1727: I think it would be worth it to add a test where Endtime=ZeroTime

rest/route/project.go
```diff
@@ -1011,6 +1014,87 @@ func (h *getProjectTasksHandler) Run(ctx context.Context) gimlet.Responder {
...
+
+	if h.opts.BuildVariant == "" || h.opts.TaskName == "" {
+		return errors.New("'build_variant' and 'task_name' are required")
+	}
+	h.startTime, err = model.ParseTime(h.opts.StartTime)
+	if err != nil {
+		return errors.Wrapf(err, "parsing 'start_time' %s", h.opts.StartTime)
+	}
+
+	// If not specified, get the latest information.
```
malikchaya2, line 1055: I was initially confused by the verbiage of the comment because it says `If not specified` but it's above the case where it is specified, (!=""). I wonder if there's a better way to phrase it (although I can't think of one, lol).

rest/route/project.go
```diff
@@ -1011,6 +1014,87 @@ func (h *getProjectTasksHandler) Run(ctx context.Context) gimlet.Responder {
...
+	if h.opts.EndTime != "" {
+		h.endTime, err = model.ParseTime(h.opts.EndTime)
+		if err != nil {
+			return errors.Wrapf(err, "parsing 'end_time' %s", h.opts.EndTime)
+		}
+	}
+	if h.startTime.After(h.endTime) {
```
malikchaya2, line 1062: How will this behave if h.endTime is not defined?

rest/route/project_test.go
```diff
@@ -1240,3 +1241,222 @@ func (s *ProjectPutRotateSuite) TestRotateProjectVars() {
...
+			err := rm.Parse(context.Background(), req)
+			assert.Error(t, err)
+		},
+		"parseNoBuildVariant": func(t *testing.T, rm *getProjectTaskExecutionsHandler) {
+			body := []byte(
+				`{
+                     "task_name": "t1",
+                     "requesters": ["gitter_request"],
+                     "start_time": "2022-11-02T00:00:00.000Z",
+                     "end_time": "2022-11-03T00:00:00.000Z"
```
malikchaya2, line 1423: Can we please add two test cases, one with no end time and one with no start time so that we can make sure that parse handles them as expected?

malikchaya2 (changes requested):

model/task/db.go
bynn: following 
what do we expect to happen if the task is aborted or unscheduled?

rest/route/project.go
```diff
@@ -1011,6 +1014,87 @@ func (h *getProjectTasksHandler) Run(ctx context.Context) gimlet.Responder {
...
+		}
+	}
+
+	h.projectName = gimlet.GetVars(r)["project_id"]
+	h.projectId, err = dbModel.GetIdForProject(h.projectName)
+	if err != nil {
+		return errors.Wrap(err, "getting id for project")
+	}
+
+	for _, requester := range h.opts.Requesters {
```
bynn, line 1072: i am adding this as a part of my pr as well (build together)

bynn (approved):

model/task/db.go
malikchaya2: That makes sense. I would advocate to use a route name that makes it clear that it's the number of `completed` executions so that someone doesn't accidentally use it if it doesn't match their use case. 

Maybe something like `num_completed_executions` or `completed_executions_count`. It's definitely more clunky but I think it's worth it.

rest/route/project.go
malikchaya2: thank you!

malikchaya2 (changes requested):
I had one comment left about the name of the route

model/task/db.go
ablack12: I'll do this for the response variable name instead maybe -- that way the route name will still work if we do want to start allowing for status to be modifiable or etc, and also will make the variable itself clearer if we ever add more

malikchaya2 (approved):

PR #6121, opened 2022-12-08
EVG-14528 fix panics 

malikchaya2:

### Description 
Fix panics 

### Testing 
reasoned about it

### DIFF
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -368,7 +368,7 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
 	if t.data.Status != evergreen.PatchSucceeded && t.data.Status != evergreen.PatchFailed {
 		return nil, nil
 	}
-	if t.event.EventType != event.VersionChildrenCompletion {
+	if t.event.EventType != event.PatchChildrenCompletion {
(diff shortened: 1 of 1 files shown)


julianedwards (approved):

Kimchelly (approved):
LGTM the hotfix, but for my own understanding, what was the original issue causing the panic?

Do we really not need to test this if it can produce panics?

julianedwards:
The `run-child-patch` subscription listens on the patch event `CHILDREN_FINISHED` because the patch trigger event handler is shared and the attributes overlap. The panics themselves are harmless and all event triggers are getting handled correctly. In the case of the the `run-child-patch` subscription, it gets picked up twice [here](https://github.com/evergreen-ci/evergreen/blob/main/trigger/process.go#L36); the second time, incorrectly, when `CHILDREN_FINISHED` patch event is logged and processed. This causes run the [patch outcome trigger ](https://github.com/evergreen-ci/evergreen/blob/main/trigger/patch.go#L84) to run when it shouldn't, and thus the panic. Chaya and I discussed refactoring the patch trigger event handler into two handlers—one for family outcome triggers and another for regular outcome triggers—that can share a base handler since there is a lot of shared code.

PR #6122, opened 2022-12-08
EVG-14528: Check event type in version completion triggers 

malikchaya2:

### Description 
Tying up loose ends


julianedwards (approved):

bynn (approved):
should we merge this change soon? i think it could be the cause of the ops alerts

PR #6123, opened 2022-12-09
EVG-18079 Correct limit functionality for get tasks by project endpoint

bynn:
[EVG-18079](https://jira.mongodb.org/browse/EVG-18079)

### Description 
previously, setting the limit on the endpoint would limit the number of tasks returned. It now correctly limits the number of versions returned, even if more tasks are returned due to duplicate names

### Testing 
unit test 
postman on staging

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -456,8 +456,7 @@ const (
 )
 
 const (
-	tasksByProjectQueryMaxTime   = 90 * time.Second
-	tasksByProjectMaxNumVersions = 1000
+	tasksByProjectQueryMaxTime = 90 * time.Second
 )
 
 var adminPermissions = gimlet.Permissions{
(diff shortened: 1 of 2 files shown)


model/project_ref.go
```diff
@@ -1198,11 +1197,10 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 	}
 	match["$and"] = []bson.M{
 		{task.RevisionOrderNumberKey: bson.M{"$lte": startingRevision}},
-		{task.RevisionOrderNumberKey: bson.M{"$gte": startingRevision - tasksByProjectMaxNumVersions}},
+		{task.RevisionOrderNumberKey: bson.M{"$gte": startingRevision - opts.Limit}},
```
ablack12, line 1200: The documentation says "the number of versions to be returned" but here it's actually that these are the number of versions we're checking (it could be that some versions don't have the task or something in which case we'd skip those versions).

I can't think of a way to actually return the latest 20 versions without _also_ querying the version collection though so that does get maybe a bit expensive; perhaps we should change the documentation (and update the user) that it'll be the limit for what we search rather than for what we return 🤔

ablack12: Also need to update the function comment.

bynn: I couldn't figure out an aggregation that would keep the behavior that we described so I changed it in the docs to searching the number of versions specified. https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-tasks-for-a-project

I don't think this would be an issue because the current behavior wasn't as described so it shouldn't impact workflow. Will update the ticket after merging

bynn: also updated function comment

model/project_ref.go
```diff
@@ -1166,7 +1165,7 @@ func CountProjectRefsWithIdentifier(identifier string) (int, error) {
 	return db.CountQ(ProjectRefCollection, byId(identifier))
 }
 
-// GetTasksWithOptions will find the last number of tasks (denoted by Limit) that exist for a given project.
+// GetTasksWithOptions will find the matching task run in the last number of versions(denoted by Limit) that exist for a given project.
```
ablack12, line 1168: nit: tasks

ablack12 (approved):

PR #6124, opened 2022-12-09
EVG-14528 Filter out aborted patches from the RunChildPatchSubscriberType subscription generation 

malikchaya2:

### DIFF
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -102,7 +102,7 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
 		anyOutcome := ps == evergreen.PatchAllOutcomes
 
 		if successOutcome || failureOutcome || anyOutcome {
-			aborted, err := t.patch.IsAborted()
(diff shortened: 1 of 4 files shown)


trigger/patch.go
```diff
@@ -102,7 +102,14 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
 		anyOutcome := ps == evergreen.PatchAllOutcomes
 
 		if successOutcome || failureOutcome || anyOutcome {
-			err := finalizeChildPatch(sub)
+			aborted, err := model.IsAborted(t.patch.Id.Hex())
```
julianedwards, line 105: Why is the required now? Shouldn't we have seen this issue before since this code hasn't changed?

malikchaya2: I'm a bit confused about this too. The only difference I see is that the subscription now matches two events, when it used to match one. 

I'm hoping that working on [EVG-18381](https://jira.mongodb.org/browse/EVG-18381) will provide more clarity.

julianedwards (approved):

PR #6125, opened 2022-12-09
EVG-17180 add cron/batchtime to build route

ablack12:
[EVG-17180](https://jira.mongodb.org/browse/EVG-17180)

### Description 
Added cron/batchtime to routes to provide visibility for cloud. Added this query to rest/model instead of parser_project_db since it's only needed there, and because it seems slightly odd for parser_project_db to return an API struct (it can't return the parserBV since it's not exported) but I'm open to suggestions here.

### Testing 
Added test.


rest/route/build_test.go
```diff
@@ -79,6 +97,9 @@ func (s *BuildByIdSuite) TestFindBuildById() {
 	s.Equal(utility.ToStringPtr("branch"), b.ProjectId)
 	s.Equal("task1", b.TaskCache[0].Id)
 	s.Equal(evergreen.TaskFailed, b.TaskCache[0].Status)
+	s.Equal("@daily", utility.FromStringPtr(b.DefinitionInfo.CronBatchTime))
+	s.NotNil(b.DefinitionInfo.BatchTime)
+	s.Equal(0, utility.FromIntPtr(b.DefinitionInfo.BatchTime))
```
Kimchelly, line 102: Can this test a batchtime number other than 0 since it's the default int value? It would also obviate the nil check right above this.

rest/model/build.go
```diff
@@ -38,14 +38,36 @@ type APIBuild struct {
 	RevisionOrderNumber int            `json:"order"`
 	TaskCache           []APITaskCache `json:"task_cache,omitempty"`
 	// Tasks is the build's task cache with just the names
-	Tasks             []string             `json:"tasks"`
+	Tasks             []*string            `json:"tasks"`
```
Kimchelly, line 41: I don't have a strong opinion on this, but is this change just to keep it consistent with `Tags`? I feel like returning just `[]string` instead of `[]*string` was nice since the users don't have to account for JSON `null`s within the array.

ablack12: Yeah sort of didn't think about it; i'll change it back

rest/route/build_test.go
ablack12: I actually think testing 0 is stronger since it verifies that explicitly setting batchtime to 0 has different behavior than not setting it

rest/model/build.go
```diff
@@ -46,6 +46,28 @@ type APIBuild struct {
...
+func (apiBuild *APIBuild) PopulateDefinitionInfo(id, variant string) {
+	pp, err := model.ParserProjectFindOne(model.ParserProjectById(id).WithFields(model.ParserProjectBuildVariantsKey))
+	if err == nil {
+		for _, bv := range pp.BuildVariants {
+			if bv.Name == variant {
+				if bv.CronBatchTime != "" {
+					apiBuild.DefinitionInfo.CronBatchTime = utility.ToStringPtr(bv.CronBatchTime)
+				}
+				apiBuild.DefinitionInfo.BatchTime = bv.BatchTime
+			}
```
bynn, line 68: should we error or warn if the bv specified is not found?

ablack12:
> Sorry, one more follow-on since I thought of it - the [docs](https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#build-variants) say that cron/batchtime can also be defined at the task level. Is this supposed to be returned as well for completeness, or just the build variant cron/batchime?

I was thinking about that 🤔 I decided against it but could go either way. For completeness it would make sense to add to the task route but cloud only seems to need it for the build variant route so I'm unsure if it's worth adding the complexity to the task route (and it would need to also grab the variant ones then to resolve all the places it could be defined at the task level).

rest/model/build.go
ablack12: I decided to skip this since it shouldn't happen and shouldn't affect the route (also since this isn't supported for matrices we can't error since it would mean the route is unusable in those cases)

ablack12:
Added support for an additional route based on cloud feedback.

operations/validate.go
```diff
@@ -174,7 +175,7 @@ func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model
 	if err != nil {
 		errs = append(errs, validator.ValidationError{
 			Level:   validator.Error,
-			Message: err.Error(),
+			Message: fmt.Sprintf(err.Error(), "creating project config"),
```
Kimchelly: Seems like this has to be `errors.Wrap(err, "creating project config").Error()` or `fmt.Sprintf("creating project config: %s", err)` since `err.Error()` is not a formatting string.

operations/validate.go
```diff
@@ -194,7 +195,7 @@ func loadProjectIntoWithValidation(ctx context.Context, data []byte, opts *model
 		}
 		errs = append(errs, validator.ValidationError{
 			Level:   validator.Error,
-			Message: err.Error(),
+			Message: fmt.Sprintf(err.Error(), "Loading project into"),
```
Kimchelly: Same as above.

rest/route/build.go
```diff
@@ -246,6 +249,6 @@ func (b *buildRestartHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 
 	buildModel := &model.APIBuild{}
-	buildModel.BuildFromService(*foundBuild)
+	buildModel.BuildFromService(*foundBuild, nil)
```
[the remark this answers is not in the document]
ablack12: I might be misunderstanding but this comment isn't on a test. We do test the BuildFromService non-nil case implicitly in the test TestFindBuildById.

operations/validate.go
ablack12: I removed this because i didn't mean for this to be part of this PR

rest/route/build.go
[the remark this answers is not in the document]
Kimchelly: Ah yeah, I meant to comment on the test file. As long as it's already covered, then it's good enough.

Kimchelly (approved):

bynn (approved):

PR #6126, opened 2022-12-09
EVG-18285: Spinning our wheels on host/task cleanup

mridul-augustine:
[EVG-18285](https://jira.mongodb.org/browse/EVG-18285)

### Description 
The Splunk error is not the reason why the host was not removed. Based on the conversation with @Kimchelly, the background cleanup job that was supposed to clean up these hosts doesn't handle the InvalidInstanceID.NotFound error yet. We already have a ticket - [EVG-18388](https://jira.mongodb.org/browse/EVG-18388) to fix this.

In this case, the error is reported even though `InvalidInstanceID.NotFound` is handled. I have fixed this part. This will fix the wrong Splunk error.

### DIFF
--- a/cloud/ec2_fleet.go
+++ b/cloud/ec2_fleet.go
@@ -241,7 +241,7 @@ func (m *ec2FleetManager) GetInstanceStatus(ctx context.Context, h *host.Host) (
 	instance, err := m.client.GetInstanceInfo(ctx, h.Id)
 	if err != nil {
 
-		if ec2err, ok := errors.Cause(err).(awserr.Error); ok && ec2err.Code() == EC2ErrorNotFound {
+		if ec2Err, ok := errors.Cause(err).(awserr.Error); ok && ec2Err.Code() == EC2ErrorNotFound {
 			return StatusNonExistent, nil
 		}
 		grip.Error(message.WrapError(err, message.Fields{
--- a/cloud/ec2_fleet.go
+++ b/cloud/ec2_fleet.go
@@ -241,7 +241,7 @@ func (m *ec2FleetManager) GetInstanceStatus(ctx context.Context, h *host.Host) (
 	instance, err := m.client.GetInstanceInfo(ctx, h.Id)
 	if err != nil {
 
-		if ec2err, ok := errors.Cause(errors.Cause(err)).(awserr.Error); ok && ec2err.Code() == EC2ErrorNotFound {
+		if ec2err, ok := errors.Cause(err).(awserr.Error); ok && ec2err.Code() == EC2ErrorNotFound {
 			return StatusNonExistent, nil
 		}
 		grip.Error(message.WrapError(err, message.Fields{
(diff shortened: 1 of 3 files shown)


cloud/ec2_fleet.go
```diff
@@ -240,15 +240,15 @@ func (m *ec2FleetManager) GetInstanceStatus(ctx context.Context, h *host.Host) (
 
 	instance, err := m.client.GetInstanceInfo(ctx, h.Id)
 	if err != nil {
+		if ec2err, ok := err.(awserr.Error); ok && ec2err.Code() == EC2ErrorNotFound {
```
Kimchelly, line 243: I think we have to use `ec2err, ok := errors.Cause(err).(awserr.Error)` here because if `GetInstanceInfo` errors, it wraps the error in a specially-defined `error` type that has to be unwrapped to get the original EC2 SDK error.

Kimchelly, line 243: Is there already a test to verify that `GetInstanceStatus` returns `StatusNonExistent` in this case? If not, can we add one?

mridul-augustine: The AWS error is wrapped twice, once by retry and then by `GetInstanceInfo.`, so I'm unwrapping it twice now.

mridul-augustine: We didn't have a test. I have added one.

Kimchelly: It shouldn't be necessary to make multiple calls to `errors.Cause`. `errors.Cause` already unwraps the base error recursively.

cloud/ec2_fleet_test.go
```diff
@@ -55,6 +57,17 @@ func TestFleet(t *testing.T) {
 			assert.NoError(t, err)
 			assert.Equal(t, "us-east-1a", hDb.Zone)
 		},
+		"GetInstanceStatusNonExistent": func(*testing.T) {
+			aws_error := awserr.New(EC2ErrorNotFound, "The instance ID 'test-id' does not exist", nil)
+			wrapped_aws_error := errors.Wrap(
```
Kimchelly, line 62: It shouldn't be necessary to wrap this twice since we assume that `errors.Cause` behaves as expected with `errors.Wrap`. Just one `errors.Wrap` should be sufficient.

ybrill:
@mridul-augustine
Sorry, what's the other ticket for responding to the `StatusNonExistent` error appropriately? ([EVG-18285](https://jira.mongodb.org/browse/EVG-18285) is this one)

mridul-augustine:
> @mridul-augustine Sorry, what's the other ticket for responding to the `StatusNonExistent` error appropriately? ([EVG-18285](https://jira.mongodb.org/browse/EVG-18285) is this one)

Sorry its [EVG-18388](https://jira.mongodb.org/browse/EVG-18388)

ybrill:
Thanks! I understand the larger picture now.
I'm going to remove myself from the reviewers list, if that's okay, because I don't think I have anything to add to what's been said.

cloud/ec2_fleet.go
mridul-augustine: Oh got it. Still getting familiar with Go. Should have read the docs.

Kimchelly (approved):

PR #6127, opened 2022-12-09
EVG-18116 Add requesters to project tasks route opts

bynn:
[EVG-18116](https://jira.mongodb.org/browse/EVG-18116)

### Description 
added flags to only show mainline or patch tasks to get tasks by project endpoint

### Testing 
unit tests

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -37,11 +37,10 @@ const (
 )
 
 type GetProjectTasksOpts struct {
-	Limit        int    `json:"num_versions"`
-	BuildVariant string `json:"build_variant"`
-	StartAt      int    `json:"start_at"`
-	PatchOnly    bool   `json:"patch_only"`
-	MainlineOnly bool   `json:"mainline_only"`
+	Limit        int      `json:"num_versions"`
+	BuildVariant string   `json:"build_variant"`
+	StartAt      int      `json:"start_at"`
+	Requesters   []string `json:"requesters"`
 }
 
 type Project struct {
--- a/model/project.go
+++ b/model/project.go
@@ -40,6 +40,8 @@ type GetProjectTasksOpts struct {
 	Limit        int    `json:"num_versions"`
 	BuildVariant string `json:"build_variant"`
 	StartAt      int    `json:"start_at"`
+	PatchOnly    bool   `json:"patch_only"`
+	MainlineOnly bool   `json:"mainline_only"`
 }
 
 type Project struct {
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1181,10 +1181,10 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 		task.DisplayNameKey: taskName,
 		task.StatusKey:      bson.M{"$in": finishedStatuses},
 	}
-	if opts.MainlineOnly {
+	if len(opts.Requesters) > 0 {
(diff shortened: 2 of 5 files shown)


model/project_ref.go
```diff
@@ -1181,6 +1181,11 @@ func GetTasksWithOptions(projectName string, taskName string, opts GetProjectTas
 		task.DisplayNameKey: taskName,
 		task.StatusKey:      bson.M{"$in": finishedStatuses},
 	}
+	if opts.MainlineOnly {
+		match[task.RequesterKey] = bson.M{"$in": evergreen.SystemVersionRequesterTypes}
+	} else if opts.PatchOnly {
+		match[task.RequesterKey] = evergreen.PatchVersionRequester
```
hadjri, line 1187: I think this should also cover the `IsGitHubPatchRequester` case

hadjri (changes requested):

model/project.go
```diff
@@ -40,6 +40,8 @@ type GetProjectTasksOpts struct {
 	Limit        int    `json:"num_versions"`
 	BuildVariant string `json:"build_variant"`
 	StartAt      int    `json:"start_at"`
+	PatchOnly    bool   `json:"patch_only"`
```
ablack12, line 43: I'd prefer if this took in a list of requesters since that's more consistent with other routes, and is more configurable (and it should probably default to just repotracker versions

ablack12 (changes requested):

model/project.go
bynn: updated

model/project_ref.go
bynn: updated to take in a list of requesters

rest/route/project.go
```diff
@@ -1057,11 +1057,11 @@ func (p *GetProjectAliasResultsHandler) Run(ctx context.Context) gimlet.Responde
 	return gimlet.NewJSONResponse(variantTasks)
 }
 
-////////////////////////////////////////////////////////////////////////
+// //////////////////////////////////////////////////////////////////////
```
bynn, line 1060: not sure why but my new golang linter is making these changes as i save :(

ablack12: LOL

model/project.go
```diff
@@ -37,9 +37,10 @@ const (
 )
 
 type GetProjectTasksOpts struct {
-	Limit        int    `json:"num_versions"`
-	BuildVariant string `json:"build_variant"`
-	StartAt      int    `json:"start_at"`
+	Limit        int      `json:"num_versions"`
+	BuildVariant string   `json:"build_variant"`
+	StartAt      int      `json:"start_at"`
+	Requesters   []string `json:"requesters"`
```
ablack12, line 43: The parse function for this route should validate that if this is specified the requesters are valid.

ablack12 (changes requested):

hadjri (approved):

model/project.go
bynn: haha i assumed it would be in your pr but it depends on who can merge first now
added

ablack12: oh LOL didn't realize this was the same route haha

ablack12: Update my PR only actually adds this for the _new_ route, whereas you're modifying an existing one

ablack12 (approved):

PR #6129, opened 2022-12-12
EVG-18516: Update RestartTask mutation to accept failedOnly

sophstad:
[EVG-18516](https://jira.mongodb.org/browse/EVG-18516)

### Description 
- Update RestartTask mutation to accept `failedOnly` parameter, which is used to restart only failed execution tasks.
- This change does not require a coordinated release with the UI; because `failedOnly` is a bool pointer, it's valid to call the resolver without the `failedOnly` argument altogether (an existing GraphQL unit test ensures this). The UI will be updated to add this field in [EVG-15703](https://jira.mongodb.org/browse/EVG-15703) and then we can change `failedOnly` to a boolean

### Testing 
- Add unit test


graphql/mutation_resolver.go
```diff
@@ -872,18 +872,26 @@ func (r *mutationResolver) OverrideTaskDependencies(ctx context.Context, taskID
...
 	}
 	if t == nil {
-		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("cannot find task with id %s", taskID))
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("cannot find task with id '%s'", taskID))
+	}
+	// TODO (EVG-15703): Convert *bool to bool once UI has been deployed.
+	if err := model.ResetTaskOrDisplayTask(t, username, evergreen.UIPackage, utility.FromBoolPtr(failedOnly), nil); err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("error restarting task '%s': %s", taskID, err.Error()))
+	}
+	t, err = task.FindOneId(taskID)
```
bynn, line 889: Do we not want to use FindOneIdAndExecutionWithDisplayStatus again here to populate the api task as we did previously?

bynn (changes requested):

graphql/mutation_resolver.go
sophstad: Good catch, I switched the two queries by accident (I don't think we need the execution/display status to restart the task—at least we don't [here](https://github.com/evergreen-ci/evergreen/blob/299bf20c8da9e353b1097183672cd5169ff3447d/rest/route/task_restart.go#L99))

bynn (approved):

PR #6130, opened 2022-12-12
EVG-16663 Suppress general notifications for aborted patches

malikchaya2:
[EVG-16663](https://jira.mongodb.org/browse/EVG-16663)

### Description 
Users do not wish to receive notifications for aborted patches from the general patch notification setting. 
image

The only notification related to patch outcomes  that someone can sign up for is patch-finish, which is handled by the function changed in this PR. So they will continue to receive notifications if they explicitly sign up for it by clicking on notify me on that specific version page. 

### Testing 
Tested the behavior on staging.

### DIFF
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -104,7 +104,7 @@ func (t *patchTriggers) patchOutcome(sub *event.Subscription) (*notification.Not
 		if successOutcome || failureOutcome || anyOutcome {
 			aborted, err := model.IsAborted(t.patch.Id.Hex())
 			if err != nil {
-				return nil, errors.Errorf("getting aborted status for patch '%s'", t.patch.Id.Hex())
+				return nil, errors.Wrapf(err, "getting aborted status for patch '%s'", t.patch.Id.Hex())
 			}
 			if aborted {
 				return nil, nil
@@ -383,10 +383,10 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
 		return nil, nil
 	}
 
-	//Don't notify the user of the patch outcome if they aborted the patch
+	// Don't notify the user of the patch outcome if they aborted the patch
 	aborted, err := model.IsAborted(t.patch.Id.Hex())
 	if err != nil {
-		return nil, errors.Errorf("getting aborted status for patch '%s'", t.patch.Id.Hex())
+		return nil, errors.Wrapf(err, "getting aborted status for patch '%s'", t.patch.Id.Hex())
 	}
 	if aborted {
 		return nil, nil
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -382,6 +382,16 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
 	if t.event.EventType != event.PatchChildrenCompletion {
 		return nil, nil
 	}
+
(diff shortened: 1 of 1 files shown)


trigger/patch.go
```diff
@@ -382,6 +382,16 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
 	if t.event.EventType != event.PatchChildrenCompletion {
 		return nil, nil
 	}
+
+	//Don't notify the user of the patch outcome if they aborted the patch
```
Kimchelly, line 386: nit: space between `//` and `Don't`.

trigger/patch.go
```diff
@@ -382,6 +382,16 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
...
+	aborted, err := model.IsAborted(t.patch.Id.Hex())
+	if err != nil {
+		return nil, errors.Errorf("getting aborted status for patch '%s'", t.patch.Id.Hex())
```
Kimchelly, line 389: Seems like this should be `errors.Wrap(err, "getting aborted status...)`?

trigger/patch.go
```diff
@@ -382,6 +382,16 @@ func (t *patchTriggers) patchFamilyOutcome(sub *event.Subscription) (*notificati
 	if t.event.EventType != event.PatchChildrenCompletion {
 		return nil, nil
 	}
+
+	//Don't notify the user of the patch outcome if they aborted the patch
+	aborted, err := model.IsAborted(t.patch.Id.Hex())
```
Kimchelly, line 387: Question for my own understanding: since this is relying on the `(Version).Aborted` field, does the version aborted field indicate that _any_ task has been aborted, or that _all_ remaining tasks to run are aborted? For example, let's say you abort single running task in the patch, but leave the rest of the ones still running alone. When the remaining tasks finish, will this suppress the notification? I imagine we would only want to suppress patch aborted notifications only in the case where the user aborted all the remaining tasks that aren't done running.

I glanced through [the update logic for version statuses](https://github.com/evergreen-ci/evergreen/blob/84603e68a746dd779161f2a089347761f1cbbb6b/model/task_lifecycle.go#L1257-L1269), and [for build statuses](https://github.com/evergreen-ci/evergreen/blob/299bf20c8da9e353b1097183672cd5169ff3447d/model/task_lifecycle.go#L1148-L1163). It seems like it follow the former definition (i.e. any task abort counts as abort), but I could be wrong.

Kimchelly (changes requested):

trigger/patch.go
malikchaya2: done

trigger/patch.go
malikchaya2: done

trigger/patch.go
malikchaya2: If I understand correctly, it's the latter. 

But I'll test this to make sure and update back here.

malikchaya2: I can confirm. I aborted a task [here](https://spruce.mongodb.com/version/6399e8e5850e613cd70aeab1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and the version's aborted field remained false.

Kimchelly (approved):

PR #6132, opened 2022-12-12
EVG-18413 Also deactivate the dependencies of newly deactivated dependencies

hadjri:
[EVG-18413](https://jira.mongodb.org/browse/EVG-18413)

### Description 
In the[ previous PR](https://github.com/evergreen-ci/evergreen/pull/6111) to address this ticket, the change was to not activate generated tasks that depend on explicitly inactive tasks.  This works in the standard cases I was testing for, but it's not sufficient in the more complex case where the tasks that get deactivated _generate further dependencies_ within the same variant.  

Ex:  

If `generator` generates `taskA` in `variantA` which is set `activate: false`, which has a dependency in the YAML on `taskB` in `variantB`, which then has a dependency on `taskB2` in `variantB`, then the previous PR will correctly prevent `taskB` from being activated, but `taskB2` gets activated incorrectly since `taskB` has not been added to the list of inactive tasks yet at the time of processing `taskB2`. 

### Testing 
Added unit test case to cover this scenario. Also tested in staging by adding a new task `shouldNotActivate` that is a dependency of task `dependencyTask`, which is a task that was added to the list of tasks to deactivate.  

Confirmed that before the change, `shouldNotActivate` [gets activated](https://spruce-staging.corp.mongodb.com/version/6397853b97b1d31ac1eb5bcb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), and with the change it [does not](https://spruce-staging.corp.mongodb.com/version/6397950c9dbe32741ac34564/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/model/dependencies.go
+++ b/model/dependencies.go
@@ -143,14 +143,19 @@ func (di *dependencyIncluder) updateDeactivationMap(pair TVPair, generatedVarian
 		// one pair that depends on this new dep being active - so we cannot deactivate it.
 		if _, foundPair := di.deactivateGeneratedDeps[pair]; !foundPair || !pairSpecifiesActivation {
 			di.deactivateGeneratedDeps[pair] = pairSpecifiesActivation
-			di.recursivelyUpdateDeactivationMap(pair, pairSpecifiesActivation)
+			di.recursivelyUpdateDeactivationMap(pair, map[TVPair]bool{}, pairSpecifiesActivation)
 		}
 	}
 }
 
 // recursivelyUpdateDeactivationMap recurses through the full dependencies of a task and updates their value
 // in the deactivateGeneratedDeps based on the pairSpecifiesActivation input.
-func (di *dependencyIncluder) recursivelyUpdateDeactivationMap(pair TVPair, pairSpecifiesActivation bool) {
+func (di *dependencyIncluder) recursivelyUpdateDeactivationMap(pair TVPair, dependencyIncluded map[TVPair]bool, pairSpecifiesActivation bool) {
+	// If we've been here before, return early to avoid infinite recursion and extra work.
+	if dependencyIncluded[pair] {
+		return
+	}
+	dependencyIncluded[pair] = true
 	bvt := di.Project.FindTaskForVariant(pair.TaskName, pair.Variant)
 	if bvt != nil {
 		deps := di.expandDependencies(pair, bvt.DependsOn)
@@ -160,7 +165,7 @@ func (di *dependencyIncluder) recursivelyUpdateDeactivationMap(pair TVPair, pair
 			if _, foundDep := di.deactivateGeneratedDeps[dep]; !foundDep || !pairSpecifiesActivation {
 				di.deactivateGeneratedDeps[dep] = pairSpecifiesActivation
 			}
-			di.recursivelyUpdateDeactivationMap(dep, pairSpecifiesActivation)
+			di.recursivelyUpdateDeactivationMap(dep, dependencyIncluded, pairSpecifiesActivation)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


model/dependencies.go
```diff
@@ -58,6 +58,14 @@ func (di *dependencyIncluder) include(initialDeps []TVPair, activationInfo *spec
 	for pair, addToActivationInfo := range di.deactivateGeneratedDeps {
 		if addToActivationInfo {
 			activationInfo.activationTasks[pair.Variant] = append(activationInfo.activationTasks[pair.Variant], pair.TaskName)
+			bvt := di.Project.FindTaskForVariant(pair.TaskName, pair.Variant)
+			// Also need to deactivate the dependencies of the newly deactivated tasks
+			if bvt != nil {
+				deps := di.expandDependencies(pair, bvt.DependsOn)
```
ablack12, line 64: Wouldn't we want to handle this using the deactivation map instead? I might be thinking about this wrong since it's getting complex but, if taskA depends on taskB which depends on taskB2 (in your example), what if taskB2 _is_ supposed to be activated because something else does depend on it?

hadjri: Introduced a new `isRoot` parameter to the `handle` function which does the work of updating the deactivation map (if `isRoot` is true).

This should work because although `handle` computes dependencies recursively based off an [initialDeps](https://github.com/evergreen-ci/evergreen/blob/main/model/dependencies.go#L44) list provided by the generate tasks function, the only tasks that could have their activation initially set are the initialDeps, so we only need to propagate the `deactivateGeneratedDeps` at the root level of each recursion, as long as we update the map for the full recursive list of dependencies that need to exist to support the initial dependency.

model/dependencies.go
```diff
@@ -127,15 +131,25 @@ func (di *dependencyIncluder) handle(pair TVPair, activationInfo *specificActiva
...
 		// If deactivateGeneratedDeps[pair] = false it signifies that there was at least
 		// one pair that depends on this new dep being active - so we cannot deactivate it.
-		if _, ok := di.deactivateGeneratedDeps[dep]; !ok || !pairSpecifiesActivation {
-			di.deactivateGeneratedDeps[dep] = pairSpecifiesActivation
+		if _, foundPair := di.deactivateGeneratedDeps[pair]; !foundPair || !pairSpecifiesActivation {
+			di.deactivateGeneratedDeps[pair] = pairSpecifiesActivation
+			bvt := di.Project.FindTaskForVariant(pair.TaskName, pair.Variant)
+			// Also need to deactivate the dependencies of the newly deactivated tasks
+			if bvt != nil {
+				deps := di.expandDependencies(pair, bvt.DependsOn)
```
ablack12, line 146: Can you remind me why we don't need to recurse on this?  We may also want to comment aggressively because we're not gonna remember how this works in a month i think haha

model/dependencies.go
```diff
@@ -115,8 +117,10 @@ func (di *dependencyIncluder) handle(pair TVPair, activationInfo *specificActiva
 	// also mark this newly generated dependency as inactive.
 	pairSpecifiesActivation := activationInfo.taskOrVariantHasSpecificActivation(pair.Variant, pair.TaskName)
 	for _, dep := range deps {
-		di.updateDeactivationMap(activationInfo, dep, generatedVariants, pairSpecifiesActivation)
-		ok, err := di.handle(dep, activationInfo, generatedVariants)
+		if isRoot {
```
ablack12, line 120: we love a basecase! We may also want to comment here about why we only want to do this for the root case though

ablack12 (approved):
I don't 100% understand yet but with more comments and assuming test coverage is good this LGTM 👍

model/dependencies.go
hadjri: We do, thanks for the catch. `expandDependencies` only finds dependencies at the first level which I missed.  Updated this to recursion and added another unit test case to confirm that the full dependency tree gets its deactivation value updated.

hadjri:
Tested this again in staging also

PR #6134, opened 2022-12-13
EVG-18474 give amboy its own db client

ybrill:
[EVG-18474](https://jira.mongodb.org/browse/EVG-18474)

### Description 
In preparation for bifurcating Evergreen and Amboy's databases we need to pass amboy a distinct db client. For the moment it will continue to point at the same cluster as evergreen, but once we're ready to cut over to the Atlas cluster we'll point it there. The cutover will also involve adding auth, so those options are added too.

Before this gets deployed we'll need to set the `db` field on the amboy config in prod's db to point to the current db. After it gets deployed we might want to clean up the supplanted "database" field.

Some considerations:
- The current client (database URI, auth) is configured from configuration files placed on the app server by chef. This seems like it's only necessary for bootstrapping the evergreen application. Once there's a connection to evergreen's db we can pull the parameters for connecting to amboy from there.
- The evergreen db client allows configuration of its write/read concern. It didn't seem worth adding separate settings for the amboy client since we've been using "majority" for forever. We could.
- The APM monitor will spit out separate messages for amboy that will only be distinguishable by the collection names. I don't think this will break any dashboards. If we want them to remain combined we could use a single instance of the logging monitor, or if we want them to be really distinguishable we could add another field to the message it's logging.

### Testing 
In staging, amboy ran jobs.


environment.go
```diff
@@ -393,12 +414,12 @@ func (e *envState) createLocalQueue(ctx context.Context) error {
 	return nil
 }
 
-func (e *envState) createApplicationQueue(ctx context.Context) error {
+func (e *envState) createApplicationQueue(ctx context.Context, client *mongo.Client) error {
 	// configure the remote mongodb-backed amboy
 	// queue.
 	opts := queue.DefaultMongoDBOptions()
-	opts.URI = e.settings.Database.Url
```
ybrill, line 400: It seemed the URI is unused when a client is passed.

config_amboy.go
```diff
@@ -17,7 +17,7 @@ import (
 type AmboyConfig struct {
 	Name                                  string                  `bson:"name" json:"name" yaml:"name"`
 	SingleName                            string                  `bson:"single_name" json:"single_name" yaml:"single_name"`
-	DB                                    string                  `bson:"database" json:"database" yaml:"database"`
+	DB                                    AmboyDBConfig           `bson:"db" json:"db" yaml:"db"`
```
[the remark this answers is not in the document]
ybrill: I changed it to `DBConnection`. Does that work?
It seemed awkward to have `AmboyConnectionConfig` inside an `AmboyConfig`.

rest/model/admin.go
```diff
@@ -355,7 +355,7 @@ func (a *APISMTPConfig) ToService() (interface{}, error) {
 type APIAmboyConfig struct {
 	Name                                  *string                    `json:"name"`
 	SingleName                            *string                    `json:"single_name"`
-	DB                                    *string                    `json:"database"`
+	DBConnection                          APIAmboyDBConfig           `json:"db"`
```
Kimchelly, line 358: Update JSON tag.

Kimchelly (approved):
LGTM mod JSON tag.

rest/model/admin.go
ybrill: done!

ybrill:
evergreen retry

Kimchelly (approved):
LGTM all the updates since my last LGTM.

For future reader context, the read preference is being explicitly set to primary because that's the only operating mode where Amboy could expect to function properly. The possibility of reading stale data isn't really an option.

PR #6135, opened 2022-12-13
EVG-18552: Remove test stats API code

julianedwards:
JIRA: https://jira.mongodb.org/browse/EVG-18552

### Description 
This removes the test stats API and all related code. This also stops sending Mongo test results to the Evergreen DB.

### Testing 
Unit tests and staging. In staging I also tested it with the [changes I made to Spruce](https://github.com/evergreen-ci/spruce/pull/1577) to ensure I did not break the UI.


julianedwards:
FYI the downstream spruce e2e tests will fail until [this commit](https://github.com/evergreen-ci/spruce/commit/c65de77b0eca94775f737fa16121d073feff7e4d) is deployed to prod.

service/templates/projects.html
```diff
@@ -999,41 +999,6 @@ <h3>Virtual Workstations Commands</h3>
...
-            <div class="col-lg-5 col-header">
-              <label class="control-label">File patterns to ignore</label>
-            </div>
-          </div>
-          <div class="form-group" ng-repeat="(index, pattern) in settingsFormData.files_ignored_from_cache">
-            <div class="col-lg-4">
-              <label class="control-label">[[pattern]]</label>
-            </div>
-            <div class="col-lg-2">
-              <button class="btn btn-default btn-danger" type="button" ng-click="removeCacheIgnoreFile(index)">
```
ybrill, line 1021: we should probably also remove the corresponding bits from projects.js, like the removeCacheIgnoreFile function.

agent/command/results_utils.go
```diff
@@ -29,14 +29,6 @@ func sendTestResults(ctx context.Context, comm client.Communicator, logger clien
 	logger.Task().Info("Attaching test results...")
 	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
 
-	// TODO (PM-2940): Stop sending Mongo project test results to the
-	// database once they can support Presto test results.
-	if model.IsServerResmokeProject(conf.ProjectRef.Identifier) {
-		if err := comm.SendTestResults(ctx, td, results); err != nil {
```
ybrill, line 35: could the `SendTestResults` method be removed from the communicator? (and the corresponding route on the server side)?

model/stats/query.go
```diff
@@ -176,23 +148,8 @@ func (f *StatsFilter) validateDates() error {
...
-	catcher.NewWhen(len(f.Tests) == 0 && len(f.Tasks) == 0, "missing tests or tasks")
-
-	return catcher.Resolve()
-}
-
-// ValidateForTasks use with test stats validates that the StatsFilter struct is valid for use with task stats.
+// ValidateForTasks validates that the StatsFilter struct is valid for use with
+// task stats.
 func (f *StatsFilter) ValidateForTasks() error {
 	catcher := grip.NewBasicCatcher()
```
ybrill, line 155: should the `ValidateCommon`s be merged into their respective validate functions?

ybrill (commented):
Could `model/testresult/*` go too?

julianedwards:
> Could model/testresult/* go too?

That is going to be a separate ticket! But yes, soon.

rest/route/stats.go
```diff
@@ -531,20 +301,7 @@ func (tsh *taskStatsHandler) Parse(ctx context.Context, r *http.Request) error {
 }
 
 func (tsh *taskStatsHandler) Run(ctx context.Context) gimlet.Responder {
-	flags, err := evergreen.GetServiceFlags()
```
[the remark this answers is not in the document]
julianedwards: I didn't think so. @ablack12 do you think we still need the service flag to disable task stats endpoint? Has it ever caused us an issue?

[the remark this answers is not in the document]
ablack12: I feel like we do just turn it off when things are bad because stats are expensive but idk if task_stats has been specifically the issue since we just turn off both. I think we might as well leave it if we can.

agent/command/results_utils.go
julianedwards: I will do that when I remove the test results code in a subsequent ticket.

units/crons_remote_hour.go
```diff
@@ -46,7 +46,7 @@ func (j *cronsRemoteHourJob) Run(ctx context.Context) {
 	}
 
 	ops := []amboy.QueueOperation{
-		PopulateCacheHistoricalTestDataJob(2),
+		PopulateCacheHistoricalTaskDataJob(2),
```
ybrill, line 49: now that it's not calculating hourly stats maybe we only need it to run once a day?

julianedwards: I am nervous that this would end up being a lot of work for the database at once if we do it this way.

model/reliability/query.go
```diff
@@ -13,25 +13,25 @@ import (
...
 
 // TaskReliabilityFilter represents search and aggregation parameters when querying the test or task statistics.
 type TaskReliabilityFilter struct {
-	stats.StatsFilter
+	taskstats.StatsFilter
 	Significance float64
 }
 
-// ValidateForTaskReliability validates that the StartAt struct is valid for use with test stats.
+// ValidateForTaskReliability validates that the StartAt struct is valid for use with test taskstats.
```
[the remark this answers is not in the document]
julianedwards: Good catch!

ybrill (approved):

PR #6137, opened 2022-12-13
EVG-18502 Update validate build variant fields to handle task groups

bynn:
[EVG-18502](https://jira.mongodb.org/browse/EVG-18502)

### Description 
we werent checking task groups when validating projects

### Testing 
unit tests

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -666,17 +666,15 @@ func validateBVFields(project *model.Project) ValidationErrors {
 		for _, task := range buildVariant.Tasks {
 			taskHasValidDistro := validateRunOn(task.RunOn)
 			if !taskHasValidDistro && task.IsGroup {
-				taskGroupTasksHaveDistro := true
 				for _, t := range project.FindTaskGroup(task.Name).Tasks {
 					pt := project.FindProjectTask(t)
 					if pt != nil {
-						if !validateRunOn(pt.RunOn) {
-							taskGroupTasksHaveDistro = false
+						if validateRunOn(pt.RunOn) {
+							taskHasValidDistro = true
 							break
 						}
 					}
 				}
-				taskHasValidDistro = taskGroupTasksHaveDistro
 			} else if !taskHasValidDistro {
 				// check for a default in the task definition
 				pt := project.FindProjectTask(task.Name)
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -614,6 +614,15 @@ func validateProjectConfigPlugins(pc *model.ProjectConfig) ValidationErrors {
 	return errs
 }
 
+func validateRunOn(runOn []string) bool {
+	for _, d := range runOn {
+		if d != "" {
+			return true
+		}
+	}
+	return false
+}
+
 // Ensures that the project has at least one buildvariant and also that all the
 // fields required for any buildvariant definition are present
 func validateBVFields(project *model.Project) ValidationErrors {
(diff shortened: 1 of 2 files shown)


validator/project_validator.go
```diff
@@ -662,7 +662,21 @@ func validateBVFields(project *model.Project) ValidationErrors {
 					break
 				}
 			}
-			if !taskHasValidDistro {
+			if !taskHasValidDistro && task.IsGroup {
```
ablack12, line 665: If we re-factor the logic from 658 to 689 to a lil helper function, we could just return true on taskHasValidDistro instead of setting to true and casing on it each time

validator/project_validator.go
```diff
@@ -662,7 +662,21 @@ func validateBVFields(project *model.Project) ValidationErrors {
...
+				taskGroupTasksHasDistro := true
+				for _, t := range project.FindTaskGroup(task.Name).Tasks {
+					pt := project.FindProjectTask(t)
+					if pt != nil {
+						for _, d := range pt.RunOn {
+							if d == "" {
```
ablack12, line 671: is there a reason why the `if d != "" {set to true}` logic won't work here? The false case is actually most likely to happen when run_on is completely empty which isn't captured here (and should be captured by the test if it's not already)

ablack12 (changes requested):
As part of this ticket I think we also need to verify that run_on for task_group tasks _is_ respected by the existing code, which you might be able to do in a unit test somewhere or could make a staging task group that defines run_on and verify that the task group task does use the distro defined in it's task instead of in the build variant.

validator/project_validator.go
bynn: done

validator/project_validator.go
bynn: it was because we need every task in the group to have a valid runOn if it's not specified in the bv or group instead of just checking 1 but i can work around it

bynn:
https://spruce-staging.corp.mongodb.com/version/639b6fb09dbe327c3b57bdc5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

task group uses the specified run_on and specifying it in the task itself will also override that as expected 

didnt write new tests for it because it seemed out of scope for the validate ticket

validator/project_validator.go
```diff
@@ -614,6 +614,15 @@ func validateProjectConfigPlugins(pc *model.ProjectConfig) ValidationErrors {
 	return errs
 }
 
+func validateRunOn(runOn []string) bool {
```
ablack12, line 617: nit: maybe hasValidRunOn since that makes more sense to return a boolean

validator/project_validator.go
```diff
@@ -655,24 +664,23 @@ func validateBVFields(project *model.Project) ValidationErrors {
...
-					break
-				}
-			}
-			if !taskHasValidDistro {
-				// check for a default in the task definition
-				pt := project.FindProjectTask(task.Name)
-				if pt != nil {
-					for _, d := range pt.RunOn {
-						if d != "" {
+			taskHasValidDistro := validateRunOn(task.RunOn)
```
ablack12, line 667: if taskHasValidDistro we can break

validator/project_validator.go
```diff
@@ -655,24 +664,23 @@ func validateBVFields(project *model.Project) ValidationErrors {
...
 				}
+			} else if !taskHasValidDistro {
+				// check for a default in the task definition
+				pt := project.FindProjectTask(task.Name)
+				if pt != nil {
+					taskHasValidDistro = validateRunOn(pt.RunOn)
+				}
 			}
 			if !taskHasValidDistro {
 				hasTaskWithoutDistro = true
```
ablack12, line 686: Sorry I'm just asking you to refactor this now but we don't need hasTaskWithoutDistro either, since we just break when we find it; we can just move the error at 691 into this and _then_ break

ablack12 (changes requested):

validator/project_validator.go
bynn: done

validator/project_validator.go
bynn: all good and tests pass

validator/project_validator.go
ablack12: don't need to check taskHasValidDistro here or in the else since we break for it above

ablack12 (approved):
LGTM with nit

PR #6142, opened 2022-12-14
EVG-18518: produce diffable YAMLs via evergreen evaluate

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18518

### Description 
Apologies for the long description, but this change has a bunch of context. The tl;dr is that `evergreen evaluate` now produces output consistently and also outputs more fields.

This is a follow-on to #6002 and is related to #6017. The downside of #6017 was that `evergreen validate` can only check if the upgraded YAML passes validation, but can't check that some runtime values (like expansions) would be exactly the same pre-upgrade and post-upgrade. Unfortunately, the YAML upgrade _may_ change the YAML parsing behavior such that the YAML is still syntactically valid, just different. Users would be most interested in knowing if their YAML is parsed _differently_, rather than simply knowing if their YAML is valid, because this can result in different behavior when the task actually runs.

This change provides a temporary tool that makes it easy for users to figure out exactly "what will change" for a given project YAML after the YAML upgrade in [EVG-17291](https://jira.mongodb.org/browse/EVG-17291) rolls out by adding a temporary feature enhancement to `evergreen evaluate`. This enhancement makes it easier to produce a side-by-side file diff of the pre-upgraded and post-upgraded YAML (similar to a `git diff` of changes, except for YAML). `evergreen evaluate` already produces the YAML after things such as Evergreen tags and YAML aliases are resolved, but currently does so in such a way that shuffles the ordering of some YAML fields (like the order of the `buildvariants:` section). For example, if you ran `evergreen evaluate` before this change, the build variants order was shuffled, so running `evergreen evaluate` multiple times on the same file would produce differently-ordered YAML results. For example:
```yaml
# Original YAML lists build variants in order: bv0, bv1, bv2
buildvariants:
    - name: bv0
    - name: bv1
    - name: bv2
```
```yaml
# evergreen evaluate YAML result lists build variants in a different order:
buildvariants:
    - name: bv2
    - name: bv0
    - name: bv1
```
This makes it difficult to compare two results from `evergreen evaluate` because it does not produce consistent output for the same input. I enhanced `evergreen evaluate` to ensure that it will produce the same output consistently, thus enabling users to "diff" what will happen pre-upgrade and post-upgrade for their YAML.

* Add `--use_upgraded_yaml` flag to `evergreen evaluate` to use the upgraded YAML version rather than the current one.
* Add `--diffable` flag to `evergreen evaluate` to produce side-by-side diffable YAMLs. This is achieved by doing the following:
    * Sort fields in `evergreen evaluate` according to their name for cases where it uses unordered lists.
    * Ensure all fields that can contain nested maps are included in `evergreen evaluate`. These fields are included because nested map aliases are the ones affected by the YAML upgrade.

### Testing 
I didn't think it was worth adding automated tests because it's a temporary feature until the YAML upgrade is rolled out. Instead, I tested this manually on our self-tests YAML and on the server YAML. For example, for the server's YAML, I ran this:
```sh
# Outputs the parsed YAML pre-upgrade
evergreen evaluate --diffable etc/evergreen.yml

# Outputs the parsed YAML post-upgrade
evergreen evaluate --use_upgraded_yaml --diffable etc/evergreen.yml
```

I verified the two desired properties:
1. `evergreen evaluate` always produces the same output for the same exact input.
2. The fields are sorted so that using a diffing tool (like `diff` or `vimdiff`) produces side-by-side results that show _only_ the different fields.

### DIFF
--- a/operations/evaluate.go
+++ b/operations/evaluate.go
@@ -72,12 +72,22 @@ func Evaluate() cli.Command {
 				}{}
 				if showTasks {
 					tmp.Functions = p.Functions
-					tmp.Tasks = p.Tasks
+					if diffable {
+						sortTasksByName := model.ProjectTasksByName(p.Tasks)
+						sort.Sort(sortTasksByName)
+						tmp.Tasks = sortTasksByName
+					} else {
+						tmp.Tasks = p.Tasks
+					}
 				}
 				if showVariants {
-					sortByName := model.BuildVariantsByName(p.BuildVariants)
-					sort.Sort(sortByName)
-					tmp.Variants = sortByName
+					if diffable {
+						sortBVsByName := model.BuildVariantsByName(p.BuildVariants)
+						sort.Sort(sortBVsByName)
+						tmp.Variants = sortBVsByName
+					} else {
+						tmp.Variants = p.BuildVariants
+					}
 				}
 				out = tmp
 			} else if diffable {
--- a/operations/evaluate.go
+++ b/operations/evaluate.go
@@ -4,17 +4,20 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"sort"
 
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 	yaml "gopkg.in/20210107192922/yaml.v3"
+	upgradedYAML "gopkg.in/yaml.v3"
 )
 
 func Evaluate() cli.Command {
 	const (
 		taskFlagName     = "tasks"
 		variantsFlagName = "variants"
+		diffableFlagName = "diffable"
 	)
 
 	return cli.Command{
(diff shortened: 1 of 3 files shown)


operations/evaluate.go
```diff
@@ -28,12 +31,21 @@ func Evaluate() cli.Command {
...
+				Usage: "show the project configuration in an ordered, diff-friendly format",
+			},
+		),
 		Before: mergeBeforeFuncs(requirePathFlag),
 		Action: func(c *cli.Context) error {
 			path := c.String(pathFlagName)
 			showTasks := c.Bool(taskFlagName)
 			showVariants := c.Bool(variantsFlagName)
+			useUpgradedYAML := c.Bool(useUpgradedYAMLFlagName)
+			diffable := c.Bool(diffableFlagName)
```
malikchaya2, line 48: It seems like this is the preferred state, or at the very least, not a downgrade to the regular state. Why not make all the output diffable instead of making it flag dependent? Is it because it's substantially slower?

malikchaya2 (changes requested):

operations/evaluate.go
Kimchelly: I don't want to change the existing behavior of the function because it's a temporary feature that I don't really think is worth maintaining, and because people use `evergreen evaluate` in their scripts. Producing a different format would probably break those scripts.

malikchaya2: Wouldn't it only help scripts by giving the results in order?

Kimchelly: It would, but we would have to maintain the promise of sorted fields. I don't think we would really want to maintain that promise in the long-term unless someone actually asks for it.

malikchaya2 (approved):

PR #6143, opened 2022-12-14
EVG-18381 Refactor Github child patch subscriptions

malikchaya2:
[EVG-18381](https://jira.mongodb.org/browse/EVG-18381)

### Description 
We no longer need to subscribe on all github pr child patches so that we can wait on it. We can use the family-outcom trigger instead. 

### Testing 
[Tested on staging ](https://github.com/evergreen-ci/commit-queue-sandbox/pull/376)

### DIFF
--- a/model/event/subscribers.go
+++ b/model/event/subscribers.go
@@ -190,11 +190,6 @@ type GithubPullRequestSubscriber struct {
 	Type     string `bson:"type"`
 }
 
-const (
-	WaitOnChild           = "wait-on-child"
-	SendChildPatchOutcome = "send-child-patch-outcome"
-)
-
 func (s *GithubPullRequestSubscriber) String() string {
 	return fmt.Sprintf("%s-%s-%d-%s-%s", s.Owner, s.Repo, s.PRNumber, s.Ref, s.ChildId)
 }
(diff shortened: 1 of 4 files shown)


ablack12 (commented):
I haven't reviewed most of these PRs so I'm a little bit behind; is it correct that even regular patches will use this FamilyOutcome instead of just TriggerOutcome? Will already existing subscriptions work for this? (Not sure if that's the migration that you did, I also notice that we have all the triggers added at the patch/etc level so maybe it's that) Also can we add some words somewhere (NewExpiringPatchOutcomeSubscription or where the triggers are defined) to explain family vs non-family?

malikchaya2:
> I haven't reviewed most of these PRs so I'm a little bit behind; is it correct that even regular patches will use this FamilyOutcome instead of just TriggerOutcome?

Yes

> Will already existing subscriptions work for this? (Not sure if that's the migration that you did, I also notice that we have all the triggers added at the patch/etc level so maybe it's that) 

It won't work for in-progress PRs that have child patches (I believe only our team) - they will have to be restarted. Anything else will not be affected. 

> Also can we add some words somewhere (NewExpiringPatchOutcomeSubscription or where the triggers are defined) to explain family vs non-family? 

I added it to the trigger definitions

ablack12 (approved):

PR #6152, opened 2022-12-20
EVG-17302 Do not allow aliases when --reuse or --repeat failed is used

malikchaya2:
[EVG-17302](https://jira.mongodb.org/browse/EVG-17302)

### Description 
Do not allow aliases when --reuse or --repeat failed is used.

### Testing 
Tested locally

### DIFF
--- a/operations/patch.go
+++ b/operations/patch.go
@@ -178,8 +178,8 @@ func Patch() cli.Command {
 			}
 			params.Description = params.getDescription()
 
-			if (params.RepeatDefinition || params.RepeatFailed) && (len(params.Tasks) > 0 || len(params.Variants) > 0) {
-				return errors.Errorf("can't define tasks/variants when reusing previous patch's tasks and variants")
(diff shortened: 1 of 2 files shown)


Kimchelly (approved):

operations/patch.go
```diff
@@ -178,8 +178,8 @@ func Patch() cli.Command {
 			}
 			params.Description = params.getDescription()
 
-			if (params.RepeatDefinition || params.RepeatFailed) && (len(params.Tasks) > 0 || len(params.Variants) > 0) {
-				return errors.Errorf("can't define tasks/variants when reusing previous patch's tasks and variants")
+			if (params.RepeatDefinition || params.RepeatFailed) && (len(params.Tasks) > 0 || len(params.Variants) > 0 || len(params.Alias) > 0) {
```
Kimchelly, line 181: Actually, I thought it over again - does this have to check RegexVariants and RegexTasks are both empty as well? They might be populated if they specify a local alias.

malikchaya2: good catch! fixed.

PR #6154, opened 2022-12-21
EVG-12434 Expose ResetWhenFinished field

khelif96:
[EVG-12434](https://jira.mongodb.org/browse/EVG-12434)

### DIFF
--- a/graphql/schema/types/task.graphql
+++ b/graphql/schema/types/task.graphql
@@ -76,6 +76,7 @@ type Task {
   projectId: String!
   projectIdentifier: String
   requester: String!
+  resetWhenFinished: Boolean!
   revision: String
   scheduledTime: Time
   spawnHostLink: String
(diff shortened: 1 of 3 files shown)


graphql/schema/types/task.graphql
```diff
@@ -76,6 +76,7 @@ type Task {
   projectId: String!
   projectIdentifier: String
   requester: String!
+  resetWhenFinished: Boolean!
```
minnakt, line 79: It could be nice to test for this field in a small GraphQL test

minnakt (approved):

graphql/schema/types/task.graphql
khelif96: I had intentionally omitted writing a test for this field. Since the field has no underlying logic/resolver and is just inherited from an `apiTask` it would panic if it was not defined.

PR #6155, opened 2022-12-28
EVG-7440 Allow updating github statuses for multiple builds at once

hadjri:
[EVG-7440](https://jira.mongodb.org/browse/EVG-7440)

### Description 

### Testing 
Tested in staging. Confirmed that restarting a PR version changes all builds to status "pending" in the GitHub UI, and that manually restarting two or more separate builds accurately changes each corresponding build variant in the GitHub UI to "pending" as well.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -409,6 +409,19 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 			event.LogTaskRestarted(t.Id, t.Execution, caller)
 		}
 	}
+
+	buildIdsMap := map[string]bool{}
+	var buildIds []string
+	for _, t := range tasksToRestart {
+		buildIdsMap[t.BuildId] = true
+	}
+	for buildId := range buildIdsMap {
+		buildIds = append(buildIds, buildId)
+	}
+	if err = UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
+		return errors.Wrapf(err, "updating build and version status for version '%s'", versionId)
+	}
+
 	if err = build.SetBuildStartedForTasks(tasksToRestart, caller); err != nil {
 		return errors.Wrap(err, "setting builds started")
 	}
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1283,10 +1283,27 @@ func updateVersionStatus(v *Version) (string, error) {
 	return versionStatus, nil
 }
 
+// UpdatePatchStatus updates the status of a patch. For PR patches, the build associated
+// with the passed in buildVariant has its GitHub status updated.
 func UpdatePatchStatus(p *patch.Patch, versionStatus, buildVariant string) error {
 	patchStatus, err := evergreen.VersionStatusToPatchStatus(versionStatus)
 	if err != nil {
 		return errors.Wrapf(err, "getting patch status from version status '%s'", versionStatus)
+	}
+
+	if p.IsGithubPRPatch() && !evergreen.IsFinishedPatchStatus(patchStatus) {
+		input := thirdparty.SendGithubStatusInput{
+			VersionId: p.Id.Hex(),
+			Owner:     p.GithubPatchData.BaseOwner,
+			Repo:      p.GithubPatchData.BaseRepo,
+			Ref:       p.GithubPatchData.HeadHash,
+			Desc:      "patch status change",
+			Caller:    "pr-task-reset",
+			Context:   fmt.Sprintf("evergreen/%s", buildVariant),
+		}
+		if err = thirdparty.SendVersionStatusToGithub(input); err != nil {
+			return errors.Wrapf(err, "sending patch '%s' status to Github", p.Id.Hex())
+		}
 	}
 
 	if patchStatus == p.Status {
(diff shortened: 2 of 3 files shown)


model/task_lifecycle.go
```diff
@@ -1429,17 +1434,16 @@ func UpdateVersionAndPatchStatusForBuilds(buildIds []string) error {
...
-		versionsToUpdate[build.Version] = build.Id
+		versionsToUpdate[build.Version] = append(versionsToUpdate[build.Version], build.Id)
 		bvMap[build.Id] = build.BuildVariant
 	}
-	for versionId, buildId := range versionsToUpdate {
+	for versionId, buildList := range versionsToUpdate {
 		buildVersion, err := VersionFindOneId(versionId)
 		if err != nil {
-			return errors.Wrapf(err, "getting version '%s' for build '%s'", versionId, buildId)
+			return errors.Wrapf(err, "getting version '%s' for build '%s'", versionId, buildList[0])
```
bynn, line 1443: could just returning the first element of the buildList be misleading sometimes?

hadjri: Sure, I'm using a `strings.Join` now instead.

model/task_lifecycle.go
```diff
@@ -1418,7 +1422,9 @@ func UpdateVersionAndPatchStatusForBuilds(buildIds []string) error {
 		return errors.Wrapf(err, "fetching builds")
 	}
 
-	versionsToUpdate := make(map[string]string)
+	// Maintain a list of builds for each version because we may
+	// be updating many builds for the same version.
+	versionsToUpdate := make(map[string][]string)
```
ablack12, line 1427: nit: Maybe versionBuildsMap now instead to make it clear what the mapping was? The comment confused me at first since I didn't know where builds became relevant

model/task_lifecycle.go
```diff
@@ -1429,17 +1435,16 @@ func UpdateVersionAndPatchStatusForBuilds(buildIds []string) error {
...
-		versionsToUpdate[build.Version] = build.Id
+		versionsToUpdate[build.Version] = append(versionsToUpdate[build.Version], build.Id)
 		bvMap[build.Id] = build.BuildVariant
 	}
-	for versionId, buildId := range versionsToUpdate {
+	for versionId, buildList := range versionsToUpdate {
 		buildVersion, err := VersionFindOneId(versionId)
 		if err != nil {
-			return errors.Wrapf(err, "getting version '%s' for build '%s'", versionId, buildId)
+			return errors.Wrapf(err, "getting version '%s' for builds '%s'", versionId, strings.Join(buildList, ","))
```
ablack12, line 1444: This is kind of an unruly error and shouldn't happen anyway so I'd preference just saying "getting version '%s'"

model/task_lifecycle.go
```diff
@@ -1454,8 +1459,10 @@ func UpdateVersionAndPatchStatusForBuilds(buildIds []string) error {
 			if p == nil {
 				return errors.Errorf("no patch found for version '%s'", buildVersion.Id)
 			}
-			if err = UpdatePatchStatus(p, newVersionStatus, bvMap[buildId]); err != nil {
-				return errors.Wrapf(err, "updating patch '%s' status", p.Id.Hex())
+			for _, buildId := range versionsToUpdate[versionId] {
```
ablack12, line 1462: can we iterate through buildList instead?

model/lifecycle.go
```diff
@@ -409,6 +409,19 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
+		buildIdsMap[t.BuildId] = true
+	}
+	for buildId := range buildIdsMap {
+		buildIds = append(buildIds, buildId)
+	}
+	if err = UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
+		return errors.Wrapf(err, "updating build and version status for version '%s'", versionId)
+	}
+
 	if err = build.SetBuildStartedForTasks(tasksToRestart, caller); err != nil {
```
ablack12, line 425: Would we need to call this before UpdateVersionAndPatchStatusForBuilds so that the builds are in the right state or?

model/task_lifecycle.go
```diff
@@ -1283,12 +1284,29 @@ func updateVersionStatus(v *Version) (string, error) {
...
 
+// UpdatePatchStatus updates the status of a patch. For PR patches, the build associated
+// with the passed in buildVariant has its GitHub status updated.
 func UpdatePatchStatus(p *patch.Patch, versionStatus, buildVariant string) error {
 	patchStatus, err := evergreen.VersionStatusToPatchStatus(versionStatus)
 	if err != nil {
 		return errors.Wrapf(err, "getting patch status from version status '%s'", versionStatus)
 	}
 
+	if p.IsGithubPRPatch() && !evergreen.IsFinishedPatchStatus(patchStatus) {
```
ablack12, line 1295: Why do we do this before the check at 1310? If that's true then I would think that this shouldn't be sent.

ablack12: I guess I just don't know why we moved it. Was this the change to fix restarts?

model/task_lifecycle.go
hadjri: Sure!

model/lifecycle.go
hadjri: It doesn't matter because `UpdateVersionAndPatchStatusForBuilds` will only update the status of builds that have had their [statuses changed](https://gist.github.com/hadjri/f1287d20a311464513c465bf7767336f#file-task_lifecycle-go-L1353) by checking the state of the tasks (which have just been reset above).

model/task_lifecycle.go
ablack12: Oh I see -- if we send this by build variant, then should this just be part of UpdateBuildStatus instead of UpdatePatchStatus? Doesn't seem right for it to be related to patch.

model/lifecycle.go
hadjri: Ok, looks like they can be safely removed.

hadjri: Actually, only `setVersionStatus` can be safely removed because  `SetBuildStartedForTasks` performs an activation update that is necessary, which is something that `updateBuildStatus` does not do.

model/task_lifecycle.go
```diff
@@ -1118,6 +1118,33 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
 	return b.UpdateGithubCheckStatus(buildStatus.status)
 }
 
+// checkUpdateBuildPrStatus checks if the build is coming from a PR, and if so
+// sends its updated status to GitHub to reflect the status of the build.
+func checkUpdateBuildPrStatus(b *build.Build) error {
```
ablack12, line 1123: nit: checkUpdateBuildPRStatus (since PR is an acronym)

model/task_lifecycle.go
```diff
@@ -1118,6 +1118,33 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
...
+	if !evergreen.IsGitHubPatchRequester(b.Requester) {
+		return nil
+	}
+	p, err := patch.FindOneId(b.Version)
```
ablack12, line 1127: should add check for nil patch

ablack12 (approved):
LGTM with nits

bynn (approved):

PR #6156, opened 2022-12-28
EVG-16480 Refactor build restarts to share functionality with RestartVersion

hadjri:
[EVG-16480](https://jira.mongodb.org/browse/EVG-16480)

### Description 
- `RestartVersion` has special logic for restarting single host task group tasks in that instead of archiving/restarting them immediately, it marks the task group to reset when all tasks in the group finish, so they can restart together. `RestartBuild` does not have any of this logic although it should, so I've modularized the logic of `RestartVersion` into some helper functions that can be used by both.  
- Made a change to `AbortTasksForBuild` to mark the in-progress tasks it is aborting to restart when they are finished. This is how it works for `RestartVersion`, whereas when we restart a build we simply abort the in-progress tasks (which results in a partial restart of the build which is incorrect).
- Removed `RestartAllBuildTasks` and replaced its invocations with `RestartBuild`
### Testing 
Updated & added unit tests.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -426,8 +426,8 @@ func restartTasks(allFinishedTasks []task.Task, caller, versionId string) error 
 		}
 	}
 
-	if err := errors.Wrap(build.SetBuildStartedForTasks(allFinishedTasks, caller), "setting builds started"); err != nil {
-		return err
+	if err := build.SetBuildStartedForTasks(allFinishedTasks, caller); err != nil {
+		return errors.Wrap(err, "setting builds started")
 	}
 
 	return errors.Wrap(setVersionStatus(versionId, evergreen.VersionStarted), "changing version status")
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -338,13 +338,40 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
 			return errors.WithStack(err)
 		}
 	}
+	tasksToReset, err := getTasksToReset(taskIds)
+	if err != nil {
+		return errors.Wrap(err, "getting finished tasks")
+	}
+	if len(tasksToReset) == 0 {
+		return nil
+	}
+	err = restartTasks(tasksToReset, caller)
+	if err != nil {
+		return errors.Wrap(err, "restarting tasks")
+	}
+	return errors.Wrap(setVersionStatus(versionId, evergreen.VersionStarted), "changing version status")
+}
+
+// getTasksToReset returns all tasks that should be reset given an initial input list of
+// finished taskIds.
+func getTasksToReset(taskIds []string) ([]task.Task, error) {
+	// Restart all the 'not in-progress' tasks
 	finishedTasks, err := task.FindAll(db.Query(task.ByIdsAndStatus(taskIds, evergreen.TaskCompletedStatuses)))
 	if err != nil {
-		return errors.WithStack(err)
+		return nil, errors.WithStack(err)
 	}
 	allFinishedTasks, err := task.AddParentDisplayTasks(finishedTasks)
 	if err != nil {
-		return errors.WithStack(err)
+		return nil, errors.WithStack(err)
+	}
+	for _, t := range allFinishedTasks {
+		if t.IsPartOfSingleHostTaskGroup() {
+			taskGroupTasks, err := task.FindTaskGroupFromBuild(t.BuildId, t.TaskGroup)
+			if err != nil {
+				return nil, errors.Wrapf(err, "getting tasks for task group '%s'", t.TaskGroup)
+			}
+			allFinishedTasks = append(allFinishedTasks, taskGroupTasks...)
+		}
 	}
 	// Remove execution tasks in case the caller passed both display and execution tasks.
 	// The functions below are expected to work if just the display task is passed.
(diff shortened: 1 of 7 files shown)


model/lifecycle.go
```diff
@@ -338,13 +338,30 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
-		return errors.WithStack(err)
+		return nil, errors.WithStack(err)
 	}
 	allFinishedTasks, err := task.AddParentDisplayTasks(finishedTasks)
 	if err != nil {
-		return errors.WithStack(err)
+		return nil, errors.WithStack(err)
 	}
 	// Remove execution tasks in case the caller passed both display and execution tasks.
 	// The functions below are expected to work if just the display task is passed.
```
ablack12, line 367: Functions below isn't accurate anymore; maybe the function comment should say that it removes execution tasks?

model/lifecycle.go
```diff
@@ -338,13 +338,30 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
+		return errors.Wrap(err, "getting finished tasks")
+	}
+	if len(allFinishedTasks) == 0 {
+		return nil
+	}
+	err = restartTasks(allFinishedTasks, caller)
+	if err != nil {
+		return errors.Wrap(err, "restarting tasks")
+	}
+	return errors.Wrap(setVersionStatus(versionId, evergreen.VersionStarted), "changing version status")
```
ablack12, line 352: TODO: don't we need to call updateBuildAndVersionStatus

model/lifecycle.go
```diff
@@ -373,47 +396,42 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
 				Build:     t.BuildId,
 				TaskGroup: t.TaskGroup,
 			}] = t
-		} else {
-			// Only hard restart non-single host task group tasks
-			restartIds = append(restartIds, t.Id)
-			if t.DisplayOnly {
-				restartIds = append(restartIds, t.ExecutionTasks...)
-			}
+			continue
```
ablack12, line 409: preference to keep this an if / else statement; continue seems more useful when we're exiting early in a function that does a lot of stuff, or when the only thing we want to do in the if case is continue

model/lifecycle.go
```diff
@@ -432,79 +450,18 @@ func RestartVersions(versionsToRestart []*VersionToRestart, abortInProgress bool
 func RestartBuild(buildId string, taskIds []string, abortInProgress bool, caller string) error {
 	if abortInProgress {
 		// abort in-progress tasks in this build
-		if err := task.AbortTasksForBuild(buildId, taskIds, caller); err != nil {
+		if err := task.AbortAndMarkResetTasksForBuild(buildId, taskIds, caller); err != nil {
```
ablack12, line 453: Is there a benefit to marking tasks reset here? We already do that in restartTasks (or is there a way to move that out of that function now?)

model/lifecycle.go
```diff
@@ -432,79 +450,18 @@ func RestartVersions(versionsToRestart []*VersionToRestart, abortInProgress bool
...
-	}
-
-	for tg, t := range taskGroupsToCheck {
-		if err := checkResetSingleHostTaskGroup(&t, caller); err != nil {
-			return errors.Wrapf(err, "resetting single host task group '%s'", tg)
-		}
-	}
-
-	return errors.Wrap(build.SetBuildStartedForTasks(tasks, caller), "setting builds started")
+	return restartTasks(tasksToReset, caller)
```
ablack12, line 464: Since etVersionStatus was moved out of restartTasks, I think that the version status stays outdated here. I'm not sure there's a time we'd want to restart tasks and _not_ update the version so we may want to move it back in.

ablack12 (changes requested):

model/lifecycle.go
hadjri: Sounds good, moved to function level comment

model/lifecycle.go
hadjri: Only single host TG tasks are marked for reset in `restartTasks` (because we need to wait for the whole group to end before we reset those), so we still need to mark the aborted tasks for reset if we want them to be considered part of the restart.

model/lifecycle.go
hadjri: That is a good point. I think moving `setVersionStatus` should fix this.

model/lifecycle.go
ablack12: As far as I was aware we don't check ResetWhenFinished for other things. Unless we need this change to make display tasks work?

model/lifecycle.go
hadjri: Not sure what you mean here?

model/lifecycle.go
hadjri: We check it [here](https://gist.github.com/hadjri/e1849bd79ce84b89907a3c3822aaa0cd#file-task_lifecycle-go-L711) when a task is ended by the agent. In this case, the running tasks will get marked `ResetWhenFinished` and `Aborted`, then later the agent will pick up on this and fail the tasks, and this line will hit causing the task to reset automatically.

model/lifecycle.go
ablack12: I think this was a thinking comment i forgot to make intelligible and idk what it meant now haha

model/lifecycle.go
```diff
@@ -395,22 +411,23 @@ func RestartVersion(versionId string, taskIds []string, abortInProgress bool, ca
...
-	for _, t := range tasksToRestart {
+	for _, t := range allFinishedTasks {
 		if !t.IsPartOfSingleHostTaskGroup() { // this will be logged separately if task group is restarted
 			event.LogTaskRestarted(t.Id, t.Execution, caller)
 		}
 	}
-	if err = build.SetBuildStartedForTasks(tasksToRestart, caller); err != nil {
-		return errors.Wrap(err, "setting builds started")
+
+	if err := errors.Wrap(build.SetBuildStartedForTasks(allFinishedTasks, caller), "setting builds started"); err != nil {
```
ablack12, line 429: Why did we move the wrap up to the if statement level? This is pretty unconventional / harder to read than just keeping the wrap in the next line

model/lifecycle.go
ablack12: Hm I'm kind of confused about how this works now then. Is this a confirmed bug (that a regular task aborted via this function doesn't get restarted bc resetWhenFinished isn't set) or is this logic handled elsewhere?

model/lifecycle.go
hadjri: I am not sure haha - reverted.

model/lifecycle.go
hadjri: Yes currently, `RestartBuild` has a bug because the in-progress tasks just get aborted and that's it, there isn't any extra logic handling. The change to also add `ResetWhenFinished` will make it work correctly.

ablack12: Sorry @hadjri I think this comment might have gotten lost since it's an old thread (or let me know if you did address it already)

ablack12 (approved):

PR #6158, opened 2023-01-03
EVG-18566 Fix project name missing in slack notifications

malikchaya2:
[EVG-18566](https://jira.mongodb.org/browse/EVG-18566)

### Description 
This happens because some api patches don't have the project identifier populated. It's unclear how that happens because after auditing the code base, I can assert that we include args with IncludeProjectIdentifier set to true whenever appropriate. 

When fetching the project identifier, we currently do nothing on error so I'm unable to tell if that's whats happening. I added logging so that it's traceable. 

To fix the problem, I added a check to makeData to fetch the project identifier when building the slack message if it doesn't exist on the patch. I added it to builds and versions as well. 

### Testing 
Reasoned about it.

### DIFF
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -151,12 +151,12 @@ func (apiPatch *APIPatch) GetCommitQueuePosition() error {
 }
 
 func (apiPatch *APIPatch) GetIdentifier() {
-	if apiPatch.ProjectIdentifier != nil {
+	if utility.FromStringPtr(apiPatch.ProjectIdentifier) != "" {
 		return
 	}
 	if utility.FromStringPtr(apiPatch.ProjectId) != "" {
 		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(apiPatch.ProjectId))
-		if err == nil {
+		if err == nil && identifier != "" {
 			apiPatch.ProjectIdentifier = utility.ToStringPtr(identifier)
 		}
 	}
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -157,13 +157,12 @@ func (apiPatch *APIPatch) GetIdentifier() {
 	}
 	if utility.FromStringPtr(apiPatch.ProjectId) != "" {
 		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(apiPatch.ProjectId))
-		if err != nil {
-			grip.Error(message.WrapError(err, message.Fields{
-				"message": "could not get identifier for project",
-				"project": apiPatch.ProjectId,
-			}))
 
-		}
+		grip.Error(message.WrapError(err, message.Fields{
+			"message": "could not get identifier for project",
+			"project": apiPatch.ProjectId,
+		}))
+
 		if err == nil && identifier != "" {
 			apiPatch.ProjectIdentifier = utility.ToStringPtr(identifier)
 		}
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -13,6 +13,7 @@ import (
 	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 )
 
@@ -156,6 +157,13 @@ func (apiPatch *APIPatch) GetIdentifier() {
 	}
 	if utility.FromStringPtr(apiPatch.ProjectId) != "" {
 		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(apiPatch.ProjectId))
+		if err != nil {
+			grip.Error(message.WrapError(err, message.Fields{
+				"message": "could not get identifier for project",
+				"project": apiPatch.ProjectId,
+			}))
+
+		}
 		if err == nil && identifier != "" {
 			apiPatch.ProjectIdentifier = utility.ToStringPtr(identifier)
 		}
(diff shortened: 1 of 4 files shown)


trigger/build.go
```diff
@@ -269,6 +270,11 @@ func (t *buildTriggers) makeData(sub *event.Subscription, pastTenseOverride stri
 	projectName := t.build.Project
 	if api.ProjectIdentifier != nil {
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
+	} else {
+		identifier, err := model.GetIdentifierForProject(*api.ProjectId)
```
Kimchelly, line 274: This fallback logic seems to be identical to what `api.BuildFromService` already does internally, which [sets the `ProjectId` based on the build's project](https://github.com/evergreen-ci/evergreen/blob/24b4506727504198dd5cbbd47fb020c3e7e0a89a/rest/model/build.go#L90), then [uses `model.GetIdentifierForProject` to get the identifier](https://github.com/evergreen-ci/evergreen/blob/24b4506727504198dd5cbbd47fb020c3e7e0a89a/rest/model/build.go#L117-L122).

trigger/version.go
```diff
@@ -104,6 +104,11 @@ func (t *versionTriggers) makeData(sub *event.Subscription, pastTenseOverride st
 	projectName := t.version.Identifier
 	if api.ProjectIdentifier != nil {
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
+	} else {
+		identifier, err := model.GetIdentifierForProject(*api.Project)
```
Kimchelly, line 108: Same thing as above - seems like the [project ID used here](https://github.com/evergreen-ci/evergreen/blob/24b4506727504198dd5cbbd47fb020c3e7e0a89a/rest/model/version.go#L55) is [using the same logic to get the project identifier](https://github.com/evergreen-ci/evergreen/blob/24b4506727504198dd5cbbd47fb020c3e7e0a89a/rest/model/version.go#L81-L86) as the if block above this.

rest/model/patch.go
```diff
@@ -151,12 +152,19 @@ func (apiPatch *APIPatch) GetCommitQueuePosition() error {
...
 func (apiPatch *APIPatch) GetIdentifier() {
-	if apiPatch.ProjectIdentifier != nil {
+	if utility.FromStringPtr(apiPatch.ProjectIdentifier) != "" {
 		return
 	}
 	if utility.FromStringPtr(apiPatch.ProjectId) != "" {
 		identifier, err := model.GetIdentifierForProject(utility.FromStringPtr(apiPatch.ProjectId))
-		if err == nil {
+		if err != nil {
+			grip.Error(message.WrapError(err, message.Fields{
```
Kimchelly, line 161: Is this meant to be temporary debug logging for fixing this one issue, or permanent? If its temporary, can we add a ticket to remove it? Otherwise, if it's permanent, will this be useful for tracing down issues in the future? It seems like this is deep enough in the implementation logic that nobody would think to look for this log.

trigger/patch.go
```diff
@@ -189,6 +189,11 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
 	projectName := t.patch.Project
 	if api.ProjectIdentifier != nil {
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
+	} else {
+		identifier, err := model.GetIdentifierForProject(*api.ProjectId)
```
Kimchelly, line 193: Assuming we're confident that `IncludeProjectIdentifier` is always set when needed for patch triggers, it seems like this wouldn't do anything, because [`api.ProjectId` is the same as the patch's project](https://github.com/evergreen-ci/evergreen/blob/24b4506727504198dd5cbbd47fb020c3e7e0a89a/rest/model/patch.go#L168), so this check seems equivalent to the if block above this. Are we certain that `t.patch.Project` is always set? If not, it seems like this would have the same issue as the old logic.

Kimchelly (changes requested):

rest/model/patch.go
malikchaya2: I meant to keep it there, with my logic being that it's good for us to have access to it, because right now there is not much traceability.

trigger/patch.go
malikchaya2: Maybe I'll just add the logging and the other checks, and see what happens.

malikchaya2: Okay, I removed them. The theory that the cause of it is because the project identifier was set to an empty string feels like a long shot, but it's the only theory at the moment. If that doesn't fix it, I'll have the logging to help trace it.

trigger/version.go
malikchaya2: removed

trigger/build.go
malikchaya2: removed

rest/model/patch.go
Kimchelly: That sounds fine to me. In that case, we can remove the `if err != nil {` check because `grip.Error(message.WrapError(err...` already has internal logic to not log if the err is nil.

trigger/patch.go
Kimchelly: Is it possible that that `ProjectId` is nil or empty? It seems like [this check](https://github.com/evergreen-ci/evergreen/blob/1e57833ee9e7c59cca434dd398574c0b3e6d98d1/rest/model/patch.go#L115) would skip all this logic if `Project` is empty (i.e. `ProjectId` is empty).

malikchaya2: It doesn't seem likely because the patch in question (which is just one example) does have it populated in the db.

rest/model/patch.go
malikchaya2: thanks, removed

PR #6160, opened 2023-01-04
EVG-16826 Add dependencies to single host task groups upon creation

hadjri:
[EVG-16826](https://jira.mongodb.org/browse/EVG-16826)

### Description 
We currently only add dependencies to single host task groups in the case that one of them fails so we can block the subsequent tasks. There is a lot of extraneous logic to handle these cases and it's simpler to assign these dependencies upon task creation. 

Change creates dependencies for single host task groups upon creation, where each task depends on the task in the group that comes before it.

### Testing 
Tested in staging and confirmed the change works even when the tasks succeed.


model/lifecycle.go
```diff
@@ -930,11 +931,42 @@ func createTasksForBuild(creationInfo TaskCreationInfo) (task.Tasks, error) {
...
+		tg := p.FindTaskGroup(t.TaskGroup)
+		if tg == nil || tg.MaxHosts > 1 {
+			continue
+		}
+		singleHostTGDeps := []TaskUnitDependency{}
+		// Iterate backwards until we find a task that exists in the taskMap. This task
+		// will be the parent dependency for the current single host TG task.
+		taskFound := false
+		for i := len(tg.Tasks) - 1; i >= 0; i-- {
+			if t.DisplayName == tg.Tasks[i] {
```
Kimchelly, line 949: Not sure how much this matters really, but would this account for tasks that are used multiple times in the same task group? For example:
```yaml
task_groups:
    - name: my_task_group
      tasks:
        - run-my-script
        - test-a-thing
        - run-my-script
```
I don't know if this is necessarily a case we need to handle or perhaps document.

model/lifecycle.go
```diff
@@ -930,11 +931,42 @@ func createTasksForBuild(creationInfo TaskCreationInfo) (task.Tasks, error) {
...
+				taskFound = true
+				continue
+			}
+			if _, ok := taskMap[taskIds.GetId(t.BuildVariant, tg.Tasks[i])]; ok && taskFound {
```
Kimchelly, line 953: Is there ever a case where we could expect `ok` would be false here? I see there's [a test for this](https://github.com/evergreen-ci/evergreen/blob/1a8abd172c2f87f084c5f765e61cce7def610f05/model/lifecycle_test.go#L1500), but I'm not sure if/when this could happen.

Is it worth logging or something if `ok` is ever false?

model/task_lifecycle.go
```diff
@@ -1682,30 +1672,6 @@ func doRestartFailedTasks(tasks []string, user string, results RestartResults) R
 	return results
 }
 
-// CheckAndBlockSingleHostTaskGroup blocks a single-host task group if the given
-// task is part of one and has finished (or is currently finishing) with a
-// failed status. This is a best-effort attempt, so if it fails, it will just
-// log the error.
-func CheckAndBlockSingleHostTaskGroup(t *task.Task, status string) {
```
Kimchelly, line 1689: We would have to keep this logic if we wanted backward compatibility with existing versions that don't have the dependencies set up. Is it worth keeping this for backward compatibility? I don't have strong opinions here.

model/lifecycle.go
hadjri: I don't think we need to handle this since it is a [validation error](https://gist.github.com/hadjri/73759d0ef290638c1b637b14b7e33699#file-project_validator-go-L1504) to do that. I'll add a comment to make it clear we can rely on this fact though

model/lifecycle.go
hadjri: This could happen if a task group contains `t1, t2, t3` and we only create `t1` and `t3`. Since `t2` doesn't exist then `t3` needs to depend directly on `t1`. The check for `ok` is just to see if the task was actually created, and if not we can just skip and look for the next task in the group when we're adding dependencies.

model/task_lifecycle.go
hadjri: Hmm, yeah I think we should do that actually. I'll make a ticket to remove it later

model/lifecycle.go
hadjri: You can schedule a subset of  a single host task group tasks in the UI. If they succeed currently, there will be no dependencies ([example](https://spruce-staging.corp.mongodb.com/version/63bc8508b237360254cdef85/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)), but if one fails, dependencies will be set ([example](https://spruce-staging.corp.mongodb.com/version/63bc853397b1d313d5bf02c3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)).

Kimchelly: Interesting... it seems a bit strange that it's allowed, but I guess it doesn't matter much as long as the dependencies are set up correctly.

Kimchelly (approved):
LGTM mod one request - can we make a ticket to [remove the legacy dispatcher version](https://github.com/evergreen-ci/evergreen/blob/502403c1ceee7548e323a9f130fb5967e57a433b/globals.go#L211)? It seems like correctness hinges upon the dispatcher taking dependencies into account, so keeping any legacy dispatcher logic seems undesirable.

hadjri:
> LGTM mod one request - can we make a ticket to [remove the legacy dispatcher version](https://github.com/evergreen-ci/evergreen/blob/502403c1ceee7548e323a9f130fb5967e57a433b/globals.go#L211)? It seems like correctness hinges upon the dispatcher taking dependencies into account, so keeping any legacy dispatcher logic seems undesirable.

I thought the same when I was debugging the smoke tests (which were failing due to this and took a long time to debug). I created https://jira.mongodb.org/browse/EVG-18706 and going to add some TODOs

PR #6161, opened 2023-01-04
EVG-18561 Fix wrong status in Github notifications 

malikchaya2:
[EVG-18561](https://jira.mongodb.org/browse/EVG-18561)

### Description 
The logic that makes sure that patch completion events are logged with the collective status filtered on patch requesters, so patch statuses for Github patches were not consistent. 

### Testing 
Reasoned about it.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1371,7 +1371,7 @@ func UpdateBuildAndVersionStatusForTask(t *task.Task) error {
 		return errors.Wrapf(err, "updating version '%s' status", taskVersion.Id)
 	}
 
-	if evergreen.IsPatchRequester(taskVersion.Requester) {
+	if evergreen.IsPatchRequester(taskVersion.Requester) || evergreen.IsGitHubPatchRequester(taskVersion.Requester) {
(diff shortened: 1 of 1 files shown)


model/task_lifecycle.go
```diff
@@ -1371,18 +1371,19 @@ func UpdateBuildAndVersionStatusForTask(t *task.Task) error {
 		return errors.Wrapf(err, "updating version '%s' status", taskVersion.Id)
 	}
 
-	if evergreen.IsPatchRequester(taskVersion.Requester) {
+	if evergreen.IsPatchRequester(taskVersion.Requester) || evergreen.IsGitHubPatchRequester(taskVersion.Requester) {
```
ablack12, line 1374: IsPatchRequester already calls IsGitHubPatchRequester so this may not be the right change

ablack12 (changes requested):

model/task_lifecycle.go
malikchaya2: :see_no_evil:  thank you! correct. 

After further investigation I closed the ticket and added some more information in a comment in the ticket.

PR #6162, opened 2023-01-09
EVG-18267 Update permissions for project create

bynn:
[EVG-18267](https://jira.mongodb.org/browse/EVG-18267)

### Description 
a couple of changes 
1. people were previously able to duplicate projects even without being admin so I added an extra check 
2. all admin roles now come with a project-create role 
3. project create role is removed if all admin roles are removed 

### Testing 
tested on staging

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2303,6 +2303,19 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) (bool, error) {
 		return false, nil
 	}
 	rm := evergreen.GetEnvironment().RoleManager()
+	projectCreateRole, err := rm.FindRoleWithPermissions(
+		evergreen.SuperUserResourceType,
+		[]string{evergreen.SuperUserPermissionsID},
+		gimlet.Permissions{
+			evergreen.PermissionProjectCreate: evergreen.ProjectCreate.Value,
+		},
+	)
+	if err != nil {
+		return false, errors.Wrap(err, "finding role with project create permissions")
+	}
+	if projectCreateRole == nil {
+		return false, errors.New("project create role not found")
+	}
 	role, err := rm.FindRoleWithPermissions(evergreen.ProjectResourceType, []string{p.Id}, adminPermissions)
 	if err != nil {
 		return false, errors.Wrap(err, "finding role with admin permissions")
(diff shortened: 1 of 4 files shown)


bynn:
would have to run a db command on prod to give all existing admins the project create role

model/project_ref.go
```diff
@@ -2303,6 +2303,19 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) (bool, error) {
 		return false, nil
 	}
 	rm := evergreen.GetEnvironment().RoleManager()
+	projectCreateRole, err := rm.FindRoleWithPermissions(
```
ablack12, line 2306: Can we add a test for this? We'd want to verify that we correctly find the role that's formatted like the one we have now, and that we correctly add it.

model/project_ref.go
```diff
@@ -2333,8 +2346,9 @@ func (p *ProjectRef) UpdateAdminRoles(toAdd, toRemove []string) (bool, error) {
 			p.removeFromAdminsList(addedUser)
 			continue
 		}
-		if err = adminUser.AddRole(role.ID); err != nil {
-			catcher.Wrapf(err, "adding role '%s' to user '%s'", role.ID, addedUser)
+		rolesToAdd := []string{role.ID, projectCreateRole.ID}
+		if err = adminUser.AddRoles(rolesToAdd); err != nil {
+			catcher.Wrapf(err, "adding roles '%s' to user '%s'", rolesToAdd, addedUser)
```
ablack12, line 2351: roles '%v'

model/user/user.go
```diff
@@ -336,6 +353,15 @@ func (u *DBUser) RemoveRole(role string) error {
 	return event.LogUserEvent(u.Id, event.UserEventTypeRolesUpdate, before, u.SystemRoles)
 }
 
+func (u *DBUser) IsAdmin() bool {
```
ablack12, line 356: I don't love this function since it assumes the naming scheme of the admin role -- it's also not accurate, since this naming scheme is only used for roles added from the project page as opposed to mana. Maybe we could do something similar to https://github.com/evergreen-ci/evergreen/blob/f21f81cc987e426ee8d0fe6095c73594e49b081c/model/user/user.go#L375 that just iterates through u.SystemRoles to check permissions and ignores ResourceType

rest/data/project_settings.go
```diff
@@ -56,6 +56,15 @@ func CopyProject(ctx context.Context, env evergreen.Environment, opts CopyProjec
 	disableStartingSettings(projectToCopy)
 
 	u := gimlet.GetUser(ctx).(*user.DBUser)
+	isAdmin := u.HasPermission(gimlet.PermissionOpts{
```
ablack12, line 59: We can't / shouldn't add this here -- we use middleware to handle permission checking before we ever get to the route logic, which is why the route currently doesn't use the createProject logic and instead uses the requireProjectAdmin logic (which should be fine to keep). If the idea is to fix this for Spruce, it may be to change the directive used here https://github.com/evergreen-ci/evergreen/blob/adacf7967c2df3c5b4db9f551b02efb3ce983e82/graphql/schema/mutation.graphql#L52 . I think it would be worth splitting that work to a separate ticket.

ablack12 (changes requested):
I think these tickets were implemented slightly out of order, since it looks like the work to ensure validation hasn't been added yet. We don't want to give users permission power until the project is more wrapped up to ensure we're ready to do so, so after we approve this work we probably want to make a new ticket to merge / migrate existing users to do at the end.

PR #6164, opened 2023-01-09
EVG-18153 fix commit queue time out and PR failure

ablack12:
[EVG-18153](https://jira.mongodb.org/browse/EVG-18153)

### Description 
1) Found a bug where passing in a PR patch ID correctly removed from queue but failed the Atoi line bc we mistakenly use it as the PR number.
2) Return StatusNotFound from the delete function so we don't retry on failures (causing the time out).
3) Refactor to not look up version, since we can rely on the item version field.

### Testing 
Tested on staging to verify


Kimchelly (approved):

Kimchelly:
LGTM mod updating commit queue delete tests.

PR #6169, opened 2023-01-10
EVG-14416 Skip mergeable check when submitting to the commit queue

malikchaya2:
[EVG-14416](https://jira.mongodb.org/browse/EVG-14416)

### Description 
There's already a mergeable check in commit queue tasks, so there's no need to check it here. Checking here can make it so that enqueuing very quickly after pushing a commit fails to enqueue

### DIFF
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -1026,7 +1026,7 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 		return nil, errors.Wrap(err, "GitHub returned an incomplete PR")
 	}
 
-	if !(utility.FromBoolTPtr(pr.Mergeable)) {
+	if !utility.FromBoolTPtr(pr.Mergeable) {
 		return pr, errors.New("PR is not mergeable")
 	}
 
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -1026,7 +1026,7 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 		return nil, errors.Wrap(err, "GitHub returned an incomplete PR")
 	}
 
-	if !(utility.FromBoolPtr(pr.Mergeable)) {
+	if !(utility.FromBoolTPtr(pr.Mergeable)) {
 		return pr, errors.New("PR is not mergeable")
 	}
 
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):

thirdparty/github.go
```diff
@@ -1026,16 +1026,6 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
...
-	if pr.Mergeable == nil {
-		if *pr.Merged {
-			return pr, errors.New("PR is already merged")
-		}
-		// GitHub hasn't yet tested if the PR is mergeable.
-		// Check back later
-		// See: https://developer.github.com/v3/pulls/#response-1
-		return pr, errors.New("GitHub hasn't yet generated a merge commit")
-	}
-
```
Kimchelly, line 1038: L1039 needs to check if pr.Mergeable is nil before dereferencing.

Kimchelly (commented):
Sorry, I noticed one thing on an unchanged line.

thirdparty/github.go
malikchaya2: thanks! switched to utility.FromBoolPtr.

Kimchelly: I think it has to be `utility.FromBoolTPtr` because if the pointer is nil, we're assuming it's mergeable.

malikchaya2: It seems to me that if the pointer is nil, FromBoolPtr returns false, so we end up with !(false), so we enter the if and return errors.New("PR is not mergeable"). 

Am I misreading it?

Kimchelly: I thought the point of removing the nil check was to assume that if the mergeability check isn't available yet, we just skip the check, and would only want to exit early if `*Mergeable` was true. If `nil` (i.e. can't check mergeability yet) is equivalent to `false`, then the if statement will make it claim that it's not mergeable, but we actually don't know if it's mergeable.

malikchaya2: ah yes :see_no_evil: . Thanks! fixed!

thirdparty/github.go
```diff
@@ -1026,7 +1026,7 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 		return nil, errors.Wrap(err, "GitHub returned an incomplete PR")
 	}
 
-	if !(utility.FromBoolPtr(pr.Mergeable)) {
+	if !(utility.FromBoolTPtr(pr.Mergeable)) {
```
Kimchelly, line 1029: nit: the parentheses aren't necessary since `!` is a unary operator.

Kimchelly (approved):
LGTM mod nit.

PR #6171, opened 2023-01-10
Ignore non-repotracker versions when activating versions

hadjri:
[EVG-18673](https://jira.mongodb.org/browse/EVG-18673)

### Description 
There is an [issue](https://mongodb.slack.com/archives/C0V896UV8/p1673021506764289) in the way activating tasks based on batchtime / cron works because the version activation job looks for the most recent version with one of the `SystemVersionRequesterTypes` which also includes periodic builds, git tags and triggers.  

If a non-repotracker version is the most recent one found at the time of version activation, and it doesn't contain the cron / batchtime tasks, then the cron / batchtime tasks from the most recent repotracker version will get skipped and will never be activated.

Updated to only look for repotracker versions.

### Testing 
Updated unit test.

### DIFF
--- a/model/version_db.go
+++ b/model/version_db.go
@@ -248,14 +248,12 @@ func VersionBySystemRequesterOrdered(projectId string, startOrder int) db.Q {
 	return db.Query(q).Sort([]string{"-" + VersionRevisionOrderNumberKey})
 }
 
-// ByMostRecentNonIgnored finds all non-ignored versions within a project,
+// VersionByMostRecentNonIgnored finds all non-ignored versions within a project,
 // ordered by most recently created to oldest, before a given time.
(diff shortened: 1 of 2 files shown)


ablack12 (approved):
LGTM -- I'm willing to risk a regression on this, let's just keep our eyes out for any issues, and make a documentation change to cron/batchtime

hadjri:
> LGTM -- I'm willing to risk a regression on this, let's just keep our eyes out for any issues, and make a documentation change to cron/batchtime

Sounds good, I've added an extra sentence to the cron & batchtime explanations here to say they only apply to mainline commits: https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files

PR #6175, opened 2023-01-11
EVG-18543 consider only activate: false for specific generate task activation

ablack12:
[EVG-18543](https://jira.mongodb.org/browse/EVG-18543)

### Description 
It's not typical to add `activate: true` to generate tasks output which is likely why we've never run into this but it's technically valid but really just means the default; we should really only consider if it's explicitly set to false.


hadjri (approved):
Love a nice straightforward fix! 🚀

PR #6178, opened 2023-01-11
EVG-18221 Update to search patch aliases

bynn:
[EVG-18221](https://jira.mongodb.org/browse/EVG-18221)

### Description 
deleted shouldExit variable when checking for aliases because that was causing us to never search project configs for patch aliases 

New alias behavior that I'm proposing : 
Instead of ignoring everything in project configs when a patch alias is defined on the UI,  we still check the project configs.
The UI will still take precedence over the project config patch alias if an alias with the same name is redefined.
My reasoning for that being that users can more easily change the names of custom patch aliases for their specific patch to not match an existing one if they really wanted. 

### Testing 
staging
unit tests

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -1782,28 +1782,29 @@ func findAliasesForPatch(projectId, alias string, patchDoc *patch.Patch) ([]Proj
 	if err != nil {
 		return nil, errors.Wrap(err, "getting alias from project")
 	}
-	if len(vars) == 0 {
-		pRef, err := FindMergedProjectRef(projectId, "", false)
+	if len(vars) > 0 {
+		return vars, nil
+	}
+	pRef, err := FindMergedProjectRef(projectId, "", false)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting project ref")
+	}
+	if pRef == nil || !pRef.IsVersionControlEnabled() {
+		return vars, nil
+	}
+	if len(patchDoc.PatchedProjectConfig) > 0 {
+		projectConfig, err := CreateProjectConfig([]byte(patchDoc.PatchedProjectConfig), projectId)
 		if err != nil {
-			return nil, errors.Wrap(err, "getting project ref")
+			return nil, errors.Wrap(err, "retrieving aliases from patched config")
 		}
-		if pRef == nil || !pRef.IsVersionControlEnabled() {
-			return vars, nil
+		vars, err = findAliasFromProjectConfig(projectConfig, alias)
+		if err != nil {
+			return nil, errors.Wrapf(err, "retrieving alias '%s' from project config", alias)
 		}
-		if len(patchDoc.PatchedProjectConfig) > 0 {
-			projectConfig, err := CreateProjectConfig([]byte(patchDoc.PatchedProjectConfig), projectId)
-			if err != nil {
-				return nil, errors.Wrap(err, "retrieving aliases from patched config")
-			}
-			vars, err = findAliasFromProjectConfig(projectConfig, alias)
-			if err != nil {
-				return nil, errors.Wrapf(err, "retrieving alias '%s' from project config", alias)
-			}
-		} else if patchDoc.Version != "" {
-			vars, err = getMatchingAliasForVersion(patchDoc.Version, alias)
-			if err != nil {
-				return nil, errors.Wrapf(err, "retrieving alias '%s' from project config", alias)
-			}
+	} else if patchDoc.Version != "" {
+		vars, err = getMatchingAliasForVersion(patchDoc.Version, alias)
+		if err != nil {
+			return nil, errors.Wrapf(err, "retrieving alias '%s' from project config", alias)
 		}
 	}
 	return vars, nil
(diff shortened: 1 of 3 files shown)


model/project_aliases.go
```diff
@@ -185,26 +185,18 @@ func findMatchingAliasForRepo(repoID, alias string) ([]ProjectAlias, error) {
 // Typically FindAliasInProjectRepoOrConfig should be used.
 // Returns true if we have an alias match or the alias doesn't match but
```
Kimchelly, line 186: Remove this part of the doc comment since it's no longer relevant.

Kimchelly (commented):
1. Merging the UI and project config aliases sounds fine to me. Do we have to update any wiki docs to mention this behavior?

2. Does this change also address the issue where the YAML aliases weren't being used even removing all the UI aliases? In the original ticket, it was mentioned that the user already tried deleting all the aliases in the UI, but it still didn't pick up the alias in their YAML. I was thinking the `shouldExit` condition would already be false once the UI aliases were all deleted.

model/project_aliases.go
bynn: done

bynn:
@Kimchelly 
1. I don't think we mention it in the wiki but I will update it here once the ticket is complete 
2. yes this should fix the issue that the ticket is based on

hadjri:
> @Kimchelly
> 
> 1. I don't think we mention it in the wiki but I will update it here once the ticket is complete
> 2. yes this should fix the issue that the ticket is based on

FYI the wiki does talk about project config version control so this paragraph will likely need to be updated in accordance of how we change the precedence rules: https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#version-control

model/project_aliases.go
```diff
@@ -355,42 +337,34 @@ func FindAliasInProjectRepoOrProjectConfig(projectID, alias string, projectConfi
 // findAliasInProjectOrRepoFromDb finds all aliases with a given name for a project without merging with parser project.
 // If the project has no aliases, the repo is checked for aliases.
 // Returns true if we should continue to look for repos from a different source.
-func findAliasInProjectOrRepoFromDb(projectID, alias string) ([]ProjectAlias, bool, error) {
-	aliases, shouldExit, err := findMatchingAliasForProjectRef(projectID, alias)
+func findAliasInProjectOrRepoFromDb(projectID, alias string) ([]ProjectAlias, error) {
```
hadjri, line 340: Need to update comment here

model/project.go
```diff
@@ -1778,11 +1778,11 @@ func (p *Project) IsGenerateTask(taskName string) bool {
 }
 
 func findAliasesForPatch(projectId, alias string, patchDoc *patch.Patch) ([]ProjectAlias, error) {
-	vars, shouldExit, err := findAliasInProjectOrRepoFromDb(projectId, alias)
+	vars, err := findAliasInProjectOrRepoFromDb(projectId, alias)
 	if err != nil {
 		return nil, errors.Wrap(err, "getting alias from project")
 	}
-	if !shouldExit && len(vars) == 0 {
+	if len(vars) == 0 {
```
hadjri, line 1785: I know you didn't write this but can we instead say on this line `if len(vars) > 0 { return vars }` and then remove the need for a big indentation below for readability purposes

model/project_aliases.go
```diff
@@ -252,14 +234,14 @@ func aliasesFromMap(input map[string]ProjectAliases) []ProjectAlias {
...
-	aliases, shouldExit, err := findAliasInProjectOrRepoFromDb(projectID, alias)
+	aliases, err := findAliasInProjectOrRepoFromDb(projectID, alias)
 	if err != nil {
 		return nil, errors.Wrap(err, "checking for existing aliases")
 	}
 	// If nothing is defined in the DB, check the project config,
 	// unless the alias defined is a patch alias and there are patch aliases
 	// defined on the project page.
-	if len(aliases) > 0 || shouldExit {
+	if len(aliases) > 0 {
```
hadjri, line 244: The 2nd part of the above comment should be removed.

hadjri (changes requested):

Kimchelly (approved):
LGTM mod comment text.

model/project_aliases.go
bynn: done

model/project_aliases.go
```diff
@@ -182,29 +182,18 @@ func findMatchingAliasForRepo(repoID, alias string) ([]ProjectAlias, error) {
 }
 
 // findMatchingAliasForProjectRef finds all aliases with a given name for a project.
-// Typically FindAliasInProjectRepoOrConfig should be used.
```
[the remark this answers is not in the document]
bynn: done

model/project.go
bynn: done

model/project_aliases.go
bynn: done

hadjri (approved):

bynn:
@hadjri just wanted to let you know that this change is already in accordance with what the wiki says so no update was needed

PR #2472, opened 2019-07-23
EVG-6383 restart failed commit queue versions 

ablack12:
Given a time range, we filter patches by Alias, Time, and Project. I'm filtering patches instead of versions because patches have the actual logic I'm looking for, and for commit queue this seems 1 to 1 (FinalizePatch creates a version with the same ID as the patch, and this is called for both PR and CLI). 

From these patches I reconstruct the commit queue item and re-add it to the project.

Because this ranges first through projects and then through project patches, it's slightly inefficient; an alternative may be to get the patches for all projects and keep a hash map of projectID -> CQ, but this would only minimize the number of patch filter calls we have to make, not the number of patches we have to range over.

PR #6179, opened 2023-01-11
EVG-18634 change range to consider started or failed within time range

ablack12:
[EVG-18634 ](https://jira.mongodb.org/browse/EVG-18634 )

### Description 
It looks like I originally very purposefully made this query only check for things that either started in range or failed in range but had no start time? I don't understand than scenario 😅 and I don't see anything enlightening [in the PR](https://github.com/evergreen-ci/evergreen/pull/2472). 

### Testing 
Added unit tests.

### DIFF
--- a/model/patch/db.go
+++ b/model/patch/db.go
@@ -318,8 +318,7 @@ func PatchesByProject(projectId string, ts time.Time, limit int) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(limit)
 }
 
-// FindFailedCommitQueuePatchesInTimeRange returns failed patches if they started within range,
-// or if they were never started but finished within time range. (i.e. timed out)
+// FindFailedCommitQueuePatchesInTimeRange returns failed patches if they started or failed within range.
 func FindFailedCommitQueuePatchesInTimeRange(projectID string, startTime, endTime time.Time) ([]Patch, error) {
 	query := bson.M{
 		ProjectKey: projectID,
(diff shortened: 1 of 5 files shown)


ablack12:
One question is: do we want to make this a _choice_ on the admin page? (Choose either started OR failed?) I think we may want to do different ones depending on what we're doing it for. However we may want to break that into a different ticket and it may not be worth adding complexity to the legacy admin page.

model/patch/db.go
```diff
@@ -331,7 +330,6 @@ func FindFailedCommitQueuePatchesInTimeRange(projectID string, startTime, endTim
 				{StartTimeKey: bson.M{"$gte": startTime}},
 			}},
 			{"$and": []bson.M{
-				{StartTimeKey: time.Time{}},
```
Kimchelly, line 334: This might be possible if you call `MarkEnd` before the task actually starts. Container tasks currently call `MarkEnd` to system fail the task if the container can't run (meaning the task can't start) so they have a finish time but no start time.

Kimchelly (approved):

Kimchelly:
Worth mentioning that with container tasks (not host tasks), it's possible for tasks to have a finish time but no start time if the container cannot run (e.g. they provide a nonexistent container), because it gets marked as system failed.

ybrill (approved):
I really like the idea of having a choice between using a started range and a finished range.

Like in my case I would probably have wanted to use specifically a finished range for the few minutes just after I turned the application back on to catch any tasks that idle timeout'ed maybe?

ablack12:
I'm going to cut a new ticket for the choice since we may want to reconsider it for the new UI instead: [EVG-18736](https://jira.mongodb.org/browse/EVG-18736)

PR #6181, opened 2023-01-12
EVG-18268 Add project create validation for enabling projects

bynn:
[EVG-18268](https://jira.mongodb.org/browse/EVG-18268)

### Description 
validation function errors when you try to enable a project if you are past the limit

### Testing 
staging
unit tests


ablack12:
Started reviewing but I'm seeing a couple of things big enough that it may be easier to discuss IRL so will sync up with you.

model/project_ref.go
```diff
@@ -2203,6 +2244,38 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+	allEnabledProjects, err := GetNumberOfProjects()
+	if err != nil {
+		return errors.Wrap(err, "getting number of projects")
+	}
+	if allEnabledProjects >= config.ProjectCreation.TotalProjectLimit {
+		catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
+	}
+	enabledOwnerRepoProjects, err := GetNumberOfProjectsForOwnerRepo(projectRef.Owner, projectRef.Repo)
+	if err != nil {
+		return errors.Wrapf(err, "getting number of projects for '%s'/'%s'", projectRef.Owner, projectRef.Repo)
```
ablack12, line 2260: doesn't make sense to put owner and repo in separate quotes, since really the owner/repo pair is one string (so '%s/%s' is preferable)

model/project_ref.go
```diff
@@ -2203,6 +2244,38 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+	}
+	if enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit && !config.ProjectCreation.IsException(projectRef.Owner, projectRef.Repo) {
```
ablack12, line 2262: Can we check exception before counting num enabled projects so we don't do unnecessary work?

model/project_ref.go
```diff
@@ -2203,6 +2244,38 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		catcher.Errorf("owner repo limit of %d reached for '%s'/'%s'", config.ProjectCreation.RepoProjectLimit, projectRef.Owner, projectRef.Repo)
+	}
+
+	// If the project is not enabled, warn the user while still allowing project creation.
+	if !utility.FromBoolPtr(projectRef.Enabled) && catcher.HasErrors() {
+		grip.Warning(message.WrapError(catcher.Resolve(), message.Fields{
+			"message":            "project creation limit reached",
```
ablack12, line 2269: maybe enabled project limit reached, since we do actually create the project. Also, this logs a message for us but not the user. Spruce work may be needed to allow for that which may involve refactoring this function to return different error codes or something? Could be split to future work. (cc @sophstad if you remember anything to add around showing spruce warnings vs errors)

ablack12 (changes requested):

model/project_ref.go
sophstad: Some changes would need to be made at the resolver level to expose this warning to users. This would involve adding a GraphQL error while also returning the created project ref. We use this pattern for copy project [here](https://github.com/evergreen-ci/evergreen/blob/9c5464e479ebb2660b0867f788c23859a43fcba1/graphql/mutation_resolver.go#L438-L443) and it seems like this could be done very similarly.

bynn: Graphql  became bigger as I worked on it so I am splitting it into a different ticket

https://jira.mongodb.org/browse/EVG-18784

model/project_ref_test.go
```diff
@@ -140,6 +140,61 @@ func TestFindMergedProjectRef(t *testing.T) {
...
+	assert.NoError(t, enabled2.Insert())
+	disabled1 := &ProjectRef{
+		Id:      "disabled1",
+		Owner:   "mongodb",
+		Repo:    "mci",
+		Enabled: utility.FalsePtr(),
+	}
+	assert.NoError(t, disabled1.Insert())
+	disabled2 := &ProjectRef{
+		Id:      "disabled2",
```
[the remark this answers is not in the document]
bynn: fixed

model/project_ref_test.go
```diff
@@ -140,6 +140,61 @@ func TestFindMergedProjectRef(t *testing.T) {
...
+		Owner:   "mongodb",
+		Repo:    "mci",
+		Enabled: utility.TruePtr(),
+	}
+	assert.NoError(t, disabled2.Insert())
+	enabledByRepo := &ProjectRef{
```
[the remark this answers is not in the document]
bynn: done

model/project_ref.go
ablack12: I think this still needs to be addressed

model/project_ref.go
```diff
@@ -2203,6 +2250,38 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		return errors.Wrap(err, "getting number of projects")
+	}
+	if allEnabledProjects >= config.ProjectCreation.TotalProjectLimit {
+		catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
+	}
+	enabledOwnerRepoProjects, err := GetNumberOfEnabledProjectsForOwnerRepo(projectRef.Owner, projectRef.Repo)
+	if err != nil {
+		return errors.Wrapf(err, "getting number of projects for '%s'/'%s'", projectRef.Owner, projectRef.Repo)
+	}
+	if !config.ProjectCreation.IsExceptionToRepoLimit(projectRef.Owner, projectRef.Repo) && enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit {
```
ablack12, line 2268: In another thread I mentioned checking exception before counting projects to avoid a query; making a new thread since that one was outdated but still needs to be addressed

model/project_ref.go
```diff
@@ -921,6 +921,47 @@ func FindMergedProjectRef(identifier string, version string, includeProjectConfi
...
+
+func countEnabledProjects(owner, repo string) (int, error) {
+	pipeline := projectRefPipelineForValueIsBool(ProjectRefEnabledKey, RepoRefEnabledKey, true)
+	if owner != "" && repo != "" {
+		// Check owner and repos in project ref or repo ref.
+		query := bson.M{
+			"$or": []bson.M{
+				{ProjectRefOwnerKey: owner, ProjectRefRepoKey: repo},
+				{"repo_ref": bson.M{
+					"$elemMatch": bson.M{RepoRefOwnerKey: owner, RepoRefRepoKey: repo},
```
ablack12, line 945: Owner/Repo isn't supposed to need to be inherited from the repo; all projects explicitly define it (this is why in the projectRef these fields aren't pointers).

ablack12 (changes requested):

model/project_ref.go
ablack12: Should be able to just use https://github.com/evergreen-ci/evergreen/blob/f4f422a9c06a80fa8069d622d943cfb7aeb514f2/model/project_ref.go#L1326

model/project_ref.go
bynn: ahh sorry

model/project_ref.go

model/project_ref.go
```diff
@@ -2203,6 +2242,40 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		enabledOwnerRepoProjects, err := GetNumberOfEnabledProjectsForOwnerRepo(projectRef.Owner, projectRef.Repo)
+		if err != nil {
+			return errors.Wrapf(err, "getting number of projects for '%s/%s'", projectRef.Owner, projectRef.Repo)
+		}
+		if enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit {
+			catcher.Errorf("owner repo limit of %d reached for '%s/%s'", config.ProjectCreation.RepoProjectLimit, projectRef.Owner, projectRef.Repo)
+		}
+	}
+
+	// If the project is not enabled, warn the user while still allowing project creation.
```
ablack12, line 2266: re https://github.com/evergreen-ci/evergreen/pull/6181#discussion_r1081957527 this doesn't actually warn the user, it just warns us. Is this error worth logging? It seems more worth it to return an error if you try to enable a project that's over the limit. There are two things we want to do:

1) At project creation, the project is _always_ disabled, so we just want to show users a warning on the UI that they won't be able to enable it since these checks aren't met.

2) When trying to enable a project, we want to check the above constraints and return an error.
Maybe we want this function to return the error no matter what, so then we can use it for (1) to create the project, but check projectRef.Enabled and return a potential warning, and (2) to enable the project, but check projectRef.Enabled and return an error.

model/project_ref.go
```diff
@@ -2203,6 +2242,40 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+	}
+	if allEnabledProjects >= config.ProjectCreation.TotalProjectLimit {
+		catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
+	}
+	if !config.ProjectCreation.IsExceptionToRepoLimit(projectRef.Owner, projectRef.Repo) {
+		enabledOwnerRepoProjects, err := GetNumberOfEnabledProjectsForOwnerRepo(projectRef.Owner, projectRef.Repo)
+		if err != nil {
+			return errors.Wrapf(err, "getting number of projects for '%s/%s'", projectRef.Owner, projectRef.Repo)
+		}
+		if enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit {
```
ablack12, line 2261: This function doesn't check that these things could be empty; if totalProjectLimit and RepoProjectLimit are empty we should return from this function early so that we don't just completely block project creation.

model/project_ref.go
bynn: return a boolean to see if we want to error or warn
will handle repo projects in another ticket
creating will also be handled in a different ticket

bynn: returns a bool to see if we want to error or warn
now validates saving project for enabling validation.

model/project_ref.go
```diff
@@ -2203,6 +2242,53 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 	return res, nil
 }
 
+// ValidateProjectCreation returns an error if the total or owner/repo project limit set by the admin settings
+// has been reached. If the project is not enabled, it will grip a warning but not error.
```
malikchaya2, line 2246: `If the project is not enabled, it will grip a warning but not error.`
Is this being done in this function, or is that how we expect the function to be used?

model/project_ref.go
```diff
@@ -1867,15 +1900,21 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+			// Allow a user to modify owner and repo only if they are editing an unattached project
+			if err := p.ValidateOwnerAndRepo(config.GithubOrgs); err != nil {
 				return false, errors.Wrap(err, "validating new owner/repo")
 			}
 
 			setUpdate[ProjectRefOwnerKey] = p.Owner
 			setUpdate[ProjectRefRepoKey] = p.Repo
+
+			// Cannot enable projects if the project creation limits have been reached.
+			err, _ = ValidateProjectCreation(projectId, config, p)
```
malikchaya2, line 1914: Why is this not doing anything with the `shouldError` bool?

malikchaya2 (changes requested):

model/project_ref.go
```diff
@@ -921,6 +921,37 @@ func FindMergedProjectRef(identifier string, version string, includeProjectConfi
...
 
+func GetNumberOfProjects() (int, error) {
+	query := db.Query(
+		bson.M{
+			ProjectRefEnabledKey: true,
+		},
+	)
+	if num, err := db.CountQ(ProjectRefCollection, query); err != nil {
+		return 0, errors.Wrap(err, "counting projects")
+	} else {
```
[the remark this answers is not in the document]
bynn: outdated

model/project_ref.go
malikchaya2: I still curious about why the shouldError field is not used here to decide if we should error on line 1910?

malikchaya2 (changes requested):
one last question

model/project_ref.go
bynn: ah its because should error doesnt matter if there is no error 
the boolean just tells us what we should do with the error (ie if it's false we will warn them instead of erroring)

malikchaya2: At this point we don't know if there is an error or not because we are checking it on line 1910. What if there's an error and shouldError is false, should we not change the behavior on 1911 to not return the error?

bynn: ah sorry, I didn't look closely enough. 
I added a check and a todo so we can warn users once that ticket is done

malikchaya2 (approved):

PR #6186, opened 2023-01-13
EVG-18507 increment idle time

ybrill:
[EVG-18507](https://jira.mongodb.org/browse/EVG-18507)

### Description 
We used to increment the idle time of a host when we terminated it so the time from when the last task to run until termination time was included in the tally. Right now, though, I think we only increment when we start a task.

I think this might have been a regression in https://github.com/evergreen-ci/evergreen/commit/4a7a32041d8bd02f4784321820eb2655f7476df2 (specifically [here](https://github.com/evergreen-ci/evergreen/commit/4a7a32041d8bd02f4784321820eb2655f7476df2#diff-6dd488a00c797fb521cd37e03c8869d4fb82c17ee75a9c79eec5648ed8f7a05bL364-L366)).

### Testing 
I'll try it out in staging when I get a chance 😄

### DIFF
--- a/units/host_termination.go
+++ b/units/host_termination.go
@@ -319,7 +319,7 @@ func (j *hostTerminationJob) incrementIdleTime(ctx context.Context) error {
 	}
 
 	idleTimeStartsAt := j.host.LastTaskCompletedTime
-	if idleTimeStartsAt.IsZero() || idleTimeStartsAt == utility.ZeroTime {
+	if utility.IsZeroTime(idleTimeStartsAt) {
 		idleTimeStartsAt = j.host.StartTime
 	}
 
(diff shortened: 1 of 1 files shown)


ybrill:
This checks out on staging: we log the sum of the idle time that includes the final time before it was terminated.
For [this host](https://spruce-staging.corp.mongodb.com/host/i-064adb4dde8672440) we logged 334 seconds, which is the sum of the time before we started running the task and the time after.

ybrill:
It's worth mentioning that this still isn't _completely_ accurate since we're counting the idle time before we start running a task from the host's start time when in reality we're only billed from when the host enters the running state.
It might be good to get a handle on this. I opened [EVG-18744](https://jira.mongodb.org/browse/EVG-18744).

julianedwards (approved):

PR #6190, opened 2023-01-17
EVG-16926  Make max retries configurable for system failed tasks on the commit queue 

malikchaya2:
[EVG-16926](https://jira.mongodb.org/browse/EVG-16926)

### Description 
We've seen a situation where a task in the commit queue ran a full 10 times because it caused system unresponsive failures, which caused the next batch to wait practically overnight. To avoid situations like that, we want to lower the amount of times system failed tasks can be restarted on the commit queue. This will be configurable on the admin page, and will be set to 5 to start out with.  

### Testing 
Set to five and merged something on staging to make sure it didn't break it.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -331,6 +331,14 @@ var TaskNonGenericFailureStatuses = []string{
 	TaskSystemTimedOut,
 }
 
+var TaskSystemFailures = []string{
+	TaskSystemFailed,
+	TaskTimedOut,
+	TaskSystemUnresponse,
+	TaskSystemTimedOut,
+	TaskTestTimedOut,
+}
+
 // TaskFailureStatuses represent all the ways that a completed task can fail,
 // inclusive of display statuses such as system failures.
 var TaskFailureStatuses = append([]string{TaskFailed}, TaskNonGenericFailureStatuses...)
@@ -355,6 +363,10 @@ func IsFinishedTaskStatus(status string) bool {
 
 func IsFailedTaskStatus(status string) bool {
 	return utility.StringSliceContains(TaskFailureStatuses, status)
+}
+
+func IsSystemFailedTaskStatus(status string) bool {
+	return utility.StringSliceContains(TaskSystemFailures, status)
 }
 
 func IsValidTaskEndStatus(status string) bool {
@@ -830,6 +842,10 @@ func IsGitHubPatchRequester(requester string) bool {
 
 func IsGitTagRequester(requester string) bool {
 	return requester == GitTagRequester
+}
+
+func IsCommitQueueRequester(requester string) bool {
+	return requester == MergeTestRequester
 }
 
 func ShouldConsiderBatchtime(requester string) bool {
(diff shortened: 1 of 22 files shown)


model/task_lifecycle.go
```diff
@@ -307,8 +307,13 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 
 	var execTask *task.Task
 
+	reachedMaxExecutions := t.Execution >= evergreen.MaxTaskExecution
+
+	if evergreen.IsCommitQueueRequester(t.Requester) && evergreen.IsSystemFailedTaskStatus(t.Status) {
+		reachedMaxExecutions = t.Execution >= evergreen.GetEnvironment().Settings().CommitQueue.MaxSystemFailedTaskRetries
```
ablack12, line 313: Can we verify that MaxSystemFailedTaskRetries isn't set to 0 before doing this?

model/task_lifecycle.go
```diff
@@ -307,8 +307,13 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
...
+	}
 	// if we've reached the max number of executions for this task, mark it as finished and failed
-	if t.Execution >= evergreen.MaxTaskExecution {
+	if reachedMaxExecutions {
 		// restarting from the UI bypasses the restart cap
 		msg := fmt.Sprintf("task '%s' reached max execution %d: ", t.Id, evergreen.MaxTaskExecution)
```
ablack12, line 318: Need to update message to use the right max execution

globals.go
```diff
@@ -832,6 +844,10 @@ func IsGitTagRequester(requester string) bool {
 	return requester == GitTagRequester
 }
 
+func IsCommitQueueRequester(requester string) bool {
+	return requester == MergeTestRequester
```
ablack12, line 848: Can we replace other instances of `t.Requester == evergreen.MergeTestRequester` with this?

ablack12 (changes requested):
We also want to check this here https://github.com/evergreen-ci/evergreen/blob/043e736edfc993f7e83898bb6d59091eba1a0858/model/task_lifecycle.go#L1944

Can we also add a test?

model/task_lifecycle.go
```diff
@@ -307,10 +307,18 @@ func TryResetTask(taskId, user, origin string, detail *apimodels.TaskEndDetail)
 
 	var execTask *task.Task
 
+	maxExecution := evergreen.MaxTaskExecution
+
+	if evergreen.IsCommitQueueRequester(t.Requester) && evergreen.IsSystemFailedTaskStatus(t.Status) {
+		maxSystemFailedTaskRetries := evergreen.GetEnvironment().Settings().CommitQueue.MaxSystemFailedTaskRetries
+		if maxSystemFailedTaskRetries != 0 {
```
ablack12, line 314: Preference for > 0 just to be safe

model/task_lifecycle_test.go
```diff
@@ -2228,6 +2228,20 @@ func TestTryResetTask(t *testing.T) {
 			So(v.Insert(), ShouldBeNil)
 			So(anotherTask.Insert(), ShouldBeNil)
 
+			commitQueueMax := evergreen.GetEnvironment().Settings().CommitQueue.MaxSystemFailedTaskRetries
```
ablack12, line 2231: Do we set this MaxSystemFailedTaskRetries to something for this test? I'm a little confused about how this reset doesn't work, since I'd assume that this number is just 0 but I think i'm missing something.

ablack12 (approved):
LGTM with a couple things

bynn (approved):

PR #6191, opened 2023-01-17
EVG-18677 Allow inactive generated tasks to be scheduled in the patch reconfigure page

hadjri:
[EVG-18677](https://jira.mongodb.org/browse/EVG-18677)

### Description 
A task generator may generate some tasks, but if they are not activated they will not show up in the Spruce UI even though they do exist. They can be viewed in the legacy UI however ([example](https://evergreen-staging.corp.mongodb.com/version/63c72372b237363f91ff4b3b)). In Spruce if you select these seemingly nonexistent tasks in the "Reconfigure tasks/variants" and hit schedule, nothing happens because this function is using the [SchedulePatch](https://github.com/evergreen-ci/evergreen/blob/main/graphql/mutation_resolver.go#L199) resolver, which will only create new tasks, and cannot activate existing inactive tasks. A current workaround is if you hit the Schedule button in the top right of Spruce instead it will work because it uses a different [ScheduleTasks](https://github.com/evergreen-ci/evergreen/blob/main/graphql/mutation_resolver.go#L900) resolver which does explicitly activate the selected tasks.

Introduced a change to `SchedulePatch` such that if we encounter an existing task, instead of skipping it, check if it is inactive and if so, activate it with `SetActiveState`
### Testing 
Added unit test confirming that if a task generator generates a bunch of tasks but does not activate all of them, that the inactive tasks can later be selected in the "Reconfigure tasks/variants" field in Spruce and they get scheduled.

Tested in staging.


model/lifecycle.go
```diff
@@ -1703,14 +1703,24 @@ func addNewTasks(ctx context.Context, creationInfo TaskCreationInfo, existingBui
 		// a record in the TVPairSet indicating that it should exist
 		tasksToAdd := []string{}
 		for _, taskName := range creationInfo.Pairs.ExecTasks.TaskNames(b.BuildVariant) {
-			if ok := existingTasksIndex[taskName]; ok {
+			if t, ok := existingTasksIndex[taskName]; ok {
+				if !t.Activated {
+					if err = SetActiveState(evergreen.DefaultTaskActivator, true, *t); err != nil {
+						return nil, errors.Wrapf(err, "setting task '%s' to active", t.Id)
```
Kimchelly, line 1709: nit: can we mention in the error wrapping message that we're activating an execution task rather than a normal task here?

rest/route/patch_test.go
```diff
@@ -865,3 +867,263 @@ buildvariants:
 	assert.NoError(t, err)
 	assert.Len(t, tasks, 4)
 }
+
+func TestSchedulePatchActivatesInactiveTasks(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	var generatedProject = []string{`
```
Kimchelly, line 874: nit: there's no need to declare this as a `var` since `generatedProject := []string{...` is short-hand for `var generatedProject = []string{...` anyways. Was this supposed to be a `const`?

model/lifecycle.go
```diff
@@ -1703,14 +1703,24 @@ func addNewTasks(ctx context.Context, creationInfo TaskCreationInfo, existingBui
 		// a record in the TVPairSet indicating that it should exist
 		tasksToAdd := []string{}
 		for _, taskName := range creationInfo.Pairs.ExecTasks.TaskNames(b.BuildVariant) {
-			if ok := existingTasksIndex[taskName]; ok {
+			if t, ok := existingTasksIndex[taskName]; ok {
+				if !t.Activated {
+					if err = SetActiveState(evergreen.DefaultTaskActivator, true, *t); err != nil {
```
ablack12, line 1708: SetActiveState also updates build/version status each time it's called which seems a bit heavy. Can we instead save these taskIds / buildIds to include in the calls for updating build activation and task activation at the end of this function?

rest/route/patch_test.go
hadjri: Yes it was, but turns out that can't be done for an array. Changed to `:=` notation

model/lifecycle.go
hadjri: That is fair - since `SetActiveState` accepts a list of tasks, I changed it to collect tasks we want to activate and then only invoke the function once after we finish collecting.

hadjri:
> I think this change sounds good to me, but one question I do have here is that, if the generated tasks exist but aren't active yet, should they be appearing in the UI? It seems like this workaround that requires the reconfigure page would be less of an issue if users could click on inactive generated tasks, then press the schedule button for them.

That is a good question, I'm not sure the answer from a UI perspective, on one hand I see why it's useful to abstract away inactive tasks from the user for readability purposes, but I have noticed sometimes users think a task does not exist when it actually does (but it's just inactive). CC @khelif96 @evergreen-ci/evg-ui is this anything that's been discussed in the past?

Kimchelly (commented):
The code itself LGTM, gonna wait for a response on the inactive task thing. FWIW, it seems like [Spruce shows inactive tasks that we set with `activate: false`](https://spruce.mongodb.com/version/evergreen_ceea5c312d4aff3e9b608413b7d1d8692288d386/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=undispatched-umbrella,unscheduled,aborted,blocked&variant=%5Eubuntu1604%24) on the one patch that I looked at. Maybe generate.tasks does something different so that it doesn't show up?

hadjri:
> The code itself LGTM, gonna wait for a response on the inactive task thing. FWIW, it seems like [Spruce shows inactive tasks that we explicitly set to `activate: false`](https://spruce.mongodb.com/version/evergreen_ceea5c312d4aff3e9b608413b7d1d8692288d386/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=undispatched-umbrella,unscheduled,aborted,blocked&variant=%5Eubuntu1604%24) on the one patch that I looked at. Maybe generate.tasks does something different so that it doesn't show up?

Mainline commits such as the link above will always show all of their tasks in Spruce, this behavior is [exclusive to patches](https://github.com/evergreen-ci/evergreen/blob/main/graphql/version_resolver.go#L309). I found [this ticket](https://jira.mongodb.org/browse/EVG-14941) that mentions that this is explicitly intended behavior, citing the fact that server has a generator that creates hundreds of tasks, but often a user only wants to schedule a small subset of them for a patch.

Kimchelly (approved):
SGTM - seems like a weird special consideration just for the server (especially since the inactive tasks show up on the waterfall), but I guess if it's already like that, then it's fine. 🤷

khelif96:
> > I think this change sounds good to me, but one question I do have here is that, if the generated tasks exist but aren't active yet, should they be appearing in the UI? It seems like this workaround that requires the reconfigure page would be less of an issue if users could click on inactive generated tasks, then press the schedule button for them.
> 
> That is a good question, I'm not sure the answer from a UI perspective, on one hand I see why it's useful to abstract away inactive tasks from the user for readability purposes, but I have also noticed sometimes users think a task does not exist when it actually does (but it's just inactive). CC @khelif96 @evergreen-ci/evg-ui is this anything that's been discussed in the past?

We intentionally hide tasks that have never been activated for patch builds. But surface them for mainline commits. Users are able to schedule inactive tasks using the "schedule" Button on a version page. 
https://github.com/evergreen-ci/evergreen/pull/6064 
image

model/lifecycle.go
```diff
@@ -1702,19 +1702,29 @@ func addNewTasks(ctx context.Context, creationInfo TaskCreationInfo, existingBui
...
-			if ok := existingTasksIndex[taskName]; ok {
+			if t, ok := existingTasksIndex[taskName]; ok {
+				if !t.Activated {
+					existingTasksToActivate = append(existingTasksToActivate, t)
+				}
 				continue
 			}
 			displayTasksToAdd = append(displayTasksToAdd, taskName)
 		}
+		if err = SetActiveState(evergreen.DefaultTaskActivator, true, existingTasksToActivate...); err != nil {
```
ablack12, line 1725: We should only call this if existingTasksToActivate isn't empty, right? Also would we then need to add this build to buildIdsToActivate?

ablack12: I'm slightly wondering if this would be less edge-casey if we just made a new function activateExistingTasks that the UI calls, rather than putting this into addNewTasks. I don't think it would cost so much more to rebuild the existing tasks index. That being said this is also fine if you prefer it and/or if we need this for all places addNewTasks is called (like I see we tested using the generate tasks job); we should update the function comment though and add a couple of comments here.

ablack12 (changes requested):

model/lifecycle.go
hadjri: `SetActiveState` would noop with an empty task list input, but to be more explicit I put it behind a conditional.  It also handles build activation [on its own](https://gist.github.com/hadjri/23262276c9ace0f05efe5ee3e804272c#file-task_lifecycle-go-L106) so I don't think the second part is necessary.

Really the only function that needs this behavior is `ConfigurePatch`, not the task generation process, so I moved this logic into a new helper as you suggested and propagated the call up one level into `addNewTasksAndBuildsForPatch` for cleanliness purposes.

model/lifecycle.go
```diff
@@ -1771,6 +1771,34 @@ func addNewTasks(ctx context.Context, creationInfo TaskCreationInfo, existingBui
 	return activatedTaskIds, nil
 }
 
+// activateExistingInactiveTasks will find existing inactive tasks in the patch that need to be activated as
+// part of the patch re-configuration.
+func activateExistingInactiveTasks(creationInfo TaskCreationInfo, existingBuilds []build.Build) error {
+	for _, b := range existingBuilds {
+		tasksInBuild, err := task.FindAll(db.Query(task.ByBuildId(b.Id)))
```
ablack12, line 1778: can we readd withFields displayName and Activated? Really it'd be nice if we could refactor out the call to find existingTasks and pass it into addNewTasks and activateExistingInactiveTasks but it may not be worth it for this relatively inexpensive query.

model/lifecycle.go
```diff
@@ -1771,6 +1771,34 @@ func addNewTasks(ctx context.Context, creationInfo TaskCreationInfo, existingBui
...
+			existingTasksIndex[tasksInBuild[i].DisplayName] = tasksInBuild[i]
+		}
+		existingTasksToActivate := []task.Task{}
+		execAndDisplayTasks := append(creationInfo.Pairs.ExecTasks.TaskNames(b.BuildVariant), creationInfo.Pairs.DisplayTasks.TaskNames(b.BuildVariant)...)
+		for _, taskName := range execAndDisplayTasks {
+			if t, ok := existingTasksIndex[taskName]; ok && !t.Activated {
+				existingTasksToActivate = append(existingTasksToActivate, t)
+			}
+		}
+		if len(existingTasksToActivate) > 0 {
```
ablack12, line 1793: Can we call this just once outside the for loop since it doesn't require all tasks be from the same build?

ablack12 (approved):
LGTM with small changes

model/lifecycle.go
hadjri: Sure, `BuildIdKey` and `VersionKey` need to be added as well for `SetActiveState` to work, I'll add those too.

PR #6192, opened 2023-01-18
EVG-18731 by the second billing for windows

ybrill:
[EVG-18731](https://jira.mongodb.org/browse/EVG-18731)

### Description 
Right now Evergreen thinks that Windows is billed by the hour so [we hesitate to terminate idle Windows hosts that aren't close to the end of an hour boundary](https://github.com/evergreen-ci/evergreen/blob/ad9cef43a0f7452cd0ea15aed2565a511429f36d/units/host_monitoring_idle_termination.go#L240-L245). This, in turn, motivates us to limit the Windows host pools so we can minimize idle time. This creates a poor user experience for Windows testing.

We've discovered that AWS switched Windows billing to by-the-second. This PR will make Evergreen terminate them at will. Once this goes in we will be able to raise max hosts for Windows.

### Testing 
Just the unit test. I'll try to get access to the billing info so I can check that this isn't making us spend radically more.

### DIFF
--- a/cloud/ec2_test.go
+++ b/cloud/ec2_test.go
@@ -1021,12 +1021,13 @@ func (s *EC2Suite) TestTimeTilNextPaymentLinux() {
 	s.Equal(time.Second, s.onDemandManager.TimeTilNextPayment(h))
 }
 
-func (s *EC2Suite) TestTimeTilNextPaymentWindows() {
+func (s *EC2Suite) TestTimeTilNextPaymentSUSE() {
 	now := time.Now()
 	thirtyMinutesAgo := now.Add(-30 * time.Minute)
 	h := &host.Host{
 		Distro: distro.Distro{
-			Arch: "windows",
+			Id:   "suse15-large",
+			Arch: "linux",
 		},
 		CreationTime: thirtyMinutesAgo,
 		StartTime:    thirtyMinutesAgo.Add(time.Minute),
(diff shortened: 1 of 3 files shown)


cloud/ec2_test.go
```diff
@@ -1021,12 +1021,13 @@ func (s *EC2Suite) TestTimeTilNextPaymentLinux() {
 	s.Equal(time.Second, s.onDemandManager.TimeTilNextPayment(h))
 }
 
-func (s *EC2Suite) TestTimeTilNextPaymentWindows() {
+func (s *EC2Suite) TestTimeTilNextPaymentSUSE() {
```
johndaniels, line 1024: Lol at the rename :D

cloud/ec2_util.go
```diff
@@ -216,19 +224,19 @@ func timeTilNextEC2Payment(h *host.Host) time.Duration {
 	return time.Second
 }
 
-// UsesHourlyBilling checks if a distro name to see if it is billed hourly,
-// and returns true if so (for example, most linux distros are by-the-minute).
```
johndaniels, line 220: Good catch on the per-minute vs per-second

johndaniels (approved):

cloud/ec2_util.go
ybrill: thank you! 
I wonder if it used to be per-minute and that also changed or if we always had it wrong.

PR #6196, opened 2023-01-19
EVG-18737: fix patch check for version notification

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18737

### Description 
This fixes an issue introduced in #6168 which assumed that all versions are associated with patches. All finalized patches have a version, but not all versions have a patch.

### Testing 
Added a unit test for version trigger.


trigger/patch.go
```diff
@@ -194,6 +194,14 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
 	if err != nil {
 		return nil, errors.Wrap(err, "getting collective status for patch")
 	}
+	grip.NoticeWhen(collectiveStatus != t.data.Status, message.Fields{
```
malikchaya2, line 197: > Add a log for when the patch's collective status differs from the event's status. If the bug described in [EVG-18561](https://jira.mongodb.org/browse/EVG-18561) is fixed, this should never log and the collective status query should be removed.

This was added as a fix to that bug, so I don't think it can be removed.

trigger/version.go
```diff
@@ -107,9 +108,21 @@ func (t *versionTriggers) makeData(sub *event.Subscription, pastTenseOverride st
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
 	}
 
-	collectiveStatus, err := patch.CollectiveStatus(t.version.Id)
-	if err != nil {
-		return nil, errors.Wrap(err, "getting collective status for patch")
+	var status string
```
malikchaya2, line 111: if it is not a patch requester, the status needs to be set to `t.data.Status`

malikchaya2 (changes requested):

trigger/patch.go
Kimchelly: I guess the issue with this fix is that it's not clear if the issue why the statuses differ? It seems like it's not supposed to change statuses once it's in a finished state, so fixing up the issue right before we send a notification seems too late.

trigger/version.go
Kimchelly: Ah yeah, I forgot to push the commit with the actual else statement here.

trigger/patch.go
malikchaya2: I agree. It was only added because the bug is not reproducible and not findable via code reading.

malikchaya2 (approved):

PR #6198, opened 2023-01-19
EVG-14186 Allow PR commit queue enqueuing via Spruce UI

hadjri:
[EVG-14186](https://jira.mongodb.org/browse/EVG-14186)

### Description 
Currently, the only way to add a PR to the commit queue is via the `evergreen merge` comment. It isn't possible to add it to the queue by clicking the "Enqueue to commit queue" in Spruce.

Change introduced to generalize the `commitQueueEnqueue` method on the github hook API into something that can also be used by the graphQL mutation resolvers that run when "Enqueue to CQ" gets clicked in Spruce. It has been refactored and renamed to a function in the data package named `EnqueuePRToCommitQueue`. It now also returns an `*restModel.APIPatch` to be consistent with the response expected by the resolver.

### Testing 
Tested in staging. I was able to add [this PR](https://github.com/evergreen-ci/commit-queue-sandbox/pull/419) to the commit queue without typing `evergreen merge`


rest/data/commit_queue.go
```diff
@@ -280,6 +283,103 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
 	return hasPermission, nil
 }
 
+// EnqueuePRToCommitQueue enqueues an existing pull request to the commit queue.
```
Kimchelly, line 286: The phrasing on the doc comment might be a little vague, since it's unclear it means to "enqueue a pull request". Can we phrase this as something more Evergreen-ish, like "enqueues a commit queue item to test and merge a PR"?

rest/data/commit_queue.go
```diff
@@ -280,6 +283,103 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+	}
+	authorized, err := sc.IsAuthorizedToPatchAndMerge(ctx, settings, userRepo)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting user info from GitHub API")
+	}
+	if !authorized {
+		return nil, errors.Errorf("user '%s' is not authorized to merge", userRepo.Username)
+	}
+
+	prNum := info.PR
```
Kimchelly, line 301: nit: It seems like the original logic only assigned this to a new variable because it had to dereference the GitHub event pointer. Can we just use `info.PR` in all the places that need it rather than making a new local variable?

rest/data/commit_queue.go
```diff
@@ -280,6 +283,103 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+	_, err = EnqueueItem(projectRef.Id, item, false)
+	if err != nil {
+		return nil, errors.Wrap(err, "enqueueing commit queue item")
+	}
+
+	if pr == nil || pr.Head == nil || pr.Head.SHA == nil {
+		return nil, errors.New("PR contains no head branch SHA")
+	}
+	pushJob := units.NewGithubStatusUpdateJobForPushToCommitQueue(userRepo.Owner, userRepo.Repo, *pr.Head.SHA, prNum, patchDoc.Id.Hex())
+	q := evergreen.GetEnvironment().LocalQueue()
```
Kimchelly, line 363: If we're reusing the global env in the function, can we pass it in as a parameter to `EnqueuePRToCommitQueue`? It makes it somewhat easier to unit test this function when you can choose which env gets passed in, such as if you want to use the mock environment instead of the global testing one. The only inconvenience is you have to pass the environment down through all the function calls, but it's worth it to get the improved testability.

rest/data/commit_queue.go
```diff
@@ -388,3 +488,30 @@ func GetAdditionalPatches(patchId string) ([]string, error) {
 		Message:    errors.Errorf("patch '%s' not found in commit queue", patchId).Error(),
 	}
 }
+
+func requireSigned(ctx context.Context, userRepo UserRepoInfo, prNum int) error {
+	settings, err := evergreen.GetConfig()
```
Kimchelly, line 493: Not that important, but we typically try not loading the settings via `evergreen.GetConfig()` unless there's a specific reason, and instead use the `evergreen.GetEnvironment().Settings()` (which is already passed into `EnqueuePRToCommitQueue`). Can we do that here, or is it critical to load the current GitHub token without restarting the app servers?

rest/data/commit_queue.go
```diff
@@ -388,3 +488,30 @@ func GetAdditionalPatches(patchId string) ([]string, error) {
 		Message:    errors.Errorf("patch '%s' not found in commit queue", patchId).Error(),
 	}
 }
+
+func requireSigned(ctx context.Context, userRepo UserRepoInfo, prNum int) error {
```
Kimchelly, line 492: nit: can we call this something like `checkSignedCommit` or something?

rest/data/commit_queue.go
```diff
@@ -388,3 +488,30 @@ func GetAdditionalPatches(patchId string) ([]string, error) {
...
+	}
+
+	commits, err := thirdparty.GetGithubPullRequestCommits(ctx, githubToken, userRepo.Owner, userRepo.Repo, prNum)
+	if err != nil {
+		return errors.Wrap(err, "getting GitHub commits")
+	}
+
+	for _, c := range commits {
+		commit := c.GetCommit()
+		if commit.Verification != nil && !utility.FromBoolPtr(commit.Verification.Verified) &&
```
Kimchelly, line 510: Are these conditions supposed to be `||` instead of `&&`? It seems like this function should error if any of the conditions are false because any one of them might indicate the commit is not signed.

Kimchelly (changes requested):

model/commitqueue/commit_queue.go
```diff
@@ -207,3 +207,11 @@ func ClearAllCommitQueues() (int, error) {
 
 	return clearedCount, nil
 }
+
+type PRInfo struct {
```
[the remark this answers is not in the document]
hadjri: Renamed to `EnqueuePRInfo` as well for clarity.

rest/data/commit_queue.go
hadjri: Done

rest/data/commit_queue.go
hadjri: Done

rest/data/commit_queue.go
hadjri: Not critical. Done

rest/data/commit_queue.go
hadjri: Done

rest/data/commit_queue.go
hadjri: Hmm, I'm not sure what the intention was when this function was created but that makes sense to me as well

hadjri: Hmm, not sure the context on the original function when it was written, but this makes sense to me as well. Going to change to `||`

rest/data/commit_queue.go
```diff
@@ -280,6 +283,103 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+				"repo":    userRepo.Repo,
+				"pr":      prNum,
+			}))
+			return nil, errors.Wrapf(err, "checking commit signing")
+		}
+	}
+
+	patchDoc, err := sc.AddPatchForPr(ctx, *projectRef, prNum, cqInfo.Modules, cqInfo.MessageOverride)
+	if err != nil {
+		sendErr := thirdparty.SendCommitQueueGithubStatus(evergreen.GetEnvironment(), pr, message.GithubStateFailure, "failed to create patch", "")
```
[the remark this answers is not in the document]
hadjri: Sure - since the logic between the GraphQL resolver and the gihub hook's use of this function has a lot of overlap so instead I've added a flag to skip this line if the caller is not the github hook.

model/commitqueue/commit_queue.go
```diff
@@ -207,3 +207,12 @@ func ClearAllCommitQueues() (int, error) {
 
 	return clearedCount, nil
 }
+
+// EnqueuePRInfo holds information necessary to enqueue a PR to the commit queue.
+type EnqueuePRInfo struct {
```
malikchaya2, line 212: (random comment placement so we can thread)

Note on testing: 
One of the motivations behind this ask was that the PR be closed after it is enqueued using the UI. Can we please test that that happens as expected?

rest/data/commit_queue.go
```diff
@@ -280,6 +283,103 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+	if err != nil {
+		return nil, errors.Wrap(err, "enqueueing commit queue item")
+	}
+
+	if pr == nil || pr.Head == nil || pr.Head.SHA == nil {
+		return nil, errors.New("PR contains no head branch SHA")
+	}
+	pushJob := units.NewGithubStatusUpdateJobForPushToCommitQueue(userRepo.Owner, userRepo.Repo, *pr.Head.SHA, prNum, patchDoc.Id.Hex())
+	q := evergreen.GetEnvironment().LocalQueue()
+	// TODO FIX THIS
```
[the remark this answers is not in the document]
hadjri: Leftover debugging 🤦

Since the github hook already [propagates up](https://github.com/evergreen-ci/evergreen/blob/main/rest/route/github.go#L233) an enqueueing error if encountered, I changed this line to return an error with a more descriptive message that will get wrapped by the GitHub hook's error template.

model/commitqueue/commit_queue.go
hadjri: Yes - I tested it on [this PR](https://github.com/evergreen-ci/commit-queue-sandbox/pull/419) and observed it was closed when I enqueued using the UI.

rest/data/commit_queue.go
```diff
@@ -280,6 +283,95 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+		return nil, errors.Wrapf(err, "getting project for '%s:%s' tracking branch '%s'", userRepo.Owner, userRepo.Repo, baseBranch)
+	}
+	if projectRef == nil {
+		return nil, errors.Errorf("no project with commit queue enabled for '%s:%s' tracking branch '%s'", userRepo.Owner, userRepo.Repo, baseBranch)
+	}
+
+	if utility.FromBoolPtr(projectRef.CommitQueue.RequireSigned) {
+		err = checkSignedCommit(ctx, settings, userRepo, info.PR)
+		if err != nil {
+			sendErr := thirdparty.SendCommitQueueGithubStatus(evergreen.GetEnvironment(), pr, message.GithubStateFailure, "can't enqueue with unsigned commits", "")
```
Kimchelly, line 327: Alright, I read this GitHub status logic over again, and it seems like it would be nice to send the GitHub failure back to the user for all error return cases after we've retrieved and validated the PR info so we can send a response but before the item has been successfully put in the commit queue. We could reduce the amount of duplicate `thirdparty.SendCommitQueueGithubStatus` calls by moving everything L321-358 into a helper function and if that helper errors, call `SendCommitQueueGithubStatus`. Something like:

```go
apiPatch, err := tryEnqueueItemForPR(...)
if err != nil {
    if sendErr := thirdparty.SendCommitQueueGithubStatus(...) { log the send error }
    return nil, errors.Wrap(err, ...)
}

func tryEnqueueItemForPR(...) {
    // check signed, check required approval count, create patch, enqueue in commit queue
}
```

rest/data/commit_queue.go
```diff
@@ -280,6 +286,134 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
...
+				"owner":   userRepo.Owner,
+				"repo":    userRepo.Repo,
+				"pr":      info.PR,
+			}))
+			return nil, errors.Wrapf(err, "checking commit signing")
+		}
+	}
+
+	requiredApprovalCount := projectRef.CommitQueue.RequiredApprovalCount
+	if requiredApprovalCount != 0 {
```
Kimchelly, line 339: nit: since we're only using `requiredApprovalCount` in this if block, we can just declare it in the one if block:
```go
if requiredApprovalCount := projectRef.CommitQueue.RequiredApprovalCount; requiredApprovalCount != 0 {
    ...
}
```

malikchaya2 (approved):
Approved with nit

rest/data/commit_queue.go
hadjri: Sounds good, `SendCommitQueueGithubStatus` has an error description input param, so I added a third return value to `tryEnqueueItemForPR` that corresponds with the type of failure encountered, since providing the full `err.Error()` could be a long string.

Kimchelly (approved):

rest/data/commit_queue.go
malikchaya2: based on our logging, it looks like we are encountering empty commits on a pR that only has one commit and it's not empty. We may want to investigate that as well, outside of this PR.

PR #6204, opened 2023-01-24
EVG-18789 testing platform to ubuntu2004

ybrill:
[EVG-18789](https://jira.mongodb.org/browse/EVG-18789)

### Description 
Upgrade the testing platform to Ubuntu 20.04 in advance of upgrading the app servers.
Since the company doesn't build mongo 4.2 for Ubuntu 20.04 this is pushing us to upgrade the db to 4.4. This is a good thing anyway since we're going to upgrade the db to 4.4 shortly.

### Testing 
goto Description


ybrill:
cc @khelif96 in case you wanted to review the removal of the pre 4.4 test workarounds.

self-tests.yml
```diff
@@ -686,7 +685,7 @@ buildvariants:
       goos: linux
       GOROOT: /opt/golang/go1.16
       RUN_EC2_SPECIFIC_TESTS: true
-      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-aarch64-ubuntu1804-4.2.18.tgz
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-aarch64-ubuntu1804-4.4.18.tgz
```
ybrill, line 688: Left this distro alone on the assumption it's not here to test the app servers

khelif96 (commented):
Could you make this change on spruce and parsley as well? I just want to ensure the e2e tests still run.

self-tests.yml
julianedwards: Is this a reason to continue testing on 18.04?

cmd/mongo-auth/replset_key.txt
```diff
@@ -0,0 +1 @@
+"replica set key"
```
julianedwards, line 1: Is this just dummy data for testing/running locally?

ybrill: yes, it seems that with the new db version mongod requires a secret key for members of the replica set to communicate with each other

self-tests.yml
ybrill: I mean, the agent runs on both ubuntu1804 and ubuntu2004.
I think the idea of this variant is just to make sure we test the ARM architecture.

julianedwards (approved):

ybrill:
Apparently ifdef $(foo) first dereferences foo and then checks if whatever foo referred to is defined ([docs](https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html)). So setting AUTH_ENABLED=1 and checking in the makefile `ifdef $(AUTH_ENABLED)` is checking if a variable named `1` is defined.

If so, my changes from [here](https://github.com/evergreen-ci/evergreen/pull/5410) should have broken the test. They didn't, I think, because anyone can connect to the db. You only need auth to do things.

ybrill:
I verified that [the test fails now when we don't have an authorized user](https://evergreen.mongodb.com/task_log_raw/evergreen_ubuntu2004_test_db_auth_patch_0bff91e1810acb2eb0b26e34e12081bcb4d898d0_63d155483627e01c122e84dc_23_01_25_16_14_08/0?type=T#L432).

ybrill:
@khelif96 are you okay with me merging without spruce/parsley changes this given the discussion on evergreen-ci/spruce#1627?

khelif96:
sgtm I'll file a build request to get node up to date on the 20.04 distros. @ybrill

PR #6206, opened 2023-01-24
EVG-17939 Update agent to cancel heartbeat correctly

bynn:
[EVG-17939](https://jira.mongodb.org/browse/EVG-17939)

### Description 
change the context to cancel in heartbeat because it was not ending before

### Testing 
tested on staging to make sure timeout works
https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu1604_bynntask_patch_7ac7bc95d6b92fe95e934a1859792407acf40be9_63d0ad1e9dbe3232f3912bcf_23_01_25_04_16_40/logs?execution=0

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -489,8 +489,8 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 
 	innerCtx, innerCancel := context.WithCancel(tskCtx)
 
-	// Pass in idle timeout context(innerCtx) to heartbeat to enforce the idle timeout.
-	// Pass in task cancel context to heartbeat because it's responsible for aborting the task.
+	// Pass in idle timeout context to heartbeat to enforce the idle timeout.
+	// Pass in the task context canceller to heartbeat because it's responsible for aborting the task.
 	heartbeat := make(chan string, 1)
 	go a.startHeartbeat(innerCtx, tskCancel, tc, heartbeat)
 	go a.startIdleTimeoutWatch(tskCtx, tc, innerCancel)
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -487,10 +487,10 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 	defer a.killProcs(ctx, tc, false)
 	defer tskCancel()
 
-	heartbeat := make(chan string, 1)
-	go a.startHeartbeat(tskCtx, tskCancel, tc, heartbeat)
-
 	innerCtx, innerCancel := context.WithCancel(tskCtx)
+
+	heartbeat := make(chan string, 1)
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -487,11 +487,11 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 	defer a.killProcs(ctx, tc, false)
 	defer tskCancel()
 
-	heartbeat := make(chan string, 1)
-	go a.startHeartbeat(tskCtx, tskCancel, tc, heartbeat)
-
 	innerCtx, innerCancel := context.WithCancel(tskCtx)
 
+	heartbeat := make(chan string, 1)
+	go a.startHeartbeat(innerCtx, tskCancel, tc, heartbeat)
```
Kimchelly, line 493: Can we add a comment noting that this intentionally passes in the idle timeout context because we want to ensure that there is a deadline on how long the agent can spend handling idle timeout?

Kimchelly (changes requested):

agent/agent.go
bynn: i tried to simplify it? i hope it makes sense

agent/agent.go
```diff
@@ -487,11 +487,12 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 	defer a.killProcs(ctx, tc, false)
 	defer tskCancel()
 
-	heartbeat := make(chan string, 1)
-	go a.startHeartbeat(tskCtx, tskCancel, tc, heartbeat)
-
 	innerCtx, innerCancel := context.WithCancel(tskCtx)
 
+	// Pass in idle timeout context(innerCtx) to heartbeat to enforce the idle timeout.
+	// Pass in task cancel context to heartbeat because it's responsible for aborting the task.
```
Kimchelly, line 493: Pass in the task context canceller

agent/agent.go
```diff
@@ -487,11 +487,12 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (bool, error) {
 	defer a.killProcs(ctx, tc, false)
 	defer tskCancel()
 
-	heartbeat := make(chan string, 1)
-	go a.startHeartbeat(tskCtx, tskCancel, tc, heartbeat)
-
 	innerCtx, innerCancel := context.WithCancel(tskCtx)
 
+	// Pass in idle timeout context(innerCtx) to heartbeat to enforce the idle timeout.
```
Kimchelly, line 492: I don't think we need to specify `innerCtx` here since it's understood to be the idle timeout context. If it's not, then we should probably rename the `innerCtx` to a better name like `idleTimeoutCtx`.

Kimchelly (approved):
LGTM mod some wording suggestions.

PR #6208, opened 2023-01-25
EVG-18811  Fix approval being hidden 

malikchaya2:
[EVG-18811](https://jira.mongodb.org/browse/EVG-18811)

### Description 
We rely on the list of reviews to know if there is the required number of approvals. The default that listReviews returns is 30. On rare occasions, there will be PRs with more than 30 reviews. For example, today we saw a PR with 37 reviews, and the approval being the 37th. This increases the amount of reviews returned to 100.

### DIFF
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -894,7 +894,10 @@ func GetGithubPullRequestReviews(ctx context.Context, token, owner, repo string,
 
 	client := github.NewClient(httpClient)
 
-	reviews, _, err := client.PullRequests.ListReviews(ctx, owner, repo, prNumber, nil)
+	opts := &github.ListOptions{
+		PerPage: 100,
+	}
+	reviews, _, err := client.PullRequests.ListReviews(ctx, owner, repo, prNumber, opts)
 	if err != nil {
 		return nil, err
 	}


bynn (approved):
if the default is 30, is 100 necessary? is there a reason for the number? 

the change looks good though!

malikchaya2:
> if the default is 30, is 100 necessary? is there a reason for the number?
> 
> the change looks good though!

it's just the highest it can go

Kimchelly (changes requested):
Would this be problematic if people submitted tons and tons of PR reviews (say for a very large/risky change)? It seems like listing all of them (even if we have to paginate through multiple `ListReviews` responses would be preferable than capping it arbitrarily at 100.

I know there are some teams that have said they sometimes submit very large code changes for their features, so it doesn't seem impossible that they could end up with lots of reviews.

malikchaya2:
> Would this be problematic if people submitted tons and tons of PR reviews (say for a very large/risky change)? It seems like listing all of them (even if we have to paginate through multiple `ListReviews` responses) would be preferable than capping it arbitrarily at 100.
> 
> I know there are some teams that have said they sometimes submit very large code changes for their features, so it doesn't seem impossible that they could end up with lots of reviews.

That would take a bit longer to implement, so let's do this first to unblock them, and consider doing this later if it feels needed.

Kimchelly (approved):
SGTM

PR #6209, opened 2023-01-25
EVG-18811 Add Pagination to GetGithubPullRequestReviews

malikchaya2:
[EVG-18811](https://jira.mongodb.org/browse/EVG-18811)

### Description 
We sometimes need to fetch more than the first 30 reviews to see all the "approval" reviews. 

### Testing 
Tested on staging with a PR with 37 reviews. 

Note, the github conflicts will go away once [6198](https://github.com/evergreen-ci/evergreen/pull/6198/files) which was temporarily reverted is merged back in.


thirdparty/github.go
```diff
@@ -888,24 +888,28 @@ func GetGithubPullRequestCommits(ctx context.Context, token, owner, repo string,
...
-		return nil, errors.New("no PR reviews received from GitHub")
+	reviews, resp, err := client.PullRequests.ListReviews(ctx, owner, repo, prNumber, opts)
+	if resp != nil {
+		defer resp.Body.Close()
+		if err != nil {
+			return nil, 0, parseGithubErrorResponse(resp)
+		}
+	} else {
+		errMsg := fmt.Sprintf("nil response from query for PR reviews in '%s/%s' prNumber %d : %v", owner, repo, prNumber, err)
+		grip.Error(errMsg)
```
Kimchelly, line 908: I imagine we would log this error anyways at a higher level. Do we need to log it here?

rest/data/commit_queue.go
```diff
@@ -385,15 +385,27 @@ func checkPRApprovals(ctx context.Context, settings *evergreen.Settings, userRep
...
-	}
-
 	var numApprovals int
-	for _, r := range reviews {
-		if r.GetState() == githubReviewApproved {
-			numApprovals += 1
+	reviewsPage := 0
+	var reviews []*github.PullRequestReview
+
+	for numApprovals < requiredApprovalCount {
```
Kimchelly, line 392: Can we turn these nested for loops into a helper like `getNumApprovals(requiredApprovalCount)` that simply returns the number of approvals that it found (up to the required amount)? Would avoid having to think about the for loop `break`'ing behavior.

Kimchelly (changes requested):

rest/data/commit_queue.go
```diff
@@ -385,15 +385,27 @@ func checkPRApprovals(ctx context.Context, settings *evergreen.Settings, userRep
...
-	reviews, err := thirdparty.GetGithubPullRequestReviews(ctx, githubToken, userRepo.Owner, userRepo.Repo, prNum)
-	if err != nil {
-		return errors.Wrap(err, "getting GitHub PR reviews")
-	}
-
 	var numApprovals int
-	for _, r := range reviews {
-		if r.GetState() == githubReviewApproved {
-			numApprovals += 1
+	reviewsPage := 0
```
Kimchelly, line 389: nit: the default int is 0, so we can declare this the same as L388 with `var reviewsPage int`.

rest/data/commit_queue.go
```diff
@@ -385,15 +385,27 @@ func checkPRApprovals(ctx context.Context, settings *evergreen.Settings, userRep
...
+		)
+		if err != nil {
+			return errors.Wrap(err, "getting GitHub PR reviews")
+		}
+		for _, r := range reviews {
+			if r.GetState() == githubReviewApproved {
+				numApprovals += 1
+			}
+			if numApprovals >= requiredApprovalCount {
+				break
```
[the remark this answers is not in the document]
malikchaya2: I'm guessing it would break out of this for loop and then not pass the check on line 392, so we wouldn't fetch the next page. Am I misreading it?

rest/data/commit_queue.go
malikchaya2: It seems to me like that would just move this for loop to another function, I don't know if that would have much of an impact on readability.

thirdparty/github.go
```diff
@@ -888,24 +888,28 @@ func GetGithubPullRequestCommits(ctx context.Context, token, owner, repo string,
...
-func GetGithubPullRequestReviews(ctx context.Context, token, owner, repo string, prNumber int) ([]*github.PullRequestReview, error) {
+func GetGithubPullRequestReviews(ctx context.Context, token, owner, repo string, prNumber int, reviewPage int) ([]*github.PullRequestReview, int, error) {
 	httpClient := getGithubClientRetryWith404s(token, "GetGithubPullRequestCommits")
 	defer utility.PutHTTPClient(httpClient)
 
 	client := github.NewClient(httpClient)
 
 	opts := &github.ListOptions{
-		PerPage: 100,
+		Page: reviewPage,
```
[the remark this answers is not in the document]
malikchaya2: true, added

thirdparty/github.go
malikchaya2: removed

rest/data/commit_queue.go
malikchaya2: true, fixed

rest/data/commit_queue.go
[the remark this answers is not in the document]
Kimchelly: Ah yeah, it would do that. The double-checking approval counts in both loops threw me off.

rest/data/commit_queue.go
Kimchelly: Eh, I don't have that strong of opinions here. Generally, I try to break up double loops that use breaks/continues into helper functions (at least, put the inner loop in its own little function) because it avoids having to think about "am I breaking out of the inner loop or outer loop, does it still work if I just break out of the inner loop" and such. I don't think it's a big deal in this case.

Kimchelly (approved):

malikchaya2:
evergreen retry

PR #6210, opened 2023-01-26
EVG-6777 Unescape description in commit queue email

hadjri:
[EVG-6777](https://jira.mongodb.org/browse/EVG-6777)

### Description 
If the description of a commit queue patch contains special characters, they will get escaped, e.g. `remove the &#39;touch&#39; command`. 

Changed the description field of `commonTemplateData` in the case of making an email payload to the `template.HTML` type which will not be escaped by default.
### Testing 
Added unit test certifying a description with quotations does not get escaped.

### DIFF
--- a/trigger/payloads.go
+++ b/trigger/payloads.go
@@ -42,7 +42,7 @@ type commonTemplateData struct {
 	DisplayName     string
 	Object          string
 	Project         string
-	Description     string
+	Description     template.HTML
 	URL             string
 	PastTenseStatus string
 	Headers         http.Header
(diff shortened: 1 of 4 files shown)


malikchaya2 (approved):

trigger/payloads.go
```diff
@@ -62,6 +62,11 @@ type commonTemplateData struct {
 	emailContent *template.Template
 }
 
+type emailTemplateData struct {
+	commonTemplateData
+	Description template.HTML
```
malikchaya2, line 67: neat!

PR #6212, opened 2023-01-26
EVG-18006 Use select statement for agent commands at risk of blocked goroutines

hadjri:
[EVG-18006](https://jira.mongodb.org/browse/EVG-18006)
### Description 
Some agent commands utilize blocking channels when performing key operations, without protection from the possibility that a value may never get sent to the channel. This would cause an indefinite blocking of the command.

Introduced a `select` statement to these goroutines to handle potential context cancellation.

### DIFF
--- a/agent/command/archive_tarball_create.go
+++ b/agent/command/archive_tarball_create.go
@@ -96,7 +96,13 @@ func (c *tarballCreate) Execute(ctx context.Context,
 		}()
 		var err error
 		filesArchived, err = c.makeArchive(ctx, logger.Execution())
-		errChan <- errors.WithStack(err)
+		select {
+		case errChan <- errors.WithStack(err):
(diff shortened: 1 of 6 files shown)


malikchaya2 (approved):

agent/command/archive_tarball_create.go
```diff
@@ -96,7 +96,13 @@ func (c *tarballCreate) Execute(ctx context.Context,
 		}()
```
Kimchelly, line 96: Won't let me comment on unchanged lines, but L94 needs to exit if the context errors as well, because the defer will hang if there's no consumer for `errChan`.

Kimchelly (changes requested):

Kimchelly (approved):
💯

PR #6214, opened 2023-01-26
EVG-18744 record the time a host starts to be billed

ybrill:
[EVG-18744](https://jira.mongodb.org/browse/EVG-18744)

### Description 
Record the time a host starts to be billed and include it in logged metrics. The AWS docs say that billing starts from when an instance goes to running, but AWS doesn't make it easy to find out exactly when that was. It isn't included in [DescribeInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html) output, for example. This PR pulls it out of the CloudWatch event when it goes to running. The docs don't explicitly say that the event time is the time the event occurred, but [they do say](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events.html) that the service originating the event _can_ set it. It stands to reason that it's at least approximately the time the host went to running.

Some things this will let us find out: 
- how much wasted time we're paying for
- how much of our paid time it's taking for provisioning to finish (~100s in staging today)
- it could be good to compare our vision of billable time against AWS's

### Testing 
Tried it out in staging. It appears to work. The time difference between start time and billing start for my ubuntu2004-small hosts were around 10-20s. Not a huge difference, but it could matter in aggregate. I also wonder if it will depend across distros.


rest/route/agent.go
```diff
@@ -996,7 +996,17 @@ func (h *startTaskHandler) Run(ctx context.Context) gimlet.Responder {
 		}
 
 		idleTimeStartAt := host.LastTaskCompletedTime
-		if idleTimeStartAt.IsZero() || idleTimeStartAt == utility.ZeroTime {
+		if utility.IsZeroTime(idleTimeStartAt) {
+			idleTimeStartAt = host.BillingStartTime
+		}
+		if utility.IsZeroTime(idleTimeStartAt) {
```
johndaniels, line 1002: I feel like doing this identical if depending on the state of idleTimeStartAt is a little awkward.

johndaniels: Does this make sense as a method on host since it repeats, and doing so might avoid this slightly awkward double "if" statement?

johndaniels (changes requested):
Willing to be convinced otherwise on the above suggestion.

rest/route/agent.go
ybrill: When I went to move it onto the host I noticed a very similar method already existed. I tried to consolidate the two.

ybrill:
Along the way I added some additional logging in the task-start-stats which I felt I would have benefitted from trying to investigate [EVG-18862](https://jira.mongodb.org/browse/EVG-18862).

ybrill:
@johndaniels I realized I need a bit more time to get this ready for re-review.
I'll ping you on the PR when it's ready.

model/host/host_test.go
```diff
@@ -2591,7 +2591,7 @@ func TestIsIdleParent(t *testing.T) {
 	assert.True(idle)
 	assert.NoError(err)
 
-	// ios a container --> false
```
johndaniels, line 2574: :D nice catch.

rest/route/host_agent_test.go
```diff
@@ -601,42 +600,16 @@ func TestTaskLifecycleEndpoints(t *testing.T) {
...
-			require.NoError(t, q.Start(ctx))
-			stat = q.Stats(ctx)
 			require.Equal(t, resp.Status(), http.StatusOK)
 			require.NotNil(t, resp)
-			require.Equal(t, stat.Total, 1)
-			amboy.WaitInterval(ctx, env.RemoteQueue(), time.Millisecond)
-
-			counter := 0
-			for job := range q.Results(ctx) {
-				require.NotNil(t, job)
```
johndaniels, line 621: Nice

johndaniels (approved):

rest/route/agent.go
```diff
@@ -1027,6 +1015,50 @@ func (h *startTaskHandler) Run(ctx context.Context) gimlet.Responder {
 	return gimlet.NewJSONResponse(msg)
 }
 
+func logTaskStartMessage(h *host.Host, t *task.Task) {
```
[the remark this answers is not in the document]
ybrill: I factored it out of the stats job.

ybrill:
I realized IdleTime would return 0 for started tasks since the task gets assigned when it's dispatched. It made sense to me to keep the definition of idle time the same wrt billed time we aren't using to run a task, so instead of moving the idle time incrementation to when the task is assigned I divided up the concept of IdleTime so it wouldn't count the currently running task when considering idle time for a host starting a task.

johndaniels (approved):

PR #6215, opened 2023-01-26
EVG-14441 Wrap long commit queue commit messages

hadjri:
[EVG-14441](https://jira.mongodb.org/browse/EVG-14441)

### Description 
Long commit queue messages don't get wrapped which is poor for readability in `git log`.  Made it such that every 72 characters the message will wrap in a way that prevents words from being chopped up in an inconvenient way. Chose 72 characters since it seems to be a [common reccommendation](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting).
### Testing 
Added new unit tests confirming that long messages will get wrapped cleanly regardless of extra whitespace.

### DIFF
--- a/rest/model/commit_queue.go
+++ b/rest/model/commit_queue.go
@@ -2,11 +2,14 @@ package model
 
 import (
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/evergreen-ci/evergreen/model/commitqueue"
 	"github.com/evergreen-ci/utility"
 )
+
+const commitMessageLineLength = 72
 
 type APICommitQueue struct {
 	ProjectID *string              `json:"queue_id"`
@@ -88,11 +91,31 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
 		if index == 1 {
 			data = parseFirstLine(line)
 		} else if index == 2 {
-			data.MessageOverride = line
+			data.MessageOverride = wrapString(line, commitMessageLineLength)
 		}
 	}
 
 	return data
+}
+
+// wrapString manually splits a string into smaller chunks and appends them to a new string.
+func wrapString(str string, limit int) string {
+	var b strings.Builder
+	words := strings.Fields(str)
+	currentLine := ""
+	for _, word := range words {
+		if len(currentLine+" "+word) > limit {
+			b.WriteString(currentLine + "\n")
+			currentLine = ""
+		}
+		if currentLine == "" {
+			currentLine += word
+		} else {
+			currentLine += " " + word
+		}
+	}
+	b.WriteString(currentLine)
+	return b.String()
 }
 
 func parseFirstLine(comment string) GithubCommentCqData {
--- a/rest/model/commit_queue.go
+++ b/rest/model/commit_queue.go
@@ -101,20 +101,32 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
 // wrapString manually splits a string into smaller chunks and appends them to a new string.
 func wrapString(str string, limit int) string {
 	var b strings.Builder
-	words := strings.Fields(str)
 	currentLine := ""
-	for _, word := range words {
-		if len(currentLine+" "+word) > limit {
-			b.WriteString(currentLine + "\n")
+	lines := strings.Split(str, "\n")
+	for i, line := range lines {
+		words := strings.Fields(line)
+		for _, word := range words {
+			lineLength := len(currentLine + word)
+			// Only factor in spaces if the current line is not empty.
+			if len(currentLine) > 0 {
+				lineLength += 1
+			}
+			if lineLength > limit {
+				b.WriteString(currentLine + "\n")
+				currentLine = ""
+			}
+			if currentLine == "" {
+				currentLine += word
+			} else {
+				currentLine += " " + word
+			}
+		}
+		b.WriteString(currentLine)
+		if i < len(lines)-1 {
+			b.WriteString("\n")
 			currentLine = ""
 		}
-		if currentLine == "" {
-			currentLine += word
-		} else {
-			currentLine += " " + word
-		}
 	}
-	b.WriteString(currentLine)
 	return b.String()
 }
 
--- a/rest/model/commit_queue.go
+++ b/rest/model/commit_queue.go
@@ -101,9 +101,9 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
 // wrapString manually splits a string into smaller chunks and appends them to a new string.
 func wrapString(str string, limit int) string {
 	var b strings.Builder
-	currentLine := ""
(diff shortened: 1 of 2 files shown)


rest/model/commit_queue.go
```diff
@@ -88,13 +91,33 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
...
+	currentLine := ""
+	for _, word := range words {
+		if len(currentLine+" "+word) > limit {
+			b.WriteString(currentLine + "\n")
+			currentLine = ""
+		}
+		if currentLine == "" {
+			currentLine += word
+		} else {
+			currentLine += " " + word
```
Kimchelly, line 114: We could use `currentLine = strings.Join(currentLine, word, " ")` to get around the if-else here.

rest/model/commit_queue.go
```diff
@@ -88,13 +91,33 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
...
 	return data
 }
 
+// wrapString manually splits a string into smaller chunks and appends them to a new string.
+func wrapString(str string, limit int) string {
+	var b strings.Builder
+	words := strings.Fields(str)
+	currentLine := ""
+	for _, word := range words {
+		if len(currentLine+" "+word) > limit {
```
Kimchelly, line 107: It seems like here we want to check `len(currentLine + word)` if `currentLine` is empty. Can we also use `strings.Join` here?

rest/model/commit_queue_test.go
```diff
@@ -89,7 +89,16 @@ some more lines
...
+	comment = `evergreen merge --unknown-option blah_blah --module module1:1234 
+This is a very long string that needs to be wrapped around 72 characters for readability.
+Here is another shorter line.
+											Whitespace over here.`
+	data = ParseGitHubComment(comment)
+	assert.Len(data.Modules, 1)
+	assert.Equal(utility.ToStringPtr("module1"), data.Modules[0].Module)
+	assert.Equal(utility.ToStringPtr("1234"), data.Modules[0].Issue)
+	assert.Equal(`This is a very long string that needs to be wrapped around 72 characters
+for readability. Here is another shorter line. Whitespace over here.`, data.MessageOverride)
```
Kimchelly, line 103: Can we check a few boundary cases like:
- Much shorter than 72 characters
- 71 characters
- 72 characters

Kimchelly (changes requested):

rest/model/commit_queue_test.go
```diff
@@ -89,7 +89,16 @@ some more lines
 	assert.Len(data.Modules, 1)
 	assert.Equal(utility.ToStringPtr("module1"), data.Modules[0].Module)
 	assert.Equal(utility.ToStringPtr("1234"), data.Modules[0].Issue)
-	assert.Equal(`this is my commit message
-some more lines
-    extra whitespace  `, data.MessageOverride)
+	assert.Equal("this is my commit message some more lines extra whitespace", data.MessageOverride)
```
malikchaya2, line 92: Can we add a test with a long word at the end of the line to make sure that it doesn't break up words?

malikchaya2, line 92: Can we also test how it behaves with code formatted text?

rest/model/commit_queue.go
hadjri: Updated with a comment.

rest/model/commit_queue_test.go
```diff
@@ -91,5 +91,75 @@ some more lines
...
+	assert.Equal(`A message with a word that is
+longggggggggggggggggggggggggggggggggggggggggggggggggggggggggggg`, data.MessageOverride)
+
+	comment = `evergreen merge --unknown-option blah_blah --module module1:1234 
+Singlewordthatis72characterslonganditshouldntbebrokenupbythewrappingfunc`
+	data = ParseGitHubComment(comment)
+	assert.Len(data.Modules, 1)
+	assert.Equal(utility.ToStringPtr("module1"), data.Modules[0].Module)
+	assert.Equal(utility.ToStringPtr("1234"), data.Modules[0].Issue)
+	assert.Equal("Singlewordthatis72characterslonganditshouldntbebrokenupbythewrappingfunc", data.MessageOverride)
```
malikchaya2, line 164: Since it only breaks up after 72, can we please add some more characters to the check?

rest/model/commit_queue.go
```diff
@@ -88,13 +91,45 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
...
 		}
 	}
 
 	return data
 }
 
+// wrapString manually splits a string into smaller chunks and appends them to a new string.
+func wrapString(str string, limit int) string {
+	var b strings.Builder
+	currentLine := ""
```
Kimchelly, line 104: nit: we can move `currentLine` into the outer for loop since it's not used at the function level. Would also avoid the need to reset `currentLine` on L127.

2nd nit: the default value of a string type is already the empty string, so we could declare this as `var currentLine string`.

rest/model/commit_queue.go
```diff
@@ -88,13 +91,45 @@ func ParseGitHubComment(comment string) GithubCommentCqData {
...
+			}
+			if currentLine == "" {
+				currentLine += word
+			} else {
+				currentLine += " " + word
+			}
+		}
+		b.WriteString(currentLine)
+		if i < len(lines)-1 {
+			b.WriteString("\n")
```
Kimchelly, line 126: Is the thinking here that we want to preserve newlines as they originally appeared in the commit message, even if we're breaking up individual long lines?

Kimchelly (approved):
Gonna approve this since I don't have strong opinions on either of the follow-up comments I made.

rest/model/commit_queue.go
hadjri: Yes, the assumption is that if someone introduced a line break in their initial message that it was intentional.

rest/model/commit_queue_test.go
```diff
@@ -162,4 +162,12 @@ Singlewordthatis72characterslonganditshouldntbebrokenupbythewrappingfunc`
 	assert.Equal(utility.ToStringPtr("module1"), data.Modules[0].Module)
 	assert.Equal(utility.ToStringPtr("1234"), data.Modules[0].Issue)
 	assert.Equal("Singlewordthatis72characterslonganditshouldntbebrokenupbythewrappingfunc", data.MessageOverride)
```
malikchaya2, line 164: I'm a bit confused about why this passes if it's only 72 characters

hadjri: It's behaving as expected since we're not breaking up words. In the extreme case of a message having a word of 72+ characters that word will take up a whole line no matter how long it is. This isn't a realistic case anyway

malikchaya2 (approved):

PR #6216, opened 2023-01-27
EVG-17334 Add logging to test 

malikchaya2:
[EVG-17334](https://jira.mongodb.org/browse/EVG-17334)

### Description 
This test fails on the repotracker but passes on the patch. The logging should help with debugging.

### DIFF
--- a/agent/command/s3_put_test.go
+++ b/agent/command/s3_put_test.go
@@ -503,7 +503,7 @@ func TestPreservePath(t *testing.T) {
 		"remote/myWebsite/assets/images/image2": false,
 	}
 	grip.Infof("expected: %s", expected)
-	grip.Infof("recieved: ")
+	grip.Infof("received: ")
 	for it.Next(ctx) {
 		grip.Infof("\n: %s", it.Item().Name())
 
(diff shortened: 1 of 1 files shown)


hadjri (approved):

agent/command/s3_put_test.go
```diff
@@ -501,7 +502,11 @@ func TestPreservePath(t *testing.T) {
 		"remote/myWebsite/assets/images/image1": false,
 		"remote/myWebsite/assets/images/image2": false,
 	}
+	grip.Infof("expected: %s", expected)
+	grip.Infof("received: ")
 	for it.Next(ctx) {
+		grip.Infof("\n: %s", it.Item().Name())
```
Kimchelly, line 508: Is this meant to be committed or just for temporary testing? We don't normally log in tests because you can typically add assert/require statements that will do the equivalent, without the logging noise.

I guess, what kind of information do we get out of the logs that we couldn't get from the assert/require statements?

malikchaya2: This is just for debugging. It's passing in patches but not the repotracker

PR #6217, opened 2023-01-27
EVG-18828 Fix failing test 

malikchaya2:
[EVG-18828](https://jira.mongodb.org/browse/EVG-18828)

### Description 
This test was failing on windows because the file structure is different. 

### Testing 
Tested on windows in a patch

### DIFF
--- a/agent/command/s3_put_test.go
+++ b/agent/command/s3_put_test.go
@@ -507,15 +507,8 @@ func TestPreservePath(t *testing.T) {
 		expected[it.Item().Name()] = true
 	}
 
-	trueForExpected := true
-	trueForExpectedWindows := true
-
-	for _, exists := range expected {
-		if !exists {
-			trueForExpected = false
-		}
+	for item, exists := range expected {
+		require.True(t, exists, item)
 	}
-
-	require.True(t, trueForExpected || trueForExpectedWindows)
 
 }
--- a/agent/command/s3_put_test.go
+++ b/agent/command/s3_put_test.go
@@ -494,12 +494,17 @@ func TestPreservePath(t *testing.T) {
 	it, err := s.bucket.List(ctx, "")
 	require.NoError(t, err)
 
-	expected := getExpected(false)
-	expectedWindows := getExpected(true)
+	expected := map[string]bool{
+		filepath.Join("remote", "foo"):                                     false,
+		filepath.Join("remote", "myWebsite", "assets", "asset1"):           false,
+		filepath.Join("remote", "myWebsite", "assets", "asset2"):           false,
+		filepath.Join("remote", "myWebsite", "assets", "asset3"):           false,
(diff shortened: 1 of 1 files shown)


agent/command/s3_put_test.go
```diff
@@ -494,24 +493,51 @@ func TestPreservePath(t *testing.T) {
...
+	}
+
+	require.True(t, trueForExpected || trueForExpectedWindows)
+
+}
+
+func getExpected(isWindows bool) map[string]bool {
+	if isWindows {
+		return map[string]bool{
+			"remote\\foo":                               false,
```
Kimchelly, line 526: Rather than having two separate maps with basically the same files, this could just be `filepath.Join("remote", "foo"): false` (and so on) so we don't have to care about if it's Windows or not.

agent/command/s3_put_test.go
```diff
@@ -494,24 +493,51 @@ func TestPreservePath(t *testing.T) {
...
+
+	trueForExpected := true
+	trueForExpectedWindows := true
+
+	for _, exists := range expected {
+		if !exists {
+			trueForExpected = false
+		}
+	}
+	for _, exists := range expectedWindows {
```
Kimchelly, line 513: I think if we did the thing I mentioned in my comment, we don't have to worry about checking both Unix systems vs. Windows paths since the filepath separator will be decided by `filepath.Join`.

Kimchelly (changes requested):

agent/command/s3_put_test.go
```diff
@@ -494,24 +493,29 @@ func TestPreservePath(t *testing.T) {
...
-		grip.Infof("\n: %s", it.Item().Name())
 
+	for it.Next(ctx) {
 		expected[it.Item().Name()] = true
 	}
 
-	for item, exists := range expected {
-		require.True(t, exists, item)
+	trueForExpected := true
+	trueForExpectedWindows := true
```
Kimchelly, line 511: We should remove this variable since it will always be true.

agent/command/s3_put_test.go
```diff
@@ -494,24 +493,29 @@ func TestPreservePath(t *testing.T) {
...
+
+	for _, exists := range expected {
+		if !exists {
+			trueForExpected = false
```
Kimchelly, line 515: Rather than define a variable to check after the loop, can we just `assert.True(t, exists, <the file name that should have existed>)`? Would also have the benefit of telling exactly which files were expected but nonexistent.

Kimchelly (changes requested):

agent/command/s3_put_test.go
malikchaya2: done

agent/command/s3_put_test.go
malikchaya2: done

Kimchelly (approved):

PR #6218, opened 2023-01-27
EVG-16434: Temporarily allow either projectID or projectIdentifier for some queries

minnakt:
[EVG-16434](https://jira.mongodb.org/browse/EVG-16434)

### Description 
This PR isn't meant to change any core functionality, it's just a PR to rename some variables. There are a few GraphQL queries that are supposed to be called with a `projectIdentifier` but the queries have the param name `projectID`. 

This PR temporarily allows either `projectID` or `projectIdentifier` to be passed to these queries. Once Spruce is updated to only use the `projectIdentifier` param, I'll remove the `projectID` param and make the `projectIdentifier` param required for these queries.

#### Note
I don't think it really matters if we send `projectID` or `projectIdentifier` to these queries. This is because the underlying `byId` function is defined in a way where we can pass either the ID or identifier and the result will be correct:

https://github.com/evergreen-ci/evergreen/blob/d1f28a88613c3789e22ad4b1ddd5f2c3935b284c/model/project_ref.go#L1380-L1387

However since we are always passing the identifier to these queries from Spruce it would make sense for these params to be named correctly.

### Testing 
- On staging
- Updated the particular queries to use `projectIdentifier` rather than `projectID` in GraphQL tests
- Spruce e2e test should pass i.e. using these queries should still work under current Spruce


khelif96 (approved):
Given this is temporary and will be updated soon it LGTM

malikchaya2 (approved):

PR #6219, opened 2023-01-27
EVG-17334 Fix uploadFiles appending remote twice 

malikchaya2:
[EVG-17334](https://jira.mongodb.org/browse/EVG-17334)

### Description 
Fix bug accidentally introduce in https://github.com/evergreen-ci/evergreen/pull/6211.

### DIFF
--- a/agent/command/s3_put.go
+++ b/agent/command/s3_put.go
@@ -444,7 +444,11 @@ retryLoop:
 					continue retryLoop
 				}
 
-				uploadedFiles = append(uploadedFiles, remoteName)
+				if s3pc.preservePath {
+					uploadedFiles = append(uploadedFiles, remoteName)
+				} else {
+					uploadedFiles = append(uploadedFiles, fpath)
+				}
 			}
 
 			break retryLoop


agent/command/s3_put.go
```diff
@@ -444,7 +444,11 @@ retryLoop:
 					continue retryLoop
 				}
 
-				uploadedFiles = append(uploadedFiles, remoteName)
+				if s3pc.preservePath {
+					uploadedFiles = append(uploadedFiles, remoteName)
+				} else {
+					uploadedFiles = append(uploadedFiles, fpath)
+				}
```
Kimchelly, line 451: Rather than add another if-else here, could we just change [the if-else in attachFiles](https://github.com/evergreen-ci/evergreen/blob/11ac2056b7b08001e5c05769d947f9b544363eb1/agent/command/s3_put.go#L487-L491) to just be `remoteFileName = fn`? It reduces the number of if-else's from 3 to 1.

Kimchelly (changes requested):

Kimchelly:
Can we test that this is working as expected after this change?

agent/command/s3_put_test.go
```diff
@@ -430,6 +430,64 @@ func TestS3LocalFilesIncludeFilterPrefix(t *testing.T) {
 	}
 }
 
+func TestFileUploadNaming(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	var err error
```
Kimchelly, line 436: Remove err since it's already declared and defined on L440.

agent/command/s3_put_test.go
```diff
@@ -430,6 +430,64 @@ func TestS3LocalFilesIncludeFilterPrefix(t *testing.T) {
...
+
+	require.NoError(t, s.Execute(ctx, comm, logger, conf))
+	attachedFiles := comm.AttachedFiles
+	expected := map[string]bool{
+		"remotebar":       false,
+		"remoteremotebar": false,
+	}
+
+	for _, file := range attachedFiles {
+		for _, file := range file {
```
Kimchelly, line 480: Can we call the variable something other than `file` since the outer loop already defines `file`?

Kimchelly (approved):
LGTM mod some nits.

PR #6221, opened 2023-01-31
EVG-17650 enable "setup failed" statuses for tasks that fail their pre-commands

hadjri:
[EVG-17650](https://jira.mongodb.org/browse/EVG-17650)

### Description 
There is currently a bug where if a task fails due to a `pre` command failure, the task itself will get marked as `setup.initial` being the command that failed, which is incorrect. This is because pre & post commands do not ever get set as the current command.

Added a change so that if a task is configured to fail if its pre/post block fails, set each pre & post task as the current command when it runs so that its failure details get correctly propagated up.

### Testing 
Added unit test confirming a failed `pre` command with `pre_error_fails_task` set results in a failure based off the specific `pre` command.

In staging, recreated a simple version of the problem in the ticket. System failure [without change](https://spruce-staging.corp.mongodb.com/version/63d937269dbe321c5d9c69e8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), setup failure [with the change](https://spruce-staging.corp.mongodb.com/task/sandbox_git_env_test_validate_commit_message_patch_37dbc82858fa1d557a1b1989db8fb6ce1f5e0d6d_63d998379dbe3242f84bac69_23_01_31_22_37_43/logs?execution=0).

### DIFF
--- a/agent/agent_command_test.go
+++ b/agent/agent_command_test.go
@@ -57,6 +57,28 @@ func (s *CommandSuite) SetupTest() {
 	}
 }
 
+func (s *CommandSuite) TestPreErrorFailsWithSetup() {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	taskID := "pre_error"
+	s.tc.task.ID = taskID
+	s.tc.ranSetupGroup = false
+
+	defer s.a.removeTaskDirectory(s.tc)
+	_, err := s.a.runTask(ctx, s.tc)
+	s.NoError(err)
+	detail := s.mockCommunicator.GetEndTaskDetail()
+	s.Equal(evergreen.TaskFailed, detail.Status)
+	s.Equal(evergreen.CommandTypeSetup, detail.Type)
+	s.Contains(detail.Description, "shell.exec")
+	s.False(detail.TimedOut)
+
+	taskData := s.mockCommunicator.EndTaskResult.TaskData
+	s.Equal(taskID, taskData.ID)
+	s.Equal(s.tc.task.Secret, taskData.Secret)
+}
+
 func (s *CommandSuite) TestShellExec() {
 	f, err := ioutil.TempFile(s.tmpDirName, "shell-exec-")
 	s.Require().NoError(err)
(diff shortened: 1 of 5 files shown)


agent/internal/client/testdata/pre_error.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - func: foo
+
+functions:
+  "foo":
+    - command: shell.exec
+      type: setup
+      params:
+        shell: bash
+        script: |
+          exit 1
```
Kimchelly, line 14: If it's a one-line shell script, you can just put it on the same line rather than using `|` to stick it on the next line.

agent/internal/client/testdata/pre_error.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+
+tasks:
+  - name: build
+    commands:
+      - command: shell.exec
+        params:
+          working_dir: src
+          script: |
```
Kimchelly, line 23: Same nit as above.

agent/agent_command_test.go
```diff
@@ -57,6 +57,28 @@ func (s *CommandSuite) SetupTest() {
...
+	taskID := "pre_error"
+	s.tc.task.ID = taskID
+	s.tc.ranSetupGroup = false
+
+	defer s.a.removeTaskDirectory(s.tc)
+	_, err := s.a.runTask(ctx, s.tc)
+	s.NoError(err)
+	detail := s.mockCommunicator.GetEndTaskDetail()
+	s.Equal(evergreen.TaskFailed, detail.Status)
+	s.Equal(evergreen.CommandTypeSetup, detail.Type)
```
Kimchelly, line 73: If `pre_error_fails_task` is true, should this respect the command/top-level `command_type`? I would have thought based on the YAML `command_type: system` that it would result in a system failure instead of setup failure.

Kimchelly (changes requested):

agent/agent_command_test.go
hadjri: The top-level command gets [overridden](https://github.com/evergreen-ci/evergreen/blob/main/model/project.go#L1180) by the current command's type, so since the failing pre command specifies `setup` as its type it will be the type that is respected.

Kimchelly (approved):

PR #6222, opened 2023-01-31
EVG-18847 Add convenience methods to get client settings for scripting

bsamek:


ablack12 (approved):
Changes look good to me -- do you intend to use this for something / would that work be temporary such that it'd be worth removing it once that work is done?

bsamek:
@ablack12 No, this came out of conversations with Max, where he has do a bit of awkward parsing of the config file in scripts he has.

PR #6223, opened 2023-01-31
EVG-18856 Update SaveProjectPageForSection to not use cached config

bynn:
[EVG-18856](https://jira.mongodb.org/browse/EVG-18856)

### Description 
update to not use cached config
also update to not error if config isnt set 

### Testing 
staging

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1897,7 +1897,10 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			setUpdate[ProjectRefTracksPushEventsKey] = p.TracksPushEvents
 		}
 		if !isRepo && !p.UseRepoSettings() && !defaultToRepo {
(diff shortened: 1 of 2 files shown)


model/project_ref.go
```diff
@@ -2245,7 +2248,7 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 // The boolean returns true if you should error. Otherwise, warn.
 func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (error, bool) {
```
ablack12, line 2249: Can we return (bool, error)

ablack12 (approved):
LGTM besides change

model/project_ref.go
ablack12: discussed in person; we'll change this to http status code in a future PR

PR #6224, opened 2023-01-31
EVG-17579: fix generate.tasks validation for generate.tasks calls

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-17579

### Description 
`generate.tasks` failed for a slightly subtle reason. The task calling `generate.tasks` in the ticket was adding tasks that were already defined YAML to new build variants. When `generate.tasks` does this, Evergreen does not allow recursive task generation, so it must check that it's not adding tasks that will end up calling `generate.tasks` again.

The bug happened because this for loop was trying to add all the existing tasks to a map for fast lookup during validation. However, it did so by assigning a pointer to a local variable defined in the for loop line, which is a common Go gotcha because the local variable will be mutated on each iteration, resulting in `&t` pointing to the last item in `p.Tasks` (and therefore the map will be filled completely with the last item in `p.Tasks`). Add to that the fact that `p.Tasks` is in random order (the parser project randomizes the order of the task list), it means that the last task is random. This time, the last task in `p.Tasks` happened to be the one running `generate.tasks`. Thus, the validation logic wrongly concluded that all the tasks were running `generate.tasks` because that was the task pointed to by `&t`.

### Testing 
Added unit tests for the bugged function and another one in generate.tasks that has historicallyh had this same issue.

### DIFF
--- a/model/generate.go
+++ b/model/generate.go
@@ -92,6 +92,10 @@ func MergeGeneratedProjects(projects []GeneratedProject) (*GeneratedProject, err
 	for i := range taskGroups {
 		g.TaskGroups = append(g.TaskGroups, *taskGroups[i])
 	}
+	grip.Info(message.Fields{
+		"message":           "kim: successfully merged generated fields",
+		"generated_project": g,
+	})
 	return g, catcher.Resolve()
 }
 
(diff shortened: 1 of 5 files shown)


model/generate.go
```diff
@@ -199,8 +199,8 @@ func cacheProjectData(p *Project) projectMaps {
 	for _, bv := range p.BuildVariants {
 		cachedProject.buildVariants[bv.Name] = struct{}{}
 	}
-	for _, t := range p.Tasks {
-		cachedProject.tasks[t.Name] = &t
+	for i, t := range p.Tasks {
+		cachedProject.tasks[t.Name] = &p.Tasks[i]
```
ablack12, line 203: should we add a comment to this or do we feel comfortable enough given the test

ablack12 (approved):

model/generate.go
Kimchelly: We assign local variables in a lot of places, so I don't know if it's worth putting a comment specifically here when it happens all over the place. Maybe we can advertise a list of common Go gotchas during a team presentation to explain why this is correct and the previous thing isn't?

ablack12: i think a team presentation isn't really necessary since the issue with gotchas is that they're easy to forget unless you run into them (although could still be a fun one so i'm not against having one haha). i'm fine not leaving a comment though since it is tested

Kimchelly: Maybe we can compile a list of common bugs/practices to avoid in Evergreen with a Google doc, and if we see other gotchas come up in other code reviews, we could add to the list?

PR #6226, opened 2023-02-02
EVG-18530 Suggest a merge conflict is likely if an enqueued CLI patch fails on setup

hadjri:
[EVG-18530](https://jira.mongodb.org/browse/EVG-18530)

### Description 
Unlike Github PRs where we [explicitly check](https://gist.github.com/hadjri/fbb86d69df6cf83c05e03106282d5772#file-github-go-L942) if GitHub has created a valid merge commit without conflicts, there is currently nothing stopping us from staging a CLI patch with a merge conflict in it, and enqueuing it to the commit queue. If this happens, the merge task will fail on the first `git.get_project` command of the task, resulting in a [setup failure](https://spruce-staging.corp.mongodb.com/task/sandbox_commit_queue_merge_merge_patch_patch_c23d49c536277965b729fc5c86c75d86a339ab26_63daecc99dbe32c5d5f54b86_23_02_01_22_51_08/logs?execution=0) with errors in the logs saying `patch does not apply`.

The logic behind this change is, rather than performing a full `git apply` on the target branch based off of all of the file patches in a given patch (which would be slightly complex and increase the already high traffic we're sending to GitHub), we can make the assumption that if a merge task failed on its setup command, there is a high likelihood that it happened due to a merge conflict, so an error mentioning this is included when this occurs so that it gets propagated to the CQ email that gets sent out.
### Testing 
Tested in staging and confirmed that enqueuing a CLI patch with a merge conflict in it causes this error to pop up in the commit queue merge email.

### DIFF
--- a/model/event/commit_queue_event.go
+++ b/model/event/commit_queue_event.go
@@ -69,6 +69,16 @@ func LogCommitQueueConcludeTest(patchID, status string) {
 	logCommitQueueEvent(patchID, CommitQueueConcludeTest, data)
 }
 
+func LogCommitQueueConcludeWithError(patchID, status string, err error) {
+	data := &CommitQueueEventData{
+		Status: status,
+	}
+	if err != nil {
+		data.Error = err.Error()
+	}
+	logCommitQueueEvent(patchID, CommitQueueConcludeTest, data)
+}
+
 func LogCommitQueueEnqueueFailed(patchID string, err error) {
 	data := &CommitQueueEventData{
 		Status: evergreen.EnqueueFailed,
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -930,7 +930,7 @@ func tryDequeueAndAbortCommitQueueVersion(p *patch.Patch, cq commitqueue.CommitQ
 	// If the commit queue merge task failed on setup, there is likely a merge conflict.
 	var mergeError error
 	if t.Details.Type == evergreen.CommandTypeSetup {
-		mergeError = errors.New("merge task failed on setup, which likely means a merge conflict was introduced")
+		mergeError = errors.New("merge task failed on setup, which likely means a merge conflict was introduced, please try merging with the base branch")
 	}
 	event.LogCommitQueueConcludeWithError(p.Id.Hex(), evergreen.MergeTestFailed, mergeError)
 	return errors.Wrap(CancelPatch(p, task.AbortInfo{TaskID: t.Id, User: caller}), "aborting failed commit queue patch")
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -814,7 +814,7 @@ func DequeueAndRestartForTask(cq *commitqueue.CommitQueue, t *task.Task, githubS
 	if p == nil {
 		return errors.Errorf("patch '%s' not found", t.Version)
 	}
-	if err := tryDequeueAndAbortCommitQueueVersion(p, *cq, t.Id, caller); err != nil {
+	if err := tryDequeueAndAbortCommitQueueVersion(p, *cq, t, caller); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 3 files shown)


model/event/commit_queue_event.go
```diff
@@ -69,6 +69,16 @@ func LogCommitQueueConcludeTest(patchID, status string) {
 	logCommitQueueEvent(patchID, CommitQueueConcludeTest, data)
 }
 
+func LogCommitQueueConcludeWithError(patchID, status string, err error) {
```
malikchaya2, line 72: Will this error show up when we try to enqueue it, or only after the patch has reached the top of the queue and tried to run tasks?

malikchaya2 (changes requested):

model/event/commit_queue_event.go
hadjri: The latter. If we want to actually prevent the patch from getting enqueued at all, we'd need to do a full server-side `git apply --check` on the whole repo of the target branch, which is a bit of a heavy lift.

hadjri: Couple other things we could try:

Incorporate this into the CLI side `patch` command, where under the hood we check out the main branch, apply the patches, and revert back the original branch with an error message if the apply failed or not, and then store that as a boolean value on the patch as something like `has_merge_conflict_with_target`, and then prevent enqueuing  based on that. 

But I'm not sure if that would be worth the effort.

ablack12 (commented):
I thought the purpose of this ticket was -- we already send an email in this case, and we just wanted to shore up the wording of it. Is that not the case?

model/event/commit_queue_event.go
malikchaya2: That's probably not worth the effort. I added a comment to the ticket saying that we should only add this help text if it failed to enqueue, but on second thought, I don't think we send out emails unless it did enqueue and then failed - so I think this is good. 

Can you please include a screenshot in the description of the error being added to the email?

model/task_lifecycle.go
```diff
@@ -927,9 +927,13 @@ func tryDequeueAndAbortCommitQueueVersion(p *patch.Patch, cq commitqueue.CommitQ
 			"patch":   p.Id.Hex(),
 		}))
 	}
-
-	event.LogCommitQueueConcludeTest(p.Id.Hex(), evergreen.MergeTestFailed)
-	return errors.Wrap(CancelPatch(p, task.AbortInfo{TaskID: taskId, User: caller}), "aborting failed commit queue patch")
+	// If the commit queue merge task failed on setup, there is likely a merge conflict.
+	var mergeError error
+	if t.Details.Type == evergreen.CommandTypeSetup {
+		mergeError = errors.New("merge task failed on setup, which likely means a merge conflict was introduced")
```
malikchaya2, line 933: Can we also add that it can be fixed by merging main?

hadjri:
> I thought the purpose of this ticket was -- we already send an email in this case, and we just wanted to shore up the wording of it. Is that not the case?

The email we send only gets picked up if the [config file is part of the merge conflict](https://github.com/evergreen-ci/evergreen/blob/main/model/patch_lifecycle.go#L376), and will go undetected if there is a merge conflict that doesn't touch the config file. If this is the case, the email sent out currently just says "Your patch has failed" with no followup message.

model/event/commit_queue_event.go
hadjri: image

hadjri: image

malikchaya2: nit: since this ends up as a full sentence in the email, can we capitalize and add a period?

malikchaya2 (approved):
approved with nit

bynn (approved):

model/task_lifecycle.go
```diff
@@ -927,9 +927,13 @@ func tryDequeueAndAbortCommitQueueVersion(p *patch.Patch, cq commitqueue.CommitQ
 			"patch":   p.Id.Hex(),
 		}))
 	}
-
-	event.LogCommitQueueConcludeTest(p.Id.Hex(), evergreen.MergeTestFailed)
-	return errors.Wrap(CancelPatch(p, task.AbortInfo{TaskID: taskId, User: caller}), "aborting failed commit queue patch")
+	// If the commit queue merge task failed on setup, there is likely a merge conflict.
+	var mergeError error
+	if t.Details.Type == evergreen.CommandTypeSetup {
+		mergeError = errors.New("Merge task failed on setup, which likely means a merge conflict was introduced. Please try merging with the base branch")
```
ablack12, line 933: nit: can we end this with a period?

model/event/commit_queue_event.go
```diff
@@ -69,6 +69,16 @@ func LogCommitQueueConcludeTest(patchID, status string) {
...
+	data := &CommitQueueEventData{
+		Status: status,
+	}
+	if err != nil {
+		data.Error = err.Error()
```
ablack12, line 77: Since we just convert this to a string anyway, can we just pass in a string? (can maybe call the function WithMessage instead)

PR #6228, opened 2023-02-02
EVG-18592 Add further splunk logging to track commit queues

hadjri:
[EVG-18592](https://jira.mongodb.org/browse/EVG-18592)

### Description 

- Added logs to monitor when we remove items from the commit queue, and logs that track the speed of processing single commit queue items as well as logs that track the speed of processing the full batch of commit queue items
- Some logs were inaccurately logging the length of the commit queue when enqueuing / dequeuing since it would log before the queue was actually updated. Moved the logs to occur after the state change.
- Changed Splunk logs that previously showed only the commit item's issue number to show the full item so we can inspect its version, source, etc more easily
- Removed some unused parameters from CQ db functions


units/commit_queue.go
```diff
@@ -187,12 +197,20 @@ func (j *commitQueueJob) Run(ctx context.Context) {
...
+			"item":                 nextItem,
+			"message":              "finished processing commit queue item",
+			"processing_time_secs": time.Since(beginProcessingTime).Seconds(),
 		})
 	}
+	grip.Info(message.Fields{
+		"source":               "commit queue",
+		"job_id":               j.ID(),
+		"items":                nextItems,
+		"message":              "finished processing batch of commit queue items",
```
malikchaya2, line 211: If we wanted to go somewhere and answer the questions: "Would we benefit from increasing the batch size from 3 to 4 based on the current speed of our commit queue?" , where would someone go to get an answer for that? And can we document that somewhere?

malikchaya2 (changes requested):

units/commit_queue.go
ablack12: We'd theoretically always benefit right? It just may not be worth the added cost? (I might be misunderstanding what's wanted from this comment though)

malikchaya2: It's the "based on the current speed of our commit queue" that's key here. And yes, it would be benefit in a way that is worth the added cost. 

The main ask of the comment is: when I created this ticket my goal was to both generate that information, but then also to think of a way to make that information easily accessible without too much additional code reading or log grepping. So the question is: has that been thought about and can we document the answer?

hadjri: The formula for the worst case number of patches that can be ran for a commit queue of size `n` (if every item fails) can be quantified by this [pretty wild formula](https://docs.google.com/document/d/1n7xog26jPzM5j87ce5AnOeGwtWaGPKIieNQ2CirQwcI/edit#) from the initial design of when batches in the CQ were initially developed. So (output of the formula - `n`) is the number of patches we risk wasting by increasing the batch size. So if queue size is 10 for example, and batch size is 3, the worst case scenario is that we run 19 patches which is 9 more patches than we'd have needed to run if we didn't do any parallel execution.

Of course items don't all fail, in fact most succeed, but to track how many failures occur for each project I added some logs to `DequeueAndRestartForTask` so we can collect statistics moving forward on how many CQ items are getting dequeued due to failures over time. As this collects stats over time we can probably get a decent idea of the rate of failures to expect in each commit queue, and plug it into the above formula to get a more quantitative idea on how much work we are re-computing as a consequence of a certain batch size.

Also, I created a [Commit Queue Wait Statistics](https://mongodb.splunkcloud.com/en-US/app/search/commit_queue_wait_time_2?form.global_time.earliest=-30d%40d&form.global_time.latest=now) dashboard which currently shows project's commit queue size over time, and the # of minutes queue items wait to get processed to get better idea for the queue's performance over time. I also added a new field to the items (`QueueLengthAtEnqueue`) which saves the length of the queue at the time of enqueueing, so we can later add charts to get an accurate rate of progress of project's commit queues by comparing this value to the time it took for the items to get processed.

ablack12: Love this dashboard! I would suggest sending to product (Maria/Cris) ; they may have something similar or may wish to hold onto it.

units/commit_queue.go
```diff
@@ -187,12 +198,20 @@ func (j *commitQueueJob) Run(ctx context.Context) {
...
+			"job_id":               j.ID(),
+			"item":                 nextItem,
+			"message":              "finished processing commit queue item",
+			"processing_time_secs": time.Since(beginProcessingTime).Seconds(),
 		})
 	}
+	grip.Info(message.Fields{
+		"source":               "commit queue",
+		"job_id":               j.ID(),
+		"items":                nextItems,
```
ablack12, line 211: Is items needed to log? I'd rather we didn't since we log this so frequently. Is the use of this the length of items?

units/commit_queue.go
```diff
@@ -187,12 +198,20 @@ func (j *commitQueueJob) Run(ctx context.Context) {
...
 		grip.Info(message.Fields{
-			"source":  "commit queue",
-			"job_id":  j.ID(),
-			"item":    nextItem,
-			"message": "finished processing item",
+			"source":               "commit queue",
+			"job_id":               j.ID(),
+			"item":                 nextItem,
+			"message":              "finished processing commit queue item",
+			"processing_time_secs": time.Since(beginProcessingTime).Seconds(),
```
ablack12, line 205: I don't really expect processing time to be a bottleneck; i'm not sure this log has value

model/commitqueue/commit_queue.go
```diff
@@ -29,12 +29,13 @@ type CommitQueueItem struct {
...
-	Modules             []Module  `bson:"modules"`
-	MessageOverride     string    `bson:"message_override"`
-	Source              string    `bson:"source"`
+	Version              string    `bson:"version,omitempty"`
+	EnqueueTime          time.Time `bson:"enqueue_time"`
+	ProcessingStartTime  time.Time `bson:"processing_start_time"`
+	Modules              []Module  `bson:"modules"`
+	MessageOverride      string    `bson:"message_override"`
+	Source               string    `bson:"source"`
+	QueueLengthAtEnqueue int       `bson:"queue_length_at_enqueue"`
```
ablack12, line 38: Can you add a comment to this field to explain its use, since it's just used for logging?

units/commit_queue.go
hadjri: It's not explicitly needed to log, I just thought it'd be helpful to give more visibility into what's being processed. I'm fine to change this to showing the length, but I think we should log the full item at least once.

units/commit_queue.go
hadjri: Sent!

malikchaya2: Thank you for typing that up. This information and dashboard are awesome! I'm personally a big fan of transferring this information and a link to the dashboard to one of our docs, probably the ops guide so that engineers can easily find it if they hit a commit queue conundrum. 

> I also added a new field to the items (QueueLengthAtEnqueue) which saves the length of the queue at the time of enqueueing, so we can later add charts to get an accurate rate of progress of project's commit queues by comparing this value to the time it took for the items to get processed.

I think it would be a good idea to do that now while the information is fresh, maybe in a new ticket.

malikchaya2 (changes requested):
This looks great! I had one more comment.

units/commit_queue.go
hadjri: Sounds good, added it to the [commit queue troubleshooting portion](https://docs.google.com/document/d/14BTuPnzbSLCuewcMXFNQivkyUPy3Dsy1TYdF_9WVaBY/edit#heading=h.lrutrlkph14l) of the ops guide.

I think we'll need at least a few days of `QueueLengthAtEnqueue` logs being available before we make dashboards based off of it, so creating a new ticket: https://jira.mongodb.org/browse/EVG-18895

malikchaya2 (approved):

PR #6230, opened 2023-02-02
EVG-16206 Use raw terminal for prompts

bynn:
[EVG-16206](https://jira.mongodb.org/browse/EVG-16206)

### Description 
use raw terminal so we can use arrow keys in prompts 
(this may change anything on a windows machine so if @Kimchelly could try it out I would appreciate it!) 

### Testing 
tried it out on my cli


bynn:
too many complications

PR #6231, opened 2023-02-02
EVG-18856 Fix test and add testing config to db

bynn:
[EVG-18856](https://jira.mongodb.org/browse/EVG-18856)

### Description 
the test was failing because we no longer read cached config so I added what we need to the db

### Testing 
tests pass


ablack12 (approved):
nit: please rename PR to fit conventions ("fix test and add" probably)

PR #6232, opened 2023-02-02
EVG-18537: implement parser project S3 storage

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18537

### Description 
* Implement parser project storage in S3. It's not used anywhere for now, there will be a follow-up to actually use it.
* Add some more docs about the `model.ParserProject` and how it's different from the `model.Project`.

### Testing
* Add integration tests for storing/retrieving parser projects from S3.
* Tested in staging by making a patch that kept its parser project in S3, then verified that generate.tasks worked with the S3 parser project.

### DIFF
--- a/model/project_parser_s3.go
+++ b/model/project_parser_s3.go
@@ -26,16 +26,16 @@ type ParserProjectS3Storage struct {
 // NewParserProjectS3Storage sets up access to parser projects stored in S3.
 // If this returns a non-nil ParserProjectS3Storage, callers are expected to
 // call Close when they are finished with it.
-func NewParserProjectS3Storage(awsConf evergreen.AWSConfig) (*ParserProjectS3Storage, error) {
+func NewParserProjectS3Storage(ppConf evergreen.ParserProjectS3Config) (*ParserProjectS3Storage, error) {
 	c := utility.GetHTTPClient()
 
 	var creds *credentials.Credentials
-	if awsConf.ParserProject.Key != "" && awsConf.ParserProject.Secret != "" {
-		creds = pail.CreateAWSCredentials(awsConf.ParserProject.Key, awsConf.ParserProject.Secret, "")
+	if ppConf.Key != "" && ppConf.Secret != "" {
+		creds = pail.CreateAWSCredentials(ppConf.Key, ppConf.Secret, "")
 	}
 	b, err := pail.NewS3MultiPartBucketWithHTTPClient(c, pail.S3Options{
-		Name:        awsConf.ParserProject.Bucket,
-		Prefix:      awsConf.ParserProject.Prefix,
+		Name:        ppConf.Bucket,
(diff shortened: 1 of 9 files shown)


model/project_parser_s3.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "reading parser project '%s'", id)
+	}
+
+	var pp ParserProject
+	if err := bson.Unmarshal(b, &pp); err != nil {
+		return nil, errors.Wrapf(err, "unmarshalling parser project '%s' from BSON", id)
+	}
+
+	// TODO: this is in the equivalent DB method, but is it necessary?
```
Kimchelly, line 78: Not sure what this is for, but findOne for the parser project from the DB does this, so I just copied the logic.

model/project_parser_s3.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return &pp, nil
+}
+
+// FindOneByIDWithFields finds a parser project using its ID from S3 and returns
+// the parser project with only the requested fields populated. This is not any
+// more efficient than FindOneByID. If the context errors, it will return the
+// context error.
+func (s *ParserProjectS3Storage) FindOneByIDWithFields(ctx context.Context, id string, fields ...string) (*ParserProject, error) {
```
bynn, line 90: Are we supposed to be doing something with fields?

model/project_parser_s3.go
```diff
@@ -0,0 +1,124 @@
...
+
+// FindOneByID finds a parser project in S3 using its ID. If the context errors,
+// it will return the context error.
+func (s *ParserProjectS3Storage) FindOneByID(ctx context.Context, id string) (*ParserProject, error) {
+	if s.closed {
+		return nil, errors.New("cannot access parser project S3 storage when it is closed")
+	}
+
+	r, err := s.bucket.Get(ctx, id)
+	if pail.IsKeyNotFoundError(err) {
```
bynn, line 60: Why don't we want to error if we don't find anything?

bynn (changes requested):

model/project_parser_s3.go
Kimchelly: It's intentionally ignored. This method only exists to make it compatible with [the DB equivalent](https://github.com/evergreen-ci/evergreen/blob/23e6c4bb142bc3273995025021c4b285bd084e60/model/project_parser_db.go#L134-L136).

model/project_parser_s3.go
Kimchelly: It's to follow the behavior of [the existing implementation for the DB](https://github.com/evergreen-ci/evergreen/blob/23e6c4bb142bc3273995025021c4b285bd084e60/model/project_parser_db.go#L71-L73).

PR #6236, opened 2023-02-03
EVG-17697 Retrieve host's tasks by id and execution

hadjri:
[EVG-17697](https://jira.mongodb.org/browse/EVG-17697)

### Description 
Since there is currently an implicit assumption that a host is always running the most recent execution of a task, it opens up the potential case where a host gets terminated after its running task got aborted / restarted. 

I don't believe the last paragraph in the original ticket is relevant any longer since the merge of this [PR](https://github.com/evergreen-ci/evergreen/pull/5888) which was implemented after this ticket makes it impossible for a task to be in an unfinished state if it is not on the latest execution. Hence, there isn't a risk of having any old tasks in an unfinished state. There is also some [documentation](https://gist.github.com/hadjri/c33c22dbb5d01e2afa57c76d6fb4dc7a#file-host_agent-go-L1256) in the codebase mentioning that having the order of operations be 1. Mark the task finished and then 2. Clear the running task would be costly.

I also removed some unused functions I noticed along the way.
### Testing 
Ran versions and tasks with restarts in staging. Updated unit tests.


model/host/db.go
```diff
@@ -898,6 +842,20 @@ func FindOneId(id string) (*Host, error) {
 	return FindOne(ById(id))
 }
 
+// FindOneTaskIdAndExecution returns a single host with the given running task ID and execution.
+func FindOneTaskIdAndExecution(id string, execution int) (*Host, error) {
```
malikchaya2, line 846: should this maybe be FindOneByTaskIdAndExecution?

model/host/host.go
```diff
@@ -62,7 +62,7 @@ type Host struct {
 
 	// the task that is currently running on the host
 	RunningTask             string `bson:"running_task,omitempty" json:"running_task,omitempty"`
-	RunningTaskExecution    int    `bson:"running_task_execution,omitempty" json:"running_task_execution,omitempty"`
+	RunningTaskExecution    int    `bson:"running_task_execution" json:"running_task_execution"`
```
malikchaya2, line 65: Did you find instances where the field didn't exist? If there are cases of that, we will still have to case against that since someone may be trying to restart and old task.

model/task_lifecycle_test.go
```diff
@@ -4055,48 +4063,52 @@ func TestClearAndResetStrandedHostTask(t *testing.T) {
...
-	assert.Equal(dt.Details, task.GetSystemFailureDetails(evergreen.TaskDescriptionStranded))
-
-	foundBuild, err = build.FindOneId("b2")
-	require.NoError(t, err)
-	assert.Equal(evergreen.BuildFailed, foundBuild.Status)
-
-	foundVersion, err = VersionFindOneId(b2.Version)
-	require.NoError(t, err)
-	assert.Equal(evergreen.VersionFailed, foundVersion.Status)
+	//assert.NoError(ClearAndResetStrandedHostTask(settings, h))
```
malikchaya2, line 4066: Did you mean to remove this?

rest/route/host.go
```diff
@@ -530,7 +530,7 @@ func (rh *hostProvisioningOptionsGetHandler) Run(ctx context.Context) gimlet.Res
 	return gimlet.NewJSONResponse(apiOpts)
 }
 
-////////////////////////////////////////////////////////////////////////
+// //////////////////////////////////////////////////////////////////////
```
malikchaya2, line 533: super nit: we might want to remove this space

malikchaya2 (changes requested):

model/host/host.go
```diff
@@ -1420,29 +1423,6 @@ func (h *Host) clearRunningTaskWithFunc(doUpdate func(update bson.M) error) erro
 	return nil
 }
 
-// UpdateRunningTask updates the running task in the host document and logs an
-// event indicating that it's been assigned a task.
-func (h *Host) UpdateRunningTask(t *task.Task) error {
```
Kimchelly, line 1425: Since this is now deleted, we can put the logic in `updateRunningTaskWithFunc` inline into `UpdateRunningTaskWithContext`.

Kimchelly (changes requested):
Does [this check](https://github.com/evergreen-ci/evergreen/blob/80a85354240cc7bebbb9724aada6f773e1be365b/units/task_monitor_execution_timeout.go#L178) have to handle the possibility of the task execution not matching the host running task execution?

hadjri:
> Does [this check](https://github.com/evergreen-ci/evergreen/blob/80a85354240cc7bebbb9724aada6f773e1be365b/units/task_monitor_execution_timeout.go#L178) have to handle the possibility of the task execution not matching the host running task execution?

The only way the execution number could be wrong is if it's lower than the current execution number, in which case that task would be in a finished state. Since right before the invocation of this function we [check if the task is finished](https://github.com/evergreen-ci/evergreen/blob/main/units/task_monitor_execution_timeout.go#L125), I think we're safe that if this line hits then it is the most recent task execution.

model/host/host.go
```diff
@@ -1471,21 +1471,17 @@ func (h *Host) clearRunningTaskWithFunc(doUpdate func(update bson.M) error) erro
...
-
-func (h *Host) updateRunningTaskWithFunc(doUpdate func(query, update bson.M) error, t *task.Task) error {
 	if t == nil {
 		return errors.New("received nil task, cannot update")
 	}
 	if t.Id == "" {
 		return errors.New("task has empty task ID, cannot update")
 	}
 
+	doUpdate := func(query, update bson.M) error {
```
Kimchelly, line 1481: We don't need this update functor anymore since we can just call `env.DB()...UpdateOne(...)` directly on L1509.

Kimchelly (approved):
I thought through it a little more and this sounds reasonable with the way that `(Task).HostId` currently behaves when a task resets. Thanks! LGTM mod one minor comment.

malikchaya2 (approved):

PR #6243, opened 2023-02-07
EVG-18863 Add validation to project modify routes

bynn:
[EVG-18863](https://jira.mongodb.org/browse/EVG-18863)

### Description 
patch project route has validation now

The ui server does not ask for repo and owner to create/copy project so, so I decided not to add validation there. (even if we did, it would only warn and not error because it would never be enabled. )
I added validation to the modify project ui service so users will not be able to enable projects if it does not pass validation

### Testing 
staging 
image

image

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -312,6 +312,9 @@ var (
 )
 
 func (p *ProjectRef) IsEnabled() bool {
+	if p.Enabled == nil {
+
+	}
 	return utility.FromBoolPtr(p.Enabled)
 }
 
@@ -1910,7 +1913,7 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 			setUpdate[ProjectRefRepoKey] = p.Repo
 
 			// Cannot enable projects if the project creation limits have been reached.
-			err, shouldError := ValidateProjectCreation(projectId, config, p)
+			shouldError, err := ValidateProjectCreation(projectId, config, p)
 			if err != nil {
 				if shouldError {
 					return false, errors.Wrap(err, "validating project creation")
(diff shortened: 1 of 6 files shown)


service/project.go
```diff
@@ -350,6 +350,23 @@ func (uis *UIServer) modifyProject(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if !utility.FromBoolPtr(origProjectRef.Enabled) && responseRef.Enabled {
+		settings, err := evergreen.GetConfig()
+		if err != nil {
+			uis.LoggedError(w, r, http.StatusInternalServerError, err)
+			return
+		}
+		err, _ = model.ValidateProjectCreation(responseRef.Identifier, settings, &model.ProjectRef{
```
ablack12, line 359: this comment may have gotten missed but we really should never return error before value; can ValidateProjectCreation be updated?

ablack12 (changes requested):
This is a UIServer function so this isn't what's used by the rest routes. Does https://github.com/evergreen-ci/evergreen/blob/46ee96d25d86db98282907eb8212c7af03ca595b/rest/route/project.go#L330 already have validation?

service/project.go
bynn: done, and will update return values in a later ticket

model/project_ref.go
```diff
@@ -2246,36 +2246,36 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 
 // ValidateProjectCreation returns an error if the total or owner/repo project limit set by the admin settings has been reached.
```
ablack12, line 2247: This comment isn't quite right since it returns an error also if there's an error, and the boolean determines if its an error or warning, but we intend to tweak this to handle the front end right (to use status codes instead)? In which case we can just handle it all later.

PR #6244, opened 2023-02-07
EVG-18835: Patch.projectMetadata GQL field

SupaJoon:
[EVG-18835](https://jira.mongodb.org/browse/EVG-18835)

### Description 
We need Patch.projectMetadata to render the repo/branch name for PatchCards in the case that Version is undefined.
This [Spruce PR](https://github.com/evergreen-ci/spruce/pull/1641) will apply Patch.projectMetadata instead of Version.projectMetadata when rendering PatchCards.

### DIFF
--- a/graphql/tests/patch/results.json
+++ b/graphql/tests/patch/results.json
@@ -330,14 +330,12 @@
       "query_file": "project-metadata.graphql",
       "result": {
         "data": {
-          "data": {
-            "patch": {
-              "id": "5e4ff3abe3c3317e352062e4",
-              "projectMetadata": {
-                "owner": "evergreen-ci",
-                "repo": "evergreen",
-                "identifier": "evergreenIdentifier"
-              }
+          "patch": {
+            "id": "5e4ff3abe3c3317e352062e4",
+            "projectMetadata": {
+              "owner": "evergreen-ci",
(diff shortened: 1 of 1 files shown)


graphql/schema/types/patch.graphql
```diff
@@ -74,6 +74,7 @@ type Patch {
   project: PatchProject
   projectID: String!
   projectIdentifier: String!
+  projectMetadata: Project
```
minnakt, line 77: I wonder if it's necessary to send the entire `Project`, if we only really need the `owner` and `repo`. It also kind of makes the `projectIdentifier` patch resolver redundant. Would it make sense to instead send an object with those two fields?

graphql/schema/types/patch.graphql
```diff
@@ -74,6 +74,7 @@ type Patch {
   project: PatchProject
```
minnakt, line 74: Can you see if it's possible to delete this field? It seems like it's unused by the frontend and it could be confusing to have two different fields with similar names. No worries if it's not possible to delete it though!

minnakt (approved):
reasoning sounds good to me!

ablack12 (approved):

PR #6245, opened 2023-02-07
EVG-16701 Allow bulk version priority setting

hadjri:
[EVG-16701](https://jira.mongodb.org/browse/EVG-16701)

### Description 
Similar to the `getProjectVersionsHandler`, introduced a new `versionsSetPriorityHandler` to perform a bulk priority set on a set of versions.

- Added a field to `GetVersionsOptions` (a revision order number to end at) because the functon `GetVersionsWithOptions` where the options are used shares a lot of the same logic we need for implementing this route, i.e. given two revision order numbers, give me all of the versions in between them.
- Refactored `SetVersionPriority` to accept a list of version IDs to reuse the same code between instances that set the priority of a single version and this new route
- Removed some unused code (and their tests) as I noticed them along the way

Will update wiki with documentation once PR discussion is resolved.
### Testing 
Added unit testing and tested in staging


rest/route/service.go
```diff
@@ -203,6 +203,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/projects/{project_id}/task_reliability").Version(2).Get().Wrap(requireUser).RouteHandler(makeGetProjectTaskReliability(opts.URL))
 	app.AddRoute("/projects/{project_id}/task_stats").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTaskStats(opts.URL))
 	app.AddRoute("/projects/{project_id}/versions").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectVersionsHandler(opts.URL))
+	app.AddRoute("/projects/{project_id}/versions").Version(2).Post().Wrap(requireUser, editTasks).RouteHandler(makeSetPriorityProjectVersionsHandler(opts.URL))
```
malikchaya2, line 206: The route we currently have that allows setting priority requires users to have `submitPatches` permissions.  I don't know if `editTasks` is enough here.

rest/route/project.go
```diff
@@ -962,6 +923,147 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
 	return resp
 }
 
+// POST /rest/v2/projects/{project_id}/versions
+
+// versionsSetPriorityHandler is a RequestHandler for setting the priority of versions.
```
malikchaya2, line 928: I know that we currently only let users bump the priority on patches up to a certain number, unless they have special permissions to bypass it. Do we have that check in place here? I know we are creating this route for disabling, but it's also giving users the power to bump the priority up on a bunch of versions, potentially blocking important versions from running for a while.

malikchaya2 (changes requested):

rest/route/service.go
hadjri: Good point, changed it to `submitPatches`

rest/route/project.go
hadjri: Yep, the route hits [this function](https://gist.github.com/hadjri/0a8e4e99a2df70aee805591d2cd1a27b#file-middleware-go-L116) which will allow those with `TasksAdmin` role to surpass the max priority

malikchaya2 (approved):

rest/route/project.go
```diff
@@ -962,6 +923,147 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+type versionsSetPriorityHandler struct {
+	projectId string
+	url       string
+	opts      dbModel.SetVersionsPriorityOptions
+}
+
+func makeSetPriorityProjectVersionsHandler(url string) gimlet.RouteHandler {
```
ablack12, line 935: Should rename this to something more general in case we want to support things besides priority (since the route name is general)

rest/route/service.go
```diff
@@ -203,6 +203,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/projects/{project_id}/task_reliability").Version(2).Get().Wrap(requireUser).RouteHandler(makeGetProjectTaskReliability(opts.URL))
 	app.AddRoute("/projects/{project_id}/task_stats").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTaskStats(opts.URL))
 	app.AddRoute("/projects/{project_id}/versions").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectVersionsHandler(opts.URL))
+	app.AddRoute("/projects/{project_id}/versions").Version(2).Post().Wrap(requireUser, submitPatches).RouteHandler(makeSetPriorityProjectVersionsHandler(opts.URL))
```
ablack12, line 206: Discussed IRL that we may want to make this a project admin permission since it could be abused.

rest/route/project.go
```diff
@@ -962,6 +923,147 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
 	return resp
 }
 
+// POST /rest/v2/projects/{project_id}/versions
+
+// versionsSetPriorityHandler is a RequestHandler for setting the priority of versions.
+type versionsSetPriorityHandler struct {
```
ablack12, line 929: Discussed also accepting date range in case that's preferred.

rest/route/project.go
```diff
@@ -962,6 +923,154 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+				"non-superusers can only set priority at or below %d", priority, evergreen.MaxTaskPriority),
+			StatusCode: http.StatusForbidden,
+		})
+	}
+	if err = dbModel.SetVersionsPriority(versionIds, priority, user.Id); err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "setting version priorities"))
+	}
+
+	// Construct the response.
+	foundVersions, err := dbModel.VersionFind(dbModel.VersionByIds(versionIds))
```
ablack12, line 987: I'd rather we just returned ok (we don't do this in all routes but I think it should be preferred)

rest/route/service.go
```diff
@@ -203,6 +203,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/projects/{project_id}/task_reliability").Version(2).Get().Wrap(requireUser).RouteHandler(makeGetProjectTaskReliability(opts.URL))
 	app.AddRoute("/projects/{project_id}/task_stats").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectTaskStats(opts.URL))
 	app.AddRoute("/projects/{project_id}/versions").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetProjectVersionsHandler(opts.URL))
+	app.AddRoute("/projects/{project_id}/versions").Version(2).Post().Wrap(requireUser, requireProjectAdmin).RouteHandler(makeModifyProjectVersionsHandler(opts.URL))
```
ablack12, line 206: Since this route modifies existing versions (rather than creates new ones) would prefer patch over post

rest/route/project.go
```diff
@@ -962,6 +923,154 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+	apiVersions := []model.APIVersion{}
+	for _, v := range foundVersions {
+		apiVersion := model.APIVersion{}
+		apiVersion.BuildFromService(v)
+		apiVersions = append(apiVersions, apiVersion)
+	}
+	return gimlet.NewJSONResponse(apiVersions)
+}
+
+// parseGetVersionsOptions parses the request body and query parameters and returns a GetVersionsOptions struct.
```
ablack12, line 1000: Should SetVersionsPriorityOptions be renamed ?

ablack12 (changes requested):

rest/route/project.go
hadjri: Right, renamed to `ModifyVersionsOptions`

model/version.go
```diff
@@ -506,17 +506,33 @@ func GetMainlineCommitVersionsWithOptions(projectId string, opts MainlineCommitV
...
-	Skip           int    `json:"skip"`
-	IncludeBuilds  bool   `json:"include_builds"`
-	IncludeTasks   bool   `json:"include_tasks"`
-	ByBuildVariant string `json:"by_build_variant"`
-	ByTask         string `json:"by_task"`
+// ModifyVersionsOptions is a struct for setting the priority of versions.
+// It uses all the options in the GetVersionsOptions struct, as well as a priority field.
+type ModifyVersionsOptions struct {
+	Priority     *int64 `json:"priority"`
+	StartTimeStr string `json:"start_time"`
```
ablack12, line 513: preference to have json strings match variable names

rest/route/project.go
```diff
@@ -883,83 +884,205 @@ func (h *getProjectVersionsHandler) Parse(ctx context.Context, r *http.Request)
...
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "getting versions for project '%s'", h.projectId))
+	}
+	var versionIds []string
+	for _, v := range versions {
+		versionIds = append(versionIds, utility.FromStringPtr(v.Id))
+	}
+
+	// Check for a valid priority and perform the update.
+	if ok := validPriority(priority, h.projectId, user); !ok {
```
ablack12, line 975: Can we check this before doing the version query?

rest/route/project.go
```diff
@@ -883,83 +884,205 @@ func (h *getProjectVersionsHandler) Parse(ctx context.Context, r *http.Request)
...
+	if requester != "" {
+		opts.Requester = requester
+	}
+	if opts.Requester == "" {
+		opts.Requester = evergreen.RepotrackerVersionRequester
+	}
+	return opts, nil
+}
 
+func getStartAndEndTime(startTimeStr, endTimeStr string) (start time.Time, end time.Time, err error) {
```
ablack12, line 1070: Can we remove the names from these return values?

ablack12 (changes requested):

model/version.go
```diff
@@ -534,8 +550,17 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
-	if opts.StartAfter > 0 {
-		match[VersionRevisionOrderNumberKey] = bson.M{"$lt": opts.StartAfter}
+	if !opts.StartTime.IsZero() {
+		match[VersionCreateTimeKey] = bson.M{
+			"$gte": opts.StartTime,
+			"$lte": opts.EndTime,
+		}
+	} else {
+		if opts.EndAt > 0 && opts.StartAfter > 0 {
+			match[VersionRevisionOrderNumberKey] = bson.M{"$lte": opts.StartAfter, "$gte": opts.EndAt}
```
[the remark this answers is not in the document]
hadjri: Good idea

rest/route/project.go
hadjri: Sure

rest/route/project.go
hadjri: Sure. I only added them since we're returning two `Time` structs and thought it'd be good to be clearer which is which. But if it's a pattern we'd like to avoid for consistency that's fine.

ablack12: That logic makes sense -- in that case maybe a function comment?

ablack12 (approved):

PR #6246, opened 2023-02-07
EVG-18384: Add task.taskFiles resolver

minnakt:
[EVG-18384](https://jira.mongodb.org/browse/EVG-18384)

### Description 
Currently `taskFiles` falls under the query resolver. However, since `taskFiles` belongs to a task, it would make more sense to fall under the task resolver. This PR makes it so that `taskFiles` can be retrieved from the task resolver.

Note that the code for the new `task.taskFiles` resolver is literally copy pasted from the old `query.taskFiles` resolver, with the exception that it does not need to do a task lookup anymore.

### Testing 
- Modified existing GraphQL tests for taskFiles
- Added test in `graphql/tests/task/queries/taskFiles.graphql`
- e2e tests should pass because I didn't delete the old resolver
- On staging


khelif96 (approved):

graphql/task_resolver.go
```diff
@@ -509,6 +509,43 @@ func (r *taskResolver) Status(ctx context.Context, obj *restModel.APITask) (stri
 	return *obj.DisplayStatus, nil
 }
 
+// TaskFiles is the resolver for the taskFiles field.
+func (r *taskResolver) TaskFiles(ctx context.Context, obj *restModel.APITask) (*TaskFiles, error) {
```
bynn, line 513: do we have plans to remove the other taskfiles function in query so we dont have the same thing twice?

bynn (approved):

graphql/task_resolver.go
minnakt: yessir I'll do it within this ticket 🫡

PR #6247, opened 2023-02-07
EVG-18577: dynamically switch parser project location

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18577

### Description 
During generate.tasks, move the parser project to S3 if it hits the document too large error (i.e. the parser project is bigger than the DB's 16 MB document limit). Unfortunately, I don't have a way of verifying whether or not relying on the document too large error will cause any DB performance issues, but my reasoning was that the document too large error would only be hit at most once per version (at which point it would move to S3), and most versions won't need to resort to using S3 except the largest ones.

An alternative implementation could be compute the parser project BSON size eagerly before attempting to put it in the DB, which would avoid the document too large issue. However, for large parser projects (~14 MB or bigger), this adds an extra ~250 ms to the runtime of generate.tasks, so I opted not to do it this way.

### Testing 
Added unit tests.


model/version.go
```diff
@@ -244,6 +244,22 @@ func (v *Version) MarkFinished(status string, finishTime time.Time) error {
 	)
 }
 
+// UpdateProjectStorageMethod updates the version's parser project storage
+// method.
+func (v *Version) UpdateProjectStorageMethod(method ParserProjectStorageMethod) error {
+	if method == v.ProjectStorageMethod {
```
hadjri, line 250: I know this is saving some DB calls but maybe we should consider removing this in case the struct has been modified but the DB has yet to update it?

Kimchelly: I guess it's possible that this could happen, but I feel like we should usually assume that the in-memory struct hasn't been modified when it's supposed to be updated only after the DB update succeeds. We make this assumption about the struct state in a lot of places.

I know some places in the code base don't follow this struct-DB update state agreement, but they probably should.

hadjri: Fair enough!

hadjri (approved):

PR #6248, opened 2023-02-07
EVG-18508 handle untracked branches with task ID generation

ablack12:
[EVG-18508](https://jira.mongodb.org/browse/EVG-18508)

### Description 
Tasks created for untracked branches (as described [here](https://github.com/evergreen-ci/evergreen/wiki/Using-Repo-Level-Settings#how-to-use-pr-testing-for-untracked-branches) don't have user-defined readable identifiers, so we should provide something else (owner_repo_branch proposed here) to provide context.


model/project.go
```diff
@@ -958,7 +958,12 @@ func generateIdsForVariant(vt TVPair, proj *Project, v *Version, tasks TVPairSet
 	return table
 }
 
+// generateId generates a unique project ID. For tasks created for untracked branches,
+// use owner/repo/branch in lieu of a user-defined project identifier.
```
hadjri, line 962: Nit: owner_repo_branch

hadjri (approved):

model/project.go
ablack12: I more meant to enumerate the fields we were using in place but I see why this isn't clear; i'll switch to owner, repo, and branch

PR #6249, opened 2023-02-08
EVG-18293: Add logkeeper resolvers

sophstad:
[EVG-18293](https://jira.mongodb.org/browse/EVG-18293)

### Description 
- Call Logkeeper client to return a build's test results, as introduced by https://github.com/evergreen-ci/plank/pull/1
- Let me know if you think this organization doesn't make sense—I opted to move what little logic was required into `rest/data` instead of handling everything in the resolver body but I don't know what would really be the most ideal place for this.

### Testing 
- Interested in any thoughts on what a useful test may look like here... the client tests itself in terms of fetching build results, and all we're doing on evergreen is really piping in the build ID and Logkeeper URL
- I verified that this successfully fetches production Logkeeper data using the GraphQL playground


rest/data/logkeeper.go
```diff
@@ -0,0 +1,16 @@
+package data
```
khelif96, line 1: I would defer to the app folks over organization but my intuition is this probably shouldn't live here since this won't be used by the rest routes at all.

khelif96 (commented):
Testing this is tricky we might be able to start a logkeeper instance and have evergreen talk to it but that can also be done in spruce (I would assume we would need to do that anyways to write e2e tests) Parsley does something similar if you want to look there for inspiration.

Apart from the comments the rest lgtm.

rest/data/logkeeper.go
ablack12: Is there a reason this needs to be in a package external to graphql at all? If it's only used by graphql then I would think this code could just live in the resolver itself, or the graphql/util.go file

sophstad: Sounds good, I wasn't sure if calling an external api from GraphQL was discouraged :o) glad to know it's not!

ablack12: actually idk maybe that's a good point, i wasn't thinking of this as an external thing 😅 we don't really have something similar that i can think of. It could live under model the same way cedar stuff seems to (https://github.com/evergreen-ci/evergreen/blob/0c54078f0a6112dd737ac88e048adcb608dd0079/model/task_log.go) but yeah idk, i guess the primary benefit of having it elsewhere would be if we wanted to use a mock version of the client for tests but it seems that wouldn't provide much value since we just return the results verbatim

sophstad: Yeah, I think for now interacting with Plank directly from the resolver is nice and lightweight, but maybe in the future this would make sense!

PR #6253, opened 2023-02-09
EVG-18577: do not put context timeout on inserting tasks/builds for generate.tasks

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18577

### Description 
Inserting tasks/builds is very expensive and oftentimes takes several minutes to finish for full server generators, so I reverted it back to the old logic of letting the insertion run for an unlimited amount of time. Probably not that great long term, but I'm not looking to optimize the build/task insertion right now.

### Testing
Don't really think a context timeout modification is worth figuring out a test.

### DIFF
--- a/units/generate_tasks.go
+++ b/units/generate_tasks.go
@@ -196,10 +196,9 @@ func (j *generateTasksJob) generate(ctx context.Context, t *task.Task) error {
 	start = time.Now()
 
 	// Don't use the job's context, because it's better to try finishing than to
-	// exit early after a SIGTERM from app server shutdown.
-	ctx, cancel := context.WithTimeout(context.Background(), model.DefaultParserProjectAccessTimeout)
-	defer cancel()
-	err = g.Save(ctx, j.env.Settings(), p, pp, v)
+	// exit early after a SIGTERM from app server shutdown. This should probably
+	// use a timeout rather than just the background context.
+	err = g.Save(context.Background(), j.env.Settings(), p, pp, v)
 
 	// If the version or parser project has changed there was a race. Another generator will try again.
 	if adb.ResultsNotFound(err) || db.IsDuplicateKey(err) {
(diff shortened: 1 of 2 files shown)


bynn (approved):
was this because server had a problem with their build time? do we have plans to change it back once they fix their problem? 

otherwise looks fine

Kimchelly:
The old logic has a background context on `Save` so it could run forever. I think this isn't something quickly fixable due to how many tasks each generate.tasks creates in the server (inserting all the server tasks can take > 5 minutes), so we would probably have to improve the performance of the build/task insertion ourselves.

bynn:
sounds good! Should we make a ticket for investigating if we could improve bulk insert times?

units/generate_tasks.go
```diff
@@ -196,10 +196,9 @@ func (j *generateTasksJob) generate(ctx context.Context, t *task.Task) error {
 	start = time.Now()
 
 	// Don't use the job's context, because it's better to try finishing than to
-	// exit early after a SIGTERM from app server shutdown.
-	ctx, cancel := context.WithTimeout(context.Background(), model.DefaultParserProjectAccessTimeout)
-	defer cancel()
-	err = g.Save(ctx, j.env.Settings(), p, pp, v)
+	// exit early after a SIGTERM from app server shutdown. This should probably
+	// use a timeout rather than just the background context.
```
ablack12, line 200: Is the timeout only used for inserting tasks/variants? In which case I don't think the ticket to optimize generating tasks/variants is going to improve this situation.

Either way, this comment isn't ideal since it doesn't really explain why there isn't currently a timeout. I think if we want to keep a timeout we should just increase it, or if we think a timeout can be reintroduced after x specific work we should link the ticket for it.

Kimchelly: I just re-added the comment that used to be there.

I'll open a follow-up ticket to put a context deadline eventually.

Kimchelly: https://jira.mongodb.org/browse/EVG-18929

PR #6254, opened 2023-02-09
EVG-18878 Remove concept of private projects, Add 2nd auth flag

hadjri:
[EVG-18878](https://jira.mongodb.org/browse/EVG-18878)

### Description 
- Made a 2nd auth toggle admin flag, such that there is one for the UI pages, and one for the REST routes, so they can be toggled separately.

Will perform a DB update on all existing non-wiredtiger projects in prod to set them all to private once this passes.

### Testing 
Updated existing unit tests, deployed to staging and confirmed that the admin toggles work as expected.
    
Notions of private / public projects need to be removed from the wiki and additional clarification should be added to emphasize that all projects are private by default.


globals.go
```diff
@@ -291,6 +291,9 @@ const (
 	// Valid types of performing git clone
 	CloneMethodLegacySSH = "legacy-ssh"
 	CloneMethodOAuth     = "oauth"
+
+	// PublicRepoName is the repo name for which a project ref is considered public.
+	PublicRepoName = "wiredtiger"
```
Kimchelly, line 296: I know it's a one-off exception, but it is a little hackish to hard-code the one special repo. Is the plan to keep this until the rest of time, or is there a plan to get them to use auth?

model/project_test.go
```diff
@@ -1416,16 +1414,15 @@ func TestFindProjectsSuite(t *testing.T) {
...
 				Owner:       "mongodb",
 				Repo:        "mongo",
 				Branch:      "main",
 			},
-			{Id: "projectD", Private: utility.FalsePtr()},
-			{Id: "projectE", Private: utility.FalsePtr()},
-			{Id: "projectF", Private: utility.TruePtr()},
+			{Id: "projectD", Repo: evergreen.PublicRepoName},
+			{Id: "projectE", Repo: evergreen.PublicRepoName},
+			{Id: "projectF"},
```
Kimchelly, line 1425: It seems like the case on L1420 would already be testing a private repo, so can we remove this project ref?

globals.go
```diff
@@ -291,6 +291,9 @@ const (
 	// Valid types of performing git clone
 	CloneMethodLegacySSH = "legacy-ssh"
 	CloneMethodOAuth     = "oauth"
+
+	// PublicRepoName is the repo name for which a project ref is considered public.
```
Kimchelly, line 295: Small suggestion to define what "public" means in this context. I would just say something like it has an exemption from auth requirements for REST routes?

rest/data/project_test.go
```diff
@@ -90,34 +90,30 @@ func TestProjectConnectorGetSuite(t *testing.T) {
...
 				Owner:       "mongodb",
 				Repo:        "mongo",
 				Branch:      "main",
 			},
-			{Id: "projectD", Private: utility.FalsePtr()},
-			{Id: "projectE", Private: utility.FalsePtr()},
-			{Id: "projectF", Private: utility.TruePtr()},
+			{Id: "projectD", Repo: evergreen.PublicRepoName},
+			{Id: "projectE", Repo: evergreen.PublicRepoName},
+			{Id: "projectF"},
```
Kimchelly, line 116: Same thing as above, this seems to be covered by the project ref on L110.

rest/route/middleware.go
```diff
@@ -214,7 +214,7 @@ func (m *projectRepoMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Reques
 
 // NewRequireAuthHandler provides middleware that wraps gimlet.NewRequireAuthHandler
 // and requires that users be authenticated generally to access the resource.
-// This middleware's active state depends on the admin service flag PartialRouteAuthDisabled
+// This middleware's active state depends on the admin service flag RestRouteRequireAuthDisabled
```
Kimchelly, line 217: nit: period at the end of the doc comment.

service/api.go
```diff
@@ -88,15 +88,15 @@ func MustHaveProject(r *http.Request) *model.Project {
 }
 
 // requireUserToggleable wraps gimlet.NewRequireAuthHandler and checks that
-// a user is generally authenticated if the PartialRouteAuthDisabled flag is set.
+// a user is generally authenticated if the RestRouteRequireAuthDisabled flag is set.
```
Kimchelly, line 91: This comment phrasing implies the opposite of what it actually does - if the `RestRouteRequireAuthDisabled` flag is true (i.e. set), then it will skip the auth requirement. Maybe phrase this as `if the RestRouteRequireAuthDisabled flag is false`?

service/middleware.go
```diff
@@ -119,7 +119,7 @@ func (uis *UIServer) GetSettings() evergreen.Settings {
 // requireUser takes a request handler and returns a wrapped version which verifies that requests
 // request are authenticated before proceeding. For a request which is not authenticated, it will
 // execute the onFail handler. If onFail is nil, a simple "unauthorized" error will be sent.
-// skipWithToggle indicates that we will skip the authentication if PartialRouteAuthDisabled flag
+// skipWithToggle indicates that we will skip the authentication if UIRequireAuthDisabled flag
 // is set. This is a temporary parameter that is in place while we determine if adding auth to all
```
Kimchelly, line 123: Same phrasing suggestion as above about setting a flag that disables.

Kimchelly (changes requested):

globals.go
ablack12: Rather than this, I'd rather we just have wiredtiger keep Private set to true. We aren't sure yet I think what the longterm plan is for wiredtiger or if other projects will end up having an argument for this treatment.

ablack12 (changes requested):

graphql/schema/types/project.graphql
```diff
@@ -182,7 +182,6 @@ type Project {
   patchTriggerAliases: [PatchTriggerAlias!]
   perfEnabled: Boolean
   periodicBuilds: [PeriodicBuild!]
-  private: Boolean
```
khelif96, line 185: This being used on the frontend, you would first need to remove its use on the UI before deleting it here.

khelif96 (changes requested):
Only reviewed the graphql parts.

graphql/schema/types/project.graphql
hadjri: Not deleting anything from the schema anymore 😅

globals.go
hadjri: Sounds fair enough, I suppose I got too excited at the concept of deleting a bunch of code centered around `private`. Reverted it back to having this flag, all instances of this `PublicRepoName` are gone. This change will now come with a proposed prod DB update needing LGTM in the ops channel to set all projects to private except the 5-6 wiredtiger projects.

ablack12: Before making the DB update we can also probably follow up with Alex on the ticket to see if he can confirm if this should be considered for all wiredtiger projects.

service/api.go
hadjri: The naming schema for our admin flags messed me up here. Since they all have "Disabled" in their variable name, yet on the actual admin page [Enabled](https://evergreen.mongodb.com/admin) really means the value is false (which is all flags default value) under the hood is unintuitive. I'm considering if a ticket to make this easier to understand would be helpful.

I think the real problem is I named the variables incorrectly. I've now changed them from `RestRouteRequireAuthDisabled ` --> `RestRoutePartialAuthDisabled `, to make the variable name actually do what it says it does when reading the code. The value will initially be false, meaning the [admin page](https://evergreen.mongodb.com/admin) will say that "partial auth" will be enabled, which means the code `if flags.RestRoutePartialAuthDisabled { do_auth() }` makes sense

graphql/schema/types/project.graphql
khelif96: sgtm if you do decide to make some changes to gql feel free to tag me again.

Kimchelly (approved):

khelif96:
Just wanted to confirm that I am not blocking this review. I'm just unable to remove my request for review.

ablack12 (approved):

hadjri:
Re-requesting review because the waterfall/patch/task/etc pages I assume users of public projects will be visiting need to also have their endpoints public as well after the `RestRoutePartialAuthDisabled` flag is toggled (or else we are just serving static HTML).  

- Changed the behavior of REST V1 from always checking the `RestRoutePartialAuthDisabled` flag to only checking the flag if the project context is nil (so routes dealing with a public project will not have auth on them)

- Changed `IsPrivate` back to being a simple bool ptr function, since the `RestRoutePartialAuthDisabled` should not affect the private field of a project ref.

ablack12 (approved):

PR #6256, opened 2023-02-10
EVG-18820: reduce agent parser project fetches

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18820

### Description 
When the agent is preparing to run the task, it fetches the expansions, project vars, and parameters, which is split across two routes (`/task/{task_id}/fetch_vars` and `/task/{task_id}/expansions`), and each of those routes fetches the parser project to get one or two fields from the project. This change reduces the number of times the parser project is accessed during this task setup from 3 down to 1, since downloading large parser projects from S3 can be expensive. Evergreen still has to load the parser project once to send the whole project to the agent, but for the expansions/vars, I pushed the logic that needed the parser project in the REST route into the agent (which already has the project on hand).

* Consolidate expansions and vars fetching from 2 routes into a 1 route, and remove any accesses to the parser project. There's going to be a follow-up PR to remove the old routes.
* Push logic in the expansions/vars route that required parser project information into the agent, which already has a copy of the project.

### Testing 
* Added unit tests.
* Tested in staging that nothing broke.
    
No documentation needed.


Kimchelly:
evergreen retry

Kimchelly:
evergreen retry

Kimchelly:
evergreen retry

rest/route/agent.go
```diff
@@ -353,14 +354,128 @@ func (h *getExpansionsHandler) Run(ctx context.Context) gimlet.Responder {
...
+}
+
+func (h *getExpansionsAndVarsHandler) Parse(ctx context.Context, r *http.Request) error {
+	if h.taskID = gimlet.GetVars(r)["task_id"]; h.taskID == "" {
+		return errors.New("missing task ID")
+	}
+	h.hostID = r.Header.Get(evergreen.HostHeader)
+	podID := r.Header.Get(evergreen.PodHeader)
+	if h.hostID == "" && podID == "" {
+		return errors.New("missing both host and pod ID")
```
hadjri, line 403: I think this is unnecessary since the `requirePodOrHost` middleware set on this route already handles this case

hadjri (changes requested):

rest/route/agent.go
Kimchelly: I just copy-pasted this from the existing expansions/vars route. It is unnecessary because of the auth middleware, but I kind of prefer to keep it around just because 1. the auth middleware is not actually important here, it simply happens to conveniently bump the last communication time, and 2. the middleware set for the route is independent of the route handler itself so I'd prefer the handler validate its own inputs. For example, if we changed the middleware behavior or removed the host/pod auth middleware, this route handler would still want to validate the correctness of the required request inputs during parsing.

hadjri (approved):

ablack12 (approved):

PR #6259, opened 2023-02-14
EVG-18784 Warn users when creating/copying projects at limit

bynn:
[EVG-18784](https://jira.mongodb.org/browse/EVG-18784)

### Description 
when creating/copying projects on spruce, users will now receive a warning toast 

the validation function has been updated to return an error code and handle all edge cases that it didn't handle before 
(test has been added for that)

### Testing 
staging
creating and copying projects that are at the limit will warn the users
    
eventually (there is a ticket for create project documentation)

### DIFF
--- a/graphql/mutation_resolver.go
+++ b/graphql/mutation_resolver.go
@@ -415,8 +415,6 @@ func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.
 				if apiErr.StatusCode == http.StatusBadRequest {
 					return nil, InputValidationError.Send(ctx, apiErr.Message)
 				}
-				// StatusNotFound and other error codes are really internal errors bc we determine this input
-				return nil, InternalServerError.Send(ctx, apiErr.Message)
 			}
 			return nil, InternalServerError.Send(ctx, err.Error())
 		}
--- a/graphql/mutation_resolver.go
+++ b/graphql/mutation_resolver.go
@@ -408,16 +408,19 @@ func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.
 	}
 	u := gimlet.GetUser(ctx).(*user.DBUser)
 
-	if err := data.CreateProject(ctx, evergreen.GetEnvironment(), dbProjectRef, u); err != nil {
-		apiErr, ok := err.(gimlet.ErrorResponse)
-		if ok {
-			if apiErr.StatusCode == http.StatusBadRequest {
-				return nil, InputValidationError.Send(ctx, apiErr.Message)
+	if created, err := data.CreateProject(ctx, evergreen.GetEnvironment(), dbProjectRef, u); err != nil {
+		if !created {
+			apiErr, ok := err.(gimlet.ErrorResponse)
+			if ok {
+				if apiErr.StatusCode == http.StatusBadRequest {
+					return nil, InputValidationError.Send(ctx, apiErr.Message)
+				}
+				// StatusNotFound and other error codes are really internal errors bc we determine this input
+				return nil, InternalServerError.Send(ctx, apiErr.Message)
 			}
-			// StatusNotFound and other error codes are really internal errors bc we determine this input
-			return nil, InternalServerError.Send(ctx, apiErr.Message)
+			return nil, InternalServerError.Send(ctx, err.Error())
 		}
-		return nil, InternalServerError.Send(ctx, err.Error())
+		graphql.AddError(ctx, PartialError.Send(ctx, err.Error()))
 	}
 
 	projectRef, err := model.FindBranchProjectRef(*project.Identifier)
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2037,6 +2037,10 @@ func DefaultSectionToRepo(projectId string, section ProjectPageSection, userId s
 	}
 
 	if section == ProjectPageGeneralSection {
+		originalMergedRef, err := GetProjectRefMergedWithRepo(before.ProjectRef)
+		if err != nil {
+			return errors.Wrap(err, "getting merged project ref for validation")
+		}
 		newMergedRef, err := GetProjectRefMergedWithRepo(ProjectRef{
 			RepoRefId: before.ProjectRef.RepoRefId,
 		})
@@ -2047,7 +2051,7 @@ func DefaultSectionToRepo(projectId string, section ProjectPageSection, userId s
 		if err != nil {
 			return errors.Wrap(err, "getting evergreen config")
 		}
-		_, err = ValidateEnabledProjectsLimit(projectId, config, *newMergedRef)
+		_, err = ValidateEnabledProjectsLimit(projectId, config, originalMergedRef, newMergedRef)
 		if err != nil {
 			return errors.Wrap(err, "validating project creation")
 		}
@@ -2260,42 +2264,39 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 	return res, nil
 }
 
+func shouldValidateOwnerRepoLimit(originalMergedRef, mergedRefToValidate *ProjectRef) bool {
+	return originalMergedRef == nil || !originalMergedRef.IsEnabled() || originalMergedRef.Owner != mergedRefToValidate.Owner || originalMergedRef.Repo != mergedRefToValidate.Repo
+}
+
 // ValidateEnabledProjectsLimit takes in a merged project ref and validates if the current project is allowed to be enabled.
-// Returns a status code and error we are already at limit with enabled projects.
-func ValidateEnabledProjectsLimit(projectId string, config *evergreen.Settings, mergedRefToValidate ProjectRef) (int, error) {
-	if config.ProjectCreation.TotalProjectLimit == 0 || config.ProjectCreation.RepoProjectLimit == 0 {
-		return 0, nil
-	}
-	originalRef, err := FindMergedProjectRef(projectId, "", false)
-	if err != nil {
-		return http.StatusInternalServerError, errors.Wrapf(err, "getting project '%s'", projectId)
+// Returns a status code and error if we are already at limit with enabled projects.
+func ValidateEnabledProjectsLimit(projectId string, config *evergreen.Settings, originalMergedRef, mergedRefToValidate *ProjectRef) (int, error) {
+	if config.ProjectCreation.TotalProjectLimit == 0 || config.ProjectCreation.RepoProjectLimit == 0 || !mergedRefToValidate.IsEnabled() {
+		return http.StatusOK, nil
 	}
 
+	isNewProject := originalMergedRef == nil
 	catcher := grip.NewBasicCatcher()
-	if originalRef == nil || !originalRef.IsEnabled() {
+	// Only validate total project limit if it's a new project or was previously disabled.
+	if isNewProject || !originalMergedRef.IsEnabled() {
 		allEnabledProjects, err := GetNumberOfEnabledProjects()
 		if err != nil {
 			return http.StatusInternalServerError, errors.Wrap(err, "getting number of projects")
 		}
-		if originalRef == nil {
-			allEnabledProjects++
-		}
-		// Only validate total project limit if it's a new project or was previously disabled.
+		allEnabledProjects++
 		if allEnabledProjects >= config.ProjectCreation.TotalProjectLimit {
-			catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
+			catcher.Errorf("total enabled project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
 		}
 	}
 
 	// Only validate owner/repo limit if it's a new project, was previously disabled, or we're updating owner/repo.
-	if originalRef == nil || !originalRef.IsEnabled() || originalRef.Owner != mergedRefToValidate.Owner || originalRef.Repo != mergedRefToValidate.Repo {
+	if isNewProject || !originalMergedRef.IsEnabled() || originalMergedRef.Owner != mergedRefToValidate.Owner || originalMergedRef.Repo != mergedRefToValidate.Repo {
 		if !config.ProjectCreation.IsExceptionToRepoLimit(mergedRefToValidate.Owner, mergedRefToValidate.Repo) {
 			enabledOwnerRepoProjects, err := GetNumberOfEnabledProjectsForOwnerRepo(mergedRefToValidate.Owner, mergedRefToValidate.Repo)
 			if err != nil {
 				return http.StatusInternalServerError, errors.Wrapf(err, "getting number of projects for '%s/%s'", mergedRefToValidate.Owner, mergedRefToValidate.Repo)
 			}
-			if originalRef == nil {
-				enabledOwnerRepoProjects++
-			}
+			enabledOwnerRepoProjects++
 			if enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit {
 				catcher.Errorf("owner repo limit of %d reached for '%s/%s'", config.ProjectCreation.RepoProjectLimit, mergedRefToValidate.Owner, mergedRefToValidate.Repo)
 			}
@@ -2305,7 +2306,7 @@ func ValidateEnabledProjectsLimit(projectId string, config *evergreen.Settings, 
 	if catcher.HasErrors() {
 		return http.StatusBadRequest, catcher.Resolve()
 	}
-	return 0, nil
+	return http.StatusOK, nil
 }
 
 func (p *ProjectRef) ValidateOwnerAndRepo(validOrgs []string) error {
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -929,7 +929,7 @@ func GetNumberOfEnabledProjects() (int, error) {
 	return getNumberOfEnabledProjects("", "")
 }
 
-// GetEnabledProjectsForOwnerRepo returns the number of enabled projects for a given owner/repo.
+// GetNumberOfEnabledProjectsForOwnerRepo returns the number of enabled projects for a given owner/repo.
 func GetNumberOfEnabledProjectsForOwnerRepo(owner, repo string) (int, error) {
 	if owner == "" || repo == "" {
 		return 0, errors.New("owner and repo must be specified")
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1904,24 +1904,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 		if p.TracksPushEvents != nil {
 			setUpdate[ProjectRefTracksPushEventsKey] = p.TracksPushEvents
 		}
-		config, error := evergreen.GetConfig()
 		if !isRepo && !p.UseRepoSettings() && !defaultToRepo {
+			config, err := evergreen.GetConfig()
 			if err != nil {
-				return false, errors.Wrap(error, "getting evergreen config")
+				return false, errors.Wrap(err, "getting evergreen config")
 			}
 			// Allow a user to modify owner and repo only if they are editing an unattached project
-			if error = p.ValidateOwnerAndRepo(config.GithubOrgs); error != nil {
-				return false, errors.Wrap(error, "validating new owner/repo")
+			if err = p.ValidateOwnerAndRepo(config.GithubOrgs); err != nil {
+				return false, errors.Wrap(err, "validating new owner/repo")
 			}
 
 			setUpdate[ProjectRefOwnerKey] = p.Owner
 			setUpdate[ProjectRefRepoKey] = p.Repo
 
-		}
-		// Cannot enable projects if the project creation limits have been reached.
-		_, error = ValidateProjectCreation(projectId, config, p)
-		if error != nil {
-			return false, errors.Wrap(error, "validating project creation")
 		}
 		// some fields shouldn't be set to nil when defaulting to the repo
 		if !defaultToRepo {
@@ -2040,6 +2035,23 @@ func DefaultSectionToRepo(projectId string, section ProjectPageSection, userId s
 
 	if err != nil {
 		return errors.Wrap(err, "getting before project settings event")
+	}
+
+	if section == ProjectPageGeneralSection {
+		newMergedRef, err := GetProjectRefMergedWithRepo(ProjectRef{
+			RepoRefId: projectId,
+		})
+		if err != nil {
+			return errors.Wrap(err, "getting merged project ref for validation")
+		}
+		config, err := evergreen.GetConfig()
+		if err != nil {
+			return errors.Wrap(err, "getting evergreen config")
+		}
+		_, err = ValidateEnabledProjectsLimit(projectId, config, *newMergedRef)
+		if err != nil {
+			return errors.Wrap(err, "validating project creation")
+		}
 	}
 
 	modified, err := SaveProjectPageForSection(projectId, nil, section, false)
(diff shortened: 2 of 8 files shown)


bynn:
evergreen retry

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 }
 
 // ValidateProjectCreation returns a boolean if you should surface the error or not.
-func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (bool, error) {
+func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (int, error) {
```
malikchaya2, line 2253: Can we rename this to include the word `limit`? Like `CheckProjectLimit` or `ValidateProjectCreationLimit`?

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
-		return true, errors.Wrap(err, "getting number of projects")
+		return http.StatusInternalServerError, errors.Wrapf(err, "getting project '%s'", projectId)
 	}
-	allEnabledProjectsId := []string{}
-	for _, p := range allEnabledProjects {
-		allEnabledProjectsId = append(allEnabledProjectsId, p.Id)
+	if originalRef != nil {
+		if projectRef == nil {
+			// Should be unreachable
+			return http.StatusInternalServerError, errors.New("cannot create project while defaulting to repo")
```
malikchaya2, line 2264: Is this reached if someone tries to copy an existing project that defaults to the repo? In that case should we change the error message to say "create or copy"?

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 }
 
 // ValidateProjectCreation returns a boolean if you should surface the error or not.
```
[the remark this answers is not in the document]
malikchaya2: +1

malikchaya2 (approved):
approved with some requests

graphql/mutation_resolver.go
```diff
@@ -408,16 +408,19 @@ func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.
...
-			if apiErr.StatusCode == http.StatusBadRequest {
-				return nil, InputValidationError.Send(ctx, apiErr.Message)
+	if created, err := data.CreateProject(ctx, evergreen.GetEnvironment(), dbProjectRef, u); err != nil {
+		if !created {
+			apiErr, ok := err.(gimlet.ErrorResponse)
+			if ok {
+				if apiErr.StatusCode == http.StatusBadRequest {
+					return nil, InputValidationError.Send(ctx, apiErr.Message)
+				}
+				// StatusNotFound and other error codes are really internal errors bc we determine this input
```
ablack12, line 418: Do we need this return statement (isn't it just the same as the one below)?

model/project_ref.go
```diff
@@ -1896,27 +1904,24 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 		if p.TracksPushEvents != nil {
 			setUpdate[ProjectRefTracksPushEventsKey] = p.TracksPushEvents
 		}
+		config, error := evergreen.GetConfig()
```
ablack12, line 1907: shouldn't use error since this is the package name. If you want to have distinct error names I would make this one specific, like configErr. It's pretty unclear to me what's going on here though.

model/project_ref.go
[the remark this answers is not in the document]
ablack12: This also needs some more details about what it's validating, as a public function. And it appears that it can be expected for the inputted projectRef to be nil. Can you expand on that?

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
 // ValidateProjectCreation returns a boolean if you should surface the error or not.
-func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (bool, error) {
+func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (int, error) {
 	if config.ProjectCreation.TotalProjectLimit == 0 || config.ProjectCreation.RepoProjectLimit == 0 {
-		return false, nil
+		return 0, nil
 	}
-	catcher := grip.NewBasicCatcher()
-	allEnabledProjects, err := GetEnabledProjects()
+	originalRef, err := findOneProjectRefQ(byId(projectId))
```
ablack12, line 2257: Why not call FindMergedProjectRef if you end up merging with repo anyway? (Otherwise FindBranchProjectRef)

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+			return http.StatusInternalServerError, errors.Wrapf(err, "getting merged project for '%s'", originalRef.Id)
+		}
 	}
-	// No need to validate if project was enabled to begin with.
-	if utility.StringSliceContains(allEnabledProjectsId, projectId) {
-		return false, nil
+	var newRef *ProjectRef
+	if projectRef == nil {
+		// Default to repo case
+		repoRef, err := FindOneRepoRef(originalRef.Id)
```
ablack12, line 2274: Should add a comment to explain why this needs repo ref.

model/project_ref.go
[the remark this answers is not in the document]
ablack12: This is also now about enabling a project, not creating one

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		}
+		originalRef, err = GetProjectRefMergedWithRepo(*originalRef)
+		if err != nil {
+			return http.StatusInternalServerError, errors.Wrapf(err, "getting merged project for '%s'", originalRef.Id)
+		}
 	}
-	// No need to validate if project was enabled to begin with.
-	if utility.StringSliceContains(allEnabledProjectsId, projectId) {
-		return false, nil
+	var newRef *ProjectRef
```
ablack12, line 2271: refToValidate might be clearer since we aren't creating a new project

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+	if projectRef == nil {
```
ablack12, line 2272: In this case we could instead just set projectRef to ProjectRef{RepoRefId: <id needed>} and then call GetProjectRefMergedWithRepo for both cases

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
-	pRef, err := findOneProjectRefQ(byId(projectId))
-	if err != nil {
-		return true, errors.Wrapf(err, "getting project '%s", projectId)
-	}
-	if pRef != nil {
-		// Defaulting to repo
-		if projectRef == nil {
-			projectRef, err = GetProjectRefMergedWithRepo(*pRef)
+	catcher := grip.NewBasicCatcher()
+	if originalRef == nil || newRef.IsEnabled() {
```
ablack12, line 2293: These if statements need some rethinking or commenting. What is going on here? Why do we check if originalRef == nil in both if statements? Does that imply "defaultToRepo"? We should set that at the top instead to make it way clearer what's going on.

model/project_ref.go
ablack12: I'm actually confused about why all of this is necessary. Why can't we just pass in mergedSection https://github.com/evergreen-ci/evergreen/blob/d80b595462048b97a7557e24f57483eecd9beb92/rest/data/project_settings.go#L217 to understand if the thing is going to be enabled or not?

ablack12 (changes requested):
Does this handle the "isRepo" case where a repo is being set to enabled and all the branches are disabled?

I'm thinking this is too complex. Do you know who is using a repo to determine whether a branch is enabled or not? It may be more worth removing that option so we can _way_ simplify this validation. I'm also not fully sure this needs to be as complex as it is though.

rest/data/project.go
```diff
@@ -108,12 +109,19 @@ func CreateProject(ctx context.Context, env evergreen.Environment, projectRef *m
 		}
 	}
 	if err := VerifyUniqueProject(projectRef.Id); err != nil {
-		return err
+		return false, err
 	}
 	// Always warn because created projects are never enabled.
-	_, err := model.ValidateProjectCreation(projectRef.Id, env.Settings(), projectRef)
+	catcher := grip.NewBasicCatcher()
```
ablack12, line 115: Can we call this warningCatcher or something?

model/project_ref.go
bynn: My reasoning for it was that when we call the validate function, we don't always have the merged ref in other places so I'd just take the minimal info and handle every case (like defaulting to repo) in the function.

But I think it does make more sense to make the function simpler and handle the other cases separately

model/project_ref.go
bynn: deleted

model/project_ref.go
bynn: the original == nil checks if the project previously exists or not and that means we're creating/copying

I don't think putting it on the top works because there's an extra condition for checking the owner/repo and that would duplicate more code. I put in comments to make this clearer

rest/data/project.go
bynn: done

model/project_ref.go
bynn: removed

model/project_ref.go
```diff
@@ -2245,54 +2250,78 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
 	if err != nil {
-		return true, errors.Wrap(err, "getting number of projects")
+		return http.StatusInternalServerError, errors.Wrapf(err, "getting project '%s'", projectId)
 	}
-	allEnabledProjectsId := []string{}
-	for _, p := range allEnabledProjects {
-		allEnabledProjectsId = append(allEnabledProjectsId, p.Id)
+	if originalRef != nil {
+		if projectRef == nil {
+			// Should be unreachable
```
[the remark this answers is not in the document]
bynn: removed

model/project_ref.go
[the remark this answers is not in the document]
bynn: updated

model/project_ref.go
bynn: renamed to `ValidateEnabledProjectsLimit`

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 	return res, nil
 }
 
-// ValidateProjectCreation returns a boolean if you should surface the error or not.
-func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (bool, error) {
+// ValidateEnabledProjectsLimit takes in a merged project ref and validates if the current project is allowed to be enabled.
+// Returns a status code and error we are already at limit with enabled projects.
+func ValidateEnabledProjectsLimit(projectId string, config *evergreen.Settings, mergedRefToValidate ProjectRef) (int, error) {
```
ablack12, line 2265: Preference for mergedRefToValidate to be a pointer since that's better for passing -- we don't need to check if it's nil if that's not correct to pass in.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 	return res, nil
 }
 
-// ValidateProjectCreation returns a boolean if you should surface the error or not.
-func ValidateProjectCreation(projectId string, config *evergreen.Settings, projectRef *ProjectRef) (bool, error) {
+// ValidateEnabledProjectsLimit takes in a merged project ref and validates if the current project is allowed to be enabled.
+// Returns a status code and error we are already at limit with enabled projects.
```
ablack12, line 2264: **if we

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+func ValidateEnabledProjectsLimit(projectId string, config *evergreen.Settings, mergedRefToValidate ProjectRef) (int, error) {
 	if config.ProjectCreation.TotalProjectLimit == 0 || config.ProjectCreation.RepoProjectLimit == 0 {
```
ablack12, line 2266: If mergedRefToValidate is disabled then we should return early.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
-		return false, nil
+		return 0, nil
```
ablack12, line 2267: Comment mentions this should be a status code so should we return http.StatusOk?

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
 	}
-	catcher := grip.NewBasicCatcher()
-	allEnabledProjects, err := GetEnabledProjects()
+	originalRef, err := FindMergedProjectRef(projectId, "", false)
```
ablack12, line 2269: If mergedRefToValidate is disabled then we can return before querying the original project.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		return http.StatusInternalServerError, errors.Wrapf(err, "getting project '%s'", projectId)
 	}
 
-	// Only error if we are enabling a previously disabled project.
-	shouldError := false
-	pRef, err := findOneProjectRefQ(byId(projectId))
-	if err != nil {
-		return true, errors.Wrapf(err, "getting project '%s", projectId)
+	catcher := grip.NewBasicCatcher()
+	if originalRef == nil || !originalRef.IsEnabled() {
```
ablack12, line 2275: originalRef shouldn't be nil right? I'm not sure I understand this. Shouldn't we just error if it's nil?

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		allEnabledProjects, err := GetNumberOfEnabledProjects()
+		if err != nil {
+			return http.StatusInternalServerError, errors.Wrap(err, "getting number of projects")
+		}
+		if originalRef == nil {
+			allEnabledProjects++
+		}
+		// Only validate total project limit if it's a new project or was previously disabled.
```
ablack12, line 2283: Can you explain the "if it's a new project" part? Shouldn't we only validate if it's enabled period?

ablack12: This is also a confusing place for this comment since we don't check anything like that in this if statement.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+		if allEnabledProjects >= config.ProjectCreation.TotalProjectLimit {
+			catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
```
ablack12, line 2285: enabled project

rest/data/project.go
```diff
@@ -108,12 +109,19 @@ func CreateProject(ctx context.Context, env evergreen.Environment, projectRef *m
...
 	// Always warn because created projects are never enabled.
-	_, err := model.ValidateProjectCreation(projectRef.Id, env.Settings(), projectRef)
+	warningCatcher := grip.NewBasicCatcher()
+	statusCode, err := model.ValidateEnabledProjectsLimit(projectRef.Id, env.Settings(), *projectRef)
 	if err != nil {
-		// TODO EVG-18784: Return graphql warning
+		if statusCode != http.StatusBadRequest {
+			return false, gimlet.ErrorResponse{
+				StatusCode: statusCode,
+				Message:    errors.Wrapf(err, "inserting project '%s'", projectRef.Identifier).Error(),
```
ablack12, line 121: this message needs to be updated

model/project_ref.go
ablack12: oh is this the "created" case so we want to check and warn? Let's think of ways to refactor this to be clearer. We could assign a variable `isNewProject := originalRef == nil` and comment to explain this, and comment in the function header that if it's a new project, we validate as though it were enabled to see if that would be a problem.

Another idea is we could pass in a variable to the top-level that indicates if we're assuming it's enabled.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
-	pRef, err := findOneProjectRefQ(byId(projectId))
-	if err != nil {
-		return true, errors.Wrapf(err, "getting project '%s", projectId)
+	catcher := grip.NewBasicCatcher()
+	if originalRef == nil || !originalRef.IsEnabled() {
+		allEnabledProjects, err := GetNumberOfEnabledProjects()
+		if err != nil {
+			return http.StatusInternalServerError, errors.Wrap(err, "getting number of projects")
+		}
+		if originalRef == nil {
```
ablack12, line 2280: I would think we'd want to do this even if original ref is disabled, since the idea is that we should be enabling the given ref.

model/project_ref.go
```diff
@@ -2244,55 +2260,52 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+			catcher.Errorf("total project limit of %d reached", config.ProjectCreation.TotalProjectLimit)
+		}
 	}
-	if pRef != nil {
-		// Defaulting to repo
-		if projectRef == nil {
-			projectRef, err = GetProjectRefMergedWithRepo(*pRef)
+
+	// Only validate owner/repo limit if it's a new project, was previously disabled, or we're updating owner/repo.
+	if originalRef == nil || !originalRef.IsEnabled() || originalRef.Owner != mergedRefToValidate.Owner || originalRef.Repo != mergedRefToValidate.Repo {
```
ablack12, line 2290: this would be a good spot for a tiny private helper function shouldValidateOwnerRepoLimit or something, and can combine this if with the below

rest/data/project_settings.go
```diff
@@ -256,6 +260,15 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 			}
 		}
 
+		config, err := evergreen.GetConfig()
+		if err != nil {
+			return nil, errors.Wrap(err, "getting evergreen config")
+		}
+		_, err = model.ValidateEnabledProjectsLimit(projectId, config, *mergedSection)
```
ablack12, line 267: should validateEnabledProjectsLimit take in the original ref as well so we don't re-query mergedbeforeRef? (for the create case can pass in an empty project ref, or could refactor that function so there are two different functions that reuse a lot of the code)

ablack12 (changes requested):
Left a lot of comments but this is looking better! I think there are a few edge cases we need to add tests for and some room to refactor a bit more.

model/project_ref.go

graphql/mutation_resolver.go
```diff
@@ -408,16 +408,17 @@ func (r *mutationResolver) CreateProject(ctx context.Context, project restModel.
...
+				if apiErr.StatusCode == http.StatusBadRequest {
+					return nil, InputValidationError.Send(ctx, apiErr.Message)
+				}
 			}
-			// StatusNotFound and other error codes are really internal errors bc we determine this input
-			return nil, InternalServerError.Send(ctx, apiErr.Message)
+			return nil, InternalServerError.Send(ctx, err.Error())
 		}
-		return nil, InternalServerError.Send(ctx, err.Error())
+		graphql.AddError(ctx, PartialError.Send(ctx, err.Error()))
```
khelif96, line 421: Can you explain this logic? Why are we adding an error to the response?

bynn: this is from sophies comment 

~		// Use AddError to bypass gqlgen restriction that data and errors cannot be returned in the same response
		// https://github.com/99designs/gqlgen/issues/1191
~

model/project_ref.go
```diff
@@ -2244,55 +2266,57 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
-			projectRef, err = GetProjectRefMergedWithRepo(*pRef)
-			if err != nil {
-				return true, errors.Wrapf(err, "getting merged project for '%s'", projectId)
-			}
+	isNewProject := originalMergedRef == nil
+	catcher := grip.NewBasicCatcher()
+	if shouldValidateTotalProjectLimit(isNewProject, originalMergedRef) {
+		allEnabledProjects, err := GetNumberOfEnabledProjects()
+		if err != nil {
+			return http.StatusInternalServerError, errors.Wrap(err, "getting number of projects")
```
ablack12, line 2299: of enabled projects

model/project_ref.go
```diff
@@ -2244,55 +2266,57 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+	if shouldValidateOwnerRepoLimit(isNewProject, config, originalMergedRef, mergedRefToValidate) {
+		enabledOwnerRepoProjects, err := GetNumberOfEnabledProjectsForOwnerRepo(mergedRefToValidate.Owner, mergedRefToValidate.Repo)
 		if err != nil {
-			return true, errors.Wrapf(err, "getting number of projects for '%s/%s'", projectRef.Owner, projectRef.Repo)
+			return http.StatusInternalServerError, errors.Wrapf(err, "getting number of projects for '%s/%s'", mergedRefToValidate.Owner, mergedRefToValidate.Repo)
 		}
-		if len(enabledOwnerRepoProjects) >= config.ProjectCreation.RepoProjectLimit {
-			catcher.Errorf("owner repo limit of %d reached for '%s/%s'", config.ProjectCreation.RepoProjectLimit, projectRef.Owner, projectRef.Repo)
+		if enabledOwnerRepoProjects >= config.ProjectCreation.RepoProjectLimit {
+			catcher.Errorf("owner repo limit of %d reached for '%s/%s'", config.ProjectCreation.RepoProjectLimit, mergedRefToValidate.Owner, mergedRefToValidate.Repo)
```
ablack12, line 2312: since this error will go to users, should be a bit clearer: "enabled project limit" might be clearer than "owner repo limit" (we specify owner/repo at the end so that clarifies it i think)

model/project_ref.go
ablack12: looks like we handle this by only calling the function if the mergedRefToValidate is enabled

ablack12: Maybe we can tweak the first comment to something like "and validates project limits, assuming the given project is going to be enabled". Also isn't quite accurate since it says it takes in 1 project ref.

model/project_ref.go
```diff
@@ -2244,55 +2266,57 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
...
+// - the original project was disabled
+func shouldValidateTotalProjectLimit(isNewProject bool, originalMergedRef *ProjectRef) bool {
+	return isNewProject || !originalMergedRef.IsEnabled()
+}
+
+// shouldValidateOwnerRepoLimit will return true if:
+// - we are creating a new project
+// - the original project was disabled
+// - the owner or repo has changed
+// - the owner/repo is part of exception
```
ablack12, line 2280: if the owner repo is NOT an exception right? Maybe "will return true if the owner/repo is not exempt from validation and: <list>"

ablack12 (approved):
LGTM with nits

model/project_ref.go
bynn: yes!

model/project_ref.go
bynn: done

model/project_ref.go
bynn: done

model/project_ref.go
bynn: done

PR #6262, opened 2023-02-14
EVG-7029 Error when multi-word commit message is provided without quotes in CLI

hadjri:
[EVG-7029](https://jira.mongodb.org/browse/EVG-7029)

### Description 
Currently the `commit-message` flag available in the `enqueue-patch` CLI command does not provide any error that unused parameters have been passed in.  For example, running `enqueue-patch --commit-message hello world` will result in the commit message being just "hello".

Added a change to check if there are more than 1 arguments passed to the `commit-message` flag, and error if so.
### Testing 
Built CLI locally and confirmed that passing in a commit message phrase without quotes causes an error, and succeeds otherwise.
    
N/A

### DIFF
--- a/operations/before.go
+++ b/operations/before.go
@@ -62,6 +62,16 @@ var (
 		return c.Set(hostFlagName, host)
 	}
 
+	checkCommitMessageFlagFlag = func(c *cli.Context) error {
+		message := c.String(commitMessageFlag)
+		if message != "" {
+			if c.NArg() > 1 {
+				return errors.New("multiple arguments passed in for commit message, please consider using quotations")
+			}
+		}
+		return nil
+	}
+
 	requireProjectFlag = func(c *cli.Context) error {
 		if c.String(projectFlagName) == "" {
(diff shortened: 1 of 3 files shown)


operations/before.go
```diff
@@ -62,6 +62,16 @@ var (
 		return c.Set(hostFlagName, host)
 	}
 
+	checkCommitMessageFlagFlag = func(c *cli.Context) error {
```
bynn, line 65: nit: because it's not really a flag (i think) people use can we call it something like validateCommitMessage?

bynn (approved):

operations/before.go
malikchaya2: or `checkExtraArgs` would work as well. 
I'm fine with `checkCommitMessageFlag` because it matches with requirePatchIDFlag, but the extra Flag confuses me a bit.

malikchaya2 (approved):
approved with nit

PR #6265, opened 2023-02-14
EVG-18700: refactor functions in preparation for PatchedParserProject deprecation

Kimchelly:
Jira:
https://jira.mongodb.org/browse/EVG-18700

### Description 
This is half of a change preparing to deprecate the `(Patch).PatchedParserProject` field. because I didn't want to add too many changes to one PR, so this one focuses on refactoring the existing logic to make it less risky to do the follow-up change. There will be a follow-on PR to make the actual change.

The current purpose of `(Patch).PatchedParserProject` is to store the entire parser project as a YAML string before the patch is finalized. Once it's finalized, this YAML string is converted into the parser project and inserted into the DB.

The new behavior (after the follow-up PR) will be avoid putting the parser project in a string field and instead insert the parser project immediately when the patch is created, rather than wait until the patch is finalized. The new field `ProjectStorageMethod` will be set when the patch is created to indicate where to get the parser project from (either from the DB or from S3). It's not used for now (i.e. paresr projects are still inserted during patch finalization rather than patch creation), but will be used in the follow-up PR.

* Refactor parser project storage method so it's in the global `evergreen` constants instead of the `model` package. (This avoids some dependency cycles between the `patch` and `model` packages.)
* Refactor some existing methods to return slightly different values in preparation to insert the parser project at the same time as patch creation.
* Add a field for a patch to access the parser project from storage after the patch is created.
* Fix an issue where project identifier was not set in the DB for parser projects except those created via GitHub PRs. (Not related to the desired change, but I spotted this bug while unit testing)

### Testing 
* Updated existing unit tests and added new ones.
* Created and finalized a new CLI patch in staging.

No documentation needed.


model/patch/patch.go
```diff
@@ -582,24 +596,33 @@ func (p *Patch) FilesChanged() []string {
 	return filenames
 }
 
-// SetActivated sets the patch to activated in the db
+// SetActivated sets the patch to activated in the DB, which effectively
```
hadjri, line 599: Seems as though it does finalize the patch, can we reword to something like "marks the patch as finalized?" We also may want to change this function name now since it does more than simply marking activated.

model/project_parser.go
```diff
@@ -512,25 +512,43 @@ func (bvt *parserBV) hasSpecificActivation() bool {
...
+// a project. This assumes that the version may not exist yet (i.e. the patch is
+// unfinalized); otherwise FindAndTranslateProjectForVersion is equivalent.
 func FindAndTranslateProjectForPatch(ctx context.Context, settings *evergreen.Settings, p *patch.Patch) (*Project, *ParserProject, error) {
-	if p.PatchedParserProject == "" {
-		v, err := VersionFindOneId(p.Version)
+	if p.ProjectStorageMethod != "" {
+		pp, err := ParserProjectFindOneByID(ctx, settings, p.ProjectStorageMethod, p.Id.Hex())
 		if err != nil {
-			return nil, nil, errors.Wrapf(err, "finding version '%s' for patch '%s'", p.Version, p.Id.Hex())
+			return nil, nil, errors.Wrapf(err, "finding parser project '%s' in storage", p.Id.Hex())
```
hadjri, line 522: Can we add the `ProjectStorageMethod` to the formatted error output?

model/project_parser.go
```diff
@@ -512,25 +512,43 @@ func (bvt *parserBV) hasSpecificActivation() bool {
...
 		}
-		if v == nil {
-			return nil, nil, errors.Errorf("version '%s' not found for patch '%s'", p.Version, p.Id.Hex())
+		if pp == nil {
+			return nil, nil, errors.Errorf("parser project '%s' not found in storage", p.Id.Hex())
```
hadjri, line 525: Same here

model/patch_lifecycle.go
```diff
@@ -251,51 +226,156 @@ func GetPatchedProject(ctx context.Context, settings *evergreen.Settings, p *pat
...
+	if err != nil {
+		return "", errors.Wrap(err, "fetching project options for patch")
+	}
+
+	projectFileBytes, err := getPatchedProjectYAML(ctx, projectRef, opts, p)
+	if err != nil {
+		return "", errors.Wrap(err, "getting patched project file as YAML")
+	}
+
+	if !projectRef.IsVersionControlEnabled() {
```
hadjri, line 350: We can move this above the `getPatchedProjectYAML` call.

hadjri (changes requested):
No qualms with the general design 👍

model/patch/patch.go
Kimchelly: Yeah, I guess a finalized patch used to be called activated, but the terminology has drifted over time. I updated this + added a doc comment to activated to say it means finalized.

model/patch_lifecycle.go
Kimchelly: Nice catch, done.

model/project_parser.go
Kimchelly: Done.

model/project_parser.go
Kimchelly: Done.

hadjri (approved):

PR #6266, opened 2023-02-14
EVG-18744 add more host fields to termination logging

ybrill:
[EVG-18744](https://jira.mongodb.org/browse/EVG-18744)

### Description 
Add `started_by` to the `host successfully terminated` message so we can filter out non-`mci` started hosts when looking at idle time. Add `user_host` so we can differentiate between non-`mci` hosts that are started by users from those started by tasks.


ybrill:
evergreen retry

julianedwards (approved):

PR #6268, opened 2023-02-15
EVG-18974 Populate log url for tasks

khelif96:
[EVG-18974](https://jira.mongodb.org/browse/EVG-18974)

### Description 
Log URL was not populated for tasks fetched using `generateBuildVariants`


malikchaya2 (approved):
the logic looks solid, but I'd suggest trying to figure out a way to test it to make sure it doesn't accidentally break anything

PR #6270, opened 2023-02-16
EVG-7440 Do not call checks API on MarkEnd unless patch status changes

hadjri:
[EVG-7440](https://jira.mongodb.org/browse/EVG-7440)

### Description 
The [previous PR](https://github.com/evergreen-ci/evergreen/pull/6155) decoupled PR status updating from the `UpdatePatchStatus` function for understandability purposes, yet when the new `checkUpdateBuildPRStatus` was placed in `UpdateBuildAndVersionStatusForTask` it would only perform the PR status update if the task's [build status had changed](https://github.com/evergreen-ci/evergreen/pull/6155/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R1377), not if the patch status had changed.  This means that every time a build in a PR changed status (whether it was a restart, or simply `MarkEnd` calling) we were calling the GitHub checks API, rather than only when the overall PR changed status. 

Although I think this isn't necessarily wrong it caused an increase in GitHub checks API usage and made it more likely for us to breach our GitHub API budget.

Since the purpose of this ticket was to address UX during manual restarts of builds & versions, I've reverted the logic in `UpdateBuildAndVersionStatusForTask` [to only do a PR update when the patch status changes ](https://github.com/evergreen-ci/evergreen/pull/6270/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R1408)to reduce GitHub API usage. This is the only difference between this PR and the previous PR.
### Testing 
Tested in staging    
N/A

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -1127,6 +1127,36 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
 	return b.UpdateGithubCheckStatus(buildStatus.status)
 }
 
+// checkUpdateBuildPRStatus checks if the build is coming from a PR, and if so
+// sends its updated status to GitHub to reflect the status of the build.
+func checkUpdateBuildPRStatus(b *build.Build) error {
+	if !evergreen.IsGitHubPatchRequester(b.Requester) {
+		return nil
+	}
+	p, err := patch.FindOneId(b.Version)
+	if err != nil {
+		return errors.Wrapf(err, "finding patch '%s'", b.Version)
+	}
+	if p == nil {
+		return errors.Errorf("patch '%s' not found", b.Version)
+	}
+	if p.IsGithubPRPatch() && !evergreen.IsFinishedPatchStatus(p.Status) {
+		input := thirdparty.SendGithubStatusInput{
+			VersionId: p.Id.Hex(),
+			Owner:     p.GithubPatchData.BaseOwner,
+			Repo:      p.GithubPatchData.BaseRepo,
+			Ref:       p.GithubPatchData.HeadHash,
+			Desc:      "patch status change",
+			Caller:    "pr-task-reset",
+			Context:   fmt.Sprintf("evergreen/%s", b.BuildVariant),
+		}
+		if err = thirdparty.SendVersionStatusToGithub(input); err != nil {
+			return errors.Wrapf(err, "sending patch '%s' status to Github", p.Id.Hex())
+		}
+	}
+	return nil
+}
+
 // updateBuildStatus updates the status of the build based on its tasks' statuses
 // Returns true if the build's status has changed or if all of the build's tasks become blocked.
 func updateBuildStatus(b *build.Build) (bool, error) {
@@ -1301,7 +1331,9 @@ func updateVersionStatus(v *Version) (string, error) {
 	return versionStatus, nil
 }
 
-func UpdatePatchStatus(p *patch.Patch, versionStatus, buildVariant string) error {
+// UpdatePatchStatus updates the status of a patch. For PR patches, the build associated
+// with the passed in buildVariant has its GitHub status updated.
+func UpdatePatchStatus(p *patch.Patch, versionStatus string) error {
 	patchStatus, err := evergreen.VersionStatusToPatchStatus(versionStatus)
 	if err != nil {
 		return errors.Wrapf(err, "getting patch status from version status '%s'", versionStatus)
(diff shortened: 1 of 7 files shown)


model/task_lifecycle.go
```diff
@@ -1301,7 +1331,9 @@ func updateVersionStatus(v *Version) (string, error) {
 	return versionStatus, nil
 }
 
-func UpdatePatchStatus(p *patch.Patch, versionStatus, buildVariant string) error {
+// UpdatePatchStatus updates the status of a patch. For PR patches, the build associated
+// with the passed in buildVariant has its GitHub status updated.
```
Kimchelly, line 1335: It seems like it doesn't update the PR patch anymore. Should the second sentence be for `UpdateVersionAndPatchStatusForBuilds`?

model/task_lifecycle.go
```diff
@@ -1127,6 +1127,36 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
 	return b.UpdateGithubCheckStatus(buildStatus.status)
 }
 
+// checkUpdateBuildPRStatus checks if the build is coming from a PR, and if so
+// sends its updated status to GitHub to reflect the status of the build.
```
Kimchelly, line 1131: It seems like this function is specifically restricted to resetting the state back to pending, rather than updating the state arbitrarily. Can we update this doc comment to clarify that?

model/task_lifecycle.go
```diff
@@ -1127,6 +1127,36 @@ func updateBuildGithubStatus(b *build.Build, buildTasks []task.Task) error {
...
+func checkUpdateBuildPRStatus(b *build.Build) error {
+	if !evergreen.IsGitHubPatchRequester(b.Requester) {
+		return nil
+	}
```
Kimchelly, line 1135: Is it worth having a check to verify that `b.Status` is an unfinished status, or is checking the patch status sufficient?

model/task_lifecycle.go
```diff
@@ -1472,7 +1498,7 @@ func UpdateVersionAndPatchStatusForBuilds(buildIds []string) error {
 			if p == nil {
 				return errors.Errorf("no patch found for version '%s'", buildVersion.Id)
 			}
-			if err = UpdatePatchStatus(p, newVersionStatus, bvMap[buildId]); err != nil {
+			if err = UpdatePatchStatus(p, newVersionStatus); err != nil {
```
Kimchelly, line 1501: Is this supposed to happen after `checkUpdateBuildPRStatus`? I would have thought the patch would have to be updated first so that we can resolve the patch's latest status before we check the patch status in `checkUpdateBuildPRStatus`.

Kimchelly (changes requested):

model/task_lifecycle.go
hadjri: Good catch. This is a symptom of moving around the logic `UpdatePatchStatus`, which had this issue in the first place. We want to be updating build statuses independently of the patch status changes. Changed the check at L1143 to `p.IsGithubPRPatch() && !evergreen.IsFinishedBuildStatus(b.Status)`

model/task_lifecycle.go
hadjri: Yes, moved to `UpdateVersionAndPatchStatusForBuilds` and updated the wording.

model/task_lifecycle.go
hadjri: With the latest fix to only update, we no longer check the patch status in  `checkUpdateBuildPRStatus ` so this is fine to happen after that function.

model/task_lifecycle_test.go
```diff
@@ -1279,6 +1282,84 @@ func TestUpdateBuildStatusForTask(t *testing.T) {
...
+	}
+
+	assert.NoError(t, b.Insert())
+	assert.NoError(t, p.Insert())
+	assert.NoError(t, v.Insert())
+	assert.NoError(t, testTask.Insert())
+	assert.NoError(t, anotherTask.Insert())
+
+	assert.NoError(t, UpdateVersionAndPatchStatusForBuilds([]string{b.Id}))
+	// Confirm that the build status was updated to pending in the PR by checking for appropriate Splunk logs
```
Kimchelly, line 1333: We don't need to specify that this is Splunk since we could change out logging backend at any point. I think it's sufficient to say something like "confirm if there was a log indicating the build status updated to pending".

Kimchelly (approved):
LGTM mod trivial test comment change.

PR #6271, opened 2023-02-16
EVG-18931 Move RespErrorf from utility to util

bynn:

### Description 

### Testing 
    
  < if yes then where >


Kimchelly (approved):

Kimchelly:
We had a conversation in Slack about this, which is a small refactor of utility so that it doesn't import other evergreen-ci repos, as those dependencies are now causing some dependency cycle headaches that are preventing the actual desired change in EVG-18931.

PR #6272, opened 2023-02-17
EVG-16701 Allow bulk version priority setting

hadjri:
[EVG-16701](https://jira.mongodb.org/browse/EVG-16701)

### Description 
Same changes as the [previous](https://github.com/evergreen-ci/evergreen/pull/6245) PR, but removes the requirement for specifying `revision_start` or `start_time` in the request.

This is the [3-line block](https://github.com/evergreen-ci/evergreen/pull/6245/files#diff-e16fb0855e8abc1277b2fa7b08b94806183d1bef59d4526cbe92a325b1e661a4R1044) that has been removed.
### Testing 
Added a new unit test for `getProjectVersionsHandler` with more cases.  Tested the route manually in staging via URL.
Updated the [modify project versions](https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#modify-versions-for-a-project) docs, as well as the [get project versions](https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#get-versions-for-a-project) route.


ablack12 (changes requested):
What's the context around why this isn't required? Seems like this route shouldn't just be able to disable versions willy nilly.

Also small nit about the PR title: it references the new change you've added but really the original title is more accurate, since this commit still adds the new route, just with a few changes.

hadjri:
> What's the context around why this isn't required? Seems like this route shouldn't just be able to disable versions willy nilly.
> 
> Also small nit about the PR title: it references the new change you've added but really the original title is more accurate, since this commit still adds the new route, just with a few changes.

The "get project versions" route and this new route were sharing some of the same request parsing rules.  Since we added a requirement to have time or revision number, calling the "get project versions" via URL with nothing specified now returned an error saying to add these params to your request.

Currently this modify route will disable the last 20 versions if no revision or time params are specified, just as the "get project versions" route will get the last 20 revisions if no params are specified. In my mind this was for consistency across routes but I'm fine to re-add the params requirement specifically to the modify route if we think this is too inexplicit.

ablack12:
Yeah I think since this is an update and not a query it's more important to make sure we don't forget to include needed parameters so we don't accidentally do something silly.

ablack12 (commented):
(didn't add that comment as a review so adding this comment so it registers)

hadjri:
> (didn't add that comment as a review so adding this comment so it registers)

Ah I already included that change in the initial re-request review, sorry if my message didn't make that clear.

ablack12 (approved):
Oh thanks i missed that!

PR #6274, opened 2023-02-17
EVG-18812 Force the commit queue to respect branch protection rules 

malikchaya2:
[EVG-18812](https://jira.mongodb.org/browse/EVG-18812)

### Description 
This will block a pr from getting enqueued to the commit queue if branch protection rules are not met. 

### Testing 
image

yes: [EVG-19096](https://jira.mongodb.org/browse/EVG-19096)

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -320,13 +320,13 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 
 	if !thirdparty.IsUnblockedGithubStatus(pr.GetMergeableState()) {
 		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", pr.GetMergeableState())
-		grip.Error(message.Fields{
+		grip.Debug(message.Fields{
 			"message":        errMsg,
-			"state":          *pr.MergeableState,
+			"state":          pr.GetMergeableState(),
 			"owner":          userRepo.Owner,
 			"repo":           userRepo.Repo,
-			"pr_title":       *pr.Title,
-			"commit_message": *pr.Number,
+			"pr_title":       pr.GetTitle(),
+			"commit_message": pr.GetNumber(),
 		})
 		sendErr := thirdparty.SendCommitQueueGithubStatus(env, pr, message.GithubStateFailure, errMsg, "")
 
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -319,7 +319,7 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 	}
 
 	if !thirdparty.IsUnblockedGithubStatus(pr.GetMergeableState()) {
-		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", *pr.MergeableState)
+		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", pr.GetMergeableState())
 		grip.Error(message.Fields{
 			"message":        errMsg,
 			"state":          *pr.MergeableState,
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -318,7 +318,7 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 		return nil, errors.New("PR contains no base branch label")
 	}
 
-	if !thirdparty.IsUnblockedGithubStatus(*pr.MergeableState) {
+	if !thirdparty.IsUnblockedGithubStatus(pr.GetMergeableState()) {
 		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", *pr.MergeableState)
 		grip.Error(message.Fields{
 			"message":        errMsg,
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -318,6 +318,28 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 		return nil, errors.New("PR contains no base branch label")
 	}
 
+	if !thirdparty.IsUnblockedGithubStatus(*pr.MergeableState) {
+		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", *pr.MergeableState)
+		grip.Error(message.Fields{
+			"message":        errMsg,
+			"state":          *pr.MergeableState,
+			"owner":          userRepo.Owner,
+			"repo":           userRepo.Repo,
+			"pr_title":       *pr.Title,
+			"commit_message": *pr.Number,
+		})
+		sendErr := thirdparty.SendCommitQueueGithubStatus(env, pr, message.GithubStateFailure, errMsg, "")
+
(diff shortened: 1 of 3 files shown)


rest/data/commit_queue.go
```diff
@@ -318,6 +318,28 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
...
+			"message":        errMsg,
+			"state":          *pr.MergeableState,
+			"owner":          userRepo.Owner,
+			"repo":           userRepo.Repo,
+			"pr_title":       *pr.Title,
+			"commit_message": *pr.Number,
+		})
+		sendErr := thirdparty.SendCommitQueueGithubStatus(env, pr, message.GithubStateFailure, errMsg, "")
+
+		grip.Error(message.WrapError(sendErr, message.Fields{
```
malikchaya2, line 333: I will add an alert for when we see a blocked PR whose status is not "blocked" so that we can assess it and add it to the list so that we can let it through.

thirdparty/github.go
```diff
@@ -36,6 +36,32 @@ const (
...
+	githubPrClean,
+	githubPrDirty,
+	githubPrDraft,
+	githubPrHas_Hooks,
+	githubPrUnknown,
+	githubPrUnstable,
+}
+
+const (
+	// all PR statuses except for "blocked" based on statuses listed here:
```
ablack12, line 50: Why is that we aren't just checking if it == blocked?

rest/data/commit_queue.go
```diff
@@ -318,6 +318,28 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 		return nil, errors.New("PR contains no base branch label")
 	}
 
+	if !thirdparty.IsUnblockedGithubStatus(*pr.MergeableState) {
```
ablack12, line 321: can we use GetMergeableState() to avoid nil pointer issues

rest/data/commit_queue.go
```diff
@@ -318,6 +318,28 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
...
+		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", *pr.MergeableState)
+		grip.Error(message.Fields{
```
ablack12, line 323: I don't think we should log this at the error level, since it's not an evergreen error but rather information to help debug for users.

rest/data/commit_queue.go
ablack12: Do you mean if the state isn't one of the ones we have in the list but also isn't equal to blocked? I'm not sure I see where we have the information for this alert.

ablack12 (changes requested):

thirdparty/github.go
```diff
@@ -36,6 +36,32 @@ const (
...
+	githubPrBehind    = "behind"
+	githubPrClean     = "clean"
+	githubPrDirty     = "dirty"
+	githubPrDraft     = "draft"
+	githubPrHas_Hooks = "has_hooks"
+	githubPrUnknown   = "unknown"
+	githubPrUnstable  = "unstable"
+)
+
+func IsUnblockedGithubStatus(status string) bool {
```
ablack12, line 61: We need a function header here

thirdparty/github.go
malikchaya2: Because the field is not documented so it can change without notice. 

If we let through anything that is not blocked, we risk it failing silently without us knowing if the field changes. With this setup, it needs to be one of the fields listed to go through, so we are less likely to merge something that shouldn't be merged.

rest/data/commit_queue.go
malikchaya2: my bad, I meant to add this comment to line 323.

malikchaya2: yes, if something is not equal to blocked but we are blocking on it, we should investigate what the status is to see if we want to add it to the list of statuses to let through, or if "blocked" was changed to something else.

rest/data/commit_queue.go
malikchaya2: done

rest/data/commit_queue.go
malikchaya2: done

thirdparty/github.go
malikchaya2: done

rest/data/commit_queue.go
```diff
@@ -318,6 +318,28 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 		return nil, errors.New("PR contains no base branch label")
 	}
 
+	if !thirdparty.IsUnblockedGithubStatus(pr.GetMergeableState()) {
```
ablack12, line 321: I think we should apply this change (GetMergeableState, getWhatever) to the other calls in this function as well.

rest/data/commit_queue.go
ablack12: I think this change may have been left out

rest/data/commit_queue.go
ablack12: That makes sense. Can we update the rotation playbook to handle this alert?

thirdparty/github.go
ablack12: can we add this rational to the function comment?

ablack12 (changes requested):
I noticed that you opened a separate ticket for documenting this change; I'm slightly against that workflow generally, since I don't think we should close tickets that users are watching without the feature being ready, and documentation is part of that. Additionally user facing documentation for a single ticket shouldn't take very long to add such that it needs to be a separate ticket; I think this makes more sense in the case of projects where we have many things to document and/or don't want to document until all the work is done, but in that case the user is watching the overall project.

thirdparty/github.go
malikchaya2: done

malikchaya2:
> I noticed that you opened a separate ticket for documenting this change; I'm slightly against that workflow generally, since I don't think we should close tickets that users are watching without the feature being ready, and documentation is part of that. Additionally user facing documentation for a single ticket shouldn't take very long to add such that it needs to be a separate ticket; I think this makes more sense in the case of projects where we have many things to document and/or don't want to document until all the work is done, but in that case the user is watching the overall project. A separate ticket for the alert seems more relevant since that isn't user-facing and is more an improvement, but this still isn't likely to take very long.

Sounds good, I'll keep 18812 open until that work is done, and convert the documentation/alert ticket to 0 points.

thirdparty/github.go
```diff
@@ -36,6 +36,37 @@ const (
...
+	githubPrHas_Hooks,
+	githubPrUnknown,
+	githubPrUnstable,
+}
+
+const (
+	// All PR statuses except for "blocked" based on statuses listed here:
+	// https://docs.github.com/en/graphql/reference/enums#mergestatestatus
+	// Because the pr.MergeableState is not documented, it can change without
+	// notice. that's why we want to only allow fields we know to be unblocked
```
ablack12, line 53: nit: That's

ablack12 (approved):

PR #6276, opened 2023-02-17
EVG-18700: insert parser project document during patch creation

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-18700

### Description 
This is a follow-up to #6265 to insert the parser project upon patch creation for all patches that go through the patch intent job, which removes the need to temporarily store the parser project in the patch document as a string. The only kind of patch that _doesn't_ go through the patch intent job are commit queue items. To reduce the riskiness and size of this PR, I'm going to make a follow-up PR to deal with commit queue items after this one is deployed.

* Insert parser projects when patches are converted from intent patches into real patches.
* Modify patch finalization to support both old patches that stored the `PatchedParserProject` string and new patches that have already stored the parser project in the DB during patch creation.
* Leave some TODOs for next improvements to wrap up the work to remove `PatchedParserProject`.

### Testing 
* Updated existing tests and added some new ones for extra coverage.
* Tested in staging:
    * Pre-created a CLI patch using `PatchedParserProject` and post-deploy, verified that it still finalized correctly.
    * Created a CLI patch post-deploy and finalized it. I also diffed the pre-deploy patch parser project and post-deploy patch parser project to confirm that the parser project itself is the same.
    * Ran a patch with a trigger child patch to verify that child patches could be created and finalize.

N/A


model/patch_lifecycle.go
```diff
@@ -491,11 +492,29 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
-			p.Githash)
+
+	// It used to be that the parser project was not stored until patches were
+	// finalized. Newer patches always store the parser project when the patch
+	// is created rather than when it's finalized. For backward compatibility
+	// with existing unfinalized patches, the parser project must be inserted
+	// now if an old patch is being finalized.
+	mustInsertParserProjectDuringFinalization := p.PatchedParserProject != ""
+	var intermediateProject *ParserProject
+	// TODO (EVG-18700): this if-else most likely is equivalent to just calling
```
malikchaya2, line 503: At what point will it be safe to remove this logic? Is the plan to keep it around for a bit of time for the case where a user created a patch and then comes back months later to finalize it? 
Or is this only meant to stay in prod just to bridge the gap of patches created before the deploy and finalized after?

malikchaya2 (changes requested):
Mostly looks good, I just had two small questions.

model/patch_lifecycle.go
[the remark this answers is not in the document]
Kimchelly: I'm removing it in the follow-up PR. I'm waiting for a deploy to verify that this change works, then going to remove the if from the if-else.

model/project_parser_storage.go
```diff
@@ -77,3 +78,40 @@ func ParserProjectUpsertOne(ctx context.Context, settings *evergreen.Settings, m
...
+		return method, err
+	}
+
+	if !db.IsDocumentLimit(err) {
+		return method, err
+	}
+
+	// Upsertion could fail for multiple reasons, but Evergreen only resorts to
+	// using S3 if upsert failed due to the DB document size limitation.
+
```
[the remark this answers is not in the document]
Kimchelly: I intentionally made it a floating comment because it's meant to apply to everything below this line. I'm just gonna remove it since the doc comment already covers when it would fall back to S3.

rest/model/patch.go
```diff
@@ -41,6 +41,7 @@ type APIPatch struct {
 	GithubPatchData         githubPatch          `json:"github_patch_data,omitempty"`
 	ModuleCodeChanges       []APIModulePatch     `json:"module_code_changes"`
 	Parameters              []APIParameter       `json:"parameters"`
+	ProjectStorageMethod    *string              `json:"project_storage_method,omitempty"`
```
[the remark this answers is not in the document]
Kimchelly: I don't think it's that important to manually test S3 for this change since the fallback logic will be tested in EVG-18701 and the S3 fallback logic has already been tested in other integration tests. It's also not a prod risk because we're not enabling S3 support yet and it would be extremely difficult to have an initial project YAML for a patch (before any generate.tasks) that is greater than 16 MB (except for child patches, for which its S3 behavior will be tested tested in EVG-18701).

Basically, as long as the existing DB parser project logic still works, testing the S3 logic doesn't matter for this change.

model/patch_lifecycle.go
```diff
@@ -506,7 +525,9 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 				p.Githash)
 		}
 	}
-	intermediateProject.Id = p.Id.Hex()
+	if mustInsertParserProjectDuringFinalization {
```
hadjri, line 528: Can this be moved into the block on L703? `if mustInsertParserProjectDuringFinalization {` is used multiple times in this function and it seems they can be consolidated so it's more clear what the function of the flag is.

model/patch_lifecycle.go
```diff
@@ -554,27 +575,28 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
...
+		ParentPatchNumber:   parentPatchNumber,
+		Branch:              projectRef.Branch,
+		RevisionOrderNumber: p.PatchNumber,
+		AuthorID:            p.Author,
+		Parameters:          params,
+		Activated:           utility.TruePtr(),
+		AuthorEmail:         authorEmail,
+	}
+	if mustInsertParserProjectDuringFinalization {
+		intermediateProject.CreateTime = patchVersion.CreateTime
```
hadjri, line 598: Same comment here, I think this could be moved to L703

model/project_parser_storage.go
```diff
@@ -77,3 +78,37 @@ func ParserProjectUpsertOne(ctx context.Context, settings *evergreen.Settings, m
...
+// DB document size limitations, it will attempt to fall back to using S3
+// to store it. If it succeeds, this returns the actual project storage method
+// used to persist the parser project; otherwise, it returns the
+// originally-requested storage method.
+func ParserProjectUpsertOneWithS3Fallback(ctx context.Context, settings *evergreen.Settings, method evergreen.ParserProjectStorageMethod, pp *ParserProject) (evergreen.ParserProjectStorageMethod, error) {
+	err := ParserProjectUpsertOne(ctx, settings, method, pp)
+	if err == nil {
+		return method, nil
+	}
+	if method == evergreen.ProjectStorageMethodS3 {
```
hadjri, line 93: Can we combine these two statements that both return `method, err`?

model/project_parser_storage.go
```diff
@@ -77,3 +78,37 @@ func ParserProjectUpsertOne(ctx context.Context, settings *evergreen.Settings, m
...
+	flags, flagErr := evergreen.GetServiceFlags()
+	if flagErr != nil {
+		return method, errors.Wrap(flagErr, "getting service flags to check ability to fall back to S3")
+	}
+	if flags.ParserProjectS3StorageDisabled {
+		return method, err
+	}
+
+	if err := ParserProjectUpsertOne(ctx, settings, evergreen.ProjectStorageMethodS3, pp); err != nil {
+		return method, errors.Wrap(err, "upserting parser project into S3")
```
hadjri, line 110: We should either return plain `err` here, or return a wrapped error above on L94 and L98

hadjri (changes requested):

model/patch_lifecycle.go
Kimchelly: Done.

model/patch_lifecycle.go
Kimchelly: Done.

model/project_parser_storage.go
Kimchelly: Sure.

model/project_parser_storage.go
Kimchelly: I wrapped the other errors so it's a little more obvious which one it chose.

hadjri (approved):

PR #6278, opened 2023-02-22
Trim trailing whitespace from PR template

Kimchelly:
This is trivial, but the random whitespaces were bothering me.

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,10 +1,10 @@
 [EVG-<number>](https://jira.mongodb.org/browse/EVG-<number>)
 
-### Description 
+### Description
 < add description, context, thought process, etc >
 
-### Testing 
+### Testing
   < add a description of how you tested it >
-    
+
 #### Does this need documentation?
-  < if yes then where > 
+  < if yes then where >


docs/pull_request_template.md
```diff
@@ -1,10 +1,10 @@
 [EVG-<number>](https://jira.mongodb.org/browse/EVG-<number>)
```
khelif96, line 1: Can we please update this as well. The spruce one auto linkifies the ticket
https://github.com/evergreen-ci/spruce/blob/main/.github/pull_request_template.md

docs/pull_request_template.md
```diff
@@ -1,10 +1,10 @@
...
 
-### Description 
+### Description
 < add description, context, thought process, etc >
```
khelif96, line 4: Can these be replaced with comments so if someone forgets to change these they don't show up

malikchaya2: controversial take: I think that it's good if forgetting to update it obvious. That way it's easy to spot and fix.

Kimchelly: I don't have strong feelings one way or the other here. I'm inclined to leave it as is since I'm not the one who came up with the template.

docs/pull_request_template.md
Kimchelly: Same thing as the other comment, I don't really want to change the PR template meaningfully here.

ablack12 (approved):

PR #6281, opened 2023-02-22
EVG-19011: Paginated Query for Recent Pod Events 

SupaJoon:
[EVG-19011](https://jira.mongodb.org/browse/EVG-19011)

### Description 
This query will be utilized in the Pod GQL resolver to return recent pod events for the Container page in Spruce. I repurposed an existing function to support Host and Pod events.


khelif96 (approved):
Changes make sense to me although we should wait for an app review!

Kimchelly (approved):

PR #6284, opened 2023-02-22
EVG-18732 surface if a patch isn't created bc all files are ignored

ablack12:
EVG-18732

### Description
Right now we log if a patch isn't created bc files are ignored but we don't tell the user; this is really confusing and often means we need to look at splunk logs to tell the user what happened. We should instead return a successful status with a message.
Cleaned up some other stuff as I went.

### Testing
Tested with a PR https://github.com/evergreen-ci/commit-queue-sandbox/pull/448
image

Document that no patch will be created for a PR containing only ignored file changes.

### DIFF
--- a/units/github_status_api.go
+++ b/units/github_status_api.go
@@ -25,6 +25,7 @@ const (
 	githubStatusUpdateJobName = "github-status-update"
 
 	githubUpdateTypeNewPatch              = "new-patch"
+	githubUpdateTypeMessage               = "send-message"
 	githubUpdateTypeRequestAuth           = "request-auth"
 	githubUpdateTypePushToCommitQueue     = "commit-queue-push"
 	githubUpdateTypeDeleteFromCommitQueue = "commit-queue-delete"
@@ -80,6 +81,19 @@ func makeGithubStatusUpdateJob() *githubStatusUpdateJob {
 	}
 	j.SetPriority(1)
 	return j
+}
+
+// NewGithubStatusUpdateJobWithMessage creates a job to send a passing status to Github with a message.
+func NewGithubStatusUpdateJobWithMessage(githubContext, owner, repo, ref, description string) amboy.Job {
+	job := makeGithubStatusUpdateJob()
+	job.GithubContext = githubContext
+	job.UpdateType = githubUpdateTypeMessage
+	job.Owner = owner
+	job.Repo = repo
+	job.Ref = ref
+	job.Description = description
+	job.SetID(fmt.Sprintf("%s:%s-%s", githubStatusUpdateJobName, job.UpdateType, time.Now().String()))
+	return job
 }
 
 // NewGithubStatusUpdateJobForNewPatch creates a job to update github's API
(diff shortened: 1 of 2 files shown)


Kimchelly (commented):
1. Would returning an error here cause an issue for projects that block merging because they require all checks to pass?

2. The [ignore docs](https://github.com/evergreen-ci/evergreen/wiki/Project-Configuration-Files#ignoring-changes-to-certain-files) don't say anything about the expected interaction between an open PR with all ignored changes and the patch. Can we document that no patch will be created for a PR containing only ignored file changes?

ablack12:
Really good thought about the error causing an issue. I'll rethink this a bit.

Kimchelly:
Could it return a success check mark instead of nothing/error? Seems like if there's no tests to run, it trivially passes the Evergreen check.

ablack12:
Updated example: https://github.com/evergreen-ci/commit-queue-sandbox/pull/448

units/patch_intent.go
```diff
@@ -1026,3 +1016,17 @@ func (j *patchIntentProcessor) sendGitHubErrorStatus(patchDoc *patch.Patch) {
...
+// sendGitHubStatusMessage sends a successful status to Github with the given message.
+func (j *patchIntentProcessor) sendGitHubStatusMessage(patchDoc *patch.Patch, msg string) {
+	update := NewGithubStatusUpdateJobWithMessage(
+		evergreenContext,
+		patchDoc.GithubPatchData.BaseOwner,
+		patchDoc.GithubPatchData.BaseRepo,
+		patchDoc.GithubPatchData.HeadHash,
+		msg,
+	)
+	update.Run(nil)
```
Kimchelly, line 1029: The [context docs](https://pkg.go.dev/context@go1.20.1#pkg-overview) say to never pass a nil context because it has the potential for bad things to happen, especially as future-proofing if one day the context does get used. Can we just pass the `patchIntentProcessor` job's context into this?

units/patch_intent.go
```diff
@@ -1026,3 +1016,17 @@ func (j *patchIntentProcessor) sendGitHubErrorStatus(patchDoc *patch.Patch) {
 
 	j.AddError(update.Error())
 }
+
+// sendGitHubStatusMessage sends a successful status to Github with the given message.
+func (j *patchIntentProcessor) sendGitHubStatusMessage(patchDoc *patch.Patch, msg string) {
```
Kimchelly, line 1021: Can we call this `sendGitHubSuccessMessage` or something, since it only means success?

units/github_status_api.go
```diff
@@ -25,6 +25,7 @@ const (
 	githubStatusUpdateJobName = "github-status-update"
 
 	githubUpdateTypeNewPatch              = "new-patch"
+	githubUpdateTypeMessage               = "send-message"
```
Kimchelly, line 28: Rather than call this just message, it seems like it's specifically encoding success + a message, can we call this `githubUpdateTypeSuccessMessage` or something?

Kimchelly (changes requested):

Kimchelly (approved):

PR #6290, opened 2023-02-24
EVG-18700: remove PatchedParserProject for commit queue items and repotracker

Kimchelly:
EVG-18700

### Description
This is a follow-up to #6276, which removed the dependency on storing the entire parser project configuration in the patch doc before finalization by inserting the parser project when the patch is created. Commit queue items follow a slightly different process where they create a patch (and enqueue it in the commit queue), then finalize it when it's being processed by the queue. The repotracker also follows a special process where it doesn't make a patch at all, and just skips straight to making a version for any commits it picks up.

* For GitHub commit queue items, create the parser project when the patch is created and is being enqueued in the commit queue.
* For CLI commit queue items, create the parser project when the patch is dequeued from the commit queue for processing.
* For repotracker versions, insert the parser project when creating the version.
* Add a few helpers for some common copy-pasted parser project initialization.

### Testing
* Added unit tests in all places that didn't have some GitHub integration logic.
* Tested in staging commit queue sandbox by enqueueing a CLI patch and a GitHub patch via the UI, then checking that the repotracker correctly made versions for the waterfall once they were committed.

N/A

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -694,8 +694,7 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		ppStorageMethod = evergreen.ProjectStorageMethodDB
 	}
 	if mustInsertParserProjectDuringFinalization {
-		intermediateProject.Id = p.Id.Hex()
-		intermediateProject.CreateTime = patchVersion.CreateTime
+		intermediateProject.Init(p.Id.Hex(), patchVersion.CreateTime)
 		ppStorageMethod, err = ParserProjectUpsertOneWithS3Fallback(ctx, settings, ppStorageMethod, intermediateProject)
 		if err != nil {
 			return nil, errors.Wrapf(err, "upserting parser project for patch '%s'", p.Id.Hex())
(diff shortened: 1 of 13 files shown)


model/patch_lifecycle.go
```diff
@@ -1205,6 +1206,14 @@ func restartDiffItem(p patch.Patch, cq *commitqueue.CommitQueue) error {
 		PatchNumber:     patchNumber,
 	}
 
+	// The parser project is typically inserted at the same time as the patch.
```
malikchaya2, line 1209: was duplicating this comment intentional?

repotracker/repotracker.go
```diff
@@ -647,11 +646,14 @@ func CreateVersionFromConfig(ctx context.Context, projectInfo *model.ProjectInfo
 			v.Errors = append(v.Errors, versionErrs.Errors...)
 		}
 		if len(v.Errors) > 0 {
-			if err = v.Insert(); err != nil {
-				return nil, errors.Wrap(err, "error inserting version")
+			env := evergreen.GetEnvironment()
```
malikchaya2, line 649: nit: would it make sense to pass down the env from the repotrackerJob like we do with the context?

malikchaya2 (approved):
Approved with nits

model/patch_lifecycle.go
Kimchelly: Hmm, it applies to two different use cases in the code (one for CLI merge patches, one for admin commit queue restarts), so I'd say it's equally important in both places.

repotracker/repotracker.go
Kimchelly: I would love to do that, but unfortunately, `CreateVersionFromConfig` is used in a lot of places, of which some of them are really deep in the task lifecycle, so propagating the environment down from the top level for all of them would be a large amount of of refactoring of task lifecycle functions.

malikchaya2: that makes sense!

model/patch_lifecycle.go
malikchaya2: sounds good!

PR #6291, opened 2023-02-24
EVG-16349 add github comment to refresh status

ablack12:
EVG-16349

### Description
Add `github status` to allow for refreshing checks. I'm torn between status, checks, and refresh for the language here. Any preferences?

Refactored a number of trigger things so they could be used in units.

### Testing
Tested a lot on staging. 
Created to test general functionality https://github.com/evergreen-ci/commit-queue-sandbox/pull/448
and https://github.com/evergreen-ci/commit-queue-sandbox/pull/449 to test child patches.
Added &refreshed_status=true to these to make it clearer that they were just generated.
(Build links aren't correct here but that was fixed in cleanup and isn't a huge deal either way)

Yes -- I don't think we have a documentation page about evergreen github commands; if not I'll add one, otherwise i'll add to it.

### DIFF
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -348,7 +348,7 @@ func (b *Build) GetFinishedNotificationDescription(tasks []task.Task) string {
 		case t.Status == evergreen.TaskFailed:
 			failed++
 
-		case statusIsSystemFailure(t.Status):
+		case evergreen.IsSystemFailedTaskStatus(t.Status):
 			systemError++
 
 		case utility.StringSliceContains(evergreen.TaskUncompletedStatuses, t.Status):
@@ -379,16 +379,6 @@ func (b *Build) GetFinishedNotificationDescription(tasks []task.Task) string {
 	}
 
 	return b.appendTime(desc)
-}
-
-func statusIsSystemFailure(status string) bool {
-	systemFailures := []string{evergreen.TaskSystemFailed,
-		evergreen.TaskTimedOut,
-		evergreen.TaskSystemUnresponse,
-		evergreen.TaskSystemTimedOut,
-		evergreen.TaskTestTimedOut,
-	}
-	return utility.StringSliceContains(systemFailures, status)
 }
 
 func taskStatusSubformat(n int, verb string) string {
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -3,6 +3,7 @@ package build
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
@@ -296,6 +297,14 @@ func (b *Build) GetTimeSpent() (time.Duration, time.Duration, error) {
 
 	timeTaken, makespan := task.GetTimeSpent(tasks)
 	return timeTaken, makespan, nil
+}
+
+func (b *Build) GetURL(uiBase string) string {
+	url := fmt.Sprintf("%s/build/%s", uiBase, url.PathEscape(b.Id))
+	if evergreen.IsPatchRequester(b.Requester) {
+		url += "?redirect_spruce_users=true"
+	}
+	return url
 }
 
 // Insert writes the b to the db.
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -299,6 +299,7 @@ func (b *Build) GetTimeSpent() (time.Duration, time.Duration, error) {
 	return timeTaken, makespan, nil
 }
 
+// GetURL returns a url to the build page.
 func (b *Build) GetURL(uiBase string) string {
 	url := fmt.Sprintf("%s/build/%s", uiBase, url.PathEscape(b.Id))
 	if evergreen.IsPatchRequester(b.Requester) {
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -301,11 +301,7 @@ func (b *Build) GetTimeSpent() (time.Duration, time.Duration, error) {
 
 // GetURL returns a url to the build page.
 func (b *Build) GetURL(uiBase string) string {
-	url := fmt.Sprintf("%s/build/%s", uiBase, url.PathEscape(b.Id))
-	if evergreen.IsPatchRequester(b.Requester) {
-		url += "?redirect_spruce_users=true"
-	}
-	return url
+	return fmt.Sprintf("%s/build/%s?redirect_spruce_users=true", uiBase, url.PathEscape(b.Id))
 }
 
 // Insert writes the b to the db.
--- a/model/build/build.go
+++ b/model/build/build.go
@@ -2,13 +2,17 @@ package build
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/db"
 	mgobson "github.com/evergreen-ci/evergreen/db/mgo/bson"
 	"github.com/evergreen-ci/evergreen/model/task"
+	"github.com/evergreen-ci/utility"
 	adb "github.com/mongodb/anser/db"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo/options"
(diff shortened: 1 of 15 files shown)


model/build/build.go
```diff
@@ -317,3 +331,78 @@ func (b Builds) InsertMany(ctx context.Context, ordered bool) error {
...
+		desc += fmt.Sprintf(", %d internal errors", systemError)
+	}
+	if other > 0 {
+		desc += fmt.Sprintf(", %d other", other)
+	}
+
+	return b.appendTime(desc)
+}
+
+func statusIsSystemFailure(status string) bool {
```
malikchaya2, line 384: Can we instead model this function like `IsSystemFailedTaskStatus` and keep it in globals?

rest/route/github.go
```diff
@@ -30,9 +31,10 @@ const (
 	githubActionReopened    = "reopened"
 
 	// pull request comments
-	retryComment   = "evergreen retry"
-	patchComment   = "evergreen patch"
-	triggerComment = "evergreen merge"
+	retryComment         = "evergreen retry"
+	refreshStatusComment = "evergreen status"
```
malikchaya2, line 35: I have a slight preference for "evergreen refresh". 
I think it works nicely with "evergreen retry". One will take the tests and retry them, and on will simply refresh them.

rest/route/github.go
```diff
@@ -645,6 +664,12 @@ func triggersPatch(action, comment string) (bool, string) {
 
 	return false, ""
 }
+func triggersStatusRefresh(action, comment string) bool {
+	if action == "deleted" {
+		return false
+	}
+	return comment == refreshStatusComment
```
malikchaya2, line 671: Can we please wrap this in a function that checks for whitespace, similar to how `isRetryComment` is set up?

rest/route/github.go
```diff
@@ -326,6 +329,23 @@ func (gh *githubHookApi) createPRPatch(ctx context.Context, owner, repo, calledB
...
+		return errors.Wrap(err, "finding patch")
+	}
+	if p == nil {
+		return errors.Errorf("couldn't find patch for PR '%s/%s:%d'", owner, repo, prNumber)
+	}
+	if p.Version == "" {
+		return errors.Errorf("patch '%s' not finalized", p.Version)
+	}
+
+	job := units.NewGithubStatusRefreshJob(p)
```
malikchaya2, line 344: How much lag are we expecting there to be between the user commenting and them seeing the refresh?

malikchaya2 (changes requested):

model/build/build.go
```diff
@@ -294,6 +299,15 @@ func (b *Build) GetTimeSpent() (time.Duration, time.Duration, error) {
 	return timeTaken, makespan, nil
 }
 
+// GetURL returns a url to the build page.
+func (b *Build) GetURL(uiBase string) string {
+	url := fmt.Sprintf("%s/build/%s", uiBase, url.PathEscape(b.Id))
+	if evergreen.IsPatchRequester(b.Requester) {
+		url += "?redirect_spruce_users=true"
```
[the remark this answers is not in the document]
ablack12: You know, I don't know -- I just stole this from version link. We also do this for build link. I wonder if this is out of date from when spruce was only built for patches? @khelif96 should we just always add this now?

model/build/build.go
ablack12: having wrote half this code a month ago i have no idea why i even did this haha, since this whole function is just TaskStatusToDesc moved except for this change. switched to use the existing IsSystemFailedTaskStatus.

rest/route/github.go
ablack12: alright refresh it is!

rest/route/github.go
ablack12: When I tested it, it was pretty instant! Although I wrote this as a job bc it seemed more consistent / reusable for different things, we're just running the job right away as opposed to queueing it so I wouldn't expect delay.

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		return false, errors.Wrap(err, "error retrieving admin settings")
+	}
+	if flags.GithubStatusAPIDisabled {
+		grip.InfoWhen(sometimes.Percent(evergreen.DegradedLoggingPercent), message.Fields{
+			"job":     j.Name,
+			"message": "GitHub status updates are disabled, not updating status",
+		})
+		return false, nil
```
hadjri, line 80: Can remove this line and simply call `return !flags.GithubStatusAPIDisabled, nil`

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	return true, nil
+}
+
+func (j *githubStatusRefreshJob) fetch() error {
```
hadjri, line 85: seems that a more fitting name would be something like `populate` since we're populating the job struct's data here

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+	if j.env == nil {
+		j.env = evergreen.GetEnvironment()
+	}
+	uiConfig := evergreen.UIConfig{}
+	var err error
+	if err := uiConfig.Get(j.env); err != nil {
+		return errors.Wrap(err, "retrieving UI config")
+	}
+	urlBase := uiConfig.Url
```
hadjri, line 94: urlBase can be removed and just use j.urlBase

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+	if urlBase == "" {
+		return errors.New("url base doesn't exist")
+	}
+	j.urlBase = urlBase
+	if j.sender == nil {
+		var err error
+		j.sender, err = j.env.GetSender(evergreen.SenderGithubStatus)
+		if err != nil {
```
hadjri, line 102: should we also add a nil check for the sender here or is that not necessary?

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+	status := &message.GithubStatus{
+		URL:     j.patch.GetURL(j.urlBase),
+		Context: evergreenContext,
+		Owner:   j.patch.GithubPatchData.BaseOwner,
+		Repo:    j.patch.GithubPatchData.BaseRepo,
+		Ref:     j.patch.GithubPatchData.HeadHash,
+	}
+
+	// Send patch status
+	patchDuration := j.patch.FinishTime.Sub(j.patch.StartTime).String()
```
hadjri, line 209: Can we move this duplicated code to a helper?

model/build/build.go
```diff
@@ -317,3 +331,78 @@ func (b Builds) InsertMany(ctx context.Context, ordered bool) error {
...
+			success++
+
+		case t.Status == evergreen.TaskFailed:
+			failed++
+
+		case statusIsSystemFailure(t.Status):
+			systemError++
+
+		case utility.StringSliceContains(evergreen.TaskUncompletedStatuses, t.Status):
+			noReport++
```
hadjri, line 355: Should we just return something / error out early here since it shows the build is not finished? 

If we do this, maybe we can use a string -> int map of statuses to count, with `TaskFailed, TaskSucceeded, TaskSystemFailed` being the keys rather than setting up these 5 variables.

model/patch/patch.go
```diff
@@ -808,6 +808,21 @@ func (p *Patch) GetPatchIndex(parentPatch *Patch) (int, error) {
...
 }
 
+// GetGithubContextForChildPatch returns the github context for the given child patch, to be used in github statuses.
+func GetGithubContextForChildPatch(projectIdentifier string, parentPatch, childPatch *Patch) (string, error) {
+	var githubContext string
+	patchIndex, err := childPatch.GetPatchIndex(parentPatch)
+	if err != nil {
+		return "", errors.Wrap(err, "getting child patch index")
+	}
+	if patchIndex == 0 || patchIndex == -1 {
```
hadjri, line 818: Can we add a small comment on why this if/else block is necessary?

units/github_status_refresh.go
```diff
@@ -0,0 +1,259 @@
...
+const (
+	githubStatusRefreshJobName = "github-status-refresh"
+)
+
+func init() {
+	registry.AddJobType(githubStatusUpdateJobName, func() amboy.Job { return makeGithubStatusRefreshJob() })
+}
+
+// NewGithubStatusRefreshJob is a job that re-sends github statuses to the PR associated with the given patch.
+func NewGithubStatusRefreshJob(p *patch.Patch) amboy.Job {
```
hadjri, line 35: Logic of the job looks sound but can we add some unit tests for it?

hadjri (changes requested):

model/build/build.go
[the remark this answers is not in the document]
khelif96: Yeah, I think for most links that have a redirect configured this makes sense.

model/build/build.go
ablack12: I didn't write this function (just moved it so it could be used in the units package) but it does look like even if the build isn't finished we still show how many have succeeded / failed so far looking at the description.

model/patch/patch.go
ablack12: I'm honestly not sure since I grabbed this from elsewhere; I'm assuming it's because the patch might not have been created yet? Or to catch the case that it's not actually being called with a child patch? cc @malikchaya2 can you weigh in

units/github_status_refresh.go
ablack12: likely unnecessary; doesn't look like we do it anywhere else or that the functions can return a nil sender without an error.

malikchaya2 (approved):

model/patch/patch.go
malikchaya2: We needed a way to identify child patches uniquely so that Github can update the status. Using `evergreen/projectIdentifier` wasn't enough because there can be two child patches in the same project. We use the index as a unique key. 
We don't need it if the index is -1 because that is used to indicate that the patch is not a child patch. 0 doesn't need it because it can just be `"evergreen/project"` instead of `"evergreen/project/index"` and still be unique. 

If we want to, we can add a comment to the effect of: 
"if the patch is a child patch and it's not the only child, use the index as a unique identifier to disambiguate the child patches". 

I'm also realizing that we can remove the -1 altogether because it checks to make sure that the patch has a parent before getting the index. I created https://jira.mongodb.org/browse/EVG-19081 to clean it up, so I can add the comment then.

ablack12: by "in the same project" do you mean two child patches within the same child patch project?

malikchaya2: yes

hadjri: Ah ok, I understand - can we add a comment on this line consolidating this explanation?

rest/route/github_test.go
```diff
@@ -302,11 +302,18 @@ func (s *GithubWebhookRouteSuite) TestRetryCommentTrigger() {
...
-	s.False(triggersPatch("deleted", commentString))
+	s.True(triggersPatch(commentString))
 
 	//test whitespace trimming
-	s.True(triggersPatch("created", "  evergreen retry "))
+	s.True(triggersPatch("  evergreen retry "))
+}
+
+func (s *GithubWebhookRouteSuite) TestRefreshStatusTrigger() {
+	s.True(triggersPatch(refreshStatusComment))
```
hadjri, line 312: I think these assertions may need to be flipped in this test?

hadjri (changes requested):
Just one final comment and fixing the failing tests - FWIW the failing smoke tests are unrelated to this PR it seems

rest/route/github_test.go
ablack12: was calling the wrong function (#copypasta) but fwiw no need for reviewers to help with debugging failing tests, since that should be on the writer

hadjri (approved):

PR #6292, opened 2023-02-27
EVG-18935: Save and Resolve Project.ExternalLinks & Version.ExternalLinksForPatchMetadata

SupaJoon:
[EVG-18935](https://jira.mongodb.com/browse/EVG-18935)

### Description
ExternalLinks are saved to Project Ref and contain a URL template and name. The URL template may contain `{version_id}` which is replaced by an actual version ID by Version.ExternalLinksForPatchMetadata. 
The links will be displayed on the Patch Metadata panel and set from the Plugins project settings tab as part of [EVG-19057](https://jira.mongodb.org/browse/EVG-19057)

### Testing
integration and unit tests

### DIFF
--- a/util/webhook_grip_test.go
+++ b/util/webhook_grip_test.go
@@ -220,7 +220,7 @@ func (t *mockWebhookTransport) RoundTrip(req *http.Request) (*http.Response, err
 	t.lastUrl = req.URL.String()
 	resp := &http.Response{
 		StatusCode: http.StatusNoContent,
-		Body:       io.NopCloser(nil),
+		Body:       http.NoBody,
 	}
 
 	if t.attemptCount < t.minAttempts {


model/project_ref_test.go
```diff
@@ -2961,6 +2961,27 @@ func TestSaveProjectPageForSection(t *testing.T) {
...
 	_, err = SaveProjectPageForSection("iden_", update, ProjectPageGeneralSection, false)
 	assert.NoError(err)
+
+	// Test successful external link update
+	update = &ProjectRef{
+		ExternalLinks: []ExternalLink{
+			{URLTemplate: "https://arnars.com/{version_id}", DisplayName: "A link"},
+		},
+	}
+	_, err = SaveProjectPageForSection("iden_", update, ProjectPagePluginSection, false)
```
malikchaya2, line 2971: Can we please add a test that checks that the `{version_id}` was replaced successfully, validating that the url is equal to the url we expect it to be?

malikchaya2 (changes requested):
[jira.mongodb.com/browse/EVG-18935](jira.mongodb.com/browse/EVG-18935) doesn't have a lot of details about external patch links. 
Can you please explain a bit more about what they are, what they link to, and why they are needed? 
Is there an example of one?

graphql/schema/types/version.graphql
```diff
@@ -69,6 +69,7 @@ type Version {
   taskStatusStats(options: BuildVariantOptions!): TaskStats
   upstreamProject: UpstreamProject
   versionTiming: VersionTiming
+  externalPatchLinks: [ExternalPatchLink!]!
```
malikchaya2, line 72: `externalPatchLinks` reads to me like it's a link to an external patch. Can we please change it to just externalLink or externalLinksForPatch? 
Similarly, can we please update the PR title and description to "external links for patches" or some other description that makes the distinction clear?

malikchaya2 (changes requested):

graphql/schema/types/version.graphql
khelif96: +1

graphql/version_resolver.go
```diff
@@ -469,6 +470,27 @@ func (r *versionResolver) VersionTiming(ctx context.Context, obj *restModel.APIV
...
+	pRef, err := data.FindProjectById(*obj.Project, false, false)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding project `%s`: %s", *obj.Project, err.Error()))
+	}
+	if pRef == nil {
+		return nil, ResourceNotFound.Send(ctx, fmt.Sprintf("Project `%s` not found", *obj.Project))
+	}
+	var externalPatchLinks []*ExternalPatchLink
+	for _, link := range pRef.ExternalLinks {
+		// replace {version_id} with the actual version id
```
khelif96, line 484: Does it make sense to support any other fields?

graphql/schema/types/version.graphql
malikchaya2: I see that you made the change to the field. Can you please also change it in the PR title and description so that it's easy to understand if we ever need to refer back to it?

malikchaya2 (approved):

graphql/version_resolver.go
khelif96: I was thinking author's name or branch or some other relevant information but we can add those in later if we need them.

khelif96 (approved):
Last comments everything else LGTM

PR #6293, opened 2023-02-27
EVG-18700: get project config with unfinalized patches from legacy route

Kimchelly:
EVG-18700

### Description
I missed that `getPatchConfig` in the legacy route depends on the `PatchedParserProject`, and that route is used by the legacy CLI client to get parser project info.

### Testing
This function already doesn't have testing, I don't want to add tests for a hotfix.

N/A

### DIFF
--- a/service/rest_patch.go
+++ b/service/rest_patch.go
@@ -77,24 +77,35 @@ func (restapi restAPI) getPatchConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if projCtx.Version == nil {
-		gimlet.WriteJSONInternalError(w, fmt.Sprintf("cannot get parser project for patch '%s' because version is nil", projCtx.Patch.Id.Hex()))
-		return
-	}
-
 	settings := restapi.GetSettings()
-	pp, err := model.ParserProjectFindOneByID(r.Context(), &settings, projCtx.Version.ProjectStorageMethod, projCtx.Version.Id)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


hadjri (approved):

service/rest_patch.go
```diff
@@ -77,24 +77,35 @@ func (restapi restAPI) getPatchConfig(w http.ResponseWriter, r *http.Request) {
...
-		gimlet.WriteJSONInternalError(w, errors.Wrapf(err, "finding parser project '%s' for patch '%s'", projCtx.Version.Id, projCtx.Patch.Id.Hex()))
-		return
-	}
-	if pp == nil {
-		gimlet.WriteJSONInternalError(w, fmt.Sprintf("parser project '%s' for patch '%s' not found", projCtx.Version.Id, projCtx.Patch.Id.Hex()))
+	var pp *model.ParserProject
+	var err error
+	if projCtx.Patch.ProjectStorageMethod != "" {
+		pp, err = model.ParserProjectFindOneByID(r.Context(), &settings, projCtx.Version.ProjectStorageMethod, projCtx.Patch.Id.Hex())
+		if err != nil {
```
bynn, line 85: nit: can we move the err and pp == nil check outside of the if so we dont duplicate it?

bynn (approved):

service/rest_patch.go
Kimchelly: I'd prefer to keep them separate so we can keep the error message wraps separate, which helps know which code path it took.

PR #6296, opened 2023-02-27
EVG-5829 Expose AMI field for hosts

khelif96:
EVG-5829

### Description
Exposes the AMI field for hosts

### DIFF
--- a/graphql/host_resolver.go
+++ b/graphql/host_resolver.go
@@ -22,11 +22,6 @@ func (r *hostResolver) Ami(ctx context.Context, obj *restModel.APIHost) (*string
 		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding host %s: %s", utility.FromStringPtr(obj.Id), err.Error()))
 	}
 	if host == nil {
-		// This should never happen, but if it does, we should log it.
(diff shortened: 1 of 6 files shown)


graphql/host_resolver.go
```diff
@@ -15,6 +15,23 @@ import (
...
 
+// Ami is the resolver for the ami field.
+func (r *hostResolver) Ami(ctx context.Context, obj *restModel.APIHost) (*string, error) {
+	host, err := host.FindOneId(utility.FromStringPtr(obj.Id))
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error finding host %s: %s", utility.FromStringPtr(obj.Id), err.Error()))
+	}
+	if host == nil {
+		// This should never happen, but if it does, we should log it.
+		grip.Warning(message.Fields{
```
khelif96, line 26: I'm not sure if this warrants logging at all or if we should even log at a higher level. If this ever happens something is probably wrong with the host.

ablack12: we log graphql errors anyway right? so idk about an additional log.

ablack12 (approved):

PR #6300, opened 2023-02-28
EVG-17384: Expose Warnings and Errors fields for versions

minnakt:
EVG-17384

### Description
This PR exposes the warnings and errors field for versions. The `Errors` field comes directly from the `APIVersion` and the `Warnings` field comes from an additional resolver to avoid exposing it in the rest route.

I also made some changes to the local data so that we have testing data for the UI.

### Testing
- Added GraphQL test
- Note that failing tasks are failing on base

- Not from this PR, but the [REST V2 version documentation](https://github.com/evergreen-ci/evergreen/wiki/REST-V2-Usage#version) needs to be updated because it seems to be missing fields like `aborted`, `order`, and `errors`.


khelif96 (approved):

ablack12 (changes requested):
Warnings can potentially be really long; I'm not sure we want to expose these with our rest route.

minnakt:
> Warnings can potentially be really long; I'm not sure we want to expose these with our rest route.

I moved `Warnings` to a resolver so it won't show in the rest route anymore

ablack12 (approved):

PR #6301, opened 2023-03-01
EVG-18931 Add more debugging to event send

bynn:
EVG-18931

### Description
adding more discussion as we discussed

### DIFF
--- a/units/event_send.go
+++ b/units/event_send.go
@@ -80,6 +80,7 @@ func (j *eventSendJob) Run(_ context.Context) {
 			"message":     "unable to set up event send job",
 			"jira_ticket": "EVG-18931",
 			"job_id":      j.ID(),
+			"error":       err.Error(),
 		})
 		j.AddError(err)
 		return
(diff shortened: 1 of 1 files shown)


units/event_send.go
```diff
@@ -118,8 +147,25 @@ func (j *eventSendJob) send(n *notification.Notification) error {
 		return err
 	}
 	if err = c.SetPriority(level.Notice); err != nil {
+		grip.Debug(message.Fields{
```
Kimchelly, line 150: Should this (and the other ones) log the error message as well?

units/event_send.go
```diff
@@ -118,8 +147,25 @@ func (j *eventSendJob) send(n *notification.Notification) error {
 		return err
```
Kimchelly, line 147: Is it worth adding temporary logging to these other return cases in `send`?

Kimchelly (approved):
LGTM mod some suggestions to make it easier to debug.

PR #6303, opened 2023-03-01
EVG-19082 Do not activate disabled tasks in repotracker job

hadjri:
EVG-19082

### Description
The version activation done by repotracker jobs has a [bug](https://gist.github.com/hadjri/dc7854c1bc80233cf3252684ca5eebef#file-lifecycle-go-L135) where it does not ignore disabled tasks when doing activation. Therefore,  setting `priority: -1 in YAML` would not have any effect for mainline versions because they get [un-disabled ](https://gist.github.com/hadjri/3e7b9681daeebadb310db6bc0fc3b2e8#file-db-go-L2687)during activation, which is behavior we don't want. Reproduced this bug in staging.

Added a change to not query disabled tasks when performing activation.
### Testing
Added new unit test affirming this, also tested in staging.
N/A

### DIFF
--- a/repotracker/repotracker_test.go
+++ b/repotracker/repotracker_test.go
@@ -282,6 +282,7 @@ buildvariants:
   - name: t1
     batchtime: 30
   - name: t2
+  - name: t3
 - name: bv2
   display_name: bv2_display
   run_on: d2
@@ -292,6 +293,7 @@ tasks:
   priority: 3
 - name: t2
   priority: -1
+- name: t3
 `
 
 	previouslyActivatedVersion := &model.Version{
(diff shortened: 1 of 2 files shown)


repotracker/repotracker_test.go
```diff
@@ -371,6 +373,13 @@ tasks:
 	build1, err := build.FindOneId(bv.BuildId)
 	assert.NoError(t, err)
 
+	// neither batchtime task nor disabled task should be activated
+	tasks, err := task.Find(task.ByBuildId(build1.Id))
+	assert.Len(t, tasks, 2)
+	for _, tsk := range tasks {
+		assert.False(t, tsk.Activated)
```
Kimchelly, line 380: Can the BV also include a task that should be activated to double-check that it activates a subset of the BV tasks?

Kimchelly (changes requested):

Kimchelly (approved):

PR #6304, opened 2023-03-02
EVG-18080 Upgrade cocoa and reduce container specs

hadjri:
EVG-18080

### Description
The container instances in `evergreen-linux-test` have lower resource capabilities than those of the previous cluster we were testing in, `evergreen-linux-development`. Their maximum CPU / memory respectively is 2048 / 3883.  When we configured resources past these numbers in the YAML file, tasks remained in PROVISIONING state indefinitely because they could not fit onto the container instances.

Reduced the default CPU/memory of our container variant in the staging YAML, and also fixed an issue where we weren't using the most recent version of Cocoa in Evergreen, which was preventing pod environment variables from populating correctly on the containers.
### Testing
Confirmed that pod environment variables are getting populated properly in ECS and tasks are getting allocated to the container instances correctly after the change.
N/A


self-tests-staging.yml
```diff
@@ -6,8 +6,8 @@ containers:
     working_dir: /
     image: "hadjri/evg-e2e-test-ubuntu"
     resources:
-      cpu: 4096
-      memory_mb: 8192
+      cpu: 2048
+      memory_mb: 3584
```
Kimchelly, line 10: Per our discussion on Slack, I don't think we need to reduce the CPU/memory available to the container since the staging cluster has large enough container instances.

Kimchelly (approved):
LGTM mod the one comment.

PR #6306, opened 2023-03-03
EVG-19098 refresh status if commit queue merge is blocked

ablack12:
EVG-19098

### Description
We added the `evergreen refresh` comment which will be cool but it would also be cool if we automatically refreshed this if we see that a merge is blocked but the patch is successful (this may over-refresh since a merge can be blocked on other things but since this is manually triggered I don't expect it to be too much).

### Testing
Tested with a PR, where the first merge fails because PR status sending is disabled, and the second merge succeeds.
https://github.com/evergreen-ci/commit-queue-sandbox/pull/453

Will add to the `evergreen merge` documentation

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -309,35 +309,9 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
 		return nil, errors.Errorf("user '%s' is not authorized to merge", userRepo.Username)
 	}
 
-	pr, err := sc.GetGitHubPR(ctx, userRepo.Owner, userRepo.Repo, info.PR)
+	pr, err := getPRAndCheckMergeable(ctx, env, sc, info)
 	if err != nil {
-		return nil, errors.Wrap(err, "getting PR from GitHub API")
-	}
-
-	if pr == nil || pr.Base == nil || pr.Base.Ref == nil {
-		return nil, errors.New("PR contains no base branch label")
-	}
-
-	if !thirdparty.IsUnblockedGithubStatus(pr.GetMergeableState()) {
-		errMsg := fmt.Sprintf("PR is not mergeable, status: %s", pr.GetMergeableState())
-		grip.Debug(message.Fields{
-			"message":        errMsg,
-			"state":          pr.GetMergeableState(),
-			"owner":          userRepo.Owner,
-			"repo":           userRepo.Repo,
-			"pr_title":       pr.GetTitle(),
-			"commit_message": pr.GetNumber(),
-		})
-		sendErr := thirdparty.SendCommitQueueGithubStatus(env, pr, message.GithubStateFailure, errMsg, "")
-
-		grip.Error(message.WrapError(sendErr, message.Fields{
-			"message": "error sending patch creation failure to github",
-			"owner":   userRepo.Owner,
-			"repo":    userRepo.Repo,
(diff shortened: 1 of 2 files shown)


rest/data/commit_queue.go
```diff
@@ -377,6 +351,72 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
...
+			// Not required that such a patch exists, since the commit queue can be enabled without PR checks, so we log at debugging.
+			grip.Debug(message.WrapError(err, message.Fields{
+				"message": "couldn't find latest PR patch when enqueuing PR",
+				"ticket":  "EVG-19098",
+				"owner":   info.Owner,
+				"repo":    info.Repo,
+				"pr":      info.PR,
+			}))
+		} else if p != nil && p.Version != "" && p.Status == evergreen.PatchSucceeded {
+			refreshJob := units.NewGithubStatusRefreshJob(p)
```
malikchaya2, line 387: Is this meant to address the potential pain point of the PR specifically being blocked because a required status didn't pass, and the user seeing an error message saying that it can't be enqueued because it's blocked, but not knowing why it's blocked because we didn't update the status from "running" to "failed"?

rest/data/commit_queue.go
```diff
@@ -377,6 +351,72 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
...
+	if err != nil {
+		return nil, err
+	}
+
+	// If the PR is blocked and the patch is successful, refresh status, and re-check PR.
+	if pr.GetMergeableState() == thirdparty.GithubPRBlocked {
+		p, err := patch.FindLatestGithubPRPatch(info.Owner, info.Repo, info.PR)
+		if err != nil {
+			// Not required that such a patch exists, since the commit queue can be enabled without PR checks, so we log at debugging.
+			grip.Debug(message.WrapError(err, message.Fields{
```
hadjri, line 379: Should we also log this if the patch is nil, since the log is stating that the patch could not be found?

rest/data/commit_queue.go
```diff
@@ -377,6 +351,72 @@ func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc C
...
+				"message": "couldn't find latest PR patch when enqueuing PR",
+				"ticket":  "EVG-19098",
+				"owner":   info.Owner,
+				"repo":    info.Repo,
+				"pr":      info.PR,
+			}))
+		} else if p != nil && p.Version != "" && p.Status == evergreen.PatchSucceeded {
+			refreshJob := units.NewGithubStatusRefreshJob(p)
+			refreshJob.Run(ctx)
```
hadjri, line 388: Worth it to add a `"ticket": "EVG-19098"` debug log here as well for traceability?

rest/data/commit_queue.go
ablack12: ah yes that makes sense

malikchaya2 (approved):

rest/data/commit_queue.go
ablack12: I actually meant to remove that elsewhere, it was just for my personal debugging. Imo tickets should only ever be used in debugging if the logging is being introduced to aid in a future or the same ticket (with the intention of removing it after the investigation is done), otherwise it's just noise.

hadjri (approved):

PR #6308, opened 2023-03-03
EVG-19109 fix panic accessing nil volume

ablack12:
EVG-19109

### Description
fix panic accessing nil volume in error


sophstad (approved):
beautiful

bynn (approved):

PR #6310, opened 2023-03-08
EVG-19135: Improve patch validate error message

julianedwards:
EVG-19135

### Description
Improve error messages when `git apply --numstat` fails to make it more clear to users why their patch fails.

[Slack archive](https://mongodb.slack.com/archives/C0V896UV8/p1678303811935539)

### Testing
N/A

No.

### DIFF
--- a/thirdparty/git.go
+++ b/thirdparty/git.go
@@ -68,13 +68,13 @@ func GitApplyNumstat(patch string) (*bytes.Buffer, error) {
 
 	// this should never happen if patch is initially validated
 	if err := cmd.Start(); err != nil {
-		return nil, errors.Wrapf(err, "starting `git apply --numstat`: 424 - %v",
+		return nil, errors.Wrapf(err, "starting 'git apply --numstat': 424 - %v",
 			summaryBuffer.String())
 	}
 
 	// this should never happen if patch is initially validated
 	if err := cmd.Wait(); err != nil {
-		return nil, errors.Wrapf(err, "running `git apply --numstat`: 562 - %v",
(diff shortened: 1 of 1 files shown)


thirdparty/git.go
```diff
@@ -138,15 +138,15 @@ func GetPatchSummaries(patchContent string) ([]Summary, error) {
 	if patchContent != "" {
 		gitOutput, err := GitApplyNumstat(patchContent)
 		if err != nil {
-			return nil, errors.Wrap(err, "couldn't validate patch")
+			return nil, errors.Wrap(err, "validating patch")
 		}
 		if gitOutput == nil {
-			return nil, errors.New("couldn't validate patch: git apply --numstat returned empty")
+			return nil, errors.New("validating patch: `git apply --numstat` returned empty")
```
ablack12, line 144: Nit: since this isn't a wrap error it can be a full sentence, i.e. should say couldn't or error at the front

ablack12 (approved):

thirdparty/git.go
julianedwards: It gets wrapped by other things higher in the call stack, something like:
```
Unexpected reply from server (500 Internal Server Error): 400 (Bad Request): building CLI patch document: getting module patch from GridFS: getting patch summaries: couldn't validate patch: Error waiting on patch: 562 - error: unrecognized input
: exit status 128
(base)
```

ablack12: This is different though because it's the lowest error -- it's not wrapping anything. In the example you sent, this error is basically the "unrecognized input part"

ablack12 (approved):

julianedwards:
evergreen refresh

PR #6314, opened 2023-03-10
EVG-16701 Add bulk version priority setting route

hadjri:
EVG-16701

### Description
Unbundled the logic between parsing this route and the `getProjectVersions` route that existed in previous PRs, and this route now stands alone. Boundaries of the bulk update may be set by revision order numbers, or start/end timestamps. Requester is also a specifiable parameter

I've added this context as the first ADR for the Evergreen repo since this reasoning might be applicable in the future. Per the [docs](https://docs.devprod.prod.corp.mongodb.com/devprod-docs/ADR-presentation/), I've initialized a `docs/decisions` folder to store ADRs and their templates.
### Testing
Added unit test, tested in staging.
Yes, will update REST v2 guide on approval with how to use the route


docs/decisions/2023-03-13_shared_rest_route_logic.md
```diff
@@ -0,0 +1,20 @@
+# 2023-03-13 Shared REST Route Logic
+
+* status: accepted
+* date: 2023-03-13
+* authors: Malik Hadjri
+
+## Context and Problem Statement
+
+A feature request was made for a new endpoint to perform a bulk operation on a group of versions. Since there was an existing GET endpoint to retrieve a list of versions based off of detailed request parameters, the
```
ablack12, line 9: Could we include the ticket number in this?

docs/decisions/adr.md.template
```diff
@@ -0,0 +1,18 @@
+# {YYYY-MM-DD} {short title of solved problem and solution}
```
ablack12, line 1: Thanks for getting this set up -- super excited to start seeing these in our codebase!! 🥳

model/lifecycle.go
```diff
@@ -273,7 +273,7 @@ func SetBuildPriority(buildId string, priority int64, caller string) error {
 
 	// negative priority - these tasks should never run, so unschedule now
 	if priority < 0 {
-		tasks, err := task.FindAll(db.Query(bson.M{task.BuildIdKey: buildId}).WithFields(task.IdKey, task.ExecutionKey))
```
ablack12, line 276: Why did we change this?

model/version.go
```diff
@@ -629,6 +632,43 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+		return nil, err
+	}
+	match := bson.M{
+		VersionIdentifierKey: projectId,
+		VersionRequesterKey:  opts.Requester,
+	}
+
+	// setting revision numbers will take precedence over setting start and end times
+	if opts.RevisionStart > 0 {
+		match[VersionRevisionOrderNumberKey] = bson.M{"$lte": opts.RevisionStart, "$gte": opts.RevisionEnd}
```
ablack12, line 658: Do we want revisionEnd / endTime required or would we just allow this to default to latest?

model/version.go
```diff
@@ -629,6 +632,43 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
...
+	} else {
+		match[VersionCreateTimeKey] = bson.M{"$gte": startTime, "$lte": endTime}
+	}
+	pipeline := []bson.M{{"$match": match}}
+	pipeline = append(pipeline, bson.M{"$sort": bson.M{VersionRevisionOrderNumberKey: -1}})
```
ablack12, line 663: Is there a benefit to having this be sorted? if this is just for tests, can we write them to not rely on sort? (could iterate through the list and validate based on ID or etc)

ablack12 (changes requested):

model/lifecycle.go
ablack12: This is. still a question I have^

model/version.go
ablack12: This is still a question I have^

ablack12 (changes requested):
Accidentally added my comments without a review -- no new comments but a couple haven't been addressed.

ablack12 (changes requested):
@hadjri reminder I'm waiting on a few open threads (https://github.com/evergreen-ci/evergreen/pull/6314#discussion_r1144950523, https://github.com/evergreen-ci/evergreen/pull/6314#discussion_r1144950941)

ablack12 (approved):

PR #6315, opened 2023-03-13
EVG-19132 Add checksum verification for Go lint installer

pierwill:
EVG-19132

### Description
(Part of a skunkworks security initiative! 😇 )

This PR adds a checksum verification for the Go lint installer script. This reduces the probability of running a malicious script. (For a recent example of such an attack, see https://blog.aquasec.com/codecovs-breach-supply-chain-attack.)

### Testing
`make lint-util` passes locally. Also ran following [these instructions](https://github.com/evergreen-ci/evergreen/pull/6315#issuecomment-1466621446).

No

### DIFF
--- a/makefile
+++ b/makefile
@@ -208,7 +208,7 @@ $(buildDir)/golangci-lint:
 	@GO_LINT_INSTALLER_CHECKSUM="0e09dedc7e35f511b7924b885e50d7fe48eef25bec78c86f22f5b5abd24976cc"
 	@curl --retry 10 --retry-max-time 120 -sSfL -O https://raw.githubusercontent.com/golangci/golangci-lint/${GO_LINT_INSTALLER_VERSION}/install.sh
 	@echo "${GO_LINT_INSTALLER_CHECKSUM} install.sh" | sha256sum --check
-	@sh install.sh -b $(buildDir) ${GO_LINT_INSTALLER_VERSION} >/dev/null 2>&1 && touch $@
+	@bash install.sh -b $(buildDir) ${GO_LINT_INSTALLER_VERSION} >/dev/null 2>&1 && touch $@
 $(buildDir)/run-linter:cmd/run-linter/run-linter.go $(buildDir)/.lintSetup
 	$(gobin) build -ldflags "-w" -o $@ $<
 # end lint setup targets
--- a/makefile
+++ b/makefile
@@ -99,6 +99,9 @@ agentVersion := $(shell grep "AgentVersion" config.go | tr -d '\tAgentVersion = 
 ldFlags := $(if $(DEBUG_ENABLED),,-w -s )-X=github.com/evergreen-ci/evergreen.BuildRevision=$(currentHash)
 gcFlags := $(if $(STAGING_ONLY),-N -l,)
 karmaFlags := $(if $(KARMA_REPORTER),--reporters $(KARMA_REPORTER),)
+
+goLintInstallerVersion := "v1.51.2"
+goLintInstallerChecksum := "0e09dedc7e35f511b7924b885e50d7fe48eef25bec78c86f22f5b5abd24976cc"
 # end evergreen specific configuration
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -204,7 +204,11 @@ coverageHtmlOutput := $(foreach target,$(packages),$(buildDir)/output.$(target).
 $(buildDir)/.lintSetup:$(buildDir)/golangci-lint
 	@touch $@
 $(buildDir)/golangci-lint:
-	@curl --retry 10 --retry-max-time 120 -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(buildDir) v1.51.2 >/dev/null 2>&1 && touch $@
+	@GO_LINT_INSTALLER_VERSION="v1.51.2"
+	@GO_LINT_INSTALLER_CHECKSUM="0e09dedc7e35f511b7924b885e50d7fe48eef25bec78c86f22f5b5abd24976cc"
+	@curl --retry 10 --retry-max-time 120 -sSfL -O https://raw.githubusercontent.com/golangci/golangci-lint/${GO_LINT_INSTALLER_VERSION}/install.sh
+	@echo "${GO_LINT_INSTALLER_CHECKSUM} install.sh" | sha256sum --check
+	@sh install.sh -b $(buildDir) ${GO_LINT_INSTALLER_VERSION} >/dev/null 2>&1 && touch $@
```
Kimchelly, line 211: I would explicitly invoke bash here if it's a bash syntax script. Using `sh` isn't advisable because the particular shell that's invoked by `sh` varies across distros.

Kimchelly (changes requested):
If we're making this change to the Evergreen repo, it also has to be applied to all the other evergreen-ci repos, as well as https://github.com/mongodb/amboy and https://github.com/mongodb/grip since they're all owned by Evergreen.

Kimchelly:
Alternatively, if you don't want to open PRs for all of the repos, you could file a follow-up EVG ticket to copy the checksum verification logic and someone else can do it.

Kimchelly:
Sorry, I forgot to mention how to test this. We typically run `make lint-<NAME_OF_PACKAGE>`, so you could test this change works locally with `make lint-util` to run the linter against `/util`.

If you want to test that all the packages lint properly in CI tests, you can just reduce [this logic](https://github.com/evergreen-ci/evergreen/blob/213d38fa25caa5d3c68e5f356319167897b27cde/cmd/generate-lint/generate-lint.go#L121-L136) down to just this:
```go
targets, err := getAllTargets()
if err != nil {
	return nil, err
}
```

Kimchelly (commented):
I ran this to double check and it failed:
```
$ make lint-util
GO_LINT_INSTALLER_VERSION="v1.51.2"
GO_LINT_INSTALLER_CHECKSUM="0e09dedc7e35f511b7924b885e50d7fe48eef25bec78c86f22f5b5abd24976cc"
curl --retry 10 --retry-max-time 120 -sSfL -O https://raw.githubusercontent.com/golangci/golangci-lint//install.sh
curl: (22) The requested URL returned error: 400
make: *** [makefile:209: bin/golangci-lint] Error 22
```

Since [make invokes a new shell with each recipe line](https://www.gnu.org/software/make/manual/html_node/Execution.html), I would recommend just defining `go_lint_installer_version` and `go_lint_installer_checksum` as plain make variables (like how [these](https://github.com/evergreen-ci/evergreen/blob/c2b09d215630d7960dbedbfe39e19fab79925e54/makefile#L79-L84) are defined) rather than trying to inject too much shell syntax into make rules.

makefile
```diff
@@ -204,7 +207,9 @@ coverageHtmlOutput := $(foreach target,$(packages),$(buildDir)/output.$(target).
 $(buildDir)/.lintSetup:$(buildDir)/golangci-lint
 	@touch $@
 $(buildDir)/golangci-lint:
-	@curl --retry 10 --retry-max-time 120 -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(buildDir) v1.51.2 >/dev/null 2>&1 && touch $@
+	@curl --retry 10 --retry-max-time 120 -sSfL -O https://raw.githubusercontent.com/golangci/golangci-lint/$(goLintInstallerVersion)/install.sh
+	@echo "$(goLintInstallerChecksum)  install.sh" | sha256sum --check
```
Kimchelly, line 211: nit: The echo string has two spaces between the args instead of one.

Kimchelly (approved):
LGTM mod whitespace nit.

I believe you can merge this by typing an `evergreen merge` comment, but if that doesn't work, I can merge it myself.

makefile
```diff
@@ -204,7 +207,9 @@ coverageHtmlOutput := $(foreach target,$(packages),$(buildDir)/output.$(target).
 $(buildDir)/.lintSetup:$(buildDir)/golangci-lint
 	@touch $@
 $(buildDir)/golangci-lint:
-	@curl --retry 10 --retry-max-time 120 -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(buildDir) v1.51.2 >/dev/null 2>&1 && touch $@
+	@curl --retry 10 --retry-max-time 120 -sSfL -O https://raw.githubusercontent.com/golangci/golangci-lint/$(goLintInstallerVersion)/install.sh
```
Kimchelly, line 210: Sorry, I actually noticed that this puts `install.sh` in the root directory. Can it be put in `$(buildDir)/install.sh` instead?

makefile
pierwill: This is something I often see. It might even be required by some implementations of `sha256sum`.

pierwill:
evergreen merge

Kimchelly:
Guess `evergreen merge` didn't work 😢 

I'll merge the PR myself. Thanks again for your contribution!

PR #6318, opened 2023-03-13
EVG-19079 Delete tasklogs resolver

khelif96:
EVG-19079

### Description
This resolver was replaced by task.tasklogs


minnakt (approved):
Nice! I think you can delete `graphql/tests/task/queries/taskLogs-unstarted.graphql` too! There's no corresponding definition for it in the results file

PR #6321, opened 2023-03-13
EVG-17595 remove gotest.parse_json

ablack12:
EVG-17595

### Description
Verified that this isn't used in any existing (non-generated) configs.

Will create a new ticket to remove this completely after some time (along with the other commands that have been deprecated a long time), and to modify shrub / notify shrub.py.

This is already not in the project commands documentation.


Kimchelly (approved):
LGTM mod bump the agent version.

PR #6325, opened 2023-03-14
EVG-18724: fix race between merge task and PR close webhook

Kimchelly:
EVG-18724

### Description
During the merge task, it merges the PR, then attempts to conclude the merge by logging the merge event and dequeueing it from the commit queue. Merging the PR would cause it to race with the PR closing webhook, which attempts to abort PR patches and dequeue commit queue items when the PR is closed, so there would be cases where the commit queue item has already been aborted (by the PR close webhook) before the merge can be concluded.

* When aborting patches for a closed GitHub PR, skip commit queue items that are actively running the merge task or already merged.
* Ensure that concluding the merge does not rely on the item actually existing in the queue. As long as it's not in the queue anymore, it doesn't matter if concluding the merge caused the removal, or if something else (like the GitHub PR close webhook) removed it.
* Add a couple `IsFinished` helper functions to make the task models consistent.

### Testing
* Added unit tests for aborting GitHub PR patches.
* Tested that the race is fixed in staging with this PR: https://github.com/evergreen-ci/commit-queue-sandbox/pull/454

No.


rest/route/github.go
```diff
@@ -192,18 +192,6 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
 				return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "aborting patches"))
 			}
 
-			// if the item is on a commit queue, remove it
-			if err = gh.tryDequeueCommitQueueItemForPR(event.PullRequest); err != nil {
```
malikchaya2, line 196: can you please explain a bit more about why we don't want to dequeue here or check if it's still on the queue?

Kimchelly: This shouldn't be necessary because `AbortPatchesForGithubPatchData` [already checks for any commit queue items and dequeues them](https://github.com/evergreen-ci/evergreen/blob/1bba8be228ded93d63b0cea395101bd1012fedee/model/patch_lifecycle.go#L933).

malikchaya2 (approved):

PR #6327, opened 2023-03-14
EVG-19068 Remove auth from dockerfile route

hadjri:
EVG-19068

### Description

N/A


Kimchelly (commented):
I'm fine with making this a one-off exception to auth.

hadjri:
> I'm fine with making this a one-off exception to auth.

Sounds good, @ablack12 does this sound acceptable?

ablack12:
Sure, can we just add a comment (or ADR?) to explain this so we don't lose track and re-add it?

hadjri:
> Sure, can we just add a comment (or ADR?) to explain this so we don't lose track and re-add it?

Added!

Kimchelly (approved):

PR #6328, opened 2023-03-15
EVG-19028 Disallow projects to inherit enabled setting from repo project

bynn:
EVG-19028

### Description
make enabled not optional for project refs anymore and do not merge with repo ref field 
made sure that no one was using this field 

### Testing
all current tests pass 

yes, we should delete the part about being able to set repo level enabled here 
https://github.com/evergreen-ci/evergreen/wiki/Project-and-Distro-Settings#general-project-settings


model/project_ref.go
```diff
@@ -941,7 +937,10 @@ func GetNumberOfEnabledProjectsForOwnerRepo(owner, repo string) (int, error) {
 }
 
 func getNumberOfEnabledProjects(owner, repo string) (int, error) {
-	pipeline := projectRefPipelineForValueIsBool(ProjectRefEnabledKey, RepoRefEnabledKey, true)
+	pipeline := []bson.M{
+		{"$match": bson.M{ProjectRefEnabledKey: true}},
+	}
+	// pipeline := projectRefPipelineForValueIsBool(ProjectRefEnabledKey, RepoRefEnabledKey, true)
```
ablack12, line 943: should remove this

model/project_ref.go
```diff
@@ -983,6 +982,7 @@ func mergeBranchAndRepoSettings(pRef *ProjectRef, repoRef *RepoRef) (*ProjectRef
 	defer func() {
 		err = recovery.HandlePanicWithError(recover(), err, "project and repo structures do not match")
 	}()
+	// Don't merge enabled field
```
ablack12, line 985: did we mean to leave this comment in?

model/project_ref.go
```diff
@@ -1037,6 +1036,7 @@ func (p *ProjectRef) createNewRepoRef(u *user.DBUser) (repoRef *RepoRef, err err
 	repoRef.Id = mgobson.NewObjectId().Hex()
 	repoRef.RepoRefId = ""
 	repoRef.Identifier = ""
+	repoRef.Enabled = false
```
ablack12, line 1039: False is the default value; we shouldn't need this.

ablack12 (changes requested):
Are there any projects that have enabled set to nil right now?

model/project_ref.go
bynn: I thought it might be nice to see it explicitly stated but seems redundant so removing

model/project_ref.go
bynn: done

model/project_ref.go
bynn: done

bynn (commented):
the person i reached out to get rid of default enabled undid the change that we made to the project so im reaching out to them again

model/project_ref.go
ablack12: yeah there are a few fields that aren't merged so it may be unclear

ablack12 (approved):

PR #6331, opened 2023-03-16
EVG-18948 add parsley logs to task rest output

ablack12:
EVG-18948

### Description
Users have requested that parsley logs be available from rest output. 

### Testing
Added to some existing unit tests and tested on staging.

Yes, to the task object on the rest route page.


rest/model/task.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	TaskLogLinkFormat  = "%s/task_log_raw/%s/%d?type=%s"
-	EventLogLinkFormat = "%s/event_log/task/%s"
+	TaskLogLinkFormat        = "%s/task_log_raw/%s/%d?type=%s"
+	ParsleyTaskLogLinkFormat = "%s/evergreen/%s/%d/%s" // URL,taskId, execution, type
```
bynn, line 20: it looks like the comment is applicable to the taskloglinkformat so can we put it above?

bynn (approved):
it looks good 😅

rest/model/task.go
ablack12: this was actually just for my own thought process so i'll just pull it out

rest/model/task_test.go
```diff
@@ -123,6 +129,12 @@ func TestTaskBuildFromService(t *testing.T) {
 						AgentLogLink:  utility.ToStringPtr("url/task_log_raw//0?type=E"),
 						EventLogLink:  utility.ToStringPtr("url/event_log/task/"),
 					},
+					ParsleyLogs: LogLinks{
+						AllLogLink:    utility.ToStringPtr("parsley/evergreen//0/all"),
+						TaskLogLink:   utility.ToStringPtr("parsley/evergreen//0/task"),
+						SystemLogLink: utility.ToStringPtr("parsley/evergreen//0/system"),
+						AgentLogLink:  utility.ToStringPtr("parsley/evergreen//0/agent"),
```
khelif96, line 136: Does this test make sense? In what situation would there not be a taskid to populate this?

rest/model/task.go
```diff
@@ -157,14 +159,15 @@ func (ad *APIOomTrackerInfo) ToService() *apimodels.OOMTrackerInfo {
 	}
 }
 
-func (at *APITask) BuildPreviousExecutions(tasks []task.Task, url string) error {
+func (at *APITask) BuildPreviousExecutions(tasks []task.Task, url, parsleyURL string) error {
```
khelif96, line 162: nit.  maybe use an opts struct for logURL and parsleyURL?

rest/model/task_test.go
ablack12: i have no idea haha but it's an existing test; I think changing it is out of scope and it's not really harming anything

rest/model/task.go
khelif96: Can we at least rename it to logURL?

PR #6335, opened 2023-03-17
EVG-15417 add handling for repos to help text

ablack12:
EVG-15417

### Description
Realized a couple of things about the help text as I started documenting.
1) even though `evergreen retry` and `evergreen patch` technically do the same work in the backend, we only do the action based on what's enabled in the project, so separated these commands and added some context.
2) added consideration for if there are no enabled PR testing projects for the branch, but the repo is enabled, since we do allow PR testing then.

### Testing
Added unit tests

Yes added to https://github.com/evergreen-ci/evergreen/wiki/Github-Integrations

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -309,26 +309,37 @@ func (gh *githubHookApi) displayHelpText(ctx context.Context, owner, repo string
 		return errors.New("PR contains no base branch label")
 	}
 	branch := pr.Base.GetRef()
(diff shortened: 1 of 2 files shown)


malikchaya2 (changes requested):
Can you please add a screenshot of what the output of "evergreen help" looks like on a sample PR after these changes?

ablack12:
Tested using https://github.com/evergreen-ci/commit-queue-sandbox/pull/444
image

malikchaya2 (approved):

PR #6336, opened 2023-03-17
EVG-18949 surface problems not finding tasks/variants for child patch

ablack12:
EVG-18949

### Description
Surface error finding tasks/variants for child patches the way we do for parent patches.

### Testing
Tested manually https://github.com/evergreen-ci/commit-queue-sandbox/pull/455
image

No


model/project.go
```diff
@@ -1302,8 +1302,8 @@ func FindLatestVersionWithValidProject(projectId string) (*Version, *Project, *P
 				revisionOrderNum = lastGoodVersion.RevisionOrderNumber // look for an older version if the returned version is malformed
 				continue
 			}
+			return lastGoodVersion, project, pp, nil
```
ablack12, line 1305: This was an unrelated change I found when my test fails -- we shouldn't be returning the first iteration of the loop, we should return the error outside the loop. Wrote a new test to catch this.

malikchaya2 (approved):

PR #6337, opened 2023-03-20
EVG-18267 Update project creation permissions for UI

bynn:
EVG-18267

### Description
Created a new directive for the create and copy project mutations. 
The create project button will show up for anyone who is an admin for a project 
Users will not be permitted to duplicate a project that they are not admins for.

Changing this for our rest routes would require new gimlet functions so I think the work should be done in another ticket 

I also think that we shouldn't allow users to create projects on the legacy page anymore because they use a ui project route that we can get rid of

### Testing
Staging 
I can now see create project and duplicate project buttons without super user perms but can only duplicate project that I am an admin in 

Update it with doc ticket for this project

### DIFF
--- a/graphql/resolver.go
+++ b/graphql/resolver.go
@@ -25,7 +25,8 @@ func New(apiURL string) Config {
 			sc: &data.DBConnector{URL: apiURL},
 		},
 	}
-	c.Directives.RequireSuperUser = func(ctx context.Context, obj interface{}, next graphql.Resolver) (interface{}, error) {
+	c.Directives.CanCreateProject = func(ctx context.Context, obj interface{}, next graphql.Resolver) (interface{}, error) {
+		// Allow if user is superuser
 		user := mustHaveUser(ctx)
 		opts := gimlet.PermissionOpts{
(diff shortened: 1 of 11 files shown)


graphql/resolver.go
```diff
@@ -25,7 +25,8 @@ func New(apiURL string) Config {
 			sc: &data.DBConnector{URL: apiURL},
 		},
 	}
-	c.Directives.RequireSuperUser = func(ctx context.Context, obj interface{}, next graphql.Resolver) (interface{}, error) {
```
malikchaya2, line 28: Have we considered adding some documentation or warnings when a new user who is not yet an admin anywhere is added as an admin to a project? Because the user adding them probably doesn't realize that they are simultaneously giving this user the power to create new evergreen projects.

malikchaya2 (approved):
Approved with question

graphql/resolver.go
bynn: I don't think we should necessarily notify them every time but I think it would make sense to add something here 
image


and it would also be added to our docs

PR #6338, opened 2023-03-20
EVG-19186 begin instrumenting evergreen

ybrill:
[EVG-19186](https://jira.mongodb.org/browse/EVG-19186)

### Description
Start a tracer provider when the environment is being initialized. Use it to add instrumentation with a bunch of prepackaged instrumentation libraries.
- mux router
- graphql server
- mongo driver

In the ticket I'd suggested adding the AWS SDK too, but afaict the option for intercepting middleware was added in v2 of the AWS SDK. It shouldn't be a big lift to upgrade, but I wanted to split it off to keep things simple ([EVG-19211](https://jira.mongodb.org/browse/EVG-19211)).

The mux and graphql spans get combined into a single trace because the trace puts itself in the request context which is passed down the stack. The mongo traces aren't combined because they get their context from the cached application context from [anser](https://github.com/mongodb/anser/blob/main/db/wrapper.go#L20). We'll fix this too, hopefully ([EVG-19190](https://jira.mongodb.org/browse/EVG-19190)).

The admin settings are pretty basic
Screenshot 2023-03-20 at 11 58 32 AM

Some considerations:
- Calling out: this will **replace** the APM we currently use to log to Splunk. My impression is that no one really uses it (e.g. [EVG-18228](https://jira.mongodb.org/browse/EVG-18228)). Also, now that the data is in Honeycomb we can get the data from there if we want to, though we don't have the premade graphs. If someone feels that we should retain both for the time being I can try to figure out how to merge the two into a single Monitor (the driver will only take one).
- This PR uses `otel.SetTracerProvider(tp)` to put the tracer provider in global state. This is what the docs suggest (e.g. [here](https://opentelemetry.io/docs/instrumentation/go/getting-started/#installing-a-tracer-provider)). If we prefer consolidating global state in the `evergreen.Environment` we could instead keep it in there. It would be messy because we'll need to initialize the one for the db before we can ask the db for the parameters we need to set up the tracer provider, but it's _possible_.

### Testing
Sends the three types of spans from staging.

N/A


environment.go
```diff
@@ -288,7 +296,7 @@ func (e *envState) initSettings(path string) error {
 func (e *envState) initDB(ctx context.Context, settings DBSettings) error {
 	opts := options.Client().ApplyURI(settings.Url).SetWriteConcern(settings.WriteConcernSettings.Resolve()).
 		SetReadConcern(settings.ReadConcernSettings.Resolve()).
-		SetConnectTimeout(5 * time.Second).SetMonitor(apm.NewLoggingMonitor(ctx, time.Minute, apm.NewBasicMonitor(nil)).DriverAPM())
+		SetConnectTimeout(5 * time.Second).SetMonitor(otelmongo.NewMonitor())
```
julianedwards, line 299: What happens here if we have the tracer disabled?

config_tracer.go
```diff
@@ -0,0 +1,59 @@
+package evergreen
+
+import (
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+type TracerConfig struct {
```
julianedwards, line 10: Could you add some documentation here?

service/service.go
```diff
@@ -87,7 +89,10 @@ func GetRouter(as *APIServer, uis *UIServer) (http.Handler, error) {
 	uiService := uis.GetServiceApp()
 	apiService := as.GetServiceApp()
 
+	router := mux.NewRouter().UseEncodedPath()
+	router.Use(otelmux.Middleware("evergreen"))
```
julianedwards, line 93: Should this be more specific, something like `evergreen-rest`?

julianedwards (changes requested):

julianedwards:
I would also tag Kim or the app team re: the removal of the APM monitor.

config_tracer.go
ybrill: done

service/service.go
ybrill: I think it's a really generic router, though. Like this router handles just about all of the incoming traffic from REST, GraphQL, the old UI, etc.

environment.go
ybrill: If we never set up a tracer provider it will default to the [NoopTracerProvider](https://pkg.go.dev/go.opentelemetry.io/otel/api/trace#NoopTracerProvider) and the traces/spans will go nowhere.

ybrill:
@evergreen-ci/evg-app can you provide thoughts on supplanting the APM in favor of the OTEL tracing?

service/service.go
julianedwards: Sound good!

julianedwards (approved):

PR #6341, opened 2023-03-20
EVG-18896 remove $where

ybrill:
[EVG-18896](https://jira.mongodb.org/browse/EVG-18896)

### Description
My understanding of why we're using `$where` is that it's hard to match on any arbitrary key. This PR tries to accomplish the same thing with an aggregation.

Aside from wanting to disallow server side javascript in general, it makes me nervous as is because it passes unvalidated user input to get executed on the db.

### Testing
The existing tests are still passing. I'm hoping this could be a drop-in replacement.

N/A


malikchaya2 (approved):
This looks okay, but is there a manual test we can do somewhere as an extra precaution to double check that it didn't break anything?

PR #6350, opened 2023-03-23
EVG-18975 check if PR submitted by installed app

ablack12:
EVG-18975 
### Description
Because we're moving to using Github apps, can't just check if the bot is a member of the organization. Instead, check all installed apps for the org and return true if it's the right permission.

### Testing
Can't think of a good way to test end-to-end but [verified the Github API call and shape via Splunk](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22EVG-18975%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=1679588520&latest=1679590337&workload_pool=standard_perf&sid=1679591196.2404273).

Will see if we have any documentation about bot users; if not will add a stack overflow for it.

### DIFF
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -843,21 +843,30 @@ func AppAuthorizedForOrg(ctx context.Context, token, requiredOrganization, name 
 	defer utility.PutHTTPClient(httpClient)
 
 	client := github.NewClient(httpClient)
-	// We can assume no org will have more than 100 installations.
-	installations, _, err := client.Organizations.ListInstallations(ctx, requiredOrganization, &github.ListOptions{PerPage: 100})
-	if err != nil {
-		return false, err
-	}
+	opts := &github.ListOptions{PerPage: 100}
+	for {
+		installations, resp, err := client.Organizations.ListInstallations(ctx, requiredOrganization, opts)
+		if err != nil {
+			return false, err
+		}
 
-	for _, installation := range installations.Installations {
-		if installation.GetAppSlug() == name {
-			prPermission := installation.GetPermissions().GetPullRequests()
-			if prPermission == githubWrite {
-				return true, nil
+		for _, installation := range installations.Installations {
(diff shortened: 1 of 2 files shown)


thirdparty/github.go
```diff
@@ -835,6 +836,31 @@ func GithubUserInOrganization(ctx context.Context, token, requiredOrganization,
...
 
+// AppAuthorizedForOrg returns true if the given app name exists in the org's installation list,
+// and has permission to write to pull requests. Returns an error if the app name exists but doesn't have permission.
+func AppAuthorizedForOrg(ctx context.Context, token, requiredOrganization, name string) (bool, error) {
+	httpClient := getGithubClientRetry(token, "AppAuthorizedForOrg")
+	defer utility.PutHTTPClient(httpClient)
+
+	client := github.NewClient(httpClient)
+	// We can assume no org will have more than 100 installations.
+	installations, _, err := client.Organizations.ListInstallations(ctx, requiredOrganization, &github.ListOptions{PerPage: 100})
```
Kimchelly, line 847: Should we put some kind of log to alert in the future if this hits 100+ installations?

malikchaya2, line 847: Is there a way for us to maybe cache this information so that we don't have to check every time. Like maybe cache it in the project and only check if the information is not cached? 
or do we expect this to change often?

malikchaya2: We can maybe add some logic to off boarding scripts if we are worried about someone having access after leaving the company.

Kimchelly: I would avoid caching because it won't respond to GitHub-side app changes. If for example, someone deletes a bot or changes its permissions, we shouldn't be allowed to authorize patches for it.

ablack12: eh I guess I'd rather just use pagination handling than that

Kimchelly (approved):

PR #6352, opened 2023-03-23
EVG-19080 Prevent tasks query from failing on large generated tasks.

khelif96:
EVG-19080

### Description
The GetTasksByVersion query was failing despite only returning 10 entries due to a large generated_json field in each document. This projects that field out to prevent the query from OOMing.

Also passes context to the queries so they can be picked up in the OTEL traces. (Helped me narrow down the bug) 
### Testing
Tested the patch in the ticket and it now works.


ybrill (commented):
Passing through the ctx lgtm, but I don't have context on anything else

hadjri (approved):

khelif96:
evergreen merge

PR #6358, opened 2023-03-27
EVG-19103 add param and author flags to patch-file

ablack12:
EVG-19103 

### Description
To support a simpler profiling workflow, Jeff's team needs a way to re-run tasks with different parameters; we settled on creating a new patch with an existing diff, and allowing them to pass in the patch author (as the person who clicks the button to run this workflow). 

### Testing
Created https://spruce-staging.corp.mongodb.com/version/6421fb86b23736777fc2d89e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC using `egss patch-file --diff-file  test_diff_file --param hello=world --base 0c0255c06 --variants ubuntu2004 --tasks test-model --project evg --author bynn.lee` (egss -> staging shortcut). This command failed before I added permissions to myself to submit on behalf of other users.

No bc CLI command but I'll add some details to the ticket for Jeff

### DIFF
--- a/operations/commit_queue.go
+++ b/operations/commit_queue.go
@@ -29,6 +29,7 @@ const (
 	commitShaFlag       = "commit-sha"
 	commitMessageFlag   = "commit-message"
 	githubAuthorFlag    = "author"
+	patchAuthorFlag     = "author"
 
 	noCommits             = "No Commits Added"
 	commitQueuePatchLabel = "Commit Queue Merge:"
--- a/service/api_patch.go
+++ b/service/api_patch.go
@@ -35,36 +35,102 @@ type PatchAPIResponse struct {
 	Patch   *patch.Patch `json:"patch"`
 }
 
+// getAuthor returns the author for the patch. If githubAuthor or patchAuthor is provided and exists, will use this instead
+// of the submitter if the submitter is authorized to submit patches on behalf of users.
+// Returns the author, status code, and error.
+func (as *APIServer) getAuthor(data patchData, dbUser *user.DBUser, projectId, patchID string) (string, int, error) {
+	author := dbUser.Id
+	if data.GithubAuthor == "" && data.PatchAuthor == "" {
+		return author, http.StatusOK, nil
+	}
+
+	opts := gimlet.PermissionOpts{
+		Resource:      projectId,
+		ResourceType:  evergreen.ProjectResourceType,
+		Permission:    evergreen.PermissionPatches,
+		RequiredLevel: evergreen.PatchSubmitAdmin.Value,
+	}
+	if !dbUser.HasPermission(opts) {
+		return "", http.StatusUnauthorized, errors.New("user is not authorized to patch on behalf of other users")
+	}
+
+	if data.GithubAuthor != "" {
+		specifiedUser, err := user.FindByGithubName(data.GithubAuthor)
+		if err != nil {
+			return "", http.StatusInternalServerError, errors.Wrap(err, "error looking for specified author")
+		}
+		if specifiedUser != nil {
+			grip.Info(message.Fields{
+				"message":               "overriding patch author as specified by the submitter",
+				"submitter":             dbUser.Id,
+				"new_author":            specifiedUser.Id,
+				"given_github_username": data.GithubAuthor,
+				"patch_id":              patchID,
+			})
+			author = specifiedUser.Id
+		}
+		grip.DebugWhen(specifiedUser == nil, message.Fields{
+			"message":         "github user not found",
+			"github_username": data.GithubAuthor,
+			"patch_id":        patchID,
+		})
+	} else if data.PatchAuthor != "" {
+		specifiedUser, err := user.FindOneById(data.PatchAuthor)
+		if err != nil {
+			return "", http.StatusInternalServerError, errors.Wrap(err, "error looking for specified author")
+		}
+		if specifiedUser != nil {
+			grip.Info(message.Fields{
+				"message":    "overriding patch author as specified by the submitter",
+				"submitter":  dbUser.Id,
+				"new_author": data.PatchAuthor,
+				"patch_id":   patchID,
+			})
+			author = specifiedUser.Id
+		}
+		grip.DebugWhen(specifiedUser == nil, message.Fields{
+			"message":  "patch user not found",
+			"username": data.PatchAuthor,
+			"patch_id": patchID,
+		})
+	}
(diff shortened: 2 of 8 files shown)


operations/commit_queue.go
```diff
@@ -29,6 +29,7 @@ const (
 	commitShaFlag       = "commit-sha"
 	commitMessageFlag   = "commit-message"
 	githubAuthorFlag    = "author"
+	patchAuthorFlag     = "author"
```
malikchaya2, line 32: It seems like the motivation behind this ask is for me for example, to be able to run a patch that you initially created, but have it say "author: Chaya Malik" and not "author: Annie Black". 
Do we know that there is specifically the need to be able to create a patch with a completely third party author? I think we should avoid that if possible, because I'd for example feel a bit weird if patches started showing up on my my patches page that I didn't create.

malikchaya2 (changes requested):

service/api_patch.go
```diff
@@ -35,36 +35,102 @@ type PatchAPIResponse struct {
 	Patch   *patch.Patch `json:"patch"`
 }
 
+// getAuthor returns the author for the patch. If githubAuthor or patchAuthor is provided and exists, will use this instead
```
malikchaya2, line 38: Can you please specify here what "this" refers to?

service/api_patch.go
```diff
@@ -35,36 +35,102 @@ type PatchAPIResponse struct {
...
+		}
+		grip.DebugWhen(specifiedUser == nil, message.Fields{
+			"message":         "github user not found",
+			"github_username": data.GithubAuthor,
+			"patch_id":        patchID,
+		})
+	} else if data.PatchAuthor != "" {
+		specifiedUser, err := user.FindOneById(data.PatchAuthor)
+		if err != nil {
+			return "", http.StatusInternalServerError, errors.Wrap(err, "error looking for specified author")
```
malikchaya2, line 80: It might be a good idea to add the `data.PatchAuthor `to the error message.

operations/commit_queue.go
hadjri: I suppose we could filter this out of the "My patches" page (or allow to toggle the viewing of them), or we could label the patches in the frontend in a way that makes it clear that they were submitted on your behalf.

service/api_patch.go
```diff
@@ -35,36 +35,102 @@ type PatchAPIResponse struct {
...
+		if specifiedUser != nil {
+			grip.Info(message.Fields{
+				"message":    "overriding patch author as specified by the submitter",
+				"submitter":  dbUser.Id,
+				"new_author": data.PatchAuthor,
+				"patch_id":   patchID,
+			})
+			author = specifiedUser.Id
+		}
+		grip.DebugWhen(specifiedUser == nil, message.Fields{
```
hadjri, line 91: If someone uses a nonexistent author name or has a typo, can we propagate down to the CLI output that the user was not found? It seems that with the exception of this debug statement it would fail silently and the author name would default to the submitter's username.

operations/commit_queue.go
ablack12: The ask here is because this command will be run by a bot user -- DAG is going to add a button to their tooling that calls this command. Rather than having all new patches created be authored by the bot user, they'd like the author to be the person who hits the button. This is a highly gated permission though, and similar to the flag we already have for the commit queue bot https://github.com/evergreen-ci/evergreen/blob/4afe0f066dd383e47d54cac87858ba7d3d417f11/operations/commit_queue.go#L175. I can open a new ticket to add a Submitter field or something to display also who submitted the patch, but I'd like to avoid making that part of this work.

service/api_patch.go
ablack12: well this error would be hit if it's a database error, not if the user isn't found, so it doesn't matter too much but I can add it

service/api_patch.go
ablack12: The behavior of this command for the commit queue and the desired behavior for this change is that it would default to the author name if the submitter name isn't configured. This behavior is in the flag description https://github.com/evergreen-ci/evergreen/blob/4afe0f066dd383e47d54cac87858ba7d3d417f11/operations/patch.go#L290

operations/commit_queue.go
hadjri: I see, if it's primarily intended for bot users then I'm fine to keep things as is. As for the ticket, I'm ambivalent on whether it would be helpful - I'll defer to your decision

service/api_patch.go
hadjri: Does the CLI output mention if the user was found or not if an author name is provided? Or is the only way for us to find out if the author override was successful to check the submitter field in the UI for the patch?

My only concern here was that if there isn't any direct CLI feedback that a default / fallback to the submitter is taking place.

hadjri: Does the CLI output mention if the user was found successfully if an author name is provided? Or is the only way for us to find out if the author override was successful to check the submitter field in the UI for the patch?

My only concern here was that if there isn't any direct CLI feedback that a default / fallback to the submitter is taking place.

hadjri: Does the CLI output mention if the user was found successfully if an author name is provided? Or is the only way for us to find out if the author override was successful to check the submitter field in the UI for the patch?

My only concern here was that if there isn't any direct CLI feedback that a default / fallback to the submitter is taking place.

hadjri (commented):
LGTM although just have one last consideration to bring up.

operations/commit_queue.go
malikchaya2: If the author is intended to be the user who clicked the button, then I think that makes, as they will expecting to see it in their my patches page. I don't think it's necessary in that case.

malikchaya2 (approved):

service/api_patch.go
ablack12: The CLI output doesn't mention or know about whether it's succeeded yet, bc this logic is happening on the app and not on the client. I'm still fine with that since this permission is only given to bots (i.e. this permission isn't grantable on mana or given to project admins or superusers) that probably aren't checking if it ended up defaulting anyway.

operations/commit_queue.go
ablack12: Actually in the spirit of not creating tickets we don't think we'll do I'll wait to see if this ends up being needed - it hasn't been for the commit queue version of the flag.

PR #6360, opened 2023-03-29
EVG-19270 Pass context to mainline commits db queries

khelif96:
EVG-19270

### Description
Dove down a honeycomb rabbit hole and wanted to instrument the mainline commits query to get better performance insights. Passes context to the db calls so they can be tracked as part of the same span.


ybrill (approved):
This is exciting

PR #6361, opened 2023-03-29
EVG-18934 Enable adding links to task annotations

hadjri:
EVG-18934

### Description
To enhance the task metadata sidebar, we're aiming to enable users to provide link titles and corresponding URLs representing build failures to their task annotations. Currently this is implemented in a way that only enables a single link, but since the data structure is a list, we could increase the limit of this list's length if needed in the future.

This change includes adding a new `task_links`, to the `TaskAnnotation` struct that accepts a description string and a URL string. For validation, we're leveraging our existing CORS list to evaluate the URLs.
### Testing
Added unit tests and GraphQL tests
I'm not aware of docs covering task annotations and their usage in the "Failure details" tab in Spruce, if they don't exist they can be added as part of this ticket.


model/annotations/task_annotations.go
```diff
@@ -25,6 +25,16 @@ type TaskAnnotation struct {
 	SuspectedIssues []IssueLink `bson:"suspected_issues,omitempty" json:"suspected_issues,omitempty"`
 	// links to tickets created from the task using a custom web hook
 	CreatedIssues []IssueLink `bson:"created_issues,omitempty" json:"created_issues,omitempty"`
+	// non-ticket links to be displayed in the UI metadata sidebar
+	TaskLinks []TaskLink `bson:"task_links,omitempty" json:"task_links,omitempty"`
```
ablack12, line 29: nit: technically these could still be tickets, so I would just say links in the comment. Also I know I used task_links in the ticket but maybe MetadataLinks is more specific? Whatever works though

rest/model/task_annotations.go
```diff
@@ -203,3 +250,18 @@ func ValidateIssues(issues []APIIssueLink) error {
 	}
 	return catcher.Resolve()
 }
+
+func ValidateTaskLinks(links []APITaskLink) error {
```
ablack12, line 254: Can this be moved to the model/annotations package so it can be used by the UI also? Should have a function comment also.

rest/model/task_annotations.go
```diff
@@ -203,3 +250,18 @@ func ValidateIssues(issues []APIIssueLink) error {
...
+	settings := evergreen.GetEnvironment().Settings()
+	catcher := grip.NewBasicCatcher()
+	for _, link := range links {
+		catcher.Add(util.CheckURL(utility.FromStringPtr(link.URL)))
+		if !utility.StringMatchesAnyRegex(utility.FromStringPtr(link.URL), settings.Ui.CORSOrigins) {
+			catcher.Errorf("task link URL '%s' must match a CORS origin", utility.FromStringPtr(link.URL))
+		}
+		if utility.FromStringPtr(link.Text) == "" {
```
ablack12, line 262: Can we validate that this isn't too long?

rest/model/task_annotations.go
```diff
@@ -203,3 +250,18 @@ func ValidateIssues(issues []APIIssueLink) error {
 	}
 	return catcher.Resolve()
 }
+
+func ValidateTaskLinks(links []APITaskLink) error {
+	settings := evergreen.GetEnvironment().Settings()
+	catcher := grip.NewBasicCatcher()
+	for _, link := range links {
```
ablack12, line 257: Shouldn't we validate here that there's only one as well?

ablack12 (changes requested):

ablack12:
There is some documentation here https://docs.devprod.prod.corp.mongodb.com/evergreen/REST-V2-Usage/#task-annotations

model/annotations/task_annotations.go
hadjri: Sure, fine to rename to MetadataLinks

malikchaya2 (commented):
Task annotations are shown in a "failure details" tab. If this will be shown in the Task Metadata sidebar, what makes this be considered a "task annotation"?

graphql/pod_resolver.go
```diff
@@ -10,13 +10,29 @@ import (
...
 	"github.com/evergreen-ci/evergreen/rest/model"
-	restModel "github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/evergreen-ci/utility"
 )
 
-// Type is the resolver for the type field.
-func (r *podResolver) Type(ctx context.Context, obj *model.APIPod) (string, error) {
-	return string(obj.Type), nil
+// Events is the resolver for the events field.
+func (r *podResolver) Events(ctx context.Context, obj *model.APIPod, limit *int, page *int) (*PodEvents, error) {
```
malikchaya2, line 17: Was this meant to be part of this PR?

malikchaya2 (changes requested):

graphql/pod_resolver.go
hadjri: This is a reorganization of the file that was needed to make `make gqlgen` succeed, which was necessary as part of this PR.

rest/model/task_annotations.go
hadjri: Refactored and moved to model package.

rest/model/task_annotations.go
hadjri: Sure, setting the limit at 1024 characters initially but fine to restrict it further if needed.

hadjri:
> Task annotations are shown in a "failure details" tab. If this will be shown in the Task Metadata sidebar, what makes this be considered a "task annotation"?

I believe it's also meant to be shown in the "failure details" panel, just also popping up in the sidebar for better visibility.

malikchaya2 (approved):

rest/model/task_annotations.go
khelif96: I would vote to restrict it to only 40 characters.
image

model/annotations/task_annotations.go
```diff
@@ -313,3 +350,26 @@ func AddCreatedTicket(taskId string, execution int, ticket IssueLink, userDispla
...
+func ValidateMetadataLinks(links ...MetadataLink) error {
+	settings := evergreen.GetEnvironment().Settings()
+	catcher := grip.NewBasicCatcher()
+	if len(links) > MaxMetadataLinks {
+		catcher.Errorf("cannot have more than %d task links per annotation", MaxMetadataLinks)
+	}
+	for _, link := range links {
+		catcher.Add(util.CheckURL(link.URL))
+		if !utility.StringMatchesAnyRegex(link.URL, settings.Ui.CORSOrigins) {
+			catcher.Errorf("task link URL '%s' must match a CORS origin", link.URL)
```
khelif96, line 365: I don't think we should do this. The CORS origins are a security feature that we use to ensure we aren't allowing different websites to directly interact with our API. We shouldn't make task annotation links be dependent on it. 
e.g. If a user wants to add a new link we shouldn't also be required to add them to our CORS origins list.

Since creating annotation links requires auth anyways that should be enough validation.

khelif96 (changes requested):

graphql/schema/mutation.graphql
```diff
@@ -26,6 +26,16 @@ type Mutation {
...
+  addAnnotationMetadataLink(
+    taskId: String!
+    execution: Int!
+    apiMetadataLink: MetadataLinkInput!
+  ): Boolean!
+  removeAnnotationMetadataLink(
+    taskId: String!
+    execution: Int!
+    apiMetadataLink: MetadataLinkInput!
+  ): Boolean!
```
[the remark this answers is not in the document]
ablack12: I was also under this assumption, although I suppose there's really no reason not to have it be consistent with everything else in a task annotation and have it be editable on the failures detail tab by people with correct permissions, especially since the work is half done now.

rest/model/task_annotations.go
```diff
@@ -170,6 +195,28 @@ func ArrAPIIssueLinkArrtaskannotationsIssueLink(t []APIIssueLink) []annotations.
 	return m
 }
 
+func ArrtaskannotationsMetadataLinkArrAPIMetadataLink(t []annotations.MetadataLink) []APIMetadataLink {
```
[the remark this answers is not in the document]
hadjri: Short for "array"

graphql/schema/mutation.graphql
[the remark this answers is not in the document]
hadjri: Yes my assumption was that both would be supported 🤷

model/annotations/task_annotations.go
khelif96: For the parsley case which wanted to allow arbitrary files, 
Anyone can generate a link that resembles the one below. There is no auth mechanism in place for generating arbitrary links because anyone can slap together a URL and share it.
`https://parsley.mongodb.com/userSuppliedLog=someShadyLogFile.com`

For this scenario, only privileged users can set these links because they need API access in order to upload them. So the links are inherently safer because of the assumption that if a user has API access we already trust them. 

If we wanted to provide an allow list for URLs using the CORs origins is not a good idea either way. The CORs origins [are used to prevent other websites from directly making requests to Evergreen's API](https://developer.mozilla.org/en-US/docs/Glossary/CORS). So if we start expanding this list to allow any arbitrary URL that a user wants to upload as a task URL we also expose evergreen to requests from those other URLs.

hadjri: I see, I guess since this is restricted to authed users I'm fine to remove this bit. Or we could introduce an "allow" list as Mohamed mentioned but maybe that'd be best as a followup to this after we determined we need it? @ablack12

hadjri: Went ahead and removed it.

rest/model/task_annotations.go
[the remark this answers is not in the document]
ablack12: I believe this is auto generated, right? (although i think this should be ArrTaskAnnotations) to be proper camelcase

model/annotations/task_annotations.go
```diff
@@ -125,6 +137,27 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	return errors.Wrapf(err, "updating note for task '%s'", taskId)
 }
 
+// AddMetadataLinkToAnnotation adds a task link to the annotation for the given task.
+func AddMetadataLinkToAnnotation(taskId string, execution int, metadataLink MetadataLink) error {
+	_, err := db.Upsert(
+		Collection,
+		ByTaskIdAndExecution(taskId, execution),
+		bson.M{
+			"$push": bson.M{MetadataLinksKey: metadataLink},
```
ablack12, line 146: preference for addToSet to avoid dups.

model/annotations/task_annotations.go
```diff
@@ -313,3 +349,22 @@ func AddCreatedTicket(taskId string, execution int, ticket IssueLink, userDispla
...
+	catcher := grip.NewBasicCatcher()
+	if len(links) > MaxMetadataLinks {
+		catcher.Errorf("cannot have more than %d task links per annotation", MaxMetadataLinks)
+	}
+	for _, link := range links {
+		catcher.Add(util.CheckURL(link.URL))
+		if link.Text == "" {
+			catcher.Errorf("link text cannot be empty")
+		}
+		if len(link.Text) > MaxMetadataTextLength {
```
ablack12, line 365: nit: this can be else if

model/annotations/task_annotations.go
```diff
@@ -125,6 +137,27 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
...
+	)
+	return errors.Wrapf(err, "updating task links for task '%s'", taskId)
+}
+
+// RemoveMetadataLinkFromAnnotation removes a task link from the annotation for the given task.
+func RemoveMetadataLinkFromAnnotation(taskId string, execution int, metadataLink MetadataLink) error {
+	return db.Update(
+		Collection,
+		ByTaskIdAndExecution(taskId, execution),
+		bson.M{"$pull": bson.M{MetadataLinksKey: metadataLink}},
```
ablack12, line 157: nit: pull should be on its own line for consistency

model/annotations/task_annotations.go
```diff
@@ -313,3 +349,22 @@ func AddCreatedTicket(taskId string, execution int, ticket IssueLink, userDispla
 	)
 	return errors.Wrapf(err, "adding ticket to task '%s'", taskId)
 }
+
+// ValidateMetadataLinks will validate the given metadata links, ensuring that they are valid URLs, that they match Evergreen's approved
```
ablack12, line 353: we should remove the comment about CORS

graphql/mutation_resolver.go
```diff
@@ -108,6 +108,37 @@ func (r *mutationResolver) RemoveAnnotationIssue(ctx context.Context, taskID str
...
+func (r *mutationResolver) AddAnnotationMetadataLink(ctx context.Context, taskID string, execution int, apiMetadataLink restModel.APIMetadataLink) (bool, error) {
+	metadataLink := restModel.APIMetadataLinkToService(apiMetadataLink)
+	if err := annotations.ValidateMetadataLinks(*metadataLink); err != nil {
+		return false, InputValidationError.Send(ctx, fmt.Sprintf("invalid metadata link: %s", err.Error()))
+	}
+	annotation, err := annotations.FindOneByTaskIdAndExecution(taskID, execution)
+	if err != nil {
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("finding annotation: %s", err.Error()))
+	}
+	if annotation == nil {
```
ablack12, line 121: Should this be an error? What if there isn't an annotation yet, wouldn't we just want to add one? (Similar to how the add issue mutations work)

model/annotations/task_annotations.go
khelif96: +1

graphql/mutation_resolver.go
```diff
@@ -108,6 +108,37 @@ func (r *mutationResolver) RemoveAnnotationIssue(ctx context.Context, taskID str
 	}
 }
 
+// AddAnnotationMetadataLink is the resolver for the addAnnotationMetadataLink field.
+func (r *mutationResolver) AddAnnotationMetadataLink(ctx context.Context, taskID string, execution int, apiMetadataLink restModel.APIMetadataLink) (bool, error) {
```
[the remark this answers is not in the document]
khelif96: Assuming we are confident we won't change our minds then one mutation resolver that lets you edit the link makes sense to me. 

If that's the case we should update this on the DB side as well to accept a field instead of an array.

rest/route/annotations.go
```diff
@@ -258,7 +258,7 @@ func (h *bulkPatchAnnotationHandler) Run(ctx context.Context) gimlet.Responder {
 // PUT /rest/v2/tasks/{task_id}/annotation
 
 // Parsing logic for task annotation put and patch routes.
-func annotationByTaskPutOrPatchParser(ctx context.Context, r *http.Request) (string, *restModel.APITaskAnnotation, error) {
+func annotationByTaskPutOrPatchParser(r *http.Request) (string, *restModel.APITaskAnnotation, error) {
```
khelif96, line 261: Threading context here allows us to capture this as part of one trace.  I think we should keep it. Jonathan is planning on doing some work to thread context for allot of our requests and db calls.

model/annotations/task_annotations.go
```diff
@@ -125,6 +137,27 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	return errors.Wrapf(err, "updating note for task '%s'", taskId)
 }
 
+// AddMetadataLinkToAnnotation adds a task link to the annotation for the given task.
+func AddMetadataLinkToAnnotation(taskId string, execution int, metadataLink MetadataLink) error {
```
khelif96, line 141: nit. If we thread context to the db call below we can capture these in an otel trace.

graphql/mutation_resolver.go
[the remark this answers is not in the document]
hadjri: If we still want to enable extending the number of links a user can add in the future I'm fine to do this while keeping the field on the annotation as a list struct.  I assume what makes the most sense is accepting a list of `MetadataLink` into this resolver.

model/annotations/task_annotations.go
```diff
@@ -125,6 +139,16 @@ func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMess
 	return errors.Wrapf(err, "updating note for task '%s'", taskId)
 }
 
+// SetAnnotationMetadataLinks sets the metadata links for a task annotation.
+func SetAnnotationMetadataLinks(ctx context.Context, taskId string, execution int, metadataLinks ...MetadataLink) error {
+	query := ByTaskIdAndExecution(taskId, execution)
```
ablack12, line 144: Looking at the functions below, it seems like we may need to update source to be consistent with how annotations are used. cc @malikchaya2

ablack12 (approved):

PR #6365, opened 2023-03-29
EVG-19167: Add origin field to GraphQL Splunk logging

sophstad:
EVG-19167

### Description
- Add origin field to Splunk logs
- Also fix existing error thrown when running `make gqlgen`: `PodAPIEventLogEntry` was being referenced using the `restModel` package when it actually resides in `model`.

### Testing
Verified on staging:
Screenshot 2023-03-29 at 3 04 15 PM


hadjri (approved):

khelif96 (approved):
Thanks for fixing this!

PR #6367, opened 2023-03-30
EVG-18971 fix retry logic for merging cq items

ablack12:
EVG-18971
### Description
Attempted to fix this already by using getGithubClientRetry but didn't realize this only retries in certain cases. Adding an explicit retry to the merge command will more reliably always retry on error. 

### Documentation
added some comments

### DIFF
--- a/agent/command/git_merge_pr.go
+++ b/agent/command/git_merge_pr.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/evergreen-ci/evergreen/thirdparty"
+	"github.com/evergreen-ci/utility"
 
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/agent/internal"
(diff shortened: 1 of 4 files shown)


agent/command/git_merge_pr.go
```diff
@@ -19,17 +20,23 @@ import (
...
 	Token string `mapstructure:"token"`
 
 	statusSender send.Sender
 	base
 }
 
-func gitMergePrFactory() Command   { return &gitMergePr{} }
-func (c *gitMergePr) Name() string { return "git.merge_pr" }
+const (
+	mergePRAttempts      = 3
```
bynn, line 31: should we use these settings since theyre defined already? 

https://github.com/evergreen-ci/evergreen/blob/2e726d4bdd0c0bede32f6e1f9ef9887f8392616a/thirdparty/github.go#L28

bynn (changes requested):

bynn (approved):
makes sense to me!

PR #6369, opened 2023-03-30
EVG-7440 Refresh github PR statuses when restarting builds and versions

hadjri:
EVG-7440

### Description

Previous PRs focused on getting version/build restarts to properly update PR statuses, and refactoring PR status logic out of `UpdatePatchStatus` into its own function.

During this process I spotted what seemed to be an existing bug, since `UpdateVersionAndPatchStatusForBuilds` was only updating [a single build per version](https://github.com/evergreen-ci/evergreen/blob/main/model/task_lifecycle.go#L1592) since the `versionsToUpdate ` map would constantly be overwritten in the linked loop, which is incorrect. Therefore in previous PRs, this was changed to make it such that each version would send a PR status update for all relevant builds, not just one single build. However, this function is called when aborting PR patches due to a new PR commit, and we saw a spike in our PR status update API usage since patches with 25-100 builds were now using 25-100x more of their previous PR status update API usage.

I think, since the logic in `UpdateVersionAndPatchStatusForBuilds` already does not make sense, and "fixing" it causes a large spike in Github API usage, it should be removed altogether since I don't think it's providing us any value. I've made the focus of this change to simply do a refactor and fix the behavior of restarting versions & builds to get them to update the PR statuses, which should not be a large API load increase.

- Modified `restartTasks` (which is used by `RestartVersion` and `RestartBuild`) to send github PR updates
- Separated updating an Evergreen build's PR status logic from `UpdatePatchStatus` into its own function
- Removed Github checks subscriptions logic from `UpdateVersionAndPatchStatusForBuilds`
- Function renaming

### Testing
In staging, confirmed that restarting builds and versions reset PR statuses as expected, and also confirmed that our GitHub checks API was not being called when old PR patches got aborted due to a new PR commit.
### Documentation
This shouldn't affect external facing docs so N/A


model/task_lifecycle.go
```diff
@@ -1425,6 +1440,12 @@ func UpdateBuildAndVersionStatusForTask(t *task.Task) error {
 		return errors.Wrapf(err, "updating version '%s' status", taskVersion.Id)
 	}
 
+	if newVersionStatus != taskVersion.Status {
```
ablack12, line 1443: Can we check more specifically that the old status was finished and the new status is not? Wondering if otherwise we'll hit these everytime we go from scheduled to in progress or something.

ablack12:
LGTM! We should make a point this time around to monitor how the github API limit is affected and update the ticket with that before closing it out.

hadjri:
> LGTM! We should make a point this time around to monitor how the github API limit is affected and update the ticket with that before closing it out.

Definitely. Will be monitoring API usage before closing

PR #6370, opened 2023-03-31
EVG-19296 skip notifications only for unfinished system unresponsive tasks

ablack12:
EVG-19296

### Description
This unblocks some critical server work.

Right now we skip notifications for all system unresponsive tasks with the idea that they could succeed on future retries, and that this can't be a user's fault. However if a task does fail all 10 runs it is likely that it's user error and it seems like at that time we should send a notification, since we won't be retrying anymore. 

### Testing
Modified unit test

### Documentation
Couldn't find something existing so added a line

### DIFF
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -425,7 +425,7 @@ func (t *taskTriggers) taskFailure(sub *event.Subscription) (*notification.Notif
 		return nil, nil
 	}
 
-	if t.task.IsSystemUnresponsive() {
+	if !t.task.IsFinishedSystemUnresponsive() {
 		return nil, nil
 	}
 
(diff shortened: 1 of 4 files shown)


trigger/task.go
```diff
@@ -547,12 +548,7 @@ func (t *taskTriggers) taskRegression(sub *event.Subscription) (*notification.No
 
 func isTaskRegression(sub *event.Subscription, t *task.Task) (bool, *alertrecord.AlertRecord, error) {
 	if t.Status != evergreen.TaskFailed || !utility.StringSliceContains(evergreen.SystemVersionRequesterTypes, t.Requester) ||
-		t.IsSystemUnresponsive() {
-		return false, nil, nil
-	}
-
-	// Regressions are not actionable if they're caused by a host that was terminated or an agent that died
-	if t.Details.Description == evergreen.TaskDescriptionStranded || t.Details.Description == evergreen.TaskDescriptionHeartbeat {
```
malikchaya2, line 555: why was this check removed?

malikchaya2 (approved):

PR #6372, opened 2023-04-03
EVG-19108: fix commit queue removal behavior

Kimchelly:
EVG-19108

### Description
The functionality to remove commit queue items used to just remove the one commit queue item, but it didn't consider the effect of that removal on later items in the commit queue - if the commit is already running in a batch, deleting that item will end up failing the later ones since their merge tasks have dependencies on each o ther. The desired behavior should be to dequeue just the one item from the batch and restart the items running after it. This is what's already done when a commit queue task fails, so I just made it apply to users manually removing whole items themselves.

* Refactor to generalize logic of dequeueing commit queue at the end of tasks so it can apply to manually removing commit queue items.
* Fix logic for manual commit queue item removal to handle items running in batches.
* Remove [a piece of code](https://github.com/evergreen-ci/evergreen/pull/6372/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9L916) that unnecessarily sends a GitHub status a second time. Dequeueing already sends a GitHub status [here](https://github.com/evergreen-ci/evergreen/pull/6372/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R868).
* Slightly improve user-facing error messages when commit queue items are dequeued.

### Testing
Added unit tests and tested in the commit queue sandbox in staging. I tried dequeueing an item that wasn't yet finalized, which simply removed the item as expected. I also tried dequeueing an item in the middle of a running batch, which removed just the one item and restarted the items after it that were running in the batch.

### Documentation
N/A


model/task_lifecycle.go
```diff
@@ -775,48 +775,104 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
+	mergeErrMsg := fmt.Sprintf("commit queue item '%s' is being dequeued: %s", t.Version, reason)
+	if t.Details.Type == evergreen.CommandTypeSetup {
+		// If the commit queue merge task failed on setup, there is likely a merge conflict.
+		mergeErrMsg = "Merge task failed on setup, which likely means a merge conflict was introduced. Please try merging with the base branch."
+	}
+
+	_, err := dequeueAndRestartItem(dequeueAndRestartOptions{
+		cq:           cq,
+		projectID:    t.Project,
+		versionID:    t.Version,
```
hadjri, line 790: suggestion to rename this to `itemVersionID` to make it clear that the item is represented by a version here

model/task_lifecycle.go
```diff
@@ -775,48 +775,104 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
+		mergeErrMsg:  mergeErrMsg,
+		githubStatus: githubState,
+	})
+	return err
+}
+
+// DequeueAndRestartForVersion restarts all items after the commit queue
+// item, aborts/dequeues this version, and sends an updated status to GitHub. If
+// it succeeds, it returns the removed item.
+func DequeueAndRestartForVersion(cq *commitqueue.CommitQueue, project, version, user, reason string) (*commitqueue.CommitQueueItem, error) {
```
hadjri, line 803: Can we consolidate this with `DequeueAndRestartForTask` to a single `DequeueAndRestartItem` function we can refer to for this behavior? Since it seems they do the same thing mod a little bit of task / error message handling if a task has been passed in.

model/task_lifecycle.go
```diff
@@ -775,48 +775,104 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
+		return nil, errors.Wrapf(err, "finding patch '%s'", opts.versionID)
 	}
 	if p == nil {
-		return errors.Errorf("patch '%s' not found", t.Version)
+		return nil, errors.Errorf("patch '%s' not found", opts.versionID)
 	}
-	if err := tryDequeueAndAbortCommitQueueVersion(p, *cq, t, caller); err != nil {
-		return err
+
+	removed, err := tryDequeueAndAbortCommitQueueVersion(p, *opts.cq, opts.taskID, opts.mergeErrMsg, opts.caller)
```
hadjri, line 852: Since we've been referring to them as items for the most part in the function names can we rename this to `tryDequeueAndAbortCommitQueueItem`?

hadjri (changes requested):

model/task_lifecycle.go
```diff
@@ -775,48 +775,104 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
+		githubStatus: githubState,
+	})
+	return err
+}
+
+// DequeueAndRestartForVersion restarts all items after the commit queue
+// item, aborts/dequeues this version, and sends an updated status to GitHub. If
+// it succeeds, it returns the removed item.
+func DequeueAndRestartForVersion(cq *commitqueue.CommitQueue, project, version, user, reason string) (*commitqueue.CommitQueueItem, error) {
+	return dequeueAndRestartItem(dequeueAndRestartOptions{
```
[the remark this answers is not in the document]
Kimchelly: We could, but it seems easier to reason about what you need to pass in when you have a more understandable publicly-exported function that enumerates all the options you must pass in and in what combination.

model/task_lifecycle.go
Kimchelly: I usually prefer making two different functions and calling a helper when it simplifies the complexity of reading it. My experience in the Evergreen code base is that too much functionality gets merged and consolidated to reduce the number of function definitions, but at the expense of increasing complexity substantially (and therefore producing misunderstandings and bugs). If possible, I prefer to lean towards obvious and easy-to-use function calls, as opposed to a few function calls that are complicated and you really have to dig to understand how to call it properly.

I did try consolidating the two initially, but they're annoyingly just a bit different enough that merging them into one function would just be more confusing. In the case where you're dequeueing an entire version, you would have to pass a nil task and know that it's safe to pass a nil task, and understand why you would pass a task vs. just a version. You could alternatively push the task-specific logic into the callers, but that doesn't seem any simpler than having two functions. Splitting them in two makes it obvious that the task one is handling dequeue after a commit queue task fails, whereas the other one is for removing entire versions for reasons not realted to tasks. This reduces overhead of understanding what conceptually is being done in the two different cases.

model/task_lifecycle.go
Kimchelly: Done.

model/task_lifecycle.go
```diff
@@ -775,48 +775,104 @@ func RestartItemsAfterVersion(cq *commitqueue.CommitQueue, project, version, cal
...
-	if err := tryDequeueAndAbortCommitQueueVersion(p, *cq, t, caller); err != nil {
-		return err
+
+	removed, err := tryDequeueAndAbortCommitQueueVersion(p, *opts.cq, opts.taskID, opts.mergeErrMsg, opts.caller)
+	if err != nil {
+		return nil, errors.Wrapf(err, "dequeueing and aborting commit queue item '%s'", opts.versionID)
 	}
+
 	grip.Info(message.Fields{
+		"message":      "commit queue item was dequeued and later items were restarted",
```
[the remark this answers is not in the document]
Kimchelly: Done, I added a couple fields.

model/task_lifecycle.go
Kimchelly: Done.

model/task_lifecycle.go
hadjri: Alright, that's a fair point, I'm on board. Thanks for the thorough response

hadjri (approved):
LGTM mod merge conflict resolution

PR #6373, opened 2023-04-04
EVG-19176  Add “verbose mode” to git clone retries

malikchaya2:
EVG-19176

### Description
This runs git clone in verbose mode on the last two retries. I only added it on the last two because if it only retries 2 or 3 times and passes, then I'm guess that the verbose logs are not necessary. And if it does retry the full five times and then fails, it's probably nice to have two iterations of verbose logs to look at in case there's any variation between them. I'm open to feedback on when we should start using verbose mode. 

### Testing
Tested on staging 

### Documentation
NA

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -74,6 +74,7 @@ type cloneOpts struct {
 	cloneParams        string
 	recurseSubmodules  bool
 	mergeTestRequester bool
+	useVerbose         bool
 	cloneDepth         int
 }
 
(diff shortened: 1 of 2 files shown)


agent/command/git.go
```diff
@@ -183,6 +184,9 @@ func (opts cloneOpts) buildHTTPCloneCommand() ([]string, error) {
 	if opts.recurseSubmodules {
 		clone = fmt.Sprintf("%s --recurse-submodules", clone)
 	}
+	if opts.useVerbose {
+		clone = fmt.Sprintf(" GIT_TRACE=1 GIT_CURL_VERBOSE=1 %s", clone)
```
Kimchelly, line 188: nit: remove leading space before `GIT_TRACE`.

agent/command/git.go
```diff
@@ -208,6 +212,9 @@ func (opts cloneOpts) buildSSHCloneCommand() ([]string, error) {
 	if opts.recurseSubmodules {
 		cloneCmd = fmt.Sprintf("%s --recurse-submodules", cloneCmd)
 	}
+	if opts.useVerbose {
+		cloneCmd = fmt.Sprintf(" GIT_TRACE=1 GIT_CURL_VERBOSE=1 %s", cloneCmd)
```
Kimchelly, line 216: Same thing as above about leading space.

agent/command/git.go
malikchaya2: done

agent/command/git.go
malikchaya2: done

Kimchelly (approved):

PR #6376, opened 2023-04-04
EVG-19306 Do not look for decommissioned pods in the terminated job

hadjri:
EVG-19306

### Description
We introduced a [safety measure](https://github.com/evergreen-ci/evergreen/pull/5908) to ensure that a pod can only run one single task in its lifecycle, whereby we mark the pod as decommissioned immediately upon setting its running task.  However, our cron job that looks for pods to terminate looks for decommissioned pods, meaning that pods marked as decommissioned that were actively running tasks would match the termination filter, resulting in pods prematurely getting terminated with running tasks.

Removed this from the filter, since it appears that a decommissioned pod will still properly get terminated once it [asks for another task](https://gist.github.com/hadjri/174b7f5a5b34a2781a73ed0cc63199a4#file-pod_agent-go-L242)
### Testing
Ran a container patch with this change and saw tasks finally completing successfully
### Documentation
n/a

### DIFF
--- a/model/pod/db.go
+++ b/model/pod/db.go
@@ -121,9 +121,6 @@ func FindByNeedsTermination() ([]Pod, error) {
 				StatusKey: StatusStarting,
 				bsonutil.GetDottedKeyName(TimeInfoKey, TimeInfoStartingKey): bson.M{"$lte": staleCutoff},
 			},
-			{
-				StatusKey: StatusDecommissioned,
-			},
 		},
 	}))
 }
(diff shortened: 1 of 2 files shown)


model/pod/db.go
```diff
@@ -121,9 +121,6 @@ func FindByNeedsTermination() ([]Pod, error) {
 				StatusKey: StatusStarting,
 				bsonutil.GetDottedKeyName(TimeInfoKey, TimeInfoStartingKey): bson.M{"$lte": staleCutoff},
 			},
-			{
-				StatusKey: StatusDecommissioned,
```
Kimchelly, line 125: We have to leave this condition in. It should be sufficient to add a condition to check that the current running task is empty.

Kimchelly (changes requested):

Kimchelly (changes requested):

PR #6380, opened 2023-04-05
EVG-18351 succeed xunit_results with valid xml

ablack12:
EVG-18351
### Description
Cloud has asked that valid xml files passed to xunit_results be considered valid, even if there are no test results. This seems reasonable. We introduced erroring on no found tests in EVG-5249 to handle invalid files, however this functionality is achievable without checking test results specifically.

### Testing
Added unit test for both cases.


agent/command/results_xunit.go
```diff
@@ -168,9 +175,8 @@ func (c *xunitResults) parseAndUploadResults(ctx context.Context, conf *internal
 			cumulative = addTestCasesForSuite(suite, idx, conf, cumulative)
 		}
 	}
-
-	if len(cumulative.tests) == 0 {
-		return errors.New("no test results found")
+	if len(reportFilePaths) == numInvalid {
```
ablack12, line 178: Can't totally decide if we should include this error but it seems consistent.

agent/command/results_xunit.go
```diff
@@ -108,6 +108,10 @@ func getFilePaths(workDir string, files []string) ([]string, error) {
 	if catcher.HasErrors() {
 		return nil, errors.Wrapf(catcher.Resolve(), "%d incorrect file specifications", catcher.Len())
 	}
+	// Only error for no files if the user provided files.
+	if len(out) == 0 && len(files) > 0 {
+		return nil, errors.New("Files provided but no files found")
```
Kimchelly, line 113: Suggestion: "Files parameter was provided but no XML files matched"

agent/command/results_xunit.go
```diff
@@ -168,9 +175,8 @@ func (c *xunitResults) parseAndUploadResults(ctx context.Context, conf *internal
 			cumulative = addTestCasesForSuite(suite, idx, conf, cumulative)
 		}
 	}
-
-	if len(cumulative.tests) == 0 {
-		return errors.New("no test results found")
+	if len(reportFilePaths) == numInvalid {
+		return errors.New("all given file paths are invalid")
```
Kimchelly, line 179: I imagine users wouldn't understand tihs. Can this specifically state that it's because the matching files were either directories or nonexistent?

Kimchelly (changes requested):

agent/command/results_xunit.go
ablack12: I'm not too concerned here since we log if it's one of the above problems

Kimchelly (approved):
LGTM mod bump the agent version.

PR #6387, opened 2023-04-06
EVG-17952 Modify FindProjectAliases to correctly merge aliases

hadjri:
EVG-17952

### Description
`FindProjectAliases` doesn't currently behave as expected because it will either return all project page aliases, or all repo page aliases before merging with what was defined in the config YAML.

- Renamed alias merging related functions to be more indicative of what they are actually doing
- Changed the logic of `FindProjectAliases` to use the same rules that `FindMergedAliasesFromProjectRepoOrProjectConfig ` was following, where merging occurs on a project page -> repo page -> config YAML order of precedence.

### Testing
Added further test cases

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -271,7 +271,7 @@ func ConstructMergedAliasesByPrecedence(projectRef *ProjectRef, projectConfig *P
 	}
 	// If all aliases are covered in the project, so there's no reason to look at other sources
 	uncoveredAliases := uncoveredAliasTypes(aliasesToReturn)
-	if len(uncoveredAliases) > 0 {
+	if len(uncoveredAliases) > 0 && repoId != "" {
 		// Get repo aliases and merge with project aliases
 		repoAliases, err := FindAliasesForRepo(repoId)
 		if err != nil {
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -394,14 +394,6 @@ func (p *ProjectRef) UseRepoSettings() bool {
 	return p.RepoRefId != ""
 }
 
-// GetRepoSettingsId returns the repo settings id if the project is using repo settings, otherwise it returns an empty string
-func (p *ProjectRef) GetRepoSettingsId() string {
-	if !p.UseRepoSettings() {
-		return ""
-	}
-	return p.RepoRefId
-}
-
 func (p *ProjectRef) DoesTrackPushEvents() bool {
 	return utility.FromBoolPtr(p.TracksPushEvents)
 }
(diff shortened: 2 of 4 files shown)


model/project_aliases.go
```diff
@@ -271,9 +271,9 @@ func FindMergedAliasesFromProjectRepoOrProjectConfig(projectRef *ProjectRef, pro
 	}
 	// If all aliases are covered in the project, so there's no reason to look at other sources
 	uncoveredAliases := uncoveredAliasTypes(aliasesToReturn)
-	if len(uncoveredAliases) > 0 && projectRef.UseRepoSettings() {
+	if len(uncoveredAliases) > 0 {
```
ablack12, line 274: should we check that repoId != "" instead to avoid doing an unnecessary query?

model/project_ref.go
```diff
@@ -394,6 +394,14 @@ func (p *ProjectRef) UseRepoSettings() bool {
 	return p.RepoRefId != ""
 }
 
+// GetRepoSettingsId returns the repo settings id if the project is using repo settings, otherwise it returns an empty string
+func (p *ProjectRef) GetRepoSettingsId() string {
```
ablack12, line 398: This function isn't really necessary -- all p.UseRepoSettings does is check if p.RepoRefId is empty, i.e. we don't set this if we aren't using repo settings

rest/data/aliases.go
```diff
@@ -9,14 +9,31 @@ import (
...
-	var err error
-	var aliases model.ProjectAliases
+// FindMergedProjectAliases returns a merged list of aliases, with the order of precedence being:
+// 1. aliases defined on the project page
+// 2. aliases defined on the repo page
+// 3. aliases defined in the project config YAML
+// The includeProjectConfig flag determines whether to include aliases defined in the project config YAML.
+// If the aliasesToAdd parameter is defined, we fold those aliases in and remove any that are marked as deleted.
+func FindMergedProjectAliases(projectId, repoId string, aliasesToAdd []restModel.APIProjectAlias, includeProjectConfig bool) ([]restModel.APIProjectAlias, error) {
+	projectRef, err := model.FindBranchProjectRef(projectId)
```
ablack12, line 19: I think we want MergedProjectRef since we pass it into the construct function and check version control

rest/data/aliases.go
```diff
@@ -9,14 +9,31 @@ import (
...
+		projectConfig, err = model.FindLastKnownGoodProjectConfig(projectId)
+		if err != nil {
+			return nil, errors.Wrapf(err, "finding project config for project '%s'", projectId)
+		}
+	}
+	aliases, err := model.ConstructMergedAliasesByPrecedence(projectRef, projectConfig, repoId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "finding merged aliases for project '%s'", projectId)
+	}
+	if aliases == nil {
```
ablack12, line 34: I don't think we want to return early since we mentioned folding in aliasesToAdd.

ablack12 (changes requested):

rest/data/aliases.go
hadjri: Right, because `VersionControlEnabled` can be overridden at the repo level as well. Thanks

ablack12 (approved):

PR #6389, opened 2023-04-07
EVG-19293 log task end stats from MarkEnd

ablack12:
[EVG-19293](https://jira.mongodb.org/browse/EVG-19293)

### Description
Right now we log task end stats to splunk in a job, and we only call it in the routes called by the agent. This means we don't log anything to splunk for system failures or other places where we end and restart the task ourselves. 

We can't call the job from anywhere in the model package bc of import cycles. Because MarkEnd is used all over the task lifecycle, cleanest solution was to not have a job for this. We don't gate the job on project settings or anything; I theorize that the HistoricalRuntime aggregation might be expensive and maybe that's why we had a job originally? I'm not convinced we need this though. We could log how much time this takes in MarkEnd if we want to see how this affects things.

### Testing
Moved job tests into model.


Kimchelly (changes requested):

Kimchelly (approved):
I gotta run, but the change generally looks good. I'm just gonna assume you'll fix the failing tests and can re-request evg-app if necessary.

PR #6394, opened 2023-04-11
EVG-19597 always validate github org when attaching to repo / saving page

ablack12:
[EVG-19597](https://jira.mongodb.org/browse/EVG-19597) 

### Description
Realized from Bynn's scope that Lisbeth was able to track a personal Github org. After some investigation, it looks like the workflow "change owner -> enable repo -> enable project" is a loophole, since we don't validate this for projects attached to repo, since we assumed that the repo had to be valid. This would also be true for attach to new repo. 

Fixed both of these functions to validate, and updated the save page to validate strictly.

### Testing
Added some unit tests; tested in staging

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -689,8 +689,13 @@ func (p *ProjectRef) DetachFromRepo(u *user.DBUser) error {
 
 // AttachToRepo adds the branch to the relevant repo scopes, and updates the project to point to the repo.
 // Any values that previously were unset will now use the repo value, unless this would introduce
-// a Github project conflict. If no repo ref currently exists, the user attaching it will be added as the repo ref admin.
+// a GitHub project conflict. If no repo ref currently exists, the user attaching it will be added as the repo ref admin.
 func (p *ProjectRef) AttachToRepo(u *user.DBUser) error {
+	// Before allowing a project to attach to a repo, verify that this is a valid GitHub organization.
+	allowedOrgs := evergreen.GetEnvironment().Settings().GithubOrgs
+	if err := p.ValidateOwnerAndRepo(allowedOrgs); err != nil {
+		return errors.Wrap(err, "validating new owner/repo")
+	}
 	before, err := GetProjectSettingsById(p.Id, false)
 	if err != nil {
 		return errors.Wrap(err, "getting before project settings event")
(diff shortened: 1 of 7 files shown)


rest/route/repo_test.go
```diff
@@ -321,7 +322,7 @@ func TestPatchHandlersWithRestricted(t *testing.T) {
 	resp := attachProjectHandler.Run(ctx)
 	assert.NotNil(t, resp)
 	assert.Equal(t, resp.Status(), http.StatusOK)
-
+	fmt.Println(resp.Data())
```
hadjri, line 325: Is this leftover debugging?

rest/data/project_settings_test.go
```diff
@@ -32,17 +32,29 @@ func TestSaveProjectSettingsForSectionForRepo(t *testing.T) {
...
-			// ensure that we're saving settings without a special case
+
+			// Shouldn't succeed if the new owner isn't in the config.
 			settings, err := SaveProjectSettingsForSection(ctx, ref.Id, apiChanges, model.ProjectPageGeneralSection, true, "me")
+			assert.Error(t, err)
+
+			config.GithubOrgs = append(config.GithubOrgs, ref.Owner) // Add the new owner
+			assert.NoError(t, config.Set())
+			
+			// ensure that we're saving settings without a special case
```
hadjri, line 56: nit: capitalize

model/project_ref.go
```diff
@@ -2334,6 +2330,8 @@ func (p *ProjectRef) ValidateOwnerAndRepo(validOrgs []string) error {
 
 func validateOwner(owner string, validOrgs []string) error {
 	if len(validOrgs) > 0 && !utility.StringSliceContains(validOrgs, owner) {
+		fmt.Println("valid orgs: ", validOrgs)
```
hadjri, line 2333: Same comment here

rest/data/project_settings.go
```diff
@@ -237,6 +237,17 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 			}
 		}
 
+		if mergedSection.Owner != "" && mergedSection.Repo != "" {
```
hadjri, line 240: Is this if block in place to avoid erroring if we called `ValidateOwnerAndRepo` directly? Can we get around this by exporting the `validateOwner` function, and then just calling that here, since it only checks the list of valid orgs and won't error if the Owner or Repo is empty?

hadjri (changes requested):

hadjri (approved):

hadjri:
LGTM mod unit test & lint fixes.

PR #6396, opened 2023-04-11
EVG-18219 Add logging for host statuses

bynn:
EVG-18219

### Description
Adding an explicit grip log with a custom message to filter for a splunk alert 

"Because people are less likely to modify grip logs when they're tied to alerts. Errors that we add using AddError we generally just have for debugging/tracking/error rate monitoring, not for specific alerts that need to look for specific conditions" - kim

### DIFF
--- a/units/host_status.go
+++ b/units/host_status.go
@@ -107,6 +107,10 @@ clientsLoop:
 					j.AddError(j.terminateUnknownHosts(ctx, err.Error()))
 					continue clientsLoop
 				}
+				grip.Error(message.WrapError(err, message.Fields{
+					"message": "getting instance statuses from AWS",
+					"job":     j.ID(),
+				}))
 				j.AddError(errors.Wrap(err, "getting host statuses for providers"))
 				continue clientsLoop
 			}


units/host_status.go
```diff
@@ -107,6 +107,10 @@ clientsLoop:
 					j.AddError(j.terminateUnknownHosts(ctx, err.Error()))
 					continue clientsLoop
 				}
+				grip.Error(message.WrapError(err, message.Fields{
```
ablack12, line 110: I think we should grip.Debug this -- since we already add the error to the job, this would be double counted in the error alert.

ablack12 (changes requested):

units/host_status.go
bynn: makes a lot of sense

units/host_status.go
```diff
@@ -107,6 +107,10 @@ clientsLoop:
 					j.AddError(j.terminateUnknownHosts(ctx, err.Error()))
 					continue clientsLoop
 				}
+				grip.Debug(message.WrapError(err, message.Fields{
+					"message": "getting instance statuses from AWS",
```
ablack12, line 111: nit: this message here isn't wrapping an error (the way that errors.Wrap does) so it does make sense to start the sentence with "error" or "problem" to explain that this is an issue, not just that you _are_ getting the instance status.

ablack12 (approved):

PR #6397, opened 2023-04-12
EVG-19293 only compute standard deviation when needed

ablack12:
EVG-19293

### Description
After introducing some logging for EVG-19293, looking for ways to improve this. Remove unnecessary function.

### DIFF
--- a/model/task/expected_duration.go
+++ b/model/task/expected_duration.go
@@ -29,9 +29,9 @@ type expectedDurationResults struct {
 	StdDev           float64 `bson:"std_dev"`
 }
 
-func getExpectedDurationsForWindow(name, project, buildvariant string, start, end time.Time) ([]expectedDurationResults, error) {
+func getExpectedDurationsForWindow(name, project, buildVariant string, start, end time.Time, withStdDev bool) ([]expectedDurationResults, error) {
 	match := bson.M{
-		BuildVariantKey: buildvariant,
+		BuildVariantKey: buildVariant,
 		ProjectKey:      project,
 		StatusKey: bson.M{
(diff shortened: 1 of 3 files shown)


ablack12:
evergreen refresh

model/task/expected_duration.go
```diff
@@ -92,21 +95,3 @@ func getExpectedDurationsForWindow(name, project, buildvariant string, start, en
 
 	return results, nil
 }
-
```
hadjri, line 95: Can we add in the PR description that this was removed only because it was unused

model/task/task.go
```diff
@@ -2803,7 +2803,8 @@ func GetAllDependencies(taskIDs []string, taskMap map[string]*Task) ([]Dependenc
 }
 
 func (t *Task) GetHistoricRuntime() (time.Duration, error) {
-	runtimes, err := getExpectedDurationsForWindow(t.DisplayName, t.Project, t.BuildVariant, t.FinishTime.Add(-oneMonthIsh), t.FinishTime.Add(-time.Second))
+	runtimes, err := getExpectedDurationsForWindow(t.DisplayName, t.Project, t.BuildVariant,
```
hadjri, line 2806: Code LGTM, so it seems `withStdDev` is false always for now? Do we have a plan to change this in the future if we see some certain behavior once this is deployed?

model/task/expected_duration.go
ablack12: comments also provide history so this thread covers that as well i think

model/task/task.go
ablack12: Oh that was a mistake; this should be true for FetchExpectedDuration since that records the std dev. I'll update.

hadjri (approved):

PR #6399, opened 2023-04-12
EVG-18732 use correct context for ignored files message

ablack12:
EVG-18732
### Description
Realized in https://github.com/evergreen-ci/evergreen/pull/6398 that I used the wrong context for this haha

### DIFF
--- a/units/github_status_api.go
+++ b/units/github_status_api.go
@@ -226,7 +226,7 @@ func (j *githubStatusUpdateJob) fetch() (*message.GithubStatus, error) {
 		status.State = message.GithubStateFailure
 
 	} else if j.UpdateType == githubUpdateTypePushToCommitQueue {
-		status.Context = commitqueue.GithubContext
+		status.Context = evergreen.GithubContext
 		status.Description = "added to queue"
 		status.State = message.GithubStatePending
 		if j.FetchID != "" {


bynn (commented):
can i ask you to explain the differences of these contexts? 

why do we use j.GithubContext sometimes and evergreenContext (where does this come from) in other times? 

and where did the evergreen.GithubContext come from?

units/github_status_api.go
```diff
@@ -209,7 +209,7 @@ func (j *githubStatusUpdateJob) fetch() (*message.GithubStatus, error) {
 		status.Description = j.Description
 
 	} else if j.UpdateType == githubUpdateTypeSuccessMessage {
-		status.Context = commitqueue.GithubContext
+		status.Context = evergreenContext
```
ablack12, line 212: > can i ask you to explain the differences of these contexts?
> why do we use j.GithubContext sometimes and evergreenContext (where does this come from) in other times?
> and where did the evergreen.GithubContext come from?
(answering this here so we have a thread in case of follow up questions) 

I actually overall made a couple of mistakes so updating the PR. evergreen.GithubContext isn't a thing haha, it should've been evergreenContext. 

For context (lol), status.Context here refers to what we have as the bolded label for a Github check. So the bug here is that "all patched files are ignored" looks like it's coming from the commit queue build variant, which isn't right (if you examine the values, this is the difference between these values: commitqueue.GithubContext is evergreen/commit_queue while evergreenContext is just "evergreen". That's the actual correct value for us not creating a patch.
image

j.GithubContext meanwhile is configured when the job is created, and is how we configure build variant statuses, but isn't as relevant to the other types of status updates (although we do tend to set it, so this isn't super necessary, but probably still makes sense as a failsafe).

bynn (approved):

PR #6400, opened 2023-04-12
EVG-19269 Reroute version page to spruce before checking for project

bynn:
EVG-19269

### Description

The legacy version page was failing but Spruce wasn't so I updated it so that we redirect to Spruce early and this should hopefully unblock cloud.

I also added some logging so we could figure out why this is happening. 
There is also a chance that this change will not fix the issue but that will also be made clearer by the logging.

### DIFF
--- a/service/version.go
+++ b/service/version.go
@@ -21,11 +21,23 @@ func (uis *UIServer) versionPage(w http.ResponseWriter, r *http.Request) {
 	projCtx := MustHaveProjectContext(r)
 	project, err := projCtx.GetProject()
 
-	if err != nil || project == nil || projCtx.Version == nil {
-		http.Error(w, "not found", http.StatusNotFound)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
 	if RedirectSpruceUsers(w, r, fmt.Sprintf("%s/version/%s", uis.Settings.Ui.UIv2Url, projCtx.Version.Id)) {
+		return
+	}
+
+	if project == nil || projCtx.Version == nil {
+		grip.Debug(message.Fields{
+			"message": "project or version not found",
+			"project": project,
+			"projCtx": projCtx,
+			"version": projCtx.Version,
+		})
+		http.Error(w, "not found", http.StatusNotFound)
 		return
 	}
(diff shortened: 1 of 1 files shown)


service/version.go
```diff
@@ -21,14 +21,26 @@ func (uis *UIServer) versionPage(w http.ResponseWriter, r *http.Request) {
 	projCtx := MustHaveProjectContext(r)
 	project, err := projCtx.GetProject()
 
-	if err != nil || project == nil || projCtx.Version == nil {
-		http.Error(w, "not found", http.StatusNotFound)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
khelif96, line 25: I don't think this would resolve the bug because `err` will probably be set if `projCtx` or `project` is `nil`. You should move the redirect above this line to avoid hitting this case entirely. 
https://github.com/evergreen-ci/evergreen/blob/cd6a9f2a659d8822f83a84086318e1d45f49b8cf/model/context.go#L80

This function gets called by GetProject() 
https://github.com/evergreen-ci/evergreen/blob/ecd41df3feffef24ceb380ad5ff5750069660b07/model/project.go#L1269
I didn't look at the actual query but based on the comment I think it is a likely culprit. There would be no mainline version for the project identifier linked to the problematic version. Because it is tracking a branch and not the mainline branch.

khelif96 (changes requested):

service/version.go
bynn: updated

service/version.go
```diff
@@ -21,12 +21,18 @@ func (uis *UIServer) versionPage(w http.ResponseWriter, r *http.Request) {
 	projCtx := MustHaveProjectContext(r)
 	project, err := projCtx.GetProject()
 
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	grip.DebugWhen(err != nil, message.Fields{
```
khelif96, line 24: Should we also log when project or projectCtx is nil as well? Since those were the conditions of the original error

khelif96 (approved):
LGTM

bynn:
evergreen refresh

PR #6402, opened 2023-04-13
EVG-14306, EVG-19287: Apply mutli sort logic to test resolvers and set HasCedarResults field in the API task model

julianedwards:
EVG-14306, EVG-19287

### Description
- Update timber and the test results interface to enable multi sort logic
- Apply multi sort logic to the Task.tests resolver
- Add the legacy `HasCedarResults` field to the API task model to fix the bug in the Task.tests resolver where older tasks using said field would not return tasks. 

### Testing
Added unit tests and tested that single sort logic works in staging. I confirmed using the graphql playground that the Task.tests resolver returns tests for older tasks using the `HasCedarResults` field.

### Documentation
N/A

### DIFF
--- a/model/testresult/local_service.go
+++ b/model/testresult/local_service.go
@@ -221,7 +221,7 @@ func (s *localService) filterAndSortTestResults(ctx context.Context, results []T
 func (s *localService) validateFilterOptions(opts *FilterOptions) error {
 	catcher := grip.NewBasicCatcher()
 
-	seenSortByKeys := map[string]int{}
+	seenSortByKeys := map[string]bool{}
 	for _, sortBy := range opts.Sort {
 		switch sortBy.Key {
 		case SortByStartKey, SortByDurationKey, SortByTestNameKey, SortByStatusKey, SortByBaseStatusKey, "":
@@ -230,13 +230,13 @@ func (s *localService) validateFilterOptions(opts *FilterOptions) error {
 			continue
 		}
 
-		if seenSortByKeys[sortBy.Key] == 1 {
+		if seenSortByKeys[sortBy.Key] {
 			catcher.Errorf("duplicate sort by key '%s'", sortBy.Key)
 		} else {
 			catcher.NewWhen(sortBy.Key == SortByBaseStatusKey && len(opts.BaseTasks) == 0, "must specify base task ID when sorting by base status")
 		}
 
-		seenSortByKeys[sortBy.Key] += 1
+		seenSortByKeys[sortBy.Key] = true
 	}
 
 	catcher.NewWhen(opts.Limit < 0, "limit cannot be negative")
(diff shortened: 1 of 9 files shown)


julianedwards:
> Were changes to go.mod and go.sum intentional?

Yep! I had to update the timber library.

model/testresult/local_service.go
```diff
@@ -266,44 +279,52 @@ func (s *localService) filterTestResults(results []TestResult, opts *FilterOptio
...
-				return results[i].GetDisplayTestName() > results[j].GetDisplayTestName()
-			}
-			return results[i].GetDisplayTestName() < results[j].GetDisplayTestName()
-		})
-	case SortByStatus:
-		sort.SliceStable(results, func(i, j int) bool {
-			if opts.SortOrderDSC {
-				return results[i].Status > results[j].Status
+	sort.SliceStable(results, func(i, j int) bool {
+		for _, sortBy := range opts.Sort {
```
ybrill, line 283: is this equivalent to `sort.SliceStable` for each of the sorts one after the other?
like
```go
for _, sortBy := range opts.Sort {
    sort.SliceStable(results, func(i, j int) bool {sortBy.lessThanFunc()}
}
```

model/testresult/local_service.go
```diff
@@ -221,11 +221,24 @@ func (s *localService) filterAndSortTestResults(ctx context.Context, results []T
 func (s *localService) validateFilterOptions(opts *FilterOptions) error {
 	catcher := grip.NewBasicCatcher()
 
-	switch opts.SortBy {
-	case SortByStart, SortByDuration, SortByTestName, SortByStatus, SortByBaseStatus, "":
-	default:
-		catcher.Errorf("unrecognized sort by criteria '%s'", opts.SortBy)
+	seenSortByKeys := map[string]int{}
```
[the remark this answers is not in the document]
julianedwards: Yes, not sure why I didn't just use a bool. I will update that!

model/testresult/local_service.go
julianedwards: No because even though it's a stable sort, it will not respect the priority of the sort criteria. For example:
```
[("julian", 5), ("jonathan", 10), ("john", 1), ("april", 9)] 
# sort by (name, int)
# after the first iteration:
[("april", 9), ("john", 1), ("jonathan", 10), ("julian", 5)] 
# and after the second:
[("john", 1), ("julian", 5), ("april", 9), ("jonathan", 10)] 
```

Basically, if we do it one after the other it completely ignores the previous sort criteria. You only want to sort collisions from the previous criteria.

ybrill: ah I understand.
Is it equivalent to doing the `SliceStable`s in reverse, then?

julianedwards: Essentially, this attempts to sort by the highest precedent criteria (returning if it's either less than or greater than depending on the sort order). If the keys are equivalent, instead of returning, it continues to the subsequent sort criteria until it can return (either all keys are equal or one is less than the other).

julianedwards: Basically, first try sorting by key X, but if the values of key X are equal then try sorting by key Y, and so on.

ybrill (approved):

julianedwards:
evergreen refresh

PR #6414, opened 2023-04-18
EVG-19630 Update correct child patch status to github

bynn:
EVG-19630

### Description
child patch statuses were not updating correctly on gituhb because we were passing in a collective status instead 
updated to pass in the correct status

### Testing
staging sandbox pr updates statuses correctly 
https://github.com/evergreen-ci/commit-queue-sandbox/pull/467

image

### DIFF
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -191,7 +191,9 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData, 
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
 	}
 
-	collectiveStatus := t.patch.Status
+	// For parent patches, we want to look at the collective status of the child patches.
+	// For child patches, we only want to look at its own status.
+	collectiveStatus := t.data.Status
 	if t.patch.IsParent() {
 		var err error
 		collectiveStatus, err = t.patch.CollectiveStatus()
--- a/trigger/patch.go
+++ b/trigger/patch.go
@@ -191,8 +191,8 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData, 
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
 	}
 
-	// For parent patches, we want to look at the collective status of the child patches.
-	// For child patches, we only want to look at its own status.
+	// For child patches, we only want to look at its own status because the collective status will be affected
+	// by other child patches' and the parent's status
 	collectiveStatus := t.data.Status
 	if t.patch.IsParent() {
 		var err error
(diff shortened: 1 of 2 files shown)


trigger/patch.go
```diff
@@ -190,10 +190,16 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
 	if api.ProjectIdentifier != nil {
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
 	}
-	collectiveStatus, err := patch.CollectiveStatus(t.patch.Id.Hex())
-	if err != nil {
-		return nil, errors.Wrap(err, "getting collective status for patch")
+
+	collectiveStatus := t.patch.Status
```
malikchaya2, line 194: Should this maybe be t.data.status? That's what it was before we switched it to collective status here.

trigger/patch.go
```diff
@@ -190,10 +190,16 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
...
+	if t.patch.IsParent() {
+		var err error
+		collectiveStatus, err = t.patch.CollectiveStatus()
```
malikchaya2, line 197: do we still need the `patch.CollectiveStatus(t.patch.Id.Hex())` function or can it be removed now?

trigger/patch.go
bynn: i think data.status is from the subscription so it would be more accurate to have patch.status maybe? open to discussion

trigger/patch.go
bynn: we need that function for versions (only used in 1 place)

ablack12 (changes requested):

trigger/patch.go
malikchaya2: we should probably use t.data.status to remain consistent with other makeData functions

bynn: updated!

trigger/patch.go
```diff
@@ -190,10 +190,16 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
 	if api.ProjectIdentifier != nil {
 		projectName = utility.FromStringPtr(api.ProjectIdentifier)
 	}
-	collectiveStatus, err := patch.CollectiveStatus(t.patch.Id.Hex())
-	if err != nil {
-		return nil, errors.Wrap(err, "getting collective status for patch")
+
+	collectiveStatus := t.patch.Status
+	if t.patch.IsParent() {
```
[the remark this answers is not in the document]
bynn: updated!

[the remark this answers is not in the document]
ablack12: Can we add a bit more detail about why, not just about what? Should we add a comment to the version trigger to explain why we _do_ look at collective there?

ablack12 (changes requested):

trigger/patch.go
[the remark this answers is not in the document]
bynn: updated

PR #6420, opened 2023-04-19
EVG-18657 refactor child patch operations to catch more cases

ablack12:
EVG-18657

### Description
Right now we handle child patch operations in the graphql helper, but this means everywhere else we consider restarts / priorities / etc we don't consider child patches. Moved this consideration into the individual functions. This may not cover everything but we can ticket new cases as they come up.

### Testing
Will add tests for child patches.


ablack12:
evergreen refresh

model/version.go
```diff
@@ -135,13 +135,22 @@ func (v *Version) SetActivated(activated bool) error {
...
+}
+
+// setVersionsActivated sets activated field to specified boolean given a list of version IDs.
+func setVersionsActivated(versionIds []string, activated bool) error {
+	if len(versionIds) == 0 {
+		return nil
+	}
 	return VersionUpdateOne(
-		bson.M{VersionIdKey: versionId},
+		bson.M{
```
hadjri, line 152: We should probably rename `VersionUpdateOne` since it just does a general update

model/version_modification.go
```diff
@@ -74,7 +77,16 @@ func ModifyVersion(version Version, user user.DBUser, modifications VersionModif
 				return http.StatusUnauthorized, errors.Errorf("insufficient access to set priority %d, can only set priority less than or equal to %d", modifications.Priority, evergreen.MaxTaskPriority)
 			}
 		}
-		if err := SetVersionsPriority([]string{version.Id}, modifications.Priority, user.Id); err != nil {
+		versionIds := []string{version.Id}
+		if evergreen.IsPatchRequester(version.Requester) {
```
hadjri, line 81: Since `GetFinalizedChildPatchIdsForPatch` will no-op if it's not a patch we can remove this if statement

model/version_db.go
```diff
@@ -405,3 +407,24 @@ func FindProjectForVersion(versionID string) (string, error) {
 	}
 	return v.Identifier, nil
 }
+
+// AbortVersion sets the abort flag on all tasks associated with the version which are in an
+// abortable state. Also handles child patches.
+func AbortVersion(versionId string, reason task.AbortInfo) error {
```
hadjri, line 413: Since this doesn't perform any operation on the version itself I think a better name would be `AbortVersionTasks`

graphql/util.go
```diff
@@ -380,22 +380,6 @@ func modifyVersionHandler(ctx context.Context, patchID string, modification mode
 	if err != nil {
 		return mapHTTPStatusToGqlError(ctx, httpStatus, err)
 	}
-
```
hadjri, line 383: Curious why was this removed? My understanding was that we'd add logic elsewhere to match what is happening here, not remove the child patch logic from the graphQL helper.

hadjri (changes requested):

model/version_db.go
ablack12: Good call

model/version_modification.go
ablack12: well it does no op but it's still a waste of a query, if we do know the requester ahead of time it makes sense to not bother with it

ablack12:
I think this PR might actually not be understanding correctly how child patches work; moving to draft for a bit.

model/version_modification.go
hadjri: Ah, I thought the `IsValidId` would make it no-op before the DB query was made but perhaps not

hadjri:
> I think this PR might actually not be understanding correctly how child patches work; moving to draft for a bit.

Sounds good, ready to re-review anytime

ablack12:
I'll probably just start over since it's been forever now.

PR #6421, opened 2023-04-19
EVG-18876 include commit messages for PR items

ablack12:
EVG-18876
### Description
For PR CQ items, we create the patch slightly differently such that commit_messages isn't populated. 

### Testing
[tested on staging](https://evergreen-staging.corp.mongodb.com/rest/v2/patches/6440287cb23736ed76eda692) / added to unit test

### DIFF
--- a/rest/data/commit_queue_test.go
+++ b/rest/data/commit_queue_test.go
@@ -334,7 +334,7 @@ func (s *CommitQueueSuite) TestWritePatchInfo() {
 			Name:      "myfile.go",
 			Additions: 1,
 			Deletions: 0,
-				},
+		},
 	}
 
 	patchContents := `diff --git a/myfile.go b/myfile.go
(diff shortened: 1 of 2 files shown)


rest/data/commit_queue_test.go
```diff
@@ -343,8 +347,10 @@ func (s *CommitQueueSuite) TestWritePatchInfo() {
 	`
 
 	s.NoError(writePatchInfo(patchDoc, patchSummaries, patchContents))
-	s.Len(patchDoc.Patches, 1)
+	s.Require().Len(patchDoc.Patches, 1)
```
bynn, line 350: most of our codebase seems to use s.len() and it looks like s.require().len seems to do the same thing.
would it be possible for us to use one version across the codebase for consistency/readability?

bynn (approved):

rest/data/commit_queue_test.go
ablack12: these do different things -- .require() errors if the line fails (which is helpful for length checks where we would have a nil pointer dereference on future lines), whereas otherwise we'll just continue and evaluate all lines, even if one fails. It would be ideal if we used Require everytime this could happen but I would say it's really not worth auditing the codebase for.

PR #6422, opened 2023-04-19
EVG-19619 Add nil check for projectRef

bynn:
EVG-19619

### Description
add nil check for projectRef

### Testing
staging 
called api without panic

### DIFF
--- a/service/api.go
+++ b/service/api.go
@@ -263,6 +263,12 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 
 	if input.Quiet {
 		errs = errs.AtLevel(validator.Error)
+	} else if projectRef != nil {
+		validationErr = validator.ValidationError{
+			Message: "project does not exist; validation will proceed without checking alias coverage",
+			Level:   validator.Warning,
+		}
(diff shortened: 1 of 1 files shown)


service/api.go
```diff
@@ -263,6 +264,12 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 
 	if input.Quiet {
 		errs = errs.AtLevel(validator.Error)
+	} else if projectRef != nil || !isConfigDefined {
```
Kimchelly, line 267: Is this supposed to be `projectRef == nil`?

Kimchelly: Also, should this skip alias validation when there's no project config? I thought that the project config was optional.

Kimchelly (changes requested):

service/api.go
bynn: you are so right

hadjri: `ConstructMergedAliasesByPrecedence` is meant to return a list of aliases regardless of whether the project config is defined so yes, we'd want to do alias validation regardless

Kimchelly (approved):

PR #6423, opened 2023-04-19
EVG-19631: remove commit queue merge dependence on patched parser project

Kimchelly:
EVG-19631

### Description
There was apparently one remaining place that was dependent on setting the `PatchedParserProject` (which is deprecated) to add the merge task to the commit queue patch. I removed that dependency, so nowhere in the code base sets the `PatchedParserProject` on the patch doc anymore. After this is deployed, I'm going to re-apply #6409, which should be safe once the pre-deploy commit queue patches all finish running.

### Testing
Ran existing tests, and ran a commit queue merge in the sandbox.

### Documentation
N/A

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -1802,7 +1802,10 @@ func activateExistingInactiveTasks(creationInfo TaskCreationInfo, existingBuilds
 
 func getTaskIdTables(creationInfo TaskCreationInfo) (TaskIdConfig, error) {
 	// The table should include only new and existing tasks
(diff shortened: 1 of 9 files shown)


bynn (approved):
all the changes look good but I'm a little curious as to why were not completely removing PatchedParserProject in this pr since we're getting rid of all uses of it and it looks like its currently just for falling back on but there haven't been any issues. 

is there another ticket for that work?

Kimchelly:
We can't delete because it will break backward compatibility. Old patches that existed before EVG-18700 that were created but not finalized can't finalize without the fallback `PatchedParserProject` handling. EVG-19636 is intended to eventually remove the field entirely.

PR #6424, opened 2023-04-19
EVG-17975: delete subprocess.scripting and metabuild system

Kimchelly:
EVG-17975

### Description
Remove the unused `subprocess.scripting` command and auto-generated YAML config features. These were an old idea that never got off the ground, so have never been used as far as I'm aware.

### Testing
All it needs to do is compile and pass remaining tests.

### Documentation
The docs related to both features were already removed a while back.


ablack12 (approved):
LGTM but a couple of questions looking at https://jira.mongodb.org/browse/EVG-19434 when this project was originally finished -- is the issue that there weren't stakeholders? Does the feature seem like it does work and we should just advertise it more or is it definitely dead? And out of curiosity are there any parts of this project that _were_ useful or are we just undoing it now?

Kimchelly:
I don't remember who the original stakeholders were, but I believe they were people working on the server metabuild system (which seems to be defunct). The thought was that people should have an alternative to writing long, complex, and difficult-to-debug YAML configurations and the server was looking for a way to reduce that complexity. This auto-generation/scripting feature specifically was meant to be a POC to prove that we could reduce that configuration complexity for simple needs and expand upon it in later projects as a more formal Evergreen metabuild system. Unfortunately, it never panned out nor was it fleshed out into handling more complex cases partly due to the server metabuild idea losing momentum, the original engineer who came up with this feature leaving, and PLT moving onto other projects. (As an aside, my personal belief is that the features proposed in the POC are generally in the right direction if we wanted to flesh out the metabuild system, but it's not clear that anyone would use it at this point.)

The feature does work as it's described, but it was just a POC for small, simple projects, which likely don't have face a YAML configuration issue since they're simple. The other feature that was created in EVG-19434 was task sync, which I don't have stats on whether it's used, but it is enabled in a bunch of projects' settings.

PR #6428, opened 2023-04-20
EVG-18942 Fix missing failing command due to missing display names 

malikchaya2:
EVG-18942

### Description
Display names were only being added for commands that were part of a function. This left tasks without failing commands. 

This PR adds failing commands for all commands, and adds where in the project yaml they came from (pre, post, etc) for extra clarity. This will also show up in task logs, hopefully providing a better user experience. 

### Testing
Tested on staging 

image
image


agent/agent.go
```diff
@@ -586,7 +586,7 @@ func (a *Agent) runTaskTimeoutCommands(ctx context.Context, tc *taskContext) {
 		return
 	}
 	if taskGroup.Timeout != nil {
-		err := a.runCommands(ctx, tc, taskGroup.Timeout.List(), runCommandsOptions{})
+		err := a.runCommands(ctx, tc, taskGroup.Timeout.List(), runCommandsOptions{}, "task-timeout")
```
hadjri, line 589: Can we make global constants for these and group them in the same place?

hadjri (approved):

agent/agent.go
malikchaya2: smart, done

PR #6430, opened 2023-04-20
Evg 19201 agent instrumentation

ybrill:
[EVG-19201](https://jira.mongodb.org/browse/EVG-19201)

### Description
Create a trace for the execution of a task. 

### Testing
If tracing is enabled this will create a trace for every task execution. The trace breaks down the execution of the task into its individual commands. Begin to instrument git.get_project so the steps of the command are also broken down. [This](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen-agent/trace/yVtoRdnUhwC) is an example of a span from running [this task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2004_test_auth_c372fc8686d12c2480c369f3d7ac5be834859adf_23_04_19_21_33_18/tests?execution=7&sortBy=STATUS&sortDir=ASC).

Used the OTEL concept of baggage to pass the task's attributes down to _every_ span in the trace. This allows queries such as [this one](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen-agent/result/2Z7SHMSe2Ao). Baggage is how Honeycomb suggests doing this [here](https://docs.honeycomb.io/getting-data-in/opentelemetry/go-distro/#multi-span-attributes).


agent/agent.go
```diff
@@ -149,19 +166,101 @@ func newWithCommunicator(ctx context.Context, opts Options, comm client.Communic
...
-func (a *Agent) Close() {
-	if a.comm != nil {
-		a.comm.Close()
+func (a *Agent) initTracerProvider(ctx context.Context) error {
+	if a.opts.TraceCollectorEndpoint == "" {
+		return nil
 	}
+
+	resource, err := resource.New(ctx,
+		resource.WithAttributes(semconv.ServiceName("evergreen-agent")),
```
opsnlops, line 200: This will work for now! Long term we might wanna think about how to make this something that's settable at the project level, maybe? 🤔

opsnlops (approved):
Looks good to me!

agent/agent.go
ybrill: It definitely seems possible, though since the agent is shared between projects I wonder if we'd want to divide the traces by project further downstream.
One potential issue that comes to mind is that HC doesn't allow queries across datasets and some stuff (like the host metrics) might straddle more than one project.

I'm good to wait and see.

PR #6431, opened 2023-04-21
EVG-19027: Support preflight OPTIONS requests for legacy UI POST routes

minnakt:
EVG-19027

### Description
On Spruce, POST requests to the `/login` and `/settings/newkey` endpoints are not working because of CORS errors. The errors are caused by a failure to satisfy the [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request).

I added OPTIONS requests to these endpoints, which essentially do nothing but return the correct CORS headers. This allows us to be able to use the endpoints on Spruce.

### Testing
- On staging, checked that I could regenerate my Evergreen API Key (this does not work on Spruce production)


Kimchelly (commented):
Does the OPTIONS route have to return a 200 response? Also, can we just plop [this logic](https://github.com/evergreen-ci/evergreen/blob/207e50de530ae2630371ac1e25363c2c43b250ee/rest/route/service.go#L261-L268) into the legacy UI router?

khelif96 (commented):
+1 to @Kimchelly's comment.

minnakt:
> Does the OPTIONS route have to return a 200 response? Also, can we just plop [this logic](https://github.com/evergreen-ci/evergreen/blob/207e50de530ae2630371ac1e25363c2c43b250ee/rest/route/service.go#L261-L268) into the legacy UI router?

Uoohh I did not know about this... thanks Kim!! 🥰 and yes it needs to be 200

service/ui.go
```diff
@@ -446,5 +444,15 @@ func (uis *UIServer) GetServiceApp() *gimlet.APIApp {
 	app.PrefixRoute("/plugin").Route("/buildbaron/note/{task_id}").Wrap(needsLogin, needsContext, viewTasks).Handler(bbGetNote).Get()
 	app.PrefixRoute("/plugin").Route("/buildbaron/note/{task_id}").Wrap(needsLogin, needsContext, editTasks).Handler(bbSaveNote).Put()
 	app.PrefixRoute("/plugin").Route("/buildbaron/file_ticket").Wrap(needsLogin, needsContext).Handler(uis.bbFileTicket).Post()
+
+	// Add an OPTIONS method to every POST request to handle preflight OPTIONS requests.
+	// These requests must not check for credentials. They exist to validate whether a route exists, and to
+	// allow requests from an origin.
+	for _, r := range app.Routes() {
+		if r.HasMethod(http.MethodPost) {
+			app.AddRoute(r.GetRoute()).Wrap(allowsCORS).Handler(func(_ http.ResponseWriter, _ *http.Request) {}).Options()
```
Kimchelly, line 453: My question about the 200 response was because this seems to return no response to the request. Is this handler supposed to `gimlet.WriteJSON(w, "")`?

minnakt: Oh, it seemed to already return 200 when I checked in the network tab 🤔

Screenshot 2023-04-25 at 11 11 50 AM

I updated it to use `gimlet.WriteJSON(w, "")` and it still works so that sounds good to me!

Kimchelly (approved):

PR #6439, opened 2023-04-24
EVG-19680 Add logging for PR mergable state 

malikchaya2:
EVG-19680

### Description
Logging the state of PRs we try to merge will help us with debugging.

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -375,14 +375,14 @@ func getPRAndCheckMergeable(ctx context.Context, env evergreen.Environment, sc C
 	}
 	mergeableState := pr.GetMergeableState()
 
-	grip.Info(message.WrapError(err, message.Fields{
+	grip.Info(message.Fields{
 		"message":        "checking PR mergeable status",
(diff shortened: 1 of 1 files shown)


hadjri (commented):
Do we have a ticket to undo this once investigation is done?

rest/data/commit_queue.go
```diff
@@ -373,10 +373,19 @@ func getPRAndCheckMergeable(ctx context.Context, env evergreen.Environment, sc C
 	if err != nil {
 		return nil, err
 	}
-
+	mergeableState := pr.GetMergeableState()
+
+	grip.Info(message.WrapError(err, message.Fields{
```
ablack12, line 378: If you wrap error here, you'll only actually log if the error isn't nil -- not sure if that's what you want.

ablack12 (approved):
LGTM with note

rest/data/commit_queue.go
ablack12: Can we also use grip.Debug since this is for debugging?

hadjri (approved):

PR #6442, opened 2023-04-25
EVG-18278 Fix slack target for build break notifications for admins 

malikchaya2:
EVG-18278

### Description
The slack target for build break notifications didn't include the "@" like it should have. I opted to fix this instead of removing it because it's used by a lot of projects and only the slack (and not email) target was broken. Additionally, this was already fixed for most slack users when we switched to using memberIds for slack. 

### Testing
Expanded existing tests.

### DIFF
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -580,7 +580,7 @@ func makeBuildBreakSubscriber(userID string) (*event.Subscriber, error) {
 		if preference == user.PreferenceEmail {
 			subscriber.Target = u.Email()
 		} else if preference == user.PreferenceSlack {
-			slackTarget := fmt.Sprintf("@%s", u.Settings.SlackUsername)
+			slackTarget := fmt.Sprintf("@%s", strings.TrimPrefix(u.Settings.SlackUsername, "@"))
 			if u.Settings.SlackMemberId != "" {
 				slackTarget = u.Settings.SlackMemberId
 			}
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -580,7 +580,7 @@ func makeBuildBreakSubscriber(userID string) (*event.Subscriber, error) {
 		if preference == user.PreferenceEmail {
 			subscriber.Target = u.Email()
 		} else if preference == user.PreferenceSlack {
-			slackTarget := u.Settings.SlackUsername
+			slackTarget := fmt.Sprintf("@%s", u.Settings.SlackUsername)
 			if u.Settings.SlackMemberId != "" {
 				slackTarget = u.Settings.SlackMemberId
 			}
--- a/repotracker/repotracker_test.go
+++ b/repotracker/repotracker_test.go
@@ -772,13 +772,7 @@ func TestBuildBreakSubscriptions(t *testing.T) {
 	}
 	assert.NoError(AddBuildBreakSubscriptions(&v3, &proj2))
 	assert.NoError(db.FindAllQ(event.SubscriptionsCollection, db.Q{}, &subs))
-	assert.Equal(subs[0].Subscriber.Target, "HELL0173")
-	if _, ok := event.(*rawAdminEventData); !ok {
-		s.NotEmpty(jsonTag, "struct %s: field '%s' must have json tag", t.Name(), f.Name)
-
-		s.NotEqual("resource_type", jsonTag, `'%s' has a json:"resource_type" tag, but should not`, t.String())
-		s.NotEqual("resource_type,omitempty", jsonTag, `'%s' has a json:"resource_type,omitempty" tag, but should not`, t.String())
-	}
+	assert.EqualValues(utility.ToStringPtr("HELL0173"), subs[0].Subscriber.Target)
 	assert.Len(subs, 2)
 }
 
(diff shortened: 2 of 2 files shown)


repotracker/repotracker.go
```diff
@@ -580,7 +580,7 @@ func makeBuildBreakSubscriber(userID string) (*event.Subscriber, error) {
 		if preference == user.PreferenceEmail {
 			subscriber.Target = u.Email()
 		} else if preference == user.PreferenceSlack {
-			slackTarget := u.Settings.SlackUsername
+			slackTarget := fmt.Sprintf("@%s", u.Settings.SlackUsername)
```
Kimchelly, line 583: I would check if it already starts with an `@` to avoid double-adding it.

repotracker/repotracker_test.go
```diff
@@ -761,9 +756,16 @@ func TestBuildBreakSubscriptions(t *testing.T) {
 		},
 	}
 	assert.NoError(u4.Insert())
+	assert.NoError(AddBuildBreakSubscriptions(&v1, &proj2))
+	assert.NoError(db.FindAllQ(event.SubscriptionsCollection, db.Q{}, &subs))
+	assert.Len(subs, 2)
```
Kimchelly, line 761: Seems like this should check that the scription for `u2` is now `@hello.itsme` to verify the PR change.

Kimchelly (changes requested):
Will this fix existing build break subscriptions that are already malformed?

Did you test this will fix the issue for pre-existing build-break subscriptions (i.e. those that already are missing the `@`) in staging?

repotracker/repotracker.go
malikchaya2: true, fixed.

repotracker/repotracker_test.go
malikchaya2: That would be ideal but I don't want to add logic for dealing with the Target interface{}. I stepped through it with a debugger though so I know that it's acting the way it should. 
The other tests are not that robust so I don't think it's crucial to add it here.

malikchaya2:
> Will this fix existing build break subscriptions that are already malformed?
> 
> Did you test this will fix the issue for pre-existing build-break subscriptions (i.e. those that already are missing the `@`) in staging?

I don't think that's necessary because these subscriptions don't stick around and get re-used the way that the ones set up [here ](https://spruce.mongodb.com/preferences/notifications) do. They are created by the repotracker.

repotracker/repotracker_test.go
Kimchelly: The point of the test is not just to verify that the code work today, but to also verify non-regressive behavior. Just spot checking is not a replacement for a test, which will verify on every PR that the resulting subscription is still correct. You should be able to compare against an `interface{}` type by using `assert.EqualValues`.

repotracker/repotracker_test.go
```diff
@@ -761,15 +756,29 @@ func TestBuildBreakSubscriptions(t *testing.T) {
...
-		Identifier: proj1.Id,
+		Identifier: proj2.Id,
 		Requester:  evergreen.RepotrackerVersionRequester,
 		Branch:     "branch",
 		AuthorID:   u4.Id,
 	}
 	assert.NoError(AddBuildBreakSubscriptions(&v3, &proj2))
 	assert.NoError(db.FindAllQ(event.SubscriptionsCollection, db.Q{}, &subs))
+	assert.Equal(subs[0].Subscriber.Target, "HELL0173")
+	if _, ok := event.(*rawAdminEventData); !ok {
```
Kimchelly, line 776: What is this for?

Kimchelly (changes requested):

repotracker/repotracker_test.go
malikchaya2: committed by mistake

repotracker/repotracker_test.go
```diff
@@ -724,14 +724,16 @@ func TestBuildBreakSubscriptions(t *testing.T) {
...
 		NotifyOnBuildFailure: utility.TruePtr(),
-		Admins:               []string{"u2", "u3"},
+		Admins:               []string{"u2", "u3", "u4"},
 	}
 	u2 := user.DBUser{
 		Id:           "u2",
 		EmailAddress: "shaw@blizzard.com",
 		Settings: user.UserSettings{
+			SlackUsername: "hello.itsme",
+			SlackMemberId: "HELL0173",
```
Kimchelly, line 734: Adding the `SlackMemberId` obfuscates whether the change being tested actually works (the `@` addition) because `SlackMemberId` overwrites `SlackUsername`, so I would remove `SlackMemberId` from this user (you could test `SlackMemberId` with another user if desired).

repotracker/repotracker_test.go
```diff
@@ -761,15 +756,23 @@ func TestBuildBreakSubscriptions(t *testing.T) {
...
 		Id:         "v3",
-		Identifier: proj1.Id,
+		Identifier: proj2.Id,
 		Requester:  evergreen.RepotrackerVersionRequester,
 		Branch:     "branch",
 		AuthorID:   u4.Id,
 	}
 	assert.NoError(AddBuildBreakSubscriptions(&v3, &proj2))
 	assert.NoError(db.FindAllQ(event.SubscriptionsCollection, db.Q{}, &subs))
+	assert.EqualValues(utility.ToStringPtr("HELL0173"), subs[0].Subscriber.Target)
```
Kimchelly, line 775: Now that I'm looking at how the `interface{}` stores the string, it seems like kind of a weird comparison since this should ideally compare strings and not pointer addresses. If it's stored as a pointer in the `interface{}`, can we do a type assert to a string pointer `subs[0].Subscriber.Target.(string)` and then `utility.FromStringPtr` it?

Kimchelly (approved):

PR #6446, opened 2023-04-25
EVG-17957 Add API to manually set pod secrets

hadjri:
EVG-17957

### Description
Currently, newly created projects are given a pod secret by default, but most existing projects do not have one. Creating a route that will allow us to provision a pod secret per project, with the ability to rotate if needed as a query parameter. This route should no-op if there is an existing pod secret and we're not rotating anything.
### Testing
Tested in staging, added unit tests


Kimchelly (changes requested):
Is this somehow different from the `PATCH /rest/v2/projects/{project_id}` route that lets you add a pod secret to an existing project? If it's the same, I'd be in favor of just using the existing route.

PR #6449, opened 2023-04-26
EVG-19716 Add log stream prefix to ECS options to fix OTEL error

hadjri:
EVG-19716

### Description
Saw that container tasks were suddenly unable to start at all since OTEL changes have been added to the agent. 

Turns out it's because [Opentelemetry's](https://gist.github.com/hadjri/11fca9d9e9629347a2257706709b6e5c#file-ecs-go-L153) implementation that parses ECS task definitions sends an error if the log stream is not specified in the ECS log options, causing our agents to fail and exit without running tasks with the following error: `[p=emergency]: constructing agent: initializing tracer provider: making otel resource: detecting resources: [the ECS Metadata v4 did not return a AwsLogStream name]`

Added "awslogs-stream-prefix" as a mandatory option to our ECS log configuration since OTEL requires it. It also makes the stream names more readable in ECS so that is a plus too. Currently all stream names appear to be some random hash but this will make them be structured in the following format: `ecs/evg-agent/<id>`
### Testing
Confirmed in staging that this fixes the failing agents.

### DIFF
--- a/cloud/pod_util.go
+++ b/cloud/pod_util.go
@@ -226,6 +226,8 @@ const (
 	awsLogsGroup = "awslogs-group"
 	// awsLogsGroup is the log configuration option name for specifying the AWS region.
 	awsLogsRegion = "awslogs-region"
+	// awsLogsStreamPrefix is the log configuration option name for specifying the log stream prefix.
+	awsLogsStreamPrefix = "awslogs-stream-prefix"
 )
 
 // ExportECSPodDefinitionOptions exports the ECS pod creation options into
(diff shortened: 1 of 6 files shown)


malikchaya2 (approved):
This looks good, but I would also advocate to, either in this ticket or another ticket, make sure that we continue on error so that if their system is ever down it down doesn’t bring us down. We can log the errors as critical if we want to increase the visibility.

hadjri:
> This looks good, but I would also advocate to, either in this ticket or another ticket, make sure that we continue on error so that if their system is ever down it down doesn’t bring us down. We can log the errors as critical if we want to increase the visibility.

https://jira.mongodb.org/browse/EVG-19718

PR #6450, opened 2023-04-27
EVG-18501: Add retry options to WebhookSubscriber in GraphQL schema

minnakt:
EVG-18501

### Description
This PR modifies the `WebhookSubscriber` and `WebhookSubscriberInput` types in GraphQL to accept the retry options that were introduced in [EVG-17448](https://jira.mongodb.org/browse/EVG-17448). 

The options are set to 0 by default because that's how they get saved in the database currently. I think that should be okay because it appears we have some defaults set for when the webhook gets sent in [`webhook_grip.go`](https://github.com/evergreen-ci/evergreen/blob/main/util/webhook_grip.go#L137-L144).

### Testing
- Added GraphQL tests


khelif96 (approved):

graphql/schema/types/project_subscriber.graphql
```diff
@@ -51,6 +54,9 @@ input WebhookSubscriberInput {
   headers: [WebhookHeaderInput]!
   secret: String!
   url: String!
+  retries: Int = 0
```
khelif96, line 57: Should these correspond to the defaults?

minnakt: I think at least for right now, it's ok to set them to 0. There are a bunch of subscriptions in the production DB with these values set to 0 (and it might be confusing for that pattern to suddenly end). I double checked with Jonathan and he said that if these values are left at 0 in the database, they will get set to the defaults when sending the webhook.

minnakt:
Going to merge as the Spruce e2e tests seems to be flaky!

PR #6452, opened 2023-04-27
EVG-19104 add logging to merge commit fallback

ablack12:
EVG-19104 

### Description
Adding some logging to see if we ever go through all of our retries and [fall back on an old commit](https://github.com/evergreen-ci/evergreen/blob/98e0fe9852cbf3c907dbc03893c3367590359cba/agent/command/git.go#L311) (and if so, is this ever a problem). Will set up a splunk alert for myself and will plan to look at EVG-19723 in about a month.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -350,9 +350,13 @@ func (c *gitFetchProject) waitForMergeableCheck(ctx context.Context, comm client
 		getPRRetryMaxDelay = 15 * time.Second
 	)
 	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
-
+	attempt := 0
 	err := utility.Retry(ctx, func() (bool, error) {
-		info, err := comm.GetPullRequestInfo(ctx, td, conf.GithubPatchData.PRNumber, opts.owner, opts.repo)
+		var lastAttempt bool
+		if attempt == getPRAttempts-1 {
+			lastAttempt = true
+		}
+		info, err := comm.GetPullRequestInfo(ctx, td, conf.GithubPatchData.PRNumber, opts.owner, opts.repo, lastAttempt)
 		if err != nil {
 			return false, errors.Wrap(err, "getting pull request data from GitHub")
 		}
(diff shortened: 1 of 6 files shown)


agent/command/git.go
```diff
@@ -350,9 +350,13 @@ func (c *gitFetchProject) waitForMergeableCheck(ctx context.Context, comm client
 		getPRRetryMaxDelay = 15 * time.Second
 	)
 	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
-
+	attempt := 0
 	err := utility.Retry(ctx, func() (bool, error) {
-		info, err := comm.GetPullRequestInfo(ctx, td, conf.GithubPatchData.PRNumber, opts.owner, opts.repo)
+		var lastAttempt bool
+		if attempt == getPRAttempts-1 {
```
Kimchelly, line 356: Seems like `attempt` is supposed to be incremented somewhere?

rest/route/agent.go
```diff
@@ -168,6 +168,15 @@ func (h *agentCheckGetPullRequestHandler) Run(ctx context.Context) gimlet.Respon
 		Mergeable:      pr.Mergeable,
 		MergeCommitSHA: pr.GetMergeCommitSHA(),
 	}
+	if h.req.LastRetry && (!utility.FromBoolPtr(resp.Mergeable) || resp.MergeCommitSHA == "") {
+		grip.Debug(message.Fields{
+			"message": "last attempt to retry getting pull request",
+			"route":   "/pull_request",
+			"ticket":  "EVG-19723",
+			"resp":    resp,
```
Kimchelly, line 176: Is there other useful information we can get out of the PR, like the `pr.State` or something?

Kimchelly (changes requested):

agent/command/git.go
ablack12: yes duh thank you

rest/route/agent.go
ablack12: sure i can add a couple of things

Kimchelly (approved):

PR #6453, opened 2023-04-27
EVG-18405: support build variant patch_only

Kimchelly:
EVG-18405

### Description
The original ticket asks for being able to specify `patch_only` and `disable` at the build variant level. This is half of the requested change and only includes `patch_only`. `disable` already exists but it appears to be non-functional (it also has some problems in that it's not a pointer, so there's no way to distinguish between it being unset and it being set to false). To reduce the riskiness of these changes, I'm going to include a follow-up PR to fix build variant level `disable`.

For completeness' sake, I will also have a follow-up PR to support other fields at the build variant level, such as `patchable`, `allow_for_git_tag` and `git_tag_only`.

* Refactor parser project translation to ensure that the build variant task unit's `Variant` is always set when it's translated to a project.
    * Add a debug log and graceful continue on error in a spot where I think `Variant` should be set but I can't confirm for certain. I'll remove the log in a follow-up PR once I'm sure it does not occur.
* Support `patch_only` at the build variant level.
* Improve some parser project doc comments where there's tricky logic and one case where I believe there's a bug but I can't tell.

### Testing
Added unit tests and tested in staging that `patch_only` worked. I also tested some extra patches to ensure that `Variant` really is being set and to spot check that it's not causing issues with configuring patches or generating tasks.

### Documentation
Added docs.

### DIFF
--- a/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
+++ b/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
@@ -251,7 +251,7 @@ Fields:
 -   `name`: an identification string for the variant
 -   `display_name`: how the variant is displayed in the Evergreen UI
 -   `run_on`: a list of acceptable distros to run tasks for that variant
-    a. The first distro in the list is the primary distro. The others
+    on. The first distro in the list is the primary distro. The others
     are secondary distros. Each distro has a primary queue, a queue of
     all tasks that have specified it as their primary distro; and a
     secondary queue, a queue of tasks that have specified it as a
--- a/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
+++ b/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
@@ -1,6 +1,4 @@
 # Set Up a Project Configuration File
-kim: TODO: add section about setting priority at different definition levels.
-
 Project configurations are how you tell Evergreen what to do. They
 contain a set of tasks and variants to run those tasks on, and are
 stored within the repository they test. Project files are written in a
@@ -538,7 +536,9 @@ task).
 
 To cause a task to only run in commit builds, set `patchable: false`.
 
-To cause a task to only run in patches, set `patch_only: true`.
+To cause a task to only run in patches, set `patch_only: true`. `patch_only: true` can also be set at the build variant
+level to apply this behavior to all tasks in the build variant. The build variant level setting can be overridden if
+it's been explicitly set in the task definition or in the task listed under the build variant.
 
 To cause a task to only run in versions NOT triggered from git tags, set
 `allow_for_git_tag: false`.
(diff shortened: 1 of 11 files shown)


model/generate.go
```diff
@@ -685,7 +685,7 @@ func (g *GeneratedProject) validateNoRedefine(cachedProject projectMaps) error {
 func isNonZeroBV(bv parserBV) bool {
 	if bv.DisplayName != "" || len(bv.Expansions) > 0 || len(bv.Modules) > 0 ||
 		bv.Disable || len(bv.Tags) > 0 ||
-		bv.BatchTime != nil || bv.Activate != nil || bv.PatchOnly != nil || bv.Stepback != nil || len(bv.RunOn) > 0 {
+		bv.BatchTime != nil || bv.PatchOnly != nil || bv.Stepback != nil || len(bv.RunOn) > 0 {
```
Kimchelly, line 688: @ablack12 I originally changed this to include `bv.Activate != nil` since it seems like we wouldn't allow overwriting the build variant level activate when it's already been set, but some of the generate tasks tests seem to be confirming that the overwriting is allowed. Is it intentional that you can overwrite an existing build variant activate setting?

docs/02-Configure-a-Project/01-Project-Configuration-Files.md
```diff
@@ -537,7 +536,9 @@ task).
 
 To cause a task to only run in commit builds, set `patchable: false`.
 
-To cause a task to only run in patches, set `patch_only: true`.
+To cause a task to only run in patches, set `patch_only: true`. `patch_only: true` can also be set at the build variant
```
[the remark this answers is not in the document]
Kimchelly: Done.

repotracker/repotracker.go
```diff
@@ -935,7 +935,6 @@ func createVersionItems(ctx context.Context, v *model.Version, metadata model.Ve
 				if !ok || !bvt.HasSpecificActivation() {
 					continue
 				}
-				bvt.Variant = buildvariant.Name
```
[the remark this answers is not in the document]
Kimchelly: This is not necessary because I changed project translation to always set `Variant` the variant, so `BuildVariantTaskUnit`s will always have it set.

docs/02-Configure-a-Project/01-Project-Configuration-Files.md
```diff
@@ -251,7 +251,7 @@ Fields:
 -   `name`: an identification string for the variant
 -   `display_name`: how the variant is displayed in the Evergreen UI
 -   `run_on`: a list of acceptable distros to run tasks for that variant
-    on. The first distro in the list is the primary distro. The others
+    a. The first distro in the list is the primary distro. The others
```
malikchaya2, line 254: nit: I think you accidentally changed "on" to "a"

Kimchelly: Fixed.

malikchaya2 (approved):

model/generate.go
Kimchelly: I'm going to merge this as-is, but I'd still be interested to know if activate can be overrridden. If so, I can update the docs in a follow-up PR to mention that activate is a special case field that _can_ be overridden in generate.tasks.

PR #6455, opened 2023-04-28
EVG-17997 Move container variant to self-tests.yml

hadjri:
EVG-17997

### Description
Now that we have confirmed that we can run full green patches in prod, we can run the same dogfooding in prod as we did in staging, and observe the behavior over some set of time.
### Testing
Ran full containerized patches with the BV in this PR in prod.


self-tests.yml
```diff
@@ -644,6 +644,29 @@ buildvariants:
...
+      - evg-container
+    expansions:
+      goos: linux
+      goarch: amd64
+      IS_DOCKER: true
+      GOROOT: /usr/local/go
+      mongodb_url: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu2004-5.0.14.tgz
+      decompress: tar zxvf
+    tasks:
+      - name: "dist-and-upload-clis"
```
Kimchelly, line 659: This shouldn't run `dist-and-upload-clis` because that's the official deployed dist, which is already created by `ubuntu2004`.

self-tests.yml
```diff
@@ -644,6 +644,29 @@ buildvariants:
...
+      - name: "dist-staging"
```
Kimchelly, line 660: Do we need this? Seems like it's not that useful since we already run `dist-unsigned`. `dist-staging` is only used because some devs preferred to run patches to create staging dists.

self-tests.yml
```diff
@@ -644,6 +644,29 @@ buildvariants:
...
+      - name: "dist-unsigned"
+        activate: false
+      - name: ".smoke"
+      - name: ".test"
+      - name: "js-test"
+      - name: ".linter"
+      - name: "docker-cleanup"
```
Kimchelly, line 667: This one cannot run because `test-thirdparty-docker` requires Docker, which is not recursively available when it's already running in a Docker container.

self-tests.yml
```diff
@@ -695,3 +718,15 @@ buildvariants:
     tasks:
       - name: ".agent .test"
       - name: ".cli .test"
+
+
+containers:
+  - name: evg-container
+    working_dir: /
+    image: "hadjri/evg-container-self-testing"
```
Kimchelly, line 726: This seems fine for the moment, but I believe we'll have to replace this with one of the security-vetted images, since we'll be requiring all users (including us) to run pre-vetted containers. Can we make a ticket to replace it?

Kimchelly (changes requested):

self-tests.yml
hadjri: Ah, thought it was passing but you're right it's just [no-opping](https://spruce.mongodb.com/task/evergreen_ubuntu2004_container_docker_cleanup_patch_65b839a80f0a9cb9523f25bbdbae4a3b93d2ffe2_644c1354850e615d1b3adaa4_23_04_28_18_41_34/logs?execution=0&sortBy=STATUS&sortDir=ASC)

self-tests.yml
hadjri: Was trying to replicate our 20.04 variant but yeah that sounds reasonable. Removed

self-tests.yml
hadjri: https://jira.mongodb.org/browse/EVG-19745

Kimchelly (approved):

PR #673, opened 2018-02-05
EVG-2601 Patch alias does not support display tasks

richardsamuels:
This brings partial support of display tasks to patch aliases. EVG-2722 has been created to track the remaining issues in display tasks.

PR #6459, opened 2023-05-01
EVG-18405: support disabling build variants

Kimchelly:
EVG-18405

### Description
Add support for `disable` at the build variant level. While this field already exists at the build variant level, I didn't find any project YAMLs using it, and I believe the existing usage was buggy, so I rewrote it.

* Allow `disable` to be specified at build variant level. It has lower precedence than disable settings specified at the task or build variant task levels.
* Update a few doc/code comments for clarity.
* Drive-by fix a couple markdown errors in the docs.

### Testing
Added unit tests and tested in staging patches that `disable: true` at the build variant level works as I described.

### Documentation
Updated docs and described precedence of conflicting settings.

### DIFF
--- a/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
+++ b/docs/02-Configure-a-Project/01-Project-Configuration-Files.md
@@ -532,15 +532,13 @@ early_termination:
 
 ### Limiting When a Task Will Run
 
-The following can be added to a task definition OR to a task listed
-under a build variant (so that it will only effect that variant's
-task).
+To limit the conditions when a task will run, the following settings can be
+added to a task definition, to a build variant definition, or to a specific task
+listed under a build variant (so that it will only affect that variant's task).
 
 To cause a task to only run in commit builds, set `patchable: false`.
 
-To cause a task to only run in patches, set `patch_only: true`. `patch_only: true` can also be set at the build variant
-level to apply this behavior to all tasks in the build variant. The build variant level setting can be overridden if
-it's been explicitly set in the task definition or in the task listed under the build variant.
+To cause a task to only run in patches, set `patch_only: true`.
 
 To cause a task to only run in versions NOT triggered from git tags, set
 `allow_for_git_tag: false`.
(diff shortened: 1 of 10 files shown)


model/project.go
```diff
@@ -1693,9 +1700,6 @@ func (p *Project) ResolvePatchVTs(patchDoc *patch.Patch, requester, alias string
 	if len(tasks) == 1 && tasks[0] == "all" {
 		tasks = []string{}
 		for _, t := range p.Tasks {
-			if !utility.FromBoolTPtr(t.Patchable) || utility.FromBoolPtr(t.GitTagOnly) {
```
Kimchelly, line 1696: Note that this check is not actually necessary (and also has a bug because it ignores precedence rules for these two fields). Checking for a matching requester on L1726 performs the same check while respecting precedence.

model/project.go
```diff
@@ -1915,7 +1926,9 @@ func (p *Project) extractDisplayTasks(pairs TaskVariantPairs) TaskVariantPairs {
 }
 
 // BuildProjectTVPairsWithAlias returns variants and tasks for a project alias.
-func (p *Project) BuildProjectTVPairsWithAlias(vars []ProjectAlias) ([]TVPair, []TVPair, error) {
+// This filters out tasks that cannot run due to being disabled or having an
+// unmatched requester (e.g. a patch-only task for a mainline commit).
+func (p *Project) BuildProjectTVPairsWithAlias(vars []ProjectAlias, requester string) ([]TVPair, []TVPair, error) {
```
hadjri, line 1931: While we're at it here can we rename `vars` to something like `aliases` or `projectAliases`?

model/project.go
```diff
@@ -1932,29 +1945,31 @@ func (p *Project) BuildProjectTVPairsWithAlias(vars []ProjectAlias) ([]TVPair, [
...
+				if bvtu := p.FindTaskForVariant(t.Name, variant.Name); bvtu != nil {
+					if bvtu.IsDisabled() || bvtu.SkipOnRequester(requester) {
 						continue
 					}
-					displayTaskPairs = append(displayTaskPairs, TVPair{variant.Name, displayTask.Name})
+					pairs = append(pairs, TVPair{variant.Name, t.Name})
+				}
+			}
+
+			if v.Task == "" {
```
hadjri, line 1965: Why do we wait until right before display task construction to do this?

docs/02-Configure-a-Project/01-Project-Configuration-Files.md
```diff
@@ -532,32 +532,51 @@ early_termination:
...
 `git_tag_only: true`.
 
-To cause a task to not run at all, set `disable: true`. Setting `disable: true` at the build variant level 
-will apply this behavior to all tasks in the build variant.
+To cause a task to not run at all, set `disable: true`.
 
 -   This behaves similarly to commenting out the task but will not
     trigger any validation errors.
--   Disabling its dependencies will still allow the task to run
+-   If a disable is disabled and is depended on by another task, the
```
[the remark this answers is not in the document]
Kimchelly: Yes, English is hard. 🤦

model/generate.go
```diff
@@ -683,8 +683,10 @@ func (g *GeneratedProject) validateNoRedefine(cachedProject projectMaps) error {
 }
 
 func isNonZeroBV(bv parserBV) bool {
+	// Note: this omits activate from consideration, but it's unclear if it's
+	// intentional or not.
```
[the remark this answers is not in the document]
Kimchelly: Sure, I made EVG-19783.

model/project.go
Kimchelly: Sure, I renamed it to `aliases` + a couple other places in this file.

model/project.go
Kimchelly: This was how the logic was already, so I just left it. The original PR (#673) didn't make any comment about it, but if I had to guess, it's indirectly checking that [`taskRegex` is non-nil](https://github.com/evergreen-ci/evergreen/blob/c89980b58abab4449e6b5c374bc506cfaad29bbd/model/project_aliases.go#L538-L540) before using it to check display tasks. I changed it to check for a non-nil `taskRegex` to make it clearer.

hadjri (approved):

PR #6461, opened 2023-05-02
EVG-18405: add other build variant requester settings

Kimchelly:
EVG-18405

### Description
* Add other build variant requester-related settings for the sake of completeness (`patchable`, `allow_for_git_tag`, `git_tag_only`).
* Remove completed TODO since the variant is always set. I left in the error log so we can trace if it ever happens again (which is a sign of a regression).

### Testing
Added unit tests.

### Documentation
The docs were already updated in #6459 to mention that all requester-related settings are supported at the build variant level.


model/project.go
```diff
@@ -1548,17 +1560,12 @@ func (p *Project) tasksFromGroup(bvTaskGroup BuildVariantTaskUnit) []BuildVarian
 	}
 	bv := p.FindBuildVariant(bvTaskGroup.Variant)
 	if bv == nil {
-		// TODO (EVG-18405): remove this and return early after confirming that
-		// this does not log. Continuing on error with an empty build variant is
-		// inconsequential for now, since it is only necessary for checking
-		// build variant level patch_only.
 		grip.Error(message.Fields{
-			"message":       "found a task group that has no associated build variant (this is not supposed to happen), using an empty build variant configuration as a temporary workaround",
+			"message":       "programmatic error: found a task group that has no associated build variant (this is not supposed to ever happen and is probably a bug)",
```
malikchaya2, line 1564: Do we want to add an alert on this so that we can fix whatever the error is?

model/project.go
```diff
@@ -321,8 +330,11 @@ type BuildVariant struct {
 	CronBatchTime string `yaml:"cron,omitempty" bson:"cron,omitempty"`
 
 	// If Activate is set to false, then we don't initially activate the build variant.
-	Activate  *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
-	PatchOnly *bool `yaml:"patch_only,omitempty" bson:"patch_only,omitempty"`
+	Activate       *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
+	Patchable      *bool `yaml:"patchable,omitempty" bson:"patchable,omitempty"`
+	PatchOnly      *bool `yaml:"patch_only,omitempty" bson:"patch_only,omitempty"`
```
malikchaya2, line 335: I think we should add the rest of the tags to the docs here since we list them all out
https://github.com/Kimchelly/evergreen/blob/main/docs/02-Configure-a-Project/01-Project-Configuration-Files.md?plain=1#L290

malikchaya2 (changes requested):

malikchaya2 (approved):

PR #6463, opened 2023-05-02
EVG-19767 Update go-github package to v52

bynn:
EVG-19767

### Description
upgraded the library,
some functions changed their return type or parameters so i added the necessary changes for it to compile 

### Testing
restarted tasks on staging to make sure nothing broke


model/version.go
```diff
@@ -741,7 +741,7 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 			// Otherwise, retrieve the module's commit from the time of the patch creation.
 			revisionTime := time.Unix(0, 0)
 			if !evergreen.IsPatchRequester(v.Requester) {
-				revisionTime = commit.Commit.Committer.GetDate()
+				revisionTime = *commit.Commit.Committer.Date.GetTime()
```
ablack12, line 744: Have we verified that this returns things in the same way? I'm also concerned about nil pointer errors so we may want to make a thirdparty helper function that returns this time and returns the empty time if it's nil or something.

ablack12 (changes requested):

model/version.go
bynn: verified by making a mainline commit on sandbox without erroring 

a wrapper function shouldn't be needed with the current method because it already nil checks on both the committer and date.

model/version.go
```diff
@@ -741,7 +741,7 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 			// Otherwise, retrieve the module's commit from the time of the patch creation.
 			revisionTime := time.Unix(0, 0)
 			if !evergreen.IsPatchRequester(v.Requester) {
-				revisionTime = *commit.Commit.Committer.Date.GetTime()
+				revisionTime = commit.Commit.Committer.GetDate().Time
```
ablack12, line 744: such nuance haha love it

ablack12 (approved):

rest/data/patch.go
```diff
@@ -187,7 +187,7 @@ func AbortPatchesFromPullRequest(event *github.PullRequestEvent) error {
 		return err
 	}
 
-	if err = model.AbortPatchesWithGithubPatchData(*event.PullRequest.ClosedAt, true, "", owner, repo, *event.Number); err != nil {
+	if err = model.AbortPatchesWithGithubPatchData(*event.PullRequest.ClosedAt.GetTime(), true, "", owner, repo, *event.Number); err != nil {
```
ablack12, line 190: If we can update this to also not access pointers that would be cool but I see we already did that so alas

bynn: draft

PR #6465, opened 2023-05-02
EVG-17989 Remove REST route toggleable auth flag

hadjri:
EVG-17989

### Description
We currently have two admin flags that allow us to toggle authentication locks in Evergreen: one that will lock the entire legacy UI behind auth, and another that will lock all other APIs behind auth. The latter has been enabled for quite a while without any complaints from users, and I don't anticipate we'll want to un-auth any REST routes in the future. Removing this flag and its associated middleware logic.

The former flag currently remains because toggling authentication for the legacy UI is a useful tool and something we can resort to during a fire (as we did during [EVG-19610](https://jira.mongodb.org/browse/EVG-19610), although I've renamed it to `LegacyUIPublicAccessDisabled` rather than `UIPartialAuthDisabled`, as it is a more understandable name (imo).
### Testing
Tested locally


operations/commit_queue.go
```diff
@@ -636,6 +636,8 @@ func (p *mergeParams) uploadMergePatch(conf *ClientSettings, ac *legacyClient, u
 	if err = patchParams.validateSubmission(diffData); err != nil {
 		return err
 	}
+	//here we go. test here tomorrow and compare it with the regular patch command
```
ablack12, line 639: we should remove this i imagine?

units/patch_intent.go
```diff
@@ -721,6 +721,7 @@ func (j *patchIntentProcessor) buildCliPatchDoc(ctx context.Context, patchDoc *p
 	}
 
 	if len(patchDoc.Patches) > 0 {
+		// hereee
```
ablack12, line 724: also this

ablack12 (changes requested):

operations/commit_queue.go
hadjri: LOL my bad. I'm glad it wasn't anything more embarrassing than this.

service/middleware.go
```diff
@@ -119,7 +119,7 @@ func (uis *UIServer) GetSettings() evergreen.Settings {
 // requireUser takes a request handler and returns a wrapped version which verifies that requests
 // request are authenticated before proceeding. For a request which is not authenticated, it will
 // execute the onFail handler. If onFail is nil, a simple "unauthorized" error will be sent.
-// If skipWithToggle is true, the request for an authenticated user will be skipped if the UIPartialAuthDisabled
+// If skipWithToggle is true, the request for an authenticated user will be skipped if the LegacyUIPublicAccessDisabled
```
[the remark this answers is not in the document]
hadjri: Updated the comment. This confusion stems from the fact that in the admin page it'll say "Enabled" when that really means `false` in the backend which is confusing.

ablack12 (approved):

PR #6466, opened 2023-05-02
EVG-18260 verify that hidden project can be accessed via rest API

ablack12:
EVG-18260 

### Description
Moved project variable creation into Add() since we assume that both project ref and project var must exist for any project in lots of places.

In thinking through this, realized we should gate hidden projects from being modifiable via route or UI since they're meant to be backend only skeleton projects.

### Testing
Verified in staging that a hidden project could be accessed via rest API if the variables doc existed.

### DIFF
--- a/rest/route/project.go
+++ b/rest/route/project.go
@@ -328,6 +328,9 @@ func (h *projectIDPatchHandler) Parse(ctx context.Context, r *http.Request) erro
 
 // Run updates a project by name.
 func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
+	if h.newProjectRef.IsHidden() {
+		return gimlet.NewJSONErrorResponse("can't patch a hidden project")
+	}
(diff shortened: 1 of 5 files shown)


rest/route/project.go
```diff
@@ -328,6 +328,9 @@ func (h *projectIDPatchHandler) Parse(ctx context.Context, r *http.Request) erro
 
 // Run updates a project by name.
 func (h *projectIDPatchHandler) Run(ctx context.Context) gimlet.Responder {
+	if h.newProjectRef.IsHidden() {
```
Kimchelly, line 331: Should this also apply to other patch creation routes like submitting commit queue patches and such? Or do we assume that users wouldn't try submitting work for a hidden project, and this is just a spot check?

Kimchelly (changes requested):

rest/route/project.go
ablack12: Submitting commit queue should hit an issue with the commit queue being disabled but yeah generally this is just a spot check.

Kimchelly (approved):

PR #6471, opened 2023-05-03
EVG-18346 Point to different links for multi-commit merge patch diff links

hadjri:
EVG-18346

### Description
Currently, HTML diff links only accept a filename parameter when templating the data. This means for commit queue merge patches that have multiple commits as part of their changes panel such as [this one](https://spruce.mongodb.com/version/637ba3350305b9100cee7294/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), if the same file is changed between both commits, they will point to the same diff, i.e. the first one matching the filename in the full diff file.

Introduced a change to track which commit we are on in the templating logic and serve the data accordingly. Also did a drive-by change fixing a bug I noticed where for multi-commit diffs the beginning of the next commit bleeds into the top of the preceding commit's changes ([example](https://evergreen.mongodb.com/filediff/637ba3350305b9100cee7294/?file_name=evergreen%2Flint_fuzzer_sanity_all.sh&patch_number=0))

### Testing
Unit testing, tested in staging and locally.

### DIFF
--- a/rest/model/patch.go
+++ b/rest/model/patch.go
@@ -314,13 +314,8 @@ func (apiPatch *APIPatch) buildModuleChanges(p patch.Patch, identifier string) {
 		htmlLink := fmt.Sprintf("%s/filediff/%s?patch_number=%d", apiURL, *apiPatch.Id, patchNumber)
 		rawLink := fmt.Sprintf("%s/rawdiff/%s?patch_number=%d", apiURL, *apiPatch.Id, patchNumber)
 		fileDiffs := []FileDiff{}
-		commitNumber := 0
-		prevCommitMsg := ""
 		for i, file := range modPatch.PatchSet.Summary {
-			if i > 0 && file.Description != prevCommitMsg {
-				commitNumber++
-			}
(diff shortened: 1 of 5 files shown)


rest/model/patch.go
```diff
@@ -314,8 +314,13 @@ func (apiPatch *APIPatch) buildModuleChanges(p patch.Patch, identifier string) {
...
 		fileDiffs := []FileDiff{}
-		for _, file := range modPatch.PatchSet.Summary {
-			diffLink := fmt.Sprintf("%s/filediff/%s?file_name=%s&patch_number=%d", apiURL, *apiPatch.Id, url.QueryEscape(file.Name), patchNumber)
+		commitNumber := 0
+		prevCommitMsg := ""
+		for i, file := range modPatch.PatchSet.Summary {
+			if i > 0 && file.Description != prevCommitMsg {
+				commitNumber++
+			}
+			diffLink := fmt.Sprintf("%s/filediff/%s?file_name=%s&patch_number=%d&commit_number=%d", apiURL, *apiPatch.Id, url.QueryEscape(file.Name), patchNumber, commitNumber)
```
ablack12, line 323: is there a reason to not just always append i as the commit number? Will these always be in order where we only need to consider the exact previous commit message (as opposed to the set of previous ones)?

hadjri: Sure, we might as well do that. This route is just reading from the full diff which will be the same order every time.

ablack12 (approved):

PR #6476, opened 2023-05-04
EVG-18533 Error on NaN and Inf testcase times

hadjri:
EVG-18533

### Description
NaN no longer causes a negative duration for tests, rather it results in a `0` value for the duration, which is still incorrect. Made a change to propagate the logger down to the test level with errors if `NaN` or `Inf` are given for the XML time. `Inf` and `-Inf` do still create nonsensical durations so added a line to set the test's end time to 0 if this occurs.
### Testing
Tested in staging with one NaN and one Inf testcase, they both [errored as expected](https://evergreen-staging.corp.mongodb.com/task_log_raw/evg_ubuntu2004_test_util_patch_631f3744785a0d5fce891a48a6be631d19909356_6454162097b1d3f48c56e2c7_23_05_04_20_31_40/0?type=T#L330)


agent/command/results_xunit_parser.go
```diff
@@ -106,6 +108,14 @@ func (tc testCase) toModelTestResultAndLog(conf *internal.TaskConfig) (testresul
 	// Replace spaces, dashes, etc. with underscores.
 	res.TestName = util.CleanForPath(res.TestName)
 
+	if math.IsNaN(float64(tc.Time)) {
+		logger.Task().Errorf("Test '%s' time was NaN, its calculated duration will be incorrect", res.TestName)
+	}
+	if math.IsInf(float64(tc.Time), 0) || math.IsInf(float64(tc.Time), -1) {
```
bynn, line 114: nit: it wasn't obvious to me why we passed in 0 to IsInf so could we add a little comment saying its to check for both -inf and inf

agent/command/results_xunit_parser_test.go
```diff
@@ -222,17 +224,19 @@ func TestXMLToModelConversion(t *testing.T) {
...
 			logs := []*model.TestLog{}
 			for _, testCase := range res[0].TestCases {
-				test, log := testCase.toModelTestResultAndLog(conf)
+				test, log := testCase.toModelTestResultAndLog(conf, logger)
 				if log != nil {
 					logs = append(logs, log)
 				}
 				tests = append(tests, test)
 			}
+			So(logger.Close(), ShouldBeNil)
```
Kimchelly, line 239: Can this test also assert somewhere that the test data added [here](https://github.com/evergreen-ci/evergreen/blob/a37cefa7b9b54209c2fc95c75116ea06db474ba8/agent/command/testdata/xunit/junit_3.xml#L7-L8) translates into sensible test start and end times?

bynn (approved):

agent/command/results_xunit_parser.go
Kimchelly, line 114: Per [the docs](https://pkg.go.dev/math@go1.20.3#IsInf), it seems like checking `math.IsInf(float64(tc.Time), 0)` checks both positive and negative infinity, so it seems like the second `-1` check can be omitted.

agent/command/results_xunit_parser.go
```diff
@@ -106,6 +108,14 @@ func (tc testCase) toModelTestResultAndLog(conf *internal.TaskConfig) (testresul
 	// Replace spaces, dashes, etc. with underscores.
 	res.TestName = util.CleanForPath(res.TestName)
 
+	if math.IsNaN(float64(tc.Time)) {
+		logger.Task().Errorf("Test '%s' time was NaN, its calculated duration will be incorrect", res.TestName)
```
Kimchelly, line 112: Is this supposed to transform the NaN into a zero value as well?

Kimchelly (changes requested):

agent/command/results_xunit_parser_test.go
hadjri: Ah yes, thanks

agent/command/results_xunit_parser.go
hadjri: Ah yes, thanks

agent/command/results_xunit_parser.go
hadjri: No because `NaN` already results in the `res.TestEndTime` below equating to whatever `TestStartTime` was, so we don't need to modify the time value here to prevent `TestEndTime` from evaluating to something strange

agent/command/testdata/xunit/junit_3.xml
```diff
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<testsuite name="nosetests" tests="648" errors="0" failures="1" skip="188">
+<testsuite name="nosetests" tests="650" errors="0" failures="1" skip="188">
     <testcase classname="test.test_auth.TestAuthURIOptions" name="test_uri_options" time="0.002">
         <skipped type="unittest.case.SkipTest" message="Authentication is not enabled on server"><![CDATA[SkipTest: Authentication is not enabled on server
 ]]></skipped>
     </testcase>
+    <testcase classname="test.test_binary.TestBinary" name="test_nan" time="NaN"></testcase>
+    <testcase classname="test.test_binary.TestBinary" name="test_inf" time="Inf"></testcase>
```
julianedwards, line 8: Could you add a case with `-Inf` here as well?

julianedwards (changes requested):

Kimchelly (approved):

julianedwards (approved):

PR #6479, opened 2023-05-05
EVG-19692: Fetch user subscriptions

sophstad:
EVG-19792

### Description
- Add subscriptions field to user resolver
- I noticed the GraphQL `ProjectSubscription` type actually represents all subscriptions and would like to reuse one type. For now, there is `ProjectSubscription` and `GenericSubscription` with a plan to merge...
  - To facilitate renaming the `ProjectSubscription` type to `GenericSubscription`, I added a `projectSubscriptions` field to the types that return subscriptions. The front end will be updated to point at this field, then `subscriptions` will be updated here to use a `GenericSubscription`, then the front end will be updated again to point at `subscriptions` with the new type, and finally `projectSubscriptions` will be removed from app code.
  - Also, `Subscription` is a reserved keyword for GraphQL which is why I went with `GenericSubscription` as the type name... open to better ideas! 😅 

### Testing
- Update `user` GraphQL unit test to return a subscription
- Add unit test verifying `projectSubscriptions` returns the same data as `subscriptions`

### Documentation
N/A


gqlgen.yml
```diff
@@ -68,6 +68,8 @@ models:
     model: github.com/evergreen-ci/evergreen/rest/model.APIFile
   FileDiff:
     model: github.com/evergreen-ci/evergreen/rest/model.FileDiff
+  GenericSubscription:
```
malikchaya2, line 71: I have a slight preference for not using the word generic here since this behaves a bit differently than generic types usually do in OOP. I'd prefer something like GeneralSubscription instead.

malikchaya2 (approved):

PR #6480, opened 2023-05-05
EVG-19806 refresh metrics exporter

ybrill:
[EVG-19806](https://jira.mongodb.org/browse/EVG-19806)

### Description
The metrics exporter was getting closed after the first task ran on the agent. This was happening because the metrics provider Shutdown ends up calling the exporter's shutdown. The solution here is simple: make a new exporter every time a task starts.

Unrelated: `instrument.With...` is deprecated in favor of `metric.With...` so this switches them over.

### Testing
I'll take it for a spin in staging when I can get the lock.


ybrill:
checks out in staging.
Also added the task attributes to the metrics provider resource so we won't have to add them to every single metric.

opsnlops (approved):

PR #6483, opened 2023-05-08
EVG-19816: properly catch scanner errors when parsing go tests

julianedwards:
[EVG-NNNNN](https://jira.mongodb.org/browse/EVG-19816)

### Description
Move logic to catch scanner errors outside the `Scan` for loop otherwise they will never get caught.

### Testing
New unit test to make sure we are actually propagating the error.

### Documentation
N/A


ablack12:
Re the js-test @julianedwards this is being investigated in https://jira.mongodb.org/browse/EVG-19773

Kimchelly (approved):

PR #6485, opened 2023-05-08
EVG-14033 Introduce option to toggle patch visibility

khelif96:
EVG-14033

### Description
Adds a flag to patches to toggle their visibility and updates the patches page queries to case on it. 
Introduce a mutation to toggle patch visibility


model/patch/patch.go
```diff
@@ -674,6 +675,18 @@ func (p *Patch) SetActivation(activated bool) error {
 	)
 }
 
+func (p *Patch) SetPatchVisibility(hidden bool) error {
```
ablack12, line 678: Can we add a function comment here?

graphql/mutation_resolver.go
```diff
@@ -223,6 +224,31 @@ func (r *mutationResolver) EnqueuePatch(ctx context.Context, patchID string, com
 	return newPatch, nil
 }
 
+// SetPatchVisibility is the resolver for the setPatchVisibility field.
+func (r *mutationResolver) SetPatchVisibility(ctx context.Context, patchIds []string, hidden bool) ([]*restModel.APIPatch, error) {
+	updatedPatches := []*restModel.APIPatch{}
+	for _, patchID := range patchIds {
+		patch, err := patch.FindOneId(patchID)
```
ablack12, line 231: Can we do this in one query instead?

ablack12 (changes requested):

graphql/mutation_resolver.go
khelif96: Great catch!

ablack12 (approved):

PR #6486, opened 2023-05-09
EVG-19827 Add logging to refresh in commit queue

bynn:
EVG-19827

### Description
add logging to see how often this happens
will be deleted after investigation

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -397,6 +397,13 @@ func getPRAndCheckMergeable(ctx context.Context, env evergreen.Environment, sc C
 				"pr":      info.PR,
 			}))
 		} else if p != nil && p.Version != "" {
+			grip.Debug(message.Fields{
+				"message": "calling refresh from commit queue",
+				"ticket":  "EVG-19827",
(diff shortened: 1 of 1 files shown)


rest/data/commit_queue.go
```diff
@@ -397,6 +397,13 @@ func getPRAndCheckMergeable(ctx context.Context, env evergreen.Environment, sc C
 				"pr":      info.PR,
 			}))
 		} else if p != nil && p.Version != "" {
+			grip.Debug(message.Fields{
+				"message": "calling refresh from commit queue",
```
ablack12, line 401: might also be interesting to log if this actually changes this status, so maybe logging after 412 instead and logging mergeableState

ablack12 (approved):
LGTM with thought

ablack12:
Ticket doesn't have a description but I imagine we intend to look at the logging and remove it before closing the ticket?

rest/data/commit_queue.go
bynn: moved and printing new mergable state!

PR #6487, opened 2023-05-09
EVG-17996 Add exec platform to system failed logs

hadjri:
EVG-17996

### Description
`execution_platform` needs to be a part of this log for it to be reflected in the "Container Tasks Finished by Status" dashboard.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1335,13 +1335,14 @@ func (t *Task) MarkSystemFailed(description string) error {
 		event.LogTaskFinished(t.Id, t.Execution, evergreen.TaskSystemFailed)
 	}
 	grip.Info(message.Fields{
-		"message":     "marking task system failed",
-		"usage":       "container task health dashboard",
-		"task_id":     t.Id,
-		"execution":   t.Execution,
-		"status":      t.Status,
-		"host_id":     t.HostId,
-		"description": description,
+		"message":            "marking task system failed",
+		"usage":              "container task health dashboard",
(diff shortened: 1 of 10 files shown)


model/task/task.go
```diff
@@ -1335,13 +1335,14 @@ func (t *Task) MarkSystemFailed(description string) error {
...
 	grip.Info(message.Fields{
-		"message":     "marking task system failed",
-		"usage":       "container task health dashboard",
-		"task_id":     t.Id,
-		"execution":   t.Execution,
-		"status":      t.Status,
-		"host_id":     t.HostId,
-		"description": description,
+		"message":            "marking task system failed",
+		"usage":              "container task health dashboard",
```
ablack12, line 1339: If this is being used for the container task health dashboard, can we just only log this for container tasks? Seems unnecessary to log this for host tasks.

ablack12 (changes requested):

model/task/task.go
hadjri: The `usage` field is just a signal that these logs are depended on to populate the container health dashboard, it's not meant to signify that it's the only purpose for the log. In this case we have been logging this for system-failed tasks for a while

ablack12: Ah -- can we refactor "container task health dashboard" to be a constant if we use this for a dashboard we expect to keep around for a while? It'd also be a bit clearer if "usage" was something like "included_on" to be clear that it's not just for that, but not sure if we rely on that.

PR #6490, opened 2023-05-10
EVG-19227 fix bug with patch vs version statuses

ablack12:
EVG-19227
### Description
The main bug here is that in the version trigger, we check for VersionSucceeded, but CollectiveStatus returns a patch status, and PatchSucceeded != VersionSucceeded (I think it's time to refactor these to be the same and will make a ticket from this).

Found a couple of other things: we use "PatchAborted" as a display status even though we should expect to see this for versions since it's a display-only status. The UI function to calculate status was modified to consider Aborted but fetched more information than it needed, and didn't convert the patch status back to version status. cc @evergreen-ci/evg-ui want to make sure the UI knows to expect "success" instead of "succeeded", if not we can tweak this and follow up in the refactor ticket.

### Testing
Moved / augmented existing resolver test.


model/patch/patch.go
```diff
@@ -765,8 +765,9 @@ func (p *Patch) IsChild() bool {
 	return p.Triggers.ParentPatch != ""
 }
 
-// CollectiveStatus returns the aggregate status of all tasks and child patches.
-func (p *Patch) CollectiveStatus() (string, error) {
+// CollectivePatchStatus returns the aggregate status of all tasks and child patches.
+// If this is meant for display on the UI, we should also consider the display status aborted.
+func (p *Patch) CollectivePatchStatus() (string, error) {
```
Kimchelly, line 770: The previous name was actually more stylistically Go because it didn't [stutter](https://google.github.io/styleguide/go/best-practices#avoid-repetition). Can we change it back?

model/patch/patch.go
```diff
@@ -1228,7 +1218,7 @@ func GetCollectiveStatus(statuses []string) string {
 	} else if hasFailure {
 		return evergreen.PatchFailed
 	} else if hasAborted {
-		return evergreen.PatchAborted
+		return evergreen.VersionAborted
```
Kimchelly, line 1221: This is kind of weird to put a version status in a patch function, and I suspect someone will one day think this is a bug (it looks like one). Can we just make a `evergreen.PatchAborted` status and set it to the same string value as `evergreen.VersionAborted`? We can just translate it to the version status if needed in `PatchStatusToVersionStatus`.

Kimchelly (changes requested):

model/patch/patch.go
```diff
@@ -1178,21 +1179,9 @@ func (p PatchesByCreateTime) Swap(i, j int) {
...
-	}
-	return p.CollectiveStatus()
-}
-
-// GetCollectiveStatus answers the question of what the patch status should be
-// when the patch status and the status of its children are different
-func GetCollectiveStatus(statuses []string) string {
+// GetCollectivePatchStatusFromStatuses answers the question of what the patch status should be
+// when the patch status and the status of its children are different, given a list of statuses.
+func GetCollectivePatchStatusFromStatuses(statuses []string) string {
```
Kimchelly, line 1184: Should there be a test for this function?

model/patch/patch.go
ablack12: hmm I just want to clarify that it returns the _patch_ status and not the version status since that's easy to forget those are different. If we make a follow up ticket to consolidate these, could we fix it at that point? Otherwise I can just leave a comment to be more aggressively specific about it.

model/patch/patch.go
ablack12: Yeah that's fair

model/version.go
```diff
@@ -129,6 +129,50 @@ func (v *Version) ActivateAndSetBuildVariants() error {
 	)
 }
 
+// GetDisplayStatus considers both child patch statuses and
+// aborted status, and returns an overall status.
+func (v *Version) GetDisplayStatus() (string, error) {
```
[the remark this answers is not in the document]
ablack12: This does ultimately return a version status -- we convert it to patch to get the cumulative but then we convert it back. I can add a comment to that effect.

graphql/version_resolver.go
```diff
@@ -217,12 +217,15 @@ func (r *versionResolver) ProjectMetadata(ctx context.Context, obj *restModel.AP
 
 // Status is the resolver for the status field.
 func (r *versionResolver) Status(ctx context.Context, obj *restModel.APIVersion) (string, error) {
```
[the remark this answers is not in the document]
ablack12: Thanks! I'm going to make a new ticket to migrate these to be the same but in the meantime we'll do a conversion here.

model/patch/patch.go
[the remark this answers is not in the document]
ablack12: Could alternatively call this GetCollectiveStatusFromPatchStatuses to indicate that should be the input?

model/version.go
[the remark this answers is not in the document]
ablack12: Okay _actually_ since this is only used by the resolver which relies on the patch status (as referenced by minna) i'll just have this return the patch status.

Kimchelly (approved):
LGTM mod non-that-important suggestion.

model/patch/patch.go
ablack12: Well the tricky thing is I think we forget that PatchStatus and VersionStatus aren't the same, since that's what the bug was I found here.

ablack12: But yeah maybe I'll just make the comment aggressive and that could do the trick, plus we now know to be suspicious of these statuses.

minnakt (approved):
patch statuses are back! 😸

PR #6491, opened 2023-05-11
EVG-18985 Deprecate unused project fields

bynn:
EVG-18985

### Description
these fields are only being set and just checked in testing 
i put a deprecation notice on the fields just in case we need them for the github project that im working on (i don't think so though)

### Testing
nothing should break

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -37,16 +37,16 @@ const (
 )
 
 type Project struct {
-	Enabled            bool                       `yaml:"enabled,omitempty" bson:"enabled"`
+	Enabled            bool                       `yaml:"enabled,omitempty" bson:"enabled"`         // deprecated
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):
LGTM mod question. Is there going to be a follow-up ticket (or maybe reuse this ticket and bump it a few sprints) to delete these fields once we've determined if we need them?

bynn:
> LGTM mod question. Is there going to be a follow-up ticket (or maybe reuse this ticket and bump it a few sprints) to delete these fields once we've determined if we need them?

Honestly I'm not sure what the protocol is for these deprecated fields because I do see a couple of them in the codebase already. I heard some of the things are kept just in case it's used in the future?

bynn:
@Kimchelly https://jira.mongodb.org/browse/EVG-19862

created a ticket to delete all deprecated fields in the future

PR #6496, opened 2023-05-15
EVG-19251 capture command traces

ybrill:
[EVG-19251](https://jira.mongodb.org/browse/EVG-19251)

### Description
Provide two ways for user commands to send their traces to the collector. 

The first is straightforward: we'll expose the necessary parameters as expansions so any test that wants them can get them and use them to send stuff to the collector themselves. 

The latter is to define a predetermined location for tests to drop a serialized trace file that we'll pick up, parse, and ship off to the collector. The [JSON protobuf format](https://opentelemetry.io/docs/specs/otel/protocol/otlp/#json-protobuf-encoding) seems to have been intended for this use case. It's output natively by the Java SDK (which I presume will work for cloud which is the primary audience for this feature anyway), and by the fileexporter in the contrib distribution of the collector. Parsing it is a bit of a hack: the supported library for parsing protobuf json is [protojson](https://pkg.go.dev/google.golang.org/protobuf@v1.30.0/encoding/protojson). But using [the generated go protobuf files](https://github.com/open-telemetry/opentelemetry-proto-go) results in the trace/span ids getting mangled since the official mapping uses base64 encoding for binary while the OTel JSON encoding uses hex 😠 (see [this discussion](https://github.com/open-telemetry/opentelemetry-specification/issues/786)). For context, the collector uses [this deprecated fork](https://github.com/gogo/protobuf) that lets you set [a custom type](https://github.com/open-telemetry/opentelemetry-collector/blob/6542100317582afe3f730471244ccc9a8c331908/pdata/internal/data/protogen/trace/v1/trace.pb.go#L330) for [parsing the id hex](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/traceid.go) and they reimplement the entire trace data structure from scratch. This PR uses the standard parsing library and manually recovers the trace/span ids by encoding them back to a hex string and decoding.

One thing I wanted to point out about the uploading: after a file's been successfully uploaded the file is deleted. This simplifies things since we don't need to keep track of which files we've uploaded already in the case of task groups that have multiple tasks share the same working directory.

### Testing
I generated a trace with local-evergreen and a local collector running on my machine configured to export with the fileexporter. I has a task write that file to the trace directory and it ended up in Honeycomb (e.g. [this trace](https://ui.honeycomb.io/mongodb-4b/environments/staging/trace/pi8XwcAJcGM)).

### Documentation
Duly added.


agent/otel.go
```diff
@@ -279,6 +288,166 @@ func hostResource(ctx context.Context) (*resource.Resource, error) {
...
+// into []byte. The unmarshaller decoded the string as a base64 encoded
+// string so we encode it back to the string and decode it again to []byte.
+func fixBinaryID(id []byte) ([]byte, error) {
+	idHex := base64.StdEncoding.EncodeToString(id)
+	return hex.DecodeString(idHex)
+}
+
+// getTraceFiles returns the full path of all the files in the [traceSuffix] directory
+// under the task's working directory.
+func getTraceFiles(taskDir string) ([]string, error) {
```
rtimmons, line 422: @julianedwards do we want to model this similarly to `build/TestLogs` - should this be `build/OTelTraces` or something?

opsnlops (approved):
Good luck, pending Ryan's questions!

agent/otel.go
julianedwards: I think that is a good idea!

PR #6497, opened 2023-05-15
EVG-18723 Add admin setting to validate images for container tasks

malikchaya2:
EVG-18723

### Description
Add an admin setting that only allows specific container images to be used. 

### Testing
Tested on staging and added to existing tests. 
Smoke tests are passing [here](https://spruce.mongodb.com/version/6462a48d3627e07154f40618/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). 

### Documentation
I don't think we are ready for documentation around this.


config_cloud.go
```diff
@@ -186,6 +186,8 @@ type ECSConfig struct {
 	// that will be used. This is not a value that can or should be configured
 	// for production, but is useful to explicitly set for testing purposes.
 	ClientType AWSClientType `bson:"client_type" json:"client_type" yaml:"client_type"`
+	// AllowedImages specifies images that are allowed to be used
```
Kimchelly, line 189: nit: Period at the end of the doc comment.

model/pod/pod.go
```diff
@@ -80,6 +80,7 @@ func (o *TaskIntentPodOptions) Validate(ecsConf evergreen.ECSConfig) error {
 	if o.OS == OSWindows {
 		catcher.Wrap(o.WindowsVersion.Validate(), "must specify a valid Windows version")
 	}
+	catcher.NewWhen(!utility.StringSliceContains(ecsConf.AllowedImages, o.Image), "image not allowed")
```
Kimchelly, line 83: @hadjri Just a heads up that this will cause the current image to fail unless it's set in staging/production.

rest/data/pod_test.go
```diff
@@ -12,6 +13,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
```
Kimchelly, line 16: I would avoid doing this because it is assuming that we're initializing global state that will never change for testing. Unless there's a _really_ strong reason to do this, we should only set the environment global state in the individual test cases.

rest/route/pod_test.go
```diff
@@ -16,6 +17,12 @@ import (
 )
 
 func TestPostPod(t *testing.T) {
+	env := evergreen.GetEnvironment()
```
Kimchelly, line 20: Instead of setting global environment state, use the `testutil.NewEnvironment` on L84 and set the settings on that one.

units/pod_creation_test.go
```diff
@@ -19,6 +19,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func init() {
```
Kimchelly, line 22: Same thing as above.

validator/project_validator_test.go
```diff
@@ -3240,6 +3240,12 @@ func TestValidateVersionControl(t *testing.T) {
 }
 
 func TestValidateContainers(t *testing.T) {
+	env := evergreen.GetEnvironment()
```
Kimchelly, line 3243: Rather than setting the global state, I would change `validateContainers` to accept an environment and pass that through rather than set the global testing state.

units/pod_allocator_test.go
```diff
@@ -223,6 +227,10 @@ func TestPodAllocatorJob(t *testing.T) {
 			}()
 			env.EvergreenSettings.PodLifecycle.MaxParallelPodRequests = 1
 			require.NoError(t, env.EvergreenSettings.PodLifecycle.Set())
+			env.EvergreenSettings.Providers.AWS.Pod.ECS.AllowedImages = []string{
```
Kimchelly, line 230: It seems weird to set the allowed images in a case which is testing something completely unrelated. Can we delete/move this? It seems like it's already set at the top of the test.

service/templates/admin.html
```diff
@@ -1893,52 +1893,58 @@ <h2 class="modal-title">[[modalTitle]]</h2>
 										<textarea ng-model="Settings.providers.aws.allowed_instance_types"
```
Kimchelly, line 1893: GitHub doesn't let me comment on unchanged lines but L1891 needs to have `width:45%; margin-left:50px;` to preserve the consistency of adming settings styling.

Kimchelly (changes requested):

model/pod/pod.go
malikchaya2: I set it in staging for "hadjri/evg-container-self-testing", so unless there's another one actively being used, we should be okay.

rest/route/pod_test.go
malikchaya2: ph.Run does not have the ecs config when it's only set on testutil. I moved it down there but it has to be a a global env.

units/pod_allocator_test.go
Kimchelly: I still have this question. If all of them need this setting, I would just specify the allowed images in the setup (i.e. above L53) and remove this spot, which sets it for just one test.

rest/route/pod_test.go
```diff
@@ -81,7 +82,14 @@ func TestPostPod(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			p := makePostPod(testutil.NewEnvironment(ctx, t))
+			env := evergreen.GetEnvironment()
```
Kimchelly, line 85: You can use `testutil.GetEnvironment()` or `mock.Environment` here since you already pass it into `makePostPod`, which will then use it when running the route.

rest/route/pod_test.go
Kimchelly: It seems like it doesn't have to be set on the global `evergreen.GetEnvironment()`, just on the environment passed into the route handler (see my other comment).

units/pod_allocator_test.go
malikchaya2: I'm having some issues with this, I'll look into it soon

malikchaya2: moved it down

malikchaya2: (this should be good now, I had forgotten to push my last change)

rest/route/pod_test.go
malikchaya2: I would think so too, but for some reason, the env does not get passed all the way through to NewTaskIntentPod.

Kimchelly (approved):

PR #6499, opened 2023-05-16
EVG-19753: write GitHub PR comment when commit queue enqueue fails

Kimchelly:
EVG-19753

### Description
Write a GitHub PR comment containing the error message when a PR can't be enqueued in the commit queue.

### Testing
Tested in staging commit queue sandbox (https://github.com/evergreen-ci/commit-queue-sandbox/pull/484) and it displayed the error message. Can't really unit test since the GitHub integration code has no mock implementation.

### Documentation
N/A


rest/route/github.go
```diff
@@ -240,8 +240,22 @@ func (gh *githubHookApi) handleComment(ctx context.Context, event *github.IssueC
 		grip.Info(gh.getCommentLogWithMessage(event, "commit queue triggered"))
 
 		_, err := data.EnqueuePRToCommitQueue(ctx, evergreen.GetEnvironment(), gh.sc, createEnqueuePRInfo(event))
-		grip.Error(message.WrapError(err, gh.getCommentLogWithMessage(event, "can't enqueue on commit queue")))
-		return errors.Wrap(err, "enqueueing in commit queue")
+		if err != nil {
+			owner := event.Repo.Owner.GetLogin()
+			repo := event.Repo.GetName()
+			prNum := event.Issue.GetNumber()
+			grip.Error(message.WrapError(gh.displayCommitQueueEnqueueError(ctx, owner, repo, prNum, err), message.Fields{
```
ablack12, line 247: This is also going to post an error for checking if the project is tracking the commit queue to begin with https://github.com/evergreen-ci/evergreen/blob/91095164b387b5216b26082e0b830cdb38fc2b16/rest/data/commit_queue.go#L322-L321 which means we'll comment this on any 10gen project, which will be confusing, and it seems wrong to comment on disabled projects. Can we refactor to check this first or only comment if we return a bool?

ablack12 (changes requested):

rest/route/github.go
ablack12: 10gen projects aren't tracked in Evergreen; they use a CR bot to listen to the `evergreen merge` comment themselves, and then use the changes to add them to the commit queue for the mongodb/mongo project (which _is_ tracked) using the CLI. I wish very much that they had used a different comment but I think they wanted users to not have to consider what project they were in. This is what the #github-pr channel is for. Although this is a bit niche, this does effect a number of projects and would result in us posting a lot of unnecessary comments that are likely to cause confusion.

Kimchelly: Ah, I see... This is definitely not the reason I was expecting and I don't know if anyone knows this, so I'll try mentioning that the disabled commit queue with `evergreen merge` is a special case.

Kimchelly (commented):
I consolidated the logic so that it only comments if the PR fails to get into the commit queue (the logic after that can error, but it seems like it's not important as long as the PR patch is in the commit queue).

ablack12 (approved):

rest/data/commit_queue.go
```diff
@@ -474,17 +484,20 @@ func tryEnqueueItemForPR(ctx context.Context, sc Connector, projectRef *model.Pr
 	return patchDoc, nil
 }
 
-func sendGitHubCommitQueueError(env evergreen.Environment, pr *github.PullRequest, userRepo UserRepoInfo, err error) {
+// sendGitHubCommitQueueError updates the GitHub status and posts a comment
```
malikchaya2, line 487: Can we please add a period to line 438 after "branch protection settings are likely not met"? Since this shows up in a comment, I think it will look better as a full sentence.

Kimchelly: None of the other statuses that Evergreen sends back to GitHub are period-ending sentences, so I don't think we should do this.

PR #6500, opened 2023-05-16
EVG-13927 remove the spot provider

ybrill:
[EVG-13927](https://jira.mongodb.org/browse/EVG-13927)

### Description
I was looking to remove OnUp for something else and I realized that the only provider that was making any material use of it was spot. Since we had [EVG-13927](https://jira.mongodb.org/browse/EVG-13927) for removing spot it seemed like an opportune time.

To Brian's point in the ticket, none of the existing configs specify [spot](https://github.com/evergreen-ci/evergreen/blob/d8d411c4e38da221976dc996308fa6b7418fb767/apimodels/agent_models.go#L155) for `host.create`.

### Testing
If tests are still broken I'll fix them🙈 . I'll try spinning up an on-demand host on staging before merging.

PR #6501, opened 2023-05-16
EVG-19091 skip provisioned not running step

ybrill:
[EVG-19091](https://jira.mongodb.org/browse/EVG-19091)

### Description
It looks like the reason we have this intermediate `provisioned-not-running` state is [so we can do whatever setup on the host first](https://github.com/evergreen-ci/evergreen/pull/2723). Post https://github.com/evergreen-ci/evergreen/pull/6500 `OnUp` is no longer a thing so we should be good to skip it, I think?

In the case where a host comes online and there are no dispatchable tasks for it we should terminate it after the idle termination threshold is passed. An issue I observed is that the idle termination job only looks for hosts that are [running](https://github.com/evergreen-ci/evergreen/blob/6dace20f6ad21225fe63f04046f7d047d085ea48/model/host/db.go#L179), so a user data host that hasn't yet transitioned to running and that hasn't passed the `provisioningCutoff` isn't eligible for termination. Because nextTask doesn't set the host to `running` unless it's been already been marked provisioned by the set-cloud-hosts-ready job it won't get terminated by the idle termination job until set-cloud-hosts-ready has found the host running, which is often on the order of minutes after the host has requested its first task.

### Testing
I'll make sure a user data host still provisions as expected before I merge this.

### DIFF
--- a/mock/environment.go
+++ b/mock/environment.go
@@ -57,6 +57,9 @@ func (e *Environment) Configure(ctx context.Context) error {
 	e.EnvContext = ctx
 
 	e.EvergreenSettings = testutil.TestConfig()
+	if err := evergreen.UpdateConfig(e.EvergreenSettings); err != nil {
+		return errors.WithStack(err)
+	}
 
 	e.Remote = queue.NewLocalLimitedSize(2, 1048)
 	if err := e.Remote.Start(ctx); err != nil {
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -1073,38 +1073,6 @@ func (h *Host) SetProvisionedNotRunning() error {
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


Kimchelly (changes requested):
We can't skip this, because it's the only way to represent the intermediate state we know the host is running but. If we remove this, a bunch of things will happen like the cloud host ready job will keep checking a host we know is up already until next task is requested (and it's provisioned). It's also used for a bunch of timing logic to check when it's acceptable to terminate a host that we know is up but failed to do whatever it's supposed to do (whether that's becoming a running spawn host, starting the agent, etc.)

I would really advise not removing this, since it will cause a bunch of bugs.

ybrill:
Ah, okay.
I'll look for a different solution, then. Thanks!

PR #6502, opened 2023-05-16
EVG-19265: set GitHub PR status for all commit queue enqueue errors

Kimchelly:
EVG-19265

### Description
Evergreen currently only sends a GitHub status back when commit queue enqueueing fails for specific reasons (e.g. branch protections not met), but it seems like users would just want to know whatever is the enqueue error without asking us to sift through Splunk for it. This change ensures that for all the error cases (except specifically one where we can't get the PR's information), we return some kind of error message back to the GitHub status.

### Testing
Tested in staging with the commit queue sandbox (https://github.com/evergreen-ci/commit-queue-sandbox/pull/485) that it propagated the errors back to the GitHub status, and that merging still worked properly. Not really that unit testable since the GitHub API has no mocking.

### Documentation
N/A

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -260,6 +260,15 @@ type UserRepoInfo struct {
 	Repo     string
 }
 
+// NewUserRepoInfo creates UserRepoInfo from PR information.
+func NewUserRepoInfo(info commitqueue.EnqueuePRInfo) UserRepoInfo {
+	return UserRepoInfo{
+		Username: info.Username,
+		Owner:    info.Owner,
+		Repo:     info.Repo,
+	}
+}
+
 func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Context, settings *evergreen.Settings, args UserRepoInfo) (bool, error) {
 	// In the org
 	token, err := settings.GetGithubOauthToken()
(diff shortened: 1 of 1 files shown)


rest/data/commit_queue.go
```diff
@@ -298,60 +307,67 @@ func (pc *DBCommitQueueConnector) IsAuthorizedToPatchAndMerge(ctx context.Contex
 
 // EnqueuePRToCommitQueue enqueues an item to the commit queue to test and merge a PR.
 func EnqueuePRToCommitQueue(ctx context.Context, env evergreen.Environment, sc Connector, info commitqueue.EnqueuePRInfo) (*restModel.APIPatch, error) {
-	settings := env.Settings()
-	userRepo := UserRepoInfo{
-		Username: info.Username,
-		Owner:    info.Owner,
-		Repo:     info.Repo,
+	patchDoc, pr, err := getAndEnqueueCommitQueueItemForPR(ctx, env, sc, info)
+	if err != nil {
```
hadjri, line 311: `sendGitHubCommitQueueError` already does a nil check on the err. You could remove this and rename the function to something like `trySendGitHubCommitQueueError`

Kimchelly: Yeah it does check it, but I like writing sanity checks in places where I think a helper function might be used in the future in a way that's different from how I use it right now. It's nice to check inputs when you're not sure how someone later on is going to use it, especially since it would weird to send a failure status back to GitHub with no actual error message attached.

PR #6504, opened 2023-05-16
EVG-19884 No-op stranded container task cleanup if task is already deallocated

hadjri:
EVG-19884

### Description
Some pods were found to be in a decommissioned state with a task still specified in their `TaskRuntimeInfo`, but the task was marked as deallocated, preventing the termination job from succeeding, saying `cannot deallocate a container task if it's not currently allocated` ([logs](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen-staging%22%20%22cannot%20deallocate%20a%20container%20task%20if%20it%27s%20not%20currently%20allocated%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-30d%40d&latest=now&sid=1684167348.432188)).

While the root cause that allowed this bad state to happen was likely fixed by this [PR](https://github.com/evergreen-ci/evergreen/pull/6426) (these were old pods erroring), we should allow the system to heal itself if it enters this bad state by going through with the pod termination even if the pod's assigned task has been deallocated.
### Testing
Updated unit test with scenario.

### DIFF
--- a/units/pod_termination.go
+++ b/units/pod_termination.go
@@ -232,9 +232,16 @@ func (j *podTerminationJob) fixStrandedRunningTask(ctx context.Context) error {
 		})
 		return nil
 	}
-
-	if err := t.MarkAsContainerDeallocated(ctx, j.env); err != nil {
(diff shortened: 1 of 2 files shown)


Kimchelly (approved):

Kimchelly (approved):
Nice catch!

PR #6474, opened 2023-05-04
EVG-19174 Remove deprecated commands

malikchaya2:
EVG-19174

### Description
This removes commands that are marked deprecated. 
Todo: remove from shrub. 

### Documentation
Updated relevant docs 

### Shrub PR
[34](https://github.com/evergreen-ci/shrub/pull/34)

### Testing 
The tests that fail here pass on [this patch ](https://spruce.mongodb.com/version/645533637742aecaa11bc75c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

PR #6507, opened 2023-05-17
EVG-19174 Remove deprecated commands

malikchaya2:
EVG-19174

### Description
This includes all the changes from [6474](https://github.com/evergreen-ci/evergreen/pull/6474) except for removing manifest.load and git.apply_patch. These two still have a lot of usage and don't seem worth the effort of removing. 

manifest.load is currently being used in 30 project yamls
git.apply_patch is being used in 62
and shell.cleanup and shell.track in 2 
gotest.parse_json and expansions.fetch are not being used. 

This is waiting on one user to remove their shell.track and shell.cleanup. This won't be merged until then. 

### Testing
Existing tests

### Documentation
Included


docs/Configure-a-Project/Project-Commands.md
```diff
@@ -745,14 +740,6 @@ Parameters:
     missing/empty will run for all
 -   `working_directory`: local path to the working directory
 
-## manifest.load
```
hadjri, line 748: We should keep this right?

scripts/agent.yml
```diff
@@ -186,7 +186,6 @@ tasks:
         params:
           file: "src/scripts/downstream_expansions.yml"
           destination: "test_num"
-      - command: manifest.load
```
hadjri, line 189: Should we put this back now?

docs/Configure-a-Project/Project-Commands.md
malikchaya2: I don't think so, because all manifest.load does is let the user know that it's deprecated, so there's nothing to test. 
We are only keeping it in the registry in order for it to not result in errors for everyone who hasn't removed it yet.

scripts/agent.yml
malikchaya2: please see my comment above

docs/Configure-a-Project/Project-Commands.md
malikchaya2: (this probably should have been removed when the command was first deprecated)

ablack12 (commented):
Can you include in the description the information you found here from validating project yamls?

hadjri (approved):

PR #6508, opened 2023-05-17
EVG-19749 Add auth method for github apps

bynn:
EVG-19749

### Description
adds function to connect to GitHub client with GitHub app auth 
what it does: 
first, we check if we have the settings set to even use this method (app id and private key) 
then we use those things to connect to a jwt client. this client can only call certain functions, one being requesting the installation id of a repo. (this does not use up any rate limit) 
then using the installation id, we create a new client where we can actually call functions 

### Testing
tested that the client functions by calling it on a test and making sure it returns values i expected it to 

	token, err := settings.CreateInstallationToken(context.Background(), "evergreen-ci", "commit-queue-sandbox")

	s.NoError(err)

	a, _, err := GetGithubCommits(context.Background(), token, "evergreen-ci", "commit-queue-sandbox", "", time.Time{}, 0)

	s.NoError(err)

1 unit test,
will be testing on staging after integrating on the next ticket

### DIFF
--- a/config.go
+++ b/config.go
@@ -581,16 +581,19 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 		return "", errors.New("Github app settings are not set")
 	}
 	httpClient := utility.GetHTTPClient()
+	defer utility.PutHTTPClient(httpClient)
 	itr, err := ghinstallation.NewAppsTransport(httpClient.Transport, authFields.AppId, authFields.privateKey)
 	if err != nil {
 		return "", errors.Wrap(err, "creating transport with JWT")
 	}
 	httpClient.Transport = itr
-	defer utility.PutHTTPClient(httpClient)
 	client := github.NewClient(httpClient)
 	installationId, _, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
-	if err != nil || installationId == nil {
+	if err != nil {
 		return "", errors.Wrapf(err, "finding installation token for '%s/%s'", owner, repo)
+	}
+	if installationId == nil {
+		return "", errors.New(fmt.Sprintf("Installation id for '%s/%s' not found", owner, repo))
 	}
 	token, _, err := client.Apps.CreateInstallationToken(ctx, installationId.GetID(), opts)
 	if err != nil || token == nil {
--- a/config.go
+++ b/config.go
@@ -10,8 +10,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/bradleyfalzon/ghinstallation"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/utility"
+	"github.com/google/go-github/v52/github"
 	"github.com/mongodb/amboy/logger"
 	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
@@ -547,6 +549,54 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
 	}
 
 	return send.NewConfiguredMultiSender(senders...), nil
+}
+
+type githubAppAuth struct {
+	AppId      int64
+	privateKey []byte
+}
+
+// getGithubAppAuth returns app id and app private key if it exists.
+func (s *Settings) getGithubAppAuth() *githubAppAuth {
+	if s.AuthConfig.Github == nil || s.AuthConfig.Github.AppId == 0 {
+		return nil
+	}
+
+	key := s.Expansions[githubAppPrivateKey]
+	if key == "" {
+		return nil
+	}
+
+	return &githubAppAuth{
+		AppId:      s.AuthConfig.Github.AppId,
+		privateKey: []byte(key),
+	}
+}
+
+// CreateInstallationToken uses the owner/repo information to request an github app installation id
+// and uses that id to create an installation token.
+func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo string, opts *github.InstallationTokenOptions) (string, error) {
+	authFields := s.getGithubAppAuth()
+	if authFields == nil {
+		return "", errors.New("Github app settings are not set")
+	}
+	httpClient := utility.GetHTTPClient()
+	itr, err := ghinstallation.NewAppsTransport(httpClient.Transport, authFields.AppId, authFields.privateKey)
+	if err != nil {
+		return "", errors.Wrap(err, "creating transport with JWT")
+	}
+	httpClient.Transport = itr
+	defer utility.PutHTTPClient(httpClient)
+	client := github.NewClient(httpClient)
+	installationId, _, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
+	if err != nil || installationId == nil {
+		return "", errors.Wrapf(err, "finding installation token for '%s/%s'", owner, repo)
+	}
+	token, _, err := client.Apps.CreateInstallationToken(ctx, installationId.GetID(), opts)
+	if err != nil || token == nil {
+		return "", errors.Wrapf(err, "creating installation token for installation id: %d", installationId.GetID())
+	}
+	return token.GetToken(), nil
 }
 
 func (s *Settings) makeSplunkSender(ctx context.Context, client *http.Client, levelInfo send.LevelInfo, fallback send.Sender) (send.Sender, error) {
--- a/globals.go
+++ b/globals.go
@@ -281,6 +281,7 @@ const (
 	DefaultJasperPort = 2385
 
 	GlobalGitHubTokenExpansion = "global_github_oauth_token"
+	githubAppPrivateKey        = "github_app_private_key"
 
 	VSCodePort = 2021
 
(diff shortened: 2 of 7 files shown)


config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+			"message": "failed to create github jwt app transport",
+			"app_id":  appId,
+		}))
+		return nil
+	}
+	jwtClient := utility.GetHTTPClient()
+	jwtClient.Transport = itr
+	defer utility.PutHTTPClient(jwtClient)
+	client := github.NewClient(jwtClient)
+	installationId, _, err := client.Apps.FindRepositoryInstallation(context.Background(), owner, repo)
```
Kimchelly, line 581: Pass a context into the function so we can avoid using `context.Background()`.

globals.go
```diff
@@ -278,6 +278,7 @@ const (
 	DefaultJasperPort = 2385
 
 	GlobalGitHubTokenExpansion = "global_github_oauth_token"
+	GithubAppPrivateKey        = "github_app_private_key"
```
Kimchelly, line 281: Does this need to be exported to other packages if we're only using it in the `evergreen` package?

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
 	return send.NewConfiguredMultiSender(senders...), nil
 }
 
+// getGithubAppAuth returns app id and app private key if it exists.
+// returns false if it does not have both of those fields.
+func (s *Settings) getGithubAppAuth() (int64, []byte, bool) {
+	key, ok := s.Expansions[GithubAppPrivateKey]
+	if s.AuthConfig.Github != nil && s.AuthConfig.Github.AppId != 0 && ok && key != "" {
```
Kimchelly, line 558: Can we simplify this if statement down a little bit? Something like:
```go
if s.AuthConfig.Github == nil || s.AuthConfig.Github.AppId == 0 { return no result }
key := s.Expansions[GithubAppPrivateKey]
if key == "" { also return no result }
return successful info
```
It also reads more smoothly when the last return is the success path (i.e. the key is returned), and all the returns before it are the erroneous conditions.

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	}
+	return 0, nil, false
+}
+
+func (s *Settings) GetGithubAppClient(caller, owner, repo string, retry utility.HTTPRetryFunction, delay utility.HTTPDelayFunction) *http.Client {
+	appId, key, exists := s.getGithubAppAuth()
+	if !exists {
+		return nil
+	}
+	itr, err := ghinstallation.NewAppsTransport(http.DefaultTransport, appId, key)
```
Kimchelly, line 569: Rather than using the HTTP package's default transporter, can we use the utility client's built-in transporter?
```go
c := utility.GetHTTPClient()
itr, err := ghinstallation.NewAppsTransport(c.Transport, appId, key)
c.Transport = itr
stuff...
```

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	if err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
+			"message": "failed to create github jwt app transport",
+			"app_id":  appId,
+		}))
+		return nil
```
Kimchelly, line 575: Should this really squash the error rather than return it? The preference is to return errors instead of `grip.Error`ing in such low-level infrastructure code, unless there's good reason for it (or it's too diffcult to change existing code). This log statement doesn't seem very traceable, and I don't think anyone would know to look for it.

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+			"repo":    repo,
+			"caller":  caller,
+		}))
+		return nil
+	}
+	return getGithubClient(caller, appId, installationId.GetID(), key, retry, delay)
+}
+
+func getGithubClient(caller string, appId, installationId int64, key []byte, retry utility.HTTPRetryFunction, delay utility.HTTPDelayFunction) *http.Client {
+	client := utility.GetHTTPClient()
```
Kimchelly, line 595: Instead of making a brand new HTTP client, can we just pass in the one that was created in `GetGithubAppClient`? Or is that not allowed?

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	itr, err := ghinstallation.New(client.Transport, appId, installationId, key)
+	if err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
```
Kimchelly, line 598: Same thing as above, don't squash errors in low-level code without good reason.

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+			"message": "error creating github client",
+			"caller":  caller,
+		}))
+		return nil
+	}
+	if retry != nil || delay != nil {
+		return utility.GetCustomHTTPRetryableClientWithTransport(itr, retry, delay)
```
Kimchelly, line 605: I believe wrapping a `utility.GetHTTPClient` (above) in another `utility.GetHTTPClient` here will produce a slow HTTP client leak, because even if we `utility.PutHTTPClient` the wrapper, it won't unwrap the underlying client that was originally used. We should only use 1 HTTP client in this code to avoid this.

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+// returns false if it does not have both of those fields.
+func (s *Settings) getGithubAppAuth() (int64, []byte, bool) {
+	key, ok := s.Expansions[GithubAppPrivateKey]
+	if s.AuthConfig.Github != nil && s.AuthConfig.Github.AppId != 0 && ok && key != "" {
+		return s.AuthConfig.Github.AppId, []byte(key), true
+	}
+	return 0, nil, false
+}
+
+func (s *Settings) GetGithubAppClient(caller, owner, repo string, retry utility.HTTPRetryFunction, delay utility.HTTPDelayFunction) *http.Client {
```
Kimchelly, line 564: Doc comment for exported functions.

config.go
```diff
@@ -549,6 +551,63 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
 	return send.NewConfiguredMultiSender(senders...), nil
 }
 
+// getGithubAppAuth returns app id and app private key if it exists.
+// returns false if it does not have both of those fields.
+func (s *Settings) getGithubAppAuth() (int64, []byte, bool) {
```
Kimchelly, line 556: It seems redundant to return a bool when you could just check for a non-zero result.

Kimchelly, line 556: I would also slightly recommend returning a struct with the two results in the return so that it's slightly more self-documenting than random unnamed int and byte slice.

Kimchelly (changes requested):

config.go
bynn: updated to use 1 client

globals.go
bynn: privatized

config.go
```diff
@@ -549,6 +551,61 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	if authFields == nil {
+		return nil, errors.New("Github app settings are not set")
+	}
+	httpClient := utility.GetHTTPClient()
+	itr, err := ghinstallation.NewAppsTransport(httpClient.Transport, authFields.AppId, authFields.privateKey)
+	if err != nil {
+		return nil, err
+	}
+	httpClient.Transport = itr
+	defer utility.PutHTTPClient(httpClient)
```
bynn, line 588: i had a question for you @Kimchelly 
when I'm using just 1 client like I'm doing here, I thought I should delete this defer because we're going to be returning the client

but when I deleted that line, I started getting auth errors and nothing worked. It works in the current state that it's in but I'm not super sure why

config.go
```diff
@@ -549,6 +551,54 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	if authFields == nil {
+		return "", errors.New("Github app settings are not set")
+	}
+	httpClient := utility.GetHTTPClient()
+	itr, err := ghinstallation.NewAppsTransport(httpClient.Transport, authFields.AppId, authFields.privateKey)
+	if err != nil {
+		return "", errors.Wrap(err, "creating transport with JWT")
+	}
+	httpClient.Transport = itr
+	defer utility.PutHTTPClient(httpClient)
```
Kimchelly, line 589: Move `defer utility.PutHTTPClient()` right below `utility.GetHTTPClient()` to ensure that it gets returned to the pool.

config.go
```diff
@@ -549,6 +551,54 @@ func (s *Settings) GetSender(ctx context.Context, env Environment) (send.Sender,
...
+	client := github.NewClient(httpClient)
+	installationId, _, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
+	if err != nil || installationId == nil {
+		return "", errors.Wrapf(err, "finding installation token for '%s/%s'", owner, repo)
```
Kimchelly, line 593: This needs to make a new error in the case where `err` is non-nil but installationId is nil (or make a separate if block).

Kimchelly (approved):
LGTM mod small things.

PR #6509, opened 2023-05-17
EVG-19912 Fix test results not fetching for past executions

khelif96:
EVG-19912

### Description
The `GetTestResults` function relied on several fields from the task model. Since we used `ToService` to generate the db task model it may not have had all of the necessary field mappings for certain tasks with older executions. Updates it to query for the task so we always have an up-to-date task. 
### Testing
Tested on a patch in staging and confirmed I can view test results regardless of execution. Also wrote unit tests.


minnakt (approved):
lgtm

PR #6510, opened 2023-05-17
EVG-19875: Modify ProjectRef struct to include Parsley filters

minnakt:
EVG-19875

### Description
This PR is just to add Parsley filters to the `ProjectRef` and `APIProjectRef` structs. Future tickets will build on this work (including handling validation and saving Parsley filters via GraphQL mutation and PATCH rest route).

#### Note
I used `expression` rather than `name` for the field because I felt like it might be more correct... The filter doesn't really have a name, it's defined by some regex expression. This change can also be made on the frontend.

### Testing
- Added some local data for Parsley filters in the `sys-perf` project and checked that it was correctly loaded into the DB and viewable via http://localhost:9090/rest/v2/projects/sys-perf


sophstad (approved):
Lgtm but would probably make sense to have an app reviewer confirm! =]

model/project_ref.go
```diff
@@ -114,6 +114,9 @@ type ProjectRef struct {
 	ContainerSizeDefinitions []ContainerResources `bson:"container_size_definitions,omitempty" json:"container_size_definitions,omitempty" yaml:"container_size_definitions,omitempty"`
 	ContainerSecrets         []ContainerSecret    `bson:"container_secrets,omitempty" json:"container_secrets,omitempty" yaml:"container_secrets,omitempty"`
 
+	// Filters settings
+	ParsleyFilters []ParsleyFilter `bson:"parsley_filters,omitempty" json:"parsley_filters,omitempty"`
```
ablack12, line 118: Preference to move this to the bottom since those are also about UI views

ablack12 (approved):
LGTM with nit

model/project_ref.go
sophstad: I'll wait until this is merged in and follow suit for #6516 👍

PR #6511, opened 2023-05-17
EVG-17508 remove backwards compatibility for latest parent execution

ablack12:
EVG-17508

### Description
This backwards compatibility is almost a year old and only affects restarting display tasks, so we should be good to remove it now. 

### Testing
Updated test to not test for backwards compatibility.


bynn (approved):
wow it's already been a year?? 👴

PR #6512, opened 2023-05-17
EVG-19091 include starting hosts in idle hosts

ybrill:
[EVG-19091](https://jira.mongodb.org/browse/EVG-19091)

### Description
The same motivation as #6501. The solution proposed here is to expand the `IdleEphemeralGroupedByDistroID` query to include user data hosts that have contacted the app to ask for a task. This is also idle in the sense that these hosts are cleared to run tasks from this time. Also adjust `h.IdleTime()` accordingly.

Another adjustment to `h.IdleTime()` is to not count the time before the host is provisioned as idle time for the purposes of idle termination. That time still counts as `WastedComputeTime` so that logic was left as its own function. This is necessary since if we count the time since the host came online the host will have that much less time before it's considered idle. And especially now that user data hosts won't have the preceding few minutes to grab a task before they're considered for idle termination.

### Testing
I'll make sure an idle user data host gets killed in staging at around the idle time from when it first called `nextTask`.

### DIFF
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -181,7 +181,6 @@ func IdleEphemeralGroupedByDistroID() ([]IdleHostsByDistroID, error) {
 					{
 						StatusKey:    evergreen.HostStarting,
 						bootstrapKey: distro.BootstrapMethodUserData,
-
 						// Idle time starts from when a host is first able
 						// to run a task. For user data hosts this is the time
 						// when they first made contact with the app instead
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -181,10 +181,17 @@ func IdleEphemeralGroupedByDistroID() ([]IdleHostsByDistroID, error) {
 					{
 						StatusKey:    evergreen.HostStarting,
 						bootstrapKey: distro.BootstrapMethodUserData,
-						// User data hosts have a grace period between creation
-						// and provisioning during which they are not considered
-						// for idle termination to give agents time to start.
-						CreateTimeKey: bson.M{"$lte": time.Now().Add(-provisioningCutoff)},
+						"$or": []bson.M{
+							// Idle time starts from when a host is first able
+							// to run a task. For user data hosts this is the time
+							// when they first made contact with the app instead
+							// of when they're marked running.
+							{AgentStartTimeKey: bson.M{"$gt": utility.ZeroTime}},
+							// User data hosts have a grace period between creation
+							// and provisioning during which they are not considered
+							// for idle termination to give agents time to start.
+							{CreateTimeKey: bson.M{"$lte": time.Now().Add(-provisioningCutoff)}},
+						},
 					},
 				},
 			},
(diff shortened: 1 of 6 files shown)


model/host/db.go
```diff
@@ -181,10 +181,12 @@ func IdleEphemeralGroupedByDistroID() ([]IdleHostsByDistroID, error) {
 					{
 						StatusKey:    evergreen.HostStarting,
 						bootstrapKey: distro.BootstrapMethodUserData,
-						// User data hosts have a grace period between creation
-						// and provisioning during which they are not considered
-						// for idle termination to give agents time to start.
-						CreateTimeKey: bson.M{"$lte": time.Now().Add(-provisioningCutoff)},
+
```
Kimchelly, line 184: nit: remove random whitespace line.

ybrill:
Tested this in staging: I commented out task dispatching and the set-cloud-hosts-ready job so when a host came up it would be idle and never marked ready. [This](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20host_id%3D%22i-02f845c3044af65e4%22%0A%7C%20sort%20-metadata.time&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1684418400&latest=1684419000&sid=1684419545.1050384) is an example of a host that came up that got marked for termination by the idle hosts job ~30s after its AgentStartTime.

model/host/host_test.go
```diff
@@ -2035,9 +2036,10 @@ func TestIdleEphemeralGroupedByDistroID(t *testing.T) {
 		Provider:              evergreen.ProviderNameMock,
 		Status:                evergreen.HostStarting,
 		CreationTime:          time.Now().Add(-100 * time.Minute),
+		AgentStartTime:        time.Now(),
 	}
-	// User data host that is not running task but has not passed the grace
-	// period to start running tasks is not idle.
+	// User data host that is not running task but has not had its
+	//  AgentStartTime set is not idle.
```
Kimchelly, line 2042: nit: remove extra whitespace.

model/host/host.go
```diff
@@ -386,21 +386,39 @@ func (h *Host) GetTaskGroupString() string {
...
-	// If the host is not running a task then the idle time is the time
-	// elapsed since the last task finished.
-	return h.SinceLastTaskCompletion()
+	// If the host has run a task it's been idle since that task finished running.
+	if h.LastTask != "" {
+		return time.Since(h.LastTaskCompletedTime)
+	}
+
+	// If the host never ran a task it's been idle since the time it was first ready
+	// to run tasks.
```
Kimchelly, line 402: Can we add a comment explaining why idle time is computed differently between the start task agent route (for `IncIdleTime`) and here? Because here, we're intentionally giving some extra time to hosts to come up and ignoring all the idle compute time that it spends waiting for provisioning steps to run. It might not be obvious to a reader that this is an intentional distinction between the calculations (since agent start time always comes after host start time).

Kimchelly (approved):
LGTM mod nits and suggestion.

model/host/db.go
ybrill: done

model/host/host.go
ybrill: done

PR #6517, opened 2023-05-18
EVG-19920 Simplify docs organization

bsamek:


bsamek:
This reduces the number of directories, simplifies the titles, and throws away the requirement that titles start with a verb.

bsamek:
Before I merge, I will fix all the broken links. I didn't want to do that until I get PR approval though, since moving things around again can re-break the links.

Kimchelly (commented):
* Do we have to number the markdown files so that they appear in prioritized order on the sidebar? Otherwise I assume it'll be alphabetical and the CLI will appear at the top, while the REST API will appear at the bottom.
* Can we just absorb the degraded mode docs into the internal Evergreen ops docs? I doubt anyone other than Evergreen admins will ever need that info.
* I don't know where the task trace docs should live, but it seems like they wouldn't be reference material. Maybe under Project Configuration since it does involve you hooking up your project to Honeycomb?

bsamek:
> * Do we have to number the markdown files so that they appear in prioritized order on the sidebar? Otherwise I assume it'll be alphabetical and the CLI will appear at the top, while the REST API will appear at the bottom.

It's possible to number them, but I found them pretty painful to maintain numbered, since you have to keep the numbers consistent across all links. I think it's better without numbering.

> * Can we just absorb the degraded mode docs into the internal Evergreen ops docs? I doubt anyone other than Evergreen admins will ever need that info.

Sure. @ablack12 Does that sound okay to you?

> * I don't know where the task trace docs should live, but it seems like they wouldn't be reference material. Maybe under Project Configuration since it does involve you hooking up your project to Honeycomb?

Done.

ablack12 (approved):
Excited for the simplification 🎉 Fine to absorb team specific docs into Ops or Dev

PR #6519, opened 2023-05-18
EVG-17583: Add DeleteProject mutation

minnakt:
EVG-17583

### Description
This PR adds the `DeleteProject` mutation. It includes:
- Refactor around the project rest route code because I wanted to reuse the code from the [`DELETE`](https://github.com/evergreen-ci/evergreen/blob/main/rest/route/project.go#L741) rest route.
- Refactor the `@canCreateProject` directive because I wanted to reuse it for the delete operation (restricted to admins only).
- Filter out hidden projects from `groupProjects` util function. Hidden projects were showing up as disabled projects with empty identifiers in the dropdown.

### Testing
- Unit tests and GraphQL tests


graphql/resolver.go
```diff
@@ -38,14 +38,28 @@ func New(apiURL string) Config {
...
 		args, isStringMap := obj.(map[string]interface{})
 		if !isStringMap {
 			return nil, ResourceNotFound.Send(ctx, "Project not specified")
 		}
-		projectIdToCopy, hasProjectId := args["project"].(map[string]interface{})["projectIdToCopy"].(string)
-		if hasProjectId {
-			// Check if the user has permission to copy the project
+		operationContext := graphql.GetOperationContext(ctx).OperationName
+
+		if operationContext == "CreateProject" {
```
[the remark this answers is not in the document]
minnakt: Yes, it's directly related. What the directive was doing before was checking if the function arg `projectIdToCopy` existed, which only belongs to the `copyProject` mutation; so if it didn't exist it would mean the `createProject` mutation was being called

rest/data/project.go
```diff
@@ -389,3 +389,54 @@ func (pc *DBProjectConnector) GetProjectFromFile(ctx context.Context, pRef model
 	}
 	return model.GetProjectFromFile(ctx, opts)
 }
+
+func DeleteBranch(ctx context.Context, pRef *model.ProjectRef) error {
```
ablack12, line 393: need to put a function comment on here -- I'd also prefer we call this "HideBranch" since we aren't actually deleting it

rest/data/project.go
```diff
@@ -389,3 +389,54 @@ func (pc *DBProjectConnector) GetProjectFromFile(ctx context.Context, pRef model
...
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Message:    errors.Errorf("project '%s' is already hidden", pRef.Id).Error(),
+		}
+	}
+
+	if !pRef.UseRepoSettings() {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Message:    errors.Errorf("project '%s' must be attached to a repo to be eligible for deletion", pRef.Id).Error(),
```
ablack12, line 404: I have no idea why this is true. Can we make a follow up ticket for app to investigate this?

rest/data/project.go
```diff
@@ -389,3 +389,54 @@ func (pc *DBProjectConnector) GetProjectFromFile(ctx context.Context, pRef model
 	}
 	return model.GetProjectFromFile(ctx, opts)
 }
+
+func DeleteBranch(ctx context.Context, pRef *model.ProjectRef) error {
+	if pRef.IsHidden() {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusBadRequest,
```
ablack12, line 396: Should we return the status code so that the graphql resolver can use this rather than wrapping it with internal server error?

rest/data/project.go
ablack12: Also are we using context anywhere?

graphql/mutation_resolver.go
```diff
@@ -517,6 +526,19 @@ func (r *mutationResolver) DefaultSectionToRepo(ctx context.Context, projectID s
 	return &projectID, nil
 }
 
+// DeleteProject is the resolver for the deleteProject field.
+func (r *mutationResolver) DeleteProject(ctx context.Context, projectID string) (bool, error) {
+	pRef, err := model.FindBranchProjectRef(projectID)
```
ablack12, line 531: Should we move this in, since it's called in both places?

graphql/mutation_resolver.go
```diff
@@ -517,6 +526,19 @@ func (r *mutationResolver) DefaultSectionToRepo(ctx context.Context, projectID s
...
+	if err != nil || pRef == nil {
+		return false, ResourceNotFound.Send(ctx, fmt.Sprintf("finding project '%s'", projectID))
+	}
+	err = data.DeleteBranch(ctx, pRef)
```
ablack12, line 535: nit: preference to combine 535 and 536

graphql/query_resolver.go
```diff
@@ -488,7 +488,14 @@ func (r *queryResolver) ViewableProjectRefs(ctx context.Context) ([]*GroupedProj
 		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting projects: %v", err.Error()))
 	}
 
-	groupedProjects, err := groupProjects(projects, true)
+	visibleProjects := []model.ProjectRef{}
```
ablack12, line 491: Should we move this into groupProjects? I think the other place it's called is also relevant; can add a comment to it or rename groupVisibleProjects

graphql/resolver.go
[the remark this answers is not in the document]
ablack12: can we make these constants? Maybe top of file?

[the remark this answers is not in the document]
ablack12: (alternatively, should we have more directives than just one? Could be a different ticket though)

ablack12 (changes requested):

rest/data/project.go
minnakt: Opened [EVG-19958](https://jira.mongodb.org/browse/EVG-19958)!

rest/data/project.go
minnakt: I ended up modifying the GraphQL resolver so it can capture the correct error — let me know if you want me to try something else though!

rest/route/project.go
```diff
@@ -758,58 +758,10 @@ func (h *projectDeleteHandler) Parse(ctx context.Context, r *http.Request) error
 }
 
 func (h *projectDeleteHandler) Run(ctx context.Context) gimlet.Responder {
-	project, err := dbModel.FindBranchProjectRef(h.projectName)
+	err := data.HideBranch(h.projectName)
```
ablack12, line 761: nit: can combine with 762 since we're only returning the error now

graphql/util.go
```diff
@@ -748,6 +748,10 @@ func groupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
 	groupsMap := make(map[string][]*restModel.APIProjectRef)
 
 	for _, p := range projects {
+		if p.IsHidden() {
+			continue
+		}
+
```
ablack12, line 754: I'd suggest adding to the function comment that we ignore hidden projects, since we don't mention it in the function title

ablack12 (approved):
LGTM with nits

PR #6520, opened 2023-05-18
EVG-19587: Omit logging "context canceled" errors from GQL

SupaJoon:
[EVG-19587](jira.mongodb.com/browes/EVG-19587)

### Description
These code changes decrease the logging level for "context canceled" errors coming from GQL. Sometimes a single query will produce thousands of "context canceled" errors when a user closes their browser before a response is received. gqlgen middleware doesn't allow a way to reduce multiple errors into 1. The "context canceled" error from GQL seems to be expected behavior so far and isn't actionable but it could still be useful to continue logging in case it help in a future diagnostic. 

We originally thought that the main reason for the uptick in these errors was because we began threading the ctx to the db resolver in this [PR](https://github.com/evergreen-ci/evergreen/pull/6360) merged on March 29 but it's probably not true because the [mainline commits query context canceled error](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen%22%20path%3D%22%2Fgraphql%2Fquery%22%20AND%20error%3D%22*%3A%20context%20canceled%22%20AND%20query%3D%22mainlineCommits.versions%5B*%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1677646800&latest=1684382400&display.events.timelineEarliestTime=1679544000&display.events.timelineLatestTime=1679630400&sid=1684439091.1101263) is the most common single error and it began on March 23rd. 
### Testing

### Documentation

### DIFF
--- a/graphql/http_handler.go
+++ b/graphql/http_handler.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"net/http"
 	"runtime/debug"
-	"strings"
 
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/graphql/handler"
@@ -16,8 +15,6 @@ import (
 	"github.com/ravilushqa/otelgqlgen"
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
-
-const ContextCanceledErrorMessage = "context canceled"
 
 // Handler returns a gimlet http handler func used as the gql route handler
 func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
@@ -60,7 +57,7 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 			queryPath = fieldCtx.Path().String()
 			args = fieldCtx.Args
 		}
-		if err != nil && !strings.HasSuffix(err.Error(), ContextCanceledErrorMessage) {
+		if !errors.Is(err, context.Canceled) {
 			grip.Error(message.WrapError(err, message.Fields{
 				"path":    "/graphql/query",
 				"query":   queryPath,


graphql/http_handler.go
```diff
@@ -57,12 +58,17 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
...
-		grip.Error(message.WrapError(err, message.Fields{
+		errorMessage := message.WrapError(err, message.Fields{
 			"path":    "/graphql/query",
 			"query":   queryPath,
 			"args":    args,
 			"request": gimlet.GetRequestID(ctx),
-		}))
+		})
+		if strings.HasSuffix(err.Error(), ": context canceled") {
+			grip.Warning(errorMessage)
```
ablack12, line 68: If we're really only logging for debugging purposes, grip.Debug seems more correct.

ablack12 (changes requested):
I thought we'd originally discussed not logging this, since it does still log 1000s of logs we don't intend to look at. Do we have actual planned follow up work to do something with these logs, with a ticket we can include in the message? Otherwise it doesn't make sense to log these until we do. 

We could also possibly do something like https://github.com/evergreen-ci/evergreen/blob/91095164b387b5216b26082e0b830cdb38fc2b16/rest/data/repotracker.go#L45 to avoid logging in a noisy way.

ablack12 (changes requested):
I thought we'd originally discussed not logging this, since it does still log 1000s of logs we don't intend to look at. Do we have actual planned follow up work to do something with these logs, with a ticket we can include in the message? Otherwise it doesn't make sense to log these until we do. 

We could also possibly do something like https://github.com/evergreen-ci/evergreen/blob/91095164b387b5216b26082e0b830cdb38fc2b16/rest/data/repotracker.go#L45 to avoid logging in a noisy way.

graphql/http_handler.go
```diff
@@ -57,12 +58,14 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 			queryPath = fieldCtx.Path().String()
 			args = fieldCtx.Args
 		}
-		grip.Error(message.WrapError(err, message.Fields{
-			"path":    "/graphql/query",
-			"query":   queryPath,
-			"args":    args,
-			"request": gimlet.GetRequestID(ctx),
-		}))
+		if !strings.HasSuffix(err.Error(), ": context canceled") {
```
ablack12, line 61: err.Error() will panic if err is nil so we need to check that explicitly (message.WrapError noops if err is nil which is why we didn't need to do taht before)

Also preference to make this a string a const somewhere (and just checking for "context canceled" would be cleaner but idk if there's something you're looking to avoid)

ablack12 (changes requested):

graphql/http_handler.go
khelif96: You can use errors.is instead of the string comparison.

khelif96: I think you should be able to use `errors.is` instead of doing a string comparison. 

https://pkg.go.dev/errors

graphql/http_handler.go
```diff
@@ -57,12 +60,14 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 			queryPath = fieldCtx.Path().String()
 			args = fieldCtx.Args
 		}
-		grip.Error(message.WrapError(err, message.Fields{
-			"path":    "/graphql/query",
-			"query":   queryPath,
-			"args":    args,
-			"request": gimlet.GetRequestID(ctx),
-		}))
+		if err != nil && !strings.HasSuffix(err.Error(), ContextCanceledErrorMessage) {
```
khelif96, line 63: Sorry for the confusion I meant something like this.
```suggestion
		if !errors.Is(err, context.Canceled) {
```

graphql/http_handler.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
+const ContextCanceledErrorMessage = "context canceled"
```
khelif96, line 20: This is not needed see below.

graphql/http_handler.go
khelif96: `errors.Is` just looks for the error in the error tree so if it appears anywhere this condition should be met.

graphql/http_handler.go
```diff
@@ -57,12 +57,14 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
...
 			args = fieldCtx.Args
 		}
-		grip.Error(message.WrapError(err, message.Fields{
-			"path":    "/graphql/query",
-			"query":   queryPath,
-			"args":    args,
-			"request": gimlet.GetRequestID(ctx),
-		}))
+		if !errors.Is(err, context.Canceled) {
+			grip.Error(message.WrapError(err, message.Fields{
```
khelif96, line 61: How do you feel about keeping these around and just logging at a lower level if they exist? Just in case a context cancelled error ever appears we could have some logs to use for an investigation instead of throwing away everything.

ablack12: I commented this above:

> I thought we'd originally discussed not logging this, since it does still log 1000s of logs we don't intend to look at. Do we have actual planned follow up work to do something with these logs, with a ticket we can include in the message? Otherwise it doesn't make sense to log these until we do.
> 
> We could also possibly do something like
> 
> https://github.com/evergreen-ci/evergreen/blob/91095164b387b5216b26082e0b830cdb38fc2b16/rest/data/repotracker.go#L45
> 
> to avoid logging in a noisy way.

I'm pretty against logging something that logs this often without explicit planned follow up work.

ablack12: *explicit planned follow up OR a specific alert and guidance on how to approach it

ablack12 (approved):

khelif96 (approved):

PR #6521, opened 2023-05-18
EVG-19325: require GitHub patch definition tasks to finish

Kimchelly:
EVG-19325

### Description
This is another attempt at achieving the same goal as #6514, but making it specific to GitHub PR patches. The new behavior is that tasks created specifically by the GitHub patch definition must finish before users get a success/failure update on their GitHub PR status.

* Require all tasks that are automatically created for GitHub PR patches to finish in order for the patch to be put in a finished state. This also applies to tasks generated by tasks in the GitHub patch definition.
* Write an ADR for this new concept.

### Testing
Added unit tests and tested in staging in the sandbox (https://github.com/evergreen-ci/commit-queue-sandbox/pull/486) to ensure that:
* Unscheduling a GitHub PR task blocked its build/patch from sending a finished status
* Running them fixed it
* Scheduling more tasks beyond the defaults and then unscheduling them did not block a build/patch from sending a finished status.

### Documentation
Updated the ADRs and mentioned it in the docs for GitHub PRs.


model/generate.go
```diff
@@ -281,6 +281,9 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		ActivationInfo: activationInfo,
 		SyncAtEndOpts:  syncAtEndOpts,
 		GeneratedBy:    g.Task.Id,
+		// If the parent generator is required to finish, then its generated
+		// tasks inherit that requirement.
+		AllTasksAreEssentialToComplete: g.Task.IsEssentialToFinish,
```
hadjri, line 286: Same for this field - can we just inspect the requester directly?

Kimchelly: Same reason as above.

Kimchelly:
I don't think we need to handle that because I can't think of a reason you'd want to use `activate: false` in a PR. The reason that field exists is because by default, everything is run for mainline commits, which is the opposite of a PR alias, where you specifically select tasks to run.

model/task/task.go
```diff
@@ -248,6 +248,13 @@ type Task struct {
 
 	CanSync       bool             `bson:"can_sync" json:"can_sync"`
 	SyncAtEndOpts SyncAtEndOptions `bson:"sync_at_end_opts,omitempty" json:"sync_at_end_opts,omitempty"`
+
+	// IsEssentialToFinish indicates that this task must finish in order for
+	// the build and version to be considered complete. For example, tasks
+	// selected by the GitHub PR alias must succeed for the GitHub PR requester
+	// before its build or version can be considered complete, but tasks
+	// manually scheduled by the user afterwards are not required.
```
ablack12, line 256: Does this mean that the version status will also be updated only off of this, or overall will the patch status wait until this is done? 

I'd prefer if we had this wording / usage be more similar to IsGithubCheck, especially since we're only using it for PR patches / statuses.

ablack12 (changes requested):

model/task/task.go
Kimchelly: All the higher-level statuses (build, patch, version) will remain in "running" until the required tasks are all done.

I was thinking this is a pretty general/extensible feature since it's not doing any GitHub-specific actions like `IsGithubCheck` is, but if we think it's only useful for GitHub PR check marks and has no other use cases, I can rename it.

model/task/task.go
```diff
@@ -248,6 +248,13 @@ type Task struct {
...
+	IsEssentialToFinish bool `bson:"is_essential_to_finish" json:"is_essential_to_finish"`
```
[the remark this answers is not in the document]
hadjri: Gotcha - I see.

hadjri:
> I don't think we need to handle that because I imagine people don't use `activate: false` in a PR alias task. The reason that field exists is because by default, everything is run for mainline commits, which is the opposite of a PR alias, where you specifically select tasks to run.
> 
> While `activate: false` doesn't make sense for static tasks, I guess there could be some workflow where you generate tasks and don't activate them, but optionally keep them around in case you want to run them later. I'm not sure if people actually do this for a GitHub PR (it seems pretty niche). If you think it's a realistic enough case, I can try encoding some extra logic to skip the requirement for tasks that have specific activation conditions.

Since it's low risk, I guess I'm fine to make this change and then observe if any complaints are made if we need to add the complexity to handle it.

model/task/task.go
ablack12: Oh I see -- I didn't read the PR closely and misunderstood that we weren't special casing this for Github statuses.

Can we reconsider that? I think it's pretty important to users that Evergreen's statuses reflect what is actually happening in Evergreen. I also think that's reasonable -- the thing we want to do here is make sure that the Github check we're sending accurately reflects what they've asked us to check for Github, which can be separate from what we're doing in Evergreen overall. It seems like we could compute a status here, or check if all tasks essential to finish have been completed at the time of sending the status in triggers (and perhaps we could send a special notification in refresh like "not all required checks are scheduled" as a red check or pending status if we're concerned about people spamming it).

Kimchelly: I feel like this would be doable but I would also think that the patch being in an unfinished state (running) _is_ the patch's true state. Sending special alternative statuses for a specific condition seems like it would add more overhead to think about than I think is needed for something which is preventing unusual/atypical user behavior (at least, I can't think of a workflow where a user would want to intentionally unschedule required PR tasks). Since the build/version updating logic is pretty far from the GitHub status job, we would have to either encode this as another field/status in the build/version to represent "done but we can't tell GitHub that you're all good" (which I'd like to avoid) or dynamically re-check all the GitHub essential tasks in the trigger notification logic (which also seems very special case).

If it turns out people have a lot of questions about this and don't see the note in the docs, I think we can revisit making it more ergonomic later and consider sending a special notification just for this state. We could do something like add a display status in Spruce representing tasks that are essential but haven't run yet (like add an asterisk to them or something).

Kimchelly: Alright, I thought it over, and I'm gonna leave the build/patch in the running state but see if I can add some way of notifying the user of which/how many essential tasks still need to run.

ablack12: Lmk when I should look at this PR again @Kimchelly !

model/build/build.go
```diff
@@ -327,14 +327,16 @@ func (b Builds) InsertMany(ctx context.Context, ordered bool) error {
...
-// GetFinishedNotificationDescription returns a description of successful/failed tasks for the build,
-// to be used by jobs and notification processing.
-func (b *Build) GetFinishedNotificationDescription(tasks []task.Task) string {
+// GetPRNotificationDescription returns a GitHub PR status description based on
+// the statuses of tasks in the build, to be used by jobs and notification
+// processing.
+func (b *Build) GetPRNotificationDescription(tasks []task.Task) string {
 	success := 0
 	failed := 0
-	systemError := 0
```
Kimchelly, line 335: I removed this because `(task.Task).Status` only stores success/failed (the specific failure reasons in the UI are computed based on the task end details). Only tasks that are many years old store specific failure statuses like system-failed and whatnot.

Kimchelly (commented):
Alright I made some substantial changes so that:
* Tasks with specific activation are not essential anymore.
* It sends a GitHub pending status when all the tasks in the patch are finished except it's still waiting for an unscheduled essential task to be scheduled.
* Slightly fiddled with the `evergreen refresh` logic to check for the essential task completion and sends the same pending status back to GitHub as the automated process does.

model/task/task.go
Kimchelly: It should be ready now.

model/task_lifecycle.go
```diff
@@ -1306,6 +1320,26 @@ func updateBuildStatus(b *build.Build) (bool, error) {
 		return true, nil
 	}
 
+	// kim: TODO: manually test in GitHub PR status update
```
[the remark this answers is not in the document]
Kimchelly: It's an old comment to myself, I removed it.

model/build/build.go
```diff
@@ -360,22 +364,36 @@ func (b *Build) GetFinishedNotificationDescription(tasks []task.Task) string {
 		"build_id": b.Id,
 	})
 
-	if success == 0 && failed == 0 && systemError == 0 && other == 0 {
+	if unfinished == unscheduledEssential && unscheduledEssential > 0 {
```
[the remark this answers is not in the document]
Kimchelly: Good idea, that seems reasonable. I updated it to send this even if the other tasks aren't done yet.

model/task/task.go
ablack12: Sorry I missed this -- will look first thing tomorrow!

docs/Project-Configuration/Project-and-Distro-Settings.md
```diff
@@ -123,6 +123,14 @@ push to each pull request. In order for this to work, you must fill out
 at least 1 row in the section titled "GitHub Patch Definitions," as
 those define what tasks will be run in this patch.
 
+The status of each build and the patch overall will appear as GitHub statuses
+under your pull request. These will update as tasks in your PR complete. All the
+tasks selected by the GitHub patch definition **must** pass for the patch to be
+considered complete (except if they have specific activation conditions like
+`cron` or `activate: false`). For example, if you unschedule some of the tasks
```
ablack12, line 130: We can remove cron from this -- it's not relevant to patches

model/task_lifecycle.go
```diff
@@ -1209,25 +1213,37 @@ func getBuildStatus(buildTasks []task.Task) buildStatus {
...
 
+	var hasUnfinishedTask bool
+	var unfinishedEssentialTasks []string
 	// Check if tasks are started but not finished.
 	for _, t := range buildTasks {
 		if t.Status == evergreen.TaskStarted {
-			return buildStatus{status: evergreen.BuildStarted}
+			hasUnfinishedTask = true
 		}
 		if t.Activated && !t.Blocked() && !t.IsFinished() {
```
ablack12, line 1223: You use this logic and the status logic above to; would it be helpful to move this to a WillRun helper function or something?

model/task_lifecycle.go
```diff
@@ -1209,25 +1213,37 @@ func getBuildStatus(buildTasks []task.Task) buildStatus {
...
+	if hasUnfinishedTask {
+		return buildStatus{status: evergreen.BuildStarted, unfinishedEssentialTasks: unfinishedEssentialTasks}
 	}
 
-	// Check if all tasks are finished but have failures.
+	// Check if tasks are finished but failed.
 	for _, t := range buildTasks {
 		if evergreen.IsFailedTaskStatus(t.Status) || t.Aborted {
-			return buildStatus{status: evergreen.BuildFailed}
+			return buildStatus{status: evergreen.BuildFailed, unfinishedEssentialTasks: unfinishedEssentialTasks}
```
ablack12, line 1237: If we're just using the  length of unfinishedEssentialTasks, should we just store that instead of the list of task IDs?

ablack12 (approved):
LGTM with a couple of nits!

ablack12 (approved):
LGTM with a couple of nits!

model/task_lifecycle.go
Kimchelly: Yeah, I changed it to just an int.

PR #6524, opened 2023-05-19
EVG-19753: check project ref for commit queue disablement before other checks

Kimchelly:
EVG-19753

### Description
Move the PR mergeability check so that it happens before checking actual commit queue enqueueing validation (i.e. user authorization and GitHub mergeable state). I don't think the project ref lookup needs the refreshed `github.PullRequest` info from `checkPRIsMergeable` since the base branch should remain the same.

### Testing
Existing unit tests pass.

### Documentation
N/A


Kimchelly:
The smoke test is failing most likely because of the recent OTel changes, not this change.

PR #6529, opened 2023-05-19
EVG-16366 add labels to skip CI testing 

ablack12:
EVG-16366
### Description
Allow users to skip CI using a label in their PR title or description.

_Open question_ : should we send a status here? I'm not sure what the right status would be, and sending a comment might be annoying since presumably the user knows they meant to skip CI if it's in their title / description; that being said I could see this being an accidental support burden. (cc @Kimchelly since we've been talking about this more lately) 

### Testing
Tested on staging with https://github.com/evergreen-ci/commit-queue-sandbox/pull/487 with both comments and pushes.

### Documentation
Added to doc, will post in channel as well


rest/route/github.go
```diff
@@ -414,6 +417,21 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
 	if err != nil {
 		return errors.Wrap(err, "creating GitHub patch intent")
 	}
+	// If there are no errors with the PR, verify that we aren't skipping CI before adding the intent (and send a message).
+	for _, label := range skipCILabels {
+		if strings.Contains(pr.GetTitle(), label) || strings.Contains(pr.GetBody(), label) {
```
bynn, line 422: would it be worth making it case insensitive in case people write something like `[Skip-ci]`

docs/Project-Configuration/Github-Integrations.md
```diff
@@ -30,6 +30,12 @@ Sometimes Evergreen has trouble creating a PR patch, due to internal server erro
 
 Note that this is specific to Github PR checks; it won't retry a commit queue patch. For that, re-type `evergreen merge` (detailed below).
 
+#### Skip CI Testing
+
+Sometimes you may want to avoid having Evergreen create patches (perhaps because the work is in progress, or testing isn't relevant yet). 
+Simply including `[skip-ci]` or `[skip ci]` in your PR title or description will prevent us from creating a patch until it's removed.  
+
```
bynn, line 37: not sure if this is necessary but maybe mentioning that it will disable comments like evergreen retry

bynn (changes requested):
to address the question in the pr, I think it makes sense to not send anything, and sending a status might make it more confusing?

hadjri:
FWIW regarding statuses, I think giving a green check with a note mentioning tests were skipped could work. For example: 
image

docs/Project-Configuration/Github-Integrations.md
```diff
@@ -30,6 +30,12 @@ Sometimes Evergreen has trouble creating a PR patch, due to internal server erro
 
 Note that this is specific to Github PR checks; it won't retry a commit queue patch. For that, re-type `evergreen merge` (detailed below).
 
+#### Skip CI Testing
+
+Sometimes you may want to avoid having Evergreen create patches (perhaps because the work is in progress, or testing isn't relevant yet). 
+Simply including `[skip-ci]` or `[skip ci]` in your PR title or description will prevent us from creating a patch until it's removed.
```
hadjri, line 36: We should highlight that old commits will still not get patches created even if the label is lifted, and this only applies for subsequent commits.

ablack12:
@hadjri out of curiosity haha how did you build that example?
I'm slightly against sending a green checkmark with the evergreen context, since if there are branch protection rules, we would be erroneously passing them. We could send it with a red check or with a evergreen-skip context or something.

docs/Project-Configuration/Github-Integrations.md
ablack12: This is a really good point!

ablack12: I actually tested this and it does work: 
https://github.com/evergreen-ci/commit-queue-sandbox/pull/487

ablack12:
I think I'd prefer to leave it this way for now and see how people who use it feel, since I do suspect this will always be intentional.

hadjri (approved):

ablack12:
I think that does add some complication because the PR item doesn't return commit titles, so we'd have to query them separately or add some logic elsewhere. Also Evergreen tests are pretty cheap so I don't think it's worth our engineering time to add this for just ourselves.

PR #6530, opened 2023-05-19
docs: add metadata links

ablack12:
### Description
Making a doc to link to for the newsletter.

Not adding a screenshot bc this should change in EVG-16363 but will make a note in that ticket to add it.

preview:
image


hadjri (approved):
🚀

PR #6516, opened 2023-05-17
EVG-19874,EVG-19881: Add ProjectHealthView field to project ref

sophstad:
EVG-19874, EVG-19881

### Description
- Add ProjectHealthView allowing project admins to set whether waterfall should show only failed tasks (current default) or all tasks
- In `BuildFromService` and `ToService`, use "failed" as the projectHealthView type if none is specified in the database.
- Expose field in `projectSettings` GraphQL query
- Saving the field will be added in EVG-19876

### Testing
- Add test to ensure that BuildFromService returns "failed"
- Update GraphQL unit test

### Documentation
N/A

PR #6540, opened 2023-05-23
EVG-19876: Save project health view

sophstad:
EVG-19876

### Description
- Save project health view setting when Views & Filters page is updated

### Testing
- Add field to save unit test

### Documentation
N/A

### DIFF
--- a/graphql/generated.go
+++ b/graphql/generated.go
@@ -60265,7 +60265,7 @@ func (ec *executionContext) unmarshalInputProjectInput(ctx context.Context, obj 
 		asMap[k] = v
 	}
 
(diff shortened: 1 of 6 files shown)


minnakt (approved):
lgtm! :)))

sophstad:
Yeah the thought is that other views are possible! I think this was something Minna considered in her design. And the front end shouldn't be seeing empty strings: because of the BuildFromService implementation in #6516, empty strings will be converted to `failed`. AFAICT this is the closest we can get to setting a default string value when adding a new field. But lmk if you have other thoughts!

PR #6543, opened 2023-05-24
EVG-18657 query by version and parent version for certain operations

ablack12:
EVG-18657

### Description
A number of things don't behave quite right for child patch tasks (or require looping through and querying patches).
Specifically:
1) Did a bit of refactoring of functions for clarity, and removed some unused things I found
2) Query with child tasks for activation/deactivation, makespan, priority, and aborts.
3) Removed special graphql handling of child patches, since we now handle it in the operations listed in (2) directly.

We'd also want to add an index here on versionID and parent patch ID because [all of the clauses need to be supported](https://www.mongodb.com/docs/manual/reference/operator/query/or/#-or-clauses-and-indexes).

### Testing
Augmented existing unit testing, will test on staging.


ablack12:
Even after adding an index, UI operations are noticeably slower. Back to the drawing board.

PR #6545, opened 2023-05-24
EVG-19702 Log event when repo is created, add events for repo attachments

hadjri:
EVG-19702

### Description
The [repo](https://spruce.mongodb.com/project/634d56d3850e610cacfe7e0b/settings/event-log) in question from the ticket has no event logs because it was created by "attach to repo" being clicked for an existing project, copying over all settings and variables to the repo. By the time the button was clicked, the project had undergone changes and had event logs, but the repo was initialized with no event logs.

This is a bit confusing, because in the DB there is nothing acknowledging the existence of the created repo at all, so I added two changes to allow us to keep a better paper trail of what's happening:

- Logging a `PROJECT_ADDED` event when a repo is newly created from the "attach to repo" command
- Created two new events to log for a project, for when it attaches to / detaches from a repo, respectively.

### Testing
Tested with repo attachment / detachment in staging, and confirmed that newly attached repos get `PROJECT_ADDED` events.

### DIFF
--- a/model/project_event.go
+++ b/model/project_event.go
@@ -224,16 +224,14 @@ func GetAndLogProjectModified(id, userId string, isRepo bool, before *ProjectSet
 	return errors.Wrap(LogProjectModified(id, userId, before, after), "logging project modified")
 }
 
-// GetAndLogProjectModifiedWithRepoAttachment retrieves the project settings before and after the change, and logs the modification
-// as an event, alongside a repo attachment/detachment event.
-func GetAndLogProjectModifiedWithRepoAttachment(id, userId, attachmentType string, isRepo bool, before *ProjectSettings) error {
+// GetAndLogProjectRepoAttachment retrieves the project settings before and after the change, and logs the modification
+// as a repo attachment/detachment event.
+func GetAndLogProjectRepoAttachment(id, userId, attachmentType string, isRepo bool, before *ProjectSettings) error {
 	after, err := GetProjectSettingsById(id, isRepo)
 	if err != nil {
 		return errors.Wrap(err, "getting after project settings event")
 	}
-	if err = LogProjectModified(id, userId, before, after); err != nil {
-		return errors.Wrap(err, "logging project modified")
-	}
+
 	return errors.Wrap(LogProjectRepoAttachment(id, userId, attachmentType, before, after), "logging project repo attachment")
 }
 
--- a/model/project_event.go
+++ b/model/project_event.go
@@ -218,6 +218,15 @@ func GetAndLogProjectModified(id, userId string, isRepo bool, before *ProjectSet
 	return errors.Wrap(LogProjectModified(id, userId, before, after), "logging project modified")
 }
 
+func GetAndLogProjectModifiedWithRepoAttachment(id, userId, attachmentType string, isRepo bool, before *ProjectSettings) error {
+	after, err := GetProjectSettingsById(id, isRepo)
+	if err != nil {
+		return errors.Wrap(err, "getting after project settings event")
+	}
+
+	return errors.Wrap(LogProjectRepoAttachment(id, userId, attachmentType, before, after), "logging project modified")
+}
+
 // resolveDefaults checks if certain project event fields are nil, and if so, sets the field's corresponding flag.
 // ProjectChangeEvents must be cast to a generic interface to utilize event logging, which casts all nil objects of array types to empty arrays.
 // Set flags if these values should indeed be nil so that we can correct these values when the event log is read from the database.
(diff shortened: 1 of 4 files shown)


model/project_event.go
```diff
@@ -158,7 +159,7 @@ func (e *ProjectChangeEventEntry) SetBSON(raw mgobson.Raw) error {
 	return nil
 }
 
-// Project Events queries
+// MostRecentProjectEvents returns the n most recent project events for the given project ID.
```
Kimchelly, line 162: Bless these doc comments 🙏 (no change)

model/project_event.go
```diff
@@ -218,6 +224,17 @@ func GetAndLogProjectModified(id, userId string, isRepo bool, before *ProjectSet
 	return errors.Wrap(LogProjectModified(id, userId, before, after), "logging project modified")
 }
 
+// GetAndLogProjectModifiedWithRepoAttachment retrieves the project settings before and after the change, and logs the modification
+// as an event, alongside a repo attachment event.
```
Kimchelly, line 228: repo attachment/detachment event

model/project_ref.go
```diff
@@ -711,7 +711,7 @@ func (p *ProjectRef) DetachFromRepo(u *user.DBUser) error {
 	}
 	catcher.Add(UpsertAliasesForProject(repoAliasesToCopy, p.Id))
 
-	catcher.Add(GetAndLogProjectModified(p.Id, u.Id, false, before))
+	catcher.Add(GetAndLogProjectModifiedWithRepoAttachment(p.Id, u.Id, event.EventTypeProjectDetachedFromRepo, false, before))
```
Kimchelly, line 714: Can we add assertions that the event gets logged? I see there's already unit tests set up for detach/attach.

Kimchelly:
Does this require any UI changes for it to display in the project event logs?

hadjri:
> Does this require any UI changes for it to display in the project event logs?

No, the purposes of this ticket are just to provide evidence for what's occurring in the DB for our investigative purposes. If we want in the future we can add a way to display it nicely in Spruce if we think it's helpful.

malikchaya2:
> * Created two new events to log for a project, for when it attaches to / detaches from a repo, respectively.
> 
> ### Testing

Does this not show up in the event log tab automatically for projects that have it logged?

malikchaya2:
> > Does this require any UI changes for it to display in the project event logs?
> 
> No, the purposes of the changes here are to provide evidence for what's occurring in the DB for our investigative purposes. If we want in the future we can add a way to display it nicely in Spruce if we think it's helpful.

Does this not show up in the event log tab automatically for projects that have it logged?

model/project_event.go
```diff
@@ -247,7 +266,23 @@ func (p *ProjectSettings) resolveDefaults() *ProjectSettingsEvent {
...
+	eventData := constructProjectChangeEvent(username, before, after)
+	if eventData == nil {
+		return nil
+	}
+	return LogProjectEvent(event.EventTypeProjectModified, projectId, *eventData)
+}
+
+// LogProjectRepoAttachment logs an event for either the attachment of a project to a repo,
+// or a detachment of a project from a repo.
+func LogProjectRepoAttachment(projectId, username, attachmentType string, before, after *ProjectSettings) error {
```
ablack12, line 280: Putting this as a random thread to make conversation easier:

@hadjri from the ticket description I think it is important that we have events for repos show up in the event log (not necessarily the PROJECT_ADDED variable since it's new and we weren't doing that before; do the repo attachment ones show up?

> By the time the button was clicked, the project had undergone changes and had event logs, but the repo was initialized with no event logs.

This confuses me a little because the ticket makes it seem like the repo itself was modified and was missing event logs. Does that not appear to be the case?

hadjri: So yes I'm pretty sure that the linked repo was never modified after its creation and what the ticket was initially implying didn't actually happen. A few requests to modify the repo have come in but all failed due to insufficient permissions by the requester. It's looking like regular repo modifications log events without issue.

I reproduced this in Spruce - attaching a project to a new repo initializes the repo with all of the variables from the attaching project. This results in the repo having custom variables set, making it seem like it was modified when it was just copied over.

hadjri: image

image

Attaching what the event log looks like for a project attaching to a repo and then subsequently detaching from it. First picture is the event logs for the project, second is for the repo. The project event logs contain logs saying it is attaching to / detaching from a repo, and the repo event logs only get a single event log for the initial attachment, i.e. creation

hadjri:
> > > Does this require any UI changes for it to display in the project event logs?
> > 
> > 
> > No, the purposes of the changes here are to provide evidence for what's occurring in the DB for our investigative purposes. If we want in the future we can add a way to display it nicely in Spruce if we think it's helpful.
> 
> Does this not show up in the event log tab automatically for projects that have it logged?

image


> > > Does this require any UI changes for it to display in the project event logs?
> > 
> > 
> > No, the purposes of the changes here are to provide evidence for what's occurring in the DB for our investigative purposes. If we want in the future we can add a way to display it nicely in Spruce if we think it's helpful.
> 
> Does this not show up in the event log tab automatically for projects that have it logged?

It does, but it just shows up as a regular project modification. What I meant above is that the UI behavior of the event page remains unchanged as part of this PR.

model/project_event.go
ablack12: Ah okay that makes sense. Thanks!

Kimchelly (approved):

Kimchelly:
Yeah, I was asking about whether the modified log would render at all in the project event log. As long as it shows something, that's good enough to me.

malikchaya2:
> > > > Does this require any UI changes for it to display in the project event logs?
> > > 
> > > 
> > > No, the purposes of the changes here are to provide evidence for what's occurring in the DB for our investigative purposes. If we want in the future we can add a way to display it nicely in Spruce if we think it's helpful.
> > 
> > 
> > Does this not show up in the event log tab automatically for projects that have it logged?
> 
> It does, but it just shows up as a regular project modification. What I meant above is that the UI behavior of the event page remains unchanged as part of this PR.

Thanks, that makes sense! 
I think it might be worth a follow up ticket that logs that change on the repo level as well, so that if a repo has not been changed since creation, the user sees a creation or attachment event, and then nothing after.

PR #6547, opened 2023-05-24
EVG-19774 Add legacy project page redirect logic

hadjri:
EVG-19774

### Description
Added Spruce redirect functionality for the legacy project page, along with an admin flag to turn it on. The redirect will not be active until we toggle it.
### Testing
Tested locally.
### Documentation
Has been documented via email and banner that the legacy project page will be deprecated in 3 weeks (June 13).

### DIFF
--- a/service/project.go
+++ b/service/project.go
@@ -48,7 +48,7 @@ func (uis *UIServer) projectsPage(w http.ResponseWriter, r *http.Request) {
 	}
 	if flags.LegacyUIProjectPageDisabled {
 		newUIProjectsLink := fmt.Sprintf("%s/projects", uis.Settings.Ui.UIv2Url)
-		http.Redirect(w, r, newUIProjectsLink, http.StatusTemporaryRedirect)
+		http.Redirect(w, r, newUIProjectsLink, http.StatusPermanentRedirect)
 	}
 
 	allProjects, err := uis.filterViewableProjects(dbUser)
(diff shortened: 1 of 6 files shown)


service/project.go
```diff
@@ -42,6 +42,15 @@ func (uis *UIServer) filterViewableProjects(u *user.DBUser) ([]model.ProjectRef,
 func (uis *UIServer) projectsPage(w http.ResponseWriter, r *http.Request) {
 	dbUser := MustHaveUser(r)
 
+	flags, err := evergreen.GetServiceFlags()
+	if err != nil {
+		gimlet.WriteResponse(w, gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "retrieving admin settings")))
+	}
+	if flags.LegacyUIProjectPageDisabled {
+		newUIProjectsLink := fmt.Sprintf("%s/projects", uis.Settings.Ui.UIv2Url)
+		http.Redirect(w, r, newUIProjectsLink, http.StatusTemporaryRedirect)
```
malikchaya2, line 51: should this be `StatusPermanentRedirect` since we are planning to fully deprecate it?

hadjri: I said temporary in that there will be a temporary redirect after which the page won't exist at all.  That being said I can change it if you prefer

malikchaya2: That's a good point actually. I'm fine with either, I'll leave it up to you.

malikchaya2 (approved):

PR #6550, opened 2023-05-25
EVG-19723 use merge commit SHA even if mergeable check isn't ready

ablack12:
EVG-19723
### Description
Introduced logging to see how often we fall back on a different SHA than the merge commit, and why. Doesn't seem to happen often, and seems to be Github inconsistency rather than a reasonable return value. Leaving logging in to verify that this doesn't cause it's own problems; will revisit this ticket in two weeks.


agent/command/git.go
```diff
@@ -352,14 +352,19 @@ func (c *gitFetchProject) waitForMergeableCheck(ctx context.Context, comm client
...
 		attempt++
 		lastAttempt := attempt == getPRAttempts
 		info, err := comm.GetPullRequestInfo(ctx, td, conf.GithubPatchData.PRNumber, opts.owner, opts.repo, lastAttempt)
 		if err != nil {
 			return false, errors.Wrap(err, "getting pull request data from GitHub")
 		}
 		if info.Mergeable == nil {
+			// TODO EVG-19723: if using the merge commit SHA here doesn't cause issues, we should remove retrying here.
+			mergeSHA = info.MergeCommitSHA
+			mergeableCheckErr = true
```
hadjri, line 367: I think it's easier to understand if we prepend `if mergeSHA != "" { return true, nil }` right before L368 and then remove the change at the end of the function.

hadjri (changes requested):

agent/command/git.go
ablack12: If we don't return an error, then the function won't retry. This should theoretically also be temporary so I don't know how much we need to optimize.

hadjri: Gotcha, thanks

hadjri (approved):

PR #6551, opened 2023-05-25
EVG-20002: Remove legacy Performance Discovery plugin

zamj:
EVG-20002

### Description
We've released the new Perf-Analyzer app which is intended to replace the plugin Perf Discovery page. This PR removes the old Perf Discovery page and also removes the `Plugin` button on the old UI (the Perf Discovery app was the only plugin inside of that dropdown anymore).

Before we took on this project, Maria did an investigation and discovered that
a) It seems like the Perf Discovery page was mostly not working (except for some metrics)
b) No one was using it anymore and it was safe to just delete it.

### Testing
I haven't yet - is there a staging environment I can push to? The main thing for this is just making sure the Evergreen UI can still start up and I haven't horrifically broken it accidentally. 

### Documentation
I'll remove all mention of the Perf Discovery app from the docs once this is done. 

One overall question: do you want to do a notification like is being done for the old project settings page? I'm fine to leave this baking for a bit while that notification lets people know this page is being removed or I'm also happy just removing it. As far as I know, no one is using this page anymore.


service/ui.go
```diff
@@ -315,10 +315,6 @@ func (uis *UIServer) GetServiceApp() *gimlet.APIApp {
 	app.AddRoute("/json/task_log/{task_id}/{execution}").Wrap(needsLogin, needsContext, viewLogs).Handler(uis.taskLog).Get()
 	app.AddRoute("/task_log_raw/{task_id}/{execution}").Wrap(needsLogin, needsContext, allowsCORS, viewLogs).Handler(uis.taskLogRaw).Get()
 
-	// Performance Discovery pages
-	app.AddRoute("/perfdiscovery/").Wrap(needsLogin, needsContext).Handler(uis.perfdiscoveryPage).Get()
```
ablack12, line 319: Answering some of your questions from the PR via thread to facilitate easier discussion:

1. I'm fine to remove it entirely; if it's not really used and you don't expect this to throw people, I don't think we need to do a whole migration, but it's really up to you. I would suggest maybe sending an email letting people know where to look if that's a consideration.
2. I added the dist-staging task to your PR patch; I can deploy it when it's done and we can poke around a bit.
3. This change doesn't have anyting to do with the Perf Plugin checkmark right? Like this is still a thing to opt into, we're just removing legacy?
image

zamj: 1. Sounds good to me. I'm happy to send out an email, wait a day, and then merge this. If people really care, 24 hours seems like a reasonable amount of time to come find me, especially as we can always revert.
2. SGTM, thank you!
3. Nope, this is something completely different. The Perf checkmark will still work as normal.

ablack12: I suppose if it seems likely we'll need to revert, I'd prefer we add this as an admin toggle first to avoid having to do that (even though we can, it's not the ideal), similar to the change Malik made here:
https://github.com/evergreen-ci/evergreen/pull/6547/files

ablack12: Although it doesn't seem very likely if it literally doesn't work, and if there's a modern version of this we can just point them to

zamj: Yeah, I would be very surprised if anyone is using this. I've spoke to the server PERF team already and they said they're not.

zamj: I checked around and the changes LGTM on staging. Honestly, I think an email is probably overkill - I've sent out previous communications about the new Perf Analyzer in the last few weeks. I really doubt anyone is using this page so I'm just to merge this in.

ablack12 (approved):

zamj:
evergreen merge

PR #6554, opened 2023-05-25
EVG-19842: Remove branch for repos

sophstad:
EVG-19842

### Description
- Remove `branch` field from RepoRef and RepoRefInput. This field is removed from the front end in https://github.com/evergreen-ci/spruce/pull/1845.
- Don't set the repo's branch field when defaulting a page to the repo.

### Testing
- Update unit tests

### Documentation
N/A

### DIFF
--- a/rest/data/project_settings.go
+++ b/rest/data/project_settings.go
@@ -275,7 +275,7 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 		}
 
 		if mergedSection.Enabled {
-			if mergedSection.Branch == "" {
+			if mergedSection.Branch == "" && !isRepo {
 				return nil, errors.New("branch not set on enabled repo")
 			}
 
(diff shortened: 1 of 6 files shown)


sophstad:
evergreen retry

rest/data/project_settings.go
```diff
@@ -275,7 +275,7 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 		}
 
 		if mergedSection.Enabled {
-			if mergedSection.Branch == "" {
+			if mergedSection.Branch == "" && !isRepo {
 				return nil, errors.New("branch not set on enabled repo")
```
ablack12, line 279: maybe we should swap this to "enabled project" so this doesn't seem like an error. Can we also put a comment here to say branch isn't defined for repos so we remember why we special-cased this?

ablack12 (approved):
LGTM with nit

PR #6557, opened 2023-05-26
EVG-17998 Containerized Task doc page

hadjri:
EVG-17998

### Description
Walkthrough for users for container tasks which aims to provide background as well as enabling them to set up a containerized variant and run container tasks for themselves.

- Overview of the offering along with motivation and proposed benefits
- Example configurations
- Supported & unsupported features
- What's new in the UI
- Disk space considerations


docs/Project-Configuration/Containerized-Tasks.md
```diff
@@ -0,0 +1,118 @@
...
+Task events such as container assignment and status changes are also recorded, as well as the clearing of a task from a container once it has run its course.
+
+![container_event_logs.png](../images/container_event_logs.png)
+
+### Disk Space Considerations
+Each instance is provisioned with 200GB of space; however, the actual disk space available for each container task can vary depending on the number and the nature of tasks sharing the same instance. The more memory and CPU a task requires, 
+the less likely it is to share an instance with other tasks. Conversely, less resource-hungry tasks are more likely to share the instance with other tasks, which means they will also be sharing the available disk space.
+
+In practical terms, this means that while each instance has a maximum of 200GB of disk space, please bear in mind that the effective disk space available to your container tasks might be less, depending on what other tasks are running concurrently on the same instance.
```
Kimchelly, line 118: If this is the case, I would say that a hard requirement of the users running container tasks is they are not allowed to run tasks that use excessive disk space (several GB or more). Sharing disk space between containers is a breach of the container isolation promise, which we should eventually find a better way to fix (if technically possible).

docs/Project-Configuration/Containerized-Tasks.md
```diff
@@ -0,0 +1,118 @@
...
+#### Container Page
+
+The link in the task metadata sidebar takes you to the container page, which details the lifecycle of a container and their tasks. Like the host page, event logs exist charting the journey of a container task from initialization to termination.
+Task events such as container assignment and status changes are also recorded, as well as the clearing of a task from a container once it has run its course.
+
+![container_event_logs.png](../images/container_event_logs.png)
+
+### Disk Space Considerations
+Each instance is provisioned with 200GB of space; however, the actual disk space available for each container task can vary depending on the number and the nature of tasks sharing the same instance. The more memory and CPU a task requires, 
+the less likely it is to share an instance with other tasks. Conversely, less resource-hungry tasks are more likely to share the instance with other tasks, which means they will also be sharing the available disk space.
```
Kimchelly, line 116: I feel like this should be rephrased to say something about isolated vs. shared resources on the host - although you get dedicated memory/CPU to your task, you have to share disk space due to current technical limitations, hence why dedicating more memory/CPU makes it less likely you'll share the disk space with another task.

docs/Project-Configuration/Containerized-Tasks.md
Kimchelly, line 118: Does the disk space get reclaimed after a container exits?

docs/Project-Configuration/Containerized-Tasks.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Important Note
+Container tasks are still undergoing active testing and, at this stage, is used by the internal Evergreen team only.
+
+If you have any questions about container tasks or are interested in exploring how this feature could benefit your project, we encourage you to reach out to us. 
+We'll discuss its potential applications and assist you in preparing for its broader release.
+
+## What's Different
+
+While tasks running on containers come with all the capabilities you're familiar with in regular host tasks, there are a couple of differences to note as we roll out this new feature:
```
Kimchelly, line 19: To make it extra clear, we should also mention that containers are _not_ the same as hosts, which may require some user adjustment. For example, things that are heavily reliant on low-level components (such as systemd or Docker) are not available within a container. Might also be worth linking some kind of Docker docs or something about the differences (if one exists, I imagine it does).

Kimchelly (changes requested):
* I would actually avoid putting this in the public-facing docs. For the initial offering, we were hoping for a limited audience until we had sufficient confidence that it was ready for general use (at this point, it's definitely not ready for that). Can we move this into a Google doc or something? We can make it visible to everyone in 10gen, but avoid the possibility of someone just up and deciding to use it without our knowledge.
* I would include a section that distinguishes the new container offering from the distro-based Docker pool container solution so that people don't ask the difference.

docs/Project-Configuration/Containerized-Tasks.md
hadjri: Closest thing I could find was the [docker architecture overview](https://docs.docker.com/get-started/overview/#docker-architecture).

docs/Project-Configuration/Containerized-Tasks.md
hadjri: Yes, changed the comment to clarify that the available space on an instance fluctuates

docs/Project-Configuration/Containerized-Tasks.md
```diff
@@ -0,0 +1,118 @@
...
+    system:
+      cpu_architecture: x86_64
+      operating_system: windows
+      windows_version: 2019
+```
+Fields:
+
+-   `name`: a user-defined name for the container that represents the task or the environment of the container
+-   `working_dir`: the working directory for your tasks within the container. In the example, it's set to the root directory
+-   `image`: the Docker image to use for the container. As part of the initial offering, the images specificable in this field will be
```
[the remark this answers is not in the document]
hadjri: Can you clarify what you mean here on how this is different than specifying that they must pick from an approved registry?

docs/Project-Configuration/Containerized-Tasks.md
```diff
@@ -0,0 +1,118 @@
+# Containerized Tasks
+
+We're excited to introduce the ability to run Evergreen tasks in containers. 
+
+This offering is designed to streamline work and reduce friction caused by software dependency requirements. 
+Greater flexibility and control over task environments is achievable with containers, ensuring that each task runs in an isolated, dedicated space with its own specific set of dependencies. 
+
+This means fewer conflicts,
```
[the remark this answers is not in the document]
hadjri: I mean that the isolation of the environment makes things more flexible and tasks aren't really able to be disrupted by others on the same host. But you're right it's vague and we already touch on that in the docs elsewhere. Removing

hadjri:
Apologies for the time, other things came to the top of my queue this week. I'm keeping this PR open for provenance of the changes and the diffs, but I've ported everything over to this [Google doc](https://docs.google.com/document/d/1g17N47KPaIUABXGs0S3edfTxj5MzpNLCX9IJqweNr5g/edit#), and we can continue the discussion there.

Kimchelly:
@hadjri Is it alright if we close this PR? We can continue review in the other doc, but we might as well remove it from the Slack notification since this PR won't get updates.

docs/Project-Configuration/Containerized-Tasks.md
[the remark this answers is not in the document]
Kimchelly: We're good, this comment is not that relevant with the new wording in the doc because it explicitly says we'll publish the images on your behalf.

Kimchelly (approved):
The internal doc looks good mod one comment about Docker pools. I'm gonna LGTM, feel free to close this whenever.

PR #6558, opened 2023-05-26
EVG-19825: make project upsert safer

Kimchelly:
EVG-19825

### Description
* If a project is not already explicitly set to public/private, set it to private during all upserts.
* Consolidate `Update()` into `Upsert()` since they're pretty much the same thing.
* Reduce usage of upsert in places where the logic can do single field updates. I couldn't remove it from some places though (such as hiding a project), because they relied on doing whole-document updates.

### Testing
Added unit tests.

### Documentation
N/A

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -37,7 +37,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
 	"go.mongodb.org/mongo-driver/bson"
-	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 // The ProjectRef struct contains general information, independent of any revision control system, needed to track a given project.
@@ -896,17 +895,6 @@ func (p *ProjectRef) addPermissions(creator *user.DBUser) error {
 	}
 	return nil
 }
-
-// kim: TODO: have to fix/remove this as well since it's the same as upsert.
-// func (projectRef *ProjectRef) Update() error {
-//     return db.Update(
-//         ProjectRefCollection,
-//         bson.M{
-//             ProjectRefIdKey: projectRef.Id,
-//         },
-//         projectRef,
-//     )
-// }
 
 func findOneProjectRefQ(query db.Q) (*ProjectRef, error) {
 	projectRef := &ProjectRef{}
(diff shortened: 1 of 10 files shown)


model/project_ref.go
```diff
@@ -1895,16 +1885,43 @@ func (projectRef *ProjectRef) CanEnableCommitQueue() (bool, error) {
...
-		ProjectRefCollection,
-		bson.M{
-			ProjectRefIdKey: projectRef.Id,
-		}, projectRef)
+// overwriting the existing ref. If no project ref exists, a new one is created.
+func (p *ProjectRef) Upsert() error {
+	if p.Private == nil {
+		// Projects are private by default unless they've been specially made
+		// public.
+		p.Private = utility.TruePtr()
```
ablack12, line 1893: Good catch -- we can also change https://github.com/evergreen-ci/evergreen/blob/11dd59ac7f6f7b824d0cf99c7590692dae1d01d9/model/project_ref.go#L348 to use FromBoolTPtr

model/project_ref.go
```diff
@@ -1895,16 +1885,43 @@ func (projectRef *ProjectRef) CanEnableCommitQueue() (bool, error) {
...
+	}
+	_, err := db.Upsert(ProjectRefCollection, bson.M{ProjectRefIdKey: p.Id}, p)
```
ablack12, line 1895: The thing I was thinking would make this safer is using Update instead of Upsert, specifically listing all of the keys that need to be updated (since not everything is implemented in the legacy page). That said if it seems like overkill given the pending deprecation, I'm also fine with that.

ablack12 (approved):

model/project_ref.go
Kimchelly: Done.

model/project_ref.go
Kimchelly: I think update vs. upsert doesn't matter much in this case (one just happens to insert if it doesn't exist already, the update behavior is otherwise identical between them). It also seems like the upsert won't be going away entirely just by deprecating the legacy UI because it's used in other places like attaching/detaching repos and the REST routes to put/patch a project ref.

I did initially try listing every single key out but it seems like 1. it would be a burden to remember to update the upsert every time we added fields and 2. it would be a headache to keep the data model consistent with the DB model without having extra logic to `$set` and `$unset` as appropriate for zeroing/setting the omitempty fields (if you just `$set` everything, the `omitempty` fields will be `null` instead of nonexistent, which can make some find queries messier since they'll have to know this). Using upsert directly respects the `omitempty` behavior without needing to explicitly `$set`/`$unset` appropriately. I couldn't think of a good reason we would intentionally set `Private` explicitly to false (except via manual DB update), so I thought this was a good enough kind of fix.

PR #6559, opened 2023-05-30
EVG-19659 Add app auth to github functions

bynn:
EVG-19659

### Description
function that connect to the github client now use github app to generate an installation token to connect to it 

### Testing
went on staging and made sure we werent hitting the debug logs with a repo that has the app installed 
all unit tests pass 
fallback works if settings arent set (it is a little noisy but will be deleted once integration is complete)


globals.go
```diff
@@ -281,7 +281,7 @@ const (
 	DefaultJasperPort = 2385
 
 	GlobalGitHubTokenExpansion = "global_github_oauth_token"
-	githubAppPrivateKey        = "github_app_private_key"
+	githubAppPrivateKey        = "github_app_test_key"
```
bynn, line 284: name only changed for testing for now, will be changed back

thirdparty/github.go
```diff
@@ -35,6 +35,9 @@ const (
 	githubWrite      = "write"
 
 	GithubInvestigation = "Github API Limit Investigation"
+
+	defaultOwner = "evergreen-ci"
+	defaultRepo  = "commit-queue-sandbox"
```
bynn, line 40: will be "evergreen"
it is sandbox for testing currently

thirdparty/github.go
```diff
@@ -252,12 +255,34 @@ func getGithubClientRetry(token, caller string) *http.Client {
...
 
+func GetInstallationToken(ctx context.Context, owner, repo string, opts *github.InstallationTokenOptions) (string, error) {
+	settings, err := evergreen.GetConfig()
+	if err != nil {
+		return "", errors.Wrap(err, "getting evergreen settings")
+	}
+	token, err := settings.CreateInstallationToken(ctx, owner, repo, opts)
+	if err != nil {
+		grip.Info(message.Fields{
+			"message": "creating token",
```
Kimchelly, line 266: style: "error creating token" or something along those lines since it's just a plain error string.

thirdparty/github.go
```diff
@@ -252,12 +255,34 @@ func getGithubClientRetry(token, caller string) *http.Client {
 	)
 }
 
+func GetInstallationToken(ctx context.Context, owner, repo string, opts *github.InstallationTokenOptions) (string, error) {
```
Kimchelly, line 258: Doc comment for exported functions.

Although, if this function is just being used in this one package, I would instead make this unexported.

thirdparty/github.go
```diff
@@ -252,12 +255,34 @@ func getGithubClientRetry(token, caller string) *http.Client {
...
+	token, err := settings.CreateInstallationToken(ctx, owner, repo, opts)
+	if err != nil {
+		grip.Info(message.Fields{
+			"message": "creating token",
+			"ticket":  "EVG-19966",
+			"owner":   owner,
+			"repo":    repo,
+			"err":     err.Error(),
+		})
+		return "", errors.Wrap(err, "installation token")
```
Kimchelly, line 272: style: `errors.Wrap(err, "creating installation token")`

thirdparty/github.go
```diff
@@ -271,16 +296,40 @@ func GetGithubCommits(ctx context.Context, oauthToken, owner, repo, ref string,
...
 		defer resp.Body.Close()
+		if err == nil {
+			return commits, resp.NextPage, nil
+		}
+	}
+
+	// fallback to not using github app on error
+	grip.Debug(message.Fields{
+		"ticket":  "EVG-19966",
+		"message": "failed to get commits from GitHub",
```
Kimchelly, line 307: Seems like this debug message should be different and say something about falling back?

thirdparty/github.go
```diff
@@ -271,16 +296,40 @@ func GetGithubCommits(ctx context.Context, oauthToken, owner, repo, ref string,
...
+		"caller":  "GetGithubCommits",
+		"owner":   owner,
+		"repo":    repo,
+		"options": options,
+		"resp":    resp,
+		"error":   err.Error(),
```
Kimchelly, line 313: Rather than put the error string in the message fields directly, this should be `grip.Debug(message.WrapError(err, message.Fields{ ... }))`

thirdparty/github.go
```diff
@@ -271,16 +296,40 @@ func GetGithubCommits(ctx context.Context, oauthToken, owner, repo, ref string,
 	commits, resp, err := client.Repositories.ListCommits(ctx, owner, repo, &options)
 	if resp != nil {
 		defer resp.Body.Close()
+		if err == nil {
+			return commits, resp.NextPage, nil
+		}
+	}
+
+	// fallback to not using github app on error
```
Kimchelly, line 304: style: `// Fallback to not using the GitHub app on error.`

thirdparty/github.go
```diff
@@ -312,11 +363,36 @@ func GetGithubFile(ctx context.Context, oauthToken, owner, repo, path, ref strin
...
-		if resp.StatusCode == http.StatusNotFound {
+		if err == nil && file != nil && file.Content != nil {
+			return file, nil
+		}
+	}
+
+	// fallback to not using github app on error
+	grip.Debug(message.Fields{
+		"ticket":  "EVG-19966",
+		"message": "failed to get file from GitHub",
```
Kimchelly, line 374: Same thing as above (+ all the others in this file).

thirdparty/github.go
```diff
@@ -312,11 +363,36 @@ func GetGithubFile(ctx context.Context, oauthToken, owner, repo, path, ref strin
...
+		"caller":  "GetGithubFile",
+		"owner":   owner,
+		"repo":    repo,
+		"path":    path,
+		"options": opt,
+		"resp":    resp,
+		"error":   err.Error(),
```
Kimchelly, line 381: Same thing as above (+ all the others in this file).

thirdparty/github.go
Kimchelly, line 40: What does the default owner and repo do for the installation token? Is it okay to hard-code these as the installation token source for production?

thirdparty/github.go
```diff
@@ -653,23 +833,52 @@ func GithubAuthenticate(ctx context.Context, code, clientId, clientSecret string
...
-	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/git/refs/tags/%s", owner, repo, tag)
-	resp, err := client.Get(url)
+	httpClient := getGithubClientRetry(installationToken, "GetTaggedCommitFromGithub")
+	defer utility.PutHTTPClient(httpClient)
+	client := github.NewClient(httpClient)
+
+	tags, resp, err := client.Repositories.ListTags(ctx, owner, repo, nil)
 	if resp != nil {
 		defer resp.Body.Close()
+		if err == nil {
```
Kimchelly, line 846: nit: seems like you could avoid the heavily nested blocks (heavily nested block are generally not preferred) by just doing the for loop after the `if resp != nil` block on `tags`. The err in this case seems kind of irrelevant since you're searching for a specific tag result, and if it errored, then `tags` will be nil anyways and the for loop will be skipped.

thirdparty/github.go
```diff
@@ -271,16 +296,40 @@ func GetGithubCommits(ctx context.Context, oauthToken, owner, repo, ref string,
 	commits, resp, err := client.Repositories.ListCommits(ctx, owner, repo, &options)
 	if resp != nil {
 		defer resp.Body.Close()
+		if err == nil {
```
Kimchelly, line 299: It's more idiomatic in Go to simplify this (and other functions in this file that do a similar thing) down to fewer nested blocks down to:
```go
if resp != nil { defer resp.Body.Close() }
if err != nil {
    return <the expected result>, nil
}
```
I believe this code is still correct, but it's likelier to cause bugs if anyone changes the logic later on compared to having fewer nested blocks.

Kimchelly (changes requested):

thirdparty/github.go
```diff
@@ -252,12 +255,34 @@ func getGithubClientRetry(token, caller string) *http.Client {
...
+		return "", errors.Wrap(err, "getting evergreen settings")
+	}
+	token, err := settings.CreateInstallationToken(ctx, owner, repo, opts)
+	if err != nil {
+		grip.Info(message.Fields{
+			"message": "creating token",
+			"ticket":  "EVG-19966",
+			"owner":   owner,
+			"repo":    repo,
+			"err":     err.Error(),
```
malikchaya2, line 270: It looks like the other debug statements are using grip.Debug, do we want to do that for this as well?

thirdparty/github.go
```diff
@@ -873,66 +1197,153 @@ func AppAuthorizedForOrg(ctx context.Context, token, requiredOrganization, name
...
-	defer utility.PutHTTPClient(httpClient)
+	installationToken, _ := GetInstallationToken(ctx, owner, repo, nil)
 
+	httpClient := getGithubClientRetryWith404s(installationToken, "GithubUserPermissionLevel")
+	defer utility.PutHTTPClient(httpClient)
 	client := github.NewClient(httpClient)
 
 	permissionLevel, _, err := client.Repositories.GetPermissionLevel(ctx, owner, repo, username)
+	// fallback to not using github app on error
+	if err == nil && permissionLevel != nil && permissionLevel.Permission != nil {
```
malikchaya2, line 1208: same as above,  I think it might feel more readable if the fallback was wrapped in an err != nil instead of exiting if err==nil.

malikchaya2 (changes requested):

thirdparty/github.go
malikchaya2: I only saw this review after submitting mine, so there is some overlap 🙈

thirdparty/github.go
malikchaya2: We might want to move this to the admin page

thirdparty/github.go
bynn: discussed!

thirdparty/github.go
bynn: discussed!

Kimchelly (changes requested):
Alright, we discussed this a little bit offline so I could understand the workflow of using a GitHub token vs. the new GitHub app. I captured the gist I got here:

Using a default owner/repo is both necessary and not problematic for the functions where it's used. To get an API token, the GitHub API requires us to pass in _some_ repo that has the app installed because using an installation ID is the only method the GitHub API will allow us get a token. Each installation ID is unique to each repo (i.e. there's no way to get a generic GitHub app token that's not specific to a repo), so you always have to specify a repo to get an installation token. However, some special API calls (such as getting the remaining rate limit and checking org membership) don't care about which particular repo's installation you use to get the token, so passing any valid installation token for the app will return the correct result. The commit queue sandbox just so happens to have been configured with the app manually installed, so we use it as an arbitrary repo.

Since this is kind of a strange/unexpected special case in the usage of a GitHub app, I suggest:
* Make a helper to wrap `GetInstallationToken` in special cases where the owner/repo is irrelevant to the API call (e.g. checking org membership, checking the app's rate limit) and add doc comments on when to use each one (the repo default when it doesn't matter vs. actually passing in an owner/repo).
* Instead of hard-coding the sandbox as the default repo, I'd still prefer to move it to an admin setting so we can change it in the future if needed (e.g. if we want to delete the sandbox one day or remove its installation). My main concern here is that anything that's hard dependent on manual configuration is risky/likely to change since it's not codified anywhere. It's safer and easier to ops in prod if we have the ability to change our mind in the future to use a different default repo or if we do something like accidentally uninstall it from the sandbox. It's also slightly better practice for an application to try not to hard-code too many external dependencies if it can be avoided.

thirdparty/github.go
malikchaya2: I think it should be, because evergreen is open source so it should be generic when it can be.

malikchaya2 (changes requested):

thirdparty/github.go
malikchaya2: (this can be done in a separate ticket though)

agent/command/git_merge_pr.go
```diff
@@ -71,6 +71,7 @@ func (c *gitMergePR) Execute(ctx context.Context, comm client.Communicator, logg
 	token := c.Token
 	if token == "" {
 		token = conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion)
+		// TODO EVG-1996: return on error
```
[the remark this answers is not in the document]
bynn: this part was removed, but yes wrong number

thirdparty/github.go
```diff
@@ -255,21 +255,21 @@ func getGithubClientRetry(token, caller string) *http.Client {
 	)
 }
 
+// GetInstallationToken creates an installation token using github app auth.
```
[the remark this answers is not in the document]
bynn: done

bynn (commented):
created a function called getInstallationTokenWithoutOwnerRepo but i am open to suggestions if there is a better name

created a new ticket to move var to admin 

tested on staging
commits are being picked up, prs run tests, and the commit queue works fine

bynn (commented):
created a function called getInstallationTokenWithoutOwnerRepo but i am open to suggestions if there is a better name

created a new ticket to move var to admin 

tested on staging
commits are being picked up, prs run tests, and the commit queue works fine

malikchaya2 (approved):

Kimchelly (changes requested):

rest/route/agent.go
```diff
@@ -368,7 +368,30 @@ func (h *getExpansionsAndVarsHandler) Run(ctx context.Context) gimlet.Responder
...
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "finding version '%s'", t.Version))
+	}
+	if v == nil {
+		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("version '%s' not found", t.Version),
+		})
+	}
+
+	appToken, err := h.settings.CreateInstallationToken(ctx, v.Owner, v.Repo, nil)
```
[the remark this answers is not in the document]
bynn: do you think it would be worth it to make an extra db call to get the projectref? 

i did notice that it wasnt being set or used so i had to make some changes to set them in version code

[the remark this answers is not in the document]
Kimchelly: A find one for project ref should be almost instantaneous and not expensive for the DB.

I wouldn't change the logic to set them in the version just for this one piece of code. Deprecating them and using the project ref seems preferrable.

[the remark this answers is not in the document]
bynn: updated

rest/data/host_create.go
```diff
@@ -183,14 +183,19 @@ func makeProjectAndExpansionsFromTask(ctx context.Context, settings *evergreen.S
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "getting GitHub OAuth token from admin settings")
 	}
-	appToken, err := settings.CreateInstallationToken(ctx, v.Owner, v.Repo, nil)
+	pRef, err := model.FindBranchProjectRef(t.Project)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "finding project ref '%s'", t.Project)
+	}
```
Kimchelly, line 189: This needs to check if `pRef` is nil since the FindOne methods in Evergreen generally `return nil, nil`.

Kimchelly (approved):
LGTM mod two small things.

PR #6563, opened 2023-05-30
EVG-20013: new log model and interfaces

julianedwards:
EVG-20013

### Description
New `LogLine` data model, log fetcher interface and log iterator interface.

### Testing
N/A

### Documentation
N/A


model/log/log.go
```diff
@@ -0,0 +1,53 @@
...
+	// represented as a Unix timestamp in nanoseconds.
+	End *int64
+	// LineLimit specifies the number of log lines to return. Invalid if
+	// specified along with Tail or if Paginate is set to true.
+	LineLimit *int
+	// Tail specifies the last N log lines to return. Invalid if specified
+	// along with LineLimit or if Paginate is set to true.
+	Tail *int
+	// Paginate indicates whether to enable byte-based pagination.
+	Paginate bool
```
ybrill, line 52: Would we also allow them to specify a maximum number of bytes?

julianedwards: I was going back and forth about that. Cedar does not allow requesters to specify page size, but this is internal so I think it makes sense to make it configurable here and we can decide later if we want to let it be configurable over the REST API.

ybrill (approved):

ybrill (approved):

PR #6570, opened 2023-05-31
EVG-19998 cache unattainable task status

ybrill:
[EVG-19998](https://jira.mongodb.org/browse/EVG-19998)

#### Problem Statement
The [schedulableHostTasksQuery](https://github.com/evergreen-ci/evergreen/blob/fbbbd42cd673b24b9e12e9eb810674a7c389c1b2/model/task/db.go#L708) is inefficient if there are many documents that have a mix of blocked and unblocked dependencies. The reason for this is that we filter out tasks with blocked dependencies [here](https://github.com/evergreen-ci/evergreen/blob/fbbbd42cd673b24b9e12e9eb810674a7c389c1b2/model/task/db.go#L720), but the index can only be used to filter out tasks that have **all** their dependencies blocked. For tasks that have some unblocked dependencies we need to fetch the actual documents and filter the array[^1]. Some distros have many (~200k) tasks like this, and potentially long dependency arrays to look through, and this throws off the query planner.

#### Proposed Solution
Add a new field to the task document that caches whether there are any unattainable dependencies. This will allow the query to use a new index that will be added to replace the current one. Once we can assume a task has this field (after old tasks have TTLed, once there's a TTL 😅) this might also make things like [this](https://github.com/evergreen-ci/evergreen/blob/fbbbd42cd673b24b9e12e9eb810674a7c389c1b2/model/task/db.go#L252-L262) more efficient too.

#### Migration plan
Once this new field is being set for some time (a week?) we can assume that any new task that would want to run will have the field set. Any reset task will also have the field set. We can then adjust the `schedulableHostTasksQuery` to match on `unattainable_dependency: false`.

### Testing
Added a test for `MarkUnattainableDependency`

[^1]: The index has a key for each of the dependency subdocuments. The best we can do is avoid the keys in the index where there is an unattainable dependency, which filters the tasks that don't have **any** unblocked dependencies.
If a task has any unblocked dependencies that task will have a key that's `depends_on.unattainable: false` so that task will still be selected and need to be filtered.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -60,6 +60,7 @@ var (
 	SecondaryDistrosKey            = bsonutil.MustHaveTag(Task{}, "SecondaryDistros")
 	BuildVariantKey                = bsonutil.MustHaveTag(Task{}, "BuildVariant")
 	DependsOnKey                   = bsonutil.MustHaveTag(Task{}, "DependsOn")
+	UnattainableDependencyKey      = bsonutil.MustHaveTag(Task{}, "UnattainableDependency")
 	OverrideDependenciesKey        = bsonutil.MustHaveTag(Task{}, "OverrideDependencies")
 	NumDepsKey                     = bsonutil.MustHaveTag(Task{}, "NumDependents")
 	DisplayNameKey                 = bsonutil.MustHaveTag(Task{}, "DisplayName")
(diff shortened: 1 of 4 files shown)


ablack12 (approved):
TIL you can add footnotes to PR descriptions 🤯

PR #6571, opened 2023-05-31
EVG-19648 avoid deactivating previous tasks if they're depended on

ablack12:
EVG-19648

### Description
[This ticket](https://spruce.mongodb.com/task/mms_v20230412_deploy_gate_PACKAGE_CLOUD_DEPLOY_GATE_8d625c0f9fdc195431b5824995467764d6149181_23_04_17_17_30_16/logs?execution=0&logtype=event&sortBy=STATUS&sortDir=ASC) was unscheduled because a dependent task was deactivated because of [a previous success](https://spruce.mongodb.com/task/mms_v20230412_code_health_CHECK_PROM_ALERT_fe010402187c7649a2bfc7e7c2c11080854229ab_23_04_17_18_18_19/logs?execution=0). 

This PR (1) checks dependencies, and (2) removes legacy logic to include display task execution tasks (SetActiveState handles that). 

### Testing
Verified existing tests pass, and added a test to verify that a task with an active dependent task wouldn't be deactivated.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -1890,7 +1890,7 @@ func AbortAndMarkResetTasksForVersion(versionId string, taskIds []string, caller
 	return err
 }
 
-// HasUnfinishedTaskForVersion returns true if there are any scheduled but
+// HasUnfinishedTaskForVersions returns true if there are any scheduled but
 // unfinished tasks matching the given conditions.
 func HasUnfinishedTaskForVersions(versionIds []string, taskName, variantName string) (bool, error) {
 	count, err := Count(
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -1574,6 +1574,18 @@ func CountActivatedTasksForVersion(versionId string) (int, error) {
 	}))
 }
 
+// HasActivatedDependentTasks returns true if there are active tasks waiting on the given task.
+func HasActivatedDependentTasks(taskId string) (bool, error) {
+	numDependentTasks, err := Count(db.Query(bson.M{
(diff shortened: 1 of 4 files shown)


model/task/db.go
```diff
@@ -1574,6 +1574,18 @@ func CountActivatedTasksForVersion(versionId string) (int, error) {
 	}))
 }
 
+// HasActivatedDependentTasks returns true if there are active tasks waiting on the given task.
+func HasActivatedDependentTasks(taskId string) (bool, error) {
+	numDependentTasks, err := Count(db.Query(bson.M{
+		bsonutil.GetDottedKeyName(DependsOnKey, DependencyTaskIdKey): taskId,
+		ActivatedKey: true,
+	}))
+	if err != nil {
```
hadjri, line 1583: Can just `return numDependentTasks > 0, err` here

model/task/db.go
```diff
@@ -1574,6 +1574,18 @@ func CountActivatedTasksForVersion(versionId string) (int, error) {
 	}))
 }
 
+// HasActivatedDependentTasks returns true if there are active tasks waiting on the given task.
+func HasActivatedDependentTasks(taskId string) (bool, error) {
+	numDependentTasks, err := Count(db.Query(bson.M{
+		bsonutil.GetDottedKeyName(DependsOnKey, DependencyTaskIdKey): taskId,
+		ActivatedKey: true,
```
hadjri, line 1581: I think we'd want to check `OverrideDependenciesKey` is not set to true here too

hadjri (changes requested):

hadjri (approved):

PR #6574, opened 2023-06-01
EVG-18670 Standardize time taken formatting between patch finish notification and Spruce UI

malikchaya2:
EVG-18670

### Description
This standardizes how the time taken is formatted to avoid a very large integer in notifications for cases where there is no start time or an end time. 

### Testing 
Additional unit test plus existing tests

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"regexp"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
@@ -496,13 +494,6 @@ func getAllTaskStatuses(tasks []task.Task) []string {
 	return statusesArr
 }
 
-func formatDuration(duration string) string {
-	regex := regexp.MustCompile(`\d*[dhms]`)
-	return strings.TrimSpace(regex.ReplaceAllStringFunc(duration, func(m string) string {
-		return m + " "
-	}))
-}
-
 func removeGeneralSubscriptions(usr *user.DBUser, subscriptions []event.Subscription) []string {
 	filteredSubscriptions := make([]string, 0, len(subscriptions))
 	for _, subscription := range subscriptions {
@@ -512,6 +503,23 @@ func removeGeneralSubscriptions(usr *user.DBUser, subscriptions []event.Subscrip
 	}
 
 	return filteredSubscriptions
+}
+
+func makePatchDuration(t, m string) *PatchDuration {
+	var tPointer *string
+	if t != "0s" {
+		tPointer = &t
+	}
+
+	var mPointer *string
+	if m != "0s" {
+		mPointer = &m
+	}
+
+	return &PatchDuration{
+		Makespan:  mPointer,
+		TimeTaken: tPointer,
+	}
 }
 
 func getResourceTypeAndIdFromSubscriptionSelectors(ctx context.Context, selectors []restModel.APISelector) (string, string, error) {
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -505,21 +505,18 @@ func removeGeneralSubscriptions(usr *user.DBUser, subscriptions []event.Subscrip
 	return filteredSubscriptions
 }
 
-func makePatchDuration(t, m string) *PatchDuration {
-	var tPointer *string
-	if t != "0s" {
-		tPointer = &t
+func makePatchDuration(timeTaken, makeSpan string) *PatchDuration {
+	res := &PatchDuration{}
+
+	if timeTaken != "0s" {
+		res.TimeTaken = &timeTaken
 	}
 
-	var mPointer *string
-	if m != "0s" {
-		mPointer = &m
+	if makeSpan != "0s" {
+		res.Makespan = &makeSpan
 	}
 
-	return &PatchDuration{
-		Makespan:  mPointer,
-		TimeTaken: tPointer,
-	}
+	return res
 }
 
 func getResourceTypeAndIdFromSubscriptionSelectors(ctx context.Context, selectors []restModel.APISelector) (string, string, error) {
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -3097,6 +3097,22 @@ func GetTimeSpent(tasks []Task) (time.Duration, time.Duration) {
 	return timeTaken, latestFinishTime.Sub(earliestStartTime)
 }
 
+func GetFormattedTimeSpent(tasks []Task) (string, string) {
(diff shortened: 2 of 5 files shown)


hadjri (changes requested):

trigger/patch.go
```diff
@@ -273,15 +273,15 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
...
-	var makespan time.Duration
-	if utility.IsZeroTime(t.patch.FinishTime) {
-		patchTasks, err := task.Find(task.ByVersion(t.patch.Id.Hex()))
-		if err == nil {
-			_, makespan = task.GetTimeSpent(patchTasks)
-		}
-	} else {
-		makespan = t.patch.FinishTime.Sub(t.patch.StartTime)
+
+	tasks, err := task.Find(task.ByVersion(t.patch.Id.Hex()))
```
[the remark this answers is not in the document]
malikchaya2: I changed it because it wasn't a pattern followed in the whole function and I don't see why this case is different than other cases where we do return on error.

model/task/task.go
```diff
@@ -3097,6 +3097,22 @@ func GetTimeSpent(tasks []Task) (time.Duration, time.Duration) {
 	return timeTaken, latestFinishTime.Sub(earliestStartTime)
 }
 
+func GetFormattedTimeSpent(tasks []Task) (string, string) {
```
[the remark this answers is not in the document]
malikchaya2: done

trigger/patch.go
ablack12, line 277: If the issue is that we do this where there's no start or end time, should we just now only do this query if one of finished / started is zero?

trigger/patch.go
```diff
@@ -273,15 +273,15 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting tasks for patch '%s'", t.patch.Id)
+	}
+	if tasks == nil {
+		return nil, errors.Wrapf(err, "no tasks found for patch '%s'", t.patch.Id)
```
ablack12, line 282: shouldn't wrap error here

ablack12 (changes requested):

graphql/util.go
```diff
@@ -514,6 +505,23 @@ func removeGeneralSubscriptions(usr *user.DBUser, subscriptions []event.Subscrip
 	return filteredSubscriptions
 }
 
+func makePatchDuration(t, m string) *PatchDuration {
```
ablack12, line 508: Despite this being a helper function, t and m aren't helpful variables with understanding what's supposed to be passed in here -- can you elaborate a bit here?

graphql/util.go
```diff
@@ -514,6 +505,23 @@ func removeGeneralSubscriptions(usr *user.DBUser, subscriptions []event.Subscrip
...
+	var tPointer *string
```
ablack12, line 509: Rather than defining tPointer and mPointer, can we have a res := &PatchDuration{} and just set res.Makespan and res.TimeTaken to &t and &m directly?

graphql/util.go
ablack12: (Also not seeing where this is used)

trigger/patch.go
```diff
@@ -273,15 +273,15 @@ func (t *patchTriggers) makeData(sub *event.Subscription) (*commonTemplateData,
...
 	}
+	_, makespan := task.GetFormattedTimeSpent(tasks)
```
ablack12, line 284: Do we only ever use the makespan value here, in which case should we just return that?

ablack12: And/or rather than having a separate function for this, GetTimeSpent could take in a formatted bool

trigger/patch.go
malikchaya2: I may be misunderstanding the question. There was a formatting issue for containers tasks that failed to start after several attempts and were missing a start time or an end time for tasks. We need the list of tasks for the patch regardless.

graphql/util.go
malikchaya2: Very much yes :) 
Thanks! Too much copy/pasting.

graphql/util.go
malikchaya2: It's used in graphql/patch_resolver.go. 
renamed for clarity

trigger/patch.go
malikchaya2: This makes it cleaner for the patch resolver, and it doesn't add much work. 
If you feel strongly about it I can change it though.

malikchaya2: (I missed the second comment, I'll switch to one function with a formatted bool)

trigger/patch.go
ablack12: Ah I was referencing: 
> This standardizes how the time taken is formatted to avoid a very large integer in notifications for cases where there is no start time or an end time.

do we need the list of tasks only because of container tasks? Is there a way we can roll this up so we can avoid needing to do this?

graphql/util.go
ablack12: Ah yes I didn't see to expand this file ,thanks!

trigger/patch.go
malikchaya2: I actually can't change it to an isFormatted bool because GetTimeSpent() returns values as time.Duration, not strings.

trigger/patch.go
malikchaya2: We need the list of tasks for all cases so that we can iterate over it and look for the task with the earliest start time and latest finish time and use it to calculate the time taken. It's not something that was introduced for container tasks.

malikchaya2: What it was doing, where it didn't calculate the time spent if the finishTime wasn't ZeroTime wasn't a good approach, it's actually what caused it to have a huge number in the notification. You can see that looking at the data for the patch (attached below). Instead, it should always be calculated based on the tasks earliest start time and latest finish time, which is what spruce does when calculating Makespan. 
image

model/task/task.go
```diff
@@ -3097,6 +3097,23 @@ func GetTimeSpent(tasks []Task) (time.Duration, time.Duration) {
 	return timeTaken, latestFinishTime.Sub(earliestStartTime)
 }
 
+// GetTimeSpent returns the total time_taken and makespan of tasks as a formatted string
```
ablack12, line 3100: function name needs to be updated here

trigger/patch.go
ablack12: Makes sense, thanks!

ablack12 (approved):

hadjri (approved):

PR #6577, opened 2023-06-02
EVG-18898 js-test to ubuntu-2204

ybrill:
[EVG-18898](https://jira.mongodb.org/browse/EVG-18898)

### Description
Given the issues with js-test aren't related to the upgrade to ubuntu22.04 it shouldn't need to be a variant unto itself.


ablack12 (approved):
Fair enough haha

PR #6580, opened 2023-06-02
docs: list REST API model statuses and improve task status doc comments

Kimchelly:
* Update the REST API docs to enumerate the possible statuses for tasks/builds/versions.
* Seems like there's a lot of confusion around DB statuses vs. UI statuses, so I tried to make it more obvious in the doc comments what each task status represents, which ones are valid `(Task).Status` values, which ones are display (i.e. UI) statuses, and which ones aren't statuses at all.


globals.go
```diff
@@ -63,31 +71,38 @@ const (
 	TaskStarted = "started"
 
 	// The task statuses below indicate that a task has finished.
+	// TaskSucceeded indicates that the task has finished and is successful.
```
hadjri, line 74: Can we move the "undispatched", "dispatched", "started", "success", and "failed" statuses in a block at the top of this section in L44 since they're the only actual DB statuses?

globals.go
```diff
@@ -41,18 +41,26 @@ const (
...
 	// TaskUndispatched indicates either:
 	//  1. a task is not scheduled to run (when Task.Activated == false)
 	//  2. a task is scheduled to run (when Task.Activated == true)
 	TaskUndispatched = "undispatched"
-	TaskUnscheduled  = "unscheduled"
-	// TaskWillRun is a subset of undispatched tasks and is only used in the UI.
+	// TaskUnscheduled indicates that the task is undispatched and is not
+	// scheduled to eventually run. This is a display status, so it's only used
+	// in the UI.
+	TaskUnscheduled = "unscheduled"
```
hadjri, line 60: Can we similarly organize all display statuses into a block?

hadjri (changes requested):
LGTM, organizational nit

globals.go
Kimchelly: Done.

globals.go
Kimchelly: Done.

globals.go
```diff
@@ -41,19 +41,12 @@ const (
...
 
 	// TaskUndispatched indicates either:
 	//  1. a task is not scheduled to run (when Task.Activated == false)
 	//  2. a task is scheduled to run (when Task.Activated == true)
 	TaskUndispatched = "undispatched"
-	TaskUnscheduled  = "unscheduled"
-	// TaskWillRun is a subset of undispatched tasks and is only used in the UI.
-	TaskWillRun = "will-run"
 
 	// TaskDispatched indicates that an agent has received the task, but
```
bynn, line 51: I know you didn't write this but can we fix this here 
Otherwise looks fine

bynn (approved):

hadjri (approved):

PR #6614, opened 2023-06-05
EVG-18002 Add a middleware for Github payload validation 

malikchaya2:
EVG-18002

### Description
This adds a middleware that makes sure that the github payload is valid. While this adds some duplication, middleware is still the right place to handle auth, and parsing the body shouldn't be expensive. 

### Testing
Testing by creating a PR on staging and ensuring that we can still receive webhooks.

### DIFF
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -943,11 +943,11 @@ func (m *githubAuthMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request
 }
 
 func setGitHubPayload(r *http.Request, payload []byte) *http.Request {
-	return r.WithContext(context.WithValue(r.Context(), payloadKey, payload))
+	return r.WithContext(context.WithValue(r.Context(), githubPayloadKey, payload))
 }
 
 func getGitHubPayload(ctx context.Context) []byte {
-	if rv := ctx.Value(payloadKey); rv != nil {
+	if rv := ctx.Value(githubPayloadKey); rv != nil {
 		if t, ok := rv.([]byte); ok {
 			return t
 		}
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -34,8 +34,8 @@ type (
 
 const (
 	// These are private custom types to avoid key collisions.
-	RequestContext requestContextKey = 0
-	payloadKey     requestContextKey = 3
+	RequestContext   requestContextKey = 0
+	githubPayloadKey requestContextKey = 3
 )
 
 type projCtxMiddleware struct{}
@@ -938,15 +938,15 @@ func (m *githubAuthMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request
 		return
 	}
 
-	r = setPayload(r, payload)
+	r = setGitHubPayload(r, payload)
 	next(rw, r)
 }
 
-func setPayload(r *http.Request, payload []byte) *http.Request {
+func setGitHubPayload(r *http.Request, payload []byte) *http.Request {
 	return r.WithContext(context.WithValue(r.Context(), payloadKey, payload))
 }
 
-func getPayload(ctx context.Context) []byte {
+func getGitHubPayload(ctx context.Context) []byte {
 	if rv := ctx.Value(payloadKey); rv != nil {
 		if t, ok := rv.([]byte); ok {
 			return t
--- a/rest/route/middleware.go
+++ b/rest/route/middleware.go
@@ -1,8 +1,10 @@
 package route
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


rest/route/middleware.go
```diff
@@ -914,6 +917,37 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
...
+			"source":  "GitHub hook",
+			"message": "rejecting GitHub webhook",
+			"msg_id":  r.Header.Get("X-Github-Delivery"),
+			"event":   r.Header.Get("X-Github-Event"),
+		}))
+		gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(errors.Wrap(err, "validating github payload")))
+		return
+	}
+
+	//reset the body so that it can be parsed again
```
Kimchelly, line 946: nit: `// Reset the body so that it can be parsed again.`

rest/route/middleware.go
```diff
@@ -914,6 +917,37 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
 	next(rw, r)
 }
 
+// NewGithubAuthMiddleware returns a middleware that verifies the payload
```
Kimchelly, line 920: nit: period at the end of doc comment sentences.

rest/route/middleware.go
```diff
@@ -914,6 +917,37 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
...
+
+	_, err = github.ValidatePayload(r, githubSecret)
+	if err != nil {
+		grip.Error(message.WrapError(err, message.Fields{
+			"source":  "GitHub hook",
+			"message": "rejecting GitHub webhook",
+			"msg_id":  r.Header.Get("X-Github-Delivery"),
+			"event":   r.Header.Get("X-Github-Event"),
+		}))
+		gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(errors.Wrap(err, "validating github payload")))
```
Kimchelly, line 942: `validating GitHub payload`

rest/route/service.go
```diff
@@ -135,8 +136,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	// client_urls is used by the agent monitor deploy job which does not pass in user info
 	app.AddRoute("/distros/{distro_id}/client_urls").Version(2).Get().RouteHandler(makeGetDistroClientURLs(env))
 
-	// Middleware is omitted for webhook routes because they cannot be called by users and validation occurs in Parse function.
```
Kimchelly, line 138: Rather than remove it, ieems like this comment should be moved down to L140 since it still applies to the AWS webhooks.

rest/route/middleware.go
```diff
@@ -914,6 +917,37 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
...
+		return
+	}
+
+	//reset the body so that it can be parsed again
+	r.Body = io.NopCloser(bytes.NewBuffer(payload))
```
Kimchelly, line 947: Rather than doing this body-resetting logic to try resetting the request's state back to what it was originally, can we instead just set the request context value using `r.WithContext(context.WithValue(r.Context(), ...)`? That way, the route's parse method can just pull the parsed payload out and it'll also avoid the need to re-parse the body a second time.

Kimchelly (changes requested):

rest/route/middleware.go
```diff
@@ -914,6 +916,46 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
...
+		}))
+		gimlet.WriteResponse(rw, gimlet.MakeJSONErrorResponder(errors.Wrap(err, "validating GitHub payload")))
+		return
+	}
+
+	r = setPayload(r, payload)
+	next(rw, r)
+}
+
+func setPayload(r *http.Request, payload []byte) *http.Request {
```
Kimchelly, line 945: Since this is just a one-liner and used in one place I would avoid making a package-wide `setPayload` function since it pollutes the package namespace (or call it something more specific like `setGitHubPayload`).

rest/route/middleware.go
```diff
@@ -34,6 +35,7 @@ type (
 const (
 	// These are private custom types to avoid key collisions.
 	RequestContext requestContextKey = 0
+	payloadKey     requestContextKey = 3
```
Kimchelly, line 38: nit: I would call this `githubPayloadKey` since payload is kind of generic (unless we think it'll be used elsewhere).

rest/route/middleware.go
```diff
@@ -914,6 +916,46 @@ func (m *EventLogPermissionsMiddleware) ServeHTTP(rw http.ResponseWriter, r *htt
...
+
+	r = setPayload(r, payload)
+	next(rw, r)
+}
+
+func setPayload(r *http.Request, payload []byte) *http.Request {
+	return r.WithContext(context.WithValue(r.Context(), payloadKey, payload))
+}
+
+func getPayload(ctx context.Context) []byte {
```
Kimchelly, line 949: Same thing as above - if this is something specific to just the GitHub webhook, I would define it as a method on `githubHookApi`.

malikchaya2: Made it more specific. I would avoid making it part of githubHookApi as that's used in the route handler, while these functions are used in both the middleware and route handler.

PR #6625, opened 2023-06-06
EVG-18657 query with child patches

ablack12:
EVG-18657

### Description
A number of things don't behave quite right for child patch tasks (or require looping through and querying patches).
Specifically:

Did a bit of refactoring of functions for clarity, and removed some unused things I found.
Query with child tasks for activation/deactivation, makespan, priority, and aborts.
Removed special graphql handling of child patches, since we now handle it in the operations listed in (2) directly.

We'd also want to add an index here on parent patch ID because [all of the clauses need to be individually supported](https://www.mongodb.com/docs/manual/reference/operator/query/or/#-or-clauses-and-indexes). I originally had this index wrong so that's why it appeared not to work in staging but this is since fixed. Additionally this index will be pretty small since this value isn't set for many tasks.

### Testing
Verified that the UI loaded quickly and that closing a PR quickly unscheduled all tasks (including [child patches](https://spruce-staging.corp.mongodb.com/version/647faca797b1d31eb48b33b1/downstream-projects?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)).


graphql/util.go
```diff
@@ -423,22 +423,6 @@ func modifyVersionHandler(ctx context.Context, patchID string, modification mode
 	if err != nil {
 		return mapHTTPStatusToGqlError(ctx, httpStatus, err)
 	}
-
-	// Restart is handled through graphql because we need the user to specify
-	// which downstream tasks they want to restart.
-	if evergreen.IsPatchRequester(v.Requester) && modification.Action != evergreen.RestartAction {
```
ablack12, line 429: We don't need this anymore since we handle child patches in each individual part of modifyVersionHandler.

model/patch_lifecycle.go
```diff
@@ -910,36 +910,39 @@ func AbortPatchesWithGithubPatchData(createdBefore time.Time, closed bool, newPa
...
-					"project":        p.Project,
-					"version":        p.Version,
-				}))
-
-				catcher.Add(err)
+		if p.Version == "" { // Skip anything unfinalized
+			continue
+		}
+
+		if p.IsCommitQueuePatch() {
```
ablack12, line 917: Niothing here is actually different, just modified if statements a bit

malikchaya2 (approved):
🎉

model/task/task.go
```diff
@@ -2225,66 +2226,41 @@ func (t *Task) hasUnattainableDependency() bool {
...
-		}},
-	)
-	if err != nil {
-		return errors.Wrapf(err, "setting aborted statuses for tasks in build '%s'", buildId)
-	}
-
-	event.LogManyTaskAbortRequests(ids, reason.User)
-
-	return nil
+	return abortTasksByQuery(q, reason)
```
hadjri, line 2238: Can we wrap the error here to make it clear it is being done for a build? `return errors.Wrapf(err, "setting aborted statuses for tasks in build '%s'", buildId)`

model/lifecycle.go
```diff
@@ -85,6 +84,10 @@ func SetVersionActivation(versionId string, active bool, caller string) error {
 		if err != nil {
 			return errors.Wrap(err, "getting tasks to deactivate")
 		}
+		taskIds := []string{}
```
hadjri, line 87: This seems to be unused?

hadjri (changes requested):

model/task/db.go
```diff
@@ -1538,20 +1561,18 @@ func FindStuckDispatching() ([]Task, error) {
 }
 
 func FindAllTaskIDsFromVersion(versionId string) ([]string, error) {
-	q := db.Query(bson.M{VersionKey: versionId}).WithFields(IdKey)
+	q := db.Query(ByVersion(versionId)).WithFields(IdKey)
```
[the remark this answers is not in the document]
ablack12: I can add it, but I do think that doing cleanup of unrelated code is probably a nice-to-have more than a requirement

hadjri (approved):

PR #6579, opened 2023-06-02
EVG-19211 migrate to v2 of the AWS go SDK

ybrill:
[EVG-19211](https://jira.mongodb.org/browse/EVG-19211)

### Description
Followed the migration guide from [here](https://aws.github.io/aws-sdk-go-v2/docs/migrating/). Most of the changes are because the new version splits off many things into separate modules. Some changes are because the SDK now uses values instead of pointers for required fields. It also seems to prefer `int32` to `int64` now 🤷 . The input fields used to have a nice `.String()` method but they no longer do, so this uses json to similar effect in the `AWS API call` log messages.

There are a couple of references to the old SDK left because it's used to make values that get passed into other libraries that use them (pail and cocoa). They'll be cleaned up once those libraries are migrated.

### Testing
On staging it was able to request a task host (using Fleet) and a spawn host, attach/detach a volume, and terminate an instance. There were no errors in Splunk.

PR #6626, opened 2023-06-06
EVG-19211 migrate to v2 of the AWS SDK

ybrill:
[EVG-19211](https://jira.mongodb.org/browse/EVG-19211)

### Description
#6579 was reverted because it caused panics in prod. This PR cherry-picks it with a fix.
The panic was for an out of range access [here](https://github.com/evergreen-ci/evergreen/blob/7f39e06a3115986934a6bd1933e172a5049bb859/cloud/ec2_fleet.go#L559) which was confusing because it's checked [here](https://github.com/evergreen-ci/evergreen/blob/7f39e06a3115986934a6bd1933e172a5049bb859/cloud/ec2_client.go#LL849C24-L849C24). The answer was that we were returning a nil error instead of returning the new error.


julianedwards (approved):
Sorry I didn't catch this! LGTM

ybrill:
Thanks!

PR #6629, opened 2023-06-07
EVG-14117: fix commit queue unsticking and blocked cross-variant dependency state

Kimchelly:
EVG-14117

### Description
[Example patch showing the problem](https://spruce.mongodb.com/version/6463d26c2fbabe8388f149cd/tasks)

This fixes three independent bugs that I found in the patch, each of which resulted in its own issue:
1. When a task finishes, it marks itself finished, blocks any dependencies that are now unattainable, and updates the status for the finished task's build. This isn't quite right because if it blocks a dependency in a different build, that build's status may need to be updated. I fixed it by just updating the build statuses after blocking dependencies.
2. Fix an issue where the `(Build).AllTasksBlocked` key is not included in the build query when updating a version's status. This caused it to always be false, meaning builds that had all blocked dependencies were not detected and prevented versions from ever updating from started to failed.
3. Fix an issue where the commit queue auto-unsticker (the thing that automatically deletes commit queue items if it finds they're in an invalid state) does not remove items correctly. The old logic would just delete the item from the commit queue, but that's not right because it needs to also restart later items in the batch (as well as do other small things like update the removed item's status). I consolidated the logic so that the commit queue auto-unsticker calls `CommitQueueRemoveItem` (the thing that the UI does) to remove the item properly.

### Testing
Added unit tests.

Tested bug 1 and 2 in staging with a patch that has one task that depends on a task in another variant (the dependency always fails). Before the changes, [the version](https://spruce-staging.corp.mongodb.com/version/6480a009b2373662b6e5f0bc/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) would stay stuck in starting forever. After my change, [the version](https://spruce-staging.corp.mongodb.com/version/6480a5869dbe32073dd0c124/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) updates to failed as expected.

Bug 3 is mostly tested just by virtue of `CommitQueueRemoveItem` being tested. I didn't put in that much effort to unit test it because the commit queue machinery is going to go away.

### Documentation
N/A

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -208,7 +208,6 @@ func FindCommitQueueForProject(name string) (*restModel.APICommitQueue, error) {
 // CommitQueueRemoveItem dequeues an item from the commit queue and returns the
 // removed item. If the item is already being tested in a batch, later items in
 // the batch are restarted.
-// kim: TODO: move most of the tests into model.CommitQueueRemoveItem tests.
 func CommitQueueRemoveItem(cqId, issue, user, reason string) (*restModel.APICommitQueueItem, error) {
 	cq, err := commitqueue.FindOneId(cqId)
 	if err != nil {
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -205,8 +205,10 @@ func FindCommitQueueForProject(name string) (*restModel.APICommitQueue, error) {
 	return apiCommitQueue, nil
 }
 
-// CommitQueueRemoveItem dequeues an item from the commit queue. If the item is
-// already being tested in a batch, later items in the batch are restarted.
+// CommitQueueRemoveItem dequeues an item from the commit queue and returns the
+// removed item. If the item is already being tested in a batch, later items in
+// the batch are restarted.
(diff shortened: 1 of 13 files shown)


rest/data/commit_queue.go
```diff
@@ -205,8 +205,9 @@ func FindCommitQueueForProject(name string) (*restModel.APICommitQueue, error) {
 	return apiCommitQueue, nil
 }
 
-// CommitQueueRemoveItem dequeues an item from the commit queue. If the item is
-// already being tested in a batch, later items in the batch are restarted.
+// CommitQueueRemoveItem dequeues an item from the commit queue and returns the
```
hadjri, line 208: Can we rename this to something like `FindAndCommitQueueRemoveItem` just we don't have two exported functions with the same name?

hadjri (changes requested):
Great stuff 💯

rest/data/commit_queue.go
Kimchelly: Done. I imagine this function will go away once we move to the GitHub merge queue anyways.

hadjri (approved):

PR #6630, opened 2023-06-07
EVG-19867 Allow host.create to use function vars

hadjri:
EVG-19867

### Description
Functions that include `host.create` that pass in variables have those variables get ignored when `Also start any hosts this task started (if applicable)` is set in the UI and `CreateHostsFromTask` is run. If the distro value is set there, it will cause the spawn host command to fail, saying that no distro was specified.

The git diff is treating it weirdly but `WithCommandVars` is the new test case. I also refactored some of the looping logic at the beginning of the function to be more succinct.
### Testing
Test-driven approach, wrote initially failing tests and added change to fix them.
### Documentation
n/a

### DIFF
--- a/rest/data/host_create.go
+++ b/rest/data/host_create.go
@@ -105,25 +105,25 @@ func CreateHostsFromTask(ctx context.Context, settings *evergreen.Settings, t *t
 	createHostCmds := []apimodels.CreateHost{}
 	catcher := grip.NewBasicCatcher()
 	for _, commandConf := range projectTask.Commands {
-		var createHost *apimodels.CreateHost
+		var cmds []model.PluginCommandConf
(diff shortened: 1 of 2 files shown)


Kimchelly (commented):
The function vars change looks good, but since we're already thinking about it, does this also have to handle version parameters? The agent [receives version parameters from the REST route](https://github.com/evergreen-ci/evergreen/blob/1d5feca21ded59fc1ac5f8272126f917f518c101/rest/route/agent.go#L401) and basically treats them like expansions.

Kimchelly (approved):
Actually never mind, it seems like the parameters are already implicitly merged in `makeProjectAndExpansionsFromTask`. LGTM

PR #6633, opened 2023-06-08
EVG-20073 Add GitHub merge intent type

bsamek:
http://go/jira/EVG-20073

### Description
This adds a new GitHub merge intent type. Intents, like CLI patch intents and GitHub PR intents, are processed by the "patch-intent-processor" amboy job. This creates a patch. This PR will eventually be joined by PRs modifying the amboy job, modifying the clone behavior, and adding behavior to post to the GitHub checks API.

### Testing
All functions in this new intent are unit tested.

### Documentation
Added code overview to the ADR.

### DIFF
--- a/docs/decisions/2023-06-05_github_merge_queue.md
+++ b/docs/decisions/2023-06-05_github_merge_queue.md
@@ -52,3 +52,15 @@ types, "statuses" and "checks". Branch protection rules are fulfilled by
 statuses, whereas the GitHub merge queue listens for a check. The merge when
 ready button waits for the branch protection rules to be fulfilled before adding
 the PR to the merge queue.
+
+## Code Overview
+
+[x] indicates this is done. [ ] is to do. These will be removed when all is done.
+
+* [x] There is a new patch intent type in
+[model/patch](https://github.com/evergreen-ci/evergreen/blob/main/model/patch/github_merge_intent.go)
+* [ ] The intent will be procssed by the amboy [patch-intent-processor
+job](https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go).
+* [ ] New clone logic in the agent will clone the merge group branch.
+* [ ] Evergreen will post the result to the GitHub checks API.
+* [ ] A new UI element will allow users to opt into the GitHub merge queue.
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -38,6 +38,7 @@ const (
 
 	refTags = "refs/tags/"
 
+	// This will be removed when EVG-19964 is ready.
 	disableMergeGroup = true
 )
 
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -37,6 +37,8 @@ const (
 	evergreenHelpComment    = "evergreen help"
 
 	refTags = "refs/tags/"
+
+	disableMergeGroup = true
 )
 
 // skipCILabels are a set of labels which will skip creating PR patch if part of the commit description or message.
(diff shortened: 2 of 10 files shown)


docs/decisions/2023-06-05_github_merge_queue.md
```diff
@@ -52,3 +52,15 @@ types, "statuses" and "checks". Branch protection rules are fulfilled by
...
+
+[x] indicates this is done. [ ] is to do. These will be removed when all is done.
+
+* [x] There is a new patch intent type in
+[model/patch](https://github.com/evergreen-ci/evergreen/blob/main/model/patch/github_merge_intent.go)
+* [ ] The intent will be procssed by the amboy [patch-intent-processor
+job](https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go).
+* [ ] New clone logic in the agent will clone the merge group branch.
+* [ ] Evergreen will post the result to the GitHub checks API.
+* [ ] A new UI element will allow users to opt into the GitHub merge queue.
```
malikchaya2, line 66: I really like this format!

rest/route/github.go
```diff
@@ -222,23 +224,58 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
 			"event":    gh.eventType,
 			"org":      event.GetOrg(),
 			"repo":     event.GetRepo(),
 			"base_sha": event.GetMergeGroup().GetBaseSHA(),
 			"head_sha": event.GetMergeGroup().GetHeadSHA(),
+			"message":  msg,
 		})
 
+		// This will be removed when EVG-19964 is ready.
+		if disableMergeGroup {
```
malikchaya2, line 246: will `disableMergeGroup` in general be removed? If yes, should we maybe move this comment up to where disableMergeGroup is set?

rest/route/github.go
```diff
@@ -222,23 +224,58 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+				"msg_id":   gh.msgID,
+				"event":    gh.eventType,
+				"org":      event.GetOrg(),
+				"repo":     event.GetRepo(),
+				"base_sha": event.GetMergeGroup().GetBaseSHA(),
+				"head_sha": event.GetMergeGroup().GetHeadSHA(),
+				"message":  "can't add intent",
+			}))
+			return gimlet.NewJSONInternalErrorResponse(errors.Wrap(err, "adding patch intent"))
+		}
```
malikchaya2, line 261: This will never run. What is the intent behind committing it?  Is it mostly for documentation?

malikchaya2 (commented):
I love how well documented this code is! 
I just had some minor comments/questions.

rest/route/github.go
bsamek: Yes. It's a feature flag https://martinfowler.com/articles/feature-toggles.html. Copied the comment there.

rest/route/github.go
bsamek: It'll run when we remove the feature flag. The idea is not to expose this code path until we're ready.

rest/route/github.go
malikchaya2: Makes sense. We should probably set it up like so: https://github.com/evergreen-ci/evergreen/commit/86a50a3828e2d30a165609da52c3fc403227b68b

bsamek: I don't think there's a need for that. For a feature like, say, a new way of doing host creation, flipping the feature flag has an instant effect on the entire system. In this case though, flipping the feature flag should change nothing at all, since it will be off for all projects. And then we can slowly roll it out for individual projects. I'm including this flag here just to be extra cautious during development.

malikchaya2: I guess that makes sense since we're expecting a short development cycle for this project.

malikchaya2 (approved):

PR #6636, opened 2023-06-08
hotfix: Comment out idle termination job populator

Kimchelly:
The server upgrade is causing the idle host query within the job to stall. For now, we don't have a fine-grained way to stop the job without just commenting it out.


Kimchelly:
It compiles locally and is simple enough, so I'm just gonna merge.

PR #6637, opened 2023-06-08
EVG-20112 hint for IdleEphemeralGroupedByDistroID

ybrill:
[EVG-20112](https://jira.mongodb.org/browse/EVG-20112)

### Description
Upgrading the db to 6.0 caused this query to stop using the expected index and the query was getting stuck. It seems that the sort on `creation_time` was causing it to prefer the `creation_time_1` index instead of using the `started_by_1_status_1`.

### Testing
The explain plan is much better now. I'll take it for a spin in staging to make sure it works.

### DIFF
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -205,7 +205,7 @@ func IdleEphemeralGroupedByDistroID() ([]IdleHostsByDistroID, error) {
 		},
 	}
 
-	if err := db.Aggregate(Collection, pipeline, &idlehostsByDistroID); err != nil {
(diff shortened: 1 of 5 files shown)


Kimchelly (approved):
I'm just gonna LGTM the changes since they seem fine for prod, but does the smoke test/local-evergreen need these indexes as well (in `load-smoke-data`)?

Kimchelly:
I'm going to merge despite `test-units` failing (the test setup is just missing the index) - we can fix the test failure (and smoke/local index issues) in a later PR once the prod issue is resolved. I confirmed that the index exists in prod.

PR #6639, opened 2023-06-09
EVG-20045 start hosts faster

ybrill:
[EVG-20045](https://jira.mongodb.org/browse/EVG-20045)

### Description
Right now when we insert a host intent we wait around for the `PopulateHostCreationJobs` job to find the host intent (it runs once a minute) and enqueue a `provisioning-create-host` for it. This PR makes that whenever we want to create a host we immediately enqueue a `provisioning-create-host` job. This is relevant in three places: task hosts, host.create hosts, and spawn hosts.

Because the `provisioning-create-host` jobs are coming from many places now there's no central place to throttle how many of these jobs are getting created. Not all is lost, however, because there's still a throttle [here](https://github.com/evergreen-ci/evergreen/blob/6dace20f6ad21225fe63f04046f7d047d085ea48/units/provisioning_create_host.go#L253-L258) that will limit how many hosts are starting at once.

One interesting thing I learned about getting the amboy group queue is that if [Get](https://github.com/mongodb/amboy/blob/151f8470357ee2a976ef75d888b2bd5307339c80/queue/group_remote_mongo.go#L259) needs to construct a new queue (i.e. we haven't called Get on this queue group since the last pruning) the context we call Get with gets propagated down to the driver ([here](https://github.com/mongodb/amboy/blob/151f8470357ee2a976ef75d888b2bd5307339c80/queue/group_remote_mongo.go#L303)) and when it's cancelled the driver will close. So we need to give it the application context. The one concern I had was that [env.Context()](https://github.com/evergreen-ci/evergreen/blob/0b38c0c784757ca57aa54a8ced532d02c49435db/environment.go#L394) returns a `WithCancel()` copy of the application's context and we never call cancel. [The docs](https://pkg.go.dev/context#WithCancel) mention calling cancel to release resources. I think this is okay, though, since as long as the parent is a cancel context (which [it is](https://github.com/evergreen-ci/evergreen/blob/0b38c0c784757ca57aa54a8ced532d02c49435db/environment.go#L177) in this case) [the child context doesn't get its own watcher goroutine](https://github.com/golang/go/blob/1c89de6a71cf5a4c6e9545582cf277bf69bf9206/src/context/context.go#L473-L487) and is just added to a list of contexts to cancel when the parent is cancelled. Maybe this is being too clever 🤷 . We can keep an eye on the goroutine count after the deploy to make sure.

### Testing
On staging we were able to start task hosts, host.create hosts, and spawn hosts.


rest/data/host_create.go
```diff
@@ -422,6 +432,15 @@ func makeEC2IntentHost(taskID, userID, publicKey string, createHost apimodels.Cr
 		return nil, errors.Wrap(err, "inserting intent host")
 	}
 
+	appCtx, _ := env.Context()
```
johndaniels, line 435: This is repeated 4 times... does it make sense to have a function to create and enqueue provisioning-create-host jobs or is that premature deduplication?

units/crons.go
```diff
@@ -760,73 +768,15 @@ func PopulateHostCreationJobs(env evergreen.Environment, part int) amboy.QueueOp
...
-						continue
-					}
-				}
-				// only increment for task hosts, since otherwise
-				// spawn hosts and hosts spawned by tasks could
-				// starve task hosts
-				submitted++
-			}
-
 			catcher.Wrapf(queue.Put(ctx, NewHostCreateJob(env, h, ts, 0, false)), "enqueueing host creation job for host '%s'", h.Id)
```
johndaniels, line 777: Is there a race condition between this and the various places where we create an intent host and then later create a NewHostCreateJob? I think no because of scopes, but wanted to ask.

rest/data/host_create_test.go
```diff
@@ -163,13 +177,8 @@ buildvariants:
...
 
-		settings := &evergreen.Settings{
-			Credentials: map[string]string{"github": "token globalGitHubOauthToken"},
-		}
-		assert.NoError(t, evergreen.UpdateConfig(settings))
-
-		assert.NoError(t, CreateHostsFromTask(ctx, settings, &t1, user.DBUser{Id: "me"}, ""))
-		createdHosts, err := host.Find(host.IsUninitialized)
+		assert.NoError(t, CreateHostsFromTask(ctx, env, &t1, user.DBUser{Id: "me"}, ""))
+		createdHosts, err := host.Find(db.Query(bson.M{host.StartedByKey: "me"}))
```
johndaniels, line 181: Is there a race with the provisioning-create-host task here? This query will only make sense if that task has run, right?

ybrill: I think there is a race (so `host.IsUninitialized` was inconsistent) but the `StartedBy` field is set [when the host intent is created](https://github.com/evergreen-ci/evergreen/blob/88bc52da38ee865940f567a33838051603be1c87/model/host/host_intent.go#L60) and persisted when it's started, so it seemed like it would cover both cases.

units/crons.go
ybrill: That's my understanding as well: this relies on scopes to prevent multiple jobs for [the same host](https://github.com/evergreen-ci/evergreen/blob/6dace20f6ad21225fe63f04046f7d047d085ea48/units/provisioning_create_host.go#L62). Because `SetEnqueueAllScopes` is [true](https://github.com/evergreen-ci/evergreen/blob/6dace20f6ad21225fe63f04046f7d047d085ea48/units/provisioning_create_host.go#L63) whichever comes second will fail to enqueue a job.

rest/data/host_create.go
ybrill: done

johndaniels (approved):

PR #6647, opened 2023-06-12
EVG-19189 Instrument calls to GitHub's API

ybrill:
[EVG-19189](https://jira.mongodb.org/browse/EVG-19189)

### Description
Add tracing for calls to GitHub's API.

### Testing
Created traces in staging like [this one](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/jJy1ztiNneH/trace/7bB6uFjjW98) from when I created a patch. This should enable things like count by API, p90(retries), etc.


ybrill:
https://github.com/mongodb/grip/pull/146 for grip

thirdparty/github.go
```diff
@@ -47,6 +61,8 @@ var UnblockedGithubStatuses = []string{
 	githubPRUnstable,
 }
 
+var tracer = otel.GetTracerProvider().Tracer(packageName)
```
julianedwards, line 64: I would move the otel-related global variables to their own file.

thirdparty/github.go
```diff
@@ -171,98 +198,62 @@ func githubShouldRetry(caller string) utility.HTTPRetryFunction {
...
+func getGithubClient(ctx context.Context, token, caller string, config retryConfig, attributes []attribute.KeyValue) (context.Context, *http.Client, func()) {
 	grip.Info(message.Fields{
 		"ticket":  GithubInvestigation,
 		"message": "called getGithubClient",
 		"caller":  caller,
 	})
 
-	return utility.GetOAuth2HTTPClient(token)
-}
+	ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(append(attributes, attribute.String(endpointAttribute, caller))...))
```
julianedwards, line 225: Nit: if order does not matter here, you can also just do:
```
ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(attribute.String(endpointAttribute, caller), attributes...))
```

julianedwards (changes requested):

thirdparty/github.go
```diff
@@ -37,6 +41,16 @@ const (
 	GithubInvestigation = "Github API Limit Investigation"
 )
 
+const (
+	packageName       = "github.com/evergreen-ci/evergreen/thirdparty"
+	endpointAttribute = "evergreen.github.endpoint"
```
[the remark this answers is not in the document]
ybrill: done

thirdparty/github.go
ybrill: done

thirdparty/github.go
```diff
@@ -171,98 +198,62 @@ func githubShouldRetry(caller string) utility.HTTPRetryFunction {
...
+	return ctx,
+		utility.SetupOauth2CustomHTTPRetryableClient(
+			token,
+			githubShouldRetry(caller, config),
+			utility.RetryHTTPDelay(utility.RetryOptions{
+				MaxAttempts: numGithubAttempts,
+				MinDelay:    githubRetryMinDelay,
+			}),
+			client,
+		), func() {
```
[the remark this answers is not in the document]
ybrill: 😞 this was such a headache!
I realized that that function was unable to extract the transport from the otelhttp transport since it's [private](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/e94326455e25cede6b4381d3f90efae18f71b29a/instrumentation/net/http/otelhttp/transport.go#L33). So the only way to reset the transport without making a whole new one is to hold on to the client we give the otelhttp transport and reinstate it before we call `PutHTTPClient`.

thirdparty/github.go
julianedwards: Oh that is annoying :(

julianedwards (approved):

PR #6650, opened 2023-06-13
EVG-19895: Add Parsley filters in local testdata

minnakt:
EVG-19895

### Description
This PR just adds some data for default filters for the <s>Spruce</s> mongodb-mongo-master project to use in Parsley's e2e tests.


ablack12 (approved):

testdata/local/project_ref.json
```diff
@@ -4,4 +4,4 @@
 {"_id" : "performance", "owner_name" : "mongodb", "repo_name" : "mongo", "branch_name" : "NOBRANCH", "repo_kind" : "github", "enabled" : true, "private" : true, "batch_time" : 0, "remote_path" : "etc/perf.yml", "identifier" : "performance", "display_name" : "MongoDB Microbenchmarks (main)", "local_config" : "", "tracked" : true, "deactivate_previous" : false, "repotracker_error" : {"exists" : false, "invalid_revision" : "", "merge_base_revision" : ""}, "admins" : ["ian.whalen", "henrik.ingo", "david.daly", "cristopher.stauffer"]}
 {"_id":"mongodb-mongo-master","owner_name":"mongodb","repo_name":"mongodb","branch_name":"NOBRANCH","repo_kind":"github","enabled":0,"private":1,"batch_time":0,"remote_path":".docker-test.yaml","identifier":"mongodb-mongo-master","display_name":"mongo","local_config":"","deactivate_previous":1,"tracked":0,"repotracker_error":null,"commit_queue":{"enabled":true}}
```
khelif96, line 5: Does this work for the resmoke logs as well?

minnakt: I was going to leave `mongodb-mongo-master` intentionally empty; I didn't feel like there was a need to duplicate testing for both ansii and resmoke logs since functionally it would be the same, but let me know if you feel differently

(but yes as long as there's a project identifier associated with the task in the log it should work! tested it locally)

khelif96: We do have precedent for duplicating some tests for resmoke and ansii logs but if you think it isn't necessary for these tests its fine with me.
image

khelif96 (approved):

testdata/local/project_ref.json
khelif96: We do have precedent for duplicating some tests for resmoke and ansii logs but if you think it isn't necessary for these tests it's fine with me.
image

minnakt: yup I think it's fine -- I decided I'm going to switch the project filters to be for Mongo rather than Spruce, since I think that's a more realistic scenario

PR #6652, opened 2023-06-13
Revert "EVG-19325: require GitHub patch definition tasks to finish (#6521)

Kimchelly:
EVG-19325

Revert because of a suspicion that it's increasing the number of GitHub statuses sent and is stressing the API limit.


ablack12 (approved):
Didn't read but if it's passing tests LGTM 👍

PR #6653, opened 2023-06-13
EVG-20029 Remove legacy UI project page and flag

hadjri:
EVG-20029

### Description
Now that the legacy project page has been deprecated, we should be free to remove the deprecation flag as well as the page itself.
### Testing
Tests should remain green.


service/project.go
```diff
@@ -40,39 +40,8 @@ func (uis *UIServer) filterViewableProjects(u *user.DBUser) ([]model.ProjectRef,
...
-		NewUILink    string
-		SlackAppName string
-	}{allProjects, uis.GetCommonViewData(w, r, true, true), newUILink, slackAppName}
-
-	uis.render.WriteResponse(w, http.StatusOK, data, "base", "projects.html", "base_angular.html", "menu.html")
+	newUIProjectsLink := fmt.Sprintf("%s/projects", uis.Settings.Ui.UIv2Url)
+	http.Redirect(w, r, newUIProjectsLink, http.StatusPermanentRedirect)
 }
 
 func (uis *UIServer) projectPage(w http.ResponseWriter, r *http.Request) {
```
ablack12, line 47: Can we remove this too? And modifyProject?

hadjri: `projectPage` returns a JSON response of the project which is why I didn't include it as the removal, but I'm fine to get rid of it. Done.

ablack12 (commented):
To test this, can you also deploy this to staging (if you didn't already) and verify that the waterfall, tasks, variants work as expected (and that the legacy project dropdown button still redirects to spruce)

hadjri:
> To test this, can you also deploy this to staging (if you didn't already) and verify that the waterfall, tasks, variants work as expected (and that the legacy project dropdown button still redirects to spruce)

Checked the redirect still works, and all the regular task, version, build pages are working as expected, and ran a few restarts as well. No odd behavior or errors in Splunk

ablack12 (approved):

PR #6655, opened 2023-06-14
EVG-19845 Allow filtering of unscheduled tasks in the tasks table 

khelif96:
EVG-19845

### Description
The unscheduled status was not being passed through because the original task statuses resolver would not fetch unactivated tasks which we now surface. 
* This PR refactors the resolver to use a dedicated query that fetches those tasks.
* I also threaded context into one of the base tasks' status queries.

Disclaimer: This query uses [`$sortArray`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sortArray/) which is a feature introduced in MongoDB 5.2 the prod and staging DBs are now on 6.0 so please upgrade your local instance. 
### Testing

image


graphql/version_resolver.go
khelif96, line 1: This file is not generated but github marks it as generated. 
https://jira.mongodb.org/browse/EVG-20166

model/task/db.go
```diff
@@ -2135,6 +2135,60 @@ func GetTasksByVersion(ctx context.Context, versionID string, opts GetTasksByVer
...
+
+	pipeline = append(pipeline, bson.M{
+		"$group": bson.M{
+			"_id": nil,
+			"statuses": bson.M{
+				"$addToSet": "$" + DisplayStatusKey,
+			},
+		},
+	})
+	pipeline = append(pipeline, bson.M{
```
ablack12, line 2162: What's the goal behind having these sorted?

graphql/version_resolver.go
```diff
@@ -329,20 +329,11 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
...
-		Sorts:                          defaultSort,
-		IncludeBaseTasks:               false,
-		FieldsToProject:                []string{task.DisplayStatusKey},
-		IncludeBuildVariantDisplayName: false,
-	}
-	tasks, _, err := task.GetTasksByVersion(ctx, *obj.Id, opts)
+	statuses, err := task.GetTaskStatusesByVersion(ctx, *obj.Id)
 	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error getting version tasks: %s", err.Error()))
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("getting task statuses for version with id '%s': %s", *obj.Id, err.Error()))
```
ablack12, line 334: I was under the impression that we started UI errors with "Error" because they actually are the errors being propagated to the user. Is this true? (We may want to shore up https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit regardless)

graphql/version_resolver.go
```diff
@@ -329,20 +329,11 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
...
 	}
-	return getAllTaskStatuses(tasks), nil
+	return statuses, nil
```
ablack12, line 336: Out of curiosity, do we just want the unique statuses to be returned here? utility.UniqueStrings can do this

ablack12 (changes requested):

model/task/db.go
khelif96: It ensures they are returned in a consistent order otherwise the order changes every iteration which makes testing this impossible.

graphql/version_resolver.go
khelif96: Yes we do! I handled this in the aggregation in the `$addToSet` stage.

graphql/version_resolver.go
ablack12: This is still a question but otherwise LGTM

khelif96: We aren't really consistent with this technically all errors appear in an Error toast which indicates that it is an error. I opened EVG-20187 to consolidate this.

PR #6662, opened 2023-06-14
EVG-19659 Auth with Github app

bynn:
EVG-19659

### Description
when this was previously merged, it was too noisy so i wrapped things in a `token != ""`


Kimchelly (approved):

agent/command/git_merge_pr.go
```diff
@@ -72,6 +73,14 @@ func (c *gitMergePR) Execute(ctx context.Context, comm client.Communicator, logg
 	if token == "" {
 		token = conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion)
 	}
+	appToken := conf.Expansions.Get(evergreen.GithubAppToken)
+	if appToken == "" {
+		logger.Task().Debug(message.Fields{
```
ablack12, line 78: this doesn't log to splunk since it's in the agent so kind of noisy for users

agent/internal/testutil/task_config.go
```diff
@@ -17,7 +19,16 @@ func MakeTaskConfigFromModelData(ctx context.Context, settings *evergreen.Settin
 	if err != nil {
 		return nil, errors.Wrap(err, "getting global GitHub OAuth token")
 	}
-	exp, err := model.PopulateExpansions(data.Task, data.Host, oauthToken)
+	appToken, err := settings.CreateInstallationToken(ctx, data.Project.Owner, data.Project.Repo, nil)
+	if err != nil {
+		grip.Debug(message.WrapError(err, message.Fields{
```
ablack12, line 24: this function is only used in tests if that changes anything (might be helpful to include that in the comment above)

thirdparty/github.go
```diff
@@ -274,19 +298,49 @@ func GetGithubCommits(ctx context.Context, oauthToken, owner, repo, ref string,
 		options.Until = until
 	}
 
-	commits, resp, err := client.Repositories.ListCommits(ctx, owner, repo, &options)
-	if resp != nil {
-		defer resp.Body.Close()
+	installationToken, _ := GetInstallationToken(ctx, owner, repo, nil)
+	if installationToken != "" {
```
ablack12, line 302: Shouldn't we check the error here? Otherwise why are we even  returning an error?

thirdparty/github.go
```diff
@@ -357,42 +358,43 @@ func parseGithubErrorResponse(resp *github.Response) error {
...
+
+		file, _, resp, err := client.Repositories.GetContents(ctx, owner, repo, path, opt)
+		if resp != nil {
+			defer resp.Body.Close()
+			if err == nil && file != nil && file.Content != nil {
+				return file, nil
+			}
 		}
+
+		grip.Debug(message.WrapError(err, message.Fields{
```
ablack12, line 382: there's a chance we won't log this if we don't have an error but file or file content is nil; that being said, shouldn't we return the same error as below in that case? (like the call worked still)

thirdparty/github.go
```diff
@@ -476,34 +478,35 @@ func SendPendingStatusToGithub(input SendGithubStatusInput) error {
...
-
-	compare, resp, err := client.Repositories.CompareCommits(ctx, owner, repo, baseRevision, currentCommitHash, nil)
-	if resp != nil {
-		defer resp.Body.Close()
-		if err == nil && compare != nil && compare.GetMergeBaseCommit() != nil && compare.GetMergeBaseCommit().GetSHA() != "" {
-			return compare.GetMergeBaseCommit().GetSHA(), nil
+		compare, resp, err := client.Repositories.CompareCommits(ctx, owner, repo, baseRevision, currentCommitHash, nil)
+		if resp != nil {
+			defer resp.Body.Close()
+			if err == nil && compare != nil && compare.GetMergeBaseCommit() != nil && compare.GetMergeBaseCommit().GetSHA() != "" {
```
ablack12, line 489: same thought here

ablack12 (changes requested):

thirdparty/github.go
bynn: it seems a little weird to check the comp before checking nil but there shouldn't be a problem compare is always nil if there's an error and this is very temporary

agent/command/git_merge_pr.go
bynn: removed

thirdparty/github.go
```diff
@@ -375,9 +391,14 @@ func GetGithubFile(ctx context.Context, token, owner, repo, path, ref string) (*
...
-	defer utility.PutHTTPClient(legacyClient)
-	client := github.NewClient(legacyClient)
+	ctx, httpLegacyClient, putLegacyClient := getGithubClient(ctx, token, "GetGithubFile", retryConfig{retry: true}, []attribute.KeyValue{
+		attribute.String(githubOwnerAttribute, owner),
+		attribute.String(githubRepoAttribute, repo),
+		attribute.String(githubRefAttribute, ref),
+		attribute.String(githubPathAttribute, path),
+	})
+	defer putLegacyClient()
+	client := github.NewClient(httpLegacyClient)
```
ablack12, line 401: Discussed below -- including `if newResp.StatusCode == http.StatusNotFound` up here should be good

ablack12 (approved):

thirdparty/github.go
bynn: added to the if statement above

PR #6663, opened 2023-06-14
EVG-20074 Post statuses for GitHub merge queue

bsamek:
EVG-20074

Description
This adds subscriptions for GitHub merge queue commits.

Testing
This validates that the subscriptions are added in the patch-intent-processor tests.

Documentation
There are no user-facing changes in this PR.

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -383,6 +383,9 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 	if patchDoc.IsGithubPRPatch() {
 		catcher.Wrap(j.createGitHubSubscriptions(patchDoc), "creating GitHub PR patch subscriptions")
 	}
+	if patchDoc.IsGithubMergePatch() {
+		catcher.Wrap(j.createGitHubMergeSubscription(patchDoc), "creating GitHub PR patch subscriptions")
+	}
 	if patchDoc.IsBackport() {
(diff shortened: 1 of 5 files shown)


bsamek:
evergreen patch

bsamek:
evergreen patch

bsamek:
Lint failed to generate in the `evergreen patch`-created patch because of how merge base is calculated. The PR linked in description has run lint.

units/patch_intent.go
```diff
@@ -485,6 +488,37 @@ func (j *patchIntentProcessor) createGitHubSubscriptions(p *patch.Patch) error {
...
+	input := thirdparty.SendGithubStatusInput{
+		VersionId: j.PatchID.Hex(),
+		Owner:     p.GithubMergeData.Org,
+		Repo:      p.GithubMergeData.Repo,
+		Ref:       p.GithubMergeData.HeadSHA,
+		Desc:      "patch created",
+		Caller:    j.Name,
+		Context:   "evergreen",
+	}
+	err := thirdparty.SendPendingStatusToGithub(input, j.env.Settings().Ui.Url)
```
malikchaya2, line 515: Can you please explain the `j.env.Settings().Ui.Url` a bit? What do we expect it to be? Why do we want it over uiConfig.Get(env).Url?

bsamek: They're the same value. `uiConfig.Get()` makes a database call. Since we already have the environment object, we don't need to make the call.

malikchaya2 (approved):

PR #6664, opened 2023-06-14
EVG-20138: explicitly set agent logging output

Kimchelly:
EVG-20138

### Description
Remove usage of the log prefix in a special case. It almost always refers to the file path prefix that the agent should use; however, in the case where it should log to stdout (e.g. for the smoke test), it used the special `STDOUT`/`LOCAL` to indicate the output. I split these two concepts so that the output source is independent of the log file prefix. There will be a follow-up PR to set these flags since it needs the next deploy to go out to ensure agents have the new CLI flag.

* Split log prefix from log output for the agent.
* Do not export the special logging output prefixes, which are now only used by the app server for its own logging.

### Testing
Tested in staging to ensure that a new host started up in staging still logs agent logs to its file as it did before. For the smoke test, I ran a smoke test and it's still is still logging to stdout.

We can't unit test this because the function uses a grip multi sender. The current implementation of grip global senders is such that you can never re-assign the global sender (i.e. use `grip.SetSender` again) after it's been assigned a multi sender. Therefore, testing the agent/monitor logging setup (i.e. assigning the multi sender as the global sender) will irrevocably break the grip logging for all other tests in the `operations` package. I kind of don't want to fix this because it's an oversight in the design of grip, and the piece of functionality is not that large, so there's no unit test.

### Documentation
N/A


Kimchelly:
Seemingly unrelated to the change, but I had to change the commit queue delete test because it seems like the delete item route or the test itself has a bug. I didn't really think it was investigating further since nobody's ever filed a ticket about the route not working.

operations/agent_monitor.go
```diff
@@ -118,6 +121,11 @@ func agentMonitor() cli.Command {
...
 				Usage: "the path to the shell for starting the agent",
 			},
+			cli.StringFlag{
+				Name:  logOutputFlagName,
+				Value: string(agent.LogOutputFile),
+				Usage: "location for the agent monitor's log output (file, stdout)",
+			},
 			cli.StringFlag{
 				Name:  logPrefixFlagName,
 				Value: monitorLoggerName,
```
malikchaya2, line 131: should usage here also be switched to `log output` instead of `log name`?

malikchaya2 (approved):
approved with nit

operations/agent_monitor.go
Kimchelly: Not sure if you're talking about the prefix flag or the output flag, but the flag value/usage fields seems appropriate for the log prefix.

Kimchelly: Oh I see, yeah I can make them say the same thing.

malikchaya2 (approved):

Kimchelly:
`version-client-version-update` is failing spuriously (this PR contains no changes to the user-facing CLI), so I filed EVG-20182.

PR #6668, opened 2023-06-15
EVG-20040 Limit data returned by /ref 

malikchaya2:
[EVG-20040](https://jira.mongodb.org/browse/EVG-20040)

### Description
Restricting this route to only people with project settings view permissions is very risky since it would affect cli commands used by a lot of users. I instead opted to limit the route to only the settings that are needed. I also created a separate route for the workstation config so that it's only fetched when needed. 

### Testing
Existing tests

### Documentation
Not needed, no downstream affects since this is a cli route and not a user facing route.

### DIFF
--- a/operations/host_spawn.go
+++ b/operations/host_spawn.go
@@ -399,7 +399,7 @@ func hostConfigure() cli.Command {
 				return errors.Wrap(err, "setting up legacy Evergreen client")
 			}
 
-			projectRef, err := ac.GetProjectRef(project)
+			projectRef, err := ac.GetProjectWithWorkstationConfig(project)
 			if err != nil {
 				return errors.Wrapf(err, "finding project '%s'", project)
 			}
--- a/operations/host_spawn.go
+++ b/operations/host_spawn.go
@@ -399,7 +399,7 @@ func hostConfigure() cli.Command {
 				return errors.Wrap(err, "setting up legacy Evergreen client")
 			}
 
-			projectRef, err := ac.GetProjectWithWorkstationConfig(project)
+			projectRef, err := ac.GetProjectRef(project)
 			if err != nil {
 				return errors.Wrapf(err, "finding project '%s'", project)
 			}
--- a/service/api.go
+++ b/service/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model"
 	"github.com/evergreen-ci/evergreen/model/host"
 	"github.com/evergreen-ci/evergreen/model/task"
+	restModel "github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/evergreen-ci/evergreen/rest/route"
 	"github.com/evergreen-ci/evergreen/validator"
 	"github.com/evergreen-ci/gimlet"
@@ -143,18 +144,18 @@ func (as *APIServer) fetchLimitedProjectRef(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	limitedRef := &model.ProjectRef{
-		Id:         p.Id,
-		Identifier: p.Identifier,
-		Owner:      p.Owner,
-		Repo:       p.Repo,
-		Branch:     p.Branch,
-		WorkstationConfig: model.WorkstationConfig{
-			GitClone:      p.WorkstationConfig.GitClone,
-			SetupCommands: p.WorkstationConfig.SetupCommands,
-		},
-		CommitQueue: model.CommitQueueParams{
-			Message: p.CommitQueue.Message,
+	wc := restModel.APIWorkstationConfig{}
+	wc.BuildFromService(p.WorkstationConfig)
+
+	limitedRef := restModel.APIProjectRef{
+		Id:                utility.ToStringPtr(p.Id),
+		Identifier:        utility.ToStringPtr(p.Identifier),
+		Owner:             utility.ToStringPtr(p.Owner),
+		Repo:              utility.ToStringPtr(p.Repo),
+		Branch:            utility.ToStringPtr(p.Branch),
+		WorkstationConfig: wc,
+		CommitQueue: restModel.APICommitQueueParams{
+			Message: utility.ToStringPtr(p.CommitQueue.Message),
 			Enabled: p.CommitQueue.Enabled,
 		},
 	}
--- a/service/api.go
+++ b/service/api.go
@@ -328,6 +328,7 @@ func (as *APIServer) GetSettings() evergreen.Settings {
 }
 
 // NewRouter returns the root router for all APIServer endpoints.
+// These routes are deprecated; any new functionality should be added to REST v2
 func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	requireProject := gimlet.WrapperMiddleware(as.requireProject)
 	requireUser := gimlet.NewRequireAuthHandler()
--- a/service/api.go
+++ b/service/api.go
@@ -130,19 +130,58 @@ func (as *APIServer) FetchTask(w http.ResponseWriter, r *http.Request) {
 	gimlet.WriteJSON(w, t)
 }
 
-// fetchProjectRef returns a project ref given the project identifier
-func (as *APIServer) fetchProjectRef(w http.ResponseWriter, r *http.Request) {
-	id := gimlet.GetVars(r)["identifier"]
-	projectRef, err := model.FindMergedProjectRef(id, "", true)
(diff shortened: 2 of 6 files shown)


service/api.go
```diff
@@ -322,7 +361,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.SimpleVersions = true
 
 	// Project lookup and validation routes
-	app.AddRoute("/ref/{identifier}").Wrap(requireUser).Handler(as.fetchProjectRef).Get()
+	app.AddRoute("/ref/{projectId}").Wrap(requireUser).Handler(as.fetchLimitedProjectRef).Get()
```
Kimchelly, line 364: Can we add a deprecation notice to this to warn devs not to extend our usage of this route any more than it's already used?

service/api.go
```diff
@@ -130,19 +130,58 @@ func (as *APIServer) FetchTask(w http.ResponseWriter, r *http.Request) {
 	gimlet.WriteJSON(w, t)
 }
 
-// fetchProjectRef returns a project ref given the project identifier
-func (as *APIServer) fetchProjectRef(w http.ResponseWriter, r *http.Request) {
-	id := gimlet.GetVars(r)["identifier"]
-	projectRef, err := model.FindMergedProjectRef(id, "", true)
+// fetchLimitedProjectRef returns a limited project ref given the project identifier
+func (as *APIServer) fetchLimitedProjectRef(w http.ResponseWriter, r *http.Request) {
```
Kimchelly, line 134: Can we add a basic unit test that this route redacts some of the other project ref information?

service/api.go
```diff
@@ -322,7 +361,8 @@ func (as *APIServer) GetServiceApp() *gimlet.APIApp {
 	app.SimpleVersions = true
 
 	// Project lookup and validation routes
-	app.AddRoute("/ref/{identifier}").Wrap(requireUser).Handler(as.fetchProjectRef).Get()
+	app.AddRoute("/ref/{projectId}").Wrap(requireUser).Handler(as.fetchLimitedProjectRef).Get()
+	app.AddRoute("/ref/{projectId}/workstation").Wrap(requireUser).Handler(as.projectWithWorkstationConfig).Get()
```
Kimchelly, line 365: Rather than add more routes to the long-deprecated legacy UI, I would just turn the workstation route into a REST v2 route in `rest/route/service.go` (and possibly the limited project ref route as well).

service/api.go
```diff
@@ -130,19 +130,58 @@ func (as *APIServer) FetchTask(w http.ResponseWriter, r *http.Request) {
...
+			Message: p.CommitQueue.Message,
+			Enabled: p.CommitQueue.Enabled,
+		},
+	}
+
+	gimlet.WriteJSON(w, limitedRef)
+}
+
+// projectWithWorkstationConfig returns a limited project ref with the workstation config given the project identifier
+func (as *APIServer) projectWithWorkstationConfig(w http.ResponseWriter, r *http.Request) {
```
Kimchelly, line 161: Same thing as above, a basic unit test of the functionality would be good here.

operations/host_spawn.go
```diff
@@ -399,7 +399,7 @@ func hostConfigure() cli.Command {
 				return errors.Wrap(err, "setting up legacy Evergreen client")
 			}
 
-			projectRef, err := ac.GetProjectRef(project)
+			projectRef, err := ac.GetProjectWithWorkstationConfig(project)
```
Kimchelly, line 402: It seems like changing this will still break existing CLIs without some transition period. If they don't upgrade their CLI immediately and they run `evergreen host configure`, it won't work anymore.

Are users supposed to have a grace period to upgrade, or are they expected to upgrade their CLIs immediately for the command to work? If they're expected to have a grace period, we have to leave the existing `/ref` route as-is and make new ones.

service/api.go
```diff
@@ -130,19 +130,58 @@ func (as *APIServer) FetchTask(w http.ResponseWriter, r *http.Request) {
...
+	limitedRef := &model.ProjectRef{
+		Id:         p.Id,
+		Identifier: p.Identifier,
+		Owner:      p.Owner,
+		Repo:       p.Repo,
+		Branch:     p.Branch,
+		WorkstationConfig: model.WorkstationConfig{
+			SetupCommands: p.WorkstationConfig.SetupCommands,
+		},
+	}
```
Kimchelly, line 182: It seems like the workstation project ref needs to [include the `GitClone` field](https://github.com/evergreen-ci/evergreen/blob/e8374cb27545c3e6cb180aa8bc5d796f1182b804/model/project_ref.go#L2600) as well?

Kimchelly (changes requested):

service/api.go
malikchaya2: What do you mean by that? This should only be used by our cli, I don't think it's documented anywhere.

Kimchelly: It's true that they're only used by the CLI, but most of the CLI has already switched over to using `setupRestCommunicator` to get a client for REST v2 to facilitate our move away from REST v1. The REST v1 API has been deprecated for years in favor of REST v2 (i.e. the routes in `/rest/route/service.go`), and the only reason we keep it around is due to lack of benefit from migrating existing routes to v2. However, since we're writing new routes here, it seems like they should follow the REST v2 style.

Kimchelly: Hmm, this is true, but I still think we should leave a deprecation notice. Since this still bypasses the permissions system (in a more limited way), warning devs that `/ref/{project_id}` is solely for the sake preserving existing functionality working seems alright, but we should still discourage devs from trying to extend the usage of legacy routes for future needs (e.g. like stuffing more CLI functionality into the route).

service/api.go
malikchaya2: done, added it up top

service/api.go
malikchaya2: There is no unit test or a good place to add one, and give that it's deprecated it doesn't feel worth going through the setup. I tested it at http://localhost:9090/api/ref/evergreen though and made sure it looks okay 
image

operations/host_spawn.go
malikchaya2: removed it and put it all in /ref

service/api.go
malikchaya2: added

service/api.go
malikchaya2: There is no unit test or a good place to add one, and give that it's deprecated it doesn't feel worth going through the setup. I tested it at http://localhost:9090/api/ref/evergreen though and made sure it looks okay 
image

service/api.go
malikchaya2: added one up top

service/api.go
Kimchelly: It seems like there are [some existing tests](https://github.com/evergreen-ci/evergreen/blob/main/service/api_patch_test.go) that you can adapt already in this package.

I think we ought to test this because it is a security/permissions concern and we ought to preserve the anti-functionality that this really doesn't return extra fields beyond what it returns today. It doesn't seem like we plan to delete this route ever, so we ought to test/verify that it returns no more than it needs today and won't try to do it in the future.

service/api.go
```diff
@@ -130,19 +131,36 @@ func (as *APIServer) FetchTask(w http.ResponseWriter, r *http.Request) {
 	gimlet.WriteJSON(w, t)
 }
 
-// fetchProjectRef returns a project ref given the project identifier
-func (as *APIServer) fetchProjectRef(w http.ResponseWriter, r *http.Request) {
-	id := gimlet.GetVars(r)["identifier"]
-	projectRef, err := model.FindMergedProjectRef(id, "", true)
+// fetchLimitedProjectRef returns a limited project ref given the project identifier
```
Kimchelly, line 134: I would also mention that we should not add any more data to this route so that people don't try to stuff more information into it in the future. Instead, people should use a REST v2 route.

Kimchelly (changes requested):

service/api.go
malikchaya2: I added it to line 331

service/api.go
malikchaya2: done

service/api.go
malikchaya2: done

service/api.go
malikchaya2: done

service/api.go
Kimchelly: Thanks!

Kimchelly (approved):

PR #6669, opened 2023-06-15
EVG-20136: propagate panic from runTask to caller

Kimchelly:
EVG-20136

### Description
* Return the panic error (if any) from `runTask`.
* Add helper to handle logging and returning the original error + the panic error more conveniently.

### Testing
Added unit test for panic safety.

### Documentation
N/A


agent/agent.go
```diff
@@ -499,16 +499,20 @@ func (a *Agent) startLogging(ctx context.Context, tc *taskContext) error {
...
 	// we want to have separate context trees for tasks and loggers, so
 	// when a task is canceled by a context, it can log its clean up.
 	tskCtx, tskCancel := context.WithCancel(ctx)
 	defer tskCancel()
 
-	var err error
 	defer func() {
-		err = recovery.HandlePanicWithError(recover(), err, "running task")
+		op := "running task"
+		pErr := recovery.HandlePanicWithError(recover(), nil, op)
```
Kimchelly, line 511: In case anyone is wondering why I didn't absorb the `recover` checking into `logPanic` (since it is duplicated code), it's because `recover()` only works when it's in the defer block where the panic happened, per [the spec](https://go.dev/ref/spec#Handling_panics):
>  The return value of recover is nil if any of the following conditions holds:
>  * panic's argument was nil
>  * the goroutine is not panicking;
>  * recover was not called directly by a deferred function.

malikchaya2 (approved):

PR #6671, opened 2023-06-15
EVG-19919  Fix the return value for /rest/v2/projects/{project_id}/versions when using start 

malikchaya2:
EVG-19919 

### Description
The route was returning everything alder than the "start" revision. This fixes it to return the start revision and anything more recent. 

### Testing
Fixed existing test and tested on staging 

### Documentation
Edited the docs to include "start" and removed some params that are part of the POST route but not the GET route.


ablack12 (changes requested):
Re the comment thread in EVG-19919 --are we certain this workflow isn't what's expected for users? I think it's correct right now that our start after returns versions that come after, because of how we paginate backwards: https://github.com/evergreen-ci/evergreen/blob/94d9e76fe8c3ff742603f9bbde62d06fe9a690d4/rest/route/project.go#L887 

We may need to introduce a new field to allow this behavior.

docs/API/REST-V2-Usage.md
```diff
@@ -1213,10 +1213,8 @@ start as query parameters to support legacy behavior).
...
-| end_time_str     | string | Optional. Timestamp to stop looking for applicable versions.                                                                                                                                                                                                                                                 |
+| start   | int    | Optional. The version order number to start at for pagination. It will return anything less than the start.                                                                                                                                                                                                                                              |
+| revisions_after     | int    | Optional. Will return the version with that revision number and anything with a greater revision number (and therefore more recent).                                                                                                                                                                                                                                                 |
```
ablack12, line 1217: start and revisions_after are different enough that I think this is pretty confusing, considering they do the same thing and both consider revisions, but yeah thinking of a name that complements start is hard. We could change the name of this to something else (but still support start since people are already using it, i.e. just assign start to our new variable name during "parse"), and then we could try to be more consistent with the names revision_start / revision_end used for modify (assuming they behave similarly).

Alternatively, the suggestion from the ticket to have a boolean flip directions for start might be clearest while maintaining backwards compatibility.

malikchaya2: (I realize there's a default version limit, I'm looking into that now)

malikchaya2: Please ignore the above. I'm looking into it some more and will comment when it's ready for re-review.

malikchaya2: Ready for re-review. I deprecated start and added `revision_start` and `revision_end`. I can't parse revision_start to start because one uses `lt` and one use `lte`.

model/version.go
```diff
@@ -565,9 +567,22 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 		match[bsonutil.GetDottedKeyName(VersionBuildVariantsKey, VersionBuildStatusVariantKey)] = opts.ByBuildVariant
 	}
 
-	if opts.StartAfter > 0 {
-		match[VersionRevisionOrderNumberKey] = bson.M{"$lt": opts.StartAfter}
+	// start is deprecated but maintained for backwards compatibility
+	if opts.Start > 0 {
```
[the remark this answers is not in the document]
bynn: it looks like cloud is the main user of this field 
https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen%22%20%20action%3Dcompleted%20%7C%20spath%20path%20%7C%20search%20path%3D%22/rest/v2/projects/*/versions%22%7C%20spath%20%22params.start%7B%7D%22%20%7C%20search%20%22params.start%7B%7D%22%3D*&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-24h@h&latest=now&sid=1687456263.2601575

bynn (approved):
it looks good i had 1 comment about ticketing and letting users know

model/version.go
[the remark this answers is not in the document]
malikchaya2: This might be too heavily used and keeping it may not have a big enough impact to be worth the work of off-boarding users.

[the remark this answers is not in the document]
bynn: then should we not deprecate this at all and keep this is a separate field?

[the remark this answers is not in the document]
malikchaya2: I think maybe we'd still want to shift people away adding any new usage.

docs/API/REST-V2-Usage.md
```diff
@@ -1213,10 +1213,9 @@ start as query parameters to support legacy behavior).
...
-| start_time_str   | string | Optional. Timestamp to start looking for applicable versions.                                                                                                                                                                                                                                                |
-| end_time_str     | string | Optional. Timestamp to stop looking for applicable versions.                                                                                                                                                                                                                                                 |
+| start   | int    | start is deprecated. It was previously used as the version order number to start at for pagination. Use revision_start instead but keep in mind that it uses lte and not lt.                                                                                                                                                                                                                                                 |
```
ablack12, line 1216: We can't mark this as deprecated if it'll be used for pagination and the others cannot, since we aren't suggesting that users stop using pagination for this route. That being said, do we need to give revision start? The user is just asking for revision_end right?

ablack12 (changes requested):

rest/route/project.go
```diff
@@ -912,11 +923,15 @@ func (h *getProjectVersionsHandler) Run(ctx context.Context) gimlet.Responder {
 	}
 
 	if len(versions) >= h.opts.Limit {
+		key := "start"
+		if h.opts.RevisionStart != 0 {
+			key = "revision_start"
```
ablack12, line 928: sorry why are we changing this? Didn't you say above that revision_start isn't being used for pagination?

ablack12 (changes requested):

docs/API/REST-V2-Usage.md
malikchaya2: I switched it to use revision_start for pagination and not allow combining revision_start or revision_end with start. I
could just add revision_end to avoid over engineering.

rest/route/project.go
malikchaya2: I edited the comment above. I switched it to deprecate start and use revision_start for pagination instead when it's provided. But I'll remove revision_start based on the comment above and only leave revision_end.

docs/API/REST-V2-Usage.md
malikchaya2: (I realize I forgot to update the documentation)

rest/route/project.go
malikchaya2: done

model/version.go
```diff
@@ -565,9 +566,17 @@ func GetVersionsWithOptions(projectName string, opts GetVersionsOptions) ([]Vers
 		match[bsonutil.GetDottedKeyName(VersionBuildVariantsKey, VersionBuildStatusVariantKey)] = opts.ByBuildVariant
 	}
 
-	if opts.StartAfter > 0 {
-		match[VersionRevisionOrderNumberKey] = bson.M{"$lt": opts.StartAfter}
+	revisionFilter := bson.M{}
+	if opts.Start > 0 {
+		revisionFilter["$lt"] = opts.Start
+		match[VersionRevisionOrderNumberKey] = revisionFilter
```
ablack12, line 572: Since we aren't using revisionFilter outside of the if statements or sharing logic, it seems simpler to read to just assign this to {"$lt": opts.Start} and {"$gte": opts.RevisionEnd} directly

ablack12 (changes requested):
LGTM with simplification

model/version.go
malikchaya2: I set it up this way because if both are defined, we'd first set `match[VersionRevisionOrderNumberKey] = bson.M{"$lt": opts.Start}` and then end up overwriting it with `VersionRevisionOrderNumberKey with bson.M{"$gte": opts.RevisionEnd}`.

ablack12: Oh I see. In that case can we move `match[VersionRevisionOrderNumberKey] = revisionFilter` out of the if statements since it's duplicate code, and this will make it clearer that both parameters are meant to be settable

ablack12 (changes requested):

model/version.go
malikchaya2: What I was think is that if neither are set, which can happen since they are both optional, it'll end up setting it to `VersionRevisionOrderNumberKey : {}`.

ablack12: Good point, thanks!

ablack12 (approved):

PR #6672, opened 2023-06-15
EVG-20161 Cache merge SHA for commit queue patches

hadjri:
EVG-20161

### Description

This is an effort to reduce the amount of GitHub API usage in light of the recent rate limit throttling we've been receiving.  There are some places where we can't meaningfully reduce the amount of usage without a more significant uplift (such as caching files to reduce our usage of `GetGithubFile`). 

However, we can make a meaningful reduction in the amount of calls that commit queue tasks make requesting the merge commit SHA of their PR. These calls are usually made 300-500 times per hour, which represents 8-10% of our usage limit, so mitigating this should be significant and helpful.

Currently all commit queue tasks ask for the same merge commit SHA which is redundant. Merge patches [already store](https://gist.github.com/hadjri/9cc805a7da5807c4a53760c1ab4489d6#file-patch-go-L1176) the merge commit sha upon their creation, and it gets [pre-validated](https://gist.github.com/hadjri/b70ad2940d18040328e606a11ea58441#file-github-go-L1108) to ensure it is mergeable when it gets added to the commit queue or else it doesn't get created at all.

Merge patches (AFAICT) cannot get reused or rescheduled once they are dequeued. Since we dequeue patches once a new commit comes in on the PR, I think it's safe to just statically reference the merge commit SHA from when the patch was created, since although a PR merge commit SHA can change (if new commits come in), it would cause the merge patch to be dequeued and discarded in favor of the new merge patch.

What In think we can actually do is get rid of the agent's `/pull_request` endpoint altogether since the task config [stores](https://gist.github.com/hadjri/5a52e3f1f04db4982d650b384f585aa1#file-task_config-go-L98) the patch data on it already, but I wanted this to be an easy thing to revert if needed without having to worry about agent rollover issues.

### Testing
Added some unit tests, and testing on staging. Ran a patch [with](https://spruce-staging.corp.mongodb.com/version/648b94b2b23736b25a588b93/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [without](https://spruce-staging.corp.mongodb.com/version/648b97a7b23736b5717e15d7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) the change, and confirmed that the correct merge SHA was used for both patches, and that the patch with the change didn't make any Github requests in Splunk.


rest/route/agent.go
```diff
@@ -153,6 +153,35 @@ func (h *agentCheckGetPullRequestHandler) Parse(ctx context.Context, r *http.Req
...
+	}
+	// This field is only set for PR merge patches.
+	if p.GithubPatchData.MergeCommitSHA != "" {
+		return gimlet.NewJSONResponse(apimodels.PullRequestInfo{
+			// Validation elsewhere in GetMergeablePullRequest ensures that all merge patches are mergeable.
+			Mergeable:      utility.TruePtr(),
+			MergeCommitSHA: p.GithubPatchData.MergeCommitSHA,
+		})
+	}
+
```
malikchaya2, line 184: random comment placement. 
for this: 

> What In think we can actually do is get rid of the agent's /pull_request endpoint altogether since the task config [stores](https://gist.github.com/hadjri/5a52e3f1f04db4982d650b384f585aa1#file-task_config-go-L98) the patch data on it already, but I wanted this to be an easy thing to revert if needed without having to worry about agent rollover issues.

Do we maybe want to add some logging to it and then create a ticket to check a few days later if it logged since it was deployed and if not remove it?

malikchaya2 (changes requested):

rest/route/agent.go
```diff
@@ -153,6 +153,35 @@ func (h *agentCheckGetPullRequestHandler) Parse(ctx context.Context, r *http.Req
 }
 
 func (h *agentCheckGetPullRequestHandler) Run(ctx context.Context) gimlet.Responder {
```
ablack12, line 155: We have MergeCommitSHA in the agent already.
https://github.com/evergreen-ci/evergreen/blob/5f7ff6c5f3eda47b655e9e5329c662d581e5000a/agent/command/git.go#L310
I see not removing this route yet in case of rollback or etc, but can we just remove the call to it on the agent-side and use MergeCommitSHA directly? Can remove a bit of this logic too.

ablack12 (changes requested):

rest/route/agent.go
hadjri: Yes, created a ticket for that: https://jira.mongodb.org/browse/EVG-20196

ablack12 (approved):
This looks good to me (discussed using /pull_request completed logging to verify if the fallback logic is being used or not)

PR #6674, opened 2023-06-16
EVG-20168 Update dependent build and version after restarting a task

bynn:
EVG-20168

### Description
Build was setting all_tasks_blocked but never unsetting it when the dependent tasks were restarted 
That was causing patches to mistakenly believe they were complete and unschedule tasks 

### Testing
created a patch on staging and saw that the build now updates when you restart tasks that are dependent 
even if they were previously blocked 

https://spruce-staging.corp.mongodb.com/version/648cae22b2373620e2ba9d00/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -809,6 +809,7 @@ func UpdateUnblockedDependencies(t *task.Task) error {
 		return errors.Wrap(err, "getting dependencies marked unattainable")
 	}
 
+	buildsToUpdate := make(map[string]bool)
 	for _, blockedTask := range blockedTasks {
 		if err = blockedTask.MarkUnattainableDependency(t.Id, false); err != nil {
 			return errors.Wrap(err, "marking dependency attainable")
@@ -818,10 +819,15 @@ func UpdateUnblockedDependencies(t *task.Task) error {
 			return errors.WithStack(err)
 		}
 
-		if err := UpdateBuildAndVersionStatusForTask(&blockedTask); err != nil {
-			return errors.Wrapf(err, "updating build and version status for task '%s'", blockedTask.Id)
-		}
+		buildsToUpdate[blockedTask.BuildId] = true
+	}
 
+	var buildIds []string
+	for buildId := range buildsToUpdate {
+		buildIds = append(buildIds, buildId)
+	}
+	if err := UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
+		return errors.Wrapf(err, "updating version and patch status for builds")
 	}
 
 	return nil
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -827,7 +827,7 @@ func UpdateUnblockedDependencies(t *task.Task) error {
 		buildIds = append(buildIds, buildId)
 	}
 	if err := UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
-		return errors.Wrapf(err, "updating version and patch status for builds")
+		return errors.Wrapf(err, "updating build, version, and patch statuses")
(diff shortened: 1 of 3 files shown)


model/task_lifecycle.go
```diff
@@ -817,6 +817,11 @@ func UpdateUnblockedDependencies(t *task.Task) error {
 		if err := UpdateUnblockedDependencies(&blockedTask); err != nil {
 			return errors.WithStack(err)
 		}
+
+		if err := UpdateBuildAndVersionStatusForTask(&blockedTask); err != nil {
```
Kimchelly, line 821: Unit tests?

Kimchelly, line 821: Rather than updating the build/version for every single iteration, can this do the same status update logic as in `UpdateBlockedDependencies` and collect a list of build IDs to update at the end? It would avoid having to update the statuses repeatedly for the same builds.

Kimchelly (changes requested):

model/task_lifecycle.go
bynn: sorry i asked for a review before i pushed the test but they are here now!

model/task_lifecycle.go
```diff
@@ -817,6 +818,16 @@ func UpdateUnblockedDependencies(t *task.Task) error {
 		if err := UpdateUnblockedDependencies(&blockedTask); err != nil {
 			return errors.WithStack(err)
 		}
+
+		buildsToUpdate[blockedTask.BuildId] = true
+	}
+
+	var buildIds []string
```
Kimchelly, line 825: It makes my eye twitch slightly that it's `buildID` above but `buildId` here, but I can get over it.

model/task_lifecycle.go
```diff
@@ -817,6 +818,16 @@ func UpdateUnblockedDependencies(t *task.Task) error {
...
+	for buildId := range buildsToUpdate {
+		buildIds = append(buildIds, buildId)
+	}
+	if err := UpdateVersionAndPatchStatusForBuilds(buildIds); err != nil {
+		return errors.Wrapf(err, "updating version and patch status for builds")
```
Kimchelly, line 830: This is updating all the build and higher statuses (build, version, and patch), so I would include the mention of build in the error message so it's a little more obvious.

model/task_lifecycle_test.go
```diff
@@ -6316,6 +6343,14 @@ func TestUpdateUnblockedDependencies(t *testing.T) {
 	dbTask5, err := task.FindOneId(tasks[5].Id)
 	assert.NoError(err)
 	assert.True(dbTask5.DependsOn[0].Unattainable)
+
+	// Restarting a dependent task should unblock builds
```
Kimchelly, line 6347: nit: this isn't really testing restarting so much as unit testing one specific piece of functionality that's runs as part of restarting. I would just phrase this as something like `Unblocking a dependent task should unblock its build` so it's not confusing.

Kimchelly (approved):
LGTM mod small suggestions.

model/task_lifecycle_test.go
bynn: done

model/task_lifecycle.go

model/task_lifecycle.go
bynn: haha i didnt even realize that you wrote the same logic up there 
maybe we can try to put it in the style guide so we all write it in the same way! 
my preference is Id because all capital makes me feel like it should stand for something

bynn: ahh ill update it to match it then!

PR #6677, opened 2023-06-16
EVG-17142 GitHub conditional requests

ybrill:
[EVG-17142](https://jira.mongodb.org/browse/EVG-17142)

### Description
Use [conditional requests](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#conditional-requests) for calls to the GitHub API. They don't count against the rate limit if we don't have to get the actual response. 

The SDK suggests using [httpcache](https://github.com/gregjones/httpcache) for the cache. If we'd use the default, in-memory, cache it would be specific to each app server. Instead, this PR uses a collection in the db so it can be shared between app servers. My plan is to make the collection with a tight TTL index on when a key was last inserted. If the cache proves useful it will be straightforward to switch it out for a redis cluster in Kanopy.

The cache implementation swallows errors (with logging, but still). This is okay because worst case the cache will have outdated stuff in it and we'll just make more calls to GitHub when we see it's no longer valid.

### Testing
In staging, the first request made a call to GitHub, but subsequent requests did not. A query like [this one](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/nL1dyyQFByn) will tell us how often the response is coming from the cache.


ybrill:
cc @bynn there were many merge conflicts with #6662. I think I've addressed them all, but if you want to make sure I've captured your intent.

ybrill:
The build will pass (hopefully) once https://github.com/evergreen-ci/utility/pull/53 merges.

thirdparty/github.go
```diff
@@ -215,44 +256,34 @@ func githubShouldRetry(caller string, config retryConfig) utility.HTTPRetryFunct
 	}
 }
 
-// getGithubClient returns an *http.Client that will retry according to supplied retryConfig.
-// Also creates a span tracking the lifespan of the client.
-// Defer the returned function to release the client to the pool and close the span.
-func getGithubClient(ctx context.Context, token, caller string, config retryConfig, attributes []attribute.KeyValue) (context.Context, *http.Client, func()) {
+func getGithubClient(token, caller string, config retryConfig) *github.Client {
```
malikchaya2, line 263: nit: do we maybe want to add a comment saying that the client will use caching?

thirdparty/github.go
```diff
@@ -352,65 +362,38 @@ func parseGithubErrorResponse(resp *github.Response) error {
...
 
-	installationToken, err := GetInstallationToken(ctx, owner, repo, nil)
-	if err != nil {
-		ctx, httpClient, putClient := getGithubClient(ctx, installationToken, "GetGithubFile", retryConfig{retry: true}, []attribute.KeyValue{
-			attribute.String(githubOwnerAttribute, owner),
-			attribute.String(githubRepoAttribute, repo),
-			attribute.String(githubRefAttribute, ref),
-			attribute.String(githubPathAttribute, path),
-		})
-		defer putClient()
```
malikchaya2, line 370: @bynn this reworks the github client code so you may want to review it to see how it impacts EVG-19966.

thirdparty/github.go
```diff
@@ -891,58 +767,31 @@ func GithubAuthenticate(ctx context.Context, code, clientId, clientSecret string
...
 	if err != nil {
 		return "", errors.Wrap(err, "failed to get tag information from GitHub")
 	}
-	if newResp == nil {
+	if resp == nil {
 		return "", errors.New("invalid github response")
 	}
 
-	respBody, err := io.ReadAll(newResp.Body)
+	respBody, err := io.ReadAll(resp.Body)
```
malikchaya2, line 798: I see the old code looping through a list of tags while this one seems to parse only one tag. However, the url seems to be the same. Where is the discrepancy from?

thirdparty/github.go
```diff
@@ -1078,30 +928,19 @@ func GetGithubTokenUser(ctx context.Context, token string, requiredOrg string) (
...
-			return int64(limits.Core.Remaining), nil
-		}
+	githubClient := getGithubClient(getInstallationTokenWithoutOwnerRepo(ctx, token), caller, retryConfig{retry: true})
 
-		grip.Debug(message.WrapError(err, message.Fields{
-			"ticket":  "EVG-19966",
-			"message": "failed to check github limit",
-			"caller":  "CheckGithubAPILimit",
-		}))
+	limits, resp, err := githubClient.RateLimits(ctx)
```
malikchaya2, line 943: Will this pull the value from the cache and give us inaccurate results?

thirdparty/github.go
```diff
@@ -1260,39 +1044,21 @@ func AppAuthorizedForOrg(ctx context.Context, token, requiredOrganization, name
...
-	})
-	defer putLegacyClient()
-	client := github.NewClient(legacyClient)
+	))
+	defer span.End()
+
+	githubClient := getGithubClient(getInstallationToken(ctx, token, owner, repo, nil), caller, retryConfig{retry404: true})
 
-	permissionLevel, _, err := client.Repositories.GetPermissionLevel(ctx, owner, repo, username)
+	permissionLevel, resp, err := githubClient.Repositories.GetPermissionLevel(ctx, owner, repo, username)
```
malikchaya2, line 1061: Do we maybe not want to use the cached value here since using an outdated value here can have bigger repercussions?

thirdparty/github.go
```diff
@@ -1581,12 +1324,24 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
...
-
 	if settings.Api.GithubWebhookSecret == "" {
 		return nil, errors.New("Evergreen is not configured for GitHub Webhooks")
 	}
 
+	caller := "CreateGithubHook"
+	ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(
+		attribute.String(githubEndpointAttribute, caller),
+		attribute.String(githubOwnerAttribute, owner),
+		attribute.String(githubRepoAttribute, repo),
```
malikchaya2, line 1335: why are we using the legacy token here instead of using it only as a fallback?

thirdparty/github.go
```diff
@@ -1652,47 +1370,25 @@ func CreateGithubHook(ctx context.Context, settings evergreen.Settings, owner, r
...
-			"caller":  "GetExistingGithubHook",
-			"owner":   owner,
-			"repo":    repo,
-		}))
-	}
+	caller := "ListGithubHooks"
+	ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(
+		attribute.String(githubEndpointAttribute, caller),
+		attribute.String(githubOwnerAttribute, owner),
+		attribute.String(githubRepoAttribute, repo),
```
malikchaya2, line 1377: same qeustion as above

thirdparty/github.go
```diff
@@ -81,6 +85,47 @@ const (
...
+
+	// Wrap in a transport that creates a new span for each request that goes over the wire.
+	otelTransport := otelhttp.NewTransport(baseTransport)
+
+	// Wrap in a transport that caches responses to reduce the number of calls that count against our rate limit.
+	memoryCacheTransport := httpcache.NewTransport(&cache.DBCache{})
+	memoryCacheTransport.Transport = otelTransport
+
+	// Wrap in a transport that overrides the cache-control header so we don't rely on max-age
+	// from GitHub's response to not ask if there's been a change.
```
ablack12, line 115: "so we don't rely on a response to not ask" -- this sentence kind of confuses me 😅  why don't we want to rely on max age? why can't we set cachingTransport here directly as opposed to having this extra variable maxAgeTransport?

thirdparty/github.go
```diff
@@ -215,44 +260,34 @@ func githubShouldRetry(caller string, config retryConfig) utility.HTTPRetryFunct
...
+
+	return github.NewClient(client)
 }
 
-// GetInstallationToken creates an installation token using Github app auth.
-func GetInstallationToken(ctx context.Context, owner, repo string, opts *github.InstallationTokenOptions) (string, error) {
+// getInstallationToken creates an installation token using Github app auth.
+// If creating a token fails it will return the legacyToken.
+func getInstallationToken(ctx context.Context, legacyToken, owner, repo string, opts *github.InstallationTokenOptions) string {
+	trace.SpanFromContext(ctx).SetAttributes(attribute.Bool(githubAppTokenAttribute, false))
```
ablack12, line 286: can we do this after getting config?

thirdparty/github.go
```diff
@@ -12,9 +12,11 @@ import (
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/model/cache"
```
malikchaya2, line 15: Random comment placement: 
I had two general questions: 
How will the ttl index be set up, aka after what amount of time will the cache expire? We might want to tweak it until we get it right, so it might be a good idea to document somewhere what it's set to and how to change it.

Once we move over to github App, will we keep this caching or remove it?

malikchaya2 (changes requested):

thirdparty/github.go
ablack12: Oh or could you explain why we do false here and true later?

thirdparty/github.go
```diff
@@ -284,57 +335,20 @@ func GetGithubCommits(ctx context.Context, token, owner, repo, ref string, until
...
-	defer putLegacyClient()
-	client := github.NewClient(legacyClient)
-
-	commits, newResp, err := client.Repositories.ListCommits(ctx, owner, repo, &options)
-	if newResp != nil {
-		defer newResp.Body.Close()
+	commits, resp, err := githubClient.Repositories.ListCommits(ctx, owner, repo, &options)
+	if resp != nil {
+		defer resp.Body.Close()
+		span.SetAttributes(attribute.Bool(githubCachedAttribute, respFromCache(resp.Response)))
```
ablack12, line 341: what's this bit about?

ablack12 (changes requested):
@bynn can you take a look and verify that this refactor of the installation token looks good to you?

thirdparty/github.go
ybrill: done

thirdparty/github.go
ybrill: I'm not sure why we were doing it differently with the app than with the legacy token. To me it seems fraught to list all the tags on the repository since the beginning of time. Also, I think if we wanted to do that we'd need to use the options to get all the pages ([the docs](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-tags) say we'll get a max of 100 at a time).

@bynn was it intentional to have them use different endpoints?

thirdparty/github.go
ybrill: discussed above

thirdparty/github.go
ybrill: 😆 .
I was thinking like a week. Like probably most things won't be useful after that. I mean, it's not really a big deal if we keep things too long unless the storage becomes an issue. And if we kill things too fast the worst case is that we'll ask GH for it again.

thirdparty/github.go
ybrill: For some routes GitHub sets a max-age in the response to 60 seconds. This would mean that if we request the same thing from GitHub within 60 seconds the cache layer wouldn't even make a conditional request and instead just pull it out of the cache. This seems like a bad idea for our use case where things can change quickly. 

The idea of this extra transport is to override the max-age from the response by telling the cache that we want a max-age of 0. If the cache layer gets a request that sets max-age to something different than the response it will go with the request and make the conditional request according to the request.

thirdparty/github.go
ybrill: The idea here was to add an attribute to the span for the GitHub call so we could count the ratio of the which token got used by route. `SetAttributes` overwrites any previous attributes with the same key so this set it to false originally and only true if the app token is used.

thirdparty/github.go
ybrill: This is to let us query the ratio of requests that get their answer from the cache (so they don't count towards the rate limit) to the requests that need to get an original response from GitHub. This will help us to gauge how effective the cache is.

thirdparty/github.go
ybrill: It seems like there's not much of a downside to using the cache and if it saves requests then why remove it? 
I think the decision will hinge on whether we see it effectively reducing the number of calls that count against the rate limit.

malikchaya2: Are we not worried about the data becoming outdated?

thirdparty/github.go
malikchaya2: yes, I forgot that getInstallationToken was reconfigured to do that, my bad.

thirdparty/github.go
ablack12: Thanks for the info! maybe we could reword to "so we don't use Github's max-age, which prevents us from asking if there's been a change if we requested recently" or something like that to remove the double negative?

thirdparty/github.go
```diff
@@ -262,18 +297,34 @@ func GetInstallationToken(ctx context.Context, owner, repo string, opts *github.
 			"owner":   owner,
 			"repo":    repo,
 		}))
-		return "", errors.Wrap(err, "creating installation token")
+		return legacyToken
 	}
-	return token, nil
+	if token == "" {
```
ablack12, line 302: we may want to add a log here for EVG-19966

thirdparty/github.go
```diff
@@ -81,6 +85,45 @@ const (
...
+// RoundTrip sets the [max-age] Cache-Control directive before passing the request off
+// to the base transport.
+//
+// [max-age]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#max-age_2
+func (t *cacheControlTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("Cache-Control", "max-age=0")
+	return t.base.RoundTrip(req)
+}
+
+func init() {
```
ablack12, line 103: async discussed adding an ADR to explain briefly why we added this, and what different pieces of it do (the span github cache attribute, the different transporters)

model/cache/cache.go
```diff
@@ -0,0 +1,84 @@
...
+	grip.Error(message.WrapError(err, message.Fields{
+		"message":   "setting cached value",
+		"key":       key,
+		"operation": "Set",
+		"source":    "DBCache",
+	}))
+}
+
+// Delete removes the value associated with the key.
+func (c *DBCache) Delete(key string) {
```
ablack12, line 75: Do values get deleted from the cache when they're stale or etc?

thirdparty/github.go
malikchaya2: this is awesome! thanks for explaining it!

thirdparty/github.go
malikchaya2: That makes sense, thank you!

malikchaya2 (approved):
LGTM from my end but I think this should wait on a review from Bynn as well.

thirdparty/github.go
bynn: the "https://api.github.com/repos/%s/%s/git/refs/tags/%s" doesn't seem to exist on the client that we're using. 

It is technically the same route but without the "tags/{tag}" part at the end. It probably makes more sense to call it directly using the route since using the client will make us do extra work

thirdparty/github.go
ybrill: @bynn and I discussed logging. I pushed f1f9d70 to address his concerns, pending his review.

thirdparty/github.go
ybrill: done

thirdparty/github.go
ybrill: done

model/cache/cache.go
ybrill: Usually they don't. The TTL index will be deleting most of them

thirdparty/github.go
ybrill: I'd originally thought about switching to [this route](https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#get-a-reference) to get the SHA for the tag (because it's documented and [implemented by the SDK](https://github.com/google/go-github/blob/ae774d0751d40497e72d1ef2e3b98160adffd93e/github/git_refs.go#L50-L68)) but decided to not change unrelated things.

docs/decisions/2023-06-21_github_api_cache.md
```diff
@@ -0,0 +1,18 @@
...
+
+[Our SDK](https://github.com/google/go-github#conditional-requests) recommends the [httpcache package](https://github.com/gregjones/httpcache) which implements most of the lower-level details for us. It's used by instantiating its transport and including it in the transport chain the GitHub client's client uses. It provides a default in-memory cache but that would only cache on a per app-server basis and would be lost on restart. The package also provides a number of cache backends (redis, memcache, s3, etc.) and a [cache interface](https://github.com/gregjones/httpcache/blob/901d90724c7919163f472a9812253fb26761123d/httpcache.go#LL30C1-L39C2) for custom backends. We implemented a cache backend using a regular collection in the database as a quick-and-dirty solution for persisting the cache and letting the app servers share state. Although [GridFS](https://www.mongodb.com/docs/manual/core/gridfs/) might have made sense for this as it allows files greater than 16MB, its updates are not atomic so files might be partially overwritten.
```
ablack12, line 16: The greater than 16 MB document thing is actually an interesting question, since that could be possible in the future (albeit unlikely for an ungenerated file) -- if the thing doesn't exist in our cache, will we still get it from Github even if it says we should've cached it (or is that a future improvement / thing to investigate)?

model/cache/cache.go
ablack12: Will we be introducing a TTL for this collection or just for later? I hesitate to not TTL these since things like project yamls are huge and could make this collection pretty big pretty fast.

thirdparty/github.go
```diff
@@ -81,6 +84,45 @@ const (
...
+	baseTransport := utility.DefaultTransport()
+
+	// Wrap in a transport that creates a new span for each request that goes over the wire.
+	otelTransport := otelhttp.NewTransport(baseTransport)
+
+	// Wrap in a transport that caches responses to reduce the number of calls that count against our rate limit.
+	memoryCacheTransport := httpcache.NewTransport(&cache.DBCache{})
+	memoryCacheTransport.Transport = otelTransport
+
+	// Wrap in a transport that overrides the cache-control header so we don't so we don't use Github's
```
ablack12, line 113: duplicate "so we don't" haha

ablack12 (approved):
Left a couple of comments but overall looks good to me!

docs/decisions/2023-06-21_github_api_cache.md
ybrill: yes, if a thing is larger than the limit the Set function will fail and we won't cache it.
This seems okay atm assuming we'll more this to redis if it's working as expected.

thirdparty/github.go
ybrill: we _really_ won't
done

model/cache/cache.go
ybrill: before this merges I'll be creating the collection and making a TTL index on it.

bynn (approved):
Thanks for waiting! I tested all the changes and I think it should look fine 
There's a couple of errors in staging caused by the fact that the app is only installed on sandbox and not evergreen but that should go away after you merge main

PR #6678, opened 2023-06-16
EVG-20204 Temporarily remove token error to support multiple repos

bynn:
EVG-20204

### Description
got webhooks to work through github app and not the one on the repo 
had to remove an error from install tokens because it's only installed on commit queue and not evergreen repo but this is temporary 
using the same github webhook secret as the one we were already using for the current webhook

### Testing
staging
1. comment on pr, triggered new patch, retry etc 
2. pushing commit triggers repo tracker 
3. statuses are updated correctly


bynn:
evergreen patch

Kimchelly (approved):

bynn:
evergreen patch

PR #6681, opened 2023-06-20
EVG-20166: Mark GQL resolver files as not generated

SupaJoon:
[EVG-20166](https://jira.mongodb.com/browse/EVG-20166)

### Description
Deleting a field from a .graphql file and running `make gqlgen` programmatically deletes the resolver function from the resolver file. As a result, Github by default marks the files as generated and hides it in the diff. This code change prevents Github from marking GQL resolver files as generated in any case. [docs](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)


.gitattributes
```diff
@@ -1,3 +1,4 @@
 *.sh text eol=lf
 ./public/static/dist/* binary
 ./public/package-lock.json binary
+graphql/*_resolver.go linguist-generated=false
```
khelif96, line 4: EOL

khelif96 (approved):

PR #6684, opened 2023-06-20
EVG-20139, EVG-20141: standardize command display names and rename some functions

Kimchelly:
EVG-20139
EVG-20141

#### Current Command Format
In the UI: `'shell.exec' in "my-func-name" (#5) in "pre"`
Which has the format: `'COMMAND_NAME' in "FUNC_NAME" (#COMMAND_NUMBER) in "BLOCK_NAME"`

In the task logs: `'shell.exec' in function 'my-func-name' (step 5.12 of 8)`
Which has the format: `'COMMAND_NAME' in function 'FUNC_NAME' (step COMMAND_NUMBER.FUNC_SUBCOMMAND_NUM of TOTAL_COMMANDS_IN_BLOCK)`

Some problems with the current command format:
* The "failing command" and the task logs represent the failing command in different formats, making it difficult to deduce which task log lines correspond to the actual failing command.
* The UI display name doesn't mention what the context of the failing command is, so it's unclear if the func name is a function or if it's the command block section in which it failed.
* The UI display name doesn't include the function sub-command number that failed, but the task logs do.

#### Proposed Command Format
If a command doesn't already have an explicit display name configured by the user, the command format is as follows for both the UI and the task logs:
`'COMMAND_NAME' in function 'FUNC_NAME' (step COMMAND_NUMBER.FUNC_SUBCOMMAND_NUM of TOTAL_COMMANDS_IN_BLOCK) in block 'BLOCK_NAME'`

For example:
`'shell.exec' in function 'my-func-name' (step 5.12 of 8) in block 'pre'`

The biggest difference is that commands in the task logs now include the `in block 'BLOCK_NAME'` at the end, like the UI currently shows. I thought it might be kind of excessively verbose, but it is more explicit in the case where you're lost in the task logs on which specific block you're looking at. It also has the side benefit of you being able to search for the "failing command" exactly as-is in the task logs.

**I'm also open to alternative display formats if you think this one can be improved.**

#### Changes
* Make the default command display name consistent between the task logs and the UI.
* Rename a couple functions to better reflect what they actually do.

### Testing
Added unit tests for task log command name, existing tests should also pass.

Tested in staging that the commands displayed as expected in the task logs, and also appeared as expected in Spruce:
[Example of succeeding task](https://spruce-staging.corp.mongodb.com/task/evg_race_detector_test_util_fbcd2ecadeaf57797fb385a49740d32814b1989e_23_06_16_02_34_46/logs?execution=2)
[Example of failed task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_js_test_fbcd2ecadeaf57797fb385a49740d32814b1989e_23_06_16_02_34_46/logs?execution=1)

### Documentation
N/A

### DIFF
--- a/agent/command/registry.go
+++ b/agent/command/registry.go
@@ -65,8 +65,13 @@ func GetCommandFactory(name string) (CommandFactory, bool) {
 	return evgRegistry.getCommandFactory(name)
 }
 
-func Render(c model.PluginCommandConf, project *model.Project, block string) ([]Command, error) {
-	return evgRegistry.renderCommands(c, project, block)
+// Render takes a command specification and returns the commands to actually
+// run. It resolves the command specification into either a single command (in
+// the case of standalone command) or a list of commands (in the case of a
+// function).
+// kim: TODO: test display name defaulting.
+func Render(c model.PluginCommandConf, project *model.Project, blockInfo BlockInfo) ([]Command, error) {
+	return evgRegistry.renderCommands(c, project, blockInfo)
 }
 
 func RegisteredCommandNames() []string { return evgRegistry.registeredCommandNames() }
@@ -127,7 +132,7 @@ func (r *commandRegistry) getCommandFactory(name string) (CommandFactory, bool) 
 }
 
 func (r *commandRegistry) renderCommands(commandInfo model.PluginCommandConf,
-	project *model.Project, block string) ([]Command, error) {
+	project *model.Project, blockInfo BlockInfo) ([]Command, error) {
 
 	var (
 		parsed []model.PluginCommandConf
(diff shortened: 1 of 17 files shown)


agent/command/registry.go
```diff
@@ -200,3 +206,50 @@ func (r *commandRegistry) renderCommands(commandInfo model.PluginCommandConf,
...
+func GetDefaultDisplayName(commandName string, blockInfo BlockInfo, funcInfo FunctionInfo) string {
+	displayName := fmt.Sprintf("'%s'", commandName)
+	if funcInfo.Function != "" {
+		displayName = fmt.Sprintf("%s in function '%s'", displayName, funcInfo.Function)
+	}
+	if blockInfo.CmdNum > 0 && blockInfo.TotalCmds > 0 {
+		if funcInfo.SubCmdNum > 0 && funcInfo.TotalSubCmds > 1 {
+			// Include the function sub-command number only if the function runs
+			// more than one command.
+			displayName = fmt.Sprintf("%s (step %d.%d of %d)", displayName, blockInfo.CmdNum, funcInfo.SubCmdNum, blockInfo.TotalCmds)
```
malikchaya2, line 246: I wonder if the "step x of x" part is useful in the UI. Do we maybe want to only add that for the task logs? 

 I also noticed in the testing staging link that the display name is `'subprocess.exec' in function 'run-make' (step 3 of 3)` and doesn't include the block. Is it because it's not in a pre or post block? If that's the case I think it would be helpful to test one that does include the block so that we can see what that would look like.

malikchaya2 (changes requested):

malikchaya2 (approved):

PR #6687, opened 2023-06-21
EVG-19932: Add distro permissions directive

sophstad:
EVG-19932

### Description
- Add `@requireDistroAccess(DistroSettingsAccess)` directive to gate future distro resolvers.
- The technical design identified multiple directives that would be added for various permissions, but I remembered that we can use an argument to specify the permission level and use just one directive.
- I also noticed that we distinguish between create and admin permissions for distros, so I added a fourth permission level. I'll update the project tickets to reflect these changes.

### Testing
- Add unit tests for various permission levels with the directive

### Documentation
N/A

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -1142,6 +1142,7 @@ func GetPermissionLevelsForPermissionKey(permissionKey string) []PermissionLevel
 		}
 	case PermissionDistroSettings:
 		return []PermissionLevel{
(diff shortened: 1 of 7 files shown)


sophstad:
@ablack12 re-requesting because I tweaked some functionality around distro create permissions.

ablack12 (approved):
good catch!

PR #6688, opened 2023-06-21
EVG-18855 Clarify difference between Settings() and GetConfig()

bynn:
EVG-18855

### Description
adds function comment to clarify when each funciton should be used

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -86,8 +86,9 @@ func SetEnvironment(env Environment) {
 // Environment provides application-level services (e.g. databases,
 // configuration, queues.
 type Environment interface {
-	// Returns the settings object. The settings object is not
-	// necessarily safe for concurrent access.
+	// Returns the cached version of the settings object.
+	// The settings object is not necessarily safe for concurrent access.
+	// GetConfig should be used to access the settings object from the db.
 	Settings() *Settings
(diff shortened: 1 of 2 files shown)


environment.go
```diff
@@ -86,8 +86,9 @@ func SetEnvironment(env Environment) {
 // Environment provides application-level services (e.g. databases,
 // configuration, queues.
 type Environment interface {
-	// Returns the settings object. The settings object is not
-	// necessarily safe for concurrent access.
+	// Returns the cached version of the settings object.
```
Kimchelly, line 89: Per [go doc style](https://go.dev/blog/godoc), this should start as `Settings returns...`.

I would also recommend mentioning that it's the admin settings, since there are many settings in Evergreen (same with above).

environment.go
```diff
@@ -86,8 +86,9 @@ func SetEnvironment(env Environment) {
...
+	// The settings object is not necessarily safe for concurrent access.
+	// Should use GetConfig() to access the settings object from the db.
```
Kimchelly, line 91: nit: `Use GetConfig to access the settings object from the DB.`

Kimchelly (approved):
LGTM mod minor suggestions/style.

PR #6689, opened 2023-06-21
EVG-20103 Add default owner and repo for github app to admin settings

bynn:
EVG-20103

### Description
moving default owner and repo to admin settings 

### Testing
works on staging

### DIFF
--- a/config.go
+++ b/config.go
@@ -611,7 +611,7 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 func (s *Settings) CreateInstallationTokenWithDefaultOwnerRepo(ctx context.Context, opts *github.InstallationTokenOptions) (string, error) {
 	if s.AuthConfig.Github == nil || s.AuthConfig.Github.DefaultOwner == "" || s.AuthConfig.Github.DefaultRepo == "" {
 		// TODO EVG-19966: Return error here
-		return "", errors.New(fmt.Sprintf("default for '%s/%s' not found", s.AuthConfig.Github.DefaultOwner, s.AuthConfig.Github.DefaultRepo))
+		return "", nil
 	}
 	return s.CreateInstallationToken(ctx, s.AuthConfig.Github.DefaultOwner, s.AuthConfig.Github.DefaultRepo, opts)
 }
--- a/config.go
+++ b/config.go
@@ -609,7 +609,7 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 }
 
 // CreateInstallationTokenWithDefaultOwnerRepo returns an installation token when we do not care about
-// the owner/repo that we are calling the Github function with. (ie checking rate limit)
+// the owner/repo that we are calling the GitHub function with (i.e. checking rate limit).
 // It will use the default owner/repo specified in the admin settings and error if it's not set.
 func (s *Settings) CreateInstallationTokenWithDefaultOwnerRepo(ctx context.Context, opts *github.InstallationTokenOptions) (string, error) {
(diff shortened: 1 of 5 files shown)


config.go
```diff
@@ -608,6 +608,14 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 	return token.GetToken(), nil
 }
 
+func (s *Settings) CreateInstallationTokenWithDefaultOwnerRepo(ctx context.Context, opts *github.InstallationTokenOptions) (string, error) {
+	if s.AuthConfig.Github == nil || s.AuthConfig.Github.DefaultOwner == "" || s.AuthConfig.Github.DefaultRepo == "" {
+		// TODO EVG-19966: Return error here
```
ablack12, line 613: Can you explain this todo? You are returning an error

config.go
```diff
@@ -608,6 +608,14 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
...
+		return "", errors.New(fmt.Sprintf("default for '%s/%s' not found", s.AuthConfig.Github.DefaultOwner, s.AuthConfig.Github.DefaultRepo))
```
ablack12, line 614: errors.Errorf

thirdparty/github.go
```diff
@@ -267,8 +263,20 @@ func GetInstallationToken(ctx context.Context, owner, repo string, opts *github.
...
 }
 
-func getInstallationTokenWithoutOwnerRepo(ctx context.Context) (string, error) {
-	return GetInstallationToken(ctx, defaultOwner, defaultRepo, nil)
+func getInstallationTokenWithDefaultOwnerRepo(ctx context.Context, opts *github.InstallationTokenOptions) (string, error) {
+	settings, err := evergreen.GetConfig()
+	if err != nil {
+		return "", errors.Wrap(err, "getting evergreen settings")
+	}
+	token, err := settings.CreateInstallationTokenWithDefaultOwnerRepo(ctx, opts)
```
ablack12, line 271: Can you explain why we need a default owner/repo? Do we intend to keep this? What are we setting this to?

ablack12 (changes requested):

config.go
bynn: sorry that was there just so I could debug something but we shouldn't have it at the moment

config.go
bynn: removed!

thirdparty/github.go
Kimchelly: The motivation for this ticket comes from [the conversation described in this comment](https://github.com/evergreen-ci/evergreen/pull/6559#pullrequestreview-1455695124). Adding a comment to the function as to the motivation for when/where to use the default owner/repo func would be helpful.

bynn: In functions like CheckGithubAPILimit, we dont provide the owner and repo because we don't need it. 
But if we want to authenticate as app, we still need an owner and repo to get the installation token for. 
So this is in that case, we would just want to use the evergreen-ci/evergreen aka the default repo. 
We agreed that it should be configurable on admin settings instead of having it be hard coded like it was previously but I separated this work out of the last pr!

config.go
```diff
@@ -608,6 +608,17 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 	return token.GetToken(), nil
 }
 
+// CreateInstallationTokenWithDefaultOwnerRepo returns an installation token when we do not care about
+// the owner/repo that we are calling the Github function with. (ie checking rate limit)
```
ablack12, line 612: tiniest grammar nit: `GitHub function with (i.e. checking rate limit).` (i.e. is an acronym, period should come after parenthesis, capitalize Hub)

ablack12 (approved):

PR #6690, opened 2023-06-21
EVG-20161 Try patch doc's cached merge commit SHA upon the first fetch attempt

hadjri:
EVG-20161

### Description
This is a rework of the idea from #6672 

When the base branch has new commits that are not in the PR branch, Github can create a new merge commit for the merging of the PR. This means that the `merge_commit_sha` would be different from the last commit in the PR branch.

This is relevant for long commit queues where items get their `merge_commit_sha` cached on their patch doc upon queueing, but tasks won't get started until changes from the batch of size N ahead of them get merged into the base branch, at which point `merge_commit_sha` will have changed. This means that the [field](https://gist.github.com/hadjri/f3b3da7de0431c03b71a66abf378ac08#file-github-go-L100) we store on the patch to save the commit sha is subject to become invalid, and since it's a static & unchanging field we can't strictly rely on it at all times.

Change here is to do the same operation as the previous PR, but instead just check on the first git fetch attempt if the cached merge commit SHA is valid, and if not, retry the original way. Also introduced a change to keept the `MergeCommitSHA` field on the patch in the pull request route updated.
### Testing
Tested in staging, by committing something to main after a commit queue item had already been queued (but not started), and confirmed that it [failed on the first try, and succeeded](https://evergreen-staging.corp.mongodb.com/task_log_raw/sandbox_ubuntu2004_bynntask_patch_ed8e6708b5e7aceaf81620df54d85df28ef69132_64933698b23736cc672082ee_23_06_21_17_43_06/0?type=ALL#L361) on retry after pinging the agent route.

Tested another item without committing anything to main in between to make sure the cached merge commit SHA still works in the happy case [on the first try](https://evergreen-staging.corp.mongodb.com/task_log_raw/sandbox_ubuntu2004_bynntask_patch_ed8e6708b5e7aceaf81620df54d85df28ef69132_6493384e9dbe321cdf7025bb_23_06_21_17_50_12/0?type=ALL#L360) and the `/pull_request` endpoint doesn't get hit.

### DIFF
--- a/model/patch/patch.go
+++ b/model/patch/patch.go
@@ -531,18 +531,18 @@ func (p *Patch) AddSyncVariantsTasks(vts []VariantTasks) error {
 
 // UpdateMergeCommitSHA updates the merge commit SHA for the patch.
 func (p *Patch) UpdateMergeCommitSHA(sha string) error {
+	if p.GithubPatchData.MergeCommitSHA == sha {
+		return nil
+	}
 	shaKey := bsonutil.GetDottedKeyName(githubPatchDataKey, thirdparty.GithubPatchMergeCommitSHAKey)
-	if err := UpdateOne(
+	return UpdateOne(
 		bson.M{IdKey: p.Id},
 		bson.M{
 			"$set": bson.M{
 				shaKey: sha,
 			},
 		},
-	); err != nil {
-		return errors.WithStack(err)
-	}
-	return nil
+	)
 }
 
 func (p *Patch) FindModule(moduleName string) *ModulePatch {
--- a/model/patch/patch.go
+++ b/model/patch/patch.go
@@ -529,6 +529,22 @@ func (p *Patch) AddSyncVariantsTasks(vts []VariantTasks) error {
 	return nil
 }
 
+// UpdateMergeCommitSHA updates the merge commit SHA for the patch.
+func (p *Patch) UpdateMergeCommitSHA(sha string) error {
+	shaKey := bsonutil.GetDottedKeyName(githubPatchDataKey, thirdparty.GithubPatchMergeCommitSHAKey)
+	if err := UpdateOne(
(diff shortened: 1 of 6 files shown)


rest/route/agent.go
```diff
@@ -152,6 +152,26 @@ func (h *agentCheckGetPullRequestHandler) Parse(ctx context.Context, r *http.Req
...
+		return gimlet.MakeJSONErrorResponder(gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("task '%s' not found", h.taskID),
+		})
+	}
+	p, err := patch.FindOne(patch.ByVersion(t.Version))
+	if err != nil {
+		return gimlet.NewJSONInternalErrorResponse(errors.Wrapf(err, "getting patch for task '%s'", h.taskID))
+	}
+	if p == nil {
```
ablack12, line 169: Can we move getting the aptch until after we get the pull request, so we don't make an unnecessary DB call?

model/patch/patch.go
```diff
@@ -529,6 +529,22 @@ func (p *Patch) AddSyncVariantsTasks(vts []VariantTasks) error {
 	return nil
 }
 
+// UpdateMergeCommitSHA updates the merge commit SHA for the patch.
+func (p *Patch) UpdateMergeCommitSHA(sha string) error {
+	shaKey := bsonutil.GetDottedKeyName(githubPatchDataKey, thirdparty.GithubPatchMergeCommitSHAKey)
+	if err := UpdateOne(
```
ablack12, line 535: Presumably this should only update if the sha doesn't equal what the patch already has

ablack12, line 535: also we can just return this error directly (I don't think we really need the stack for this)

ablack12 (approved):
LGTM, just some small things

PR #6691, opened 2023-06-22
EVG-19933: Add distro resolver

sophstad:
EVG-19933

### Description
- Add resolver to fetch distro by ID. Use of the resolver requires the user to have distro view permissions.
- Spruce E2E tests are breaking because of these changes. A separate PR is open to fix them: https://github.com/evergreen-ci/spruce/pull/1930

### Testing
- Add GraphQL unit test

### Documentation
N/A


graphql/query_resolver.go
sophstad, line 1: Note that this file is _not_ autogenerated (not sure why https://github.com/evergreen-ci/evergreen/pull/6681 doesn't seem to be catching this)

sophstad:
evergreen retry

graphql/schema/types/distro.graphql
```diff
@@ -4,3 +4,113 @@ enum DistroSettingsAccess {
   EDIT
   VIEW
 }
+
+"""
+Distro models an environment configuration for a host.
+"""
+type Distro {
```
minnakt, line 11: Should we return an `Id` as well? I feel like the front-end GraphQL linter will complain about not having an id when we fetch the distro

sophstad: Interestingly name = ID at this point. The conversion happens here: https://github.com/evergreen-ci/evergreen/blob/0dc1abcfedaaa901c7641989063a2404f77e62d1/rest/model/distro.go#L370

minnakt: oh I see...! 🤔 I don't love it since it deviates from many of our other GraphQL objects, but if that is what's already in place 🤷‍♀️

minnakt (approved):

hadjri (approved):

PR #6692, opened 2023-06-22
EVG-20056: Update karma and peer dependencies to fix js-test failure

SupaJoon:
[EVG-20056](https://jira.mongodb.com/browse/EVG-20056)


khelif96 (approved):
Great find!

PR #6695, opened 2023-06-23
EVG-20059: Expose gitTags field for versions

minnakt:
EVG-20059

### Description
This PR adds `gitTags` to the GraphQL `Version` object. Please ignore the changes with `externalLinksForMetadata`, I just moved it to be in alphabetical order.

### Testing
- Added GraphQL test


graphql/version_resolver.go
```diff
@@ -149,6 +149,46 @@ func (r *versionResolver) ChildVersions(ctx context.Context, obj *restModel.APIV
...
+func (r *versionResolver) GitTags(ctx context.Context, obj *restModel.APIVersion) ([]*GitTag, error) {
+	v, err := model.VersionFindOneId(utility.FromStringPtr(obj.Id))
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding version with id `%s`: %s", *obj.Id, err.Error()))
+	}
+	if v == nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding version with id `%s`: %s", *obj.Id, "version not found"))
+	}
+	var gitTags []*GitTag
+	for _, tag := range v.GitTags {
```
malikchaya2, line 183: since GitTags is already part of the original version object and not in a different collection, why don't we get it from obj so that we don't need an additional db call?

malikchaya2 (changes requested):

graphql/version_resolver.go
minnakt: Sure I could do that, but I think it would require exposing the git tags via the rest route and I wasn't sure if that's desirable or not. I added a commit that represents this work!

malikchaya2: Since the end result is the data being exposed to the user along with the patch without any special restrictions, I don't personally see a downside to adding it to the restAPI.  

We may want to expand the version graphql tests to query for the tag field as well.

malikchaya2 (approved):

khelif96 (approved):

PR #6696, opened 2023-06-23
EVG-20306: improve error message for conflicting definitions in parser project

Kimchelly:
EVG-20306

### Description
* Mention the build variant name when there are multiple conflicting tasks defined under a build variant.
* Change formatting so it includes the field names in the error message when the deep equality check fails.

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -1116,7 +1116,7 @@ func evaluateBuildVariants(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluato
 					if old, ok := existing[t.Name]; ok {
 						if !reflect.DeepEqual(t, *old) {
 							evalErrs = append(evalErrs, errors.Errorf(
-								"conflicting definitions of added tasks '%s': %v != %v", t.Name, t, old))
+								"conflicting definitions of added tasks '%s': %#v != %#v", t.Name, t, old))
 						}
 					} else {
 						bv.Tasks = append(bv.Tasks, t)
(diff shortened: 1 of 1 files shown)


model/project_parser.go
```diff
@@ -1259,7 +1259,7 @@ func evaluateBVTasks(tse *taskSelectorEvaluator, tgse *tagSelectorEvaluator, vse
 				// it's already in the new list, so we check to make sure the status definitions match.
 				if !reflect.DeepEqual(t, old) {
 					evalErrs = append(evalErrs, errors.Errorf(
-						"conflicting definitions of build variant tasks '%s': %v != %v", name, t, old))
+						"conflicting definitions of task '%s' listed under build variant '%s': %#v != %#v", name, pbvt.Name, t, old))
```
malikchaya2, line 1262: %#v will also print the package name, I don't know if that's helpful here.

malikchaya2: it would be `model.parserTask{ fields }` here

Kimchelly: The package/struct name isn't helpful, but the other formatters are less than ideal in their own ways, so I picked the one with the most field-level user-readable information. The only other alternative is `%+v` which is less human-readable.

malikchaya2: It seems like %+v would yield {x:1 y:2} and  %#v which would yield model.parserTask{x:1, y:2}.

malikchaya2 (approved):

PR #6697, opened 2023-06-23
EVG-20314: improve smoke test string checks

Kimchelly:
EVG-20314

### Description
The smoke test is failing spuriously because it's checking for string output that is too ambiguous/short (i.e. someone made a commit that coincidentally included the word "first" in it, and the smoke test is wrongly thinking that's the relevant output of the smoke test). I made the strings much more unique so there's a much lower chance that the test coincidentally fails because it happened to output a conflicting string.

### Testing
I just changed some strings, and the smoke test can't pass because it's using the `agent.yml` committed to main. This will be substantially reworked in EVG-19590, but for now, there's no way to easily get it to pass. It seems easier to just commit this and check the (currently red) smoke test turns green.

### Documentation
N/A

### DIFF
--- a/operations/service_deploy_smoke_utils.go
+++ b/operations/service_deploy_smoke_utils.go
@@ -344,7 +344,7 @@ func checkTaskLogContent(body []byte, mode agent.Mode) error {
 			return errors.New("did not find setup_task in task logs")
 		}
 		const teardownTaskLog = "smoke test is running the teardown task"
-		if !strings.Contains(page, "teardown_task") {
+		if !strings.Contains(page, teardownTaskLog) {
 			return errors.New("did not find teardown_task in task logs")
 		}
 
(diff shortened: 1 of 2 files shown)


hadjri (approved):
Wow, what an odd way for the smoke test to break. Thanks for fixing/mitigating 🚀

operations/service_deploy_smoke_utils.go
```diff
@@ -325,38 +326,45 @@ func checkTaskLogContent(body []byte, mode agent.Mode) error {
...
 			}
 		}
 
 		// Validate that setup_task and teardown_task run for all tasks
-		if !strings.Contains(page, "setup_task") {
+		const setupTaskLog = "smoke test is running the setup task"
+		if !strings.Contains(page, setupTaskLog) {
 			return errors.New("did not find setup_task in task logs")
 		}
+		const teardownTaskLog = "smoke test is running the teardown task"
```
hadjri, line 346: Nit: need to reference this on the line below as it's unused

hadjri (changes requested):

operations/service_deploy_smoke_utils.go
Kimchelly: Good catch, fixed it.

hadjri (approved):

PR #6700, opened 2023-06-26
EVG-19965 Thread context to user patches page queries

khelif96:
EVG-19965

### Description
In order to aid in investigating the performance of the getTasksByVersion query I am threading context to the user patches page queries which fail on large patches in order to get traces on honeycomb.

### DIFF
--- a/graphql/project_resolver.go
+++ b/graphql/project_resolver.go
@@ -39,7 +39,7 @@ func (r *projectResolver) Patches(ctx context.Context, obj *restModel.APIProject
 		IncludeHidden:   false,
 	}
 
-	patches, count, err := patch.ByPatchNameStatusesCommitQueuePaginated(opts)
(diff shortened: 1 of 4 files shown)


khelif96:
Failing tests have been failing on the waterfall.

model/task/db.go
```diff
@@ -2203,7 +2203,7 @@ type GroupedTaskStatusCount struct {
 	StatusCounts []*StatusCount `bson:"status_counts"`
 }
 
-func GetTaskStatsByVersion(versionID string, opts GetTasksByVersionOptions) (*TaskStats, error) {
+func GetTaskStatsByVersion(ctx context.Context, versionID string, opts GetTasksByVersionOptions) (*TaskStats, error) {
```
ybrill, line 2206: It looks like this function is still using the wrapped Aggregation and not passing the context through to the driver.

khelif96: Good catch!

ybrill (approved):

PR #6701, opened 2023-06-26
EVG-18280: consolidate termination logic in cloud host ready job

Kimchelly:
EVG-18280

### Description
This consolidates some AWS special host status handling from `DescribeInstances`. If the host does not exist, **or** `DescribeInstances` returns an error because it cannot find any info about 1 or more of the requested host IDs (because the host has been terminated for a long time), Evergreen treats both cases as a host that failed to start up, and therefore must be terminated. I consolidated the common termination logic for those two cases in the cloud host ready job.

### Testing
Existing tests pass.

### Documentation
N/A


units/host_status.go
```diff
@@ -312,3 +300,21 @@ func (j *cloudHostReadyJob) logHostStatusMessage(h *host.Host, cloudStatus cloud
 		})
 	}
 }
+
+// prepareToTerminateHost handles a host that is an unrecoverable state by
+// marking it for termination and handling any other.
```
hadjri, line 305: did you mean to add to the end of this comment?

hadjri (approved):

units/host_status.go
Kimchelly: Yeah it should have more words at the end, I fixed it.

PR #6703, opened 2023-06-26
EVG-19998 fix the UnattainableDependency field

ybrill:
[EVG-19998](https://jira.mongodb.org/browse/EVG-19998)

### Description
#6570 added a field to cache whether any of a task's dependencies are unattainable. Two things were overlooked there:
1. The cached status was coming from the task struct in the app, which made it very susceptible to races where we got the tasks from the database [here](https://github.com/evergreen-ci/evergreen/blob/a32d74340f0d04a237512d02557339e43f9fa506/model/task_lifecycle.go#L807-L810). For example, task `a` has two dependencies, `b` and `c`, that have both been marked as unattainable because they both depend on `d` which failed. If `b` and `c` both become attainable at the same time because task `d` was restarted then if [UpdateUnblockedDependencies](https://github.com/evergreen-ci/evergreen/blob/a32d74340f0d04a237512d02557339e43f9fa506/model/task_lifecycle.go#L806) starts for `b` before it starts for `c` but the db write for `c` happens first then `b` will set `a` to have `UnattainableDependency: true`. The fix in this PR is to get the `UnattainableDependency` status from the db itself so it won't race (at least not more than the current recursive updates already race 😭). This is because the updates to the same document are atomic so `UnattainableDependency` will always be in sync with the dependencies in the array.
2. To avoid a migration #6570 sets the field when a task is restarted so even if it didn't have the `UnattainableDependency` field set before it'll get picked up by the scheduler. There was another case that was missed, namely if an old task is activated for the first time. This PR will set the field for that case as well.

### Testing
Existing tests cover this code. Before we vert [the scheduler query switchover](https://github.com/evergreen-ci/evergreen/commit/d3c4a6b697900bdf210b468867b3bb323aad2b4e) to use this new field we should do a diff of the tasks returned by each of the queries.


ybrill:
cc @ablack12 (if you have time to review it'd be awesome since you have the context)

model/task/db.go
```diff
@@ -1820,31 +1820,41 @@ func FindProjectForTask(taskID string) (string, error) {
...
-			IdKey: taskId,
-		},
-		bson.M{
-			"$set": bson.M{bsonutil.GetDottedKeyName(DependsOnKey, "$[elem]", DependencyUnattainableKey): unattainable},
+			IdKey: t.Id,
 		},
-		options.FindOneAndUpdate().SetArrayFilters(options.ArrayFilters{Filters: []interface{}{
-			bson.M{
-				bsonutil.GetDottedKeyName("elem", DependencyTaskIdKey): dependencyId,
+		[]bson.M{
```
ablack12, line 1832: Can we drop a comment to explain with words why we do this and what we're finding/updating? (What is $$this here?)

model/task/db.go
```diff
@@ -1820,31 +1820,41 @@ func FindProjectForTask(taskID string) (string, error) {
 	return t.Project, nil
 }
 
-func updateAllMatchingDependenciesForTask(taskId, dependencyId string, unattainable bool) error {
+func (t *Task) updateAllMatchingDependenciesForTask(dependencyId string, unattainable bool) error {
```
ablack12, line 1823: i see that tests cover MarkUnattainableDependency; is there any case we'd want to introduce for that or for this function that might catch more of what we're trying to fix here? (i.e. that would fail with the old function)

model/task/db.go
ybrill: done.
Renamed `this` (the default name) to `dependency` for clarity.

model/task/db.go
ybrill: done

model/task/db.go
```diff
@@ -1820,31 +1820,48 @@ func FindProjectForTask(taskID string) (string, error) {
...
 
-func updateAllMatchingDependenciesForTask(taskId, dependencyId string, unattainable bool) error {
+func (t *Task) updateAllMatchingDependenciesForTask(dependencyID string, unattainable bool) error {
 	env := evergreen.GetEnvironment()
 	ctx, cancel := env.Context()
 	defer cancel()
+
+	// Update the matching dependencies in the DependsOn array and the UnattainableDependency field that caches
+	// whether any of the dependencies are blocked. Combining both these updates in a single update operation makes it
+	// impervious to races because updates to single documents are atomic.
```
ablack12, line 1830: i love having "impervious" in our DB

ablack12 (approved):

PR #6705, opened 2023-06-27
EVG-20169: Prevent panics for missing operation context

minnakt:
EVG-20169

### Description
This PR is trying to reduce the panics in the ops alerts channel.

The error "missing operation context" seems to be coming from some call to [`GetOperationContext`](https://github.com/99designs/gqlgen/blob/master/graphql/context_operation.go#L61-L66). I'm not sure exactly what is calling this but to be safe I guarded all of our calls to `GetOperationContext` by checking `HasOperationContext` first.

I downgraded the panic to an error because it doesn't represent an error with the queries we've written, and doesn't seem to have an effect on the app or the UI. I mainly want to be able to see how often we see these errors in Splunk after these changes are applied.

### Testing
- Added a test case in `directive_test.go`


khelif96:
https://jira.mongodb.org/browse/EVG-20337 Introduced request number logging to honeycomb logs so it might be possible to get a better understanding of what is actually calling that to produce the panics.

hadjri (approved):

graphql/splunktracing.go
```diff
@@ -23,15 +23,16 @@ func (SplunkTracing) Validate(graphql.ExecutableSchema) error {
 }
 
 func (SplunkTracing) InterceptResponse(ctx context.Context, next graphql.ResponseHandler) *graphql.Response {
-	rc := graphql.GetOperationContext(ctx)
-	if rc == nil {
+	if hasOperationContext := graphql.HasOperationContext(ctx); !hasOperationContext {
 		// There was an invalid operation context, so we can't do anything
 		grip.Critical(message.Fields{
```
ablack12, line 28: We shouldn't log critical, because we still log this to the ops-alerts channel (if we do still want to catch this via ops alerts, better to make an alert so we can toggle it or update it without a code change if we need to)

graphql/resolver.go
```diff
@@ -96,6 +96,10 @@ func New(apiURL string) Config {
 		if !isStringMap {
 			return nil, ResourceNotFound.Send(ctx, "Project not specified")
 		}
+
+		if hasOperationContext := graphql.HasOperationContext(ctx); !hasOperationContext {
```
ablack12, line 100: This might not actually be enough because HasOperationContext checks if the value exists but it doesn't check that the value is nil (which would still panic) but this would still narrow it down. You could alternative add a panic handler 
https://github.com/evergreen-ci/evergreen/blob/f2381fe506626079db16089a11ab98bb7f690181/model/task/task.go#L1660

graphql/http_handler.go
```diff
@@ -38,6 +38,17 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 	srv.SetRecoverFunc(func(ctx context.Context, err interface{}) error {
 		queryPath := graphql.GetFieldContext(ctx).Path()
 
+		if queryPath == nil {
+			grip.Error(message.Fields{
```
ablack12, line 42: We're already logging critical right below (which I don't love re the info below but alas) and it already logs what queryPath is (if you want to really know if it's nil and not just empty you could add a `"query_is_nil": queryPath==nil` below)

ablack12 (changes requested):

minnakt:
> https://jira.mongodb.org/browse/EVG-20337 Introduced request number logging to honeycomb logs so it might be possible to get a better understanding of what is actually calling that to produce the panics.
> 
> There appears to be a user associated with the errors maybe you could try reaching out. The error claims something is wrong with the request body. @minnakt I wonder if it's possible some value or param would cause us to generate invalid JSON for requests.
> 
> https://mongodb.splunkcloud.com/en-US/app/search/search?earliest=-8h@h&latest=now&q=search%20index%3Devergreen%20request%3D894239%20host%3D%22evergreenapp-15.build.10gen.cc%22&display.page.search.mode=smart&dispatch.sample_ratio=1&workload_pool=standard_perf&display.page.search.tab=events&display.general.type=events&display.visualizations.charting.chart=line&sid=1687951051.3815629

I'm not sure, I feel like the user wouldn't be able to share any useful information given that this seems like a transient error. When I talked to Arjun (who was the first person who encountered this error), he said that he didn't see any errors on the UI. I also manually triggered this error multiple times (by scrolling really fast on the history page) and never saw any errors on the UI.

But I agree with you that it seems like there are two separate problems:
1. The `could not get json request body` error
2. The `missing operation context` panic which is probably caused by the lack of request body

minnakt:
I think I have more information about this panic now. Whenever I query the `/graphql/query` endpoint with malformed body e.g. `"{"`, the app will panic (you can try testing this on staging and check Splunk for the panics). This seems not ideal because if we ever officially support the GraphQL endpoint for users, that would mean we would panic whenever a user makes a malformed request.

I've looked at the stack again and the panic seems to come from the ApolloTracing extension. When I comment out the extension, the panics disappear. I think what happens is a user makes a malformed request, the operation context fails to get created, and then the ApolloTracing extension calls `GetOperationContext` which panics. Note that the ApolloTracing extension is not something we've written, it's provided by 99designs and so I don't believe it's directly editable. 

One thing I could do is make a wrapper extension around the ApolloTracing extension and make sure that we check the `OperationContext` exists before intercepting the response. I think I want to wait on this ticket and talk to a more senior engineer on UI before proceeding.

bynn:
evergreen retry

khelif96:
> I think I have more information about this panic now. Whenever I query the `/graphql/query` endpoint with malformed body e.g. `"{"`, the app will panic (you can try testing this on staging and check Splunk for the panics). This seems not ideal because if we ever officially support the GraphQL endpoint for users, that would mean we would panic whenever a user makes a malformed request.
> 
> I've looked at the stack again and the panic seems to come from the ApolloTracing extension. When I comment out the extension, the panics disappear. I think what happens is a user makes a malformed request, the operation context fails to get created, and then the ApolloTracing extension calls `GetOperationContext` which panics. Note that the ApolloTracing extension is not something we've written, it's provided by 99designs and so I don't believe it's directly editable.
> 
> One thing I could do is make a wrapper extension around the ApolloTracing extension and make sure that we check the `OperationContext` exists before intercepting the response. I think I want to wait on this ticket and talk to a more senior engineer on UI before proceeding.
> 
> About the deleted comment below: I just deleted Bynn's `evergreen retry` comment since he wanted to use my PR to test something!

 Great find! I think now that we have honeycomb traces we could remove this extension entirely. It was also [deprecated by the Apollo team ](https://github.com/apollographql/apollo-tracing#apollo-tracing) last year and they even recommend using OTEL instead.

khelif96 (changes requested):
Left a comment about removing the Apollo Tracing extension in favor of using the OTEL traces since that is where the error is originating.

graphql/resolver.go
minnakt: I think it might be okay! I tried testing it manually and it seemed fine, and in the corresponding unit tests I wrote I never defined the operation context but the tests didn't panic

graphql/http_handler.go
minnakt: removed!

graphql/resolver.go
khelif96: Could we remove this check? If the error was originating from apollo tracing this presumably is no longer needed. If we do still run into it its probably some other error.

graphql/splunktracing.go
```diff
@@ -23,15 +23,16 @@ func (SplunkTracing) Validate(graphql.ExecutableSchema) error {
 }
 
 func (SplunkTracing) InterceptResponse(ctx context.Context, next graphql.ResponseHandler) *graphql.Response {
-	rc := graphql.GetOperationContext(ctx)
-	if rc == nil {
+	if hasOperationContext := graphql.HasOperationContext(ctx); !hasOperationContext {
```
khelif96, line 26: Same comment.

graphql/resolver.go
minnakt: No, this will still occur if the request is malformed, because the operation context is not created

the only reason this was not reached was because the ApolloTracer was declared before the SplunkTracer

but if it's not important we can remove it

khelif96 (approved):

PR #6707, opened 2023-06-29
EVG-20097 Use non-cached admin settings for containers validation

hadjri:
EVG-20097

### Description
We have been modifying this field frequently and will probably continue to do so, changing so that we dynamically fetch the allowed images and don't require a prod bounce for the changes to take effect.

### Testing
Ran a container task in staging based off of a new image and the task went through and succeeded without error.

### DIFF
--- a/units/pod_allocator.go
+++ b/units/pod_allocator.go
@@ -202,13 +202,11 @@ func (j *podAllocatorJob) populate() error {
 	if j.env == nil {
 		j.env = evergreen.GetEnvironment()
 	}
-
-	// Use the latest service flags instead of those cached in the environment.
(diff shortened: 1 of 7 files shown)


units/pod_allocator.go
```diff
@@ -202,13 +202,11 @@ func (j *podAllocatorJob) populate() error {
 	if j.env == nil {
 		j.env = evergreen.GetEnvironment()
 	}
-
-	// Use the latest service flags instead of those cached in the environment.
```
malikchaya2, line 206: Can we please keep this comment and update it?

malikchaya2 (approved):
approved with nit

PR #6709, opened 2023-06-29
EVG-20355 Always allow HTTPS clone for modules

hadjri:
EVG-20355

### Description
Distros that use `legacy-ssh` are currently unable to perform a clone from an HTTPS url in their module section, because they do not pass a token to Github in the clone command. This change modifies that logic as well as removes the `GetRepoOwnerAndName` in favor of `thirdparty.ParseGitUrl` as they do the same thing.

### Testing
Tested before and after in staging, confirmed that a distro host with a legacy SSH default clone value was able to successfully HTTPS clone a private repository in `git.get_project`


agent/command/git.go
```diff
@@ -810,7 +809,7 @@ func (c *gitFetchProject) fetch(ctx context.Context,
 		if err := ctx.Err(); err != nil {
 			return errors.Wrapf(err, "canceled while applying module '%s'", moduleName)
 		}
-		err = c.fetchModuleSource(ctx, conf, logger, jpm, opts.method, opts.token, p, moduleName)
+		err = c.fetchModuleSource(ctx, conf, logger, jpm, opts.token, p, moduleName)
```
malikchaya2, line 812: It looks like these were already passing in opts.token. Where does the fix happen for 
"they do not pass a token to Github in the clone command"?

malikchaya2 (changes requested):

model/version.go
```diff
@@ -734,7 +734,10 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 			}
 		}
 		var sha, url string
-		owner, repo := module.GetRepoOwnerAndName()
+		owner, repo, err := thirdparty.ParseGitUrl(module.Repo)
+		if err != nil {
+			return nil, errors.Wrapf(err, "can't parse git url '%s'", module.Repo)
```
Kimchelly, line 739: nit: should be `"parsing git url '%s'` since that's what it was in the other diffed line.

Kimchelly (approved):
LGTM mod nit.

agent/command/git.go
hadjri: The fix is [here](https://github.com/evergreen-ci/evergreen/pull/6709/files#diff-1e8d12ae20a5922169a78aaae96aab11adb323b77baf8910c3cbfba80b1d0012R697), to make the clone method HTTPS if we encounter an HTTPS url. Beforehand if an HTTPS url is detected the default distro clone method would get set, which is often legacy SSH. This causes a problem because when the clone command is built, if the method is legacy SSH it will create a clone command that [does not use the token](fetchModuleSource)

malikchaya2 (approved):

PR #6712, opened 2023-06-30
EVG-16283 Add evergreen keep-definitions and evergreen reset-definitions

malikchaya2:
[EVG-16283](https://jira.mongodb.org/browse/EVG-16283)

### Description
This adds a PR-level option where users comment on the PR with evergreen keep-definitions to reuse the definitions in the most recent patch that ran on subsequent pushes. It also adds an option to reset to default with evergreen reset-definitions. 

### Testing
Added unit tests and [tested on staging](https://github.com/evergreen-ci/commit-queue-sandbox/pull/545). 
image

### DIFF
--- a/model/patch/db.go
+++ b/model/patch/db.go
@@ -250,6 +250,13 @@ func MostRecentPatchByUserAndProject(user, project string) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(1)
 }
 
+// MostRecentPatchByPR returns the latest patch for a pr number
+func MostRecentPatchByPR(prNumber int) db.Q {
+	return db.Query(bson.M{
+		bsonutil.GetDottedKeyName(githubPatchDataKey, thirdparty.GithubPatchPRNumberKey): prNumber,
+	}).Sort([]string{"-" + CreateTimeKey}).Limit(1)
+}
+
 // ByVersion produces a query that returns the patch for a given version.
 func ByVersion(version string) db.Q {
 	return db.Query(bson.M{VersionKey: version})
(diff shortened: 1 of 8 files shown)


model/patch/db.go
```diff
@@ -250,6 +250,13 @@ func MostRecentPatchByUserAndProject(user, project string) db.Q {
 	}).Sort([]string{"-" + CreateTimeKey}).Limit(1)
 }
 
+// MostRecentPatchByPR returns the latest patch for a pr number
+func MostRecentPatchByPR(prNumber int) db.Q {
+	return db.Query(bson.M{
```
hadjri, line 255: This should probably also check for the project name, because we can have conflicting PR numbers

model/patch/github.go
```diff
@@ -155,24 +158,39 @@ func NewGithubIntent(msgDeliveryID, patchOwner, calledBy string, pr *github.Pull
...
+		PushedAt:      pr.Head.Repo.PushedAt.Time.UTC(),
+		CalledBy:      calledBy,
+		RepeatPatchId: repeat,
 	}, nil
 }
 
+// getRepeatPatchId returns the patch id to repeat the definitions from
+// this information is found on the most recent pr patch
+func getRepeatPatchId(prNumber *int) string {
+	p, err := FindOne(MostRecentPatchByPR(*prNumber))
```
hadjri, line 186: should use `utility.FromInt` here

model/patch/github.go
```diff
@@ -155,24 +158,39 @@ func NewGithubIntent(msgDeliveryID, patchOwner, calledBy string, pr *github.Pull
...
+	if err != nil || p == nil {
```
hadjri, line 187: Seems we should still be aware of errors that happen here. Can we just return an empty string if `p == nil` and return the error otherwise?

hadjri (changes requested):

model/patch/github.go
malikchaya2: switched it to int

PR #6731, opened 2023-07-03
EVG-20078: Add mergeQueue field to project ref

minnakt:
EVG-20078

### Description
This PR adds the `mergeQueue` field to a project which specifies whether to use the Evergreen or GitHub merge queue. 

I decided to limit the scope of this PR to just adding the project field since it will be easier to review, and the GitHub route appears to require more changes than I expected. I think this will still unblock other work on the project though, since it seems to depend on the project field existing.

### Testing
- Added GraphQL test

### Documentation
- Updated CommitQueueParams


docs/API/REST-V2-Usage.md
```diff
@@ -1040,6 +1040,7 @@ exception of project variables, task annotation settings, workstation settings,
 |--------------|--------|-------------------------------------------|
 | enabled      | bool   | Enable/disable the commit queue           |
 | merge_method | string | method of merging (squash, merge, rebase) |
+| merge_queue  | string | merge queue to use (EVERGREEN or GITHUB ) |
```
bsamek, line 1043: I think lowercase follows the conventions better, e.g., "squash", from above.

bsamek (approved):
lgtm with a casing nit

malikchaya2 (approved):

docs/API/REST-V2-Usage.md
minnakt: Sadly this is consequence of how enums are defined in all caps in GraphQL, which trickles down into the API 😔 (the `merge_method` field above is defined as `String` rather than `Enum` which is why those fields are lowercase)

I confirmed with another UI engineer that the practice right now is to just keep the uppercase form in the API route. But I do agree it looks weird, so I'm going to spend some time thinking about if there's anything we can change about it in the future!

minnakt:
evergreen refresh

minnakt:
merging b/c failing tests are the same failing tests from the distro resolver PR which should now be fixed on Spruce

PR #6734, opened 2023-07-05
EVG-20372 Add len check to GetTaskStatusesByVersion

bynn:
EVG-20372

### Description
this was causing a panic so we added a len check


sophstad (approved):
LGTM, would probably be worth investigating the underlying cause of a task-less version but this is a good stopgap.

PR #6736, opened 2023-07-06
EVG-20015: Implement grip Sender for the Evergreen logger

julianedwards:
EVG-20015

### Description
A new Sender implementation for the Evergreen logging service.

Notable features:
- Injectable line parser: this makes the sender agnostic to the different log line formats it ingests—the parser converts the line format into the service representation of the a log line. We will support varying line formats for test logs written to the reserved directory.
- Injectable log writer: this makes the sender agnostic to the underlying logging service.

I also simplified the Evergreen logging service by removing the predefined prefixes, this will enable more flexibility and keep the service as simple as possible. Tasks will determine prefix structure beyond how the service distinguishes individual task runs in the actual data store. This will hold true for anything else storing logs in the future (like hosts).

### Testing
Unit tests.

### Documentation
N/A


model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+	"sync"
+	"time"
+
+	"github.com/mongodb/grip/level"
+	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/send"
+	"github.com/pkg/errors"
+)
+
+const defaultMaxBufferSize = 1e7 //nolint: unused
```
ybrill, line 15: It seems like this _is_ used later

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+}
+
+func (s *sender) flush(ctx context.Context) error {
+	if err := s.write(ctx, s.buffer); err != nil {
+		return errors.Wrap(err, "writing log")
+	}
+
+	s.buffer = []LogLine{}
```
ybrill, line 205: could it make sense to do something like
```go
s.buffer = s.buffer[:0]
```
to hold on to the allocated array?

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			s.opts.Local.Send(message.NewErrorMessage(level.Error, errors.Wrap(err, "parsing log line")))
+			return
+		}
+		if logLine.Priority == 0 {
+			logLine.Priority = m.Priority()
+		}
+		if !logLine.Priority.IsValid() {
+			s.opts.Local.Send(message.NewErrorMessage(level.Error, errors.Errorf("invalid log line priority %d", logLine.Priority)))
+			return
```
ybrill, line 119: could it make sense to `continue` here?

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		if logLine.Timestamp == 0 {
+			logLine.Timestamp = ts
+		}
+		if logLine.Timestamp < 0 {
+			s.opts.Local.Send(message.NewErrorMessage(level.Error, errors.Errorf("invalid log line timestamp %d", logLine.Timestamp)))
+			return
```
ybrill, line 126: could it make sense to `continue` here too?

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+
+		s.buffer = append(s.buffer, logLine)
+		s.bufferSize += len(line)
+		if s.bufferSize > s.opts.MaxBufferSize {
```
ybrill, line 131: maybe we'd want `>=`?

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+			return errors.Wrap(err, "flushing buffer")
+		}
+	}
+
+	return nil
+}
+
+func (s *sender) timedFlush() {
+	s.mu.Lock()
+	s.timer = time.NewTimer(s.opts.FlushInterval)
```
ybrill, line 179: Would a ticker work?

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+// LoggerOptions support the use and creation of an Evergreen log sender.
+type LoggerOptions struct {
+	// LogName is the identifying name of the log to use when persisting
+	// data.
+	LogName string
+	// Parse is the function for parsing raw log lines collected by the
+	// sender.
+	Parse LineParser
+	// Local is the sender for "fallback" operations and to collect the
+	// location of the logger output.
```
ybrill, line 31: What is meant by
> to collect the location of the logger output

model/log/sender.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	for _, line := range strings.Split(m.String(), "\n") {
+		if line == "" {
+			continue
+		}
+
+		logLine, err := s.opts.Parse(line)
+		if err != nil {
+			s.opts.Local.Send(message.NewErrorMessage(level.Error, errors.Wrap(err, "parsing log line")))
```
ybrill, line 111: could it make sense to use `s.ErrorHandler()` for these?

model/log/sender.go
julianedwards: Yeah this will get used once I create the exported functions, but I just didn't want the linter to complain. It trickles down, since the unexported function isn't exported it complains that this const isn't used either.

model/log/sender.go
julianedwards: We could, but then we risk handling corrupt data on retrieve. We would just have to make sure to handle it properly there.

model/log/sender.go
julianedwards: I think the only risk there is that we hold on to a large array that we may not end up using, wasting memory.

model/log/sender.go
julianedwards: It doesn't actually matter, either way is technically correct. This just makes it a bit easier to test "overflowing" the buffer.

model/log/sender.go
julianedwards: Good call!

model/log/sender.go
julianedwards: I don't like to cause it forces us to use an `error` and a `message.Composer` which feels like overkill here since we can just log these errors.

ybrill (approved):

PR #6738, opened 2023-07-06
EVG-20282 Use correct context 

malikchaya2:
EVG-20282 

### Description
All the logic after the exec timeout ticker starts should respect the exec timeout. This fixes the wrong context being used in some cases.


agent/task.go
```diff
@@ -67,6 +67,7 @@ func (a *Agent) startTask(ctx context.Context, tc *taskContext, complete chan<-
 	go a.startMaxExecTimeoutWatch(ctx, tc, innerCancel)
 
 	// set up the system stats collector
+	//start here
```
Kimchelly, line 70: Remove comment.

Kimchelly (approved):
LGTM mod remove comment.

PR #6741, opened 2023-07-06
EVG-20147 Add generate tasks timing trackers to OTEL

hadjri:
EVG-20147

### Description
Now that we have access to OTEL and honeycomb, we can replace our existing generate tasks timing Splunk timestamps with OTEL traces to get a more fine tuned view into the telemetry of the generate.tasks job to determine which areas need the most focus of improvement (if possible).

Added traces across the job, and removed the existing Splunk timing logs.

### Testing
Tested in staging, example [trace](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/47r3oqjeCuT/trace/hPKuXaCX5XK?fields[]=c_name&fields[]=c_service.name&span=4fc12f95026cd735) of a generate tasks job


model/generate.go
```diff
@@ -20,6 +19,14 @@ const (
...
 )
 
+const (
+	findTasksWithActivationSecondsAttribute  = "evergreen.find_tasks_with_activation.seconds"
+	getTasksWithDependenciesSecondsAttribute = "evergreen.get_tasks_with_dependencies.seconds"
+	addNewTasksSecondsAttribute              = "evergreen.add_new_tasks.seconds"
+	addNewBuildsSecondsAttribute             = "evergreen.add_new_builds.seconds"
+	addDependenciesSecondsAttribute          = "evergreen.add_dependencies.seconds"
+)
+
```
malikchaya2, line 29: Since I'm guessing we'll be doing more of this down the line, I'd advocate against storing the strings fully formed. I think an ideal setup would be where we store the "evergreen" part, the middle part, and the "seconds" part separately, and have a function build the string when we need it. I think that would be more scalable.

model/generate.go
```diff
@@ -304,7 +319,9 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 // CheckForCycles builds a dependency graph from the existing tasks in the version and simulates
 // adding the generated tasks, their dependencies, and dependencies on the generated tasks to the graph.
 // Returns a DependencyCycleError error if the resultant graph contains dependency cycles.
-func (g *GeneratedProject) CheckForCycles(v *Version, p *Project, projectRef *ProjectRef) error {
+func (g *GeneratedProject) CheckForCycles(ctx context.Context, v *Version, p *Project, projectRef *ProjectRef) error {
+	ctx, span := tracer.Start(ctx, "simulate-dependencies")
```
malikchaya2, line 323: Why "simulate-dependencies" over "Check-for-cycles"?

units/generate_tasks.go
```diff
@@ -19,12 +19,25 @@ import (
...
+const (
+	taskIdAttribute                        = "evergreen.task.id"
+	taskVersionAttribute                   = "evergreen.task.version"
+	parseProjectSecondsAttribute           = "evergreen.parse_project.seconds"
+	mergeGeneratedProjectsSecondsAttribute = "evergreen.merge_generated_projects.seconds"
+	createNewVersionSecondsAttribute       = "evergreen.create_new_version.seconds"
+	validateConfigSecondsAttribute         = "evergreen.validate_config.seconds"
+	simulateDependenciesSecondsAttribute   = "evergreen.simulate_dependencies.seconds"
+	saveTasksSecondsAttribute              = "evergreen.save_tasks.seconds"
+)
```
malikchaya2, line 39: same comments as above about breaking up the strings

units/generate_tasks.go
```diff
@@ -113,93 +130,43 @@ func (j *generateTasksJob) generate(ctx context.Context, t *task.Task) error {
...
-	grip.Debug(message.Fields{
-		"message":       "generate.tasks timing",
-		"function":      "generate",
-		"operation":     "parseProjects",
-		"duration_secs": time.Since(start).Seconds(),
-		"task":          t.Id,
-		"job":           j.ID(),
-		"version":       t.Version,
-	})
-	start = time.Now()
```
malikchaya2, line 129: do we want to keep both for a bit or are we confident enough that the honeycomb trace will be a sufficient replacement?

validator/otel.go
```diff
@@ -0,0 +1,7 @@
+package validator
+
+import "go.opentelemetry.io/otel"
+
+const packageName = "github.com/evergreen-ci/evergreen/validator"
```
malikchaya2, line 5: should we maybe move `github.com/evergreen-ci/evergreen/` to a global const somewhere?

malikchaya2 (changes requested):

units/generate_tasks.go
hadjri: I think it should be fine to go straight to honeycomb

model/generate.go
hadjri: That's what it was called in the old splunk logs, but you're right we should change it

model/generate.go
hadjri: Oh we actually don't need this block anymore. Removed

units/generate_tasks.go
```diff
@@ -19,12 +19,25 @@ import (
...
+	"go.opentelemetry.io/otel/trace"
 )
 
 const (
 	generateTasksJobName = "generate-tasks"
 )
 
+const (
+	taskIdAttribute                        = "evergreen.task.id"
+	taskVersionAttribute                   = "evergreen.task.version"
```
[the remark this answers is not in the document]
hadjri: Yes, thank you

validator/project_validator.go
```diff
@@ -298,8 +299,10 @@ func CheckProjectSettings(settings *evergreen.Settings, p *model.Project, ref *m
 	return errs
 }
 
-// checks if the project configuration has errors
-func CheckProjectConfigurationIsValid(settings *evergreen.Settings, project *model.Project, pref *model.ProjectRef) error {
+// CheckProjectConfigurationIsValid checks if the project configuration has errors
+func CheckProjectConfigurationIsValid(ctx context.Context, settings *evergreen.Settings, project *model.Project, pref *model.ProjectRef) error {
+	ctx, span := tracer.Start(ctx, "check-configuration")
```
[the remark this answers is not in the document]
hadjri: Hmm good idea, added

units/generate_tasks.go
hadjri: I changed to not use these anymore (because OTEL already tells you how many seconds, that's kind of the point 😅 ) so the only attributes we'll need are the existing ones on [globals.go](https://gist.github.com/hadjri/993d8b2c9bd025a8c44be1c76d054b85#file-globals-go-L467-L477)

units/generate_tasks.go
```diff
@@ -66,7 +68,15 @@ func NewGenerateTasksJob(versionID, taskID string, ts string) amboy.Job {
...
-	start := time.Now()
+	ctx, span := tracer.Start(ctx, "task-generation", trace.WithAttributes(
+		attribute.String(evergreen.TaskIDOtelAttribute, t.Id),
+		attribute.Int(evergreen.TaskExecutionOtelAttribute, t.Execution),
+		attribute.String(evergreen.VersionIDOtelAttribute, t.Version),
+		attribute.String(evergreen.BuildIDOtelAttribute, t.BuildId),
+		attribute.String(evergreen.ProjectIDOtelAttribute, t.Project),
+		attribute.String(evergreen.VersionRequesterOtelAttribute, t.Requester),
+	))
+	defer span.End()
```
ybrill, line 79: Idk if it's desirable, but if we use [ContextWithAttributes](https://github.com/evergreen-ci/utility/blob/4d91d64f59fb653893f0f0b0f301971ce050b389/otel.go#L37-L41) it could make all the child spans have all these attributes. For example, you could query how long `add-new-builds` takes by project.

units/generate_tasks.go
```diff
@@ -113,93 +123,43 @@ func (j *generateTasksJob) generate(ctx context.Context, t *task.Task) error {
...
-	})
-
-	start = time.Now()
-
 	// Don't use the job's context, because it's better to try finishing than to
 	// exit early after a SIGTERM from app server shutdown.
 	saveCtx, saveCancel := context.WithTimeout(context.Background(), 30*time.Minute)
 	defer saveCancel()
-	err = g.Save(saveCtx, j.env.Settings(), p, pp, v)
+	err = g.Save(ctx, saveCtx, j.env.Settings(), p, pp, v)
```
ybrill, line 162: an alternative to passing through the job context could be to store the `SpanContext` in the `saveCtx`.
This could look something like
```go
saveCtx, saveCancel := context.WithTimeout(context.Background(), 30*time.Minute)
defer saveCancel()

// Inject the span context into the saveCtx.
saveCtx = trace.ContextWithSpanContext(saveCtx, span.SpanContext())
err = g.Save(ctx, saveCtx, j.env.Settings(), p, pp, v)
...
```

model/generate.go
```diff
@@ -325,7 +334,7 @@ func (g *GeneratedProject) CheckForCycles(v *Version, p *Project, projectRef *Pr
 // simulateNewTasks adds the tasks we're planning to add to the version to the graph and
 // adds simulated edges from each task that depends on the generator to each of the generated tasks.
 func (g *GeneratedProject) simulateNewTasks(graph task.DependencyGraph, v *Version, p *Project, projectRef *ProjectRef) (task.DependencyGraph, error) {
-	newTasks := g.getNewTasksWithDependencies(v, p, nil)
+	newTasks := g.getNewTasksWithDependencies(context.Background(), v, p, nil)
```
ybrill, line 337: Why not pass through the context from `CheckForCycles`?

units/generate_tasks.go
hadjri: Oh that's great, thanks!

units/generate_tasks.go
hadjri: Ah that is much cleaner thanks

ybrill (approved):
🐝 !

PR #6742, opened 2023-07-06
EVG-19147 SES Emails

ybrill:
[EVG-19147](https://jira.mongodb.org/browse/EVG-19147)

### Description
Send emails directly with the SES instead of sending to Sendmail on localhost (that then sends emails with SES 🙃). Fetches credentials from the instance profile assigned to the app server. The order is described [here](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials). Of note is that it looks at `AWS_WEB_IDENTITY_TOKEN_FILE` first which is set by IRSA, so this should just work™️  when we switch.

### Testing
I sent myself a notification when a task finished on staging.
Screenshot 2023-07-06 at 6 35 02 PM


config_notify.go
```diff
@@ -0,0 +1,76 @@
...
+		return errors.Errorf("maximum notification jobs per second is %d", maxNotificationsPerSecond)
+
+	}
+
+	return nil
+}
+
+// SESConfig configures the SES email sender.
+type SESConfig struct {
+	From string `bson:"from" json:"from" yaml:"from"`
```
julianedwards, line 75: Nit: can we call this `SenderAddress`?

config_notify.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (c *NotifyConfig) ValidateAndDefault() error {
+	if c.BufferIntervalSeconds <= 0 {
+		c.BufferIntervalSeconds = DefaultBufferIntervalSeconds
+	}
+	if c.BufferTargetPerInterval <= 0 {
+		c.BufferTargetPerInterval = DefaultBufferTargetPerInterval
+	}
+
+	// cap to 100 jobs/sec per server
```
julianedwards, line 63: Nit: `Cap to 100 jobs/sec per server.`

rest/route/admin_test.go
```diff
@@ -175,9 +172,6 @@ func (s *AdminRouteSuite) TestAdminRoute() {
 	s.EqualValues(testSettings.LoggerConfig.Buffer.Count, settings.LoggerConfig.Buffer.Count)
 	s.EqualValues(testSettings.LoggerConfig.Buffer.UseAsync, settings.LoggerConfig.Buffer.UseAsync)
 	s.EqualValues(testSettings.LoggerConfig.Buffer.IncomingBufferFactor, settings.LoggerConfig.Buffer.IncomingBufferFactor)
-	s.EqualValues(testSettings.Notify.SMTP.From, settings.Notify.SMTP.From)
-	s.EqualValues(testSettings.Notify.SMTP.Port, settings.Notify.SMTP.Port)
-	s.Equal(len(testSettings.Notify.SMTP.AdminEmail), len(settings.Notify.SMTP.AdminEmail))
```
julianedwards, line 180: Should we check that the SES settings are properly fetched?

julianedwards (changes requested):

rest/route/admin_test.go
ybrill: done

julianedwards (approved):

PR #6744, opened 2023-07-07
EVG-16363: Add requesters field to external links

minnakt:
EVG-16363

### Description
The purpose of this ticket is to allow users to specify where metadata links will show up. The `requesters` field is going to be used for that specification (so you can choose to show external links on commits, patches, both, etc).

Note: Following this ticket, I will need to update all existing external links definitions to have requesters of `["patch_request", "github_pull_request", "merge_test"]`. This is because we currently show external links for patches only (i.e. anything that matches [`IsPatchRequester`](https://github.com/evergreen-ci/evergreen/blob/20a22068d7cd254c667ca8d7c3d51982d59f3c4b/globals.go#L901)).

### Testing
- Added GraphQL tests

### Documentation
- Updated Metadata Links section

### DIFF
--- a/docs/Project-Configuration/Project-and-Distro-Settings.md
+++ b/docs/Project-Configuration/Project-and-Distro-Settings.md
@@ -382,9 +382,6 @@ Options:
 Customize additional links to specify for your project under the Plugins section
 of the project page, by specifying a link and title. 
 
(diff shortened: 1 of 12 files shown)


khelif96 (approved):
Makes sense to me!

hadjri (approved):

PR #6746, opened 2023-07-07
EVG-20280 Fix wrong current command for post block

malikchaya2:
[EVG-20280](https://jira.mongodb.org/browse/EVG-20280)

### Description
This fixes a bug where setIdleTimeout logged the wrong command. It also sets the end task failure type and description when the task succeeded but post failed. 

### Testing
Added unit tests 
Also [tested on staging](https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu2004_chayaTest_patch_dc7231007f634a2fed4da3eee74c584b3cb40b7d_64ac4a139dbe329c10117110_23_07_10_18_12_47/logs?execution=0&logtype=agent). The first execution ran on main and shows Set idle timeout for 'shell.exec' (step 1 of 1), and the second execution ran with these changes and correctly shows Set idle timeout for 'shell.exec' (step 1 of 1) in block 'post'.
### Documentation
N/A


agent/command.go
```diff
@@ -156,12 +156,10 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 		tc.taskConfig.Expansions.Put(key, newVal)
 	}
 
+	tc.setCurrentCommand(cmd)
 	if options.isTaskCommands || options.failPreAndPost {
-		tc.setCurrentCommand(cmd)
 		tc.setCurrentIdleTimeout(cmd)
```
Kimchelly, line 161: Seems like the idle timeout still needs to be set for the post block to mirror the old `else` block.

agent/agent_test.go
```diff
@@ -928,6 +1138,20 @@ func (s *AgentSuite) getPanicLogs() []string {
 	return panicLogs
 }
 
+// getIdleTimeoutLogs returns logs setting the idle timeout.
+func (s *AgentSuite) getIdleTimeoutLogs() []string {
```
Kimchelly, line 1143: This can be replaced by just using `checkMockLogs` to look for the strings `Set idle timeout for ...` instead of making its own function.

agent/agent_test.go
```diff
@@ -928,6 +1138,20 @@ func (s *AgentSuite) getPanicLogs() []string {
...
+	var panicLogs []string
+	msgsByTask := s.mockCommunicator.GetMockMessages()
+	for _, msgs := range msgsByTask {
+		for _, msg := range msgs {
+			if strings.Contains(msg.Message, "Set idle timeout") {
+				panicLogs = append(panicLogs, msg.Message)
```
Kimchelly, line 1149: This function is already going away, but I would avoid reusing this variable name because this function has nothing to do with panics.

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+		BuildVariant: "some_build_variant",
+		DisplayName:  name,
+		Version:      "my_version",
+	}
+	s.tc.taskModel = t
+	s.tc.taskConfig.Task = t
+	_, err = s.a.runTask(s.ctx, s.tc)
+
+	s.NoError(err)
+	s.Equal("success", s.mockCommunicator.EndTaskResult.Detail.Status)
```
Kimchelly, line 498: `evergreen.TaskSucceeded` instead of `"success"`.

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+		DisplayName:  name,
+		Version:      "my_version",
+	}
+	s.tc.taskModel = t
+	s.tc.taskConfig.Task = t
+	_, err = s.a.runTask(s.ctx, s.tc)
+
+	s.NoError(err)
+	s.Equal("failed", s.mockCommunicator.EndTaskResult.Detail.Status)
+	// it should show main as the failing command when both failed.
```
Kimchelly, line 584: Rather than put a comment, you can make this more useful by putting it a string at the end of `s.Equal` indicating your expectation, which will helpfully show up in the Go test output logs if the assertion fails. Something like:
`s.Equal(expected, actual, "should show main block command as the failing command if both main and post block commands fail")`

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+		BuildVariant: "some_build_variant",
+		DisplayName:  name,
+		Version:      "my_version",
+	}
+	s.tc.taskModel = t
+	s.tc.taskConfig.Task = t
+	_, err = s.a.runTask(s.ctx, s.tc)
+
+	s.NoError(err)
+	s.Equal("failed", s.mockCommunicator.EndTaskResult.Detail.Status)
```
Kimchelly, line 626: `evergreen.TaskFailed`

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+	s.Equal("'shell.exec' (step 1 of 1) in block 'post'", s.mockCommunicator.EndTaskResult.Detail.Description)
+
+	logs := s.getIdleTimeoutLogs()
+	s.Equal("Set idle timeout for 'shell.exec' (step 1 of 1) (test) to 2h0m0s.", logs[0])
```
Kimchelly, line 461: This line is probably going to go away, but in the future, you should probably check the length of `logs` before indexing the slice, or else the test will panic (and you'll have to debug the test to figure out why it panics). You can do something like `s.Require().Len(logs, 2)`.

agent/agent_test.go
```diff
@@ -420,6 +420,217 @@ post:
...
+		BuildVariant: "some_build_variant",
+		DisplayName:  name,
+		Version:      "my_version",
+	}
+	s.tc.taskModel = t
+	s.tc.taskConfig.Task = t
+	_, err = s.a.runTask(s.ctx, s.tc)
+
+	s.NoError(err)
+	s.Equal("success", s.mockCommunicator.EndTaskResult.Detail.Status)
```
Kimchelly, line 541: In the case of the task succeeding, it seems like we don't need to check the task end description since it's only relevant in the case of failure. I think we can omit this check (and same with the other tests that succeed).

Kimchelly (changes requested):
* Did you test this in staging?
* Can we update the docs for idle timeout to describe which command blocks it applies to?
* While we're at it, can we also update the docs to better describe how `pre_error_fails_task` and `post_error_fails_task` are expected to behave? For example, if `pre_error_fails_task` is set and it fails, do the main and post block run even though the task is already known to fail, or does it exit early?

agent/agent_test.go
malikchaya2: done

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
```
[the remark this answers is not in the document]
malikchaya2: done

agent/agent_test.go
malikchaya2: done

agent/agent_test.go
malikchaya2: yeah, I was just thinking that too.

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+		BuildVariant: "some_build_variant",
+		DisplayName:  name,
+		Version:      "my_version",
+	}
+	s.tc.taskModel = t
+	s.tc.taskConfig.Task = t
+	_, err = s.a.runTask(s.ctx, s.tc)
+
+	s.NoError(err)
+	s.Equal("failed", s.mockCommunicator.EndTaskResult.Detail.Status)
```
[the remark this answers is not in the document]
malikchaya2: done

agent/agent_test.go
malikchaya2: thanks, done

agent/agent_test.go
malikchaya2: done

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
```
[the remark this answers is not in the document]
malikchaya2: done

agent/command.go
malikchaya2: fixed

agent/agent_test.go
malikchaya2: yeah I have since cleaned it up and then removed it

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
 	)
 }
 
+func (s *AgentSuite) TestFailingPostAfterMainSucceeds() {
+	name := "test_failing_command_after_main_succeeds"
```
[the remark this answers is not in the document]
malikchaya2: done

agent/agent_test.go
```diff
@@ -420,6 +420,216 @@ post:
...
+	s.NoError(err)
+	s.Equal("failed", s.mockCommunicator.EndTaskResult.Detail.Status)
+	s.Equal("'shell.exec' (step 1 of 1) in block 'post'", s.mockCommunicator.EndTaskResult.Detail.Description)
+
+	logs := s.getIdleTimeoutLogs()
+	s.Equal("Set idle timeout for 'shell.exec' (step 1 of 1) (test) to 2h0m0s.", logs[0])
+	s.Equal("Set idle timeout for 'shell.exec' (step 1 of 1) in block 'post' (test) to 2h0m0s.", logs[1])
+}
+
+func (s *AgentSuite) TestFailingPostErrorFailsNotTrue() {
```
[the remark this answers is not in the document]
malikchaya2: done. It gets very long, but adds clarity.

agent/command.go
Kimchelly: I'm pretty sure we can just `tc.setCurrentIdleTimeout` without any condition. It will default to the currently-configured idle timeout or fall back to the default anyways.

agent/task.go
```diff
@@ -207,10 +207,15 @@ func (tc *taskContext) setCurrentIdleTimeout(cmd command.Command) {
...
 	defer tc.Unlock()
 
-	var timeout time.Duration
 	if cmd == nil {
-		timeout = defaultIdleTimeout
-	} else if dynamicTimeout := tc.taskConfig.GetIdleTimeout(); dynamicTimeout != 0 {
+		tc.setIdleTimeout(defaultIdleTimeout)
+		tc.logger.Execution().Debugf("Set idle timeout for %s (%s) to %s.",
+			tc.currentCommand.DisplayName(), tc.currentCommand.Type(), tc.getIdleTimeout())
+		return
```
Kimchelly, line 214: As described in the above comment, we don't need to re-add the nil check as long as we always pass in the current command (i.e. remove the if-else conditional above and just `tc.setCurrentIdleTimeout`.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -464,22 +464,27 @@ timeout:
...
 You may also force a specific command to trigger a failure if it does
 not appear to generate any output on `stdout`/`stderr` for more than a
 certain threshold, using the `timeout_secs` setting on the command. As
 long as the command does not appear to be idle it will be allowed to
 continue, but if it does not write any output for longer than
-`timeout_secs` then the timeout handler will be triggered. This timeout
-defaults to 2 hours.
+`timeout_secs` then the timeout handler will be triggered. If timeout_secs
+is specified for a command that runs in a pre or post block, it will
+only be set in pre or post if pre_error_fails_task or post_error_fails_task
```
Kimchelly, line 485: Since it's set at the command level, it seems like it would be more consistent/desirable behavior to have the idle timeout always be respected as the timeout, regardless of whether `pre/post_error_fails_task` is set. Whether or not pre/post is relevant to the end task status is an independent concern from how/when the command times out.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -509,7 +514,14 @@ cause the task to fail, set the field `pre_error_fails_task` to true.
 Likewise, setting the field `post_error_fails_task` to true will enforce
 that failures in `post` cause the task to fail.
 
-``` yaml
+If pre_error_fails_task is set to true and pre fails, it will exit early 
+and not run the main task commands, but it will still run the post block.
+
+If post_error_fails_task is set to true and both the main task and post 
+block fail, the failing command in the main task will be displayed as the 
+failing command, not the failing post command.
```
Kimchelly, line 522: Can we also mention that `pre_error_fails_task` and `post_error_fails_task` has no effect for tasks in task groups? Users can instead use the analogous task group YAML settings.

Kimchelly (changes requested):

docs/Project-Configuration/Project-Configuration-Files.md
malikchaya2: I changed the code above to set tc.setCurrentIdleTimeout without any condition and removed that line in the docs.

docs/Project-Configuration/Project-Configuration-Files.md
malikchaya2: done

agent/agent_timeout_test.go
```diff
@@ -111,7 +111,7 @@ func (s *TimeoutSuite) TestExecTimeoutProject() {
 
 	detail := s.mockCommunicator.GetEndTaskDetail()
 	s.Equal(evergreen.TaskFailed, detail.Status)
-	s.Equal("test", detail.Type)
+	s.Equal("system", detail.Type)
```
Kimchelly, line 114: Hmm, this seems like the wrong assertion because [the YAML that it used](https://github.com/evergreen-ci/evergreen/blob/9e3b301c551c6bc6ed0dc0ecde9022679051b17f/agent/internal/client/testdata/exec_timeout_task.yaml#L6) hits the exec timeout, then runs [the timeout block](https://github.com/Kimchelly/evergreen/blob/349ec5b72098ab1785544b2aaf45736732d58965/agent/internal/client/testdata/exec_timeout_task.yaml#L13-L17), but the timeout block . I suspect it's recording the timeout block's command as the failing command since the timeout block runs after the main block, but we want the main block's command to supercede the timeout block's command (just like the post block).

I imagine one way to fix this is that, since the timeout block has no influence on the end task response and post with `post_error_fails_task` is the only way to override the end task details, we can just call `a.endTaskResponse` in `a.wait` before the timeout block runs and pass those end task details into `a.handleTaskResponse` (you might also have to call `a.endTaskResponse` in a couple other usages of `a.handleTaskResponse` as well).

Kimchelly (changes requested):

agent/agent_timeout_test.go
malikchaya2: done

agent/agent.go
```diff
@@ -648,6 +635,19 @@ func (a *Agent) wait(ctx, taskCtx context.Context, tc *taskContext, heartbeat ch
 			tc.logger.Execution().Debugf("Found no OOM kill (in %.3f seconds).", time.Since(startTime).Seconds())
```
Kimchelly, line 635: These lines still have to obey the order of operations that they had before to avoid changing the order that timeout/post/OOM kill checking run, so the timeout and OOM tracker handling have to run after `endTaskResponse` and before the post block runs.

Kimchelly (changes requested):

agent/agent.go
malikchaya2: endTaskResponse runs in  a.finishTask on line 609, so this runs after that. This whole thing runs in runTask, which runs before runPostGroupCommands in the loop.

Kimchelly: This is talking about the order of [`runPostTaskCommands`](https://github.com/evergreen-ci/evergreen/blob/main/agent/agent.go#L680-L684) (post and teardown task) and `runEndTaskSync` with respect to these lines. Before this PR, these lines ran before `runPostTaskCommands`, but now they run after.

agent/agent.go
```diff
@@ -770,6 +773,7 @@ func (a *Agent) endTaskResponse(tc *taskContext, status string, message string)
 	if tc.taskConfig != nil {
 		detail.Modules.Prefixes = tc.taskConfig.ModulePaths
 	}
+	a.handleTimeoutAndOOM(ctx, tc, status)
```
Kimchelly, line 776: This needs to be moved because it should only run in [cases of success/failure](https://github.com/evergreen-ci/evergreen/blob/0acd94f7c3d69739f39198527bfaa44af1f5e433/agent/agent.go#L678-L690). In the system failed case, timeout cannot run because the task has encountered an unfixable error while gathering necessary task information (such as due to an agent bug, or inability to get information from the app server). Also, it would be a weird combining of responsibilities to run more task logic in a function that's meant to just calculate an end task response. I would move `a.handleTimeoutAndOOM` to call them in the individual success/failed cases.

Kimchelly (changes requested):

agent/agent.go
```diff
@@ -686,12 +686,14 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 			setEndTaskCommand(tc, detail, "", "")
 		}
 		a.runEndTaskSync(ctx, tc, detail)
+		a.handleTimeoutAndOOM(ctx, tc, status)
```
Kimchelly, line 689: This has to run before `runPostTaskCommands` and before the `Task completed` log to preserve the original order of operations (same with below).

Kimchelly (changes requested):

agent/agent.go
malikchaya2: moved

agent/agent.go
```diff
@@ -680,20 +679,21 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 	switch detail.Status {
 	case evergreen.TaskSucceeded:
 		tc.logger.Task().Info("Task completed - SUCCESS.")
+		a.handleTimeoutAndOOM(ctx, tc, status)
```
Kimchelly, line 682: I has to go right before L681, as I stated in my comment (same as the other one).

agent/agent.go
```diff
@@ -680,20 +679,21 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
...
 		if err := a.runPostTaskCommands(ctx, tc); err != nil {
 			tc.logger.Task().Info("Post task completed -- FAILURE. Overall task status changed to FAILED.")
 			detail.Status = evergreen.TaskFailed
 			setEndTaskCommand(tc, detail, "", "")
 		}
 		a.runEndTaskSync(ctx, tc, detail)
-		a.handleTimeoutAndOOM(ctx, tc, status)
+
```
Kimchelly, line 689: nit: remove random newline.

Kimchelly (changes requested):

agent/agent.go
malikchaya2: my bad, I misread your previous comment. 
fixed

agent/agent.go
malikchaya2: done

Kimchelly (approved):

PR #6756, opened 2023-07-10
EVG-19325, EVG-20113: require GitHub patch definition tasks to finish

Kimchelly:
EVG-19325
EVG-20113

### Description
This is the same change as #6521, with some small changes to reduce the usage of GitHub API calls and fix a small bug. You can refer to the original PR for most of the changes, but the new changes from that PR are:

* This no longer sends a status update when the only unfinished tasks blocking patch completion are unscheduled essential tasks. While it would be a nice to have, it stresses the GitHub API rate limit because it will try to send when PRs are closed with unfinished patches, which will abort all of them. The thinking is that it's not that important to alert users because unscheduling your PR tasks is not something a user would typically do, and they can type `evergreen refresh` if they want to see what's holding up their patch from being finished.
* Fix an issue where a version with unfinished essential tasks might be marked finished because one build finished all its tasks and another build was entirely unscheduled before it could run any of its tasks. In this case, the version should _not_ be marked finished because there are still some essential tasks to run.

### Testing
Added unit tests. I also tested the same conditions as in the original PR in the sandbox.

### Documentation
Added ADR and updated GitHub PR integration docs (same as original PR).


model/build/build.go
```diff
@@ -343,12 +366,13 @@ func (b *Build) GetFinishedNotificationDescription(tasks []task.Task) string {
 		case t.Status == evergreen.TaskFailed:
 			failed++
 
-		case evergreen.IsSystemFailedTaskStatus(t.Status):
-			systemError++
-
 		case utility.StringSliceContains(evergreen.TaskUncompletedStatuses, t.Status):
-			noReport++
-
+			if utility.StringSliceContains([]string{evergreen.TaskDispatched, evergreen.TaskStarted}, t.Status) || (t.Activated && !t.Blocked() && !t.IsFinished()) {
```
bynn, line 370: can we use the TaskInProgressStatuses?

bynn (changes requested):
If I understand correctly, if a user sees that they failed 1 essential task and then aborts the patch, the status will say "essential tasks not scheduled" but I think the preferable behavior for it to say that there was a failure. 

It doesn't seem hugely important do you think it would be worth changing?

Kimchelly (commented):
I waffled on this a bunch because I can't really think of a valid case where a user would want to manually abort the patch for the commit they want to merge (because they want it to pass to merge), unless they planned on closing the PR anyways (in which case the patch result is irrelevant to them). Nonetheless, it doesn't cause any harm, so I loosened the rules so that essential tasks only block a build/patch from being marked successful, not being marked failed (and renamed stuff to reflect that meaning).

bynn (approved):

PR #6757, opened 2023-07-10
EVG-20419 log otel errors

ybrill:
[EVG-20419](https://jira.mongodb.org/browse/EVG-20419)

### Description
If the client fails to upload spans it will log the error with the otel error handler. By default this just goes to stdout which ends up in the service logs on the host. It would make things a lot easier if they went with the rest of our logging to Splunk.

We already do this for the agent [here](https://github.com/evergreen-ci/evergreen/blob/0b38c0c784757ca57aa54a8ced532d02c49435db/agent/otel.go#L89-L91).

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -883,7 +883,7 @@ func (e *envState) initTracer(ctx context.Context) error {
 	tp.RegisterSpanProcessor(utility.NewAttributeSpanProcessor())
 	otel.SetTracerProvider(tp)
 	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
-		grip.Error(errors.Wrap(err, "encountered otel error"))
+		grip.Error(errors.Wrap(err, "otel error"))
 	}))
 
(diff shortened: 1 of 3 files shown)


environment.go
```diff
@@ -882,6 +882,9 @@ func (e *envState) initTracer(ctx context.Context) error {
 	)
 	tp.RegisterSpanProcessor(utility.NewAttributeSpanProcessor())
 	otel.SetTracerProvider(tp)
+	otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
+		grip.Error(errors.Wrap(err, "encountered otel error"))
```
julianedwards, line 886: I would set the message as `otel error`.

julianedwards (changes requested):

environment.go
ybrill: done

opsnlops (approved):
LGTM!

PR #6759, opened 2023-07-11
EVG-20358: update Dependabot for support rotation

Kimchelly:
EVG-20358

### Description
Since we're absorbing Dependabot into the weekly support rotation, I tuned it so that it no longer pings evg-app (it's on the person on support to review the Dependabot PRs) and reduce the number of open PRs to limit the amount that a single person on rotation has to review. Eventually, we'll get to a point where there's very few Dependabot PRs, but for now, our dependencies are very outdated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,8 +7,6 @@ updates:
       interval: monthly
       time: "06:00"
       timezone: "America/New_York"
-    open-pull-requests-limit: 99
+    open-pull-requests-limit: 5
     commit-message:
       prefix: "CHORE: "
-    reviewers:
-      - evergreen-ci/evg-app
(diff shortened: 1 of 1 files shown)


khelif96:
Is this only for the app primary/secondary?  If it's for all primary's regardless of team would it make sense to also do this for the UI repos?

bynn:
should we consider leaving the open pr limit 99 until more of our dependencies are up to date later?

Kimchelly:
This is only for app because we manage the Evergreen repo's dependencies. I don't know what is the policy on UI repos, so I assume @ablack12 can speak to this when she's back.

We could leave the PR limit at 99, but having a lower limit + more frequent update is nice to amortize the amount of work every person on rotation has to do to get us up to date, rather than leave the first unfortunate person on rotation with a huge amount of PRs to merge/triage due to how out of date everything is. I expect that once we're more caught up, most rotations will only have a few Dependabot PRs, if any.

bynn (approved):
I think what you're saying makes sense!

PR #6760, opened 2023-07-11
EVG-20416 Calculate base version outside of getTasksByVersion aggregation

khelif96:
EVG-20416

### Description
This introduces yet another optimization to the `getTasksByVersion` aggregation in the form of pre-populating the base task. This is a fairly significant optimization, in my testing it has led to a > 20x performance improvement in the load time of the query. This works by first precalculating the base version id and using that as a basis for populating the base task status. 
Previously the `$lookup` would determine the base status by trying to match several factors including build variant, task name, order, and revision this would cause the `$lookup` to use an inefficient index and slow index. Followed by a slow operation to match the fields. 

Instead, I went with the approach of fetching all tasks for both the versions and base versions in the initial match and then grouping base tasks and tasks. This is far more performant than a `$lookup` due to being able to use an index to fetch all fields and doing simple matches on the data to ensure things are grouped together correctly.
### Testing
Before:
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/gM5yNcsufZZ/trace/zFSgTuokHrc?fields%5B%5D=c_name&span=860385508eb12651

7.603s
After:
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/zCPJaU1LQg3/trace/EfPTf7bCaCd?fields%5B%5D=c_name&span=a4e5ae46990ffb79
377ms

### Testing
This logic is tested extensively by existing tests and none of them needed to be modified apart from adding the additional fields necessary for prepopulating the base version id.

The main patch I used to test this feature which features > 10k tasks which represents an extreme case.
https://spruce-staging.corp.mongodb.com/version/62db074797b1d369e3e28ded/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


model/task/db.go
```diff
@@ -2599,60 +2602,110 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
+			},
+		})
+
+		// Project out the root task since it is no longer needed
+		pipeline = append(pipeline, bson.M{
+			"$project": bson.M{
+				"root_task": 0,
+			},
+		})
+
```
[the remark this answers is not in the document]
khelif96: Good catch!

model/task/db.go
```diff
@@ -2592,60 +2595,103 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
+			"$project": bson.M{
+				"base_task": 0,
+				"tasks":     0,
+			},
+		})
+
+		// Use display status from the base task as its status
+		pipeline = append(pipeline, bson.M{
+			"$addFields": bson.M{
+				bsonutil.GetDottedKeyName("root_task", BaseTaskKey, StatusKey): "$" + bsonutil.GetDottedKeyName("root_task", BaseTaskKey, DisplayStatusKey),
```
minnakt, line 2675: I wonder if you can use `BaseTaskStatusKey` here... I would double check though lol 

```suggestion
				bsonutil.GetDottedKeyName("root_task", BaseTaskStatusKey): "$" + bsonutil.GetDottedKeyName("root_task", BaseTaskKey, DisplayStatusKey),
```

model/task/db.go
```diff
@@ -2592,60 +2595,103 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
+			},
+		})
+
+		// Project out the base task and the tasks array since they are no longer needed
+		pipeline = append(pipeline, bson.M{
+			"$project": bson.M{
+				"base_task": 0,
+				"tasks":     0,
+			},
+		})
```
minnakt, line 2670: Just curious for my education... I assume that these would get naturally dropped when you replace root at the end, but is doing this faster? :o

graphql/version_resolver.go
```diff
@@ -302,26 +302,25 @@ func (r *versionResolver) Tasks(ctx context.Context, obj *restModel.APIVersion,
 			taskSorts = append(taskSorts, task.TasksSortOrder{Key: key, Order: order})
 		}
 	}
-	v, err := model.VersionFindOne(model.VersionById(versionId).WithFields(model.VersionRequesterKey))
+	baseVersionID := ""
+	baseVersion, err := model.FindBaseVersionForVersion(utility.FromStringPtr(obj.Id))
 	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding version with id: '%s': %s", versionId, err.Error()))
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding base version id for version with id: '%s': %s", versionId, err.Error()))
```
minnakt, line 308: nit. shorten error message & remove the first colon
```suggestion
		return nil, InternalServerError.Send(ctx, fmt.Sprintf("finding base version id for version '%s': %s", versionId, err.Error()))
```

minnakt (approved):
yay! looking forward to the performance benefits

model/task/db.go
khelif96: It was mainly as a precaution to prevent there from being 4 task documents nested in one task document at this stage of the pipeline. But I moved it around a little bit to get rid of the tasks array earlier.

PR #6761, opened 2023-07-11
EVG-19935: Add SaveDistroSection mutation

minnakt:
EVG-19935

### Description
This PR introduces the SaveDistroSection resolver, which currently only works for the general section. Other sections can be added incrementally. The implementation follows a similar pattern to project settings, in that a subset of fields will be updated per section.

### Testing
- Unit tests &  GraphQL tests


minnakt:
**Note:** Since the implementation diverged significantly from the technical design, I'm going to wait for feedback from the UI team before requesting an app review, to confirm that this is going on the right track. Tagging Sophie since you have the most context about why I had to make these changes 🙏

graphql/mutation_resolver.go
```diff
@@ -123,6 +125,41 @@ func (r *mutationResolver) SetAnnotationMetadataLinks(ctx context.Context, taskI
 	return true, nil
 }
 
+// SaveDistroSection is the resolver for the saveDistroSection field.
+func (r *mutationResolver) SaveDistroSection(ctx context.Context, distroID string, changes *restModel.APIDistro, section distro.DistroSettingsSection, onSave DistroOnSaveOperation) (*DistroWithHostCount, error) {
```
sophstad, line 129: I'll defer to an app engineer's opinion on this, but it may be preferable to move the heavy logic out of the resolver/GraphQL utils and into `rest/data/distros.go`

graphql/util.go
```diff
@@ -1032,3 +1033,44 @@ func getBaseTaskTestResultsOptions(ctx context.Context, dbTask *task.Task) ([]te
 
 	return taskOpts, nil
 }
+
+func handleOnSaveOperation(ctx context.Context, distroID string, userID string, onSave DistroOnSaveOperation) (int, error) {
```
sophstad, line 1037:
```suggestion
func handleDistroOnSaveOperation(ctx context.Context, distroID string, userID string, onSave DistroOnSaveOperation) (int, error) {
```

graphql/util.go
```diff
@@ -1032,3 +1033,44 @@ func getBaseTaskTestResultsOptions(ctx context.Context, dbTask *task.Task) ([]te
...
+			return 0, InternalServerError.Send(ctx, fmt.Sprintf("marking hosts for reprovision for distro with id '%s': %s", distroID, err))
+		}
+	case DistroOnSaveOperationRestartJasper:
+		catcher := grip.NewBasicCatcher()
+		for _, h := range hosts {
+			_, err = api.GetRestartJasperCallback(ctx, evergreen.GetEnvironment(), userID)(&h)
+			catcher.Wrapf(err, "marking host '%s' as needing Jasper service restarted", h.Id)
+		}
+		if catcher.HasErrors() {
+			return 0, InternalServerError.Send(ctx, fmt.Sprintf("marking hosts as needing Jasper service restarted for distro with id '%s': %s", distroID, err))
```
sophstad, line 1071:
```suggestion
			return 0, InternalServerError.Send(ctx, fmt.Sprintf("marking hosts as needing Jasper service restarted for distro '%s': %s", distroID, err))
```

graphql/util.go
```diff
@@ -1032,3 +1033,44 @@ func getBaseTaskTestResultsOptions(ctx context.Context, dbTask *task.Task) ([]te
...
+			catcher.Wrapf(err, "marking host '%s' as needing to reprovision", h.Id, distroID)
+		}
+		if catcher.HasErrors() {
+			return 0, InternalServerError.Send(ctx, fmt.Sprintf("marking hosts for reprovision for distro with id '%s': %s", distroID, err))
+		}
+	case DistroOnSaveOperationRestartJasper:
+		catcher := grip.NewBasicCatcher()
+		for _, h := range hosts {
+			_, err = api.GetRestartJasperCallback(ctx, evergreen.GetEnvironment(), userID)(&h)
+			catcher.Wrapf(err, "marking host '%s' as needing Jasper service restarted", h.Id)
```
sophstad, line 1068: Maybe we could just add the host ID to the catcher here since the error is wrapped in a similar message (and same above)

malikchaya2 (approved):
Approved with some nits

graphql/mutation_resolver.go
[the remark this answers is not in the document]
khelif96: @SupaJoon do you know why this is still happening?

validator/distro_validator.go
```diff
@@ -549,3 +550,41 @@ func ensureHasValidVirtualWorkstationSettings(ctx context.Context, d *distro.Dis
...
+		validationErrs = append(validationErrs, ensureNoAliases(d, allDistroAliases)...)
+	}
+	if len(d.Aliases) > 0 {
+		validationErrs = append(validationErrs, ensureValidAliases(d)...)
+	}
+	return validationErrs
+}
+
+// ValidateDistroSection validates that the incoming changes to a given section are valid.
+func ValidateDistroSection(ctx context.Context, originalDistro *distro.Distro, changes *distro.Distro, section distro.DistroSettingsSection, s *evergreen.Settings) error {
```
sophstad, line 567: [nit] I would probably grab settings from the Evergreen environment in this function directly, your call though!

validator/distro_validator.go
```diff
@@ -549,3 +550,41 @@ func ensureHasValidVirtualWorkstationSettings(ctx context.Context, d *distro.Dis
 	}
 	return errs
 }
+
+func validateAliases(ctx context.Context, d *distro.Distro, allDistroAliases []string) ValidationErrors {
```
sophstad, line 554: Would you mind adding a unit test for this validator?

graphql/util.go
```diff
@@ -996,6 +998,50 @@ func getBaseTaskTestResultsOptions(ctx context.Context, dbTask *task.Task) ([]te
 	return taskOpts, nil
 }
 
+func handleDistroOnSaveOperation(ctx context.Context, distroID string, userID string, onSave DistroOnSaveOperation) (int, error) {
```
sophstad, line 1001: This is tricky stuff, nice work Minna 😍

graphql/schema/types/distro.graphql
```diff
@@ -1,10 +1,142 @@
...
+
+enum DistroOnSaveOperation {
+  DECOMMISSION
+  RESTART_JASPER
+  REPROVISION
+  NONE
+}
+
+###### INPUTS ######
+input SaveDistroOpts {
```
sophstad, line 25: Not sure if this is a practice worth keeping around, but we do usually name inputs with `Input` at the end
```suggestion
input SaveDistroInput {
```

sophstad (changes requested):

graphql/schema/types/distro.graphql
minnakt: oops sorry yes!

model/distro/distro.go
```diff
@@ -876,3 +877,49 @@ func AllDistros(ctx context.Context) ([]Distro, error) {
...
+const (
+	DistroSettingsGeneral  DistroSettingsSection = "GENERAL"
+	DistroSettingsProvider DistroSettingsSection = "PROVIDER"
+	DistroSettingsTask     DistroSettingsSection = "TASK"
+	DistroSettingsHost     DistroSettingsSection = "HOST"
+	DistroSettingsProject  DistroSettingsSection = "PROJECT"
+)
+
+// UpdateDistroSection saves the changes for a given section and logs the events with the associated user.
+func UpdateDistroSection(ctx context.Context, originalDistro *Distro, changes *Distro, section DistroSettingsSection, userID string) (*Distro, error) {
```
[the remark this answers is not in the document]
minnakt: It might be tricky to combine them at this point since only one section (the general section) is complete and there may be some unexpected curve balls down the road -- we could definitely revisit it though!

sophstad (approved):

PR #6762, opened 2023-07-11
EVG-19587: Update GQL context canceled error comparison

SupaJoon:
EVG-19587

### Description
The current comparison using `errors.Is(err, context.Canceled)` doesn't work because the error propagated to GQL diverges from the original context.Canceled error. The original PR can be viewed here: https://github.com/evergreen-ci/evergreen/pull/6520

### DIFF
--- a/graphql/http_handler.go
+++ b/graphql/http_handler.go
@@ -16,8 +16,6 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
-const ContextCanceledErrorMessage = "context canceled"
-
 // Handler returns a gimlet http handler func used as the gql route handler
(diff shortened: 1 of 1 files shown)


graphql/http_handler.go
```diff
@@ -15,6 +16,8 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
+const ContextCanceledErrorMessage = "context canceled"
```
Kimchelly, line 19: Rather than duplicate the `context.Canceled` error, it seems like you could just use `context.Canceled.Error()`.

Kimchelly (approved):
LGTM mod suggestion to use the standard library error.

PR #6764, opened 2023-07-12
EVG-19937: Add CopyDistro mutation

sophstad:
EVG-19937

### Description
- Add CopyDistro mutation. The operation will fail if the distro validator returns any errors.
- Add new SSH key to test data to facilitate e2e testing

### Testing
- Unit test copy function and GraphQL mutation

### Documentation
N/A


graphql/tests/mutation/copyDistro/results.json
```diff
@@ -0,0 +1,48 @@
...
+          {
+            "message": "copying distro: 404 (Not Found): distro 'fake-distro' not found",
+            "path": [
+              "copyDistro"
+            ],
+            "extensions": {
+              "code": "INTERNAL_SERVER_ERROR"
+            }
+          }
+        ]
```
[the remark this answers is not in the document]
sophstad: Good catch! Interestingly it looks like there's precedent for returning a "not found" error as an internal server error here:

https://github.com/evergreen-ci/evergreen/blob/bcb5efb3b9581e36b456b88c6d570ec3d3b4ae81/graphql/mutation_resolver.go#L491-L492

The `CopyProject` implements a similar StatusNotFound error that is returned as an InternalServerError by GraphQL. So I'm kinda inclined to leave this as is 🤷‍♀️

graphql/schema/types/distro.graphql
```diff
@@ -5,6 +5,22 @@ enum DistroSettingsAccess {
...
+input CopyDistroInput {
+  newDistroId: String!
+  distroIdToCopy: String!
+}
+
+"""
+Return type representing whether a distro was created and any validation errors
+"""
+type NewDistroPayload {
+  newDistroId: String!
```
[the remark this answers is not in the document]
sophstad: Mohamed linked to [this article](https://www.apollographql.com/blog/graphql/basics/designing-graphql-mutations/#designing-the-mutation-payload) a while back that proposes always using objects for payloads. I think it's a good idea tbh because it facilitates future changes and backwards compatibility.

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
 	}
 	return nil
 }
+
+type CopyDistroOpts struct {
+	DistroIdToCopy string
+	NewDistroId    string
+}
+
+func CopyDistro(ctx context.Context, env evergreen.Environment, u *user.DBUser, opts CopyDistroOpts) error {
```
Kimchelly, line 87: Documentation for exported methods.

rest/data/distro_test.go
```diff
@@ -40,3 +46,106 @@ func TestDeleteDistroById(t *testing.T) {
...
+	} {
+		t.Run(tName, func(t *testing.T) {
+
+			tctx, tcancel := context.WithCancel(context.Background())
+			defer tcancel()
+
+			assert.NoError(t, db.ClearCollections(distro.Collection, event.EventCollection, user.Collection))
+
+			env := &mock.Environment{}
+			require.NoError(t, env.Configure(ctx))
```
Kimchelly, line 114: This should use `tctx`, or else the env may have background processes (such as Amboy) still lingering between tests.

rest/data/distro_test.go
```diff
@@ -40,3 +46,106 @@ func TestDeleteDistroById(t *testing.T) {
...
+			assert.Equal(t, err.Error(), "404 (Not Found): distro 'my-distro' not found")
+
+			events, err := event.FindLatestPrimaryDistroEvents("new-distro", 10)
+			assert.NoError(t, err)
+			require.Equal(t, len(events), 0)
+		},
+	} {
+		t.Run(tName, func(t *testing.T) {
+
+			tctx, tcancel := context.WithCancel(context.Background())
```
Kimchelly, line 108: `context.WithCancel(ctx)` to use the root context of this test.

rest/data/distro_test.go
```diff
@@ -40,3 +46,106 @@ func TestDeleteDistroById(t *testing.T) {
...
+				},
+				PlannerSettings: distro.PlannerSettings{
+					Version: evergreen.PlannerVersionTunable,
+				},
+				Provider: evergreen.ProviderNameStatic,
+				SSHKey:   "abc",
+				WorkDir:  "/tmp",
+				User:     "admin",
+			}
+			assert.NoError(t, d.Insert())
```
Kimchelly, line 141: nit: this uses assert but an equivalent insert on L146 uses require. Can we make it consistent one or the other?

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
...
+		return errors.Wrapf(err, "finding distro '%s'", opts.DistroIdToCopy)
+	}
+	if distroToCopy == nil {
+		return gimlet.ErrorResponse{
+			StatusCode: http.StatusNotFound,
+			Message:    fmt.Sprintf("distro '%s' not found", opts.DistroIdToCopy),
+		}
+	}
+
+	distroToCopy.Id = opts.NewDistroId
```
Kimchelly, line 99: This will be caught by the validator, but it may be preferable if we check that `opts.NewDistroId` is different from `distroToCopy.Id` explicitly, which would let us pass back a 400 error instead of 500 in this case to indicate that it's a user input error.

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
...
+	return newDistro(ctx, env, distroToCopy, u)
+}
+
+func newDistro(ctx context.Context, env evergreen.Environment, d *distro.Distro, u *user.DBUser) error {
+	settings, err := evergreen.GetConfig()
```
Kimchelly, line 104: I'm a bit curious - is there anything that we need from the DB's config here for validation? It seems like `env` was possibly passed in to use `env.Settings()`.

Either way, `env` is currently an unused parameter.

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
 	}
 	return nil
 }
+
+type CopyDistroOpts struct {
```
Kimchelly, line 82: Documentation for exported types.

Kimchelly (changes requested):

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
...
+		return errors.Wrap(err, "getting admin settings")
+	}
+
+	vErrs, err := validator.CheckDistro(ctx, d, settings, true)
+	if err != nil {
+		return errors.Wrapf(err, "validating distro '%s'", d.Id)
+	}
+	if len(vErrs) != 0 {
+		return errors.Errorf("validator encountered errors: '%s'", vErrs.String())
+	}
```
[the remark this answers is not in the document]
sophstad: 500 is the default which is okay here afaik!

rest/data/distro.go
```diff
@@ -74,3 +78,49 @@ func DeleteDistroById(distroId string) error {
...
+
+type CopyDistroOpts struct {
+	DistroIdToCopy string
+	NewDistroId    string
+}
+
+func CopyDistro(ctx context.Context, env evergreen.Environment, u *user.DBUser, opts CopyDistroOpts) error {
+	distroToCopy, err := distro.FindOneId(opts.DistroIdToCopy)
+	if err != nil {
+		return errors.Wrapf(err, "finding distro '%s'", opts.DistroIdToCopy)
```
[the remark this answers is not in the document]
sophstad: The default 500 makes sense here too I think—textbook internal server error

rest/data/distro.go
sophstad: The validator uses the provided `env.Settings()` to check that the distro has valid SSH keys. But good catch, I removed it as a param here.

Kimchelly (approved):

minnakt (approved):

PR #6766, opened 2023-07-12
EVG-20434 Toggle merge queue with project ref

bsamek:
EVG-20434

### Description
This executes the correct merge queue behavior (Evergreen or GitHub) based on
the field in the project ref.

### Testing
I added a unit test for the behavior.

### Documentation
TBD in a later ticket.

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -342,6 +342,10 @@ func getAndEnqueueCommitQueueItemForPR(ctx context.Context, env evergreen.Enviro
 		return nil, pr, errors.Wrapf(errNoCommitQueueForBranch, "repo '%s:%s', branch '%s'", info.Owner, info.Repo, baseBranch)
 	}
 
+	if projectRef.CommitQueue.MergeQueue == model.MergeQueueGitHub {
+		return nil, pr, errors.Wrapf(errors.New("project is using GitHub merge queue"), "repo '%s:%s', branch '%s'", info.Owner, info.Repo, baseBranch)
+	}
+
 	authorized, err := sc.IsAuthorizedToPatchAndMerge(ctx, env.Settings(), NewUserRepoInfo(info))
 	if err != nil {
 		return nil, pr, errors.Wrap(err, "getting user info from GitHub API")
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -272,7 +272,6 @@ func (gh *githubHookApi) handleMergeGroupEvent(event *github.MergeGroupEvent) gi
 			"head_sha": event.GetMergeGroup().GetHeadSHA(),
 			"message":  "no matching project ref",
 		})
-		fmt.Println("org/repo/branch", org, repo, branch)
 		return gimlet.NewJSONInternalErrorResponse(errors.Wrap(err, "no matching project ref"))
 	}
 	if ref.CommitQueue.MergeQueue == model.MergeQueueGitHub {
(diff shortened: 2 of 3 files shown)


rest/data/commit_queue.go
```diff
@@ -342,6 +342,10 @@ func getAndEnqueueCommitQueueItemForPR(ctx context.Context, env evergreen.Enviro
 		return nil, pr, errors.Wrapf(errNoCommitQueueForBranch, "repo '%s:%s', branch '%s'", info.Owner, info.Repo, baseBranch)
 	}
 
+	if projectRef.CommitQueue.MergeQueue == model.MergeQueueGitHub {
+		return nil, pr, errors.Wrapf(errors.New("project is using GitHub merge queue"), "repo '%s:%s', branch '%s'", info.Owner, info.Repo, baseBranch)
+	}
+
```
malikchaya2, line 348: Is there any way for the "Add to commit queue" button  on the patch page to still work for for projects that use the github merge queue (I see this is called by the EnqueuePatch resolver)? 
If not, do we have a ticket that will hide that button when appropriate?

rest/route/github.go
```diff
@@ -39,9 +39,6 @@ const (
 	resetDefinitionsComment = "evergreen reset-definitions"
 
 	refTags = "refs/tags/"
-
-	// This will be removed when EVG-19964 is ready.
-	disableMergeGroup = true
```
malikchaya2, line 44: 🎉

rest/route/github.go
```diff
@@ -230,43 +227,72 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
-			return gimlet.NewJSONInternalErrorResponse(errors.Wrap(err, "adding patch intent"))
-		}
+		return gimlet.NewJSONInternalErrorResponse(errors.Wrap(err, "no matching project ref"))
 	}
-
-	return gimlet.NewJSONResponse(struct{}{})
+	if ref.CommitQueue.MergeQueue == model.MergeQueueGitHub {
+		err = gh.AddIntentForGithubMerge(event)
+	} else {
+		return gimlet.NewJSONResponse(struct{}{})
```
malikchaya2, line 280: What does this use case look like? What buttons would a user click to get them into a state where we are getting MergeGroupEvents but it's not enabled? I'm guessing they enabled it on github but not on evergreen? 
This might be something to think about at a later time, but I'm wondering if there's something we should do to provide feedback to the user letting them know that there is one more step they need to do to complete the setup.

rest/route/github.go
```diff
@@ -230,43 +227,72 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+	}
+	if err != nil {
```
malikchaya2, line 282: Do we want this on line 279?

malikchaya2 (changes requested):

rest/data/commit_queue.go
bsamek: @minnakt mentioned this as well. 

I think users can only add to the GitHub merge queue from the GitHub UI. Disabling the commit queue doesn't hide that button today, so I don't think it's a regression to leave it. I have a note to myself to either open a ticket to hide this or to defer this to later, when/if we remove Evergreen's native commit queue.

rest/route/github.go
bsamek: I think there might be an edge case where a user is using both Evergreen and another CI system. If they prefer to use the other CI system for the merge queue, Evergreen would still receive merge group events, since it would be registered for web hooks, but it shouldn't build them.

rest/route/github.go
bsamek: I think they have the same effect, but I found it easier to read like this. I thought it made it clear there are two branches, and to see the effects of both branches at once.

malikchaya2 (approved):

PR #6770, opened 2023-07-13
EVG-20326 Spawn job in generate.tasks command

hadjri:
EVG-20326

### Description
This change should on average save 30 seconds per `generate.tasks` command because instead of setting `GeneratedJSONString` for a task and waiting up to a minute for a cron job to later pick up on this change, we can just kick off the job in the same call and relegate the usage of the cron to a fallback mechanism that performs cleanup of lost generator jobs

### Testing
Tested in staging and confirmed that the job spawned from the original request is what generated the task and that the subsequent cron-spawned job was created but no-opped.

Updated unit test


units/crons.go
```diff
@@ -718,8 +718,10 @@ func PopulateAgentMonitorDeployJobs(env evergreen.Environment) amboy.QueueOperat
 
 }
 
-// PopulateGenerateTasksJobs populates generate.tasks jobs for tasks that have started running their generate.tasks command.
-func PopulateGenerateTasksJobs(env evergreen.Environment) amboy.QueueOperation {
+// PopulateFallbackGenerateTasksJobs populates generate.tasks jobs for tasks that have started running their generate.tasks command.
+// Since the original generate.tasks request kicks off a job immediately, this function only serves as a fallback in case the
+// original job fails to run. If the original job succeeds, the jobs created here will no-op.
```
Kimchelly, line 723: This comment is mostly right, but the job shouldn't be created if the job already completed (with success or error) - the cron populator onlys scrapes up those tasks that still have to generate tasks, because `GenerateNotRun` evaluates to [false after a job tries to generate and save the generated tasks](https://github.com/evergreen-ci/evergreen/blob/638e37f61359d4a5357669a3ff99de6d170a9af7/units/generate_tasks.go#L299-L303).

rest/route/task_generate.go
```diff
@@ -47,6 +52,29 @@ func (h *generateHandler) Run(ctx context.Context) gimlet.Responder {
...
+		return gimlet.MakeJSONErrorResponder(errors.Errorf("task '%s' not found", h.taskID))
+	}
+	ts := utility.RoundPartOfMinute(1).Format(units.TSFormat)
+	j := units.NewGenerateTasksJob(t.Version, t.Id, ts)
+	queueName := fmt.Sprintf("service.generate.tasks.version.%s", t.Version)
+	queue, err := evergreen.GetEnvironment().RemoteQueueGroup().Get(ctx, queueName)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "getting generate tasks queue '%s' for version '%s'", queueName, t.Version))
+	}
+	grip.Info(message.Fields{
```
Kimchelly, line 69: Do we need this log?

rest/route/task_generate.go
```diff
@@ -47,6 +52,29 @@ func (h *generateHandler) Run(ctx context.Context) gimlet.Responder {
...
+		"message": "enqueueing generate tasks job",
+		"version": t.Version,
+		"task_id": t.Id,
+		"job_id":  j.ID(),
+	})
+	if err = amboy.EnqueueUniqueJob(ctx, queue, j); err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "enqueueing catchup job"))
```
Kimchelly, line 76: `"enqueueing generate tasks job"`

I think it's actually okay to just log a warning and continue, rather than return an error. This could fail for non-problematic reasons, like a race between this and the cron populator to populate the job (since after L52, the cron populator may already have detected the task needs generation). At this point, we've already persisted the generate tasks JSON state to the DB, so this is just a nice-to-have for speeding up the process.

Kimchelly (approved):

Kimchelly:
LGTM mod small suggestions.

rest/route/task_generate.go
```diff
@@ -47,6 +52,29 @@ func (h *generateHandler) Run(ctx context.Context) gimlet.Responder {
...
 		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "generating tasks for task '%s'", h.taskID))
 	}
+	t, err := task.FindOneId(h.taskID)
+	if err != nil {
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "getting task '%s'", h.taskID))
+	}
+	if t == nil {
+		return gimlet.MakeJSONErrorResponder(errors.Errorf("task '%s' not found", h.taskID))
+	}
+	ts := utility.RoundPartOfMinute(1).Format(units.TSFormat)
```
Kimchelly, line 62: There's some duplication between here and the cron populator, so I would turn this into a helper function in `units` that both can just call to make a job for one task.

units/crons.go
hadjri: Yeah, I meant it would no-op if if gets kicked off before the first job finished. Updated the comment

hadjri: Yeah, I meant it would no-op if if gets kicked off before the first job finished. Updated the comment

rest/route/task_generate.go
hadjri: I think it's nice to have but should probably move it to after we successfully enqueue, that being said if you think it's too noisy I can remove.

units/crons.go
Kimchelly: The job supposedly shouldn't be created if it's already in the queue though because there's a scope on the job (which prevents duplicate jobs for the same task from being in the queue simultaneously). Did you see a case in your tests that indicated that there were two jobs in the queue for the same task at the same time?

rest/route/task_generate.go
Kimchelly: Eh, I don't have strong feelings, it just seems like it's a duplicate since we could trace that it succeeded based on the REST request started/completed logs that already go to Splunk (and the presence/absence of Amboy job logs).

units/generate_tasks.go
```diff
@@ -320,6 +320,17 @@ func (j *generateTasksJob) Run(ctx context.Context) {
...
+// GetGenerateTasksJobAndQueue returns a job and generate.tasks queue for the given task.
+func GetGenerateTasksJobAndQueue(ctx context.Context, env evergreen.Environment, t task.Task, ts string) (amboy.Job, amboy.Queue, error) {
+	j := NewGenerateTasksJob(t.Version, t.Id, ts)
+	queueName := fmt.Sprintf("service.generate.tasks.version.%s", t.Version)
+	queue, err := env.RemoteQueueGroup().Get(ctx, queueName)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "getting generate tasks queue '%s' for version '%s'", queueName, t.Version)
+	}
+	return j, queue, nil
+}
```
Kimchelly, line 332: You can absorb the `amboy.EnqueueUniqueJob` into this function as well, since the REST request logic should continue on any returned error, and only the cron populator cares to accumulate the error.

rest/route/task_generate.go
```diff
@@ -47,6 +50,23 @@ func (h *generateHandler) Run(ctx context.Context) gimlet.Responder {
...
+	j, err := units.CreateAndEnqueueGenerateTasks(ctx, evergreen.GetEnvironment(), *t, ts)
+	if err != nil {
```
Kimchelly, line 62: nit: you can omit the `if err != nil` since `message.WrapError` is smart enough to suppress the log when it's a nil error.

Kimchelly (approved):

PR #6771, opened 2023-07-13
EVG-20286: continue heartbeating after abort

Kimchelly:
EVG-20286

### Description
If the task gets an abort signal from the heartbeat, the agent is still expected to run the post task, so it should still heartbeat back to the app server to indicate that the task is alive. The heartbeat will continue to receive the abort message from the app server on later heartbeats, but the agent only needs to process the abort message once.

* Continue attempting to heartbeat after a task receives the abort signal.
* Drive-by clean up some existing agent tests.

### Testing
Added unit tests and tested in staging to verify that a long sleep in post continued running after the heartbeat timeout ([example](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_test_util_patch_b6222da4640b0199bb11a37c93f6b8192ff473dc_64aef970b2373697b2ba2c69_23_07_12_19_05_21/logs?execution=1&logtype=agent)).

### Documentation
Updated docs to better describe how pre/main/post interact with abort.

I also drive-by clarified a doc that's a little ambiguous about the precedence of YAML fields.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -62,6 +62,7 @@ func (s *AgentSuite) SetupTest() {
 	const bvName = "some_build_variant"
 	tsk := &task.Task{
 		Id:           "task_id",
+		DisplayName:  "some task",
 		BuildVariant: bvName,
 		Version:      versionID,
 	}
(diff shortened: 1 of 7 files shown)


agent/background.go
```diff
@@ -23,10 +23,14 @@ func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, t
 	ticker := time.NewTicker(heartbeatInterval)
 	defer ticker.Stop()
 
+	var hasSentHeartbeat bool
 	for {
 		select {
 		case <-ticker.C:
 			signalBeat, err = a.doHeartbeat(ctx, tc)
+			if hasSentHeartbeat {
```
hadjri, line 31: We should move this [block](https://github.com/evergreen-ci/evergreen/pull/6771/files#diff-569d49a4c046eb603023b873eea7110382401df292e83230a95862469c2029f9R46-R56) before we continue in case the err is not nil right?

hadjri (changes requested):

agent/agent_test.go
```diff
@@ -1091,40 +1052,88 @@ func (s *AgentSuite) TestFetchProjectConfig() {
...
+
+tasks:
+- name: this_is_a_task_name
+  commands:
+   - command: shell.exec
+     params:
+       script: sleep 5
+
+post:
+- command: shell.exec
```
malikchaya2, line 1071: For the sake of completeness, I think it would be a good idea to add a test that makes sure that this also works for a teardown task without a post block.

malikchaya2 (approved):

hadjri (approved):

PR #6773, opened 2023-07-13
EVG-19963  Add option to Patch-file to fetch diff and modules from patch ID 

malikchaya2:
EVG-19963

### Description
This adds an option to patch-file to specify a patchId to fetch the diff and module diffs from. 

### Testing
tested on staging. [This patch](https://spruce-staging.corp.mongodb.com/version/64b078459dbe321b6be4a496/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) was created from this [patch](https://spruce-staging.corp.mongodb.com/version/64b077049dbe321b6be4a48b/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) using the command evergreen patch-file --diff-from-patch 64b077049dbe321b6be4a48b.

### Documentation
added


operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
 	return reply.Patch, nil
 }
 
+func (ac *legacyClient) GetRawPatchWithModules(patchId string) (data.RawPatch, error) {
```
Kimchelly, line 609: nit: should this return a `*data.RawPatch`? That way, you can return nil in case of error.

rest/data/patch.go
```diff
@@ -197,35 +197,56 @@ func AbortPatchesFromPullRequest(ctx context.Context, event *github.PullRequestE
...
+}
+type RawModule struct {
+	Name    string `json:"name"`
+	Diff    string `json:"diff"`
+	Githash string `json:"githash"`
+}
+
 // GetPatchRawPatches fetches the raw patches for a patch
-func GetPatchRawPatches(patchID string) (map[string]string, error) {
+func GetPatchRawPatches(patchID string) (RawPatch, error) {
```
Kimchelly, line 211: nit: Seems like you could just call this `GetRawPatches` since it's already fairly clear from the `patchID` parameter name that it's for that specific patch.

rest/data/patch.go
```diff
@@ -197,35 +197,56 @@ func AbortPatchesFromPullRequest(ctx context.Context, event *github.PullRequestE
...
+	Patch      RawModule   `json:"patch"`
+	RawModules []RawModule `json:"raw_modules"`
+}
+type RawModule struct {
+	Name    string `json:"name"`
+	Diff    string `json:"diff"`
+	Githash string `json:"githash"`
+}
+
 // GetPatchRawPatches fetches the raw patches for a patch
```
Kimchelly, line 210: nit: I know this line wasn't changed, but period at the end of doc comment sentence.

rest/route/patch_test.go
```diff
@@ -574,6 +574,63 @@ func (s *PatchesByUserSuite) TestEmptyTimeShouldSetNow() {
...
+			},
+		},
+	}
+	assert.NoError(t, patchToInsert.Insert())
+
+	route := &moduleRawHandler{
+		patchID: patchId,
+	}
+
+	response := route.Run(context.Background())
```
Kimchelly, line 619: `context.WithCancel(context.Background())` or `context.WithTimeout` to ensure that operations stop when the test returns.

rest/route/patch_test.go
```diff
@@ -574,6 +574,63 @@ func (s *PatchesByUserSuite) TestEmptyTimeShouldSetNow() {
...
+
+	rawModulesResponse := response.Data().(data.RawPatch)
```
Kimchelly, line 621: This can panic, so I would put a [comma-ok](https://dev.to/saurabh975/comma-ok-in-go-l4f) here and check that the type assertion is okay before proceeding.

operations/patch.go
```diff
@@ -344,23 +363,50 @@ func PatchFile() cli.Command {
...
 				return err
 			}
-			newPatch, err := params.createPatch(ac, diffData)
+			newPatch, err := params.createPatch(ac, &diffData)
 			if err != nil {
 				return err
 			}
+
+			if len(rp.RawModules) > 0 {
+
```
Kimchelly, line 393: nit: remove random newline.

operations/patch.go
```diff
@@ -292,10 +298,21 @@ func PatchFile() cli.Command {
...
+			println("Hi 307")
+
+			diffFromPatch := c.String(diffFromPatchFlagName)
+			if diffFromPatch == "" {
+				path := c.String(diffPathFlagName)
+				if _, err := os.Stat(path); os.IsNotExist(err) {
+					return errors.Errorf("file '%s' does not exist %s ", path, diffFromPatch)
+				}
+			}
+			println("Hi 316")
```
Kimchelly, line 315: Remove debug print.

operations/patch.go
```diff
@@ -311,6 +328,7 @@ func PatchFile() cli.Command {
 				SyncTasks:       utility.SplitCommas(c.StringSlice(syncTasksFlagName)),
 				PatchAuthor:     c.String(patchAuthorFlag),
 			}
+			println("Hi 332")
```
Kimchelly, line 331: Remove debug print.

operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		return rp, NewAuthError(resp)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return rp, NewAPIError(resp)
+	}
+
+	err = utility.ReadJSON(resp.Body, &rp)
```
Kimchelly, line 627: nit: you can combine L627-628 into one `if err :=`.

operations/patch.go
```diff
@@ -292,10 +298,21 @@ func PatchFile() cli.Command {
...
+			mutuallyExclusiveArgs(false, baseFlagName, diffFromPatchFlagName),
 		),
 		Action: func(c *cli.Context) error {
+			println("Hi 307")
+
+			diffFromPatch := c.String(diffFromPatchFlagName)
+			if diffFromPatch == "" {
+				path := c.String(diffPathFlagName)
+				if _, err := os.Stat(path); os.IsNotExist(err) {
+					return errors.Errorf("file '%s' does not exist %s ", path, diffFromPatch)
```
Kimchelly, line 312: I don't quite understand what this error is saying. Is it saying that both the requested diff path does not exist and there is no patch to diff from? Can we rephrase this?

operations/patch.go
```diff
@@ -344,23 +363,50 @@ func PatchFile() cli.Command {
...
+
+				for _, module := range rp.RawModules {
+					moduleParams := UpdatePatchModuleParams{
+						patchID: newPatch.Id.Hex(),
+						module:  module.Name,
+						patch:   module.Diff,
+						base:    module.Githash,
+					}
+					//todo: write a route to take in all module params
+					err = ac.UpdatePatchModule(moduleParams)
```
Kimchelly, line 402: nit: combine L402-403.

rest/data/patch.go
Kimchelly, line 211: nit: seems like returning a pointer is more idiomatic with our code base so that we can return nil in the case of error (same with the other functions in this PR).

rest/data/patch.go
```diff
@@ -197,35 +197,56 @@ func AbortPatchesFromPullRequest(ctx context.Context, event *github.PullRequestE
...
-		patchMap[raw.ModuleName] = raw.PatchSet.Patch
+		module := RawModule{
+			Name:    raw.ModuleName,
+			Diff:    raw.PatchSet.Patch,
+			Githash: raw.Githash,
+		}
+		if raw.ModuleName == "" {
+			rawPatch.Patch = module
+		} else {
+			rawModules = append(rawModules, module)
```
Kimchelly, line 244: nit: since you already have a `rawPatch` here, you can get rid of the `rawModules` variable entirely and just append straight to `rawPatch.RawModules`.

Kimchelly (changes requested):
I'm trying to understand the original request described in EVG-19963. Which of these is the user asking for?
1. A command that will take an existing patch and make a copy of that patch + its modules (i.e. this PR's change)
2. A command to submit a patch from a local diff file on that user's machine, then apply particular module diff files on top of that patch diff?
3. A command to submit a patch from a local diff on that user's machine, then apply modules from an existing patch to it?

It kind of sounds like they're asking for 2 instead of 1, but I may be missing some context. Regardless of which one it is, allowing patches to be repeated with modules (i.e. the first one) sounds like something that may be useful, but may be more appropriate to absorb into the `evergreen patch --repeat-patch` command instead of adding to `evergreen patch-file`.

Kimchelly:
Resolved the main comment above in Slack - the user request is described more clearly in [this Jira comment](https://jira.mongodb.org/browse/EVG-19622?focusedCommentId=5405104&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5405104). Basically, they want the existing patch raw route to include modules, and have a way to reuse a patch's diff, but with different variants/tasks (basically the opposite of `evergreen patch --reuse-patch`, which reuses the patch's variants/tasks but with the local diff of changes).

operations/http.go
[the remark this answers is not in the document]
malikchaya2: done

operations/http.go
malikchaya2: done

rest/route/patch.go
```diff
@@ -142,12 +142,54 @@ func (p *patchRawHandler) Parse(ctx context.Context, r *http.Request) error {
...
+func (p *moduleRawHandler) Parse(ctx context.Context, r *http.Request) error {
+	p.patchID = gimlet.GetVars(r)["patch_id"]
+	return nil
+}
+
+func (p *moduleRawHandler) Run(ctx context.Context) gimlet.Responder {
+
+	rawPatches, err := data.GetPatchRawPatches(p.patchID)
+	if err != nil {
+		return gimlet.MakeJSONErrorResponder(errors.Wrapf(err, "getting raw patches for patch '%s'", p.patchID))
```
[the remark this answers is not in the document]
malikchaya2: done

rest/route/patch_test.go
malikchaya2: done

rest/route/patch_test.go
malikchaya2: done

rest/data/patch.go
malikchaya2: done

rest/route/patch_test.go
```diff
@@ -574,6 +574,63 @@ func (s *PatchesByUserSuite) TestEmptyTimeShouldSetNow() {
...
+func TestPatchRawModulesHandler(t *testing.T) {
+	require.NoError(t, db.ClearCollections(patch.Collection))
+	require.NoError(t, db.ClearGridCollections(patch.GridFSPrefix))
+	patchString := `main diff`
+	require.NoError(t, db.WriteGridFile(patch.GridFSPrefix, "testPatch", strings.NewReader(patchString)))
+	patchString = `module1 diff`
+	require.NoError(t, db.WriteGridFile(patch.GridFSPrefix, "module1Patch", strings.NewReader(patchString)))
+	patchString = `module2 diff`
+	require.NoError(t, db.WriteGridFile(patch.GridFSPrefix, "module2Patch", strings.NewReader(patchString)))
+	patchId := "aabbccddeeff001122334455"
```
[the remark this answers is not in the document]
malikchaya2: done

operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
...
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		return &rp, NewAuthError(resp)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return &rp, NewAPIError(resp)
+	}
+
+	if err = utility.ReadJSON(resp.Body, &rp); err != nil {
+		return &rp, err
```
Kimchelly, line 629: Seems like most of the error cases in this function should return nil. It's only the last return where you actually succeeded that you would return the actual results.

operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
 	return reply.Patch, nil
 }
 
+// GetRawPatchWithModules fetches the raw patch and module diffs for a given patch ID
+func (ac *legacyClient) GetRawPatchWithModules(patchId string) (*data.RawPatch, error) {
+
```
Kimchelly, line 611: nit: remove random newline.

operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
 	return reply.Patch, nil
 }
 
+// GetRawPatchWithModules fetches the raw patch and module diffs for a given patch ID
+func (ac *legacyClient) GetRawPatchWithModules(patchId string) (*data.RawPatch, error) {
```
Kimchelly, line 610: I didn't notice this on the previous code review, but it's better to not add any more new routes to the legacy REST client because it's deprecated. Methods should instead be added to the new REST communicator, so we should move this method over.

rest/data/patch.go
Kimchelly: I get it, but the line is the same as it was before 😅

operations/http.go
```diff
@@ -605,6 +606,31 @@ func (ac *legacyClient) PutPatch(incomingPatch patchSubmission) (*patch.Patch, e
 	return reply.Patch, nil
 }
 
+// GetRawPatchWithModules fetches the raw patch and module diffs for a given patch ID
```
Kimchelly, line 609: nit: period at the end of doc comment sentence.

rest/data/patch.go
```diff
@@ -197,35 +197,57 @@ func AbortPatchesFromPullRequest(ctx context.Context, event *github.PullRequestE
...
+	Githash string `json:"githash"`
+}
+
+// GetRawPatches fetches the raw patches for a patch.
+func GetRawPatches(patchID string) (*RawPatch, error) {
+	var rawPatch RawPatch
 	patchDoc, err := patch.FindOneId(patchID)
 	if err != nil {
-		return nil, gimlet.ErrorResponse{
+		return &rawPatch, gimlet.ErrorResponse{
```
Kimchelly, line 218: Same comment as in the other place - return nil in the error cases is more idiomatic, unless somehow a pointer to an empty struct has some special meaning that's distinct from nil.

Kimchelly (changes requested):

rest/data/patch.go
malikchaya2: I think it's now "// GetRawPatches fetches the raw patches for a patch."

operations/http.go
malikchaya2: done

rest/model/patch.go
```diff
@@ -94,6 +94,19 @@ type APIParameter struct {
 	Value *string `json:"value"`
 }
 
+// RawPatch contains a patch diff along with its module diffs.
```
Kimchelly, line 97: `APIRawPatch`

rest/model/patch.go
```diff
@@ -94,6 +94,19 @@ type APIParameter struct {
...
+type APIRawPatch struct {
+	Patch      APIRawModule   `json:"patch"`
+	RawModules []APIRawModule `json:"raw_modules"`
+}
+
+// RawModule contains a module diff.
```
Kimchelly, line 103: `APIRawModule`

Kimchelly (approved):
LGTM mod small doc comment renames.

PR #6777, opened 2023-07-14
EVG-20016: New admin bucket settings

julianedwards:
EVG-20016

### Description
Admin config for application-wide bucket storage.

### Testing
Added unit tests and tested that the admin page works in staging.

### Documentation
N/A


config_bucket.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (c *BucketConfig) Get(env Environment) error {
+	ctx, cancel := env.Context()
+	defer cancel()
+	coll := env.DB().Collection(ConfigCollection)
+
+	res := coll.FindOne(ctx, byId(c.SectionId()))
+	if err := res.Err(); err != nil {
+		if err == mongo.ErrNoDocuments {
+			fmt.Println("HERE")
```
ybrill, line 39: :)

config.go
```diff
@@ -66,6 +66,7 @@ type Settings struct {
 	AWSInstanceRole     string                  `yaml:"aws_instance_role" bson:"aws_instance_role" json:"aws_instance_role"`
 	Banner              string                  `bson:"banner" json:"banner" yaml:"banner"`
 	BannerTheme         BannerTheme             `bson:"banner_theme" json:"banner_theme" yaml:"banner_theme"`
+	Bucket              BucketConfig            `bson:"bucket" json:"bucket" yaml:"bucket" id:"bucket"`
```
ybrill, line 69: I would think to call it something like LoggingBucket or something to make it plain what the bucket is for.

rest/model/admin_test.go
```diff
@@ -122,6 +122,8 @@ func TestModelConversion(t *testing.T) {
 	assert.EqualValues(testSettings.AuthConfig.Github.ClientId, utility.FromStringPtr(apiSettings.AuthConfig.Github.ClientId))
 	assert.EqualValues(testSettings.AuthConfig.Multi.ReadWrite[0], apiSettings.AuthConfig.Multi.ReadWrite[0])
 	assert.Equal(len(testSettings.AuthConfig.Github.Users), len(apiSettings.AuthConfig.Github.Users))
+	assert.EqualValues(testSettings.Bucket.LogBucket, utility.FromStringPtr(apiSettings.Bucket.LogBucket))
```
ybrill, line 125: should we also test the data level ([here](https://github.com/evergreen-ci/evergreen/blob/276f46116f490f1fda52987e5c32cc5e61ab8151/rest/data/admin_test.go#L88))? and the other direction ([here](https://github.com/evergreen-ci/evergreen/blob/2ae72ff6826e8ad21e25fd862cb3531d7e273f16/rest/model/admin_test.go#L262))?

config_bucket.go
julianedwards: Ooof... good catch!

config.go
julianedwards: The idea is that we can set up all of our buckets with this single config (say, in the future, we may want to add the test results bucket here).

rest/model/admin_test.go
julianedwards: I added tests going in the other direction (for the cedar and data pipes configs as well). As for the test at the data level, that may be unnecessary since the test itself just does a spot check on the values—the more thorough testing is done here.

config.go
ybrill: Ah, I understand.
Maybe `Buckets`, then?

config_bucket.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+const (
+	BucketTypeLocal = "local"
+	BucketTypeS3    = "s3"
+)
+
+// BucketConfig represents the admin config section for bucket storage.
+type BucketConfig struct {
+	LogBucket     string `bson:"log_bucket" json:"log_bucket" yaml:"log_bucket"`
```
ybrill, line 18: could it make sense to make the bucket config a struct of structs? Like
```go
type BucketConfig struct {
    LoggingBucket Bucket
}

type Bucket struct {
    Name string
    Type string
}
```
(to facilitate adding more buckets)

rest/model/admin.go
```diff
@@ -687,6 +689,29 @@ func (a *APIAuthConfig) ToService() (interface{}, error) {
 	}, nil
 }
 
+type APIBucketConfig struct {
+	LogBucket     *string `json:"log_bucket"`
+	LogBucketType *string `json:"log_bucket_type"`
```
ybrill, line 694: do we need to pass through the bucket type? Is the type ever configured outside of testing?

config_bucket.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+var (
+	bucketConfigLogBucketKey     = bsonutil.MustHaveTag(BucketConfig{}, "LogBucket")
+	bucketConfigLogBucketTypeKey = bsonutil.MustHaveTag(BucketConfig{}, "LogBucketType")
+)
+
+func (*BucketConfig) SectionId() string { return "bucket" }
+
+func (c *BucketConfig) Get(env Environment) error {
```
ybrill, line 29: I changed the interface for Get/Set in my gargantuan PR (https://github.com/evergreen-ci/evergreen/pull/6732).
It should be there if you merge main.

rest/model/admin.go
julianedwards: I think it is worth it for local development.

config_bucket.go
julianedwards: Thanks for the head's up!

config.go
```diff
@@ -66,6 +66,7 @@ type Settings struct {
 	AWSInstanceRole     string                  `yaml:"aws_instance_role" bson:"aws_instance_role" json:"aws_instance_role"`
 	Banner              string                  `bson:"banner" json:"banner" yaml:"banner"`
 	BannerTheme         BannerTheme             `bson:"banner_theme" json:"banner_theme" yaml:"banner_theme"`
+	Buckets             BucketConfig            `bson:"buckets" json:"buckets" yaml:"bucket" id:"buckets"`
```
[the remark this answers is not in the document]
julianedwards: Thanks!

ybrill (approved):

PR #6778, opened 2023-07-14
EVG-20456 Add owner/repo field to all patches 

bynn:
EVG-20456

### Description
GitHub app was failing because owner/repo field wasn't set for all types of patches 

### Testing
this was caught in a unit test when i return an error that is currently being suppressed by fallback
the unit test is correctly passing now

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -169,16 +169,6 @@ func (j *patchIntentProcessor) Run(ctx context.Context) {
 }
 
 func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.Patch) error {
-	// // Add owner and repo for child patches
-	// if patchDoc.IsChild() {
-	// 	p, err := model.FindBranchProjectRef(patchDoc.Project)
-	// 	if err != nil {
-	// 		return errors.Wrapf(err, "finding project '%s'", patchDoc.Project)
-	// 	}
-	// 	patchDoc.GithubPatchData.BaseOwner = p.Owner
-	// 	patchDoc.GithubPatchData.BaseRepo = p.Repo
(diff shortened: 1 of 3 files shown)


units/patch_intent.go
```diff
@@ -159,6 +159,16 @@ func (j *patchIntentProcessor) Run(ctx context.Context) {
 }
 
 func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.Patch) error {
+	// Add owner and repo for child patches
+	if patchDoc.Author == evergreen.ParentPatchUser {
```
malikchaya2, line 163: Should we maybe use p.IsParent() here?

malikchaya2 (changes requested):

units/patch_intent.go
bynn: is think you meant isChild but yes! i think thats way better

units/patch_intent.go
```diff
@@ -159,6 +159,16 @@ func (j *patchIntentProcessor) Run(ctx context.Context) {
 }
 
 func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.Patch) error {
+	// Add owner and repo for child patches
+	if patchDoc.IsChild() {
```
malikchaya2, line 163: We should probably move this logic to where the patch intent is created for child patches

malikchaya2: looking at the code, FinalizePatch might be a good place for it since we already have an isChild check and access to the project.

bynn: finalize patch wouldnt hit certain use cases i think like backport but i moved it out to running patch intent processor

PR #6780, opened 2023-07-14
EVG-20459 Lookup project id for RequireProjectAccess directive variables

khelif96:
EVG-20459

### Description
The RequireProjectAccess would take in 3 different argument types and look up the project id in only one of the cases. This could be problematic due to the differences between how ids vs identifiers are treated as seen in this thread: https://mongodb.slack.com/archives/C0V896UV8/p1689352076642169

At some point, we should consolidate the argument types for these mutations so that they are consistent in the argument naming.

### DIFF
--- a/graphql/util.go
+++ b/graphql/util.go
@@ -762,12 +762,9 @@ func groupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
 
 // getProjectIdFromArgs extracts a project ID from the requireProjectAccess directive args.
 func getProjectIdFromArgs(ctx context.Context, args map[string]interface{}) (res string, err error) {
+	// id should always be a repo ID.
 	if id, hasId := args["id"].(string); hasId {
-		pid, err := model.GetIdForProject(id)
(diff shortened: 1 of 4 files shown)


sophstad (approved):

graphql/util.go
```diff
@@ -762,11 +762,16 @@ func groupProjects(projects []model.ProjectRef, onlyDefaultedToRepo bool) ([]*Gr
 
 // getProjectIdFromArgs extracts a project ID from the requireProjectAccess directive args.
 func getProjectIdFromArgs(ctx context.Context, args map[string]interface{}) (res string, err error) {
+	// id should always be a repo ID.
 	if id, hasId := args["id"].(string); hasId {
 		return id, nil
 	}
 	if projectId, hasProjectId := args["projectId"].(string); hasProjectId {
-		return projectId, nil
+		pid, err := model.GetIdForProject(projectId)
```
julianedwards, line 770: Nit: you could combine this logic with that of the following if statement.

julianedwards (approved):

PR #6781, opened 2023-07-14
EVG-20311 Remove agent loop labels

malikchaya2:
EVG-20311

### Description
This removes the LOOP label because just continue keyword does the same thing.

### DIFF
--- a/config.go
+++ b/config.go
@@ -38,7 +38,7 @@ var (
 	ClientVersion = "2023-07-06"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-07-12b"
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -38,7 +38,7 @@ var (
 	ClientVersion = "2023-07-06"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-07-12b"
+	AgentVersion = "2023-07-14a"
```
Kimchelly, line 41: nit: by convention, just `2023-07-14` when there's no additional change made on that date.

Kimchelly (approved):
LGTM mod nit.

PR #6786, opened 2023-07-17
EVG-20286: lock mock heartbeat count

Kimchelly:
EVG-20286

### Description
The test had a race to access the mock `HeartbeatCount` because the heartbeat goroutine was still going by the time the test assertions started checking the count.

### Testing
Race detector passed the test locally and I configured the PR patch to run the race detector.

### Documentation
N/A


malikchaya2 (approved):

Kimchelly:
`verify-agent-version-update` is falsely failing because I modified the mock communicator.

PR #6788, opened 2023-07-17
EVG-20310 Add initialization to the agent loop

malikchaya2:
EVG-20310

### Description
This initializes the agent loop in its own function and stores all the values it needs in a struct. This cleans up the function and will make it easier to break up the function more. 

### Testing
existing tests

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -256,7 +256,19 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 	a.ec2InstanceID = instanceID
 }
 
-func (a *Agent) loop(ctx context.Context) error {
+type loop struct {
+	tc                    *taskContext
+	needPostGroup         bool
+	jitteredSleep         time.Duration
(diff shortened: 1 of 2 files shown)


Kimchelly (changes requested):
It seems like this doesn't really change how readable/maintainable the agent loop is, it just moves the variables from function variables into struct fields. Is there something more you plan to do with this? If not, I'm inclined to say we should just leave it how it originally was.

malikchaya2:
> It seems like this doesn't really change how readable/maintainable the agent loop is, it just moves the variables from function variables into struct fields. Is there something more you plan to do with this? If not, I'm inclined to say we should just leave it how it originally was.

yes, it makes it easier to do this: https://github.com/evergreen-ci/evergreen/pull/6791

Kimchelly:
In that case, it seems like this PR can just be closed and absorbed into #6791 since it's already included in the PR where it's used.

PR #6790, opened 2023-07-18
EVG-19918 get only scheduler config and get it only once

ybrill:
[EVG-19918](https://jira.mongodb.org/browse/EVG-19918)

### Description
The OTel traces revealed that the job was slow because we were calling GetConfig in a loop (once for each idle host). Since GetConfig makes ~30 synchronous round trips to the db this was adding up.

### Testing
I'll take it for a spin on staging before I merge it to make sure the job is not broken.


ybrill:
This checks out on staging.

julianedwards (approved):
Nice catch/optimization!

PR #115, opened 2017-09-12
EVG-2035: remove legacy agent code

tychoish:

PR #6791, opened 2023-07-18
EVG-20312 Create a processNextTask function

malikchaya2:
EVG-20312

### Description
The loop did a lot of processing of the nextTask. This has grown a lot since it was first introduced [here](https://github.com/evergreen-ci/evergreen/pull/115/files). To keep the loop closer to just looping logic, this moves a lot of it out into its own processNextTaskFunction.

### Testing
existing tests

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -256,7 +256,19 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 	a.ec2InstanceID = instanceID
 }
 
-func (a *Agent) loop(ctx context.Context) error {
+type loop struct {
+	tc                    *taskContext
+	needPostGroup         bool
+	jitteredSleep         time.Duration
+	taskCtx               context.Context
+	cancel                context.CancelFunc
+	timer                 *time.Timer
+	minAgentSleepInterval time.Duration
+	maxAgentSleepInterval time.Duration
+	agentSleepInterval    time.Duration
+}
+
+func (a *Agent) initLoop(ctx context.Context) *loop {
 	minAgentSleepInterval := defaultAgentSleepInterval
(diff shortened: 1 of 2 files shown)


hadjri (approved):

agent/agent.go
```diff
@@ -265,20 +277,102 @@ func (a *Agent) loop(ctx context.Context) error {
...
+	}
+
+	return lc
+}
+
+func (a *Agent) processNextTask(ctx context.Context, l *loop, nextTask *apimodels.NextTaskResponse) (shouldContinue bool, shouldExit bool) {
+	if nextTask.ShouldExit {
+		grip.Notice("Next task response indicates agent should exit.")
+		shouldExit = true
+		return
```
Kimchelly, line 300: It's considered better practice to explicitly return the values rather than rely on setting the output variable names. This is because it becomes difficult to trace what's true vs. false if you have to track the flow of variable logic. (Same with the other places in this function.)

agent/agent.go
```diff
@@ -265,20 +277,102 @@ func (a *Agent) loop(ctx context.Context) error {
...
+		// Running the post group commands implies exiting the group, so
+		// destroy prior task information.
+		l.tc = &taskContext{}
+	}
+	return
+}
 
-	timer := time.NewTimer(0)
-	defer timer.Stop()
+func (a *Agent) loop(ctx context.Context) error {
```
Kimchelly, line 368: I actually don't see this loop as much cleaner than how it was written before - the original desire was to take the complex part (the `processNextTask` logic) and turn that into a helper function, not the loop itself. This change makes things more stateful than they were before, because now you have to track the `loop` parameters as the agent iterates.

Kimchelly (changes requested):

PR #6792, opened 2023-07-18
EVG-20477 Remove no longer necessary legacy build variant display name lookup

khelif96:
EVG-20477

### Description
Missed removing this one. We no longer need to look up the build variant display name the legacy way because of the TTL

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -1216,7 +1216,6 @@ func FindUniqueBuildVariantNamesByTask(projectId string, taskName string, repoOr
 			{RevisionOrderNumberKey: bson.M{"$lte": repoOrderNumber}},
 		},
 	}
-	query[BuildVariantDisplayNameKey] = bson.M{"$exists": true, "$ne": ""}
 	pipeline := []bson.M{{"$match": query}}
 
 	// group the build variants by unique build variant names and get a build id for each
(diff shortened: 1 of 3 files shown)


model/task/db.go
```diff
@@ -1232,9 +1216,7 @@ func FindUniqueBuildVariantNamesByTask(projectId string, taskName string, repoOr
 			{RevisionOrderNumberKey: bson.M{"$lte": repoOrderNumber}},
 		},
 	}
-	if !legacyLookup {
-		query[BuildVariantDisplayNameKey] = bson.M{"$exists": true, "$ne": ""}
-	}
+	query[BuildVariantDisplayNameKey] = bson.M{"$exists": true, "$ne": ""}
```
Kimchelly, line 1219: nit: seems like you could absorb this into the `query` initialization right above this since it's not conditional anymore.

model/task_build_variant_names_test.go
```diff
@@ -13,11 +13,6 @@ import (
 
 func TestFindUniqueBuildVariantNamesByTask(t *testing.T) {
 	assert.NoError(t, db.ClearCollections(task.Collection, build.Collection))
```
Kimchelly, line 15: Seems like this no longer needs to clear the build collection since it doesn't rely on it.

Kimchelly (approved):
LGTM mod nits.

model/task/db.go
khelif96: Its no longer necessary so I removed it

PR #6793, opened 2023-07-18
EVG-20300: submit manual patch for host smoke test

Kimchelly:
EVG-20300

### Description
Instead of relying on the flaky interaction between the smoke test setup fixture and the repotracker, submit a manual patch with the CLI to test whether host tasks can run. This has some benefits like no longer having the smoke test depend on the state of the `agent.yml` file committed to Evergreen's main branch, not having to worry about container task failures due to using up unauthenticated GitHub API rate, and testing the standard patching lifecycle better.

There's going to be a follow-up PR to organize the smoke test files better, since they're all scattered in random places currently. For now, I'm leaving the files where they are so I can reduce the amount of changes introduced in this PR.

As an aside, the smoke test _can_ be tested locally now, but it's quite difficult to set up. Running patches in Evergreen is probably still the best avenue for debugging the smoke test.

* Change the smoke test to submit a manual patch rather than rely on mainline commit versions.
* Small code/doc improvements and improvements to make the smoke test more readable and documented.
* Quickfix an issue described in EVG-20474 where the permissions on the legacy route to submit patches don't actually work. EVG-20474 will address this issue better, but for now, it's better to remove the (currently non-working) patch submit permission middleware.

### Testing
Ran the smoke tests and they passed.

### Documentation
I wrote an ADR.

I'm going to document the modified smoke test in the dev guide after the follow-up PR.


docs/decisions/2023-07-18_smoke_test_patch.md
```diff
@@ -0,0 +1,26 @@
+# 2023-07-18 Evergreen Smoke Test with Manual Patch
```
malikchaya2, line 1: 🎉

operations/service_deploy_smoke.go
```diff
@@ -264,29 +268,55 @@ func smokeTestEndpoints() cli.Command {
...
 				Name:  userNameFlagName,
-				Usage: "username to use with the API",
+				Usage: "username to use with the Evergreen API",
+				Value: "admin",
 			},
 			cli.StringFlag{
 				Name:  userKeyFlagName,
-				Usage: "key to use with the API",
+				Usage: "key to use with the Evergreen API",
+				Value: "abb623665fdbf368a1db980dde6ee0f0",
```
malikchaya2, line 299: Are there any concerns with this being in a public repo? Can this be used to run other evergreen operations by someone who shouldn't?

operations/service_deploy_smoke_utils.go
```diff
@@ -178,31 +168,131 @@ func triggerRepotracker(username, key string, client *http.Client) error {
...
+	}
+
+	return errors.Errorf("timed out waiting for repotracker to pick up new commits and create versions after %d attempts", repotrackerAttempts)
+}
+
+// submitSmokeTestPatch submits a manual patch to the app server to run the
+// smoke test and selects tasks in the given build variant.
+// Note that this requires using the CLI because there's currently no way to
+// create a patch from the REST API.
+func submitSmokeTestPatch(cliPath, cliConfigPath, bvToRun string) error {
```
malikchaya2, line 221: Should we maybe eventually create a cli command option like --smoke-test for manual testing purposes?

malikchaya2 (changes requested):
Once this is done, we should also update the smoke test section in our internal docs.

operations/service_deploy_smoke.go
Kimchelly: This isn't a secret because [it's the publicly visible admin secret](https://github.com/evergreen-ci/evergreen/blob/1039fe8ff1a85cfb68346ab3df25faa929e84a90/testdata/smoke/users.json#L1) used to populate the smoke test app server's data. This isn't insecure unless you open up your local Evergreen servers to public traffic (which you should never do since local Evergreen is not secure).

operations/service_deploy_smoke_utils.go
Kimchelly: I don't think this is possible because the main difficult with running the smoke test locally is populating secrets.

scripts/agent-cli.yml
```diff
@@ -0,0 +1,5 @@
+api_server_host: http://localhost:9090/api
+ui_server_host: http://localhost:9090
+api_key: abb623665fdbf368a1db980dde6ee0f0
```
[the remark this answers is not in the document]
Kimchelly: Same as above.

operations/service_deploy_smoke_utils.go
malikchaya2: What I meant is to create an "alias" that allows people to easily submit a smoke test patch to evergreen.

malikchaya2 (approved):
I have an open question but it's for future work, so approving.

operations/service_deploy_smoke_utils.go
Kimchelly: Done.

Kimchelly:
I already mentioned in the description that the follow-up PR will update the docs, going to merge.

PR #6794, opened 2023-07-19
EVG-18734 only activate deactivated tasks

ybrill:
[EVG-18734](https://jira.mongodb.org/browse/EVG-18734)

### Description
I'm looking to track the time it takes from when a task is activated until the time begins to run. This will allow us to evaluate timing improvements that we can't see today. I ran into cases where tasks were getting activated after they were already activated, sometimes even after a task has started to run, and it was throwing off the numbers.

This PR makes activating an already activated task a noop. If there's an issue with this I could add a separate field to track the last time it was actually activated, but if not it seemed to make sense to only activate a deactivated task.

Idk if we want to do the same for deactivating tasks, but I'm happy to if someone suggests they should be consistent. 

### Testing
Added to the `ActivateTask` test.

### DIFF
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -2054,7 +2054,7 @@ func TestActivateTasks(t *testing.T) {
 		assert.Len(t, dbTasks, 6)
 
 		for _, task := range dbTasks {
-			assert.Equal(t, task.Priority, int64(0))
+			assert.EqualValues(t, 0, task.Priority)
 			if utility.StringSliceContains(updatedIDs, task.Id) {
 				assert.True(t, task.Activated)
 				events, err := event.FindAllByResourceID(task.Id)
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -2035,6 +2035,7 @@ func TestActivateTasks(t *testing.T) {
 		{Id: "t2", DependsOn: []Dependency{{TaskId: "t0"}, {TaskId: "t1"}}, Activated: false, DeactivatedForDependency: true},
 		{Id: "t3", DependsOn: []Dependency{{TaskId: "t0"}}, Activated: false, DeactivatedForDependency: true},
 		{Id: "t4", DependsOn: []Dependency{{TaskId: "t0"}, {TaskId: "t3"}}, Activated: false, DeactivatedForDependency: true},
+		{Id: "t5", Activated: true},
 	}
 	for _, task := range tasks {
 		require.NoError(t, task.Insert())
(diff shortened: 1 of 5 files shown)


model/task/task_test.go
```diff
@@ -2035,6 +2035,7 @@ func TestActivateTasks(t *testing.T) {
 		{Id: "t2", DependsOn: []Dependency{{TaskId: "t0"}, {TaskId: "t1"}}, Activated: false, DeactivatedForDependency: true},
 		{Id: "t3", DependsOn: []Dependency{{TaskId: "t0"}}, Activated: false, DeactivatedForDependency: true},
 		{Id: "t4", DependsOn: []Dependency{{TaskId: "t0"}, {TaskId: "t3"}}, Activated: false, DeactivatedForDependency: true},
+		{Id: "t5", Activated: true},
```
Kimchelly, line 2038: Is this supposed to depend on `t0`, so we can ascertain that `ActivateTasks` did not try to change its activation state? Otherwise, it seems like the test to `ActivateTasks` would have no effect on `t5`, even before this change.

Kimchelly: Also, is it worth testing direct task activation with the task ID vs. indirect activation due to dependencies?

Kimchelly (changes requested):
Change itself looks good, just a question on the test.

model/task/task_test.go
ybrill: fixed and donw

model/task/task_test.go
```diff
@@ -2027,46 +2027,76 @@ func TestTopologicalSort(t *testing.T) {
...
-			assert.True(t, task.Activated)
-			events, err := event.FindAllByResourceID(task.Id)
-			require.NoError(t, err)
-			require.Len(t, events, 1)
-		} else {
-			for _, origTask := range tasks {
-				if origTask.Id == task.Id {
-					assert.Equal(t, origTask.Activated, task.Activated, fmt.Sprintf("task '%s' mismatch", task.Id))
+		for _, task := range dbTasks {
+			assert.Equal(t, task.Priority, int64(0))
```
Kimchelly, line 2057: nit: the order for `Equal` is expected value, actual value, so the two args should be reversed. I imagine you could also just `assert.EqualValues` so you don't have to explicitly make an `int64`, but it's not that important.

Kimchelly (approved):

model/task/task_test.go
ybrill: done and done

PR #6798, opened 2023-07-19
EVG-20276  Unset function vars after the function is done 

malikchaya2:
EVG-20276

### Description
This unsets function vars after its done 

### Testing
Added a test and [tested on staging ](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_chaya_test_patch_3cb34df8ed50b147a31f017f465778eb613918f5_64b98d799dbe327410db6396_23_07_20_19_39_47/logs?execution=0)


agent/command.go
```diff
@@ -146,15 +146,22 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
 // single sub-command within a function.
 func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.LoggerProducer, commandInfo model.PluginCommandConf,
 	cmd command.Command, displayName string, options runCommandsOptions) error {
-
+	prevDef := map[string]string{}
```
Kimchelly, line 149: nit: I would call this `prevExp` or something since def is kind of vague here.

agent/command.go
```diff
@@ -146,15 +146,22 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
...
+		prevDef[key] = prevVal
+
 		var newVal string
 		newVal, err := tc.taskConfig.Expansions.ExpandString(val)
 		if err != nil {
 			return errors.Wrapf(err, "expanding '%s'", val)
 		}
 		tc.taskConfig.Expansions.Put(key, newVal)
 	}
+	// unset the function vars once the function finished running
```
Kimchelly, line 161: I don't think we need this comment since it's just stating what the code is doing. However, if you want to explain the way, you can change the comment to say you want to ensure function vars do not persist in the expansions after the function is over (inside the defer block).

Kimchelly, line 161: Also, comment style generally expects capitalization/grammar/maybe even a period if it's a full sentence.

agent/command_test.go
```diff
@@ -168,3 +195,56 @@ func TestEndTaskSyncCommands(t *testing.T) {
 		})
 	}
 }
+
+func (s *CommandSuite) TestVarsAreUnsetAfterRunning() {
+	ctx, cancel := context.WithCancel(context.Background())
```
Kimchelly, line 200: nit: since this is a suite, it's generally preferred to just absorb ctx into `SetUpTest` and initialize suite fields with the ctx/cancel func, then cancel the context in `TearDownTest`.

agent/command_test.go
```diff
@@ -168,3 +195,56 @@ func TestEndTaskSyncCommands(t *testing.T) {
...
+	defer cancel()
+
+	var err error
```
Kimchelly, line 203: nit: move this to L232 so it's defined closer to where it actually gets initialized.

agent/command_test.go
```diff
@@ -168,3 +195,56 @@ func TestEndTaskSyncCommands(t *testing.T) {
...
+		Vars:        map[string]string{"globalVar": "functionVar"},
+	}
+
+	cmds := []model.PluginCommandConf{func1}
+	err = s.a.runCommandsInBlock(ctx, s.tc, cmds, runCommandsOptions{}, "")
+	s.NoError(err)
+
+	globalVarValue := s.tc.taskConfig.Expansions.Get("globalVar")
+	s.Equal("expansionVar", globalVarValue, "globalVar should be set back to what it was before the function ran")
+	s.NoError(s.tc.logger.Close())
```
Kimchelly, line 249: It's not necessary to `s.tc.logger.Close()` unless you want to read data out of it (this is not true of loggers in production code, but is true in tests that use the mock communicator's logger because the mock logger is just writing to an in-memory buffer).

Kimchelly (changes requested):

Kimchelly (approved):
LGTM mod fix the `lint-agent` error.

PR #6802, opened 2023-07-20
EVG-20352: allow generate.tasks to run in non-main blocks

Kimchelly:
EVG-20352

### Description
The issue was that `generate.tasks` was not picked up in the timeout block because the cron populator that picks up tasks to generate didn't count those where `(Task).GenerateTask` is not set. That field is used to indicate to _the scheduler_ that the task runs `generate.tasks` and should therefore get a bump in its task queue position. However, it's only set if the `generate.tasks` command is in the main task block, not in other blocks like pre, post, timeout etc. Checking this field also happens to not not necessary because it is only used for ordering the task queue, not for actually deciding which tasks are waiting for generation, so I removed it. There's also no impact on the query performance because `GenerateTask` is not utilized by any index.

One thing worth noting here is that we _don't_ want to set `GenerateTasks` for tasks that run `generate.tasks` in the non-main block because it would potentially give many tasks the priority bump (e.g. a project running `generate.tasks` in `post` would give it the bump to all regular tasks, which is excessive).

### Testing
Added unit tests for the function and submitted a [patch](https://spruce-staging.corp.mongodb.com/version/64b980ac9dbe326e0d1b8d0d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) in staging to verify that `generate.tasks` ran in the timeout block.

### Documentation
N/A


model/task/task.go
```diff
@@ -233,7 +233,9 @@ type Task struct {
 	DisplayTaskId *string `bson:"display_task_id,omitempty" json:"display_task_id,omitempty"`
 
 	// GenerateTask indicates that the task generates other tasks, which the
-	// scheduler will use to prioritize this task.
+	// scheduler will use to prioritize this task. This will not be set for
+	// tasks where the generate.task command runs outside of the main task block
```
hadjri, line 237: nit: generate.tasks

hadjri (approved):

PR #6806, opened 2023-07-20
EVG-19190 thread through ctx for admin queries

ybrill:
[EVG-19190](https://jira.mongodb.org/browse/EVG-19190)

### Description
The next installment in threading through contexts. This PR is for the admin collection which stores the configuration.
Although it changes a lot of files, there's not too much here besides threading through the ctx.


units/crons_event_test.go
```diff
@@ -112,20 +112,17 @@ func (s *cronsEventSuite) SetupTest() {
 }
 
 func (s *cronsEventSuite) TestDegradedMode() {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
 	// Reset to original flags after the test finishes.
-	originalFlags, err := evergreen.GetServiceFlags()
+	originalFlags, err := evergreen.GetServiceFlags(s.ctx)
```
johndaniels, line 116: This switches to using a context that is torn down at the end of the suite rather than the end of the test I think?

I'm guessing that's fine.

johndaniels (approved):

units/crons_event_test.go
ybrill: ah, that's true. I hadn't noticed it.
It seems okay.

PR #6807, opened 2023-07-21
EVG-20473 Prevent the same trigger from trying to create the same version twice

hadjri:
EVG-20473

### Description
Each trigger definition creates the same version ID. For triggers that run off of a successful build in an upstream version, multiple requests to create the same version with the same ID will run. Normally this doesn't cause a problem because they all fail with "duplicate key" errors, but we recently ran into an issue where these duplicate requests wiped the update a `generate.tasks` command performed on the version's parser project.

This was because the build triggers were ignoring the `SatisfiedTriggers` field, unlike the task triggers. Also did a drive by fix removing an unused field in a helper function.

### Testing
Created new unit test confirming this prevents the second run for the build triggers from erroring

### DIFF
--- a/repotracker/repotracker.go
+++ b/repotracker/repotracker.go
@@ -593,6 +593,8 @@ func makeBuildBreakSubscriber(userID string) (*event.Subscriber, error) {
 	return subscriber, nil
 }
 
+// CreateVersionFromConfig will create a version document from a project config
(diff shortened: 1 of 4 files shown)


malikchaya2 (approved):
good catch! 🎉

PR #6808, opened 2023-07-21
EVG-20500 Add an additional check for unused filter params

khelif96:
EVG-20500

### Description
I noticed that the aggregation was still being built with these stages included despite them being empty which would result in stages like the below being included in the aggregation. These stages are superfluous since an empty regex will always match everything. 
```js
{
  $or: [
    {
      build_variant_display_name: {
        $regex: "",
        $options: "i",
      },
    },
    {
      build_variant: {
        $regex: "",
        $options: "i",
      },
    },
  ],
}
```

Testing:
Existing Unit and e2e tests guarantee the filter functionality still works

Looking at the aggregation generated in this honeycomb trace proves that we are no longer doing any unnecessary matching on fields that are not populated
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/kHJoswnXcf/trace/vqkj6kb8suG?fields%5B%5D=c_name&span=6748071048b52fec


model/task/db.go
```diff
@@ -2559,7 +2559,7 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 	}
 
 	// Filter on Build Variants matching on display name or variant name if it exists
-	if len(opts.Variants) > 0 {
+	if len(opts.Variants) > 0 && opts.Variants[0] != "" {
```
sophstad, line 2562: Is it possible that an empty string would appear in a non-zero position in the list of task/variant filters? If so it may make more sense to filter out the empty strings instead.

model/task/db.go
```diff
@@ -2545,7 +2545,7 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
 		GeneratedJSONAsStringKey: 0,
 	}
 
-	if len(opts.TaskNames) > 0 {
+	if len(opts.TaskNames) > 0 && opts.Variants[0] != "" {
```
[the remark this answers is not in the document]
khelif96: Yes I did! I was wondering why this was panicking good catch!

model/task/db.go
khelif96: Good catch!

khelif96:
> Can you please include a description of how you tested this change?

Done

sophstad (approved):

malikchaya2 (approved):

PR #6809, opened 2023-07-21
EVG-20310 Remove unused agent opts

malikchaya2:
[EVG-20310](https://jira.mongodb.org/browse/EVG-20310)

### Description
The opts AgentSleepInterval and MaxAgentSleepInterval were ready but never set (except for one test). They were removed to simplify the code. 

### Testing
Existing tests

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -63,8 +63,6 @@ type Options struct {
 	LogOutput              LogOutputType
 	WorkingDirectory       string
 	HeartbeatInterval      time.Duration
-	AgentSleepInterval     time.Duration
-	MaxAgentSleepInterval  time.Duration
 	Cleanup                bool
 	SetupData              apimodels.AgentSetupData
 	CloudProvider          string
@@ -258,13 +256,6 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 
 func (a *Agent) loop(ctx context.Context) error {
 	minAgentSleepInterval := defaultAgentSleepInterval
-	maxAgentSleepInterval := defaultMaxAgentSleepInterval
-	if a.opts.AgentSleepInterval != 0 {
-		minAgentSleepInterval = a.opts.AgentSleepInterval
-	}
-	if a.opts.MaxAgentSleepInterval != 0 {
-		maxAgentSleepInterval = a.opts.MaxAgentSleepInterval
-	}
 	agentSleepInterval := minAgentSleepInterval
 
 	var jitteredSleep time.Duration
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -255,12 +255,7 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 }
 
 func (a *Agent) loop(ctx context.Context) error {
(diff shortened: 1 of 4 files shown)


agent/agent.go
```diff
@@ -258,13 +256,6 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 
 func (a *Agent) loop(ctx context.Context) error {
 	minAgentSleepInterval := defaultAgentSleepInterval
```
Kimchelly, line 258: This is a constant because it never changes once set, so this can be further simplified down by renaming the const `defaultAgentSleepInterval` to `minAgentSleepInterval`.

agent/agent.go
```diff
@@ -258,13 +256,6 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
...
-	maxAgentSleepInterval := defaultMaxAgentSleepInterval
-	if a.opts.AgentSleepInterval != 0 {
-		minAgentSleepInterval = a.opts.AgentSleepInterval
-	}
-	if a.opts.MaxAgentSleepInterval != 0 {
-		maxAgentSleepInterval = a.opts.MaxAgentSleepInterval
-	}
 	agentSleepInterval := minAgentSleepInterval
 
 	var jitteredSleep time.Duration
```
Kimchelly, line 261: This `jitteredSleep` variable declaration can also be removed because the only place is used is within a single iteration of the agent loop. We only need to declare/define it on L376, and then later loop iterations never need to reuse it.

Kimchelly, line 261: GitHub doesn't let me comment on unchanged lines that are far from the diff lines (😭) but the `tskCtx` defined on L271 (right below `jitteredSleep`) is also not necessary, and can be removed. The reason we can know this is true is because the only place that actually starts an agent already passes in a context whose lifetime is equivalent to this context. In other words, when this loop returns (and the defer cancels the context) is pretty much in the same condition as when [the agent's `ctx` _also_ defer cancels](https://github.com/evergreen-ci/evergreen/blob/5215dd3c022eadd3ff0c109291963d90586823b1/operations/agent.go#L188) (i.e. `(Agent).Start` is the only caller of `(Agent).loop`, and when the loop ends, so does the agent CLI command itself).

Kimchelly (changes requested):

agent/agent.go
malikchaya2: good catch! thanks!

agent/agent.go
malikchaya2: true! done.

malikchaya2: aaah yes! removing.

agent/global.go
```diff
@@ -5,13 +5,13 @@ import (
 )
 
 const (
-	// defaultAgentSleepInterval is the default amount of time an agent sleeps in between
+	// minAgentSleepInterval is the default amount of time an agent sleeps in between
```
Kimchelly, line 8: `minAgentSleepInterval is the minimum amount of time`

Kimchelly (approved):
LGTM mod update doc comment.

PR #6812, opened 2023-07-24
EVG-18734 run fifteen second jobs every fifteen seconds

ybrill:
[EVG-18734](https://jira.mongodb.org/browse/EVG-18734)

### Description
A bunch of the crons that are supposed to be running every fifteen seconds were actually running once a minute. This is because [RoundPartOfMinute](https://github.com/evergreen-ci/utility/blob/ea9a6354cebb635c026d8dbc7621c35622bb10c4/time.go#L87) rounds down to the nearest minute if n is 0. Providing `n == 15` will do what we want: floor to the previous 15 second interval.

### Testing
On staging the scheduler, alias scheduler, and host allocator jobs ran [every 15 seconds](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/4yzKaJQWYpE) instead of [every minute](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/hvreU5BzWoM).

### DIFF
--- a/units/crons.go
+++ b/units/crons.go
@@ -417,7 +417,7 @@ func PopulateHostAllocatorJobs(env evergreen.Environment) amboy.QueueOperation {
 			return errors.Wrap(err, "finding distros that need planning")
 		}
 
-		ts := utility.RoundPartOfMinute(0)
+		ts := utility.RoundPartOfMinute(15)
 		catcher := grip.NewBasicCatcher()
 
 		for _, d := range distros {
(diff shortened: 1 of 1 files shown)


julianedwards (changes requested):
I think you are missing [PopulateAgentDeployJobs](https://github.com/evergreen-ci/evergreen/blob/main/units/crons_remote_fifteen_second.go#L52). 

I think it would be safer to refactor these functions to accept a timestamp so [this function](https://github.com/evergreen-ci/evergreen/blob/main/units/crons_remote_fifteen_second.go#L29) could just pass it in rather than relying on the individual populate functions to correctly set the timestamp. But that can be a refactor for another day.

ybrill:
@julianedwards doesn't [PopulateAgentDeployJobs](https://github.com/evergreen-ci/evergreen/blob/main/units/crons_remote_fifteen_second.go#L52) already do this [here](https://github.com/evergreen-ci/evergreen/blob/2cb4004971943238be79bf06b4ff50365d880fa3/units/crons.go#L590)?

julianedwards:
Oh, you are right! I missed that.

julianedwards (approved):

PR #6813, opened 2023-07-24
EVG-20300: reorganize smoke test files

Kimchelly:
EVG-20300

### Description
Reorganize the smoke test files to consolidate them into `testdata/smoke`. Along the way, I fixed up a couple of trivial issues, like the smoke test admin settings for Amboy being malformed, smoke test data models defined in a weird location, and a random Go test depending on smoke test files.

### Testing
This can't pass until it's merged because the app server has to get the project YAML from GitHub to compare against the smoke patch's YAML, but the renamed file doesn't exist in GitHub yet. It will pass once it's merged, and given that it's just renaming files, it seems safe to merge and verify after merging that it works.

### Documentation
I updated the Evergreen dev guide section on the smoke test so it mentions how to see if your changes pass the smoke test. We don't need the complex workarounds described in the guide anymore, just running a patch is sufficient.


hadjri (approved):
Nit: looks like CLI version needs an uptick

Kimchelly:
Eh, I don't think incrementing the CLI version is necessary because it's not a user-facing change (the linter has false positives). The changes are to the smoke test commands, which are only used internally (and we always compile the CLI for the smoke test anyways).

PR #6815, opened 2023-07-25
EVG-20276 Unset function vars Plus edge case

malikchaya2:
EVG-20276

### Description
This contains the unsetting function vars logic that was reverted [here](https://github.com/evergreen-ci/evergreen/pull/6814), plus an additional check that makes sure that if the same function uses expansions.update to update a var that is also set for the function, the expansions.update persists and doesn't get reset. 

I accidentally squashed my new changes with the changes that were reverted. The new changes are line 163-169 in agent/comand.go. I also added another test to test this use case and pulled some logic out into a helper function so that the two tests are cleaner. 

### Testing
Added a test for the edge case 

### Note 
After we rolled it out we realized that some teams were relying on this bug for their tests. We should send out some communication before merging it in this time and give people time to look at their setup. Once approved I will work on that and give it some time before merging.


hadjri (approved):

agent/command.go
```diff
@@ -146,15 +146,29 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
...
 		}
 		tc.taskConfig.Expansions.Put(key, newVal)
 	}
+	defer func() {
+		// This defer ensures that the function vars do not persist in the expansions after the function is over.
+		for key, functionValue := range commandInfo.Vars {
+			currentValue := tc.taskConfig.Expansions.Get(key)
+			if currentValue != functionValue {
+				// if an expansion has been updated to something that's not a function var, persist it.
+				prevExp[key] = currentValue
```
Kimchelly, line 167: Seems like this could just `continue` the loop rather than update the value to the same value it already has.

Kimchelly (approved):
LGTM mod nit and sending necessary comms for this change.

agent/command.go
```diff
@@ -146,15 +146,29 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
...
 			return errors.Wrapf(err, "expanding '%s'", val)
 		}
 		tc.taskConfig.Expansions.Put(key, newVal)
 	}
+	defer func() {
+		// This defer ensures that the function vars do not persist in the expansions after the function is over.
+		for key, functionValue := range commandInfo.Vars {
+			currentValue := tc.taskConfig.Expansions.Get(key)
+			if currentValue != functionValue {
+				// if an expansion has been updated to something that's not a function var, persist it.
```
Kimchelly, line 166: nit: capitalization

Kimchelly: Rather than describe this in terms of function vars, I would just phrase this as something like "If the expansions.update command updates the expansion value, persist it."

agent/command.go
malikchaya2: commandInfo.Vars already has currentValue, but prevExp has the value it persisted up top before the function ran, it doesn't have the currentValue yet.

malikchaya2:
@Kimchelly @hadjri  re-requesting review because I added a feature flag. I tested the flag on staging to make sure it works as expected.

hadjri (approved):

hadjri:
> @Kimchelly @hadjri re-requesting review because I added a feature flag. I tested the flag on staging to make sure it works as expected.

Is the flag intended to be temporary?

malikchaya2:
> > @Kimchelly @hadjri re-requesting review because I added a feature flag. I tested the flag on staging to make sure it works as expected.
> 
> Is the flag intended to be temporary?

Yes, but it will probably be in for around 2 months. I'll create a ticket to remove it.

PR #6816, opened 2023-07-25
EVG-20432: simplify heartbeat

Kimchelly:
EVG-20432

### Description
This reworks the heartbeat so that it's substantially less complicated than it was before, but equivalent in functionality. The key assumption here is that if we can assume that 1. abort cancels the task context, 2. `startTask` will respect the context cancellation and therefore eventually exits, then we don't need superfluous `heartbeat` channel at all. This assumption is true by inspection because all potentially long-running operations in `startTask` use the context (or have their own internal timeouts) and all the command-running functions have been shored up long ago to respect the context cancellation signal. Therefore, when the heartbeat cancels the context, the task will eventually abort.

* Simplify down the heartbeat to signal pre and main to abort. This also reduces the amount of channel signaling the agent has to do.
* Add better doc comments for the heartbeat/`startTask`.
* Reorganize `agent/task.go` file because it contained a bunch of random things that didn't really go together. I split it so that `startTask`, `runPreTaskCommands` are in the file that contains the rest of the task-running logic, and put the task context/config stuff in a new file.

### Testing
I added a bunch of unit tests for heartbeat behavior. Existing tests should all pass.

I also [tested in staging](https://spruce-staging.corp.mongodb.com/version/64bff23db237360c22b04d74/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) that the heartbeat behaved correctly when I aborted a task, aborted and restarted a task, and just let the task run normally.

### Documentation
I wanted to document what `Failing command: heartbeat` means when a user sees it in Spruce, because system-failing due to heartbeat is difficult to interpret (since it's caused by Evergreen). Unfortunately, docs for the patches/tasks page don't exist yet, so there's not really any appropriate place to put it. We can consider adding better docs for `Failing command` in Spruce after EVG-20514.


agent/background.go
```diff
@@ -10,63 +10,64 @@ import (
 	"github.com/mongodb/grip/recovery"
 )
 
-func (a *Agent) startHeartbeat(ctx context.Context, cancel context.CancelFunc, tc *taskContext, heartbeat chan<- string) {
+// startHeartbeat runs the task heartbeat. The heartbeat is responsible for two
+// things:
+//  1. It communicates with the app server to indicate that the agent is still
+//     alive and running its task. If the app server does not receive a heartbeat
+//     for a long time while the task is still running, then it can assume task
+//     has somehow stopped running and can choose system-fail the task.
```
malikchaya2, line 18: nit: can assume "the" task 
and 
choose "to" system-fail

malikchaya2 (approved):
This is terrifying, but seems sound! 
approving with nit.

PR #6823, opened 2023-07-26
EVG-19917 run idle-host-termination more frequently

ybrill:
[EVG-19917](https://jira.mongodb.org/browse/EVG-19917)

### Description
Now that the idle-host-termination job is consistently faster than 15s (https://github.com/evergreen-ci/evergreen/pull/6790) we can run it more frequently. Hopefully this will help us to decrease idle time without a corresponding increase in wait time. I'll need to check on this a day or two after it deploys.

### Testing
In staging these changes made the job run [every fifteen seconds instead of once a minute](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/32rYi4CRRvi).

### DIFF
--- a/units/crons.go
+++ b/units/crons.go
@@ -277,26 +277,6 @@ func PopulateHostTerminationJobs(env evergreen.Environment) amboy.QueueOperation
 	}
 }
 
-func PopulateIdleHostJobs(env evergreen.Environment) amboy.QueueOperation {
-	return func(ctx context.Context, queue amboy.Queue) error {
-		flags, err := evergreen.GetServiceFlags(ctx)
-		if err != nil {
-			return errors.Wrap(err, "getting service flags")
-		}
-
-		if flags.MonitorDisabled {
-			grip.InfoWhen(sometimes.Percent(evergreen.DegradedLoggingPercent), message.Fields{
-				"message": "monitor is disabled",
-				"impact":  "not submitting detecting idle hosts",
-				"mode":    "degraded",
-			})
-			return nil
-		}
-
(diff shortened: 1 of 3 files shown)


units/crons.go
```diff
@@ -499,6 +479,26 @@ func PopulateAliasSchedulerJobs(env evergreen.Environment) amboy.QueueOperation
...
+		if flags.MonitorDisabled {
+			grip.InfoWhen(sometimes.Percent(evergreen.DegradedLoggingPercent), message.Fields{
+				"message": "monitor is disabled",
+				"impact":  "not submitting detecting idle hosts",
+				"mode":    "degraded",
+			})
+			return nil
+		}
+
+		return queue.Put(ctx, NewIdleHostTerminationJob(env, utility.RoundPartOfMinute(15).Format(TSFormat)))
```
opsnlops, line 498: Is this the only actual change?

ybrill: also to move it to [the 15 second populator](https://github.com/evergreen-ci/evergreen/blob/f5af362e71f9cd8dc606cbf12814d4b0c660e77e/units/crons_remote_fifteen_second.go#L29) instead of the [1 minute one](https://github.com/evergreen-ci/evergreen/blob/f5af362e71f9cd8dc606cbf12814d4b0c660e77e/units/crons_remote_minute.go#L29).

The idea of the populators is that they're [a goroutine](https://github.com/mongodb/amboy/blob/94858bb64f00e0a01ed226ceb1ecaed85b4ea2ff/periodic.go#L108) that inserts a job every interval.

opsnlops: Cool, just wanted to make sure I didn't miss another change. This diff did you no favors. 🤣

opsnlops (approved):

PR #6824, opened 2023-07-26
EVG-20516 Only search for project when one is known

bsamek:
EVG-20516

### Description

This fixes a bug introduced by
https://github.com/evergreen-ci/evergreen/pull/6778. That PR always attempted to
set owner and repo using the project name. However, only CLI and trigger types
already set a name. GitHub PR and merge group types do not set a name until
later in the process. This means that the code was finding a project by passing
an empty identifier.

Note that this PR also adds some returns, whereas before we continued on error.

This PR also passes the GitHub merge group data instead of PR data to the update 
message.

### Testing

Ran existing unit tests.

### Documentation

N/A

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -100,10 +100,21 @@ func (j *patchIntentProcessor) Run(ctx context.Context) {
 	}
 
 	patchDoc := j.intent.NewPatch()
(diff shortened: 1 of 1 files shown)


bsamek:
@bynn I added some `return`s to your code. Could you make sure that's okay?

malikchaya2 (approved):

PR #6826, opened 2023-07-26
EVG-20283: let commands gracefully shut down

Kimchelly:
EVG-20283

### Description
* Do not run pre and main blocks in a goroutine anymore. They can run in the foreground.
* When commands are cancelled (i.e. most likely due to abort or hitting a timeout) try to give commands a short grace period to finish up just in case the context cancellation is slow to propagate to it, or because the command ignores the context.

### Testing
Added unit test.

Tested in staging to basic verify that aborting a task does not hang forever.

### Documentation
Wrote an ADR to describe the original reasoning for the goroutine and why it is no longer necessary.

### DIFF
--- a/docs/decisions/2027-07-26_agent_command_commands.md
+++ b/docs/decisions/2027-07-26_agent_command_commands.md
@@ -1,0 +1,18 @@
+# 2023-07-26 Agent Task Execution and Command Graceful Exit
+
+* status: accepted
+* date: 2023-07-26
+* authors: Kim Tao
+
+## Context and Problem Statement
+
+The agent was implemented run pre and main block commands in a separate goroutine and then waited on either it to
+return or for the task to hit a context error (e.g. due to aborting the task). It seems to have been put in a goroutine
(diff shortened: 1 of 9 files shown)


docs/decisions/2027-07-26_agent_command_commands.md
```diff
@@ -0,0 +1,18 @@
+# 2023-07-26 Agent Task Execution and Command Graceful Exit
+
+* status: accepted
+* date: 2023-07-26
+* authors: Kim Tao
+
+## Context and Problem Statement
+
+The agent was implemented run pre and main block commands in a separate goroutine and then waited on either it to
```
malikchaya2, line 9: *to run

malikchaya2 (approved):

docs/decisions/2027-07-26_agent_command_commands.md
Kimchelly: I reworked this sentence a little to have less awkward phrasing.

PR #6827, opened 2023-07-26
EVG-19960 Create fallback for annotation lookup when it hits the $facet size limit

khelif96:
EVG-19960

### Description
On certain very large patches, there is a chance  that the `getTasksByVersion` aggregation encounters the following error `(Location4031700) PlanExecutor error during aggregation :: caused by :: document constructed by $facet is 104858682 bytes, which exceeds the limit of 104857600 bytes`  

This error stumped me because it lead me to believe that this was something that could be resolved by toggling the `allowDiskUsage` flag to get around MongoDB's 100mb stage limit. In reality, this is actually a limitation with the `$facet` stage that is described in [SERVER-59837](https://jira.mongodb.org/browse/SERVER-59837)

In order to get past this I decided to introduce an additional parameter that will not use the optimization introduced in EVG-15279 - #4997 and instead fallback to looking up annotations for all tasks not just the failed ones. 
If the original query fails this aggregation will use the fallback. 
 
### Testing
Created 200 MB worth of task documents and asserted the aggregation returns task results.

### Documentation

### DIFF
--- a/db/errors.go
+++ b/db/errors.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/evergreen-ci/evergreen/db/mgo"
 	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/mongo"
 )
 
 func IsDuplicateKey(err error) bool {
@@ -29,3 +30,15 @@ func IsDocumentLimit(err error) bool {
 	}
 	return strings.Contains(errors.Cause(err).Error(), "an inserted document is too large")
 }
+
+func IsErrorCode(err error, errorCode int) bool {
+	var mongoError mongo.CommandError
+	if errors.As(err, &mongoError) && mongoError.HasErrorCode(errorCode) {
+		return true
+	}
+	return false
+}
+
+// ErrorCode constants for mongo errors
+
+const FACET_PIPELINE_STAGE_TOO_LARGE_CODE = 4031700 // https://github.com/mongodb/mongo/blob/a1732172ed5d66d98582ea1059c0ede9d8cd5065/src/mongo/db/pipeline/document_source_facet.cpp#L165
--- a/db/errors.go
+++ b/db/errors.go
@@ -44,4 +44,4 @@ func IsErrorCode(err error, errorCode int) bool {
 
 // FacetPipelineStageTooLargeCode is the error code for when a facet pipeline stage is too large.
 // https://github.com/mongodb/mongo/blob/a1732172ed5d66d98582ea1059c0ede9d8cd5065/src/mongo/db/pipeline/document_source_facet.cpp#L165
-const FacetPipelineStageTooLargeCode = 4031700 //
+const FacetPipelineStageTooLargeCode = 4031700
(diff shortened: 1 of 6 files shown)


db/errors.go
```diff
@@ -29,3 +30,15 @@ func IsDocumentLimit(err error) bool {
 	}
 	return strings.Contains(errors.Cause(err).Error(), "an inserted document is too large")
 }
+
+func IsErrorCode(err error, errorCode int) bool {
```
Kimchelly, line 34: nit: doc comment for exported method.

db/errors.go
```diff
@@ -29,3 +30,15 @@ func IsDocumentLimit(err error) bool {
...
+	var mongoError mongo.CommandError
+	if errors.As(err, &mongoError) && mongoError.HasErrorCode(errorCode) {
+		return true
+	}
+	return false
+}
+
+// ErrorCode constants for mongo errors
+
+const FACET_PIPELINE_STAGE_TOO_LARGE_CODE = 4031700 // https://github.com/mongodb/mongo/blob/a1732172ed5d66d98582ea1059c0ede9d8cd5065/src/mongo/db/pipeline/document_source_facet.cpp#L165
```
Kimchelly, line 44: nit: Go style in Evergreen uses `PascalCase` for constants rather than C-style `YELLING_CASE`.

model/task/db.go
```diff
@@ -2112,7 +2144,13 @@ func GetTasksByVersion(ctx context.Context, versionID string, opts GetTasksByVer
 
 	env := evergreen.GetEnvironment()
 	cursor, err := env.DB().Collection(Collection).Aggregate(ctx, pipeline)
+
 	if err != nil {
+		// If the pipeline stage is too large we should use the slow annotations lookup
+		if db.IsErrorCode(err, db.FACET_PIPELINE_STAGE_TOO_LARGE_CODE) {
+			opts.UseSlowAnnotationsLookup = true
+			return GetTasksByVersion(ctx, versionID, opts)
```
Kimchelly, line 2152: Since this is a recursive call, can we make sure we don't have a chance of infinite recursion by checking if `UseSlowAnnotationsLookup` was already set for this `GetTasksByVersion` call?

Kimchelly (approved):
LGTM mod small suggestions.

db/errors.go
```diff
@@ -41,4 +42,6 @@ func IsErrorCode(err error, errorCode int) bool {
 
 // ErrorCode constants for mongo errors
 
-const FACET_PIPELINE_STAGE_TOO_LARGE_CODE = 4031700 // https://github.com/mongodb/mongo/blob/a1732172ed5d66d98582ea1059c0ede9d8cd5065/src/mongo/db/pipeline/document_source_facet.cpp#L165
+// FacetPipelineStageTooLargeCode is the error code for when a facet pipeline stage is too large.
+// https://github.com/mongodb/mongo/blob/a1732172ed5d66d98582ea1059c0ede9d8cd5065/src/mongo/db/pipeline/document_source_facet.cpp#L165
+const FacetPipelineStageTooLargeCode = 4031700 //
```
Kimchelly, line 47: nit: remove the trailing `//`.

Kimchelly (approved):

PR #6828, opened 2023-07-27
EVG-20299 Disable CLI commit queue when GitHub merge queue is on

bsamek:
EVG-20299

### Description

If the merge queue is set to GitHub, you must merge from a GitHub PR. This PR
returns an error to the user if they try to merge with the Evergreen CLI while
the queue is set to GitHub.

### Testing

I didn't see a straightforward way to test this. This code does not appear to be
directly under test, and it makes some API calls that would make it difficult to mock. I think the change is small enough that the cost of writing new tests is not worth it.

### Documentation

Added a sentence about this constraint.

### DIFF
--- a/config.go
+++ b/config.go
@@ -35,10 +35,10 @@ var (
 	BuildRevision = ""
 
 	// ClientVersion is the commandline version string used to control auto-updating.
-	ClientVersion = "2023-07-25"
+	ClientVersion = "2023-07-27"
 
 	// Agent version to control agent rollover.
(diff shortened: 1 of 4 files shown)


bsamek:
> I agree that its probably not worth adding a unit test. Can we test this manually instead by trying to merge with the cli and making sure we get an error?

Thanks for nudging! There was, in fact, a bug. I tested in staging, fixed the bug, and tested again. It now works.

malikchaya2 (approved):

PR #6829, opened 2023-07-27
EVG-20309 Add comment to agent loop

malikchaya2:
EVG-20309

### Description
Add comment to agent loop


agent/agent.go
```diff
@@ -254,6 +254,9 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
 	a.ec2InstanceID = instanceID
 }
 
+// loop is responsible for continually polling for new tasks and processing them at.
```
Kimchelly, line 257: and processing them.

agent/agent.go
```diff
@@ -254,6 +254,9 @@ func (a *Agent) populateEC2InstanceID(ctx context.Context) {
...
+// If there are no more tasks, it sleeps at interval agentSleepInterval with jitteredSleep
```
Kimchelly, line 258: I don't think this comment needs to be so specific, because the internal sleep implementation between tasks when there is none available is not that critical to understanding its behavior of processing tasks. I would omit the second sentence.

Kimchelly (approved):
LGTM mod phrasing suggestions.

PR #6830, opened 2023-07-27
EVG-20312 Create a processNextTask function

malikchaya2:
EVG-20312

### Description
This extracts some of the next task response processing to its own function to simplify the main agent loop. 

### Testing
Existing tests pass

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -344,6 +344,7 @@ type ProcessNextResponse struct {
 	tc               *taskContext
 }
 
+// ProcessNextTask takes in the NextTaskResponse and processes it.
 func (a *Agent) ProcessNextTask(nt *apimodels.NextTaskResponse, ctx context.Context, tc *taskContext, needPostGroup bool) (ProcessNextResponse, error) {
 	if nt.ShouldExit {
 		grip.Notice("Next task response indicates agent should exit.")
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -305,67 +305,23 @@ func (a *Agent) loop(ctx context.Context) error {
 				}
 				return errors.Wrap(err, "getting next task")
 			}
-			if nextTask.ShouldExit {
-				grip.Notice("Next task response indicates agent should exit.")
+			ntr, err := a.ProcessNextTask(nextTask, ctx, tc, needPostGroup)
+			if err != nil {
+				return errors.Wrap(err, "processing next task")
+			}
+			needPostGroup = ntr.needPostGroup
+			if ntr.tc != nil {
+				tc = ntr.tc
+			}
+			if ntr.ShouldResetTimer {
+				timer.Reset(0)
+				agentSleepInterval = minAgentSleepInterval
+			}
+			if ntr.ShouldContinue {
+				continue
+			}
+			if ntr.ShouldExit {
 				return nil
-			}
-			// if the host's current task group is finished we teardown
-			if nextTask.ShouldTeardownGroup {
-				a.runPostGroupCommands(ctx, tc)
-				needPostGroup = false
-				// Running the post group commands implies exiting the group, so
-				// destroy prior task information.
-				tc = &taskContext{}
-				timer.Reset(0)
-				agentSleepInterval = minAgentSleepInterval
-				continue
-			}
-			if nextTask.TaskId != "" {
-				if nextTask.TaskSecret == "" {
-					grip.Critical(message.WrapError(err, message.Fields{
-						"message": "task response missing secret",
-						"task":    tc.task.ID,
-					}))
-					timer.Reset(0)
-					agentSleepInterval = minAgentSleepInterval
-					continue
-				}
-				prevLogger := tc.logger
-				tc = a.prepareNextTask(ctx, nextTask, tc)
-				if prevLogger != nil {
-					grip.Error(errors.Wrap(prevLogger.Close(), "closing the previous logger producer"))
-				}
-
-				grip.Error(message.WrapError(a.fetchProjectConfig(ctx, tc), message.Fields{
-					"message": "error fetching project config; will attempt at a later point",
-					"task":    tc.task.ID,
-				}))
-
-				a.jasper.Clear(ctx)
-				tc.jasper = a.jasper
-				shouldExit, err := a.runTask(ctx, tc)
-				if err != nil {
-					grip.Critical(message.WrapError(err, message.Fields{
-						"message": "error running task",
-						"task":    tc.task.ID,
-					}))
-					timer.Reset(0)
-					agentSleepInterval = minAgentSleepInterval
-					continue
-				}
-				if shouldExit {
-					return nil
-				}
-				needPostGroup = true
-				timer.Reset(0)
-				agentSleepInterval = minAgentSleepInterval
-				continue
-			} else if needPostGroup {
-				a.runPostGroupCommands(ctx, tc)
-				needPostGroup = false
-				// Running the post group commands implies exiting the group, so
-				// destroy prior task information.
-				tc = &taskContext{}
 			}
 
 			jitteredSleep := utility.JitterInterval(agentSleepInterval)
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -379,6 +335,93 @@ func (a *Agent) loop(ctx context.Context) error {
...
+type ProcessNextResponse struct {
+	ShouldExit       bool
+	ShouldContinue   bool
+	ShouldResetTimer bool
+	needPostGroup    bool
+	tc               *taskContext
+}
+
+// ProcessNextTask takes in the NextTaskResponse and processes it.
+func (a *Agent) ProcessNextTask(nt *apimodels.NextTaskResponse, ctx context.Context, tc *taskContext, needPostGroup bool) (ProcessNextResponse, error) {
```
ablack12, line 348: drive by review but can we make this function and ProcessNextResponse private?

agent/agent.go
```diff
@@ -379,6 +335,93 @@ func (a *Agent) loop(ctx context.Context) error {
...
+	if nt.ShouldExit {
+		grip.Notice("Next task response indicates agent should exit.")
+		return ProcessNextResponse{ShouldExit: true}, nil
+	}
+	// if the host's current task group is finished we teardown
+	if nt.ShouldTeardownGroup {
+		a.runPostGroupCommands(ctx, tc)
+		return ProcessNextResponse{
+			ShouldContinue:   true,
```
Kimchelly, line 357: Rather than pass back `ShouldContinue` and `ShouldResetTimer` as identical values in basically every single case, it seems like we could capture a single field meaning "the agent should sleep again" (i.e. a `noTaskToRun` field or something to that effect) and make that the default rather than "the agent should continue". This would be cleaner and more meaningful for the reader because it makes it clear that having no new task to run means we sleep for a little while, waiting for another task. It also makes the default case better then the default case is to reset the timer to 0 (and ask for another task), and only in the one special case (when there's no new task to run) do you choose to sleep longer. Something like this:
```go
ntr, err := processNextTask(...)
// error handling and whatever
if ntr.noTaskToRun {
    // Increase sleep interval and sleep until next round of asking for a task
    continue
}

// Happy path - we ran something, so we should reset the timer and immediately ask for a new task
```

agent/agent.go
Kimchelly: +1 to this

agent/agent.go
```diff
@@ -379,6 +335,93 @@ func (a *Agent) loop(ctx context.Context) error {
 	}
 }
 
+// ProcessNextResponse contains information for what should be done after processing next task
+type ProcessNextResponse struct {
+	ShouldExit       bool
+	ShouldContinue   bool
+	ShouldResetTimer bool
```
Kimchelly, line 342: Like the struct itself, all these fields should not be exported.

agent/agent.go
```diff
@@ -305,67 +305,23 @@ func (a *Agent) loop(ctx context.Context) error {
 				}
```
Kimchelly, line 305: I would absorb `GetNextTask` into `processNextTask` as well, since the timer resetting logic is the same as the `ShouldResetTimer`/`ShouldContinue` case. It also seems like `processNextTask` is the only thing that uses `GetNextTask`'s output.

agent/agent.go
```diff
@@ -379,6 +335,93 @@ func (a *Agent) loop(ctx context.Context) error {
...
+	} else if needPostGroup {
+		a.runPostGroupCommands(ctx, tc)
+		return ProcessNextResponse{
+			needPostGroup: false,
+			// Running the post group commands implies exiting the group, so
+			// destroy prior task information.
+			tc: &taskContext{},
+		}, nil
+	}
+	return ProcessNextResponse{}, nil
```
Kimchelly, line 422: It's generally preferrable to have your common path (i.e. having a task to run) to be the last return in the function rather than an if case in an `if else if`. Can we restructure the if-elseif and final return so that it returns early if there's no new task to run? Something like this:
```go

if nt.TaskId == "" && needPostGroup {
    // No task but need to run post group, so run it, return, and sleep.
}
if nt.TaskId == "" {
    // No task to run and no post group to run, just return and sleep.
    return processNextResponse{
        noNewTaskToRun: true,
    }
}

// Task ID was returned, so we should actulaly run it.
a.prepareNextTask(...)
a.runTask(...)

return processNextTaskResponse{
    tc: tc,
    // Other fields...
}
```

Kimchelly (changes requested):

agent/agent.go
malikchaya2: If you don't have strong feelings about it, I would prefer to keep them separate, I think it helps with readability.

agent/agent.go
```diff
@@ -305,80 +305,122 @@ func (a *Agent) loop(ctx context.Context) error {
...
-			if agentSleepInterval > maxAgentSleepInterval {
-				agentSleepInterval = maxAgentSleepInterval
+			if ntr.noTaskToRun {
+				// Increase sleep interval and sleep until next round of asking for a task
+				jitteredSleep := utility.JitterInterval(agentSleepInterval)
+				grip.Debugf("Agent sleeping %s.", jitteredSleep)
+				timer.Reset(jitteredSleep)
+				agentSleepInterval = agentSleepInterval * 2
+				if agentSleepInterval > maxAgentSleepInterval {
+					agentSleepInterval = maxAgentSleepInterval
```
malikchaya2, line 328: Lint is failing with 
`agent/agent.go:328:6: ineffectual assignment to agentSleepInterval (ineffassign)`.
I don't  think that's true though. Setting agentSleepInterval here has an impact on the next timer.C run.

Kimchelly: It is an ineffectual assignment though, because L330 has to `continue`. Otherwise, the assignment always evaluates to `minAgentSleepInterval` on L334.

agent/agent.go
```diff
@@ -305,80 +305,122 @@ func (a *Agent) loop(ctx context.Context) error {
...
+				}
 			}
+
+			// reset the timer and immediately ask for a new task
+			timer.Reset(0)
+			agentSleepInterval = minAgentSleepInterval
+			continue
```
Kimchelly, line 336: Remove continue since it doesn't do anything.

agent/agent.go
```diff
@@ -305,80 +305,122 @@ func (a *Agent) loop(ctx context.Context) error {
...
+				jitteredSleep := utility.JitterInterval(agentSleepInterval)
+				grip.Debugf("Agent sleeping %s.", jitteredSleep)
+				timer.Reset(jitteredSleep)
+				agentSleepInterval = agentSleepInterval * 2
+				if agentSleepInterval > maxAgentSleepInterval {
+					agentSleepInterval = maxAgentSleepInterval
+				}
 			}
+
+			// reset the timer and immediately ask for a new task
```
Kimchelly, line 333: I don't think we need this comment.

agent/agent.go
```diff
@@ -305,80 +305,122 @@ func (a *Agent) loop(ctx context.Context) error {
...
 			}
 
-			jitteredSleep := utility.JitterInterval(agentSleepInterval)
-			grip.Debugf("Agent sleeping %s.", jitteredSleep)
-			timer.Reset(jitteredSleep)
-			agentSleepInterval = agentSleepInterval * 2
-			if agentSleepInterval > maxAgentSleepInterval {
-				agentSleepInterval = maxAgentSleepInterval
+			if ntr.noTaskToRun {
+				// Increase sleep interval and sleep until next round of asking for a task
```
Kimchelly, line 322: Rather than put a comment restating what the code is doing, can we just add more to the debug log to say something like `"Agent found no task to run, sleeping %s."`?

agent/agent.go
```diff
@@ -305,80 +305,122 @@ func (a *Agent) loop(ctx context.Context) error {
...
 }
 
+type processNextResponse struct {
+	shouldExit    bool
+	noTaskToRun   bool
+	needPostGroup bool
+	tc            *taskContext
+}
+
+func (a *Agent) processNextTask(nt *apimodels.NextTaskResponse, ctx context.Context, tc *taskContext, needPostGroup bool) (processNextResponse, error) {
```
Kimchelly, line 349: nit: `context.Context` in Go is conventionally the first argument when passed to a function that takes a context. I would order the parameters as `ctx, tc, nt, needsPostGroup`.

Kimchelly (changes requested):

agent/agent.go
malikchaya2: done

agent/agent.go
malikchaya2: done

agent/agent.go
malikchaya2: done

agent/agent.go
malikchaya2: added continue

agent/agent.go
malikchaya2: removed

agent/agent.go
Kimchelly: Also remove the `//nolint` directive.

agent/agent.go
```diff
@@ -348,7 +345,7 @@ type processNextResponse struct {
 	tc            *taskContext
 }
 
-func (a *Agent) processNextTask(nt *apimodels.NextTaskResponse, ctx context.Context, tc *taskContext, needPostGroup bool) (processNextResponse, error) {
+func (a *Agent) processNextTask(ctx context.Context, tc *taskContext, nt *apimodels.NextTaskResponse, needPostGroup bool) (processNextResponse, error) {
```
Kimchelly, line 348: nit: I would make the next task response the 2nd argument since it's the main thing being processed in the function, and the taskContext and needPostGroup are just data retained from the previous task that ran.

Kimchelly (approved):
LGTM mod nits about nolint and parameter order.

PR #6832, opened 2023-07-28
EVG-18952: remove unused Jira notification routes

Kimchelly:
EVG-18952

### Description
I looked through the Splunk logs for the last 60 days but saw no usages of `/notifications/jira_issue` or `/notifications/jira_comment` at all. It seems pretty safe to assume that no recent usage means nobody is depending on it for anything.

### Testing
Removed tests related to Jira notification routes.

### Documentation
Removed documentation related to the Jira notification routes.


rest/route/notification.go
```diff
@@ -60,107 +56,6 @@ func (h *notificationPostHandler) Run(ctx context.Context) gimlet.Responder {
 	return h.handler.Run(ctx)
 }
 
-///////////////////////////////////////////////////////////////////////
-//
-// POST /rest/v2/notifications/jira_comment
-
-type jiraCommentNotificationPostHandler struct {
-	APIJiraComment *model.APIJiraComment
```
ablack12, line 68: I think we can also get rid of this type and its service functions

ablack12 (approved):
lgtm with comment

rest/route/notification.go
Kimchelly: Good catch, removed.

PR #6836, opened 2023-07-28
EVG-19650 use mark end for system failed tasks

ablack12:
EVG-19650 
### Description
We were getting tasks logged to task-end-stats that said they'd been running for 53 years, i.e. the start time was set to the zero time. It looks like this isn't unexpected for tasks that are stranded after dispatching, and we have handling for it in the MarkEnd function, but we don't do that for systsem failures, so modified this to call it, and verified we don't update a task to be finished anywhere else.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -1350,7 +1350,6 @@ func (t *Task) MarkFailed() error {
 }
 
 func (t *Task) MarkSystemFailed(description string) error {
-	t.Status = evergreen.TaskFailed
 	t.FinishTime = time.Now()
 	t.Details = GetSystemFailureDetails(description)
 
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):
This change LGTM regardless since it does fix the bug, but it seems kind of weird for `MarkEnd` to default to the ingest time. Would it make more sense to try setting it to the activated time (if it's non-zero) instead?

ablack12:
It does seem weird but I'm not sure what the right alternative is; activated time I think is more likely to be wrong, since the task could be activated way before it starts if it's on a long queue, and since this number is used for cost things that could cause problems, so we have a worst-case of the standard timeout. All the examples I looked at did appear to be timeouts so this might be the best average we can make.

PR #6837, opened 2023-07-28
EVG-19723 remove pull request function logging

ablack12:
EVG-19723
### Description
Added this logging some time ago to reconsider some fallback logic since this function was being called too much. Now that we aren't calling this function much at all and are moving to a different queue, there's not really any value in figuring out how to avoid hitting this case, so removing this logging and closing out.


Kimchelly (approved):
LGTM mod we can delete `LastRetry` entirely since it's unused now.

PR #6839, opened 2023-07-30
EVG-20455: add new `ExcludeDisplayNames` field for test result filtering

julianedwards:
EVG-20445

### Description
This updates the test results service's `FilterOptions` to include and pipe through the new field `ExcludeDisplayNames` as well as update the graphql schema to include said field in the `TestFilterOptions` input. When set, the new field excludes display names when matching on the test name regex, the default behavior will continue to favor matching on display test name—so the changes are backwards compatible. 

### Testing
Tested in staging that this works with parsley, does not break existing filtering functionality on spruce (which is uses the default filtering on display test name), and added unit tests for the new field in local test results service.

### Documentation
N/A

### DIFF
--- a/model/testresult/testresult.go
+++ b/model/testresult/testresult.go
@@ -138,7 +138,7 @@ func (tr TestResult) GetLogURL(env evergreen.Environment, viewer evergreen.LogVi
 			}
 		}
 
-		return fmt.Sprintf("%s/test/%s/%d/%s?shareLine=%d", parsleyURL, tr.TaskID, tr.Execution, tr.TestName, tr.LineNum)
+		return fmt.Sprintf("%s/test/%s/%d/%s?shareLine=%d", parsleyURL, tr.TaskID, tr.Execution, tr.GetLogTestName(), tr.LineNum)
 	default:
 		if tr.RawLogURL != "" {
 			// Some test results may have internal URLs that are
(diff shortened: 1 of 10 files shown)


model/testresult/testresult.go
```diff
@@ -138,7 +138,7 @@ func (tr TestResult) GetLogURL(env evergreen.Environment, viewer evergreen.LogVi
 			}
 		}
 
-		return fmt.Sprintf("%s/test/%s/%d/%s?shareLine=%d", parsleyURL, tr.TaskID, tr.Execution, tr.GetLogTestName(), tr.LineNum)
+		return fmt.Sprintf("%s/test/%s/%d/%s?shareLine=%d", parsleyURL, tr.TaskID, tr.Execution, tr.TestName, tr.LineNum)
```
ybrill, line 141: It looks to me like the rest of the changes in this PR are meant to go with https://github.com/evergreen-ci/cedar/pull/520 and that this is the only change related to the PR description. Did you mean for all the rest of the changes to be in this PR?

julianedwards: Yeah I need to update the description, sorry about that!

julianedwards: Done.

ybrill (approved):

PR #6842, opened 2023-07-31
EVG-20130 add fields to rest routes

ablack12:
EVG-20130
### Description
Add a few small fields to the project versions route, and artifacts to the project tasks route.


ablack12:
Since these routes work with a lot of versions it may still be beneficial to keep projection so we aren't working with as many fields. I audited the API version struct and it does seem like a lot of necessary information is being sent.

PR #6843, opened 2023-07-31
EVG-18861: warn for missing dependencies with same implicit build variant

Kimchelly:
EVG-18861

### Description
If a task lists a dependency, the build variant of the task and its dependency are implicitly the same. Therefore, the dependency of the task must also be listed in the build variant task list. There was no validation for the case of an implicit build variant for a dependency, so I added a warning.

### Testing
* Added unit test for implicit dependency.
* Tested in staging with `evergreen validate` that a task that depended on another task in the same implicit build variant produced a validation warning.

### Documentation
N/A

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -1487,9 +1487,9 @@ func validateTaskDependencies(project *model.Project) ValidationErrors {
 	return errs
 }
 
-// checkRequestersForTaskDependencies checks that that each task's dependencies will run for
-// the same requesters. For example, a task that runs in a mainline commit
-// cannot depend on a patch only task since the dependency will only be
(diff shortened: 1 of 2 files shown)


validator/project_validator.go
```diff
@@ -1457,12 +1458,40 @@ func validateTaskDependencies(project *model.Project) ValidationErrors {
...
+				}
+			}
+		}
+	}
+
 	return errs
 }
 
-func checkTaskDependencies(task *model.ProjectTask, allTasks map[string]model.ProjectTask) ValidationErrors {
+// checkRequestersForTaskDependencies checks that that each task's dependencies will run for
```
malikchaya2, line 1490: nit: that that

PR #6847, opened 2023-08-01
EVG-17933 Support adding patch alias parameters via the UI

khelif96:
EVG-17933

### Description
Adds support for adding patch alias parameters on the UI

### Testing
Created a [patch](https://spruce-staging.corp.mongodb.com/version/64c938339dbe327ad9eaf5be/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) which used the default params specified in project settings
image


hadjri (approved):

.graphqlrc
```diff
@@ -0,0 +1,2 @@
+schema: "graphql/schema/**/*.graphql"
```
minnakt, line 1: nice 🙌 going to be much easier to find things now

graphql/schema/types/project_settings.graphql
```diff
@@ -188,4 +188,5 @@ type ProjectAlias {
   taskTags: [String!]!
   variant: String!
   variantTags: [String!]!
+  parameters: [Parameter!]
```
minnakt, line 191: tbh I'm not sure what our stance is on nullable vs non-nullable fields but I like making lists non-nullable, unless there's a reason it should be null. i like it because on the frontend you won't have to check if the array is null before checking if it's empty ([ref](https://yelp.github.io/graphql-guidelines/nullability.html#lists))


```suggestion
  parameters: [Parameter!]!
```

minnakt (approved):

graphql/schema/types/project_settings.graphql
khelif96: Good idea!

PR #6851, opened 2023-08-02
EVG-19206: Allow filtering hosts by DNS name

minnakt:
EVG-19206

### Description
This PR makes it possible to filter hosts by DNS name. Instead of making a new parameter, it reuses the existing `HostID` parameter. The reason behind this is:
1. Similar to the ID, the DNS name should uniquely identify the host (in the database, the field for the DNS name is called `host_id`). In fact, the legacy UI [hosts table](https://evergreen.mongodb.com/hosts#/filter/) shows the DNS name in the ID column when it's available.
2. It probably would not be worthwhile to add the DNS name as a new column on the already crowded Spruce [hosts page](https://spruce.mongodb.com/hosts).
<br/>

On the UI, it will look like this:

Screenshot 2023-08-01 at 4 43 10 PM

### Testing
* Added GraphQL test
* Added unit tests for `TestGetPaginatedRunningHosts`
* Edited local hosts data to enable e2e testing for the Spruce PR


model/host/host.go
```diff
@@ -2986,9 +2986,21 @@ func GetHostByIdOrTagWithTask(hostID string) (*Host, error) {
 	return &hosts[0], nil
 }
 
-// GetPaginatedRunningHosts gets running hosts with pagination and applies any
-// filters.
-func GetPaginatedRunningHosts(hostID, distroID, currentTaskID string, statuses []string, startedBy string, sortBy string, sortDir, page, limit int) ([]Host, *int, int, error) {
+// HostsFilterOptions represents the filtering arguments for fetching hosts.
+type HostsFilterOptions struct {
```
khelif96, line 2990: Nice!

khelif96 (approved):

model/host/host.go
```diff
@@ -3033,53 +3045,60 @@ func GetPaginatedRunningHosts(hostID, distroID, currentTaskID string, statuses [
...
-	if len(hostID) > 0 {
+	if len(opts.HostID) > 0 {
 		hasFilters = true
 
 		runningHostsPipeline = append(runningHostsPipeline, bson.M{
 			"$match": bson.M{
-				IdKey: hostID,
+				"$or": []bson.M{
+					{
+						IdKey: bson.M{"$regex": opts.HostID, "$options": "i"},
```
ablack12, line 3055: Is there a user reason we're making this regex or did it just seem like a possible improvement? This is worse performance-wise and I don't think is necessary for this -- when it comes to searching by host name or DNS name, users (probably always) copy paste the thing they need to search for and paste it.

model/host/host.go
```diff
@@ -3033,53 +3045,60 @@ func GetPaginatedRunningHosts(hostID, distroID, currentTaskID string, statuses [
...
+					},
+					{
+						DNSKey: bson.M{"$regex": opts.HostID, "$options": "i"},
```
ablack12, line 3058: Do we have an index on DNS name already? If not we likely need to add one (you can post this in the ops channel)

https://www.mongodb.com/docs/manual/reference/operator/query/or/#:~:text=%24or%20Clauses%20and%20Indexes&text=That%20is%2C%20for%20MongoDB%20to,will%20perform%20a%20collection%20scan.&text=MongoDB%20can%20use%20all%20but,index%20to%20support%20%24or%20clauses.

model/host/host.go
khelif96: I wanted to comment on this as well. But I figured a regex search might allow users to perform searches like the below search to match all hosts that match a certain pattern. For example, matching on all Macos hosts or Zseries hosts that have these static ids
image

minnakt: Geoff was using the legacy hosts page to do regex search, so I just figured that's what he would want on new UI too. I can remove it though

ablack12: Does the legacy host page support regex ? I'm inclined to remove it as unsupported myself; presumably you could filter on distro for other things but if we think there's a use case we can hold it.

model/host/host.go
```diff
@@ -2986,9 +2986,21 @@ func GetHostByIdOrTagWithTask(hostID string) (*Host, error) {
...
+	Statuses      []string
+	StartedBy     string
+	SortBy        string
+	SortDir       int
+	Page          int
+	Limit         int
+}
+
+// GetPaginatedRunningHosts gets running hosts with pagination and applies any filters.
+func GetPaginatedRunningHosts(opts HostsFilterOptions) ([]Host, *int, int, error) {
```
khelif96, line 3003: @minnakt Since you are currently refactoring this can you thread context here so that it doesn't use the global app context to perform this query? See more [here](https://jira.mongodb.org/browse/EVG-20600)

model/host/host.go
minnakt: the legacy page doesn't support regex but it supports includes (like `hostName.includes("my-filter")`) — but other than the examples Mohamed mentioned, it’s true that it might not be that useful, so I removed it for now 🫨

PR #6859, opened 2023-08-02
EVG-19601: allow subprocess.exec binary to search command PATH

Kimchelly:
EVG-19601

### Description
* Extend the places where `subprocess.exec` can search for a `binary` to run. First, try finding binary to run for `subprocess.exec` from either the agent's `PATH` (i.e. the current behavior), then fall back to searching the command's `PATH` (if any).
* Slightly refactor how `env` is handled for `PATH`. (No behavioral change, just reorganization)

### Testing
* Added unit tests.
* Tested in staging that `subprocess.exec` behaved the same as it did before when no `PATH` is given, but additionally searches the command `PATH` for matching executables.

### Documentation
Updated docs for `subprocess.exec` and its special interaction with `PATH`. I also improved docs for how `add_to_path` actually works.

### DIFF
--- a/agent/command/exec.go
+++ b/agent/command/exec.go
@@ -297,6 +297,8 @@ func (c *subprocessExec) getExecutablePath(logger client.LoggerProducer) (absPat
 		return "", err
 	}
 
+	logger.Execution().Debug("could not find executable binary in the default runtime environment PATH, falling back to trying the command's PATH")
+
 	originalPath := os.Getenv("PATH")
 	defer func() {
 		// Try to reset the PATH back to its original state. If this fails, then
--- a/agent/command/exec.go
+++ b/agent/command/exec.go
@@ -292,7 +292,12 @@ func (c *subprocessExec) getExecutablePath(logger client.LoggerProducer) (absPat
 		return defaultPath, err
 	}
 
+	// For non-Windows platforms, the filepath.Separator is always '/'. However,
+	// for Windows, Go accepts both '\' and '/' as valid file path separators,
+	// even though the native filepath.Separator for Windows is really '\'. This
+	// detects both '\' and '/' as valid Windows file path separators.
 	binaryIsFilePath := strings.Contains(c.Binary, string(filepath.Separator)) || runtime.GOOS == "windows" && strings.Contains(c.Binary, "/")
+
 	if len(cmdPath) == 0 || binaryIsFilePath {
 		return "", err
 	}
--- a/agent/command/exec.go
+++ b/agent/command/exec.go
@@ -138,28 +138,19 @@ func (c *subprocessExec) doExpansions(exp *util.Expansions) error {
 		catcher.Wrap(err, "expanding args")
 	}
 
-	// kim: NOTE: env is lower priority than add_to_path
-	// kim: TODO: add test for env var precedence
-	// * If add_to_path is specified: add_to_path, default PATH, explicit PATH.
+	// kim: TODO: update docs about precedence of PATH env var.
+	// * If add_to_path is specified: add_to_path, agent PATH (explicit PATH is
+	// overwritten).
 	// * If explicit PATH is specified, only explicit PATH is available.
 	// * Otherwise (PATH not specified), does not have PATH defined.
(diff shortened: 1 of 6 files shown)


agent/command/exec.go
```diff
@@ -263,13 +280,41 @@ func (c *subprocessExec) getProc(ctx context.Context, taskID string, logger clie
...
+// default PATH locations, the command will fall back to checking the command's
+// PATH environment variable for a matching executable location (if any).
+func (c *subprocessExec) getExecutablePath(logger client.LoggerProducer) (absPath string, err error) {
+	cmdPath := c.Env["PATH"]
+	defaultPath, err := exec.LookPath(c.Binary)
+	if defaultPath != "" {
+		return defaultPath, err
+	}
+
+	binaryIsFilePath := strings.Contains(c.Binary, string(filepath.Separator)) || runtime.GOOS == "windows" && strings.Contains(c.Binary, "/")
```
hadjri, line 295: Should this instead be `runtime.GOOS != "windows" && strings.Contains(c.Binary, string(filepath.Separator)) || runtime.GOOS == "windows" && strings.Contains(c.Binary, "/")`

agent/command/exec.go
```diff
@@ -263,13 +280,41 @@ func (c *subprocessExec) getProc(ctx context.Context, taskID string, logger clie
...
+// then that path will be used. Otherwise if it can't find the command in the
+// default PATH locations, the command will fall back to checking the command's
+// PATH environment variable for a matching executable location (if any).
+func (c *subprocessExec) getExecutablePath(logger client.LoggerProducer) (absPath string, err error) {
+	cmdPath := c.Env["PATH"]
+	defaultPath, err := exec.LookPath(c.Binary)
+	if defaultPath != "" {
+		return defaultPath, err
+	}
+
```
hadjri, line 294: Perhaps we should an execution log in this function saying we are falling back to the commands path?

agent/command/exec.go
```diff
@@ -163,13 +159,25 @@ type modifyEnvOptions struct {
...
 
+	if len(opts.addToPath) > 0 {
+		// Prepend paths to the runtime environment's PATH. More reasonable
+		// behavior here would be to respect the PATH env var if it's explicitly
+		// set for the command, but changing this could break existing
+		// workflows, so we don't do that.
+		path := make([]string, 0, len(opts.addToPath)+1)
+		path = append(path, opts.addToPath...)
+		path = append(path, os.Getenv("PATH"))
+		env["PATH"] = strings.Join(path, string(filepath.ListSeparator))
```
hadjri, line 178: Does this cover windows case?

hadjri (changes requested):

agent/command/exec.go
Kimchelly: No, the first case is unconditional because this statement is covering three different valid file path casesthe. For the non-Windows case, the only allowed filepath separator is `/`. However, for Windows, Go accepts either `\` or `/` as a file path separator, so we have to check for either `\` (the native `filepath.Separator`) or `/` (not native to Windows, but still accepted by Go as a file path for Windows) as potential separators.

agent/command/exec.go
Kimchelly: Good idea, added one.

agent/command/exec.go
Kimchelly: Not sure if you mean `PATH` or the `filepath.ListSeparator`, but yes, this is compatible with both Windows and Linux. The `filepath.ListSeparator` is set to the correct native value for Windows (`;`) vs. Linux (`:`). And both platforms accept `PATH` to search for executables (as confirmed by the [docs](https://pkg.go.dev/os/exec?GOOS=windows#LookPath)).

agent/command/exec.go
hadjri: Ah gotcha, thanks for that. Can you add a short comment mentioning this since it's not immediately clear?

hadjri (approved):
LGTM with one last req

agent/command/exec.go
Kimchelly: Sounds good, added it.

PR #6860, opened 2023-08-02
EVG-19712 add job to check if volume is stuck with terminated host

ablack12:
EVG-19712
### Description
I recognize this job might be overkill, but I spent a good amount of time reading how this state could happen; it seems to me like if host terminate is ever called without also checkAndTerminateCloudHost this could happen, or if that gets interrupted. This also hasn't occurred in the last year from the query I ran (details in ticket) so there's nothing for us to look at on splunk to see how this happened, so it's hard to confirm for sure how these got in this state -- some have an actual uptime, and some seem terminated immediately, so it's possible more could go into this state.

Because the query is pretty cheap (<500 volumes match the original match condition in prod) I think it may be worth keeping this in here to try and keep data consistent. I can be convinced to not do this if there's another idea about how to ensure we don't get here which doesn't involve a major refactor.

### Testing
Added unit test; will ensure hosts (listed in EVG-19712) are marked as unmounted before closing this ticket.


hadjri (commented):
Implementation looks good to me but maybe we should create a followup ticket to check back up on this and see if the job caught anything, and if so we can use Splunk to try to re-investigate how this state occurs in the first place (and if patchable, remove this job)?

units/crons.go
```diff
@@ -1030,6 +1030,23 @@ func PopulateVolumeExpirationJob() amboy.QueueOperation {
...
+func PopulateUnstickVolumesJob() amboy.QueueOperation {
+	return func(ctx context.Context, queue amboy.Queue) error {
+		catcher := grip.NewBasicCatcher()
+		volumes, err := host.FindVolumesWithTerminatedHost()
+		if err != nil {
+			return errors.Wrap(err, "finding volumes to delete")
+
+		}
+		for _, v := range volumes {
+			ts := utility.RoundPartOfHour(0).Format(TSFormat)
```
malikchaya2, line 1042: I wonder if this should maybe be done daily instead of hourly?

model/host/volume.go
```diff
@@ -124,6 +125,33 @@ func FindVolumesToDelete(expirationTime time.Time) ([]Volume, error) {
...
+			"from":         Collection,
+			"localField":   "host",
+			"foreignField": "_id",
+			"as":           "host_doc",
+		}}
+	matchTerminatedHosts := bson.M{
+		"$match": bson.M{
+			"host_doc.status": evergreen.HostTerminated,
+		}}
+	project := bson.M{"$project": bson.M{"host_doc": 0}}
```
malikchaya2, line 146: It looks like we are not querying on this field rn. Should we add an index?

malikchaya2 (changes requested):

model/host/volume.go
ablack12: Which field are you referencing Chaya?

malikchaya2: scratch that, I misread what "host_doc" was.

ablack12:
> Implementation looks good to me but maybe we should create a followup ticket to check back up on this and see if the job caught anything, and if so we can use Splunk to try to re-investigate how this state occurs in the first place (and if patchable, remove this job)? 

Not planning on a follow up ticket; going to keep this ticket open for a couple of days after deploy to make sure it's working. I could also make a splunk alert after that to let us know if this occurs again!

units/crons.go
ablack12: I thought about this but I'd say it's as unlikely to occur as some other hourly jobs, and the query isn't very expensive so I don't think it's worth adding a new kind of daily job.

malikchaya2: It looks like we have one here: https://github.com/evergreen-ci/evergreen/blob/main/units/crons.go#L948

hadjri:
> > Implementation looks good to me but maybe we should create a followup ticket to check back up on this and see if the job caught anything, and if so we can use Splunk to try to re-investigate how this state occurs in the first place (and if patchable, remove this job)?
> 
> Not planning on a follow up ticket; going to keep this ticket open for a couple of days after deploy to make sure it's working. I could also make a splunk alert after that to let us know if this occurs again!

No followup ticket sounds good, I think a quick splunk alert would be helpful though 👍

hadjri (approved):

units/crons.go
ablack12: oh haha that's funny bc it's in the hourly list. I'd rather make a split ticket to move both this job and the stats job to an actual daily cron; it's really not expensive enough that hourly is going to be problematic as it is.

ablack12: https://jira.mongodb.org/browse/EVG-20611

malikchaya2: I agree. Thank you for creating the ticket!

PR #6865, opened 2023-08-04
EVG-20505: Save distro resolver

sophstad:
EVG-20505

### Description
- Update save mutation to expect an entire distro object as input.
  - We no longer save by section since this made validation more challenging (i.e. when validating fields that are dependent on a field in another section)
  - Instead of using `$set` for the fields present in a section, replace the entire distro document on save
- All fields are required for input and output distro objects
- Replace Birch's `MapInterfaceErr()` with JSON marshaling/unmarshaling since this is how it's done in [legacy](https://github.com/evergreen-ci/evergreen/blob/01d2c44969e7a1169a581a9325cbbcefc80d1e62/service/distros.go#L132). The Birch method panicked when it encountered a slice value in a map.

### Testing
- Add GraphQL unit tests

### Documentation
N/A


graphql/mutation_resolver.go
sophstad, line 1: This file needs review, not autogenerated 😢

khelif96: I opened a ticket with GitHub support about this. Since AFAICT we have this configured correctly.

graphql/distro_resolver.go
```diff
@@ -25,21 +28,50 @@ func (r *distroResolver) ProviderSettingsList(ctx context.Context, obj *model.AP
...
 	}
 	obj.ProviderSettingsList = settings
 	return nil
 }
 
+// AcceptableHostIdleTime is the resolver for the acceptableHostIdleTime field.
+func (r *hostAllocatorSettingsInputResolver) AcceptableHostIdleTime(ctx context.Context, obj *model.APIHostAllocatorSettings, data int) error {
+	obj.AcceptableHostIdleTime = model.NewAPIDuration(time.Duration(obj.AcceptableHostIdleTime))
+	return nil
+}
```
khelif96, line 50: This confuses me `AcceptableHostIdleTime` is already an `APIDuration` why is it being recast into `Duration` and then an `APIDuration` again?

graphql/distro_resolver.go
```diff
@@ -25,21 +28,50 @@ func (r *distroResolver) ProviderSettingsList(ctx context.Context, obj *model.AP
...
+
+// TargetTime is the resolver for the targetTime field.
+func (r *plannerSettingsInputResolver) TargetTime(ctx context.Context, obj *model.APIPlannerSettings, data int) error {
+	obj.TargetTime = model.NewAPIDuration(time.Duration(obj.TargetTime))
```
khelif96, line 54: Same comment here.

graphql/distro_resolver.go
sophstad: This is a resolver for the input field specifically, which is an `Int`. I ran into trouble passing a `Duration` object to the resolver as an input field; we don't do that anywhere else in the codebase and I'm not sure what it would look like. If you know how that could be done we can avoid this casting from `Int` to `APIDuration`!

graphql/mutation_resolver.go
```diff
@@ -178,34 +177,28 @@ func (r *mutationResolver) CreateDistro(ctx context.Context, opts CreateDistroIn
 // SaveDistroSection is the resolver for the saveDistroSection field.
```
ablack12, line 177: Can we add to this comment to indicate that we do input the entire distro object as opposed to just the individual section?

ablack12 (changes requested):

graphql/mutation_resolver.go
```diff
@@ -175,37 +174,32 @@ func (r *mutationResolver) CreateDistro(ctx context.Context, opts CreateDistroIn
...
 
-	updatedDistro, err := distro.UpdateDistroSection(ctx, originalDistro, distroChanges, opts.Section, usr.Username())
-	if err != nil {
-		return nil, InternalServerError.Send(ctx, fmt.Sprintf("updating distro '%s': %s", opts.DistroID, err.Error()))
+	if err = d.ReplaceOne(ctx); err != nil {
+		return nil, InternalServerError.Send(ctx, err.Error())
 	}
-	apiDistro := &restModel.APIDistro{}
-	apiDistro.BuildFromService(*updatedDistro)
+	event.LogDistroModified(d.Id, usr.Username(), d.NewDistroData())
```
ablack12, line 194: One thing to call out here is that this function only logs the new distro -- it doesn't log both the original and the new and handle things intelligently the way that Project Settings does, and fixing this was initially intended to be a goal of changing to the new UI so we wouldn't have to handle this multiple times (https://jira.mongodb.org/browse/EVG-14885). That being said, I don't see it as an explicit scope goal, but it is still in the project. Is this still something we're going to consider improving? (can still be work outside this ticket but then we may want to add a todo here to log more intelligently)

graphql/mutation_resolver.go
```diff
@@ -175,37 +174,32 @@ func (r *mutationResolver) CreateDistro(ctx context.Context, opts CreateDistroIn
 	}, nil
 }
 
-// SaveDistroSection is the resolver for the saveDistroSection field.
+// SaveDistroSection is the resolver for the saveDistroSection field. The entire distro object is provided as input (not just the updated fields) in order to validate all distro settings.
```
ablack12, line 177: nit: can we make this two lines since it's a bit long?

ablack12 (approved):

graphql/mutation_resolver.go
sophstad: Sadly gqlgen seems to impose a limit of one line of comments before a resolver. When I use two lines it tries to force this:
```go
// SaveDistroSection is the resolver for the saveDistroSection field. The entire distro object is provided as input
(not just the updated fields) in order to validate all distro settings.
	func (r *mutationResolver) SaveDistroSection(ctx context.Context, opts SaveDistroInput) ( *SaveDistroSectionPayload,  error){
```

graphql/distro_resolver.go
minnakt: hmm I'm also confused because like Mohamed said, `AcceptableHostIdleTime` should already be an `APIDuration`. Are we not supposed to be using the `data` field (which is an `int`) instead here?

khelif96: I think there may be a typo here it doesn't look like you are doing anything with an int. Did you mean to pass in data?

PR #6867, opened 2023-08-04
EVG-20440 Move getNewTasksWithDependencies to its own function

hadjri:
EVG-20440

### Description
Currently, the command runs `getNewTasksWithDependencies` twice: once when simulating dependencies to check for cycles, and another time when actually creating the generated tasks, with the only difference being that the latter call passes in the generator's activation info, which determines if tasks are created in an active or inactive state, but since that doesn't matter for computing dependency cycles, it's fine to consolidate the two calls.  For large JSONs, this step can take a solid chunk of time (~30 seconds or more), so running this function once and saving its output to the generator struct will speed things up.
### Testing
Created a dummy generated JSON in the sandbox project with ~20,000 tasks and 20 variants to emulate what server does. Honeycomb trace shows a reduction in execution time from 2 minutes to ~80 seconds (first spike is with change, second is without change):

Screenshot 2023-08-04 at 4 08 23 PM

[Trace with change
](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/pambB5RCdi6/trace/wpwt8yRAhPH?fields[]=c_name&fields[]=c_service.name&span=926bc920ee2817c7)

image

[Trace without change](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/pambB5RCdi6/trace/t2dDUTmnnc6?fields[]=c_name&fields[]=c_service.name&span=ded0ab12fd66170d)

image

Inspecting these two links you can see that the reason the trace with the change is shorter is because the `get-new-tasks-with-dependencies` step doesn't run twice

### DIFF
--- a/model/generate.go
+++ b/model/generate.go
@@ -31,8 +31,8 @@ type GeneratedProject struct {
 	TaskGroups    []parserTaskGroup          `yaml:"task_groups"`
 
 	Task           *task.Task
-	ActivationInfo specificActivationInfo
-	NewTVPairs     TaskVariantPairs
+	ActivationInfo *specificActivationInfo
+	NewTVPairs     *TaskVariantPairs
 }
 
 // MergeGeneratedProjects takes a slice of generated projects and returns a single, deduplicated project.
@@ -234,17 +234,18 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		buildSet[b.BuildVariant] = struct{}{}
 	}
 
+	newTVPairs, activationInfo := g.GetNewTasksAndActivationInfo(ctx, v, p)
 	// Group into new builds and new tasks for existing builds.
 	newTVPairsForExistingVariants := TaskVariantPairs{}
 	newTVPairsForNewVariants := TaskVariantPairs{}
-	for _, execTask := range g.NewTVPairs.ExecTasks {
+	for _, execTask := range newTVPairs.ExecTasks {
 		if _, ok := buildSet[execTask.Variant]; ok {
 			newTVPairsForExistingVariants.ExecTasks = append(newTVPairsForExistingVariants.ExecTasks, execTask)
 		} else {
 			newTVPairsForNewVariants.ExecTasks = append(newTVPairsForNewVariants.ExecTasks, execTask)
 		}
 	}
-	for _, dispTask := range g.NewTVPairs.DisplayTasks {
+	for _, dispTask := range newTVPairs.DisplayTasks {
 		if _, ok := buildSet[dispTask.Variant]; ok {
 			newTVPairsForExistingVariants.DisplayTasks = append(newTVPairsForExistingVariants.DisplayTasks, dispTask)
 		} else {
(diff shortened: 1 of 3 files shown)


model/generate.go
```diff
@@ -312,6 +299,15 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 	return nil
 }
 
+// SetNewTasksAndActivationInfo sets the NewTVPairs and ActivationInfo fields on the generator.
+func (g *GeneratedProject) SetNewTasksAndActivationInfo(ctx context.Context, v *Version, p *Project) {
```
Kimchelly, line 303: One suggestion on how to write this a little more cleanly is to change this to a `GetNewTasks` function, then check if `g.NewTVPairs` is initialized, and just return it if so. If not, then initialize it once. Something like:
```go
if g.NewTVPairs != nil {
    // Already initialized
    return g.NewTVPairs
}

// First initialization
g.NewTVPairs = g.getNewTasks...
return g.NewTVPairs
// And then make some equivalent function for activation info as well
```
It would also get around having explicitly call `Set` to ensure the data is set before we actually need the fields.

model/generate.go
```diff
@@ -312,6 +299,15 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
...
+	// Only consider batchtime for mainline builds. We should always respect activate if it is set.
```
Kimchelly, line 304: I know this was copy pasted from the original, but is this comment supposed to be here? Seems kind of a random place to put it. Is it supposed to be inside `findTasksAndVariantsWithSpecificActivations`?

Kimchelly (changes requested):

model/generate.go
hadjri: Yeah it initially prepended the function call so I kept it that way but true it makes more sense to move into `findTasksAndVariantsWithSpecificActivations `

model/generate.go
hadjri: Sounds good, changed!

model/generate.go
```diff
@@ -334,19 +338,20 @@ func (g *GeneratedProject) CheckForCycles(ctx context.Context, v *Version, p *Pr
 // simulateNewTasks adds the tasks we're planning to add to the version to the graph and
 // adds simulated edges from each task that depends on the generator to each of the generated tasks.
 func (g *GeneratedProject) simulateNewTasks(ctx context.Context, graph task.DependencyGraph, v *Version, p *Project, projectRef *ProjectRef) (task.DependencyGraph, error) {
+	newTVPairs, _ := g.GetNewTasksAndActivationInfo(ctx, v, p)
```
Kimchelly, line 341: Since you only generally need one or the other, you can split the `Get` function into two functions (one for the TV pairs, the other for the activatio ninfo) so you don't have to omit values. It also makes the initialization of the two values a little cleaner, so you only initialize the value of relevance to you the first time you need it, rather than initializing both at once.

model/generate_test.go
```diff
@@ -1373,7 +1373,7 @@ func TestSimulateNewDependencyGraph(t *testing.T) {
 				},
 			},
 		}
-		g.SetNewTasksAndActivationInfo(context.Background(), v, project)
+		g.GetNewTasksAndActivationInfo(context.Background(), v, project)
```
Kimchelly, line 1376: nit: since all these tests are using `context.Background()`, I would recommend just initializing the context once at the top of the test:
```go
defer cancel()

// Test...
```

Kimchelly (approved):
LGTM, I made some not-that-important organizational suggestions.

model/generate.go
hadjri: Unfortunately the two values can't be initialized independently because they're intertwined in a way that computes them both at the same time. 

`newTasks := g.getNewTasksWithDependencies(ctx, v, p, &activationInfo)` is the line that finalizes both values, so the only way to accomplish this is to run this function twice which nullifies the performance improvement of this ticket

hadjri:
Re-requesting just to make sure we're on the same page about this [comment](https://github.com/evergreen-ci/evergreen/pull/6867#discussion_r1287609662)

model/generate.go
Kimchelly: Hmm yeah, I see that. Oh well, guess this looks good as is then.

Kimchelly (approved):

PR #6875, opened 2023-08-08
EVG-20597 Only process one GitHub webhook event depending on app settings

bynn:
EVG-20597

### Description
 github events from apps are now ignored if app auth is not set up 
also deletes the admin setting validation for the GitHub section because it seems irrelevant 

### Testing
created a staging pr and made sure that none of the events went through 
adding the app id back to the settings lets the event go through again

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -102,10 +102,12 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 }
 
 func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
+	hasApp := gh.settings.GetGithubAppAuth() != nil
 	switch event := gh.event.(type) {
 	case *github.PingEvent:
 		// Ignore events from GitHub app if app is not set up.
-		if event.GetInstallation() != nil && gh.settings.GetGithubAppAuth() == nil {
+		// If the app is set up, ignore events from repo webhooks.
+		if (event.GetInstallation() != nil) != hasApp {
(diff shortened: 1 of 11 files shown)


ablack12 (commented):
Discussed adding an else to ignore events from regular webhook if app is set up

rest/route/github.go
```diff
@@ -102,8 +102,14 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 }
 
 func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
+	hasApp := gh.settings.GetGithubAppAuth() != nil
 	switch event := gh.event.(type) {
 	case *github.PingEvent:
+		// Ignore events from GitHub app if app is not set up.
+		// If the app is set up, ignore events from repo webhooks.
+		if (event.GetInstallation() != nil) != hasApp {
```
ablack12, line 110: this logic is a little awkward; if they're both not nil or both nil then we want to continue? We may want to move this all out to a helper function that just explains itself (could pass in the event and re-switch on it above this or something)

ablack12 (changes requested):

rest/route/github.go
bynn: its logically equivalent to 
if (fromApp && !hasApp) ||  (!fromApp && hasApp) {skip} 
which the comment describes more directly

would it be better to make it more clear?

ablack12 (approved):

PR #6877, opened 2023-08-08
EVG-20524: upload smoke test logs to S3 and fix closed sender

Kimchelly:
EVG-20524

### Description
This improves the task log debugging for the smoke test but also fixes a bug related to the agent file log that was being relied upon for both the smoke test and prod.

* Fix an issue where the global sender (i.e. the agent file log) was being closed because `SetSender` (which closes the current logger and sets the new one) was called after the task logger was initialized. For files, this caused the file to be closed, and prevented further logging. Incidentally, this prematurely closed logger bug was helping prod maintain the correct behavior (i.e. not logging task logs to the agent file log), while allowing the smoke test logs to stdout (because closing the stdout logger is a no-op). Instead of having prod and the smoke test rely on this bug to unintentionally maintain the desired behavior, I exposed a new agent CLI flag that will let us log task logs conveniently to the agent file log for smoke tests (so we can read them directly), while ensuring that prod still does not send task logs to this file.
* Send agent and agent monitor logs in the smoke test to S3 files. This is kind of preferable because before this change, the logs would just go to stdout in an unorganized way, so organizing the logs into one file per task run is easier to look through.
* Include the task ID/execution in the agent log file name so that for smoke test debugging (and live prod debug sessions in hosts), we can more easily deduce which agent log file is related to which task.
* Add some comments about task logger flushing vs. closing so future readers know to be careful not to do one or the other in the wrong scenario (this might go away with the logging rework, but it's good to have just in case).
* Drive-by tweak a small but potentially confusing comment in subprocess.exec.

### Testing
Ran the smoke test and tested in staging that the agent log file behaved as desired.

### Documentation
I added a mention to the smoke test doc about checking the files tab for the agent logs.


agent/agent.go
```diff
@@ -588,12 +595,6 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (shouldExit bool,
 	tc.taskConfig.WorkDir = tc.taskDirectory
 	tc.taskConfig.Expansions.Put("workdir", tc.taskConfig.WorkDir)
 
-	grip.Info(message.Fields{
-		"message":     "running task",
-		"task_id":     tc.task.ID,
-		"task_secret": tc.task.Secret,
-	})
-
```
malikchaya2, line 596: why is this being removed?

agent/internal/client/logger_producer.go
```diff
@@ -45,6 +48,10 @@ func (l *logHarness) Flush(ctx context.Context) error {
 	return catcher.Resolve()
 }
 
+// Close closes the all the task loggers and prevents further writes to it. Note
```
malikchaya2, line 51: nit: "the task logger" or "all the task loggers

local_smoke_test.sh
```diff
@@ -0,0 +1 @@
+/home/kim/go/src/github.com/evergreen-ci/evergreen/kim/local_smoke_test.sh
```
malikchaya2, line 1: Did you mean to leave this in?

malikchaya2 (changes requested):

agent/agent.go
Kimchelly: I removed it because it didn't seem like it served any purpose and it logs the task secret, which we should avoid logging if possible.

agent/internal/client/logger_producer.go
Kimchelly: Fixed.

local_smoke_test.sh
Kimchelly: Hmm, no I removed it.

agent/agent.go
malikchaya2: We can just remove the secret part. I think it could be helpful when debugging things.

Kimchelly: Sure, I re-added without the secret.

malikchaya2 (approved):

PR #6880, opened 2023-08-08
EVG-20586 Fix flaky test-agent on OSX 

malikchaya2:
EVG-20586

### Description
This rewrites the test to check the task missing a secret condition using the processNext function instead of relying on the agent loop context to test the condition which is more error prone.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -160,26 +160,16 @@ func (s *AgentSuite) TestNextTaskResponseShouldExit() {
 }
 
 func (s *AgentSuite) TestTaskWithoutSecret() {
-	s.mockCommunicator.NextTaskResponse = &apimodels.NextTaskResponse{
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):

agent/agent_test.go
```diff
@@ -160,26 +160,16 @@ func (s *AgentSuite) TestNextTaskResponseShouldExit() {
...
-	select {
-	case err := <-errs:
-		s.NoError(err)
-	case <-ctx.Done():
-		s.FailNow(ctx.Err().Error())
-	}
+	ntr, err := s.a.processNextTask(s.ctx, nextTask, s.tc, false)
+
+	s.NoError(err)
+	s.NotNil(ntr)
```
Kimchelly, line 171: `s.Require().NotNil` or else L172 can panic.

PR #6881, opened 2023-08-08
EVG-19745 Move container image uri to immutable tag

hadjri:
EVG-19745

### Description
Followup from [this PR](https://github.com/evergreen-ci/container-initial-offering-dockerfiles/pull/6) in the container-initial-offering-dockerfiles repo, we no longer need to store the Dockerfile in this repo and we should use an immutable tag rather than `:latest`


Kimchelly (approved):
I dunno if you already did this, but before merging, make sure you update the allowed images (or do the thing where we allow images by repo name instead of specific tags).

hadjri:
> I dunno if you already did this, but before merging, make sure you update the allowed images (or do the thing where we allow images by repo name instead of specific tags).

Will update the list, I created a ticket to do the smarter version of validation: https://jira.mongodb.org/browse/EVG-20609

hadjri:
evergreen retry

PR #6882, opened 2023-08-08
EVG-17126 Remove temporary debug logs

hadjri:
EVG-17126

### Description
Removing these logs that were intended to be temporary and only for debugging / verification purposes.


bynn (approved):
lgtm

PR #6883, opened 2023-08-08
EVG-20608 Reduce the noise for the task missing secret condition

malikchaya2:
EVG-20608 

### Description
If the agent encounters a task that is missing a secret, there is no point in immediately trying to get another task, because it will keep getting the same task with the missing secret until enough time has passed. Instead, we should make it sleep for longer to reduce the amount of times it logs "task response missing secret". 

### Testing
Added to the unit test.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -169,8 +169,8 @@ func (s *AgentSuite) TestTaskWithoutSecret() {
 
 	s.NoError(err)
 	s.Require().NotNil(ntr)
-	s.Equal(false, ntr.shouldExit)
-	s.Equal(true, ntr.noTaskToRun)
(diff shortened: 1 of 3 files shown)


agent/agent_test.go
```diff
@@ -170,6 +170,7 @@ func (s *AgentSuite) TestTaskWithoutSecret() {
 	s.NoError(err)
 	s.Require().NotNil(ntr)
 	s.Equal(false, ntr.shouldExit)
+	s.Equal(true, ntr.noTaskToRun)
```
Kimchelly, line 173: nit: rather than `s.Equal` you can just use `s.False` above this and `s.True` here.

Kimchelly (approved):
LGTM mod nit.

PR #6885, opened 2023-08-09
EVG-19779 instrument Evergreen tests

ybrill:
[EVG-19779](https://jira.mongodb.org/browse/EVG-19779)

### Description
Add a framework for instrumenting tests. Instrument `test-units` as a POC.
To instrument a test you call `testutil.TestSpan(ctx, t)` and it will 
- initialize the tracer so traces from the tested code are captured, such as calls to the db (that use the ctx)
- create a span for the test (pass the context to a subtest to create a nested span).

The testing framework has a [Cleanup](https://pkg.go.dev/testing#B.Cleanup) function which runs after a test and all its subtests complete. This is used for closing the span (for convenience) and flushing/closing the span exporter (if it's the root span for the test). The tracer provider is set globally and closing it would be a problem if something tried to send a span after it was closed, but this is okay because cleanup is guaranteed to run after the test is finished and tests aren't running in parallel.

There were some changes to the makefile/self-tests.yml to separate out building the test binary (`go test -c`) from the actual test.


ybrill:
An [task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_test_units_patch_6b2e6a1eb98afbf71c1cae3558acfc6754083b61_64d3ab86b23736ade5ae6438_23_08_09_15_07_41/tests?execution=0&sortBy=STATUS&sortDir=ASC) and [its trace](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen-agent/trace/2NUE4XkGygY)

globals.go
```diff
@@ -468,6 +468,8 @@ type ModificationAction string
 const (
 	PackageName = "github.com/evergreen-ci/evergreen"
 
+	OtelAttributeMaxLength = 10000
```
opsnlops, line 471: This is Honeycomb's limit, right?

units/cache_historical_task_data_test.go
```diff
@@ -247,22 +249,24 @@ func TestCacheHistoricalTaskDataJob(t *testing.T) {
...
+
 			require.NoError(t, db.ClearCollections(task.Collection, taskstats.DailyTaskStatsCollection, taskstats.DailyStatsStatusCollection))
 			if test.pre != nil {
-				test.pre(t)
+				test.pre(tctx, t)
 			}
 
 			j := NewCacheHistoricalTaskDataJob("id", "p0")
 			if test.hasErr {
-				j.Run(context.TODO())
```
opsnlops, line 257: What was this?

testutil/otel.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
```
opsnlops, line 20: The current version of this package is v1.20.0. Is there any reason we're pinned to 1.17?

testutil/otel.go
```diff
@@ -0,0 +1,149 @@
...
+		return spanForChildTest(ctx, t)
+	}
+}
+
+func spanForRootTest(ctx context.Context, t *testing.T) context.Context {
+	collectorEndpoint := os.Getenv(otelCollectorEndpointEnvVar)
+	traceIDString := os.Getenv(otelTraceIDEnvVar)
+	spanIDString := os.Getenv(otelParentIDEnvVar)
+	if collectorEndpoint == "" || traceIDString == "" || spanIDString == "" {
+		return ctx
```
opsnlops, line 45: No error message?

opsnlops (commented):
I only glanced over all of the test changes, since they all seem to more-or-less be changing the same thing.

globals.go
ybrill: This limit is self-imposed (HC doesn't care as long as the whole event is under 100kb). 
In https://github.com/evergreen-ci/evergreen/pull/6769 we limited the string length to prevent batches from being over the gRPC message size limit and it was determined that this limit should be enough for p99 of the `db.statement` strings.

units/cache_historical_task_data_test.go
ybrill: do you mean what is the purpose of a `context.TODO()`?

globals.go
opsnlops: Okay, good. It's better that we code to our _own_ limits, and not ones that Honeycomb imposes. That helps us keep flexible in case we want to move OTel providers in the future.

testutil/otel.go
ybrill: I think it should be okay to go to the latest.
done

testutil/otel.go
ybrill: This condition is like if someone is running the test locally so these variables won't be set in the environment

units/cache_historical_task_data_test.go
ybrill: yeah, idk why they stuck `context.TODO()`s randomly in various places. At least now there will be fewer of them 😄

testutil/otel.go
opsnlops: Okay, cool.

testutil/otel.go
opsnlops: 👍🏻

units/host_termination_test.go
```diff
@@ -238,6 +238,7 @@ func TestHostTerminationJob(t *testing.T) {
 			require.NoError(t, db.ClearCollections(host.Collection, event.EventCollection))
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
+			ctx = testutil.TestSpan(ctx, t)
```
julianedwards, line 241: Should there be a parent test span?

julianedwards (changes requested):

testutil/otel.go
ybrill: done

units/host_termination_test.go
ybrill: fixed

units/host_drawdown_test.go
```diff
@@ -42,7 +43,10 @@ func numHostsTerminated(ctx context.Context, env evergreen.Environment, drawdown
...
 
 func TestTerminatingHosts(t *testing.T) {
-	ctx := context.Background()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ctx = testutil.TestSpan(ctx, t)
+
 	env := evergreen.GetEnvironment()
 
 	t.Run("SimpleTerminationTest", func(t *testing.T) {
```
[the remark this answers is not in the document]
ybrill: we could make another span for the subtest. 
It seems like a waste, though, since it's just this one subtest.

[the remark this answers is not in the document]
julianedwards: It seems like the subtest span will include the subtest name, and the parent span wouldn't, correct? Also, if someone updates this test in the future it will be more clear that a new subtest should have its own span.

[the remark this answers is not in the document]
ybrill: ah, okay. That makes sense
done

PR #6887, opened 2023-08-09
EVG-19944: Change APIPlannerSettings.GroupVersions from *bool -> bool

minnakt:
EVG-19944

### Description
While working on the general settings page, I kept running into this issue with `groupVersions`:

Screenshot 2023-08-09 at 1 25 23 PM

Changing `APIPlannerSettings.GroupVersions` to be `bool` rather than `*bool` resolves this error.


sophstad (approved):
LGTM but makes sense to have an app engineer confirm! (Also noting that every instance of this field in the prod DB is `false` so this change shouldn't be impactful)

rest/model/distro.go
```diff
@@ -14,7 +14,7 @@ import (
 type APIPlannerSettings struct {
 	Version                   *string     `json:"version"`
 	TargetTime                APIDuration `json:"target_time"`
-	GroupVersions             *bool       `json:"group_versions"`
```
ablack12, line 17: I'm fine with this in this case since we aren't very consistent about using *bool for API structs, but do we understand why this error is occurring? This change would be a harder sell for strings for example, or structs that are omit empty.

minnakt: I don't really understand why this happens... I feel like the boolean pointer should be getting coerced by GraphQL into `false` when it's `nil`, but it does not 😭 The string pointers are properly getting coerced into `""` so they are okay.

The reason the GraphQL schema error occurs is that `groupVersions` is required to be a [`Boolean!`](https://github.com/evergreen-ci/evergreen/blob/1e2de6c2077425d29d40f90aa8c31c8039f8fc45/graphql/schema/types/distro.graphql#L268) (a non-null boolean), which makes `groupVersions: null` an invalid value for the schema. I think it does make more sense for it to be a plain boolean through, since on the UI it's a concrete choice between "Yes" or "No":

Screenshot 2023-08-10 at 10 25 24 AM

sophstad: What Minna said, plus I'd add that we definitely can work with a `null`/`bool` field (there's a lot of that in Project Settings), but it makes sense to tighten up the typing here since group versions will be represented as `true` or `false` in the UI.

ablack12 (approved):

PR #6888, opened 2023-08-09
EVG-20018: Add new task field `LogServiceVersion` 

julianedwards:
EVG-20018

### Description
This adds a new field `LogServiceVersion` to the `Task` struct and creates a function to atomically set the field. Since this field should only be set once during a task run, a subsequent call to the field will fail.

### Testing
Added unit tests

### Documentation
N/A


model/task/task.go
```diff
@@ -1442,6 +1443,40 @@ func (t *Task) SetStepbackDepth(stepbackDepth int) error {
...
+	res, err := env.DB().Collection(Collection).UpdateByID(ctx, t.Id, []bson.M{
+		{
+			"$set": bson.M{LogServiceVersionKey: bson.M{
+				"$ifNull": bson.A{
+					"$" + LogServiceVersionKey,
+					version,
+				}},
+			},
+		},
+	})
```
ybrill, line 1465: Why is it preferable to include the Null check in the update instead of in the filter?
Like why couldn't something like this work?
```go
res, err := env.DB().Collection(Collection).UpdateOne(ctx,
    bson.M{IdKey: t.Id, LogServiceVersionKey: nil},
    bson.M{"$set": bson.M{LogServiceVersionKey: version}},
)
```

julianedwards: It could, but then we wouldn't know whether there is a programmatic error (the task ID should should exist does not exist) or the field is NULL.

ybrill (approved):

PR #6878, opened 2023-08-08
EVG-20603: Improve distro modified logging

sophstad:
EVG-20603

### Description
add description, context, thought process, etc

### Testing
add a description of how you tested it

### Documentation
remember to add or edit docs in the docs/ directory if relevant

PR #6889, opened 2023-08-09
EVG-20621: Fix old distro pointing to new distro struct

sophstad:
EVG-20621

### Description
- An existing bug set `newDistro` to equal the pointer of `oldDistro` when making a distro update. This meant that applying the user's changes to `newDistro` also applied them to `oldDistro`. In the updated distro modified event logger from #6878, we stop logging if the "before" and "after" structs are equal, so an event was not being added.
- I also removed the equality check for safety in the meantime (without it, the legacy distro event log functions correctly since it does not rely on a comparison between the old and new structs)

### Testing
- Tested manually since there's no test for `modifyDistro()`

### Documentation
N/A

### DIFF
--- a/model/event/distro_event.go
+++ b/model/event/distro_event.go
@@ -1,7 +1,6 @@
 package event
 
 import (
-	"reflect"
 	"time"
 
 	"github.com/mongodb/grip"
@@ -62,11 +61,6 @@ func LogDistroAdded(distroId, userId string, data interface{}) {
 
 // LogDistroModified should take in DistroData in order to preserve the ProviderSettingsList
 func LogDistroModified(distroId, userId string, before, after interface{}) {
-	// Stop if there are no changes
-	if reflect.DeepEqual(before, after) {
-		return
-	}
-
 	data := DistroEventData{
 		UserId: userId,
 		User:   userId,
--- a/model/event/distro_event.go
+++ b/model/event/distro_event.go
@@ -1,6 +1,7 @@
 package event
 
 import (
+	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


model/event/distro_event.go
```diff
@@ -62,11 +61,6 @@ func LogDistroAdded(distroId, userId string, data interface{}) {
 
 // LogDistroModified should take in DistroData in order to preserve the ProviderSettingsList
 func LogDistroModified(distroId, userId string, before, after interface{}) {
-	// Stop if there are no changes
```
ablack12, line 65: re removing this for safety -- does it seem probable that this fixed it? Or do we just want to log here if we hit this line for a bit since presumably this doesn't happen? We will want to add it in eventually i imagine.

service/distros.go
```diff
@@ -126,7 +126,8 @@ func (uis *UIServer) modifyDistro(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	newDistro := oldDistro
+	newDistro := &distro.Distro{}
+	*newDistro = *oldDistro
```
ablack12, line 130: Does newDistro need to be a pointer at all? We may just be able to set it to *oldDistro directly (I'd also suggest adding a lil comment here so it's clear that this was intentional and not wonky)

ablack12 (changes requested):

model/event/distro_event.go
sophstad: It does seem probable; the only other place this function is called is from REST routes that don't do any struct copying like this. Stopping + logging is a good idea, added!

PR #6890, opened 2023-08-10
EVG-20436: Add DistroEvents query resolver

sophstad:
EVG-20436

### Description

### Testing

### Documentation


graphql/query_resolver.go
sophstad, line 1: Not generated, please review!

graphql/mutation_resolver.go
sophstad, line 1: Also needs review

.gitattributes
```diff
@@ -1,4 +1,4 @@
 *.sh text eol=lf
 ./public/static/dist/* binary
 ./public/package-lock.json binary
-graphql/*_resolver.go linguist-detectable linguist-language=go
```
khelif96, line 4: I've escalated this with github again.

graphql/schema/types/distro.graphql
```diff
@@ -166,6 +178,20 @@ type DeleteDistroPayload {
   deletedDistroId: String!
 }
 
+
+type DistroEvent {
+  after: Map
+  before: Map
+  data: Map
```
khelif96, line 185: I think this is an anti-pattern and we should reduce our usage of the Map type. I understand why it's used here but we should try to reduce our usage/dependence on it. We lose all of the benefits of Graphql for these fields when we use it. (Nonactionable just wanted to state my thoughts)

khelif96 (changes requested):
Great job just had one comment on how we return errors.

khelif96 (approved):
LGTM!

graphql/util.go
```diff
@@ -1060,3 +1061,58 @@ func userHasDistroPermission(u *user.DBUser, distroId string, requiredLevel int)
...
+	}
+
+	mapField := map[string]interface{}{}
+	marshalledData, err := bson.Marshal(data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "marshalling data")
+	}
+
+	err = bson.Unmarshal(marshalledData, &mapField)
+	if err != nil {
```
ablack12, line 1113: nit; combine with 1112

model/event/event_finder.go
```diff
@@ -161,9 +161,9 @@ func TaskEventsInOrder(id string) db.Q {
 // Distro Events
 
 // FindLatestPrimaryDistroEvents return the most recent non-AMI events for the distro.
-func FindLatestPrimaryDistroEvents(id string, n int) ([]EventLogEntry, error) {
+func FindLatestPrimaryDistroEvents(id string, n int, before *time.Time) ([]EventLogEntry, error) {
```
ablack12, line 164: I'm assuming before is used for pagination? Can we add that to the function comment?

ablack12 (changes requested):

graphql/query_resolver.go
```diff
@@ -108,17 +108,39 @@ func (r *queryResolver) SubnetAvailabilityZones(ctx context.Context) ([]string,
 func (r *queryResolver) Distro(ctx context.Context, distroID string) (*restModel.APIDistro, error) {
```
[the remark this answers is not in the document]
sophstad: I did :)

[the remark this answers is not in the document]
ablack12: ohhh i see, Github marked them as outdated so they stopped showing up in the files tab. That's so annoying 🙃 not reasonable to expect you to catch that, disregard

ablack12 (changes requested):

graphql/query_resolver.go
```diff
@@ -121,7 +121,17 @@ func (r *queryResolver) Distro(ctx context.Context, distroID string) (*restModel
 
 // DistroEvents is the resolver for the distroEvents field.
 func (r *queryResolver) DistroEvents(ctx context.Context, opts DistroEventsInput) (*DistroEventsPayload, error) {
-	events, err := event.FindLatestPrimaryDistroEvents(opts.DistroID, utility.FromIntPtr(opts.Limit), opts.Before)
+	before := time.Now()
+	if opts.Before != nil {
+		before = *opts.Before
```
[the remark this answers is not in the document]
sophstad: That makes sense. I also updated a lot of the usages of `FindLatestPrimaryDistroEvents` to pass in `utility.ZeroTime` instead of `time.Now()` to avoid the extra clause. Both should work though (and both are tested).

ablack12 (approved):
🚀

sophstad:
evergreen refresh

PR #6893, opened 2023-08-10
EVG-20278 Reorganize setup.initial for task setup

malikchaya2:
EVG-20278

### Description
This moves a bunch of logic that is conceptually part of setting up the task into a setupTask function. It also removes some things that are not needed because they are duplicates. 

### Testing
Added a test and fixed existing tests


globals.go
```diff
@@ -320,6 +320,10 @@ const (
 
 	DefaultShutdownWaitSeconds = 10
 
+	// HeartbeatTimeoutThreshold is the default timeout for how long a task can run without sending
```
Kimchelly, line 323: Remove "default" since it's not a configurable value - the app server will system-fail a task that exceeds this.

agent/agent.go
```diff
@@ -557,53 +626,19 @@ func (a *Agent) runTask(ctx context.Context, tc *taskContext) (shouldExit bool,
...
-
-	var taskConfig *internal.TaskConfig
-	taskConfig, err = a.makeTaskConfig(ctx, tc)
-	if err != nil {
-		err = errors.Wrap(err, "making task config")
-		grip.Error(err)
-		grip.Infof("Task complete: '%s'.", tc.task.ID)
-		tc.logger = client.NewSingleChannelLogHarness("agent.error", a.defaultLogger)
-		return a.handleTaskResponse(tskCtx, tc, evergreen.TaskSystemFailed, err.Error())
+	// setup should happen before the heartbeat so that it's not abortable
```
Kimchelly, line 629: `// Setup occurs before the task is actually running, so it's not abortable. If setup is taking a long time, timing it out after the heartbeat timeout ensures the operation will give up within the same timeout as the usual amount of time a task is allowed to run before it must heartbeat.`

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	var taskConfig *internal.TaskConfig
+	taskConfig, err = a.makeTaskConfig(setupCtx, tc)
+	if err != nil {
+		setupFailed = true
+		err = errors.Wrap(err, "making task config")
+		grip.Error(err)
+		grip.Infof("Task complete: '%s'.", tc.task.ID)
+		tc.logger = client.NewSingleChannelLogHarness("agent.error", a.defaultLogger)
+		shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
+		return tc, setupFailed, shouldExit, err
```
Kimchelly, line 471: Seems like this can just return true rather than setting a variable to true and returning it.

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	}
+	tc.setTaskConfig(taskConfig)
+	if err = a.startLogging(agentCtx, tc); err != nil {
+		setupFailed = true
+		err = errors.Wrap(err, "setting up logger producer")
+		grip.Error(err)
+		grip.Infof("Task complete: '%s'.", tc.task.ID)
+		tc.logger = client.NewSingleChannelLogHarness("agent.error", a.defaultLogger)
+		shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
+		return tc, setupFailed, shouldExit, err
```
Kimchelly, line 481: Same thing as above.

agent/agent.go
Kimchelly: We should also explicitly return false for shouldExit, rather than assume that another piece of code higher up will set it for us. Generally, returning the named output parameter is a code smell because it makes it difficult to deduce when another unrelated return statement changes it value due to setting it.

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	if !tc.ranSetupGroup {
+		tc.taskDirectory, err = a.createTaskDirectory(tc)
+		if err != nil {
+			setupFailed = true
+			err = errors.Wrap(err, "creating task directory")
+			grip.Error(err)
+			grip.Infof("Task complete: '%s'.", tc.task.ID)
+			tc.logger.Execution().Error(errors.Wrap(err, "creating task directory"))
+			shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
+			return tc, setupFailed, shouldExit, err
```
Kimchelly, line 493: Same thing as above.

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	tc.logger.Task().Infof("Task logger initialized (agent version '%s' from Evergreen build revision '%s').", evergreen.AgentVersion, evergreen.BuildRevision)
+	tc.logger.Execution().Info("Execution logger initialized.")
+	tc.logger.System().Info("System logger initialized.")
+
+	if err := setupCtx.Err(); err != nil {
+		setupFailed = true
+		grip.Error(err)
+		tc.logger.Execution().Infof("Stopping task execution: %s", err)
+		shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
+		return tc, setupFailed, shouldExit, err
```
Kimchelly, line 508: Same thing as above.

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	}
+
+	hostname, err := os.Hostname()
+	tc.logger.Execution().Info(errors.Wrap(err, "getting hostname"))
+	if hostname != "" {
+		tc.logger.Execution().Infof("Hostname is '%s'.", hostname)
+	}
+	tc.logger.Task().Infof("Starting task '%s', execution %d.", tc.taskConfig.Task.Id, tc.taskConfig.Task.Execution)
+
+	return tc, setupFailed, shouldExit, nil
```
Kimchelly, line 518: It seems like the value of `setupFailed` is always equivalent to the returned error being non-nil. Can we consolidate the boolean?

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
-		ranSetupGroup: !shouldSetupGroup,
-		taskDirectory: taskDirectory,
-		oomTracker:    jasper.NewOOMTracker(),
+	if tc.logger != nil {
+		grip.Error(errors.Wrap(tc.logger.Close(), "closing the previous logger producer"))
+	}
+	return shouldSetupGroup, taskDirectory
+}
+
+func (a *Agent) setupTask(agentCtx, setupCtx context.Context, tcInput *taskContext, nt *apimodels.NextTaskResponse, shouldSetupGroup bool, taskDirectory string) (tc *taskContext, setupFailed bool, shouldExit bool, err error) {
```
Kimchelly, line 434: I would put a comment above this stating what this is generally trying to accomplish (like initialize the logger and load the task config data).

Kimchelly, line 434: nit: rather than just call it something like `tcInput`, something like `initialTC` or whatnot might be more illustrative that it's not just an input but also is the "seed" task context.

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+
+	var taskConfig *internal.TaskConfig
+	taskConfig, err = a.makeTaskConfig(setupCtx, tc)
+	if err != nil {
+		setupFailed = true
+		err = errors.Wrap(err, "making task config")
+		grip.Error(err)
+		grip.Infof("Task complete: '%s'.", tc.task.ID)
+		tc.logger = client.NewSingleChannelLogHarness("agent.error", a.defaultLogger)
+		shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
```
Kimchelly, line 470: This return logic (and the others that are the same in this function) seem like they're copy-pasting the same logic. Can we consolidate them into a helper function?

agent/agent.go
```diff
@@ -699,25 +709,26 @@ func (a *Agent) startTask(ctx context.Context, tc *taskContext) (status string)
 
 	// notify API server that the task has been started.
 	tc.logger.Execution().Info("Reporting task started.")
-	if err = a.comm.StartTask(innerCtx, tc.task); err != nil {
+
+	if err := a.comm.StartTask(innerCtx, tc.task); err != nil {
```
Kimchelly, line 713: Since we aren't absorbing some things that apply only to pre and main (like the stats collector) in `runPreAndMain`, can we make a small helper function to also wrap the setup logic in `runPreAndMain`?

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+			ranSetupGroup:             !shouldSetupGroup,
+			taskDirectory:             taskDirectory,
+			oomTracker:                jasper.NewOOMTracker(),
+			unsetFunctionVarsDisabled: nt.UnsetFunctionVarsDisabled,
+		}
+	} else {
+		tc = tcInput
+	}
+	a.jasper.Clear(setupCtx)
+	tc.jasper = a.jasper
```
Kimchelly, line 451: It seems like this `tc.jasper` has no usages. Can we delete the field entirely?

agent/agent.go
```diff
@@ -429,24 +416,106 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+			taskGroup:                 nt.TaskGroup,
+			ranSetupGroup:             !shouldSetupGroup,
+			taskDirectory:             taskDirectory,
+			oomTracker:                jasper.NewOOMTracker(),
+			unsetFunctionVarsDisabled: nt.UnsetFunctionVarsDisabled,
+		}
+	} else {
+		tc = tcInput
+	}
+	a.jasper.Clear(setupCtx)
```
Kimchelly, line 450: It seems like this is trying to clear out any of the old processes from the previous task. Rather than doing that here (when we're setting up a new task), can we do it in `finishPrevTask` (because we're already tying up loose ends from the previous task)?

agent/agent_test.go
```diff
@@ -707,22 +744,67 @@ func (s *AgentSuite) TestEndTaskResponse() {
...
+		TaskId:     s.tc.task.ID,
+		TaskSecret: s.tc.task.Secret,
+		TaskGroup:  s.tc.taskGroup,
+	}
+	shouldSetupGroup := !s.tc.ranSetupGroup
+	taskDirectory := s.tc.taskDirectory
+	_, _, err := s.a.runTask(s.ctx, s.tc, nextTask, shouldSetupGroup, taskDirectory)
 	s.NoError(err)
 	s.Equal(evergreen.TaskSucceeded, s.mockCommunicator.EndTaskResult.Detail.Status)
+	// oom tracker is nil
```
Kimchelly, line 781: Remove comment to self, or if it's intended to stay, make this comment a little more meaningful.

agent/agent_test.go
```diff
@@ -707,22 +744,67 @@ func (s *AgentSuite) TestEndTaskResponse() {
...
+	s.NoError(s.tc.logger.Close())
+	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
+		"Current command set to initial task setup (system).",
+		"Making new folder",
+		"Task logger initialized",
+		"Execution logger initialized.",
+		"System logger initialized.",
+		"Starting task 'task_id', execution 0.",
+	}, []string{panicLog})
+
```
Kimchelly, line 805: nit: remove random newline.

agent/agent_test.go
```diff
@@ -707,22 +744,67 @@ func (s *AgentSuite) TestEndTaskResponse() {
...
+}
```
Kimchelly, line 806: Do we have to test the proper handling of the error conditions in `setupTask` (such as getting bad data about a task, or the context erroring while it's setting up)? Might need to tweak the mock communicator to return fake errors.

agent/agent_test.go
```diff
@@ -786,10 +865,9 @@ func (s *AgentSuite) TestPrepareNextTask() {
 	tc.taskGroup = "bar"
 	tc.taskDirectory = "task_directory"
 	tc.taskModel = &task.Task{}
-	tc = s.a.prepareNextTask(s.ctx, nextTask, tc)
-	s.False(tc.ranSetupGroup, "if the next task in the same version but a different build, ranSetupGroup should be false")
-	s.Equal("bar", tc.taskGroup)
-	s.Empty(tc.taskDirectory)
+	shouldSetupGroup, taskDirectory = s.a.finishPrevTask(s.ctx, nextTask, tc)
+	s.True(shouldSetupGroup, "if the next task in the same version but a different build, shouldSetupGroup should be true")
```
Kimchelly, line 869: "in the same version and task group name"?

Kimchelly (changes requested):
Did you test this in staging and verify that the logs at the beginning of the task look appropriate?

agent/agent.go
malikchaya2: Returned true. But shouldExit should be whatever handleTaskResponse returns.

agent/agent.go
malikchaya2: a.handleTaskResponse overwrites err, so we can't rely on err being non-nil

agent/agent.go
malikchaya2: Is there enough that we can move over that would make it worthwhile? It seems like it would just be the stats collector.

agent/agent_test.go
malikchaya2: I don't know how crucial that is because it's not doing too complex error handling and it's mostly moved from other functions that didn't test it either. The mock communicator is not set up for this to be tested easily. If you feel strongly about it I can try to add it.

agent/agent.go
Kimchelly: I still have this question.

agent/agent.go
```diff
@@ -518,6 +500,15 @@ func (a *Agent) setupTask(agentCtx, setupCtx context.Context, tcInput *taskConte
 	return tc, setupFailed, shouldExit, nil
 }
 
+func (a *Agent) handleSetupError(err error, tc *taskContext, ctx context.Context, shouldSetLogger bool) (*taskContext, bool, bool, error) {
```
Kimchelly, line 503: `setupCtx` should be the first parameter for Go conventions.

I would also order it as `ctx, tc, err, shouldUseFallbackLogger` since we generally pass in `tc` to a lot of functions.

Kimchelly, line 503: nit: I would call `shouldSetLogger` something more meaningful like `shouldUseFallbackLogger` (i.e. the logger when we can't set up the logger). Alternatively, initializing and passing in the logger in the error cases where the logger is not set up might be cleaner than a bool.

agent/agent.go
Kimchelly: Hmm, I guess it's not that much setup logic now that stuff has moved out. Never mind!

agent/agent_test.go
Kimchelly: Eh, we'd ideally test it since we want to know if end in tasks system failure as expected to indicate that there's a bug (rather than the agent doing something undesirable like run the task forever, silently succeed, panic the whole agent, or something else).

If it's way too difficult though, it's alright to skip.

agent/agent.go
Kimchelly, line 503: I would name the output parameters since they're not obvious.

agent/agent.go
Kimchelly: It seems like `setupTask` should return an error though (even if `handleTaskResponse` doesn't produce an error). We can simplify and remove the `setupFailed` boolean by using an error catcher inside `handleSetupError`, catching the error passed into `handleSetupError`, and then catching the error from `handleTaskResponse`, so that returning an error from `setupTask` is equivalent to `setupFailed`.

agent/agent.go
```diff
@@ -501,11 +484,10 @@ func (a *Agent) setupTask(agentCtx, setupCtx context.Context, tcInput *taskConte
 	tc.logger.System().Info("System logger initialized.")
 
 	if err := setupCtx.Err(); err != nil {
-		setupFailed = true
 		grip.Error(err)
 		tc.logger.Execution().Infof("Stopping task execution: %s", err)
 		shouldExit, err := a.handleTaskResponse(setupCtx, tc, evergreen.TaskSystemFailed, err.Error())
-		return tc, setupFailed, shouldExit, err
```
Kimchelly: This can also call `a.handleSetupError`.

agent/agent.go
```diff
@@ -462,35 +462,18 @@ func (a *Agent) setupTask(agentCtx, setupCtx context.Context, tcInput *taskConte
...
 	}
 
 	if !tc.ranSetupGroup {
 		tc.taskDirectory, err = a.createTaskDirectory(tc)
 		if err != nil {
-			setupFailed = true
-			err = errors.Wrap(err, "creating task directory")
-			grip.Error(err)
-			grip.Infof("Task complete: '%s'.", tc.task.ID)
 			tc.logger.Execution().Error(errors.Wrap(err, "creating task directory"))
```
Kimchelly, line 475: I don't think we actually need this log because the error is already passed to `handleSetupError`, which could just log the error to the `tc.logger` in there (this would work for the earlier cases as well, where the logger is not set up and we use the fallback one).

Kimchelly (changes requested):

Kimchelly:
I still also have the staging question.

agent/agent.go
malikchaya2: I removed it, it's not there anymore

agent/agent_test.go
malikchaya2: yeah I think it would be difficult. We can maybe create a ticket to do it separately.

agent/agent.go
malikchaya2: but then we'd start logging errors that we weren't logging before like "making task config" and "setting up logger producer"

agent/agent.go
malikchaya2: The complexity it would add to not 
`grip.Infof("Task complete: '%s'.", tc.task.ID)`
and instead `tc.logger.Execution().Infof("Stopping task execution: %s", err)` made it feel not worth it.

agent/agent.go
malikchaya2: it's now only one bool that's named when it's returned

malikchaya2: I moved `tc.logger = client.NewSingleChannelLogHarness("agent.error", a.defaultLogger)` out so that bool is not needed.

agent/agent.go
```diff
@@ -482,6 +482,14 @@ func (a *Agent) setupTask(agentCtx, setupCtx context.Context, initialTC *taskCon
 	tc.taskConfig.WorkDir = tc.taskDirectory
 	tc.taskConfig.Expansions.Put("workdir", tc.taskConfig.WorkDir)
 
+	factory, ok = command.GetCommandFactory("setup.initial")
```
malikchaya2, line 485: I tested it on staging. I had to add this back in because the otherwise it didn't log like it does now. It's because the first one is before startLogging. I can probably remove the first one but I'm too scared it will have some hidden unintended consequence. 

The rest of the logs are all there. The order differs each time tasks are run so it's hard to compare the order to main.

malikchaya2: the logs on main: https://parsley.mongodb.com/evergreen/evergreen_build_simple_test_patch_1072528ebb0ab31bda6733e16a6ff8bbd19651e7_64df9cada4cf4733f348ed92_23_08_18_16_30_45/1/all?bookmarks=0,1,2,3,4,5,6,7,417

the logs with the changes: 
https://parsley-staging.corp.mongodb.com/evergreen/evg_build_simple_test_patch_1072528ebb0ab31bda6733e16a6ff8bbd19651e7_64dfaa4f9dbe323bd245bf8c_23_08_18_17_28_59/0/all?bookmarks=0,1,2,3,4,5,6,7,213

malikchaya2: while the order is different between main and with the changes, they are also not in the same order with every run. 
for example, here is another run with the changes, and the order is different than the other staging link above: 

https://parsley-staging.corp.mongodb.com/evergreen/evg_build_simple_test_patch_1072528ebb0ab31bda6733e16a6ff8bbd19651e7_64dfafc19dbe323bd245bf96_23_08_18_17_52_12/0/all?bookmarks=0,210

Kimchelly: SGTM. Don't really know why the logs would be out of order, but I have noticed before that logs that come in at the same time with the same exact timestamp oftentimes are mixed up.

agent/agent.go
```diff
@@ -431,26 +416,110 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	return tc, shouldExit, nil
 }
 
+func (a *Agent) handleSetupError(ctx context.Context, tc *taskContext, err error) (*taskContext, bool, error) {
+	catcher := grip.NewBasicCatcher()
+	grip.Error(err)
+	catcher.Add(err)
+	grip.Infof("Task complete: '%s'.", tc.task.ID)
+	shouldExit, err := a.handleTaskResponse(ctx, tc, evergreen.TaskSystemFailed, err.Error())
+	catcher.Add(err)
```
Kimchelly, line 520: I would `catcher.Wrap(err, "handling task response")` (same with the error caught above) so that we can distinguish between this error and the error caught above this. Otherwise, it'll be weird because the resolved catcher errors, when logged, will look like one error happened (the setup error), then it caused a completely random set of errors happened (but it didn't because it's a separate error).

agent/agent_test.go
Kimchelly: I'll leave it to you to make the ticket.

agent/agent.go
```diff
@@ -431,26 +416,110 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+		}
+	} else {
+		tc = initialTC
+	}
+
+	// If the heartbeat aborts the task immediately, we should report that
+	// the task failed during initial task setup.
+	factory, ok := command.GetCommandFactory("setup.initial")
+	if !ok {
+		return tc, shouldExit, errors.New("setup.initial command is not registered")
```
Kimchelly, line 458: Can we actually `grip.Alert` in this case, and potentially continue on error? This error case should never happen unless there's a programmatic bug in the agent command registration setup. It's also the case that `setup.initial` is just used to populate the "Failing command" during setup, where it's not actually running any actual agent commands.

agent/agent.go
```diff
@@ -431,26 +416,110 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+	}
+	tc.setCurrentCommand(factory())
+	a.comm.UpdateLastMessageTime()
+
+	var taskConfig *internal.TaskConfig
```
Kimchelly, line 463: nit: there's no need to declare this variable because you already define it in L464.

agent/agent.go
```diff
@@ -431,26 +416,110 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+			return a.handleSetupError(setupCtx, tc, errors.Wrap(err, "creating task directory"))
+		}
+	}
+	tc.taskConfig.WorkDir = tc.taskDirectory
+	tc.taskConfig.Expansions.Put("workdir", tc.taskConfig.WorkDir)
+
+	factory, ok = command.GetCommandFactory("setup.initial")
+	if !ok {
+		tc.logger.Execution().Error("Marking task as system-failed because setup.initial command is not registered.")
+		return tc, shouldExit, errors.New("setup.initial command is not registered")
```
Kimchelly, line 488: Like above, I'm pretty sure we can continue on error if `setup.initial` is not registered.

agent/agent_test.go
```diff
@@ -707,22 +734,61 @@ func (s *AgentSuite) TestEndTaskResponse() {
 }
 
 func (s *AgentSuite) TestOOMTracker() {
-	s.setupRunTask(defaultProjYml)
-	s.tc.project.OomTracker = true
+	projYml := `
+oom_tracker: true
+buildvariants:
+- name: mock_build_variant
```
Kimchelly, line 740: mega nit: since you indent with one space in the other blocks, should also indent here (although YAML generally has 2 or 4 spaces for indentation, not 1).

Kimchelly (changes requested):

agent/agent_test.go
malikchaya2: https://jira.mongodb.org/browse/EVG-20708

agent/agent.go
```diff
@@ -431,26 +416,110 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
...
+		tc.taskDirectory, err = a.createTaskDirectory(tc)
+		if err != nil {
+			tc.logger.Execution().Error(errors.Wrap(err, "creating task directory"))
+			return a.handleSetupError(setupCtx, tc, errors.Wrap(err, "creating task directory"))
+		}
+	}
+	tc.taskConfig.WorkDir = tc.taskDirectory
+	tc.taskConfig.Expansions.Put("workdir", tc.taskConfig.WorkDir)
+
+	factory, ok = command.GetCommandFactory("setup.initial")
```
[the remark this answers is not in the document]
malikchaya2: :see_no_evil:  true

agent/agent.go
Kimchelly: Seems like this can use `handleSetupError` now that it logs both.

agent/agent.go
Kimchelly: If this is going to alert, we have to do `else { tc.setCurrentCommand(factory()) }` since the factory will be nil when `!ok`.

agent/agent.go
Kimchelly: Same thing as above about checking `if factory != nil` before dereferencing.

Kimchelly (approved):
I left a few small comments, but you can easily resolve them without another review.

PR #6894, opened 2023-08-10
EVG-20281: improve callback timeout behavior

Kimchelly:
EVG-20281

### Description
* Move callback timeout into a helper function that waits for the timeout. If the timeout is hit, it will log that it was hit to the task log (so users can see) and stop the command that was running in the block. Furthermore, if timing out the command can fail the task, it marks the task as having hit a timeout failure.
* Use the timeout helper for exec timeout handler. It should be logically the same as before, just consolidating duplicate logic.
* Add more timeout types. If a task times out (i.e. setup group or callback timeout), the new timeout reasons will be shown in the UI. These will be documented more thoroughly in EVG-20629, when callback timeout is reworked.
* Drive-by remove a random dead struct.

### Testing
Added unit tests, and existing timeout tests pass.

### Documentation
Updated documentation on special behavior of pre, post, and timeout to be a little clearer. I didn't expand on the explanation of callback timeout that deeply beyond what it already says, because we're going to rework callback timeout in EVG-20629 anyways.


docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -523,28 +516,37 @@ tasks:
...
 ```
 
+#### Special Behavior for Pre, Post, and Timeout
+By default, a command failing during the `pre` block will not cause the entire
+task to fail. If you want to enforce that failures during `pre` cause the task
+to fail, set the field `pre_error_fails_task` to true. If `pre_error_fails_task`
+is set to true and a command in `pre` fails, it will exit early and not run the
+main task commands, but it will still run the `post` block.
+`pre_error_fails_task` has no effect on tasks run in task groups because they do
+not run `pre`; instead, those tasks follow the settings defined for that task
```
malikchaya2, line 533: nit: I think the "because they do not run `pre`" part is a bit confusing

malikchaya2 (approved):
Approved with some small non-blocking nits about documentation.

docs/Project-Configuration/Project-Configuration-Files.md
Kimchelly: The documentation is true, but I clarified that `pre` is ignored because task groups run `setup_group` and `setup_task` instead.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -523,28 +516,37 @@ tasks:
...
+cause the entire task to fail. If you want to enforce that failures during
+`post` cause the task to fail, set the field `post_error_fails_task` to true. If
+`post_error_fails_task` is set to true and both the main task and post block
+have failing commands, the task's failing command will be the command that
+failed in the main task block, not the failing post command.
+`post_error_fails_task` has no effect on tasks run in task groups because they
+do not run `post`; instead, those tasks, follow the settings defined for that
+task group.
+
+By default, commands in `post` or `timeout` will time out after 15 minutes. You
```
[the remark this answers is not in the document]
Kimchelly: This is not a silent failure, because the task logs explicitly state that the command fails and on top of that, now shows "Hit <type of timeout> timeout" before the command is failed.

PR #6897, opened 2023-08-11
EVG-20652 TestAgentEndTaskShouldExit should pass because shouldExit is true 

malikchaya2:
EVG-20652

### Description
This test was passing, but it also hit an unrelated error of the build variant missing from the project. This adds a check to make sure that it's not exiting for other reasons so we know that it's exiting because shouldExit is true. 

### Testing
The test itself

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -33,7 +33,7 @@ func init() {
 
 const defaultProjYml = `
 buildvariants:
-- name: some_build_variant
+- name: mock_build_variant
 tasks: 
  - name: this_is_a_task_name
    commands: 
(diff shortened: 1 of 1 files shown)


agent/agent_test.go
```diff
@@ -224,6 +225,10 @@ func (s *AgentSuite) TestAgentEndTaskShouldExit() {
 	case <-ctx.Done():
 		s.FailNow(ctx.Err().Error())
 	}
+
+	endDetail := s.mockCommunicator.EndTaskResult.Detail
+	s.Equal("", endDetail.Message, "the end message should not include any errors")
```
Kimchelly, line 230: `s.Empty`

Kimchelly (approved):
LGTM mod nit.

PR #6900, opened 2023-08-14
EVG-19646 Include container tasks in checked blocked tasks job

hadjri:
EVG-19646

### Description

- Modified the check blocked tasks job to check for container tasks alongside its checks for distro tasks
- Refactored an existing container task query to allow us to reuse the same container task dependency query in multiple places
- Removed some unused logic to count `numTasksModified` in `checkUnmarkedBlockingTasks`

### Testing
Since the check blocked tasks job only intends to clean up dependencies of parent tasks that failed before they got to [UpdateBlockedDependencies](https://gist.github.com/hadjri/7dd0e4105c3ad9cefc8e915493e1f39e#file-task_lifecycle-go-L611), I temporarily commented that line out in staging and ran tasks on a container variant and a distro variant.

[Test patch](https://spruce-staging.corp.mongodb.com/version/64fa38519dbe3283b59df38c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

The dependency chain goes `task1 -> task2 -> task3`, and task1 is configured to fail. I confirmed that for both the distro and container variant the child tasks got blocked by the cron job on the 15 minute mark as expected.


ablack12:
@hadjri this has been up for a long time; are you still pursuing this change?

units/check_blocked_tasks.go
```diff
@@ -84,29 +96,29 @@ func (j *checkBlockedTasksJob) Run(ctx context.Context) {
...
-		numModified, err := checkUnmarkedBlockingTasks(&t, dependencyCache)
-		j.AddError(err)
-		numTasksModified += numModified
-		if numTasksModified > 0 {
-			numChecksThatUpdatedTasks++
-		}
+func (j *checkBlockedTasksJob) getTasksToCheckContainer() []task.Task {
+	query := task.IsUndispatchedContainerTasksQuery()
+	query[task.ContainerAllocatedKey] = false
+	tasksToCheck, err := task.FindAll(db.Query(query).Sort([]string{task.ActivatedTimeKey}))
```
Kimchelly, line 113: Does this need to sort on activated time? The sort is needed for the container task queue because that's the only queue-ordering mechanism. But I imagine their order in the queue doesn't matter with regards to dependency checking, unless I'm missing something.

model/task/db.go
```diff
@@ -788,25 +788,36 @@ func needsContainerAllocation() bson.M {
 //     have been met.
 //  2. The container is allocated but the agent has not picked up the task yet.
 func IsContainerTaskScheduledQuery() bson.M {
+	query := IsUndispatchedContainerTasksQuery()
+	query["$or"] = dependenciesAreMetQuery()
+	return query
+}
+
+// IsUndispatchedContainerTasksQuery returns a query returning all tasks that
+// are undispatched container tasks.
```
Kimchelly, line 797: nit: the phrasing is a little awkward, maybe something like "returns all undispatched container tasks"?

units/check_blocked_tasks.go
```diff
@@ -54,6 +53,19 @@ func NewCheckBlockedTasksJob(distroId string, ts time.Time) amboy.Job {
...
 func (j *checkBlockedTasksJob) Run(ctx context.Context) {
+	var tasksToCheck []task.Task
+	if j.DistroId != "" {
+		tasksToCheck = j.getTasksToCheckDistro()
+	} else {
+		tasksToCheck = j.getTasksToCheckContainer()
+	}
+	dependencyCache := map[string]task.Task{}
+	for _, t := range tasksToCheck {
+		j.AddError(checkUnmarkedBlockingTasks(&t, dependencyCache))
```
Kimchelly, line 64: nit: can we wrap this error with the task ID so we know which particular one errored while updating dependencies?

model/task/db.go
```diff
@@ -788,25 +788,36 @@ func needsContainerAllocation() bson.M {
...
-				// running and are still attainable (i.e. the dependency's
-				// required status matched the task's actual ending status).
-				bsonutil.GetDottedKeyName(DependsOnKey, DependencyFinishedKey):     true,
-				bsonutil.GetDottedKeyName(DependsOnKey, DependencyUnattainableKey): bson.M{"$ne": true},
-			},
-			{OverrideDependenciesKey: true},
+	}
+}
+
+func dependenciesAreMetQuery() []bson.M {
```
ablack12, line 807: Since this is only used in IsContainerTaskScheduledQuery, I think there's not a benefit to factoring it out; easier to read it all together

model/task/db.go
```diff
@@ -788,25 +788,36 @@ func needsContainerAllocation() bson.M {
 //     have been met.
 //  2. The container is allocated but the agent has not picked up the task yet.
 func IsContainerTaskScheduledQuery() bson.M {
+	query := IsUndispatchedContainerTasksQuery()
+	query["$or"] = dependenciesAreMetQuery()
+	return query
+}
+
+// IsUndispatchedContainerTasksQuery returns a query returning all tasks that
```
ablack12, line 796: comment nit: "returning all undispatched container tasks" has less redundancy 
function title nit ( that also applies to the existing query above so my bad): functions starting with "Is" usually imply that we're returning a boolean. Really it's the ScheduledContainerTasksQuery and the UndispatchedContainerTasksQuery.

units/check_blocked_tasks.go
```diff
@@ -117,10 +129,10 @@ func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.T
 			"activated_by": t.ActivatedBy,
 			"depends_on":   t.DependsOn,
 		})
-		return 0, errors.Wrapf(err, "checking if dependencies met for task '%s'", t.Id)
+		return errors.Wrapf(err, "checking if dependencies met for task '%s'", t.Id)
 	}
 	if dependenciesMet {
-		return 0, nil
+		return nil
```
Kimchelly, line 135: GitHub doesn't let me comment on unchanged lines but for L144, if this is updating the blocked dependency state for finished tasks to block them in case `UpdateBlockedDependencies` fails to run, should we also call `MarkDependenciesFinished` to indicate that the blocking task is both blocking its dependency children and also finished running?

I'm also fine if we just want to make a separate ticket to investigate this later.

units/check_blocked_tasks.go
```diff
@@ -84,29 +96,29 @@ func (j *checkBlockedTasksJob) Run(ctx context.Context) {
...
 	}
+	return tasksToCheck
 }
 
-func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.Task) (int, error) {
+func checkUnmarkedBlockingTasks(t *task.Task, dependencyCaches map[string]task.Task) error {
 	catcher := grip.NewBasicCatcher()
 
 	dependenciesMet, err := t.DependenciesMet(dependencyCaches)
```
Kimchelly, line 124: Is it necessary to double-check `DependenciesMet` here? I imagine the query itself for tasks could filter out tasks whose dependencie are already met.

I'm also fine if we just want to make a separate ticket to investigate this later.

Kimchelly (changes requested):

units/check_blocked_tasks.go
```diff
@@ -54,6 +53,19 @@ func NewCheckBlockedTasksJob(distroId string, ts time.Time) amboy.Job {
...
+	} else {
+		tasksToCheck = j.getTasksToCheckContainer()
+	}
+	dependencyCache := map[string]task.Task{}
+	for _, t := range tasksToCheck {
+		j.AddError(checkUnmarkedBlockingTasks(&t, dependencyCache))
+	}
+}
+
+func (j *checkBlockedTasksJob) getTasksToCheckDistro() []task.Task {
```
ablack12, line 68: nit: getDistroTasksToCheck is more of a function name (and getContainerTasksToCheck)

ablack12 (changes requested):

model/task/db.go
hadjri: This was more done to allow the `UndispatchedContainerTasksQuery` function to be reused in multiple places (one invocation needs this `dependenciesAreMetQuery ` appended to it, and the other one (part of the check blocked tasks job) needs to exclude it

units/check_blocked_tasks.go
hadjri: Good catch, removed

units/check_blocked_tasks.go
hadjri: Sounds good, I consolidated this comment and the other one into https://jira.mongodb.org/browse/EVG-20855

units/check_blocked_tasks.go
hadjri: Part of https://jira.mongodb.org/browse/EVG-20855

model/task/db.go
```diff
@@ -773,12 +773,12 @@ func FindNeedsContainerAllocation() ([]Task, error) {
 // needsContainerAllocation returns the query that filters for a task that
 // currently needs a container to be allocated to run it.
 func needsContainerAllocation() bson.M {
```
Kimchelly, line 775: It seems like this query is equivalent to the `task.FindAll` query in the check blocked tasks job, so it would be simpler to export it + use it instead of setting `ContainerAllocated = false` manually.

Kimchelly (approved):
LGTM mod one consolidation.

model/task/db.go
hadjri: It is distinct because the check blocked tasks job looks at [UndispatchedContainerTasksQuery](https://github.com/evergreen-ci/evergreen/pull/6900/files#diff-bb5e4c87753aeaf8606e661b6f8a86d27ed96e48981b45490846bec566ca54f8R111) which includes tasks that have yet to have their dependencies met, rather than `ScheduledContainerTasksQuery`

hadjri: Re-requesting review just to make sure this is resolved

Kimchelly (approved):

model/task/db.go
ablack12: Right but since only one needs it, we can just define it there (i.e. in ScheduledContainerTasksQuery) explicitly (make sense to keep undispatchedContainerTasksQuery factored out since it's used in more places)

ablack12 (approved):
LGTM with last thought

PR #6901, opened 2023-08-14
EVG-20463: allow Docker host.create to read file to stdin

Kimchelly:
EVG-20463

### Description
* Add support for reading a file from a host task, and passing the entire content of that file to a container command created via host.create.
    * This only works during in the context of running a task - if you start a spawn host and check the UI option to also start any host.create'd containers, the started container will _not_ read the original stdin file from the host. I considered the case of starting a host.create container from a spawn host niche enough that it wasn't worth supporting.

### Testing
* Added a few unit tests.
* Tested in staging by passing in a file to stdin, running a container with `tee` (which acts as a passthrough to send the data from the container's stdin to stdout), and verifying that the container stdout logs included the file.

### Documentation
* Updated docs for host.create.
* Drive-by fixed an issue where the project commands file is incorrectly escaping the less than and greater than symbols (`<`, `>`) in the markdown.

### DIFF
--- a/cloud/docker.go
+++ b/cloud/docker.go
@@ -1,10 +1,13 @@
 package cloud
 
 import (
+	"bytes"
 	"context"
+	"io"
 	"strings"
 	"time"
 
+	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"github.com/evergreen-ci/evergreen"
 	"github.com/evergreen-ci/evergreen/model/distro"
@@ -12,6 +15,7 @@ import (
 	"github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/recovery"
 	"github.com/pkg/errors"
 )
 
@@ -61,6 +65,10 @@ func (m *dockerManager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host
 		return nil, err
 	}
 
+	if err := m.attachStdinStream(parentHost, h); err != nil {
+		return nil, errors.Wrapf(err, "attaching stdin to container for host '%s'", h.Id)
+	}
+
 	if err = h.SetAgentRevision(ctx, evergreen.AgentVersion); err != nil {
 		return nil, errors.Wrapf(err, "setting agent revision on host '%s' to '%s'", h.Id, evergreen.AgentVersion)
 	}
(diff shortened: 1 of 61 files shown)


cloud/docker.go
```diff
@@ -93,6 +102,76 @@ func (m *dockerManager) SpawnHost(ctx context.Context, h *host.Host) (*host.Host
 	return h, nil
 }
 
+// attachStdinStream attaches to the container's stdin and streams data to it.
+// Note that the stream to stdin has to run in the background because stdin is
+// not open until the container command is up and running.
+func (m *dockerManager) attachStdinStream(parent *host.Host, container *host.Host) error {
```
Kimchelly, line 108: Worth noting that I stole the general stdin attachment logic from myself when I wrote it [in Jasper](https://github.com/mongodb/jasper/blob/6d9ec4246500a4c9c7d4006d29f053c920e9dc18/internal/executor/docker.go#L152-L158).

rest/data/host_create.go
```diff
@@ -309,6 +310,7 @@ func makeDockerIntentHost(ctx context.Context, env evergreen.Environment, taskID
 		RegistryName:     createHost.Registry.Name,
 		RegistryUsername: createHost.Registry.Username,
 		RegistryPassword: createHost.Registry.Password,
+		StdinData:        createHost.StdinFileContents,
```
ablack12, line 313: Could this be potentially a very very large thing we're now storing in the host document? Should it have some sort of limit? And/or should we unset it once the stream is done?

Kimchelly: It could theoretically be large, but:

* We already seemingly allow very large unbounded arrays in lots of our data models (like the list of all batchtime tasks in the version document) that can be many MB in size, and they haven't caused problems (it has to be ~5 bibles in length).
* Realistically the input won't be large, because the one user who wanted this said they wanted to pass in a few secrets through stdin instead of having integration with a shared secret store (like Secrets Manager).
* If the input really is that large, their host.create will just fail, and they'll have to come up with their own workaround anyways.

In any case, this one is just a holdover until Docker host.create is deprecated in favor of container tasks, so no matter what is the use case, this is supposedly temporary.

ablack12: ~5 Bibles LOL alright that works

ablack12 (approved):
I am still in favor of unsetting the document after streaming is done if possible, since I don't think we delete the host document when we terminate it and it's just not necessary to keep it in the DB, but overall LGTM

Kimchelly:
I changed it to clear it from the host doc.

ablack12 (approved):

PR #6904, opened 2023-08-14
EVG-20653 Add a temporary flag that lets projects to test unsetting FunctionVars

malikchaya2:
EVG-20653

### Description
This allows users to specify `unset_function_vars: true` in their project yaml to test how they will be affected by this change. 

### Testing
unit test 

### Documentation
An email will be sent out. This will not be formally documented because it's temporary

### DIFF
--- a/agent/command_test.go
+++ b/agent/command_test.go
@@ -230,7 +230,7 @@ func (s *CommandSuite) setUpConfigAndProject(projYml string) {
 	s.tc.taskConfig.Project = p
 }
 
-func (s *CommandSuite) TestUnsetFunctionVarsProjectFlag() {
+func (s *CommandSuite) TestFunctionVarsUnsetWhenProjectFlagTrue() {
 	s.tc.unsetFunctionVarsDisabled = true
 	projYml := `
 unset_function_vars: true 
@@ -260,9 +260,12 @@ functions:
 	key1Value := s.tc.taskConfig.Expansions.Get("key1")
 	s.Equal("expansionVar", key1Value, "globalVar should be set back to what it was before the function ran")
 
+}
+
+func (s *CommandSuite) TestFunctionVarsDontUnsetWithoutFlag() {
 	s.tc.unsetFunctionVarsDisabled = true
 
-	projYml = `
+	projYml := `
 functions:
   yes:
     vars: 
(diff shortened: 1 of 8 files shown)


agent/command_test.go
```diff
@@ -229,6 +229,58 @@ func (s *CommandSuite) setUpConfigAndProject(projYml string) {
...
+	}
+
+	cmds := []model.PluginCommandConf{func1}
+	err := s.a.runCommandsInBlock(s.ctx, s.tc, cmds, runCommandsOptions{}, "")
+	s.NoError(err)
+
+	key1Value := s.tc.taskConfig.Expansions.Get("key1")
+	s.Equal("expansionVar", key1Value, "globalVar should be set back to what it was before the function ran")
+
+	s.tc.unsetFunctionVarsDisabled = true
```
Kimchelly, line 263: nit: the test above this seems unrelated to the test below this, so I would split it into two test cases.

Generally, keeping your test cases as specific, short, and precise as possible is preferable compared to a longer test since it reduces the possibility of having a bug in the test due to it doing a lot of things (especially where the later tests somehow depend on the state modifications of the previous tests).

model/project_parser_db.go
```diff
@@ -18,6 +18,7 @@ var (
 	ParserProjectEnabledKey           = bsonutil.MustHaveTag(ParserProject{}, "Enabled")
 	ParserProjectStepbackKey          = bsonutil.MustHaveTag(ParserProject{}, "Stepback")
 	ParserProjectPreErrorFailsTaskKey = bsonutil.MustHaveTag(ParserProject{}, "PreErrorFailsTask")
+	ParserProjectUnsetFunctionVarsKey = bsonutil.MustHaveTag(ParserProject{}, "UnsetFunctionVars")
```
Kimchelly, line 21: We don't need this key since we never use it.

Kimchelly (approved):
LGTM mod bump the agent version + non-blocking suggestions.

agent/command_test.go
malikchaya2: It's two sides of the same coin. It tests that it does work when set to true and doesn't work when set to false. Keeping it in one test will also make it easier when removing it, which we will do since it's temporary. This also reuses the function and commands definition.

model/project_parser_db.go
malikchaya2: removed

PR #6910, opened 2023-08-15
EVG-20292, EVG-20630: improve idle timeout behavior and logging

Kimchelly:
EVG-20292
EVG-20630

### Description
Idle timeout appears to have been implemented such that it stays alive for the entire task duration, but it can only actually time out a command that runs in pre or main (for example, if a command hits the idle timeout during post, the idle timeout is logged but nothing happens). I didn't want to change this behavior, so instead, I solidified this as the official behavior of idle timeout.

* Run the idle timeout for pre and main only, and stop it when pre and main are over.
* Don't log the "set idle timeout" in blocks where the idle timeout is not relevant (e.g. `post`, `teardown_task`, `teardown_group`, `timeout`).
* As part of conditionally logging the "set idle timeout" log depending on the block, I decided to refactor some of the block naming so it's an actual enum (rather than untyped string) and removed `isTaskCommands` (since you can infer that it's the main task block from the command block type).
* Slightly improve logging for commands so they go to the task logs instead of execution logs (since command failures are important to see) and so they're a little more consistent in the error cases (especially in post group).

### Testing
* Added a bunch of unit tests. Refactors of existing behavior should be covered by the existing agent test suite.
* Tested in staging patches to ensure the idle timeout is as described.

### Documentation
Improved documentation for idle timeout and exec timeout.


agent/agent.go
```diff
@@ -1092,13 +1099,8 @@ func (a *Agent) runEndTaskSync(ctx context.Context, tc *taskContext, detail *api
 	}
 	defer cancel()
 
-	if err := a.runCommandsInBlock(syncCtx, tc, taskSyncCmds.List(), runCommandsOptions{}, taskSyncBlock); err != nil {
-		tc.logger.Task().Error(message.WrapError(err, message.Fields{
-			"message":    "error running task sync",
-			"total_time": time.Since(start).String(),
```
malikchaya2, line 1098: why do we no longer want to log the total time?

agent/agent_test.go
```diff
@@ -484,7 +487,57 @@ post:
...
 		"Finished command 'shell.exec' (step 1 of 1) in block 'post'",
 		"Finished running post-task commands",
-	}, []string{panicLog})
+	}, []string{
+		panicLog,
+		"Set idle timeout for 'shell.exec'",
+	})
+}
+
+func (s *AgentSuite) TestMainTask() {
```
malikchaya2, line 496: can we maybe either make the task name more descriptive or add a comment explaining what it is that we are asserting with this test?

agent/agent_test.go
```diff
@@ -484,7 +487,57 @@ post:
...
+      script: exit 1
+`
+	s.setupRunTask(projYml)
+
+	s.Error(s.a.runTaskCommands(s.ctx, s.tc))
+	s.NoError(s.tc.logger.Close())
+	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
+		"Running task commands",
+		"Set idle timeout for 'shell.exec'",
+		"Running command 'shell.exec' (step 1 of 1)",
```
malikchaya2, line 536: Does the failure case not log `"Finished command 'shell.exec' (step 1 of 1)",`?

agent/task_context.go
```diff
@@ -30,7 +30,16 @@ func (tc *taskContext) getCurrentCommand() command.Command {
...
 
-func (tc *taskContext) setCurrentIdleTimeout(cmd command.Command) {
+// setCurrentIdleTimeout sets the idle timeout for the current running command.
+// This timeout only applies to commands running in specific blocks where idle
+// timeout is allowed.
+func (tc *taskContext) setCurrentIdleTimeout(cmd command.Command, block command.BlockType) {
+	switch block {
+	case command.PreBlock, command.SetupGroupBlock, command.SetupTaskBlock, command.MainTaskBlock:
+	default:
+		return
```
malikchaya2, line 40: Should this return an error so that it doesn't result in progammer error down the road?

malikchaya2 (changes requested):

agent/agent.go
Kimchelly: It already logs on L1102 because the return is removed.

agent/agent_test.go
Kimchelly: Done, also fixed the other ones already in this file with the same issue.

agent/agent_test.go
Kimchelly: It doesn't. EVG-20634 is supposed to fully fix this.

agent/task_context.go
Kimchelly: `setCurrentIdleTimeout` is called even for commands blocks where idle timeout doesn't apply. I changed it to move this block check to the outer function.

agent/command.go
```diff
@@ -179,7 +179,14 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
...
 	tc.setCurrentCommand(cmd)
-	tc.setCurrentIdleTimeout(cmd, options.block)
+	switch options.block {
+	case command.PreBlock, command.SetupGroupBlock, command.SetupTaskBlock, command.MainTaskBlock:
+		// Only set the idle timeout in cases where the idle timeout is actually
+		// respected. In all other blocks, setting the idle timeout should have
+		// no effect.
+		tc.setCurrentIdleTimeout(cmd, options.block)
+	default:
+	}
```
malikchaya2, line 189: since we don't have a default case, can this be an `if` instead?

malikchaya2 (changes requested):

malikchaya2 (approved):

Kimchelly:
Re-requesting review since I had to resolve merge conflicts for two different commits into this PR.

PR #6911, opened 2023-08-15
EVG-20673 Allow expansions.update to set expansions to the same value as function vars

malikchaya2:
EVG-20673

### Description
Function vars wouldn't unset any vars whose values were changed by expansions.update. However, if it was set to the same value as the function var, it didn't work. We don't want that to be the case because we want users to be able to do this to persist function vars. 
image
 

### Testing
Unit test and tested on staging.

### DIFF
--- a/agent/command.go
+++ b/agent/command.go
@@ -167,21 +167,17 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 		if !tc.unsetFunctionVarsDisabled || tc.project.UnsetFunctionVars {
 			// This defer ensures that the function vars do not persist in the expansions after the function is over
 			// unless they were updated using expansions.update
-
-			if cmd.Name() == "expansions.update" {
-				c, ok := cmd.(*command.Update)
-				if !ok {
-					tc.logger.Task().Errorf("unable to unset function vars, unexpected error getting commands.")
-					return
-				}
-				for _, v := range c.Updates {
-					if _, ok := commandInfo.Vars[v.Key]; ok {
+			if cmd.Name() == "expansions.update" && tc.taskConfig.DynamicExpansions != nil {
+				updatedExpansions := tc.taskConfig.DynamicExpansions.Map()
+				for k := range updatedExpansions {
+					if _, ok := commandInfo.Vars[k]; ok {
 						// If expansions.update updated this key, don't reset it
-						delete(prevExp, v.Key)
+						delete(prevExp, k)
 					}
 				}
 			}
 			tc.taskConfig.Expansions.Update(prevExp)
+			tc.taskConfig.DynamicExpansions.Update(map[string]string{})
 		}
 	}()
 
--- a/agent/command.go
+++ b/agent/command.go
@@ -165,12 +165,20 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 	}
 	defer func() {
 		if !tc.unsetFunctionVarsDisabled || tc.project.UnsetFunctionVars {
-			// This defer ensures that the function vars do not persist in the expansions after the function is over.
-			for key, functionValue := range commandInfo.Vars {
-				currentValue := tc.taskConfig.Expansions.Get(key)
-				if currentValue != functionValue {
-					// If a command in the func updates the expansion value, don't reset it.
-					delete(prevExp, key)
+			// This defer ensures that the function vars do not persist in the expansions after the function is over
+			// unless they were updated using expansions.update
+
+			if cmd.Name() == "expansions.update" {
+				c, ok := cmd.(*command.Update)
+				if !ok {
+					tc.logger.Task().Errorf("unable to unset function vars, unexpected error getting commands.")
+					return
+				}
+				for _, v := range c.Updates {
+					if _, ok := commandInfo.Vars[v.Key]; ok {
+						// If expansions.update updated this key, don't reset it
+						delete(prevExp, v.Key)
+					}
 				}
 			}
 			tc.taskConfig.Expansions.Update(prevExp)
--- a/agent/command.go
+++ b/agent/command.go
@@ -10,6 +10,7 @@ import (
 	"github.com/evergreen-ci/evergreen/agent/internal/client"
 	"github.com/evergreen-ci/evergreen/apimodels"
 	"github.com/evergreen-ci/evergreen/model"
+	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/utility"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/recovery"
@@ -167,7 +168,7 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 		if !tc.unsetFunctionVarsDisabled || tc.project.UnsetFunctionVars {
 			// This defer ensures that the function vars do not persist in the expansions after the function is over
 			// unless they were updated using expansions.update
-			if cmd.Name() == "expansions.update" && tc.taskConfig.DynamicExpansions != nil {
+			if cmd.Name() == "expansions.update" {
 				updatedExpansions := tc.taskConfig.DynamicExpansions.Map()
 				for k := range updatedExpansions {
 					if _, ok := commandInfo.Vars[k]; ok {
@@ -177,7 +178,8 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 				}
 			}
 			tc.taskConfig.Expansions.Update(prevExp)
-			tc.taskConfig.DynamicExpansions.Update(map[string]string{})
+			tc.taskConfig.DynamicExpansions = util.EmptyExpansion()
+
 		}
 	}()
 
--- a/agent/command/expansion_test.go
+++ b/agent/command/expansion_test.go
@@ -38,7 +38,8 @@ func TestExpansionsPlugin(t *testing.T) {
 		expansions.Put("topping", "bacon")
 
 		taskConfig := internal.TaskConfig{
-			Expansions: &expansions,
+			Expansions:        &expansions,
+			DynamicExpansions: &util.Expansions{},
 		}
 
 		So(updateCommand.ExecuteUpdates(ctx, &taskConfig), ShouldBeNil)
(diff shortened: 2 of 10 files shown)


agent/command.go
```diff
@@ -165,12 +165,20 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
...
-			for key, functionValue := range commandInfo.Vars {
-				currentValue := tc.taskConfig.Expansions.Get(key)
-				if currentValue != functionValue {
-					// If a command in the func updates the expansion value, don't reset it.
-					delete(prevExp, key)
+			// This defer ensures that the function vars do not persist in the expansions after the function is over
+			// unless they were updated using expansions.update
+
+			if cmd.Name() == "expansions.update" {
+				c, ok := cmd.(*command.Update)
```
Kimchelly, line 172: I don't think this will work if the expansions are specified via a YAML file rather than explicitly listing them in `updates`. This is also a little janky because we're closely depending on the exact way that `expansions.update` is implemented and trying to reproduce its behavior here again.

Instead of doing it like this, can we consider an alternative implementation? One idea is adding a `TaskConfig.DynamicExpansions` field and modifying `expansions.update` to duplicate its writes there (and then unsetting the dynamic expansions when `runCommand` returns). Then, so long as `expansions.update` replicates its updates to the dynamic expansions, it gives a way for this defer to read the updated expansions, without relying on how exactly `expansions.update` is implemented or awkwardly type-casting.

agent/command_test.go
```diff
@@ -328,20 +328,29 @@ functions:
...
 
 	key1Value := s.tc.taskConfig.Expansions.Get("key1")
-	s.Equal("expansion-update-value", key1Value, "key1 should be set to what it was updated to with expansions.update")
+	s.Equal("functionVar1", key1Value, "key1 should be set to what it was updated to with expansions.update")
+
+	key2Value := s.tc.taskConfig.Expansions.Get("key2")
+	s.Equal("functionVar2", key2Value, "key2 should be set to what it was updated to with expansions.update")
+
+	key3Value := s.tc.taskConfig.Expansions.Get("key3")
+	s.Equal("expansionVar3", key3Value, "key3 should be the original expansion value")
```
Kimchelly, line 354: I would also add a test for expansions.update from YAML files since it missed that case.

agent/command_test.go
```diff
@@ -328,20 +328,29 @@ functions:
...
+
```
Kimchelly, line 355: nit: remove random whitespace.

Kimchelly (changes requested):

agent/command/expansion_update.go
```diff
@@ -73,6 +73,7 @@ func (c *update) ExecuteUpdates(ctx context.Context, conf *internal.TaskConfig)
 				return errors.WithStack(err)
 			}
 			conf.Expansions.Put(update.Key, newValue)
+			conf.DynamicExpansions.Put(update.Key, newValue)
```
Kimchelly, line 76: Since `expansions.update` is the only usage of `DynamicExpansions`, can we have the command initialize `DynamicExpansions` in the case that it's nil? This also protects the agent from potential panics, where it assumes the `DynamicExpansions` exist.

Alternatively, explicitly testing a code path that runs `NewTaskConfig` and asserting that it must return a non-nil `DynamicExpansions` after initialization may be a reasonable substitute.

agent/command_test.go
```diff
@@ -327,21 +328,56 @@ functions:
...
-	s.Equal("expansion-update-value", key1Value, "key1 should be set to what it was updated to with expansions.update")
+	key1Value := s.tc.taskConfig.Expansions.Get("key_1")
+	s.Equal("functionVar1", key1Value, "key_1 should be set to what it was updated to with expansions.update")
+
+	key2Value := s.tc.taskConfig.Expansions.Get("key_2")
+	s.Equal("functionVar2", key2Value, "key_2 should be set to what it was updated to with expansions.update")
+
+	key3Value := s.tc.taskConfig.Expansions.Get("key_3")
+	s.Equal("expansionVar3", key3Value, "key_3 should be the original expansion value")
+}
```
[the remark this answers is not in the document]
Kimchelly: Can we also check that the dynamic expansions are reset to empty after the expansions.update command ends? Would avoid a potential bug of a kind similar to the original func vars bug if a dev changes the code in the future and forgets to reset the dynamic expansions (thus letting them carry over to later `expansions.update`s).

agent/command/expansion_update.go
```diff
@@ -118,6 +120,10 @@ func (c *update) Execute(ctx context.Context,
 		if err != nil {
 			return errors.WithStack(err)
 		}
+		err = conf.DynamicExpansions.UpdateFromYaml(filename)
+		if err != nil {
```
Kimchelly, line 124: nit: you can just combine L123-124 (same with L119-120).

agent/internal/task_config.go
```diff
@@ -28,6 +28,7 @@ type TaskConfig struct {
 	Task               *task.Task
 	BuildVariant       *model.BuildVariant
 	Expansions         *util.Expansions
+	DynamicExpansions  *util.Expansions
```
Kimchelly, line 31: If you want this to be extra safe, having this be a `util.Expansions` without the pointer is better since you'll only have to deal with one level of nilness rather than two - for example, a nil `*util.Expansion` vs. a non-pointer `util.Expansions` that happens to be default initialize (i.e. a nil map).

agent/command/expansion_test.go
```diff
@@ -38,7 +38,8 @@ func TestExpansionsPlugin(t *testing.T) {
 		expansions.Put("topping", "bacon")
 
 		taskConfig := internal.TaskConfig{
-			Expansions: &expansions,
+			Expansions:        &expansions,
+			DynamicExpansions: &util.Expansions{},
```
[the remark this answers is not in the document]
malikchaya2: This is only used for unsetting vars and that is already tested extensively. Do you specifically want a test here?

agent/internal/task_config.go
malikchaya2: Doesn't setting it in NewTaskConfig take care of this concern?

agent/command/expansion_update.go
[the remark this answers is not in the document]
malikchaya2: I set it in NewTaskConfig for prod, and here I set it when setting up the test.

agent/command.go
```diff
@@ -165,15 +165,19 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
...
 		if !tc.unsetFunctionVarsDisabled || tc.project.UnsetFunctionVars {
-			// This defer ensures that the function vars do not persist in the expansions after the function is over.
-			for key, functionValue := range commandInfo.Vars {
-				currentValue := tc.taskConfig.Expansions.Get(key)
-				if currentValue != functionValue {
-					// If a command in the func updates the expansion value, don't reset it.
-					delete(prevExp, key)
+			// This defer ensures that the function vars do not persist in the expansions after the function is over
+			// unless they were updated using expansions.update
+			if cmd.Name() == "expansions.update" && tc.taskConfig.DynamicExpansions != nil {
```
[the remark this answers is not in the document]
malikchaya2: true, I'll add a length check. But I'd also rather keep both checks because there are no guarantees that DynamicExpansions won't be used in some other way in the future.

agent/command/expansion_update.go
malikchaya2: added to setup test

agent/command/expansion_test.go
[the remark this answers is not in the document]
malikchaya2: I added a check here. But the test is not updated to test the file case. I don't think it's crucial to add considering the above.

agent/command.go
```diff
@@ -165,15 +166,20 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
...
+				updatedExpansions := tc.taskConfig.DynamicExpansions.Map()
+				for k := range updatedExpansions {
+					if _, ok := commandInfo.Vars[k]; ok {
+						// If expansions.update updated this key, don't reset it
+						delete(prevExp, k)
+					}
 				}
 			}
 			tc.taskConfig.Expansions.Update(prevExp)
+			tc.taskConfig.DynamicExpansions = util.EmptyExpansion()
```
Kimchelly, line 181: I'm pretty sure you can just call `util.NewExpansions(map[string]string{})` rather than make a brand new function for a specific case of it.

agent/command/expansion_update.go
Kimchelly: I still have this question, I meant adding something like this to the command's `Execute`, not just changing the test setup:
```
func Execute(...) {
// ...
if conf.DynamicExpansions == nil {
    // Initialize the map
}
// Actually do the expansions.update logic
}
```

agent/command_test.go
```diff
@@ -328,20 +330,61 @@ functions:
...
 	key1Value := s.tc.taskConfig.Expansions.Get("key1")
-	s.Equal("expansion-update-value", key1Value, "key1 should be set to what it was updated to with expansions.update")
+	s.Equal("newValue1", key1Value, "key1 should be set to what it was updated to with expansions.update")
+
+	key2Value := s.tc.taskConfig.Expansions.Get("key2")
+	s.Equal("newValue2", key2Value, "key2 should be set to what it was updated to with expansions.update")
+
+	key3Value := s.tc.taskConfig.Expansions.Get("key3")
+	s.Equal("expansionVar3", key3Value, "key3 should be the original expansion value")
+	s.Equal(&util.Expansions{}, s.tc.taskConfig.DynamicExpansions)
```
Kimchelly, line 389: nit: you can just check `s.Empty` rather than comparing it to the empty struct (same with the others).

agent/command_test.go
```diff
@@ -82,6 +82,7 @@ func (s *CommandSuite) SetupTest() {
 		Identifier: "project_identifier",
 	}, &patch.Patch{}, util.Expansions{})
 	s.Require().NoError(err)
+	s.Equal(&util.Expansions{}, taskConfig.DynamicExpansions)
```
Kimchelly, line 85: This isn't wrong, but it's a little weird to drive-by unit test the behavior of `internal.NewTaskConfig` in the agent suite. It seems like if we want to test the behavior of `internal.NewTaskConfig` itself, we should explicitly test that in the `internal` package, rather than drive-by test it in the setup here.

Kimchelly (changes requested):

Kimchelly (approved):
I'm going to LGTM this a little prematurely so we can deploy it and continue with the function vars bugfix timeline. I read it over and it seems like it's overall a correct piece of code - we can address the other style/best practices comments in a follow-up PR.

PR #6913, opened 2023-08-16
EVG-19949: Use enums for finder, planner, and dispatcher version fields

minnakt:
EVG-19949

### Description
Uses enums for the `FinderSettings.Version`, `PlannerSettings.Version`, and `DispatcherSettings.Version` fields because they have defined global consts in Evergreen so it makes sense to tie them together.

### Testing
- Updated tests


graphql/distro_resolver.go
```diff
@@ -16,6 +16,22 @@ import (
 	"github.com/evergreen-ci/utility"
```
minnakt, line 16: This is not generated and needs review

graphql/mutation_resolver.go
```diff
@@ -193,8 +193,12 @@ func (r *mutationResolver) SaveDistro(ctx context.Context, opts SaveDistroInput)
 		return nil, InputValidationError.Send(ctx, fmt.Sprintf("validating changes for distro '%s': '%s'", d.Id, validationErrs.String()))
 	}
 
-	if err = d.ReplaceOne(ctx); err != nil {
-		return nil, InternalServerError.Send(ctx, err.Error())
+	if err = data.UpdateDistro(ctx, oldDistro, d); err != nil {
```
minnakt, line 196: This is not generated and needs review. Also, I changed this to [`UpdateDistro`](https://github.com/evergreen-ci/evergreen/blob/105191bacaf22f9c5d800eb9a4be4d3ffac89730/rest/data/distro.go#L20) because it does a check for the dispatcher version which I thought might be important? But will remove if it's incorrect.

graphql/distro_resolver.go
```diff
@@ -42,6 +58,55 @@ func (r *distroResolver) ProviderSettingsList(ctx context.Context, obj *model.AP
...
+		return FinderVersionAlternate, nil
+	default:
+		return FinderVersionLegacy, InternalServerError.Send(ctx, fmt.Sprintf("finder version '%s' is invalid", utility.FromStringPtr(obj.Version)))
+	}
+}
+
+// Version is the resolver for the version field.
+func (r *plannerSettingsResolver) Version(ctx context.Context, obj *model.APIPlannerSettings) (PlannerVersion, error) {
+	if obj == nil {
+		return PlannerVersionTunable, InternalServerError.Send(ctx, "distro undefined when attempting to resolve planner version")
```
ablack12, line 84: if these nil cases are considered an error, can we just return the empty string? Returning a value implies that we're doing something with it (not sure if that's the case).

graphql/mutation_resolver.go
ablack12: Cool catch -- I actually see no reason why UpdateDistro is in rest/data as opposed to model/distro/distro.go; could you try moving it? If that doesn't cause too many problems, it would make it more discoverable.

ablack12 (changes requested):

graphql/mutation_resolver.go
minnakt: Sadly it gave me an import cycle 😔 But I kind of believe `UpdateDistro` belongs in rest/data because of how it is returning gimlet error responses (which are then consumed by the packages in rest/route), so I feel like it's okay 🤔

ablack12 (approved):

PR #6915, opened 2023-08-16
Revert "EVG-20463: allow Docker host.create to read file to stdin (#6901)"

Kimchelly:
Reverting because of a user report that Docker host.create is behaving strangely. I don't really understand how the new option could have caused this since they didn't specify the new option (and the existing behavior without the option should be the same as before), but reverting seems safe enough.

### DIFF
--- a/agent/command/host_create.go
+++ b/agent/command/host_create.go
@@ -95,14 +95,6 @@ func (c *createHost) Execute(ctx context.Context, comm client.Communicator,
 			return err
 		}
 	}
-	if c.CreateHost.StdinFile != "" {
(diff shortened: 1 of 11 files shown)


ablack12:
No need for approval on a revert, fwiw

ablack12 (approved):

Kimchelly:
I'm gonna close this actually - after reverting the deploy and some more investigation, it appears that Docker host.create still has the same user repoted issue in the reverted agent + app. Since nobody else is using Docker host.create, it's possible that Docker host.create has had issues for some time now. EVG-20686 will investigate this further.

PR #6916, opened 2023-08-16
EVG-19126 add lifecycle policy to docs

ablack12:
EVG-19126
### Description
In hindsight I don't think this needed to be an App ticket (or a ticket at all, can just update the docs at the time of responding to Stack Overflow since the info is all written out). Feel free to modify if anything here changes in the future.


ablack12:
Changes can be viewed here: https://github.com/evergreen-ci/evergreen/blob/90cf614b13297b54343a5d70e50fe00fdc711390/docs/Project-Configuration/Project-Commands.md

PR #6772, opened 2023-07-13
EVG-20280 Break up GetTaskGroup function

malikchaya2:
EVG-20280

### Description
GetTaskGroup was used as a way to return the "pre"/"post" blocks for regular tasks and task group which was confusing.  This breaks it up so that it doesn't have to pass back a fake "task group".

### Testing
This doesn't change the functionality, it just rearranges the code, so the existing tests should suffice.

PR #6918, opened 2023-08-16
EVG-20629: split callback timeout for pre and post

Kimchelly:
EVG-20629

### Description
Instead of using callback timeout for a bunch of command blocks, make a separate timeout for pre and post. To reduce the PR size, I'm going to save task group timeouts for a separate PR.

* Add pre and post block timeouts as a replacement for callback timeout.
* Set a default timeout for pre (2 hours). It previously had no timeout, so it could run until it hit the exec/idle timeout. I intentionally set it high because we don't know how users will want it to behave, so we want it to not time out in any reasonable case. Users can configure it lower if they expect it to finish faster.
* Increase the default timeout for post (15 min -> 2 hours). The previous timeout caused some tasks to silently time out in post because post doesn't fail the task by default, meaning some users may have been missing important post cleanup. I intentionally set it high to accommodate general user needs. Users can configure it lower if they expect it to finish faster, but Evergreen shouldn't be in charge of telling them how long their task ought to run.

### Testing
Added a lot of unit tests.

### Documentation
Added docs for new configurable timeouts.


agent/internal/task_config.go
```diff
@@ -186,10 +186,13 @@ func (tc *TaskConfig) GetPre(taskGroup string) (*CommandBlock, error) {
 		return nil, err
 	}
 
-	canFailTask := tc.Project.Pre == nil || tc.Project.PreErrorFailsTask
```
malikchaya2, line 189: this used to be returned like this: 
https://github.com/evergreen-ci/evergreen/pull/6772/files#diff-48f1ebe9d546f0e438a5d95b0d9c62328bc8586c35f7c241d538883ae4648b1eL157

why is this no longer necessary?

model/project.go
```diff
@@ -37,6 +37,7 @@ const (
 	waterfallTasksQueryMaxTime = 90 * time.Second
 )
 
+// Project represents the fully hydrated project configuration
```
malikchaya2, line 40: What does "fully hydrated" mean in this sense?

malikchaya2 (changes requested):

agent/internal/task_config.go
Kimchelly: This boolean is superfluous because `canFailTask` is only used to return the command block on L191, but that return can simply return `PreErrorFailsTask` as-is - whether or not pre is nil is irrelevant to whether or not pre fails the task.

model/project.go
Kimchelly: You can read about data hydration [here](https://stackoverflow.com/a/20787106). Basically, we're starting with a `ParserProject` (without tags expanded to all the matching tasks, so it's only partially hydrated), and then it gets translated to a `Project` (with tags expanded, so it's fully hydrated) so it's actually usable.

malikchaya2 (approved):
LGTM once tasks pass or confirmed unrelated.

Kimchelly:
It's unrelated, it's the new distro settings being flaky.

PR #6919, opened 2023-08-17
EVG-20474 introduce new middleware for update bulk annotations

ablack12:
EVG-20474

### Description
Add a middleware for the bulk annotations route that verifies that the user can update annotations for all projects they're asking for.

---
Background context:
Kim amazingly found that middleware validation is a bit broken because we use RequireProjectPermission in a lot of places, and this doesn't care if you don't pass in a project bc it'll match on something with an empty identifier.

Before fixing the underlying bug of not allowing ProjectPermission to validate with an empty identifier, I'm adding proper auth to all the routes that are effected by this. (And will maybe add an ADR or some comments to try and prevent this going forward). I'm going to do this one PR at a time for all of our sanity.

### Testing
Will test the bulk annotations route.

### Documentation
Added like three words


malikchaya2 (changes requested):
The code looks good! 
I don't see any tests though. Has this been tested on staging?

malikchaya2 (approved):
LGTM once tested

ablack12:
From testing, I learned that this route has never worked (bc it reads annotation as a task ID for the /tasks/{task_id} route) and the original requester has no plan for it.

PR #6921, opened 2023-08-17
EVG-20694 Update git clone string method

bynn:
EVG-20694

### Description
if the github app is defined, we should use the github app as the default auth for git clone 
else, fallback to legacy ssh 

### Testing
tested manually 
cloning private repos with old token works 
cloning private repos with new token works

created new private repo on staging called evergreen-ci/private-sandbox
tested that the patches created on sandbox (through pr or cli) can use git.get_project correctly with the app


Kimchelly (changes requested):
This seems kind of risky since it applies to OAuth token clones and GitHub app clones. Would changing this break all git clones (especially for private repos) if we have to revert the GitHub app back to the OAuth token?

bynn:
No, this change makes git clone compatible with both oauth tokens and github app tokens. 
I was unable to find any documentation about it but I was able to manually clone private repos using this string so it is backwards compatible

Kimchelly (commented):
This tentatively sounds fine if you manually tested it for private repos, but I hesitate to rely on something that has no documentation (all the scant docs I found online talk about `x-oauth-basic`). Gonna wait on GitHub support's response about `x-access-token` before approving.

bynn:
image

@Kimchelly it looks like documentation does not exist

agent/command/git.go
```diff
@@ -186,17 +192,25 @@ func (opts cloneOpts) getCloneCommand() ([]string, error) {
...
 }
 
-func (opts cloneOpts) buildHTTPCloneCommand() ([]string, error) {
+func (opts cloneOpts) buildHTTPCloneCommand(forApp bool) ([]string, error) {
 	urlLocation, err := url.Parse(opts.location)
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing URL from location")
 	}
-	clone := fmt.Sprintf("git clone %s '%s'", thirdparty.FormGitUrl(urlLocation.Host, opts.owner, opts.repo, opts.token), opts.dir)
+	gitUrl := ""
```
Kimchelly, line 207: nit: some people prefer to use `var gitURL string` to indicate that the empty string is not a meaningful value, and will be actually initialized later on (i.e. in the if-else)

thirdparty/git.go
```diff
@@ -309,3 +309,11 @@ func FormGitUrl(host, owner, repo, token string) string {
 
 	return fmt.Sprintf("git@%s:%s/%s.git", host, owner, repo)
 }
+
+func FormGitUrlForApp(host, owner, repo, token string) string {
```
Kimchelly, line 313: nit: I know that the old function was already using `Url` but can we make it `URL` to be more consistent with our style guide?

Kimchelly (approved):
The change itself seems fine mod small things. I have a couple questions about the roll-out process:
* Is enabling/disabling this change as simple as setting/clearing the GitHub app ID from the admin settings?
* Do we have a concrete plan for how to monitor `git.get_project` after the deploy goes out?

thirdparty/git.go
bynn: updated

bynn:
> Is enabling/disabling this change as simple as setting/clearing the GitHub app ID from the admin settings?
Yes, but we would also have to restart the servers which is documented in the playbook and the github app ops guide 

> Do we have a concrete plan for how to monitor git.get_project after the deploy goes out?
Yes, I plan to manually monitor git.get_project errors on honeycomb

I'm merging this because this change won't affect current prod until the app is set up again

PR #6929, opened 2023-08-21
EVG-20287 Remove unnecessary conflict check in next task

malikchaya2:
EVG-20287
### Description
The check is unnecessary as it will never be hit. It cannot receive a task conflict from next task. It could receive a host conflict, but it seems like that specific failure is never returned back to the agent from GetNextTask anyways. Therefore, that error block never runs.


hadjri (changes requested):
Looks like we can remove [this](https://gist.github.com/hadjri/abb921aae2f42202ac6098337f748b40#file-request-go-L27) and [this](https://gist.github.com/hadjri/00df27e209daae01b0fcf29b0e0b4ecd#file-base_client-go-L322) as well since they're unused now

malikchaya2:
> Looks like we can remove [this](https://gist.github.com/hadjri/abb921aae2f42202ac6098337f748b40#file-request-go-L27) and [this](https://gist.github.com/hadjri/00df27e209daae01b0fcf29b0e0b4ecd#file-base_client-go-L322) as well since they're unused now

TaskConflict is still used by doHeartbeat in background.go. 
I removed HTTPConflictError.

hadjri (approved):

Kimchelly (commented):
Malik is actually right, and `TaskConflict` is never set anywhere, so it can also be deleted.

PR #6896, opened 2023-08-11
EVG-19952: Use enum for CloneMethod field

sophstad:
Reopening because I misnamed the branch in #6895 😭 

Updates field used in UI ticket [EVG-19952](https://jira.mongodb.org/browse/EVG-19952)

### Description

- Make a distro's CloneMethod field typing more predictable by using an enum. The field resolvers convert between the GraphQL enum and Evergreen's own consts.

### Testing

- Update tests

### Documentation

N/A

PR #6930, opened 2023-08-21
EVG-19945: Use enum for distro provider

sophstad:
EVG-19945

### Description
- Use an enum for more predictable typing of a distro's provider field. This PR is in the same vein as #6896 and #6913

### Testing
- Update unit test

### Documentation
N/A

### DIFF
--- a/graphql/distro_resolver.go
+++ b/graphql/distro_resolver.go
@@ -48,6 +48,26 @@ func (r *distroResolver) CloneMethod(ctx context.Context, obj *model.APIDistro) 
 	}
 }
 
+// Provider is the resolver for the provider field.
+func (r *distroResolver) Provider(ctx context.Context, obj *model.APIDistro) (Provider, error) {
+	if obj == nil {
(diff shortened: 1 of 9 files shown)


graphql/distro_resolver.go
```diff
@@ -48,6 +48,26 @@ func (r *distroResolver) CloneMethod(ctx context.Context, obj *model.APIDistro)
...
+		return "", InternalServerError.Send(ctx, "distro undefined when attempting to resolve provider")
+	}
+
+	switch utility.FromStringPtr(obj.Provider) {
+	case evergreen.ProviderNameDocker:
+		return ProviderDocker, nil
+	case evergreen.ProviderNameEc2Fleet:
+		return ProviderEc2Fleet, nil
+	case evergreen.ProviderNameEc2OnDemand:
+		return ProviderEc2Ondemand, nil
```
ablack12, line 63: nit: can this be OnDemand? or is this generated in such a way that we'd then have to do ON_DEMAND for the enum? I might still prefer that

PR #6931, opened 2023-08-21
EVG-18992 Allow terminating static hosts 

malikchaya2:
EVG-18992

### Description
This shores up the soundness of the static provider's interaction with the host termination job by ensuring TerminateInstances allows terminating a host.  

### Testing
Tested on staging to ensure a static host can be terminated. I created a fake static host, changed the status to "terminated" in the UI, and ensured that the host is deleted from the db.

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -475,6 +475,10 @@ func (h *Host) IsEphemeral() bool {
 	return utility.StringSliceContains(evergreen.ProviderSpawnable, h.Provider)
 }
 
+func (h *Host) IsStatic() bool {
+	return h.Provider == evergreen.ProviderNameStatic
+}
+
 func (h *Host) IsContainer() bool {
 	return utility.StringSliceContains(evergreen.ProviderContainer, h.Provider)
 }
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -476,7 +476,7 @@ func (h *Host) IsEphemeral() bool {
 }
 
 func (h *Host) IsStatic() bool {
-	return h.Provider == evergreen.ProviderNameStatic
+	return h.Distro.Provider == evergreen.ProviderNameStatic
 }
 
 func (h *Host) IsContainer() bool {
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -475,10 +475,6 @@ func (h *Host) IsEphemeral() bool {
 	return utility.StringSliceContains(evergreen.ProviderSpawnable, h.Provider)
 }
 
-func (h *Host) IsStatic() bool {
-	return h.Distro.Provider == evergreen.ProviderNameStatic
-}
-
 func (h *Host) IsContainer() bool {
 	return utility.StringSliceContains(evergreen.ProviderContainer, h.Provider)
 }
--- a/units/host_termination.go
+++ b/units/host_termination.go
@@ -108,7 +108,6 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 			"status":   j.host.Status,
 			"provider": j.host.Distro.Provider,
 			"message":  "host termination for a non-spawnable distro",
-			"cause":    "programmer error",
 		})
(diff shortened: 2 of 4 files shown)


model/host/host.go
```diff
@@ -475,6 +475,10 @@ func (h *Host) IsEphemeral() bool {
 	return utility.StringSliceContains(evergreen.ProviderSpawnable, h.Provider)
 }
 
+func (h *Host) IsStatic() bool {
+	return h.Provider == evergreen.ProviderNameStatic
```
Kimchelly, line 479: Should probably check `h.Distro.Provider` instead of `h.Provider` because `h.Distro.Provider` is more widely used, and we may deprecate `h.Provider` one day (it's the same thing as `h.Distro.Provider).`

units/host_termination.go
```diff
@@ -100,7 +100,7 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 		j.env = evergreen.GetEnvironment()
 	}
 
-	if !j.host.IsEphemeral() {
+	if !j.host.IsEphemeral() && !j.host.IsStatic() {
```
Kimchelly, line 103: Rather than run through all the logic in the host termination job for static hosts, I'd prefer to _just_ call `j.host.Terminate` in this block. The host termination job is really 99% written for ephemeral host termination, not static hosts. So for static hosts, since we aren't really doing anything other than setting its state to terminated, I think we should just do it directly here.

Kimchelly (changes requested):

units/host_termination.go
```diff
@@ -100,6 +100,13 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 		j.env = evergreen.GetEnvironment()
 	}
 
+	if j.host.IsStatic() {
```
Kimchelly, line 103: This is equivalent to `j.host.IsEphemeral`, so it can be absorbed into the block below this. It will also help future investigations to preserve the grip notice in case we one day somehow introduce a bug where static hosts get selected for automatic termination (rather than by manual action).

model/host/host.go
```diff
@@ -475,6 +475,10 @@ func (h *Host) IsEphemeral() bool {
 	return utility.StringSliceContains(evergreen.ProviderSpawnable, h.Provider)
 }
 
+func (h *Host) IsStatic() bool {
```
Kimchelly, line 478: This functions is equivalent to `j.host.IsEphemeral`. I also don't think we need to worry about future-proofing that much since there's no real plans to introduce some third kind of cloud provider that's neither static nor ephemeral.

Kimchelly (changes requested):

units/host_termination.go
```diff
@@ -110,6 +110,9 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 			"message":  "host termination for a non-spawnable distro",
 			"cause":    "programmer error",
```
Kimchelly, line 111: Remove the "programmer error" bit since this is now an allowed workflow, not a bug.

Kimchelly (approved):
LGTM mod one deletion.

units/host_termination.go
```diff
@@ -108,8 +108,10 @@ func (j *hostTerminationJob) Run(ctx context.Context) {
 			"status":   j.host.Status,
 			"provider": j.host.Distro.Provider,
 			"message":  "host termination for a non-spawnable distro",
-			"cause":    "programmer error",
 		})
+		if err := j.host.Terminate(ctx, evergreen.User, j.TerminationReason); err != nil {
+			j.AddError(errors.Wrapf(err, "terminating intent host '%s' in DB", j.host.Id))
```
Kimchelly, line 113: "terminating host '%s' in the DB" - there are no static intent hosts.

Kimchelly (approved):
LGTM mod one nit.

PR #6933, opened 2023-08-21
EVG-20315: Convert smoke test to Go test

Kimchelly:
EVG-20315

### Description
The smoke test is now a plain Go test, not a custom-made binary. This gives us some advantages like being able to clearly see failed assertions as soon as they happen rather than reading through many task log lines for the failure log, getting nicely parsed test results, and the test result cut out random noise in the logs like background `mongod` logs.

The large majority of this PR is moving the logic from being a plain executable to being a Go test. Since this is a large PR, I'm going to save some cleanup for a follow-up PR.

The tests are renamed as follows:
* `smoke-test-host-task` -> `test-smoke-host`
* `smoke-test-container-task` -> `test-smoke-container`
* `smoke-test-agent-monitor` -> `test-smoke-agentmonitor`
* `smoke-test-endpoints` -> `test-smoke-endpoint`

#### Changes
* Transcribe smoke test checks into Go test. The checks are the same as before, just written in Go test style.
* Move smoke test setup (like creating app servers/agents) into Go test rather than leave it in the makefile. Each smoke test is in its own package and they share common smoke test utility code.
* Copy-paste smoke test files into the `smoke/testdata` directory - the duplicate `testdata/smoke` is going to be deleted in a follow-up PR.

### Testing
Ran the smoke tests in a patch and they still tested the same conditions.

### Documentation
N/A - the existing docs on how to read the smoke test still apply.


Kimchelly:
Not gonna bump the CLI version despite the linter failure because the change to remove the smoke sub-command from `operations/service.go` does not affect users at all.

smoke/host/util.go
```diff
@@ -0,0 +1,295 @@
...
+	ExecModeSecret string
+	CLIConfigPath  string
+	ProjectID      string
+	BVName         string
+}
+
+// GetSmokeTestParamsFromEnv gets the parameters for the host smoke test from
+// the environment. It sets defaults where possible. Note that the default data
+// depends on the setup test data for the smoke test.
+func GetSmokeTestParamsFromEnv(t *testing.T) SmokeTestParams {
```
ablack12, line 36: Is it really worth having multiple packages in this smoke directory? I'd advocate for flattening it -- then we can keep everything private, and I think it's a bit clearer that everything in here exists only to be used here. This package isn't going to grow to the extent that I think we need. (I could see a case for keeping host and container separate but it seems like that could just handled with file names)

smoke/internal/util.go
```diff
@@ -0,0 +1,349 @@
...
+		return nil, errors.Wrap(err, "reading response body")
+	}
+	if resp.StatusCode >= 400 {
+		return body, errors.Errorf("got HTTP response code %d with error %s", resp.StatusCode, string(body))
+	}
+
+	return body, nil
+}
+
+func SmokeRunBinary(ctx context.Context, name, wd, bin string, cmdParts ...string) (*exec.Cmd, error) {
```
ablack12, line 284: Can we add a comment here and describe the inputs?

ablack12 (changes requested):
This change is wild! 🚀

ablack12:
There are a few references to the smoke tests in the dev guide we may want to update also.

smoke/internal/util.go
```diff
@@ -0,0 +1,349 @@
...
+	grip.Infof("Running command: %s", append([]string{bin}, cmdParts...))
+	cmd := exec.CommandContext(ctx, bin, cmdParts...)
+	cmd.Env = append(os.Environ(), fmt.Sprintf("EVGHOME=%s", wd))
+	cmdSender := send.NewWriterSender(send.MakeNative())
+	cmd.Dir = wd
+	cmdSender.SetName(name)
```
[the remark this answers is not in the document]
Kimchelly: Yeah, it's so that we can see which output logs came from which process.

smoke/internal/util.go
```diff
@@ -0,0 +1,349 @@
...
+		"start-evergreen",
+		"--agent",
+		fmt.Sprintf("--mode=%s", mode),
+		fmt.Sprintf("--exec_mode_id=%s", execModeID),
+		fmt.Sprintf("--exec_mode_secret=%s", execModeSecret),
+		fmt.Sprintf("--api_server=%s", params.AppServerURL),
+		fmt.Sprintf("--binary=%s", params.CLIPath),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "starting Evergreen smoke test app server")
```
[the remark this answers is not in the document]
Kimchelly: Sure, we can require no error here. I also changed some of the others that could also require no error and exit early.

smoke/internal/util.go
Kimchelly: Done.

Kimchelly:
> There are a few references to the smoke tests in the dev guide we may want to update also.

I already updated the section about the smoke test, and I also just removed the other reference to the smoke test.

smoke/host/util.go
ablack12: The last bit works for me, thanks!

smoke/internal/host/util.go
```diff
@@ -0,0 +1,298 @@
...
+	const repotrackerAttempts = 10
+	for i := 0; i < repotrackerAttempts; i++ {
+		time.Sleep(2 * time.Second)
+		respBody, err := internal.MakeSmokeRequest(ctx, params.APIParams, http.MethodGet, client, fmt.Sprintf("/rest/v2/projects/%s/versions?limit=1", params.ProjectID))
+		if err != nil {
+			grip.Error(errors.Wrapf(err, "requesting latest version for project '%s'", params.ProjectID))
+			continue
+		}
+		if len(respBody) == 0 {
+			grip.Error(errors.Errorf("did not find any latest revisions yet for project '%s'", params.ProjectID))
```
ablack12, line 165: Can we update this (and the following) to grip.Errorf as well?

ablack12 (approved):

smoke/internal/host/util.go
Kimchelly: Sure, I updated the others in the smoke tests.

PR #6934, opened 2023-08-22
EVG-19629 disallow jira comment/issue subscription for every task

ablack12:
EVG-19629

### Description
This is already disallowed on the frontend but ensures it can't be worked around on routes or etc.

### Testing
Tested via rest routes that we receive the following error for both issues and comments:
```
{
    "status": 400,
    "message": "JIRA comment/issue subscription not allowed for all tasks in the project",
    "metadata": {
        "time": "0001-01-01T00:00:00Z"
    }
}
```


model/event/subscriptions.go
```diff
@@ -530,6 +530,7 @@ func RemoveSubscription(id string) error {
 	})
 }
 
+// IsSubscriptionAllowed disallows subscriptions that abuse Jira or webhooks.
```
hadjri, line 533: Can we absorb this into Validate rather than have its own exported function?

hadjri (changes requested):

model/event/subscriptions.go
ablack12: Yeah seems reasonable; I want to avoid doing too much cleanup work here since subscriptions are their own world, but this function isn't individually unit tested or anything so idk why it lives alone 🤷‍♀️

ablack12: I also removed some of the old logic; I think it's trying to prevent us from commenting on Jira or creating a webhook for particular versions,tasks,builds, but it doesn't seem to work, and there's no reason for that really; this logic is a bit stronger.

ablack12: Jk I'm going to keep it in;  I think it's actually checking something different (there are certain options, like creating a jira issue, that we don't show on the new UI even for regular tasks)

hadjri (approved):

model/event/subscriptions.go
hadjri: Makes sense, good catch!

PR #6935, opened 2023-08-22
EVG-20474 remove bulk annotation route

ablack12:
EVG-NNNNN

### Description
Originally this ticket was to fix the middleware for this route; however in testing, we learned that this route never worked and has never been used, and the requester has no intention of using it. We should verify closer with users that they actually intend to do follow up work for new routes in the future.

### Testing

### Documentation
Removed documentation


rest/route/middleware.go
```diff
@@ -599,6 +600,13 @@ func updateHostAccessTime(ctx context.Context, h *host.Host) {
 // canAlwaysSubmitPatchesForProject returns true if the user is a superuser or project admin,
 // or is authorized specifically to patch on behalf of other users.
 func canAlwaysSubmitPatchesForProject(user *user.DBUser, projectId string) bool {
+	if projectId == "" {
+		grip.Error(message.Fields{
```
Kimchelly, line 604: nit: you can simplify this down to `grip.ErrorWhen(projectId == "", ...)`

Kimchelly (approved):
LGTM mod nit.

rest/route/middleware.go
ablack12: Oh I meant to also return here since the follow up lines are useless

PR #6936, opened 2023-08-22
EVG-20474 add validation to submit patch

ablack12:
EVG-20474 
### Description
Added permission check for the submit patch (originally we had a SubmitPatch middleware, but Kim removed this since it didn't work, bc it relied on the query parameters). Since this is a pretty special case and the APIServer routes are already a bit wild west I think adding this to the route directly is fine.

### Testing
Will create a patch with and without permissions to test.


hadjri (commented):
Code LGTM, were the test patches created in staging?

ablack12:
Just tested:
With an unrestricted project and basic permissions, I was [able to patch](https://spruce-staging.corp.mongodb.com/version/64e518d7b237369f9f71c05f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). 
With a restricted project and special permissions, I was [able to patch](https://spruce-staging.corp.mongodb.com/patch/64e51a759dbe32521805d388/configure/tasks).
With an restricted project and no patch permissions, I was unable to patch:
```
Unexpected reply from server (401 Unauthorized): 400 (Bad Request): not authorized to patch for project 'evg' (Possibly user credentials are expired, try logging in again via the Evergreen web UI.)
```

PR #6937, opened 2023-08-22
EVG-20400: fix race in test-rest-data

Kimchelly:
EVG-20400

### Description
This test began racing because of a few changes:

* The test is set up to use [a local queue group](https://github.com/evergreen-ci/evergreen/blob/63cc5d7f42e4f2c74ecaf550c53490c3cd7a9df8/rest/data/host_create_test.go#L134-L137). This behaves differently from the remote queue because in the local queue, all jobs are stored in-memory (unexported fields like the `evergreen.Environment` do not get cleared in between the job enqueueing and the job running because there's no roundtrip through the DB).
* The [host creation job is enqueued immediately](https://github.com/evergreen-ci/evergreen/blob/63cc5d7f42e4f2c74ecaf550c53490c3cd7a9df8/rest/data/host_create.go#L448-L450) when host.create inserts the host document.
* The job runs in the local queue, and multiple jobs run in parallel because the local queue group has 2 workers. These jobs both read and [modify the underlying hostinit settings](https://github.com/evergreen-ci/evergreen/blob/63cc5d7f42e4f2c74ecaf550c53490c3cd7a9df8/units/provisioning_create_host.go#L113). Since they all have the same underlying `evergreen.Environment`, this modifies the copy of settings shared between all the enqueued jobs.
* Race occasionally happens when one job is reading while another job is modifying.

I basically just fixed it by not modifying the shared admin settings copy.

### Testing
Ran the unit test a bunch of times.

### Documentation
N/A

### DIFF
--- a/units/provisioning_create_host.go
+++ b/units/provisioning_create_host.go
@@ -38,9 +38,10 @@ type createHostJob struct {
 	BuildImageStarted bool   `bson:"build_image_started" json:"build_image_started" yaml:"build_image_started"`
 	job.Base          `bson:"metadata" json:"metadata" yaml:"metadata"`
 
-	start time.Time
-	host  *host.Host
-	env   evergreen.Environment
+	start    time.Time
+	host     *host.Host
+	env      evergreen.Environment
+	settings evergreen.Settings
 }
 
 func makeCreateHostJob() *createHostJob {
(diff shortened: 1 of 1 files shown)


ablack12:
Reminder Jonathan is on family leave now; is there anyone else on PLT you'd want to look at this or was that just for visibility?

Kimchelly:
Oh yeah I forgot. He doesn't need to be the one to review.

ablack12 (approved):

PR #6939, opened 2023-08-22
EVG-20681 Add function to check if app is installed on a repo

bynn:
EVG-20681

### Description
Added a function that returns a bool whether the github app is installed on a given repo 
also updated skipWebhook function to use this instead of GetGithubAppAuth because it's more accurate. 

### Testing
tested on staging

### DIFF
--- a/config.go
+++ b/config.go
@@ -556,6 +556,17 @@ func (s *Settings) GetGithubAppAuth() *githubAppAuth {
 
 // HasAppInstalled returns true if the GitHub app is installed on the repo.
 func (s *Settings) HasAppInstalled(ctx context.Context, owner, repo string) bool {
+	if owner == "" || repo == "" {
+		grip.Debug(message.Fields{
+			"message": "owner/repo not specified",
+			"caller":  "HasAppInstalled",
+			"owner":   owner,
+			"repo":    repo,
+			"ticket":  "EVG-19966",
+		})
+		return false
+	}
+
 	authFields := s.GetGithubAppAuth()
 	if authFields == nil {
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -554,6 +554,76 @@ func (s *Settings) GetGithubAppAuth() *githubAppAuth {
...
+			"appId":   authFields.AppId,
+			"ticket":  "EVG-19966",
+		}))
+		return false
+	}
+
+	itr := ghinstallation.NewAppsTransportFromPrivateKey(httpClient.Transport, authFields.AppId, key)
+	httpClient.Transport = itr
+	client := github.NewClient(httpClient)
+	installationId, _, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
```
ablack12, line 601: Does this use our github API limit? Calling this every time we need a token is going to be really expensive. In general this will probably be time consuming also -- should we be caching the installationId or is this likely to change?

config.go
```diff
@@ -554,6 +554,76 @@ func (s *Settings) GetGithubAppAuth() *githubAppAuth {
...
+	if err != nil {
```
ablack12, line 602: A lot of this also seems duplicated with CreateInstallationToken; should this function return the installation ID, and then CreateInstallationToken can call it, and use the installationID to create?

ablack12 (changes requested):

bynn:
The ticket seemed unnecessary now that we had a talk with IT and every repo that uses Evergreen has organizational access to every repo.

PR #6940, opened 2023-08-22
EVG-20633 Send pending status for merge queue variants

bsamek:
EVG-20633

### Description

Please see the added section in the ADR in this PR.

### Testing

I added unit tests for the new functions this PR depends on. I will also test
the new pending behavior in staging.

### Documentation

Edited ADR.

### DIFF
--- a/docs/decisions/2023-06-05_github_merge_queue.md
+++ b/docs/decisions/2023-06-05_github_merge_queue.md
@@ -64,3 +64,28 @@ job](https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go).
 * New clone logic in the agent will clone the merge group branch.
 * Evergreen will post the result to the GitHub checks API.
 * A new UI element will allow users to opt into the GitHub merge queue.
+
+## Pending Status Checks
+
+Evergreen doesn't behave in the way that GitHub designers intended: For PRs,
+instead of sending a pending and then a success/fail status for each check, it
+sends a single "evergreen" pending status, and then it sends "evergreen"
+success/failure as well as "evergreen/\<variant\>" status for each variant. This
+is fine for PRs because "evergreen" links to the version in Evergreen.
+
+However, for GitHub merge queue builds, sending an "evergreen" pending check
+only works if "evergreen" is configured as a branch protection rule. If instead
+the user has configured an "evergreen/\<variant\>" branch protection rule but
+_not_ an "evergreen" branch protection rule, the "evergreen" pending check does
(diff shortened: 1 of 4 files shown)


docs/decisions/2023-06-05_github_merge_queue.md
```diff
@@ -64,3 +64,28 @@ job](https://github.com/evergreen-ci/evergreen/blob/main/units/patch_intent.go).
...
+the user has configured an "evergreen/\<variant\>" branch protection rule but
+_not_ an "evergreen" branch protection rule, the "evergreen" pending check does
+not show up in the GitHub UI, which means the user can't click a link to the
+Evergreen version. They eventually can, when it completes, but not while it's running.
+
+There are several options to handle this:
+
+1. Accept the problem that there is no link to the version until it finishes.
+2. Send status pending for variants in the version.
+3. Send status pending for branch protection rules.
```
malikchaya2, line 87: if we send status pending for branch protection rules, what happens if "evergreen" is not set up as a branch protection rule? Is this pain point solved?

> However, for GitHub merge queue builds, sending an "evergreen" pending check
only works if "evergreen" is configured as a branch protection rule. If instead
the user has configured an "evergreen/\<variant\>" branch protection rule but
_not_ an "evergreen" branch protection rule, the "evergreen" pending check does
not show up in the GitHub UI, which means the user can't click a link to the
Evergreen version.

bsamek: Yes, the logic is to send a pending status for the first 10 branch protection rules with the prefix "evergreen". If there are none, or if we can't get them, just send "evergreen" and hope for the best.

units/patch_intent.go
```diff
@@ -511,19 +511,41 @@ func (j *patchIntentProcessor) createGitHubMergeSubscription(ctx context.Context
...
+	// protection rules, so log the error, but don't return it.
+	grip.Error(message.WrapError(err, message.Fields{
+		"job":      j.ID(),
+		"job_type": j.Type,
+		"message":  "failed to get branch protection rules",
+		"org":      p.GithubMergeData.Org,
+		"repo":     p.GithubMergeData.Repo,
+		"branch":   p.GithubMergeData.BaseBranch,
+	}))
+	if len(rules) == 0 {
```
malikchaya2, line 535: Can you please add a comment here explaining that this is meant to avoid a regression in case we don't have permission to get the rules?

malikchaya2 (approved):

PR #6942, opened 2023-08-22
Rename materialized intelligence to Evergreen Data for Analytics.

johndaniels:
It's unclear what materialized intelligence means. The new name makes it clearer that it's evergreen data to be used for analytics.


julianedwards (changes requested):
Also need to update the reference to this page [here](https://github.com/evergreen-ci/evergreen/blob/main/docs/API/REST-V2-Usage.md?plain=1#L1549).

PR #6945, opened 2023-08-23
EVG-20595: set task description for failure in test results and setup.initial

Kimchelly:
EVG-20595

### Description
If the test results contain a failure, the end task response is updated right before marking the task finished to indicate the test failed (changing success -> failure). When this happened, the failure included the wrong "failing command" because it kept the task end description that it originally received.

* Ensure that tasks ending with success do not record a failing command type/description. The one exception is the case where it's success and the user explicitly set the task status and description, which we do record because it's their custom description.
* Drive-by fix an issue with handling task `setup.initial` failures where it didn't set the command failure type to system failure.

### Testing
Updated unit tests.
Spot checked in staging that a task containing a Go test failure showed the failure message and a task that succeeded did not show any failure message.

### Documentation
N/A


Kimchelly:
evergreen retry

PR #6946, opened 2023-08-23
EVG-20360: Check user permissions in RemoveItemFromCommitQueue resolver

minnakt:
EVG-20360

### Description
Right now, the `RemoveItemFromCommitQueue` resolver does not check any permissions before removing a patch from the commit queue. This means that a basic user can remove anyone else's patch from the commit queue.

The rest route (which is also used by the CLI) has a middleware called `requireCommitQueueItemOwner` which checks if a given user can remove a patch from the commit queue. I just extracted this code into a separate function called `CheckCanRemoveCommitQueueItem` and reused that function in the rest route and the resolver.

### Testing
- Added unit tests for `CheckCanRemoveCommitQueueItem`
- Updated GraphQL tests
- Manually on staging, tested three different scenarios:
    - I can remove my own patch if I'm a basic user
    - I cannot remove Bynn's patch if I'm a basic user
    - I can remove Bynn's patch if I'm a superuser

### Screenshot
Me failing to remove Bynn's patch as a basic user:

not_authorized


graphql/mutation_resolver.go
```diff
@@ -1265,8 +1265,23 @@ func (r *mutationResolver) UpdateUserSettings(ctx context.Context, userSettings
 
 // RemoveItemFromCommitQueue is the resolver for the removeItemFromCommitQueue field.
 func (r *mutationResolver) RemoveItemFromCommitQueue(ctx context.Context, commitQueueID string, issue string) (*string, error) {
-	username := gimlet.GetUser(ctx).DisplayName()
-	result, err := data.FindAndRemoveCommitQueueItem(ctx, commitQueueID, issue, username, fmt.Sprintf("removed by user '%s'", username))
+	user := mustHaveUser(ctx)
+
+	project, err := data.FindProjectById(commitQueueID, false, false)
```
ablack12, line 1270: includeRepo should be true, since a project could inherit commit queue settings from the repo (might be worth adding a test for this)

graphql/mutation_resolver.go
```diff
@@ -1265,8 +1265,23 @@ func (r *mutationResolver) UpdateUserSettings(ctx context.Context, userSettings
...
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("could not find project '%s': %s",
+			commitQueueID, err.Error()))
+	}
+
+	if err = data.CheckCanRemoveCommitQueueItem(ctx, r.sc, user, project, issue); err != nil {
```
ablack12, line 1276: Are we able to check this in a directive the way we check project settings?

ablack12 (changes requested):

graphql/mutation_resolver.go
minnakt: oops yes... added directive with new tests in which I tested for the `includeRepo` case 😃

rest/data/commit_queue.go
```diff
@@ -577,3 +580,87 @@ func GetAdditionalPatches(patchId string) ([]string, error) {
...
+		if err != nil {
+			return gimlet.ErrorResponse{
+				StatusCode: http.StatusInternalServerError,
+				Message:    errors.Wrapf(err, "finding patch '%s'", itemId).Error(),
+			}
+		}
+		if usr.Id != utility.FromStringPtr(patch.Author) {
+			return gimlet.ErrorResponse{
+				StatusCode: http.StatusUnauthorized,
+				Message:    "not authorized to patch on behalf of author",
```
ablack12, line 608: Sorry to only catch this late, but this error message (and the Github one below) seem wrong; they may be outdated from when this check was only for patching? I'm not sure. It would probably more accurate to say "not authorized to perform action on behalf of author" since this is used in a few places.

ablack12 (approved):

rest/data/commit_queue.go
minnakt: makes sense, I will update this text!

minnakt:
Tested again just for sanity checking (error message is a bit wonk because Spruce appends an error string, probably can just remove at some point):
Screenshot 2023-08-31 at 9 32 56 AM

Merging!

PR #6948, opened 2023-08-23
EVG-20474 check if project is empty for middleware

ablack12:
EVG-20474
### Description
This change is a bit risky, since if it's being called somewhere where we aren't actually using urlVars then this will start failing, whereas before it just succeeded using an empty project identifier. However after auditing our code base for it, I believe I've caught the cases.

### Testing
Can't really test every route, but will poke around in staging.

### DIFF
--- a/model/version_db.go
+++ b/model/version_db.go
@@ -152,7 +152,7 @@ func VersionByProjectIdAndOrder(projectId string, revisionOrderNumber int) db.Q 
 		}).Sort([]string{"-" + VersionRevisionOrderNumberKey})
 }
 
-// ByLastVariantActivation finds the most recent non-patch, non-ignored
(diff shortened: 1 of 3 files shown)


ablack12:
@malikchaya2 I'll take a look at tests; do the changes look good to you besides that?

malikchaya2:
> @malikchaya2 I'll take a look at tests; do the changes look good to you besides that?

Because it's TestProjectViewPermission that's failing, I'd want to re-review after the tests pass.

malikchaya2 (approved):

ablack12:
Tested some rest routes that use requireProjectAdmin and confirmed they still work. There is risk that users could've been relying on some of these routes allowing non-admins, but in that case they should be admins or have a use case for an admin-powered bot.

PR #6953, opened 2023-08-24
EVG-20633 Log for default merge queue status

bsamek:
EVG-20633

### Description

I'd like to use this debug logging to check to see what projects, if any, either aren't setting branch protection rules or don't have permission for Evergreen to view them. This will help with the migration.

### Testing

None needed. It's debug logging.

### Documentation

N/A

### DIFF
--- a/units/patch_intent.go
+++ b/units/patch_intent.go
@@ -534,6 +534,14 @@ func (j *patchIntentProcessor) createGitHubMergeSubscription(ctx context.Context
 	}))
 	// If we don't find any rules, send the default.
 	if len(rules) == 0 {
+		grip.Debug(message.Fields{
+			"job":      j.ID(),
+			"job_type": j.Type,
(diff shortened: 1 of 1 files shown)


units/patch_intent.go
```diff
@@ -534,6 +534,14 @@ func (j *patchIntentProcessor) createGitHubMergeSubscription(ctx context.Context
 	}))
 	// If we don't find any rules, send the default.
 	if len(rules) == 0 {
+		grip.Debug(message.Fields{
```
malikchaya2, line 537: non blocking: 
you might want to log p.Project as well, it makes it a bit easier to search for things and figure out who to talk to (so that you can search for the project page and don't need to query the db)

malikchaya2 (approved):

units/patch_intent.go
bsamek: oh good point thanks!

PR #6954, opened 2023-08-25
EVG-20629: split callback timeout into more granular task group timeouts

Kimchelly:
EVG-20629

### Description
* Split callback timeout into individual user-configurable timeouts per task group block (`setup_group`, `setup_task`, `teardown_task`, `teardown_group`).
    * Callback timeout only applies to the `timeout` block now.
    * `setup_group` and `setup_task` default timeouts are both 2 hours to be consistent with `pre`'s default. For `setup_group`, this is an increase from its default of 15 minutes, but I consider this not particularly problematic since the vast majority (over 99%) of pre-task commands finish in under 5 minutes, and the remaining few still finished in ~20 minutes or so.
    * `teardown_task` default timeout is 30 minutes to remain consistent with `post`'s default.
    * `teardown_group` has an unchanged default timeout of 15 minutes.
* Reorganize some task context/task config functions to better share the timeout logic.
* Small rename of a couple variables and shifting around of agent methods to belong in more appropriate places.

### Testing
* Updated unit tests.
* Tested in staging with explicit timeout settings for each block in a task group and the timeouts triggered command/task failures as described in the docs.

### Documentation
* Updated docs for task group for new settings.
* Updated existing `pre`, `post`, and `timeout` docs to list the parameters in a way that's consistent with the task group docs.
* Write an ADR for the callback timeout rework.


config.go
```diff
@@ -35,10 +35,10 @@ var (
 	BuildRevision = ""
 
 	// ClientVersion is the commandline version string used to control auto-updating.
-	ClientVersion = "2023-08-23"
+	ClientVersion = "2023-08-25"
```
Kimchelly, line 38: Had to bump the CLI because `evergreen validate` uses the CLI's local data model of the parser project for validation. Therefore, if we don't bump it, `evergreen validate` will spuriously warn users that the new project fields don't exist when they add it to their YAMLs.

agent/task_context.go
```diff
@@ -191,14 +192,262 @@ func (tc *taskContext) getExecTimeout() time.Duration {
...
+// getTeardownTaskTimeout returns the timeout for the teardown task block.
+func (tc *taskContext) getTeardownTaskTimeout(tg *model.TaskGroup) func() time.Duration {
+	return func() time.Duration {
+		tc.RLock()
+		defer tc.RUnlock()
+
+		if tg.TeardownTaskTimeoutSecs != 0 {
+			return time.Duration(tg.TeardownTaskTimeoutSecs) * time.Second
+		}
+		return defaultPostTimeout
```
malikchaya2, line 416: https://github.com/evergreen-ci/evergreen/pull/6918 had set the defaultPostTimeout to 2hrs. I think we might be running the risk of missing important post cleanup things by setting this to the no 30 minute defaultPostTimeout. Are there any plans to put it back to 2 hours?

Kimchelly: I thought about it and while it's not perfect to have invisible timeouts, 30 minutes is still somewhat more generous than the original 15 minutes, and we aren't in a rush to bump it. Letting a task run longer can increase spending and increased patch waiting time for existing projects who got used to relying on the 15 minute timeout, so I think it's better to pause on bumping it further until any projects who hit the timeout today have had a time window to set any block timeouts they want to suit their needs and/or optimize their post blocks to not hit the timeout.

From [the HC stats](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/result/i77eTtodCGG), it seems like ~90 tasks hit the default timeout per day, and they're clustered around a few projects. While it doesn't sound like a lot of tasks, a small number of tasks per day stuck running for an extra few hours can add up over time.

agent/agent_test.go
```diff
@@ -650,25 +644,13 @@ post:
...
-// have to include that hard-coded task information for the test to run.
 func (s *AgentSuite) setupRunTask(projYml string) {
 	p := &model.Project{}
 	_, err := model.LoadProjectInto(s.ctx, []byte(projYml), nil, "", p)
 	s.Require().NoError(err)
 	s.tc.taskConfig.Project = p
 	s.tc.project = p
 	s.mockCommunicator.GetProjectResponse = p
-
-	t := &task.Task{
```
malikchaya2, line 663: I'm surprised that the tests pass without this!

agent/background.go
```diff
@@ -157,37 +156,3 @@ func (a *Agent) startTimeoutWatcher(ctx context.Context, operationCancel context
 		}
 	}
 }
-
-// getCallbackTimeout returns the callback timeout for the task.
-func (tc *taskContext) getCallbackTimeout() time.Duration {
```
malikchaya2, line 162: nice :)

agent/task_context.go
```diff
@@ -191,14 +192,262 @@ func (tc *taskContext) getExecTimeout() time.Duration {
...
+	return &commandBlock{
+		block:       command.TeardownGroupBlock,
+		commands:    tg.TeardownGroup,
+		timeoutKind: teardownGroupTimeout,
+		getTimeout:  tc.getTeardownGroupTimeout(tg),
+		canFailTask: false,
+	}, nil
+}
+
+// getTimeout returns a command block containing the timeout handler commands.
```
malikchaya2, line 317: Do we want to specify in the comment that this is for the task group callback timeout?

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -435,33 +435,51 @@ Fields:
...
+### Pre and Post
 
-All projects can have a `pre` and `post` field which define a list of
-command to run at the start and end of every task that isn't in a task
-group. For task groups, `setup_task` and `teardown_task` will run instead of
-`pre` and `post`. These are incredibly useful as a place for results commands or
-for cleanup and setup tasks.
+All projects can have a `pre` and `post` field which define a list of commands
+to run at the start and end of every task that isn't in a task group. For task
+groups, `setup_task` and `teardown_task` will run instead of `pre` and `post`
```
malikchaya2, line 442: I think it would be helpful to add a few words about "teardown_group" in case someone is grepping to figure out the difference between post, teardown_task and teardown_group.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -471,9 +489,17 @@ timeout:
...
+Parameters:
+
+- `timeout`: commands to run when the task hits a timeout. The timeout commands
+  will only run if the timeout occurs in `pre`, `setup_group`, `setup_task`, or
+  the task commands. Furthermore, for `pre`, `setup_group`, and `setup_task`,
+  because they do not fail the task by default, they must be explicitly set to
+  fail for the timeout to trigger. For example, if a command in `pre`
+  hits the default 2 hour timeout for `pre` but `pre_error_fails_task` is not
+  set to true, then the timeout block will not trigger.
+- `callback_timeout_secs`: set a timeout for the `timeout` block. Defaults to
```
malikchaya2, line 501: I wish we could rename this, but alas.

agent/task_context.go
Kimchelly: They are already aware that post has a slow command and have a few tickets (such as [SERVER-79972](https://jira.mongodb.org/browse/SERVER-79972) and the related tickets) to make their post task run more efficiently.

docs/decisions/2023-08-23_split_callback_timeout.md
```diff
@@ -0,0 +1,45 @@
...
+  default, failing commands in post do not fail the task. Therefore, tasks can invisibly time out in post without
+  affecting the task status, meaning they could fail to do some cleanup and nobody would notice.
+* Increasing the callback timeout is bad for teardown group - Evergreen discourages people from increasing callback
+  timeout because it applies to the teardown group. This is due to a design issue where the teardown group only runs
+  once the agent has already been assigned the next task to run. The consequence is that giving the teardown group more
+  time to run holds up the next task it's supposed to run.
+
+Therefore, the callback timeout exists and can be configured but there is no reasonable setting for it - it can be
+neither increased (due to teardown group) nor reduced (due to post) without causing unwanted side effects, and in
+practice, nobody ever set this value.
```
malikchaya2, line 23: Should we also mentioned that you noticed a bunch of tasks hitting the timeout for post and exiting prematurely?

docs/decisions/2023-08-23_split_callback_timeout.md
```diff
@@ -0,0 +1,45 @@
...
+The solution was to split the callback timeout so that there's instead one user-configurable timeout per command block.
+Each block should be given a large enough default timeout that existing tasks will not unintentionally hit the timeout
+and fail the task (or silently fail to run some commands). The downside of providing a more generous default timeout is
+that users may potentially spend more time waiting for tasks than necessary in edge cases (e.g. where their post task is
+taking unreasonably long), but this is less problematic than the task invisibly timing out.
+
+The takeaway here is that choosing reasonable defaults are important since they're difficult to change later on, and
+when designing a feature, consider how its default behavior could potentially interact with other default behaviors.
+
+## More Information
```
malikchaya2, line 45: nit: remove heading?

malikchaya2 (changes requested):
great work! 
Just a couple of small comments.

agent/agent_test.go
Kimchelly: Yeah, this isn't needed because the suite's `SetupTask` already sets up the test with a task that's almost the same as this (except with a different display name).

docs/decisions/2023-08-23_split_callback_timeout.md
Kimchelly: Sure, I removed.

PR #6964, opened 2023-08-28
EVG-19951: Use enums for arch, method, and communication fields

minnakt:
EVG-19951

### Description
This PR adds enums for the `arch`, `method`, and `communication` fields.

There are some more enums I have to add for host allocator but I think I will do it in a separate PR so that this one is not overloaded.


graphql/distro_resolver.go
```diff
@@ -12,10 +12,47 @@ import (
```
minnakt, line 12: This file is not generated

rest/data/distro.go
```diff
@@ -117,7 +117,8 @@ func CreateDistro(ctx context.Context, u *user.DBUser, newDistroId string) error
 		Id:   newDistroId,
 		Arch: evergreen.ArchLinuxAmd64,
 		BootstrapSettings: distro.BootstrapSettings{
-			Method: distro.BootstrapMethodNone,
+			Method:        distro.BootstrapMethodLegacySSH,
+			Communication: distro.CommunicationMethodLegacySSH,
```
minnakt, line 121: I changed this because the code says that `BootstrapMethodNone` is for [internal use only](https://github.com/evergreen-ci/evergreen/blob/4055cd62aeebf77af84749e9eedd58a2547c71be/model/distro/distro.go#L279), and it's not a selectable option on the legacy UI page

sophstad (approved):

graphql/schema/types/distro.graphql
```diff
@@ -42,6 +42,31 @@ enum DispatcherVersion {
   REVISED_WITH_DEPENDENCIES
 }
 
+enum Arch {
```
sophstad, line 45: Also noticing that a few of these are no longer in use! Here's what's being used:
```
> db.distro.distinct("arch")
> [
  'darwin_amd64',
  'darwin_arm64',
  'linux_amd64',
  'linux_arm64',
  'linux_ppc64le',
  'linux_s390x',
  'windows_amd64'
]
```
So neither 32-bit architecture is in use (Windows or Linux).

Brian McCarthy is ooo this week but I'm going to follow up when he's back to see if we can remove these. If you'd like to omit them in this PR that's probably okay, and we can add them in if necessary.

rest/data/distro.go
Kimchelly: 👍 Sounds good, and yeah, this one isn't an option that users or admins can directly use.

graphql/schema/types/distro.graphql
```diff
@@ -42,6 +42,31 @@ enum DispatcherVersion {
...
+  RPC
+}
+
+enum BootstrapMethod {
+  LEGACY_SSH
+  SSH
+  USER_DATA
+}
+
+
```
Kimchelly, line 69: nit: did you intend to put 3 newlines in between this + the next section?

minnakt: fixed!

Kimchelly (approved):

PR #6966, opened 2023-08-29
EVG-19951: Use enums for host allocator fields

minnakt:
EVG-19951

### Description
This just adds more enums for the host settings page, in a separate PR since these fields are thematically in the same group.

I'm pretty sure that the `DEFAULT` values means that the values will be [inherited](https://github.com/evergreen-ci/evergreen/blob/2c322c18832281119433281f9026724975e9a74a/model/distro/distro.go#L665C1-L673) from Evergreen settings (see the "Scheduler" section in the admin settings).

### DIFF
--- a/graphql/distro_resolver.go
+++ b/graphql/distro_resolver.go
@@ -180,20 +180,20 @@ func (r *hostAllocatorSettingsResolver) FeedbackRule(ctx context.Context, obj *m
 }
 
 // HostsOverallocatedRule is the resolver for the hostsOverallocatedRule field.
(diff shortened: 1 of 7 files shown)


graphql/distro_resolver.go
```diff
@@ -98,6 +98,74 @@ func (r *finderSettingsResolver) Version(ctx context.Context, obj *model.APIFind
 	}
```
minnakt, line 98: This file is not generated

Kimchelly (approved):
The explicit "default" enums are kind of odd because the defaults are already [determined within the scheduler](https://github.com/evergreen-ci/evergreen/blob/bed59fe469fe55b4a532afced50f2ed8b15acd67/scheduler/host_allocator.go#L32). However, given that this is how the distro page already is, it seems alright to duplicate that behavior.

sophstad (approved):

PR #6969, opened 2023-08-29
EVG-20622: support allowed requesters list

Kimchelly:
EVG-20622

### Description
This adds a new project YAML configuration, which lets a user enumerate the particular requesters (e.g. GitHub PR, git tag version, periodic build) for which they want to run a task. This is a generalization of the filters currently provided by `patchable`, `patch_only`, `allow_for_git_tag` and `git_tag_only` (e.g. `git_tag_only: true` is equivalent to `allowed_requesters: ["github_tag"]`). We have no plans to actually deprecate the boolean requester options though; I chose to instead add a warning that mixing requester booleans with `allowed_requesters` is incompatible.

There will be a follow-up PR to update Shrub for `generate.tasks` after this is merged.

* Add `allowed_requesters` configuration.
* Add some helpers to translate the user-facing requesters that we expose via expansions and `allowed_requesters`.

### Testing
* Added unit tests.
* Tested in staging by submitting patches with `allowed_requesters` and checking `evergreen validate`.

### Documentation
Updated docs.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -653,6 +653,86 @@ var (
 	}
 )
 
+// UserRequester represents the allowed user-facing requester types.
+type UserRequester string
+
+const (
+	// User-facing requester types. These are equivalent in meaning to the above
+	// requesters, but are more user-friendly. These should only be used for
+	// user-facing functionality such as YAML configuration and expansions and
+	// should be translated into the true internal requester types so they're
+	// actually usable.
+	PatchVersionUserRequester       UserRequester = "patch"
+	GithubPRUserRequester           UserRequester = "github_pr"
+	GitTagUserRequester             UserRequester = "github_tag"
+	RepotrackerVersionUserRequester UserRequester = "commit"
+	TriggerUserRequester            UserRequester = "trigger"
+	MergeTestUserRequester          UserRequester = "commit_queue"
+	AdHocUserRequester              UserRequester = "ad_hoc"
+	GithubMergeUserRequester        UserRequester = "github_merge_queue"
+)
+
+var AllUserRequesterTypes = []UserRequester{
+	PatchVersionUserRequester,
+	GithubPRUserRequester,
+	GitTagUserRequester,
+	RepotrackerVersionUserRequester,
+	TriggerUserRequester,
+	MergeTestUserRequester,
+	AdHocUserRequester,
+	GithubMergeUserRequester,
+}
+
+// InternalRequesterToUserRequester translates an internal requester type to a
+// user-facing requester type.
+func InternalRequesterToUserRequester(requester string) UserRequester {
+	switch requester {
+	case PatchVersionRequester:
+		return PatchVersionUserRequester
+	case GithubPRRequester:
+		return GithubPRUserRequester
+	case GitTagRequester:
+		return GitTagUserRequester
+	case RepotrackerVersionRequester:
+		return RepotrackerVersionUserRequester
+	case TriggerRequester:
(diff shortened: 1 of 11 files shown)


globals.go
```diff
@@ -653,6 +653,86 @@ var (
...
+// UserRequester represents the allowed user-facing requester types.
+type UserRequester string
+
+const (
+	// User-facing requester types. These are equivalent in meaning to the above
+	// requesters, but are more user-friendly. These should only be used for
+	// user-facing functionality such as YAML configuration and expansions and
+	// should be translated into the true internal requester types so they're
+	// actually usable.
+	PatchVersionUserRequester       UserRequester = "patch"
```
ablack12, line 665: Welp I wanted to panic about having these special requester names but it looks like I approved them initially for expansions so I blame myself 🙃

model/project.go
```diff
@@ -282,25 +289,49 @@ func (bvt *BuildVariantTaskUnit) UnmarshalYAML(unmarshal func(interface{}) error
...
+		allowed, _ := utility.StringSliceSymmetricDifference(bvt.AllowedRequesters, evergreen.PatchRequesters)
+		return len(allowed) == 0
+	}
+
 	return utility.FromBoolPtr(bvt.PatchOnly)
 }
 
 func (bvt *BuildVariantTaskUnit) SkipOnGitTagBuild() bool {
+	if len(bvt.AllowedRequesters) != 0 {
+		allowed := utility.StringSliceIntersection(bvt.AllowedRequesters, []string{evergreen.GitTagRequester})
```
ablack12, line 322: nit: feels cleaner to just use stringSliceContains for git tags

validator/project_validator.go
```diff
@@ -1374,33 +1374,73 @@ func validateProjectTaskIdsAndTags(project *model.Project) ValidationErrors {
 func checkTaskRuns(project *model.Project) ValidationErrors {
 	var errs ValidationErrors
 	for _, bvtu := range project.FindAllBuildVariantTasks() {
+		if len(bvtu.AllowedRequesters) != 0 {
+			for _, requester := range bvtu.AllowedRequesters {
+				if !utility.StringSliceContains(evergreen.AllRequesterTypes, requester) {
+					errs = append(errs, ValidationError{
+						Level: Error,
```
ablack12, line 1381: In the work to make project validation more consistent, validators were split by warning/error (the names here should maybe include the word warning to make that clearer, but checkTaskRuns is in the projectWarningValidators). We don't check warning validators everywhere we check error ones, I believe. This error may best live in [validateBVFields](https://github.com/evergreen-ci/evergreen/blob/96b3f970f8e436d639cc8ee19456d225dda3384d/validator/project_validator.go#L661)

ablack12 (changes requested):

globals.go
Kimchelly: Yeah, I was wondering if I could just use the actual requester values, but the existing docs already described the requesters like this. 🤷

model/project.go
Kimchelly: Done.

validator/project_validator.go
Kimchelly: Hmm, I just removed this validation since it's a duplicate. The requesters already get checked for valid values when translating the parser project to the project.

Kimchelly:
evergreen refresh

Kimchelly:
evergreen retry

ablack12 (approved):
Excited about this!

PR #6970, opened 2023-08-30
EVG-20700: bump timeout for flaky ps tests

Kimchelly:
EVG-20700

### Description
The test output shows that some machines are occasionally slow at running `ps` to get the running process, so it's hitting the 5-second context timeout. Just bumping the timeout and checking for no context timeout seems sufficient to stop the test from flaking.

* Increase flaky test timeout.
* Refactor agent process-related tests a little bit.

### Testing
I ran it locally, but if it's really fixed, we will no longer see this test flake on the waterfall.

### DIFF
--- a/agent/util/subtree_unix_test.go
+++ b/agent/util/subtree_unix_test.go
@@ -72,9 +72,8 @@ func TestGetPIDsToKill(t *testing.T) {
 }
 
 func TestKillSpawnedProcs(t *testing.T) {
-	ctx := context.Background()
-	for testName, test := range map[string]func(*testing.T){
-		"expired context": func(t *testing.T) {
+	for testName, test := range map[string]func(ctx context.Context, t *testing.T){
+		"ErrorsWithContextTimeout": func(ctx context.Context, t *testing.T) {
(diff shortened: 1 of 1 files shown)


agent/util/subtree_unix_test.go
```diff
@@ -72,58 +72,62 @@ func TestGetPIDsToKill(t *testing.T) {
...
-
-	for testName, test := range map[string]func(*testing.T){
-		"non-existent process": func(t *testing.T) {
-			waitCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
-			defer cancel()
-			pids, err := waitForExit(waitCtx, []int{1234567890})
+	for testName, test := range map[string]func(ctx context.Context, t *testing.T){
+		"DoesNotReturnNonexistentProcess": func(ctx context.Context, t *testing.T) {
+			pids, err := waitForExit(ctx, []int{1234567890})
+			assert.NoError(t, ctx.Err())
```
ablack12, line 111: if the context is cancelled, then this will still be flaky right? do we want it to be like, if there's no context error, then check the below conditions? (Or do we expect this to now only happen consistently?)

ablack12 (approved):

PR #6971, opened 2023-08-30
EVG-20433: move user-defined end task data to task config

Kimchelly:
EVG-20433

### Description
Users can define explicit task end statuses by sending a request to the agent's `/task_status` endpoint. I moved the data from the `Agent` into the `taskConfig`, and fixed some small bugs in its behavior along the way.

* Move the user-defined end task data from the agent into the task config.
* Make the end task data safe against concurrent reads/writes (e.g. a background request to `/task_status` writes the task status data while the agent is reading the end task data).
* Fix a small issue where the user-defined task description is not set if the task succeeds. Setting a task description on success seems rather niche, but it's not invalid for users to set one for success either.

### Testing
* Added unit tests.
* Submitted patches in staging to explicitly set end task response.

### Documentation
N/A


agent/agent.go
```diff
@@ -966,24 +976,25 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 func (a *Agent) endTaskResponse(ctx context.Context, tc *taskContext, status string, message string) *apimodels.TaskEndDetail {
```
ZackarySantana, line 976: I'm not sure what our convention is but don't use the agent object anymore- so we could remove it off of the agent struct.

ZackarySantana (approved):
LGTM

agent/agent.go
```diff
@@ -453,6 +459,10 @@ func (a *Agent) setupTask(agentCtx, setupCtx context.Context, initialTC *taskCon
 
 	a.comm.UpdateLastMessageTime()
 
+	a.setEndTaskRespMutex.Lock()
```
malikchaya2, line 462: do we need this lock since the function itself already has one?

malikchaya2 (changes requested):

malikchaya2 (approved):

PR #6974, opened 2023-08-31
EVG-20634: refactor common block timeout and logs

Kimchelly:
EVG-20634

### Description
* Refactor common logic like setting up the block timeout watcher (e.g. for pre timeout, post timeout, etc.) into `runCommandsInBlock`.
* Make logs at beginning and end of a task block log consistently. For example, when a pre command errored and `pre_error_fails_task` was set, it wouldn't log how long the command ran (`Finished 'shell.exec' (step 1 of 1) in block 'pre' in 10s`).
* Slightly change the names of the block name that logs. For example, when running `setup_task` for a task group task, it used to log `Running pre-task commands`, but now it logs `Running setup-task commands` to reflect the actual block of commands that it's running.

### Testing
* Added more log checks to the agent suite.
* Tested in staging to verify that the output looked reasonable ([example for task group task](https://evergreen-staging.corp.mongodb.com/task_log_raw/evg_ubuntu2204_test_util_patch_71c599fbee3717ab5c8bd9492659762aded9a233_64f0a370b237363adda14d56_23_08_31_14_28_01/1?type=T)).

### Documentation
N/A


agent/agent.go
```diff
@@ -873,31 +895,6 @@ func (a *Agent) handleTimeoutAndOOM(ctx context.Context, tc *taskContext, status
...
-		defer timeoutCancel()
-		timeoutOpts := timeoutWatcherOptions{
-			tc:         tc,
-			kind:       timeout.timeoutKind,
-			getTimeout: timeout.getTimeout,
-		}
-		go a.startTimeoutWatcher(timeoutCtx, timeoutCancel, timeoutOpts)
-
-		err := a.runCommandsInBlock(timeoutCtx, tc, timeout.commands.List(), runCommandsOptions{block: timeout.block})
-		tc.logger.Task().Error(errors.Wrap(err, "Running timeout commands failed"))
```
hadjri, line 896: Why was this line removed and the error is ignored now? Same question for `runEndTaskSync` etc

Kimchelly: Hm, so [the command failure already gets logged further down](https://github.com/evergreen-ci/evergreen/blob/a2958b866cf95565441d94b807834220c8a45449/agent/command.go#L294) in the command logic, so the user will still see that a command failed. From what I can see, the `"Running <block name> commands failed"` is specifically used in cases where the command both failed _and_ that failure will cause the whole task to fail (which will also stop regular command execution and skip to `post`).

In the case of `timeout`, the agent is already going to mark the task as failed since it timed out, so `timeout` can't change an already-failed task. For task sync, they're considered "optional" commands because they don't influence the task status.

hadjri: Got it, that makes sense, thanks

hadjri (approved):

Kimchelly (commented):
Sorry, I did a bit of flip-flopping and decided to actually log the error even if the block can't fail the task. The reason being that if the block fails for some reason other than the command failing (e.g. somehow the command doesn't render, so it never even got to run), then we should still log the error for us/the user to see. I also added some comments to explain why ignoring the error is okay in the cases where it is ignored (i.e. they're situations where you just log the error and otherwise ignore it).

hadjri (approved):

PR #6975, opened 2023-08-31
EVG-20654: deprecate duplicate GetWorkingDirectory

Kimchelly:
EVG-20654

### Description
The agent commands that need to work in a directory can use either `GetWorkingDirectory` or `getJoinedWithWorkDir`, both of which accomplish the general goal of resolving a working directory. The major difference is that `getJoinedWithWorkDir` allows absolute paths outside of the task working directory, whereas `GetWorkingDirectory` will force all paths to be prefixed by the task working directory, even if they're specified like absolute paths. For example, if you set a working directory of `/home/mci/directory`, `getJoinedWithWorkDir` would keep it as-is while `GetWorkingDirectory` would convert this into `${workdir}/home/mci/directory`.

Since `GetWorkingDirectory` behaves differently for absolute paths and is also used in `subprocess.exec` and `shell.exec`, it would be difficult to get rid of `GetWorkingDirectory` entirely because of the widespread usage of those commands. I instead opted to just mark it as legacy so we don't try to extend usage of it further.

### Testing
Existing tests pass.

### Documentation
N/A

### DIFF
--- a/agent/command/util.go
+++ b/agent/command/util.go
@@ -3,6 +3,7 @@ package command
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/evergreen-ci/evergreen/agent/internal"
 	"github.com/evergreen-ci/evergreen/agent/internal/client"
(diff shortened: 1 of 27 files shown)


ablack12 (changes requested):
Would it make sense to rename getJoinedWithWorkDir to GetWorkingDirectory so it's clear it's the actual one? And maybe move this to also be a private function in util.go so it's more consistent and they can be easily found? Otherwise this feels kind of halfway done.

Kimchelly:
Sounds good, I moved `GetWorkingDirectory` to the command package and made it clear that one is the new one and the other is the legacy one.

agent/command/util.go
```diff
@@ -61,19 +62,47 @@ func expandModulePrefix(conf *internal.TaskConfig, module, prefix string, logger
...
 	}
 	return filepath.Join(conf.WorkDir, path)
 }
 
+// getWorkingDirectoryLegacy is a legacy function to get the working directory
+// for a path, enforce that the path is always prefixed with the task working
+// directory, and check that the directory exists. This is a legacy function, so
+// should not be used anymore. Commands that need to get the working directory
+// should instead use getWorkingDirectory.
+func getWorkingDirectoryLegacy(tc *internal.TaskConfig, dir string) (string, error) {
```
ablack12, line 83: LGTM -- just to check, would we still want to use this function if we wanted to do some handling around directories (in which case we might not want to actually deprecate but call forDirectories or something), or is that unnecessary bc we'd get errors when we actually tried to use the path?

ablack12 (approved):

agent/command/util.go
Kimchelly: It kind of depends on the command, but generally, if the directory doesn't already exist but it has to exist, both functions will error, this one will just error out a little earlier than the other one. The reason this one is slightly less preferable is because 1. many commands do stuff to the working directory that doesn't mandate that the directory must exist (e.g. s3.put with `optional: true`) and 2. if the directory needs to exist, the newer one will still have some error about the directory not existing, just further down the line when it needs the directory to exist.

PR #6978, opened 2023-08-31
EVG-20052 Add event logs when a build/version's priority changes 

ZackarySantana:
EVG-20052

### Description
We are not logging events when versions (or builds) are changed in priority. This adds the event logs when either are changed in priority.

I added it to the build priority change call as well although I think that is only used via a REST endpoint and the old UI, but it was an identical change and might help us debug in the future niche cases.

### Testing
Existing tests already test these functions. Tested (before and after) setting priority of a version and looking at event log of individual tasks (locally and in staging)

Example of setting priority for a version and it showing up in the event logs for a task:
https://github.com/evergreen-ci/evergreen/assets/64446617/33813def-9a0d-4f09-9d00-7d02f3ac240d

### DIFF
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -105,8 +105,8 @@ func LogJiraIssueCreated(taskId string, execution int, jiraIssue string) {
 	logTaskEvent(taskId, TaskJiraAlertCreated, TaskEventData{Execution: execution, JiraIssue: jiraIssue})
 }
 
-func LogTaskPriority(taskId string, execution int, user string, priority int64) {
-	logTaskEvent(taskId, TaskPriorityChanged, TaskEventData{Execution: execution, UserId: user, Priority: priority})
+func LogTaskPriority(taskId string, execution int, userId string, priority int64) {
+	logTaskEvent(taskId, TaskPriorityChanged, TaskEventData{Execution: execution, UserId: userId, Priority: priority})
 }
 
 func LogTaskCreated(taskId string, execution int) {
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -198,9 +198,9 @@ func LogManyTaskAbortRequests(taskIds []string, userId string) {
 		TaskEventData{UserId: userId})
 }
 
-func LogManyTaskPriority(taskIds []string, execution int, userId string, priority int64) {
+func LogManyTaskPriority(taskIds []string, userId string, priority int64) {
 	logManyTaskEvents(taskIds, TaskPriorityChanged,
-		TaskEventData{Execution: execution, UserId: userId, Priority: priority})
+		TaskEventData{UserId: userId, Priority: priority})
 }
 
 func LogTaskContainerAllocated(taskId string, execution int, containerAllocatedTime time.Time) {
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -198,6 +198,11 @@ func LogManyTaskAbortRequests(taskIds []string, userId string) {
 		TaskEventData{UserId: userId})
 }
 
+func LogManyTaskPriority(taskIds []string, execution int, userId string, priority int64) {
+	logManyTaskEvents(taskIds, TaskPriorityChanged,
+		TaskEventData{Execution: execution, UserId: userId, Priority: priority})
+}
+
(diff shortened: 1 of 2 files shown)


model/event/task_event.go
```diff
@@ -105,8 +105,8 @@ func LogJiraIssueCreated(taskId string, execution int, jiraIssue string) {
 	logTaskEvent(taskId, TaskJiraAlertCreated, TaskEventData{Execution: execution, JiraIssue: jiraIssue})
 }
 
-func LogTaskPriority(taskId string, execution int, user string, priority int64) {
-	logTaskEvent(taskId, TaskPriorityChanged, TaskEventData{Execution: execution, UserId: user, Priority: priority})
+func LogTaskPriority(taskId string, execution int, userId string, priority int64) {
```
ZackarySantana, line 108: Renamed to keep consistency with other methods in the same file.

hadjri:
Note: PRs need a descriptive title along with the JIRA ticket number ([example](https://github.com/evergreen-ci/evergreen/pull/6981))

model/lifecycle.go
```diff
@@ -262,23 +262,27 @@ func SetTaskPriority(ctx context.Context, t task.Task, priority int64, caller st
...
+	query := bson.M{task.BuildIdKey: buildId}
+	_, err := task.UpdateAll(query,
 		bson.M{"$set": bson.M{task.PriorityKey: priority}},
 	)
 	if err != nil {
 		return errors.Wrapf(err, "setting priority for build '%s'", buildId)
 	}
 
-	// negative priority - these tasks should never run, so unschedule now
+	tasks, err := task.FindAll(db.Query(query))
```
hadjri, line 273: Since this logic is duplicated in two places can we move it into a helper function?

model/lifecycle.go
```diff
@@ -294,18 +298,21 @@ func SetVersionsPriority(ctx context.Context, versionIds []string, priority int6
...
+	if err != nil {
+		return errors.Wrap(err, "getting tasks for versions")
+	}
+	var taskIds []string
+	for _, t := range tasks {
+		taskIds = append(taskIds, t.Id)
+	}
+	event.LogManyTaskPriority(taskIds, caller, priority)
+
+	// Deactivate tasks with negative priority
```
hadjri, line 311: Can we change this comment (and the other one that is identical) back to something closer to the original comment, like `Tasks with negative priority should never run, so we unschedule them`

hadjri (changes requested):

ZackarySantana:
I attached a working video of the event logs getting updated and added a title!

hadjri:
Nit: The PR title should be describing what was actually done in the code, instead of copying the JIRA ticket tile. Something good here would be like "Add priority event logs when setting version priority"

ablack12:
FYI I think [the new hire introduction section](https://docs.google.com/document/d/1LJHPjLb_bUZxv-l_9Y61kitzAg7u-13I5DjOhZ-deBA/edit) on how to choose a PR title is outdated (the ticket title is not always a good PR title, per Malik's advice); I'll update

ZackarySantana:
That makes sense! I updated it to align with the code changes rather than the JIRA ticket

model/lifecycle.go
```diff
@@ -262,50 +263,38 @@ func SetTaskPriority(ctx context.Context, t task.Task, priority int64, caller st
...
+	return errors.Wrap(setTasksPriority(ctx, query, priority, caller), "setting priority for build")
 }
 
 // SetVersionsPriority updates the priority field of all tasks and child tasks associated with the given version ids.
 func SetVersionsPriority(ctx context.Context, versionIds []string, priority int64, caller string) error {
 	query := task.ByVersionsWithChildTasks(versionIds)
+	return errors.Wrap(setTasksPriority(ctx, query, priority, caller), "setting priority for versions")
+}
+
+func setTasksPriority(ctx context.Context, query primitive.M, priority int64, caller string) error {
```
hadjri, line 276: We instead use `bson.M` rather than `primitive`, for consistency

hadjri (changes requested):

hadjri (approved):

PR #6980, opened 2023-08-31
EVG-19099 update module git.get_project docs

ablack12:
EVG-19099 
### Description

Added some docs to try and explain the behavior of module revision/ref/branch in the git.get_project docs. This isn't perfect -- I don't totally understand why ref isn't used for patches (although it's user confirmed and already mentioned in documentation) but I believe it has to do with the manifest, which is checked before ref -- but I think addresses the problem that spawned this ticket.

### DIFF
--- a/docs/Project-Configuration/Project-Commands.md
+++ b/docs/Project-Configuration/Project-Commands.md
@@ -352,8 +352,8 @@ by a patch submission.
 
 Parameters:
 
-- `dir`: the directory to clone into
-- `revisions`: For commit builds, each module should be passed as
+-   `dir`: the directory to clone into
+-   `revisions`: For commit builds, each module should be passed as
     `<module_name> : ${<module_name>_rev}` (these are loaded from the [manifest](../API/REST-V2-Usage.md#manifest) 
     at the beginning of the command). 
     For patch builds, the hash
@@ -361,17 +361,17 @@ Parameters:
     means that for patch builds, editing the
     ["modules"](Project-Configuration-Files.md#modules)
     section of the project config will not change the checked out hash.
-- `token`: Use a token to clone instead of the ssh key on the host.
+-   `token`: Use a token to clone instead of the ssh key on the host.
     Since this is a secret, it should be provided as a project
     expansion. For example, you could provide an expansion called
     "github_token" and then set this field to \${github_token}.
     Evergreen will populate the expansion when it parses the project
     yaml.
-- `clone_depth`: Clone with `git clone --depth <clone_depth>`. For
+-   `clone_depth`: Clone with `git clone --depth <clone_depth>`. For
     patch builds, Evergreen will `git fetch --unshallow` if the base
-    commit is older than `<clone_depth>` commits.
-- `shallow_clone`: Sets `clone_depth` to 100.
-- `recurse_submodules`: automatically initialize and update each
+    commit is older than `<clone_depth>` commits. `clone_depth` takes precedence over `shallow_clone`.
+-   `shallow_clone`: Sets `clone_depth` to 100, if not already set.
+-   `recurse_submodules`: automatically initialize and update each
     submodule in the repository, including any nested submodules.
 
 The parameters for each module are:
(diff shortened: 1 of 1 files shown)


docs/Project-Configuration/Project-Commands.md
```diff
@@ -352,24 +352,26 @@ by a patch submission.
 
 Parameters:
 
--   `dir`: the directory to clone into
--   `revisions`: For commit builds, each module should be passed as
-    `<module_name> : ${<module_name>_rev}`. For patch builds, the hash
+- `dir`: the directory to clone into
+- `revisions`: For commit builds, each module should be passed as
```
Kimchelly, line 356: nit: do we want to keep the follow-on lines for a bullet point at the same indentation level? I slightly preferred when the markdown bullet stayed at the same indentation level. The other markdown lists in the file also keep the indentation the same.

Maybe one day we should have a markdown linter or something 🤔

docs/Project-Configuration/Project-Commands.md
```diff
@@ -352,24 +352,26 @@ by a patch submission.
...
     "github_token" and then set this field to \${github_token}.
     Evergreen will populate the expansion when it parses the project
     yaml.
--   `clone_depth`: Clone with `git clone --depth <clone_depth>`. For
+- `clone_depth`: Clone with `git clone --depth <clone_depth>`. For
     patch builds, Evergreen will `git fetch --unshallow` if the base
     commit is older than `<clone_depth>` commits.
--   `shallow_clone`: Sets `clone_depth` to 100.
--   `recurse_submodules`: automatically initialize and update each
+- `shallow_clone`: Sets `clone_depth` to 100.
```
Kimchelly, line 373: I know you didn't change this doc, but while we're at it, should we mention that implies you can't set both `clone_depth` and `shallow_clone` simultaneously?

docs/Project-Configuration/Project-Commands.md
```diff
@@ -379,10 +381,18 @@ The parameters for each module are:
 -   `prefix`: the subdirectory to clone the repository in. It will be
     the repository name as a top-level directory in `dir` if omitted
 -   `ref`: must be a commit hash, takes precedence over the `branch`
-    parameter if both specified
+    parameter if both specified (for commits)
 -   `branch`: must be the name of branch, commit hashes _are not
     accepted_.
 
+
```
Kimchelly, line 388: nit: remove extra newline.

docs/Project-Configuration/Project-Commands.md
```diff
@@ -379,10 +381,18 @@ The parameters for each module are:
...
+More specifically, module hash priority is as follows:
+* For commit queue patches, we always use the module branch name, to ensure accurate testing.
+* For other patches, we initially default to the githash in set-module, if specified.
+* For both commits and patches, we next default to the <module_name> set in revisions for the command.
```
Kimchelly, line 392: `<module_name>` has to be wrapped in tildes (or you can use `&lt;` + `&gt;`. Otherwise, markdown will interpret this as HTML tags.

docs/Project-Configuration/Project-Commands.md
```diff
@@ -379,10 +381,18 @@ The parameters for each module are:
...
     the repository name as a top-level directory in `dir` if omitted
 -   `ref`: must be a commit hash, takes precedence over the `branch`
-    parameter if both specified
+    parameter if both specified (for commits)
 -   `branch`: must be the name of branch, commit hashes _are not
     accepted_.
 
+
+More specifically, module hash priority is as follows:
+* For commit queue patches, we always use the module branch name, to ensure accurate testing.
```
Kimchelly, line 390: nit: Instead of saying a nebulous "we", should we refer to it as "Evergreen" instead? I think of the Evergreen CI system as a single entity at least.

Kimchelly (changes requested):
The hash selection priority seems a little chaotic, but I won't try to suggest we change it at all. 😨

docs/Project-Configuration/Project-Commands.md
Kimchelly, line 390: I'm not that familiar with how modules interact with the commit queue. Does this comment + the above docs about `branch` mean that if your project uses teh commit queue and you have modules, you must explicitly set `branch` in the YAML, or does it get auto-populated (by set-module or something)?

docs/Project-Configuration/Project-Commands.md
ablack12: Ah yeah I'll have to adjust my editor to not do this

docs/Project-Configuration/Project-Commands.md
ablack12: Sure thing

docs/Project-Configuration/Project-Commands.md
ablack12: Haha fair -- I can rework

ablack12: It looks like we expect branch: https://github.com/evergreen-ci/evergreen/blob/96b3f970f8e436d639cc8ee19456d225dda3384d/validator/project_validator.go#L984

I'm not sure if the new commit queue is set up to handle modules right. I can file this away as a question for Brian when he's back.

docs/Project-Configuration/Project-Commands.md
ablack12: Fixed; that reads better anyway

ablack12:
It does seem chaotic assuming I code read correctly 😅 simplifying modules seems like a worthy next step. For example, I only saw one enabled config that actually uses "revision" for modules so idk if we should still be supporting this

docs/Project-Configuration/Project-Commands.md
Kimchelly: Sounds good, thanks!

Kimchelly (approved):
LGTM, and simplifying modules also SGTM!

PR #6981, opened 2023-08-31
EVG-19285 avoid unneeded error for fixing stale display task

ablack12:
EVG-19285

### Description
For Archive, we check if archiving has a duplicate key error, I believe to avoid races. This doesn't work for ArchiveMany bc transactions don't respect error handling, they just abort if there's an issue (GODRIVER-1721).

We only actually are hitting this error for ArchiveMany for display tasks that have multiple execution tasks fail heartbeat (i.e. all of them are racing to archive the display task). 

I think we should just avoid erroring in the case that the display task actually was restarted, since there's no error here. Open question I have is whether we should log something special in this case.

### Testing
Hard to test since it requires a specific race; going to reason over and keep an eye out for this error for a week or so.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -2429,12 +2429,6 @@ func ArchiveMany(tasks []Task) error {
 		}
 	}
 
-	grip.DebugWhen(len(utility.UniqueStrings(allTaskIds)) != len(allTaskIds), message.Fields{
-		"ticket":           "EVG-17261",
-		"message":          "archiving same task multiple times",
-		"tasks_to_archive": allTaskIds,
-	})
-
 	return archiveAll(allTaskIds, execTaskIds, toUpdateExecTaskIds, archivedTasks)
 }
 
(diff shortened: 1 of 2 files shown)


model/task/task.go
```diff
@@ -2429,12 +2429,6 @@ func ArchiveMany(tasks []Task) error {
 		}
 	}
 
-	grip.DebugWhen(len(utility.UniqueStrings(allTaskIds)) != len(allTaskIds), message.Fields{
```
ablack12, line 2432: looks like we closed this ticket and this line isn't logged

hadjri (approved):
I think we could prevent the race errors from happening at all by implementing a new task field, something similar to [CanReset](https://gist.github.com/hadjri/e91e31ad7fe0daa22c945018403726de#file-task-go-L166), something like `HasFailedExecTask`, which we could include in the query for `archiveAll` and tell it to skip the archive attempt if it's true for the given execution number. That being said, I can't think of any real harm these race failures are doing so it may unnecessary to do that work.

We could add a debug grip statement that checks if the error includes "transaction aborted" in its string so we can monitor short term how prevalent this issue is, but I don't feel too strongly about whether that's necessary

ablack12:
Yeah I think it's a bit overkill; we also do want the first one to succeed, and yeah this doesn't cause any behavioral issues for users since these transactions do correctly fail, it just incorrectly contributes to our error count.

PR #6982, opened 2023-09-01
EVG-20289 Simplify down taskContext/TaskConfig usage where possible

malikchaya2:
EVG-20289

### Description
This removes fields that were duplicated between task config and task context. It also fetches the task group when making the task config. that way it doesn't need to be fetched. 

### Testing
Existing tests


malikchaya2:
evergreen retry

PR #6984, opened 2023-09-01
EVG-19946: Expose container pool settings via SpruceConfig

minnakt:
EVG-19946

### Description
This PR exposes the container pools field that is set through the admin settings page. Note that [SpruceConfig is mapped to APIAdminSettings](https://github.com/evergreen-ci/evergreen/blob/5df5c07ddd9bd75953acdefd3c9b6acaabeee319/gqlgen.yml#L304-L305) which is why I put this field here. 

Also adds some testdata for Spruce.

### Testing
- Added GraphQL tests


graphql/config_resolver.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+
+	"github.com/evergreen-ci/evergreen/rest/model"
+)
+
+// Port is the resolver for the port field.
+func (r *containerPoolResolver) Port(ctx context.Context, obj *model.APIContainerPool) (int, error) {
+	return int(obj.Port), nil
+}
```
minnakt, line 16: This file is not generated. I wasn't really sure how to handle `uint16` through GraphQL, but I casted it to an `int` because I figured that given the max possible value of `uint16`, it should be able to fit within the `int` type...

rest/model/admin.go
```diff
@@ -182,6 +182,12 @@ func (as *APIAdminSettings) BuildFromService(h interface{}) error {
 			return errors.Wrap(err, "converting slack config to API model")
 		}
 		as.Slack = &slackConfig
+		containerPoolsConfig := APIContainerPoolsConfig{}
+		err = containerPoolsConfig.BuildFromService(v.ContainerPools)
+		if err != nil {
```
sophstad, line 187: [nit] Combine into one line

sophstad (approved):

graphql/config_resolver.go
Kimchelly: `uint16` isn't significant in this case, I believe it was just originally used to prevent inputting negative numbers. The conversion LGTM!

graphql/tests/query/spruceConfig/queries/container_pools.graphql
```diff
@@ -0,0 +1,12 @@
+query {
+  spruceConfig {
+    containerPools {
+      pools {
+        id
+        distro
+        maxContainers
+        port
```
Kimchelly, line 8: Is this supposed to include max images? It doesn't matter that much given that we don't use any value but the default, but I was just wondering.

Kimchelly (approved):

ablack12 (commented):
I'm sorry I forgot to hit submit on this comment I wrote forever ago but just a question about the file name

graphql/tests/query/spruceConfig/queries/container_pools.graphql
minnakt: I think I might just leave it out for now since `APIContainerPools` doesn't have it! 🤔

graphql/config_resolver.go
```diff
@@ -0,0 +1,21 @@
...
+// This file will be automatically regenerated based on the schema, any resolver implementations
+// will be copied through when generating and any unknown code will be moved to the end.
+// Code generated by github.com/99designs/gqlgen version v0.17.36
+
+import (
+	"context"
+
+	"github.com/evergreen-ci/evergreen/rest/model"
+)
+
```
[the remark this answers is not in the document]
minnakt: it actually gets generated from the GraphQL filename; so because the value is declared in `config.graphql`, it gets put in `config_resolver.go`. Do you want me to try renaming GraphQL file to something else? 🤔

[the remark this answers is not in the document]
ablack12: Ah if this is coming from the admin settings then it would make sense; is this just consumed by the distro page?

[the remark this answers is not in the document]
minnakt: yep thats right, when you have the "Docker" provider it lets you pick from the container pools which are defined in the admin settings:

Screenshot 2023-09-06 at 2 41 46 PM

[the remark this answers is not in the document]
ablack12: Makes sense, thanks!

ablack12 (approved):

graphql/tests/query/spruceConfig/queries/container_pools.graphql
Kimchelly: SGTM! This setting will hopefully go away one day anyways 🙏

PR #6985, opened 2023-09-01
EVG-20289 Simplify down taskContext/TaskConfig usage where possible

malikchaya2:
EVG-20289

### Description
This removes fields that were duplicated between task config and task context. It also fetches the task group when making the task config. that way it doesn't need to be fetched. 

### Testing
Existing tests


agent/logging_test.go
```diff
@@ -130,30 +128,33 @@ func TestStartLogging(t *testing.T) {
...
-		},
 	}
 
 	ctx := context.Background()
-	assert.NoError(agt.fetchProjectConfig(ctx, tc))
-	require.NotNil(t, tc.project)
-	assert.EqualValues(model.EvergreenLogSender, tc.project.Loggers.Agent[0].Type)
-	assert.EqualValues(model.SplunkLogSender, tc.project.Loggers.System[0].Type)
-	assert.EqualValues(model.FileLogSender, tc.project.Loggers.Task[0].Type)
+	task, project, expansion, _, err := agt.fetchProjectConfig(ctx, tc)
```
Kimchelly, line 134: `makeTaskConfig` below this already calls `fetchProjectConfig` when the project is not initialized in the `taskContext`. Do we still need to call this directly here?

agent/task_context.go
```diff
@@ -127,16 +144,18 @@ func (tc *taskContext) getTimeoutType() timeoutType {
 
 // makeTaskConfig fetches task configuration data required to run the task from the API server.
 func (a *Agent) makeTaskConfig(ctx context.Context, tc *taskContext) (*internal.TaskConfig, error) {
-	if tc.project == nil {
-		grip.Info("Fetching project config.")
-		err := a.fetchProjectConfig(ctx, tc)
-		if err != nil {
-			return nil, err
-		}
+	if tc.taskConfig != nil && tc.taskConfig.Project != nil && tc.taskConfig.Project.Identifier != "" {
```
Kimchelly, line 147: Since this function is itself making the `taskConfig`, is it sufficient to just check `tc.taskConfig != nil`? The tests presumably already have to populate `taskConfig` themselves.

agent/task_context.go
```diff
@@ -348,9 +350,9 @@ func (tc *taskContext) getTimeout(taskGroup string) (*commandBlock, error) {
 
 	return &commandBlock{
 		block:       command.TaskTimeoutBlock,
-		commands:    tg.Timeout,
+		commands:    tc.taskConfig.TaskGroup.Timeout,
 		timeoutKind: callbackTimeout,
-		getTimeout:  tc.getTaskGroupCallbackTimeout(tg),
+		getTimeout:  tc.getTaskGroupCallbackTimeout(&tc.taskConfig.TaskGroup),
```
Kimchelly, line 355: It seems like these functions don't need to pass in the `tc.taskConfig.TaskGroup` (and in the other places in this file) because `tc.getTaskGroupCallbackTimeout` already has the task group available to check.

agent/task_context.go
```diff
@@ -154,7 +173,11 @@ func (a *Agent) makeTaskConfig(ctx context.Context, tc *taskContext) (*internal.
 	}
 
 	var confPatch *patch.Patch
-	if evergreen.IsGitHubPatchRequester(tc.taskModel.Requester) {
+	requester := ""
+	if tc.taskConfig != nil && tc.taskConfig.Task != nil {
+		requester = tc.taskConfig.Task.Requester
```
Kimchelly, line 178: This line will never run in production because `tc.taskConfig` isn't populated until [`makeTaskConfig` returns and sets `tc.taskConfig`](https://github.com/evergreen-ci/evergreen/blob/a68cebae20ca387a5ba4d97cc086b044fc7cc6b1/agent/agent.go#L440-L445). It seems like you want to check `task` for the requester instead?

agent/internal/task_config.go
```diff
@@ -87,6 +88,19 @@ func NewTaskConfig(workDir string, d *apimodels.DistroView, p *model.Project, t
 		return nil, errors.Errorf("cannot find build variant '%s' for task in project '%s'", t.BuildVariant, t.Project)
 	}
 
+	var taskGroupPtr *model.TaskGroup
+	if t.TaskGroup != "" {
+		taskGroupPtr = p.FindTaskGroup(t.TaskGroup)
+		if taskGroupPtr == nil {
+			return nil, errors.New("programmatic error: task group is nil")
```
Kimchelly, line 92: nit: include the task group name, and maybe make the comment a little more descriptive of why it's bad for the task group to be nil. Something along the lines of `programmatic error: task is part of task group '%s' but no such task group is defined in the project`.

agent/command_test.go
```diff
@@ -145,10 +145,10 @@ func (s *CommandSuite) TestShellExec() {
 	nextTask := &apimodels.NextTaskResponse{
 		TaskId:     s.tc.task.ID,
 		TaskSecret: s.tc.task.Secret,
-		TaskGroup:  s.tc.taskGroup,
+		TaskGroup:  "",
 	}
 	shouldSetupGroup := !s.tc.ranSetupGroup
-	taskDirectory := s.tc.taskDirectory
+	taskDirectory := ""
```
Kimchelly, line 151: If `s.tc.taskDirectory` supposed to be empty? It seems like for the agent suite, it pre-initializes a temporary directory for the task to use, but this test suite does not. This `s.tc` likely needs to have a temporary directory set as well so it can get cleaned up after the test finishes.

agent/command_test.go
```diff
@@ -145,10 +145,10 @@ func (s *CommandSuite) TestShellExec() {
 	nextTask := &apimodels.NextTaskResponse{
 		TaskId:     s.tc.task.ID,
 		TaskSecret: s.tc.task.Secret,
-		TaskGroup:  s.tc.taskGroup,
+		TaskGroup:  "",
```
Kimchelly, line 148: If task group is not needed for this test, we can just remove it.

agent/agent_test.go
```diff
@@ -1100,9 +1098,38 @@ task_groups:
 	}, []string{panicLog})
 }
 
+func (s *AgentSuite) TestSetupGroupTimeout() {
```
Kimchelly, line 1237: Is this test needed? It's already covering the same condition as `TestSetupGroupTimeoutDoesNotFailTask`.

agent/agent_test.go
```diff
@@ -1305,90 +1335,89 @@ task_groups:
...
-	s.setupRunTask(projYml)
+	p := &model.Project{}
+	_, err := model.LoadProjectInto(s.ctx, []byte(projYml), nil, "", p)
+	s.Require().NoError(err)
+	s.tc.taskConfig.Project = p
 	s.tc.taskConfig.Task.TaskGroup = taskGroup
+	s.tc.taskConfig.TaskGroup = *p.FindTaskGroup(taskGroup)
 
-	s.Error(s.a.runPostTaskCommands(s.ctx, s.tc))
+	s.NoError(s.a.runPreTaskCommands(s.ctx, s.tc))
```
Kimchelly, line 1515: It seems like this has to call `runPostTaskCommands`, not `runPreTaskCommands`, because it's testing teardown task.

agent/agent_test.go
```diff
@@ -1305,90 +1335,89 @@ task_groups:
...
+func (s *AgentSuite) TestTeardownTaskTimeoutDoesNotFailTask() {
 	const taskGroup = "task_group_name"
-	s.tc.taskGroup = taskGroup
 	projYml := `
 task_groups:
   - name: task_group_name
-    teardown_task_can_fail_task: true
-    teardown_task:
+    teardown_task_timeout_secs: 1
+    setup_task:
```
Kimchelly, line 1503: `teardown_task`?

agent/agent_test.go
```diff
@@ -1305,90 +1335,89 @@ task_groups:
...
 	s.tc.taskConfig.Task.TaskGroup = taskGroup
+	s.tc.taskConfig.TaskGroup = *p.FindTaskGroup(taskGroup)
 
-	s.Error(s.a.runPostTaskCommands(s.ctx, s.tc))
+	s.NoError(s.a.runPreTaskCommands(s.ctx, s.tc))
 
 	s.NoError(s.tc.logger.Close())
 	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
-		"Running post-task commands",
+		"Running pre-task commands",
```
Kimchelly, line 1519: What happened to the check to ensure the timeout happened?

agent/agent_test.go
```diff
@@ -1305,90 +1335,89 @@ task_groups:
...
+		"Running command 'shell.exec' (step 1 of 1) in block 'setup_task'",
+		"Finished command 'shell.exec' (step 1 of 1) in block 'setup_task'",
+		"Finished running pre-task commands",
 	}, []string{panicLog})
 }
 
-func (s *AgentSuite) TestTeardownTaskTimeoutDoesNotFailTask() {
-	const taskGroup = "task_group_name"
-	s.tc.taskGroup = taskGroup
+func (s *AgentSuite) TestTeardownTaskSucceeds() {
```
Kimchelly, line 1527: nit: this test was originally the first test in the agent suite. Can we move it back to its original position?

agent/agent_test.go
```diff
@@ -1430,8 +1460,12 @@ task_groups:
         params:
           script: echo hi
 `
-	s.setupRunTask(projYml)
-	s.tc.taskConfig.Task.TaskGroup = s.tc.taskGroup
+	p := &model.Project{}
+	_, err := model.LoadProjectInto(s.ctx, []byte(projYml), nil, "", p)
+	s.Require().NoError(err)
+	s.tc.taskConfig.Project = p
```
Kimchelly, line 1602: It seems like replacing `setupRunTask` with `LoadProjectInto` + setting the project is unnecessary.

agent/agent_test.go
```diff
@@ -1305,90 +1335,89 @@ task_groups:
...
-	s.tc.taskConfig.Task.TaskGroup = s.tc.taskGroup
+	s.tc.taskConfig.Task.TaskGroup = taskGroup
+	s.tc.taskConfig.TaskGroup = *s.tc.taskConfig.Project.FindTaskGroup(taskGroup)
 
-	s.NoError(s.a.runPostTaskCommands(s.ctx, s.tc))
+	s.Error(s.a.runPostTaskCommands(s.ctx, s.tc))
 
 	s.NoError(s.tc.logger.Close())
 	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
 		"Running post-task commands",
```
Kimchelly, line 1493: What happened to the other `checkMockLogs` checks in this test?

agent/agent_test.go
```diff
@@ -1167,11 +1194,14 @@ task_groups:
         params:
           script: sleep 5
 `
-	s.setupRunTask(projYml)
-	s.tc.taskConfig.Task.TaskGroup = taskGroup
+	p := &model.Project{}
+	_, err := model.LoadProjectInto(s.ctx, []byte(projYml), nil, "", p)
+	s.Require().NoError(err)
```
Kimchelly, line 1335: It seems like replacing `setupRunTask` with `LoadProjectInto` + setting the project is unnecessary.

agent/agent_test.go
```diff
@@ -649,8 +646,17 @@ func (s *AgentSuite) setupRunTask(projYml string) {
...
 	s.mockCommunicator.GetProjectResponse = p
+
+	t := &task.Task{
+		Id:           "task_id",
+		BuildVariant: "mock_build_variant",
+		DisplayName:  "this_is_a_task_name",
+		Version:      "my_version",
+	}
+	s.mockCommunicator.GetTaskResponse = t
+	s.tc.taskConfig.Task = t
```
Kimchelly, line 659: It seems like this is redundant because this task is a duplicate of the `s.task` already set in `SetupTest`.

Kimchelly (changes requested):

agent/agent_test.go
malikchaya2: I think it somehow got added when resolving the merge conflict

agent/agent_test.go
malikchaya2: got messed up in the the merge, I reset it.

agent/agent_test.go
malikchaya2: reset it

agent/agent_test.go
malikchaya2: it got messed up in the merge, I reset it

agent/agent_test.go
malikchaya2: I"m not sure which check you are referring to but it should be reset now

agent/agent_test.go
malikchaya2: done

agent/agent_test.go
malikchaya2: reset it

agent/command_test.go
malikchaya2: this is no longer necessary, I reset it

agent/task_context.go
malikchaya2: It was failing for a bunch of tests when the check was just tc.taskConfig != nil. If I remember correctly, some had the taskConfig defined but not the project and were relying on this fetching it.

agent/logging_test.go
malikchaya2: removed

agent/internal/task_config.go
```diff
@@ -87,6 +88,19 @@ func NewTaskConfig(workDir string, d *apimodels.DistroView, p *model.Project, t
...
+		}
+	}
+
+	var taskGroup model.TaskGroup
+	if taskGroupPtr != nil {
+		taskGroup = *taskGroupPtr
+	}
```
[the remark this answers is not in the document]
malikchaya2: removed it

agent/task_context.go
```diff
@@ -2,19 +2,41 @@ package agent
...
+	logger                    client.LoggerProducer
+	task                      client.TaskData
+	ranSetupGroup             bool
+	taskConfig                *internal.TaskConfig
+	timeout                   timeoutInfo
+	oomTracker                jasper.OOMTracker
+	traceID                   string
+	unsetFunctionVarsDisabled bool
+	// TODO (EVG-20289): see if taskDirectory can be replaced with
+	// TaskConfig.WorkDir.
```
Kimchelly, line 32: I intentionally left this for a follow-up PR to reduce the surface area of potential breakages from this change.

Kimchelly (commented):
Putting this up for evg-app re-review since Chaya is out and I've already made my own changes from my original code review. The changes are:

* De-duplicate some fields that were replicated across `taskContext` and `taskConfig`.
* Changed some pointer fields in `taskConfig` to non-pointers because they're not optional and are always populated when the `taskConfig` is made (e.g. expansions).
* Convert the `taskContext.taskGroup` string to a `*model.TaskGroup` that gets looked up once when initializing the `taskConfig`, rather than having to look up the task group every time the agent needs to use it.

agent/task_context.go
```diff
@@ -127,34 +149,41 @@ func (tc *taskContext) getTimeoutType() timeoutType {
...
+	if tc.taskConfig != nil {
+		// This is only relevant in tests. For convenience, tests can
+		// pre-initialize a task config to use instead of fetching the task
+		// setup data using the communicator.
+		return tc.taskConfig, nil
 	}
+
+	grip.Info("Fetching project config.")
+	var err error
+	task, project, expansions, redacted, err := a.fetchProjectConfig(ctx, tc)
```
ablack12, line 161: nit: preference not to call this task since that's a package name

agent/task_context.go
```diff
@@ -127,34 +149,41 @@ func (tc *taskContext) getTimeoutType() timeoutType {
...
-		}
+	if tc.taskConfig != nil {
+		// This is only relevant in tests. For convenience, tests can
+		// pre-initialize a task config to use instead of fetching the task
+		// setup data using the communicator.
+		return tc.taskConfig, nil
 	}
+
+	grip.Info("Fetching project config.")
+	var err error
```
ablack12, line 160: don't need this line i think

agent/agent.go
```diff
@@ -482,41 +464,36 @@ func (a *Agent) handleSetupError(ctx context.Context, tc *taskContext, err error
...
-		}
+	} else if nextTask.TaskGroup != previousTaskGroup { // next task has a different task group
 		return true
 	}
-
 	return false
 }
 
-func (a *Agent) fetchProjectConfig(ctx context.Context, tc *taskContext) error {
+func (a *Agent) fetchProjectConfig(ctx context.Context, tc *taskContext) (*task.Task, *model.Project, util.Expansions, map[string]bool, error) {
```
ablack12, line 483: this name seems inaccurate now (or maybe always was) that we don't add things to the task context. Maybe fetchTaskInfo?

agent/internal/task_config.go
```diff
@@ -20,54 +20,67 @@ import (
 
 type TaskConfig struct {
 	Distro             *apimodels.DistroView
-	ProjectRef         *model.ProjectRef
-	Project            *model.Project
-	Task               *task.Task
-	BuildVariant       *model.BuildVariant
-	Expansions         *util.Expansions
+	ProjectRef         model.ProjectRef
```
ablack12, line 23: Why don't we want these to be pointers anymore? This is a lot more stuff to pass around

agent/internal/task_config.go
```diff
@@ -111,39 +131,21 @@ func (c *TaskConfig) GetCloneMethod() string {
...
-	if tg == nil {
-		return nil, errors.Errorf("couldn't find task group '%s' in project '%s'", taskGroup, tc.Project.Identifier)
-	}
-
-	return tg, nil
-}
-
 // Validate validates that the task config is populated with the data required
 // for a task to run.
+// Note that this is here only as legacy code. These checks are not sufficient
```
ablack12, line 136: do we want to remove these calls then? Or shore up the validate function as future work?

agent/agent.go
```diff
@@ -1267,15 +1065,15 @@ func (a *Agent) shouldKill(tc *taskContext, ignoreTaskGroupCheck bool) bool {
...
+	if tc.taskConfig.TaskGroup == nil {
 		return true
 	}
 	// kill if ignoreTaskGroupCheck is true
 	if ignoreTaskGroupCheck {
 		return true
 	}
 	// do not kill if share_processes is set
-	if tc.taskConfig.TaskGroup.Name != "" && tc.taskConfig.TaskGroup.ShareProcs {
+	if tc.taskConfig.TaskGroup != nil && tc.taskConfig.TaskGroup.ShareProcs {
```
hadjri, line 1076: This can just be `if tc.taskConfig.TaskGroup.ShareProcs` since an earlier block guarantees the task group is not nil

hadjri: We actually can just remove this if statement and call `return !tc.taskConfig.TaskGroup.ShareProcs`

agent/agent.go
Kimchelly: I renamed it. And yeah, it always fetched more stuff than just the project model.

agent/agent.go
Kimchelly: Done.

agent/task_context.go
Kimchelly: Hmm yeah, I actually forgot to push to the correct branch since I don't own this PR b ranch. This should go away after the push.

agent/task_context.go
Kimchelly: This will also go away with the push.

agent/internal/task_config.go
Kimchelly: This avoids an issue where I've noticed a lot of agent code is paranoia nil checking everything because nobody's sure if the pointer indicates they're optional values. In reality, after the task is set up, there's no case where these could be nil and possibly run the task, so they're not optional.

The thing that gets passed around between functions a lot is the `*TaskConfig`, so it's not copying structs very much. For the case of expansions, the `util.Expansions` is already implicitly a pointer type since it's just a wrapper type for a map.

agent/internal/task_config.go
Kimchelly: We could remove these (they seem to only apply to tests that don't load the task config from the app server), but I didn't want to increase the amount of changes made by this PR. We can have a follow-up to delete the validation.

agent/internal/task_config.go
```diff
@@ -80,7 +80,8 @@ func (t *TaskConfig) GetExecTimeout() int {
 }
 
 // NewTaskConfig validates that the required inputs are given and populates the
-// information necessary for a task to run.
+// information necessary for a task to run. It is generally preferred to use
```
ablack12, line 83: Out of curiosity, are there still places where we're initializing the struct manually?

agent/internal/task_config.go
ablack12: That works for me!

ablack12 (approved):

agent/internal/task_config.go
Kimchelly: I forgot to mention, but the ticket to get rid of the extra validation is EVG-20852.

agent/internal/task_config.go
Kimchelly: It's only in tests. The problem is that when they do this, the risk of the test panicking is pretty high if we change the task running logic (e.g. if we use a new field that is populated for tasks in prod but the unit test forgets to set it because it hard-codes `taskConfig`). Using `internal.NewTaskConfig` instead of hard-coding `taskConfig` makes it harder for the tests to panic due to a slight change in assumptions.

hadjri (approved):

Kimchelly (approved):
GitHub merge queue says it can't merge unless I approve my own PR. 🤷

PR #6987, opened 2023-09-01
EVG-20360: Check user permissions in RemoveItemFromCommitQueue resolver

minnakt:
EVG-20360

### Description
(Note: The [original PR](https://github.com/evergreen-ci/evergreen/pull/6946) for this was reverted. This is the same PR but with an adjustment, described below.)

Sometimes service users are listed as the author of commit queue patches. In this case, it's not possible to tell who actually submitted the patch. With the previous changes, this meant that users with basic permissions were not able to remove any commit queue patches created by service users, even if the patch originally belonged to them.

The fix is to allow all users to remove any commit queue patches created by service users(`User.OnlyAPI = true`).

### Testing
* Tested on staging by changing my `.evergreen-staging.yml` to correspond to a service user, submitted a commit queue patch via the CLI, and successfully removed it through the UI with basic permissions (below is a debug statement I sent to Splunk to make sure the new condition was reached):
Screenshot 2023-09-01 at 3 13 18 PM

* Augmented unit tests and GraphQL tests


rest/data/commit_queue.go
```diff
@@ -602,7 +602,23 @@ func CheckCanRemoveCommitQueueItem(ctx context.Context, sc Connector, usr *user.
...
 			}
 		}
-		if usr.Id != utility.FromStringPtr(patch.Author) {
+
+		// TODO: Remove user lookup and OnlyAPI condition after EVG-20118 is completed.
+		// Since we don't require users to save their GitHub information before submitting patches,
+		// we must allow them to remove any patches created by service users (OnlyAPI = true).
+		patchAuthor := utility.FromStringPtr(patch.Author)
+		patchUsr, err := user.FindOneById(patchAuthor)
+		if patchUsr == nil || err != nil {
```
ablack12, line 611: We handle these cases in separate if statements because errors.Wrapf defaults to nil if error isn't nil, and so nil.Error will panic (in this case I'd also expect different error messages -- one is an error finding, one is that user isn't found).

ablack12 (approved):
LGTM with error fix

rest/data/commit_queue.go
minnakt: oops, thanks annie!

PR #6989, opened 2023-09-01
EVG-20675 Move distro retrieval for create host to separate function

hadjri:
EVG-20675

### Description
The error that rises from a host.create distro not being compatible with a docker provider should be considered an improper request, not an internal server error on Evergreen's end. The fact that it is currently intertwined with the `MakeHost` means that [this retry loop](https://gist.github.com/hadjri/f0f75045653f4043b7805a1f86a96211#file-http-go-L331-L350) in the agent will keep looping until it times out, rather than returning the error immediately as it should. 

Made a refactor to separate the distro retrieval / validation from the actual host creation function to keep this as part of the route parsing logic.
### Testing
Verified in staging that the `distro '%s' provider must support Docker but actual provider is '%s'` error now is shown when host.create tries to use the docker provider for an improper distro.

### DIFF
--- a/rest/data/host_create.go
+++ b/rest/data/host_create.go
@@ -152,7 +152,7 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
 			continue
 		}
 		for i := 0; i < numHosts; i++ {
-			_, err := MakeHost(ctx, env, t.Id, user.Username(), keyVal, createHost, d)
+			_, err := MakeHost(ctx, env, t.Id, user.Username(), keyVal, createHost, *d)
 			if err != nil {
 				return errors.Wrap(err, "creating intent host")
 			}
@@ -303,18 +303,18 @@ func createHostFromCommand(cmd model.PluginCommandConf) (*apimodels.CreateHost, 
 }
 
 // MakeHost creates a host or container to run for host.create.
-func MakeHost(ctx context.Context, env evergreen.Environment, taskID, userID, publicKey string, createHost apimodels.CreateHost, distro *distro.Distro) (*host.Host, error) {
+func MakeHost(ctx context.Context, env evergreen.Environment, taskID, userID, publicKey string, createHost apimodels.CreateHost, distro distro.Distro) (*host.Host, error) {
 	if evergreen.IsDockerProvider(createHost.CloudProvider) {
 		return makeDockerIntentHost(ctx, env, taskID, userID, createHost, distro)
 	}
 	return makeEC2IntentHost(ctx, env, taskID, userID, publicKey, createHost, distro)
 }
 
-func makeDockerIntentHost(ctx context.Context, env evergreen.Environment, taskID, userID string, createHost apimodels.CreateHost, d *distro.Distro) (*host.Host, error) {
+func makeDockerIntentHost(ctx context.Context, env evergreen.Environment, taskID, userID string, createHost apimodels.CreateHost, d distro.Distro) (*host.Host, error) {
 	// Do not provision task-spawned hosts.
 	d.BootstrapSettings.Method = distro.BootstrapMethodNone
 
-	options, err := getHostCreationOptions(*d, taskID, userID, createHost)
+	options, err := getHostCreationOptions(d, taskID, userID, createHost)
 	if err != nil {
 		return nil, errors.Wrap(err, "making intent host options")
 	}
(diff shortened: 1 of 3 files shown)


rest/data/host_create.go
```diff
@@ -157,6 +162,44 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
 	return catcher.Resolve()
 }
 
+// GetDistro returns the distro based on the name and provider.
+func GetDistro(ctx context.Context, createHost apimodels.CreateHost) (*distro.Distro, error) {
```
Kimchelly, line 166: nit: since this is only usable with `host.create`, can we call this something more specific like `GetHostCreateDistro`?

rest/data/host_create.go
```diff
@@ -343,34 +372,15 @@ func makeDockerIntentHost(ctx context.Context, env evergreen.Environment, taskID
...
 
-func makeEC2IntentHost(ctx context.Context, env evergreen.Environment, taskID, userID, publicKey string, createHost apimodels.CreateHost) (*host.Host, error) {
+func makeEC2IntentHost(ctx context.Context, env evergreen.Environment, taskID, userID, publicKey string, createHost apimodels.CreateHost, d distro.Distro) (*host.Host, error) {
 	if createHost.Region == "" {
 		createHost.Region = evergreen.DefaultEC2Region
 	}
-	// get distro if it is set
-	d := distro.Distro{}
 	ec2Settings := cloud.EC2ProviderSettings{}
 	var err error
```
Kimchelly, line 380: nit: it shouldn't be necessary to pre-define `err` anymore since the `if createHost.Distro != ""` block doesn't need it.

rest/data/host_create.go
```diff
@@ -157,6 +162,44 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
 	return catcher.Resolve()
 }
 
+// GetDistro returns the distro based on the name and provider.
```
Kimchelly, line 165: Can we mention that for EC2 `host.create`, the distro is optional and this is allowed to return an empty distro if the user did not specify one?

Kimchelly (changes requested):

rest/data/host_create.go
```diff
@@ -157,6 +162,56 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
...
+				return nil, errors.Errorf("distro '%s' not found", createHost.Distro)
+			}
+			d = foundDistro
+		}
+	}
+	// Do not provision task-spawned hosts.
+	d.BootstrapSettings.Method = distro.BootstrapMethodNone
+	if !isDockerProvider {
+		// Set provider only if it's not a Docker provider.
+		d.Provider = evergreen.ProviderNameEc2OnDemand
```
Kimchelly, line 210: nit: rather than make a separate `if` block just for this, you can just absorb this into the `else` (i.e. EC2) case above.

rest/data/host_create.go
```diff
@@ -157,6 +162,56 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
...
+			var dat distro.AliasLookupTable
+			dat, err := distro.NewDistroAliasesLookupTable(ctx)
+			if err != nil {
+				return nil, errors.Wrap(err, "getting distro lookup table")
+			}
+			distroIDs := dat.Expand([]string{createHost.Distro})
+			if len(distroIDs) == 0 {
+				return nil, errors.Wrap(err, "distro lookup returned no matching distro IDs")
+			}
+			foundDistro, err := distro.FindOneId(ctx, distroIDs[0])
```
Kimchelly, line 196: nit: since `foundDistro` and `d` are the same type and `err` has been pre-declared, you can just do `d, err = distro.FindOneId` here.

rest/data/host_create.go
```diff
@@ -157,6 +162,56 @@ func CreateHostsFromTask(ctx context.Context, env evergreen.Environment, t *task
...
+			return nil, errors.Wrapf(err, "finding distro '%s'", createHost.Distro)
+		}
+		if d == nil {
+			return nil, errors.Errorf("distro '%s' not found", createHost.Distro)
+		}
+		if !evergreen.IsDockerProvider(d.Provider) {
+			return nil, errors.Errorf("distro '%s' provider must support Docker but actual provider is '%s'", d.Id, d.Provider)
+		}
+	} else {
+		d = &distro.Distro{}
```
Kimchelly, line 185: nit: this if-else + defaulting of distro reads a little awkwardly since you have `d` starting out as a nil pointer, then for EC2, it gets initialized as an empty distro, then gets further overridden to an actual distro if the user specifies a distro name. It might read a little nicer if, rather than setting it 3 times, we either set the initial `d` value to `&distro.Distro{}` unconditionally, or move this defaulting into an `else` following `if createHost.Distro != ""`.

Kimchelly (approved):
LGTM mod some nitpick suggestions.

PR #6995, opened 2023-09-05
EVG-19885 Show errors correctly for patch command

ZackarySantana:
EVG-19885

### Description
The patch command ouputs warnings when it runs in to an error. Instead, the patch command should only output errors- never warnings. If users want to view present warnings, they can run `evergreen validate` or look at the UI.

Makes a change to filter by errors only at the 'error' level and only output's those at that 'error' level.

### Testing
Forced warnings/errors in staging and it displays only the error- not the warning.

Before
Screen Shot 2023-09-05 at 4 38 43 PM
(Shows error, invalid patched config, and version control for project warnings)

After:
Screen Shot 2023-09-05 at 4 38 20 PM
(Shows only error)

### Documentation
Add documentation for patch command not displaying warnings and to use the validate command if you want to see warnings.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -53,6 +53,8 @@ To skip all (y/n) prompts, the `-y` keyword can be given:
 evergreen patch -y
 ```
 
+The patch command ignores warnings. If you want to view warnings, look at the [validate](#validating-changes-to-config-files) command.
+
 Working Tree Changes
 ---
 By default patches will include only committed changes, not changes in Git's working tree or index. To include changes from the working tree use the `--uncommitted, -u` flag or set a default by inserting `patch_uncommitted_changes: true` in the `~/.evergreen.yml` file.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -14,6 +14,8 @@ On macOS, the evergreen binary is currently not notarized. To allow running it, 
 Basic Patch Usage
 --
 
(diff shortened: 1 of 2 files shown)


docs/CLI.md
```diff
@@ -53,6 +53,8 @@ To skip all (y/n) prompts, the `-y` keyword can be given:
 evergreen patch -y
 ```
 
+The patch command ignores warnings. If you want to view warnings, look at the [validate](#validating-changes-to-config-files) command.
```
Kimchelly, line 56: This assumes that the user already is aware that `evergreen patch` will output project validation errors, which I assume most people aren't aware of. The placement is also a little weird because it comes after the "basic usage" tutorial. I would move this to the beginning of the "basic usage" section and add a brief sentence prefacing what the patch command does to start off. For example, something like:

> `evergreen patch` allows you to submit patches to test your local changes. It will also check your project YAML any for validation errors before submission.

Kimchelly (changes requested):

docs/CLI.md
ZackarySantana: @Kimchelly should I reference the validate command at all?

> evergreen patch allows you to submit patches to test your local changes. It will also check your project YAML any for validation errors before submission. If you want to view warnings, look at the [validate]() command.

Or should I keep it succinct like yours?

Kimchelly: Hmm, I think it still makes sense to reference `evergreen validate` since people may want a more complete validation rather than just patch-blocking errors.

Kimchelly (approved):

PR #6996, opened 2023-09-05
EVG-20020:  Add new test log info field to the test result model

julianedwards:
EVG-20020

### Description
New field to store log info for a test result.

### Testing
Not needed yet.

### Documentation
N/A

### DIFF
--- a/model/testresult/testresult.go
+++ b/model/testresult/testresult.go
@@ -29,19 +29,32 @@ type TaskTestResultsStats struct {
 
 // TestResult represents a single test result from an Evergreen task run.
 type TestResult struct {
-	TaskID          string    `json:"task_id" bson:"task_id"`
-	Execution       int       `json:"execution" bson:"execution"`
-	TestName        string    `json:"test_name" bson:"test_name"`
-	DisplayTestName string    `json:"display_test_name" bson:"display_test_name"`
-	GroupID         string    `json:"group_id" bson:"group_id"`
-	Status          string    `json:"status" bson:"status"`
-	BaseStatus      string    `json:"base_status" bson:"base_status"`
-	LogTestName     string    `json:"log_test_name" bson:"log_test_name"`
-	LogURL          string    `json:"log_url" bson:"log_url"`
-	RawLogURL       string    `json:"raw_log_url" bson:"raw_log_url"`
-	LineNum         int       `json:"line_num" bson:"line_num"`
-	TestStartTime   time.Time `json:"test_start_time" bson:"test_start_time"`
-	TestEndTime     time.Time `json:"test_end_time" bson:"test_end_time"`
+	TaskID          string       `json:"task_id" bson:"task_id"`
+	Execution       int          `json:"execution" bson:"execution"`
+	TestName        string       `json:"test_name" bson:"test_name"`
+	DisplayTestName string       `json:"display_test_name" bson:"display_test_name"`
+	GroupID         string       `json:"group_id" bson:"group_id"`
+	Status          string       `json:"status" bson:"status"`
+	BaseStatus      string       `json:"base_status" bson:"base_status"`
+	LogInfo         *TestLogInfo `json:"log_info" bson:"log_info"`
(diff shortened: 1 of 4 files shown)


model/testresult/testresult.go
```diff
@@ -29,19 +29,32 @@ type TaskTestResultsStats struct {
...
+	LineNum     int    `json:"line_num" bson:"line_num"`
+}
+
+// TestLogInfo describes a metadata for a test result's log stored using
+// Evergreen logging.
+type TestLogInfo struct {
+	LogName       string    `json:"log_name" bson:"log_name"`
+	LogsToMerge   []*string `json:"logs_to_merge" bson:"logs_to_merge"`
+	LineNum       int32     `json:"line_num" bson:"line_num"`
+	RenderingType *string   `json:"rendering_type" bson:"rendering_type"`
```
Kimchelly, line 56: Should this be an enum, or is a string the best choice?

model/testresult/testresult.go
```diff
@@ -29,19 +29,32 @@ type TaskTestResultsStats struct {
...
+	RawLogURL   string `json:"raw_log_url" bson:"raw_log_url"`
+	LineNum     int    `json:"line_num" bson:"line_num"`
+}
+
+// TestLogInfo describes a metadata for a test result's log stored using
+// Evergreen logging.
+type TestLogInfo struct {
+	LogName       string    `json:"log_name" bson:"log_name"`
+	LogsToMerge   []*string `json:"logs_to_merge" bson:"logs_to_merge"`
+	LineNum       int32     `json:"line_num" bson:"line_num"`
```
Kimchelly, line 55: Is there a reason the ints are specifically 32-bit? Is 64-bit not acceptable?

model/testresult/testresult.go
```diff
@@ -29,19 +29,32 @@ type TaskTestResultsStats struct {
...
-	GroupID         string    `json:"group_id" bson:"group_id"`
-	Status          string    `json:"status" bson:"status"`
-	BaseStatus      string    `json:"base_status" bson:"base_status"`
-	LogTestName     string    `json:"log_test_name" bson:"log_test_name"`
-	LogURL          string    `json:"log_url" bson:"log_url"`
-	RawLogURL       string    `json:"raw_log_url" bson:"raw_log_url"`
-	LineNum         int       `json:"line_num" bson:"line_num"`
-	TestStartTime   time.Time `json:"test_start_time" bson:"test_start_time"`
-	TestEndTime     time.Time `json:"test_end_time" bson:"test_end_time"`
+	TaskID          string       `json:"task_id" bson:"task_id"`
```
Kimchelly, line 32: Are we intentionally making all the BSON tags not `omitempty` for a reason? I thought the recommendation from the Mongo driver was to use `omitempty` where possible.

Kimchelly (changes requested):
The original ticket mentioned writing docs for the fields - will they be added in a follow-up PR?

model/testresult/testresult.go
julianedwards: This is to follow the parquet schema (which will hopefully get moved into Evergreen) that uses int32s.

model/testresult/testresult.go
julianedwards: I can add omitempty tags but we only use Mongo for local development.

julianedwards:
> The original ticket mentioned writing docs for the fields - will they be added in a follow-up PR?

Yes, I am going to update the docs separately. Some of this may change slightly so I wanted to wait until I actually start testing this E2E before updating any public-facing documentation.

model/testresult/testresult.go
julianedwards: It could be an enum, but the choices really depend on Parsley which feels external enough that I don't want to hardcode the rendering type options here.

model/testresult/testresult.go
Kimchelly: So just to clarify - the BSON tags are only relevant for local development and prod never stores these?

Kimchelly (approved):
SGTM! Just one lingering question for my own understanding.

PR #6997, opened 2023-09-05
EVG-20823 Pass in data connector to make versions route

hadjri:
EVG-20823

### Description
Updating version create route to pass in the connector it needs to initialize.


Kimchelly (approved):

ablack12 (commented):
Where you able to verify on staging that this route now works? (we should do this ticket rn but worth making sure it's done fully)

hadjri:
> Where you able to verify on staging that this route now works? (we should do this ticket rn but worth making sure it's done fully)

Just confirmed in staging the panics go away with this change: 
image

ablack12 (approved):

PR #6998, opened 2023-09-05
EVG-20800 Ensure content type is cached for task uploaded files

khelif96:
EVG-20800

### Description
Ensure `content_type` (mime type) is saved in the DB. When writing files to s3 we do not maintain a record of the file's mime type. This is problematic if we want to perform operations that depend on this mime type.  In order to avoid querying s3 every time we want to use the mime type this pr saves it in the artifact_files collection alongside the file reference.

### Testing
Created several patches that upload files to s3 and determined that the task file document contains the mime type.
 Example task: https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_upload_sample_log_patch_462647a590a92231b96e6a5b3cae1473e87cc663_64f78a5e9dbe3274877316da_23_09_05_20_07_12/logs?execution=0
Document:
image

I would appreciate suggestions on how to write unit tests for this. I'm not sure if the agent tests actually write to the test db, or at least they didn't when I tried.


agent/command/s3_copy.go
```diff
@@ -346,8 +346,9 @@ func (c *s3copy) attachFiles(ctx context.Context, comm client.Communicator,
 	}
 	logger.Execution().Infof("Attaching file '%s'.", displayName)
 	file := artifact.File{
-		Name: displayName,
-		Link: fileLink,
+		Name:        displayName,
+		Link:        fileLink,
+		ContentType: c.ContentType,
```
Kimchelly, line 351: I'm not sure using this content type field is actually valid because 1. this appears to be a dead field just from looking for Go references (the [docs](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Commands#s3copycopy) also support that these are not supported fields), and 2. would be incompatible with the fact that S3 copy can copy multiple files (which may have different content types). It seems like the content type would have to be in the `s3CopyFile` struct so that the user could specify the content type for each copied file.

agent/command/s3_copy.go
```diff
@@ -346,8 +346,9 @@ func (c *s3copy) attachFiles(ctx context.Context, comm client.Communicator,
 	}
 	logger.Execution().Infof("Attaching file '%s'.", displayName)
 	file := artifact.File{
-		Name: displayName,
-		Link: fileLink,
+		Name:        displayName,
```
Kimchelly, line 349: GH doesn't let me comment on unchanged lines, but is [the S3 bucket](https://github.com/evergreen-ci/evergreen/blob/d7daa79cf9adb8145cc0bbe23905bb320d05c150/agent/command/s3_copy.go#L274-L279) also supposed to set the content type for uploading files? Seems like if we were to use the content type for our own purposes, we might as well set it for the actual S3 file as well.

model/artifact/db.go
```diff
@@ -9,15 +9,16 @@ import (
...
+	TaskIdKey      = bsonutil.MustHaveTag(Entry{}, "TaskId")
+	TaskNameKey    = bsonutil.MustHaveTag(Entry{}, "TaskDisplayName")
+	BuildIdKey     = bsonutil.MustHaveTag(Entry{}, "BuildId")
+	FilesKey       = bsonutil.MustHaveTag(Entry{}, "Files")
+	ExecutionKey   = bsonutil.MustHaveTag(Entry{}, "Execution")
+	CreateTimeKey  = bsonutil.MustHaveTag(Entry{}, "CreateTime")
+	NameKey        = bsonutil.MustHaveTag(File{}, "Name")
+	LinkKey        = bsonutil.MustHaveTag(File{}, "Link")
+	ContentTypeKey = bsonutil.MustHaveTag(File{}, "ContentType")
+	AwsSecretKey   = "aws_secret"
```
Kimchelly, line 21: This isn't related to the change, but it's gonna bother me - can we change this to `bsonutil.MustHaveTag(File{}, "AwsSecret")`?

agent/command/s3_put.go
```diff
@@ -509,13 +509,14 @@ func (s3pc *s3put) attachFiles(ctx context.Context, comm client.Communicator, lo
...
-			Bucket:     bucket,
-			FileKey:    fileKey,
+			Name:        displayName,
+			Link:        fileLink,
+			Visibility:  s3pc.Visibility,
+			AwsKey:      key,
+			AwsSecret:   secret,
+			Bucket:      bucket,
+			FileKey:     fileKey,
+			ContentType: s3pc.ContentType,
```
Kimchelly, line 519: Can we add a test for this? There's already [a test that inspects the artifacts that would be sent to the app server](https://github.com/evergreen-ci/evergreen/blob/d7daa79cf9adb8145cc0bbe23905bb320d05c150/agent/command/s3_put_test.go#L342).

agent/command/s3_copy.go
Kimchelly, line 351: I would prefer to also test this, but it seems like the s3Copy.copy tests are kind of nonexistent... 🤷

Kimchelly (changes requested):
* Does the user always have to specify the content type to have it be saved? What happens if they don't set one?
* I made EVG-20843 to delete the dead fields that I noticed in s3Copy.copy.

agent/command/s3_copy.go
ablack12: We have https://jira.mongodb.org/browse/EVG-19764 also

Kimchelly: Ah, I didn't see that. Not sure how easy it is to actually deprecate without somehow giving warning in advance, because I see multiple projects in `evergreen all-configs` using it.

khelif96: I'm going to not add this functionality to s3copy if the plan is to eventually deprecate it.

ablack12: What projects do you see Kim? I actually don't see that but my configs dir could be out of date. But yeah deprecation could take a minute / might need to be axed if it's not just server projects like Brian mentioned in the linked build ticket

khelif96:
> * Does the user always have to specify the content type to have it be saved? What happens if they don't set one?
https://github.com/evergreen-ci/evergreen/blob/6b17bba18159f3b6c141b62a9a671fdbd87ac547/agent/command/s3_put.go#L165-L167
Yes there is validation logic that guarantees ContentType is provided.

agent/command/s3_put.go
khelif96: Thanks done!

agent/command/s3_copy.go
khelif96: I removed the changes for s3 copy.

agent/command/s3_copy.go
khelif96: Thanks for the clarification. I've removed support for this on s3.copy operations.

ablack12: @khelif96 what I meant by this thread is that I think we _should_ support content type for s3 copy since we aren't sure that removing it is going to be feasible.

agent/command/s3_put_test.go
```diff
@@ -356,6 +356,45 @@ func TestSignedUrlVisibility(t *testing.T) {
...
+	logger, err := comm.GetLoggerProducer(ctx, client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}, nil)
+	require.NoError(t, err)
+
+	localFiles := []string{"file1", "file2"}
+	remoteFile := "remote file"
+
+	require.NoError(t, s.attachFiles(ctx, comm, logger, localFiles, remoteFile))
+
+	attachedFiles := comm.AttachedFiles
+	if v, found := attachedFiles[""]; found {
```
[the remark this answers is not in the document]
khelif96: Made this change assuming I did the thing you meant.

agent/command/s3_copy.go
khelif96: I took a look at this I would probably push back on supporting it for s3Copy.Copy. It is used .0004% of the time compared to other S3 operations. I don't think it would be worth the effort to implement given it will likely not be used at all.
https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/result/fspuN6naiEn

khelif96: I did a little more research into the implementation of `s3Copy.Copy` it looks it relies on an [s3 method](https://github.com/evergreen-ci/pail/blob/74d684cc93493d325861a2d0777b519eced027ea/s3_bucket.go#L998) to just move the files between s3 buckets and does not actually ever download the file. Caching the mime type for this would require us to perform a [`GetHeadObject` ](https://github.com/evergreen-ci/pail/blob/74d684cc93493d325861a2d0777b519eced027ea/s3_bucket.go#L1387) operation for each file we wish to copy to get its ContentType and then save it. Given such low usage of the method, I don't think it's worth the additional complexity required to implement it. We can reinvestigate if people start asking for it. Although [most usage I found](https://github.com/search?q=%22s3copy.copy%22+language:YAML+owner:evergreen-ci+OR+owner:mongodb+OR+owner:10gen+&type=code) seems to only use it to move archived files around so I don't think people will notice.

agent/command/s3_put_test.go
```diff
@@ -386,13 +387,13 @@ func TestContentTypeSaved(t *testing.T) {
 	require.NoError(t, s.attachFiles(ctx, comm, logger, localFiles, remoteFile))
 
 	attachedFiles := comm.AttachedFiles
-	if v, found := attachedFiles[""]; found {
+	if v, ok := attachedFiles["mock_id"]; ok {
```
Kimchelly, line 390: This has to be unconditional. Otherwise, if the task ID isn't in the mock data map, then the test won't check the assertions.
```go
v, ok := attachedFiles["mock_id"]
// Must be able to find an attached file
require.True(t, ok)
// Needs to produce 1 file with the expected content type.
assert.Len(t, v, 1)
```

Kimchelly, line 390: Also `mock_id` here can just be `conf.Task.id`.

Kimchelly (approved):
LGTM mod small lingering test things.

agent/command/s3_put_test.go
khelif96: Thanks for the clarification!

agent/command/s3_copy.go
ablack12: Wow honeycomb to the rescue. That works for me, we should make a note to include documentation for this or include it in the mouse over for why the button isn't available

khelif96: Done I created a [ticket](https://jira.mongodb.org/browse/EVG-20851) to write some docs for this feature and I'll make a note of it there.

PR #6999, opened 2023-09-06
EVG-19128 Case on non mainline commits for FindTaskOnPreviousCommit calculation

khelif96:
EVG-19128

### Description
It's possible for some versions to increment the order number value used to keep track of the order of commits. This query would sometimes fail because we only compared against order number - 1 and if the previous order number was a non-mainline commit this query would fail. This change adds a `$lt` so that we attempt to fetch the most recent `gitter_request` commit

### Testing
Wrote unit tests.

### DIFF
--- a/model/task/task.go
+++ b/model/task/task.go
@@ -852,7 +852,7 @@ func (t *Task) FindTaskOnBaseCommit() (*Task, error) {
 }
 
 func (t *Task) FindTaskOnPreviousCommit() (*Task, error) {
-	return FindOne(db.Query(ByPreviousCommit(t.BuildVariant, t.DisplayName, t.Project, evergreen.RepotrackerVersionRequester, t.RevisionOrderNumber)))
+	return FindOne(db.Query(ByPreviousCommit(t.BuildVariant, t.DisplayName, t.Project, evergreen.RepotrackerVersionRequester, t.RevisionOrderNumber)).Sort([]string{"-" + RevisionOrderNumberKey}))
 }
 
 // FindIntermediateTasks returns the tasks from most recent to least recent between two tasks.
--- a/model/task/task_test.go
+++ b/model/task/task_test.go
@@ -3774,7 +3774,7 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
 }
 
 func TestFindTaskOnPreviousCommit(t *testing.T) {
-	require.NoError(t, db.ClearCollections(Collection, OldCollection))
+	require.NoError(t, db.ClearCollections(Collection))
 	t1 := Task{
 		Id:                  "t1",
 		Version:             "v1",
@@ -3802,7 +3802,7 @@ func TestFindTaskOnPreviousCommit(t *testing.T) {
 
 	task, err := t2.FindTaskOnPreviousCommit()
 	assert.NoError(t, err)
-	assert.NotNil(t, task)
+	require.NotNil(t, task)
 	assert.Equal(t, t1.Id, task.Id)
 	assert.Equal(t, t1.Version, task.Version)
 	t3 := Task{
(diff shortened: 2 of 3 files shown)


model/task/task.go
```diff
@@ -852,7 +852,7 @@ func (t *Task) FindTaskOnBaseCommit() (*Task, error) {
 }
 
 func (t *Task) FindTaskOnPreviousCommit() (*Task, error) {
-	return FindOne(db.Query(ByPreviousCommit(t.BuildVariant, t.DisplayName, t.Project, evergreen.RepotrackerVersionRequester, t.RevisionOrderNumber)))
+	return FindOne(db.Query(ByPreviousCommit(t.BuildVariant, t.DisplayName, t.Project, evergreen.RepotrackerVersionRequester, t.RevisionOrderNumber)).Sort([]string{"-" + RevisionOrderNumberKey}))
```
Kimchelly, line 855: Does this sort utilize an index, or is the filtered set small enough that it doesn't need an index? I glanced at the existing indexes in the DB to see which one might fit, but none of them jumped out to me immediately.

model/task/task_test.go
```diff
@@ -3773,6 +3773,72 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
...
+	}
+	assert.NoError(t, db.Insert(Collection, t4))
+
+	// Should fetch the latest mainline commit task and should not consider non gitter tasks
+	task, err = t4.FindTaskOnPreviousCommit()
+	assert.NoError(t, err)
+	assert.NotNil(t, task)
+	assert.Equal(t, t2.Id, task.Id)
+	assert.Equal(t, t2.Version, task.Version)
+
```
Kimchelly, line 3839: nit: remove random newline.

model/task/task_test.go
```diff
@@ -3773,6 +3773,72 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
...
+		Requester:           evergreen.RepotrackerVersionRequester,
+		BuildVariant:        "bv",
+		DisplayName:         "dn",
+		Project:             "p",
+	}
+	assert.NoError(t, db.Insert(Collection, t2))
+
+	task, err := t2.FindTaskOnPreviousCommit()
+	assert.NoError(t, err)
+	assert.NotNil(t, task)
```
Kimchelly, line 3805: `require.NotNil` or else the following lines will panic because if they dereference a nil task.

model/task/task_test.go
```diff
@@ -3773,6 +3773,72 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
...
+		BuildVariant:        "bv",
+		DisplayName:         "dn",
+		Project:             "p",
+	}
+	assert.NoError(t, db.Insert(Collection, t4))
+
+	// Should fetch the latest mainline commit task and should not consider non gitter tasks
+	task, err = t4.FindTaskOnPreviousCommit()
+	assert.NoError(t, err)
+	assert.NotNil(t, task)
```
Kimchelly, line 3836: `require.NotNil`

model/task/task_test.go
```diff
@@ -3773,6 +3773,72 @@ func TestByExecutionTasksAndMaxExecution(t *testing.T) {
 	})
 }
 
+func TestFindTaskOnPreviousCommit(t *testing.T) {
+	require.NoError(t, db.ClearCollections(Collection, OldCollection))
```
Kimchelly, line 3777: I'm pretty sure it's only necessary to clear the `tasks` collection and not the `old_tasks` collection because this test doesn't involve archiving tasks.

Kimchelly (changes requested):

model/task/task.go
khelif96: When I tried performing this build_variant_1_display_name_1_order_1 was the index the query planner used and it didn't have to do any in memory sorts or scan more documents then necessary. 
image

Kimchelly (approved):

PR #7004, opened 2023-09-06
EVG-20612 add logging to understand batchtime calls

ablack12:
EVG-20612

### Description
I looked into adding an index for this query but it'd be very expensive. I also started digging into some of the traces and it looks like these queries don't return any versions, which shouldn't be the case; I'm wondering if there's a bug in the code so that we're calling this function too often or something but it's hard to know without more info, so my next thought is to add some info to continue investigating.


ablack12:
I plan to use it in investigation for this ticket so I'd remove them before closing this ticket out.

hadjri (approved):
Sounds good!

PR #7006, opened 2023-09-07
EVG-20790 Add stepback last failed, last passed, and next task id to the database

ZackarySantana:
EVG-20790

### Description
This adds the fields for last failed stepback task id, last passed stepback task id, next stepback task id, to the task struct and corresponding database calls. This is to support bisect stepback.

### Testing
Existing tests


model/task/db.go
```diff
@@ -87,6 +87,9 @@ var (
 	PriorityKey                    = bsonutil.MustHaveTag(Task{}, "Priority")
 	ActivatedByKey                 = bsonutil.MustHaveTag(Task{}, "ActivatedBy")
 	StepbackDepthKey               = bsonutil.MustHaveTag(Task{}, "StepbackDepth")
+	LastFailingStepbackTaskId      = bsonutil.MustHaveTag(Task{}, "LastFailingStepbackTaskId")
```
ablack12, line 90: Should end these variable names with key, as above, to indicate that it's a bson tag key and not storing an actual ID

model/task/task.go
```diff
@@ -273,6 +277,14 @@ type Task struct {
 	IsEssentialToSucceed bool `bson:"is_essential_to_succeed" json:"is_essential_to_succeed"`
 }
 
+// StepbackInformation is used to
```
ablack12, line 280: comment haha

model/task/task.go
```diff
@@ -226,6 +222,14 @@ type Task struct {
 	ExecutionTasks        []string `bson:"execution_tasks,omitempty" json:"execution_tasks,omitempty"`
 	LatestParentExecution int      `bson:"latest_parent_execution" json:"latest_parent_execution"`
 
+	// Used to indictate what task to stepback to if needed.
+	LastFailingStepbackTaskId string `bson:"last_failing_stepback_task_id,omitempty" json:"last_failing_stepback_task_id"`
```
ablack12, line 226: May make sense to just have *StepbackInformation in the task instead of each field individually (I think it needs to be a pointer for omitempty to work but worth checking)

model/task/task.go
```diff
@@ -1434,15 +1446,21 @@ func (t *Task) SetAborted(reason AbortInfo) error {
 }
 
 // SetStepbackDepth adds the stepback depth to the task.
```
ablack12, line 1448: update comment

model/task/task.go
ablack12: I think we can just call this StepbackInfo also

model/task_lifecycle.go
```diff
@@ -256,11 +256,14 @@ func activatePreviousTask(ctx context.Context, taskId, caller string, originalSt
 	}
 
 	grip.Debug(message.Fields{
-		"ticket":         "EVG-17949",
-		"message":        "stepping back task",
-		"stepback_depth": stepbackDepth,
-		"project_id":     t.Project,
-		"task_id":        t.Id,
+		"ticket":                        "EVG-17949",
```
ablack12, line 259: We should update this ticket to be a new ticket if we intend to remove this logging at the end of the project; otherwise we shouldn't be logging at all

model/task_lifecycle.go
```diff
@@ -553,8 +556,15 @@ func doStepback(ctx context.Context, t *task.Task) error {
 		return nil
 	}
 
+	s := task.StepbackInformation{
+		StepbackDepth:             t.StepbackDepth + 1,
+		LastFailingStepbackTaskId: "",
```
ablack12, line 561: The default value for strings is the empty string so you don't need to define it here. Is it correct though that we don't want to be setting these values? Do you intend to start updating these in a future PR?

ablack12 (changes requested):

ablack12:
Looks like the base status for your Spruce e2e_test is also failed; may help to merge main and push that

model/task_lifecycle.go
ZackarySantana: Yes. For now, this ticket just adds the fields to the database to separate the logic of setting them appropriately and this.

model/task/task.go
```diff
@@ -273,6 +271,16 @@ type Task struct {
 	IsEssentialToSucceed bool `bson:"is_essential_to_succeed" json:"is_essential_to_succeed"`
 }
 
+// StepbackInfo is used to indicate what task to stepback to.
```
ablack12, line 274: this isn't quite right, is it? It stores information about stepback for the task more generally. (I imagine this comment as is would be the definition of NextStepbackTaskId)

model/task/task.go
```diff
@@ -1433,16 +1441,21 @@ func (t *Task) SetAborted(reason AbortInfo) error {
...
+func (t *Task) SetStepbackInfo(s StepbackInfo) error {
+	t.StepbackInfo = &s
 	return UpdateOne(
 		bson.M{
 			IdKey: t.Id,
 		},
 		bson.M{
 			"$set": bson.M{
-				StepbackDepthKey: stepbackDepth,
+				StepbackInfoKey: bson.M{
```
ablack12, line 1453: does it not work to just set StepbackInfoKey to s?

model/task_lifecycle.go
ablack12: I'm a little uncertain about the ticket that's now linked here, since EVG-20792 looks like it's about _adding_ logging rather than removing it, but I also don't have context about what the logs to be added in that ticket are for. @hadjri can you weigh in here?

model/task_lifecycle.go
```diff
@@ -553,8 +556,17 @@ func doStepback(ctx context.Context, t *task.Task) error {
 		return nil
 	}
 
+	depth := 1
+	if t.StepbackInfo != nil {
```
ablack12, line 560: I'd be pro a lil function called t.GetStepbackDepth() so we could avoid needing this here.

hadjri (changes requested):

model/task_lifecycle.go
hadjri: Yes, [EVG-20792](https://jira.mongodb.org/browse/EVG-20792) is about adding logging to specifically allow us to observe what next task is being chosen by the bisection algo once it is implemented.

It actually looks like [EVG-17949](https://jira.mongodb.org/browse/EVG-17949) was closed because enforcing a maximum stepback depth is no longer going to be implemented in favor of simply implementing stepback bisection. Given that we should just remove this log entirely.

model/task/task.go
```diff
@@ -273,6 +271,25 @@ type Task struct {
...
+	// this is the potentially the previous head if that passed the task and if it had not passed
+	// the task, this may be a commit further back in the history. As bisect stepback continues,
+	// it will either stay or become the NextStepbackTaskId.
+	LastPassingStepbackTaskId string `bson:"last_passing_stepback_task_id,omitempty" json:"last_passing_stepback_task_id"`
+	// NextStepbackTaskId stores the next task id to stepback to when doing bisect stepback. This
+	// is the middle of LastFailingStepbackTaskId and LastPassingStepbackTaskId.
+	NextStepbackTaskId string `bson:"next_stepback_task_id,omitempty" json:"next_stepback_task_id"`
+
+	// StepbackDepth indicates how far into stepback this task was activated, starting at 1 for stepback tasks.
+	StepbackDepth int `bson:"stepback_depth,omitempty" json:"stepback_depth,omitempty"`
```
ablack12, line 290: Given that we removed the logging for StepbackDepth, I don't think we need to actually store this field at all anymore, unless we think it'd be helpful to show a comparison of how deeply we stepback or something but given the math Brian had to do, that's probably too complicated (cc @hadjri )

ablack12 (approved):
Approving with last couple comments + approval from Malik

model/task/task.go
hadjri: Yes, seeing now that the field actually only was ever meant to serve EVG-17949 and its logging, so now that we've removed the log statement, let's go ahead and remove this field and its usages.

hadjri (changes requested):

model/task_lifecycle.go
```diff
@@ -273,8 +263,8 @@ func activatePreviousTask(ctx context.Context, taskId, caller string, originalSt
 			return errors.Wrap(err, "setting generated tasks to activate")
 		}
 	}
-	if stepbackDepth > 0 {
-		if err = prevTask.SetStepbackDepth(stepbackDepth); err != nil {
+	if s.LastFailingStepbackTaskId != "" {
+		if err = prevTask.SetStepbackInfo(s); err != nil {
```
hadjri, line 267: nit: setting stepback info

units/spawnhost_stop.go
```diff
@@ -50,6 +53,11 @@ func NewSpawnhostStopJob(h *host.Host, user, ts string) amboy.Job {
 	j.SetEnqueueAllScopes(true)
 	j.CloudHostModification.HostID = h.Id
 	j.CloudHostModification.UserID = user
+	j.UpdateRetryInfo(amboy.JobRetryOptions{
```
hadjri, line 56: Wrong PR :)

hadjri (commented):
Changes LGTM, just need to remove the amboy changes, and merge with main again so the spruce e2e test passes

units/spawnhost_stop.go
ZackarySantana: nice catch.... between moving in to my new place and switching up on the tickets I didn't notice!

hadjri (approved):

PR #7008, opened 2023-09-07
EVG-20799 Stream task uploaded files via evergreen proxy

khelif96:
EVG-20799

### Description
This PR introduces a new endpoint designed to act as a proxy for task-uploaded files via Evergreen. The primary goal is to enable gzipped downloads, addressing an issue with S3 where files are transmitted without compression, resulting in larger downloads for users.
**Key Benefits:**

**Improved Download Speed**: By streaming files from S3 through Evergreen, we leverage Evergreen's capability to gzip request responses. This optimization can lead to significant reductions in file size, potentially up to 95%. As a result, users will experience faster downloads and an enhanced overall user experience. 

**Custom CORS Rules**: This approach offers the added advantage of allowing us to define our own CORS (Cross-Origin Resource Sharing) rules. This flexibility ensures seamless compatibility with our web apps, eliminating the need to rely on bucket owners to configure CORS settings.

- [X] Request validation to ensure a user is requesting a valid file
- [X] <del>Validate request origin to ensure only allowed requesters (Our UI's) are downloading the file</del>
- [x] Mime type validation
- [x] Investigate if Pail can be used to perform the download
- [x] Handle Signed URLs

Crossed out goal number 2. After discussing with @bsamek and @julianedwards there didn't seem to be a good way to do this. It is fairly simple to replicate a request so this wouldn't prevent anyone from performing this request. We can feature flag this route if we notice it is causing any problems and we will also have logging to determine if there are any users that are abusing this endpoint.

### Memory concerns

> "Is there a risk here about memory on the app servers? Should there be some kind of limit? I think we're trading off speed for users against risk to the system." - Brian Samek

> I made a few changes to help guard against this. The download should not pose an additional memory risk to the app servers apart from temporarily allocating 1 MB of buffer for the download itself, AFAICT. We do not keep the response in memory it is immediately piped from the download to the request-response.

### Testing
Uploaded a 701 MB log file to an Evergreen [task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_upload_sample_log_patch_bd6166843081055dcb9ffb0da1cf64912370aad6_64f5aadbb23736caa7c9473f_23_09_04_10_01_14/files?execution=0) and attempted to curl the file directly via S3 and via the Evergreen proxy approach. 
Using the Proxy led to a 94% reduction in file size over the network and was 80% faster despite having a slower download speed. 

Benchmark script:
```


curl '<SIGNED_S3_URL or EVG PROXY ROUTE>' \
  -H 'authority: evergreen-staging.corp.mongodb.com' \
  -H 'accept: */*' \
  -H 'accept-language: en-US,en;q=0.9' \
  -H 'content-type: application/json' \
  -H 'cookie: <AUTH_TOKEN_HERE> \
  -H 'dnt: 1' \
  -H 'origin: https://spruce-staging.corp.mongodb.com' \
  -H 'referer: https://spruce-staging.corp.mongodb.com/' \
  -H 'sec-ch-ua: "Not)A;Brand";v="24", "Chromium";v="116"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36' \
  -o downloaded_file.ext \
  --compressed \
  -w "HTTP Response Code: %{http_code}\nDownload Size: %{size_download} bytes\nDownload Speed: %{speed_download} B/s\n" URL \
  > output.log 2>&1
```
Download Speed and Size Via S3:
Attempt 1: 
image

Attempt 2:
image

Download speed and size via Evergreen proxy:
Attempt 1:
image

Attempt 2:
image

_Note: Download speeds may vary based on things such as geography and internet speeds but compression ratios should remain the same_

### Documentation
Potentially don't want to document this endpoint to reduce the chance of abuse.


service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if !utility.StringMatchesAnyRegex(tFile.ContentType, uis.Settings.Ui.FileStreamingContentTypes) {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New(fmt.Sprintf("File content type: %s not supported", tFile.ContentType)))
+		return
+	}
+
+	response, err := http.Get(tFile.Link)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "Error downloading file:"))
```
julianedwards, line 683: Nit: the error message should just be `downloading file`

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		_, err := io.Copy(pw, response.Body)
+		if err != nil {
+			pw.CloseWithError(err)
+		}
+	}()
+
+	// Copy data from the pipe to the response writer
+	_, err = io.Copy(w, pr)
+	if err != nil {
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
```
julianedwards, line 706: I would use `uis.LoggedError` here and wrap the error message from `Copy`.

service/ui.go
```diff
@@ -316,6 +316,8 @@ func (uis *UIServer) GetServiceApp() *gimlet.APIApp {
 	app.AddRoute("/json/task_log/{task_id}/{execution}").Wrap(needsLogin, needsContext, viewLogs).Handler(uis.taskLog).Get()
 	app.AddRoute("/task_log_raw/{task_id}/{execution}").Wrap(needsLogin, needsContext, allowsCORS, viewLogs).Handler(uis.taskLogRaw).Get()
 
+	// Proxy downloads for task uploaded files via S3
+	app.AddRoute(("/task_file_raw/{task_id}/{execution}/{file_name}")).Wrap(needsLogin, needsContext, allowsCORS, viewLogs).Handler(uis.taskFileRaw).Get()
```
julianedwards, line 320: Do we need to add a middleware handler to make this route gzip the response?

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	taskFiles, err := artifact.GetAllArtifacts([]artifact.TaskIDAndExecution{{TaskID: projCtx.Task.Id, Execution: projCtx.Task.Execution}})
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrapf(err, "unable to find artifacts for %s", projCtx.Task.Id))
+	}
+	taskFiles, err = artifact.StripHiddenFiles(taskFiles, true)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrapf(err, "unable to strip hidden files for %s", projCtx.Task.Id))
```
julianedwards, line 663: Nit: `...files for task '%s'`

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+
+	fileName := gimlet.GetVars(r)["file_name"]
+	if fileName == "" {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New("file name not specified"))
+		return
+	}
+
+	taskFiles, err := artifact.GetAllArtifacts([]artifact.TaskIDAndExecution{{TaskID: projCtx.Task.Id, Execution: projCtx.Task.Execution}})
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrapf(err, "unable to find artifacts for %s", projCtx.Task.Id))
```
julianedwards, line 659: Nit: `...artifacts for task '%s'`

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+			tFile = &taskFile
+			break
+		}
+	}
+	if tFile == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("File not found"))
+		return
+	}
+	if !utility.StringMatchesAnyRegex(tFile.ContentType, uis.Settings.Ui.FileStreamingContentTypes) {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New(fmt.Sprintf("File content type: %s not supported", tFile.ContentType)))
```
julianedwards, line 677: Nit: `unsupported file content type '%s'`

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	response, err := http.Get(tFile.Link)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "Error downloading file:"))
+		return
+
+	}
+	if response.StatusCode != http.StatusOK {
+		uis.LoggedError(w, r, response.StatusCode, errors.New(fmt.Sprintf("Failed to download file. Status code: %d", response.StatusCode)))
```
julianedwards, line 688: Nit: `errors.Errorf("failed to download file with status code %d", response.StatusCode)`

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	if tFile == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("File not found"))
+		return
+	}
+	if !utility.StringMatchesAnyRegex(tFile.ContentType, uis.Settings.Ui.FileStreamingContentTypes) {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New(fmt.Sprintf("File content type: %s not supported", tFile.ContentType)))
+		return
+	}
+
+	response, err := http.Get(tFile.Link)
```
julianedwards, line 681: Won't this support non-S3 links? Is that something we want now?

julianedwards (changes requested):
Does this need tests?

service/ui.go
khelif96: Nope, this happens automatically!

service/task.go
khelif96: I don't think it will ever reach this point. Since we perform a check on `content_type` above and that value [only gets set for files uploaded](https://github.com/evergreen-ci/evergreen/pull/6998) with s3.put it should never happen for non-s3 links.

service/task.go
```diff
@@ -640,6 +641,74 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if response.StatusCode != http.StatusOK {
+		uis.LoggedError(w, r, response.StatusCode, errors.Errorf("failed to download file with status code: %d", response.StatusCode))
+		return
+	}
+
+	pr, pw := io.Pipe()
+
+	// Create a goroutine to copy data from the body to the pipe
+	go func() {
```
julianedwards, line 695: Add a defer statement with panic recovery like [here](https://github.com/evergreen-ci/evergreen/blob/ef5e4e0f90c3e4ff92c55bc963b24b4731fe889f/model/log/chunk_iterator.go#L141-L144).

julianedwards: Actually, per our conversation offline, this goroutine may not be necessary. Let me know how it goes when just copying the response body directly into the response writer!

service/task.go
```diff
@@ -640,6 +641,74 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+
+	fileName := gimlet.GetVars(r)["file_name"]
+	if fileName == "" {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New("file name not specified"))
+		return
+	}
+
+	taskFiles, err := artifact.GetAllArtifacts([]artifact.TaskIDAndExecution{{TaskID: projCtx.Task.Id, Execution: projCtx.Task.Execution}})
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrapf(err, "unable to find artifacts for task: %s", projCtx.Task.Id))
```
bsamek, line 659: Should there be a `return` after this?

service/task.go
```diff
@@ -640,6 +641,74 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	}
+	taskFiles, err = artifact.StripHiddenFiles(taskFiles, true)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrapf(err, "unable to strip hidden files for task: %s", projCtx.Task.Id))
```
bsamek, line 663: Same question about `return`

service/task.go
```diff
@@ -640,6 +641,74 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		if taskFile.Name == fileName {
+			tFile = &taskFile
+			break
+		}
+	}
+	if tFile == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("File not found"))
+		return
+	}
+	if !utility.StringMatchesAnyRegex(tFile.ContentType, uis.Settings.Ui.FileStreamingContentTypes) {
```
bsamek, line 676: Should this really match a regex? I imagine it's a small enough number that it doesn't super matter, but I would expect this to have been an exact string match.

service/task.go
```diff
@@ -640,6 +641,74 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New(fmt.Sprintf("unsupported file content type: %s", tFile.ContentType)))
+		return
+	}
+
+	response, err := http.Get(tFile.Link)
```
bsamek, line 681: Is there a risk here about memory on the app servers? Should there be some kind of limit? I think we're trading off speed for users against risk to the system.

bsamek (changes requested):

bsamek:
This should likely have an app reviewer. I added @ablack12 explicitly.

service/task.go
khelif96: This allows us to support partial matches https://docs.google.com/document/d/10qJmO3lkMvUBpOApXu9ovy9m0ARBQGbaOvoFUbRW6FE/edit?disco=AAAA21_7eZc

service/task.go
khelif96: I made a few changes to help guard against this. The download should not pose an additional memory risk to the app servers apart from temporarily allocating 1 MB of buffer for the download itself, AFAICT. We do not keep the response in memory it is immediately piped from the download to the request-response.

service/task.go
khelif96: Thanks removed this and actually replaced it with a buffered read.

service/task.go
bsamek: Ah, that makes sense. Would all these cases be prefix matches, in which case you could use `strings.HasPrefix()`?

service/task.go
```diff
@@ -640,6 +641,81 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	// Create a buffer to stream the file in chunks
+	bufferSize := 1000000 // 1mb
+	buffer := make([]byte, bufferSize)
+
+	w.Header().Set("Content-Type", tFile.ContentType)
+	// Stream the file content to the response writer
+	for {
+		n, err := response.Body.Read(buffer)
+		if err == io.EOF {
+			break // End of file
```
bsamek, line 704: nit: this is slightly easier to read as a return

service/task.go
```diff
@@ -640,6 +641,81 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		}
+		if err != nil {
+			uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "reading file:"))
```
bsamek, line 707: nit: colons are not conventional errors.Wrap()

service/task.go
```diff
@@ -640,6 +641,81 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		_, err = w.Write(buffer[:n])
+		if err != nil {
+			uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "writing to response:"))
```
bsamek, line 712: same comment about colons

bsamek (approved):
lgtm with a followup question and some nits

service/task.go
```diff
@@ -640,6 +641,73 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+func (uis *UIServer) taskFileRaw(w http.ResponseWriter, r *http.Request) {
+	projCtx := MustHaveProjectContext(r)
+	if projCtx.Task == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("task not found"))
+		return
+	}
+
+	fileName := gimlet.GetVars(r)["file_name"]
+	if fileName == "" {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New("file name not specified"))
```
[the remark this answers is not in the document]
julianedwards: I think you forgot to change this to call `http.Error`.

service/task.go
julianedwards: Nit: `...artifacts for task '%s'`

service/task.go
julianedwards: Nit: `...hidden files for task '%s'`

service/task.go
```diff
@@ -640,6 +641,81 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		return
+
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		uis.LoggedError(w, r, response.StatusCode, errors.Errorf("failed to download file with status code: %d", response.StatusCode))
+		return
+	}
+
+	// Create a buffer to stream the file in chunks
```
julianedwards, line 695: Nit: comments should be proper sentences with periods.

service/task.go
julianedwards: I think the `End of file` comment is unnecessary since the if statement is already checking for an `io.EOF` error.

service/task.go
```diff
@@ -640,6 +641,81 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	bufferSize := 1000000 // 1mb
+	buffer := make([]byte, bufferSize)
+
+	w.Header().Set("Content-Type", tFile.ContentType)
+	// Stream the file content to the response writer
+	for {
```
julianedwards, line 701: Could you use [io.CopyBuffer](https://pkg.go.dev/io#CopyBuffer) here instead handling the reading/buffering yourself?

julianedwards (changes requested):

service/task.go
khelif96: Good call-out thank you!

service/task.go
```diff
@@ -640,6 +641,70 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+			tFile = &taskFile
+			break
+		}
+	}
+	if tFile == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("File not found"))
+		return
+	}
+	if !utility.StringMatchesAnyRegex(tFile.ContentType, uis.Settings.Ui.FileStreamingContentTypes) {
+		uis.LoggedError(w, r, http.StatusBadRequest, errors.New(fmt.Sprintf("unsupported file content type: %s", tFile.ContentType)))
```
julianedwards, line 679: Nit: `unsupported file content type '%s'`

service/task.go
```diff
@@ -640,6 +641,70 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	response, err := http.Get(tFile.Link)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "downloading file:"))
```
julianedwards, line 685: Nit: remove colon

service/task.go
```diff
@@ -640,6 +641,70 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+	}
+	var tFile *artifact.File
+	for _, taskFile := range taskFiles {
+		if taskFile.Name == fileName {
+			tFile = &taskFile
+			break
+		}
+	}
+	if tFile == nil {
+		uis.LoggedError(w, r, http.StatusNotFound, errors.New("File not found"))
```
julianedwards, line 675: Nit: `file not found`

julianedwards (approved):
LGTM % nits

service/task.go
```diff
@@ -640,6 +642,78 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+
+	response, err := http.Get(tFile.Link)
+	if err != nil {
+		uis.LoggedError(w, r, http.StatusInternalServerError, errors.Wrap(err, "downloading file"))
+		return
+
+	}
+	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		uis.LoggedError(w, r, response.StatusCode, errors.Errorf("failed to download file with status code: %d", response.StatusCode))
```
ablack12, line 700: can we return some of the error as well or nah?

service/task.go
```diff
@@ -640,6 +642,78 @@ func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	// Create a buffer to stream the file in chunks.
+	bufferSize := 1000000 // 1mb
```
ablack12, line 705: nit: can this be a local const

service/ui.go
ablack12, line 320: is it possible for filenames to be wonky such that a query parameter would be chaotic? (we could have this be part of the body instead if so)

ablack12 (approved):
Looks good to me -- I don't know a lot about the memory risk but the buffer description seems reasonable to me.

Agreed with not documenting on the docs site.

service/task.go
khelif96: Done!

service/task.go
khelif96: I think we wouldn't want to. If for whatever reason the response body is super large I wouldn't want to flush it to memory just to print it out.

service/ui.go
khelif96: Nope as long as the URLs are properly encoded by whatever service is making the request this isn't an issue. The HTTP client handles decoding them automatically.

service/task.go
khelif96: I also added this thread to the PR description to make it easier if anyone wants to reference this in the future.

PR #7009, opened 2023-09-07
EVG-19914 remove duplicate patch statuses and deprecate patch success

ablack12:
EVG-19914
### Description
Contains a lot of the same changes as the [original PR](https://github.com/evergreen-ci/evergreen/pull/6567/files): patch and version statuses are duplicated (except for success) which is a huge headache (so is deprecating it evidently but it might be worth it).

This PR removes the easy duplicates, marks "succeeded" as deprecated, and reads both statuses across the code. The next PR will write the new status to the DB (but this PR should ensure that we have backwards compatibility in case we need to revert). Sophie already updated the UI to be able to read both statuses for success as well. Then in EVG-20032 (after the change has been in for a while) we can remove the legacy status entirely.

### Testing
Ensure existing tests pass, that a task can be created and successful and etc.

### Documentation
Not relevant until we actually update the status we're writing


globals.go
```diff
@@ -164,15 +164,10 @@ const (
 	VersionFailed    = "failed"
 	VersionSucceeded = "success"
 
-	PatchCreated     = "created"
-	PatchStarted     = "started"
-	PatchSucceeded   = "succeeded"
-	PatchFailed      = "failed"
-	PatchAllOutcomes = "*"
+	LegacyPatchSucceeded = "succeeded" // deprecated
```
hadjri, line 167: Can we mention EVG-20032 here?

globals.go
```diff
@@ -428,48 +427,36 @@ func IsValidTaskEndStatus(status string) bool {
...
-
 func IsFinishedBuildStatus(status string) bool {
 	return status == BuildFailed || status == BuildSucceeded
 }
 
+// IsFinishedVersionStatus returns true if the version or patch is true.
+// Also handles the legacy status, to be removed in EVG-20032.
 func IsFinishedVersionStatus(status string) bool {
-	return status == VersionFailed || status == VersionSucceeded
+	return status == VersionFailed || status == VersionSucceeded || status == LegacyPatchSucceeded
```
hadjri, line 437: Can we change to `return status == VersionFailed || IsSuccessfulVersionStatus(status)`?

trigger/commit_queue.go
```diff
@@ -109,7 +109,7 @@ func (t *commitQueueTriggers) makeData(sub *event.Subscription) (*commonTemplate
 	}
 
 	slackColor := evergreenFailColor
-	if t.data.Status == evergreen.PatchSucceeded || t.data.Status == evergreen.MergeTestStarted {
+	if t.data.Status == evergreen.LegacyPatchSucceeded || t.data.Status == evergreen.MergeTestStarted {
```
hadjri, line 112: We'd want to use `IsSuccessfulVersionStatus` here right?

agent/command/git_merge_pr.go
```diff
@@ -81,11 +81,11 @@ func (c *gitMergePR) Execute(ctx context.Context, comm client.Communicator, logg
 		return errors.Wrap(err, "setting up GitHub status logger")
 	}
 
-	status := evergreen.PatchFailed
-	if patchDoc.MergeStatus == evergreen.PatchSucceeded {
-		status = evergreen.PatchSucceeded
+	status := evergreen.VersionFailed
+	if evergreen.IsSuccessfulVersionStatus(patchDoc.MergeStatus) {
```
hadjri, line 85: We can just simplify this to `if !evergreen.IsSuccessfulVersionStatus(patchDoc.MergeStatus) { // do logging }` and return nil here

model/patch/patch.go
```diff
@@ -1273,15 +1273,15 @@ func GetCollectiveStatusFromPatchStatuses(statuses []string) string {
...
+		return evergreen.VersionCreated
 	} else if hasFailure {
-		return evergreen.PatchFailed
+		return evergreen.VersionFailed
 	} else if hasAborted {
-		return evergreen.PatchAborted
+		return evergreen.VersionAborted
 	} else if hasSuccess {
-		return evergreen.PatchSucceeded
+		return evergreen.LegacyPatchSucceeded
```
hadjri, line 1284: Do we want to return VersionSucceeded here instead?

rest/route/agent.go
```diff
@@ -1211,7 +1211,7 @@ func (h *gitServePatchHandler) Run(ctx context.Context) gimlet.Responder {
 			return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "finding tasks for version"))
 		}
 
-		status := evergreen.PatchSucceeded
+		status := evergreen.LegacyPatchSucceeded
```
hadjri, line 1214: Same q: should we change this to VersionSucceeded?

hadjri (changes requested):

model/patch/patch.go
ablack12: I'd like to stick with returning the legacy value right now, in case this value is written to the DB anywhere. Writing with the new status is what I'd like to address in the next PR.

trigger/commit_queue.go
ablack12: good catch

model/patch/patch.go
hadjri: SGTM

hadjri (approved):

PR #7011, opened 2023-09-07
EVG-20594: Replace DefaultTaskActivator with more accurate names

hadjri:
EVG-20594

### Description

- Deprecating the `DefaultTaskActivator ` while keeping it as a valid system activator for backwards compatibility purposes
- Added a more specifcally named activator for all internal non-userID activations that were using the `DefaultTaskActivator` before
- Add an explicit requirement for user ID retrieval during patch scheduling to use for the activator name, which shouldn't be an issue since users must be signed in to do this

### Testing
Existing tests passing


globals.go
```diff
@@ -213,9 +213,29 @@ const (
 	// app server should log to stdout.
 	standardOutputLoggingOverride = "STDOUT"
 
-	DefaultTaskActivator   = ""
-	StepbackTaskActivator  = "stepback"
+	// DefaultTaskActivator is a deprecated legacy activator that used
```
ablack12, line 216: should we make a ticket so we don't forget to get rid of this at some point? Presumably we can drop it a year from now

globals.go
```diff
@@ -213,9 +213,29 @@ const (
...
+	// to be a majority of non-stepback and non-API activations.
+	DefaultTaskActivator = ""
+	// APIServerTaskActivator represents Evergreen's internal API activator
 	APIServerTaskActivator = "apiserver"
+	// StepbackTaskActivator represents the activator for tasks activated
+	// due to stepback.
+	StepbackTaskActivator = "stepback"
```
ablack12, line 223: would it be chaotic to have all of these say activator at the end? Or does that seem redundant for the use cases where we see these?

globals.go
```diff
@@ -213,9 +213,29 @@ const (
 	// app server should log to stdout.
 	standardOutputLoggingOverride = "STDOUT"
 
-	DefaultTaskActivator   = ""
-	StepbackTaskActivator  = "stepback"
+	// DefaultTaskActivator is a deprecated legacy activator that used
+	// to be a majority of non-stepback and non-API activations.
+	DefaultTaskActivator = ""
```
ablack12, line 218: can we change the name of this to LegacyTaskActivator so new code knows not to use it?

model/version_activation.go
```diff
@@ -120,7 +120,7 @@ func ActivateElapsedBuildsAndTasks(v *Version) (bool, error) {
 	}
 	if len(buildIdsToActivate) > 0 {
 		// Don't need to set the version in here since we do it ourselves in a single update
-		if err := build.UpdateActivation(buildIdsToActivate, true, evergreen.DefaultTaskActivator); err != nil {
+		if err := build.UpdateActivation(buildIdsToActivate, true, evergreen.BuildActivator); err != nil {
```
ablack12, line 123: I'm actually maybe confused about the difference between the activation here and below but is it confusing to have two different activator names here?

ablack12: Oh it seems like we activate the build if either the whole build or individual tasks are being activated. Maybe for traceability it's fine to leave this as is.

globals.go
hadjri: Created https://jira.mongodb.org/browse/EVG-20869

globals.go
hadjri: I think it's a good suggestion, it'll help to be more explicit with our naming and make it clear that the only non-users that activate things in Evergreen must be "activators". Added

ablack12 (approved):
Sorry for the massive review delay!

PR #7012, opened 2023-09-08
EVG-20532 Create getter and setter for branch protection

bsamek:
EVG-20532

### Description

This creates a getter and setter for branch protection for the merge queue. However, I'm unsure how to use these, so I wanted to open an early PR to get opinions. Here are some options:

1. Hook these functions into the UI. When someone turns on the merge queue, if they haven't set "evergreen" branch protection, set it for them, with a warning in a modal.
2. Add these functions to the CLI under `evergreen admin`. When we migrate people, let them know we're going to set a branch protection rule. Set it for them.
3. Ask people to set a branch protection rule themselves.

The argument for 3 is the risk of doing this. Unlike setting a webhook, it's possible we might inadvertently remove someone's branch protection if there is a bug.

### Testing

This has not been tested yet.

### Documentation

Will add docs once we decide how we will wire this up.

### DIFF
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -1768,35 +1768,12 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 // HasEvergreenBranchProtections checks if any branch protection rule begins "evergreen".
 // We deliberately check for the prefix because users may have installed variant-level branch protections.
 func HasEvergreenBranchProtection(ctx context.Context, token, owner, repo, branch string) (bool, error) {
-	caller := "HasEvergreenBranchProtection"
-	ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(
-		attribute.String(githubEndpointAttribute, caller),
-		attribute.String(githubOwnerAttribute, owner),
-		attribute.String(githubRepoAttribute, repo),
-		attribute.String(githubRefAttribute, branch),
-	))
-	defer span.End()
-
-	if token == "" {
-		var err error
-		token, err = getInstallationToken(ctx, owner, repo, nil)
-		if err != nil {
-			return false, errors.Wrap(err, "getting installation token")
-		}
-	}
-	githubClient := getGithubClient(token, caller, retryConfig{})
-
-	protection, resp, err := githubClient.Repositories.GetBranchProtection(ctx, owner, repo, branch)
-	if resp != nil {
-		defer resp.Body.Close()
-		span.SetAttributes(attribute.Bool(githubCachedAttribute, respFromCache(resp.Response)))
-	}
+	rules, err := GetEvergreenBranchProtectionRules(ctx, token, owner, repo, branch)
 	if err != nil {
-		return false, errors.Wrap(err, "getting branch protection")
+		return false, errors.Wrap(err, "getting branch protection rules")
 	}
-
-	for _, check := range protection.GetRequiredStatusChecks().Checks {
-		if strings.HasPrefix(check.Context, "evergreen") {
+	for _, rule := range rules {
+		if strings.HasPrefix(rule, "evergreen") {
 			return true, nil
 		}
 	}
--- a/thirdparty/github.go
+++ b/thirdparty/github.go
@@ -1805,8 +1805,9 @@ func HasEvergreenBranchProtection(ctx context.Context, token, owner, repo, branc
 }
 
 // InstallEvergreenBranchProtections installs "evergreen" branch protection.
-func InstallEvergreenBranchProtections(ctx context.Context, token, owner, repo, branch string) error {
-	caller := "InstallEvergreenBranchProtections"
+// We deliberately check for the prefix because users may have installed variant-level branch protections.
+func InstallEvergreenBranchProtection(ctx context.Context, token, owner, repo, branch string) error {
+	caller := "InstallEvergreenBranchProtection"
 	ctx, span := tracer.Start(ctx, caller, trace.WithAttributes(
 		attribute.String(githubEndpointAttribute, caller),
 		attribute.String(githubOwnerAttribute, owner),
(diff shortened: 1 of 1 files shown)


bsamek:
@ablack12 @malikchaya2 @minnakt I added you as reviewers, since you've all typed merge queue code or done a code review, and I'm not sure what the right approach for migration should be. I'm curious if you have thoughts about the right approach.

thirdparty/github.go
```diff
@@ -1765,6 +1765,104 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 	return pr, nil
 }
```
ablack12, line 1766: Starting a random thread for this so we can add comments in one place:
>@ablack12 @malikchaya2 @minnakt I added you as reviewers, since you've all typed merge queue code or done a code review, and I'm not sure what the right approach for migration should be. I'm curious if you have thoughts about the right approach.

I think what you've said about going the route of (3) is reasonable given the risk. Maybe we let users know (with a warning message or something, Minna would know better) if they toggle the merge queue on that it won't be effective without the Evergreen branch protection? Or maybe best scenario, we don't allow them to turn it on if they don't have the check in place.

thirdparty/github.go
```diff
@@ -1765,6 +1765,104 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
...
 
+// HasEvergreenBranchProtections checks if any branch protection rule begins "evergreen".
+// We deliberately check for the prefix because users may have installed variant-level branch protections.
+func HasEvergreenBranchProtection(ctx context.Context, token, owner, repo, branch string) (bool, error) {
```
ablack12, line 1770: I'm seeing you already wrote [GetEvergreenBranchProtection](https://github.com/evergreen-ci/evergreen/blob/5df5c07ddd9bd75953acdefd3c9b6acaabeee319/thirdparty/github.go#L2016-L2015) -- can we just adapt that one?

ablack12: I have a mild preference for this just returning the branch protection so it can also just be used in the install function( if we keep it) as opposed to duplicating that logic

bsamek: I removed the duplicate logic. I suspect we do want a function that returns a boolean. This now just wraps the other function.

thirdparty/github.go
bsamek: Oh I didn't consider a hybrid approach. We could check to see if they have a branch protection rule, and then if they don't, we could offer to try to set it for them. This would mean that they'd be explicitly opting in to letting Evergreen mess with their settings.

ablack12: I like that approach!

thirdparty/github.go
```diff
@@ -1765,6 +1765,81 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
 	return pr, nil
 }
 
+// HasEvergreenBranchProtections checks if any branch protection rule begins "evergreen".
+// We deliberately check for the prefix because users may have installed variant-level branch protections.
```
ablack12, line 1769: To verify, is variant-level branch protection enough? For the CQ results then would we only consider the outcome of those variants or does the whole cq patch have to be successful? (apologies if i've asked you this already haha)

thirdparty/github.go
```diff
@@ -1765,6 +1765,81 @@ func GetMergeablePullRequest(ctx context.Context, issue int, githubToken, owner,
...
+func HasEvergreenBranchProtection(ctx context.Context, token, owner, repo, branch string) (bool, error) {
+	rules, err := GetEvergreenBranchProtectionRules(ctx, token, owner, repo, branch)
+	if err != nil {
+		return false, errors.Wrap(err, "getting branch protection rules")
+	}
+	for _, rule := range rules {
```
ablack12, line 1775: GetEvergreenBranchProtectionRules already checks the Evergreen prefix, and so the only rules returned here are those. 

That being said, I have a preference for keeping these third party functions really simple -- they make the Github client, call the Github function and handle it, and return the result. Then we can use the result however we want (in helper functions within thirdparty and outside, so in this function we could keep the logic of iterating through the rules). If we were to do that, then we could also call this hypothetical GetBranchProtection function in the Install function, and then just loop through the checks.

ablack12 (changes requested):

thirdparty/github.go
bsamek: True. I think in that case we can remove this function entirely. Removed.

thirdparty/github.go
bsamek: We're not checking the branch protection rules in order to decide whether or not to merge a PR. GitHub does that. We're checking to make sure that the user has set at least one branch protection rule that begins "evergreen". Users would either want the whole build to finish, so they would use "evergreen", or they would only want a variant (Cloud wants this), in which case they would use "evergreen/<variant>".

thirdparty/github.go
ablack12: okay sorry to keep commenting on https://github.com/evergreen-ci/evergreen/pull/7012 i think i’m not explaining myself well — the thing i’m trying to do away with is that both the Install and the Get function set up and call GetBranchProtection https://github.com/evergreen-ci/evergreen/blob/5df5c07ddd9bd75953acdefd3c9b6acaabeee319/thirdparty/github.go#L2035 

I feel like ideally this code would be written only once and would just return the result of GetBranchProtection (or return all of the checks at least) so then both the Install and Get functions would be able to use the result. I'm going to go ahead and approve though since the gist works.

ablack12 (approved):

ablack12:
LGTM but want to highlight new comment: https://github.com/evergreen-ci/evergreen/pull/7012#discussion_r1323549039

bsamek:
After discussing with @minnakt, I'd like to hold off on merging this PR. There are some complexities around installing installing branch protections for users. Instead, I'd like to get feedback from early adopters to figure out the best approach. I'm going to close this PR for now.

PR #7014, opened 2023-09-08
EVG-20701: add timeout to heartbeat

Kimchelly:
EVG-20701

### Description
I added a mechanism to time out the heartbeat if the task is running for a suspiciously long time past the timeout it should be obeying (i.e. so long that the task is most likely permanently stuck). It doesn't actually stop the heartbeat for now, because it's somewhat risky. I want to give it a bit of time to run tasks (and log if it hits the heartbeat timeout) to ensure it's not somehow going to spuriously hit the heartbeat timeout during normal task runtime for a reason that I didn't think of. There will be a follow-up PR to uncomment the bit of logic that actually stops the heartbeat after it's run in prod for a few days.

### Testing
* Added unit tests for heartbeat timeout.
* Tested in staging to verify that the heartbeat timeout log appeared (in Splunk) when I manually logged it, but otherwise did not log during normal task operation.

### Documentation
Wrote an ADR.


agent/agent.go
```diff
@@ -665,6 +677,17 @@ func (a *Agent) runPreAndMain(ctx context.Context, tc *taskContext) (status stri
 	}
 	go a.startTimeoutWatcher(timeoutWatcherCtx, execTimeoutCancel, timeoutOpts)
 
+	tc.logger.Execution().Infof("Setting heartbeat timeout to type '%s'.", execTimeout)
```
Kimchelly, line 680: I intentionally put the set/reset for the heartbeat timeout in the foreground rather than absorb it into `startTimeoutWatcher` goroutine because if it runs in the background, the heartbeat set/reset could race between background timers (and set the wrong timeout value).

agent/task_context.go
```diff
@@ -194,11 +229,12 @@ func (tc *taskContext) getExecTimeout() time.Duration {
...
-	block       command.BlockType
-	commands    *model.YAMLCommandSet
-	timeoutKind timeoutType
-	getTimeout  func() time.Duration
-	canFailTask bool
+	block               command.BlockType
+	commands            *model.YAMLCommandSet
+	timeoutKind         timeoutType
+	getTimeout          func() time.Duration
+	canTimeoutHeartbeat bool
```
ablack12, line 236: canHeartbeatTimeout?  unless i'm confused about what this is

agent/task_context.go
```diff
@@ -194,11 +229,12 @@ func (tc *taskContext) getExecTimeout() time.Duration {
...
 type commandBlock struct {
-	block       command.BlockType
-	commands    *model.YAMLCommandSet
-	timeoutKind timeoutType
-	getTimeout  func() time.Duration
-	canFailTask bool
+	block               command.BlockType
+	commands            *model.YAMLCommandSet
+	timeoutKind         timeoutType
+	getTimeout          func() time.Duration
```
ablack12, line 235: should we just embed heartbeatTimeoutOptions in here ?

ablack12 (commented):
Wanted to get a review started here but I think I'm missing a piece of what's going on here

agent/background.go
```diff
@@ -33,9 +36,39 @@ func (a *Agent) startHeartbeat(ctx context.Context, preAndMainCancel context.Can
...
 		case <-ticker.C:
+			if tc.hadHeartbeatTimeout() {
+				// If the heartbeat hits its maximum reasonable timeout, that
+				// means something unrecoverably wrong has most likely occurred
+				// in the task runtime (e.g. a deadlock in the agent has stalled
+				// the task).
+				timeoutOpts := tc.getHeartbeatTimeout()
+				timeout := timeoutOpts.getTimeout()
+				if !loggedTimeout {
+					msg := fmt.Sprintf("Heartbeat has hit maximum allowed '%s' timeout of %s, task is at risk of timing out if it runs for much longer.", timeoutOpts.kind, timeout.String())
```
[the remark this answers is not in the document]
Kimchelly: Done.

agent/background.go
```diff
@@ -33,9 +36,39 @@ func (a *Agent) startHeartbeat(ctx context.Context, preAndMainCancel context.Can
...
+					grip.Alert(message.Fields{
```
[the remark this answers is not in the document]
Kimchelly: It's not necessary, but it would make filtering the Splunk logs slightly easier. The message includes this info solely because it's going to be for the execution log sent to the user in the follow-up PR.

agent/background.go
```diff
@@ -33,9 +36,39 @@ func (a *Agent) startHeartbeat(ctx context.Context, preAndMainCancel context.Can
...
+						"task_id":        tc.taskConfig.Task.Id,
+						"task_execution": tc.taskConfig.Task.Execution,
+						"timeout_type":   timeoutOpts.kind,
+						"timeout_start":  timeoutOpts.startAt,
+						"timeout":        timeout,
+					})
+					loggedTimeout = true
+				}
+				// TODO (EVG-20701): uncomment the timeout handling once there
+				// is sufficient confidence that this case is never hit during
```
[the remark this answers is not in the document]
Kimchelly: Logging alert and higher in the agent logs to Splunk.

agent/global.go
```diff
@@ -28,6 +28,10 @@ const (
 	// heartbeat to API server.
 	defaultHeartbeatInterval = 30 * time.Second
 
+	// defaultHeartbeatTimeout is how long the agent can perform operations when
```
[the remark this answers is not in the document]
Kimchelly: I intentionally made this short because the default 6 hour exec timeout is too generous a timeout. This is only meant to fill in the gaps where the task is in between commands, so there's no explicit user-configurable timeout that it can use. If it's using this default, tasks have 8 minutes (= 1 minute + 7 minutes grace period) to do anything, which is fairly generous, given that the task shouldn't be kept waiting for more than a couple seconds between commands.

agent/task_context.go
Kimchelly: Hmm, we can't because 1. the exec timeout works in a different way and 2. timeout options are required for the regular (non-heartbeat) timeout mechanism, but are only conditionally required if the timeout ticker can time out the heartbeat as well.

agent/task_context.go
Kimchelly: I'm using time out as a verb instead of noun, but I can change it to `TimeOut` to make it clearer.

PR #7015, opened 2023-09-08
EVG-20612 avoid querying if we don't need last activation time

ablack12:
EVG-20612
### Description
1) Fix the main problem of running the query when there's no reason to
2) Log the stack (with string so it logs correctly) to understand why we'd be doing this in the first place
3) Log how long the query takes in the case that we _are_ supposed to be running it

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2279,6 +2279,20 @@ func (p *ProjectRef) GetActivationTimeForVariant(variant *BuildVariant) (time.Ti
 // Temporarily takes in the task ID that prompted this query, for logging.
 func (p *ProjectRef) GetActivationTimeForTask(t *BuildVariantTaskUnit, taskId string) (time.Time, error) {
 	defaultRes := time.Now()
+	// Verify that we mean to be getting activation time for task
+	if !t.HasSpecificActivation() {
+		grip.Debug(message.Fields{
+			"ticket":         "EVG-20612",
+			"message":        "incorrectly called GetActivationTimeForTask",
+			"task_id":        taskId,
+			"variant":        t.Variant,
+			"task_name":      t.Name,
+			"bvtu_batchtime": t.BatchTime,
+			"bvtu_activate":  t.Activate,
+			"stack":          string(debug.Stack()),
+		})
+		return defaultRes, nil
+	}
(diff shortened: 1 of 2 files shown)


repotracker/repotracker.go
```diff
@@ -931,6 +931,7 @@ func createVersionItems(ctx context.Context, v *model.Version, metadata model.Ve
 			// add only tasks that require activation times
 			for _, bvt := range buildvariant.Tasks {
 				tId, ok := taskNameToId[bvt.Name]
+				// TODO: remove activation check
```
ablack12, line 934: Keeping this activation check in for now so that we only log the debug statement for the cases I didn't already know about; will remove this when I remove the logging

model/project_ref.go
```diff
@@ -2279,6 +2279,20 @@ func (p *ProjectRef) GetActivationTimeForVariant(variant *BuildVariant) (time.Ti
...
 	defaultRes := time.Now()
+	// Verify that we mean to be getting activation time for task
+	if !t.HasSpecificActivation() {
+		grip.Debug(message.Fields{
+			"ticket":         "EVG-20612",
+			"message":        "incorrectly called GetActivationTimeForTask",
+			"task_id":        taskId,
+			"variant":        t.Variant,
+			"task_name":      t.Name,
+			"bvtu_batchtime": t.BatchTime,
```
Kimchelly, line 2290: Would logging `CronBatchTime` help as well since it's part of `HasSpecificActivation`?

repotracker/repotracker.go
```diff
@@ -931,6 +931,7 @@ func createVersionItems(ctx context.Context, v *model.Version, metadata model.Ve
 			// add only tasks that require activation times
 			for _, bvt := range buildvariant.Tasks {
 				tId, ok := taskNameToId[bvt.Name]
+				// TODO EVG-20162: remove activation check
```
Kimchelly, line 934: EVG-20612?

Kimchelly (approved):
LGTM mod non-blocking suggestions.

repotracker/repotracker.go
ablack12: so true

PR #7017, opened 2023-09-08
EVG-19715 terminate oudated AMI hosts faster

ablack12:
EVG-19715
### Description
We discussed in the ticket terminating idle hosts more aggressively. Right now we wait for a minute, which is a long time if the queue isn't empty / has a lot of tasks that started. I arbitrarily dropped it to 5 seconds.

If we don't think this is enough, then we could alternatively mark all hosts as decommissioned when an AMI is updated.

### DIFF
--- a/units/host_monitoring_idle_termination.go
+++ b/units/host_monitoring_idle_termination.go
@@ -24,7 +24,7 @@ const (
 	idleWaitingForAgentCutoff = 10 * time.Minute
 
 	// outdatedIdleTimeCutoff is the amount of time we wait for an outdated idle host to be marked idle.
-	outdatedIdleTimeCutoff = time.Minute
+	outdatedIdleTimeCutoff = 10 * time.Second
 	// MaxTimeNextPayment is the amount of time we wait to have left before marking a host as idle
 	maxTimeTilNextPayment = 5 * time.Minute
 )
--- a/units/host_monitoring_idle_termination.go
+++ b/units/host_monitoring_idle_termination.go
@@ -168,17 +168,12 @@ func (j *idleHostJob) checkAndTerminateHost(ctx context.Context, schedulerConfig
 	if idleThreshold == 0 {
 		idleThreshold = time.Duration(schedulerConfig.AcceptableHostIdleTimeSeconds) * time.Second
 	}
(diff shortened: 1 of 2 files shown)


units/host_monitoring_idle_termination.go
```diff
@@ -24,7 +24,7 @@ const (
 	idleWaitingForAgentCutoff = 10 * time.Minute
 
 	// outdatedIdleTimeCutoff is the amount of time we wait for an outdated idle host to be marked idle.
-	outdatedIdleTimeCutoff = time.Minute
+	outdatedIdleTimeCutoff = 5 * time.Second
```
bsamek, line 27: Would it make sense to make this 0 (and change the `>=`s to `>`?

bsamek (approved):
lgtm with a question

units/host_monitoring_idle_termination.go
ablack12: Like just don't wait any amount of time? That would make sense if we want to be super aggressive about the termination, is that what you're thinking?

bsamek: I guess I'm not sure what the difference really is between 0 and 5 seconds. Do you have a sense how fast nextTask is? Like will 5 seconds actually allow nextTask to run?

units/host_monitoring_idle_termination.go
```diff
@@ -174,12 +172,15 @@ func (j *idleHostJob) checkAndTerminateHost(ctx context.Context, schedulerConfig
...
-		idleThreshold = outdatedIdleTimeCutoff
+		// Since tasks created after the AMI is updated will only run on new hosts,
+		// we want to terminate outdated hosts aggressively to ensure we're respecting task priorities.
+		idleThreshold = 0
 	}
 
-	// if we haven't heard from the host or it's been idle for longer than the cutoff, we should terminate
 	var terminateReason string
-	if hasOutdatedAMI {
+	if hasOutdatedAMI && communicationTime >= idleThreshold {
```
bsamek, line 181: Is `communicationTime >= idleThreshold` redundant, since `idleThreshold` will always be 0 if `hasOutdatedAMI` is true?

bsamek (approved):

PR #7024, opened 2023-09-11
docs: Add bash warning for SSH availability check

bsamek:


bsamek:
@ablack12 FYI this comes out of https://mongodb.slack.com/archives/C0V896UV8/p1694462193008229

docs/Project-Configuration/Project-Commands.md
```diff
@@ -540,7 +540,14 @@ permissions:
 Certain instances require more time for SSH access to become available.
 If the user plans to execute commands on the remote host, then waiting
 for SSH access to become available is mandatory. Below is an Evergreen
-function that probes for SSH connectivity:
+function that probes for SSH connectivity.
+
+Note, however, an important bash caveat! By default Evergreen implements
```
Kimchelly, line 545: I feel like someone reading this without reading the entire original thread + having some pre-existing bash knowledge (+ knowing that the default shell for `shell.exec` is _not_ bash) won't understand what they're supposed to take away from this warning. It feels like this either needs its own YAML example separate from the one described above or apply the real fix to the `script` example below and explain why there.

Kimchelly (changes requested):

docs/Project-Configuration/Project-Commands.md
bsamek: I edited the example. I also edited some of the "bash" to read "shell".

PR #7025, opened 2023-09-11
EVG-20177: Add GQL field canUpdateAdminSettings to Permissions type

SupaJoon:
EVG-20177

### Description
This field will determine if a link to the EVG Admin Settings page is available from Spruce. 
I updated the hidden file setting because it was marking resolver files as generated.

### DIFF
--- a/graphql/permissions_resolver.go
+++ b/graphql/permissions_resolver.go
@@ -43,7 +43,17 @@ func (r *permissionsResolver) CanCreateProject(ctx context.Context, obj *Permiss
 
 // CanEditAdminSettings is the resolver for the canEditAdminSettings field.
 func (r *permissionsResolver) CanEditAdminSettings(ctx context.Context, obj *Permissions) (bool, error) {
-	panic(fmt.Errorf("not implemented: CanEditAdminSettings - canEditAdminSettings"))
+	usr, err := user.FindOneById(obj.UserID)
+	if err != nil {
+		return false, ResourceNotFound.Send(ctx, "user not found")
+	}
+	opts := gimlet.PermissionOpts{
+		Resource:      evergreen.SuperUserPermissionsID,
+		ResourceType:  evergreen.SuperUserResourceType,
+		Permission:    evergreen.PermissionAdminSettings,
+		RequiredLevel: evergreen.AdminSettingsEdit.Value,
+	}
+	return usr.HasPermission(opts), nil
 }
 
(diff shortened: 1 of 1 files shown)


graphql/schema/types/permissions.graphql
```diff
@@ -12,6 +12,7 @@ type DistroPermissions {
 type Permissions {
   canCreateDistro: Boolean!
   canCreateProject: Boolean!
+  canUpdateAdminSettings: Boolean!
```
khelif96, line 15: I don't feel too strongly because this pattern is used here but I feel isAdmin is a better name for this.

khelif96 (approved):
Left a comment but I won't block the PR

graphql/schema/types/permissions.graphql
minnakt: +1 I also like `isAdmin` 🤔 at the very least, I think I'd prefer using the [same verbiage](https://github.com/evergreen-ci/evergreen/blob/4d5228b8e7e87d06b5ff55b0e5fcf8c663069a0f/globals.go#L1090-L1093) from the permission levels like `canEditAdminSettings`

.gitattributes
```diff
@@ -1,4 +1,4 @@
 *.sh text eol=lf
 ./public/static/dist/* binary
 ./public/package-lock.json binary
-graphql/*_resolver.go -linguist-generated
+graphql/*_resolver.go linguist-detectable linguist-generated=false linguist-language=go
```
sophstad, line 4: I'm not sure that `=false` is valid linguist syntax. In [this](https://github.com/github-linguist/linguist/blob/master/docs/overrides.md#detectable) example, they demonstrate unmarking a path as detectable with `-linguist-detectable` which suggests that `-` is how negation is done.

graphql/schema/types/permissions.graphql
sophstad: I think using `isAdmin` would be fine. Imo there's no comparable equivalent for the other fields (`isAbleToCreateProject` is awk, `isProjectCreator` is not ideal either) so breaking the pattern makes sense.

.gitattributes
sophstad: Wow nice!! @khelif96 should we raise this in the GitHub Support thread?

sophstad (approved):

minnakt (approved):
niceee :)

PR #7026, opened 2023-09-12
EVG-20856 Fix getMatchingAliasForVersion to take in project and version

hadjri:
EVG-20856

### Description
There's a bug in `FindAliasInProjectRepoOrConfig` in the project config retrieval step as it passes a `projectID` to `getMatchingAliasForVersion` which prevents the project config from being retrieved correctly.

Changed `getMatchingAliasForVersion` to accept both a project ID and an optional version ID, allowing it to retrieve the project config by id if version ID is provided, and otherwise default to the most recent project config for the given project.

### Testing
Confirmed [without change](https://evergreen-staging.corp.mongodb.com/version/65009595b237360dc882e498), periodic build does not respect the alias defined in config, and that [with the change](https://evergreen-staging.corp.mongodb.com/version/65009911b2373618aa36ba17) the alias is respected.

image

### DIFF
--- a/model/project_aliases_test.go
+++ b/model/project_aliases_test.go
@@ -522,7 +522,7 @@ func (s *ProjectAliasSuite) TestFindAliasInProjectRepoOrConfig() {
 	// Test project config
 	found, err = FindAliasInProjectRepoOrConfig(pRef1.Id, "alias-6")
 	s.NoError(err)
-	s.Len(found, 1)
+	s.Require().Len(found, 1)
 	s.Equal(found[0].Alias, "alias-6")
 	s.Equal(found[0].Task, "*")
 	s.Equal(found[0].Variant, "*")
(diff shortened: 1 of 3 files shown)


Kimchelly (changes requested):
Does this need unit testing for handling project/repo/YAML precedence?

hadjri:
> Does this need unit testing for handling project/repo/YAML precedence?

Sure, I added some additional tests confirming the alias levels work as intended

model/project_aliases_test.go
```diff
@@ -493,6 +518,16 @@ func (s *ProjectAliasSuite) TestFindAliasInProjectOrRepo() {
 	found, err = FindAliasInProjectRepoOrConfig(pRef1.Id, "alias-5")
 	s.NoError(err)
 	s.Len(found, 0)
+
+	// Test project config
+	found, err = FindAliasInProjectRepoOrConfig(pRef1.Id, "alias-6")
+	s.NoError(err)
+	s.Len(found, 1)
```
Kimchelly, line 525: Is it worth checking that `found[0]` is the one you expect? Same with the otherone.

hadjri: Sure, added

Kimchelly: Need to `s.Require().Len` to avoid panicking in case found is empty.

Kimchelly (approved):
LGTM mod one thing.

PR #7027, opened 2023-09-12
EVG-20786 Add stepback type to project settings and parser

ZackarySantana:
EVG-20786

### Description
Add a new StepbackType field to the project config struct to support users switching between linear, bisect, and disabled. By default, the value should be disabled (0). We should also keep stepback field (bool) for backwards compatibility for now.

### Testing
Existing tests (adding this field)

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -41,7 +41,6 @@ const (
 // the ParserProject.
 type Project struct {
 	Stepback           bool                       `yaml:"stepback,omitempty" bson:"stepback"`
-	BisectStepback     bool                       `yaml:"bisect_stepback,omitempty" bson:"bisect_stepback,omitempty"`
 	UnsetFunctionVars  bool                       `yaml:"unset_function_vars,omitempty" bson:"unset_function_vars,omitempty"`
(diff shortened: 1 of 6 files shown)


hadjri:
Need to update description as the implementation's changed to simply introducing a new `BisectStepback` boolean flag

hadjri (changes requested):
Since stepback can be enabled in YAML at the [base](https://github.com/evergreen-ci/evergreen/blob/main/model/project.go#L43), [variant](https://github.com/evergreen-ci/evergreen/blob/main/model/project.go#L396), and [task](https://github.com/evergreen-ci/evergreen/blob/main/model/project.go#L707) levels, I think a simpler approach here would be to simply add a `BisectStepback` [project page](https://spruce.mongodb.com/project/evergreen/settings/general) setting (defined in the project_ref.go [here](https://github.com/evergreen-ci/evergreen/blob/main/model/project_ref.go#L62)), so the algorithm can be globally set no matter what type of stepback is occurring.

hadjri (approved):
🚀

PR #7028, opened 2023-09-12
EVG-20622: do not translate user requester for project

Kimchelly:
EVG-20622

### Description
I forgot that the commit queue does a weird thing where [it takes the already-translated project, converts it to YAML, adds more YAML configuration to add the merge task, and re-translates the already-translated project again as if it's a parser project](https://github.com/evergreen-ci/evergreen/blob/bb4161f86a67469d8d696f7d88e769c8b4045756/units/commit_queue.go#L622-L653). This is not really supposed to be done since it requires that the project can be subbed in at any place where a parser project is supposed to be used, but I don't feel like it's worth trying to fix the commit queue given that the commit queue is going to go away soon. Instead, I changed project translation so that it does not translate the user-facing requester into the internal requester.

For example, originally, if we had `allowed_requesters: ["patch"]` for the parser project (i.e. the user-facing requester string), the project would represent this as `allowed_requesters: ["patch_request"]` (i.e. the internal requester). After this change, both the parser project and the project will store `allowed_requesters: ["patch"]`.

* Do not translate user-facing requester to internal requester during project translation to avoid mismatches between the parser project and project.
* Translate user-facing requester to internal requester at the last second when comparing internal requesters against `AllowedRequesters`.

### Testing
* Updated unit tests.
* Checked all-configs to verify that the one project using this field will not be affected - new versions created after this is deployed are not affected, but existing versions that modify the parser project after the version is created (e.g. via `generate.tasks`) would be affected. From a glance, it seems like that project will not do that.

### Documentation
N/A

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -1374,7 +1374,7 @@ func validateProjectTaskIdsAndTags(project *model.Project) ValidationErrors {
 func checkTaskRuns(project *model.Project) ValidationErrors {
 	var errs ValidationErrors
 	for _, bvtu := range project.FindAllBuildVariantTasks() {
-		var hasValidAllowedRequester bool
+		hasValidAllowedRequester := len(bvtu.AllowedRequesters) == 0
 		if len(bvtu.AllowedRequesters) != 0 {
 			if bvtu.PatchOnly != nil {
 				errs = append(errs, ValidationError{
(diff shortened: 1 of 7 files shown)


validator/project_validator.go
```diff
@@ -1403,30 +1404,44 @@ func checkTaskRuns(project *model.Project) ValidationErrors {
...
+				} else {
+					hasValidAllowedRequester = true
+				}
+			}
+		} else {
+			hasValidAllowedRequester = true
 		}
-		if bvtu.SkipOnPatchBuild() && bvtu.SkipOnNonPatchBuild() {
+
+		if hasValidAllowedRequester && bvtu.SkipOnPatchBuild() && bvtu.SkipOnNonPatchBuild() {
```
hadjri, line 1422: Since these all are contingent on `hasValidAllowedRequester` we can call `if hasValidAllowedRequester { // old if statements here }` instead

validator/project_validator.go
```diff
@@ -1403,30 +1404,44 @@ func checkTaskRuns(project *model.Project) ValidationErrors {
...
+					errs = append(errs, ValidationError{
+						Level: Warning,
+						Message: fmt.Sprintf("task '%s' in build variant '%s' specifies invalid allowed_requester '%s'",
+							bvtu.Name, bvtu.Variant, requester),
+					})
+				} else {
+					hasValidAllowedRequester = true
+				}
+			}
+		} else {
```
hadjri, line 1418: Can remove this else block and initialize the variable as `hasValidAllowedRequester := len(bvtu.AllowedRequesters) == 0`

hadjri (changes requested):

validator/project_validator.go
Kimchelly: Sure, done.

validator/project_validator.go
Kimchelly: Done.

hadjri (approved):

PR #7029, opened 2023-09-12
EVG-20729: use end task message

Kimchelly:
EVG-20729

### Description
If a task fails during `setup.initial`, it never even gets to running the task, so the agent needs to immediately end the task with system failure. However, the error message parameter was never used, so the error cause of the `setup.initial` failure never made it into the failure description (it would just show the `setup.initial` command name as the failure reason with no further details). I changed it to set the system failure description when it's given.

* Set the system failure description as the failing command description when `setup.initial` fails.
* Rename the parameter a little bit so it's clear that it's specifically for the system failure case.

### Testing
Added unit tests.

### Documentation
N/A

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -877,8 +877,8 @@ func (a *Agent) runEndTaskSync(ctx context.Context, tc *taskContext, detail *api
 	_ = a.runCommandsInBlock(ctx, tc, taskSync)
 }
 
-func (a *Agent) handleTaskResponse(ctx context.Context, tc *taskContext, status string, description string) (bool, error) {
-	resp, err := a.finishTask(ctx, tc, status, description)
+func (a *Agent) handleTaskResponse(ctx context.Context, tc *taskContext, status string, systemFailureDescription string) (bool, error) {
+	resp, err := a.finishTask(ctx, tc, status, systemFailureDescription)
 	if err != nil {
 		return false, errors.Wrap(err, "marking task complete")
 	}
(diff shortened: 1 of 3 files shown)


agent/agent.go
```diff
@@ -979,8 +980,8 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 	return resp, nil
 }
 
-func (a *Agent) endTaskResponse(ctx context.Context, tc *taskContext, status string, message string) *apimodels.TaskEndDetail {
-	var userDefinedDescription string
+func (a *Agent) endTaskResponse(ctx context.Context, tc *taskContext, status string, systemFailureDescription string) *apimodels.TaskEndDetail {
+	description := systemFailureDescription
```
hadjri, line 984: Nit: can remove this line in favor of using `systemFailureDescription`

hadjri (changes requested):

agent/agent.go
Kimchelly: Hmm I kind of want to leave it as a separate variable just because there's multiple failure descriptions at play, so reusing the parameter might get confusing. The priority is like this:
* User-defined task description (if any)
* System failure description (if any)
* Default failing command description (if the command failed)
* No description at all if the task succeeded _and_ would have used the default failing command description.

hadjri: Ok, that's all good with me!

hadjri (approved):

PR #7030, opened 2023-09-12
EVG-20289: consolidate task directory field

Kimchelly:
EVG-20289

### Description
Follow-up PR to #6985 to also consolidate the task directory fields between `taskContext`/`taskConfig`. They refer to the same working directory (i.e. the task directory within the agent working directory).

* Consolidate task working directory.
* Fix a flaky agent test (which has also bit me in the past) where Windows will not allow removal of the temp dir at the end of the agent suite because some running process (e.g. `shell.exec`) is still accessing it. This is not really fixable in the agent itself because the tests are testing that the agent will not wait for commands to finish. (The Go developers [are doing the same thing for Windows](https://github.com/golang/go/blob/4ee1d542eda8d8aab7ca2024a4a0a9068d3cad70/src/testing/testing.go#L1248-L1256), but their arbitrary timeout isn't long enough.)

### Testing
* Added unit tests for keeping/clearing working directory state between tasks.
* Tested in staging that the working directory was still created for tasks, and still got cleared between normal tasks, but was preserved in between task group tasks.

### Documentation
N/A


Kimchelly:
evergreen keep-definitions

PR #7034, opened 2023-09-13
EVG-20867 Update shouldSkipWebhook to check if token can be created

bynn:
EVG-20867

### Description
the github webhooks are currently skipped for orgs without the github app if the github app is defined in the evergreen admin side.

instead of just checking that the app is defined on the evergreen side, we should check that the specific repos are able to create a token, meaning that the app can actually be used by the repo

### Testing
made sure that the app still worked

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -101,8 +101,8 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 	return nil
 }
 
-// shouldSkipWebhook returns true if the event is from a GitHub app and the app is not set up or,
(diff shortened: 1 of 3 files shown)


rest/route/github.go
```diff
@@ -103,16 +103,17 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 
 // shouldSkipWebhook returns true if the event is from a GitHub app and the app is not set up or,
```
ablack12, line 104: would update this comment for the new behavior (the app is available for the owner/repo)

ablack12 (approved):

PR #7036, opened 2023-09-14
EVG-20670: check project ID/identifier has valid characters

Kimchelly:
EVG-20670

### Description
Since Spruce uses the project identifier in the project settings URL and there's currently no character limitations on the identifier, we have to either 1. prevent special characters from being used for project identifiers, or 2. percent-encode the project ID/identifier in Spruce project settings URLs. I chose to just do option 1.

For valid characters, I referred to the RFC, which describes that [reserved characters should be percent-encoded for safety](https://www.rfc-editor.org/rfc/rfc3986#section-2.2):
>   If data for a URI component would conflict with a reserved character's purpose as a delimiter, then the conflicting data must be percent-encoded before the URI is formed.

And then chose a regexp pattern that matches [the set of unreserved characters](https://www.rfc-editor.org/rfc/rfc3986#section-2.3). I also had to include some special exceptions (space and parentheses) which are probably not valid characters, but are currently in use; we could opt to not allow setting these in new identifiers if we wanted (users can still include special characters in the display name).

As an aside, it's also worth noting that a lot of existing projects (> 100) currently have empty project identifiers, which Spruce doesn't have a way of handling since it uses the identifier in the project settings URL. The changes introduced in this PR still allow empty identifiers.

### Testing
* Added unit tests.
* Tested my regexp against all existing projects - only one didn't match the regexp (it had an `@`, which made the project settings unable to load anyways because it requires percent encoding).

### Documentation
N/A

### DIFF
--- a/rest/data/project.go
+++ b/rest/data/project.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"regexp"
 	"time"
 
 	"github.com/evergreen-ci/evergreen"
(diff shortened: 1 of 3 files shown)


hadjri (approved):

ablack12 (approved):
LGTM -- is the @ project enabled / should we let them know it's worth changing their identifier?

Kimchelly:
The project is disabled and it has "deleteme 2023-10-01" in the display name, so I assume they're going to get rid of it (or ask us to do it).

PR #7037, opened 2023-09-14
EVG-20018: Set task output metadata upon task dispatch

julianedwards:
EVG-20018

### Description
- Update the task output metadata to hold information for each type of output
- Refactor the task output interface
- Have the task set its task output metadata upon dispatch
- Remove now unnecessary agent routes to set the task output information 

### Testing
Unit testing and tested in staging.

### Documentation
N/A


taskoutput/task_log.go
```diff
@@ -80,29 +71,27 @@ func (o TaskLogs) Get(ctx context.Context, env evergreen.Environment, taskOpts T
...
-		filename = "agent"
+		logTypePrefix = "agent"
 	case TaskLogTypeSystem:
-		filename = "system"
+		logTypePrefix = "system"
 	case TaskLogTypeTask:
-		filename = "task"
+		logTypePrefix = "task"
 	default:
 		return prefix
```
Kimchelly, line 86: For my own understanding, is "all" logs treated the same as an empty/unset `TaskLogType`? Or will `TaskLogType` always be set/always be validated before `getLogName` is called?

taskoutput/task_log.go
```diff
@@ -80,29 +71,27 @@ func (o TaskLogs) Get(ctx context.Context, env evergreen.Environment, taskOpts T
...
 	switch logType {
 	case TaskLogTypeAgent:
-		filename = "agent"
+		logTypePrefix = "agent"
 	case TaskLogTypeSystem:
-		filename = "system"
+		logTypePrefix = "system"
 	case TaskLogTypeTask:
-		filename = "task"
+		logTypePrefix = "task"
```
Kimchelly, line 84: Is there a reason we would prefer using special almost-the-same-but-not-quite strings instead of just using the `TaskLogType` string value in the prefix? Seems like it could cut down on some of the duplication.

Kimchelly (changes requested):

model/task/task.go
```diff
@@ -1001,6 +1010,24 @@ func (t *Task) markAsHostDispatchedWithFunc(doUpdate func(update bson.M) error,
 	return nil
 }
 
+// newTaskOutput returns a new task output with the most up-to-date
+// configuration for the task run. This function should only be used to set the
+// task output field upon task dispatch.
+func (t *Task) newTaskOutput() *taskoutput.TaskOutput {
```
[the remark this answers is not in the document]
julianedwards: So I originally had it that, but thought it was misleading because it is not exactly just "getting" the task output, rather it is "initializing" it or "setting" it. Maybe `initializeTaskOutput` or `populateTaskOutput` would be more appropriate?

taskoutput/task_log.go
```diff
@@ -28,27 +28,18 @@ func (t TaskLogType) validate() error {
...
+// storage of a task run's task log data.
+type TaskLogOutput struct {
+	Version int `bson:"version" json:"version"`
 }
 
-// GetOptions represents the arguments for fetching task logs belonging to a
-// task run.
-type GetOptions struct {
+// ID returns the unique identifier of the task log output type.
+func (TaskLogOutput) ID() string { return "task_logs" }
```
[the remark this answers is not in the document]
julianedwards: Yeah this is a consequence of bad naming, I hate that we call one of the task log subtypes "task". I can add a comment!

taskoutput/task_log.go
julianedwards: I did this to make the log prefix generation explicit. This may just be me being paranoid, but my fear is that using the `TaskLogType` string values can leave us vulnerable to bugs. For example, someone could easily come along in the future and change the string values without realizing they are used to create the prefix for log storage. This would break log fetching and any other potential downstream systems that use S3 directly (like, potentially, a data warehouse). And for posterity, it may not be immediately obvious why that is happening which can lead to a time consuming investigation and fix.

taskoutput/task_log.go
julianedwards: It should always be set and validated before `getLogName` is called!

taskoutput/task_log.go
```diff
@@ -80,29 +71,27 @@ func (o TaskLogs) Get(ctx context.Context, env evergreen.Environment, taskOpts T
...
 }
 
-func (o TaskLogs) getLogServiceVersion() int {
+func (o TaskLogOutput) getLogServiceVersion() int {
 	switch {
-	case o >= 0:
-		return 0
 	default:
-		return -1
+		return 0
```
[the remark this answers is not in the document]
julianedwards: I was going to use `-1` for cedar, but it was too difficult to implement a Cedar backed `LogService` interface. So this will just start at `0` (the v0 bucket logger implementation).

taskoutput/task_log.go
```diff
@@ -14,9 +14,9 @@ type TaskLogType string
 
 const (
 	TaskLogTypeAll    TaskLogType = "all"
-	TaskLogTypeAgent  TaskLogType = "agent"
-	TaskLogTypeSystem TaskLogType = "system"
-	TaskLogTypeTask   TaskLogType = "task"
+	TaskLogTypeAgent  TaskLogType = "agent_log"
+	TaskLogTypeSystem TaskLogType = "system_log"
+	TaskLogTypeTask   TaskLogType = "task_log"
```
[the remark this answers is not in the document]
julianedwards: Agreed!

model/task/task.go
```diff
@@ -970,15 +972,21 @@ func (t *Task) MarkAsHostDispatchedWithContext(ctx context.Context, env evergree
...
+
+	set := bson.M{
+		DispatchTimeKey:  dispatchTime,
+		StatusKey:        evergreen.TaskDispatched,
+		HostIdKey:        hostID,
+		LastHeartbeatKey: dispatchTime,
+		DistroIdKey:      distroID,
+		AgentVersionKey:  agentRevision,
+	}
+	if output != nil {
```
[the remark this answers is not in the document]
julianedwards: I think it does [here](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L954-L957).

model/task/task.go
```diff
@@ -909,20 +907,23 @@ func (t *Task) cacheExpectedDuration() error {
...
-			AgentVersionKey:  agentVersion,
-		},
+	set := bson.M{
+		StatusKey:        evergreen.TaskDispatched,
+		DispatchTimeKey:  dispatchedAt,
+		LastHeartbeatKey: dispatchedAt,
+		PodIDKey:         podID,
+		AgentVersionKey:  agentVersion,
+	}
+	if output != nil {
```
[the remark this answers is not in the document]
julianedwards: I did this to be safe since `newTaskOutput()` _can_ return `nil`, and who knows if in the future there will be another case where it doesn't make sense to have a task output for a task document.

model/task/task.go
```diff
@@ -160,8 +158,8 @@ type Task struct {
 	ExecutionPlatform ExecutionPlatform `bson:"execution_platform,omitempty" json:"execution_platform,omitempty"`
 
 	// The version of the agent this task was run on.
-	AgentVersion      string `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
-	TaskOutputVersion *int   `bson:"task_output_version,omitempty" json:"task_output_version,omitempty"`
+	AgentVersion string                 `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
+	TaskOutput   *taskoutput.TaskOutput `bson:"task_output,omitempty" json:"task_output,omitempty"`
```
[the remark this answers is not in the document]
julianedwards: With the current architecture there are three scenarios:
1. The task is a display task, and therefore it does not make sense to have a task output (the `nil` case).
2. The task is a standalone or execution task but predates the field, in which case we should assume the "zero" value. I created the `output()` function to set the zero value when appropriate (which can be removed in a year since we will be implementing TTLs soon).
3. The task is a standalone or execution task and was dispatched with the task output field set, in which case we can just use the field as intended.

Using the "zero" value struct for first versions of any task output type is cleaner and safer. As we support for more task output types (e.g., test results), relying on the zero value is much safer than having to remember to initialize it for older tasks with the correct version. The `nil` case is distinct because it means this task should not have task output ever (i.e., display tasks).

model/task/task.go
```diff
@@ -1001,6 +1010,24 @@ func (t *Task) markAsHostDispatchedWithFunc(doUpdate func(update bson.M) error,
 	return nil
 }
 
+// newTaskOutput returns a new task output with the most up-to-date
+// configuration for the task run. This function should only be used to set the
+// task output field upon task dispatch.
+func (t *Task) newTaskOutput() *taskoutput.TaskOutput {
+	if t.DisplayOnly {
+		return nil
```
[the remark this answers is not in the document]
julianedwards: I don't think I want to error since it does look like the dispatch functions are also called for display tasks and I want the logic for whether a task has task output to live here, centrally. What if I return a `bool` as well to indicate whether the task has task output?

taskoutput/task_log.go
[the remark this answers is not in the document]
julianedwards: Done.

model/log/empty_iterator.go
```diff
@@ -0,0 +1,19 @@
+package log
+
+var empty = &emptyIterator{}
+
+type emptyIterator struct{}
+
+// EmptyIterator returns a convenience log iterator with no data that returns
+// false immediately.
+func EmptyIterator() *emptyIterator { return empty }
```
Kimchelly, line 9: It seems like this could just return a `&emptyIterator{}` rather than store the empty iterator once in a package-level field.

model/log/empty_iterator.go
```diff
@@ -0,0 +1,19 @@
+package log
+
+var empty = &emptyIterator{}
+
+type emptyIterator struct{}
+
+// EmptyIterator returns a convenience log iterator with no data that returns
+// false immediately.
```
Kimchelly, line 8: Can you disambiguate what `return false` mean in this case? It already says that it holds "no data" so it seems like that would be sufficient explanation.

model/task/task.go
```diff
@@ -158,8 +158,23 @@ type Task struct {
...
-	AgentVersion string                 `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
-	TaskOutput   *taskoutput.TaskOutput `bson:"task_output,omitempty" json:"task_output,omitempty"`
+	AgentVersion string `bson:"agent_version,omitempty" json:"agent_version,omitempty"`
+	// TaskOutputInfo holds the information for the interface that
+	// coordinates persistent storage of a task's output data.
+	// There are four possible scenarios:
+	//     1. The task will never have output data (e.g., display tasks)
+	//        and, therefore, the value is and always will be nil.
+	//     2. The task does not have output data yet, but can in the future
+	//        after running, and, therefore, the value is currently nil.
```
Kimchelly, line 168: "once it is dispatched" rather than "after running" since the value is no longer nil after it's dispatched. This seems like this case encompasses when the task is dispatched/currently running, in which case this field is not nil and it should have output data eventually, but may not have the complete output yet. I would split these two cases into scheduled (i.e. it's nil and may eventually run), dispatched/running (i.e. the field is non-nil but the output may not be done), and finished (i.e. the field is non-nil and output is finished).

model/task/task.go
```diff
@@ -158,8 +158,23 @@ type Task struct {
...
+	//     3. The task ran with the task output information initialized
```
Kimchelly, line 169: Same as above - case 2 and 3 sound pretty similar.

model/task/task.go
```diff
@@ -158,8 +158,23 @@ type Task struct {
...
+	//        upon getting dispatched and the application can safely use
+	//        this field to fetch any output data.
+	//     4. The task has data but was run before the introduction of the
+	//        field and should be initialized to the zero value on the
+	//        application to safely fetch any output data.
```
Kimchelly, line 174: I would rephrase this since zero value is the nil pointer in this case, which does not require initialization to set. This is already the value for all tasks run before this field existed.

model/task/task.go
```diff
@@ -1476,15 +1473,44 @@ func (t *Task) SetStepbackDepth(stepbackDepth int) error {
...
+		ProjectID: t.Project,
+		TaskID:    taskID,
+		Execution: t.Execution,
+	}), true
+}
+
+// getTaskOutputSafe returns an instantiation of the task output interface and
+// whether it is safe to fetch task output data. This function should always
+// be called to access task output data.
+func (t *Task) getTaskOutputSafe() (*taskoutput.TaskOutput, bool) {
```
Kimchelly, line 1500: I guess if this is going to return a bool and not an error, it's not really that much better than just passing the ponter back and letting it be nil/non-nil. That being said, I'm also fine if you want to leave it like this (to avoid unintentional nil pointer derefs).

model/task/task.go
```diff
@@ -158,8 +158,23 @@ type Task struct {
...
+	//     2. The task does not have output data yet, but can in the future
+	//        after running, and, therefore, the value is currently nil.
+	//     3. The task ran with the task output information initialized
+	//        upon getting dispatched and the application can safely use
+	//        this field to fetch any output data.
+	//     4. The task has data but was run before the introduction of the
+	//        field and should be initialized to the zero value on the
+	//        application to safely fetch any output data.
+	// This field should *never* be accessed directly, instead call
+	// `Task.getTaskOutputSafe()` to instantiate the task output interface.
```
Kimchelly, line 176: `getTaskOutputSafe` can't used to instantiate this field because it checks for an undispatched task status. Is this supposed to be `initializeTaskOutputInfo` to write the field and `getTaskOutputSafe` to read it?

model/task/task.go
```diff
@@ -1493,6 +1519,13 @@ func (t *Task) GetTaskLogs(ctx context.Context, env evergreen.Environment, getOp
 		return nil, errors.New("cannot get task logs for a display task")
 	}
 
+	output, ok := t.getTaskOutputSafe()
+	if !ok {
+		// We know there task cannot have task output, likely because
+		// it has not run yet. Return an empty iterator.
+		return log.EmptyIterator(), nil
```
Kimchelly, line 1526: Did UI already confirm that in the case of no output, the UI will show the "No logs available" message even without an error?

Kimchelly (changes requested):

model/log/empty_iterator.go
julianedwards: Oh that was referring to the `Next` command, I will just remove it.

model/task/task.go
julianedwards: Yes! I looked at the UI code and confirmed with @khelif96 that this is the expected behavior.

taskoutput/task_output.go
```diff
@@ -17,3 +19,15 @@ type TaskOptions struct {
 	// Execution is the execution number of the task run.
 	Execution int
 }
+
+// InitializeTaskOutput initializes the task output for a new task run.
+func InitializeTaskOutput(env evergreen.Environment, opts TaskOptions) *TaskOutput {
```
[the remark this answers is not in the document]
julianedwards: It will very shortly in the follow up PR!

model/task/task.go
julianedwards: It does instantiate it when appropriate (in cases where the task is undispatched it will return `nil, false` since it cannot instantiate it, hence why it is "safe"). The `initializeTaskOutputInfo` function is really only used for setting up the values to store in the DB upon dispatch.

model/task/task.go
```diff
@@ -1476,15 +1473,44 @@ func (t *Task) SetStepbackDepth(stepbackDepth int) error {
...
+// field upon task dispatch.
+func (t *Task) initializeTaskOutputInfo(env evergreen.Environment) (*taskoutput.TaskOutput, bool) {
 	if t.DisplayOnly {
-		return nil
+		return nil, false
+	}
+
+	taskID := t.Id
+	if t.Archived {
+		taskID = t.OldTaskId
```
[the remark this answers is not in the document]
julianedwards: Nope! Going to update this to return `false, nil` if the task is archived.

model/task/task.go
julianedwards: Yeah I think it is more clear in the case, similar to the map convention of returning a bool if the key does not exist.

model/task/task.go
Kimchelly: The phrasing is kind of because it talks about accessing (i.e. reading) + instantiating (i.e. writing/giving it an initial value), but they're supposedly different things, right? Even if you call `getTaskOutputSafe`, this field will still remain uninitialized. Maybe say something like "use `getTaskOutputSafe` to access this value, use `initializeTaskOutputInfo` to initialize it"?

taskoutput/task_log.go
Kimchelly: Makes sense, and sounds good!

taskoutput/task_output.go
```diff
@@ -1,32 +1,33 @@
...
-func (o TaskOutput) TaskLogs() tasklogs.TaskLogs {
-	switch {
-	case o >= 0:
-		return tasklogs.TaskLogs(0)
-	default:
-		return tasklogs.TaskLogs(-1)
-	}
+// TaskOptions represents the task-level information required for accessing
+// task logs belonging to a task run.
+type TaskOptions struct {
```
Kimchelly, line 14: nit: would be worth adding `bson:"-" json:"-"` to this (+ `TaskLogGetOptions`) to make it clear that it's not meant to be serialized (or otherwise, add tags names).

Kimchelly (approved):
LGTM mod small suggestions.

model/task/task.go
julianedwards: I ended up just simplifying it:
```
// This field should *never* be accessed directly, instead call
// `Task.getTaskOutputSafe()`.
```
The function definition itself explains how the instantiation works.

PR #7038, opened 2023-09-14
EVG-20019: Export log service interface

julianedwards:
EVG-20019

### Description
Updates the code with the following improvements/fixes:
- Export the log service interface and have the task output implementations set up the log service directly
- Stop relying on the env to fetch the log bucket config, instead use bucket info stored in the task output info
- Refactor the log sender
- Fix small bug in the log service v0 implementation
- Do not include new lines in the service representation of a log line 

### Testing
Updated and added unit tests

### Documentation
N/A


taskoutput/util.go
```diff
@@ -0,0 +1,36 @@
...
+
+func newBucket(bucketName, bucketType string) (pail.Bucket, error) {
+	var b pail.Bucket
+	var err error
+
+	switch bucketType {
+	case evergreen.BucketTypeS3:
+		b, err = pail.NewS3Bucket(pail.S3Options{
+			Name:        bucketName,
+			Region:      "us-east-1",
```
malikchaya2, line 18: Why do we want this hardcoded? 
If we do, we should probably use a constant for "us-east-1".

malikchaya2 (changes requested):

taskoutput/util.go
julianedwards: Good catch! Thanks.

julianedwards: We only use `us-east-1` for our buckets so we can hard code this.

malikchaya2 (approved):

PR #7042, opened 2023-09-15
EVG-20838 Update githook to only handle checks required merge group events

bynn:
EVG-20838

### Description
merge group seems to be sending 2 events with different actions 

image

1. checks_requested
2. destroyed

it says that github will send the destroy event when a pull request is removed from a queue in the docs 
https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=checks_requested#merge_group

it looks like the merge group destroy event is sent if there's only 1 PR in the queue and it merges successfully (couldn't find documentation for this behavior) 

I updated the githook to only look at the checks_requested action for merge_group events

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -24,10 +24,11 @@ import (
 )
 
 const (
-	githubActionClosed      = "closed"
-	githubActionOpened      = "opened"
-	githubActionSynchronize = "synchronize"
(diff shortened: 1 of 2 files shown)


bsamek (approved):
Good find! Thank you!

Kimchelly (approved):

PR #7051, opened 2023-09-20
EVG-19914 write using new patch status

ablack12:
EVG-19914
### Description
Broke this migration into a few steps. We now handle the old and new patch status in both the UI and the front end; so we should be able to start writing with the new status (and if we need to revert, this will be backwards compatible).

### Testing
Unit tests (and added a statement), and will test on the UI with old and new patches.

### Documentation


ablack12:
Verified with Spruce, notifications, and legacy that things look to be behaving correctly with old and [new patch](https://spruce-staging.corp.mongodb.com/version/650df28b97b1d3c6d2ed3b4d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
image

model/patch/patch_test.go
```diff
@@ -946,19 +946,23 @@ func TestSetChildPatches(t *testing.T) {
 }
 
 func TestGetCollectiveStatusFromPatchStatuses(t *testing.T) {
-	successful := []string{evergreen.LegacyPatchSucceeded}
-	assert.Equal(t, evergreen.LegacyPatchSucceeded, GetCollectiveStatusFromPatchStatuses(successful))
+	successful := []string{evergreen.VersionSucceeded}
+	assert.Equal(t, evergreen.VersionSucceeded, GetCollectiveStatusFromPatchStatuses(successful))
```
Kimchelly, line 950: Since `GetCollectiveStatusFromPatchStatuses` is used in the REST API to [decide the API model patch's status](https://github.com/evergreen-ci/evergreen/blob/dc44657541b9a8eb6b746bd03679cde291061b13/rest/model/patch.go#L319), it seems like this will change the output from the REST API from "succeeded" to "success". I wonder if people have existing scripts (or maybe use libraries like [evergreen.py](https://github.com/evergreen-ci/evergreen.py/blob/d989b8f84dd370ee42315256ed95dca100d12c78/tests/evergreen/data/patch.json#L11)) and are still looking for "succeeded" but aren't aware that this is about to change to "success" - is it worth doing any comms to raise awareness?

Kimchelly (changes requested):

model/patch/patch_test.go
ablack12: Yeaaaahh we should either do some sort of comms, or we could translate this back to "succeeded" for the rest response 🤔

ablack12: @Kimchelly re-requesting review -- with the assumption that we're going to do comms, could you review the changes?

ablack12: I added something to ensure that API response is consistently the new thing but if we're concerned we could change it to be consistently the old thing.

rest/model/patch.go
```diff
@@ -203,12 +203,16 @@ func (apiPatch *APIPatch) buildBasePatch(p patch.Patch) {
...
 	apiPatch.Author = utility.ToStringPtr(p.Author)
 	apiPatch.Version = utility.ToStringPtr(p.Version)
-	apiPatch.Status = utility.ToStringPtr(p.Status)
 	apiPatch.Hidden = p.Hidden
 	apiPatch.CreateTime = ToTimePtr(p.CreateTime)
 	apiPatch.StartTime = ToTimePtr(p.StartTime)
 	apiPatch.FinishTime = ToTimePtr(p.FinishTime)
 	apiPatch.MergedFrom = utility.ToStringPtr(p.MergedFrom)
+	apiPatch.Status = utility.ToStringPtr(p.Status)
+	// Ensure we're returning a consistent successful status.
```
Kimchelly, line 212: I assume the answer is yes, but one day we're going to delete `evergreen.LegacyPatchSucceeded` from the codebase entirely, right?

Kimchelly (approved):
LGTM assuming sufficient comms go out before this PR is deployed.

rest/model/patch.go
ablack12: yes, in a year haha (https://jira.mongodb.org/browse/EVG-20032)

PR #7053, opened 2023-09-20
EVG-20883 Create function to find the task inbetween two given tasks 

ZackarySantana:
EVG-20833

### Description
To support bisect stepback, we need a function that takes two tasks of the same lineage (build variant, display name, on the mainline commits (aka a specific requester)). And find the task that would be between them in the middle.

### Testing
Add unit testing


model/task/db.go
```diff
@@ -591,22 +602,6 @@ func ByVersionsForNameAndVariant(versions, displayNames []string, buildVariant s
...
-		DisplayNameKey:  displayName,
-		RequesterKey:    requester,
-		RevisionOrderNumberKey: bson.M{
-			"$lt": currentRevisionOrder,
-			"$gt": previousRevisionOrder,
-		},
-		ProjectKey: project,
-	}
-}
-
```
ZackarySantana, line 609: This was dead code

model/task/task.go
```diff
@@ -169,10 +169,14 @@ type Task struct {
...
-	Archived            bool   `bson:"archived,omitempty" json:"archived,omitempty"`
-	RevisionOrderNumber int    `bson:"order,omitempty" json:"order,omitempty"`
+	Execution int    `bson:"execution" json:"execution"`
+	OldTaskId string `bson:"old_task_id,omitempty" json:"old_task_id,omitempty"`
+	Archived  bool   `bson:"archived,omitempty" json:"archived,omitempty"`
+
+	// RevisionOrderNumber for user submitted patches is their count of patches
+	// that they have submitted and attached to their user. For mainline commits,
+	// it's the amount of versions for that repository so far.
+	RevisionOrderNumber int `bson:"order,omitempty" json:"order,omitempty"`
```
ZackarySantana, line 179: Adding comment for future documentation

model/task/task.go
```diff
@@ -855,22 +859,6 @@ func (t *Task) FindTaskOnPreviousCommit() (*Task, error) {
...
-	}
-
-	// reverse the slice of tasks
-	intermediateTasksReversed := make([]Task, len(intermediateTasks))
-	for idx, t := range intermediateTasks {
-		intermediateTasksReversed[len(intermediateTasks)-idx-1] = t
-	}
-	return intermediateTasksReversed, nil
-}
-
```
ZackarySantana, line 873: This was dead code

model/task/task.go
```diff
@@ -1264,6 +1252,12 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 	return nil
 }
 
+// GetTaskIdBetweenIds retrieves the task id between two tasks that are of the same.
```
Kimchelly, line 1255: "of the same"...? Same task but different IDs?

model/task/task.go
```diff
@@ -1264,6 +1252,12 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
...
+func GetTaskIdBetweenTasks(t1, t2 Task) (*Task, error) {
```
Kimchelly, line 1256: This could imply that it would get any task ID in between the revisions for t1 and t2. Can we make this function more explicit that it gets the task midway between the two?

model/task/task_test.go
```diff
@@ -1253,6 +1253,43 @@ func TestBulkInsert(t *testing.T) {
...
+			BuildVariant:        buildVarient,
+			Requester:           requester,
+			Project:             project,
+			RevisionOrderNumber: i,
+		}
+		assert.NoError(task.Insert())
+		tasks = append(tasks, task)
+	}
+	t10, err := GetTaskIdBetweenTasks(tasks[0], tasks[19])
+	assert.Equal(t10.RevisionOrderNumber, 10)
```
Kimchelly, line 1277: Have to `require.NotZero(t10)` first or else this can panic when dereferencing `t10`. It's also a little more idiomatic to check `err` before checking the result.

Same with the others below.

model/task/task_test.go
```diff
@@ -1253,6 +1253,43 @@ func TestBulkInsert(t *testing.T) {
...
+	assert.Equal(t5.RevisionOrderNumber, 5)
+	assert.NoError(err)
+
+	t15, err := GetTaskIdBetweenTasks(tasks[10], tasks[19])
+	assert.Equal(t15.RevisionOrderNumber, 15)
+	assert.NoError(err)
+
+	t19, err := GetTaskIdBetweenTasks(tasks[17], tasks[19])
+	assert.Equal(t19.RevisionOrderNumber, 19)
+	assert.NoError(err)
```
Kimchelly, line 1290: May be worth testing other valid input cases like:
* Passing a higher order number task as the first arg.
* Two of the same task.
* Tasks that are adjacent in terms of revision number.
* Passing a task that has no lower revision order number/no other matching BV, etc (i.e. it's the first version in the repo).

model/task/task_test.go
```diff
@@ -1253,6 +1253,43 @@ func TestBulkInsert(t *testing.T) {
 	}
 }
 
+func TestGetTaskIdBetweenIds(t *testing.T) {
+	assert := assert.New(t)
+	require.NoError(t, db.ClearCollections(Collection))
```
Kimchelly, line 1258: nit: should be consistent about using a `assert.New` for `require.New` as well, or otherwise omit the `assert.New` entirely.

model/task/task.go
```diff
@@ -169,10 +169,14 @@ type Task struct {
...
 
-	Execution           int    `bson:"execution" json:"execution"`
-	OldTaskId           string `bson:"old_task_id,omitempty" json:"old_task_id,omitempty"`
-	Archived            bool   `bson:"archived,omitempty" json:"archived,omitempty"`
-	RevisionOrderNumber int    `bson:"order,omitempty" json:"order,omitempty"`
+	Execution int    `bson:"execution" json:"execution"`
+	OldTaskId string `bson:"old_task_id,omitempty" json:"old_task_id,omitempty"`
+	Archived  bool   `bson:"archived,omitempty" json:"archived,omitempty"`
+
+	// RevisionOrderNumber for user submitted patches is their count of patches
```
Kimchelly, line 176: Phrasing suggestion - `"RevisionOrderNumber for user-submitted patches is the user's current patch submission count"`.

model/task/task.go
```diff
@@ -1264,6 +1252,12 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 	return nil
 }
 
+// GetTaskIdBetweenIds retrieves the task id between two tasks that are of the same.
+func GetTaskIdBetweenTasks(t1, t2 Task) (*Task, error) {
+	mid := (t1.RevisionOrderNumber + t2.RevisionOrderNumber) / 2
+	return FindOne(db.Query(ByRevisionOrderNumber(t1.BuildVariant, t2.DisplayName, t1.Project, t1.Requester, mid)))
```
Kimchelly, line 1258: Is it significant that it's using `t2`'s display name but `t1`'s fields for everything else?

model/task/task.go
Kimchelly, line 1256: Are we assuming that `BuildVariant`, `DisplayName`, `Project`, and `Requester` are all the same between `t1` and `t2`, or can they differ? If so, can we validate that?

Kimchelly (changes requested):

model/task/db.go
```diff
@@ -579,6 +579,17 @@ func ByPreviousCommit(buildVariant, displayName, project, requester string, orde
 	}
 }
 
+// ByPreviousCommit creates a query on Evergreen as the requester on a previous revision with the same buildVariant, displayName and project
```
Kimchelly, line 582: nit: period at the end of doc comment sentence.

Kimchelly: This doc comment is also kind of the inverse of what it should be - it's implying how it's actually used under the assumption that it's called by `GetTaskIdBetweenTasks`, when it should be stating what it is querying for in a more general way. In this case, this is just an equality query on the given fields, so I would phrase it as something like `returns a query for the given requester, BV, etc.`.

model/task/task.go
Kimchelly, line 1256: We should call this `FindTaskBetweenTasks` (or `FindMidwayTask` or something) since this is getting the whole task data model rather than just its ID.

model/task/task.go
ZackarySantana: Not at all, that is a typo 😆

model/task/task.go
```diff
@@ -1273,6 +1260,35 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 	return nil
 }
 
+// GetTaskIdBetweenIds gets the task between two task given that they are
+// from the same project, requester, build variant, and display name. The
+// order of the ID's does not matter and if the task passed cannot have a
+// middle (i.e. it is sequential tasks or the same task) it will return the
+// the first task given.
+func GetMidwayTask(t1, t2 Task) (*Task, error) {
```
Kimchelly, line 1268: nit: it's slightly more conventional in Evergreen for helper functions wrapping find queries to use `Find` instead of `Get` in the name.

model/task/task.go
```diff
@@ -1273,6 +1260,35 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
...
+	// The tasks should be the same build variant, display name, project, and requester.
+	if t1.BuildVariant != t2.BuildVariant {
```
Kimchelly, line 1270: Small suggestion to use a Grip error catcher to accumulate all the errors and return them, rather than return on first error (so that the caller gets all the errors, not just the first error).

```go
catcher := grip.NewBasicCatcher() // Makes an error accumulator
catcher.ErrorfWhen(t1.BuildVariant != t2.BuildVariant, "given tasks have differing build variants '%s' and '%s', t1.BuildVariant, t2.BuildVariant`) // Conditionally adds formatted error
catcher.ErrorfWhen(...)
...
if catcher.HasErrors() {
    return nil, catcher.Resolve()
}

// Continue now that we know input is valid
```

model/task/task_test.go
```diff
@@ -1253,6 +1253,63 @@ func TestBulkInsert(t *testing.T) {
...
+			Requester:           requester,
+			Project:             project,
+			RevisionOrderNumber: i,
+		}
+		assert.NoError(task.Insert())
+		tasks = append(tasks, task)
+	}
+	t10, err := GetMidwayTask(tasks[0], tasks[19])
+	assert.NoError(err)
+	assert.NotNil(t10)
```
Kimchelly, line 1278: This has to be `require` - the difference between `assert` and `require` is that `require` will stop if the condition is not met, whereas `assert` will continue. Since this uses `assert`, it'll continue even if `t10` is nil (and report the error at the end of the test), but then the line after this will panic.

Same with the others in this file.

model/task/task_test.go
```diff
@@ -1253,6 +1253,63 @@ func TestBulkInsert(t *testing.T) {
...
+	assert.Equal(t10.RevisionOrderNumber, 10)
+
+	t5, err := GetMidwayTask(tasks[0], tasks[9])
+	assert.NoError(err)
+	assert.NotNil(t5)
+	assert.Equal(t5.RevisionOrderNumber, 5)
```
Kimchelly, line 1284: nit: I missed this in the original review, but the expected value is the 1st argument and the actual value is the 2nd. If the assertion fails, doing it like this will print the expected/actual values flipped and may cause confusion.

model/task/task_test.go
```diff
@@ -1253,6 +1253,63 @@ func TestBulkInsert(t *testing.T) {
...
+	t12, err := GetMidwayTask(tasks[11], tasks[11])
+	assert.NoError(err)
+	assert.NotNil(t12)
+	assert.Equal(t12.RevisionOrderNumber, 12)
+
+	t16, err := GetMidwayTask(tasks[15], tasks[16])
+	assert.NoError(err)
+	assert.NotNil(t16)
+	assert.Equal(t16.RevisionOrderNumber, 16)
+
```
Kimchelly, line 1310: nit: remove random newline.

model/task/task.go
```diff
@@ -1273,6 +1260,35 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
...
+		return nil, errors.Errorf("given tasks have differing build variants '%s' and '%s'", t1.BuildVariant, t2.BuildVariant)
+	}
+	if t1.DisplayName != t2.DisplayName {
+		return nil, errors.Errorf("given tasks have differing display names '%s' and '%s'", t1.DisplayName, t2.DisplayName)
+	}
+	if t1.Project != t2.Project {
+		return nil, errors.Errorf("given tasks have differing projects '%s' and '%s'", t1.Project, t2.Project)
+	}
+	if t1.Requester != t2.Requester {
+		return nil, errors.Errorf("given tasks have differing requesters '%s' and '%s'", t1.Requester, t2.Requester)
```
Kimchelly, line 1280: I think it would be worth testing the validation logic at least a little bit, since we're now requiring that they're the same task definition.

Kimchelly (changes requested):

model/task/task_test.go
ZackarySantana: I'm not sure what you mean by the last one, like it's the one and only task ever for that project, etc..?
I think the new test cases for display name and etc handle this, since if a task is the only one for that project, there cannot be another one with the same display name, project, bv, etc

model/task/task_test.go
```diff
@@ -1273,41 +1273,88 @@ func TestGetTaskIdBetweenIds(t *testing.T) {
...
+	t16, err := FindMidwayTask(tasks[15], tasks[16])
 	assert.NoError(err)
-	assert.NotNil(t16)
-	assert.Equal(t16.RevisionOrderNumber, 16)
-
+	require.NotNil(t, t16)
+	assert.Equal(16, t16.RevisionOrderNumber)
+
+	otherDisplayName := Task{
+		Id:           "otherTask",
```
Kimchelly, line 1312: This is gonna cause a write error when inserting inot the  DB since it's invalid to insert multiple docs with the same `_id`. You can use `utility.RandomString()` if you just want a random name, or just fill in some ID without any other dups in the test.

model/task/task_test.go
Kimchelly: Yeah, the last case isn't really possible now that the validation is there, so I don't think it needs to be tested explicitly.

Kimchelly (approved):
LGTM mod unique task IDs. Nice job!

PR #7054, opened 2023-09-20
EVG-20369: allow activate: true for new build variants in generate.tasks

Kimchelly:
EVG-20369

### Description
The user's patch was not activating their generated tasks because they explicitly set `activate: true` in their generate.tasks JSON. The logic in generate.tasks assumed that if `activate` is set to anything, it must have specific activation conditions. However, this is only true if `activate` is set to false - a generated task that sets `activate: true` should just activate by default, the same as if `activate` was not specified at all. This only affected new BVs that explicitly set `activate: true` because existing BV definitions can't be modified.

* Fix handling of `activate: true` in new BVs for generate.tasks.
* Fix a small bug I spotted where the task creation logic didn't filter out requesters for which batchtime doesn't apply.

### Testing
* Added more unit tests for generate.tasks interacting with `activate` + other specific activation settings.
* Tested in prod with `activate: true` when generated tasks added to an existing BV ([patch](https://spruce.mongodb.com/version/6509f71032f417e430a47b6b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) - tasks were activated as expected.
* Tested in prod with `activate: true` when generated tasks were in a new BV ([patch](https://spruce.mongodb.com/version/6509fe98d6d80a9434fad2e9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) - tasks were not activated, which is a bug.
   * Tested in staging with the same conditions + the changes in this PR ([patch](https://spruce-staging.corp.mongodb.com/version/650b0a8fb23736dcc0932004/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) - tasks were activated.

### Documentation
N/A

### DIFF
--- a/model/generate.go
+++ b/model/generate.go
@@ -530,7 +530,9 @@ func (b *specificActivationInfo) taskOrVariantHasSpecificActivation(variant, tas
 func (g *GeneratedProject) findTasksAndVariantsWithSpecificActivations(requester string) specificActivationInfo {
 	res := newSpecificActivationInfo()
 	for _, bv := range g.BuildVariants {
-		// Only consider batchtime for mainline builds. We should always respect activate if it is set.
+		// Only consider batchtime for mainline builds. A task/BV will have
+		// specific activation if activate if it is explicitly set to false;
+		// otherwise, if it's explicitly set to true, activate it immediately.
 		if evergreen.ShouldConsiderBatchtime(requester) && bv.hasSpecificActivation() {
 			res.activationVariants = append(res.activationVariants, bv.name())
 		} else if !utility.FromBoolTPtr(bv.Activate) {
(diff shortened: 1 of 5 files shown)


model/generate.go
```diff
@@ -532,7 +533,7 @@ func (g *GeneratedProject) findTasksAndVariantsWithSpecificActivations(requester
 		// Only consider batchtime for mainline builds. We should always respect activate if it is set.
 		if evergreen.ShouldConsiderBatchtime(requester) && bv.hasSpecificActivation() {
 			res.activationVariants = append(res.activationVariants, bv.name())
-		} else if bv.Activate != nil {
+		} else if !utility.FromBoolTPtr(bv.Activate) {
```
hadjri, line 536: Nit: we should update that comment on L533 to clarify we only care if activate is false

hadjri (approved):
LGTM with nit. Great thorough testing 🚀

model/generate.go
Kimchelly: I modified the wording. This comment is a little vague on the difference between activate being explicitly true, explicitly false, and unset, so I tried to clarify it a bit.

PR #7056, opened 2023-09-20
EVG-18354 Escape file names when attaching to tasks 

malikchaya2:
EVG-18354

### Description
We saw an example of a [task](https://spruce.mongodb.com/task/mms_e2e_cypress_E2E_CYPRESS_UI_patch_6a989832deb87ea30ff5c117945cc10c2d96b25b_637ce21d32f417389e6c7f9a_22_11_22_14_52_17/files?execution=2&sortBy=STATUS&sortDir=ASC) where the first file gad a broken link because the "#" character was not escaped correctly. This change ensures that the "#" at the end of the link is replaced with "%23" so that the file loads properly.

### Testing
Added a unit test for the function and [tested it on staging ](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_dist_staging_patch_a1e18399674027393f3f71e1de87cf079617b66a_650b6f1c97b1d3bc9cd5b382_23_09_20_22_16_18/files?execution=0) with a file with # in the link.

### DIFF
--- a/model/artifact/artifact_file.go
+++ b/model/artifact/artifact_file.go
@@ -154,13 +154,13 @@ func RotateSecrets(toReplace, replacement string, dryRun bool) (map[TaskIDAndExe
 func EscapeFiles(files []File) []File {
 	var escapedFiles []File
 	for _, file := range files {
-		file.Link = EscapeFile(file.Link)
+		file.Link = escapeFile(file.Link)
 		escapedFiles = append(escapedFiles, file)
 	}
 	return escapedFiles
 }
 
-func EscapeFile(path string) string {
+func escapeFile(path string) string {
 	base := filepath.Base(path)
 	i := strings.LastIndex(path, base)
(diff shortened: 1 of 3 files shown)


model/artifact/artifact_file.go
```diff
@@ -147,3 +150,18 @@ func RotateSecrets(toReplace, replacement string, dryRun bool) (map[TaskIDAndExe
...
+func EscapeFiles(files []File) []File {
+	var escapedFiles []File
+	for _, file := range files {
+		file.Link = EscapeFile(file.Link)
+		escapedFiles = append(escapedFiles, file)
+	}
+	return escapedFiles
+}
+
+func EscapeFile(path string) string {
```
bynn, line 163: can this be a private function if we only use it for EscapeFiles

bynn (changes requested):

model/artifact/artifact_file.go
malikchaya2: good call. fixed.

bynn (approved):

PR #7057, opened 2023-09-21
EVG-20910 abort with display tasks

ablack12:
EVG-20910
### Description
If the display task isn't included in the list of taskIds for RestartVersion, then we don't mark it as ResetWhenFinished, and then we bail [here](https://github.com/evergreen-ci/evergreen/blob/89de5ebcb3801d5750db68dcfad11e467818d00c/model/task_lifecycle.go#L2308) and never restart the task. 

### Testing
Added a unit test.

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -320,39 +320,50 @@ func RestartTasksInVersion(ctx context.Context, versionId string, abortInProgres
 // RestartVersion restarts completed tasks associated with a versionId.
 // If abortInProgress is true, it also sets the abort flag on any in-progress tasks.
 func RestartVersion(ctx context.Context, versionId string, taskIds []string, abortInProgress bool, caller string) error {
-	if abortInProgress {
-		if err := task.AbortAndMarkResetTasksForVersion(versionId, taskIds, caller); err != nil {
-			return errors.WithStack(err)
-		}
-	}
-	allFinishedTasks, err := getTasksToReset(taskIds)
+	// Fetch tasks to reset before aborting to ensure we're including any needed display tasks.
+	tasks, err := getTasksWithDisplayTasks(taskIds)
 	if err != nil {
 		return errors.Wrap(err, "getting finished tasks")
 	}
-	if len(allFinishedTasks) == 0 {
-		return nil
-	}
-	return restartTasks(ctx, allFinishedTasks, caller, versionId)
-}
-
-// getTasksToReset returns all finished tasks that should be reset given an initial input list of
-// finished taskIds. If a display task is in the list, its execution tasks are removed if they also exist.
-func getTasksToReset(taskIds []string) ([]task.Task, error) {
-	finishedTasks, err := task.FindAll(db.Query(task.ByIdsAndStatus(taskIds, evergreen.TaskCompletedStatuses)))
-	if err != nil {
-		return nil, errors.WithStack(err)
-	}
-	allFinishedTasks, err := task.AddParentDisplayTasks(finishedTasks)
-	if err != nil {
-		return nil, errors.WithStack(err)
-	}
-	for i := len(allFinishedTasks) - 1; i >= 0; i-- {
-		t := allFinishedTasks[i]
-		if t.DisplayTask != nil {
-			allFinishedTasks = append(allFinishedTasks[:i], allFinishedTasks[i+1:]...)
+	tasksToReset := []task.Task{}
+	abortIds := []string{}
+	for i, t := range tasks {
+		if utility.StringSliceContains(evergreen.TaskCompletedStatuses, t.Status) {
+			tasksToReset = append(tasksToReset, tasks[i])
+		} else {
+			abortIds = append(abortIds, t.Id)
 		}
 	}
-	return allFinishedTasks, nil
+	if abortInProgress {
+		if err := task.AbortAndMarkResetTasksForVersion(versionId, abortIds, caller); err != nil {
+			return errors.WithStack(err)
+		}
+	}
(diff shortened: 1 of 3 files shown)


model/lifecycle.go
```diff
@@ -320,39 +320,50 @@ func RestartTasksInVersion(ctx context.Context, versionId string, abortInProgres
...
-		return nil, errors.WithStack(err)
+		return nil, errors.Wrap(err, "finding parent display tasks")
 	}
-	for i := len(allFinishedTasks) - 1; i >= 0; i-- {
-		t := allFinishedTasks[i]
-		if t.DisplayTask != nil {
-			allFinishedTasks = append(allFinishedTasks[:i], allFinishedTasks[i+1:]...)
+	allTasks := []task.Task{}
+	for i, t := range tasks {
+		if !t.IsPartOfDisplay() {
```
julianedwards, line 361: We may want to include exec tasks so they abort too, otherwise the parent display task will abort but won't restart until the children exec tasks finish. This is actually the current behavior when restarting a version on the UI.

julianedwards (changes requested):

model/lifecycle.go
```diff
@@ -320,39 +320,50 @@ func RestartTasksInVersion(ctx context.Context, versionId string, abortInProgres
...
-	if len(allFinishedTasks) == 0 {
+	if len(tasksToReset) == 0 {
 		return nil
 	}
-	return restartTasks(ctx, allFinishedTasks, caller, versionId)
+	return restartTasks(ctx, tasksToReset, caller, versionId)
 }
 
-// getTasksToReset returns all finished tasks that should be reset given an initial input list of
+// getTasksWithDisplayTasks returns all finished tasks that should be reset given an initial input list of
```
hadjri, line 348: Need to modify comment since we're no longer just returning finished tasks

model/lifecycle.go
```diff
@@ -320,39 +320,50 @@ func RestartTasksInVersion(ctx context.Context, versionId string, abortInProgres
...
 // finished taskIds. If a display task is in the list, its execution tasks are removed if they also exist.
-func getTasksToReset(taskIds []string) ([]task.Task, error) {
-	finishedTasks, err := task.FindAll(db.Query(task.ByIdsAndStatus(taskIds, evergreen.TaskCompletedStatuses)))
+func getTasksWithDisplayTasks(taskIds []string) ([]task.Task, error) {
+	tasks, err := task.FindAll(db.Query(task.ByIds(taskIds)))
 	if err != nil {
 		return nil, errors.WithStack(err)
```
hadjri, line 353: Nit: we should wrap the error here with a descriptive string

model/lifecycle.go
ablack12: Ah good point, ty

model/lifecycle.go
```diff
@@ -442,16 +453,30 @@ func RestartVersions(ctx context.Context, versionsToRestart []*VersionToRestart,
 // RestartBuild restarts completed tasks associated with a given buildId.
 // If abortInProgress is true, it also sets the abort flag on any in-progress tasks.
 func RestartBuild(ctx context.Context, build *build.Build, taskIds []string, abortInProgress bool, caller string) error {
+	// Fetch tasks to reset before aborting to ensure we're including any needed display tasks.
+	tasks, err := getTasksWithDisplayTasks(taskIds)
```
hadjri, line 457: It seems this [logic](https://github.com/evergreen-ci/evergreen/pull/7057/files#diff-f50207584681d142f6e119549a8e2e22249b45dcbe3cb6f7570eb906be64299dR458-R469)  is duplicated across `RestartBuild` and `RestartVersion`, can we absorb it into `getTasksWithDisplayTasks`?

model/lifecycle.go
```diff
@@ -320,39 +320,50 @@ func RestartTasksInVersion(ctx context.Context, versionId string, abortInProgres
 // RestartVersion restarts completed tasks associated with a versionId.
 // If abortInProgress is true, it also sets the abort flag on any in-progress tasks.
 func RestartVersion(ctx context.Context, versionId string, taskIds []string, abortInProgress bool, caller string) error {
+	// Fetch tasks to reset before aborting to ensure we're including any needed display tasks.
+	tasks, err := getTasksWithDisplayTasks(taskIds)
+	if err != nil {
+		return errors.Wrap(err, "getting finished tasks")
```
hadjri, line 326: Same comment about the err message here

hadjri (changes requested):
Can we also update this PR title to be more descriptive and include what the change is in the description?

model/lifecycle.go
```diff
@@ -442,16 +453,30 @@ func RestartVersions(ctx context.Context, versionsToRestart []*VersionToRestart,
...
+	abortIds := []string{}
+	for i, t := range tasks {
+		if utility.StringSliceContains(evergreen.TaskCompletedStatuses, t.Status) {
+			tasksToReset = append(tasksToReset, tasks[i])
+		} else {
+			abortIds = append(abortIds, t.Id)
+		}
+	}
 	if abortInProgress {
+		allIds := []string{}
```
hadjri, line 471: Seems like `allIds` can be removed?

hadjri (changes requested):

model/lifecycle.go
ablack12: I thought about that but it's kind of messy and I don't think actually simplifies it.

I'm actually going to approach this change differently so I'll be reverting these functions to closer to what they were.

ablack12:
Got distracted and Julian picked up a better version haha so closing.

PR #7058, opened 2023-09-21
EVG-201910: fix restarting display tasks by version and build

julianedwards:
EVG-20910

### Description
Taking over from where @ablack12 left off in https://github.com/evergreen-ci/evergreen/pull/7057/

This refactors and consolidates the code to reset tasks by version and build. There were multiple codepaths making the behavior inconsistent for this functionality depending on how it was called. I also fixed a commit queue blocking bug in `ResetVersion` that did not properly abort display tasks.

### Testing
I tested in staging the restarting a version behaves consistently when invoked from the both legacy and Spruce UI as well as the rest API.

I updated the existing unit tests. Since this is a fix for a P2 bug, I will do a more thorough test refactor in a follow-up PR.

### Documentation
N/A

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -2848,19 +2848,19 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
 // AbortAndMarkResetTasksForBuild aborts and marks in-progress tasks to reset
 // from the specified task IDs and build ID. If no task IDs are specified, all
 // in-progress tasks belonging to the build are aborted and marked to reset.
-func AbortAndMarkResetTasksForBuild(ctx context.Context, env evergreen.Environment, buildID string, taskIDs []string, caller string) error {
-	return abortAndMarkResetTasks(ctx, env, ByBuildId(buildID), taskIDs, caller)
+func AbortAndMarkResetTasksForBuild(ctx context.Context, buildID string, taskIDs []string, caller string) error {
+	return abortAndMarkResetTasks(ctx, ByBuildId(buildID), taskIDs, caller)
 }
 
 // AbortAndMarkResetTasksForVersion aborts and marks in-progress tasks to reset
 // from the specified task IDs and version ID. If no task IDs are specified,
 // all in-progress tasks belonging to the version are aborted and marked to
 // reset.
-func AbortAndMarkResetTasksForVersion(ctx context.Context, env evergreen.Environment, versionID string, taskIDs []string, caller string) error {
-	return abortAndMarkResetTasks(ctx, env, ByVersion(versionID), taskIDs, caller)
+func AbortAndMarkResetTasksForVersion(ctx context.Context, versionID string, taskIDs []string, caller string) error {
+	return abortAndMarkResetTasks(ctx, ByVersion(versionID), taskIDs, caller)
 }
 
-func abortAndMarkResetTasks(ctx context.Context, env evergreen.Environment, filter bson.M, taskIDs []string, caller string) error {
+func abortAndMarkResetTasks(ctx context.Context, filter bson.M, taskIDs []string, caller string) error {
 	filter[StatusKey] = bson.M{"$in": evergreen.TaskInProgressStatuses}
 	if len(taskIDs) > 0 {
 		filter["$or"] = []bson.M{
@@ -2870,7 +2870,7 @@ func abortAndMarkResetTasks(ctx context.Context, env evergreen.Environment, filt
 		}
 	}
 
-	_, err := env.DB().Collection(Collection).UpdateMany(
+	_, err := evergreen.GetEnvironment().DB().Collection(Collection).UpdateMany(
 		ctx,
 		filter,
 		bson.M{"$set": bson.M{
@@ -2886,18 +2886,18 @@ func abortAndMarkResetTasks(ctx context.Context, env evergreen.Environment, filt
 // FindCompletedTasksByVersion returns all completed tasks belonging to the
 // given version ID. Excludes execution tasks. If no task IDs are specified,
 // all completed tasks belonging to the version are returned.
-func FindCompletedTasksByVersion(ctx context.Context, env evergreen.Environment, versionID string, taskIDs []string) ([]Task, error) {
-	return findCompletedTasks(ctx, env, bson.M{VersionKey: versionID}, taskIDs)
+func FindCompletedTasksByVersion(ctx context.Context, versionID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, bson.M{VersionKey: versionID}, taskIDs)
 }
 
 // FindCompletedTasksByBuild returns all completed tasks belonging to the
 // given build ID. Excludes execution tasks. If no taskIDs are specified, all
 // completed tasks belonging to the build are returned.
-func FindCompletedTasksByBuild(ctx context.Context, env evergreen.Environment, buildID string, taskIDs []string) ([]Task, error) {
-	return findCompletedTasks(ctx, env, bson.M{BuildIdKey: buildID}, taskIDs)
+func FindCompletedTasksByBuild(ctx context.Context, buildID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, bson.M{BuildIdKey: buildID}, taskIDs)
 }
 
-func findCompletedTasks(ctx context.Context, env evergreen.Environment, filter bson.M, taskIDs []string) ([]Task, error) {
+func findCompletedTasks(ctx context.Context, filter bson.M, taskIDs []string) ([]Task, error) {
 	filter[StatusKey] = bson.M{"$in": evergreen.TaskCompletedStatuses}
 	filter[DisplayTaskIdKey] = ""
 	if len(taskIDs) > 0 {
@@ -2907,7 +2907,7 @@ func findCompletedTasks(ctx context.Context, env evergreen.Environment, filter b
 		}
 	}
 
-	cur, err := env.DB().Collection(Collection).Find(ctx, filter)
+	cur, err := evergreen.GetEnvironment().DB().Collection(Collection).Find(ctx, filter)
 	if err != nil {
 		return nil, errors.Wrap(err, "finding completed tasks")
 	}
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -2883,18 +2883,18 @@ func abortAndMarkResetTasks(ctx context.Context, filter bson.M, taskIDs []string
 	return err
 }
 
-// FindCompletedTasksByVersion returns all completed tasks belonging to the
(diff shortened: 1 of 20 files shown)


model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+	filter[StatusKey] = bson.M{"$in": evergreen.TaskInProgressStatuses}
+	if len(taskIDs) > 0 {
+		filter["$or"] = []bson.M{
+			{IdKey: bson.M{"$in": taskIDs}},
+			{DisplayTaskIdKey: bson.M{"$in": taskIDs}},
+			{ExecutionTasksKey: bson.M{"$in": taskIDs}},
+		}
+	}
+
+	_, err := env.DB().Collection(Collection).UpdateMany(
```
ablack12, line 2873: I'm pretty against having to pass env all the way in here; is there any particular reason we can't just use db?

model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+	)
+
+	return err
+}
+
+// FindCompletedTasksByVersion returns all completed tasks belonging to the
+// given version ID. Excludes execution tasks. If no task IDs are specified,
+// all completed tasks belonging to the version are returned.
+func FindCompletedTasksByVersion(ctx context.Context, env evergreen.Environment, versionID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, env, bson.M{VersionKey: versionID}, taskIDs)
```
ablack12, line 2890: We can use ByVersion

model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+}
+
+// FindCompletedTasksByBuild returns all completed tasks belonging to the
+// given build ID. Excludes execution tasks. If no taskIDs are specified, all
+// completed tasks belonging to the build are returned.
+func FindCompletedTasksByBuild(ctx context.Context, env evergreen.Environment, buildID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, env, bson.M{BuildIdKey: buildID}, taskIDs)
```
ablack12, line 2897: We can use ByBuildId

model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+	filter[StatusKey] = bson.M{"$in": evergreen.TaskCompletedStatuses}
+	filter[DisplayTaskIdKey] = ""
+	if len(taskIDs) > 0 {
+		filter["$or"] = []bson.M{
+			{IdKey: bson.M{"$in": taskIDs}},
+			{ExecutionTasksKey: bson.M{"$in": taskIDs}},
+		}
+	}
+
+	cur, err := env.DB().Collection(Collection).Find(ctx, filter)
```
ablack12, line 2910: similar preference not to need env

ablack12 (changes requested):

model/task/db.go
julianedwards: So in Cedar we always prefer passing down the env in the same way that we pass down the context. This allows us to pass a context (with timeouts and cancels) to the DB functions as well as have more control over the environment for testing and local development.

I spoke with @Kimchelly to confirm that this was the preference here. Happy to follow whatever convention app has!

ablack12: We can pass the context without the env though, and this adds a lot of overhead, especially considering in most of the functions we're just calling evergreen.GetEnvironment, which we can just do here. I'll let Kim weigh in.

model/task/db.go
```diff
@@ -1905,41 +1905,41 @@
...
-func AbortAndMarkResetTasksForVersion(versionId string, taskIds []string, caller string) error {
-	_, err := UpdateAll(
-		bson.M{
-			VersionKey: versionId, // Include to improve query.
-			IdKey:      bson.M{"$in": taskIds},
-			StatusKey:  bson.M{"$in": evergreen.TaskInProgressStatuses},
-		},
+	_, err := env.DB().Collection(Collection).UpdateMany(
+		ctx,
+		filter,
```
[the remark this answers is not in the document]
ablack12: what in the world

model/task/db.go
ablack12: Well I guess we can't really use the context, but still we aren't actually using a special value for either of these really anywhere in the codebase. If we're going to introduce a change like this I think we should have a team discussion about it; I'd really prefer to keep this simple for now.

Kimchelly: The reason to prefer using the MongoDB client rather than use Anser is because Anser was meant to be glue code to make it less difficult to migrate off of the legacy Go driver (but we really shouldn't use it anymore if we can help it). The only reason we use Anser is because of the legacy code that largely doesn't pass in the context, even though it should. Since Julian already bothered to plumb the ctx/env down to the query, I think it's alright to keep the PR like this, and I do agree with Julian's reasoning that passing an explicit ctx/env is much better for testing + constraining the query duration to the lifetime of the request.

I didn't review any of the rest of the PR.

ablack12: This is going to mean we essentially need to pass env and context like everywhere if we want to make this consistent in the codebase. I think that's messy enough to not have enough gain. Can't we just construct the env here, even if we don't want to use the db package? Idk if I'm alright with making the codebase inconsistent just bc the plumbing is done; this would be easy to un-plumb.

model/task/db.go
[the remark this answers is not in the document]
julianedwards: Yeah I saw this was interesting... I mean all of our queries rely on user provided values. I guess it is worried about query injection haha

model/task/db.go
julianedwards: So I think the idea is to start improving the code by passing the env where we can. Wrt to messiness, are you concerned about the extra param or the plumbing? We already set up the env in our REST apps (both in `rest` and `service`) so we get it for free in most places—and I would like to set it up in the graphql resolver soon. Once the graphql case is covered, we will always have it for our DB funcs. We do this in Cedar and it has worked really well.

FYI: I already started passing in the env to the  new task methods that I have written for fetching task output (for example, [here](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L2609-L2649)).

Kimchelly: If we want some in the middle option, we can make it so that the common usage of env where the env is specifically used in a DB query and specifically just to call `env.DB().Collection().<METHOD>` (like here) can forego passing the env down from the top-level. This wouldn't apply in situations where you want to use the env for other things like enqueueing Amboy jobs (because the default Amboy testing queue state is globally affected by tests in a way that should use the non-default env) or getting service flags, but would still make it convenient enough for the simple case of accessing the DB to CRUD something.

ablack12: Both the extra param and the plumbing considering the extra param has to start being passed typically super early in the stack and then isn't used until later. I'm in favor of foregoing passing env down from top level for these kinds of simple cases.

julianedwards: Yeah for DB to CRUD data, I think this middle option is fine if you feel strongly about this. I still think passing the env from the top level is a good practice, overall. Much of the refactoring I have been working on is aiming towards isolating components of the system to reduce dependency nightmares and bug-prone spaghetti logic. Using small/clean interfaces encapsulates and isolates these deps and implementation logic, but there often needs to be something that tells the application what version/implementation of the interfaces/components to use (essentially service location). Using the env is a really clean and uniform way of doing that, it also allows us to use different implementations for testing and local development (like in smoke tests). We can easily swap out things like the type of test results storage (DB vs Cedar) or the bucket location for logs (filesystem vs S3) without compromising other parts of the system. And when we do need to update implementation or swap out the underlying services entirely, only the component itself needs to be updated reducing risk of introduce unrelated bugs elsewhere in the codebase. Finally, with this architecture the application can handle multiple implementations of the same component simultaneously, facilitating cutovers and flexibility with the third party services we use.

ablack12: It's really just that if we have to start passing the env around for _every_ DB call, we're going to be riddled with envs. I guess that's fine if you really think the improvements are worth it but it seems to me for DB calls to just be doing something for the sake of it rather than getting anything out of it. We call db.Whatever _everywhere_ so I'd rather we just stick with that (or change it everywhere but that would be _so_ many envs when we don't really need any special sort of environment for DB calls).

model/task/db.go
ablack12: These have thumbs up but maybe a commit hasn't been pushed?

model/task/db.go
Kimchelly: I made EVG-20917 to address changing the style conventions to more formally delineate when/where it's appropriate to use `evergreen.GetEnvironment` vs passing down the env.

For this PR, since the env is only used for DB ops, the style guide says it's alright to use `evergreen.GetEnvironment` at the exact spot where you need to make the DB call.

julianedwards: Ok I updated the code to fetch the env within the DB calls!

model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+	)
+
+	return err
+}
+
+// FindCompletedTasksByVersion returns all completed tasks belonging to the
+// given version ID. Excludes execution tasks. If no task IDs are specified,
+// all completed tasks belonging to the version are returned.
+func FindCompletedTasksByVersion(ctx context.Context, versionID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, bson.M{VersionKey: versionID}, taskIDs)
```
ablack12, line 2890: also ByVersion

model/task/db.go
```diff
@@ -2882,3 +2844,78 @@ func CountNumExecutionsForInterval(input NumExecutionsForIntervalInput) (int, er
...
+}
+
+// FindCompletedTasksByBuild returns all completed tasks belonging to the
+// given build ID. Excludes execution tasks. If no taskIDs are specified, all
+// completed tasks belonging to the build are returned.
+func FindCompletedTasksByBuild(ctx context.Context, buildID string, taskIDs []string) ([]Task, error) {
+	return findCompletedTasks(ctx, bson.M{BuildIdKey: buildID}, taskIDs)
```
ablack12, line 2897: also byBuildId

ablack12 (approved):
LGTM with nits

model/task/db.go
julianedwards: Oh damn I missed these!

julianedwards: Done.

model/task/db.go
julianedwards: Done.

PR #7060, opened 2023-09-21
EVG-20721 Add chmod before rm to allow for all/any file(s) to be deleted

ZackarySantana:
EVG-20721

### Description
We recently had some errors and tasks failing because of git.get_project not having permissions to delete files. This adds permissions to to delete the files.

### Testing
Deployed in stage and tested removing, adding, and both with agent.


agent/command/git.go
```diff
@@ -579,6 +579,7 @@ func (c *gitFetchProject) fetchSource(ctx context.Context,
 	if err != nil {
 		return err
 	}
+	gitCommands = append([]string{fmt.Sprintf("chmod +w %s", c.Directory)}, gitCommands...)
```
Kimchelly, line 582: This solution is somewhat awkward + non-obvious because it tweaks the git commands at the last second, well after the problem line (`rm -rf`), so I'm not sure we can trust this is a stable long-term solution. It's preferrable to move this into `buildCloneCommand` and then either 1. quickfix the tests (to more fully complete and resolve this ticket) and test this change, or 2. fix the test command slice indexing and test this change.

Kimchelly, line 582: Also, this needs to `chmod -R` because `rm -rf` can fail due to permissions if any file inside the directory doesn't have write permissions.

Kimchelly (changes requested):

agent/command/git.go
```diff
@@ -565,7 +565,7 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.Int(cloneRetriesAttribute, attemptNum))
 
-	return err
+	return errors.Wrapf(err, "retrying git.get_project failed")
```
Kimchelly, line 568: 1. This is duplicate wrapping since the higher-level logic already wraps the error with the failing command name.
2. This should follow the style guide for error wrapping: `errors.Wrap(err, "retrying git clone`).

agent/command/git.go
```diff
@@ -308,6 +308,8 @@ func (c *gitFetchProject) buildCloneCommand(ctx context.Context, comm client.Com
 	gitCommands := []string{
 		"set -o xtrace",
 		"set -o errexit",
+		fmt.Sprintf("mkdir -p %s", c.Directory),
```
Kimchelly, line 311: This is actually a little interesting because in the original ticket, we stated that we may not have the necessary permissions to `rm -rf` the directory, so this `mkdir -p` can also fail for the same reasons as `rm -rf` due to permissions. If the directory structure doesn't exist yet, this will work, but if one of the parent directories _does_ exist and the agent doesn't have sufficient permissions to create a subdirectory within that directory, then `mkdir` will fail with a permissions error. For example:
```bash
mkdir -p parent # Make parent directory
chmod 111 parent # Make the directory read-only, causing later clone attempts to fail
// Later on, try cloning...

mkdir -p parent/child # Making the child directory will fail
chmod -R 755 parent/child # We won't get to chmod since the directory creation failed
```

Since we know this is a shell script, one thing we could do is move the `chmod` _before_ `set -o errexit`, so the script continues even if chmod fails due a nonexistent directory, and still fixes the permissions if it is able to. (And if we can't chmod for some other reason, we're doomed no matter what we try since we can't remove the directory at all).

If the above sounds reasonable, I'd still double-check in staging to make sure the script continues on error.

ZackarySantana: Hm but if that's the case, if the parent directory exists but the child does not exist, we would need to do `chmod -R 755 parent` to make sure we can do mkdir -p parent/child inside that parent directory

Would something like:

chmod -R 755 dir/..
set -o errexit 
mkdir -p dir

Be good? Essentially, just moving the chmod above the set and pushing it back a directory?

Kimchelly: I guess my point is that `mkdir -p` doesn't really solve anything for this command when `chmod` is already a best-effort attempt to fix directory permissions. If I understand, the only reason `mkdir -p` is being used here is because the directory may not exist, causing `chmod` to fail. If you do the above ordering, then `mkdir -p` is no longer needed since you're about to remove the directory anyways.

As long as we have `chmod` continue on error (i.e. above `set -o errexit`), the mkdir is not needed. If it fails because the directory doesn't exist, it's fine because it'll continue on error. But if it fails because it's unable to change the permissions, then there's nothing we can do, and continuing on error is not a problem (`rm -rf` will fail later anyways, so the additional `mkdir -p` won't have any effect on git.get_project succeeding).

ZackarySantana: Ahh that makes sense, ty!

ZackarySantana:
Deployed on an agent, the chmod doesn't stop the commands: https://spruce-staging.corp.mongodb.com/task/evg_lint_generate_lint_patch_9f3350a1017ac4c63faa2480e273af4c7706a294_65173b34b237367150586c85_23_09_29_21_01_46/logs?execution=0&logtype=agent

Kimchelly (approved):

hadjri (approved):

PR #7061, opened 2023-09-21
EVG-20914 Use parsley log link when there's no HTML link 

malikchaya2:
EVG-20914 

### Description
A [BF ticket ](https://jira.mongodb.org/browse/BF-28597) that was generated included empty log links for tests that didn't have HTML logs. We should check if the log is empty, and if it is, use parsley.

### DIFF
--- a/trigger/task_jira.go
+++ b/trigger/task_jira.go
@@ -386,9 +386,14 @@ func (j *jiraBuilder) getDescription() (string, error) {
 	tests := []jiraTestFailure{}
 	for _, test := range j.data.Task.LocalTestResults {
 		if test.Status == evergreen.TestFailedStatus {
+			url := test.GetLogURL(evergreen.GetEnvironment(), evergreen.LogViewerHTML)
+			if url == "" {
+				url = test.GetLogURL(evergreen.GetEnvironment(), evergreen.LogViewerParsley)
+			}
+
 			tests = append(tests, jiraTestFailure{
 				Name:       cleanTestName(test.GetDisplayTestName()),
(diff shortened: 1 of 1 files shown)


trigger/task_jira.go
```diff
@@ -386,9 +386,14 @@ func (j *jiraBuilder) getDescription() (string, error) {
 	tests := []jiraTestFailure{}
 	for _, test := range j.data.Task.LocalTestResults {
 		if test.Status == evergreen.TestFailedStatus {
+			url := test.GetLogURL(evergreen.GetEnvironment(), evergreen.LogViewerHTML)
+			if url == "" {
+				url = test.GetLogURL(evergreen.GetEnvironment(), evergreen.LogViewerParsley)
```
Kimchelly, line 391: If we're going to call `evergreen.GetEnvironment()`, it would be preferable to call it once and store it as a var.

Kimchelly (changes requested):
Is there a test for this?

trigger/task_jira.go
malikchaya2: done. 
And re a test: It's not easy to test, and it's simple enough that I think it's okay without one.

Kimchelly: Hmm, it seems like this test would rely on global state from the env because it uses UI settings from `evergreen.GetEnvironment()`... I guess it's alright to not test it given that it relies on global testing env state.

cc @ablack12 since this is another case of `evergreen.GetEnvironment` making changes untestable (there are a lot of other ones, but this is one example).

Kimchelly (approved):

PR #7062, opened 2023-09-22
EVG-20296 Allow project triggers to be triggered without a upstream version being created

hadjri:
EVG-20296

### Description
This change introduces a new type of trigger, "push", to go alongside the task and build triggers. This allows for projects to configure upstream projects for repos that don't run Evergreen tasks, by introducing logic into the github webhook that will trigger a downstream version immediately upon a push event from the upstream repo.

This change also includes some necessary refactoring to accomodate for the fact that `SourceVersion` for this new type of trigger will be nil since there is no actual upstream Evergreen version.

This will require a followup Spruce PR to enable configuring this new type of trigger in the UI.
### Testing
In staging, I configured the sandbox project to have a [project trigger of this new push type](https://spruce-staging.corp.mongodb.com/project/sandbox/settings/project-triggers) on a new project that tracks a different branch in the sandbox project, `test-trigger`. Confirmed that upon a push to the `test-trigger` branch a [version for the sandbox project is triggered](https://evergreen-staging.corp.mongodb.com/rest/v2/versions/sandbox_d631380317a8478b879cc27d6171508dea2b11df_c6db985b8c9ffb231c10085b09efa348), even though it is an empty project where repotracker is disabled and no config file is defined

Added unit test
### Documentation
Will need to update Docusaur trigger docs on this new type upon approval


rest/data/repotracker.go
```diff
@@ -101,6 +102,19 @@ func TriggerRepotracker(ctx context.Context, q amboy.Queue, msgID string, event
...
+			if refs[i].IsRepotrackerDisabled() {
+				triggeredVersions, err := trigger.TriggerDownstreamProjectsForPush(ctx, refs[i].Id, event, trigger.TriggerDownstreamVersion)
+				catcher.Wrapf(err, "triggering downstream projects for push event for project '%s'", refs[i].Id)
+				versions := []string{}
+				for _, version := range triggeredVersions {
+					versions = append(versions, version.Id)
+				}
+				grip.InfoWhen(len(versions) > 0, message.Fields{
+					"source":   "GitHub hook",
+					"message":  fmt.Sprintf("triggered versions for push event for project '%s'", refs[i].Id),
```
ablack12, line 114: preference to just have project_id be a message field, so it's easier to grep on message if we want to

rest/data/repotracker.go
```diff
@@ -101,6 +102,19 @@ func TriggerRepotracker(ctx context.Context, q amboy.Queue, msgID string, event
 	for i := range refs {
 		if !refs[i].DoesTrackPushEvents() || !refs[i].Enabled || refs[i].IsRepotrackerDisabled() {
 			unactionable = append(unactionable, refs[i].Id)
+			if refs[i].IsRepotrackerDisabled() {
+				triggeredVersions, err := trigger.TriggerDownstreamProjectsForPush(ctx, refs[i].Id, event, trigger.TriggerDownstreamVersion)
```
ablack12, line 106: Can this function log so that we don't need to return the versions?

trigger/process.go
```diff
@@ -96,7 +97,13 @@ type ProcessorArgs struct {
 	EventID                      string
 	DefinitionID                 string
 	Alias                        string
+	ProjectID                    string
 	UnscheduleDownstreamVersions bool
+	PushInfo                     PushInfo
+}
+
+type PushInfo struct {
```
ablack12, line 105: Should comment on this as a public struct (and/or make private); would be curious to know why we went with a struct her also

trigger/process.go
```diff
@@ -289,3 +296,57 @@ projectLoop:
...
+// have repotracker enabled.
+func TriggerDownstreamProjectsForPush(ctx context.Context, projectId string, event *github.PushEvent, processor projectProcessor) ([]model.Version, error) {
+	downstreamProjects, err := model.FindDownstreamProjects(projectId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "finding downstream projects of project '%s'", projectId)
+	}
+
+	catcher := grip.NewBasicCatcher()
+	versions := []model.Version{}
+projectLoop:
```
ablack12, line 310: Do we need this? I think we can just break at line 347

trigger/process.go
```diff
@@ -289,3 +296,57 @@ projectLoop:
...
+	for _, ref := range downstreamProjects {
+
+		for _, trigger := range ref.Triggers {
+			if trigger.Level != model.ProjectTriggerLevelPush {
```
ablack12, line 314: preference to combine with 317 since these are short if statements

trigger/process.go
```diff
@@ -289,3 +296,57 @@ projectLoop:
 
 	return versions, catcher.Resolve()
 }
+
+// TriggerDownstreamProjectsForPush triggers downstream projects for a push event from a repo that does not
```
ablack12, line 300: Do we intend to make repotracker disabled a requirement on the UI? Is there a reason why they couldn't define it this way with repotracker enabled?

ablack12 (changes requested):
I didn't get all the way through my review but wanted to post what I have:
Existing triggered versions mention what project they were triggered from in the task details page and on the waterfall ([example](https://spruce.mongodb.com/version/spruce_1c5c69285fe152a5180102f61d617d91a74de882_ec1a0750bc24e181123a204a3aa6dc6b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) which i don't see for your version, so it may be missing some metadata.

trigger/process.go
hadjri: Nope, will remove - I was following the patterns from `triggerDownstreamProjectsForBuild` and `triggerDownstreamProjectsForTask`. I changed them both to remove the projectLoop as well

trigger/process.go
hadjri: Ah, during testing I believe I was using more fields here. I'll just remove this struct entirely.

trigger/process.go
hadjri: That was what I was envisioning, where disabling repotracker and having no config file would indicate that it was an empty project that only exists to trigger downstream versions from its pushes.  But I guess it is overly limiting, and not really necessary to do it that way.

I also for some reason thought it wasn't possible to enable repotracker with an empty config file in the settings but it looks like that's not the case

hadjri:
> Existing triggered versions mention what project they were triggered from in the task details page and on the waterfall ([example](https://spruce.mongodb.com/version/spruce_1c5c69285fe152a5180102f61d617d91a74de882_ec1a0750bc24e181123a204a3aa6dc6b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)) which i don't see for your version, so it may be missing some metadata.

This is enabled [here](https://gist.github.com/hadjri/bc73e0dc6860f58f1a2cd4100c656a89#file-index-tsx-L71-L83) and [here](https://gist.github.com/hadjri/9f8bd786d470b489a3a94336b498bd40#file-metadata-tsx-L169-L180) on the Spruce side, so I think the work to fix this will need to be done in a followup UI ticket to accomodate this new push trigger type.

trigger/process.go
```diff
@@ -283,9 +286,68 @@ projectLoop:
...
+	}
+
+	versionIds := []string{}
+	for _, v := range versions {
+		versionIds = append(versionIds, v.Id)
+	}
+	grip.InfoWhen(len(versions) > 0, message.Fields{
+		"source":     "GitHub hook",
+		"message":    "triggered versions for push event for project",
+		"versions":   versions,
```
ablack12, line 348: should just log ids here

trigger/project_triggers.go
```diff
@@ -41,20 +42,28 @@ func TriggerDownstreamVersion(ctx context.Context, args ProcessorArgs) (*model.V
 	if err != nil {
 		return nil, errors.Wrap(err, "creating version")
 	}
-	err = args.SourceVersion.AddSatisfiedTrigger(args.DefinitionID)
-	if err != nil {
-		return nil, err
+	if args.SourceVersion != nil {
+		err = args.SourceVersion.AddSatisfiedTrigger(args.DefinitionID)
```
ablack12, line 46: nit, combine with 47

trigger/project_triggers.go
```diff
@@ -41,20 +42,28 @@ func TriggerDownstreamVersion(ctx context.Context, args ProcessorArgs) (*model.V
...
+		}
 	}
 	settings, err := evergreen.GetConfig(ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "getting Evergreen settings")
 	}
-	upstreamProject, err := model.FindMergedProjectRef(args.SourceVersion.Identifier, args.SourceVersion.Id, true)
+	var versionID string
+	projectID := args.ProjectID
+	if projectID == "" {
```
ablack12, line 57: Sorry if I didn't catch this earlier but this change is a bit confusing to me; wouldn't we want to assign versionID to args.SourceVersion.Id unless it's empty? Where does args.ProjectID come from that we wouldn't want to overwrite it?

ablack12: Might be worth adding some comments here so we remember this special case

trigger/project_triggers.go
```diff
@@ -104,9 +118,13 @@ func metadataFromVersion(args ProcessorArgs) (model.VersionMetadata, error) {
 		return metadata, errors.Errorf("repo '%s' not found", args.DownstreamProject.Id)
 	}
 	metadata.Revision.Revision = repo.LastRevision
-	author, err := user.FindOneById(args.SourceVersion.AuthorID)
+	var authorId string
+	if args.SourceVersion != nil {
+		authorId = args.SourceVersion.AuthorID
+	}
+	author, err := user.FindOneById(authorId)
```
ablack12, line 125: I think we should only call this in the if statement as well, or else it's a waste of a DB call; we could just set author as a var outside so we can still use the if author != nil check either way

ablack12 (changes requested):
Can we link a follow up ticket for Spruce to this ticket to complete the UI change?

trigger/process.go
hadjri: Created! https://jira.mongodb.org/browse/EVG-20985

hadjri:
> Can we link a follow up ticket for Spruce to this ticket to complete the UI change?

Yep, was waiting on discussion here confirming we should make one. Created now: EVG-20987

trigger/project_triggers.go
hadjri: Sure, I added a clarifying comment

trigger/process.go
```diff
@@ -332,20 +332,15 @@ func TriggerDownstreamProjectsForPush(ctx context.Context, projectId string, eve
...
-	versionIds := []string{}
-	for _, v := range versions {
-		versionIds = append(versionIds, v.Id)
-	}
-	grip.InfoWhen(len(versions) > 0, message.Fields{
+	grip.InfoWhen(len(versionIds) > 0, message.Fields{
 		"source":     "GitHub hook",
 		"message":    "triggered versions for push event for project",
-		"versions":   versions,
+		"versions":   versionIds,
```
ablack12, line 343: nit: version_ids

trigger/project_triggers.go
```diff
@@ -53,6 +52,9 @@ func TriggerDownstreamVersion(ctx context.Context, args ProcessorArgs) (*model.V
 		return nil, errors.Wrap(err, "getting Evergreen settings")
 	}
 	var versionID string
+	// For build and task level triggers, there is a source version from which we can extract a project ID.
+	// For push triggers, since there is no source version and the ProjectID field gets populated from
```
ablack12, line 56: this isn't quite a real sentence, could you tweak? Maybe like " Since there is no source version for push triggers,  the ProjectID field gets populated from..."

ablack12 (approved):
LGTM -- if you aren't intending to do the UI work yourself though, I would add a lot more detail to that ticket about what needs doing (you linked some files in this PR for example that I imagine would be helpful).

hadjri:
> LGTM -- if you aren't intending to do the UI work yourself though, I would add a lot more detail to that ticket about what needs doing (you linked some files in this PR for example that I imagine would be helpful).

I am picking up the UI ticket next sprint, but just for ticket documentation purposes I added two relevant links to the ticket.

PR #7064, opened 2023-09-22
docs: add example of expansions.write output

Kimchelly:
[Slack thread](https://mongodb.slack.com/archives/C0V896UV8/p1695410142917209)

Someone asked for an example of the output format for the `expansions.write` file. Also, there's already [a test for the exact output format](https://github.com/evergreen-ci/evergreen/blob/ded75d09f9a7ac790d11fe0dd42dad9e712b6a0d/agent/command/expansion_test.go#L76), so we don't need to add any new tests for this.

### DIFF
--- a/docs/Project-Configuration/Project-Commands.md
+++ b/docs/Project-Configuration/Project-Commands.md
@@ -228,11 +228,11 @@ Parameters:
 
 ## expansions.write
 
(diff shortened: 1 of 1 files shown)


bynn (approved):
looks good!

PR #7072, opened 2023-09-25
EVG-20924 add higher context cancel for commit queue

ablack12:
EVG-20924

### Description
As discussed in ops, increase timeout and see how that goes.

### DIFF
--- a/rest/data/commit_queue.go
+++ b/rest/data/commit_queue.go
@@ -132,7 +132,7 @@ func getPatchInfo(ctx context.Context, settings *evergreen.Settings, githubToken
 	}
 
 	// Fetch the latest config file.
-	// Temporarily set a high timeout.
+	// Set a higher timeout for this operation. 
 	fetchCtx, ctxCancel := context.WithTimeout(context.Background(), 5*time.Minute)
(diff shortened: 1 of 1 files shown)


rest/data/commit_queue.go
```diff
@@ -131,8 +131,11 @@ func getPatchInfo(ctx context.Context, settings *evergreen.Settings, githubToken
 		return "", nil, nil, nil, errors.Wrap(err, "getting GitHub PR diff")
 	}
 
-	// fetch the latest config file
-	config, patchConfig, err := model.GetPatchedProject(ctx, settings, patchDoc, githubToken)
+	// Fetch the latest config file.
+	// Temporarily set a high timeout.
```
hadjri, line 135: Seems like the plan is to make this permanent if it succeeds right? If so can we mention [EVG-20924](https://jira.mongodb.org/browse/EVG-20924) in the comment here with a todo to make it untemporary if it works?

ablack12: true, i might just remove the word temporary since i'll just remove it if it doesn't work as part of the ticket

PR #7076, opened 2023-09-25
EVG-20317 Stop on error if module cloning fails

hadjri:
EVG-20317

### Description
Now returning an err upon failure to clone a module rather than continuing the task's execution. 

The ticket's discussion mentions introducing a feature flag in case that this breaks some behavior server is depending on, but since this is an agent side change I believe we should just be ready to revert if needed.


Kimchelly (approved):
LGTM - no rollback feature flag also SGTM since toggling the feature flag doesn't offer substantial benefits over just reverting the deploy.

PR #7079, opened 2023-09-26
EVG-20014: Implement Cedar-backed log fetching for task and test log output

julianedwards:
EVG-20014

### Description
Implement Cedar-backed log fetching for the new `TaskOutput` interface.
Nothing will actually start using this yet until EVG-20019 is complete.

### Testing
Tested in staging.

### Documentation
N/A

### DIFF
--- a/apimodels/cedar_models.go
+++ b/apimodels/cedar_models.go
@@ -46,6 +46,7 @@ type GetBuildloggerLogsOptionsV2 struct {
 // log iterator.
 // TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
 // and the original GetBuildloggerLogs function is deleted.
+// TODO (EVG-21010): Remove this once Cedar logs have TTL'ed.
 func GetBuildloggerLogsV2(ctx context.Context, opts GetBuildloggerLogsOptionsV2) (log.LogIterator, error) {
 	usr := gimlet.GetUser(ctx)
 	if usr == nil {
@@ -85,6 +86,7 @@ func GetBuildloggerLogsV2(ctx context.Context, opts GetBuildloggerLogsOptionsV2)
 	return newBuildloggerIterator(r), nil
 }
 
+// TODO (EVG-21010): Remove this once Cedar logs have TTL'ed.
 type buildloggerIterator struct {
 	readCloser io.ReadCloser
 	reader     *bufio.Reader
--- a/apimodels/cedar_models.go
+++ b/apimodels/cedar_models.go
@@ -42,8 +42,8 @@ type GetBuildloggerLogsOptionsV2 struct {
 	Tail      int      `json:"-"`
 }
 
-// GetBuildloggerLogs makes request to Cedar for a specifc log and returns an
-// io.ReadCloser.
+// GetBuildloggerLogs makes request to Cedar for a specifc log and returns a
+// log iterator.
 // TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
 // and the original GetBuildloggerLogs function is deleted.
 func GetBuildloggerLogsV2(ctx context.Context, opts GetBuildloggerLogsOptionsV2) (log.LogIterator, error) {
--- a/apimodels/cedar_models.go
+++ b/apimodels/cedar_models.go
@@ -1,9 +1,155 @@
 package apimodels
 
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/evergreen-ci/evergreen/model/log"
+	"github.com/evergreen-ci/gimlet"
+	"github.com/evergreen-ci/timber"
+	"github.com/evergreen-ci/timber/buildlogger"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/level"
+	"github.com/pkg/errors"
+)
+
 type CedarConfig struct {
 	BaseURL  string `json:"base_url"`
 	RPCPort  string `json:"rpc_port"`
 	Username string `json:"username"`
 	APIKey   string `json:"api_key,omitempty"`
 	Insecure bool   `json:"insecure"`
+}
+
+// GetBuildloggerLogsOptions represents the arguments passed into the
+// GetBuildloggerLogs function.
+// TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
+// and the original GetBuildloggerLogsOptions struct is deleted.
+type GetBuildloggerLogsOptionsV2 struct {
+	BaseURL   string   `json:"_"`
+	TaskID    string   `json:"-"`
+	Execution *int     `json:"-"`
+	TestName  string   `json:"-"`
+	Tags      []string `json:"-"`
+	Start     int64    `json:"-"`
+	End       int64    `json:"-"`
+	Limit     int      `json:"-"`
+	Tail      int      `json:"-"`
+}
+
+// GetBuildloggerLogs makes request to Cedar for a specifc log and returns an
+// io.ReadCloser.
+// TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
+// and the original GetBuildloggerLogs function is deleted.
+func GetBuildloggerLogsV2(ctx context.Context, opts GetBuildloggerLogsOptionsV2) (log.LogIterator, error) {
+	usr := gimlet.GetUser(ctx)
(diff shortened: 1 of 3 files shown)


apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+	if usr == nil {
+		return nil, errors.New("error getting user from context")
+	}
+
+	var start, end time.Time
+	if opts.Start > 0 {
+		start = time.Unix(0, opts.Start).UTC()
+	}
+	if opts.End > 0 {
+		end = time.Unix(0, opts.Start).UTC()
```
Kimchelly, line 60: `opts.End`?

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+	End       int64    `json:"-"`
+	Limit     int      `json:"-"`
+	Tail      int      `json:"-"`
+}
+
+// GetBuildloggerLogs makes request to Cedar for a specifc log and returns a
+// log iterator.
+// TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
+// and the original GetBuildloggerLogs function is deleted.
+func GetBuildloggerLogsV2(ctx context.Context, opts GetBuildloggerLogsOptionsV2) (log.LogIterator, error) {
```
Kimchelly, line 49: Is there a way to test this?

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+	BaseURL   string   `json:"_"`
+	TaskID    string   `json:"-"`
+	Execution *int     `json:"-"`
+	TestName  string   `json:"-"`
+	Tags      []string `json:"-"`
+	Start     int64    `json:"-"`
+	End       int64    `json:"-"`
+	Limit     int      `json:"-"`
+	Tail      int      `json:"-"`
+}
```
Kimchelly, line 43: Also "-" for BSON tags?

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+// TODO (EVG-20019): Remove the "V2" suffix once log fetching is consolidated
+// and the original GetBuildloggerLogsOptions struct is deleted.
+type GetBuildloggerLogsOptionsV2 struct {
+	BaseURL   string   `json:"_"`
+	TaskID    string   `json:"-"`
+	Execution *int     `json:"-"`
+	TestName  string   `json:"-"`
+	Tags      []string `json:"-"`
+	Start     int64    `json:"-"`
+	End       int64    `json:"-"`
```
Kimchelly, line 40: Is it possible to have this instead pass `time.Time`s rather than int64s representing Unix nanoseconds? It already gets converted into a `time.Time` anyways.

If not, can we document that these are nanoseconds since the Unix epoch?

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+type buildloggerIterator struct {
+	readCloser io.ReadCloser
+	reader     *bufio.Reader
+	item       log.LogLine
+	catcher    grip.Catcher
+	exhausted  bool
+	closed     bool
+}
+
+func newBuildloggerIterator(r io.ReadCloser) *buildloggerIterator {
```
Kimchelly, line 97: Can we unit test the iterator? Especially since it seems like iteration is highly sensitive to having the proper format log lines.

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+
+	line, err := it.reader.ReadString('\n')
+	if err != nil {
+		it.exhausted = err == io.EOF
+		it.catcher.AddWhen(err != io.EOF, errors.Wrap(err, "reading log lines"))
+		return false
+	}
+
+	// Parse prefixed priority with format "[P:3%d] %s".
+	priority, err := strconv.Atoi(strings.TrimSpace(line[3:6]))
```
Kimchelly, line 118: Are we confident that `line` will definitely have this length, and there's no way to get a log line that won't have this format? Otherwise it'll panic.

If we're not 100% confident that the log lines will always be formatted like this, I think we should length check it, and also verify that the log line follows the expected format.

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+	if err != nil {
+		it.catcher.Wrap(err, "parsing log line priority")
+		return false
+	}
+	line = line[8:]
```
Kimchelly, line 123: Same thing as above about validating format.

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+
+	// Parse prefixed timestamp with format "[2006/01/02 15:04:05.000] %s".
+	ts, err := time.Parse("2006/01/02 15:04:05.000", line[1:24])
```
Kimchelly, line 126: Same thing as above about validating format.

apimodels/cedar_models.go
```diff
@@ -1,9 +1,155 @@
...
+	if err != nil {
+		it.catcher.Wrap(err, "parsing log line timestamp")
+		return false
+	}
+	line = line[26:]
```
Kimchelly, line 131: Same thing as above about validating format.

Kimchelly (changes requested):

apimodels/cedar_models.go
julianedwards: Since this is `apimodels` do we need to add the BSON tags? I thought it was just JSON.

apimodels/cedar_models.go
julianedwards: This only gets used by `taskoutput` logs where the start and end times are `int64`s representing Unix nanoseconds, [for example](https://github.com/evergreen-ci/evergreen/blob/main/taskoutput/task_log.go#L46-L51). I can add documentation, but the idea is that no other part of the code should use this (it will also go away in a year or so after we cut over and logs are TTL'ed).

apimodels/cedar_models.go
julianedwards: Good catch!

apimodels/cedar_models.go
julianedwards: Yeah I can write tests for this! We never wrote tests for [the code](https://github.com/evergreen-ci/evergreen/blob/main/apimodels/task_log.go#L126-L184) that this replaces or for any code that lives in `apimodels` for that matter, I just always test in staging. Since it is only reading data for rendering, the fixes are quick to catch in staging and easy to fix if a bug does go out.

apimodels/cedar_models.go
julianedwards: Good idea! Added a check above that will fail the iterator if the line does not have at least 34 characters.

apimodels/cedar_models.go
julianedwards: See above.

apimodels/cedar_models.go
julianedwards: See above.

apimodels/cedar_models.go
julianedwards: I can add tests but we don't really test this package, see my [comment above](https://github.com/evergreen-ci/evergreen/pull/7079#discussion_r1342940545).

apimodels/cedar_models.go
Kimchelly: If these are going away in a year, then this SGTM.

Since it's a temporary piece of logic, can we have a ticket to track deletion of this code in a year?

apimodels/cedar_models.go
Kimchelly: Eh, if the other API models don't include it, I guess it's fine to not have them here.

apimodels/cedar_models.go
Kimchelly: SGTM given that this code will be deleted in a year.

apimodels/cedar_models.go
Kimchelly: SGTM assuming we're deleting this code in a year.

Kimchelly (approved):
LGTM - I feel comfortable merging with just staging tests given that this logic is only temporary until the cutover in a year. Just one request to please make a ticket for the cutover and deletion of this logic so we don't forget!

opsnlops (approved):

apimodels/cedar_models.go
julianedwards: Created https://jira.mongodb.org/browse/EVG-21010 and added a TODO.

PR #7083, opened 2023-09-27
EVG-20924 Get project file includes asynchonously

bsamek:
EVG-20924

### Description

Getting files serially from GitHub was causing projects with many file includes
to exceed a timeout.

### Testing

This is a refactor, so it should be tested by existing tests.

### Documentation

N/A

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -650,70 +650,71 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 		return nil, errors.Wrapf(err, LoadProjectError)
 	}
 
-	if opts == nil {
-		err = errors.New("trying to open include files with empty options")
-		return nil, errors.Wrapf(err, LoadProjectError)
-	}
-
-	wg := sync.WaitGroup{}
-	catcher := grip.NewBasicCatcher()
-	yamlNamePairs := make(chan yamlNamePair, len(intermediateProject.Include))
-
-	for _, path := range intermediateProject.Include {
-		wg.Add(1)
-		go func(include Include, projectOpts *GetProjectOpts) {
-			defer wg.Done()
-			// Make a copy of opts because otherwise parts opts would be
-			// modified concurrenly.  Note, however, that Ref and PatchOpts are
-			// themselves pointers, so should not be modified.
-			localOpts := &GetProjectOpts{
-				Ref:             opts.Ref,
-				PatchOpts:       opts.PatchOpts,
-				LocalModules:    opts.LocalModules,
-				RemotePath:      projectOpts.RemotePath,
-				Revision:        projectOpts.Revision,
-				Token:           projectOpts.Token,
-				ReadFileFrom:    projectOpts.ReadFileFrom,
-				Identifier:      projectOpts.Identifier,
-				UnmarshalStrict: projectOpts.UnmarshalStrict,
-			}
-			localOpts.UpdateForFile(include.FileName)
-
-			var yaml []byte
-			localOpts.Identifier = identifier
-			localOpts.RemotePath = include.FileName
-			grip.Debug(message.Fields{
-				"message":     "retrieving included YAML file",
-				"remote_path": localOpts.RemotePath,
-				"read_from":   localOpts.ReadFileFrom,
-				"module":      include.Module,
-			})
-			if include.Module != "" {
-				yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
-				if err == nil && yaml != nil {
-					yamlNamePairs <- yamlNamePair{yaml, include.FileName}
-				}
-			} else {
-				yaml, err = retrieveFile(ctx, *localOpts)
-				if err == nil && yaml != nil {
-					yamlNamePairs <- yamlNamePair{yaml, include.FileName}
-				}
-			}
-			catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
-		}(path, opts)
-	}
-	wg.Wait()
-	close(yamlNamePairs)
-
-	// return intermediateProject even if we run into issues to show merge progress
-	for pair := range yamlNamePairs {
-		add, err := createIntermediateProject(pair.yaml, opts.UnmarshalStrict)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, pair.name)
+	if len(intermediateProject.Include) > 0 {
+		if opts == nil {
+			err = errors.New("trying to open include files with empty options")
+			return nil, errors.Wrapf(err, LoadProjectError)
 		}
-		err = intermediateProject.mergeMultipleParserProjects(add)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, pair.name)
+		wg := sync.WaitGroup{}
+		catcher := grip.NewBasicCatcher()
+		yamlNamePairs := make(chan yamlNamePair, len(intermediateProject.Include))
+
+		for _, path := range intermediateProject.Include {
+			wg.Add(1)
+			go func(include Include, projectOpts *GetProjectOpts) {
+				defer wg.Done()
+				// Make a copy of opts because otherwise parts opts would be
+				// modified concurrenly.  Note, however, that Ref and PatchOpts are
+				// themselves pointers, so should not be modified.
+				localOpts := &GetProjectOpts{
+					Ref:             opts.Ref,
+					PatchOpts:       opts.PatchOpts,
+					LocalModules:    opts.LocalModules,
+					RemotePath:      projectOpts.RemotePath,
+					Revision:        projectOpts.Revision,
+					Token:           projectOpts.Token,
+					ReadFileFrom:    projectOpts.ReadFileFrom,
+					Identifier:      projectOpts.Identifier,
+					UnmarshalStrict: projectOpts.UnmarshalStrict,
+				}
+				localOpts.UpdateForFile(include.FileName)
+
+				var yaml []byte
+				localOpts.Identifier = identifier
+				localOpts.RemotePath = include.FileName
+				grip.Debug(message.Fields{
+					"message":     "retrieving included YAML file",
+					"remote_path": localOpts.RemotePath,
+					"read_from":   localOpts.ReadFileFrom,
+					"module":      include.Module,
+				})
+				if include.Module != "" {
+					yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				} else {
+					yaml, err = retrieveFile(ctx, *localOpts)
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				}
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
+			}(path, opts)
+		}
+		wg.Wait()
+		close(yamlNamePairs)
+
+		// return intermediateProject even if we run into issues to show merge progress
+		for pair := range yamlNamePairs {
+			add, err := createIntermediateProject(pair.yaml, opts.UnmarshalStrict)
+			if err != nil {
+				return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, pair.name)
+			}
+			err = intermediateProject.mergeMultipleParserProjects(add)
+			if err != nil {
+				return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, pair.name)
+			}
 		}
 	}
 
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -703,6 +703,7 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 		for _, path := range intermediateProject.Include {
 			includesToProcess <- path
 		}
+		close(includesToProcess)
 
 		// Be polite. Don't make more than 10 concurrent requests to GitHub.
 		const maxWorkers = 10
@@ -718,10 +719,8 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 		}
 
 		// This order is deliberate:
-		// 1. Close `includesToProcess` so that the workers stop `range`ing over it.
-		// 2. Wait for the workers, since sending on a `nil` `outputYAMLs` would panic.
-		// 3. Close `outputYAMLs` so that the later `range` statement over it will stop after it's is drained.
-		close(includesToProcess)
+		// 1. Wait for the workers, since sending on a `nil` `outputYAMLs` would panic.
+		// 2. Close `outputYAMLs` so that the later `range` statement over it will stop after it's is drained.
 		wg.Wait()
 		close(outputYAMLs)
 
@@ -740,6 +739,9 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 
 		// We promise to iterate over includes in the order they are defined.
 		for _, path := range intermediateProject.Include {
+			if _, ok := yamlMap[path.FileName]; !ok {
+				return intermediateProject, errors.WithStack(errors.Errorf("yaml was nil in map for %s, but it never should be", path.FileName))
+			}
 			add, err := createIntermediateProject(yamlMap[path.FileName], opts.UnmarshalStrict)
 			if err != nil {
 				// Return intermediateProject even if we run into issues to show merge progress.
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -631,9 +631,48 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 	return TranslateProject(pp)
 }
 
-type yamlNamePair struct {
+func processIntermediateProjectInludes(ctx context.Context, identifier string, intermediateProject *ParserProject,
+	include Include, yamlChan chan<- yamlTuple, projectOpts *GetProjectOpts) {
+	// Make a copy of opts because otherwise parts of opts would be
+	// modified concurrently.  Note, however, that Ref and PatchOpts are
+	// themselves pointers, so should not be modified.
+	localOpts := &GetProjectOpts{
+		Ref:             projectOpts.Ref,
+		PatchOpts:       projectOpts.PatchOpts,
+		LocalModules:    projectOpts.LocalModules,
+		RemotePath:      include.FileName,
+		Revision:        projectOpts.Revision,
+		Token:           projectOpts.Token,
+		ReadFileFrom:    projectOpts.ReadFileFrom,
+		Identifier:      identifier,
+		UnmarshalStrict: projectOpts.UnmarshalStrict,
+	}
+	localOpts.UpdateReadFileFrom(include.FileName)
+
+	var yaml []byte
+	var err error
+	grip.Debug(message.Fields{
+		"message":     "retrieving included YAML file",
+		"remote_path": localOpts.RemotePath,
+		"read_from":   localOpts.ReadFileFrom,
+		"module":      include.Module,
+	})
+	if include.Module != "" {
+		yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
+	} else {
+		yaml, err = retrieveFile(ctx, *localOpts)
+	}
+	if err != nil {
+		yamlChan <- yamlTuple{nil, include.FileName, errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName)}
+	} else {
+		yamlChan <- yamlTuple{yaml, include.FileName, nil}
+	}
+}
+
+type yamlTuple struct {
 	yaml []byte
 	name string
+	err  error
 }
 
 // LoadProjectInto loads the raw data from the config file into project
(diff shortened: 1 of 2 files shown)


model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					}
+				}
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
+			}(path, opts)
 		}
-		err = intermediateProject.mergeMultipleParserProjects(add)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+		wg.Wait()
+		close(yamlNamePairs)
```
Kimchelly, line 706: It's not necessary to close this channel since it already waits for the wait group to finish. The garbage collector will clean up `yamlNamePairs` because there's no remaining usages of it.

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					"module":      include.Module,
+				})
+				if include.Module != "" {
+					yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				} else {
+					yaml, err = retrieveFile(ctx, *localOpts)
+					if err == nil && yaml != nil {
```
Kimchelly, line 698: If `yaml` is non-nil, do we care about `err` being nil? Same thing with the other one.

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				}
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
```
Kimchelly, line 702: Rather than doing this in an if-else, where there are further if blocks within the if else, can we just make this more streamlined so there's one case and another case? The only shared line is the lingering `catcher` adding at the end, so it doesn't seem worth it to complicate the logic:
```go
if include.Module != "" {
    yaml, err := ...
    if yaml != nil {
        yamlNamePairs <- ...
    }
    catcher.Wrapf(...)
    return
}

yaml, err := retrieveFile(...)
if yaml != nil {
    yamlNamePairs <- ...
}
catcher.Wrapf(...)
```

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					Token:           projectOpts.Token,
+					ReadFileFrom:    projectOpts.ReadFileFrom,
+					Identifier:      projectOpts.Identifier,
+					UnmarshalStrict: projectOpts.UnmarshalStrict,
+				}
+				localOpts.UpdateForFile(include.FileName)
+
+				var yaml []byte
+				localOpts.Identifier = identifier
+				localOpts.RemotePath = include.FileName
```
Kimchelly, line 684: nit: can L683-684 just be absorbed into the initialization on L669?

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
+			}(path, opts)
 		}
-		err = intermediateProject.mergeMultipleParserProjects(add)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+		wg.Wait()
+		close(yamlNamePairs)
+
+		// return intermediateProject even if we run into issues to show merge progress
```
Kimchelly: nit: to follow the style guide: `// Return intermediateProject even if Evergreen runs into issues to show merge progress.`

Also, I would put this comment a little further in so it's clear that it's taking about the returns inside this for loop.

ablack12:
I'm not sure this is fully covered by existing tests, since the case we're concerned about involves fetching files from GitHub. Ideally we could test this on staging with a few tiny yamls and verify we're able to fetch them and continue on.

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+		wg.Wait()
+		close(yamlNamePairs)
+
+		if catcher.HasErrors() {
+			return intermediateProject, catcher.Resolve()
+		} else {
+			// return intermediateProject even if we run into issues to show merge progress
+			for pair := range yamlNamePairs {
```
ablack12, line 712: Verification questions because I don't know much about this:
Will the include files be added to yamlNamePairs in order? We're supposed to merge these in order, so if not, we may want to add these to a map instead and then iterate through the include list

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		err = intermediateProject.mergeMultipleParserProjects(add)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+		wg.Wait()
+		close(yamlNamePairs)
+
+		if catcher.HasErrors() {
+			return intermediateProject, catcher.Resolve()
+		} else {
+			// return intermediateProject even if we run into issues to show merge progress
```
ablack12, line 711: tiny nit: preference for capitalizing comment

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+		wg := sync.WaitGroup{}
+		catcher := grip.NewBasicCatcher()
+		yamlNamePairs := make(chan yamlNamePair, len(intermediateProject.Include))
+
+		for _, path := range intermediateProject.Include {
+			wg.Add(1)
+			go func(include Include, projectOpts *GetProjectOpts) {
+				defer wg.Done()
+				// Make a copy of opts because otherwise parts opts would be
```
ablack12, line 666: parts *of opts?

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+				// modified concurrenly.  Note, however, that Ref and PatchOpts are
```
ablack12, line 667: concurrently :)

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+		wg := sync.WaitGroup{}
+		catcher := grip.NewBasicCatcher()
+		yamlNamePairs := make(chan yamlNamePair, len(intermediateProject.Include))
+
+		for _, path := range intermediateProject.Include {
+			wg.Add(1)
+			go func(include Include, projectOpts *GetProjectOpts) {
```
Kimchelly, line 664: Suggestion to move the goroutine logic into a helper function to make it a little easier to read.

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					PatchOpts:       opts.PatchOpts,
+					LocalModules:    opts.LocalModules,
+					RemotePath:      projectOpts.RemotePath,
+					Revision:        projectOpts.Revision,
+					Token:           projectOpts.Token,
+					ReadFileFrom:    projectOpts.ReadFileFrom,
+					Identifier:      projectOpts.Identifier,
+					UnmarshalStrict: projectOpts.UnmarshalStrict,
+				}
+				localOpts.UpdateForFile(include.FileName)
```
ablack12, line 680: Maybe we should rename this UpdateReadFileFrom to make it less likely that someone will change this in the future to update ref or patchOpts

model/project_parser.go
Kimchelly, line 664: It seems like `opts` and `projectOpts` are one and the same, and never get modified. Do we need to pass it in as a separate parameter?

model/project_parser.go
```diff
@@ -644,40 +650,74 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+		wg.Wait()
+		close(yamlNamePairs)
+
+		// return intermediateProject even if we run into issues to show merge progress
+		for pair := range yamlNamePairs {
+			add, err := createIntermediateProject(pair.yaml, opts.UnmarshalStrict)
+			if err != nil {
+				return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, pair.name)
+			}
+			err = intermediateProject.mergeMultipleParserProjects(add)
```
Kimchelly: nit: combine L714-715.

Kimchelly (changes requested):
Did you try out the include change in staging?

model/project_parser.go
```diff
@@ -705,17 +705,22 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		// return intermediateProject even if we run into issues to show merge progress
-		for pair := range yamlNamePairs {
-			add, err := createIntermediateProject(pair.yaml, opts.UnmarshalStrict)
-			if err != nil {
-				return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, pair.name)
-			}
-			err = intermediateProject.mergeMultipleParserProjects(add)
-			if err != nil {
-				return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, pair.name)
+		if catcher.HasErrors() {
```
Kimchelly, line 708: It's more idiomatic in Go to omit the else when the if returns.

Kimchelly (changes requested):

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+				grip.Debug(message.Fields{
+					"message":     "retrieving included YAML file",
+					"remote_path": localOpts.RemotePath,
+					"read_from":   localOpts.ReadFileFrom,
+					"module":      include.Module,
+				})
+				if include.Module != "" {
+					yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
```
ablack12, line 694: I feel like it's generally frowned upon to rely on ordering to create structs as opposed to using the variable names, but it's not in our style guide so that could just be me

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					}
+				}
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
+			}(path, opts)
 		}
-		err = intermediateProject.mergeMultipleParserProjects(add)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+		wg.Wait()
+		close(yamlNamePairs)
```
ablack12, line 706: would we want to just defer this at the time of opening it to avoid potential bugs later?

ablack12 (changes requested):

model/project_parser.go
ablack12: Personally i think this is still good practice

bsamek:
@Kimchelly 

> Did you try out the include change in staging?

@ablack12 

> I'm not sure this is fully covered by existing tests, since the case we're concerned about involves fetching files from GitHub. Ideally we could test this on staging with a few tiny yamls and verify we're able to fetch them and continue on.

I haven't tried it in staging. I will before merging.

model/project_parser.go
bsamek: done

bsamek: no longer applicable as a side effect of moving into a helper function

model/project_parser.go
bsamek: fixed

model/project_parser.go
bsamek: fixed

model/project_parser.go
bsamek: done

model/project_parser.go
bsamek: done

model/project_parser.go
bsamek: I think this is fine for brevity in a case like this. You also can't order them incorrectly because the compiler will yell at you, because they're disparate types.

model/project_parser.go
bsamek: this became cleaner with the helper function

Kimchelly:
> I haven't tried it in staging. I will before merging.

Rather than waiting until the last second to test in staging, can this be tested in staging before the next round of review? It's better practice to confirm that a change to something as critical as included files is at least basic tested for working-ness before we can comfortably LGTM it. In general, it's more standard practice in Evergreen that the PR creator has shown their change works in at least one simple case (e.g. via a test in staging) before getting reviews.

bsamek:
> > I haven't tried it in staging. I will before merging.
> 
> Rather than waiting until the last second to test in staging, can this be tested in staging before the next round of review? It's better practice to confirm that a change to something as critical as included files is at least basic tested for working-ness before we can comfortably LGTM it. In general, it's more standard practice in Evergreen that the PR creator has shown their change works in at least one simple case (e.g. via a test in staging) before getting reviews.

Will do!

model/project_parser.go
bsamek: I think it's necessary to close the channel because otherwise the `range` loop over it would deadlock.

model/project_parser.go
bsamek: This went away with the helper function.

model/project_parser.go
bsamek: done

model/project_parser.go
bsamek: done

model/project_parser.go
bsamek: fixed

model/project_parser.go
bsamek: I didn't realize this! Thanks. I switched it to use a map and iterate through the include list again.

model/project_parser.go
bsamek: fixed

model/project_parser.go
```diff
@@ -644,40 +650,79 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				} else {
+					yaml, err = retrieveFile(ctx, *localOpts)
+					if err == nil && yaml != nil {
+						yamlNamePairs <- yamlNamePair{yaml, include.FileName}
+					}
+				}
+				catcher.Add(errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName))
```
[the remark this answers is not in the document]
bsamek: no longer relevant with helper function

model/project_parser.go
Kimchelly: I'm actually gonna +1 what Annie said, using named outputs is generally preferred over ordered parameters, unless there's some specific reason that I've never heard of. The [Go style suggests using named parameters for exported types](https://google.github.io/styleguide/go/decisions#field-names), but realistically in our case, there's not much of a distinction between usage of exported vs. unexported except for sharing between packages.

bsamek: fixed

bsamek:
@ablack12 @Kimchelly This is ready for a re-review. Note that I made one new change to the logic, which is that I limit the number of concurrent workers to 10.

Staging build showing includes working https://spruce-staging.corp.mongodb.com/version/651587729dbe327796685b74/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC. Splunk query showing them happening in the same second https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%20%22retrieving%20included%20YAML%20file%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-30m%40m&latest=now&sid=1695909895.5074433.

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
 		}
-		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+
+		close(includeChan)
+		wg.Wait()
+		close(yamlChan)
+
+		// We must apply the includes in order, so we indirect through a map.
```
Kimchelly, line 728: I'm not sure that this comment actually helps with understanding. The reason the includes can be applied in order is because of the combination of `wg.Wait` waiting for all includes to be loaded into `yamlChan`, then iterating in order on L744 through `intermediateProjet.Include` again.

I would recommend either remove or rephrase + move down.

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+			if err != nil {
+				// Return intermediateProject even if we run into issues to show merge progress.
+				return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+			}
+			if err = intermediateProject.mergeMultipleParserProjects(add); err != nil {
+				// Return intermediateProject even if we run into issues to show merge progress.
+				return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+			}
+		}
+
```
Kimchelly, line 755: mega nit: remove random whitespace

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		if path.Module != "" {
-			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module)
-		} else {
-			yaml, err = retrieveFile(ctx, *opts)
+		wg := sync.WaitGroup{}
+		yamlChan := make(chan yamlTuple, len(intermediateProject.Include))
+		includeChan := make(chan Include)
+
+		// Be polite. Don't make more than 10 concurrent requests to GitHub.
+		for i := 0; i < 10; i++ {
```
Kimchelly, line 710: nit: since this is a magic and arbitrarily-chosen number, can this be a `const`?

Kimchelly (changes requested):

model/project_parser.go
```diff
@@ -630,6 +631,58 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
 	return TranslateProject(pp)
 }
 
+func processIntermediateProjectInludes(ctx context.Context, identifier string, intermediateProject *ParserProject,
```
ablack12, line 634: Includes :)

model/project_parser.go
```diff
@@ -630,6 +631,58 @@ func GetProjectFromBSON(data []byte) (*Project, error) {
...
+	if include.Module != "" {
+		yaml, err = retrieveFileForModule(ctx, *localOpts, intermediateProject.Modules, include.Module)
+	} else {
+		yaml, err = retrieveFile(ctx, *localOpts)
+	}
+	if err != nil {
+		yamlChan <- yamlTuple{
+			yaml: nil,
+			name: include.FileName,
+			err:  errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, include.FileName),
```
ablack12, line 669: I'd refer if this error was a bit different for 660 / 662 so that we could better trace the path, if there is an error. Maybe we can set the wrapped error within the if/else so we can add a "with modules"

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		})
-		if path.Module != "" {
-			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module)
-		} else {
-			yaml, err = retrieveFile(ctx, *opts)
+		wg := sync.WaitGroup{}
+		yamlChan := make(chan yamlTuple, len(intermediateProject.Include))
+		includeChan := make(chan Include)
+
+		// Be polite. Don't make more than 10 concurrent requests to GitHub.
```
ablack12, line 709: this is a great comment

model/project_parser.go
ablack12: love this self awareness

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-			return intermediateProject, errors.Wrapf(err, "%s: retrieving file '%s'", LoadProjectError, path.FileName)
+
+		for _, path := range intermediateProject.Include {
+			includeChan <- path
 		}
-		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+
+		close(includeChan)
```
ablack12, line 724: From discussion: can we include a comment here (or somewhere) about the order of things here, so engineers don't come in here and try to reorder things bc it doesn't seem significant

ablack12 (approved):

model/project_parser.go
```diff
@@ -644,40 +697,64 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-			"read_from":   opts.ReadFileFrom,
-			"module":      path.Module,
-		})
-		if path.Module != "" {
-			yaml, err = retrieveFileForModule(ctx, *opts, intermediateProject.Modules, path.Module)
-		} else {
-			yaml, err = retrieveFile(ctx, *opts)
+		wg := sync.WaitGroup{}
+		yamlChan := make(chan yamlTuple, len(intermediateProject.Include))
+		includeChan := make(chan Include)
```
Kimchelly, line 707: Small suggestion to also buffer this channel with `len(intermediateProject.Include)`.

model/project_parser.go
bsamek: Fixed, though I wasn't sure if you'd have an opinion about where the const is defined. I defined it inline.

model/project_parser.go
```diff
@@ -644,40 +691,67 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
 		}
-		add, err := createIntermediateProject(yaml, opts.UnmarshalStrict)
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: loading file '%s'", LoadProjectError, path.FileName)
+
+		// This order is deliberate:
+		// 1. Close `includesToProcess` so that the workers stop `range`ing over it.
+		// 2. Wait for the workers, since sending on a `nil` `outputYAMLs` would panic.
+		// 3. Close `outputYAMLs` so that the later `range` statement over it will stop after it's is drained.
+		close(includesToProcess)
```
Kimchelly, line 724: nit: Slight preference for closing the channel after putting in the values on L706 to make it clearer that putting in the includes is the end of the input. Fixed-size input channels are a common-ish pattern that I've seen in other places (like Cedar), and generally the input channel is closed ASAP once all the input is filled.

model/project_parser.go
Kimchelly: Unfortunately, it's not smart enough to make style decisions when it comes to single newlines between statements/closing brackets, so I just comment on them in the PR if they look like stray whitespace. 🤷

model/project_parser.go
Kimchelly: Defining it inline is fine, just naming a var representing our arbitrary limit is good enough.

model/project_parser.go
```diff
@@ -644,40 +691,67 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-		if err != nil {
-			return intermediateProject, errors.Wrapf(err, "%s: merging file '%s'", LoadProjectError, path.FileName)
+
+		if catcher.HasErrors() {
+			return intermediateProject, errors.Wrap(catcher.Resolve(), "getting includes")
+		}
+
+		// We promise to iterate over includes in the order they are defined.
+		for _, path := range intermediateProject.Include {
+			add, err := createIntermediateProject(yamlMap[path.FileName], opts.UnmarshalStrict)
```
Kimchelly, line 743: For a bit of extra precaution against future changes, comma-ok checking the `yamlMap[path.FileName]` and returning a programmatic error might be preferable to avoid potential nil YAMLs causing panics during unmarshalling.

Kimchelly (approved):
LGTM mod non-blocking suggestions.

model/project_parser.go
bsamek: done, yeah this is was mostly because it hadn't originally been a buffered channel

model/project_parser.go
bsamek: done

PR #7084, opened 2023-09-27
EVG-20940 Update module cloning to use GitHub app

bynn:
EVG-20940

### Description
modules were previously defaulting to use the oauth token before so we want to fix it to use the correct method
adds an agent route to create new tokens 
caches tokens per owner so we don't generate new tokens for every module unnecessarily 

### Testing
forced token generation on a container task for a private repo module on staging 

https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu2004_container_bynntask_patch_d631380317a8478b879cc27d6171508dea2b11df_651afa2197b1d3728e1718be_23_10_02_17_13_13/logs?execution=2&logtype=all


Kimchelly (changes requested):
I'm wondering if this could potentially break module clones for some projects, which is especially problematic given that module clone failures currently don't cause git.get_project to fail (EVG-20317). Do we already know that all modules used in Evergreen can be cloned by the GitHub app token? And if not, does this fall back to using some other cloning strategy if the app can't clone?

agent/command/git.go
```diff
@@ -190,8 +190,18 @@ func (opts cloneOpts) getCloneCommand() ([]string, error) {
 	}
 	switch opts.method {
 	case "", evergreen.CloneMethodLegacySSH:
+		grip.Debug(message.Fields{
```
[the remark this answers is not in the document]
bynn: Oh yeah, I have a different ticket for logging that these fields are deprecated, deleting 
Is there an easier way to keep track of us ever calling this?

agent/command/git.go
```diff
@@ -707,13 +712,24 @@ func (c *gitFetchProject) fetchModuleSource(ctx context.Context,
...
 		branch:   "",
 		dir:      moduleBase,
+		method:   cloneMethod,
 	}
 	// Module's location takes precedence over the project-level clone
 	// method.
 	if strings.Contains(opts.location, "git@github.com:") {
 		opts.method = evergreen.CloneMethodLegacySSH
+	} else if opts.method == evergreen.CloneMethodAccessToken {
+		if appToken, ok := appTokens[opts.owner]; ok {
```
Kimchelly, line 722: I'm kind of curious how the `else` case ever gets hit, because it seems impossible - [`appTokens` is set in the same case where `opts.method == evergreen.CloneMethodAccessToken`](https://github.com/evergreen-ci/evergreen/blob/2062dece757ef093f0c7948e7b27853368000273/agent/command/git.go#L847-L850), in which case this if block always evaluates to true. Doesn't this devolve into:

```go
else if opts.method == evergreen.CloneMethodAccessToken {
    opts.token = appToken // If we've reached this else-if, this is always going to be set and will never dynamically fetch the installation token?
}
```
}
Am I missing something?

If you're trying to make an in-memory cache of app tokens for modules to save calls to `CreateInstallationToken` for later module clones, you have to move the `appTokens` initialization outside of the module loop.

agent/command/git.go
```diff
@@ -707,13 +712,24 @@ func (c *gitFetchProject) fetchModuleSource(ctx context.Context,
...
+			opts.token = appToken
+		} else {
+			appToken, err := comm.CreateInstallationToken(ctx, td, opts.owner, opts.repo)
+			if err != nil {
+				return errors.Wrapf(err, "creating installation token for '%s/%s'", opts.owner, opts.repo)
+			}
+			appTokens[opts.owner] = appToken
```
Kimchelly, line 729: Seems like we don't need to set this token because it's never used, unless it's for the module cache thing I mentioned in the above comment.

agent/command/git.go
```diff
@@ -828,7 +844,11 @@ func (c *gitFetchProject) fetch(ctx context.Context,
 		if err := ctx.Err(); err != nil {
 			return errors.Wrapf(err, "canceled while applying module '%s'", moduleName)
 		}
-		err = c.fetchModuleSource(ctx, conf, logger, jpm, opts.token, p, moduleName)
+		appTokens := map[string]string{}
```
Kimchelly, line 847: Assuming we're using `appTokens` as a cache, can we add a clarifying comment that this is acting as a cache mapping repo owners to GitHub app tokens to reduce the number of app tokens generated?

I guess as a follow-up question, is there enough benefit to caching them vs. the potential issue that the token could expire on later clones? I'm wondering if there are edge cases where git clones are particularly slow like [this one](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/result/f4mjgFopujE/trace/gjMb6XsDpYB?fields[]=c_name&fields[]=c_service.name&span=10a8f70440cdb60d), which took 2 hours to clone one of the modules (but it did succeed eventually). Are GitHub app tokens valid for long enough to cover slow (but working) git clones, or will it expire by the time it's reused for the next module?

agent/internal/client/interface.go
```diff
@@ -120,6 +120,8 @@ type SharedCommunicator interface {
 	GetAdditionalPatches(ctx context.Context, patchId string, td TaskData) ([]string, error)
 
 	SetDownstreamParams(ctx context.Context, downstreamParams []patchmodel.Parameter, taskData TaskData) error
+
+	CreateInstallationToken(ctx context.Context, td TaskData, owner, repo string) (string, error)
```
Kimchelly, line 124: Doc comment for exported interface method.

apimodels/agent_models.go
```diff
@@ -183,6 +183,10 @@ type RegistrySettings struct {
 	Password string `mapstructure:"registry_password" json:"registry_password" yaml:"registry_password"`
 }
 
+type InstallationToken struct {
+	Token string `bson:"token" json:"token"`
```
Kimchelly, line 187: Is the BSON tag needed? Seems like it shouldn't need the BSON tags, unless it's stored in the DB.

rest/route/github_agent_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateInstallationToken(t *testing.T) {
+	validOwner := "owner"
+	validRepo := "repo"
+
+	for tName, tCase := range map[string]func(ctx context.Context, t *testing.T, gh *createInstallationToken, env *mock.Environment){
+		"parse should error on empty owner and repo": func(ctx context.Context, t *testing.T, handler *createInstallationToken, env *mock.Environment) {
```
Kimchelly, line 22: nit: I haven't commented on this a lot in other PRs, but can we stick to one convention for test names (seems like most of them use `PascalCase`, as does the top-level function. The problem with using spaces is that `go test` auto-translates spaces to underscores, so it both looks inconsistent with the other tests (it looks like `TestCreateInstallationToken/parse_should_error_on_empty_owner_and_repo` whereas most of the tests `TestFunctionName/TestsLookLikeThis`) and you have to know that if you're grepping for this particular test name, you have to translate all the underscores back to spaces.

I can add a mention to the style guide about this.

Kimchelly (changes requested):
* Re the original issue where it didn't handle clones in different orgs, did you test cloning a private module for a different org in staging?
* Re the original concern that the app may not have permission to clone every private module (and it's going to after this change), do we have a plan to monitor git clones after this is deployed?

agent/command/git.go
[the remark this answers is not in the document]
Kimchelly: You could probably add attributes to [the Honeycomb tracer](https://github.com/evergreen-ci/evergreen/blob/2062dece757ef093f0c7948e7b27853368000273/agent/command/git.go#L604-L608) to get this info.

bynn:
now, if the github app does not exist on a repo (which is unlikely) we will fallback to the legacy global token so it should fix that issue 

added clone method to honeycomb but it can't really be relied on because the oauth method will be specified for both the fallback and user specified tokens

agent/command/git.go
Kimchelly: While it only happens a few times, it seems like mandating fast git clones isn't something we should do. The GitHub app token usage is supposed to be something that's transparent to the end user and should "just work", so absent some explicit configuration they've set (like setting exec timeout to 10 mins or something), they shouldn't be worried about their clones being too slow and then having to make their own workarounds because Evergreen is flaky about cloning.

The 1 hour TTL thing may also be troublesome for the app token in the expansion as well, because the time between when the GitHub app generates the token for `PopulateExpansions` and when the task finally clones could be any amount of time, meaning we could have already used up some of the TTL time and the original app token has already expired. If the GitHub app token can only be used for git.get_project, I would make a follow-up ticket to remove the app token from the expansions entirely (which also gets rid of the security risk of leaking them) and just fetch it dynamically like with the module token.

apimodels/agent_models.go
Kimchelly: Seems like the tag is still here.

rest/route/github_agent_test.go
```diff
@@ -0,0 +1,76 @@
...
+	} {
+		t.Run(tName, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			env := &mock.Environment{}
+			require.NoError(t, env.Configure(ctx))
+
+			r, ok := makeCreateInstallationToken(env).(*createInstallationToken)
+			r.env = env
```
Kimchelly, line 70: It's not necessary to set this since L69 already set it.

rest/route/service.go
```diff
@@ -86,6 +86,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/task/{task_id}/").Version(2).Get().Wrap(requireTask).RouteHandler(makeFetchTask())
 	app.AddRoute("/task/{task_id}/log").Version(2).Post().Wrap(requireTask, requirePodOrHost).RouteHandler(makeAppendTaskLog(settings))
 	app.AddRoute("/task/{task_id}/start").Version(2).Post().Wrap(requireTask, requirePodOrHost).RouteHandler(makeStartTask(env))
+	app.AddRoute("/github/installation_token/{owner}/{repo}").Version(2).Get().Wrap(requireTask).RouteHandler(makeCreateInstallationToken(env))
```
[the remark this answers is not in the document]
Kimchelly: The `requireTask` middleware does read the `task_id` parameter to auth, but I don't think we need to explicitly use it in the route handler itself. This just makes it easier to trace the API calls that a particular task makes while it runs if the task ID is included. If we keep it as-is and we ever have a problem with this particular route, we won't know which task is calling it and why to trace the issue.

Kimchelly:
> added clone method to honeycomb but it can't really be relied on because the oauth method will be specified for both the fallback and user specified tokens

This is a temporary field for your own purposes while we deprecate the legacy token and SSH cloning, and not something we intend to keep around forever, gith? If so, one workaround is to add some temporary conditional logic, like if the legacy token/SSH cloning is used, set an attribute on the tracer, which you can look for later.

bynn (commented):
created https://jira.mongodb.org/browse/EVG-21022 to dynamically create tokens 

also made the change for every module to dynamically create a token every time to allow for edge cases. I looked at all the modules and there weren't so many that it would drastically affect the times + they all use ssh anyway.

I went through all the modules by looking at all the configs and it looks like 99% of them use the ssh method (the only exception being ryans project that couldn't because it was a container) so this shouldn't affect any existing users 

I decided to add the clone method to honeycomb tracing permanently because I think it would be useful to have it all the time for the "remove ssh cloning" ticket, but I will comment on the ticket to make it conditional along with that change

Kimchelly (approved):
That all sounds reasonable to me.

LGTM mod one small thing.

rest/route/github_agent.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"net/http"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/apimodels"
+	"github.com/evergreen-ci/gimlet"
+	"github.com/pkg/errors"
+)
+
+// GET /rest/v2/task/{task_id}/installation_token/{owner}/{repo}
```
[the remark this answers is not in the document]
bynn: updated

PR #4689, opened 2021-05-19
EVG-14603: randomly use an alt github token

hhoke:

PR #7085, opened 2023-09-27
EVG-20184 Remove round-robin code

malikchaya2:
EVG-20184

### Description
This reverts [4689](https://github.com/evergreen-ci/evergreen/pull/4689) because we will not be using it.


Kimchelly (approved):
LGTM mod fix the trigger error message test.

PR #7086, opened 2023-09-27
EVG-20945 Add logging for deprecating ssh clone method

bynn:
EVG-20945

### Description
Adding a log to tell users to stop using ssh clone methods as they will not work anymore

### Documentation
updated docs

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -184,12 +184,13 @@ func parseToken(token string) (string, error) {
 	return splitToken[1], nil
 }
 
-func (opts cloneOpts) getCloneCommand() ([]string, error) {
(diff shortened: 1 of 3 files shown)


Kimchelly (commented):
I don't know what is the plan around deprecating SSH cloning, but:

* I imagine nobody will see the warning log, since it's in the execution logs (which people only read when looking for specific things), and nobody will really know to look for a random log line in their task/execution logs anyways.
* Are we planning to remove SSH cloning one day (fully and completely), or are we just trying to discourage new usages of it? If we plan to remove, do we have a specific timeframe + comms plan for removing SSH cloning?
* If I remember correctly, some distros couldn't use OAuth cloning because their version of git was too old. Is that still the case, or have those distros all been deleted at this point?

PR #7090, opened 2023-09-28
EVG-20579: support multiple criteria for alias tags

Kimchelly:
EVG-20579

### Description
The docs describe how you can use [tag selector syntax](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files/#task-and-variant-tags) in project YAMLs to express various behaviors like set union, intersect, and negation. For example:

```yaml
buildvariants:
- name: some-bv
  tasks:
    # Select all tasks that are tagged as both "animal" and "brown", but not "mammal".
    - .animal .brown !.mammal
    # Select all tasks that are tagged as "mammal" but not "large".
    - .mammal !.large
    # The above two selectors are unioned together to get the list of tasks to run in the BV.
```

Aliases currently support a limited version of this. For one tag selector, it currently only allows a single tag in the selector and supporting negation (e.g. `animal` will select just those items tagged with "animal", but you can't do something like `animal brown` to get all items tagged with both "animal" and "brown"). I extended it to support the full syntax described in the project YAML docs, so you can list multiple space-separated criteria for the tag selector.

### Testing
* Checked all the project aliases in prod to verify that none of them are relying on having spaces in their alias tags for something else.
* Added unit tests.
* Tested in staging that the alias multi-criteria selection worked.

### Documentation
Updated docs for project aliases with examples.

### DIFF
--- a/model/project_aliases.go
+++ b/model/project_aliases.go
@@ -66,13 +66,15 @@ const (
 // variants/tasks, assuming the tag matches the defined git_tag regex.
 // In this way, users can define different behavior for different kind of tags.
 type ProjectAlias struct {
-	ID          mgobson.ObjectId  `bson:"_id,omitempty" json:"_id" yaml:"id"`
-	ProjectID   string            `bson:"project_id" json:"project_id" yaml:"project_id"`
-	Alias       string            `bson:"alias" json:"alias" yaml:"alias"`
-	Variant     string            `bson:"variant,omitempty" json:"variant" yaml:"variant"`
-	Description string            `bson:"description" json:"description" yaml:"description"`
-	GitTag      string            `bson:"git_tag" json:"git_tag" yaml:"git_tag"`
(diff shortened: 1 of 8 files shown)


model/project_aliases.go
```diff
@@ -585,13 +595,45 @@ func isValidRegexOrTag(curItem string, curTags, aliasTags []string, aliasRegex *
...
+				// Intersection - tag does not match the item's tags.
+				allCriteriaSatisfied = false
+				break
+			}
 		}
-		// a negated tag
-		if len(tag) > 0 && tag[0] == '!' && !utility.StringSliceContains(curTags, tag[1:]) {
-			return true
+
+		if !allCriteriaSatisfied {
```
ablack12, line 632: readability nit: could we just say if allCriteriaSatisifed return true since we'll be continuing anyway?

ablack12 (approved):

PR #7095, opened 2023-09-28
EVG-19841 Error on adding github intent for repos with no config path

hadjri:
EVG-19841

### Description
Stemming from the finding in the comment of this ticket, return an error if a repo has PR testing enabled but has an empty remote config path when creating PR patches for untracked branches.

### Testing
In staging, confirmed that an[ error surfaces](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22evergreen-staging%22%20%7C%20spath%20%22metadata.level%22%20%7C%20search%20%22metadata.level%22%3D70%7C%20spath%20repo%20%7C%20search%20repo%3D%22commit-queue-sandbox%22&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1696002688&latest=1696003588&sid=1696004850.5318722) when attempting to create a PR patch for an untracked branch when the remote config path is set to empty.
### Documentation
Updated docs.

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1554,7 +1554,7 @@ func FindOneProjectRefByRepoAndBranchWithPRTesting(owner, repo, branch, calledBy
 	if err != nil {
 		return nil, errors.Wrapf(err, "finding merged repo refs for repo '%s/%s'", owner, repo)
 	}
-	if repoRef == nil || !repoRef.IsPRTestingEnabledByCaller(calledBy) || repoRef.RemotePath == "" {
+	if repoRef == nil || !repoRef.IsPRTestingEnabledByCaller(calledBy) {
 		grip.Debug(message.Fields{
 			"source":  "find project ref for PR testing",
 			"message": "repo ref not configured for PR testing untracked branches",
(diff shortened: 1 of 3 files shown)


model/project_ref.go
```diff
@@ -1564,6 +1564,16 @@ func FindOneProjectRefByRepoAndBranchWithPRTesting(owner, repo, branch, calledBy
...
+	if repoRef.RemotePath == "" {
+		grip.Error(message.Fields{
+			"source":  "find project ref for PR testing",
+			"message": "repo ref has no remote path, cannot use for PR testing",
+			"owner":   owner,
+			"repo":    repo,
+			"branch":  branch,
+		})
+		return nil, errors.Wrapf(err, "repo ref '%s' has no remote path, cannot use for PR testing", repoRef.Id)
+	}
```
malikchaya2, line 1576: Can we add a check for this case to `TestFindOneProjectRefByRepoAndBranchWithPRTesting`?

malikchaya2 (changes requested):

model/project_ref.go
hadjri: Sure, updated one of the existing test cases to check this case

PR #7096, opened 2023-09-29
EVG-20809 Resolve task id and execution for task files

khelif96:
EVG-20809

### Description
There isn't currently a way to tie a file to a specific execution task id or execution when looking at tasks in a display task.


graphql/tests/task/files/queries/task_files.graphql
minnakt, line 1: Would you mind adding a test specifically for a display task with execution tasks since that's what seems to be the primary motivation for this PR?

graphql/util.go
```diff
@@ -55,7 +55,7 @@ func getGroupedFiles(ctx context.Context, name string, taskID string, execution
 		apiFile.BuildFromService(file)
 		apiFileList = append(apiFileList, &apiFile)
 	}
-	return &GroupedFiles{TaskName: &name, Files: apiFileList}, nil
+	return &GroupedFiles{TaskName: &name, Files: apiFileList, TaskID: taskID, Execution: execution}, nil
```
minnakt, line 58: I just have a question about what these fields are going to be used for — changes seems fine I am just slightly confused 😇... The linked PR in Parsley does not appear to use these new fields. So are these new fields maybe meant for generating the links in Spruce?

khelif96: Originally these new fields were gonna be used for that, but not anymore. I instead plan on resolving the parsley URL here as well so that it can be resolved by the rest API so that some other teams can also use them. That work will be completed in https://jira.mongodb.org/browse/EVG-20804

khelif96: I figured it couldn't hurt to include these fields here anyways.

minnakt (approved):

PR #7102, opened 2023-10-02
EVG-19760 Check project id and identifier are the same when enabling performance plugin

ZackarySantana:
EVG-19760

### Description
Adds a check to make sure that the project id and identifier are the same, to match the verbiage on the UI and as required.

### Testing
Add unit testing and existing tests.


model/project_ref_test.go
```diff
@@ -1658,22 +1658,22 @@ func TestFindProjectRefIdsWithCommitQueueEnabled(t *testing.T) {
 func TestValidatePeriodicBuildDefinition(t *testing.T) {
 	assert := assert.New(t)
 	testCases := map[PeriodicBuildDefinition]bool{
-		PeriodicBuildDefinition{
```
ZackarySantana, line 1661: My IDE just changed this when saving bc it's getting it from the map definition

model/project_ref.go
```diff
@@ -2032,6 +2032,14 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 		if catcher.HasErrors() {
 			return false, errors.Wrapf(catcher.Resolve(), "validating external links")
 		}
+		var pRef *ProjectRef
+		pRef, err = FindBranchProjectRef(projectId)
+		if err != nil {
+			return false, errors.Wrapf(err, "error getting project")
+		}
```
hadjri, line 2039: `pRef == nil` should be a second error case that we check right after this err check

hadjri (changes requested):

model/project_ref.go
```diff
@@ -2032,6 +2032,14 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+		if p != nil && pRef != nil && p.IsPerfEnabled() != pRef.IsPerfEnabled() && pRef.Identifier != pRef.Id {
```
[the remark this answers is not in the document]
ZackarySantana: It actually can return nil because (for some reason) internally, we return nil, nil if it is not found. Should this be a separate error case?
I didn't make them error cases because before they weren't but it does make sense since now they are required.

model/project_ref.go
```diff
@@ -2032,6 +2032,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+		var pRef *ProjectRef
+		pRef, err = FindBranchProjectRef(projectId)
+		if err != nil {
+			return false, errors.Wrapf(err, "error getting project")
+		}
+		if pRef == nil {
+			return false, errors.Errorf("project '%s' was not found", projectId)
+		}
+		// If the performance plugin is not currently enabled, and we are trying to
+		// change it to enabled but the id and identifier ar edifferent, no-op.
```
hadjri, line 2044: nit: typo here

hadjri, line 2044: Also we are erroring here, not no-opping if the id and identifier are different so we should change the comment above

model/project_ref.go
```diff
@@ -2032,6 +2032,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 		if catcher.HasErrors() {
 			return false, errors.Wrapf(catcher.Resolve(), "validating external links")
 		}
+		var pRef *ProjectRef
+		pRef, err = FindBranchProjectRef(projectId)
+		if err != nil {
+			return false, errors.Wrapf(err, "error getting project")
```
hadjri, line 2038: Error message should just be "getting project" per our error [style guide](https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit#heading=h.sp5f9pecb7lw)

hadjri (changes requested):

model/project_ref.go
```diff
@@ -2032,6 +2032,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+		var pRef *ProjectRef
+		pRef, err = FindBranchProjectRef(projectId)
+		if err != nil {
+			return false, errors.Wrapf(err, "getting project '%s'", projectId)
+		}
+		if pRef == nil {
+			return false, errors.Errorf("project '%s' was not found", projectId)
+		}
+		// If the performance plugin is not currently enabled, and we are trying to
+		// change it to enabled but the id and identifier ar different, we error.
```
hadjri, line 2044: nit: are instead of ar

hadjri (approved):
Approved w nit

PR #7103, opened 2023-10-02
EVG-19636 Remove remaining PatchedParserProject fallback usage and delete field

malikchaya2:
EVG-19636

### Description
The parser project used to be stored in the patch document as a YAML string called the PatchedParserProject field. However, it has since been deprecated so it should be removed.

Note: patches created before March 2023 will no longer be able to be finalized after this change. 

### Testing
Existing tests


graphql/tests/patch/results.json
```diff
@@ -254,18 +254,13 @@
...
-                  "displayName": "Ubuntu 16.04",
-                  "tasks": [
-                    "batchtime_task",
-                    "display_task",
-                    "display_task2",
-                    "patch_only_task"
-                  ]
+                  "name": "ubuntu1804",
+                  "displayName": "Ubuntu 18.04",
+                  "tasks": ["compile"]
```
Kimchelly, line 263: I assume these changes come from an unrelated commit? Did you merge main in a weird way?

model/patch_lifecycle.go
```diff
@@ -264,14 +264,12 @@ func GetPatchedProject(ctx context.Context, settings *evergreen.Settings, p *pat
 			return nil, nil, errors.Wrap(err, "finding and translating project")
```
Kimchelly, line 264: GH doesn't let me comment on the line but the comment on L261 has to be updated since the parser project is no longer stored as a string in the patch. The project to remove the parser project limit changed the parser project string to instead just store the document in the DB at the same time as the patch.

model/patch_lifecycle.go
```diff
@@ -685,13 +680,6 @@ func FinalizePatch(ctx context.Context, p *patch.Patch, requester string, github
 		ppStorageMethod = evergreen.ProjectStorageMethodDB
 	}
 
-	if p.PatchedParserProject != "" {
```
Kimchelly, line 688: GH doesn't let me comment on unchanged lines, but the comment on L677 has to be removed since the parser project is never stored as a string in the patch doc anymore.

Kimchelly: The fallback of setting a default storage method on L685 also should be removed because [the storage method is always set on patch creation for newer patches](https://github.com/evergreen-ci/evergreen/blob/b029ef66fa020c5c19294224103c4af794adc510/units/patch_intent.go#L376-L380), and we're assuming that nobody's going to finalize old patches from before March 2023. The version document can just inherit the patch document's project storage method as-is.

model/patch_lifecycle.go
```diff
@@ -327,16 +325,14 @@ func GetPatchedProjectConfig(ctx context.Context, settings *evergreen.Settings,
...
-		// check and get the PatchedProjectConfig after the patch is already
-		// created.
+		// already saved the project config as a string (if any) or stored
+		// the parser project document (i.e. ProjectStorageMethod). Since
+		// the project config is optional, the patch may be created and have
+		// already evaluated the patched project config, but there simply
+		// was none. Therefore, this is a valid way to check and get the
+		// PatchedProjectConfig after the patch is already created.
 		return p.PatchedProjectConfig, nil
 	}
```
Kimchelly, line 338: Small suggestion to mention here in a doc comment that if we get past this point, then the patch has not been created yet, so this is the first-time initialization to get the parser project and project config.

rest/route/patch_test.go
```diff
@@ -829,9 +830,17 @@ buildvariants:
...
 		Project:              projectRef.Id,
 		Githash:              "3c7bfeb82d492dc453e7431be664539c35b5db4b",
-		PatchedParserProject: config,
+		ProjectStorageMethod: evergreen.ProjectStorageMethodDB,
+		PatchedProjectConfig: config,
 	}
 	require.NoError(t, unfinalized.Insert())
+
+	pp := &serviceModel.ParserProject{}
+	err := util.UnmarshalYAMLWithFallback([]byte(config), &pp)
```
Kimchelly, line 839: Go is smart enough to make it work anyways, but it's not necessary to take the pointer of a pointer here. Just passing in `&serviceModel.ParserProject{}` is sufficient.

Same with the test below.

rest/route/patch_test.go
```diff
@@ -829,9 +830,17 @@ buildvariants:
...
+	require.NoError(t, err)
```
Kimchelly, line 840: nit: You can just combine L839-840.

Same with below.

model/patch_lifecycle_test.go
```diff
@@ -388,6 +329,28 @@ modules:
...
+				{
+					Name:   "evergreen",
+					Branch: "main",
+					Repo:   "git@github.com:evergreen-ci/evergreen.git",
+				},
+			}
+
+			patchConfig.PatchedParserProject.Id = p.Id.Hex()
+			patchConfig.PatchedParserProject.Modules = modules
+			patchConfig.PatchedParserProject.Identifier = &p.Project
```
Kimchelly, line 348: nit: `utility.ToStringPtr(p.Project)` for consistency

Kimchelly (changes requested):

graphql/tests/patch/results.json
malikchaya2: no I changed which patch "project-variants-and-tasks.graphql" uses so the results are different.

model/patch_lifecycle.go
malikchaya2: good catch, thanks!

model/patch_lifecycle.go
malikchaya2: done

model/patch_lifecycle.go
malikchaya2: done

rest/route/patch_test.go
malikchaya2: oops. Removed the `&`.

rest/route/patch_test.go
malikchaya2: done

Kimchelly (approved):
Thanks for removing this! 🙏

PR #7104, opened 2023-10-02
EVG-6867: Modify MainlineCommits query to accept revision param

minnakt:
EVG-6867

### Description
This PR includes the backend changes that makes it possible for users to "jump" to a certain commit. Users can input a specific commit hash, and they will be taken to that point in the commit history. Ideally, the commit should show up towards the center of the page so that users can compare what happened before and after that commit.

Note that pagination with filters applied may not work entirely correctly due to [EVG-20965](https://jira.mongodb.org/browse/EVG-20965). It's worth mentioning that this resolver will likely need significant refactoring in the future to fix this bug.

### Testing
- GraphQL tests
- On staging. Here is a video of what this looks like:

https://github.com/evergreen-ci/evergreen/assets/47064971/c2549b59-3a7d-42fb-8cd2-117d7e0deffc


graphql/query_resolver.go
```diff
@@ -807,9 +808,24 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+	revision := utility.FromStringPtr(options.Revision)
+	if options.SkipOrderNumber == nil && revision != "" {
+		found, err := model.VersionFindOne(model.VersionByProjectIdAndRevisionPrefix(projectId, revision))
+		if err != nil {
+			graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("could not find git commit '%s'", revision)))
+		} else if found == nil {
+			graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("could not find git commit '%s'", revision)))
+		} else {
+			// Offset the order number so the specified revision lands nearer to the center of the page.
+			skipOrderNumber = found.RevisionOrderNumber + 1 + limit/2
```
minnakt, line 822: This works best if there are mostly uncollapsed commits on the project's waterfall. Then it's likely that this `skipOrderNumber` belongs to an uncollapsed commit, so you will be able to see commits before & after the specified revision. 

However, it's possible for this `skipOrderNumber` to belong to a collapsed commit, in which case you might not see any commits after the specified revision. To get around this I believe I would have to keep "stepping back" and querying until I find an uncollapsed commit, but this is difficult and probably expensive because there could be more than a hundred collapsed commits. To ensure that a commit is not collapsed, I would have to check that that each commit is activated and matches the applied filters (another aggregation pipeline).

At the very least, I think this should guarantee that the commit will be somewhere on the page. Another option would just be to not offset this number, but then you would not ever be able to compare the before & after.

graphql/query_resolver.go
```diff
@@ -807,9 +808,29 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 	if len(requesters) == 0 {
 		requesters = evergreen.SystemVersionRequesterTypes
 	}
+
+	skipOrderNumber := utility.FromIntPtr(options.SkipOrderNumber)
+	revision := utility.FromStringPtr(options.Revision)
+
+	if options.SkipOrderNumber == nil && revision != "" {
+		if len(revision) < 7 {
```
hadjri, line 816: Can this be moved to a constant?

graphql/query_resolver.go
```diff
@@ -807,9 +808,29 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+			graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("at least 7 characters must be provided for the revision")))
+		} else {
+			found, err := model.VersionFindOne(model.VersionByProjectIdAndRevisionPrefix(projectId, revision))
+			if err != nil {
+				graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("could not find git commit '%s'", revision)))
+			} else if found == nil {
+				graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("could not find git commit '%s'", revision)))
+			} else {
+				// Offset the order number so the specified revision lands nearer to the center of the page.
+				skipOrderNumber = found.RevisionOrderNumber + 1 + limit/2
```
hadjri, line 826: Just confirming that only the limit is meant to be halved here, not the entire sum?

graphql/query_resolver.go
```diff
@@ -807,9 +808,29 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+
+	skipOrderNumber := utility.FromIntPtr(options.SkipOrderNumber)
+	revision := utility.FromStringPtr(options.Revision)
+
+	if options.SkipOrderNumber == nil && revision != "" {
+		if len(revision) < 7 {
+			graphql.AddError(ctx, PartialError.Send(ctx, fmt.Sprintf("at least 7 characters must be provided for the revision")))
+		} else {
+			found, err := model.VersionFindOne(model.VersionByProjectIdAndRevisionPrefix(projectId, revision))
+			if err != nil {
```
hadjri, line 820: We should modify the err message here so it's distinct from the case below

graphql/query_resolver.go
```diff
@@ -822,8 +843,8 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
 	matchingVersionCount := 0
 
 	// We only want to return the PrevPageOrderNumber if a user is not on the first page
-	if options.SkipOrderNumber != nil {
-		prevPageCommit, err := model.GetPreviousPageCommitOrderNumber(ctx, projectId, utility.FromIntPtr(options.SkipOrderNumber), limit, requesters)
+	if skipOrderNumber != 0 {
```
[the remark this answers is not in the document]
hadjri: Is setting `options.SkipOrderNumber` intentionally at zero a valid case? Seems that this conditional would evaluate true before this change, just making sure that we do want to make a change to skip that case

hadjri (changes requested):

graphql/query_resolver.go
minnakt: yep, the limit actually represents how many commits a user is currently viewing on the page. By adding half of the limit, it means that the commit will end up roughly in the center of the page

graphql/query_resolver.go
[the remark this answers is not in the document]
minnakt: Yes, the if-statements using a different check was intentional. The new logic means that the revision only gets used when `options.skipOrderNumber == nil`. It’s done this way because the way to paginate commits is through the skipOrderNumber, and so `options.revision` and `options.skipOrderNumber` need to share that variable with each other.

So with this new logic, even if `options.skipOrderNumber` is `nil`, there could be previous pages if we've jumped to a particular commit, and we'd want to calculate the previous page offset. I hope that made sense 😅

hadjri (approved):

PR #7109, opened 2023-10-04
EVG-20795 avoid overwriting patch task/variants

ablack12:
EVG-20795
### Description
Discovered that when we update description only (using the lil UI pencil) we overwrite existing patch task/variant definitions.

Considered making a new mutation instead, but that would be a lot more work and I think would just leave the technical debt of potentially hitting this bug with how SchedulePatch is written. 

### Testing
Added a unit test, and [tested on staging](https://evergreen-staging.corp.mongodb.com/rest/v2/patches/650e02db97b1d3cc9cd2e3da) that the fields weren't erased after updating a description, and were before.

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -116,7 +116,7 @@ type PatchUpdate struct {
 	VariantsTasks       []patch.VariantTasks `json:"variants_tasks,omitempty"`
 }
 
-// ConfigurePatch validates and creates the updated tasks/variants, and updates description if needed.
+// ConfigurePatch validates and creates the updated tasks/variants if given, and updates description if needed.
 // Returns an http status code and error.
 func ConfigurePatch(ctx context.Context, settings *evergreen.Settings, p *patch.Patch, version *Version, proj *ProjectRef, patchUpdateReq PatchUpdate) (int, error) {
 	var err error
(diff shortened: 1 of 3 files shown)


model/patch_lifecycle.go
```diff
@@ -142,17 +142,17 @@ func ConfigurePatch(ctx context.Context, settings *evergreen.Settings, p *patch.
...
-	// update the description for both reconfigured and new patches
-	if err = p.SetVariantsTasks(tasks.TVPairsToVariantTasks()); err != nil {
-		return http.StatusInternalServerError, errors.Wrap(err, "setting description")
+	patchVariantTasks := tasks.TVPairsToVariantTasks()
+	if len(patchVariantTasks) > 0 {
+		if err = p.SetVariantsTasks(patchVariantTasks); err != nil {
+			return http.StatusInternalServerError, errors.Wrap(err, "setting description")
+		}
 	}
-	p.Activated = true
```
Kimchelly, line 155: If this isn't set here anymore, when does it get set? In `addNewTasksAndBuildsForPatch`?

Also for my own understanding, does a patch being activated mean that it's finalized (i.e. tasks will be scheduled)? Or does it mean something else?

model/patch_lifecycle_test.go
```diff
@@ -1158,3 +1158,56 @@ func TestAbortPatchesWithGithubPatchData(t *testing.T) {
...
+		Id: mgobson.NewObjectId().Hex(),
+	}
+	req := PatchUpdate{
+		Description: "updating the description only!",
+	}
+	pp := ParserProject{
+		Id: id.Hex(),
+	}
+	assert.NoError(t, pp.Insert())
+	_, err := ConfigurePatch(ctx, evergreen.GetEnvironment().Settings(), p, v, pRef, req)
```
Kimchelly, line 1204: Use `mock.Environment` instead of the global environment if possible when testing with the env (unless you need to depend on the global testing env for some reason), or just pass in an empty `evergreen.Settings{}` if the test doesn't depend on any particular settings (seems like it doesn't).

model/patch_lifecycle_test.go
```diff
@@ -1158,3 +1158,56 @@ func TestAbortPatchesWithGithubPatchData(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	p, err = patch.FindOneId(id.Hex())
+	assert.NoError(t, err)
+	require.NotNil(t, p)
+	assert.Equal(t, p.Description, req.Description)
+	assert.NotEmpty(t, p.VariantsTasks)
+	assert.NotEmpty(t, p.Tasks)
```
Kimchelly, line 1212: Should this also check for `p.Activated`?

Kimchelly (changes requested):

model/patch_lifecycle.go
Kimchelly: Feel like the best part of that doc comment is that I wrote it and apparently forgot.

Kimchelly (approved):

PR #7110, opened 2023-10-04
EVG-19086 Introduce option to match attach.artifacts with exact file names

hadjri:
EVG-19086

### Description
Introducing a new `exact_file_names` flag on the attach.artifacts command, which will cause the command to match exact file names rather than performing regex style matching on the entire work directory.
### Testing
Added some new unit tests to the attach artifacts test suite.
### Documentation
Added a piece to the attach.artifacts documentation for this new field and its functionality.

### DIFF
--- a/agent/command/attach_artifacts.go
+++ b/agent/command/attach_artifacts.go
@@ -86,7 +86,7 @@ func (c *attachArtifacts) Execute(ctx context.Context,
 	for idx := range c.Files {
 		segment, err = readArtifactsFile(getWorkingDirectory(conf, c.Prefix), c.Files[idx])
 		if err != nil {
-			if (c.Optional || c.ExactFileNames) && os.IsNotExist(errors.Cause(err)) {
+			if c.Optional && os.IsNotExist(errors.Cause(err)) {
 				// pass;
 			} else {
 				catcher.Add(err)
--- a/agent/command/attach_artifacts.go
+++ b/agent/command/attach_artifacts.go
@@ -20,6 +20,10 @@ type attachArtifacts struct {
 	// Files is a list of files, using gitignore syntax.
 	Files []string `mapstructure:"files" plugin:"expand"`
 
+	// ExactFileNames, when set to true, causes this command to treat the files array as an array of exact filenames to match,
+	// rather than the default behavior, which treats files as an array of gitignore file globs.
+	ExactFileNames bool `mapstructure:"exact_file_names"`
+
 	// Prefix is an optional directory prefix to start file globbing in, relative to Evergreen's working directory.
 	Prefix string `mapstructure:"prefix" plugin:"expand"`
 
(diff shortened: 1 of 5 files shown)


agent/command/attach_artifacts.go
```diff
@@ -80,7 +86,7 @@ func (c *attachArtifacts) Execute(ctx context.Context,
 	for idx := range c.Files {
 		segment, err = readArtifactsFile(getWorkingDirectory(conf, c.Prefix), c.Files[idx])
 		if err != nil {
-			if c.Optional && os.IsNotExist(errors.Cause(err)) {
+			if (c.Optional || c.ExactFileNames) && os.IsNotExist(errors.Cause(err)) {
```
malikchaya2, line 89: Why do we want to ignore the errors for exactFileNames? I don't know if exactFileNames should automatically be considered optional.

agent/command/attach_artifacts_test.go
```diff
@@ -102,6 +113,17 @@ func (s *ArtifactsSuite) TestArtifactErrorsIfDoesNotExist() {
 	s.cmd.Files = []string{"foo"}
 	s.Error(s.cmd.Execute(s.ctx, s.comm, s.logger, s.conf))
 	s.Len(s.cmd.Files, 0)
+	s.Len(s.mock.AttachedFiles, 0)
```
malikchaya2, line 116: I see that this is tested here, but I'd still strongly advocate to test this manually in staging as well, since historically file changes have been prone to bugs that were hard to discover from unit tests.

malikchaya2 (changes requested):

agent/command/attach_artifacts.go
hadjri: I put this to make it such that the command won't fail if the files specified in the YAML don't exist, since `ExactFileNames` is explicit about what files to pick and the current way the command works is more forgiving in this regard since it automatically detects files based on the gitignore style matching. 

That being said I'm fine to remove this and have the expectation that the user is going to be sure the files with these exact names are going to exist or else it will fail.

malikchaya2: Since this request came from a specific user, I'd suggest consulting them to see which they prefer. 
If they prefer that if behave like optional does, I'd suggest adding that to the docs.

malikchaya2 (approved):
LGTM after this is tested on staging. 

I also suggest reaching out to the user before merging to get their preference on if it should be treated as optional.

hadjri:
> LGTM after this is tested on staging.
> 
> I also suggest reaching out to the user before merging to get their preference on if it should be treated as optional.

Confirmed [working](https://spruce-staging.corp.mongodb.com/task/sandbox_git_env_test_validate_commit_message_patch_d631380317a8478b879cc27d6171508dea2b11df_65203eba97b1d37768e33633_23_10_06_17_07_17/logs?execution=0) on staging with the new parameter

PR #7113, opened 2023-10-05
EVG-20916 Remove Unused GitHub hook code

bynn:
EVG-29016

### Description
We no longer require github webhooks to be created and set up for repos since they go through the github app 
there is no need to create and store them or validate them anymore 

### Testing
tested that all github functionality still works with disabled webhook, and validations work even if webhook is deleted from the DB 

created follow up ticket to update the warning about webhooks on spruce 
 https://jira.mongodb.org/browse/EVG-21038


config.go
```diff
@@ -604,7 +606,10 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 			"appId":   authFields.AppId,
 			"ticket":  "EVG-19966",
 		}))
-		return "", errors.Wrap(err, "finding installation id")
+		if strings.Contains(err.Error(), "404") {
```
ablack12, line 609: Do we need this error at all? Seems like we could just not return an error here instead, and then we'll hit line 633 normally

ablack12: I am also wondering why this isn't an error though. Don't we need to know if the github app isn't installed?

graphql/project_settings_resolver.go
```diff
@@ -21,11 +21,13 @@ func (r *projectSettingsResolver) Aliases(ctx context.Context, obj *restModel.AP
 
 // GithubWebhooksEnabled is the resolver for the githubWebhooksEnabled field.
 func (r *projectSettingsResolver) GithubWebhooksEnabled(ctx context.Context, obj *restModel.APIProjectSettings) (bool, error) {
-	hook, err := model.FindGithubHook(utility.FromStringPtr(obj.ProjectRef.Owner), utility.FromStringPtr(obj.ProjectRef.Repo))
+	owner := utility.FromStringPtr(obj.ProjectRef.Owner)
+	repo := utility.FromStringPtr(obj.ProjectRef.Repo)
+	hasApp, err := evergreen.GetEnvironment().Settings().HasGitHubApp(ctx, owner, repo, nil)
 	if err != nil {
-		return false, InternalServerError.Send(ctx, fmt.Sprintf("Database error finding github hook for project '%s': %s", *obj.ProjectRef.Id, err.Error()))
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("Error GitHub app installation for project '%s' in '%s/%s': '%s'", utility.FromStringPtr(obj.ProjectRef.Id), owner, repo, err.Error()))
```
ablack12, line 28: nit: "error github app installation" doesn't work as a sentence

graphql/repo_settings_resolver.go
```diff
@@ -21,11 +21,13 @@ func (r *repoSettingsResolver) Aliases(ctx context.Context, obj *restModel.APIPr
 
 // GithubWebhooksEnabled is the resolver for the githubWebhooksEnabled field.
 func (r *repoSettingsResolver) GithubWebhooksEnabled(ctx context.Context, obj *restModel.APIProjectSettings) (bool, error) {
-	hook, err := model.FindGithubHook(utility.FromStringPtr(obj.ProjectRef.Owner), utility.FromStringPtr(obj.ProjectRef.Repo))
+	owner := utility.FromStringPtr(obj.ProjectRef.Owner)
+	repo := utility.FromStringPtr(obj.ProjectRef.Repo)
+	hasApp, err := evergreen.GetEnvironment().Settings().HasGitHubApp(ctx, owner, repo, nil)
 	if err != nil {
-		return false, InternalServerError.Send(ctx, fmt.Sprintf("Database error finding github hook for project '%s': %s", *obj.ProjectRef.Id, err.Error()))
+		return false, InternalServerError.Send(ctx, fmt.Sprintf("Error GitHub app installation for project '%s' in '%s/%s': '%s'", utility.FromStringPtr(obj.ProjectRef.Id), owner, repo, err.Error()))
```
ablack12, line 28: same here

graphql/project_settings_resolver.go
ablack12: shouldn't this just say issue getting github app though?  Seems like the error _wouldn't_ be about installation, since we don't error in that case (same with the grip statement below)

model/project_ref.go
```diff
@@ -1658,27 +1658,18 @@ func FindOneProjectRefWithCommitQueueByOwnerRepoAndBranch(owner, repo, branch st
...
 	}
 
-	hook, err = SetupNewGithubHook(ctx, *settings, projectRef.Owner, projectRef.Repo)
-	if err != nil {
+	hasApp, err := settings.HasGitHubApp(ctx, projectRef.Owner, projectRef.Repo, nil)
+	if err != nil || !hasApp {
 		// don't return error:
 		// sometimes people change a project to track a personal
 		// branch we don't have access to
 		grip.Error(message.WrapError(err, message.Fields{
```
ablack12, line 1671: This isn't going to log in the !hasApp case, because grip.Error will noop if err is nil

model/project_ref.go
```diff
@@ -1803,10 +1791,11 @@ func GetProjectSettingsById(projectId string, isRepo bool) (*ProjectSettings, er
 
 // GetProjectSettings returns the ProjectSettings of the given identifier and ProjectRef
 func GetProjectSettings(p *ProjectRef) (*ProjectSettings, error) {
-	hook, err := FindGithubHook(p.Owner, p.Repo)
-	if err != nil {
-		return nil, errors.Wrapf(err, "finding GitHub hook for project '%s'", p.Id)
-	}
+
+	// GetProjectSettings shouldn't error if GitHub app is not set up, or installed on a repo.
+	// It will return false for githubHooksEnabled though.
```
ablack12, line 1796: I think we can skip this second sentence

ablack12: Should we be caching whether or not the app is installed? Can we just use the existing GithubHooks collection for this? Seems kind of expensive to keep asking github for it (even if they don't count it towards our rate limit, seems slow)

ablack12 (changes requested):

graphql/project_settings_resolver.go
bynn: i meant to write error *verifying* app installation

config.go
bynn: discussed to keep it since its useful to know when the app is not installed vs there was a problem getting the app

graphql/repo_settings_resolver.go
bynn: updated as above

model/project_ref.go
bynn: updated

model/project_ref.go
bynn: caching installation id now

model/project_ref.go
```diff
@@ -1803,10 +1802,10 @@ func GetProjectSettingsById(projectId string, isRepo bool) (*ProjectSettings, er
 
 // GetProjectSettings returns the ProjectSettings of the given identifier and ProjectRef
 func GetProjectSettings(p *ProjectRef) (*ProjectSettings, error) {
-	hook, err := FindGithubHook(p.Owner, p.Repo)
-	if err != nil {
-		return nil, errors.Wrapf(err, "finding GitHub hook for project '%s'", p.Id)
-	}
+
+	// GetProjectSettings shouldn't error if GitHub app is not set up, or installed on a repo.
```
ablack12, line 1806: hasApp only returns an error if there's an issue getting the installation token, so this comment isn't quite right. I think we can also drop the function name here.

github_app.go
```diff
@@ -0,0 +1,235 @@
...
+	// Check cache for installation ID first.
+	installationID, err := getInstallationID(ctx, owner, repo)
+	if err != nil {
+		return false, errors.Wrapf(err, "getting cached installation id for '%s/%s'", owner, repo)
+	}
+	if installationID != 0 {
+		return true, nil
+	}
+
+	token, err := s.CreateInstallationToken(ctx, owner, repo, opts)
```
ablack12, line 212: We don't actually need the token here. we just need the installation ID. can we refactor this a bit? (or include it in the future refactor for thirdparty mentioned above)

ablack12 (changes requested):

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	if err := res.Decode(&installation); err != nil {
+		return 0, errors.Wrapf(err, "decoding installation ID for '%s/%s", owner, repo)
+	}
+
+	return installation.InstallationID, nil
+}
+
+type githubAppAuth struct {
+	AppId      int64
```
ablack12, line 98: this can be a private var too

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+	privateKey []byte
+}
+
+// getGithubAppAuth returns app id and app private key if it exists.
```
ablack12, line 102: returns **the app ID and private key if **they exist 💃

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+			InstallationID: installationID.GetID(),
+		}
+		if err := installation.Upsert(ctx); err != nil {
+			return "", errors.Wrapf(err, "saving installation id for '%s/%s'", owner, repo)
+		}
+	}
+
+	token, _, err := client.Apps.CreateInstallationToken(ctx, installationID, opts)
+	if err != nil || token == nil {
+		return "", errors.Wrapf(err, "creating installation token for installation id: %d", installationID)
```
ablack12, line 189: reminder this is going to return no error if token is nil and err is nil

ablack12: Also "installation id '%d'"

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return token.GetToken(), nil
+}
+
+// HasGitHubApp returns true if the GitHub app is installed on given owner/repo.
+// Only returns an error if app is installed.
```
ablack12, line 195: if **the app is installed

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+func (s *Settings) HasGitHubApp(ctx context.Context, owner, repo string, opts *github.InstallationTokenOptions) (bool, error) {
+	// Check cache for installation ID first.
```
ablack12, line 197: Check **the cache

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+	installationID, err := getInstallationID(ctx, owner, repo)
+	if err != nil {
+		return false, errors.Wrapf(err, "getting cached installation id for '%s/%s'", owner, repo)
+	}
+	if installationID != 0 {
+		return true, nil
+	}
+
+	grip.Info(s)
```
ablack12, line 206: remove?

github_app.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// getInstallationID returns the installation ID for GitHub app if it's installed for the given owner/repo.
+func getInstallationID(ctx context.Context, owner, repo string) (int64, error) {
+	if err := validateOwnerRepo(owner, repo); err != nil {
+		return 0, err
+	}
+
+	installation := &GitHubAppInstallation{}
+	res := GetEnvironment().DB().Collection(GitHubAppCollection).FindOne(ctx, byOwnerRepo(owner, repo))
```
ablack12, line 83: i assume this is copied code, but we can just use the classic
```
doc := &structure{}
err := db.FindOneQ(Collection, query, doc)
	if adb.ResultsNotFound(err) {
		return nil, nil
	}
```

github_app.go
ablack12: Can you add a TODO here: refactor to get installation ID instead of token

github_app.go
```diff
@@ -0,0 +1,231 @@
+package evergreen
```
ablack12, line 1: this shouldn't be top-level; this should be under model, same as the deleted files.

model/project_ref.go
```diff
@@ -1658,27 +1658,16 @@ func FindOneProjectRefWithCommitQueueByOwnerRepoAndBranch(owner, repo, branch st
 
 // EnableWebhooks returns true if a hook for the given owner/repo exists or was inserted.
 func EnableWebhooks(ctx context.Context, projectRef *ProjectRef) (bool, error) {
```
ablack12, line 1660: this function name and description is wrong now; seems like we don't neven need this function? Maybe just SetTracksPushEvents or something.

model/project_ref.go
```diff
@@ -1687,10 +1676,21 @@ func EnableWebhooks(ctx context.Context, projectRef *ProjectRef) (bool, error) {
...
 		return false, nil
 	}
-
-	if err = hook.Insert(); err != nil {
-		return false, errors.Wrapf(err, "inserting new webhook for project '%s'", projectRef.Id)
+	// don't return error:
+	// sometimes people change a project to track a personal
+	// branch we don't have access to
+	if !hasApp {
+		grip.Error(message.Fields{
```
ablack12, line 1683: not necessarily an error; should log at a lower level.

graphql/project_settings_resolver.go
```diff
@@ -21,11 +22,16 @@ func (r *projectSettingsResolver) Aliases(ctx context.Context, obj *restModel.AP
...
 func (r *projectSettingsResolver) GithubWebhooksEnabled(ctx context.Context, obj *restModel.APIProjectSettings) (bool, error) {
-	hook, err := model.FindGithubHook(utility.FromStringPtr(obj.ProjectRef.Owner), utility.FromStringPtr(obj.ProjectRef.Repo))
-	if err != nil {
-		return false, InternalServerError.Send(ctx, fmt.Sprintf("Database error finding github hook for project '%s': %s", *obj.ProjectRef.Id, err.Error()))
-	}
-	return hook != nil, nil
+	owner := utility.FromStringPtr(obj.ProjectRef.Owner)
+	repo := utility.FromStringPtr(obj.ProjectRef.Repo)
+	hasApp, err := evergreen.GetEnvironment().Settings().HasGitHubApp(ctx, owner, repo, nil)
+	grip.Error(message.WrapError(err, message.Fields{
```
bynn, line 28: does not error as it breaks the project settings page but I still had to return error because I believe that's just what GraphQL resolvers expect.

github_app.go
ablack12: we do wrap IDs -- in this case the id happens to be an int but typically they are strings.

github_app.go
ablack12: Can you add a comment to that ticket? And maybe assign to your next sprint, I think we should do this ASAP so this doesn't get any worse

github_app.go
ablack12: same for below too

ablack12 (approved):

ablack12:
I'd suggest merging main again, could help with tests

PR #7114, opened 2023-10-05
EVG-20892 Set users setting for spruce UI when clearing them

ZackarySantana:
EVG-20892

### Description
Recent returning interns have had their settings wiped when offboarded which caused them to use the old ui when they returned. This prevents that issue in the future by setting the spruce UI setting when offboarding a user.

### Testing
Unit tests

### DIFF
--- a/model/user/db.go
+++ b/model/user/db.go
@@ -488,7 +488,7 @@ func ClearLoginCache(user gimlet.User) error {
 
 // ClearUser clears one user's settings, roles, and login cache.
 func ClearUser(userId string) error {
-	update := bson.M{
+	unsetUpdate := bson.M{
 		"$unset": bson.M{
 			SettingsKey: bson.M{
 				userSettingsGithubUserKey:    1,
@@ -498,19 +498,21 @@ func ClearUser(userId string) error {
 			RolesKey:      1,
 			LoginCacheKey: 1,
 		},
+	}
+	query := bson.M{IdKey: userId}
+	if err := UpdateOne(query, unsetUpdate); err != nil {
+		return errors.Wrap(err, "unsetting user settings")
+	}
+	setUpdate := bson.M{
 		"$set": bson.M{
 			SettingsKey: bson.M{
 				UseSpruceOptionsKey: bson.M{
 					SpruceV1Key: 1,
 				},
 			},
-		}}
-	query := bson.M{IdKey: userId}
-	if err := UpdateOne(query, update); err != nil {
-		return errors.Wrap(err, "unsetting user settings")
+		},
 	}
-
-	return nil
+	return errors.Wrap(UpdateOne(query, setUpdate), "unsetting user settings")
 }
 
 // ClearAllLoginCaches clears all users' login caches, forcibly logging them
--- a/model/user/db.go
+++ b/model/user/db.go
@@ -487,6 +487,9 @@ func ClearLoginCache(user gimlet.User) error {
 }
 
 // ClearUser clears one user's settings, roles, and login cache.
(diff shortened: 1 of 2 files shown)


model/user/db.go
```diff
@@ -488,18 +488,31 @@ func ClearLoginCache(user gimlet.User) error {
 
 // ClearUser clears one user's settings, roles, and login cache.
```
ablack12, line 489: we should update this to explain why we only clear some settings.

model/user/db.go
```diff
@@ -488,18 +488,31 @@ func ClearLoginCache(user gimlet.User) error {
...
 		return errors.Wrap(err, "unsetting user settings")
 	}
-
-	return nil
+	setUpdate := bson.M{
+		"$set": bson.M{
+			SettingsKey: bson.M{
+				UseSpruceOptionsKey: bson.M{
+					SpruceV1Key: 1,
+				},
```
ablack12, line 511: 1 is only used for unset; for set, you want to use the actual value (true, false). In this case, I'm not actually sure waht you're going for. You can also just do this in one update: it would contain both the set and the unset (example: https://github.com/evergreen-ci/evergreen/blob/9f3350a1017ac4c63faa2480e273af4c7706a294/model/task/db.go#L176)

ablack12 (changes requested):

model/user/db.go
ZackarySantana: Let me know if I'm too wordy... Thought it better so no one has to investigate whyy this function does what it does.

model/user/db.go
ablack12: Ah that makes sense, alas. What if we just didn't update this at all? Let them keep their default?

model/user/db.go
```diff
@@ -487,19 +487,35 @@ func ClearLoginCache(user gimlet.User) error {
 }
 
 // ClearUser clears one user's settings, roles, and login cache.
```
ablack12, line 489: Delete this line

model/user/db.go
```diff
@@ -487,19 +487,35 @@ func ClearLoginCache(user gimlet.User) error {
...
+// ClearUser clears the users sensitive settings (GitHub and Slack credentials),
+// roles, and invalidates their login cache. It sets their settings to use Spruce,
+// or else if this user ever becomes active, they will revert to the old UI.
 func ClearUser(userId string) error {
-	update := bson.M{
+	unsetUpdate := bson.M{
 		"$unset": bson.M{
-			SettingsKey:   1,
+			SettingsKey: bson.M{
```
ablack12, line 496: We should also clear notifications, timezone, and region.

model/user/db.go
ablack12: Or if we're going to reset this anyway, we can just unset the entire user settings.

ablack12: You can just set this to true, not "true". SpruceV1 is a boolean.

model/user/db.go
```diff
@@ -486,20 +486,30 @@ func ClearLoginCache(user gimlet.User) error {
...
+	setUpdate := bson.M{
+		"$set": bson.M{
+			SettingsKey: bson.M{
+				UseSpruceOptionsKey: bson.M{
+					SpruceV1Key: "true",
+				},
+			},
+		},
+	}
+	return errors.Wrap(UpdateOne(query, setUpdate), "unsetting user settings")
```
ablack12, line 512: this error message should be like "defaulting spruce setting" or something

ablack12 (approved):
LGTM with nits

PR #7117, opened 2023-10-06
EVG-20853 Enable runtime expansions to determine a version's manifest

hadjri:
EVG-20853

### Description
The request here essentially boils down to:
1. Dynamically retrieving and setting some expansions for a task in the `pre` block
2. Using those fetched expansions for the `branch` field in for the task's modules

Currently however, manifest documents representing which branch / commit of the modules to use are created upon creation of the [mainline commit version](https://gist.github.com/hadjri/884ae233018cf56cb7f00cc83af1a313#file-repotracker-go-L367-L376) or [patch](https://github.com/evergreen-ci/evergreen/blob/main/model/patch_lifecycle.go#L571-L574), and not during task runtime.

Therefore the proposed solution here is to directly send over a list of modules as part of the `manifest/load` agent route which, if expansions are set for the modules, will be a correctly expanded representation of the modules list. This is because the manifest load handler in the agent already is set up to [create a manifest if none exists](https://gist.github.com/hadjri/b26130b99a691a7ff75b45fa0b933476#file-agent-go-L1294-L1300)

The tradeoff here is that, if a user wants to do this it would mean that their manifest cannot be created upon patch creation, and will have to wait until the first task for a patch runs. Currently if manifest construction fails for a patch, the patch creation will error out. Because of this I wanted to raise here as an [open ended question](https://github.com/evergreen-ci/evergreen/pull/7117/files#diff-db2ef846d75486c0e998561c1e2b98dbf5536368af5c37b0c17b76938b09bf50R571-R573) whether it'd be best to change patch finalization to no longer depend on a successful manifest creation, or to somehow skip the manifest creation for patches that have expansions set for their modules.

### Testing
[Sample manifest](blob:https://spruce-staging.corp.mongodb.com/1865d5d8-0516-432c-b526-04a1501fa950) that was created for the sandbox repotracker version from this [commit](https://github.com/evergreen-ci/commit-queue-sandbox/commit/d9322c0388be7ef49243bf48a2f02efd37f45df1) that added expansions to the modules

### Documentation
Will update docs on expansions for modules if approval is granted.

### DIFF
--- a/agent/command/git.go
+++ b/agent/command/git.go
@@ -268,9 +268,14 @@ func moduleRevExpansionName(name string) string { return fmt.Sprintf("%s_rev", n
 func (c *gitFetchProject) manifestLoad(ctx context.Context,
 	comm client.Communicator, logger client.LoggerProducer, conf *internal.TaskConfig) error {
 
+	modules := conf.Project.Modules
+	for i := range modules {
+		if err := util.ExpandValues(&modules[i], &conf.Expansions); err != nil {
+			return errors.Wrap(err, "applying expansions")
+		}
+	}
 	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
-
-	manifest, err := comm.GetManifest(ctx, td)
+	manifest, err := comm.GetManifest(ctx, td, modules)
 	if err != nil {
 		return errors.Wrapf(err, "loading manifest for task '%s'", td.ID)
 	}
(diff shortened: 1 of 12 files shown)


ablack12 (commented):
Marking as reviewed; discussing open question in DM.

hadjri:
Closing in favor of https://github.com/evergreen-ci/evergreen/pull/7144

PR #7124, opened 2023-10-09
EVG-20788 Implement stepback bisection when it is enabled

ZackarySantana:
[EVG-20788](https://jira.mongodb.org/browse/EVG-20788)

### Description
This implements the stepback bisection algorithm. This should only run when stepback bisect is enabled, and should get the midway task between tasks and activate them, until it has confirmed that which task is the problem task.

### Testing
Add unit testing. Going to do more advance testing but want some review on initial thoughts


model/task/task.go
```diff
@@ -1288,7 +1288,7 @@ func SetTasksScheduledTime(tasks []Task, scheduledTime time.Time) error {
 	return nil
 }
 
-// GetTaskIdBetweenIds gets the task between two task given that they are
+// FindMidwayTask gets the task between two task given that they are
```
hadjri, line 1291: this can now be unexported and we should update the comment on `FindMidwayTaskFromIds` to just be descriptive of what the pure functionality of the function is.

model/task_lifecycle.go
```diff
@@ -1203,7 +1271,6 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
 
 // evalBisectStepback performs bisect stepback on the task.
 func evalBisectStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
-	// TODO: EVG-20788 implement stepback bisection.
```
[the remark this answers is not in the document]
hadjri: It's worth noting that in this case, we'd want to check that the task is failed _or_ that the `StepbackInfo` is not nil for the task to account for the case in the bisection model where we encounter a successful task but still need to continue the bisection. If `StepbackInfo` is non-nil that would show us that it was triggered as part of stepback bisection and we still need to hit `doBisectStepback`.

model/task_lifecycle.go
```diff
@@ -579,7 +588,66 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
...
+	}
+
+	// If this is a genearator task, activate generated tasks.
+	if nextTask.GenerateTask {
+		if err = nextTask.SetGeneratedTasksToActivate(nextTask.BuildVariant, nextTask.DisplayName); err != nil {
+			return errors.Wrap(err, "setting generated tasks to activate")
+		}
+	}
+
+	return errors.Wrap(nextTask.SetStepbackInfo(s), "setting stepback info")
```
hadjri, line 650: We should set the stepback info for this next task before we call `SetActiveState` in order to ensure the task has the proper stepback info immediately once it is activated

hadjri (changes requested):

model/task/task.go
```diff
@@ -1313,6 +1313,23 @@ func FindMidwayTask(t1, t2 Task) (*Task, error) {
 	return FindOne(db.Query(ByRevisionOrderNumber(t1.BuildVariant, t1.DisplayName, t1.Project, t1.Requester, mid)))
 }
 
+// FindMidwayTaskFromIds gets the task between two task given that they are
```
hadjri, line 1316: nit: two tasks. I also think its implied that the order of ids does not matter so let's remove that and just start the next sentence with `If the task passed cannot have...`

model/task/task.go
```diff
@@ -1313,6 +1313,23 @@ func FindMidwayTask(t1, t2 Task) (*Task, error) {
...
+// from the same project, requester, build variant, and display name. The
+// order of the ID's does not matter and if the task passed cannot have a
+// middle (i.e. it is sequential tasks or the same task) it will return the
+// the first task given.
+func FindMidwayTaskFromIds(t1Id, t2Id string) (*Task, error) {
+	t1, err := FindOneId(t1Id)
+	if err != nil || t1 == nil {
```
hadjri, line 1323: the task being nil and finding an error looking it up from the DB are two separate error cases, so we should treat them as such with separate err messages, so it's best to separate them. Here's an [example](https://gist.github.com/hadjri/cc38c4ed5758423c54714adca823ce78#file-task_lifecycle-go-L130-L136)

model/task_lifecycle.go
```diff
@@ -563,9 +563,21 @@ func doLinearStepback(ctx context.Context, t *task.Task) error {
...
+// If there are no tasks to bisect, it no-ops.
 func doBisectStepback(ctx context.Context, t *task.Task) error {
+	// If we are a generated task, do stepback on our generator.
+	if t.GeneratedBy != "" {
+		generator, err := task.FindOneId(t.GeneratedBy)
+		if err != nil {
+			return errors.Wrapf(err, "finding generator with id '%s'", t.GeneratedBy)
+		}
+		if generator == nil {
+			return errors.Errorf("generator is nil with id '%s'", t.GeneratedBy)
```
hadjri, line 576: nit: we can just say `generator '%s' not found`

model/task_lifecycle.go
```diff
@@ -563,9 +563,21 @@ func doLinearStepback(ctx context.Context, t *task.Task) error {
...
-// doBisectStepback performs a bisect stepback on the task. If there are no tasks
-// to bisect, it returns nothing.
+// doBisectStepback performs a bisect stepback on the task.
+// If there are no tasks to bisect, it no-ops.
 func doBisectStepback(ctx context.Context, t *task.Task) error {
+	// If we are a generated task, do stepback on our generator.
+	if t.GeneratedBy != "" {
+		generator, err := task.FindOneId(t.GeneratedBy)
+		if err != nil {
+			return errors.Wrapf(err, "finding generator with id '%s'", t.GeneratedBy)
```
hadjri, line 573: nit: we can delete "with id"

model/task_lifecycle.go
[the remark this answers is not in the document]
hadjri: Discussed offline - but my comment was referring to the fact that `evalBisectStepback` needs to be kicked off only by an initial failure of a task, as it currently will run for every task which we don't want.

hadjri (changes requested):
I just noticed that `activatePreviousTask` checks for [bisection stepback fields](https://gist.github.com/hadjri/b82fb21931084e9d976805b996282715#file-task_lifecycle-go-L266-L270). I think this was introduced before we decided to remove stepbackDepth altogether, but those lines can be removed as  `activatePreviousTask` is exclusive to linear-stepback.

model/task_lifecycle.go
```diff
@@ -579,7 +591,74 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
...
+	} else if t.Status == evergreen.TaskFailed {
+		s.LastFailingStepbackTaskId = t.Id
+	} else {
+		return errors.Errorf("stopping task stepback due to status '%s'", t.Status)
+	}
+
+	// The midway task is our next stepback target.
+	nextTask, err := task.FindMidwayTaskFromIds(s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId)
+	if err != nil {
+		return errors.Wrapf(err, "finding midway task between tasks '%s' '%s'", s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId)
```
hadjri, line 626: nit: '%s' and '%s'

hadjri (changes requested):

model/task_lifecycle.go
```diff
@@ -1203,8 +1282,18 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
 
 // evalBisectStepback performs bisect stepback on the task.
 func evalBisectStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
```
hadjri, line 1284: We also will need to pass in `stepback` as a bool param to this function just as [evalLinearStepback does](https://github.com/evergreen-ci/evergreen/pull/7124/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R1243) to ensure that we're [respecting](https://github.com/evergreen-ci/evergreen/pull/7124/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R1251-R1253) if stepback is set to false for the project/BV/task in question.

model/task_lifecycle.go
```diff
@@ -1203,8 +1282,18 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
...
-	// TODO: EVG-20788 implement stepback bisection.
-	return errors.Wrap(doBisectStepback(ctx, t), "performing bisect stepback")
+	// If the task is aborted or it isn't a mainline commit or existing stepback, no-op.
+	if t.Aborted || (t.ActivatedBy != evergreen.RepotrackerVersionRequester && t.ActivatedBy != evergreen.StepbackTaskActivator) {
+		return nil
+	}
+
+	newStepback := t.StepbackInfo == nil && evergreen.IsFailedTaskStatus(t.Status)
```
hadjri, line 1290: passing in t.Status is fine, we should just remove the status param from this function now since it's unused. We still are gonna need caller and deactivatePrevious though so let's keep them for now (gonna create another ticket regarding this)

model/task_lifecycle.go
```diff
@@ -1203,8 +1282,18 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
 
 // evalBisectStepback performs bisect stepback on the task.
 func evalBisectStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
-	// TODO: EVG-20788 implement stepback bisection.
-	return errors.Wrap(doBisectStepback(ctx, t), "performing bisect stepback")
+	// If the task is aborted or it isn't a mainline commit or existing stepback, no-op.
+	if t.Aborted || (t.ActivatedBy != evergreen.RepotrackerVersionRequester && t.ActivatedBy != evergreen.StepbackTaskActivator) {
+		return nil
+	}
+
```
hadjri, line 1289: Can we add a brief comment explaining these two variables for future readability?

hadjri:
> I just noticed that `activatePreviousTask` checks for [bisection stepback fields](https://gist.github.com/hadjri/b82fb21931084e9d976805b996282715#file-task_lifecycle-go-L266-L270). I think this was introduced before we decided to remove stepbackDepth altogether, but those lines can be removed as `activatePreviousTask` is exclusive to linear-stepback.

I think this may have been missed

model/task_lifecycle.go
```diff
@@ -1203,8 +1282,18 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
 
 // evalBisectStepback performs bisect stepback on the task.
 func evalBisectStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
-	// TODO: EVG-20788 implement stepback bisection.
-	return errors.Wrap(doBisectStepback(ctx, t), "performing bisect stepback")
+	// If the task is aborted or it isn't a mainline commit or existing stepback, no-op.
+	if t.Aborted || (t.ActivatedBy != evergreen.RepotrackerVersionRequester && t.ActivatedBy != evergreen.StepbackTaskActivator) {
```
hadjri, line 1286: We actually already have a guarantee [here](https://gist.github.com/hadjri/ec0ec88da8a1e76c1b6b0fe4ecfc0bf2#file-task_lifecycle-go-L766) that the requester is not a patch so we can drop the check for `RepotrackerVersionRequester` (also because we do want to support `TriggerRequester` which is a different kind of requester that shows up on the mainline commit page)

So this can actually just be `if t.Aborted{ return nil }`

hadjri (changes requested):
Looking good! I expect this to be the last CR. Also there is a unit test that needs fixing

model/task_lifecycle.go
```diff
@@ -1281,13 +1274,16 @@ func evalLinearStepback(ctx context.Context, t *task.Task, caller, status string
 }
 
 // evalBisectStepback performs bisect stepback on the task.
-func evalBisectStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
+func evalBisectStepback(ctx context.Context, t *task.Task, caller string, stepback, deactivatePrevious bool) error {
 	// If the task is aborted or it isn't a mainline commit or existing stepback, no-op.
```
hadjri, line 1278: Now that we're passing the variable in, we are also going to need to return nil if stepback is not enabled, just [as linear stepback does](https://github.com/evergreen-ci/evergreen/pull/7124/files/4e7eaa3469be3dc5647347961051a496d4648684..50facde7df89b085ef2b29648fb1be3c503f51a3#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R1244-R1246). We should also update this comment

hadjri (changes requested):

model/task_lifecycle.go
ZackarySantana: oops.... totally missed adding that, my bad!

hadjri: np!

hadjri (approved):
LGTM once the build passes

PR #7125, opened 2023-10-09
EVG-19725: split IsGroup

Kimchelly:
EVG-19725

### Description
Evergreen project YAMLs allow you to list either tasks or task groups under the build variant task list. This causes a bit of confusion when dealing with the `model.Project` because it means the thing in the task list may be a task group or a task. There's an `IsGroup` field to indicate whether it's a task group or just a task. Unfortunately, `IsGroup` is also a little confusing because depending on the context, it can refer to a task group, or a task within a task group.

During project translation, Evergreen sets `IsGroup` in-memory to indicate that the item in the build variant list is a task group (this is the most common case for `IsGroup`). However, a couple of helper functions (e.g. `CreateTasksFromGroup`, `FindAllBuildVariantTasks`) explicitly break down task groups listed under BVs into their individual tasks and set a combination of `IsGroup` + `GroupName` to indicate that you're dealing _within_ a task in the task group, and not the task group as an aggregate construct. To reduce some confusion around determining "is this a task group or a task within a task group", I split these two meanings into separate fields. `IsGroup` means that it's a task group. `IsPartOfGroup` indicates that it's a task _within_ a task group, not the task group itself.

In order to reduce the risk of breaking things in case my understanding of `IsGroup`'s usage is somehow flawed, I added some logging to confirm that `IsGroup` and `IsPartOfGroup` are being used as expected in all places. There will be a follow-up PR to use `IsPartOfGroup` where it's relevant to replace `IsGroup` checks and remove the remaining TODOs/logging.

### Testing
* Added tests for `IsGroup` + `IsPartOfGroup` to project translation.
* Tested in staging patch that task groups still behaved the same.

### Documentation
N/A

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -22,7 +22,6 @@ import (
 	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/utility"
-	"github.com/k0kubun/pp"
 	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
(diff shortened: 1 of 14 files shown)


model/project.go
```diff
@@ -92,13 +93,19 @@ type BuildVariantTaskUnit struct {
...
-	// GroupName is the task group name if this is a task in a task group. If
-	// it is the task group itself, it is not populated (Name is the task group
-	// name).
+	// IsPartOfGroup indicates that this unit is a task within a task group. If
+	// this is set, then GroupName is also set. Note that project translation
+	// does not expand task groups into their individual tasks, so this is only
+	// set for special functions that explicitly expand task groups into
+	// individual task units (such as FindAllBuildVariantTasks).
+	IsPartOfGroup bool `yaml:"-" bson:"-"`
+	// GroupName is the task group name if this is a task in a task group. This
```
hadjri, line 105: If this field is only ever set if `IsPartOfGroup` is true, would it be possible to instead  remove the `IsPartOfGroup` field entirely and instead implement a function like:
```
func (bvt *BuildVariantTaskUnit) IsPartOfGroup() bool {
	return bvt.GroupName != ""
}
``` 
instead?

Kimchelly: This would also work, but I considered that we could (legitimately) change the behavior of `GroupName` one day so that for convenience, it's set for both when it's part of a task group or when it's the task group itself, which would break the distinction of task group vs. part of a task group.

I did also consider just not setting `IsGroup` + `GroupName` at all if it's part of a group (a task unit being part of a group is only relevant in [this one spot](https://github.com/evergreen-ci/evergreen/blob/main/model/lifecycle.go#L1242-L1246)), but it could make it difficult to use `CreateTasksFromGroup` and `FindAllBuildVariantTasks` in the future if we can't distinguish a regular task from one in a task group.

hadjri: Sounds good, makes sense if you want the `GroupName` to potentially take on a new meaning in the future. As for removing the `IsGroup` + `GroupName` entirely, the last solution you brought up sounds the most elegant, but I can see how it would be a bigger refactor as you said. I'm fine with how this works in this PR, and I'll leave it to you if you want to make a followup ticket to remove the fields.

hadjri (approved):

PR #7126, opened 2023-10-10
EVG-19064 Permit unbounded max_hosts for task groups

bynn:
EVG-19064

### Description
max_hosts for task groups will be set to the number of the tasks the task group has if set to -1 
it is not 0 like the ticket requests because that would change our default behavior that some teams could be relying on 

### Testing
unit test

created this patch on staging using max hosts = -1
https://spruce-staging.corp.mongodb.com/version/652573c29dbe3269b855e6be/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

and it was set to 2, the length of the tasks
https://evergreen-staging.corp.mongodb.com/rest/v2/tasks/sandbox_ubuntu2004_bynntask1_patch_d631380317a8478b879cc27d6171508dea2b11df_652573c29dbe3269b855e6be_23_10_10_15_54_53

### Documentation
updated doc


docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1398,7 +1398,8 @@ Parameters:
     commands but will not cause the task to fail unless
     `teardown_task_can_fail_task` is true.
 -   `max_hosts`: number of hosts across which to distribute the tasks in
-    this group. This defaults to 1. There will be a validation warning
+    this group. This defaults to 1. If set to -1, it will be updated to the 
+    number of tasks in this task group. There will be a validation warning
     if max hosts is less than 1 or greater than the number of tasks in
```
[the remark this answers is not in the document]
bynn: updated doc

bynn:
I added the patch and task created in staging to the description 
I want to not change the validation at that point because that validation happens after we parse the config so a negative value at this point in the code will be invalid. Do you think that makes sense or should we still add the special case to the validation?

Kimchelly (commented):
1. While negative numbers after project translation are invalid, it's a little weird to validate that we don't introduce a bug, since it's not checking for valid user input but rather for our own buggy code. Supposedly, if project translation stopped turning -1 into a positive value, it would be caught by the test for `max_hosts` of -1, making the validation redundant.
2. I only thought about it just now, but you can also define task groups inline in the build variant task definition [here](https://github.com/evergreen-ci/evergreen/blob/47d96dc924aa2c9bff1c69e378cdf2bbe10536f6/model/project_parser.go#L1399). For example:
```
buildvariants:
- name: bv
  tasks:
    - name: some-task-group
      task_group:
        # Redefine max_hosts in this one instance to overwrite the task group's max_hosts
        max_hosts: -1
```
Can we add a test for this case?

bynn (commented):
in line bv task groups was never defaulting to host of 1 for non-positive values so I added that and the new special case and a test for it 

updated validation and also tested it

docs/Project-Configuration/Project-Configuration-Files.md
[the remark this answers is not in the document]
Kimchelly: No worries, looks good now! 😎

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1398,10 +1398,11 @@ Parameters:
     commands but will not cause the task to fail unless
     `teardown_task_can_fail_task` is true.
 -   `max_hosts`: number of hosts across which to distribute the tasks in
-    this group. This defaults to 1. There will be a validation warning
-    if max hosts is less than 1 or greater than the number of tasks in
-    task group. When max hosts is 1, this is a special case where the tasks
-    will run serially on a single host. If any task fails, the task group
+    this group. This defaults to 1. If set to -1, it will be updated to the 
+    number of tasks in this task group. There will be a validation warning
+    if max hosts is less than 1 (apart from -1 )or greater than the number of
```
Kimchelly, line 1403: nit: ")or" -> ") or"

model/project_parser_test.go
```diff
@@ -1411,6 +1411,43 @@ buildvariants:
...
+  - name: inline_task_group
+    task_group:
+      max_hosts: -1
+      tasks:
+      - example_task_1
+      - example_task_2
+`
+		proj := &Project{}
+		_, err := LoadProjectInto(ctx, []byte(validMaxHostYml), nil, "id", proj)
+		assert.NotNil(t, proj)
```
Kimchelly, line 1440: From [the style guide](https://docs.google.com/document/d/1-DB1FQ_zb5fX_0pnVcwyU0HmJPrloFBu6qSE8dkk8PU/edit#heading=h.9946u3sq3e2n): `require.NotNil` instead of assert because if any logic that this test relies upon changes and makes the assertion wrong, `proj` will be nil, the later dereferences of `proj` fields will panic the entire Go runtime and stop all tests.

For example, this could happen if someone in the future changes the logic of `LoadProjectInto` or project translation that makes this error.

model/project_parser_test.go
```diff
@@ -1411,6 +1411,43 @@ buildvariants:
...
+		assert.Nil(t, err)
+		assert.Len(t, proj.TaskGroups, 1)
```
Kimchelly, line 1442: Same as above `require.Len` or else the line below could panic.

model/project_parser_test.go
```diff
@@ -1411,6 +1411,43 @@ buildvariants:
...
+		assert.Equal(t, "example_task_group", proj.TaskGroups[0].Name)
+		assert.Equal(t, proj.TaskGroups[0].MaxHosts, len(proj.TaskGroups[0].Tasks))
+
+		assert.Len(t, proj.BuildVariants, 1)
+		assert.Len(t, proj.BuildVariants[0].Tasks, 2)
```
Kimchelly, line 1447: Same for these two lines - `require.Len`

validator/project_validator_test.go
```diff
@@ -3278,8 +3280,8 @@ buildvariants:
 	assert.Contains(validationErrs[0].Message, "attach.results cannot be used in the group teardown stage")
 	validationErrs = checkTaskGroups(&proj)
 	assert.Len(validationErrs, 2)
```
Kimchelly, line 3282: I know you didn't write this, but `require.Len` here for panic safety.

validator/project_validator_test.go
```diff
@@ -3217,7 +3217,9 @@ func TestTaskGroupValidation(t *testing.T) {
 			},
 		},
 	}
-	assert.Len(checkTaskGroups(&proj), 3)
+	validationErrs = checkTaskGroups(&proj)
+	assert.Len(validationErrs, 1)
```
Kimchelly, line 3221: `require.Len`

model/project_parser_test.go
```diff
@@ -1411,6 +1411,43 @@ buildvariants:
...
+- name: bv
+  display_name: "bv_display"
+  tasks:
+  - name: example_task_group
+  - name: inline_task_group
+    task_group:
+      max_hosts: -1
+      tasks:
+      - example_task_1
+      - example_task_2
```
Kimchelly, line 1436: Would be slightly preferable to test with a different task group definition than the one for `example_task_group` so we know that `max_hosts` is set based on the tasks define inline rather than the `example_task_group` tasks.

Kimchelly (approved):
LGTM mod small nits/suggestions.

PR #7127, opened 2023-10-10
EVG-20987 Add push trigger sha to version doc

hadjri:
EVG-20987

### Description
This PR is a required precursor to [this Spruce PR](https://github.com/evergreen-ci/spruce/pull/2098). 

Spruce's construction of links pointing to upstream triggers is dependent on the [UpstreamProject](https://gist.github.com/hadjri/a00d402d3f674bf0301cfe0752b71a48#file-version_resolver-go-L374) resolver, which requires both `TriggerType` and `TriggerID` to be populated for the downstream version. For downstream versions triggered by builds / tasks, this TriggerID is set to the upstream task ID or upstream version ID respectively, but for push level triggers there is no Spruce page that can be associated with a github commit so prior to this change `TriggerID` was empty.

I've made a change to assign the upstream project ID as the `TriggerID` when creating versions from push triggers, and also added a new `TriggerSHA` field for the version, which will be set when it is a downstream version triggered from a push, to allow Spruce to later construct an external GitHub link to the commit push that started the trigger. 
### Testing
Confirmed that the `UpstreamProject` resolver used to populate links from triggered versions to the source of their trigger is now populated for push trigger versions, where it was previously returned as null (see screenshots)

Before change (id: `sandbox_d631380317a8478b879cc27d6171508dea2b11df_c6db985b8c9ffb231c10085b09efa348`):
image

After change (id: `sandbox_2ab3e8f46ad29e30432bbe30c4da723d879c3044_c6db985b8c9ffb231c10085b09efa348`): 
image

### DIFF
--- a/model/version.go
+++ b/model/version.go
@@ -79,10 +79,14 @@ type Version struct {
 
 	SatisfiedTriggers []string `bson:"satisfied_triggers,omitempty" json:"satisfied_triggers,omitempty"`
 	// Fields set if triggered by an upstream build
+	// TriggerID is the ID of the entity that triggered the downstream version. Depending on the trigger type, this
+	// could be a build ID, a task ID, or a project ID, for build, task, and push triggers respectively.
 	TriggerID    string `bson:"trigger_id,omitempty" json:"trigger_id,omitempty"`
 	TriggerType  string `bson:"trigger_type,omitempty" json:"trigger_type,omitempty"`
 	TriggerEvent string `bson:"trigger_event,omitempty" json:"trigger_event,omitempty"`
-	TriggerSHA   string `bson:"trigger_sha,omitempty" json:"trigger_sha,omitempty"`
+	// TriggerSHA is the SHA of the untracked commit that triggered the downstream version,
+	// this field is only populated for push level triggers.
+	TriggerSHA string `bson:"trigger_sha,omitempty" json:"trigger_sha,omitempty"`
 
 	// this is only used for aggregations, and is not stored in the DB
 	Builds []build.Build `bson:"build_variants,omitempty" json:"build_variants,omitempty"`
--- a/model/version.go
+++ b/model/version.go
@@ -82,6 +82,7 @@ type Version struct {
 	TriggerID    string `bson:"trigger_id,omitempty" json:"trigger_id,omitempty"`
 	TriggerType  string `bson:"trigger_type,omitempty" json:"trigger_type,omitempty"`
 	TriggerEvent string `bson:"trigger_event,omitempty" json:"trigger_event,omitempty"`
+	TriggerSHA   string `bson:"trigger_sha,omitempty" json:"trigger_sha,omitempty"`
 
 	// this is only used for aggregations, and is not stored in the DB
 	Builds []build.Build `bson:"build_variants,omitempty" json:"build_variants,omitempty"`
--- a/trigger/process.go
+++ b/trigger/process.go
@@ -97,7 +97,6 @@ type ProcessorArgs struct {
 	EventID                      string
 	DefinitionID                 string
 	Alias                        string
-	ProjectID                    string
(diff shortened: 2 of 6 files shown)


trigger/process.go
```diff
@@ -314,12 +313,12 @@ func TriggerDownstreamProjectsForPush(ctx context.Context, projectId string, eve
...
 				ConfigFile:                   trigger.ConfigFile,
 				TriggerType:                  model.ProjectTriggerLevelPush,
+				TriggerID:                    trigger.Project,
 				DefinitionID:                 trigger.DefinitionID,
 				Alias:                        trigger.Alias,
-				ProjectID:                    trigger.Project,
 				UnscheduleDownstreamVersions: trigger.UnscheduleDownstreamVersions,
 				PushRevision: model.Revision{
-					Revision:        utility.FromStringPtr(event.GetHeadCommit().SHA),
+					Revision:        utility.FromStringPtr(event.GetHeadCommit().ID),
```
ablack12, line 321: Why this change?

model/version.go
```diff
@@ -82,6 +82,7 @@ type Version struct {
 	TriggerID    string `bson:"trigger_id,omitempty" json:"trigger_id,omitempty"`
 	TriggerType  string `bson:"trigger_type,omitempty" json:"trigger_type,omitempty"`
 	TriggerEvent string `bson:"trigger_event,omitempty" json:"trigger_event,omitempty"`
+	TriggerSHA   string `bson:"trigger_sha,omitempty" json:"trigger_sha,omitempty"`
```
ablack12, line 85: Since this is only used for the case where it's _not_ triggered by an upstream build but rather an untracked commit, I think it'd be helpful to include this.

Also in the other case, is TriggerID used to store the versionID as opposed to the project ID? Might be helpful also to clarify that this field could have one of those two values then. I'm also a little hesitant about this though -- it seems like an ID field should be unique.

trigger/project_triggers.go
```diff
@@ -51,11 +51,14 @@ func TriggerDownstreamVersion(ctx context.Context, args ProcessorArgs) (*model.V
...
 		return nil, errors.Wrap(err, "getting Evergreen settings")
 	}
-	var versionID string
 	// Since push triggers have no source version (unlike build and task level triggers), we need to
-	// extract the project ID from the trigger definition's project ID, which is populated in the ProjectID field.
-	projectID := args.ProjectID
-	if projectID == "" {
+	// extract the project ID from the trigger definition's project ID, which is populated in the TriggerID field
+	// for push triggers.
+	var versionID string
```
ablack12, line 57: var versionID, projectID string

ablack12 (changes requested):
I think it'd be helpful before approving this PR to see the Spruce screenshot that this is the correct information needed for the UI.

model/version.go
hadjri: Added clarifying comments. Can you clarify what you mean by an ID field should be unique? They are all referring to unique IDs of other documents. Do you mean the naming of this field is unideal? We can name it `TriggeringEntity` or something similar

trigger/process.go
hadjri: It is the correct field to reference, `SHA` is not guaranteed to be defined: [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push)

hadjri:
Re-requesting review with code changes for the time being - will update Spruce screenshots once I am unblocked by the VPN 😅

hadjri:
image

image

Confirmed both link [here](https://github.com/evergreen-ci/commit-queue-sandbox/commit/2ab3e8f46ad29e30432bbe30c4da723d879c3044) as expected

model/version.go
ablack12: Ah i actually misunderstood what triggerID was for, this works for me.

ablack12 (approved):

PR #7128, opened 2023-10-10
docs: Add FAQ on variants in the merge queue

bsamek:


docs/Project-Configuration/Merge-Queue.md
```diff
@@ -73,3 +73,21 @@ starting point, and link to Evergreen builds from there.
...
+add the PR to the merge queue) and the merge queue itself (which determines
+whether or not GitHub will merge the PR).  Nevertheless, you can run different
+tasks in PRs and the merge queue with the separate settings on the GitHub &
+Commit Queue project settings page.
+
+Evergreen will post a status check called "evergreen" when the entire merge
+queue build has finished, and it will post a status for each variant that
+finishes called "evergreen/variant_name". Typically projects should set a branch
+protection rule for "evergreen". However, it's also possible to instead to set a
+branch protection rule for one or more "evergree/variant_name" statuses. You
```
Kimchelly, line 89: spelling: `"evergreen/variant_name"`

docs/Project-Configuration/Merge-Queue.md
```diff
@@ -73,3 +73,21 @@ starting point, and link to Evergreen builds from there.
...
+
+**Q:** How do branch protection rules apply to PRs and the merge queue?
+
+The same branch protection rules apply to PRs (which determine whether you can
+add the PR to the merge queue) and the merge queue itself (which determines
+whether or not GitHub will merge the PR).  Nevertheless, you can run different
+tasks in PRs and the merge queue with the separate settings on the GitHub &
+Commit Queue project settings page.
+
+Evergreen will post a status check called "evergreen" when the entire merge
```
Kimchelly, line 85: The info in this paragraph seems like it's relevant to both the merge queue workflow and just using GitHub PR patches (without a merge queue). Since we always send the same kind of version + individual bv statuses (whether it's for PRs, GitHub checks, or the merge queue), should we document this in a shared place in the docs to better advertise that GitHub patches all have the same status-posting behavior?

bsamek: For Evergreen's commit queue, whether through the CLI or GitHub PR comment, Evergreen will merge the change if all tasks pass, and reject it otherwise. Branch protection rules are orthogonal (though they could as a side effect prevent a merge). In the GitHub merge queue case, however, a user _must_ define branch protection rules to turn on the merge queue. I think for that reason that this is a pain point for GitHub's merge queue—it was something that 10gen/mms had to consider when they migrated to it—but it's not a pain point for Evergreen's commit queue.

Kimchelly: Hmm the thing I was originally referring to wasn't how the statuses posted by Evergreen to GitHub interact with branch protections, but rather how the statuses are displayed themselves. The [GitHub PR testing docs already discuss](https://github.com/evergreen-ci/evergreen/blob/8d47499234f2a5e237d737ed562d500b3faa01c9/docs/Project-Configuration/Project-and-Distro-Settings.md?plain=1#L135-L136) how the statuses are posted, and those statuses are the same for all types of GitHub patches (for PRs, merge queue, and GitHub checks), so I was advocating for de-duplicating the info about "how does Evergreen show its status in GitHub" into a common place and just linking to it.

Anyways, I don't think this needs to block merging, it's just a thought that having too much duplicated info in the docs rather than linking to common sources could lead to problems keeping documentation consistent/correct in the future.

Kimchelly (approved):

docs/Project-Configuration/Merge-Queue.md
bsamek: Although I agree, I think in this case the duplication makes sense because it's addressing a specific pain point. I think your point is in general correct though, and appreciate the criticism!

PR #7130, opened 2023-10-11
EVG-17662: support logs for local evergreen

julianedwards:
EVG-17662

### Description
- Support GridFS bucket type for logs
- Remove `load-smoke-data` command 
- Create new local and smoke data via `mongodump`
- Use `mongorestore` to load local and smoke data

This change requires that Evergreen developers have installed mongo tools locally. To run smoke tests or evergreen local on an Evergreen host, the makefile will take care of downloading mongo tools before running `mongorestore`.

Most of the files changed here are a result of moving the test data around via `mongodump`.

### Testing
Smoke tests must pass in Evergreen and tested that `make load-local-data` and `make load-smoke-data` worked locally. This will also require adding a new field with the mongo tools download link to the Evergreen configs in Spruce and Parsley. 

### Documentation
Will need to update the new hire and developer guides.


makefile
```diff
@@ -162,19 +160,19 @@ set-smoke-git-config:
 	git config user.email email
 load-smoke-data:$(buildDir)/.load-smoke-data
 load-local-data:$(buildDir)/.load-local-data
-$(buildDir)/.load-smoke-data:$(buildDir)/load-smoke-data
-	./$<
+$(buildDir)/.load-smoke-data:$(buildDir)/.get-mongotools smoke/internal/testdata/mongodump
+	$(call run-mongotools,mongorestore) --drop smoke/internal/testdata/mongodump
```
Kimchelly, line 164: nit: I know how this works, but it would be helpful for other maintainers who aren't as familiar with Make to explain that this works by invoking the `run-mongotools` target with the `mongorestore` argument.

makefile
```diff
@@ -387,6 +385,15 @@ scramble:
 	python cmd/scrambled-eggs/scramble.py $(file) $(multiarg)
 
 # mongodb utility targets
+mongotoolsDir := mongotools
+run-mongotools = $(if $(wildcard $(mongotoolsDir)), ./$(mongotoolsDir)/$(1), $(1))
```
Kimchelly, line 389: `:=` is preferred in Make unless you have a specific need for `=` because `=` can lead to infinite recursion.

Kimchelly, line 389: Same thing as above - would help to explain in words what this is doing (using the locally-downloaded mongo-tools if possible, and otherwise using the globally-installed relevant mongo-tool executable?

makefile
```diff
@@ -387,6 +385,15 @@ scramble:
...
+$(buildDir)/.get-mongotools:
+ifeq (,$(shell which mongorestore))
```
Kimchelly, line 391: Is this meant to be used by regular devs, or is it only for the CI testing?

If we're going to have regular devs use it to get mongo-tools, instead of doing conditional `if` logic in Make, it's more normal to just make `.get-mongotools` have a dependency on the mongotools directory existing. Otherwise, if this is only used for CI testing, I would just follow the same model as `get-mongodb`/`get-mongosh` for consistency of our CI tests.

self-tests.yml
```diff
@@ -227,6 +227,8 @@ functions:
         OTEL_COLLECTOR_ENDPOINT: ${otel_collector_endpoint}
         OTEL_TRACE_ID: ${otel_trace_id}
         OTEL_PARENT_ID: ${otel_parent_id}
+        MONGOTOOLS_URL: ${mongotools_url}
+        MONGOTOOLS_DECOMPRESS: ${mongotools_decompress|*decompress}
```
Kimchelly, line 231: It seems like `mongotools_decompress` is never set. Do we need a separate decompress env var just for mongo-tools if it's always just `decmopress`?

cmd/load-smoke-data/load-smoke-data.go
```diff
@@ -1,269 +0,0 @@
-package main
```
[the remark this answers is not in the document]
julianedwards: Supporting GridFS was definitely the catalyst for the change to use use mongo-tools but I also feel that it is much cleaner and worth the change:

1. I have always found this code to be spaghetti-ish and confusing, it is one less thing to maintain and simplifies smoke IMO.
2. Manually editing JSON files where each document is a single line and there may be 100s of documents is tedious and error-prone. Whereas updating the data via the Mongo shell and using `mongodump` is a much less painful process for devs (no need to ever touch the test data files manually). I always find it easier to deal with this data via the DB and load it into JSON using `mongoexport` (I believe that is how all of the smoke and local data was created in the first place).
3. We already store the data as extended JSON which has fields such as `$date` and `$numberInt`.
4. This data is not touched very often, merge conflicts should be rare, if at all. In the rare cases of a merge conflict devs can coordinate their changes (also probably means they are working on a similar part of the codebase anyways).
5. Finally, I did note in the description that we would need to update our internal guides to explain this new workflow. I don't think that should be a big deal, though.

If you feel strongly about keeping JSON files for PR/CR purposes, I can look into using `mongoexport` and `mongorestore` instead.

makefile
julianedwards: If I use `:=` it won't expand the arg param `$(1)`.

makefile
```diff
@@ -387,6 +385,15 @@ scramble:
 	python cmd/scrambled-eggs/scramble.py $(file) $(multiarg)
 
 # mongodb utility targets
+mongotoolsDir := mongotools
```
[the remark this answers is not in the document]
julianedwards: Devs are expected to have it downloaded, this is only for CI testing. I can add comments.

self-tests.yml
julianedwards: Some are zip other tar. Right now we are only downloading tar but if we decide to download for other variants we may need to unzip instead. I can remove it and we can always add it back in if needed?

cmd/load-smoke-data/load-smoke-data.go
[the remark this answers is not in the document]
Kimchelly: That seems reasonable, the change SGTM as long as the dev docs are updated + comms to the team on how to follow the new workflow for loading smoke/local data.

go.mod
```diff
@@ -165,7 +165,7 @@ require (
 	github.com/mattn/go-xmpp v0.0.0-20211029151415-912ba614897a // indirect
 	github.com/mholt/archiver/v3 v3.5.1
 	github.com/mongodb/ftdc v0.0.0-20220401165013-13e4af55e809 // indirect
-	github.com/montanaflynn/stats v0.0.0-20180911141734-db72e6cae808 // indirect
```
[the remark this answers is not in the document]
Kimchelly: I still have this question. I'm not opposed to bumping them, but they seem kind of random to bump.

Kimchelly (commented):
Did you push your changes? The PR looks the same.

khelif96 (commented):
Do the changes here make it more difficult to manually seed test data? The UI team often manually generates test data to be used in UI tests but since this now uses mongodump what is the recommended approach to change data in these collections? It also seems like the test data code changes will no longer be reviewable in PRs since they are binaries. 

cc @SupaJoon  since I think your test changes also make use of mongodump but on the e2e test side of things.

julianedwards:
> Do the changes here make it more difficult to manually seed test data? The UI team often manually generates test data to be used in UI tests but since this now uses mongodump what is the recommended approach to change data in these collections? It also seems like the test data code changes will no longer be reviewable in PRs since they are binaries.

I think it makes it easier to manually seed test data because you can do it via `mongosh` instead of editing large JSON files where each document is a single line. You can create new documents and the run `mongodump` and everything will just work. 

As for test data, any new indexes created for a collection will be part of the JSON metadata and still be reviewable. Changes to documents in a collection will not be as reviewable (you can always pull the branch and load the data locally to inspect it), but the E2E tests should take care of validating data. With that being said, the local data should be a representation of real data and it may be better to have a separate dump for the UI tests if you are doing lots of manual updates and adding dummy data.

go.mod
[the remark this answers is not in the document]
julianedwards: Going to fix the upgrades once https://github.com/mongodb/anser/pull/91 goes through!

makefile
julianedwards: Done.

makefile
julianedwards: Done.

makefile
julianedwards: Done.

go.mod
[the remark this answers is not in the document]
julianedwards: I think I killed the upgrade half way through the last time, it looks like this got bumped when I upgraded the go driver, amboy, and anser.

makefile
```diff
@@ -297,8 +299,8 @@ lint-%:$(buildDir)/output.%.lint
 
 
 # start test and coverage artifacts
-#    This varable includes everything that the tests actually need to
-#    run. (The "build" target is intentional and makes these targetsb
+#    This variable includes everything that the tests actually need to
+#    run. (The "build" target is intentional and makes these targets
```
Kimchelly, line 303: Spelling 👌 (no change)

makefile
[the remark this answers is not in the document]
Kimchelly: Comments look good. I assume the updated dev docs will also include instructions on how to download mongo-tools?

self-tests.yml
Kimchelly: Makes sense. I'm alright with just leaving it as-is since it doesn't really add any complexity to have a separate env var.

Kimchelly (approved):
LGTM assuming dev docs will be updated + comms for the necessary dev workflow changes will go out.

julianedwards:
> LGTM assuming dev docs will be updated + comms for the necessary dev workflow changes will go out.

Yep! I will also provide a link to the downloads page—for most people it will just be brew installing mongo tools.

julianedwards:
Closing in favor of https://github.com/evergreen-ci/evergreen/pull/7171

PR #7134, opened 2023-10-12
docs: SSH into older hosts from macOS

bsamek:


michael-wb:
Maybe you could add a new "Troubleshooting" section with the following recommendations (something like):
If you are having difficulty connecting to the Spawn Host via SSH:
* Make sure you are connected to the MongoDB network or VPN
* Verify the host is in the "RUNNING" state and the DNS host name for the spawn host is correct.
* Verify you are using the correct user. This can be found on the "My Hosts" page under the Spawn Host detail drop down.
* Verify you are using the correct ssh key used when creating the Spawn Host. Use the `-i` argument to specify the location of your local private ssh key to use.
* <the new section about Mac OS hosts>

bsamek:
@michael-wb Thanks! Edited.

PR #7137, opened 2023-10-13
EVG-19725: do not set IsGroup for a task in a task group

Kimchelly:
EVG-19725

### Description
Follow-up from #7133 and #7125 to complete the transition away from `IsGroup`'s double meaning by splitting into `IsGroup` and `IsPartOfGroup`.

* Only set `IsGroup` if it's representing a task group. Use `IsPartOfGroup` when checking a task in a task group.
* Remove temporary logs.
* Update doc comments to clarify when a task unit can represent a task group vs. a task in a task group.

### Testing
* Monitored prod for usages, they all seemed to align with my understanding.
* Updated tests.

### Documentation
N/A

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -93,18 +93,23 @@ type BuildVariantTaskUnit struct {
 	// the project level, or an error will be thrown
 	Name string `yaml:"name,omitempty" bson:"name"`
 	// IsGroup indicates that it is a task group. This is always populated for
-	// task groups after translating the parser project to the project.
+	// task groups after project translation.
 	IsGroup bool `yaml:"-" bson:"-"`
 	// IsPartOfGroup indicates that this unit is a task within a task group. If
-	// this is set, then GroupName is also set. Note that project translation
-	// does not expand task groups into their individual tasks, so this is only
-	// set for special functions that explicitly expand task groups into
-	// individual task units (such as FindAllBuildVariantTasks).
+	// this is set, then GroupName is also set.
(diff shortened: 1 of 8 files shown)


model/project.go
```diff
@@ -94,18 +93,23 @@ type BuildVariantTaskUnit struct {
...
 	// IsGroup indicates that it is a task group. This is always populated for
-	// task groups after translating the parser project to the project.
+	// task groups after project translation.
 	IsGroup bool `yaml:"-" bson:"-"`
 	// IsPartOfGroup indicates that this unit is a task within a task group. If
-	// this is set, then GroupName is also set. Note that project translation
-	// does not expand task groups into their individual tasks, so this is only
-	// set for special functions that explicitly expand task groups into
-	// individual task units (such as FindAllBuildVariantTasks).
+	// this is set, then GroupName is also set.
```
ablack12, line 99: If this is only true for IsPartOfGroup and GroupName, it might be reasonable to move these to the bottom and just say the comment once for both fields, but whichever works.

ablack12 (approved):

model/project.go
Kimchelly: Both are part of the same special functions, but I didn't consolidate because Go code inspection tools (like the ones to automatically show docs for a field in IDEs) are only smart enough to display the one comment directly above the field. Duplicating the note makes sure that readers can't miss the special-ness comment by only reading the doc for one field + makes it obvious that the special case is true for specifically just these two fields.

PR #7144, opened 2023-10-16
EVG-20853 Allow project variables to expand modules

hadjri:
EVG-20853

### Description
Spoke with user async and confirmed that for their use case, it actually would be sufficient to just expand their module branch names from manually changing the project variables side in the UI, and programmatically setting it isn't a hard requirement.

Updated modules to accept expansions and added a change to expand modules with project vars when the manifest is created.
### Testing
Tested in staging, and confirmed that with `${var1}` set to `test-smoke`, a version correctly [creates a manifest](https://evergreen-staging.corp.mongodb.com/version/sandbox_b92cc103abfbef75b1529fde02952b0b48570e99) with the expanded branch name. Updated unit test
### Documentation
Will update upon approval


model/project.go
```diff
@@ -437,11 +437,11 @@ type ContainerSystem struct {
 }
 
 type Module struct {
-	Name       string `yaml:"name,omitempty" bson:"name"`
-	Branch     string `yaml:"branch,omitempty" bson:"branch"`
-	Repo       string `yaml:"repo,omitempty" bson:"repo"`
-	Prefix     string `yaml:"prefix,omitempty" bson:"prefix"`
-	Ref        string `yaml:"ref,omitempty" bson:"ref"`
+	Name       string `yaml:"name,omitempty" bson:"name" plugin:"expand"`
```
ablack12, line 440: We need plugin here for expansions, right? Since this isn't something we typically expand, a comment would be helpful.

model/version.go
```diff
@@ -717,8 +718,18 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 		IsBase:      v.Requester == evergreen.RepotrackerVersionRequester,
 	}
 
+	projVars, err := FindMergedProjectVars(projectRef.Id)
+	if err != nil {
+		return nil, errors.Wrapf(err, "getting project vars for project '%s'", projectRef.Id)
+	}
+	expansions := util.NewExpansions(projVars.Vars)
+	for i := range moduleList {
+		if err = errors.Wrap(util.ExpandValues(&moduleList[i], expansions), "could not expand module"); err != nil {
```
ablack12, line 727: We don't need to wrap this error twice

ablack12 (commented):
Changes LGTM but holding final approval for docs! Thanks for checking in with the user on this.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -407,12 +407,16 @@ the git revision for modules by setting a module manually with
 by specifying the `auto_update` option (as described below) to use the
 latest revision available for a module.
 
+Module fields support the expansion of variables defined in the [Variables](Project-and-Distro-Settings.md#variables)
```
ablack12, line 410: Can we add that all fields (except auto_update) can be expanded (maybe under the Fields section) and also indicate that these are expanded at version creation time?

ablack12 (approved):
LGTM with doc update

PR #7146, opened 2023-10-16
EVG-21083 Update git.get_project to create GitHub app token

bynn:
EVG-21083

### Description
Instead of reading in a github_app_token, the command will call an agent route to create a github token to clone 

### Testing
updated unit tests
tested that git.get_project does not break on staging 
https://spruce-staging.corp.mongodb.com/task/sandbox_git_env_test_validate_commit_message3_patch_d631380317a8478b879cc27d6171508dea2b11df_652ef5e497b1d3c100812ee7_23_10_17_21_00_27/logs?execution=0


agent/command/git.go
```diff
@@ -137,19 +137,34 @@ func (opts *cloneOpts) setLocation() error {
...
+	// TODO EVG-21022: Remove fallback once we delete GitHub tokens as expansions.
+	if err != nil {
+		grip.Warning(message.WrapError(err, message.Fields{
+			"message": "error creating GitHub app token, falling back to legacy clone methods",
+			"owner":   owner,
+			"repo":    repo,
+			"task":    td.ID,
+			"ticket":  "EVG-21022",
+		}))
+	} else if appToken == "" {
```
Kimchelly, line 156: nit: might read more easily if-elseif was simplified down so the debug didn't need an explicit case:
```go
appToken, err := comm.CreateInstallationToken(...)
grip.Warning(...)
if appToken != "" {
    return evergreen.CloneMethodAccessToken, appToken, nil
}

grip.Debug(...)

// Continue with fallback
```

agent/command/git.go
```diff
@@ -512,7 +527,9 @@ func (c *gitFetchProject) Execute(ctx context.Context, comm client.Communicator,
 		return errors.Wrap(err, "applying expansions")
 	}
 
-	projectMethod, projectToken, err := getProjectMethodAndToken(c.Token, conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion), conf.Expansions.Get(evergreen.GithubAppToken), conf.GetCloneMethod())
+	td := client.TaskData{ID: conf.Task.Id, Secret: conf.Task.Secret}
+
+	projectMethod, projectToken, err := getProjectMethodAndToken(ctx, comm, td, conf.ProjectRef.Owner, conf.ProjectRef.Repo, c.Token, conf.Expansions.Get(evergreen.GlobalGitHubTokenExpansion), conf.GetCloneMethod())
```
ablack12, line 532: this is a _lot_ of variables; can you just make this a private method of c and pass in all of conf to avoid many strings in a row?

config.go
```diff
@@ -35,7 +35,7 @@ var (
 	ClientVersion = "2023-09-06"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-10-09"
+	AgentVersion = "2023-17-09"
```
ablack12, line 38: so european :)

agent/command/git.go
ablack12, line 156: we can just do `else if appToken != "" { return evergreen.CloneMethodAccessToken, appToken, nil }` and then move the debug down to live with the legacy clone method.

ablack12 (changes requested):

agent/internal/client/mock.go
```diff
@@ -528,5 +533,8 @@ func (c *Mock) GetPullRequestInfo(ctx context.Context, taskData TaskData, prNum
 }
 
 func (c *Mock) CreateInstallationToken(ctx context.Context, td TaskData, owner, repo string) (string, error) {
-	return "token", nil
+	if owner != "" && repo != "" && repo != NotValidRepo {
```
Kimchelly, line 536: Rather than having the mock itself hard-code the result and having to use the hard-coded constant in the test, mocks usually do the opposite - the test that depends on the mock pre-defines a value for the mock to echoed back to it. The dependency injection lets the caller define what the mock is supposed to do, rather than have to write the test to conform to what the mock does, which gives more flexibility in your tests than if the mock decides the output.

[This](https://github.com/evergreen-ci/evergreen/blob/ff46048d63feb595caedb68fd4d9fc85ab73a96b/agent/internal/client/mock.go#L384-L390) is one example of that pattern being used. The test can set `(MockCommunicator).GetTaskPatchResponse` and whatever testing code depends on the patch will use the test's desired patch.

config.go
bynn: baguette

agent/internal/client/mock.go
bynn: updated

agent/command/git.go
bynn: couldnt make it a method on c because both git fetch and push uses it but passes in conf now

agent/internal/client/mock.go
```diff
@@ -28,6 +28,11 @@ import (
 	"google.golang.org/grpc"
 )
 
+const (
+	MockedGitHubAppToken = "mocked_github_app_token"
```
Kimchelly, line 32: Move these over to `agent/command/git_test.go` and don't export the fields since these constants are used only for that one test suite.

agent/internal/client/mock.go
```diff
@@ -528,5 +534,8 @@ func (c *Mock) GetPullRequestInfo(ctx context.Context, taskData TaskData, prNum
 }
 
 func (c *Mock) CreateInstallationToken(ctx context.Context, td TaskData, owner, repo string) (string, error) {
-	return "token", nil
+	if c.CreateInstallationTokenFail {
+		return "", nil
+	}
+	return MockedGitHubAppToken, nil
```
Kimchelly, line 540: For consistency with how the mock is usually implemented, a more common/testing-friendly mock would be to have a field `c.CreateInstallationToken` that the test can set that as the returned token. Like this:
```go
// In agent/command/git_test.go
func (s *GitGetProjectSuite) SetupTest() {
    s.comm.CreateInstallationTokenResult = "mocked_github_app_token"
}

// Tests then will always get that token back (unless you set CreateInstallationTokenFail)

// In the mock
func (c *Mock) CreateInstallationToken(...) {
    if c.CreateInstallationTokenFail { /* return error */ }
    return c.CreateInstallationTokenResult, nil
}
```

It would also let you explicitly encode the various cases of what `CreateInstallationToken` can return like returning an error, vs. returning an empty result, vs. returning a non-empty token.

agent/internal/client/mock.go
```diff
@@ -528,5 +534,8 @@ func (c *Mock) GetPullRequestInfo(ctx context.Context, taskData TaskData, prNum
 }
 
 func (c *Mock) CreateInstallationToken(ctx context.Context, td TaskData, owner, repo string) (string, error) {
-	return "token", nil
+	if c.CreateInstallationTokenFail {
+		return "", nil
```
Kimchelly, line 538: Seems a little odd to return nil error in the case of a "failure". Seems like it could be covered more nicely by the suggestion below.

agent/internal/client/mock.go
```diff
@@ -28,6 +28,11 @@ import (
 	"google.golang.org/grpc"
 )
 
+const (
+	MockedGitHubAppToken = "mocked_github_app_token"
+	NotValidRepo         = "not_valid_repo"
```
Kimchelly, line 33: Seems like this constant is no longer used.

agent/internal/client/mock.go
bynn: updated

agent/internal/client/mock.go
Kimchelly: Makes sense! Keeping the mock simple and just returning the values that the test sets is preferable compared to adding a lot of special business logic like validating inputs.

Kimchelly (approved):

PR #7147, opened 2023-10-17
EVG-21092 Pass in installation token to GitHub client

bynn:
EVG-21092

### Description
we werent properly passing in the installation id after getting it previously 

### Testing
staging
made a patch on a project with no installation id to see that it did not error and cached it correctly
https://spruce-staging.corp.mongodb.com/task/private_sandbox_git_env_test_bynntask_patch_a95ba03b40ce0076c039db9e4e15af44e7e634cc_652eb101b23736fbe668f04e_23_10_17_16_06_38/logs?execution=3&logtype=task

### DIFF
--- a/github_app.go
+++ b/github_app.go
@@ -157,7 +157,7 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 		return "", errors.Wrapf(err, "getting cached installation id for '%s/%s'", owner, repo)
 	}
 	if installationID == 0 {
-		installationID, resp, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
+		installation, resp, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


github_app.go
```diff
@@ -171,17 +171,19 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
...
 
+		installationID = installation.GetID()
+
 		// Cache the installation ID for owner/repo.
-		installation := GitHubAppInstallation{
+		cachedInstallation := GitHubAppInstallation{
 			Owner:          owner,
 			Repo:           repo,
-			InstallationID: installationID.GetID(),
+			InstallationID: installationID,
```
hadjri, line 184: How is this different from the previous code where we were just assigning `InstallationID: installationID.GetID(),` directly in the struct?

bynn: talked offline but essentially installationID was being defined twice. Once inside the if and once outside, so changing it on the inside did not do anything. Changing the names solves this problem

hadjri (approved):

PR #7148, opened 2023-10-17
EVG-20984 Update to latest utility package and replace some tests with utility function

ZackarySantana:
EVG-20984

### Description
Some tests manually check the index of many items in a slice when the test itself only cares about the order in which the elements in the slice appear. This rewrites some tests to use a couple utility functions in our utility package that don't care about the index but care about the order in which the elements appear. 

I groomed through some of our other tests (by searching for tests that use a high index) and I didn't find many other tests that take a similar anti-pattern of checking index + order but only require order, so the scope of tests that changed were only in one file.

### Testing
Existing unit tests and unit tests in utility package for the utility functions used.

### DIFF
--- a/agent/command/git_test.go
+++ b/agent/command/git_test.go
@@ -188,7 +188,7 @@ func (s *GitGetProjectSuite) TestBuildCloneCommandUsesHTTPS() {
 	}
 	s.Require().NoError(opts.setLocation())
 	cmds, _ := c.buildCloneCommand(s.ctx, s.comm, logger, conf, opts)
-	s.Equal("git clone https://PROJECTTOKEN:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main'", cmds[6])
+	s.True(utility.StringSliceContains(cmds, "git clone https://PROJECTTOKEN:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main'"))
 }
 
 func (s *GitGetProjectSuite) TestBuildCloneCommandWithHTTPSNeedsToken() {
@@ -231,7 +231,7 @@ func (s *GitGetProjectSuite) TestBuildCloneCommandUsesSSH() {
 	s.Require().NoError(opts.setLocation())
 	cmds, err := c.buildCloneCommand(s.ctx, s.comm, logger, conf, opts)
 	s.Require().NoError(err)
-	s.Equal("git clone 'git@github.com:evergreen-ci/sample.git' 'dir' --branch 'main'", cmds[4])
+	s.True(utility.StringSliceContains(cmds, "git clone 'git@github.com:evergreen-ci/sample.git' 'dir' --branch 'main'"))
 }
 
 func (s *GitGetProjectSuite) TestBuildCloneCommandDefaultCloneMethodUsesSSH() {
@@ -251,7 +251,7 @@ func (s *GitGetProjectSuite) TestBuildCloneCommandDefaultCloneMethodUsesSSH() {
 	s.Require().NoError(opts.setLocation())
 	cmds, err := c.buildCloneCommand(s.ctx, s.comm, logger, conf, opts)
 	s.Require().NoError(err)
-	s.Equal("git clone 'git@github.com:evergreen-ci/sample.git' 'dir' --branch 'main'", cmds[4])
+	s.True(utility.StringSliceContains(cmds, "git clone 'git@github.com:evergreen-ci/sample.git' 'dir' --branch 'main'"))
 }
 
 func (s *GitGetProjectSuite) TestBuildCloneCommandCloneDepth() {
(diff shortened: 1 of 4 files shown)


agent/command/git_test.go
```diff
@@ -461,22 +461,25 @@ func (s *GitGetProjectSuite) TestBuildHTTPCloneCommand() {
 	cmds, err := opts.buildHTTPCloneCommand(false)
 	s.NoError(err)
 	s.Require().Len(cmds, 5)
-	s.Equal("set +o xtrace", cmds[0])
-	s.Equal("echo \"git clone https://[redacted oauth token]:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main' --filter=tree:0 --single-branch\"", cmds[1])
-	s.Equal("git clone https://PROJECTTOKEN:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main' --filter=tree:0 --single-branch", cmds[2])
-	s.Equal("set -o xtrace", cmds[3])
-	s.Equal("cd dir", cmds[4])
-
+	s.True(utility.ContainsOrderedSubset([]string{
```
Kimchelly, line 464: nit: this test seems like it pretty much boils down to checking for string slice equality. Is this test interested in exact equality of the output, or just subset ordering? If it's the latter, then this is fine as-is, but if it's the former, it should just do something like:
```go
s.Equal([]string{
    "set +o trace",
    "echo \"git clone https://[redacted oauth token]:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main' --filter=tree:0 --single-branch\""
    /* + the others... */
    }, cmds)
```

agent/command/git_test.go
```diff
@@ -461,22 +461,25 @@ func (s *GitGetProjectSuite) TestBuildHTTPCloneCommand() {
...
 	opts.branch = ""
 	cmds, err = opts.buildHTTPCloneCommand(false)
 	s.NoError(err)
 	s.Require().Len(cmds, 5)
-	s.Equal("set +o xtrace", cmds[0])
-	s.Equal("echo \"git clone https://[redacted oauth token]:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --filter=tree:0 --single-branch\"", cmds[1])
-	s.Equal("git clone https://PROJECTTOKEN:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --filter=tree:0 --single-branch", cmds[2])
-	s.Equal("set -o xtrace", cmds[3])
-	s.Equal("cd dir", cmds[4])
+	s.True(utility.ContainsOrderedSubset([]string{
```
Kimchelly, line 476: Same question as above about exact equality vs. subset order.

agent/command/git_test.go
```diff
@@ -546,14 +553,16 @@ func (s *GitGetProjectSuite) TestBuildCommand() {
...
 	s.Require().Len(cmds, 8)
-	s.Equal("set -o xtrace", cmds[0])
-	s.Equal("chmod -R 755 dir", cmds[1])
-	s.Equal("set -o errexit", cmds[2])
-	s.Equal("rm -rf dir", cmds[3])
-	s.Equal("git clone 'git@github.com:evergreen-ci/sample.git' 'dir' --branch 'main'", cmds[4])
-	s.Equal("cd dir", cmds[5])
-	s.Equal("git reset --hard ", cmds[6])
-	s.Equal("git log --oneline -n 10", cmds[7])
+	s.True(utility.ContainsOrderedSubset([]string{
```
Kimchelly, line 556: Same question as above about exact equality vs. subset order.

agent/command/git_test.go
```diff
@@ -564,17 +573,19 @@ func (s *GitGetProjectSuite) TestBuildCommand() {
...
-	s.Equal("set -o errexit", cmds[2])
-	s.Equal("rm -rf dir", cmds[3])
-	s.Equal("set +o xtrace", cmds[4])
-	s.Equal("echo \"git clone https://[redacted oauth token]:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main'\"", cmds[5])
-	s.Equal("git clone https://PROJECTTOKEN:x-oauth-basic@github.com/evergreen-ci/sample.git 'dir' --branch 'main'", cmds[6])
-	s.Equal("set -o xtrace", cmds[7])
-	s.Equal("cd dir", cmds[8])
-	s.Equal("git reset --hard ", cmds[9])
-	s.Equal("git log --oneline -n 10", cmds[10])
+	s.True(utility.ContainsOrderedSubset([]string{
```
Kimchelly, line 576: Same question as above about exact equality vs. substring order.

agent/command/git_test.go
```diff
@@ -676,11 +691,13 @@ func (s *GitGetProjectSuite) TestBuildModuleCommand() {
 	cmds, err := c.buildModuleCloneCommand(conf, opts, "main", nil)
 	s.NoError(err)
 	s.Require().Len(cmds, 5)
-	s.Equal("set -o xtrace", cmds[0])
-	s.Equal("set -o errexit", cmds[1])
-	s.Equal("git clone 'git@github.com:evergreen-ci/sample.git' 'module'", cmds[2])
-	s.Equal("cd module", cmds[3])
-	s.Equal("git checkout 'main'", cmds[4])
+	s.True(utility.ContainsOrderedSubset([]string{
```
Kimchelly, line 694: Same question as above about exact equality vs. subset order.

agent/command/git_test.go
```diff
@@ -720,13 +741,15 @@ func (s *GitGetProjectSuite) TestBuildModuleCommand() {
...
-	s.Equal("cd module", cmds[3])
-	s.Regexp("^git fetch origin \"pull/1234/merge:evg-merge-test-", cmds[4])
-	s.Regexp("^git checkout 'evg-merge-test-", cmds[5])
-	s.Equal("git reset --hard 1234abcd", cmds[6])
+	s.True(utility.StringSliceContainsOrderedPrefixSubset([]string{
+		"set -o xtrace",
+		"set -o errexit",
+		"git clone 'git@github.com:evergreen-ci/sample.git' 'module'",
+		"cd module",
+		"^git fetch origin \"pull/1234/merge:evg-merge-test-",
```
Kimchelly, line 749: Since the `^` is a regexp pattern, does this work with `utility.StringSliceContainsOrderedPrefixSubset`? Seems like we want to get rid of the leading `^`, since it's already matching a prefix, no?

agent/command/git_test.go
```diff
@@ -720,13 +741,15 @@ func (s *GitGetProjectSuite) TestBuildModuleCommand() {
...
+		"^git checkout 'evg-merge-test-",
```
Kimchelly, line 750: Same question the one above.

Kimchelly (changes requested):

agent/command/git_test.go
ZackarySantana: It's a good question. My thinking was the tests are just testing if it compiles a command list that does contain that command - disregarding the other commands (as they are tested in some of the other functions). That the intention of this test was to make sure it correctly compiles this part

agent/command/git_test.go
ZackarySantana: Yeah, I meant to remove those and I used the prefix version to emulate that, I'll remove them

agent/command/git_test.go
Kimchelly: Sounds good. Thanks!

Kimchelly (approved):
LGTM mod tests passing. Thanks for the testing improvements!

PR #7149, opened 2023-10-17
EVG-20908: handle cross-variant dependencies in generate.tasks

Kimchelly:
EVG-20908

### Description
The gist of the bug is that task creation in generate.tasks had an issue with its order of operations, resulting in cross-variant dependencies not working in some cases.

Once generate.tasks has found the set of tasks that need to be created, generate.tasks creates tasks in two separate steps:

1. It creates newly-generated tasks in existing BVs.
2. It creates newly-generated tasks to newly-generated BVs.

For each steps, it adds dependencies between tasks by looking up the task ID of the dependency (so it knows what task ID a particular task depends on). Unfortunately, step 1 only has the task IDs for generated tasks in _existing_ BVs, and step 2 only has the task IDs for generated tasks in _new_ BVs. If step 1 adds a task that depends on a task created in step 2, the dependency won't be created because it can't look up the dependency's task ID (and vice versa if a step 2 task depends on a step 1 task). I slightly changed the way the two steps initialize their task ID lookup table so that each step can look up all the task IDs to be generated, rather than just the one step that's currently running.

* Fix issue with cross-variant dependencies.
* Refactor generate.tasks tests a little bit + remove some dead testing logic.
* Add a couple doc comments so the parameters to pass to task-creating functions are clearer.

### Testing
* Wrote a bunch of unit tests for cross-variant dependencies where step 1 and step 2 need to know about all the tasks about to be generated. I also verified that these tests failed when run against the original logic, and pass with the new logic.
* Tested in staging that [the original logic didn't make the cross-variant dependency](https://spruce-staging.corp.mongodb.com/task/evg_lint_lint_util_patch_dc56f975b82980ae10dc85e5515011b00cf3e957_652d8ed8b2373698aadfa340_23_10_16_19_28_25/logs?execution=0), but with the new logic, [it created the cross-variant dependency (under `Depends On`)](https://spruce-staging.corp.mongodb.com/task/evg_lint_lint_util_patch_dc56f975b82980ae10dc85e5515011b00cf3e957_652eceaeb237360c54c02db4_23_10_17_18_13_03/logs?execution=0&sortBy=STATUS&sortDir=ASC).

### Documentation
N/A

### DIFF
--- a/model/generate.go
+++ b/model/generate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/evergreen-ci/evergreen/model/task"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/utility"
+	"github.com/k0kubun/pp"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
@@ -271,9 +272,13 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 	}
 
 	creationInfo := TaskCreationInfo{
-		Project:        p,
-		ProjectRef:     projectRef,
-		Version:        v,
+		Project:    p,
+		ProjectRef: projectRef,
+		Version:    v,
+		// kim: NOTE: this only has the TV pairs for newly generated tasks in
+		// existing BVs, but doesn't handle the case of an existing BV with no
+		// new generated tasks. I'm not sure if adding existing BVs and existing
+		// tasks will cause anything bad to happen here.
 		Pairs:          newTVPairsForExistingVariants,
 		ActivationInfo: *activationInfo,
 		SyncAtEndOpts:  syncAtEndOpts,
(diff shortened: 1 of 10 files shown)


model/generate.go
```diff
@@ -270,10 +270,16 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		return errors.Errorf("project '%s' not found", p.Identifier)
 	}
 
+	allTasksToBeCreatedIncludingDeps, err := NewPatchTaskIdTable(p, v, *newTVPairs, projectRef.Identifier)
```
hadjri, line 273: Tangential but it seems `NewPatchTaskIdTable` can run even for mainline commits that use generate.tasks, we should probably rename the function to something that doesn't imply it's just for patches, like `NewTaskIdTable`

hadjri, line 273: The variable name is good but can we also add a comment saying that this represents all tasks, both for existing builds and new builds

model/generate.go
```diff
@@ -282,6 +288,7 @@ func (g *GeneratedProject) saveNewBuildsAndTasks(ctx context.Context, v *Version
 		// tasks inherit that requirement.
 		ActivatedTasksAreEssentialToSucceed: g.Task.IsEssentialToSucceed,
 	}
+
```
hadjri, line 291: can we rename `addNewTasks` to `addNewTasksToExistingBuilds`?

model/generate.go
hadjri, line 273: If `allTasksToBeCreatedIncludingDeps` is created from `newTVPairs`, which is the global set of tasks to create, and we're passing either `newTVPairsForExistingVariants` or `newTVPairsForNewVariants` (depending whether we're calling `addNewTasks` or `addNewBuilds`) to `getTaskIdTables` [here](https://github.com/evergreen-ci/evergreen/pull/7149/files#diff-f50207584681d142f6e119549a8e2e22249b45dcbe3cb6f7570eb906be64299dR1803), then wouldn't the merging logic [here](https://github.com/evergreen-ci/evergreen/pull/7149/files#diff-f50207584681d142f6e119549a8e2e22249b45dcbe3cb6f7570eb906be64299dR1819-R1841) just no-op?

From my reading at least, it looks like any new table we calculate in `getTaskIdTables` would already be existing in the larger map that we're passing in [here](https://github.com/evergreen-ci/evergreen/pull/7149/files#diff-96655072263d498b6dc1571c45fc69e886abcc4bc9994343a80af2f54375c7bcR282)

hadjri: Hmmm, I see. Looks like `addNewTasks` [does it too](https://gist.github.com/hadjri/591b868070b8e73aaf2355fb385188c4#file-lifecycle-go-L1708) IIUC. That makes sense, thanks for the explanation and for grappling with the general complexity here

hadjri (approved):

PR #7150, opened 2023-10-17
EVG-21064 Refactor CreateInstallationToken

bynn:
EVG-21064

### Description
could not separate github logic out to thirdparty package because of dependencies 
the github call, db call, and the logic to use them both are all in separate functions now 

the top level methods are still a method on the settings object (CreateInstallationToken and HasGitHubApp)
they could just be a function that takes in a settings but I think it looks better to have less parameters 

### Testing
Staging patch works
all tests pass


github_app.go
```diff
@@ -1,237 +1,237 @@
...
-		// TODO EVG-19966: Return error here
-		grip.Debug(message.Fields{
-			"message": "no default owner/repo",
-			"ticket":  "EVG-19966",
-		})
-		return "", nil
-	}
-	return s.CreateInstallationToken(ctx, s.AuthConfig.Github.DefaultOwner, s.AuthConfig.Github.DefaultRepo, opts)
-}
+// // TODO EVG-21064: Refactor into model package.
```
ablack12, line 3: // lol

thirdparty/github.go
```diff
@@ -1919,3 +1923,138 @@ func GetBranchProtectionRules(ctx context.Context, token, owner, repo, branch st
...
+	itr := ghinstallation.NewAppsTransportFromPrivateKey(httpClient.Transport, authFields.appId, key)
+	httpClient.Transport = itr
+	client := github.NewClient(httpClient)
+
+	installationID, err := dbGithub.GetInstallationID(ctx, owner, repo)
+	if err != nil {
+		return "", errors.Wrapf(err, "getting cached installation id for '%s/%s'", owner, repo)
+	}
+	if installationID == 0 {
+		installation, resp, err := client.Apps.FindRepositoryInstallation(ctx, owner, repo)
```
ablack12, line 1983: ideally thirdparty functions just have the logic of calling Github, so a FindInstallation function and a CreateInstallationToken function, and then some other function could do this additional logic of checking the cache, updating the cache, calling those thirdparty functions (like this function for example: https://github.com/evergreen-ci/evergreen/blob/ff46048d63feb595caedb68fd4d9fc85ab73a96b/model/patch/patch.go#L1038  -- in this case, it seems like that could live in the new model/github/installation.go library)

model/github/installation.go
```diff
@@ -0,0 +1,88 @@
...
+
+// GetInstallationID returns the installation ID for GitHub app if it's installed for the given owner/repo.
+func GetInstallationID(ctx context.Context, owner, repo string) (int64, error) {
+	if err := validateOwnerRepo(owner, repo); err != nil {
+		return 0, err
+	}
+
+	installation := &GitHubAppInstallation{}
+	err := db.FindOneQ(GitHubAppCollection, db.Query(byOwnerRepo(owner, repo)), installation)
+	if adb.ResultsNotFound(err) {
```
ablack12, line 73: Seems like we should just return an error here -- this is a mongoDB document thing, not a typical error response

model/github/installation.go
ablack12, line 1: Noted on slack but we don't need a sub folder for this, the file could just be called github_app.go or something

rest/route/github.go
```diff
@@ -105,8 +105,10 @@ func (gh *githubHookApi) Parse(ctx context.Context, r *http.Request) error {
 // shouldSkipWebhook returns true if the event is from a GitHub app and the app is available for the owner/repo or,
 // the event is from webhooks and the app is not available for the owner/repo.
 func (gh *githubHookApi) shouldSkipWebhook(ctx context.Context, owner, repo string, fromApp bool) bool {
-	token, _ := gh.settings.CreateInstallationToken(ctx, owner, repo, nil)
-	hasApp := token != ""
+	hasApp, err := gh.settings.HasGitHubApp(ctx, owner, repo)
+	if err != nil {
```
ablack12, line 109: I think we can keep the original logic; it doesn't look like we return hasApp = true if there's an error (which seems correct -- that should be an error)

github_app.go
```diff
@@ -16,31 +14,136 @@ import (
...
+	}
+	if err := cachedInstallation.Upsert(ctx); err != nil {
+		return 0, errors.Wrapf(err, "caching installation id for '%s/%s'", owner, repo)
+	}
+
+	return installationID, nil
+
+}
+
+type GitHubAppInstallation struct {
```
ablack12, line 139: nit: can you move this towards the top of the file?

ablack12 (approved):

github_app.go
```diff
@@ -96,142 +196,70 @@ func getInstallationID(ctx context.Context, owner, repo string) (int64, error) {
...
 	}
 
 	itr := ghinstallation.NewAppsTransportFromPrivateKey(httpClient.Transport, authFields.appId, key)
 	httpClient.Transport = itr
 	client := github.NewClient(httpClient)
+	return client, nil
+}
 
-	installationID, err := getInstallationID(ctx, owner, repo)
+func getInstallationIDFromGitHub(ctx context.Context, authFields *githubAppAuth, owner, repo string) (int64, error) {
```
ablack12, line 218: I would suggest adding a comment to this function or file explaining that we have GitHub calls outside of thirdparty because they're used for initiatlizing the environment, so we don't try to fix the import cycle again alter

PR #7151, opened 2023-10-17
EVG-20948 Add ‘create_check_run’ field and validate parameters

malikchaya2:
EVG-20948

### Description
This adds the  ‘create_check_run’ field to the build variant task unit, and validates the parameters. This will allow users to specify it in their yamls like so: 

image

### Testing
Added tests

### Update

-  Some things changed and we decided not to support users passing in credentials as part of this project 

- I played with it some more and decided against adding validation to check if the path field is provided. Doing so won’t protect against instances where someone just adds `create_check_run:` and nothing else because then it’s null and validation doesn’t kick in. It only picks it up when another field is specified, such as  
 ```
         create_check_run:
           github_app_id: "some_id"
 ``` 

However, for the case when another field is provided we can just not make the path a pointer and it’ll be set to “” automatically, so we don't need the user to do that. (We also aren’t adding more fields anyway). I think the best we can do in this case is add good documentation to help users figure out how to get started. 

### Note
While it would seem that it makes more sense for create_check_run to be a string instead of a struct, I made it a struct so that we can later accommodate [EVG-21103](https://jira.mongodb.org/browse/EVG-21103) without requiring users to change their config files.

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -418,7 +418,7 @@ type BuildVariant struct {
 	DisplayTasks []patch.DisplayTask    `yaml:"display_tasks,omitempty" bson:"display_tasks,omitempty"`
 }
 
-// CheckRun is used to provide extra information about a github check run.
+// CheckRun is used to provide information about a github check run.
 type CheckRun struct {
 	// PathToOutputs is a local file path to an output json file for the checkrun
 	PathToOutputs string `yaml:"path_to_outputs" bson:"path_to_outputs"`
--- a/model/project.go
+++ b/model/project.go
@@ -144,6 +144,9 @@ type BuildVariantTaskUnit struct {
 	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
 	// TaskGroup is set if an inline task group is defined on the build variant.
 	TaskGroup *TaskGroup `yaml:"task_group,omitempty" bson:"task_group,omitempty"`
+
+	// CreateCheckRun will create a check run on GitHub if set
+	CreateCheckRun *CheckRun `yaml:"create_check_run,omitempty" bson:"create_check_run,omitempty"`
 }
 
 func (b BuildVariant) Get(name string) (BuildVariantTaskUnit, error) {
@@ -413,6 +416,16 @@ type BuildVariant struct {
 	// all of the tasks/groups to be run on the build variant, compile through tests.
 	Tasks        []BuildVariantTaskUnit `yaml:"tasks,omitempty" bson:"tasks"`
 	DisplayTasks []patch.DisplayTask    `yaml:"display_tasks,omitempty" bson:"display_tasks,omitempty"`
+}
+
+// CheckRun is used to provide extra information about a github check run.
+type CheckRun struct {
+	// PathToOutputs is a local file path to an output json file for the checkrun
+	PathToOutputs *string `yaml:"path_to_outputs" bson:"path_to_outputs"`
+	// GithubAppId and GitubPrivateKey are the user's credentials for
+	// authenticating with github for check runs.
+	GithubAppId      string `yaml:"github_app_id" bson:"github_app_id"`
+	GithubPrivateKey string `yaml:"github_private_key" bson:"github_private_key"`
 }
 
 // ParameterInfo is used to provide extra information about a parameter.
--- a/model/project.go
+++ b/model/project.go
@@ -145,7 +145,7 @@ type BuildVariantTaskUnit struct {
 	// TaskGroup is set if an inline task group is defined on the build variant.
 	TaskGroup *TaskGroup `yaml:"task_group,omitempty" bson:"task_group,omitempty"`
 
-	// CreateCheckRun will create a check run on GitHub if set
+	// CreateCheckRun will create a check run on GitHub if set.
 	CreateCheckRun *CheckRun `yaml:"create_check_run,omitempty" bson:"create_check_run,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


model/project.go
```diff
@@ -415,6 +418,16 @@ type BuildVariant struct {
 	DisplayTasks []patch.DisplayTask    `yaml:"display_tasks,omitempty" bson:"display_tasks,omitempty"`
 }
 
+// CheckRun is used to provide extra information about a github check run.
+type CheckRun struct {
+	// PathToOutputs is a local file path to an output json file for the checkrun
+	PathToOutputs *string `yaml:"path_to_outputs" bson:"path_to_outputs"`
```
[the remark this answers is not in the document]
bynn: Because an empty string is a valid input for this field, it may be worth differentiating users not specifying anything vs specifying an empty string. 

Maybe the comment can mention that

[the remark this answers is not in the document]
malikchaya2: I don't think it needs to be a pointer. I added an explanation with updates in the PR description. thanks!

model/project_parser.go
```diff
@@ -466,6 +468,19 @@ func (pbvt *parserBVTaskUnit) UnmarshalYAML(unmarshal func(interface{}) error) e
 		}
 		copy.RunOn, copy.Distros = copy.Distros, nil
 	}
+
+	// validate check runs
```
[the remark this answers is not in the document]
malikchaya2: I decided against validation for this. I added more details in the PR description.

model/project.go
```diff
@@ -415,6 +418,12 @@ type BuildVariant struct {
 	DisplayTasks []patch.DisplayTask    `yaml:"display_tasks,omitempty" bson:"display_tasks,omitempty"`
 }
 
+// CheckRun is used to provide information about a github check run.
+type CheckRun struct {
+	// PathToOutputs is a local file path to an output json file for the checkrun
+	PathToOutputs string `yaml:"path_to_outputs" bson:"path_to_outputs"`
```
hadjri, line 424: If this struct only holds one string now, does it need to be a struct anymore? Or are we going to be adding new fields in the future potentially

malikchaya2: yes, I added a note up top. 

> While it would seem that it makes more sense for create_check_run to be a string instead of a struct, I made it a struct so that we can later accommodate [EVG-21103](https://jira.mongodb.org/browse/EVG-21103) without requiring users to change their config files.

hadjri (approved):

model/project.go
```diff
@@ -144,6 +144,9 @@ type BuildVariantTaskUnit struct {
 	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
 	// TaskGroup is set if an inline task group is defined on the build variant.
 	TaskGroup *TaskGroup `yaml:"task_group,omitempty" bson:"task_group,omitempty"`
+
+	// CreateCheckRun will create a check run on GitHub if set
```
bynn, line 148: nit: can we add a period to make it a complete sentence

model/project.go
```diff
@@ -415,6 +418,12 @@ type BuildVariant struct {
 	DisplayTasks []patch.DisplayTask    `yaml:"display_tasks,omitempty" bson:"display_tasks,omitempty"`
 }
 
+// CheckRun is used to provide information about a github check run.
+type CheckRun struct {
+	// PathToOutputs is a local file path to an output json file for the checkrun
```
bynn, line 423: nit: can we add a period here too

model/project_parser.go
```diff
@@ -436,6 +436,8 @@ type parserBVTaskUnit struct {
 	Activate *bool `yaml:"activate,omitempty" bson:"activate,omitempty"`
 	// TaskGroup is set if an inline task group is defined on the build variant config.
 	TaskGroup *parserTaskGroup `yaml:"task_group,omitempty" bson:"task_group,omitempty"`
+	// CreateCheckRun will create a check run on GitHub if set
```
bynn, line 439: sorry for all the nits about periods :shipit:

bynn (changes requested):
would it be possible to add some tests where we read from a yaml like [this](https://github.com/evergreen-ci/evergreen/blob/0f8e56882cb0b5d61ea2ca0373847510b5fae74f/model/lifecycle_test.go#L1646)?

model/project_parser.go
malikchaya2: np! fixed the periods and added a test that reads from a yaml.

model/project_parser_test.go
```diff
@@ -727,6 +744,53 @@ func TestParserTaskSelectorEvaluation(t *testing.T) {
...
+    create_check_run:
+      path_to_outputs: ""
+tasks:
+- name: t1
+`
+
+	_, err = LoadProjectInto(ctx, []byte(ymlWithEmptyString), nil, "id", proj)
+	assert.NotNil(proj)
+	assert.Nil(err)
+	assert.Len(proj.BuildVariants, 1)
```
bynn, line 786: can we use require.len here so it doesn't panic

model/project_parser_test.go
```diff
@@ -727,6 +744,53 @@ func TestParserTaskSelectorEvaluation(t *testing.T) {
...
+tasks:
+- name: t1
+`
+
+	proj := &Project{}
+	ctx := context.Background()
+	_, err := LoadProjectInto(ctx, []byte(yml), nil, "id", proj)
+	assert.NotNil(proj)
+	assert.Nil(err)
+	assert.Len(proj.BuildVariants, 1)
```
bynn, line 765: can we use require.len here

bynn (approved):
approved with 2 small changes!

PR #7152, opened 2023-10-18
EVG-20612 remove activation logging

ablack12:
EVG-20612
### Description
I _think_ it's time to timebox this investigation -- the TL;DR is that sometimes we're calling this function with something that doesn't have activation info in the parser project, so we should avoid running the query. (The underlying problem is calling the function in a weird way but this doesn't have user-facing impact outside of slowness from running the query unnecessarily)

Confirmed also that the query isn't too expensive when we call it reasonably.

----

Other information I found in case we want to look into this again or others find it helpful:
- This happens in generate tasks regularly but not often. It's been hit for task groups and not-task-groups 
    - (I'd stopped logging for the non-generated-tasks case and i don't remember why)
- Seems to be mostly or entirely patches
- We get activationInfo using the generated project, but we get activationTime later using the parser project, but I don't see why these would have discrepancies. 

If others have ideas we can leave the logging in and make a new ticket to re-investigate this, since the work of keeping the query from tanking us is done.


model/project_ref.go
```diff
@@ -2309,24 +2308,8 @@ func (p *ProjectRef) GetActivationTimeForVariant(variant *BuildVariant) (time.Ti
 
 // GetActivationTimeForTask returns the time at which this task should next be activated.
 // Temporarily takes in the task ID that prompted this query, for logging.
-func (p *ProjectRef) GetActivationTimeForTask(t *BuildVariantTaskUnit, taskId string) (time.Time, error) {
+func (p *ProjectRef) GetActivationTimeForTask(t *BuildVariantTaskUnit) (time.Time, error) {
 	defaultRes := time.Now()
-	// Verify that we mean to be getting activation time for task
-	if !t.HasSpecificActivation() {
```
ablack12, line 2315: Moved this out of the function and into the generate tasks usage case to avoid adding unnecessary build variant task statuses to the version.

model/lifecycle.go
```diff
@@ -536,9 +536,11 @@ func addTasksToBuild(ctx context.Context, creationInfo TaskCreationInfo) (*build
 		if !utility.StringSliceContains(tasksWithActivationTime, t.DisplayName) {
 			continue
 		}
-		activateTaskAt, err := creationInfo.ProjectRef.GetActivationTimeForTask(
-			creationInfo.Project.FindTaskForVariant(t.DisplayName, creationInfo.Build.BuildVariant),
-			t.Id)
+		bvtu := creationInfo.Project.FindTaskForVariant(t.DisplayName, creationInfo.Build.BuildVariant)
+		if !bvtu.HasSpecificActivation() {
```
Kimchelly, line 540: I always have a bit of difficulty understanding how specific activation works when defined in the BV vs BVTU, since the settings precedence doesn't work the same as the other task fields - Does this also handle the case when you set the specific activation condition at the BV level rather than the BVTU level? I thought that specific activation fields like cron/batchtime weren't propagated down to the BVTU level and we had to [check the BV settings explicitly](https://github.com/evergreen-ci/evergreen/blob/332598845126ee2c235813a87661043598c880ab/model/generate.go#L538) before [checking the BVTU settings](https://github.com/evergreen-ci/evergreen/blob/332598845126ee2c235813a87661043598c880ab/model/generate.go#L551).

Kimchelly (commented):
I think I actually have a clue because I saw this debug log while debugging EVG-20369. Is this happening because `activate: true` is set at the BV level? I imagine it may be considered "specific activation" just like cron/batchtime but it applies to all version types, including patches, so the logic may not be considering that difference.

ablack12:
I've vowed to revisit this ticket before the end of the year haha:
The bug in EVG-20612 is that there are too many tasks listed in the tasks with specific activation list, which is separate from the variant one. Specific activation levels _aren't_ propagated down to the bvtu level -- it's valid to define those separately. So for example this shouldn't activate the task even though the variant has a cron:
```
name: "hello_variant"
cron: @daily
tasks:
    - name: "world_task"
       activate: false
```
Something weird to me about the task above is that even though its marked as generated, I don't see any mention of it in the evergreen_config.json. I wonder if when we generate things are marked as deactivated but then we don't add them to the parser project.

ablack12:
Some other things actually are that https://spruce.mongodb.com/task/mongodb_mongo_master_linux_x86_dynamic_compile_generate_buildid_to_debug_symbols_mapping_patch_d169f5ea1f275adbae4300825d2d4135f48cbb83_65329d65a4cf47b1a4b4dc57_23_10_20_15_32_15/logs?execution=0 says it was generated by version_gen but it doesn't show up in the generated.json attached to the project, or the generated json string in the task document (nor does .symbolizer), and [we list it here in the variant](https://github.com/mongodb/mongo/blob/1e7d1941bca71294636cdf48f4b6e31a2d3cbc61/etc/evergreen_yml_components/variants/rhel/build_dev.yml#L62). Are we generating this by mistake?

ablack12:
Re the PR description and my complete lack of time I'm going to make a new ticket to look into this more, close the original ticket (since the query problem is fixed). I think we can also still remove the logging since we've got a good time period, assuming we investigate this in the near term.

model/lifecycle.go
ablack12: This only checks the build variant task level but that's the correct thing to check for getting the task activation time -- if it's set for the variant, that'll be part of the variant's activation info (this allows us to set one activation for the variant, and a different one for individual tasks that are overriding it).

Kimchelly (approved):

PR #7154, opened 2023-10-18
EVG-21060: Remove field-level permissions for project event logs

sophstad:
EVG-21060

### Description
The linked thread shows a user receiving lots of errors because the permissions directive `@requireProjectFieldAccess` threw an error for multiple fields in an event log.

Regardless of why that specific case failed, I don't think we really need to verify that a user has field-level permissions for event logs. We already verify that they have project view access when the query is called:

https://github.com/evergreen-ci/evergreen/blob/ff46048d63feb595caedb68fd4d9fc85ab73a96b/graphql/schema/query.graphql#L55

I also think it's superfluous to check permissions for _every_ field in a Project Ref. Evergreen doesn't grant permissions on a granular level in this way. Moving the directive to the `ProjectRef` struct significantly reduces the number of permission checks made in a call to the project settings query, and allows us to reuse the same `ProjectRef` struct between project queries and project events queries while enforcing different permissioning.

### Testing
Updated existed directive tests to ensure that they pass when applied to project ref.

### Documentation
N/A


graphql/resolver.go
```diff
@@ -179,11 +179,12 @@ func New(apiURL string) Config {
 	c.Directives.RequireProjectFieldAccess = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
```
ablack12, line 179: should this directive be renamed to like RequireProjectViewAccess or etc, since we aren't calling it by field anymore?

ablack12 (approved):

khelif96 (approved):
Makes sense to me mod Annies comment.

PR #7159, opened 2023-10-20
EVG-20792 Add logging to bisect stepback

ZackarySantana:
[EVG-20792](https://jira.mongodb.org/browse/EVG-20792)

### Description
Add simple logging to bisect stepback

### Testing
Existing unit tests. Secondary ticket [EVG-20794](https://jira.mongodb.org/browse/EVG-20794) is going to test bisect stepback algorithm.

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -639,6 +639,14 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
 		return errors.Wrapf(err, "setting stepback info for task '%s'", nextTask.Id)
 	}
 
+	grip.Debug(message.Fields{
+		"message":                       "bisect stepback",
+		"last_failing_stepback_task_id": s.LastFailingStepbackTaskId,
+		"last_passing_stepback_task_id": s.LastPassingStepbackTaskId,
+		"next_task_id":                  nextTask.Id,
+		"project_id":                    t.Project,
+	})
+
 	// Activate the next task.
 	if err = SetActiveState(ctx, evergreen.StepbackTaskActivator, true, *nextTask); err != nil {
 		return errors.Wrapf(err, "setting task '%s' active", nextTask.Id)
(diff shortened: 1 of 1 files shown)


model/task_lifecycle.go
```diff
@@ -639,6 +639,14 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
 		return errors.Wrapf(err, "setting stepback info for task '%s'", nextTask.Id)
 	}
 
+	grip.Debug(message.Fields{
```
Kimchelly, line 642: Is this is a temporary log? If so, is there a ticket to remove it eventually?

model/task_lifecycle.go
```diff
@@ -639,6 +639,14 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
...
+		"message":                       "bisect stepback",
+		"last_failing_stepback_task_id": s.LastFailingStepbackTaskId,
+		"last_passing_stepback_task_id": s.LastPassingStepbackTaskId,
+		"next_task_id":                  nextTask.Id,
```
Kimchelly, line 646: I dunno what this log is for exactly, but would it be helpful to also include the current task that's being stepped back (like current task ID /display name/build ID)?

Kimchelly (approved):
LGTM mod some non-blocking questions.

model/task_lifecycle.go
ZackarySantana: There's no ticket in the epic or that I can find that is to remove the logging, there also hasn't been any discussion to remove it

model/task_lifecycle.go
ZackarySantana: The log is to be able to trace tasks in stepback and the path they take / should take. I think adding some of those fields and the status of the task that's having stepback performed might be useful actually, let me add those.

model/task_lifecycle.go
Kimchelly: I guess the question is more about what we will plan to do with the log. If this is a thing just to verify that the algorithm is correct before the project is finished, then it seems like we would want to eventually remove it. But if we have an ongoing purpose to keep it around, then we should probably upgrade it to a info log so we know that it's not a stray debug log that nobody uses.

ZackarySantana: I talked to Malik about it and it's more so an ongoing purpose, so I'll update it to an info log

Kimchelly: Cool, I already LGTM'd so feel free to merge whenever!

PR #7162, opened 2023-10-20
EVG-21044 Add temporary fix for ssh module cloning

malikchaya2:
EVG-21044

### Description
If a user specifies a module with the following format (which is the suggested format in the docs): 

```
modules:
  name: example
  repo: git@github.com:owner/repo.git
  branch: main
```

extract the owner and repo from the string and use that to clone over https instead. This is a temporary workaround that will eventually be deprecated. EVG-21113 will take care of letting users specify a repo/owner pair instead, updating the docs, and send out communications to users. 

### Testing
Tested by spinning up a [task](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_small_my_task_patch_0f8e56882cb0b5d61ea2ca0373847510b5fae74f_6531a71297b1d3b98840e872_23_10_19_22_00_59/logs?execution=0) on staging with [this config](https://github.com/malikchaya2/evergreen/blob/14dab3753c518cb6d636b05424be6b7c3211d592/self-tests.yml). With these changes, [it clones over https](https://parsley-staging.corp.mongodb.com/evergreen/evg_ubuntu2204_small_my_task_patch_0f8e56882cb0b5d61ea2ca0373847510b5fae74f_6531a71297b1d3b98840e872_23_10_19_22_00_59/0/all?bookmarks=0,209,210,211,212,213,230,231,232,233,311&shareLine=209). 
On main, [it clones over ssh](https://parsley.mongodb.com/evergreen/evergreen_ubuntu2204_small_my_task_patch_0f8e56882cb0b5d61ea2ca0373847510b5fae74f_65316e6f3e8e86fc947dc913_23_10_19_17_59_23/0/all?bookmarks=0,206,207,208,209,227,296).


thirdparty/git.go
```diff
@@ -302,6 +302,22 @@ func ParseGitUrl(url string) (string, string, error) {
 	return owner, repo, nil
 }
 
+// parseModuleLocation extracts the owner and repo from a string
+// in the format "git@github.com:owner/repo.git"
+func ParseModuleLocation(moduleLocation string) (string, string, error) {
```
hadjri, line 307: Can we add some testcases for this?

agent/command/git.go
```diff
@@ -733,14 +733,26 @@ func (c *gitFetchProject) fetchModuleSource(ctx context.Context,
...
 	}
-	// Module's location takes precedence over the project-level clone
-	// method.
+	// If the module repo is using the deprecated ssh cloning method, extract the owner
+	// and repo from the string and construct a https cloning link manually.
+	// This is a temporary workaround which will be removed once users have switched over.
 	if strings.Contains(opts.location, "git@github.com:") {
-		opts.method = evergreen.CloneMethodLegacySSH
-	} else if opts.method == evergreen.CloneMethodOAuth {
+		owner, repo, err := thirdparty.ParseModuleLocation(opts.location)
```
hadjri, line 740: Can we add a log here explaining that we're manually converting the SSH cloning to OAuth? I feel like this inevitably may break some clones on older distros, in which case we'd need some sort of way to notify why a task's behavior has suddenly changed.

hadjri (commented):
Should we add this as a temporary validation warning as well?

agent/command/git.go
malikchaya2: We only do ssh cloning on modules right now. So for the distro that is broken (amazon1-2018-small), we won't be able to get the project either, so the whole patch will fail anyway. I therefore don't expect this to break anything that's not already broken. 

I think it's a good idea to log anyway though.  

Re 

> Should we add this as a temporary validation warning as well?
I have a separate ticket to figure out the best way to warn users.

agent/command/git.go
```diff
@@ -733,14 +733,28 @@ func (c *gitFetchProject) fetchModuleSource(ctx context.Context,
...
 	}
-	// Module's location takes precedence over the project-level clone
-	// method.
+	// If the module repo is using the deprecated ssh cloning method, extract the owner
+	// and repo from the string and construct a https cloning link manually.
+	// This is a temporary workaround which will be removed once users have switched over.
 	if strings.Contains(opts.location, "git@github.com:") {
-		opts.method = evergreen.CloneMethodLegacySSH
-	} else if opts.method == evergreen.CloneMethodOAuth {
+		logger.Task().Infof("ssh cloning is being deprecated. We are manually converting '%s'"+
```
hadjri, line 740: Should we add a mention of some EVG ticket to remove all this logic once the migration is done (if there is such a ticket)?

thirdparty/git.go
```diff
@@ -302,6 +302,22 @@ func ParseGitUrl(url string) (string, string, error) {
 	return owner, repo, nil
 }
 
+// parseModuleLocation extracts the owner and repo from a string
```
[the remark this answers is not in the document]
malikchaya2: done

agent/command/git.go
malikchaya2: Not yet. This is part of a larger effort so I don't think that's necessary right now.

hadjri (approved):

PR #7164, opened 2023-10-20
EVG-19763 Validate requester-level fields with override hierarchy

hadjri:
EVG-19763

### Description
Refactored the warning validations to respect the [override hierarchy](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#task-fields-override-hierarchy) for requester-level settings by checking the merged `BuildVariantTaskUnit` rather than the non-merged `ProjectTask `
### Testing
Added a new unit test that checks for all 3 of the levels described here to be respected when checking requester level fields like patchable, patch_only, etc.

### DIFF
--- a/validator/project_validator.go
+++ b/validator/project_validator.go
@@ -155,6 +155,7 @@ var projectWarningValidators = []projectValidator{
 	checkTaskRuns,
 	checkModules,
 	checkTasks,
+	checkRequestersForTaskDependencies,
 	checkBuildVariants,
 }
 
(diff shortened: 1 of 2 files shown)


validator/project_validator.go
```diff
@@ -1533,31 +1534,48 @@ func validateTaskDependencies(project *model.Project) ValidationErrors {
...
+				// Implicit build variant - if no build variant is explicitly
+				// stated, the task and dependency should both run in the same
+				// build variant.
+				dep.Variant = bvtu.Variant
+			}
+			dependent := project.FindTaskForVariant(dep.TaskName, dep.Variant)
+			if dependent == nil {
+				continue
+			}
+			if utility.FromBoolPtr(dependent.PatchOnly) && !utility.FromBoolPtr(bvtu.PatchOnly) {
```
Kimchelly, line 1561: Not exactly related to the original ticket, but this may not handle some incompatible requester cases such as a task with `patch_only: true` + a dependency with `git_tag_only: true`. It also should probably handle `allowed_requesters`, which is a generalization of the boolean requester fields.

I'm also fine if we want to keep this change short and make a separate ticket for handling those validation cases.

validator/project_validator_test.go
```diff
@@ -649,22 +649,74 @@ buildvariants:
 }
 
 func TestCheckRequestersForTaskDependencies(t *testing.T) {
```
Kimchelly, line 651: Since this is only tests for the case where it should emit a warning, can we write a test for the success case, like having a `patch_only: true` task + a `patch_only: true` dependency?

validator/project_validator_test.go
```diff
@@ -649,22 +649,74 @@ buildvariants:
...
+    display_name: Ubuntu 22.04
+    patchable: false
+    run_on:
+      - localhost
+    tasks:
+      - name: "task1"
+      - name: "dep1"
+      - name: "dep2"
+        patchable: false
+      - name: "dep3"
```
Kimchelly, line 708: Would it be possible to split this test down into simpler YAML tests by making more test case functions, so that it doesn't get too tricky to understand what's being tested? It seems like this one test case is testing multiple different conditions here, each of which results in a validation warning for different reasons (one because `dep1` sets patchable on the task level, one because `dep2` sets `patchable` is set at the BVTU level, and one because `dep3` inherits `patchable` from the BV level). It's generally preferable to try keeping test cases small and focused on testing a particular condition, rather than trying to cover as many cases as possible within a single test, since splitting up test cases reduces complexity.

Kimchelly (changes requested):

validator/project_validator_test.go
hadjri: Sure, done

validator/project_validator_test.go
hadjri: Sure, added

validator/project_validator.go
hadjri: Yep that makes sense, I created https://jira.mongodb.org/browse/EVG-21116 given that we'd need to both confirm what these cases all are, and write a bunch of test cases for them, which is enough for its own ticket

Kimchelly (approved):
LGTM mod suggestion about testing a non-default value.

PR #7171, opened 2023-10-23
DEVPROD-49: Support loading local logs

julianedwards:
[DEVPROD-49
](https://jira.mongodb.org/browse/DEVPROD-49)

### Description
Add support for GridFS-backed buckets for writing and fetching logs. This will allow us to continue using the same infrastructure for loading local data for smoke tests and local evergreen.

### Testing
Updated relevant unit tests.

### Documentation
N/A


config_bucket.go
```diff
@@ -11,32 +11,45 @@ import (
...
+	LogBucket BucketConfig `bson:"log_bucket" json:"log_bucket" yaml:"log_bucket"`
+}
+
+var bucketsConfigLogBucketKey = bsonutil.MustHaveTag(BucketsConfig{}, "LogBucket")
+
+// BucketConfig represents the admin config for an individual bucket.
 type BucketConfig struct {
-	LogBucket Bucket `bson:"log_bucket" json:"log_bucket" yaml:"log_bucket"`
+	Name   string `bson:"name" json:"name" yaml:"name"`
+	Type   string `bson:"type" json:"type" yaml:"type"`
```
Kimchelly, line 29: nit: if this only allows the `BucketType` constants to be set here, then consider not using a plain string and making a `BucketType` enum + making those constants of that type.

config_bucket.go
```diff
@@ -11,32 +11,45 @@ import (
...
+	DBName string `bson:"db_name" json:"db_name" yaml:"db_name"`
 }
 
-var bucketConfigLogBucketKey = bsonutil.MustHaveTag(BucketConfig{}, "LogBucket")
+func (c *BucketConfig) validate() error {
+	if c.Type == "" {
+		c.Type = BucketTypeS3
```
Kimchelly, line 35: Is this a reasonable default in all scenarios? I would have thought that S3 only applies in prod, we would explicitly want it to be S3, and in tests, we would explicitly want to set it to something other than S3. Instead of setting a default type, should we just validate that the type must be set to something before using the bucket (and have the UI default it to S3)?

config_bucket.go
```diff
@@ -11,32 +11,45 @@ import (
...
 const (
-	BucketTypeLocal = "local"
-	BucketTypeS3    = "s3"
+	BucketTypeGridFS = "gridfs"
+	BucketTypeLocal  = "local"
+	BucketTypeS3     = "s3"
 )
 
-// BucketConfig represents the admin config section for bucket storage.
+// BucketsConfig represents the admin config section for bucket storage.
```
Kimchelly, line 19: Since S3 is used in various contexts throughout Evergreen, can we make this a little more explicit and state that this is bucket storage for storing internally-owned Evergreen data (as opposed to other potential usages like public buckets that all users can directly store data in)?

Kimchelly (approved):
LGTM mod nits.

config_bucket.go
julianedwards: I did this as a safeguard for prod—I don't actually allow setting the bucket type in the UI to avoid someone accidentally changing it.

Kimchelly (approved):

PR #7178, opened 2023-10-30
DEVPROD-57: Fetch logs using the log fetching task methods

julianedwards:
[DEVPROD-57](https://jira.mongodb.org/browse/DEVPROD-57)

### Description
Fetch logs using the log fetching task methods.

For the sake of readable PRs and safer deploys, I decided to break this work up. The first part (this PR) temporarily adds fallback code to support DB logs in the task output interface, a subsequent PR will remove support for DB logs, update the agent code, and migrate the local and smoke data to use a GridFS backed pail bucket for log data.

- Move task log model into its own package `model/tasklog` to avoid import cycle in `taskoutput`
- Move test log model into its own package `model/testlog` to avoid import cycle in `taskoutput`
- Create temporary log iterator for the `apimodels.LogMessage` type
- Add temporary DB fallback logic for fetching task and test log fetching in the `taskoutput` package
- Replace all log fetching code with the new log fetching task models in the `service` and `graphql` packages
- Remove buildlogger-specific logic from legacy UI code

### Testing
Tested in staging that we can still write and fetch task and test logs.

### Documentation
N/A


apimodels/task_log.go
```diff
@@ -54,149 +45,121 @@ type TaskLog struct {
...
-type GetBuildloggerLogsOptions struct {
-	BaseURL       string `json:"_"`
-	TaskID        string `json:"-"`
-	TestName      string `json:"-"`
-	GroupID       string `json:"-"`
-	Execution     *int   `json:"-"`
-	PrintPriority bool   `json:"-"`
-	Tail          int    `json:"-"`
-	LogType       string `json:"-"`
+func GetPriority(prefix string) level.Priority {
```
ablack12, line 63: can you add a function comment here (unless this is mega temporary)

apimodels/task_log.go
```diff
@@ -54,149 +45,121 @@ type TaskLog struct {
...
 
-// ReadBuildloggerToChan parses Cedar buildlogger log lines by message and
-// severity and reads into a channel.
-func ReadBuildloggerToChan(ctx context.Context, taskID string, r io.ReadCloser, lines chan<- LogMessage) {
-	var (
-		line string
-		err  error
-	)
+// StreamFromLogIterator streams log lines from the given iterator to the
+// returned log message channel. It is the responsibility to close the log
```
ablack12, line 95: "It is the responsibility to close the log" -- can you fix this wording?

graphql/util.go
```diff
@@ -915,6 +917,35 @@ func getProjectMetadata(ctx context.Context, projectId *string, patchId *string)
...
 }
 
+//////////////////////////////////
+// Helper functions for task logs.
+//////////////////////////////////
+
+func getTaskLogs(ctx context.Context, obj *TaskLogs, logType taskoutput.TaskLogType) ([]*apimodels.LogMessage, error) {
+	dbTask, err := task.FindOneIdAndExecution(obj.TaskID, obj.Execution)
+	if err != nil {
+		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Finding task %s: %s", obj.TaskID, err.Error()))
```
ablack12, line 927: task '%s'

model/log/service_v0.go
```diff
@@ -113,7 +113,7 @@ func (s *logServiceV0) getLogChunks(ctx context.Context, logNames []string) (map
 
 	for _, chunks := range logChunks {
 		sort.Slice(chunks, func(i, j int) bool {
-			return chunks[i].start > chunks[j].start
+			return chunks[i].start < chunks[j].start
```
ablack12, line 116: what's the thought behind this?

ablack12 (approved):

model/log/service_v0.go
julianedwards: We want to iterate through chunks of log lines in ascending order rather than descending order.

apimodels/task_log.go
julianedwards: It doesn't need to be exported so I just made it private, it will also be temporary.

apimodels/task_log.go
julianedwards: Done.

graphql/util.go
julianedwards: Good catch! Done.

service/ui.go
```diff
@@ -319,7 +319,6 @@ func (uis *UIServer) GetServiceApp() *gimlet.APIApp {
 	// Proxy downloads for task uploaded files via S3
 	app.AddRoute(("/task_file_raw/{task_id}/{execution}/{file_name}")).Wrap(needsLogin, needsContext, allowsCORS, viewLogs).Handler(uis.taskFileRaw).Get()
 	// Test Logs
-	app.AddRoute("/test_log/{log_id}").Wrap(needsLogin, needsContext, allowsCORS).Handler(uis.testLog).Get()
```
khelif96, line 322: Are we sure this route is no longer in use?

service/task.go
```diff
@@ -507,140 +493,122 @@ func getTaskDependencies(t *task.Task) ([]uiDep, string, error) {
...
 func (uis *UIServer) taskLogRaw(w http.ResponseWriter, r *http.Request) {
 	projCtx := MustHaveProjectContext(r)
 	if projCtx.Task == nil {
-		http.Error(w, "Not found", http.StatusNotFound)
+		http.Error(w, "task not found", http.StatusNotFound)
 		return
 	}
 
-	raw := (r.FormValue("text") == "true") || (r.Header.Get("Content-Type") == "text/plain")
-	printPriority := r.FormValue("priority") == "true" || !raw
```
khelif96, line 584: Looks like this is missing priority information if the `priority=true` is set.

Before:
image

After:
image

khelif96: Parsley passes the following query params to request task logs.
`?priority=true&text=true&type=T`
It looks like priority might be getting handled correctly.

khelif96 (changes requested):

graphql/tests/task/taskLogs/results.json
```diff
@@ -8,11 +8,11 @@
             "taskLogs": {
               "systemLogs": [
                 {
-                  "type": "S",
+                  "type": "",
                   "severity": "I",
                   "message": "    [javac]     protected void finalize() {",
                   "timestamp": "2019-12-07T13:31:19.393-05:00",
-                  "version": 1
+                  "version": 0
```
khelif96, line 15: What is the motivation behind these changes?

julianedwards: That was the DB version of the logs, which we no longer support.

service/task.go
julianedwards: Fixed!
Screenshot 2023-11-01 at 11 48 06 AM

khelif96:
The Parsley e2e tests are now failing it looks like the timestamps are no longer formatted the same way In the log file and may even be in a different timezone. @julianedwards

PR #7163, opened 2023-10-20
EVG-19801 add optional param to s3.get

ablack12:
EVG-19801 

### Description
Added "optional" param to s3.get, implemented similarly to s3.put. 

### Testing
Added to unit test and tested in staging -- [after changes](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_test_model_patch_ff46048d63feb595caedb68fd4d9fc85ab73a96b_6532989ab23736908cf98daa_23_10_20_15_11_33/logs?execution=1), failed only because I removed attaching test results,  but s3.get succeeded

### Documentation
Added documentation.

PR #7179, opened 2023-10-31
Revert "EVG-19801 add optional param to s3.get (#7163)"

Kimchelly:
This reverts commit 4139a1e56d2796433dcc71ba0ebce10eddd7d56d.

DEVPROD-1734

Reverting the change due to a panic and it'll go in with the daily deploy.

### DIFF
--- a/agent/command/s3_get.go
+++ b/agent/command/s3_get.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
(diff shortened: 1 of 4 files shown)


agent/command/s3_get.go
```diff
@@ -151,7 +127,7 @@ func (c *s3get) Execute(ctx context.Context,
 
 	// expand necessary params
 	if err := c.expandParams(conf); err != nil {
-		return errors.Wrap(err, "expanding params")
+		return errors.Wrap(err, "applying expansions")
```
ZackarySantana, line 130: I know we are reverting but do we want to revert this error message as well? I think it makes sense not to in terms of single responsibility PR's but also it makes sense to to have better logging. I'm not sure if this revert is going to be 'unreverted' eventually / if the ticket is still going to be implemented though.

ZackarySantana: LGTM but I am 2nd

Kimchelly:
The one failing test was because the original commit didn't update the agent version, going to merge without waiting for checks.

PR #7181, opened 2023-10-31
DEVPROD-68: statistics of free and used inodes

jasonwalsh:
[DEVPROD-68](https://jira.mongodb.org/browse/DEVPROD-68)

### Description

This pull request adds the `df -i` command to the [StatsCollector](https://github.com/evergreen-ci/evergreen/blob/main/agent/stats.go#L20-L26) struct for reporting statistics on the number of free and used inodes.

### Testing

N/A

### Documentation

N/A

### DIFF
--- a/agent/command/s3_get.go
+++ b/agent/command/s3_get.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
@@ -47,7 +48,13 @@ type s3get struct {
 	LocalFile string `mapstructure:"local_file" plugin:"expand"`
 	ExtractTo string `mapstructure:"extract_to" plugin:"expand"`
 
-	bucket pail.Bucket
+	// Optional, when set to true, causes this command to be skipped over without an error when
+	// the path specified in remote_file does not exist. Defaults to false, which triggers errors
+	// for missing files.
+	Optional string `mapstructure:"optional" plugin:"expand"`
+
+	bucket      pail.Bucket
+	skipMissing bool
 
 	base
 }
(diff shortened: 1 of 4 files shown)


agent/agent.go
```diff
@@ -675,6 +675,7 @@ func (a *Agent) runPreAndMain(ctx context.Context, tc *taskContext) (status stri
 		"uptime",
 		"df -h",
 		"${ps|ps}",
+		"df -i",
```
julianedwards, line 678: I think we can just combine this with L676 `df -h -i`.

julianedwards (changes requested):

julianedwards:
You will also need to update the agent version in `config.go`, just bump the date string to today.

jasonwalsh:
Thanks for the review @julianedwards. I made the necessary changes in a03b8be01.

agent/agent.go
```diff
@@ -673,7 +673,7 @@ func (a *Agent) runPreAndMain(ctx context.Context, tc *taskContext) (status stri
 		a.jasper,
 		defaultStatsInterval,
 		"uptime",
-		"df -h",
+		"df -h -i",
```
julianedwards, line 676: So actually, it looks like running `df -h -i` on Linux machines will suppress the block size stats and only print the inode stats, whereas, for Unix-based systems, `df -h` already prints both the block size and inode stats (and does the same when you add the `-i` flag). 

Printing `df -h -i` on a separate line would be the easiest solution, but on the Unix-based systems it would print the same thing twice adding a significant amount of noise to these logs. We _could_ check the OS and only add the `df -h -i` command if it is linux (and maybe windows? not sure what happens with cygwin).

jasonwalsh: @julianedwards perhaps 655b27bca will solve the problem? I'm open to suggestions!

agent/agent.go
```diff
@@ -677,6 +678,10 @@ func (a *Agent) runPreAndMain(ctx context.Context, tc *taskContext) (status stri
 		"${ps|ps}",
 	)
 
+	if runtime.GOOS == "linux" || runtime.GOOS == "windows" {
```
julianedwards, line 681: Nit: remove newline above.

julianedwards (approved):
That should do it! Could you test it in staging before merging and make sure that it prints out what we expect on linux, mac, and windows?

PR #7183, opened 2023-10-31
DEVPROD-1909 Add check for inherited projects to ignore fake potential conflicts

ZackarySantana:
DEVPROD-1909 

### Description
This adds a simple check for projects that inherit from another repo project. It also adds testing to the function because there previously wasn't any.

### Testing
Add unit testing


model/project_ref.go
```diff
@@ -2390,7 +2392,8 @@ func (p *ProjectRef) GetGithubProjectConflicts() (GithubProjectConflicts, error)
 	}
 
 	for _, conflictingRef := range matchingProjects {
-		if conflictingRef.Id == p.Id {
+		// If this is the same project ref or this project ref will inherit from it, it is not conflicting.
+		if conflictingRef.Id == p.Id || conflictingRef.Id == p.RepoRefId {
```
hadjri, line 2396: I believe we want the opposite of the 2nd boolean statement here, i.e. `if conflictingRef.Id == p.Id || conflictingRef.RepoRefId == p.Id`, because `RepoRefId` is the ID of the repo that a project ref inherits its values from.

That way, if we are modifying some settings on the repo, we skip checking conflicts for all `matchingProjects` that are project refs attached to the repo, since checking for conflicts isn't relevant/helpful.

model/project_ref.go
```diff
@@ -117,6 +117,8 @@ type ProjectRef struct {
 	ContainerSizeDefinitions []ContainerResources `bson:"container_size_definitions,omitempty" json:"container_size_definitions,omitempty" yaml:"container_size_definitions,omitempty"`
 	ContainerSecrets         []ContainerSecret    `bson:"container_secrets,omitempty" json:"container_secrets,omitempty" yaml:"container_secrets,omitempty"`
 
+	// RepoRefId is the project ref id that this project ref tracks, if any (i.e. this is a branch
```
hadjri, line 120: This field is actually the ID of the repo ref (which is a separate document stored in a [separate collection](https://gist.github.com/hadjri/ffd4ad5aeb0ddc6d51d99a160dd04745#file-repo_ref-go-L18)) that a project is attached to, so we shouldn't refer to it as a project ref.

I think we can remove the parentheses statement, the first statement is succinct enough.

hadjri (changes requested):
In addition to the unit tests, can we replicate the error in staging and confirm that it goes away with this change?

hadjri (changes requested):
Can we get some evidence of this working on staging + fixing unit tests?

ZackarySantana:
Screenshot 2023-11-09 at 9 18 00 AM

I tried adding a patch definition without my changes, it failed with conflicts. With my changes is screenshot above!

ZackarySantana:
Child:
Screenshot 2023-11-09 at 9 19 07 AM

hadjri (approved):

PR #7184, opened 2023-10-31
DEVPROD-715 Prevent repotracker from being enabled when remote path (config) is not set

ZackarySantana:
DEVPROD-715

### Description
When remote path (config) is not set, the repotracker should not be allowed to be enabled.

### Testing
Add unit tests


ablack12:
nit: can you re-write config title to better fit guidelines? https://docs.google.com/document/d/1LJHPjLb_bUZxv-l_9Y61kitzAg7u-13I5DjOhZ-deBA/edit#heading=h.i9oftc5069w9
(i.e. title should describe what the PR does, not what it does not do)

model/project_ref.go
```diff
@@ -49,15 +49,19 @@ type ProjectRef struct {
...
+	Private     *bool  `bson:"private,omitempty" json:"private,omitempty" yaml:"private"`
+	Restricted  *bool  `bson:"restricted,omitempty" json:"restricted,omitempty" yaml:"restricted"`
+	Owner       string `bson:"owner_name" json:"owner_name" yaml:"owner"`
+	Repo        string `bson:"repo_name" json:"repo_name" yaml:"repo"`
+	Branch      string `bson:"branch_name" json:"branch_name" yaml:"branch"`
+
+	// RemotePath is the path to the Evergreen config file.
+	RemotePath       string `bson:"remote_path" json:"remote_path" yaml:"remote_path"`
+	PatchingDisabled *bool  `bson:"patching_disabled,omitempty" json:"patching_disabled,omitempty"`
+
```
ablack12, line 63: nit: I'd remove this new line -- makes sense for the flag to be grouped with the flags. It's okay if comments bump against each other in structs.

model/project_ref.go
```diff
@@ -1977,6 +1981,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
 	var err error
 	switch section {
 	case ProjectPageGeneralSection:
+		if utility.FromBoolPtr(p.RepotrackerDisabled) && p.RemotePath == "" {
```
ablack12, line 1984: we can use p.IsRepotrackerDisabled() 
Also we'd want to check if it's enabled here, not disabled, right?

model/project_ref.go
```diff
@@ -1977,6 +1981,19 @@ func SaveProjectPageForSection(projectId string, p *ProjectRef, section ProjectP
...
+			var pRef *ProjectRef
+			pRef, err = FindBranchProjectRef(projectId)
```
ablack12, line 1986: We don't want to do a query to see what's _currently_ true in the DB -- this is the update function, so we're going to overwrite whatever  currently is in the DB with whatever p.RemoteFile is. So we should just be able to error here, since this is the bad state (i.e. the repotracker is enabled and config file isn't set).

ablack12 (changes requested):

model/project_ref.go
ZackarySantana: Okay sounds good, I wasn't sure about the bumping

rest/data/project_settings.go
```diff
@@ -241,6 +241,10 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
 			}
 		}
 
+		if !mergedSection.IsRepotrackerDisabled() && mergedSection.RemotePath == "" {
```
ablack12, line 244: Actually since project repotrackers are enabled by default, this should probably also check IsEnabled

rest/data/project_settings.go
ablack12, line 1: It looks like there aren't tests anymore for this change; can you add something?

ablack12 (changes requested):

rest/data/project_settings.go
ZackarySantana: Yes! Just wanted to make sure these changes looked proper beforehand. I'll add them next

rest/data/project_settings.go
ablack12: I think it's okay if their settings get in a bad state, since they haven't enabled the project yet it won't actually matter until they enable it (at which case we will then error). The project technically starts in a bad state (since it starts with repotracker enabled and config undefined).

model/project_ref.go
```diff
@@ -49,15 +49,17 @@ type ProjectRef struct {
...
-	DisplayName            string              `bson:"display_name" json:"display_name,omitempty" yaml:"display_name"`
-	Enabled                bool                `bson:"enabled,omitempty" json:"enabled,omitempty" yaml:"enabled"`
-	Private                *bool               `bson:"private,omitempty" json:"private,omitempty" yaml:"private"`
-	Restricted             *bool               `bson:"restricted,omitempty" json:"restricted,omitempty" yaml:"restricted"`
-	Owner                  string              `bson:"owner_name" json:"owner_name" yaml:"owner"`
-	Repo                   string              `bson:"repo_name" json:"repo_name" yaml:"repo"`
-	Branch                 string              `bson:"branch_name" json:"branch_name" yaml:"branch"`
-	RemotePath             string              `bson:"remote_path" json:"remote_path" yaml:"remote_path"`
-	PatchingDisabled       *bool               `bson:"patching_disabled,omitempty" json:"patching_disabled,omitempty"`
+	DisplayName string `bson:"display_name" json:"display_name,omitempty" yaml:"display_name"`
```
ablack12, line 52: this whitespace change isn't ideal, we'd prefer to have the variable types lined up like they were before. I assume your editor auto did this, any way to tweak the settings?

model/project_ref.go
```diff
@@ -1921,6 +1923,15 @@ func FindProjectRefs(key string, limit int, sortDir int) ([]ProjectRef, error) {
 	return projectRefs, err
 }
 
+// CanEnableRepotracker checks if the project can enable the repotracker
```
ablack12, line 1926: copy pasta comment function title

model/project_ref.go
```diff
@@ -1921,6 +1923,15 @@ func FindProjectRefs(key string, limit int, sortDir int) ([]ProjectRef, error) {
...
+// with the current state of their project settings.
+func (p *ProjectRef) ValidateModifiedRepotracker() error {
```
ablack12, line 1928: in both places we call this function, we also check if the repotracker is enabled. It'd make sense to just check that in here.

model/project_ref.go
```diff
@@ -1921,6 +1923,15 @@ func FindProjectRefs(key string, limit int, sortDir int) ([]ProjectRef, error) {
...
+	if p.Enabled && p.RemotePath == "" {
+		return errors.Errorf("Your project '%s' must have a confguration set to enable repotracker", p.Identifier)
```
ablack12, line 1930: i think it's not quite right to return an error here as though it's written to the user, since it's a helper function. I'd suggest just "remote path can't be empty for enabled repotracker project" or something (it's possible that they didn't just enable the repotracker or something for example, they may have just enabled the project or just removed the config, so this comment might be confusing)

model/project_ref.go
ablack12, line 1928: The user can't modify the repotracker, that's a job that tracks the repo. We also don't know when calling this function that it's been modified. Could we tweak the title?

ablack12 (changes requested):

model/project_ref.go
```diff
@@ -49,15 +49,17 @@ type ProjectRef struct {
...
+	Enabled     bool   `bson:"enabled,omitempty" json:"enabled,omitempty" yaml:"enabled"`
+	Private     *bool  `bson:"private,omitempty" json:"private,omitempty" yaml:"private"`
+	Restricted  *bool  `bson:"restricted,omitempty" json:"restricted,omitempty" yaml:"restricted"`
+	Owner       string `bson:"owner_name" json:"owner_name" yaml:"owner"`
+	Repo        string `bson:"repo_name" json:"repo_name" yaml:"repo"`
+	Branch      string `bson:"branch_name" json:"branch_name" yaml:"branch"`
+	// RemotePath is the path to the Evergreen config file.
+	RemotePath       string `bson:"remote_path" json:"remote_path" yaml:"remote_path"`
+	PatchingDisabled *bool  `bson:"patching_disabled,omitempty" json:"patching_disabled,omitempty"`
+	// RepotrackerDisabled is if repotracker is disabled (i.e. true = repotracker disabled. false = repotracker enabled).
```
ablack12, line 62: can we remove this comment? It seems redundant

model/project_ref.go
ablack12: Ah okay fair. In that case can you move RemotePath above DisplayName?

ablack12 (approved):
🚀 thanks for iterating on this with me!

PR #7187, opened 2023-11-01
DEVPROD-706 avoid checking for module with unpatched project

ablack12:
DEVPROD-706 

### Description
Removed checking for module in updatePatchModule.

Originally changed it to use the correct project, but realized we always check if the module exists in order to pass in the module name (and even if we didn't, we'd error further down the line), so it's really not necessary.

### Testing
Tested on staging with a new module: https://spruce-staging.corp.mongodb.com/patch/6542705e9dbe320c2eb26024/configure/changes

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,7 +32,7 @@ var (
 	BuildRevision = ""
 
 	// ClientVersion is the commandline version string used to control auto-updating.
-	ClientVersion = "2023-10-02"
+	ClientVersion = "2023-10-30"
 
 	// Agent version to control agent rollover.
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -32,7 +32,7 @@ var (
 	BuildRevision = ""
 
 	// ClientVersion is the commandline version string used to control auto-updating.
-	ClientVersion = "2023-10-02"
+	ClientVersion = "2023-10-30"
```
hadjri, line 35: Nit: this can be reverted

hadjri (approved):

PR #7189, opened 2023-11-01
DEVPROD-964: fix ambiguous skip CI interaction with Dependabot PRs

Kimchelly:
DEVPROD-964

### Description
Some Dependabot PRs were not running Evergreen PR tests because the PR description occasionally happened to include the label `skip ci` in the commit list (I didn't type out the actual `skip ci` label or else Evergreen will skip CI testing on this PR...). For example, in [this PR](https://github.com/evergreen-ci/spruce/pull/1985), if you toggle the commit list, you'll see one of the commits in the upgrade has the `skip ci` label, so Evergreen skips CI testing for the Dependabot PR (even though `skip ci` was only intended for the dependency repo's testing, not ours).

I considered removing the ability to put `skip ci` in the description since it could be very ambiguous (using the PR description to skip CI is not something that I've seen other CI system offer, likely for this very reason). Unfortunately, some users are already relying on putting the `skip ci` phrase in their PR description. Rather than entirely remove the ability to `skip ci` using the description, I tried narrowing it down so Evergreen will only respect `skip ci` if it's very early on in the description, which will mitigate the Dependabot PR issue.

### Testing
Tested in the sandbox to verify that `skip ci` still works in the PR title + beginning of description, but putting the `skip ci` phrase much further down in the description does not skip.

### Documentation
Updated docs.

### DIFF
--- a/docs/Project-Configuration/Github-Integrations.md
+++ b/docs/Project-Configuration/Github-Integrations.md
@@ -52,8 +52,8 @@ If you used `evergreen keep-definitions`, then `evergreen reset-definitions` wil
 #### Skip CI Testing
 
 Sometimes you may want to avoid having Evergreen create patches (perhaps because the work is in progress, or testing isn't relevant yet). 
-Simply including `[skip-ci]` or `[skip ci]` in your PR title or description will prevent us from creating a patch (both from commits and `evergreen retry` comments) 
-until the label is removed and a new comment or commit is pushed.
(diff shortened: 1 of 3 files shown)


ablack12 (approved):
I think it'd be cool to put a comment or tiny ADR to explain that this is replicating circle's behavior (or a comment on https://jira.mongodb.org/browse/EVG-16366)

Kimchelly:
I wrote the ADR, requesting re-review since it's more than a trivially small change.

ablack12 (approved):
Lovely thank you!

PR #7193, opened 2023-11-02
DEVPROD-990 Remove unsetFunctionVarsDisabled flag

malikchaya2:
DEVPROD-990

### Description
This was a temporary flag and is no longer needed.


agent/command.go
```diff
@@ -250,21 +250,19 @@ func (a *Agent) runCommand(ctx context.Context, tc *taskContext, logger client.L
 		tc.taskConfig.Expansions.Put(key, newVal)
 	}
 	defer func() {
-		if !tc.unsetFunctionVarsDisabled || tc.taskConfig.Project.UnsetFunctionVars {
```
Kimchelly, line 253: Since the project YAML `unset_function_vars` now is a no-op, is there a ticket to delete it?

rest/route/host_agent.go
```diff
@@ -245,7 +245,6 @@ func (h *hostAgentNextTask) Run(ctx context.Context) gimlet.Responder {
 	}
 
 	setNextTask(nextTask, &nextTaskResponse)
-	nextTaskResponse.UnsetFunctionVarsDisabled = flags.UnsetFunctionVarsDisabled
```
Kimchelly, line 248: Removing this in the same change as the agent change will temporarily cause a reversion in behavior for old agents still on version 2023-11-02, and those agents will stop unsetting function vars until the new version is deployed. If `UnsetFunctionVarsDisabled` is not set with the next task request anymore (i.e. the app server updates) but the old agent is still running (e.g. because it's running a task group), the 2023-11-02 agent will interpret this to mean it should keep the function vars (i.e. it will not unset them when the function ends). This could potentially break someone's tasks if they've been updated to depend on the new behavior (i.e. unset function vars) and it temporarily reverts back to keeping function vars.

If we want to do this more safely without the temporary reversion to not unsetting function vars, then we have to roll out the agent change first (to always unset function vars) and then delete the admin setting in a separate commit once the agents have all deployed.

Kimchelly (changes requested):
The field still needs to be removed from:
* [The admin settings page](https://github.com/evergreen-ci/evergreen/blob/0e4178aa0c4bd3b7690a351fca870efcb31bf41a/service/templates/admin.html#L250-L256)
* [Local Evergreen data](https://github.com/evergreen-ci/evergreen/blob/1a97b4a87851cab70bd3f28ad131da9da7fb427a/testdata/local/admin.json#L4)
* [Smoke test data](https://github.com/evergreen-ci/evergreen/blob/main/smoke/internal/testdata/db/admin.json#L1)

agent/command.go
malikchaya2: This PR deletes that too.

Kimchelly: Thanks for clarifying! In that case, we should also delete it from the [self-tests YAML](https://github.com/evergreen-ci/evergreen/blob/1a97b4a87851cab70bd3f28ad131da9da7fb427a/self-tests.yml#L1).

rest/route/host_agent.go
malikchaya2: good catch, thank you! 
opened https://github.com/evergreen-ci/evergreen/pull/7206

agent/command.go
malikchaya2: done

Kimchelly (approved):

PR #7200, opened 2023-11-03
DEVPROD-948: Return repo level filters with project

sophstad:
DEVPROD-948

### Description
- `FindMergedProjectRef` should return a merged list of repo- and project-level Parsley filters (similar to how variables are merged) instead of the default "if null then use repo settings" pattern

### Testing
- Update unit test

### Documentation
N/A

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -1029,7 +1029,25 @@ func mergeBranchAndRepoSettings(pRef *ProjectRef, repoRef *RepoRef) (*ProjectRef
 	reflectedRepo := reflect.ValueOf(repoRef).Elem().Field(0) // specifically references the ProjectRef part of RepoRef
 
 	util.RecursivelySetUndefinedFields(reflectedBranch, reflectedRepo)
+
+	// Include Parsley filters defined at repo level alongside project filters.
+	mergeParsleyFilters(pRef, repoRef)
+
 	return pRef, err
+}
+
+func mergeParsleyFilters(pRef *ProjectRef, repoRef *RepoRef) {
+	if repoRef.ParsleyFilters == nil || len(repoRef.ParsleyFilters) == 0 {
+		return
+	}
+
+	if pRef.ParsleyFilters == nil {
+		pRef.ParsleyFilters = []ParsleyFilter{}
+	}
+
+	for _, filter := range repoRef.ParsleyFilters {
+		pRef.ParsleyFilters = append(pRef.ParsleyFilters, filter)
+	}
 }
 
 func setRepoFieldsFromProjects(repoRef *RepoRef, projectRefs []ProjectRef) {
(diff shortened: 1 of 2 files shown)


minnakt (approved):
makes sense to me!

model/project_ref.go
```diff
@@ -1029,9 +1029,25 @@ func mergeBranchAndRepoSettings(pRef *ProjectRef, repoRef *RepoRef) (*ProjectRef
...
+func mergeParsleyFilters(pRef *ProjectRef, repoRef *RepoRef) {
+	if repoRef.ParsleyFilters == nil || len(repoRef.ParsleyFilters) == 0 {
+		return
+	}
+
+	if pRef.ParsleyFilters == nil {
+		pRef.ParsleyFilters = []ParsleyFilter{}
+	}
+
+	pRef.ParsleyFilters = append(pRef.ParsleyFilters, repoRef.ParsleyFilters...)
```
ablack12, line 1048: Can we write a test where just repoRef filters are set? I'm wondering if we're going to end up accidentally duplicating them because we'll have set the pRef's undefined fields at line 1031

sophstad: Good catch, switching the order of `mergeParsleyVariables` and `RecursivelySetUndefinedFields` fixes this.

model/project_ref.go
```diff
@@ -1028,10 +1028,26 @@ func mergeBranchAndRepoSettings(pRef *ProjectRef, repoRef *RepoRef) (*ProjectRef
...
+	// Include Parsley filters defined at repo level alongside project filters.
+	mergeParsleyFilters(pRef, repoRef)
+
 	util.RecursivelySetUndefinedFields(reflectedBranch, reflectedRepo)
+
 	return pRef, err
 }
 
+func mergeParsleyFilters(pRef *ProjectRef, repoRef *RepoRef) {
+	if repoRef.ParsleyFilters == nil || len(repoRef.ParsleyFilters) == 0 {
```
ablack12, line 1040: We can just check length here; if it's nil, this will be 0 implicitly.

model/project_ref_test.go
```diff
@@ -138,6 +152,20 @@ func TestFindMergedProjectRef(t *testing.T) {
 	assert.True(t, mergedProject.WorkstationConfig.ShouldGitClone())
 	assert.Len(t, mergedProject.WorkstationConfig.SetupCommands, 1)
 	assert.Equal(t, "random2", mergedProject.TaskAnnotationSettings.FileTicketWebhook.Endpoint)
+	assert.Len(t, mergedProject.ParsleyFilters, 2)
+
+	projectRef.ParsleyFilters = []ParsleyFilter{}
```
ablack12, line 157: I'd suggest making a separate test for this, or adding a comment, to make it clear this is testing the MergeParsleyFilters behavior.

ablack12 (approved):

sophstad:
Spruce failure is more E2E flake

PR #7206, opened 2023-11-06
DEVPROD-990 Always unset function vars 

malikchaya2:
DEVPROD-990 

### Description
This is part one of[7193]( https://github.com/evergreen-ci/evergreen/pull/7193)  to prevent [this](https://github.com/evergreen-ci/evergreen/pull/7193#discussion_r1383492570).


Kimchelly (approved):
LGTM mod delete the agent tests as well, since unsetting function vars is the only allowed behavior after this change. (I imagine they won't pass if left in.)

PR #7207, opened 2023-11-06
DEVPROD-1007: create REST route to list generated tasks

Kimchelly:
DEVPROD-1007

### Description
Given a generated task, it's easy to get the generator task that created it, but finding the inverse (starting with the generator task and deriving the generated tasks) is somewhat difficult without just getting all the tasks in a version and scanning them. I created a route to make it easier to find the generated tasks.

I opted not to expose all the task info possible since this could return a lot of tasks. Users can `GET /tasks/{task_id}` after getting the task IDs from this route if they're interested in the more specific details on specific generated tasks.

Before this is merged, this will need a new sparse index on `generated_by` since we do not have one currently. I built one in staging ([link to indexes](https://cloud.mongodb.com/v2/5f6906add65b40252b525f21#/metrics/replicaSet/6064ec8d9b5eaf1d22cbd2f3/explorer/mci/tasks/indexes)). Given that the new index took 1/10th of the disk space compared to the other task indexes, I'm inclined to say it's not a problem to build it.

* Create new REST route to get generated tasks.
* Fix some swaggo docs for `/users/permissions` which showed the wrong input parameters (looked like a copy-paste error).

### Testing
* Added unit tests.
* Tested in staging that the returned the relevant info about tasks.

### Documentation
Swaggo docs:
image


malikchaya2 (approved):
🎉

rest/route/service.go
```diff
@@ -241,6 +241,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/tasks/{task_id}/tests").Version(2).Get().Wrap(addProject, viewTasks).RouteHandler(makeFetchTestsForTask(env, sc))
 	app.AddRoute("/tasks/{task_id}/tests/count").Version(2).Get().Wrap(addProject, viewTasks).RouteHandler(makeFetchTestCountForTask())
 	app.AddRoute("/tasks/{task_id}/sync_path").Version(2).Get().Wrap(requireUser).RouteHandler(makeTaskSyncPathGetHandler())
+	app.AddRoute("/tasks/{task_id}/generated_tasks").Version(2).Get().Wrap(requireUser, viewTasks).RouteHandler(makeGetGeneratedTasks())
```
ablack12, line 244: Rather than creating a new route, did we consider adding this as an optional parameter to the existing get task route?

ablack12: I was thinking it could be part of the APITask struct and just conditionally populated, similar to how we have the APITaskArgs here https://github.com/evergreen-ci/evergreen/blob/0e4178aa0c4bd3b7690a351fca870efcb31bf41a/rest/model/task.go#L369 
I mostly ask this because having optional parameters is pretty consistent with a lot of behavior, and could be less bespoke. I'm fine with whichever though since this works and is kind of its own thing, just wanted to mention it.

Kimchelly: I guess we could return an APITask but it would be kind of awkward still because you would end up with a long list of JSON docs with most of the fields empty, which would make an argument for "why not just return the entire task". Anyways, I was hoping to keep this route's results simple. Since this returns a strict subset of the fields in a task, we can always adapt it later if need arises to return all the results + change the route name.

PR #7208, opened 2023-11-06
DEVPROD-961 improve build notification failed task attachments

ablack12:
DEVPROD-961
### Description
As frequently happens, caching only some statuses makes it easy to have bugs happen -- because we're caching these task docs to be used wherever, and task docs aren't large, there's no reason to just grab some statuses.

Also used the correct GetDisplayStatus, as referenced by Kim in the ticket comment. Not too important if all statuses are caught by the "toHumanSpeak" since there's a catch all.

### DIFF
--- a/trigger/build.go
+++ b/trigger/build.go
@@ -62,13 +62,13 @@ func (t *buildTriggers) Fetch(ctx context.Context, e *event.EventLogEntry) error
 
 	var tasks []task.Task
 	if e.EventType == event.BuildGithubCheckFinished {
-		query := db.Query(task.ByBuildIdAndGithubChecks(t.build.Id)).WithFields(task.StatusKey, task.DependsOnKey)
+		query := db.Query(task.ByBuildIdAndGithubChecks(t.build.Id))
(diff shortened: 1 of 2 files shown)


trigger/build.go
```diff
@@ -62,13 +62,13 @@ func (t *buildTriggers) Fetch(ctx context.Context, e *event.EventLogEntry) error
 
 	var tasks []task.Task
 	if e.EventType == event.BuildGithubCheckFinished {
-		query := db.Query(task.ByBuildIdAndGithubChecks(t.build.Id)).WithFields(task.StatusKey, task.DependsOnKey)
```
malikchaya2, line 65: This looks good, but is there any way we can test it to see if it fixes the bug?

malikchaya2 (changes requested):

trigger/build.go
ablack12: image

(I'll add unscheduled to statuses we skip)

malikchaya2 (approved):

PR #7210, opened 2023-11-06
DEVPROD-1450 Add patchable false to swaggo task

bsamek:
DEVPROD-1450

### Description

The swaggo task pushes to production pine, so it should only run on commit.

### Testing

The task shouldn't appear in the PR!


self-tests.yml
```diff
@@ -612,6 +612,7 @@ tasks:
           RUN_TEST: "TestEnvironmentSuite/TestInitDB"
           MONGO_CREDS_FILE: testdata/test_auth.yml
   - name: generate-api-docs
+    patchable: false
```
Kimchelly, line 615: If this doesn't run for patches anymore, can we have a separate task for swaggo format linting checks in PRs?

Kimchelly (changes requested):

self-tests.yml
bsamek: Oh shoot, sorry, that was my original intent. I also added an execution number to the non-latest part, so you can see the file in a patch build if you want. I also added it to the PR definitions.

Kimchelly (approved):

PR #7211, opened 2023-11-06
DEVPROD-2230 Add nil check for taskoutput

khelif96:
DEVPROD-2230

### Description
This line panics because test logs can sometimes be nil so this just adds a check.


julianedwards (approved):

julianedwards:
Nit: update the description.

PR #7213, opened 2023-11-06
DEVPROD-1661 add /data partition to task end details

ybrill:
[DEVPROD-1661](https://jira.mongodb.org/browse/DEVPROD-1661)

### Description
We need to hold on to the device name of the device mounted on /data so we can make graphs for it in Honeycomb. If this catches on we can add graphs for other disks as well.

An alternative could be to make the choice of disk to graph configurable on the distro. That seems like it'd be a lot of work and it would be subject to breaking if something changes about the distro that causes the disks to be named differently.

### Testing
[An example graph](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/FpiYpiTnevV?omitMissingValues)


agent/agent.go
```diff
@@ -602,6 +602,9 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
 	defer span.End()
 	tc.traceID = span.SpanContext().TraceID().String()
 
+	tc.dataDisk, err = dataPartition(ctx)
```
Kimchelly, line 605: Seems like this should use `tskCtx` instead of `ctx`.

Kimchelly, line 605: Since this is part of task setup, it would be preferable if we could 1. absorb this into `setupTask` for consistency with how the rest of the task config is initialized and 2. write a unit test to protect against regressions. One of the big pain points in the agent refactor was fixing bugs in task setup due to the slow initialization of `tc` over the course of several steps and the fact that the critical setup steps were untested.

Probably same thing with the metrics server setup below this, but I won't argue that strongly for moving a piece of code not directly related to the intended change.

agent/otel.go
```diff
@@ -42,6 +43,7 @@ const (
 	packageName    = "github.com/evergreen-ci/evergreen/agent"
 	traceSuffix    = "build/OTelTraces"
 	maxLineSize    = 1024 * 1024
+	dataMountpoint = "/data"
```
Kimchelly, line 46: Since this comes from a distro setting, rather than hardcoding our assumed distro settings into the agent implementation, can we have this be a distro setting and pull this from the distro info? Having the filesystem info live in one place would be nice.

agent/agent.go
```diff
@@ -602,6 +602,9 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
...
+	grip.Error(errors.Wrapf(err, "getting device name for '%s' partition", dataMountpoint))
```
Kimchelly, line 606: Since this step comes after the task logger is already set up, it would be preferable to send this log to the task logs, since those are things we can easily inspect for debugging if it fails to get the partitions.

Kimchelly (changes requested):

agent/agent.go
ybrill: this is using the setupCtx now

agent/otel.go
ybrill: done

agent/agent.go
ybrill: sent the error to the agent logs

agent/agent.go
ybrill: moved to setupTask.
I'm unsure about a unit test. To make it actually get the device names for whatever its running on seems fragile. It could be broken up to just test the loop that pulls out device names from the partitions list, if that seems valuable.

ybrill:
@Kimchelly do you know if we're still adding things to the legacy distro page? Or should I not bother and just add it to the new page in Spruce?

ablack12:
@ybrill we have a plan to deprecate the legacy distro page very soon but haven't yet so I think it'd be risky to only define it on the Spruce page. Alternatively you could hold off adding to the distro page until the deprecation is done (which we could prioritize).

ybrill:
@ablack12 I can add it to the legacy page too. It shouldn't be a big deal.

ybrill:
Added to the legacy distros page
Screenshot 2023-11-07 at 3 23 40 PM

agent/agent.go
Kimchelly: Hmm, I guess the question here would be, if this functionality broke for some reason (e.g. we upgraded gopsutil, which has in the past broken existing behavior), would anyone notice, ask for a fix, etc.? If it's something that we or users expect to come to rely upon, a simple unit test to verify that it can collect device names would be useful.

agent/task_context.go
```diff
@@ -519,3 +520,20 @@ func (tc *taskContext) getUserEndTaskResponse() *triggerEndTaskResp {
 
 	return tc.userEndTaskResp
 }
+
+func (tc *taskContext) deviceNames(ctx context.Context) error {
```
Kimchelly, line 524: nit: since this both gets the device names and populates the field, could we call this `getDeviceNames`/`populateDeviceNames` or something?

agent/task_context.go
```diff
@@ -519,3 +520,20 @@ func (tc *taskContext) getUserEndTaskResponse() *triggerEndTaskResp {
...
+	if tc.taskConfig == nil || tc.taskConfig.Distro == nil {
+		return nil
+	}
+
+	partitions, err := disk.PartitionsWithContext(ctx, false)
```
Kimchelly, line 529: nit: consider returning early if `tc.taskConfig.Distro.Mountpoints` is empty to avoid asking for the partitions unnecessarily.

Kimchelly (commented):
Only small requested changes, but other housekeeping questions:

* Is there a ticket tracking adding the mountpoints to track into Spruce?
* Does this work on Windows/MacOS as well? (Maybe the unit test would cover this?)

agent/task_context.go
ybrill: done

agent/task_context.go
ybrill: done

ybrill:
> Is there a ticket tracking adding the mountpoints to track into Spruce?

I was going to do it on this ticket after this PR closes out. I could make a separate ticket for it if that's preferred.

> Does this work on Windows/MacOS as well? (Maybe the unit test would cover this?)

Added a test for it and added test-agent for windows and macos in the patch.

agent/agent.go
ybrill: added a test

Kimchelly (approved):

PR #7214, opened 2023-11-07
DEVPROD-1450 Finish generated docs

bsamek:
DEVPROD-1450

### Description

This PR does three things

1. Finish documenting the routes. I'm not sure why I thought the remaining couldn't be documented.
2. Remove the use-case guide. I think these notes are redundant and not useful.
2. Remove the right-side table of contents. I think this makes the screen busy, and squeezes the contents horizontally.

### Testing

You can view the results in [staging](https://docs.devprod.staging.corp.mongodb.com/evergreen/API/REST-V2-Usage/), and compare it to [production](https://docs.devprod.prod.corp.mongodb.com/evergreen/API/REST-V2-Usage/).

### DIFF
--- a/docs/API/REST-V2-Usage.md
+++ b/docs/API/REST-V2-Usage.md
@@ -1,4 +1,6 @@
-import ApiDocMdx from '@theme/ApiDocMdx';
+---
+hide_table_of_contents: true
+---
 
 # REST v2 API
 
@@ -38,7 +40,9 @@ A returned object will always contain its complete list of fields. Any
 field that does not have an associated value will be filled with JSON's
 null value.
 
-## Generated API Docs
+## API Docs
+
+import ApiDocMdx from '@theme/ApiDocMdx';
 
 <ApiDocMdx id="evergreen-openapi" />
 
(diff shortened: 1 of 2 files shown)


bsamek:
@ablack12 I added you explicitly to this PR, since we talked over Slack, and you thought that the right-side TOC makes sense, but I think with the use case guide removed, it makes less sense, and clutters the page. I think this version is easier to use and easier to understand. I think this would also make DEVPROD-2228 unnecessary.

docs/API/REST-V2-Usage.md
```diff
@@ -51,163 +55,4 @@ null value.
...
-
-## Additional Endpoints
-
-### Users
-
-#### Endpoints
-
-##### Get Users for Role
-
-NOTE: These roles are not part of the OpenAPI spec, since they are part of another package.
```
ablack12, line 64: I feel like these were the routes you'd meant to keep up here https://github.com/evergreen-ci/evergreen/blob/739214c74f0eb04f1efe34d9fa9911aa6911f251/rest/route/service.go#L225, or were these never documented?

ablack12 (approved):
I'm good to get rid of the TOC, thanks for checking!

docs/API/REST-V2-Usage.md
bsamek: They weren't documented. Only GET /roles/<role_id>/users was documented https://github.com/evergreen-ci/evergreen/blob/ac7582937dbc56e46a925857c72e25a2cca77a69/docs/REST-V2-Usage.md#give-roles-to-user.

PR #7215, opened 2023-11-07
DEVPROD-971 Remove REST codegen tool

bynn:
DEVPROD-971

### Description
deleted the rest codegen tool since we do not use it


Kimchelly (changes requested):
Since the codegen'd files ([1](https://github.com/evergreen-ci/evergreen/blob/a77002508ba2d7a1b9286797b64b6547a5efbafd/rest/model/generated.go#L1), [2](https://github.com/evergreen-ci/evergreen/blob/a77002508ba2d7a1b9286797b64b6547a5efbafd/rest/model/generated_converters.go)) are no longer managed by a codegen tool, presumably we would have to manually edit those files if the need ever arose to change them. For some additional cleanup, it would be good to:
1. Remove the `// Code generated by rest/model/codegen.go. DO NOT EDIT.` comment.
2. Put a replacement comment at the top for future maintainers of what to do with them. In particular, clarifying that these models _used_ to be code-generated but are no longer (hence why they look out of place compared to the human-written code), and manual edits as needed are okay (and please don't replicate them anymore).
3. Potentially rename the files/structs/methods to `rest/model/generated_legacy.go`/ `rest/model/generated_converters_legacy.go` or something to mark them as legacy remnants of the REST codegen tool.

bynn:
Decided against renaming the files/structs/methods because we're allowing edits so it could be confusing if they are manually changed in the future but if you feel strongly about specific things like filenames I could still update them!

Kimchelly (approved):
Sounds good, and LGTM!

PR #7216, opened 2023-11-07
DEVPROD-974 Add project validation warnings for projects using ssh cloning for modules

malikchaya2:
DEVPROD-974

### Description

The original plan for this ticket was to add project level warning banners in the UI for any projects that use ssh cloning for modules. I decided against this for a few reasons. The biggest reason is that we are not currently set up to add project level warning banners programmatically in a way that reflects the real time state of the project config and in a way that doesn't interfere with a manually set up banner that may take higher precedence (it's a one banner per project setup right now). If we did indeed add a project banner if a project failed validation, we wouldn't really have a setup to scan the project config after it was fixed to see if it now passes and remove the banner. I think implementing that would be a bit of too much effort for what it is trying to accomplish and too much of a deviation from how project banners are intended to be used. I therefore opted for a validation warning instead. 

### Testing
Added to the test 

### Note
I will hold off merging this until an email is sent out to users.

### DIFF
--- a/validator/project_validator_test.go
+++ b/validator/project_validator_test.go
@@ -1485,7 +1485,7 @@ func TestCheckModules(t *testing.T) {
 			So(len(cm), ShouldEqual, 2)
 		})
 
-		Convey("An error should be returned when the module's repo is empty, in the deprecated format or invalid", func() {
+		Convey("An error should be returned when the module's repo is empty or invalid", func() {
 			project := &model.Project{
 				Modules: model.ModuleList{
 					model.Module{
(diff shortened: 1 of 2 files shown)


validator/project_validator_test.go
```diff
@@ -1501,7 +1511,8 @@ func TestCheckModules(t *testing.T) {
 					},
```
Kimchelly, line 1511: nit: remove the `// should fail` comments above, and/or split the last module (with the SSH clone format) into a separate test case. That way, we can know that the warning comes from the SSH format and not some other reason.

Kimchelly (approved):
Validation warning SGTM!

LGTM mod minor suggestion to split off the test case.

malikchaya2:
evergreen help

malikchaya2:
I'm closing this for now because we are not ready to merge these chanegs.

PR #7219, opened 2023-11-08
DEVPROD-957 increase GitHub installation page limit

ablack12:
DEVPROD-957

### Description
100 was too small a page limit here; increasing to something larger. I'm hesitant to say we should need pagination here; I don't expect us to have so many installations that we can't handle it in a call. 

### Testing
No testing needed


bynn (approved):
😠

PR #7220, opened 2023-11-08
DEVPROD-2285 Create Decomissioned Host tickets in DEVPROD project

khelif96:
DEVPROD-2285

### Description
Makes Decomissioned Host tickets go to the DEVPROD Jira project (Setting will be updated in admin settings)
Assigns to the runtimeEnvironments team
### Testing
https://jira.mongodb.org/browse/DEVPROD-2474

cc @brooookemiller

### DIFF
--- a/rest/route/host.go
+++ b/rest/route/host.go
@@ -577,7 +577,7 @@ func (rh *hostProvisioningOptionsGetHandler) Run(ctx context.Context) gimlet.Res
 
 // //////////////////////////////////////////////////////////////////////
 //
-// POST /rest/v2/host/{host_id}/disable
+// POST /rest/v2/hosts/{host_id}/disable
 type disableHost struct {
 	env evergreen.Environment
 
--- a/rest/route/host.go
+++ b/rest/route/host.go
@@ -577,7 +577,7 @@ func (rh *hostProvisioningOptionsGetHandler) Run(ctx context.Context) gimlet.Res
 
 // //////////////////////////////////////////////////////////////////////
 //
-// GET /rest/v2/host/{host_id}/disable
+// POST /rest/v2/host/{host_id}/disable
 type disableHost struct {
 	env evergreen.Environment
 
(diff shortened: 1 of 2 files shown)


rest/route/host.go
```diff
@@ -577,7 +577,7 @@ func (rh *hostProvisioningOptionsGetHandler) Run(ctx context.Context) gimlet.Res
 
 // //////////////////////////////////////////////////////////////////////
 //
-// GET /rest/v2/host/{host_id}/disable
+// POST /rest/v2/host/{host_id}/disable
```
bsamek, line 580: nit, it's `hosts`, not `host`

units/deco_host_notify.go
```diff
@@ -134,7 +136,10 @@ func (j *decoHostNotifyJob) Run(_ context.Context) {
 		Summary:     fmt.Sprintf("investigate automatically decommissioned host '%s'", j.Host.Id),
 		Type:        "Incident",
 		Description: strings.Join(descParts, "\n"),
-		Components:  []string{"Evergreen"},
+		Components:  []string{"Static Host Management"},
+		Fields: map[string]interface{}{
+			fieldAssignedTeams: []map[string]string{{"id": runtimeEnvironmentsId}}, // Assigned Teams: Runtime Environments
```
bsamek, line 141: I thought you had said "value", not "id", in Slack.

bsamek (approved):
lgtm with a nit and a question

units/deco_host_notify.go
khelif96: @brooookemiller mentioned that its actually id since the value we are using for runtimeEnvironments is an id

ablack12: I had no idea we made tickets this way. It looks like we could set up a Splunk alert to create a ticket like this which I think would be way better than hardcoding it. I get the sense that this is a quick fix to get us over to the DEVPROD project but could you make a follow up ticket?

khelif96: Yes that was the intent behind this. Great idea! https://jira.mongodb.org/browse/DEVPROD-2486

PR #7222, opened 2023-11-09
DEVPROD-998 Add single retry option to agent's local status endpoint

hadjri:
DEVPROD-998

### Description
This change allows tasks to automatically reset themselves if a given command fails by allowing commands to have a new field, `retry_on_failure `, which will automatically reset the failed task once it completes (up to one time - though this can be arbitrarily increased). 

### Testing
[confirmed](https://spruce-staging.corp.mongodb.com/task/sandbox_git_env_test_bynntask_patch_81dd13ebce8b573651e1d613febc2fb82d8887e9_65530f01b23736664991a98a_23_11_14_06_09_18/logs?execution=1) in staging that a task failing on a command with the flag set will retry exactly once
### Documentation
Updated docs describing the new field


model/task/task.go
```diff
@@ -2788,6 +2790,31 @@ func (t *Task) SetResetWhenFinished() error {
...
+// SetResetWhenFinishedWithInc requests that a task (that was marked to
+// automatically reset when finished via the agent status server) reset itself
+// when finished. It will also increment the number of automatic resets the task
+// has performed.
+func (t *Task) SetResetWhenFinishedWithInc() error {
+	if t.ResetWhenFinished {
+		return nil
+	}
+	t.ResetWhenFinished = true
+	t.NumAutomaticResets = t.NumAutomaticResets + 1
```
Kimchelly, line 2802: nit: these should usually go after the update since the update could error and allow later logic involving this task to act on bad data.

model/task_lifecycle.go
```diff
@@ -798,6 +803,26 @@ func MarkEnd(ctx context.Context, settings *evergreen.Settings, t *task.Task, ca
...
+	}
+	taskToReset := t
+	if t.IsPartOfDisplay() {
+		dt, err := t.GetDisplayTask()
+		if err != nil {
+			return errors.Wrap(err, "getting display task")
+		}
+		taskToReset = dt
+	}
+	if taskToReset.NumAutomaticResets < evergreen.MaxAutomaticRestarts {
```
Kimchelly, line 818: Would be nice if there was a way to send an HTTP error back if they try requesting more than the maximum auto-restarts.

Kimchelly (changes requested):
I didn't read through the entire PR, but have some food for thought on the UX:

* While this would work in normal operating conditions, it seems like sending the flag with the end task response wouldn't always be work in cases like if the task system fails due to something catastrophic like the agent getting OOM-killed (since the app server will fix up the dead task). It would be reasonable from a user perspective ask for it to restart even if something goes wrong on the host's end.
* I would make an argument for having a REST endpoint or a command (that pings a REST endpoint) to set the task to auto-restart on completion as soon as the user sends their request to fix the above. The only downside is that with the current proposed change, that option poses a weird situation since we have a local agent endpoint that basically just forwards a user request over to the Evergreen app server. If we wanted just a REST endpoint (and no command), the only argument I could see for keeping the local agent route (instead of asking users to ping the endpoint themselves) is to work around users not having a bot user to auth into the REST API, but I don't think it would make sense long-term to have the local agent routes become a way to conveniently bypass REST API auth. (I don't have an answer to the weird auth situation, but it is a thought that the agent's local REST API wasn't really intended for this kind of thing.)
* Is there a way that a user can identify (without reading all the task logs) that a task auto-restarted because of this option? Can the auto-restart display the source (i.e. the task) in the event logs or the UI?
* Does this restart if the user aborts the task? Should it? (I don't have an answer here, but my thought is that we could argue either way as long as we're clear about it.)
* Would it make sense to have an ADR describing this feature's UX + the overall technical design choices made? (If there weren't any technical tradeoffs to make, you can feel free to skip this.)

agent/command.go
```diff
@@ -215,6 +215,11 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
 			commandSpan.SetStatus(codes.Error, "running command")
 			commandSpan.RecordError(err, trace.WithAttributes(tc.taskConfig.TaskAttributes()...))
 			commandSpan.End()
+			if commandInfo.AutoRetry {
+				if err2 := a.comm.MarkTaskToRestart(ctx, tc.task); err2 != nil {
```
Kimchelly, line 219: Should the command still auto-restart if it can't fail the task? Let's say a shell.exec fails, but it's in one of those commands that can't fail the task (say it fails in the pre block, but the YAML has set `pre_error_fails_task: false`). The task then continues on happily and ends in success, but still restarts at the end of teh task, even though the command's failure had no bearing on the task's status.

I guess if I were to phrase the question, is auto-restart meant to kick in when a command fails, or when a command fails and will subsequently cause the task to fail?

model/task/task.go
```diff
@@ -2798,6 +2804,37 @@ func (t *Task) SetResetWhenFinished() error {
...
+// when finished. It will also increment the number of automatic resets the task
+// has performed.
+func (t *Task) SetResetWhenFinishedWithInc() error {
+	if t.ResetWhenFinished {
+		return nil
+	}
+	err := UpdateOne(
+		bson.M{
+			IdKey:      t.Id,
+			AbortedKey: bson.M{"$ne": true},
```
Kimchelly, line 2818: On the same train of thought that a command may fail but may not stop the task immediately, does this still work if a command fails, and the task is aborted later on?

Let's say (same as before) a shell.exec fails in pre but this time the user configured `pre_error_fails_task: true` in their YAML. The task then skips to post. Then later on, the user aborts the task, but the task commands ignore it because aborts don't stop the post commands from running. After the task ends, will it auto-restart because it was originally marked to auto-restart, or will it not because the task was aborted later on? I imagine for reliability, when checking to reset the task, we would have to check if the task was aborted right at the moment when we're deciding whether to reset, rather than here where we're not sure if the task may eventually abort later.

Kimchelly (changes requested):
The idea itself of adding a command parameter SGTM.

Can we test this in staging + add unit tests?

model/project.go
```diff
@@ -545,6 +545,9 @@ type PluginCommandConf struct {
 	// Vars defines variables that can be used within commands.
 	Vars map[string]string `yaml:"vars,omitempty" bson:"vars,omitempty"`
 
+	// AutoRetry indicates whether the command should be retried if it fails.
```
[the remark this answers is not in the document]
hadjri: Yes 🤦

globals.go
```diff
@@ -179,8 +179,10 @@ const (
 	MergeTestFailed    = "failed"
 	EnqueueFailed      = "failed to enqueue"
 
-	// maximum task (zero based) execution number
+	// MaxTaskExecution is the maximum task (zero based) execution number
 	MaxTaskExecution = 9
+	// MaxAutomaticRestarts is the maximum number of automatic restarts allowed for a task
+	MaxAutomaticRestarts = 1
```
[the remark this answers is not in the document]
hadjri: Sure, going with restart

agent/command.go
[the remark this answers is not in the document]
hadjri: Sure, makes sense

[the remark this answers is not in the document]
hadjri: Sgtm, added some tests for those cases

hadjri: > is auto-restart meant to kick in when a command fails, or when a command fails and will subsequently cause the task to fail?
My intention is the latter. In the case that `pre_error_fails_task` is set, the `runCommand` actually will not fail because this [conditional](https://gist.github.com/hadjri/e2c7bb1b22b261488d0c4737058a6c6c#file-command-go-L304) won't execute (which is derived from PreErrorFailsTask)

rest/route/agent.go
```diff
@@ -317,6 +317,53 @@ func (h *newPushHandler) Run(ctx context.Context) gimlet.Responder {
...
+	}
+	taskToReset := t
+	if t.IsPartOfDisplay() {
+		dt, err := t.GetDisplayTask()
+		if err != nil {
+			return gimlet.MakeJSONInternalErrorResponder(errors.Wrapf(err, "getting display task for execution task '%s'", h.taskID))
+		}
+		taskToReset = dt
+	}
+	if taskToReset.NumAutomaticResets < evergreen.MaxAutomaticRestarts {
```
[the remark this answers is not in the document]
hadjri: Sure, I'll change that case to a 400 bad request now since we're no longer failing the command if marking to restart fails, instead we just log the error in the task logs

hadjri:
Tested new implementation and confirmed working on staging with:
- a [single task](https://spruce-staging.corp.mongodb.com/task/sandbox_git_env_test_test_restart_on_fail_patch_81dd13ebce8b573651e1d613febc2fb82d8887e9_655658369dbe32608cd03318_23_11_16_17_58_38/logs?execution=1)
- a [display task](https://spruce-staging.corp.mongodb.com/task/sandbox_cron_test_display_displayTask1_patch_81dd13ebce8b573651e1d613febc2fb82d8887e9_655658369dbe32608cd03318_23_11_16_17_58_38?execution=1&sorts=STATUS%3AASC)
- a [single host task group](https://spruce-staging.corp.mongodb.com/task/sandbox_release_test_restart_on_fail_patch_81dd13ebce8b573651e1d613febc2fb82d8887e9_655658369dbe32608cd03318_23_11_16_17_58_38/logs?execution=1)

model/task/task.go
hadjri: You're right that this only prevents the task from being auto restarted if the command with `retry_on_fail` set fails due to an abort rather than failing organically, and if a user sends in a manual abort in between the duration where the command fails and the post-task commands complete, the task will still auto-restart.

I'd say that this would be the behavior we should expect though - if a command with `retry_on_fail` fails organically, that should serve as the agreed criteria for restarting the task, so even if the user sends in a manual abort in the small window where the post commands are running (which doesn't seem like something that would happen too often), the task should still auto-restart.

model/project.go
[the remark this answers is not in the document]
Kimchelly: I think this comment got missed, it's relevant to docs + follow-up work.

agent/internal/client/base_client.go
```diff
@@ -1059,16 +1059,16 @@ func (c *baseCommunicator) CreateInstallationToken(ctx context.Context, td TaskD
 	return token.Token, nil
 }
 
-// MarkTaskToRestart
-func (c *baseCommunicator) MarkTaskToRestart(ctx context.Context, td TaskData) error {
+// MarkFailedTaskToRestart will mark the task to be automatically reset upon completion
```
Kimchelly, line 1062: nit: automatically restart

agent/internal/client/base_client.go
```diff
@@ -1059,16 +1059,16 @@ func (c *baseCommunicator) CreateInstallationToken(ctx context.Context, td TaskD
...
+func (c *baseCommunicator) MarkFailedTaskToRestart(ctx context.Context, td TaskData) error {
 	info := requestInfo{
 		method:   http.MethodPost,
 		taskData: &td,
 	}
 	info.setTaskPathSuffix("reset")
```
Kimchelly, line 1068: nit: "reset" -> "restart"

agent/agent_test.go
```diff
@@ -794,6 +794,115 @@ post:
...
+
+	s.NoError(err)
+	s.Equal(evergreen.TaskFailed, s.mockCommunicator.EndTaskResult.Detail.Status)
+
+	s.NoError(s.tc.logger.Close())
+	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
+		"Running task commands",
+		"Running command 'shell.exec' (step 1 of 1)",
+		"Running task commands failed",
+		"Command is set to retry on failure, marking task to restart upon completion.",
```
Kimchelly, line 823: I mentioned it in another spot, but it seems like this test + the `WithoutPreErrorFailsTask` is testing the opposite of what we'd expect it to do (i.e. not auto-restart for a command that continues on error).

agent/command.go
Kimchelly: That makes sense, but [this test](https://github.com/evergreen-ci/evergreen/blob/1e9f3acd58c48176a5d4ce1fcd1fe30066f5b864/agent/agent_test.go#L823) (which is implicitly using the default of `pre_error_fails_task: false`) seems like it's expecting (and actually sees) the opposite, because it's claiming that the task will auto-restart even though the pre command is not one that can fail the task.

It seems like you would have to check `canFailTask` explicitly before marking the task for restart, no? `canFailTask` only returns on error at the block level, not at the individual command level.

model/task/task.go
Kimchelly: Hmm, I think that it would make more sense to obey the abort though no matter when it happens. It seems reasonable that abort would mean stop and don't auto-restart at all since the user is trying to stop their patch. It would also be complex from the perspective of understanding the feature's behavior, if users see their task have to split their thinking for an aborted task between "did the task abort and cause the auto-restart, or did the auto-restart already get triggered earlier, and then I aborted it later". It may also be a little difficult because they have to abort once, then potentially unschedule/abort again to get the patch to really stop.

agent/agent_test.go
```diff
@@ -794,6 +794,115 @@ post:
...
+
+	s.NoError(err)
+	s.Equal(evergreen.TaskFailed, s.mockCommunicator.EndTaskResult.Detail.Status)
+
+	s.NoError(s.tc.logger.Close())
+	checkMockLogs(s.T(), s.mockCommunicator, s.tc.taskConfig.Task.Id, []string{
+		"Running pre-task commands",
+		"Running command 'shell.exec' (step 1 of 1)",
+		"Running pre-task commands failed",
+		"Command is set to retry on failure, marking task to restart upon completion.",
```
Kimchelly, line 862: One small improvement we could make here is along with checking the task logs (which we kind of only do to check the user-facing output or in cases there's no other visible side effect), we could add a flag to the mock that gets set when `MarkFailedTaskToRestart`is called, then check that the flag is set.

As an example of that model, the mock's `LastMessageSent` field does something like this in that every time `UpdateLastMessageTime` is called, it updates the field, so you can use it to test that it was updated.

Kimchelly (commented):
Left some more comments, and a couple follow-up questions:

* Re the Shrub thing, is that going to be part of a follow-up ticket or are you just gonna open a separate PR with this ticket?

* I also saw this on the task metadata panel of the 1st execution for the tasks in staging. Will this be fixed/updated to show the correct message (in a follow-up ticket)?
> This task will restart when all of the tasks in its task group have finished running.
* I dunno if you still plan to do it, but since the behavior overall SGTM, an ADR would be helpful!

I'm also gonna be out after today, so if this isn't merged, I'm gonna pass it along to someone else to review.

Kimchelly:
I'm gonna be out next week, so re-requesting evg-app so someone else can take a look. I don't have any further comments other than the ones I already put up.

ablack12 (commented):
Dropping a comment so I can review once Kim's changes are addressed.

hadjri:
> Re the Shrub thing, is that going to be part of a follow-up ticket or are you just gonna open a separate PR with this ticket?



> I also saw this on the task metadata panel of the 1st execution for the tasks in staging. Will this be fixed/updated to show the correct message (in a follow-up ticket)?

Created a ticket [DEVPROD-2776](https://jira.mongodb.org/browse/DEVPROD-2776) to fix this UI bug

> I dunno if you still plan to do it, but since the behavior overall SGTM, an ADR would be helpful!

Created ADR!

model/project.go
```diff
@@ -546,6 +546,9 @@ type PluginCommandConf struct {
 	// Vars defines variables that can be used within commands.
 	Vars map[string]string `yaml:"vars,omitempty" bson:"vars,omitempty"`
 
+	// RetryOnFailure indicates whether the task should be retried if this command fails.
+	RetryOnFailure bool `yaml:"retry_on_failure,omitempty" bson:"retry_on_failure,omitempty"`
```
malikchaya2, line 550: [non-blocking]
It looks like this was changed from `auto_retry` to `retry_on_failure` since the PR was created. Can you please update the PR description when you have a chance? 
Thank you!

model/task/task.go
hadjri: Added checks such that if a task / display task / single host TG is marked for automatic restart but has _also_ since been aborted, forgo the restart, which should protect against this scenario.

model/project.go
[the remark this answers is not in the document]
hadjri: Created shrub PR and linked it a comment below.

model/project.go
hadjri: Done

agent/agent_test.go
```diff
@@ -794,6 +794,115 @@ post:
...
+        params:
+          script: exit 0
+`
+	s.setupRunTask(projYml)
+
+	nextTask := &apimodels.NextTaskResponse{
+		TaskId:     s.tc.task.ID,
+		TaskSecret: s.tc.task.Secret,
+	}
+	_, _, err := s.a.runTask(s.ctx, s.tc, nextTask, false, s.testTmpDirName)
```
ablack12, line 891: Placeholder comment -- is there a flag we should call here?

agent/command.go
```diff
@@ -215,6 +215,12 @@ func (a *Agent) runCommandOrFunc(ctx context.Context, tc *taskContext, commandIn
 			commandSpan.SetStatus(codes.Error, "running command")
 			commandSpan.RecordError(err, trace.WithAttributes(tc.taskConfig.TaskAttributes()...))
 			commandSpan.End()
+			if commandInfo.RetryOnFailure {
+				logger.Task().Info("Command is set to retry on failure, marking task to restart upon completion.")
```
ablack12, line 219: retry once** on failure perhaps? Just so its clear why we wouldn't see this another time

agent/internal/client/base_client.go
```diff
@@ -1058,3 +1058,18 @@ func (c *baseCommunicator) CreateInstallationToken(ctx context.Context, td TaskD
 
 	return token.Token, nil
 }
+
+// MarkFailedTaskToRestart will mark the task to be automatically restart upon completion
```
ablack12, line 1062: nit: "to automatically restart" or "to be automatically restarted"

globals.go
```diff
@@ -625,6 +627,7 @@ const (
 	MergeTestRequester          = "merge_test"           // Evergreen commit queue
 	AdHocRequester              = "ad_hoc"               // periodic build
 	GithubMergeRequester        = "github_merge_request" // GitHub merge queue
+	AutoResetRequester          = "automatic_reset"
```
ablack12, line 630: I think this might not be the right place for this; requesters are meant to associate with types of versions. Seems like this is more a different kind of activator, so might live better here:
https://github.com/evergreen-ci/evergreen/blob/a0dbc9709914466344b06d1c3d04d341b7d142c6/globals.go#L234

model/task/db.go
```diff
@@ -2880,6 +2882,8 @@ func abortAndMarkResetTasks(ctx context.Context, filter bson.M, taskIDs []string
 			AbortedKey:           true,
 			AbortInfoKey:         AbortInfo{User: caller},
 			ResetWhenFinishedKey: true,
+			// this function should override the automatic restart key if it is set
+			IsAutomaticRestartKey: false,
```
ablack12, line 2886: todo: what's this for?

model/task/task.go
```diff
@@ -245,9 +245,13 @@ type Task struct {
 	// ResetWhenFinished indicates that a task should be reset once it is
 	// finished running. This is typically to deal with tasks that should be
 	// reset but cannot do so yet because they're currently running.
-	ResetWhenFinished       bool  `bson:"reset_when_finished,omitempty" json:"reset_when_finished,omitempty"`
-	ResetFailedWhenFinished bool  `bson:"reset_failed_when_finished,omitempty" json:"reset_failed_when_finished,omitempty"`
-	DisplayTask             *Task `bson:"-" json:"-"` // this is a local pointer from an exec to display task
+	ResetWhenFinished       bool `bson:"reset_when_finished,omitempty" json:"reset_when_finished,omitempty"`
+	ResetFailedWhenFinished bool `bson:"reset_failed_when_finished,omitempty" json:"reset_failed_when_finished,omitempty"`
+	// NumAutomaticRestarts is the number of times the task has been programmatically restart via a failed agent command.
```
ablack12, line 250: programmatically restarted**

model/task/task.go
```diff
@@ -1365,40 +1369,6 @@ func UnscheduleStaleUnderwaterHostTasks(ctx context.Context, distroID string) (i
 	return info.Updated, nil
 }
 
-// LegacyDeactivateStepbackTasksForProject deactivates and aborts any scheduled/running tasks
-// for this project that were activated by stepback.
-// TODO: remove as part of EVG-17947
```
ablack12, line 1370: good catch !

model/task_lifecycle.go
```diff
@@ -2468,10 +2477,19 @@ func checkResetSingleHostTaskGroup(ctx context.Context, t *task.Task, caller str
...
 	shouldReset := false
+	isAutoRestart := false
 	for _, tgTask := range tasks {
 		if tgTask.ResetWhenFinished {
 			shouldReset = true
 		}
+		if tgTask.IsAutomaticRestart {
+			// Do not restart the task group if it contains a task marked for auto restart
+			// that has since been aborted
+			if tgTask.Aborted {
```
ablack12, line 2488: Instead of considering aborted in each of these places, can we just unset IsAutomaticRestart when aborting? Or is that more complicated? (Sorry if this has already been asked)

model/task_lifecycle.go
```diff
@@ -2468,10 +2477,19 @@ func checkResetSingleHostTaskGroup(ctx context.Context, t *task.Task, caller str
...
+				return nil
+			}
+			isAutoRestart = true
```
ablack12, line 2491: can we just set requester here to AutoReset? Then we don't need the variable

model/task/task.go
```diff
@@ -2797,6 +2757,38 @@ func (t *Task) SetResetWhenFinished() error {
...
+			"$set": bson.M{
+				ResetWhenFinishedKey:  true,
+				IsAutomaticRestartKey: true,
+			},
+			"$inc": bson.M{
+				NumAutomaticRestartsKey: 1,
+			},
+		},
+	)
+	if err != nil {
```
ablack12, line 2784: This is going to return an error for aborted tasks, since its not going to match anything. Is that okay?

model/task/task.go
```diff
@@ -2797,6 +2757,38 @@ func (t *Task) SetResetWhenFinished() error {
...
+		return err
+	}
+	t.ResetWhenFinished = true
```
ablack12, line 2787: I think we should just drop these updates, since the inc might be wrong unless we include it in the query (which we might want to do if this can be set in more than one command, since it could race), and its inconsistent with the other retry functions (and the fields aren't used)

rest/route/agent.go
```diff
@@ -317,6 +317,65 @@ func (h *newPushHandler) Run(ctx context.Context) gimlet.Responder {
...
+		taskToRestart = dt
+	}
+	// If the task is a display task that has already been marked for an automatic restart
+	// by another execution task, we don't want to mark it for a restart again nor do we
+	// want to error out since the display task has not been automatically restarted yet.
+	if taskToRestart.IsAutomaticRestart {
+		return gimlet.NewJSONResponse(struct{}{})
+	}
+
+	if taskToRestart.NumAutomaticRestarts < evergreen.MaxAutomaticRestarts {
```
ablack12, line 366: nit: I'd reorder this to avoid the nested if statement, and having the error away from the if statement:

```
if we've exceeded restarts {
    return error 
}
if SetResetWhenFinished errors {
   return error
}
return no error
```

ablack12 (changes requested):

ablack12:
I didn't read Kim's comments carefully so apologies if there's some overlap!

agent/command.go
hadjri: Updated it to a formatted string that references `MaxAutomaticRestarts` as this feature's been set up to change the number of allowed restarts moving forward if necesary

agent/internal/client/base_client.go
hadjri: Done.

globals.go
hadjri: Sounds good, moved!

model/task/task.go
hadjri: Fixed.

model/task/task.go
hadjri: Ty!

rest/route/agent.go
hadjri: Done.

model/task/task.go
hadjri: Fine by me, removed

model/task_lifecycle.go
hadjri: True! fixed.

agent/agent_test.go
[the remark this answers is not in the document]
hadjri: True, here the mock flag `TaskShouldRetryOnFail` should not get set here since pre_error_fails_task is false, so I added a line checking that.

model/task/task.go
hadjri: I think we should return an error there to inform in the logs [here](https://gist.github.com/hadjri/aed360cffb5610ec81dd256cae5c354a#file-command-go-L221) that the auto marking for reset did not go through. That being said I updated this function with a more descriptive error if the task has been aborted, rather than propagating a vague "matched no documents" error down to the user

model/task_lifecycle.go
hadjri: Yeah, this logic was put in place because there is a valid case where we [abort a version/build and mark all of its tasks to reset when finished](https://gist.github.com/hadjri/0b567f31c51f78548838c6d691504751#file-db-go-L2853-L2867), but we also need a way to ensure that if we abort a task marked for auto restart before it gets a chance to finish, the scheduled auto restart is ignored.

In lieu of this logic I changed the task abort DB updates to also ensure `ResetWhenFinished` and `IsAutomaticRestart` are marked false, as we should expect those requests to stop the task in its tracks even if it has been marked to reset when finished by an automatic restart. This should achieve the same effect.

model/task/db.go
hadjri: This was to ensure that clicking "restart version & abort in progress tasks" still gets respected if some of the tasks were marked for auto restart, but hadn't completed yet.  

This isn't necessary anymore given the changes I made to the `SetAborted` DB update so I'm removing this

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1546,6 +1546,38 @@ Full gitignore syntax is explained
...
 failure stepback.
 
+### Auto restarting tasks upon failure
+
+A given command can be configured to automatically restart the task upon failure
+by setting the `retry_on_failure` field on the command to true. The automatic
+restart will process after the command has failed and the task has completed its
+subsequent post task commands.
+
+The retry will only occur if the task has _not_ been aborted, and that the failing command would have caused the overall task
```
ablack12, line 1556: nit: and "if" rather than "that"

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1546,6 +1546,38 @@ Full gitignore syntax is explained
...
+- The failing command exists in the `pre` or `post` section of the task and `pre_error_fails_task`
+    or `post_error_fails_task` are (respectively) unset
+- The failing command exists in the `setup_group`, `setup_task`, or `teardown_task` sections of the task
+and `setup_group_can_fail_task`, `setup_task_can_fail_task`, or `teardown_task_can_fail_task` are (respectively) unset
+
+Otherwise, once a command with `retry_on_failure` set to true fails, the task will restart
+when it completes, regardless of the failure type.
+
+This is only recommended for commands that are known to be flaky, or fail intermittently.
+In order to prevent overuse of this feature, the number of times a single
```
ablack12, line 1567: I'm thinking we might just want to straight bold this

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1546,6 +1546,38 @@ Full gitignore syntax is explained
...
+task can be automatically restarted on failure is limited to 1 time.
+
+An example is:
+
+``` yaml
```
ablack12, line 1572: love this!

docs/decisions/2023-11-20_auto_retry_tasks.md
```diff
@@ -0,0 +1,44 @@
+# 2023-11-20 Auto Retry Tasks
+
+* status: accepted
+* date: 2023-11-20
+* authors: Malik Hadjri
+
+## Context and Problem Statement
```
ablack12, line 7: Can we include the ticket number in here?

docs/decisions/2023-11-20_auto_retry_tasks.md
```diff
@@ -0,0 +1,44 @@
...
+
+* Introduce a new agent command named something like `auto.retry` which will make the agent hit an app server endpoint
+that will mark the task for restart upon failure
+
+* Add a new boolean parameter on the generic command struct, such that if that command fails and the flag is set, we send a
+request to the app server to mark the task for restart upon completion
+
+## Decision Outcome
+
+Option 1 was rejected because it would require interested teams to request service users with the needed permissions to hit the
```
ablack12, line 32: I'm actually a little concerned about this outcome because it seems like this is a reason we're pushing for Self Service Users. Are there other reasons this would cause issues?

ablack12 (approved):
LGTM with some comments!

docs/decisions/2023-11-20_auto_retry_tasks.md
hadjri: Ideally Self Service Users will be getting some permissions that we'd carefully check don't have risk to cause potential harm before allowing them to be autoprovisioned (just read operations. + annotation write operations).

Whereas a service user that has the power to do something as high-level as restarting a task could be more likely to be misused by using it for other routes.

PR #7225, opened 2023-11-09
DEVPROD-982 Point test repository to Evergreen

ZackarySantana:
DEVPROD-982

### Description
When testing patches, we test against a random other repository. This changes to test against Evergreen.

### Testing
Existing tests

### DIFF
--- a/model/patch_lifecycle_test.go
+++ b/model/patch_lifecycle_test.go
@@ -32,13 +32,13 @@ import (
 
 var (
 	patchTestConfig = testutil.TestConfig()
-	configFilePath  = "testing/mci.yml"
+	remotePath      = "model/testdata/project.yml"
 	patchedProject  = "mci-config"
-	patchedRevision = "3578f10b95fb82183662387048b268c54fac50fb"
+	patchedRevision = "17746cb296670f53ee326deb83ac8cc4dffe55dd"
 	patchFile       = "testdata/patch2.diff"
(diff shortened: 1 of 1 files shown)


hadjri (approved):

model/testdata/patch2.diff
```diff
@@ -20,11 +20,16 @@ index 0000000..1111111 100644
...
      push_path: linux
      push_bucket: downloads.mongodb.org
-@@ -23,3 +33,4 @@ buildvariants:
+@@ -21,4 +30,5 @@ buildvariants:
+     has_debugsymbols: true
      test_flags: --continue-on-failure
    tasks:
-   - name: taskOne
+-  - name: taskOne
+\ No newline at end of file
```
hadjri, line 31: Can we remove this line and the one below?

hadjri (changes requested):

model/testdata/patch2.diff
ZackarySantana: I can try, this was autogenerated by Git for reference.

ZackarySantana: I can't do that since it changes the file version (it's a thing in the patch at the top that git uses to verify the patch successfully went through). I think it's best to leave as is. I'm reverting the "Shorter diff file" which is the one that failed because git couldn't apply the patch (and verify it is correct).

hadjri (approved):

PR #7226, opened 2023-11-09
DEVPROD-1003 Allow skip missing when local_files_include_filter for s3.put

ZackarySantana:
DEVPROD-1003

### Description
It was unclear when s3.put with `local_files_include_filter` is used, if it will succeed when no files are found. We have an 'optional' parameter that does that but is incompatable with `local_files_include_filter`. By default, when using `local_files_include_filter` it does succeed when no files are found (as intended).

### Testing
I tested s3.put with `local_files_include_filter` with no files and it succeeds, I tested it with some files and it also succeeds. I tested it by causing it to throw other errors (invalid aws credentials) and it fails. 

Updated existing tests relating to validation for s3.put to test specifically what they should. Before, some of these tests would pass regardless of what they tried to test- because of other invalid options (e.g. testing for conflicting configuration options to produce an error but an error was produced anyways because the test was missing an aws key).

### Documentation
Updated the documentation around `local_files_include_filter` so users do not get confused on if it should or will succeed when no files are found.


config.go
```diff
@@ -35,7 +35,7 @@ var (
 	ClientVersion = "2023-11-01"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-11-05"
+	AgentVersion = "2023-11-8"
```
Kimchelly, line 38: mega nit: `08`, the format is always YYYY-MM-DD.

Kimchelly (changes requested):
If we didn't have a prior good reasoning for having it like this and the command works with this combination, then I'm fine with changing s3.put to allow `optional: true` + `local_files_include_filter`. Adding a unit test for this combination + testing in staging that it works as expected would be good next steps.

On the topic of testing, is [this](https://github.com/evergreen-ci/evergreen/blob/f143d7fa036647cd7f923454cd1440f283626f23/agent/command/s3_put_test.go#L59-L72) supposed to pass, even post-change? Seems like it's testing the wrong thing if it still fails after this change.

config.go
ZackarySantana: I'll fix this, as for the test suppose to pass, I made the tests more specific and ran a patch and it was passing because it is missing an AWS key- not because of what it is testing for. I'll push a commit with correct tests that should have that particular one failing, then I will fix the test to represent the new behavior.

ZackarySantana: I contacted the user and did some more testing. I revamped the PR description with all the details!

docs/Project-Configuration/Project-Commands.md
```diff
@@ -1093,7 +1093,8 @@ Parameters:
 -   `display_name`: the display string for the file in the Evergreen UI
 -   `local_files_include_filter`: used in place of local_file, an array
     of gitignore file globs. All files that are matched - ones that
-    would be ignored by gitignore - are included in the put.
+    would be ignored by gitignore - are included in the put. If no
+    files are found, the task continues execution.
 -   `local_files_include_filter_prefix`: an optional path to start
     processing the LocalFilesIncludeFilter, relative to the working
```
Kimchelly, line 1099: nit: I know this is unrelated to the change, but can we also fix this while we're at it?
```suggestion
    processing the `local_files_include_filter`, relative to the working
```

Kimchelly (commented):
So to double-check my understanding - the user thought you had to use `optional` + `local_files_include_filter` to get the command to pass when it uploaded no files, but that's already the behavior of `local_files_include_filter`, regardless of `optional`.

In that case, it sounds like two different decisions were made with regards to local file fields, where `local_file` can be used with `optional` to make it optional (but it's by default not optional). However, `local_file_include_filter` is always optional and there's no way to change make it non-optional (hence `optional:true` has no effect and `optional: false` won't work without a code change). If that's the case, can we:
1. Clarify this distinction in the `optional` docs to make it clear that `optional` is meant to be combined with `local_file`, because it's default non-optional, whereas `local_files_include_filter` is always optional no matter what?
2. Update the validation error to clarify that `local_files_include_filter` + `optional` doesn't work because it's always optional anyways?

The other changes LGTM otherwise. (and nice tests!)

ZackarySantana:
I removed a test because it was testing if parsing a string as a boolean would cause an error (which it did, it wouldn't even get to test what it said it was testing because it stops there).

Kimchelly (approved):

PR #7229, opened 2023-11-09
DEVPROD-712: fix display task update race

Kimchelly:
DEVPROD-712

### Description
There were two user reports of display task status bugs, but they have different causes. This PR addresses the original bug report (I'm going to spin off the second one as a separate ticket.)

The bug basically boiled down to two execution tasks (let's call them ET0 and ET1) finishing within ~2 seconds of each other and racing to update the display task's status. The order of operations resulting in the bug is:

* ET0 and ET1 are running.
* ET0 finishes and marks itself complete.
* ET0 determines that the display task status should be "started" (because ET1 is not finished).
* ET1 finishes and marks itself complete.
* ET1 determines that the display task status should be "success" (because all execution tasks are finished and successful).
* ET1 updates the DB display task status to "success".
* ET0 updates the DB display task status back to "started".

Based on the above example, if any other execution task intervenes in between the status calculation and the DB update, the display task update for one execution task can produce (wrong) final display task status because its update was based on stale information. To fix this, I made the update operation atomic (by checking the display task status hasn't changed in the update condition) - if the original status it saw is not the same as its current status in the DB, that means the update is acting on stale execution task information, so it should try again. The race is transient and not that common, so just retrying a couple times is likely good enough to resolve the occasional race.

* Fix race between conflicting execution task updates for parent display task.
* Refactor display task status calculation and update into a helper function.
* Add temporary log (to be removed in a follow-up PR) to confirm that the race is handled properly.

### Testing
* Added a unit test for concurrent display task updates.
* Verified in staging that basic display task updates still work.

### Documentation
N/A

### DIFF
--- a/model/task_lifecycle.go
+++ b/model/task_lifecycle.go
@@ -2298,17 +2298,62 @@ func UpdateDisplayTaskForTask(t *task.Task) error {
 		return errors.Errorf("task '%s' is not a display task", dt.Id)
 	}
 
-	var timeTaken time.Duration
-	var statusTask task.Task
+	// kim: TODO: explain why we need to retry here (i.e. this logic can race
+	// between end tasks, and there's no backup measure that will fix up the
+	// status. There's no exact number of times that this update can fail
+	// theoretically, but number of retries is intentionally small because:
+	// - There are not a lot of different task statuses.
+	// - Status are not frequently updated.
+	// - Once they reach a non-transient state such as finished.
+	// kim: TODO: add basic race test for UpdateDisplayTaskForTask
+	const maxUpdateAttempts = 3
+	var updatedDisplayTask *task.Task
+	for i := 0; i < maxUpdateAttempts; i++ {
+		updatedDisplayTask, err = tryUpdateDisplayTaskAtomically(*dt)
+		if err == nil {
+			break
+		}
+		grip.Debug(message.WrapError(err, message.Fields{
+			"message":                      "failed to update display task due to concurrent update contention",
+			"execution_task":               t.Id,
+			"display_task_id":              dt.Id,
+			"original_display_task_status": dt.Status,
+			"updated_display_task_status":  updatedDisplayTask.Status,
+			"attempt_num":                  i,
+			"ticket":                       "DEVPROD-712",
+		}))
+		if i >= maxUpdateAttempts-1 {
+			return err
+		}
+	}
+
+	// kim: NOTE: it's okay to not set the display task info on the original display
+	// task because it's not an input parameter or a returned output.
+	if !dt.IsFinished() && updatedDisplayTask.IsFinished() {
+		event.LogTaskFinished(dt.Id, dt.Execution, dt.GetDisplayStatus())
+		// kim: NOTE: this logged at 5:32:07.
+		grip.Info(message.Fields{
+			"message":   "display task finished",
+			"task_id":   dt.Id,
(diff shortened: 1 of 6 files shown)


model/task_lifecycle.go
```diff
@@ -2282,33 +2282,93 @@ func UpdateDisplayTaskForTask(t *task.Task) error {
...
+				"display_task_id": t.DisplayTaskId,
+			})
+			return errors.Errorf("display task not found for task '%s'", t.Id)
+		}
+		if !originalDisplayTask.DisplayOnly {
+			return errors.Errorf("task '%s' is not a display task", originalDisplayTask.Id)
+		}
+
+		updatedDisplayTask, err = tryUpdateDisplayTaskAtomically(*originalDisplayTask)
+		if err == nil {
```
hadjri, line 2323: If an execution task is operating on stale data I'd expect err to still be nil, so would this just be checking for a DB operation / read error?

hadjri (commented):
I'm a bit unsure why the retry is necessary - in the given scenario:
> ...
> ET1 updates the DB display task status to "success".
> ET0 updates the DB display task status back to "started".

Simply having the update of ET0 no-op would be sufficient correct?

Kimchelly:
It's correct in this one example to not retry, but I don't think it's sufficient in general because we have to assume that tasks can revert out of the finished state if it restarts. For example, say one execution task is ending the task (to set it to success) but at the same time, the user restarted the display task. In the above scenario, you could modify it by adding a third actor (the user restarting the failed tasks), and the correct state to be in would be "started", so neither of the updates to finish should be valid.

model/task_lifecycle.go
Kimchelly: If it's operating on stale data, the UpdateOne should fail because [it will modify zero documents](https://github.com/mongodb/anser/blob/2a589a5299e680bcd5014909df16209967010365/db/wrapper.go#L206-L208).

hadjri:
> It's correct in this one example to not retry, but I don't think it's sufficient in general because we have no way of forcing an ordering of operations between such that the last thread that updated the execution tasks (i.e. the one with the most up-to-date view of the execution tasks) is the one that updates the display task. For example, we have to assume that tasks can revert out of the finished state if it restarts. Say one execution task is ending the task (to set it to success) but at the same time, the user restarted the display task.
> 
> In the above scenario, you could replace ET0 finishing the task (which updates the execution task, then updates the display task) with a "user restarted failed tasks in the display task" (which also updates the execution tasks, then updates the display task). The restarting thread will fail to update once (because ET1 already updated the status to success), but must try again because the restart has to set some execution tasks back to undispatched/started.
> 
> With a restart + execution task finishing, this would happen:
> 
> * ET is running.
> * User restarts failed tasks under the display task (we'll call this thread R).
> * ET finishes and marks itself complete. Now it has to update the display task.
> * ET determines that the display task status should be "failed" (because some other execution task failed).
> * R restarts all the execution tasks under the display task. Now it has to update the display task.
> * R determines that the display task status should be "started" (because it restarted the failed execution tasks).
> * ET updates the DB display task status to "failed".
> * R tries updating the DB display task back to "started". This fails once because ET just updated it, so it tries again and on the second try, it successfully sets it to "started".

Gotcha, alright that does make sense. Thanks for the detail 👍

hadjri (approved):

PR #7231, opened 2023-11-09
DEVPROD-959 Remove UpdateBlockedDependencies from stale/stranded task cleanup

hadjri:
DEVPROD-959

### Description
`ClearAndResetStrandedContainerTask` and `FixStaleTask` redundantly update the blocked dependencies of the task, because their following logic will always either:

1. update the blocked dependencies again in the end task logic
2. immediately reset the task, which renders updating blocked dependencies unnecessary.
### Testing
`ClearAndResetStrandedContainerTask` already has cases that check for proper behavior of blocked task dependencies, but added them to the test cases for `FixStaleTask`.  I also did check that commenting out the end task logic where blocked dependencies get updated causes the tests to fail.

### DIFF
--- a/model/task_lifecycle_test.go
+++ b/model/task_lifecycle_test.go
@@ -5260,6 +5260,11 @@ func TestResetStaleTask(t *testing.T) {
 			require.NoError(t, err)
 			require.NotZero(t, dbVersion)
 			assert.Equal(t, evergreen.VersionCreated, dbVersion.Status, "version status should be updated for restarted task")
+
+			dependencyTask, err := task.FindOne(db.Query(task.ById("dependencyTask")))
+			require.NoError(t, err)
+			assert.False(t, dependencyTask.DependsOn[0].Unattainable)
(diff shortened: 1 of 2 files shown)


model/task_lifecycle_test.go
```diff
@@ -5260,6 +5260,11 @@ func TestResetStaleTask(t *testing.T) {
 			require.NoError(t, err)
 			require.NotZero(t, dbVersion)
 			assert.Equal(t, evergreen.VersionCreated, dbVersion.Status, "version status should be updated for restarted task")
+
+			dependencyTask, err := task.FindOne(db.Query(task.ById("dependencyTask")))
```
Kimchelly, line 5264: nit: this can just be condensed down to `task.FindOneId`.

Kimchelly, line 5264: Also have to `require.NotNil(t, dependencyTask)` or else L5266 can panic.

Same with the check below this.

Kimchelly (approved):
LGTM mod testing nits.

PR #7232, opened 2023-11-10
DEVPROD-940: Expose RepotrackerError field from ProjectRef

minnakt:
DEVPROD-940

### Description
This PR exposes the `RepotrackerError` field from `ProjectRef`.

I only added this field to `ProjectRef` and not `RepoRef` because fixing the repotracker error involves updating the data in the `repo_revisions` collection. After inspecting production data, it appears that the `repo_revisions` collection only has entries corresponding to projects (attached or unattached) and does not have any entries corresponding to repos.

### Testing
- Augmented GraphQL tests


ablack12 (approved):
We also may need to confirm there's a way for users to fix the project if it gets in the bad state; I don't remember if we added that to the new UI.

PR #7236, opened 2023-11-13
DEVPROD-1000 Allow empty set-module patches

hadjri:
DEVPROD-1000

### Description
`set-module` actually does already work the way that was requested in the ticket (as it will [override](https://gist.github.com/hadjri/a061cb18107bd6043cb91da240ccf223#file-git-go-L686-L711) the manifest module revision if it exists), with the caveat that a diff must be staged on the module repo first. 

 I don't see any issue with allowing empty module patches as we already allow empty regular patches, so I allowed set-module to allow a patch to use the user's local checked out module revision regardless of a diff present.

Also deleted a block of unused code that I stumbled across for a struct that we don't ever save to the DB

### Testing
Ran `evergreen -c ~/evergreen-staging.yml set-module -m evergreen -i 6551caf69dbe328b3cf21b34` in staging with no staged evergreen changes for a sandbox patch which has evergreen as one of its modules.

[Confirmed](https://evergreen-staging.corp.mongodb.com/task_log_raw/sandbox_git_env_test_bynntask_patch_81dd13ebce8b573651e1d613febc2fb82d8887e9_6551caf69dbe328b3cf21b34_23_11_13_07_15_08/0?type=E#L44) in staging to confirm a task cloned the module revision from set-module and overrided the revision listed in the manifest
### Documentation
In order to make the nuances of module hash hierarchies clearer, I updated the docs on the project configuration and project commands side to make them clearer.

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -1100,18 +1100,6 @@ func generateId(name string, projectIdentifier string, projBV *BuildVariant, rev
 		v.CreateTime.Format(build.IdTimeLayout))
 }
 
-var (
-	// bson fields for the project struct
-	ProjectIdentifierKey    = bsonutil.MustHaveTag(Project{}, "Identifier")
-	ProjectPreKey           = bsonutil.MustHaveTag(Project{}, "Pre")
-	ProjectPostKey          = bsonutil.MustHaveTag(Project{}, "Post")
-	ProjectModulesKey       = bsonutil.MustHaveTag(Project{}, "Modules")
-	ProjectBuildVariantsKey = bsonutil.MustHaveTag(Project{}, "BuildVariants")
-	ProjectFunctionsKey     = bsonutil.MustHaveTag(Project{}, "Functions")
-	ProjectStepbackKey      = bsonutil.MustHaveTag(Project{}, "Stepback")
-	ProjectTasksKey         = bsonutil.MustHaveTag(Project{}, "Tasks")
-)
-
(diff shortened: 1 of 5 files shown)


model/project.go
```diff
@@ -1100,18 +1100,6 @@ func generateId(name string, projectIdentifier string, projBV *BuildVariant, rev
...
-	ProjectIdentifierKey    = bsonutil.MustHaveTag(Project{}, "Identifier")
-	ProjectPreKey           = bsonutil.MustHaveTag(Project{}, "Pre")
-	ProjectPostKey          = bsonutil.MustHaveTag(Project{}, "Post")
-	ProjectModulesKey       = bsonutil.MustHaveTag(Project{}, "Modules")
-	ProjectBuildVariantsKey = bsonutil.MustHaveTag(Project{}, "BuildVariants")
-	ProjectFunctionsKey     = bsonutil.MustHaveTag(Project{}, "Functions")
-	ProjectStepbackKey      = bsonutil.MustHaveTag(Project{}, "Stepback")
-	ProjectTasksKey         = bsonutil.MustHaveTag(Project{}, "Tasks")
-)
-
```
malikchaya2, line 1114: Were these deleted by mistake?

malikchaya2 (changes requested):

model/project.go
hadjri: No, they were unused :)

malikchaya2 (approved):

PR #7238, opened 2023-11-13
DEVPROD-63 add Dockerfile

ybrill:
[DEVPROD-63](https://jira.mongodb.org/browse/DEVPROD-63)

### Description
Add a Dockerfile to tell docker how to build evergreen. The first stage of the build builds all the clients and signs the macos ones. We'll need to pass in the args/secrets for that from drone.
The second stage copies static assets to where evergreen expects them and the executable to run is symlinked.

### Testing
Builds locally.


jayakasadev (commented):
testing access

ybrill:
Added a compose.yml that mimics what `make local-evergreen` does.

jayakasadev (approved):

PR #7239, opened 2023-11-14
DEVPROD-968 Add a function to read a checkrun output path and validate

malikchaya2:
DEVPROD-968

### Description
Add ReadAndValidateOutputPath to read from the given output path, validate it according to the design and return an object with output parameters. 

### Testing
Added a unit test

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -21,7 +21,7 @@ import (
 	"github.com/evergreen-ci/evergreen/thirdparty"
 	"github.com/evergreen-ci/evergreen/util"
 	"github.com/evergreen-ci/utility"
-	"github.com/google/go-github/v29/github"
+	"github.com/google/go-github/v53/github"
 	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
--- a/model/project.go
+++ b/model/project.go
@@ -2319,39 +2319,33 @@ func validateCheckRuns(checkRun *github.CheckRunOutput) error {
 		return errors.New("checkRun Output is nil")
 	}
 
-	if checkRun.Title == nil {
-		return errors.New("checkRun has no title")
-	}
+	catcher := grip.NewBasicCatcher()
 
-	if checkRun.Summary == nil {
-		return errors.New(fmt.Sprintf("the checkRun '%s' has no summary", utility.FromStringPtr(checkRun.Title)))
-	}
+	catcher.NewWhen(checkRun.Title == nil, "checkRun has no title")
+	summaryErrMsg := fmt.Sprintf("the checkRun '%s' has no summary", utility.FromStringPtr(checkRun.Title))
+	catcher.NewWhen(checkRun.Summary == nil, summaryErrMsg)
+
 	for _, an := range checkRun.Annotations {
 		annotationErrorMessage := fmt.Sprintf("checkRun '%s' specifies an annotation '%s' with no ", utility.FromStringPtr(checkRun.Title), utility.FromStringPtr(an.Title))
 
-		if an.Path == nil {
-			return errors.New(annotationErrorMessage + "path")
-		}
-		if an.StartLine == nil {
-			return errors.New(annotationErrorMessage + "start line")
-		}
-		if an.EndLine == nil {
-			return errors.New(annotationErrorMessage + "end line")
-		}
-		if an.AnnotationLevel == nil {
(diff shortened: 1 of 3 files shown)


model/project.go
```diff
@@ -2296,3 +2297,61 @@ func GetVariantsAndTasksFromPatchProject(ctx context.Context, settings *evergree
...
+		if an.AnnotationLevel == nil {
+			return errors.New(annotationErrorMessage + "annotation level")
+		}
+		if an.Message == nil {
+			return errors.New(annotationErrorMessage + "message")
+		}
+
+		if an.EndColumn != nil || an.StartColumn != nil {
+			if utility.FromIntPtr(an.StartLine) != utility.FromIntPtr(an.EndLine) {
+				errMessage := fmt.Sprintf("The annotation '%s' in checkRun '%s' should not include a column when start_line and end_line have different values", utility.FromStringPtr(an.Title), utility.FromStringPtr(checkRun.Title))
```
bynn, line 2350: can we specify it as 'StartColumn or EndColumn' instead of just saying column

model/project.go
```diff
@@ -2296,3 +2297,61 @@ func GetVariantsAndTasksFromPatchProject(ctx context.Context, settings *evergree
...
+	if checkRun.Summary == nil {
+		return errors.New(fmt.Sprintf("the checkRun '%s' has no summary", utility.FromStringPtr(checkRun.Title)))
+	}
+	for _, an := range checkRun.Annotations {
+		annotationErrorMessage := fmt.Sprintf("checkRun '%s' specifies an annotation '%s' with no ", utility.FromStringPtr(checkRun.Title), utility.FromStringPtr(an.Title))
+
+		if an.Path == nil {
+			return errors.New(annotationErrorMessage + "path")
+		}
+		if an.StartLine == nil {
```
bynn, line 2335: can we validate that start and endline > 0 because 1 is the lowest it can be?

model/project_test.go
```diff
@@ -2715,3 +2716,69 @@ tasks:
...
+	assert.Len(t, output.Annotations, 1)
+	assert.Equal(t, "a message", utility.FromStringPtr(output.Annotations[0].Message))
+}
+
+func TestValidateOutput(t *testing.T) {
+	f, err := os.CreateTemp(os.TempDir(), "")
+	require.NoError(t, err)
+	defer os.Remove(f.Name())
+
+	// a valid output
```
bynn, line 2760: comment says valid

bynn (changes requested):

model/project.go
```diff
@@ -2296,3 +2297,61 @@ func GetVariantsAndTasksFromPatchProject(ctx context.Context, settings *evergree
...
+
+	if err := utility.ReadJSONFile(file, &checkRunOutput); err != nil {
+		return nil, err
+	}
+
+	return checkRunOutput, validateCheckRuns(checkRunOutput)
+}
+
+func validateCheckRuns(checkRun *github.CheckRunOutput) error {
+	if checkRun == nil {
```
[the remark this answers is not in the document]
malikchaya2: Because it would hit this case when someone specified a path, but it didn't parse to a checkrun object. For example, if they did something like: 

{
   output: <the output>
}

instead of just 

{
   <the output>
}

So it would be an error case because they attempted to create it but didn't

model/project.go
```diff
@@ -2296,3 +2297,55 @@ func GetVariantsAndTasksFromPatchProject(ctx context.Context, settings *evergree
...
+	}
+
+	catcher := grip.NewBasicCatcher()
+
+	catcher.NewWhen(checkRun.Title == nil, "checkRun has no title")
+	summaryErrMsg := fmt.Sprintf("the checkRun '%s' has no summary", utility.FromStringPtr(checkRun.Title))
+	catcher.NewWhen(checkRun.Summary == nil, summaryErrMsg)
+
+	for _, an := range checkRun.Annotations {
+		annotationErrorMessage := fmt.Sprintf("checkRun '%s' specifies an annotation '%s' with no ", utility.FromStringPtr(checkRun.Title), utility.FromStringPtr(an.Title))
```
bynn, line 2329: This might be okay but when title is empty, this message is empty. I don't know if there's a better way to do this though.

bynn (approved):
added a small concern that might not be a problem but it looks good overall

model/project.go
malikchaya2: Yeah, I thought about that too. I don't think there's a better way either. At least the ' ' will indicate to the user that the title is empty.

PR #7241, opened 2023-11-15
DEVPROD-989 Wrap GitHub create/update checkrun APIs

malikchaya2:
DEVPROD-989

### Description
Expose the check run apis as third party functions by adding a createCheckrun and updateCheckrun function to thirdparty/github.go.

We do not need to add fallback because the API can only be used by GitHub apps

### Testing
We can't unit test this because it modifies, (our GitHub tests are only for gets). It will be easier to test this on staging once we are set up to call it in the agent.


thirdparty/github.go
```diff
@@ -1918,3 +1918,72 @@ func GetBranchProtectionRules(ctx context.Context, token, owner, repo, branch st
...
+		Output: output,
+		Name:   name,
+	}
+
+	checkRun, resp, err := githubClient.Checks.UpdateCheckRun(ctx, owner, repo, checkRunID, opts)
+	if resp != nil {
+		defer resp.Body.Close()
+		span.SetAttributes(attribute.Bool(githubCachedAttribute, respFromCache(resp.Response)))
+	}
+	if err != nil {
```
bynn, line 1984: can we wrap this error

bynn (approved):
it looks good!

PR #7191, opened 2023-11-01
DEVPROD-976 Let users specify an owner and repo for modules

malikchaya2:
DEVPROD-976

### Description
Right now, we expect users to specify a git url for modules. This is the example we show in our docs:

```
modules:
  name: example
  repo: git@github.com:10gen/mongo-example-modules.git
  prefix: src/mongo/db/modules
  ref: 12341a65256ff78b6d15ab79a1c7088443b9abcd
  branch: master
```

Now that we are deprecating ssh cloning, users should switch over to this:

```
modules:
  name: example
  owner: 10gen
  repo: mongo-example-modules
  prefix: src/mongo/db/modules
  ref: 12341a65256ff78b6d15ab79a1c7088443b9abcd
  branch: master
```

so that we can build the clone url ourselves.

### Testing
Updated tests 

### Documentation
Doc updates included

PR #7242, opened 2023-11-15
DEVPROD-2665 Deprecated https urls for modules

malikchaya2:
DEVPROD-2665

### Description
The one project that used an https url for modules switched over to specifying an owner/repo, and our docs no longer show an https example (removed in [7191](https://github.com/evergreen-ci/evergreen/pull/7191/files)). We therefore no longer need to support it.


hadjri:
LGTM assuming we have ensured that there are definitely no projects using HTTPS in their modules anymore

malikchaya2:
> LGTM assuming we have ensured that there are definitely no projects using HTTPS in their modules anymore

I just double checked again and it's not being used

PR #7244, opened 2023-11-15
DEVPROD-1966: create auto-archive command

Kimchelly:
DEVPROD-1966

### Description
* Add command to create an arbitrary output archive in any of the desired archive/compression schemes. This is a more general form of `archive.targz_pack` and `archive.zip_pack`.

### Testing
* Added unit tests and smoke test.
* Tested in staging that it:
    1. Correctly archived a source directory.
    2. Was in general more performant when compared to `archive.targz_pack` for gzipped tarballs of various sizes. The original request asked for auto-archive to use the multi-threaded gzip algorithm (which it does) instead of single-threaded gzip (i.e. what is currently used by `archive.targz_pack`) to make it faster. The one caveat to parallel gzip is that [it's a little slower than plain gzip for small archives](https://github.com/klauspost/pgzip#pgzip). I confirmed that despite this, parallel gzip is more likely to be a benefit than not when creating `.tar.gz` files - when I tested, it's much more performant on larger files than single-threaded gzip, but ever so slightly slower on small archives under 1 MB in size. Looking through existing usages of `archive.targz_pack`, I don't think the slight slowdown on small archives really matters given that most archives in CI tests are likely >1 MB (which I checked in Honeycomb) and the fact that 1MB is a very small amount of data to compress anyways.

### Documentation
Updated docs.


agent/command/archive_auto_create.go
```diff
@@ -0,0 +1,91 @@
...
+	// Target is the archive file that will be created. The target file's
+	// extension will determine the archiving format to use.
+	Target string `mapstructure:"target" plugin:"expand"`
+
+	// SourceDir is the directory to compress.
+	SourceDir string `mapstructure:"source_dir" plugin:"expand"`
+
+	// Include is a list of filename blobs to include from the source directory,
+	// e.g. "*.tgz", "file.txt", "test_*"
+	Include []string `mapstructure:"include" plugin:"expand"`
```
hadjri, line 27: Can we specify that the whole source dir will be used if this is empty in the comment here (I know it's already stated in the docs)

hadjri (commented):
Do we have a followup plan to deprecate targz_pack and zip_pack eventually as a result of this?

agent/command/archive_auto_create.go
Kimchelly: Done.

Kimchelly:
There's no plan to deprecate `archive.targz_pack` since a lot of people already use it, and I hesitate to do so because it has some custom logic for certain things (like symlinks) such that I can't be sure actually works the same as `archive.auto_pack`. There's no real downside to users of keeping both, so they can use whichever one works for them, although we may encourage them to use the `archive.auto_pack` rather than `archive.targz_pack` in new usages.

hadjri (approved):

PR #7245, opened 2023-11-15
DEVPROD-2597 Update tests to use GitHub App

bynn:
DEVPROD-2597

### Description
Tests can now use github app credentials instead of a token 

### Testing
current tests pass


config.go
```diff
@@ -587,7 +587,7 @@ func (s *Settings) GetGithubOauthToken() (string, error) {
 
 	oauthString, err := s.GetGithubOauthString()
 	if err != nil {
-		return "", err
```
bynn, line 590: Because we don't require this token anymore, we shouldn't return an error when a token is not set

testutil/testing.go
```diff
@@ -50,6 +51,15 @@ func ConfigureIntegrationTest(t *testing.T, testSettings *evergreen.Settings, te
 	integrationSettings, err := evergreen.NewSettings(*settingsOverride)
 	require.NoError(t, err, "Error opening settings override file '%s'", *settingsOverride)
 
+	// Manually update admin settings in DB for GitHub App credentials.
```
bynn, line 54: save settings override in DB

config.go
ablack12: the TODO for this function says that we'll be removing this function, if this isn't true anymore, can we update?

ablack12: I think we can also not have this return an error anymore then; 581 could just return ""

github_app.go
```diff
@@ -119,10 +119,7 @@ func (s *Settings) CreateInstallationToken(ctx context.Context, owner, repo stri
 
 func getInstallationID(ctx context.Context, authFields *githubAppAuth, owner, repo string) (int64, error) {
 	cachedID, err := getInstallationIDFromCache(ctx, owner, repo)
-	if err != nil {
-		return 0, errors.Wrapf(err, "getting cached installation id for '%s/%s'", owner, repo)
-	}
-	if cachedID != 0 {
+	if err == nil && cachedID != 0 {
```
ablack12, line 122: why don't we want to return an error here anymore? (In which case the function shouldnt)

ablack12 (changes requested):

scripts/setup-credentials.sh
```diff
@@ -65,9 +65,19 @@ auth:
...
-  manifest:
-    github_token: "$GITHUB_TOKEN"
 github_pr_creator_org: "10gen"
+
+# Do not edit below this line
+expansions:
+  github_app_key: |
 EOF
+
+echo "$GITHUB_APP_KEY" > app_key.txt
```
ablack12, line 80: Can we comment here what we're doing ?

config.go
bynn: talked about this offline but we are switching the TODO of this ticket to DEVPROD-1429 to deprecate using github tokens

trigger/patch_test.go
```diff
@@ -271,25 +271,25 @@ func (s *patchSuite) TestRunChildrenOnPatchOutcome() {
 	// there is no token set up in settings, but hitting this error
```
ablack12, line 271: We can remove this comment since its not relevant now

ablack12 (approved):

bynn:
added a spruce PR to fix test https://github.com/evergreen-ci/spruce/pull/2165

PR #7247, opened 2023-11-15
DEVPROD-962 clarify dependency space-separation logic

ablack12:
DEVPROD-962 
### Description
Realized that the behavior I'm seeing is actually what I see elsewhere in the docs, namely that we take an _intersection_ of all the criteria in the variant. So it's the example case we gave that was broken rather than the feature itself.  This matches the behavior I saw while testing.

### Documentation
Doc change only

### DIFF
--- a/docs/Project-Configuration/Project-Configuration-Files.md
+++ b/docs/Project-Configuration/Project-Configuration-Files.md
@@ -1486,11 +1486,18 @@ supported as a space-separated list. For example,
   depends_on:
   - name: test
     variant: "* !E"
+```
 
+Notably, selectors return items that satisfy all of the criteria. That is,
+they return the *set intersection* of each individual criterion. So the below yaml,
+while technically valid, wouldn't match anything given that these are static variant names, so the set 
+intersection will be nothing.
+
+``` yaml
 - name: push
   depends_on:
   - name: test
-    variant: "A B C D"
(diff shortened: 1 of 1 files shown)


docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1486,11 +1486,18 @@ supported as a space-separated list. For example,
   depends_on:
   - name: test
     variant: "* !E"
+```
```
ablack12, line 1489: image

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -1486,11 +1486,18 @@ supported as a space-separated list. For example,
...
+
+Notably, selectors return items that satisfy all of the criteria. That is,
+they return the *set intersection* of each individual criterion. So the below yaml,
+while technically valid, wouldn't match anything given that these are static variant names, so the set 
+intersection will be nothing.
```
malikchaya2, line 1494: Should we maybe add another example where the set intersection would be something?

ablack12: The example of "* !E" covers this case.

malikchaya2 (approved):

PR #7250, opened 2023-11-16
DEVPROD-709

ZackarySantana:
DEVPROD-709

### Description

### Testing

### Documentation

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -635,6 +635,14 @@ func (m *ec2Manager) ModifyHost(ctx context.Context, h *host.Host, opts host.Hos
 	}
 	defer m.client.Close()
 
+	grip.Info(message.Fields{
+		"message":   "Modifying host",
+		"message 2": "Prevalidate",
+		"hours":     opts.AddHours.String(),
+		"positive:": opts.AddHours > 1*time.Second,
+		"negative:": opts.AddHours < 1*time.Second,
+	})
+
 	// Validate modify options for user errors that should prevent all modifications
 	if err := validateEC2HostModifyOptions(h, opts); err != nil {
 		return errors.Wrap(err, "validating EC2 host modify options")
@@ -654,7 +662,13 @@ func (m *ec2Manager) ModifyHost(ctx context.Context, h *host.Host, opts host.Hos
 	if opts.NoExpiration != nil {
 		catcher.Add(m.setNoExpiration(ctx, h, *opts.NoExpiration))
 	}
-	if opts.AddHours > 0 {
+	grip.Info(message.Fields{
+		"message":   "Modifying host",
+		"hours":     opts.AddHours.String(),
+		"positive:": opts.AddHours > 1*time.Second,
+		"negative:": opts.AddHours < 1*time.Second,
+	})
+	if opts.AddHours != 0 {
 		if err := h.PastMaxExpiration(opts.AddHours); err != nil {
 			catcher.Add(err)
 		} else {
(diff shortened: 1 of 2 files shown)


cloud/ec2.go
```diff
@@ -654,7 +654,7 @@ func (m *ec2Manager) ModifyHost(ctx context.Context, h *host.Host, opts host.Hos
 	if opts.NoExpiration != nil {
 		catcher.Add(m.setNoExpiration(ctx, h, *opts.NoExpiration))
 	}
-	if opts.AddHours > 0 {
+	if opts.AddHours != 0 {
```
hadjri, line 657: We'll now need protection / testcases to ensure if the expiry length is shortened, it isn't shortened to before the current timestamp

hadjri (changes requested):
There are a handful of `MaxSpawnHostExpirationDurationHours` references elsewhere in the codebase (such as in routes like `/rest/v2/hosts/{host_id}/extend_expiration`) that are still operating off of the 14 day validation rule, that we'll need to update

cloud/ec2.go
ZackarySantana: The route you mentioned is actually internally calling the same function that checks for 30 days (and it double checks for 14 days so I just removed that check and kept the 30 days check that it was doing).

There were four other instances of MaxSpawnHostExpirationDurationHours. Two that I just removed because of a double check and two in our ModifyVolumeHandler or PATCH /rest/v2/volumes/{volume_id} (https://github.com/ZackarySantana/evergreen/blob/c460a923baacf7002a86a435c72f4b7c88923c5b/rest/route/host_spawn.go#L814). 

I'm not that educated on our volumes and their use cases and what people do with them, or whatnot- should I just change the variable to 30 days to extend volume expiration too? I think it makes sense to align volume expiration with host expiration. The volumes also cannot be reduced in expiration time.

model/host/host.go
```diff
@@ -2156,16 +2156,20 @@ func (h *Host) UpdateParentIDs(ctx context.Context) error {
 }
 
 // For spawn hosts that have never been set unexpirable, this will
-// prevent spawn hosts from being set further than 30 days.
+// prevent spawn hosts from being set further than 30 days and earlier
```
hadjri, line 2159: nit: 30 days from creation

hadjri, line 2159: nit: start comment with name of the function

model/host/host.go
```diff
@@ -2156,16 +2156,20 @@ func (h *Host) UpdateParentIDs(ctx context.Context) error {
...
+// then the current time.
 // For unexpirable hosts, this will prevent them from being extended any further
-// than the new 30 day expiration (unless it is set to unexpirable again #loophole)
-func (h *Host) PastMaxExpiration(extension time.Duration) error {
+// than the new 30 day expiration (unless it is set to unexpirable again #loophole).
+func (h *Host) ValidExpirationExtension(extension time.Duration) error {
```
hadjri, line 2163: nit: should change the function name to ValidateExpirationExtension since we're returning an error instead of a boolean

cloud/ec2.go
hadjri: let's just keep the scope of this ticket to spawn host extension, we can potentially make a new ticket to investigate if we should do this for volumes too or not.

That being said lets rename `MaxSpawnHostExpirationDurationHours` --> `MaxVolumeExpirationDurationHours` since that's all it's used for now

rest/route/host_spawn.go
```diff
@@ -1148,13 +1148,6 @@ func (h *hostExtendExpirationHandler) Parse(ctx context.Context, r *http.Request
 	}
 	h.addHours = time.Duration(addHours) * time.Hour
 
-	if h.addHours <= 0 {
```
hadjri, line 1151: We should still at least still return a 400 bad request if the user sends in an invalid expiration extension. Since `MakeExtendedSpawnHostExpiration` in `Run` below is what now handles this, how about we change L1165 to return a 400 code rather than an internal error responder?

hadjri (changes requested):

hadjri (approved):

PR #7251, opened 2023-11-16
DEVPROD-981 Support notifications when SUCCESSFUL task exceeds some duration

malikchaya2:
DEVPROD-981

### Description
This allows projects to set up notifications for a task succeeding a duration only if the task is successful. 

### Testing
Tested on staging 

### Spruce Changes 
The UI change to add the option to the dropdown will be merged after this backend change is merged. 
PR: [2155](https://github.com/evergreen-ci/spruce/pull/2155)

### DIFF
--- a/trigger/task.go
+++ b/trigger/task.go
@@ -52,6 +52,7 @@ func makeTaskTriggers() eventHandler {
 		event.TriggerFailure:                     t.taskFailure,
 		event.TriggerSuccess:                     t.taskSuccess,
 		event.TriggerExceedsDuration:             t.taskExceedsDuration,
+		event.TriggerSuccessfulExceedsDuration:   t.taskSuccessfulExceedsDuration,
 		event.TriggerRuntimeChangeByPercent:      t.taskRuntimeChange,
 		event.TriggerRegression:                  t.taskRegression,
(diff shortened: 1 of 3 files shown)


trigger/task.go
```diff
@@ -665,6 +666,31 @@ func shouldSendTaskRegression(sub *event.Subscription, t *task.Task, previousTas
...
+func (t *taskTriggers) taskSuccessfulExceedsDuration(sub *event.Subscription) (*notification.Notification, error) {
+	if t.task.IsPartOfDisplay() {
+		return nil, nil
+	}
+
+	if t.task.Status != evergreen.TaskSucceeded {
+		return nil, nil
+	}
+
+	thresholdString, ok := sub.TriggerData[event.TaskDurationKey]
```
ablack12, line 678: Since this logic is duplicated from below, we can just call taskExceedsDuration after checking the status

ablack12 (changes requested):

trigger/task.go
malikchaya2: 🤦🏻 yes, fixed

ablack12 (approved):

PR #7258, opened 2023-11-17
DEVPROD-1965: use pgzip for archive.targz_pack and archive.targz_extract

Kimchelly:
DEVPROD-1965

### Description
Use parallel gzip (pgzip) instead of single-threaded gzip when the files to archive are at least 1 MB in total size. Parallel gzip should be a drop-in replacement for gzip, so the only difference is performance.

I included some more info below for how I determined this was better (**feel free to skip reading the performance analysis unless you're really interested**).

* Use pgzip for `archive.targz_pack` when total file size to archive exceeds 1 MB.
* Refactor to move archive helper functions from agent util into the command package. (The refactor was necessary due to the fact that without moving the functions, we couldn't switch between pgzip/gzip based on file size.)

### Testing
* Tested in staging for basic correctness that pgzip produced the same targz contents when extracted as regular gzip.
* Updated unit tests to test building a targz archive with gzip and pgzip.

#### Performance
Since this is a performance change, I investigated whether it was true that pgzip would speed up performance. There's two factors to consider with a compression algorithm, which is speed of compression (faster = better) and how much it compresses the files (smaller archive = better) and in general, compression algorithms face a tradeoff between these two. I based my analysis on [the pgzip benchmarks given by the repo](https://github.com/klauspost/pgzip#compression-1), my own empirical testing in staging to verify, and some Honeycomb stats on how people use `archive.targz_pack`.

##### Decompression
[pgzip is at least as good as gzip at decompression](https://github.com/klauspost/pgzip#decompression-1), so we can always use it to decompress and it'll be better than gzip decompression.

##### Compression
**For files <1 MB in size, [pgzip is strictly worse at compression](https://github.com/klauspost/pgzip#compression)** - it's both slower and produces a compressed result that's ~2% larger than gzip. Based on Honeycomb data, the vast majority of `archive.targz_pack`s create small archives, so we definitely shouldn't use it if the files to be archived are under 1 MB.

**For files 1 MB and larger, pgzip is better in general at compression** - [it's significantly faster at the compression but still produces a compressed archive that's ~2% larger than one created by gzip](https://github.com/klauspost/pgzip#compression-1). This means that there's a tradeoff where the initial compression time is faster, but each upload/download of the targz is a bit slower due to the bigger size (we assume targz files always get uploaded to somewhere like S3 after creation, and are downloaded later on for things like dependent tasks). I did a bit of number crunching based on Honeycomb data + some staging tests on whether the larger compressed archive would affect the choice to use pgzip or not, and concluded that while it does degrade the performance a little bit for subsequent uploads/downloads, it's still worth it to use pgzip when archiving files > 1 MB as the compression speedup is substantial. The 2% increased size would be more detrimental if the task (or dependent tasks) uploaded/downloaded the targz many many times, but I assumed that this was not the case.

### Documentation
N/A


agent/command/archive_util.go
```diff
@@ -222,14 +221,137 @@ func TarGzReader(path string) (f, gz io.ReadCloser, tarReader *tar.Reader, err e
 	return f, gz, tarReader, nil
 }
 
-// TarGzWriter returns a file, gzip writer, and tarWriter for the path.
+// tarGzWriter returns a file, gzip writer, and tarWriter for the path.
```
ablack12, line 224: Could this be updated to explain useParallelGzip?

agent/command/archive_util.go
```diff
@@ -222,14 +221,137 @@ func TarGzReader(path string) (f, gz io.ReadCloser, tarReader *tar.Reader, err e
...
+	}
+
+	if catcher.HasErrors() {
+		return nil, 0, catcher.Resolve()
+	}
+
+	return out, totalSize, nil
+}
+
+func streamArchiveContents(ctx context.Context, rootPath string, includes, excludes []string) (files []archiveContentFile, totalSize int, err error) {
```
ablack12, line 272: It's a little hard to tell what changed since you moved files around -- I'd generally suggest avoiding big refactors when making changes to simplify reviewing. It seems here like we only added addUniqueFile in order to compute totalSize, to use later in making a distnction for pgzip? Could we add a comment here to that effect?

ablack12 (approved):

agent/command/archive_util.go
ablack12: Eh actually disregard, the refactor is obviously a good one. A comment for the inner function would still be helpful I think though.

agent/command/archive_util.go
Kimchelly: Done.

PR #7268, opened 2023-11-20
DEVPROD-980 Initialize stepback info if it does not exist before copying over to API

ZackarySantana:
DEVPROD-980

### Description
Adds a nil check to make sure an unwanted exception is not thrown

### Testing
Existing unit tests

### DIFF
--- a/rest/model/task.go
+++ b/rest/model/task.go
@@ -537,12 +537,11 @@ func (at *APITask) ToService() (*task.Task, error) {
 	}
 
 	if at.StepbackInfo != nil {
-		if st.StepbackInfo == nil {
-			st.StepbackInfo = &task.StepbackInfo{}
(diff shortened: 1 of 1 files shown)


rest/model/task.go
```diff
@@ -537,6 +537,9 @@ func (at *APITask) ToService() (*task.Task, error) {
 	}
 
 	if at.StepbackInfo != nil {
+		if st.StepbackInfo == nil {
```
bynn, line 540: is there a reason why we dont do something like this here?
`st.StopbackInfo = stepbackinfo{
  lastfailing 
  lastpassing
...
}`

bynn (changes requested):

bynn (approved):

PR #7269, opened 2023-11-20
DEVPROD-716 Consolidate duplicate validate logic in API and validate command

ZackarySantana:
DEVPROD-716

### Description
This consolidates some logic that we had in our API /validate route and our evergreen validate command to be the same. Previously they were the same except the validate command would surface errors from parsing the yaml and check aliases as well. Now it is only parsing the yaml errors, which isn't available when we are setting the version warnings/errors.

As well, I noticed how we were a little inconsistent with our warnings/errors. We say some duplicate messages when a project is not found for example, but if it errored earlier- we would say it errored for one message and for the other we would say the project does not exist.

We also did not follow the same ordering (we had like one of them switched I think). And this makes the ordering the same.

### Testing
Existing unit testing, I was debating about adding more unit testing for this function but I'm not sure if the overhead of testing is just duplicate testing from the functions it calls and the caller function,

### DIFF
--- a/service/api.go
+++ b/service/api.go
@@ -230,61 +230,12 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	errs := validator.ValidationErrors{}
-	var projectRef *model.ProjectRef
-	if input.ProjectID != "" {
-		projectRef, err = model.FindMergedProjectRef(input.ProjectID, "", false)
-		if err != nil {
-			validationErr = validator.ValidationError{
-				Message: "error finding project; validation will proceed without checking project settings",
-				Level:   validator.Warning,
-			}
-			errs = append(errs, validationErr)
-		} else if projectRef == nil {
-			validationErr = validator.ValidationError{
-				Message: "project does not exist; validation will proceed without checking project settings",
-				Level:   validator.Warning,
-			}
-			errs = append(errs, validationErr)
-		} else {
-			isConfigDefined := projectConfig != nil
-			errs = append(errs, validator.CheckProjectSettings(ctx, evergreen.GetEnvironment().Settings(), project, projectRef, isConfigDefined)...)
-		}
-	} else {
-		validationErr = validator.ValidationError{
-			Message: "no project specified; validation will proceed without checking project settings",
-			Level:   validator.Warning,
-		}
-		errs = append(errs, validationErr)
-	}
-
-	errs = append(errs, validator.CheckProjectErrors(r.Context(), project, input.IncludeLong)...)
-	if projectConfig != nil {
-		errs = append(errs, validator.CheckProjectConfigErrors(projectConfig)...)
-	}
+	projectRef, err := model.FindMergedProjectRef(input.ProjectID, "", false)
+	errs := validator.CheckProject(ctx, project, projectConfig, projectRef, evergreen.GetEnvironment().Settings(), input.IncludeLong, input.ProjectID != "", err != nil)
 
 	if input.Quiet {
 		errs = errs.AtLevel(validator.Error)
-	} else if projectRef == nil {
-		validationErr = validator.ValidationError{
-			Message: "no project specified; validation will proceed without checking alias coverage",
-			Level:   validator.Warning,
-		}
-		errs = append(errs, validationErr)
-	} else {
-		errs = append(errs, validator.CheckProjectWarnings(project)...)
-		// Check project aliases
-		aliases, err := model.ConstructMergedAliasesByPrecedence(projectRef, projectConfig, projectRef.RepoRefId)
-		if err != nil {
-			errs = append(errs, validator.ValidationError{
-				Message: "problem finding aliases; validation will proceed without checking alias coverage",
-				Level:   validator.Warning,
-			})
-		} else {
-			errs = append(errs, validator.CheckAliasWarnings(project, aliases)...)
-		}
 	}
-
 	if len(errs) > 0 {
 		gimlet.WriteJSONError(w, errs)
 		return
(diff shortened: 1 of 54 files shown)


validator/project_validator.go
```diff
@@ -222,6 +222,60 @@ func getDistrosForProject(ctx context.Context, projectID string) (ids []string,
 	return ids, aliases, nil
 }
 
+func CheckProject(ctx context.Context, project *model.Project, config *model.ProjectConfig, ref *model.ProjectRef, settings *evergreen.Settings, includeLong, projectRefProvided, projectRefErrored bool) ValidationErrors {
```
ablack12, line 225: A couple of things here:
1) Exported function needs a function comment
2) We want to avoid having this many variables passed in that have the same type, since its easy to mistake which one is which and pass things in the wrong order. If we need this many variables, we should use different types. 
3) projectRefProvided boolean is a little confusing; would explain how this ties into err and project = nil in the comments
4) We can just call evergreen.GetEnvironment().Settings() in this function instead of passing it in, since we just call that in both places.

validator/project_validator.go
```diff
@@ -222,6 +222,60 @@ func getDistrosForProject(ctx context.Context, projectID string) (ids []string,
...
+	isConfigDefined := config != nil
+	verrs := CheckProjectErrors(ctx, project, includeLong)
+	if !projectRefProvided {
+		verrs = append(verrs, ValidationError{
```
ablack12, line 229: We can combine this into one warning (also below)

validator/project_validator.go
```diff
@@ -222,6 +222,60 @@ func getDistrosForProject(ctx context.Context, projectID string) (ids []string,
...
+				})
+			} else {
+				verrs = append(verrs, CheckAliasWarnings(project, aliases)...)
+			}
+		}
+	}
+	if config != nil {
+		verrs = append(verrs, CheckProjectConfigErrors(config)...)
+	}
+	verrs = append(verrs, CheckProjectWarnings(project)...)
```
ablack12, line 275: I suggest moving this and the config != nil statements to the top; then in the if projectRef not provided case and the error cases we can just return verrs directly and remove some of these nested statements

ablack12 (changes requested):
title nit: consolidate

service/api.go
```diff
@@ -230,61 +230,12 @@ func (as *APIServer) validateProjectConfig(w http.ResponseWriter, r *http.Reques
...
-			Level:   validator.Warning,
-		}
-		errs = append(errs, validationErr)
-	}
-
-	errs = append(errs, validator.CheckProjectErrors(r.Context(), project, input.IncludeLong)...)
-	if projectConfig != nil {
-		errs = append(errs, validator.CheckProjectConfigErrors(projectConfig)...)
-	}
+	projectRef, err := model.FindMergedProjectRef(input.ProjectID, "", false)
```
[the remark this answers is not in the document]
ZackarySantana: I pass the error in to the check project function because it handles the output relating to the error. I updated the implementation with a better flow and different data types.

validator/project_validator.go
ZackarySantana: I combined them, originally they were different warnings so I kept it to keep behavior- I think it makes sense to combine if you think it isn't a big deal to change the output of it.

validator/project_validator.go
ZackarySantana: This is much cleaner and much less nesting! I like this way of doing it. I originally kept it in the same ordering that we previously had it so there's not that much change to the output but this makes the code a lot cleaner.

service/api.go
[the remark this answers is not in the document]
ablack12: Oh yeah I forgot to update this comment. Since its unusual not to check the error here, we should add a comment, or else someone else is gonna assume this is a mistake and change it. Ideally we'd also name it something less generic than err since we're using it as a variable, like "projectRefErr"

validator/project_validator.go
```diff
@@ -222,58 +222,50 @@ func getDistrosForProject(ctx context.Context, projectID string) (ids []string,
...
 }
 
-func CheckProject(ctx context.Context, project *model.Project, config *model.ProjectConfig, ref *model.ProjectRef, settings *evergreen.Settings, includeLong, projectRefProvided, projectRefErrored bool) ValidationErrors {
+// CheckProject calls the validating logic for a Project's configuration.
+// That is, ProjectErrors, ProjectWarnings, ProjectConfigErrors,
+// ProjectSettings, and AliasWarnings. If a respective item is nil,
+// it will not check it (e.g. if the config is nil, it does not check the config).
+// projectRefId is used to determine if there is a project specified and
+// projectRefErr is used to determine if there was a problem retrieving
+// the ref, both output different warnings for the project.
```
ablack12, line 231: tiny nit: "use -- or ; instead of , here to avoid runon sentence.

validator/project_validator.go
```diff
@@ -222,58 +222,50 @@ func getDistrosForProject(ctx context.Context, projectID string) (ids []string,
...
-	if !projectRefProvided {
+	verrs = append(verrs, CheckProjectWarnings(project)...)
+	if config != nil {
+		verrs = append(verrs, CheckProjectConfigErrors(config)...)
+	}
+
+	if projectRefId == "" {
 		verrs = append(verrs, ValidationError{
-			Message: "no project specified; validation will proceed without checking project settings",
+			Message: "no project specified; validation will proceed without checking project settings and alias coverage",
```
ablack12, line 242: nit: maybe we should change this wording to "will not check ..." instead of "will proceed", since the latter wording relies on ordering (like, we're doing this check last now so there's no other validation to proceed to)

ablack12 (approved):
LGTM with nits! Also Consolodate -> Consolidate

ablack12:
Also I'd merge main to see if the spruce error is flaky or not. I could see this being a real failure if it has to do with the error output returned, but didn't look into it.

PR #7270, opened 2023-11-20
DEVPROD-769 add doc change for project page flags

ablack12:
DEVPROD-769

### Description
We don't have a doc section for this, so adding something small. I thought about adding screenshots and etc but I don't think we should need that -- these are small enough that the page itself should be explanatory, and that increases the risk of things getting outdated, so keeping it simple.

### DIFF
--- a/docs/Project-Configuration/Project-and-Distro-Settings.md
+++ b/docs/Project-Configuration/Project-and-Distro-Settings.md
@@ -65,6 +65,19 @@ Admins can also set the branch project to inherit values from a
 repo-level project settings configuration. This can be learned about at
 ['Using Repo Level Settings'](Repo-Level-Settings.md).
 
+### Project Flags
+
+Under project flags, admins have a number of options for users to configure what
+runs for their project. For example, prevent projects from creating mainline 
+commits by disabling repotracking (by default, this is enabled. Evergreen occasionally 
+misses commits due to misconfiguration or GitHub outages, in which case admins can 
+Force Repotracker Run to unblock this). 
+
+Admins can also enable the ability to unschedule old tasks if a more recent
+commit passes, or configure tasks to stepback on failure to isolate the cause.
+
+Check out the settings on the page to see more options.
+
 ### Access and Admin Settings
 
 Admins can set a project as private or public. A private project can
@@ -82,11 +95,6 @@ via the REST API. The default for this setting is to allow logged-in
 users basic access to this project. Note this is different from the
 Private/Public setting above, which restricts access for users that are
 not logged in.
-
-### Scheduling Settings
-
-Admins can enable the ability to unschedule old tasks if a more recent
-commit passes.
 
 ### Variables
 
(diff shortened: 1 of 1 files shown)


docs/Project-Configuration/Project-and-Distro-Settings.md
```diff
@@ -65,6 +65,19 @@ Admins can also set the branch project to inherit values from a
 repo-level project settings configuration. This can be learned about at
 ['Using Repo Level Settings'](Repo-Level-Settings.md).
 
+### Project Flags
```
ablack12, line 68: image

malikchaya2: I have a feeling that users would grep for things like "disable waterfall" or "disable cron builds". I think people might end up here because they don't realize that the repotracker settings is what they need. it might be a good idea to include those phrases somewhere so this bit comes up when that is searched.

docs/Project-Configuration/Project-and-Distro-Settings.md
```diff
@@ -65,6 +65,19 @@ Admins can also set the branch project to inherit values from a
...
+
+Under project flags, admins have a number of options for users to configure what
+runs for their project. For example, prevent projects from creating mainline 
+commits by disabling repotracking (by default, this is enabled. Evergreen occasionally 
+misses commits due to misconfiguration or GitHub outages, in which case admins can
```
hadjri, line 73: Perhaps we should create a repotracker settings subsection here as the latter half of this block is getting more into what the repotracker is and how the repotracker settings specifically work

ablack12: I can do this but going to leave screenshots out re comment above

docs/Project-Configuration/Project-and-Distro-Settings.md
ablack12: reworded a bit

ablack12: image

hadjri (approved):

docs/Project-Configuration/Project-and-Distro-Settings.md
```diff
@@ -65,6 +65,25 @@ Admins can also set the branch project to inherit values from a
...
+tasks if a more recent commit passes, or configure tasks to stepback on failure to 
+isolate the cause.
+
+Check out the settings on the page to see more options.
+
+#### Repotracker Settings
+By default, Evergreen creates mainline commits (also known as waterfall versions or 
+cron builds) for enabled projects. 
+
+Admins can prevent projects from creating mainline commits by **disabling repotrackin**g,
```
malikchaya2, line 81: nit: **disabling repotrackin**g --> **disabling repotracking**

docs/Project-Configuration/Project-and-Distro-Settings.md
malikchaya2: thank you!

malikchaya2 (approved):
approved with nit

PR #7271, opened 2023-11-21
DEVPROD-977 Add stepback bisection to graphql project ref

ZackarySantana:
DEVPROD-977

### Description
Adds stepback bisection from the APIProjectRef to the graphql schema.

### Testing
Existing tests, are more needed for this scope of change?

### DIFF
--- a/graphql/tests/repoSettings/projectRef/queries/project_ref.graphql
+++ b/graphql/tests/repoSettings/projectRef/queries/project_ref.graphql
@@ -15,6 +15,7 @@
       patchingDisabled
       prTestingEnabled
       stepbackDisabled
+      stepbackBisection
       batchTime
 
       gitTagAuthorizedUsers
--- a/graphql/generated.go
+++ b/graphql/generated.go
@@ -68394,7 +68394,7 @@ func (ec *executionContext) unmarshalInputProjectInput(ctx context.Context, obj 
 		asMap[k] = v
 	}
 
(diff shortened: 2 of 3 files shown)


graphql/tests/repoSettings/projectRef/queries/project_ref.graphql
```diff
@@ -15,6 +15,7 @@
       patchingDisabled
       prTestingEnabled
       stepbackDisabled
+      stepbackBisection
```
malikchaya2, line 18: The way these tests work is they make a query and then compare the results it returns to the expected values. The expected value is in a results.json file under the name of this file (project_ref.graphl), so I usually look for it by grepping for the file name. This one is here: https://github.com/evergreen-ci/evergreen/blob/main/graphql/tests/projectSettings/projectRef/results.json#L4

malikchaya2 (changes requested):

graphql/tests/repoSettings/projectRef/queries/project_ref.graphql
malikchaya2: If you need to edit the data, you can do it in the data.json file in the same location as the results.json file, which in this case is [here](https://github.com/evergreen-ci/evergreen/blob/main/graphql/tests/projectSettings/projectRef/data.json).

ZackarySantana: I think I got it, I added it to the test data and then I added it to the query too

hadjri (changes requested):
You'll also need to update the existing Evergreen graphQL schemas such as [this](https://gist.github.com/hadjri/2263e8e0ba4130963c418664a9627823#file-project-graphql-L251) and [this](https://gist.github.com/hadjri/8b4a56066975c842a7a959e907a0b13d#file-repo_ref-graphql-L34), and then run the `make gqlgen` makefile command

I [updated](https://docs.google.com/document/d/1KxK2-JhKiHg7ydoEJN6rAf63k94KE_5-DqlYBj48pig/edit#heading=h.ap0zlvs711tl) the dev guide to make it clearer that this is needed for future reference

hadjri (approved):

hadjri:
LGTM - we'll also need a UI reviewer here since we're touching GraphQL

sophstad (approved):

malikchaya2 (approved):

PR #7274, opened 2023-11-22
DEVPROD-2921 Use SHA of push trigger if it is a module of the downstream project

hadjri:
DEVPROD-2921

### Description
- When triggering a downstream patch via upstream push, set the downstream patch's manifest to the revision of the upstream project push, if the upstream project is a module of the downstream project.
- Refactored CreateManifest to just take in a list of modules for simplicity.

Note: While this would be doable by adding `auto_update: true` to the module in YAML, it would also mean that it would no longer be possible for users to make patches off of old versions in a way that replicates the old version's behavior fully, as its module would pull from the most recent revision rather than the module revision that existed at time of the base revision.

### Testing
Confirmed [before](https://spruce-staging.corp.mongodb.com/version/sandbox_6d59ba219b13b10cce97beec439977bc88639aa6_c6db985b8c9ffb231c10085b09efa348/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) the change that a push trigger does not necessarily use the same push revision for the manifest (and instead defaults to the manifest of the base revision).

Confirmed [after](https://spruce-staging.corp.mongodb.com/version/sandbox_fa8f1ad2f504826cfe0f6ad8052c9aba337d46b4_c6db985b8c9ffb231c10085b09efa348/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) change that a push trigger where the upstream project is a module of the downstream project has the downstream manifest pulling from the upstream push revision.
### Documentation
Updated trigger docs to mention this.

### DIFF
--- a/docs/Project-Configuration/Project-and-Distro-Settings.md
+++ b/docs/Project-Configuration/Project-and-Distro-Settings.md
@@ -259,7 +259,7 @@ Options:
 -   Level: Accepted values are task, build, and push. Task and build levels will trigger
     based on the completion of either a task or a build in the upstream project. 
     - Push level triggers do not require any upstream build or task to run, but instead trigger a downstream version once
-        a commit is pushed to the upstream project.
+    a commit is pushed to the upstream project.
     - For push level triggers, if the upstream project is a module of the downstream project's YAML,
     the manifest of the downstream version will use the commit hash of the upstream project's commit.
 -   Status: Only applicable to build and task level triggers. Specify which status of the upstream 
--- a/docs/Project-Configuration/Project-and-Distro-Settings.md
+++ b/docs/Project-Configuration/Project-and-Distro-Settings.md
@@ -257,9 +257,11 @@ Options:
 
 -   Project: The upstream project.
 -   Level: Accepted values are task, build, and push. Task and build levels will trigger
-    based on the completion of either a task or a build in the upstream project. Push level triggers do 
-    not require any upstream build or task to run, but instead trigger a downstream version once
-    a commit is pushed to the upstream project.
+    based on the completion of either a task or a build in the upstream project. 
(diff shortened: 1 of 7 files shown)


docs/Project-Configuration/Project-and-Distro-Settings.md
```diff
@@ -257,9 +257,11 @@ Options:
 
 -   Project: The upstream project.
 -   Level: Accepted values are task, build, and push. Task and build levels will trigger
-    based on the completion of either a task or a build in the upstream project. Push level triggers do 
-    not require any upstream build or task to run, but instead trigger a downstream version once
-    a commit is pushed to the upstream project.
+    based on the completion of either a task or a build in the upstream project. 
+    - Push level triggers do not require any upstream build or task to run, but instead trigger a downstream version once
+        a commit is pushed to the upstream project.
```
Kimchelly, line 262: nit: leading alignment whitespace is weird on L262 and L264.

Kimchelly (approved):

trigger/process_test.go
```diff
@@ -626,6 +626,11 @@ func TestProjectTriggerIntegrationForPush(t *testing.T) {
 	require.NotNil(mani)
 	assert.Equal(downstreamProjectRef.Id, mani.ProjectName)
 	assert.Equal(uptreamProjectRef.Branch, mani.Branch)
+	assert.Len(mani.Modules, 1)
+	assert.Equal("3585388b1591dfca47ac26a5b9a564ec8f138a5e", mani.Modules["sample"].Revision)
```
Kimchelly, line 630: To avoid nil pointer panicking, you would have to `require.NotNil` for `mani.Modules["sample"]` before dereferencing `Revision`.

PR #7276, opened 2023-11-22
DEVPROD-2919 allow authorized external collaborators to patch

ablack12:
DEVPROD-2919
### Description
Verify external contributors. Added log for future reference.

### Testing
Verified existing unit test passes. We don't have an external collaborator to test this with end-to-end.


thirdparty/github.go
```diff
@@ -1392,8 +1396,8 @@ func authorizedForOrg(ctx context.Context, token, requiredOrganization, name str
 	return false, nil
 }
 
-func GitHubUserPermissionLevel(ctx context.Context, token, owner, repo, username string) (string, error) {
-	level, err := permissionLevel(ctx, "", owner, repo, username)
+func GitHubUserHasWritePermission(ctx context.Context, token, owner, repo, username string) (bool, error) {
```
hadjri, line 1399: Can we document this exported func?

units/patch_intent.go
```diff
@@ -955,6 +955,18 @@ func (j *patchIntentProcessor) buildGithubPatchDoc(ctx context.Context, patchDoc
 			"intent_id":   j.IntentID,
 		}))
 		return false, err
+	} else if !isMember {
+		grip.Debug(message.Fields{
```
hadjri, line 959: Is this intended to be a short term log?

ablack12: Long term -- mentioned this in the PR comment but I think generally this log is helpful for future reference, since we occasionally hit unauthorized errors and its helpful to know a bit more about why so we can communicate with the user. That said, I could attach it specifically to a ticket if we'd like to keep the logs clear, which I can see an argument for (maybe could just be removed after [DEVPROD-707](https://jira.mongodb.org/browse/DEVPROD-707))

hadjri: If we think [DEVPROD-707](https://jira.mongodb.org/browse/DEVPROD-707)'s completion could be reason to remove this log then it'd be helpful to add this as either a comment or a ticket field in the log itself, just so we can remember

hadjri (approved):
Approved w one suggestion

units/patch_intent.go
ablack12: I'll add a comment, since we still might want to hold onto this log.

ablack12:
Spruce failure looks unrelated.

PR #7288, opened 2023-11-28
DEVPROD-762 remove doc mentions of build baron / build failure

ablack12:
DEVPROD-762 
### Description
We renamed this UI section to Ticket Creation, and these tickets are no longer strictly build failure. 
Also added docs that we didn't have before.

### DIFF
--- a/docs/Project-Configuration/Project-and-Distro-Settings.md
+++ b/docs/Project-Configuration/Project-and-Distro-Settings.md
@@ -392,17 +392,16 @@ Options:
 Users can enable the performance plugin for tracking historical
 performance of tasks and tests.
 
-### Build Baron
+### Ticket Creation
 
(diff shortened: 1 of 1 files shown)


ablack12:
Also, didn't document Jira custom fields since no one is using them and they don't appear to be used anywhere. Should we remove this?  cc @malikchaya2 
image

hadjri (approved):
Change LGTM - will let @malikchaya2 weigh in on custom jira fields

malikchaya2:
> Also, didn't document Jira custom fields since no one is using them and they don't appear to be used anywhere. Should we remove this? cc @malikchaya2 

This was specifically requested by cloud, so I'd maybe check in with Geoff to see if they still want to use it.

PR #7289, opened 2023-11-29
DEVPROD-764 enable oom tracker by default

ablack12:
DEVPROD-764 
### Description
Switch oom tracker to be true by default. Will send out an email to communicate this.

### Testing
Added a unit test.

### Documentation
Updated doc

### DIFF
--- a/docs/Project-Configuration/Project-Configuration-Files.md
+++ b/docs/Project-Configuration/Project-Configuration-Files.md
@@ -968,7 +968,7 @@ PIDs. A message with PIDs will also be displayed in the metadata panel in the UI
 To disable the OOM tracker, add the following to the top-level of your yaml.
 
 ``` yaml
-- oom_tracker: false
+oom_tracker: false
 ```
 
 ### Matrix Variant Definition
--- a/docs/Project-Configuration/Project-Configuration-Files.md
+++ b/docs/Project-Configuration/Project-Configuration-Files.md
@@ -959,12 +959,18 @@ task within a specific build variant).
 
 ### Out of memory (OOM) Tracker
 
-This is set to true at the top level if you'd like to enable the OOM Tracker for your project.
+By default, the OOM tracker is enabled. 
 
 If there is an OOM kill, immediately before the post-task starts, there will be
 a task log message saying whether it found any OOM killed processes, with their
 PIDs. A message with PIDs will also be displayed in the metadata panel in the
 UI.
+
+To disable the OOM tracker, add the following to the top-level of your yaml.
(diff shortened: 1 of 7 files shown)


Kimchelly (commented):
If I remember correctly, @malikchaya2 had some trouble with the OOM tracker because it was enabled in one of the agent tests and running the agent tests locally triggered the security team's alerts because the (real) OOM tracker in MacOS runs `sudo log erase --all`, which is considered suspicious activity.

Not really a perfect fix, but since we're now making it default enabled, can we either make the OOM tracker never run in tests, or change the agent tests that could trip the OOM tracker ([here](https://github.com/evergreen-ci/evergreen/blob/531efa60199309956d37d70161e8dc3c5bfbfca0/agent/agent.go#L710-L715) and [here](https://github.com/evergreen-ci/evergreen/blob/531efa60199309956d37d70161e8dc3c5bfbfca0/agent/agent.go#L909)) to use [the mock OOM tracker](https://github.com/evergreen-ci/evergreen/blob/531efa60199309956d37d70161e8dc3c5bfbfca0/agent/agent_test.go#L147)?

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -959,12 +959,17 @@ task within a specific build variant).
...
-a task log message saying whether it found any OOM killed processes, with their
-PIDs. A message with PIDs will also be displayed in the metadata panel in the
-UI.
+an agent log message saying whether it found any OOM killed processes, with their
+PIDs. A message with PIDs will also be displayed in the metadata panel in the UI.
+
+To disable the OOM tracker, add the following to the top-level of your yaml.
+
+``` yaml
+- oom_tracker: false
```
Kimchelly, line 971: Remove `- ` since it's a top-level key rather than a list item.

ablack12: good catch thanks!

Kimchelly (approved):

ablack12:
@Kimchelly I added some stuff to use the mock; unfortunately I can't think of a good way to enforce this but it seems likely to stick given that we usually copy paste or add to the existing test suite. I could add something into the code to check if we're running locally if such a thing exists? Feels hacky in a different way though

Kimchelly:
I think if the concern arose again for new tests accidentally running the OOM tracker, we could always resort to something like checking for test-specific env var flags (such as looking for `EVGHOME`) in the future. Just shoring up the existing tests seems sufficient for now.

PR #7291, opened 2023-11-29
DEVPROD-2923 Delete Fallback for Some GitHub Functions

bynn:
DEVPROD-2923

### Description
Deleting fallback for some functions in thirdparty/github file 

### Testing
all unit tests still pass 
tasks in staging works as expected


hadjri (commented):
What's the reasoning for why this deletion is safe / wouldn't cause any breaking changes?

bynn:
@hadjri Yes, the fallbacks have not been utilized because we haven't been getting the splunk alerts so it seems safe to delete them. The reason why this is only SOME functions and not all is because I'm splitting this work into pieces but eventually this will be done to every github function.

malikchaya2 (approved):

bynn:
closing because smoke tests need to not rely on fallback first

PR #7292, opened 2023-11-29
DEVPROD-2473 use instance profile to connect to the db

ybrill:
[DEVPROD-2473](https://jira.mongodb.org/browse/DEVPROD-2473)

### Description
Right now we're connecting to the db using a static username/password stored on the host (chef puts a file with the creds where the application is expecting it). This isn't the best and would also be a challenge to adapt for when we're running on Kanopy so this PR makes it use the instance profile assigned to the app hosts. A subsequent PR will add getting a role through IRSA.

We need to remember to add the role db user to the production project in Atlas before this is deployed.

### Testing
Works on staging (once I created the role DB user for staging)


environment.go
```diff
@@ -406,6 +394,20 @@ func (e *envState) createRemoteQueues(ctx context.Context) error {
...
 }
 
+func dbCreds(ctx context.Context) (options.Credential, error) {
+	awsCreds, err := ec2rolecreds.New().Retrieve(ctx)
+	if err != nil {
+		return options.Credential{}, errors.Wrap(err, "getting AWS credentials")
+	}
+	return options.Credential{
+		AuthMechanism:           "MONGODB-AWS",
+		AuthMechanismProperties: map[string]string{"AWS_SESSION_TOKEN": awsCreds.SessionToken},
```
malikchaya2, line 404: nit: Can we please use a const for AWSSessionToken and MONGODB-AWS?  (We have AWSSessionToken as a const in assume_ec2_role.go)

cmd/load-smoke-data/load-smoke-data.go
```diff
@@ -212,20 +212,6 @@ func main() {
 	const dbURI = "mongodb://localhost:27017"
```
malikchaya2, line 212: (random comment placement to allow for threading)

I'm assuming that part of the work for this ticket will be to remove the static username/password stored on the host? Can we please document how to connect to the hosts locally after this change in our internal docs?

environment.go
ybrill: done

cmd/load-smoke-data/load-smoke-data.go
malikchaya2: ah yes, my bad, all good there

environment.go
```diff
@@ -348,21 +353,12 @@ func (e *envState) initDB(ctx context.Context, settings DBSettings) error {
 		SetConnectTimeout(5 * time.Second).
 		SetMonitor(apm.NewMonitor(apm.WithCommandAttributeDisabled(false), apm.WithCommandAttributeTransformer(redactSensitiveCollections)))
 
-	if settings.HasAuth() {
-		ymlUser, ymlPwd, err := settings.GetAuth()
+	if settings.AWSAuthEnabled {
```
malikchaya2, line 356: What's the downside to this always being enabled?

ybrill: Ah, I was just about to comment on the PR: I discovered from testing that in CI the tests were able to get AWS creds from the EC2 metadata and then when they tried to use them to connect to the test db the db rejected the connection because it hadn't been configured with AWS auth (e.g. [here](https://parsley.mongodb.com/evergreen/evergreen_ubuntu2204_test_cloud_patch_d46bf5ad9bb68de98cdb491c2d802515a73a7dd3_6568a7b7c9ec44e11507355f_23_11_30_15_18_18/0/task?bookmarks=0,610&shareLine=566)).

The solution I tried to implement is to add a flag that will allow us to only attempt to use AWS auth when it's specifically been configured. This way if the app is running in an environment that has the db configured to use auth we'll explicitly enable it.
The alternative was to let us turn it off for testing, but it made more sense to me to turn it on specifically when we want it.

malikchaya2: nice! thank you!

malikchaya2 (changes requested):
This looks good, but I think we should try to get more clarity on the staging panics before merging to prod, so that we know we won't face the same issue there.

ybrill:
@malikchaya2 I clarified with the driver team: it turns out this was going about providing AWS creds in the wrong way. We're supposed to just specify AWS auth and let the driver do the rest. Otherwise, the credentials we pass in expire and the next time the driver needs to make a new connection to the db auth fails.

[bab08e1](https://github.com/evergreen-ci/evergreen/pull/7292/commits/bab08e1b10886ab25545e4fce569745ba6a6182f) does it correctly. This has been up in staging for ~24 hours with no issue.

ybrill:
(and the failing test appears to be failing on the waterfall too 😅 )

PR #7303, opened 2023-12-08
DEVPROD-1758: support large JSON files for generate.tasks

Kimchelly:
DEVPROD-1758

### Description
Follow-up from #7287 to allow a task to submit input JSON files larger than 16 MB to generate.tasks. If the input file can be stored in the task collection document, it'll do that (same as how it has always done it). Otherwise if it's too large to put in a DB document, it'll be stored in S3.

* Integrate the logic from the original PR into generate.tasks.
* Fix a couple small bugs in the logic from the original PR.
* Add admin setting to UI for generated JSON S3 prefix.

### Testing
* Added unit and integration tests.
* Tested in staging to verify that [a sample patch](https://spruce-staging.corp.mongodb.com/task/evg_lint_generate_lint_patch_3c10c5e814535dc8e6d06553cce69a60e4f58c91_6573926197b1d3283b562278_23_12_08_22_02_10/logs?execution=0) used S3 to store the file when the generated JSON file exceeded 16 MB.
* Tested that a patch with a small generate.tasks file still gets stored in the DB.

### Documentation
N/A

### DIFF
--- a/model/task/generated_json_storage.go
+++ b/model/task/generated_json_storage.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/db"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
@@ -30,7 +31,7 @@ func GetGeneratedJSONFileStorage(settings *evergreen.Settings, method evergreen.
 	switch method {
 	case "", evergreen.ProjectStorageMethodDB:
 		return generatedJSONDBStorage{}, nil
-	case method:
+	case evergreen.ProjectStorageMethodS3:
 		return newGeneratedJSONS3Storage(settings.Providers.AWS.ParserProject)
 	default:
 		return nil, errors.Errorf("unrecognized generated JSON storage method '%s'", method)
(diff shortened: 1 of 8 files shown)


model/task/generated_json_storage.go
```diff
@@ -31,7 +31,7 @@ func GetGeneratedJSONFileStorage(settings *evergreen.Settings, method evergreen.
 	switch method {
 	case "", evergreen.ProjectStorageMethodDB:
 		return generatedJSONDBStorage{}, nil
-	case method:
```
ablack12, line 34: oops nice

model/task/generated_json_storage.go
```diff
@@ -82,7 +82,7 @@ func GeneratedJSONInsertWithS3Fallback(ctx context.Context, settings *evergreen.
 	}
 
 	grip.Info(message.Fields{
-		"message":            "successfully upserted generated JSON files into S3 as fallback due to document size limitation",
+		"message":            "successfully inserted generated JSON files into S3 as fallback due to document size limitation",
```
ablack12, line 85: I have a concern that it fails not because it hits the document limit, but because we run out of time passing everything over (because we hit the "unexpected end of JSON input" limit. Should we instead just case on if the files are greater than the document size limit rather than trying to insert, since we have the byte size already? Or check more error cases here?

ablack12 (commented):
overall LGTM, just have a question

ablack12 (approved):

model/task/generated_json_storage.go
ablack12: Awesome thanks!

PR #7308, opened 2023-12-11
DEVPROD-1661 add disk devices to task end detail

ybrill:
[DEVPROD-1661](https://jira.mongodb.org/browse/DEVPROD-1661)

### Description
When this field was added it wasn't included in the API version.
This PR rectifies that oversight.

### DIFF
--- a/config.go
+++ b/config.go
@@ -35,7 +35,7 @@ var (
 	ClientVersion = "2023-11-30"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-11-29"
+	AgentVersion = "2023-12-08"
 )
 
 // ConfigSection defines a sub-document in the evergreen config
--- a/graphql/schema/types/task.graphql
+++ b/graphql/schema/types/task.graphql
@@ -159,7 +159,7 @@ type TaskEndDetail {
   timeoutType: String
   type: String!
   traceID: String
-  diskDevices: [String]
+  diskDevices: [String!]!
 }
 
 type OomTrackerInfo {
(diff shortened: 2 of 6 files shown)


graphql/schema/types/task.graphql
```diff
@@ -159,6 +159,7 @@ type TaskEndDetail {
   timeoutType: String
   type: String!
   traceID: String
+  diskDevices: [String]
```
minnakt, line 162: I'm not sure what our formal guidelines are around GraphQL types, but I like typing lists like the following:
```suggestion
  diskDevices: [String!]!
```
because it prevents `diskDevices` and any of its entries from being `null`, which is generally easier to handle on the frontend. 😃

config.go
```diff
@@ -35,7 +35,7 @@ var (
 	ClientVersion = "2023-11-30"
 
 	// Agent version to control agent rollover.
-	AgentVersion = "2023-11-29"
+	AgentVersion = "2023-12-08"
```
minnakt, line 38: Just want to double check that this change was necessary — I tried undoing it and ran a patch with `verify-agent-version-update` task which passed, but I also don't quite know how this works. 😅

minnakt (approved):

graphql/schema/types/task.graphql
ybrill: done!

config.go
ybrill: I _think_ it's necessary: the change to `apimodels.TaskEndDetails` should make a difference on the agent [here](https://github.com/evergreen-ci/evergreen/blob/f81dbfb9d61dc2d30d4b7ab6d2e465146e10cfc0/agent/agent.go#L1017).

I think the same is true for any change to the agent models, so I've added it to the test (cc @Kimchelly in case this is wrong).

Kimchelly: Yes, you're correct that the JSON tag matters since it changes the payload that the agent send to the app server when the task finishes.

PR #7309, opened 2023-12-12
docs: add more links for workstation commands

Kimchelly:
A user asked some questions about how to set up a project for workstations, so I added some clarification.

* Add more links to make it clearer on how to use workstation commands defined in the project settings.
* Fix a couple links (that I wrote) that hard-code the link to the docs website rather than use relative paths.

### DIFF
--- a/docs/Hosts/Developer-Workstations.md
+++ b/docs/Hosts/Developer-Workstations.md
@@ -22,7 +22,7 @@ workstations can use the same configuration as build hosts, but also have:
   subjected to the termination deadline of most spawn hosts. While
   there is a global per-user limit for unexpirable hosts,
   workstations will, by default. 
-  
+
 - Evergreen supports a start/stop mode for spawn hosts. This makes it
   possible for users to change to another instance type, though
   administrators must configure the allowable instance types. Users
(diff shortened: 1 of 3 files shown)


bynn (approved):

docs/Hosts/Developer-Workstations.md
```diff
@@ -46,16 +46,15 @@ checked out and running on workstations, though this feature is not
...
 systems.
 
-Project configuration include a "Workstation Setup" section where
-administrators declare a number of simple commands (and directory
-contexts) that will run a project's setup. These commands are *not*
-shell interpolated, and are *not* meant to provision the development
-environment (e.g. install system packages or modify configuration
-files in `~/`). Use these commands to clone a repository, generate
-index files, and/or run a test build.
+Project configuration include a ["Workstation Setup" section](../Project-Configuration/Project-and-Distro-Settings.md#virtual-workstation-commands)
```
ablack12, line 49: includes**

docs/Hosts/Developer-Workstations.md
```diff
@@ -46,16 +46,15 @@ checked out and running on workstations, though this feature is not
...
+where administrators declare a number of simple commands (and directory
+contexts) that will run a project's setup. These commands are *not* shell
+interpolated, and are *not* meant to provision the development environment (e.g.
+install system packages or modify configuration files in `~/`). Use these
+commands to clone a repository, generate index files, and/or run a test build.
```
ablack12, line 54: If we have a handy tight example from server or something we can add that might be nice.

ablack12 (commented):
not blocking

docs/Hosts/Developer-Workstations.md
ablack12: also might suggest saying "The project page" since we often rfer to the yaml as the project configuration (which is non ideal but)

docs/Hosts/Developer-Workstations.md
Kimchelly: I didn't write this text, so I'm inclined to leave it. The server workstation commands aren't super generalizable since it's a clone of a specific repo, then just runs a script in that repo.

PR #7310, opened 2023-12-12
DEVPROD-3068 Fix outdated revision being picked for modules for commit queue patches

malikchaya2:
[DEVPROD-3068](https://jira.mongodb.org/browse/DEVPROD-3068)

### Description
This fixes the bug where an outdated revision is being chosen for modules when the module has had commits since the base commit was created. This was happening because when auto update is not set, the manifest is created based on the base commit. Since the base commit was wrong, the modules selected where based on what was the latest for the wrong base commit. 

The way the cli queue works is that it at first sets the githash to the base, but then when it processes the patch for the queue, it updates the hash to the head. The pr queue didn’t do that. I standardized that between the two commit queues. It also tries to streamline some more things. The CLI commit queue patch processing had some checks in place that the PR one didn't, so this should hopefully help prevent other bugs from being discovered. 

### Testing
Tested on staging. Without these changes a PR that had not merged main was added to the queue, it used an old revision. [This PR ](https://github.com/evergreen-ci/commit-queue-sandbox/pull/613) used [d5379](https://github.com/evergreen-ci/commit-queue-sandbox/commit/d537988416e28fa48abce47c5424931d70603fab) even though [bb2f1c](https://github.com/evergreen-ci/commit-queue-sandbox/commit/bb2f1cb0f108ae116ba5607a89f367d030ba57ab) was the latest. It used https://github.com/evergreen-ci/sample/commit/bd4f82954702ca72ad0c8639a72915bb8e771ee3 for the module, even though the “sample” module associated with head for the sandbox repo at that time was https://github.com/evergreen-ci/sample/commit/9bdedd0990e83e328e42f7bb8c2771cab6ae0145 which is newer.
 
When tested with the changes, the pr https://github.com/evergreen-ci/commit-queue-sandbox/pull/614 used this revision https://github.com/evergreen-ci/commit-queue-sandbox/commit/24de063bd89a0a68ac06436c78b158e201a6c14d which is the latest and it used the latest commit for the sample module as well https://github.com/evergreen-ci/sample/commit/9bdedd0990e83e328e42f7bb8c2771cab6ae0145.

### DIFF
--- a/units/commit_queue.go
+++ b/units/commit_queue.go
@@ -382,34 +382,12 @@ func (j *commitQueueJob) processGitHubPRItem(ctx context.Context, cq *commitqueu
 		return
 	}
 
-	patchDoc, err := patch.FindOneId(nextItem.PatchId)
+	v, projectConfig, err := updateAndFinalizeCqPatch(ctx, j.env.Settings(), nextItem.PatchId, commitqueue.SourcePullRequest, projectRef, true, githubToken)
 	if err != nil {
-		j.AddError(errors.Wrapf(err, "finding patch '%s'", nextItem.Version))
-		j.AddError(thirdparty.SendCommitQueueGithubStatus(ctx, j.env, pr, message.GithubStateFailure, "no patch found", ""))
-		j.dequeue(ctx, cq, *nextItem, fmt.Sprintf("could not find patch '%s'", nextItem.PatchId))
-		return
-	}
-	if patchDoc == nil {
-		errMsg := fmt.Sprintf("patch '%s' not found", nextItem.Version)
-		j.AddError(errors.New(errMsg))
-		j.AddError(thirdparty.SendCommitQueueGithubStatus(ctx, j.env, pr, message.GithubStateFailure, "no patch found", ""))
-		j.dequeue(ctx, cq, *nextItem, errMsg)
-		return
-	}
-	projectConfig, _, err := model.GetPatchedProject(ctx, j.env.Settings(), patchDoc, githubToken)
-	if err != nil {
-		err = errors.Wrap(err, "getting patched project")
 		j.logError(err, *nextItem)
 		j.dequeue(ctx, cq, *nextItem, err.Error())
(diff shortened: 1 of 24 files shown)


units/commit_queue.go
```diff
@@ -432,53 +410,66 @@ func (j *commitQueueJob) processGitHubPRItem(ctx context.Context, cq *commitqueu
...
-		return
+		return nil, nil, errors.Wrapf(err, "finding patch '%s'", patchId)
+	}
+	if patchDoc == nil {
+		return nil, nil, errors.Errorf("patch '%s' not found", patchId)
+	}
+
+	var projectConfig *model.Project
+	if returnProjectConfig {
+		projectConfig, _, err = model.GetPatchedProject(ctx, settings, patchDoc, githubToken)
```
ablack12, line 455: I think we're using projectConfig wrong here (I recognize this was misnamed from the other file though). GetPatchedProject returns project and projectConfig (and projectConfig is yaml project ref settings, not the translated parser project) so we should rename this to just project. 
i.e. we really have `proj, projectConfig, err = model.GetPatchedProject(...` 

Also, we should just return the one from updatePatch rather than re-calling this here. Then we can remove the return boolean.

units/commit_queue.go
```diff
@@ -487,39 +478,18 @@ func (j *commitQueueJob) processCLIPatchItem(ctx context.Context, cq *commitqueu
 	// compared to the latest project configuration. For the commit queue, it's
 	// best to test against the latest project configuration.
 	pp.Init(patchDoc.Id.Hex(), patchDoc.CreateTime)
-	ppStorageMethod, err := model.ParserProjectUpsertOneWithS3Fallback(ctx, j.env.Settings(), evergreen.ProjectStorageMethodDB, pp)
+	ppStorageMethod, err := model.ParserProjectUpsertOneWithS3Fallback(ctx, settings, evergreen.ProjectStorageMethodDB, pp)
```
ablack12, line 481: can we update the comment above to generalize CLI to just commit queue?

ablack12 (changes requested):

units/commit_queue.go
malikchaya2: thanks! done!

ablack12 (approved):

PR #7311, opened 2023-12-12
DEVPROD-722 Check upstream branch before local branch when creating patch

ZackarySantana:
[DEVPROD-722](https://jira.mongodb.org/browse/DEVPROD-722)

### Description
When doing `evergreen patch`, we attempt to reference a local branch that tracks upstream with a given branch name (namely branch+"@{upstream}" which can result in grabbing a git commit that does not exist on upstream (or incorrectly grabbing a branch). Instead, we should try to do `upstream/`+branch which will try to get the remote branch related commit. This correctly gets the common ancestor.

### Testing
The reason this ticket was created, was a user had a diverging fork and a branch based on that- then we tried getting a common ancestor commit between the current branch and the diverging fork. Instead, this tries to get the common ancestor between the branch and the main branch (which we can properly do git.get_project based on, rather than a commit only on a diverging fork).

I replicated this error state with a diverging fork and then I applied these changes (testing this also locally with what commit it tries to base the get_proejct on) and it correctly cloned and completed the tasks.


operations/patch_util.go
```diff
@@ -628,10 +628,13 @@ func loadGitData(dir, branch, ref, commits string, format bool, extraArgs ...str
 	// In the case a range is passed, we only need one commit to determine the base, so we use the first commit.
 	// For details see: https://git-scm.com/docs/gitrevisions
 
-	mergeBase, err := gitMergeBase(dir, branch+"@{upstream}", ref, commits)
+	mergeBase, err := gitMergeBase(dir, "upstream/"+branch, ref, commits)
```
malikchaya2, line 631: Can you please add some words to the comment above (on line 635) explaining `upstream/"+branch` in addition to `branch@{upstream}`?

malikchaya2 (changes requested):

Kimchelly:
Not gonna give a full review since I'm gonna be OOO, but thought it was worth asking - is this making a new assumption that the user sets up their local git repos with a remote called `upstream`? I thought that, unlike the `<branch>@upstream` (which is special and doesn't require you to have a remote called upstream), using `upstream/${branch}` assumes that the user names their remote `upstream`. Unfortunately, git doesn't have a reason you can't just name the remote pointing to the official repo something else, so while we do usually use `upstream` in Evergreen, I don't know if that's generalizable to all other teams. If users are required to set up upstream when they may not have needed to before, it may disrupt their usual patching workflow.

FWIW, I tried cloning Evergreen on my local computer, named my remote pointing to evergreen-ci/evergreen as `evg`, then submitted a [patch](https://spruce.mongodb.com/version/657c75ed32f4174ede5659d3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), which worked.

Anyways, I don't have a simple answer here and I don't actually know if this will actually be an issue (maybe everyone uses the `upstream` remote at MongoDB?), but I thought it was worth considering.

ZackarySantana:
After thinking about it more, the expected behavior of upstream/branch is not what we want. This will conflict if the user does not name their local branches the same as the remote branch. It is safer to stick with only branch@{upstream}.

That being said, I'm going to make it not 'guess' that the local branch that we want is upstream, but include it as an argument and write a helper function to get the correct remote on the user's machine. So if the remote that tracks their Evergreen project is evg, it will attempt to get the local branch tracking the remote evg

ZackarySantana:
Here is the same patch, one with the[old](https://spruce-staging.corp.mongodb.com/version/657ca46397b1d3ddcd42fa86/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) implementation and one with the [new one](https://spruce-staging.corp.mongodb.com/task/sandbox_ubuntu2004_task1_patch_edc6cd0b9f8a1b8481bcf14bc75f0a819a71aa7a_657ca5e79dbe3249fde60403_23_12_15_19_15_57/logs?execution=0). You can see the new one succeeded in running its task, while the old one couldn't clone the unknown commit.

operations/patch_util.go
ZackarySantana: Yes! let me add some wording

operations/patch_util.go
```diff
@@ -676,6 +681,24 @@ func loadGitData(dir, branch, ref, commits string, format bool, extraArgs ...str
...
 }
 
+func gitGetRemote(dir, owner, repo string) (string, error) {
+	out, err := gitCmdWithDir("remote", dir, "-v")
+	if err != nil {
+		return "", errors.Wrap(err, "getting git remotes")
+	}
+
+	lines := strings.Split(out, "\n")
+	partial := strings.ToLower("github.com/" + owner + "/" + repo)
```
Kimchelly, line 691: nit: might be more readable to use `fmt.Sprintf("github.com/%s/%s", owner, repo)` rather than concatenate.

operations/patch_util.go
```diff
@@ -676,6 +681,24 @@ func loadGitData(dir, branch, ref, commits string, format bool, extraArgs ...str
...
+	for _, line := range lines {
+		f := strings.Fields(line) // ["remote-name", "remote-url"]
```
Kimchelly, line 693: It may be helpful to add a comment above this to mention that the format of each line of `git remote -v` should be `<remote_name> <url> <fetch_or_pull>` to make the string parsing logic easier to reason through.

operations/patch_util.go
```diff
@@ -621,17 +621,22 @@ Continue?`, uncommittedChangesFlag), true), nil
...
-func loadGitData(dir, branch, ref, commits string, format bool, extraArgs ...string) (*localDiff, error) {
-	// branch@{upstream} refers to the branch that the branch specified by branchname is set to
-	// build on top of. This allows automatically detecting a branch based on the correct remote,
-	// if the user's repo is a fork, for example. This also works with a commit hash, if given.
+func loadGitData(dir, remote, branch, ref, commits string, format bool, extraArgs ...string) (*localDiff, error) {
+	if remote == "" {
+		remote = "upstream"
+	}
+	// remote/branch refers directly to the remote branch and does not require a local branch.
+	// branch@{remote} referes to a local branch that has a tag of a remote branch attached to it.
```
Kimchelly, line 629: nit: refers*

Kimchelly: Since git tags are a separate concept, can we use the original wording or add a little bit for the remote branch? The [git docs](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches) (see `Upstream shorthand`) describe branch@{upstream} as a specific syntactic sugar in git

operations/patch_util.go
```diff
@@ -621,17 +621,22 @@ Continue?`, uncommittedChangesFlag), true), nil
...
 	// In the case a range is passed, we only need one commit to determine the base, so we use the first commit.
 	// For details see: https://git-scm.com/docs/gitrevisions
 
-	mergeBase, err := gitMergeBase(dir, branch+"@{upstream}", ref, commits)
+	mergeBase, err := gitMergeBase(dir, remote+"/"+branch, ref, commits)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Error getting merge base, "+
-			"may need to create local branch '%s' and have it track upstream", branch)
+		mergeBase, err = gitMergeBase(dir, branch+"@{"+remote+"}", ref, commits)
```
Kimchelly, line 635: re above: I think this still has to be `branch@{upstream}` because of the `@{upstream}` thing I mentioned above.

Kimchelly (commented):
Non-blocking review since I'm gonna be out.

operations/patch_util.go
ZackarySantana: I did not know this wasn't referencing a git tag! I got a little confused because I was thinking that users probably dont tag their branches with the remote they track- and I was thinking if it's a process we enforce. Thank you for linking the docs for that, helped a lot!

operations/patch_util.go
```diff
@@ -676,6 +681,27 @@ func loadGitData(dir, branch, ref, commits string, format bool, extraArgs ...str
 	}, nil
 }
 
+func gitGetRemote(dir, owner, repo string) (string, error) {
```
malikchaya2, line 684: Can we please add a test for this function?

malikchaya2: or at least break out the regex manipulation bit to its own function and test that?

operations/patch_util.go
```diff
@@ -621,17 +621,22 @@ Continue?`, uncommittedChangesFlag), true), nil
...
+	if remote == "" {
+		remote = "upstream"
+	}
+	// remote/branch refers directly to the remote branch and does not require a local branch.
+	// branch@{remote} refers to the remote branch that "branch" is tracking.
 	// In the case a range is passed, we only need one commit to determine the base, so we use the first commit.
 	// For details see: https://git-scm.com/docs/gitrevisions
 
-	mergeBase, err := gitMergeBase(dir, branch+"@{upstream}", ref, commits)
+	mergeBase, err := gitMergeBase(dir, remote+"/"+branch, ref, commits)
```
malikchaya2, line 633: nit: can we use fmt.Sprintf() here instead of concatenation?

malikchaya2 (approved):
Approved with nit

PR #7313, opened 2023-12-12
DEVPROD-979: terminate spawn hosts that fail to spawn

Kimchelly:
DEVPROD-979

### Description
When a spawn host encounters an issue while Evergreen tries to create it (like having no host capacity available in the AWS region), it will just stay stuck in that state until the user manually terminates it. However, if Evergreen fails to even create the host in EC2, it makes more sense to just terminate the host and notify the user.

This does result in a small UX problem where terminating the host is often faster than the notification system telling the user, so the host that failed to start up may be terminated before the user knows why, but I didn't think it was worth addressing given that auto-terminating the host is convenient and they will know why after the notification is processed.

* Terminate spawn hosts that fail to build immediately if possible.
* Terminate spawn hosts that are stuck in building for too long. This generally won't be necessary, but is a failsafe to ensure that failed spawn hosts get terminated in case the above fails.
* Add notification when host creation fails to spawn a host.
* Fix a minor issue in the host event logs where host creation events are not logged consistently when the host is first inserted into the DB.

### Testing
* Added unit tests.
* Tested in staging that the host hit an error while spawning, was terminated, and sent a notification. Also tested that if it spawned successfully, everything went as per usual.

### Documentation
N/A

### DIFF
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -756,7 +756,6 @@ func MarkStaleBuildingAsFailed(ctx context.Context, distroID string) error {
 	spawnedByTaskKey := bsonutil.GetDottedKeyName(SpawnOptionsKey, SpawnOptionsSpawnedByTaskKey)
 	query := bson.M{
 		distroIDKey:      distroID,
-		UserHostKey:      false,
 		spawnedByTaskKey: bson.M{"$ne": true},
 		ProviderKey:      bson.M{"$in": evergreen.ProviderSpawnable},
 		StatusKey:        evergreen.HostBuilding,
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -1314,6 +1314,13 @@ func UnsafeReplace(ctx context.Context, env evergreen.Environment, idToRemove st
 		if err := toInsert.InsertWithContext(sessCtx, env); err != nil {
 			return nil, errors.Wrapf(err, "inserting new host '%s'", toInsert.Id)
 		}
+		event.LogHostCreated(toInsert.Id)
+		grip.Info(message.Fields{
+			"message":  "inserted host to replace intent host",
+			"host_id":  toInsert.Id,
+			"host_tag": toInsert.Tag,
+			"distro":   toInsert.Distro.Id,
+		})
 		return nil, nil
 	}
 
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -759,7 +759,7 @@ func MarkStaleBuildingAsFailed(ctx context.Context, distroID string) error {
 		spawnedByTaskKey: bson.M{"$ne": true},
 		ProviderKey:      bson.M{"$in": evergreen.ProviderSpawnable},
 		StatusKey:        evergreen.HostBuilding,
-		CreateTimeKey:    bson.M{"$lt": time.Now().Add(-5 * time.Minute)},
+		CreateTimeKey:    bson.M{"$lt": time.Now().Add(-15 * time.Minute)},
 	}
 
 	if distroID != "" {
--- a/model/host/db.go
+++ b/model/host/db.go
@@ -759,7 +759,7 @@ func MarkStaleBuildingAsFailed(ctx context.Context, distroID string) error {
 		spawnedByTaskKey: bson.M{"$ne": true},
 		ProviderKey:      bson.M{"$in": evergreen.ProviderSpawnable},
 		StatusKey:        evergreen.HostBuilding,
-		CreateTimeKey:    bson.M{"$lt": time.Now().Add(-15 * time.Minute)},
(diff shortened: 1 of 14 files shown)


model/host/db.go
```diff
@@ -756,11 +756,10 @@ func MarkStaleBuildingAsFailed(ctx context.Context, distroID string) error {
 	spawnedByTaskKey := bsonutil.GetDottedKeyName(SpawnOptionsKey, SpawnOptionsSpawnedByTaskKey)
 	query := bson.M{
 		distroIDKey:      distroID,
-		UserHostKey:      false,
 		spawnedByTaskKey: bson.M{"$ne": true},
 		ProviderKey:      bson.M{"$in": evergreen.ProviderSpawnable},
 		StatusKey:        evergreen.HostBuilding,
-		CreateTimeKey:    bson.M{"$lt": time.Now().Add(-15 * time.Minute)},
```
malikchaya2, line 763: Why are we lowering it to 5 minutes from 15?

units/provisioning_create_host.go
```diff
@@ -216,9 +216,22 @@ func (j *createHostJob) Run(ctx context.Context) {
...
-			if err := task.AddHostCreateDetails(j.host.StartedBy, j.host.Id, j.host.SpawnOptions.TaskExecutionNumber, j.Error()); err != nil {
-				j.AddError(errors.Wrapf(err, "adding host create error details"))
+		if j.RetryInfo().GetRemainingAttempts() == 0 && j.HasErrors() && (j.host.Status == evergreen.HostUninitialized || j.host.Status == evergreen.HostBuilding) {
+			if j.host.SpawnOptions.SpawnedByTask {
+				if err := task.AddHostCreateDetails(j.host.StartedBy, j.host.Id, j.host.SpawnOptions.TaskExecutionNumber, j.Error()); err != nil {
+					j.AddError(errors.Wrapf(err, "adding host create error details"))
+				}
+			}
+			if j.host.UserHost {
+				// Terminate spawn hosts immediately if they have no creation
```
malikchaya2, line 226: I'm a little bit worried that if some is not subscribed to receive notifications for spawn host outcome, there will be no feedback indicating that something went wrong and it will simply disappear. 
However, I don't have any better solutions (because sending a notification without them being subscribed is not a good idea), and it's still better than the current setup.

units/provisioning_user_data_done.go
```diff
@@ -134,7 +134,6 @@ func (j *userDataDoneJob) Run(ctx context.Context) {
 				TerminateIfBusy:   true,
 				TerminationReason: "failed to mount volume",
 			})
-			terminateJob.SetPriority(100)
```
malikchaya2, line 137: Why is this being removed?

malikchaya2 (changes requested):

model/host/db.go
Kimchelly: I reverted it back to 15. I only lowered it for the sake of testing a bit faster.

units/provisioning_create_host.go
Kimchelly: If they're subscribed, they'll receive a notification regardless of whether the host gets terminated here or later (when it hits the stale timeout), so they do get feedback as long as they're subscribed. The issue is that the host may be terminated before they get the notification and before Spruce displays the event log, which may result in a couple minutes of confusion on where their requested spawn host went. Regardless, they can also see their terminated host in the UI for a few minutes after it's terminated.

I can remove this for now since the spawn host will eventually go away, it'll just take 15 minutes of waiting.

units/provisioning_user_data_done.go
Kimchelly: The Amboy job priority setting never did anything, so it has no effect (and has never had an effect). I didn't want to keep a piece of code that people may believe has an effect when it doesn't.

We could also remove it from all the other places that set it for no reason, but I don't consider it that urgent, it's just a very small piece of tech debt.

units/provisioning_create_host.go
malikchaya2: I'm less worried about the couple of minutes of confusion, I'm more worried about the person who is not subscribed to notifications. Because those people before had the extra step of terminating it themselves so they knew what was happening, but now I thought there would be no feedback or sign of what's happening. 

But I had forgotten that they can still see their terminated host in the UI for a few minutes after it's terminated. In that case I'm less worried about it.

units/provisioning_user_data_done.go
malikchaya2: Makes sense, it's probably worth a ticket to do eventually.

malikchaya2 (approved):

units/provisioning_user_data_done.go
Kimchelly: I filed DEVPROD-3234.

PR #7314, opened 2023-12-12
DEVPROD-3166 Commit Queue should use the latest module revision 

malikchaya2:
DEVPROD-3166

### Description
The commit queue (both the CLI queue and the PR queue after https://github.com/evergreen-ci/evergreen/pull/7310, unless auto update is set to true, gets the module revision from the base manifest. Instead, we should fetch the latest.

Otherwise, if the module has moved ahead, it can be outdated.

### Testing
Tested on staging

### DIFF
--- a/model/version.go
+++ b/model/version.go
@@ -745,9 +745,10 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 		}
 	}
 
+	isMergePatch := v.Requester == evergreen.MergeTestRequester
 	modules := map[string]*manifest.Module{}
 	for _, module := range moduleList {
-		if isPatch && !module.AutoUpdate && baseManifest != nil {
+		if isPatch && !isMergePatch && !module.AutoUpdate && baseManifest != nil {
 			if baseModule, ok := baseManifest.Modules[module.Name]; ok {
 				modules[module.Name] = baseModule
 				continue
--- a/model/version.go
+++ b/model/version.go
@@ -745,7 +745,7 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


model/version.go
```diff
@@ -745,9 +745,10 @@ func constructManifest(v *Version, projectRef *ProjectRef, moduleList ModuleList
 		}
 	}
 
+	isMergePatch := v.Requester == evergreen.MergeTestRequester
```
bynn, line 748: https://github.com/evergreen-ci/evergreen/blob/cad3a44bc6a2eff74ce9a742960f3bb8dd102962/globals.go#L1023

we can probably just use this function instead

bynn (approved):
with nit

ablack12 (commented):
Do we need to update any module documentation to indicate this?

malikchaya2:
> Do we need to update any module documentation to indicate this?

This is already sort of implied in the docs. It mentions "manual patches" using the git revisions from its base version, with, what I would think would be the implication that non manual patches (like the commit queue patches) don't do that. I can add a sentence saying that patches created by the commit queue already to that automatically at the end, but I don't know if we want to spread commit queue documentation into our docs more considering that it will go away. 

> For manual patches and GitHub PRs, by default, the git revisions in the
> version manifest will be inherited from its base version (i.e. the mainline commit version of the patch's base git revision). 
> You can change the git revision for modules by setting a module manually with 
> [evergreen set-module](../CLI.md#operating-on-existing-patches) or
> by specifying the `auto_update` option (as described below) to use the
> latest revision available for a module. The full hierarchy of how
> module revisions are determined is available in the [git.get_project](Project-Commands.md#module-hash-hierarchy)
> docs.

ablack12 (approved):

PR #7317, opened 2023-12-13
DEVPROD-714: remove default distro setting for Docker host.create

Kimchelly:
DEVPROD-714

### Description
The agent validates the user's host.create inputs, but if the user uses Docker host.create without specifying a distro, it will fail because the validation logic tries to query the admin settings, even though the agent doesn't have access to the DB. Rather than try to fix a very small convenience feature and given the fact that the default distro is blank in both staging and prod, I decided it's better to remove the default distro and have the user always specify one for Docker host.create.

### Testing
Existing tests pass.

### Documentation
It's never explicitly stated how the `distro` parameter to host.create interacts with Docker, so I clarified the wording so the user knows they have to give a distro for Docker host.create.

### DIFF
--- a/config_cloud.go
+++ b/config_cloud.go
@@ -331,6 +331,5 @@ const (
 
 // DockerConfig stores auth info for Docker.
 type DockerConfig struct {
-	APIVersion    string `bson:"api_version" json:"api_version" yaml:"api_version"`
-	DefaultDistro string `bson:"default_distro" json:"default_distro" yaml:"default_distro"`
+	APIVersion string `bson:"api_version" json:"api_version" yaml:"api_version"`
 }
--- a/docs/Project-Configuration/Project-Commands.md
+++ b/docs/Project-Configuration/Project-Commands.md
@@ -480,7 +480,7 @@ Parameters:
 
 ## host.create
 
-`host.create` starts a host from a task.
+`host.create` starts a host or a Docker container from a task.
 
 ``` yaml
 - command: host.create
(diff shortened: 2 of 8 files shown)


docs/Project-Configuration/Project-Commands.md
```diff
@@ -520,16 +520,18 @@ Agent Parameters:
...
 -   `aws_access_key_id` - AWS access key ID. May set to use a
     non-default account. Must set if `aws_secret_access_key` is set.
 -   `aws_secret_access_key` - AWS secret key. May set to use a
     non-default account. Must set if `aws_access_key_id` is set.
 -   `device_name` - name of EBS device
--   `distro` - Evergreen distro to start. Must set `ami` or `distro` but
-    must not set both. Note that the distro setup script will not run for 
-    hosts spawned by this command, so any required initial setup must be done manually.
+-   `distro` - Evergreen distro to start. For the `ec2` provider, must set
+    either `ami` only) or `distro` but must not set both. For the `docker`
```
hadjri, line 531: Nit: need to remove unclosed parentheses here

config_cloud.go
```diff
@@ -331,6 +331,5 @@ const (
 
 // DockerConfig stores auth info for Docker.
 type DockerConfig struct {
-	APIVersion    string `bson:"api_version" json:"api_version" yaml:"api_version"`
-	DefaultDistro string `bson:"default_distro" json:"default_distro" yaml:"default_distro"`
+	APIVersion string `bson:"api_version" json:"api_version" yaml:"api_version"`
```
hadjri, line 334: Can we flatten this struct into a single string now, something like `DockerVersion`? Or do we anticipate further fields here

docs/Project-Configuration/Project-Commands.md
Kimchelly: Nice catch, fixed.

hadjri (approved):

PR #7322, opened 2023-12-14
DEVPROD-3223: add limit on max tasks per version

Kimchelly:
DEVPROD-3223

### Description
Evergreen already has a guard in place to prevent a single call to generate.tasks from adding too many tasks at once. However, it does not check if the entire version overall just has too many tasks. It's hypothesized that multiple calls to generate.tasks are incrementally adding too many tasks to some patches, which may be contributing to system instability and ops alerts right now.

* Add admin setting to limit the total number of tasks in a single version.
* Add validation to enforce total number of tasks per version limit.

### Testing
* Tested admin setting can be added via UI.
* Tried submitting some test patches with a low max task limit to verify they got validation errors.
* Added unit tests for validation and admin setting.

### Documentation
N/A


service/templates/admin.html
```diff
@@ -554,6 +555,23 @@ <h2 class="modal-title">[[modalTitle]]</h2>
...
+									<md-card-title-text>
+										<span>Task Limits</span>
+									</md-card-title-text>
+									<md-button ng-click="clearSection('task_limits')">
+										<i class="fa fa-trash"></i>
+									</md-button>
+								</md-card-title>
+								<md-card-content>
+									<md-input-container class="control" style="width:45%;">
+										<label>Max Tasks per Version</label>
```
ZackarySantana, line 570: Do we need a ticket for Spruce changes?

validator/project_validator.go
```diff
@@ -958,6 +961,17 @@ func checkRunOn(runOnHasDistro, runOnHasContainer bool, runOn []string) []Valida
 	return nil
 }
 
+func validateProjectLimits(_ context.Context, settings *evergreen.Settings, project *model.Project, _ *model.ProjectRef, _ bool) ValidationErrors {
+	errs := ValidationErrors{}
+	if settings.TaskLimits.MaxTasksPerVersion > 0 && len(project.Tasks) > settings.TaskLimits.MaxTasksPerVersion {
```
malikchaya2, line 966: Will this already include all the generated tasks?

service/templates/admin.html
bynn: there isn't an admin settings page in spruce right now so this is just to update our legacy admin page. I believe there is planned work to make a spruce admin page but no need to ticket for something like this.

validator/project_validator.go
Kimchelly: Yes, project validation is always called on the resulting project from generate.tasks.

malikchaya2 (approved):

PR #7323, opened 2023-12-14
DEVPROD-731 Add include limits to admin settings

bynn:
DEVPROD-731

### Description
add limit for the number of includes users can specify 
image

### Testing
local evergreen saves 
unit tests 
staging with low include number to fail 
image
image

### DIFF
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -679,6 +679,21 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 			err = errors.New("trying to open include files with empty options")
 			return nil, errors.Wrapf(err, LoadProjectError)
 		}
+
+		// Validate that there are less includes than the limit.
+		settings, err := evergreen.GetConfig(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, LoadProjectError)
+		}
+		if settings == nil {
+			err = errors.New("no settings found")
+			return nil, errors.Wrap(err, LoadProjectError)
+		}
+		if settings.TaskLimits.MaxIncludesPerVersion > 0 && len(intermediateProject.Include) > settings.TaskLimits.MaxIncludesPerVersion {
+			err = errors.Errorf("project's total number of includes (%d) exceeds maximum limit (%d)", len(intermediateProject.Include), settings.TaskLimits.MaxIncludesPerVersion)
+			return nil, errors.Wrap(err, LoadProjectError)
+		}
+
 		wg := sync.WaitGroup{}
 		outputYAMLs := make(chan yamlTuple, len(intermediateProject.Include))
 		includesToProcess := make(chan Include, len(intermediateProject.Include))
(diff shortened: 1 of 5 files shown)


model/project_parser.go
```diff
@@ -679,6 +679,21 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
+		// Validate that there are less includes than the limit.
+		settings, err := evergreen.GetConfig(ctx)
+		if err != nil {
+			return nil, errors.Wrap(err, LoadProjectError)
+		}
+		if settings == nil {
+			err = errors.New("no settings found")
+			return nil, errors.Wrap(err, LoadProjectError)
+		}
+		if settings.TaskLimits.MaxIncludesPerVersion > 0 && len(intermediateProject.Include) > settings.TaskLimits.MaxIncludesPerVersion {
```
ablack12, line 692: Can we add this as a project error rather than validating this here? https://github.com/evergreen-ci/evergreen/blob/f1ad47d558369748b427d3f79a69ffe006aead04/validator/project_validator.go#L123

bynn: I don't think that's possible because that's validation on the project but at that stage, all the includes are already merged into the project. Includes only exist on the parser project level

model/project_parser.go
```diff
@@ -679,6 +679,21 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 			err = errors.New("trying to open include files with empty options")
 			return nil, errors.Wrapf(err, LoadProjectError)
 		}
+
+		// Validate that there are less includes than the limit.
```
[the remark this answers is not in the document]
bynn: :shipit:

model/project_parser.go
ablack12: Oof okay that's a good point. Can we move this into createIntermediateProject maybe?

ablack12 (changes requested):

model/project_parser.go
ablack12: Actually maybe that's okay. We could have this for loop limit to x number of files so we don't read too many files, but then we could still run project validation on the files that we _do_ read and also include an error that's like "you asked for too many files so we only gave you the first x"

bynn: updated so that the catcher returns an error when max includes is exceeded if max includes is set

model/project_parser.go
```diff
@@ -679,12 +679,30 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
 		outputYAMLs := make(chan yamlTuple, len(intermediateProject.Include))
 		includesToProcess := make(chan Include, len(intermediateProject.Include))
 
-		for _, path := range intermediateProject.Include {
-			includesToProcess <- path
+		// Only process includes until max limit is reached.
+		catcher := grip.NewBasicCatcher()
+		numIncludes := len(intermediateProject.Include)
+		if settings.TaskLimits.MaxIncludesPerVersion > 0 && numIncludes > settings.TaskLimits.MaxIncludesPerVersion {
+			catcher.Add(errors.Errorf("project's total number of includes (%d) exceeds maximum limit (%d)", numIncludes, settings.TaskLimits.MaxIncludesPerVersion))
```
ablack12, line 700: catcher.Errorf

ablack12 (approved):

PR #7324, opened 2023-12-15
DEVPROD-754: Add service flag for deprecating legacy distros page

minnakt:
DEVPROD-754

### Description
Adds a flag for auto-redirecting to the new distro page. The flag will not be turned on until some time in January and I will also need to put up a banner.

Screenshot 2023-12-15 at 4 27 49 PM
<br /><br />

I also added a "Open Page in the New UI" button on the distros page to encourage the transition during this time. I wanted to link to a particular distro from this button, but I couldn't figure out how to do that (and I don't know if it's possible).

Note that this PR depends on the Spruce PR because the Spruce PR adds a new `/distros` route to redirect to.

### Testing
- On staging


service/distros.go
```diff
@@ -25,6 +25,16 @@ import (
 
 func (uis *UIServer) distrosPage(w http.ResponseWriter, r *http.Request) {
 	u := MustHaveUser(r)
+
+	flags, err := evergreen.GetServiceFlags(r.Context())
+	if err != nil {
+		gimlet.WriteResponse(w, gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "retrieving admin settings")))
```
hadjri, line 31: We'll need to return after this line

service/distros.go
```diff
@@ -25,6 +25,16 @@ import (
...
+	}
+	if flags.LegacyUIDistroPageDisabled {
+		newUIProjectsLink := fmt.Sprintf("%s/distros", uis.Settings.Ui.UIv2Url)
+		http.Redirect(w, r, newUIProjectsLink, http.StatusPermanentRedirect)
```
hadjri, line 35: also need to return after this line

service/distros.go
```diff
@@ -25,6 +25,16 @@ import (
 
 func (uis *UIServer) distrosPage(w http.ResponseWriter, r *http.Request) {
 	u := MustHaveUser(r)
+
+	flags, err := evergreen.GetServiceFlags(r.Context())
+	if err != nil {
+		gimlet.WriteResponse(w, gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "retrieving admin settings")))
+	}
+	if flags.LegacyUIDistroPageDisabled {
+		newUIProjectsLink := fmt.Sprintf("%s/distros", uis.Settings.Ui.UIv2Url)
```
hadjri, line 34: We can remove this line and move `spruceLink` on L89 right before this block to prevent initializing the same value twice

hadjri (changes requested):

service/distros.go
minnakt: oops thank you!

hadjri (approved):

minnakt:
Note: I notice that Chrome seems to be caching the pages, leading to the changes from admin page not being registered sometimes. When that happens, I enable "Disable cache" in the Network tab and refresh the page with the console still open. (The console must be left open because the "Disable cache" setting is only active if it's open.) This will invalidate the cache and register the changes.

Screenshot 2023-12-20 at 11 20 33 AM

You can also open an incognito tab and that should also correctly reflect the changes.

PR #7333, opened 2023-12-18
DEVPROD-2912: Handle family trigger conversions when saving subscriptions

minnakt:
DEVPROD-2912

### Description
Right now, the frontend always is always assigning the "family-" trigger to VERSION resource type subscriptions. The "family-" triggers were originally created to deal with child patches and they listen for a `CHILDREN_FINISHED` event. This causes VERSION resource type subscriptions for periodic builds/commits to be broken because they will never have child patches and therefore will never dispatch a `CHILDREN_FINISHED` event.

I'm changing the frontend to always assign the regular trigger to VERSION resource type subscriptions. This PR makes it so that we re-evaluate the subscription before saving it to the DB, converting triggers to family triggers if necessary.

### Testing
- Added a bunch of unit tests
- Tested on staging:
Screenshot 2023-12-18 at 5 06 42 PM


rest/data/subscription.go
```diff
@@ -12,6 +14,45 @@ import (
...
+	if s.OwnerType == event.OwnerTypeProject {
+		for _, selector := range s.Selectors {
+			if selector.Type == event.SelectorRequester {
+				requester = selector.Data
+				break
+			}
+		}
+	}
+	// Handle personal subscriptions.
+	if s.OwnerType == event.OwnerTypePerson {
```
hadjri, line 33: nit: can we make this an else if

hadjri (approved):
Approved w nit

minnakt:
Fairly sure that the Parsley and Spruce failures are unrelated, so I will merge

PR #7334, opened 2023-12-18
DEVPROD-3230 Fix shouldKill panics when makeTaskConfig errors 

malikchaya2:
DEVPROD-3230

### Description
The shouldKill function relies on the taskConfig to know if the task is not part of a taskgroup and should therefore kill processes. If makeTaskConfig fails, the taskConfig will be nil, and then when it eventually get's to a.shouldKill() via handleSetupError, it will panic when it tries to check tc.taskConfig.TaskGroup. 

### Testing
There's no good way to test this, and the changes are simple enough so I think we are okay without it.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1068,13 +1068,13 @@ func setEndTaskFailureDetails(tc *taskContext, detail *apimodels.TaskEndDetail, 
 
 }
 
-func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string, additionalKillOverride bool) {
+func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string, additionalShouldKillCheck bool) {
 	logger := grip.NewJournaler("killProcs")
 	if tc.logger != nil && !tc.logger.Closed() {
 		logger = tc.logger.Execution()
 	}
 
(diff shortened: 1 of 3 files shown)


hadjri (commented):
A few questions:

- How did you determine that this was the issue causing panics during the outage, and how did we end up in this panicking state?
- Have we confirmed that all hosts that were stuck panicking have been fixed?

malikchaya2:
> * How did you determine that this was the issue causing panics during the outage, and how did we end up in this panicking state?

I determined that this is what caused the panics because the stack trace leads us to handleSetupError after makeTaskConfig. We know why we ended up in the panicking state, but we don't know with certainty which call in makeTaskConfig errored because we don’t have logs in honeycomb since when the agent panics and dies, the trace data is lost. 

However, if I were to speculate on the cause: this was happening at a time when evergreen was down, so makeTaskConfig probably errored either because the agent’s last attempt to talk to evergreen (GetNextTask) failed and makeTaskConfig didn’t have the task information necessary to fetch the data  and therefore errored, or it simply couldn’t talk to evergreen when it was trying to fetch the data, and errored because of that. Accessing tc.taskConfig.TaskGroup in killprocs has been in production since September 12, and I’m guessing this is the first time we hit this edge case.

> * Have we confirmed that all hosts that were stuck panicking have been fixed?
Looking at splunk, this hasn't happened since 12/19.

hadjri (approved):

PR #7335, opened 2023-12-19
DEVPROD-726 Add validation for module usage in the git.get_project command

ZackarySantana:
DEVPROD-726

### Description
It was possible for the project configuration to have an bad state in which a user uses a task that requires a module but the build variant does not use the module, so it silently errors without notifying the user. This adds logic to our validate route/package to produce an error when their configuration is in this state.

### Testing
I manually tested it with an improper configuration- applied the changes and it worked. I also added some unit tests.


validator/project_validator.go
```diff
@@ -1951,6 +1952,40 @@ func validateGenerateTasks(p *model.Project) ValidationErrors {
 	return validateTimesCalledPerTask(p, ts, evergreen.GenerateTasksCommandName, 1, Error)
 }
 
+func validateModuleUsageInGitGetProject(p *model.Project) ValidationErrors {
+	var errs ValidationErrors
+
+	for _, bv := range p.BuildVariants {
```
malikchaya2, line 1958: We have a function called `bvsWithTasksThatCallCommand`  that creates a mapping from build variants to tasks that run a given command. Can that (or that with some modification) be reused here?

malikchaya2 (commented):
None blocking review since I will be out

validator/project_validator.go
ZackarySantana: I just adapted it, it's a bit longer _but_ it now covers the edge case of pre & post functions which it did not before! 
I'm new to our project yaml and how users might use it, did not think of checking those! Thank you for the suggestion :)

model/project.go
```diff
@@ -1515,6 +1515,26 @@ func (p *Project) FindProjectTask(name string) *ProjectTask {
 	return nil
 }
 
+// FindAllProjectTasksWithName finds either the individual task or
+// all tasks associated with a task group.
+func (p *Project) FindAllProjectTasksWithName(name string) []*ProjectTask {
```
malikchaya2, line 1520: This is not needed anymore

validator/project_validator.go
```diff
@@ -1951,6 +1952,58 @@ func validateGenerateTasks(p *model.Project) ValidationErrors {
...
+	bvs, _, err := bvsWithTasksThatCallCommand(p, "git.get_project")
+	if err != nil {
+		errs = append(errs, ValidationError{
+			Message: fmt.Sprintf("build variants could not be mapped to tasks that contain 'git.get_project' for project %s", p.Identifier),
+			Level:   Error,
+		})
+		return errs
+	}
+
+	for bvName, bv := range bvs {
```
malikchaya2, line 1967: nit: I think this would be clearer if bvs here was named something `bvTaskCmds` or something to indicate that this is in a special format.

validator/project_validator_test.go
```diff
@@ -2437,6 +2437,65 @@ func TestCheckTaskCommands(t *testing.T) {
...
+		"Valid git.get_project": func(t *testing.T, p *model.Project) {
+			errs := validateModuleUsageInGitGetProject(p)
+			assert.Len(t, errs, 0)
+		},
+		"Invalid git.get_project": func(t *testing.T, p *model.Project) {
+			// Add the task to a build variant that does not have all the modules needed.
+			p.BuildVariants[1].Tasks = append(p.BuildVariants[1].Tasks, model.BuildVariantTaskUnit{Name: "test"})
+			errs := validateModuleUsageInGitGetProject(p)
+			require.Len(t, errs, 1)
+			assert.Contains(t, errs[0].Message, "bar")
```
malikchaya2, line 2451: Since we know what the error will be in full, I'd preferrer adding the whole error message 
`"build variant 'bv2' with task 'test' with 'git.get_project' command uses module/revision 'bar' that is not present in build variant"` or a bigger part of it. It makes it easier to read the test and future proof against changes.

malikchaya2 (approved):
Approved with some nits. 

Apologies for the late review. There were no notifications for this. In the future, please feel free to re-request review after changes, or add evg-app back if the person is away and the review is non-blocking.

validator/project_validator.go
ZackarySantana: renamed bv and bvs to be more descriptive on the data they hold!

PR #7341, opened 2023-12-20
DEVPROD-1122 Populate next task id for previous stepback tasks

ZackarySantana:
DEVPROD-1122

### Description
Right now, stepback bisection does not populate the parent tasks (aka the last failing, last passing tasks) with what task is executing next in stepback. This adds the appropriate saving to those tasks to enable card metadata about a task's next stepback iteration.

As well, this adds to the GraphQL queries/models/data (please excuse the outrageous file changed, it is mostly gqlgen doing it's thing!)

### Testing
Retest in staging that stepback works as intended. Also added GraphQL tests.


model/task_lifecycle.go
```diff
@@ -628,6 +628,34 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
 	if nextTask == nil {
 		return errors.Errorf("midway task could not be found for tasks '%s' '%s'", s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId)
 	}
+	// Log our next task to our last passing and last failing.
+	for _, id := range []string{s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId} {
+		if t, err := task.FindOneId(id); err != nil {
```
malikchaya2, line 633: nit: This might do well with the logic from 633 --> 656 extracted into a helper called `setStepbackTaskId` or something like that and calling it on `s.LastFailingStepbackTaskId` and `s.LastPassingStepbackTaskId`

malikchaya2 (commented):
dropping a quick comment, but non blocking review

model/task_lifecycle.go
```diff
@@ -628,6 +628,34 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
...
+			}
+			// If the next task is already set, then this task has been used for stepback before
+			// and should not be updated (e.g. the right bound will have the same next stepback taskid
+			// as the left bound narrows the scope in which stepback happens).
+			if t.StepbackInfo.NextStepbackTaskId != "" {
+				continue
+			}
+			info := *t.StepbackInfo
+			info.NextStepbackTaskId = nextTask.Id
+			if err = task.SetStepbackInfo(s.LastFailingStepbackTaskId, info); err != nil {
```
ZackarySantana, line 653: The argument here should be id instead of s.LastFailingStepbackTaskId.

model/task_lifecycle.go
```diff
@@ -628,6 +628,34 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
 	if nextTask == nil {
 		return errors.Errorf("midway task could not be found for tasks '%s' '%s'", s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId)
 	}
+	// Log our next task to our last passing and last failing.
+	for _, id := range []string{s.LastFailingStepbackTaskId, s.LastPassingStepbackTaskId} {
```
hadjri, line 632: I think it'll definitely be necessary for tasks to be able to both point to the next task that they activated in the stepback process, as well as point to the previous task in stepback that activated them.

I'm noticing now that [setting NextStepbackTaskId](https://github.com/evergreen-ci/evergreen/pull/7341/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9L641) is only getting set to the ID of the task itself, which is redundant. We should be setting this field for `t` instead.

Given that `NextStepbackTaskId` only exists for auditability & UI purposes, I ultimately think we can simplify things a bit, remove this loop and instead:
1. Introduce a new field to StepbackInfo, call it `PrevStepbackTaskId` and have it point to the previous task in the stepback process
2. Split `SetStepbackInfo` into two functions -- one that just sets `NextStepbackTaskId` for a task, and another that sets the other 3 fields.  Then, [here](https://github.com/evergreen-ci/evergreen/pull/7341/files#diff-4bef865c41d6d16e1056e61806c23fb8bbc896b3739472dd60f0dd463ef5d2b9R668) we should call the former function on the original task, and the latter function on `nextTask`

hadjri (changes requested):

model/task_lifecycle.go
ZackarySantana: It used to set the next task as it's own id, but now it never gets set since this loop makes a copy and sets the next stepback id inside.

I think adding the two functions makes sense and will make it cleaner. I don't think we need PrevStepbackTaskId. 

If we have task 1 -> 10 (where 10 is the latest task) and we only know 1 passed, 10 failed (leaving 2-9 unknown). Bisect stepback will choose 5, in which both tasks are the previous tasks. While processing 5, it will set the next task id for 1 & 10 to 5. Then If 5 fails, then it will run stepback on 1 -> 5. In which 5 will not have a next stepback id, because it was the latest in the line of stepback. While 1 will have one, because while processing 5; it got set.

Does this make sense? We can also huddle/zoom to talk about it

model/task/task_test.go
```diff
@@ -3140,7 +3140,7 @@ func TestSetStepbackInfo(t *testing.T) {
 		NextStepbackTaskId:        "t4",
 	}
 
-	assert.NoError(t, task.SetStepbackInfo(s))
+	assert.NoError(t, SetStepbackInfo(task.Id, s))
```
ZackarySantana, line 3143: Need to update these tests now!

ZackarySantana: And the stepback tests should be added to test the functionality

model/task_lifecycle.go
ZackarySantana: Talked a bit ofline, going to add the prev stepback task not for the algorithm, but for usability in the UI (navigating back and forth between stepback iterations)

model/task/task.go
```diff
@@ -1510,17 +1512,50 @@ func taskAbortUpdate(reason AbortInfo) bson.M {
...
+	)
+}
+
+// SetNextStepbackId sets the NextStepbackTaskId for a given task only
+// if it does not have it set yet.
+func SetNextStepbackId(taskId string, s StepbackInfo) error {
+	return UpdateOne(
+		bson.M{
+			IdKey: taskId,
+			StepbackInfoKey: bson.M{
```
hadjri: We don't need this query parameter, all we need is `IdKey`

model/task_lifecycle.go
```diff
@@ -628,32 +630,17 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
...
-			}
-			info := *t.StepbackInfo
-			info.NextStepbackTaskId = nextTask.Id
-			if err = task.SetStepbackInfo(s.LastFailingStepbackTaskId, info); err != nil {
-				return errors.Wrapf(err, "setting stepback info for task '%s'", nextTask.Id)
-			}
-		}
+	s.NextStepbackTaskId = nextTask.Id
+	// Store our next task to our last passing and last failing.
+	if err := task.SetNextStepbackId(s.LastFailingStepbackTaskId, s); err != nil {
```
hadjri, line 635: We don't need to set the next stepback task Id field on both `LastFailingStepbackTaskId` and `LastPassingStepbackTaskId`, just the current task `t`, as the field should just be pointing to the next task being activated in the stepback process

hadjri (changes requested):

model/task_lifecycle.go
hadjri: We just need `NextStepbackTaskId ` and `PrevStepbackTaskId ` to construct a linkage of activation between the stepback tasks (i.e. 0<->4<->2<->3 for the first example in this [doc](https://docs.google.com/document/d/17bMlUDupaom5p9uIL1UbECD_XNf7POWhDNf0efv_VZk/edit))

ZackarySantana: I had assumed that for the case of the first iteration, we want to link the last passing task as well to the stepback- I removed it now, so it can only be tracked from that head task that initiated stepback

hadjri (changes requested):

graphql/task_resolver.go
```diff
@@ -675,6 +675,38 @@ func (r *taskResolver) VersionMetadata(ctx context.Context, obj *restModel.APITa
 	return apiVersion, nil
 }
 
+// StepbackInfo is the resolver for the stepbackInfo field.
+func (r *taskResolver) StepbackInfo(ctx context.Context, obj *restModel.APITask) (*StepbackInfo, error) {
```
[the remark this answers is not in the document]
ZackarySantana: I switched everything over to the same name, and I added in our dev guide the gqlgen documentation link- before this I thought it was an internal tool completely...

hadjri (approved):
LGTM once tests pass

PR #7342, opened 2023-12-20
DEVPROD-754: Add deprecation banner on legacy distros page

minnakt:
DEVPROD-754

### Description
This PR adds a warning banner on the legacy distros page. Usually we might use the admin banner for this, but the distros page receives such limited usage that it doesn't make sense to show a banner across the entire website. It also makes sense to keep the admin banner available in case an incident occurs.

Note: I changed some CSS so I regenerated the dist CSS files.

### Question
* Maybe the banner should not be dismissible? I'm not sure. 🤔

### Testing
- Confirmed changes locally:
Screenshot 2023-12-20 at 2 02 38 PM


minnakt:
oops heh 🧍‍♂️

hadjri (approved):

PR #7350, opened 2023-12-28
DEVPROD-1122 Set stepback information only when continuing bisect stepback or it has finished

ZackarySantana:
DEVPROD-1122

### Description
This moves some logic to block setting some fields when stepback is happening

### Testing
Tested via staging manually with the UI ticket that uses this code


hadjri (approved):
LGTM once tests pass

PR #7352, opened 2023-12-28
DEVPROD-3233 build-and-push task

ybrill:
[DEVPROD-3233](https://jira.mongodb.org/browse/DEVPROD-3233)

### Description
In order that every build of evergreen should have built its own executables without stepping on each other's toes this adds patchable tasks that build the binary and push to S3 under a prefix that's unique to the version (as opposed to the current solution which is scoped to the SHA of the commit). The plan is for k8s to pass the version_id to the executable when it runs so it knows where to pull the clients from. There will be a follow-on PR to read it from the environment and use it in the S3 prefix (adding it to ldflags or using [go:embed](https://gobyexample.com/embed-directive) have a certain appeal, but passing it explicitly seems better).

Once the switchover to k8s is complete we can decommission the dist/upload-clis tasks. In the meantime we'll leave them so  deploys continue to work in the meantime.

Since we were starting from scratch it seemed like a worthy experiment to cross compile in a display task so they could build in parallel. It's hard to tell how much speedup this will give since the main bottleneck will be scheduling and right now the queues are pretty empty. If it's not considerably faster than we can go back to doing it all in one task. Preliminary results are looking good: [this run took ~7 minutes](https://spruce.mongodb.com/task/evergreen_ubuntu2204_display_build_and_push_patch_33cbcb7b986a68dadb74bc25a56e96b4869cb999_658dca740ae6067d6c4032e3_23_12_28_19_20_25/execution-tasks?execution=0&sorts=STATUS%3AASC).

### Testing
I'll add this task to the PR. It earlier testing it was able to upload the CLI to S3.


hadjri (approved):
LGTM

ablack12 (commented):
Overall LGTM -- Don't know a lot about this but do we need to worry about signing for this new version of the task? Will we have a staging-only version of the task that can be run in patches later on? (Also if we're getting close to switching over to the new deploy process we should set up a cross-team meeting to make sure we feel safe about it, and if we're not close to switching over I'm wondering if we want to keep this task patch only for a while)

ybrill:
@ablack12 those are all very good points.
Added a task for staging and unsigned. Made all of them `activated: false` for now (I think that accomplishes the same as `patch_only`?). It'll be on the order of weeks before we're ready to start serving prod traffic, but when we're getting close we'll definitely loop your team in.

Separated each of them out into their own variants since the display tasks in a single variant couldn't have overlapping tasks (an alternative here was to have multiple copies of each task, but then the thing with using the task name as the make target wouldn't have worked). The result looks like [this patch](https://spruce.mongodb.com/patch/659475ed1e2d177ca379ca41/configure/tasks).

self-tests.yml
```diff
@@ -750,6 +789,51 @@ buildvariants:
       - name: ".agent .test"
       - name: ".cli .test"
 
+  - name: build-and-push
```
ablack12, line 792: do we want this main task to also be in its own variant, or should this one be in the main variant? (I thought long and hard about how to avoid the multiple variants but i see what you're saying haha, i also think that the "build-staging" tag is pretty elegant)

ablack12 (approved):

self-tests.yml
ybrill: Thanks! 😄 
An alternative could be to go back to having a single task which builds serially, and we could have different variants of it for non-signing and staging, but parallelizing it does seem like a substantive improvement (today it's been like 3 minutes to build the whole enchilada).

I'm okay to put the main one in the ubuntu variant, though I kind of like having it separate (though not that there are three of them), since it separates testing the thing from building it. It'll be good to be able to see at a glance when the build has finished. On the other hand, a lot of variants might be too much clutter on the patch configuration page. I'm thinking to leave the main one in its own variant.

ybrill:
afaict the `test-service-graphql` is because the order of [these two lines](https://github.com/evergreen-ci/evergreen/blob/63fca24149abd21fd6e414d007d6cc04e384d0e1/graphql/tests/task/taskLogs/results.json#L85-L98) is non-deterministic: the timestamps are the same.

I'm going to merge this now.

PR #7353, opened 2023-12-29
DEVPROD-67: Implement log directory watcher in the agent (WIP)

julianedwards:
DEVPROD-NNNNN

### Description

### Testing

### Documentation


agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,299 @@
...
+	defer func() {
+		h.logger.Task().Critical(recovery.HandlePanicWithError(recover(), nil, "test log file follower"))
+	}()
+	defer h.wg.Done()
+
+	h.once.Do(h.getSpecFile)
+
+	h.logger.Task().Infof("new test log file '%s' found, initiating automated ingestion", event.Path)
+
+	logPath, err := filepath.Rel(h.dir, event.Path)
```
ybrill, line 202: what is this for?

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,299 @@
...
+type testLogSpec struct {
+	SchemaVersion string        `yaml:"schema_version"`
+	Format        testLogFormat `yaml:"format"`
+}
+
+const testLogSpecFilename = "log_spec.yaml"
+
+func (s testLogSpec) getParser() taskoutput.LogLineParser {
+	switch s.Format {
+	case testLogFormatTextTimestamp:
```
ybrill, line 263: is this the resmoke format?

agent/internal/taskoutput/test_log.go
julianedwards: We do not want to include the reserved directory name in the name (path in S3) that the log will be stored under. I will add a comment.

agent/internal/taskoutput/test_log.go
julianedwards: Yes, this is the format that resmoke will be writing log lines in.

We are starting with support for two formats: 
- `text`: lines are appended as is and timestamps are added upon ingestion by the agent
- `text-timestamp`: lines should be prefixed with the nanosecond UNIX timestamp and are parsed accordingly upon ingestion

You can imagine that we expand these to include priority and other types of accepted formats.

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	started := make(chan struct{})
+	go func() {
+		h.watcher.Wait()
+		close(started)
+	}()
+	startErr := make(chan error)
+	go func() {
+		startErr <- h.watcher.Start(time.Millisecond)
```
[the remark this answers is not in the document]
julianedwards: Yes, good catch! I hate the way this start mechanism returns an error but also blocks on success, instead of spawning a goroutine and returning immediately. Alas, this strange workaround.

agent/internal/taskoutput/test_log.go
ybrill: Ah I get it now.
It took me forever to even understand what Rel _does_

ybrill (approved):

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,330 @@
...
+				seenFiles[path] = true
+				h.followFileCount++
+				h.wg.Add(1)
+				go h.followFile(workerCtx, path)
+			}
+
+			return nil
+		})
+		if err != nil {
+			h.logger.Execution().Debug("context canceled, exiting test log directory watcher")
```
ybrill, line 160: how would cancelling the context make `err` non-nil?

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,330 @@
...
+// watch recursively watches the reserved test log directory for newly created
+// log files. When a new file is detected, an asynchronous file follower is
+// created to ingest the log data as it is written for the duration of the
+// task.
+func (h *testLogDirectoryHandler) watch(ctx context.Context) {
+	defer func() {
+		h.logger.Execution().Critical(recovery.HandlePanicWithError(recover(), nil, "test log directory watcher"))
+	}()
+
+	workerCtx, cancel := context.WithCancel(ctx)
```
ybrill, line 133: why do we need an internal context for watch?

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,330 @@
...
+		MustExist: true,
+		Follow:    true,
+		// Setting this field may lead to some data loss in the case of
+		// non-atomic line writes: if the last read line does not
+		// terminate with a newline character (e.g., the follower
+		// exits early or the test process crashes), it will not get
+		// returned. This is ok because the complete line must be read
+		// to guarantee proper line format for the parser (assumming
+		// the complete line has the expected format).
+		CompleteLines: true,
```
ybrill, line 220: does this mean the last line in the file will only be read if the file is terminated with a \n?
Like in a regular case: I wrote a file and exited before the task finishes.

agent/internal/taskoutput/test_log.go
```diff
@@ -0,0 +1,330 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case line := <-t.Lines:
+			if line.Err != nil {
+				// If there is a line error it is safe to
+				// continue because the follower maintains the
+				// seek position and will pick up where it left
+				// off. These errors usually occur when rate
+				// limiting is configured and a cool down
```
ybrill, line 256: could you explain more what you mean by rate limiting?

agent/internal/taskoutput/test_log.go
julianedwards: Ah that was from before, I will remove it.

agent/internal/taskoutput/test_log.go
julianedwards: It is the only way the err will be non-nil (on L144 above).

agent/internal/taskoutput/test_log.go
julianedwards: Exactly, all log lines written in the "happy path" scenario should be terminated with \n.

agent/internal/taskoutput/test_log.go
julianedwards: You can configure the file follower to rate limit the amount of data you read: https://pkg.go.dev/github.com/nxadm/tail#Config.

ybrill: Ah, got it. But do we need to check the err, then? (since we haven't enabled ratelimiting)

agent/internal/taskoutput/test_log.go
ybrill: But are we comfortable with not potentially not getting the last line?
Could we want `close` to read the rest of the file from the position of the last read?

agent/internal/taskoutput/test_log.go
ybrill: ah I see now

agent/internal/taskoutput/test_log.go
julianedwards: Yeah, log lines should always end in a newline so we would only not get it in the case that something went wrong. If we don't get complete lines, we risk a parsing a malformed line and breaking the parser, in which case, the line would fail to get persisted anyways.

agent/internal/taskoutput/test_log.go
julianedwards: Probably not, but I just added it to be safe and for posterity (in case we end up adding rate limiting).

agent/internal/taskoutput/test_log.go
julianedwards: This will also be documented for users so they understand the behavior.

ybrill (approved):

PR #7356, opened 2024-01-02
DEVPROD-979: terminate spawn hosts that fail to spawn

Kimchelly:
DEVPROD-979

### Description
This is the same as #7313 except:
* I changed the host trigger processor to deal with the fact that intent hosts are deleted and re-inserted into the DB (with a different `_id`) when the host starts in EC2. A host event's resource ID could be either its intent host ID (i.e. the original `_id` for the host) or its EC2 instance ID (i.e. the new `_id` after the host doc is replaced), so it now handles both interchangeably.
* I changed it to log a new event type (`HOST_CREATED_ERROR`) instead of reusing the existing event type (`HOST_CREATED`). This is a small optimization to reduce the amount of events that the notification system has to process. Instead of having to process all host created events (1 per host), it only processes host creation errors (which only happen if the host fails to start up in EC2).

### Testing
* Tested in staging that a notification got sent if a spawn host could not be spawned in EC2.
* Tested in staging that success notification was sent if a spawn host was successfully started up.

### Documentation
N/A

### DIFF
--- a/model/event/host_event.go
+++ b/model/event/host_event.go
@@ -101,10 +101,12 @@ func LogHostEvent(hostId string, eventType string, eventData HostEventData) {
 	}
 }
 
+// LogHostCreated logs an event indicating that the host was created.
 func LogHostCreated(hostId string) {
 	LogHostEvent(hostId, EventHostCreated, HostEventData{Successful: true})
 }
 
+// LogManyHostsCreated is the same as LogHostCreated but for multiple hosts.
 func LogManyHostsCreated(hostIDs []string) {
 	events := make([]EventLogEntry, 0, len(hostIDs))
 	for _, hostID := range hostIDs {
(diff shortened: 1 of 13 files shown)


model/event/host_event.go
```diff
@@ -103,10 +105,31 @@ func LogHostCreated(hostId string) {
 	LogHostEvent(hostId, EventHostCreated, HostEventData{Successful: true})
 }
 
-// LogHostCreationFailed logs an event indicating that the host errored while it
+func LogManyHostsCreated(hostIDs []string) {
```
hadjri, line 108: Nit: doc exported function

hadjri (approved):

model/event/host_event.go
Kimchelly: Good catch, added.

PR #7358, opened 2024-01-03
DEVPROD-731 Add include limits to admin settings

bynn:
DEVPROD-731

### Description
Add a limit for how many include files a patch can have 

Because we can't access settings in the cli and we don't want to change over 100 instances of loadProjectInto to pass in a settings, I added a new field to project called `NumIncludes` which is just used for validation. But because of the way this works, `evergreen validate` won't error when there are too many includes. It will still go through every includes file and give the proper warning and errors. I think this behavior is okay. 

However, it errors when trying to create patches as shown below

### Testing
tested on staging and unit tests
image
image

### DIFF
--- a/model/project.go
+++ b/model/project.go
@@ -70,7 +70,7 @@ type Project struct {
 	Private bool `yaml:"private,omitempty" bson:"private"`
 
 	// Number of includes in the project cached for validation
-	NumIncludes int
+	NumIncludes int `yaml:"-" bson:"-"`
 }
 
 type ProjectInfo struct {
--- a/model/project.go
+++ b/model/project.go
@@ -68,6 +68,9 @@ type Project struct {
 
 	// Flag that indicates a project as requiring user authentication
 	Private bool `yaml:"private,omitempty" bson:"private"`
+
+	// Number of includes in the project cached for validation
+	NumIncludes int
 }
 
 type ProjectInfo struct {
--- a/model/project_parser.go
+++ b/model/project_parser.go
@@ -745,7 +745,9 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 	}
 	project.Identifier = identifier
 
-	// Remove includes once the project is translated.
+	// Remove includes once the project is translated since translate project saves number of includes.
+	// Intermediate project is used to save parser project as a YAML so removing the includes verifies that
+	// they have been processed.
 	intermediateProject.Include = nil
 
 	return intermediateProject, errors.Wrapf(err, LoadProjectError)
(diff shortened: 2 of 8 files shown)


model/project.go
```diff
@@ -68,6 +68,9 @@ type Project struct {
 
 	// Flag that indicates a project as requiring user authentication
 	Private bool `yaml:"private,omitempty" bson:"private"`
+
+	// Number of includes in the project cached for validation
+	NumIncludes int
```
Kimchelly, line 73: Typically in marshalling/unmarshalling between different data formats (YAML, BSON, etc.), the marshalling logic will try to use any exported field with a default field name as long as we haven't explicitly told it to ignore the field. From [the BSON marshalling docs](https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.13.1/bson#hdr-Structs):

>1. Only exported fields in structs will be marshalled or unmarshalled.
>2. When marshalling a struct, each field will be lowercased to generate the key for the corresponding BSON element. For example, a struct field named "Foo" will generate key "foo". This can be overridden via a struct tag (e.g. `bson:"fooField"` to generate key "fooField" instead). 

Since we don't plan on persisting this field, we should use `"-"` for the struct tags because we don't want this field to persist:
```go
NumIncludes int `yaml:"-" bson:"-"`
```

model/project_parser.go
```diff
@@ -737,14 +738,16 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
...
-
-	// return project even with errors
+	// Return project even with errors.
 	p, err := TranslateProject(intermediateProject)
 	if p != nil {
 		*project = *p
 	}
 	project.Identifier = identifier
+
+	// Don't remove includes until project is translated.
```
Kimchelly, line 748: nit: should this instead say "Remove includes once the project is translated", since it's doing the removal right here?

Can we also clarify why we need to remove them?

model/project_parser.go
```diff
@@ -737,14 +738,16 @@ func LoadProjectInto(ctx context.Context, data []byte, opts *GetProjectOpts, ide
 		}
 	}
 
-	intermediateProject.Include = nil
-
-	// return project even with errors
+	// Return project even with errors.
```
Kimchelly, line 741: Nice! 👌 (no change)

Kimchelly (changes requested):
Just small things, it overall looks good.

model/project.go
bynn: Thank you for explaining!

model/project_parser.go
bynn: It looks like the parser project that's returned is still used [to save the project yaml](https://github.com/evergreen-ci/evergreen/blob/DEVPROD-731/model/patch_lifecycle.go#L313) and I don't want to change any expected behavior 

But it makes much more sense to update the comment

Kimchelly: Thanks! Can we also include a mention of what removing the includes is intended to do? Like, why not just leave it in as-is?

Kimchelly (approved):
LGTM mod minor clarification.

PR #7349, opened 2023-12-27
DEVPROD-3233 client links to S3

ybrill:
[DEVPROD-3233](https://jira.mongodb.org/browse/DEVPROD-3233)

### Description
Our clients are uploaded to S3 and [most](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen%20path%3D%22*%2Fclients%2F*%22%20%0A%7C%20stats%20count%20by%20path&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-7d%40h&latest=now&display.page.search.tab=statistics&display.general.type=statistics&sid=1703709994.7428056) client downloads are from there already (ostensibly the outliers are CLI downloads from the UI).

This PR stops linking to the downloads from the  app servers and links only to the S3 builds. 

**NOTE**: a custom build will not correspond with anything in S3 and no binaries will be available for it. Only `dist` (and `dist-staging`) deploys will be able to provision new hosts. This is portent of deploys post the switchover to Kanopy, when only `dist`/`dist-staging`s will be deployable at all.

### Testing
On staging a new host provisioned and ran a task and I was able to download the CLI from the UI for the various architectures.

PR #7361, opened 2024-01-03
DEVPROD-3233 download clients from S3 only

ybrill:
[DEVPROD-3233](https://jira.mongodb.org/browse/DEVPROD-3233)

### Description
This goes together with https://github.com/evergreen-ci/evergreen/pull/7352 which added an S3 build per version (or at least it will when we remove `activate: false`)

This should basically be a noop unless the EVG_VERSION_ID env variable is set. If it is set then this will make s3 downloads point to
```
https://mciuploads.s3.amazonaws.com/evergreen/clients/<version_id>/<platform>/evergreen[.exe]
```
Additionally, it will make the CLI download links in the UI point to S3 instead of the app servers. Agent downloads will still fall back to the app servers for now.

Once the version ID is always set (i.e. after the migration to Kanopy) we'll remove the `S3BinaryDownloadsDisabled` flag, remove the fallbacks to local downloads, and consolidate ClientConfig to only contain S3 binaries.

### Testing
In staging the UI links and the [cli_version route](https://github.com/evergreen-ci/evergreen/blob/91e6e7d413c54234266574b6e6ed7d1ed73c1e9d/rest/route/service.go#L221) (used by the CLI to update) and [client_urls](https://github.com/evergreen-ci/evergreen/blob/91e6e7d413c54234266574b6e6ed7d1ed73c1e9d/rest/route/service.go#L146) (used by the agent monitor) pointed to the correct S3 links.


config_api.go
```diff
@@ -17,9 +20,47 @@ type ClientBinary struct {
...
+		name := strings.Split(item, "/")
+		if len(name) != 2 || !strings.Contains(name[1], "evergreen") {
+			continue
+		}
+		displayName, ok := ValidArchDisplayNames[name[0]]
+		if !ok {
+			continue
+		}
+		osArchParts := strings.Split(name[0], "_")
+		if len(osArchParts) != 2 {
```
hadjri, line 49: nit: why do we skip if we find an underscore / can we make a quick comment explaining?

hadjri (approved):
Changes conceptually LGTM - I can't think of anything in these changes / refactor that I'd consider to be a concern, but it's probably worthwhile for another @evergreen-ci/evg-app person to look too.

config_api.go
ybrill: The truth is, that check is kind of redundant given the previous check that the name is a key in the `ValidArchDisplayNames` map. I've removed it.
The idea of the check has been to make sure the name conformed to the convention that it have two parts (os and arch) separated by an underscore.

environment.go
```diff
@@ -1237,15 +1230,42 @@ func (e *envState) Close(ctx context.Context) error {
...
+	}
+
+	e.clientConfig = c
+	return nil
+}
+
+func (e *envState) populateLocalClientConfig() error {
+	c := &ClientConfig{LatestRevision: ClientVersion}
+	if e.settings.HostInit.S3BaseURL != "" {
+		c.S3URLPrefix = fmt.Sprintf("%s/%s", e.settings.HostInit.S3BaseURL, BuildRevision)
```
ablack12, line 1266: Just for my own education, could you explain this part? I assumed that this function would just be the same as the old code.

ablack12: Also will this function go away after we switch to kanopy? Or do you think App will need to maintain it? In which case it may be helpful to add some comments here about whats' going on

config_api.go
```diff
@@ -17,9 +20,44 @@ type ClientBinary struct {
...
+		prefix += "/"
+	}
+
+	iter, err := bucket.List(ctx, prefix)
+	if err != nil {
+		return errors.Wrap(err, "listing client bucket")
+	}
+	for iter.Next(ctx) {
+		item := strings.TrimPrefix(iter.Item().Name(), prefix)
+		name := strings.Split(item, "/")
```
ablack12, line 40: It seems like all of this string splitting relies on the bucket names being a certain format -- just for reference, could we put an example or the template of what we're expecting in a comment? (Seems like <prefix/<display_name_that's_of_some_specific_format>/<os>_<arch>/evergreen if i'm parsing that right)

ablack12 (approved):
LGTM with a couple of thoughts

environment.go
ybrill: The idea of the `S3URLPrefix ` is that after the initial setup they can be treated identically. So no matter whether the S3 binaries are in the `upload-clis` place or the `build-and-push` place we can get a URL by appending the path to a prefix.

This is indeed intended to be temporary. After the switchover it will look more like https://github.com/evergreen-ci/evergreen/pull/7349, where there was a single source for the S3 binaries. I'll try to add some more color with comments, though.

config_api.go
ybrill: added

PR #7362, opened 2024-01-03
DEVPROD-1943: do not use expired tokens for GitHub status sender

Kimchelly:
DEVPROD-1943

### Description
Fixes two issues with the GitHub status sender:
* Ensure GitHub app tokens are refreshed well in advance of when they actually expire. GitHub senders are cached to reduce the amount of tokens requested, but the TTL is currently set to the same duration as the token's total lifetime (1 hour). This would cause very occasional flakes when using a token that's close to expiring. For example, if Evergreen is trying to send a GitHub status back to a PR but the sender it got back had a token that was going to expire in 1 millisecond, it would use the almost-expired sender, the token would expire, and GitHub would reject the request.
* The GitHub status sender didn't set the fallback error handler like the rest of the senders do, so it wouldn't send errors to Splunk. Any errors produced from sending to the `GetGitHubSender` sender should now appear in Splunk.
* Also added some docs for clarity around global logging/sender initialization in the env.

I'm gonna be monitoring after this is deployed to ensure that 1. GitHub app errors appear in Splunk and 2. the 401 bad credential errors for the `CreateStatus` requests go away.

### Testing
* Added unit tests for setting the fallback error handler.
  * The GitHub sender is awkwardly tested in the `thirdparty` package because `testutil.ConfigureIntegrationTest` can't be called from the root `evergreen` package. Didn't think it was worth it to move `ConfigureIntegrationTest` into the root package for a single use.
* Test in staging that patches can still get data from GitHub as per usual.

### Documentation
N/A

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -1137,6 +1137,7 @@ func (e *envState) GetGitHubSender(owner, repo string) (send.Sender, error) {
 	grip.Error(message.WrapError(e.setSenderErrorHandler(sender, owner), message.Fields{
 		"message": "could not set fallback error handler for GitHub status sender",
 		"owner":   owner,
+		"repo":    repo,
 	}))
 
 	e.githubSenders[owner] = cachedGitHubSender{
--- a/environment.go
+++ b/environment.go
@@ -229,7 +229,7 @@ func NewEnvironment(ctx context.Context, confPath string, db *DBSettings) (Envir
 
 	catcher.Add(e.initJasper())
 	catcher.Add(e.initDepot(ctx))
-	catcher.Add(e.initSenders(ctx))
+	catcher.Add(e.initThirdPartySenders(ctx))
 	catcher.Add(e.createLocalQueue(ctx))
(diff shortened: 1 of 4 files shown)


environment.go
```diff
@@ -698,6 +702,10 @@ func (e *envState) initClientConfig() {
 	}
 }
 
+// initSenders initializes the senders that are used to send payloads to
+// external services such as sending GitHub statuses and Jira messages. These
+// are meant to enable specific Evergreen behavior and are distinct from the
+// global application-wide logging system (see (*Settings).GetSender).
 func (e *envState) initSenders(ctx context.Context) error {
```
bynn, line 709: Maybe unnecessary because of the nice comments you put in but should we also change the name of the function to like `initThirdpartySenders` or something to make it clearer?

bynn (approved):
Tiny suggestion but senders look prettier now!

environment.go
Kimchelly: Good idea, renamed it.

PR #7365, opened 2024-01-05
DEVPROD-3230 Agent hitting repeated panics after outage

malikchaya2:
DEVPROD-3230

This reopens https://github.com/evergreen-ci/evergreen/pull/7334 with a fix. The first four commits are from the original PR. The last commit is new. 

I would have loved for this to be a mysterious quirk or an elusive bug, but alas, it is not. It was an `||` in place of an `&&`.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1068,13 +1068,13 @@ func setEndTaskFailureDetails(tc *taskContext, detail *apimodels.TaskEndDetail, 
 
 }
 
-func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string, additionalKillOverride bool) {
+func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string, additionalShouldKillCheck bool) {
 	logger := grip.NewJournaler("killProcs")
 	if tc.logger != nil && !tc.logger.Closed() {
 		logger = tc.logger.Execution()
 	}
 
-	if !additionalKillOverride || !a.shouldKill(tc, ignoreTaskGroupCheck) {
+	if !additionalShouldKillCheck || !a.shouldKill(tc, ignoreTaskGroupCheck) {
 		return
 	}
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1074,7 +1074,7 @@ func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupC
 		logger = tc.logger.Execution()
 	}
 
-	if !additionalShouldKillCheck || !a.shouldKill(tc, ignoreTaskGroupCheck) {
+	if !a.shouldKill(tc, ignoreTaskGroupCheck) && !additionalShouldKillCheck {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


agent/agent.go
```diff
@@ -1061,13 +1068,13 @@ func setEndTaskFailureDetails(tc *taskContext, detail *apimodels.TaskEndDetail,
 
 }
 
-func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string) {
+func (a *Agent) killProcs(ctx context.Context, tc *taskContext, ignoreTaskGroupCheck bool, reason string, additionalShouldKillCheck bool) {
```
Kimchelly, line 1071: Rather than add an additional flag that we have to set to false everywhere, did we consider just checking the `taskConfig` is set or not in `shouldKill`? It seems like just checking `taskConfig` for nilness in `shouldKill` is logically the same (i.e. it'll return early) without needing to introduce a lot of plumbing of one special case value.

agent/agent.go
```diff
@@ -1061,13 +1068,13 @@ func setEndTaskFailureDetails(tc *taskContext, detail *apimodels.TaskEndDetail,
...
 	logger := grip.NewJournaler("killProcs")
 	if tc.logger != nil && !tc.logger.Closed() {
 		logger = tc.logger.Execution()
 	}
 
-	if !a.shouldKill(tc, ignoreTaskGroupCheck) {
+	if !a.shouldKill(tc, ignoreTaskGroupCheck) && !additionalShouldKillCheck {
```
Kimchelly, line 1077: 1. If the original problem was that `a.shouldKill` dereferences a nil `taskConfig` but this PR already changed the logic to check for nil `taskConfig`, is the `additionalShouldKillCheck` flag still needed? It seems specially skipping `shouldKill` wouldn't matter the nil `taskConfig` check is already in place.
2. Is the goal of this flag to skip running `a.shouldKill` specifically when `makeTaskConfig` fails? If so, the two clauses in this if statement have to be swapped (to `!additionalShouldKillCheck && !a.shouldKill(tc, ignoreTaskGroupCheck)`) because Go evaluates each `&&`-separated clause in order.

Kimchelly (changes requested):

agent/agent.go
malikchaya2: If I do that, I won't be able to use nt and the initial taskConfig:

> taskGroupIsNil := nt.TaskGroup == ""
		if initialTC != nil && initialTC.taskConfig != nil {
			taskGroupIsNil = initialTC.taskConfig.TaskGroup == nil
		}
		
Which is what was happening before we [simplified task config](https://github.com/evergreen-ci/evergreen/pull/6985/files#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8b)

agent/agent.go
malikchaya2: It doesn't need to be swapped because there's no longer a nil pointer dereference issue. a.shouldkill is not a heavy process so there's no need to save that unit of work in the case where !additionalShouldKillCheck  is false.

Kimchelly (changes requested):
I already wrote some follow-up comments but GH apparently doesn't count it as a review unless I post a new comment like this. 🤷

agent/agent.go
malikchaya2: Hmm, it’s true that if the task config is nil, shouldKill will return true anyway. So this wouldn’t help the case were we shouldn’t be killing procs (because it’s part of a task group), but we lost that information by the tc not being populated. I'll remove this.

Kimchelly (approved):

PR #7372, opened 2024-01-08
DEVPROD-2255 Fix broken doc links 

malikchaya2:
DEVPROD-2255

### Description
While these links were working in the github viewer, they weren't working on the docs site because they were resolving to `/CLI.md#operating-on-existing-patches `instead of `/CLI#operating-on-existing-patches. `

### Testing
This is hard to test completely, but in the viewer, the link opens to `https://github.com/evergreen-ci/evergreen/blob/main/docs/CLI/#operating-on-existing-patches`, so I am guessing this will solve the issue by opening `https://docs.devprod.prod.corp.mongodb.com/evergreen/CLI#operating-on-existing-patches` once it's deployed.

### DIFF
--- a/docs/Project-Configuration/Project-Configuration-Files.md
+++ b/docs/Project-Configuration/Project-Configuration-Files.md
@@ -407,10 +407,10 @@ manifest will be created that uses the latest revision available for each module
 For manual patches and GitHub PRs, by default, the git revisions in the
 version manifest will be inherited from its base version (i.e. the mainline commit version of the patch's base git revision). 
 You can change the git revision for modules by setting a module manually with 
-[evergreen set-module](../CLI.md#operating-on-existing-patches) or
+[evergreen set-module](../CLI/#operating-on-existing-patches) or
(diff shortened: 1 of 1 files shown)


Kimchelly (approved):
This seems fine, but one thing I did notice is that it seems like a lot of the links (but not all of them?) using `.md` after them are broken, whereas previously they did work (as I have used them before and confirmed they work, and using `.md` was previously the only way to link to URLs in a different sub-directory using `..`). We should probably fix all these broken URLs, so I filed DEVPROD-3598.

malikchaya2:
> This seems fine, but one thing I did notice is that it seems like a lot of the links (but not all of them?) using `.md` after them are broken, whereas previously they did work (as I have used them before and confirmed they work, and using `.md` was previously the only way to link to URLs in a different sub-directory using `..`). We should probably fix all these broken URLs, so I filed [DEVPROD-3598](https://jira.mongodb.org/browse/DEVPROD-3598).

thanks! I think it's only the links with `../` as part of it.

PR #7374, opened 2024-01-09
DEVPROD-3545: Check for nil patches to prevent panics in FinalizePatch

minnakt:
DEVPROD-3545

### Description
There was an ops alert on Monday about accessing fields on a nil object in the `FinalizePatch` function. The trace led to [evergreen/model/patch_lifecycle.go:483](https://github.com/evergreen-ci/evergreen/blob/24172ec36bde475d89a2c01453e3f6fac85704ae/model/patch_lifecycle.go#L483) and Kim suggested that the patch may be nil because we don't check for nil in some places before `FinalizePatch`.

I think this makes sense and I am suspicious that the problem was not checking if the childPatchDoc was nil. Around the same time this alert occurred, a user reported that they couldn't run any Evergreen tasks on their PR. This was because child patches were misconfigured in their Project Settings and could not be created properly. 

I went through the Evergreen repo looking for places where we didn’t check for a nil patch and added if statements for them. Let me know if I missed any!

### Testing
- Added a small unit test for SetPatchActivated which panics without the added code
- I didn't update the patch lifecycle test because it seemed kind of involved for a small nil check update


service/patch.go
```diff
@@ -125,6 +125,11 @@ func (uis *UIServer) diffPage(w http.ResponseWriter, r *http.Request) {
 			http.StatusInternalServerError)
 		return
 	}
+	if fullPatch == nil {
```
hadjri, line 128: I think we need to add a check on L43 of this file as well

hadjri (changes requested):
Only other place I could find that could use a nil check for a patch is [here](https://gist.github.com/hadjri/f6ffc922ac9dc192aa813bb23b632284#file-github-go-L534)

service/patch.go
minnakt: Thank you!

hadjri (approved):

PR #7375, opened 2024-01-09
DEVPROD-3613 Add HasCheckRun field to task

malikchaya2:
DEVPROD-3613

### Description
Add a HasCheckRun field to tasks that indicate if the task should create a checkrun

### DIFF
--- a/model/lifecycle.go
+++ b/model/lifecycle.go
@@ -537,6 +537,9 @@ func addTasksToBuild(ctx context.Context, creationInfo TaskCreationInfo) (*build
 			continue
 		}
 		bvtu := creationInfo.Project.FindTaskForVariant(t.DisplayName, creationInfo.Build.BuildVariant)
+		if bvtu.HasCheckRun() {
+			t.HasCheckRun = true
+		}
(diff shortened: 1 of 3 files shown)


model/lifecycle.go
```diff
@@ -537,6 +537,9 @@ func addTasksToBuild(ctx context.Context, creationInfo TaskCreationInfo) (*build
 			continue
 		}
 		bvtu := creationInfo.Project.FindTaskForVariant(t.DisplayName, creationInfo.Build.BuildVariant)
+		if bvtu.HasCheckRun() {
```
bynn, line 540: I think most of the other fields in task is copied from the bvtu here https://github.com/evergreen-ci/evergreen/blob/92a6e0255c64ef8a879aad4c1a7f2661e67cf7bf/model/lifecycle.go#L1162

is there a reason that we should add it here?

bynn (changes requested):

model/lifecycle.go
malikchaya2: nope! good catch, thanks! moved

bynn (approved):

PR #7377, opened 2024-01-09
DEVPROD-3615 fix test-thirdparty on Ubuntu 22.04

malikchaya2:
DEVPROD-3615

### Description
This fix uses a user that is in the organization. An argument can also be made to remove this test now. More information in the ticket. 

### Testing
The test itself.

### DIFF
--- a/thirdparty/github_test.go
+++ b/thirdparty/github_test.go
@@ -269,7 +269,7 @@ func (s *githubSuite) TestGetPullRequestMergeBase() {
 }
 
 func (s *githubSuite) TestGithubUserInOrganization() {
-	isMember, err := GithubUserInOrganization(s.ctx, s.token, "evergreen-ci", "evrg-bot-webhook")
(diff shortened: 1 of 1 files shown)


bynn (changes requested):
I'm hesitant about this because this might be the whole deafgoat situation again a couple years later. One work around may be to call the get org users API from github and grab a random user. Then, we know for sure that the test should never fail. If this seem like too much, we could also just delete this test. 

https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#list-organization-members

malikchaya2:
> I'm hesitant about this because this might be the whole deafgoat situation again a couple years later. One work around may be to call the get org users API from github and grab a random user. Then, we know for sure that the test should never fail. If this seem like too much, we could also just delete this test.
> 
> https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#list-organization-members

I agree with deleting it. I removed the assert true part.

bynn (approved):

PR #7384, opened 2024-01-11
DEVPROD-761 Specify backport commit-sha option only allows from same github project

ZackarySantana:
DEVPROD-761

### Description
A user was attempting to specify commit-sha for a different repository than the project they were using it on, which caused the sha not to be found

### DIFF
--- a/operations/commit_queue.go
+++ b/operations/commit_queue.go
@@ -362,11 +362,11 @@ func backport() cli.Command {
 				},
 				cli.StringFlag{
 					Name:  joinFlagNames(existingPatchFlag, "e"),
-					Usage: "existing commit queue patch",
+					Usage: "existing commit queue patch from another repository",
 				},
 				cli.StringFlag{
 					Name:  joinFlagNames(commitShaFlag, "s"),
-					Usage: "existing commit SHA to backport",
+					Usage: "existing commit SHA from the same repository to backport",
 				},
 				cli.StringFlag{
 					Name:  joinFlagNames(backportProjectFlag, "b"),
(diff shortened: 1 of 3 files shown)


ablack12 (changes requested):
Can we also add something to the output for `--commit-sha` from the help text for the backport command? 
https://github.com/evergreen-ci/evergreen/blob/f29c49c4c4336df43df5dd80e8774c5ce0cc1d8e/operations/commit_queue.go#L369

ZackarySantana:
> Can we also add something to the output for `--commit-sha` from the help text for the backport command?
> 
> https://github.com/evergreen-ci/evergreen/blob/f29c49c4c4336df43df5dd80e8774c5ce0cc1d8e/operations/commit_queue.go#L369

I added it to commit sha and the patch alias flags to the command. I was thinking of adding that they are mutually exclusive as well like we state in the docs, but I wasn't sure how to word it and wasn't sure if someone would try it + if they did it would error.

operations/commit_queue.go
```diff
@@ -362,11 +362,11 @@ func backport() cli.Command {
 				},
 				cli.StringFlag{
 					Name:  joinFlagNames(existingPatchFlag, "e"),
-					Usage: "existing commit queue patch",
+					Usage: "existing commit queue patch from another repository",
```
ablack12, line 365: I think this isn't likely to be a different repository but a different _project_ (or a different branch) bc backports happen to different branches on the same repo

ablack12 (changes requested):

operations/commit_queue.go
ZackarySantana: Oh! That was my misunderstanding, let me rephrase that

ablack12 (approved):

PR #7388, opened 2024-01-11
DEVPROD-932 Consistent naming for config in project trigger docs

ZackarySantana:
DEVPROD-932

### Description
This is just to make the list be in the same order as Spruce project settings page. As well as renaming 'definition file' to 'config file' to be more consistent with what we call it in other places (I haven't seen definition file anywhere else...)

EDIT: here is [the spruce PR ](https://github.com/evergreen-ci/spruce/pull/2222)that affects what the ticket is targeted towards


ablack12 (changes requested):
DM-ed; i think the ticket here is wrong (we should also add a title)

ablack12 (approved):

PR #7390, opened 2024-01-11
DEVPROD-772 shore up project trigger and patch trigger alias docs

ablack12:
DEVPROD-772 

### Description
Added some pictures that we frequently send users as examples

PR #7391, opened 2024-01-11
DEVPROD-3159: document project modify inputs

Kimchelly:
DEVPROD-3159

### Description
* Include sample input body schema for modifying a project.
* Write more detailed docs on the input fields to modify a project. I skipped a couple fields that seemed unimportant and I couldn't tell what they were supposed to do without going on a long slog through the code.

### Testing
Checked that the docs outputted by `make swaggo` produced reasonable-looking docs for the project modify route.

### DIFF
--- a/rest/model/project.go
+++ b/rest/model/project.go
@@ -419,11 +419,17 @@ type APIWorkstationConfig struct {
 }
 
 type APIContainerSecret struct {
-	Name         *string `json:"name"`
+	// Name of the container secret.
+	Name *string `json:"name"`
+	// External name of the container secrets. Cannot be modified by users.
 	ExternalName *string `json:"external_name"`
-	ExternalID   *string `json:"external_id"`
-	Type         *string `json:"type"`
-	Value        *string `json:"value"`
+	// External identifier for the container secret. Cannot be modified by
+	// users.
+	ExternalID *string `json:"external_id"`
+	// Type of container secret.
+	Type *string `json:"type"`
+	// Container secret value to set.
+	Value *string `json:"value"`
 	// ShouldRotate indicates that the user requested the pod secret to be
 	// rotated to a new value. This only applies to the project's pod secret.
 	ShouldRotate *bool `json:"should_rotate"`
--- a/rest/model/project.go
+++ b/rest/model/project.go
@@ -100,9 +100,13 @@ type APIPatchTriggerDefinition struct {
 	// List of task specifiers.
 	TaskSpecifiers []APITaskSpecifier `json:"task_specifiers"`
 	// Status for the parent patch to conditionally kick off the child patch.
-	Status         *string       `json:"status,omitempty"`
-	ParentAsModule *string       `json:"parent_as_module,omitempty"`
-	VariantsTasks  []VariantTask `json:"variants_tasks,omitempty"`
+	Status *string `json:"status,omitempty"`
+	// Name of the module corresponding to the upstream project in the
+	// downstream project's YAML.
+	ParentAsModule *string `json:"parent_as_module,omitempty"`
+	// The list of variants/tasks from the alias that will run in the downstream
+	// project.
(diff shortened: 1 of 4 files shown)


rest/model/project.go
```diff
@@ -88,13 +91,18 @@ func (t *APITriggerDefinition) BuildFromService(triggerDef model.TriggerDefiniti
...
+	Alias *string `json:"alias"`
+	// ID of the downstream project.
+	ChildProjectId *string `json:"child_project_id"`
+	// Identifier of the downstream project.
+	ChildProjectIdentifier *string `json:"child_project_identifier"`
+	// List of task specifiers.
+	TaskSpecifiers []APITaskSpecifier `json:"task_specifiers"`
+	// Status for the parent patch to conditionally kick off the child patch.
+	Status         *string       `json:"status,omitempty"`
+	ParentAsModule *string       `json:"parent_as_module,omitempty"`
```
ablack12, line 104: I just wrote some better docs for this in https://github.com/evergreen-ci/evergreen/pull/7390 -- this means the module name corresponding to the upstream project in the downstream project's yaml (this needs to be included if you want the child patch to test with the upstream project's changes)

rest/model/project.go
```diff
@@ -88,13 +91,18 @@ func (t *APITriggerDefinition) BuildFromService(triggerDef model.TriggerDefiniti
...
+	VariantsTasks  []VariantTask `json:"variants_tasks,omitempty"`
```
ablack12, line 105: Denotes the subset of tasks/variants that will run in the downstream project as part of the alias.

rest/model/project.go
```diff
@@ -535,87 +580,123 @@ func (t *APIRepositoryErrorDetails) BuildFromService(h model.RepositoryErrorDeta
...
-	UseRepoSettings        *bool   `json:"use_repo_settings"`
-	RepoRefId              *string `json:"repo_ref_id"`
-	// Options for commit queue
-	CommitQueue            APICommitQueueParams      `json:"commit_queue"`
-	TaskSync               APITaskSyncOptions        `json:"task_sync"`
+	// Enable GitHub automated pull request testing.
+	PRTestingEnabled *bool `json:"pr_testing_enabled"`
+	// Enable GitHub manual pull request testing.
+	ManualPRTestingEnabled *bool `json:"manual_pr_testing_enabled"`
+	// Enable testing when git tag are pushed.
```
ablack12, line 618: git tags*

rest/model/project.go
```diff
@@ -535,87 +580,123 @@ func (t *APIRepositoryErrorDetails) BuildFromService(h model.RepositoryErrorDeta
...
+	// Disable stats caching.
+	DisabledStatsCache *bool `json:"disabled_stats_cache"`
 	// Usernames of project admins. Can be null for some projects (EVG-6598).
 	Admins []*string `json:"admins"`
-	// Usernames of project admins to remove
-	DeleteAdmins                []*string `json:"delete_admins,omitempty"`
-	GitTagAuthorizedUsers       []*string `json:"git_tag_authorized_users" bson:"git_tag_authorized_users"`
+	// Usernames of project admins to remove.
+	DeleteAdmins []*string `json:"delete_admins,omitempty"`
+	// Usernames authorized to submit git tag versions.
```
ablack12, line 659: Should we move the git tag disabled flag closer to this since they're so related?

ablack12 (approved):

rest/model/project.go
Kimchelly: I dunno, the organization seems alright to me since it groups by the enabled/disabled flags rather than related features. Anyways, the fields are always listed in alphabetical order in the autogenerated docs so I don't think it matters to users.

rest/model/project.go
Kimchelly: Thanks, updated!

PR #7393, opened 2024-01-11
DEVPROD-3716: Change API file extension to .mdx

sophstad:
DEVPROD-3716

### Description
- This file uses MDX to render the OpenAPI spec. Give it a `.mdx` extension so that Docusaurus v3 can use the appropriate markdown parser.

### Testing
- I don't really know how we could test this on the current deployment of Pine since it builds content based on the GitHub main branch... but I think it will work...


bsamek (approved):
LGTM. It's easy to revert if it doesn't work.

PR #7400, opened 2024-01-11
DEVPROD-3140: document notifications in project settings

Kimchelly:
DEVPROD-3140

### Description
* Clarify that the notifications section in the docs is for user notifications.
* Add docs and links to project-level notifications.

### Testing
Checked the rendered Markdown looked reasonable.


ablack12 (approved):
LGTM, can we comment the new docs on the ticket?

PR #7405, opened 2024-01-12
DEVPROD-3739: Downgrade incorrect git remote to CLI warning

hadjri:
DEVPROD-3739

### Description
In order to avoid potential complexity with v7.1 and v7.2 of server (where engineers may run into this error on submitting patches), temporarily downgrading this to a warning. This also requires we revert the [loadGitData](https://github.com/evergreen-ci/evergreen/pull/7311/files#diff-8b8cc6bdad49868ce616124c6c2b682e973c6db9e89111473a319bbefda656a2R624) function change from #7311 with the understanding that it'll need to be added back along with the changes from DEVPROD-3740.

### Testing
Tested locally and confirmed warning outputs.

In sandbox, I set my local remote and origin to point to my fork, `hadjri/commit-queue-sandbox` rather than the original repo. Confirmed that without this change, the patch fails to be created, and after this change, patch creation succeeds.

(i.e., my locally built CLI makes the patch correctly, and my default `evergreen` binary does not)
Screenshot 2024-01-12 at 5 49 01 PM

### DIFF
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -621,10 +621,11 @@ Continue?`, uncommittedChangesFlag), true), nil
 // If no dir is provided, we use the current working directory.
 // The branch argument is used to determine where to generate the merge base from, and any extra
 // arguments supplied are passed directly in as additional args to git diff.
+// TODO: DEVPROD-3740 Re-implement the changes for this function from https://github.com/evergreen-ci/evergreen/pull/7311
 func loadGitData(dir, remote, branch, ref, commits string, format bool, extraArgs ...string) (*localDiff, error) {
 	// branch@{upstream} refers to the branch that the branch specified by branchname is set to
-	// build on top of. This allows automatically detecting a branch based on the correct remote,
-	// if the user's repo is a fork, for example. This also works with a commit hash, if given.
+	// build on top of. For example, if a user's repo is a fork, this allows automatic detection
+	// of a branch based on the correct remote. This also works with a commit hash, if given.
 	// In the case a range is passed, we only need one commit to determine the base, so we use the first commit.
 	// For details see: https://git-scm.com/docs/gitrevisions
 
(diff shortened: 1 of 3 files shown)


hadjri:
Not ready for review yet

operations/patch_util.go
```diff
@@ -622,21 +622,16 @@ Continue?`, uncommittedChangesFlag), true), nil
 // The branch argument is used to determine where to generate the merge base from, and any extra
 // arguments supplied are passed directly in as additional args to git diff.
 func loadGitData(dir, remote, branch, ref, commits string, format bool, extraArgs ...string) (*localDiff, error) {
-	if remote == "" {
-		remote = "upstream"
-	}
-	// remote/branch refers directly to the remote branch and does not require a local branch.
-	// branch@{remote} refers to the remote branch that "branch" is tracking.
+	// branch@{upstream} refers to the branch that the branch specified by branchname is set to
+	// build on top of. This allows automatically detecting a branch based on the correct remote,
```
bynn, line 626: nit:  This seems confusing. Could we say something like `for example, if a user's repo is a fork, this allows automatic detection....`

bynn (commented):
I was a bit confused by the comment

bynn (approved):

operations/patch_util.go
```diff
@@ -621,22 +621,18 @@ Continue?`, uncommittedChangesFlag), true), nil
...
 	// In the case a range is passed, we only need one commit to determine the base, so we use the first commit.
 	// For details see: https://git-scm.com/docs/gitrevisions
 
-	mergeBase, err := gitMergeBase(dir, fmt.Sprintf("%s/%s", remote, branch), ref, commits)
+	mergeBase, err := gitMergeBase(dir, branch+"@{upstream}", ref, commits)
 	if err != nil {
-		mergeBase, err = gitMergeBase(dir, branch+"@{upstream}", ref, commits)
-		if err != nil {
-			return nil, errors.Wrapf(err, "Error getting merge base, "+
-				"may need to create local branch '%s' and have it track your Evergreen project", branch)
```
malikchaya2, line 638: Why don't we move this logic into gitMergeBase?

PR #7406, opened 2024-01-12
DEVPROD-1158 Include project name for patch details on CLI

bynn:
DEVPROD-NNNNN

### Description
Creating patch through the CLI now returns project 

### Testing
image

### DIFF
--- a/operations/patch_util.go
+++ b/operations/patch_util.go
@@ -30,8 +30,9 @@ const largePatchThreshold = 1024 * 1024 * 16
 // This is the template used to render a patch's summary in a human-readable output format.
 var patchDisplayTemplate = template.Must(template.New("patch").Parse(`
          ID : {{.Patch.Id.Hex}}
+    Project : {{.Patch.Project}}
     Created : {{.Patch.CreateTime}}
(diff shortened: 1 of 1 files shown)


operations/patch_util.go
```diff
@@ -30,8 +30,9 @@ const largePatchThreshold = 1024 * 1024 * 16
 // This is the template used to render a patch's summary in a human-readable output format.
 var patchDisplayTemplate = template.Must(template.New("patch").Parse(`
          ID : {{.Patch.Id.Hex}}
+    Project : {{.Patch.Project}}
```
Kimchelly, line 33: If I remember correctly, `(Patch).Project` stores the project ID (i.e. the `_id`), not its human-readable identifier. They can be used interchangeably for the most part, but I imagine we want to display the human-readable identifier here.

Kimchelly (changes requested):

operations/patch_util.go
bynn: image
 
Updated to show identifier but I'm not sure if this might slow down the patch creation process because it makes an extra call to evergreen to get the project ref

Kimchelly (approved):
Making one (fast and quite small) request shouldn't slow down the CLI in a meaningful way. LGTM!

PR #7414, opened 2024-01-16
DEVPROD-781 Error on unauthorized git tag user

hadjri:
DEVPROD-781

### Description
It was suggested in the comments on this JIRA ticket that we could introduce a new notification type and send it to the user when they push a git tag but are not part of the authorized git tag user fields in the project settings page, however, I think the potential work involved as part of that may be more work than it is worth for this issue. e.g:
- Creating a new notification / subscription type
- Allowing users to opt into it in Spruce
- Should we create notifications for other things that could also prevent a git tag, like having [git tag versions disabled](https://gist.github.com/hadjri/4611524c8f15d0ea64280b7c3cc0fb99#file-github-go-L752) or [incorrectly configured alias regexes](https://gist.github.com/hadjri/4611524c8f15d0ea64280b7c3cc0fb99#file-github-go-L768-L774)?)

Instead, as part of this PR:
- Encountering an unauthorized git tag user has been upgraded from a debug log to an error returned to the webhook + a relevant Splunk error to allow us to more clearly see in the future that an unauthorized git tag user was the reason a version was not created
- The git tag versions docs have been generally improved with images, more detail, and emphasis that you must be authorized to create git tags in your project settings or else versions will _not_ be created.

### Documentation
Described above


rest/route/github.go
```diff
@@ -754,7 +754,7 @@ func (gh *githubHookApi) createVersionForTag(ctx context.Context, pRef model.Pro
 	}
 
 	if !pRef.AuthorizedForGitTag(ctx, tag.Pusher, token, pRef.Owner, pRef.Repo) {
-		grip.Debug(message.Fields{
+		grip.Error(message.Fields{
```
ablack12, line 757: I'm actually against making this an error (or returning an error), because it _doesn't_ point to an error in Evergreen behavior, which we should keep these limited to. People often use tags for things that aren't Evergreen.

rest/route/github.go
```diff
@@ -763,7 +763,7 @@ func (gh *githubHookApi) createVersionForTag(ctx context.Context, pRef model.Pro
 			"tag":                tag,
 			"message":            "user not authorized for git tag version",
 		})
-		return nil, nil
+		return nil, errors.Errorf("user '%s' not authorized to create git tag versions for project '%s'", tag.Pusher, pRef.Id)
```
ablack12, line 766: So re your PR comment, I don't think we need to do a new subscription and etc -- I _think_ we could just send an email or something directly here by, assuming senders are set up (like here https://github.com/evergreen-ci/evergreen/blob/276cd4ec09474324c8fbd6ad7fcd9bf20c963d8a/rest/route/notification.go#L156-L162). 

I don't think we'd want it to be something users need to opt in to, because that doesn't really solve the problem.

I'm also fine with just going with the docs improvements. (In which case maybe we could also make a tiny spruce PR to make this text a little more explicit)

image

ablack12 (changes requested):

rest/route/github.go
hadjri: Gotcha. I knew this would error but thought that returning an error from the webhook response wouldn't be that disruptive or even visible to the user, but it does make sense that this isn't the kind of thing we should error out on.

rest/route/github.go
hadjri: Ah, my assumption was for sending notifications we always go the event --> check subscription --> send notification pipeline detailed [here](https://docs.google.com/document/d/1KxK2-JhKiHg7ydoEJN6rAf63k94KE_5-DqlYBj48pig/edit#heading=h.sffkah3u5q9), since I didn't see any examples in our codebase of us just sending notifications out arbitrarily.

rest/route/github.go
ablack12: This grip statement still needs to be changed (we don't want this to count towards our error count since it's not actually an error)

hadjri: Changed back

ablack12 (approved):

PR #7415, opened 2024-01-16
DEVPROD-3779: Use timer instead of sleep for the test log directory watcher loop

julianedwards:
DEVPROD-3779

### Description
Use a `time.Timer` instead of `time.Sleep` to throttle the test log directory watcher loop. This should improve goroutine clean up and avoid leaks.

Also added the `taskoutput` package tests to CI since they never got added. 

### Testing
Put the `TestTestLogDirectoryHandler` test in a loop and ran it across a bunch of platforms to ensure that it is not flakey.([patch build](https://spruce.mongodb.com/version/65a6b794d6d80a7ced34fbc0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

### Documentation
N/A

### DIFF
--- a/agent/internal/taskoutput/test_log.go
+++ b/agent/internal/taskoutput/test_log.go
@@ -132,39 +132,47 @@ func (h *testLogDirectoryHandler) watch(ctx context.Context) {
 		h.logger.Execution().Critical(recovery.HandlePanicWithError(recover(), nil, "test log directory watcher"))
 	}()
 
+	timer := time.NewTimer(0)
+	defer timer.Stop()
 	ignore := filepath.Join(h.dir, testLogSpecFilename)
 	seenFiles := map[string]bool{}
 	for {
-		err := filepath.WalkDir(h.dir, func(path string, info fs.DirEntry, err error) error {
+		select {
+		case <-timer.C:
+			err := filepath.WalkDir(h.dir, func(path string, info fs.DirEntry, err error) error {
+				if err != nil {
+					h.logger.Execution().Warning(errors.Wrap(err, "watching the test log directory, continuing watching"))
+					return nil
+				}
+				if ctx.Err() != nil {
+					return ctx.Err()
+				}
+				if info.IsDir() {
+					return nil
+				}
+				if path == ignore {
+					return nil
+				}
+
+				if !seenFiles[path] {
+					seenFiles[path] = true
+					h.followFileCount++
+					h.wg.Add(1)
+					go h.followFile(ctx, path)
+				}
+
+				return nil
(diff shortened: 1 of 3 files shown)


agent/internal/taskoutput/test_log.go
```diff
@@ -132,39 +132,47 @@ func (h *testLogDirectoryHandler) watch(ctx context.Context) {
...
 
-			if !seenFiles[path] {
-				seenFiles[path] = true
-				h.followFileCount++
-				h.wg.Add(1)
-				go h.followFile(ctx, path)
+				return nil
+			})
+			if err != nil {
+				h.logger.Execution().Debug("context canceled, exiting test log directory watcher")
```
ybrill, line 167: I think I asked something related last time, but it seems like this could also happen if we hit L144.
(`h.logger.Execution().Warning(errors.Wrap(err, "watching the test log directory, continuing watching"))`)

Do we need to check the ctx in the WalkDir function now that we're checking it on L172?

agent/internal/taskoutput/test_log.go
```diff
@@ -132,39 +132,47 @@ func (h *testLogDirectoryHandler) watch(ctx context.Context) {
 		h.logger.Execution().Critical(recovery.HandlePanicWithError(recover(), nil, "test log directory watcher"))
 	}()
 
+	timer := time.NewTimer(0)
```
ybrill, line 135: could a ticker work?

ybrill:
How does the switch from sleep to timer prevent goroutine leakage?

agent/internal/taskoutput/test_log.go
julianedwards: I think so, what would be the advantage?

julianedwards:
> How does the switch from sleep to timer prevent goroutine leakage?

It wasn't an actual leak, it was a race condition—the goroutine leak check can get triggered while the thread is sleeping (before it wakes up and realizes the context is canceled and exits), so the check thinks that it is leaked.

agent/internal/taskoutput/test_log.go
ybrill: ah, got it

agent/internal/taskoutput/test_log.go
ybrill: I mean, we wouldn't need to `timer.Reset(time.Millisecond)` later.
Actually, on second thought maybe we're better off with the timer. That way if walking takes longer than the timer's period we'll still get a break between walks.

ybrill (approved):

PR #7332, opened 2023-12-18
DEVPROD-2470 skip stdout logging

ybrill:
[DEVPROD-2470](https://jira.mongodb.org/browse/DEVPROD-2470)

### Description
In Kanopy we won't want to log to stdout since that'll just land up in Splunk and we already have it in Splunk 😄 .
There's maybe a marginal benefit to having the logs go to Splunk from stdout since there the Kanopy control plane is handling sending the logs so maybe even if our logs don't make it to Splunk those will, but the volume of logs would probably make it prohibitive. Fallback logs (when sending to Splunk/Slack) will still go to stderr and end up in Kanopy splunk.

### Testing
In staging application logs went to [our Splunk index](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Devergreen-staging%0A%7C%20stats%20count%20by%20host&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1702937190&latest=1702938090&display.page.search.tab=statistics&display.general.type=statistics&sid=1702938256.4561002) and [not Kanopy's](https://mongodb.splunkcloud.com/en-US/app/search/search?q=search%20index%3Ddevprod-evergreen-staging%20%0A%7C%20stats%20count%20by%20host&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-15m&latest=now&display.page.search.tab=statistics&display.general.type=statistics&sid=1702938260.4561012) (after the logger was configured).

PR #7416, opened 2024-01-16
DEVPROD-2470 log to stdout again

ybrill:
[DEVPROD-2470](https://jira.mongodb.org/browse/DEVPROD-2470)

### Description
https://github.com/evergreen-ci/evergreen/pull/7332 was supposed to make that we don't log to stdout when `disableLocalLoggingEnvVar` isn't set or if it's set to false. There was a mistake in the logic: as is it won't log if it's unset, which isn't what we want.

### Testing
`make local-evergreen` spits out logs to stdout.


julianedwards (approved):
Good catch!

PR #7417, opened 2024-01-16
DEVPROD-3552 fix communication between devprod-evergreen and devprod-platform

ybrill:
[DEVPROD-3552](https://jira.mongodb.org/browse/DEVPROD-3552)

### Description
Communication between namespaces within Kanopy needs to use an internal DNS name. Additionally, internal communication doesn't use TLS unless the application specifically enables it (and we do not). When we're running in Kanopy (signified by the versionID being set in the environment) we need to change how we make the connection to the OTel collector running in devprod-platform.

Screenshot 2024-01-16 at 3 46 30 PM

### Testing
Got some [traces](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/xQuPr91VhZT?tab=traces) into Honeycomb 🐝


config_test.go
```diff
@@ -962,8 +962,9 @@ func (s *AdminSuite) TestTracerConfig() {
 	defer cancel()
 
 	config := TracerConfig{
-		Enabled:           true,
-		CollectorEndpoint: "localhost:4316",
+		Enabled:                   true,
+		CollectorEndpoint:         "localhost:4316",
+		CollectorInternalEndpoint: "svc.cluster.local:4317",
```
opsnlops, line 967: 🎉

config_tracer.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // TracerConfig configures the OpenTelemetry tracer provider. If not enabled traces will not be sent.
 type TracerConfig struct {
-	Enabled           bool   `yaml:"enabled" bson:"enabled" json:"enabled"`
-	CollectorEndpoint string `yaml:"collector_endpoint" bson:"collector_endpoint" json:"collector_endpoint"`
+	Enabled                   bool   `yaml:"enabled" bson:"enabled" json:"enabled"`
+	CollectorEndpoint         string `yaml:"collector_endpoint" bson:"collector_endpoint" json:"collector_endpoint"`
+	CollectorInternalEndpoint string `yaml:"collector_internal_endpoint" bson:"collector_internal_endpoint" json:"collector_internal_endpoint"`
```
julianedwards, line 16: I am assuming that this is temporary until we cut over to Kanopy—can we add a TODO here with the JIRA ticket number to remove this field and the conditional logic in the init function once we move to Kanopy?

julianedwards (changes requested):

config_tracer.go
ybrill: The `CollectorEndpoint` field will still be necessary after the cutover: the agents running in EC2 will still use it.
I'm undecided on the fate of the conditional logic: I'm inclination to leave it so evergreen will still be able to run outside Kanopy if we want it to.

PR #7418, opened 2024-01-17
DEVPROD-810: Avoid using app-level context for GraphQL queries

sophstad:
DEVPROD-810

### Description
Removed invocations of app-level context throughout the `model/` directory as indicated in the ticket.

Here were a few that proved too tricky to remove—let me know if they are crucial enough to revisit as part of this fix. I don't know much about how context is created in the application so apologies in advance.
- [`user.GetOrCreateUser`](https://github.com/evergreen-ci/evergreen/blob/812185e19e4e118b685086162c92f9e86ca8d6c6/model/user/db.go#L306) - function signature is defined as in [Gimlet](https://github.com/evergreen-ci/gimlet/blob/e7de42b0623c008a8eeca167a7190edcc364ac51/auth_interfaces.go#L72) so I couldn't add `ctx`
- [`model.GetFailedTests`](https://github.com/evergreen-ci/evergreen/blob/812185e19e4e118b685086162c92f9e86ca8d6c6/model/task_history.go#L236) - only used in `services/` directory for legacy UI
- [`task.PopulateTestResults`](https://github.com/evergreen-ci/evergreen/blob/812185e19e4e118b685086162c92f9e86ca8d6c6/model/task/task.go#L2710) - a huge cascade of changes were required to add context. I couldn't find exactly how GraphQL would call this function either.
- [`task.getExpectedDurationsForWindow`](https://github.com/evergreen-ci/evergreen/blob/812185e19e4e118b685086162c92f9e86ca8d6c6/model/task/expected_duration.go#L79) - `ExpectedDuration` seems to be used [only for legacy](https://github.com/evergreen-ci/evergreen/blob/812185e19e4e118b685086162c92f9e86ca8d6c6/model/task/task.go#L231-L233)

**Update:** Created DEVPROD-3925 to handle above issues

### Testing
- Updated lots of tests

### Documentation
N/A

### DIFF
--- a/graphql/query_resolver.go
+++ b/graphql/query_resolver.go
@@ -198,7 +198,7 @@ func (r *queryResolver) DistroTaskQueue(ctx context.Context, distroID string) ([
 
 // Host is the resolver for the host field.
 func (r *queryResolver) Host(ctx context.Context, hostID string) (*restModel.APIHost, error) {
-	host, err := host.GetHostByIdOrTagWithTask(hostID)
+	host, err := host.GetHostByIdOrTagWithTask(ctx, hostID)
 	if err != nil {
 		return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error Fetching host: %s", err.Error()))
 	}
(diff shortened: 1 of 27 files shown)


malikchaya2 (approved):
Nice work! 🎉 

I think the ones you listed in the description are still worth doing. However, since they are the more tricky ones, and this PR is already pretty big, I'd suggest creating a separate ticket for it. 

(add it would probably make sense to assign it to the app team for those ones)

PR #7421, opened 2024-01-17
DEVPROD-697: Cache whether a task has annotations

hadjri:
DEVPROD-697

### Description

### Testing

### Documentation

### DIFF
--- a/model/annotations/task_annotations.go
+++ b/model/annotations/task_annotations.go
@@ -7,7 +7,6 @@ import (
 	"github.com/evergreen-ci/birch"
 	"github.com/evergreen-ci/evergreen/db"
 	"github.com/evergreen-ci/evergreen/util"
-	"github.com/mongodb/anser/bsonutil"
 	"github.com/mongodb/grip"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
@@ -83,36 +82,6 @@ func GetLatestExecutions(annotations []TaskAnnotation) []TaskAnnotation {
 	return res
 }
 
-func MoveIssueToSuspectedIssue(taskId string, taskExecution int, issue IssueLink, username string) error {
-	newIssue := issue
-	newIssue.Source = &Source{Requester: UIRequester, Author: username, Time: time.Now()}
-	q := ByTaskIdAndExecution(taskId, taskExecution)
-	q[bsonutil.GetDottedKeyName(IssuesKey, IssueLinkIssueKey)] = issue.IssueKey
-	return db.Update(
-		TaskAnnotationsCollection,
-		q,
-		bson.M{
-			"$pull": bson.M{IssuesKey: bson.M{IssueLinkIssueKey: issue.IssueKey}},
-			"$push": bson.M{SuspectedIssuesKey: newIssue},
-		},
-	)
-}
-
-func MoveSuspectedIssueToIssue(taskId string, taskExecution int, issue IssueLink, username string) error {
-	newIssue := issue
-	newIssue.Source = &Source{Requester: UIRequester, Author: username, Time: time.Now()}
-	q := ByTaskIdAndExecution(taskId, taskExecution)
-	q[bsonutil.GetDottedKeyName(SuspectedIssuesKey, IssueLinkIssueKey)] = issue.IssueKey
-	return db.Update(
-		TaskAnnotationsCollection,
-		q,
-		bson.M{
-			"$pull": bson.M{SuspectedIssuesKey: bson.M{IssueLinkIssueKey: issue.IssueKey}},
-			"$push": bson.M{IssuesKey: newIssue},
-		},
-	)
-}
-
 func UpdateAnnotationNote(taskId string, execution int, originalMessage, newMessage, username string) error {
 	note := Note{
 		Message: newMessage,
@@ -158,23 +127,6 @@ func SetAnnotationMetadataLinks(ctx context.Context, taskId string, execution in
 	return errors.Wrapf(err, "setting task links for task '%s'", taskId)
 }
 
-func AddIssueToAnnotation(taskId string, execution int, issue IssueLink, username string) error {
-	issue.Source = &Source{
-		Author:    username,
-		Time:      time.Now(),
-		Requester: UIRequester,
-	}
-
-	_, err := db.Upsert(
-		TaskAnnotationsCollection,
-		ByTaskIdAndExecution(taskId, execution),
-		bson.M{
-			"$push": bson.M{IssuesKey: issue},
-		},
-	)
-	return errors.Wrapf(err, "adding task annotation issue for task '%s'", taskId)
-}
-
 func AddSuspectedIssueToAnnotation(taskId string, execution int, issue IssueLink, username string) error {
 	issue.Source = &Source{
 		Author:    username,
@@ -192,14 +144,6 @@ func AddSuspectedIssueToAnnotation(taskId string, execution int, issue IssueLink
 	return errors.Wrapf(err, "adding task annotation suspected issue for task '%s'", taskId)
 }
 
-func RemoveIssueFromAnnotation(taskId string, execution int, issue IssueLink) error {
-	return db.Update(
-		TaskAnnotationsCollection,
-		ByTaskIdAndExecution(taskId, execution),
-		bson.M{"$pull": bson.M{IssuesKey: issue}},
-	)
-}
-
 func RemoveSuspectedIssueFromAnnotation(taskId string, execution int, issue IssueLink) error {
 	return db.Update(
 		TaskAnnotationsCollection,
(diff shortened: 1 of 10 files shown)


model/task/task.go
```diff
@@ -3522,3 +3526,187 @@ func (t *Task) FindAbortingAndResettingDependencies() ([]Task, error) {
...
+	if _, err := db.Upsert(
+		annotations.TaskAnnotationsCollection,
+		annotations.ByTaskIdAndExecution(a.TaskId, a.TaskExecution),
+		bson.M{
+			"$set": update,
+		},
+	); err != nil {
+		return errors.Wrapf(err, "adding task annotation for task '%s'", a.TaskId)
+	}
+	if a.Issues != nil {
```
malikchaya2, line 3665: we should check that the length is not zero as well

model/task/task.go
```diff
@@ -3522,3 +3526,187 @@ func (t *Task) FindAbortingAndResettingDependencies() ([]Task, error) {
...
+	if err = db.Update(
+		annotations.TaskAnnotationsCollection,
+		annotations.ByTaskIdAndExecution(a.TaskId, a.TaskExecution),
+		bson.M{
+			"$push": update,
+		},
+	); err != nil {
+		return errors.Wrapf(err, "updating task annotation for '%s'", a.TaskId)
+	}
+	if a.Issues != nil {
```
malikchaya2, line 3708: should check length

model/task/db.go
```diff
@@ -195,10 +198,13 @@ var (
...
-						"$ne": []interface{}{
-							bson.M{
-								"$size": bson.M{"$ifNull": []interface{}{"$annotation_docs", []bson.M{}}},
-							}, 0,
+						"$or": []bson.M{
+							{"$eq": []interface{}{"$" + HasAnnotationsKey, true}},
+							{"$ne": []interface{}{
+								bson.M{
+									"$size": bson.M{"$ifNull": []interface{}{"$annotation_docs", []bson.M{}}},
+								}, 0,
```
malikchaya2, line 206: Do we still need this now that we can use hasAnnotation to calculate the display status?

model/task/db.go
```diff
@@ -292,11 +298,15 @@ var (
...
-				// We skip annotation lookup for non-failed tasks, because these can't have annotations
+				// We skip annotation lookup for non-failed tasks, because these can't have annotations,
+				// and lookup is not necessary for tasks that already have HasAnnotationsKey set
 				"not_failed": []bson.M{
 					{
 						"$match": bson.M{
-							StatusKey: bson.M{"$nin": evergreen.TaskFailureStatuses},
+							"$or": []bson.M{
+								{HasAnnotationsKey: true},
+								{StatusKey: bson.M{"$nin": evergreen.TaskFailureStatuses}},
```
malikchaya2, line 308: Once we remove the backwards compatibility, we should remove this because we'll only ever have to look it up if it's set to true (and it'll already only be set to true if it's failed) . It may be a good idea to add a todo comment so we don't forget.

model/task/db.go
```diff
@@ -2619,11 +2634,20 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
-
-	if opts.UseSlowAnnotationsLookup {
-		pipeline = append(pipeline, AddAnnotationsSlowLookup...)
-	} else {
-		pipeline = append(pipeline, AddAnnotations...)
+	// TODO: DEVPROD-___ Remove this conditional and annotation lookup pipelines
+	// Pull one task, any task associated with this versionID. A task's CreateTime is
+	// derived from the version commit time or the patch creation time, allowing us to treat it
+	// as a proxy for the version creation time.
+	dbTask, err := FindOne(db.Query(bson.M{VersionKey: versionID}))
```
malikchaya2, line 2641: Why don't we pass in the `create_time` from the version when calling this function so that we can save this lookup?

model/task/task_annotations_test.go
```diff
@@ -0,0 +1,155 @@
+package task
```
malikchaya2, line 1: Can we add some checks to these tests to make sure that they are setting HasAnnotations as expected? Maybe include some examples of when they shouldn't set it?

graphql/tests/mutation/addAnnotationIssue/data.json
```diff
@@ -1 +1,8 @@
-{}
+{
```
malikchaya2, line 1: I don't see the result files to complete these tests. Did you forget to push them?

hadjri:
evergreen retry

model/task/db.go
hadjri: Yes because this will still be needed for old versions in the period that we're maintaining backward compatability.

model/task/db.go
hadjri: That would be the ideal way of doing this but the full version document isn't always available in all of the usages of this function. Since this is a temporary addition, rather than adding version db queries in all of those places I just added one query here to be removed later.

model/task/db.go
hadjri: Sounds good - added

model/task/task_annotations_test.go
hadjri: Sure, added some more checks & cases

graphql/tests/mutation/addAnnotationIssue/data.json
hadjri: This was just added to ensure that `task-1234` exists at the time of doing [this](https://gist.github.com/hadjri/b6cf1c8b37c84a932cdeb6f4313d87f8) mutation, or else the graphql test would error.

model/task/task.go
```diff
@@ -3516,3 +3520,187 @@ func (t *Task) FindAbortingAndResettingDependencies() ([]Task, error) {
 	})
 	return FindAll(q)
 }
+
+// MoveIssueToSuspectedIssue removes an issue from an existing annotation and adds it to its suspected issues,
```
ablack12, line 3524: Could we have a new file called task/annotations.go or something, just for organization?

ablack12 (changes requested):

model/task/db.go
```diff
@@ -25,6 +25,8 @@ const (
 	OldCollection = "old_tasks"
 )
 
+var annotationLookupDeprecationTime = time.Date(2024, time.January, 18, 23, 0, 0, 0, time.UTC)
```
khelif96, line 28: I'm assuming this will be updated to the date this is merged? Just wanted to flag.

model/task/db.go
```diff
@@ -292,11 +298,16 @@ var (
 	AddAnnotations = []bson.M{
 		{
 			"$facet": bson.M{
```
khelif96, line 300: I don't think this is the correct approach. If the version was created after the annotations cache key was added we shouldn't even try to do this `$facet`. `$facets` have a lot of [limitations](https://www.mongodb.com/docs/manual/reference/operator/aggregation/facet/#considerations) and should generally be avoided. This was originally added because we wanted to reduce the amount of tasks we perform a `$lookup` for. 
After this is deployed we will hit the [100 MB aggregation limit](https://arc.net/l/quote/fcvemzqz) for facets on every single query we perform for `mongodb-mongo-master` project and it will [spill over to](https://github.com/evergreen-ci/evergreen/blob/c5495d494db06a2fb5a0422194dddd1df63ccb0f/model/task/db.go#L2431) the [`AddAnnotationsSlowLookup`](https://github.com/evergreen-ci/evergreen/blob/c5495d494db06a2fb5a0422194dddd1df63ccb0f/model/task/db.go#L362) and it will eliminate all performance gains this pr introduces. (I think it would actually become dramatically slower)

khelif96 (changes requested):
In the follow-up ticket can we also make a note to remove this piece of code.
https://github.com/evergreen-ci/evergreen/blob/df897c37f21e6787759555d4f42fa67813e8b3e7/model/task/db.go#L2132-L2137

model/task/db.go
```diff
@@ -292,11 +298,16 @@ var (
...
-				// We skip annotation lookup for non-failed tasks, because these can't have annotations
+				// We skip annotation lookup for non-failed tasks, because these can't have annotations,
+				// and lookup is not necessary for tasks that already have HasAnnotationsKey set
+				// TODO: DEVPROD-___ simplify this step to only check HasAnnotationsKey
 				"not_failed": []bson.M{
 					{
 						"$match": bson.M{
-							StatusKey: bson.M{"$nin": evergreen.TaskFailureStatuses},
+							"$or": []bson.M{
+								{HasAnnotationsKey: true},
```
khelif96, line 308: I'm confused by what this does? 
https://github.com/hadjri/evergreen/blob/c5495d494db06a2fb5a0422194dddd1df63ccb0f/model/task/db.go#L2639 avoids this check if the task was stage if the version was created after this change.

model/annotations/task_annotations.go
```diff
@@ -194,89 +144,15 @@ func AddSuspectedIssueToAnnotation(taskId string, execution int, issue IssueLink
...
-
-		if err != nil {
-			return errors.Wrap(err, "bulk inserting annotations")
-		}
-	}
-	return nil
-}
-
-func createAnnotationUpdate(annotation *TaskAnnotation, userDisplayName string) bson.M {
+func CreateAnnotationUpdate(annotation *TaskAnnotation, userDisplayName string) bson.M {
```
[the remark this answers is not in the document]
hadjri: It's only used by one other function in a separate package so I'm actually going to move it and put it back as private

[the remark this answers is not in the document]
hadjri: Fixed

model/task/task.go
hadjri: Done

model/task/db.go
hadjri: Thanks, updated to today.

model/task/db.go
```diff
@@ -2612,11 +2628,20 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions)
...
-
-	if opts.UseSlowAnnotationsLookup {
-		pipeline = append(pipeline, AddAnnotationsSlowLookup...)
-	} else {
-		pipeline = append(pipeline, AddAnnotations...)
+	// TODO: DEVPROD-___ Remove this conditional and annotation lookup pipelines
+	// Pull one task, any task associated with this versionID. A task's CreateTime is
+	// derived from the version commit time or the patch creation time, allowing us to treat it
+	// as a proxy for the version creation time.
+	dbTask, err := FindOne(db.Query(bson.M{VersionKey: versionID}))
```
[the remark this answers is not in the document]
hadjri: I added a `.WithFields(CreateTimeKey)` since it's the only referenced field here

model/task/db.go
hadjri: Yeah this would allow old versions to run the faster version of the pipeline should they get annotations added to them moving forward, but this is a negligible benefit / probably not worth worrying about. Changing it back to how it was.

hadjri:
[DEVPROD-3994](https://jira.mongodb.org/browse/DEVPROD-3994)

model/task/task_annotations.go
```diff
@@ -0,0 +1,195 @@
...
+
+// MoveIssueToSuspectedIssue removes an issue from an existing annotation and adds it to its suspected issues,
+// and unsets its associated task document as having annotations if this was the last issue removed from the
+// annotation.
+func MoveIssueToSuspectedIssue(taskId string, taskExecution int, issue annotations.IssueLink, username string) error {
+	newIssue := issue
+	newIssue.Source = &annotations.Source{Requester: annotations.UIRequester, Author: username, Time: time.Now()}
+	q := annotations.ByTaskIdAndExecution(taskId, taskExecution)
+	q[bsonutil.GetDottedKeyName(annotations.IssuesKey, annotations.IssueLinkIssueKey)] = issue.IssueKey
+	if err := db.Update(
```
ablack12, line 21: Have you considered using FindAndModify for this (and for RemoveIssue) instead of update and then a subsequent find? It combines those calls and makes the updates a bit more atomic.

ablack12 (approved):
LGTM with last suggestion

khelif96 (approved):
LGTM reminder to update the deprecation date with the deploy date before merging.

malikchaya2 (approved):

model/task/task_annotations.go
hadjri: Sounds good, I added it for those two functions.

PR #7422, opened 2024-01-18
DEVPROD-1947 Handle GitHub triggering two patch intents with same SHA

ZackarySantana:
DEVPROD-1947

#### Problem
GitHub only allows for one status per commit SHA. When there are two PR's open with the HEAD branch pointing to the same SHA, Evergreen currently creates two patches and sends statuses for both. When both finish, whatever finished last is what gets displayed for both PR's. Not only is this a waste of resources in most cases, if they have tests that require the context of a particular PR, they may be getting invalid test results than what they expect (or don't expect).

#### Solution
When Evergreen recieves a 2nd github event that it would create another patch for, but one exists, instead comment on the PR explaining why a new one will not be created and link the other PR. In most cases, tests won't depend on the PR context. If the test does, the user can type 'evergreen retry' on any PR to cancel other patches and it will inform on all affected PR's. 

#### Video
I only did a video of a working use case, I also tested 'evergreen retry' to make sure it behaves the same as before. I checked the links as well and they work. I also tested 'evergreen retry' on both PR's (one with a patch, one without one). And doing retry multiple times.

https://github.com/evergreen-ci/evergreen/assets/64446617/d0fd7344-6a9c-49dd-9bac-3d76102384e7

### Testing
Manual testing 

### Documentation
Added


globals.go
```diff
@@ -604,7 +604,7 @@ var (
 )
 
 const (
-	DefaultServiceConfigurationFileName = "/etc/mci_settings.yml"
+	DefaultServiceConfigurationFileName = "/etc/mci_settings.yml" // Unused
```
ZackarySantana, line 607: Should I delete these unused variables in this PR? I stumbled upon them while working on this

ZackarySantana:
If you are interested in the branches I was testing on and want to see it, here are two of them: 
https://github.com/evergreen-ci/commit-queue-sandbox/pull/634

https://github.com/evergreen-ci/commit-queue-sandbox/pull/654

Please note they have a sloppy history from testing. 

And as well, users can get in a state where 'evergreen retry' just makes a new patch on top of an existing one- we don't have any checks if one is currently running (or will run) and we don't cancel it as such. So in one of my tests, there displays the same PR but two different patches, that was from testing 'evergreen retry' when a patch was already made. I was going to make a ticket to address this.

rest/route/github.go
```diff
@@ -583,6 +589,102 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
...
+			"ref":               pr.Head.GetRef(),
+			"pr_num":            pr.GetNumber(),
+			"hash":              pr.Head.GetSHA(),
+			"override_existing": overrideExisting,
+			"err":               err.Error(),
+		})
+		return errors.Wrapf(err, "getting same hash patches")
+	}
+	// Remove a patch associated with this PR if applicable.
+	var filtered = []patch.Patch{}
```
Kimchelly, line 607: nit: this is equivalent to `filtered := []patch.Patch{}`, so the `var` can be omitted. Alternatively, it seems like your logic doesn't really care if the slice is nil vs empty (both would be treated the same in this function), so you could just do `var filtered []patch.Patch` to have a nil unallocated slice, which is auto-allocated when it's `append`ed to.

rest/route/github.go
```diff
@@ -558,7 +559,12 @@ func (gh *githubHookApi) refreshPatchStatus(ctx context.Context, owner, repo str
 	return nil
 }
 
-func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy string) error {
+// AddIntentForPR takes a pr and checks if that pr is applicable for CI and if it is, does one of two things.
+// If overrideExisting is false, it only creates a PR if no PR has the same head sha. If any exist, it
```
Kimchelly, line 563: This comment overall looks good (and thanks for adding a comment!), but a brief explanation of what the head SHA is would help clarify what the issue is that we're trying to solve/why it can happen so we understand this special case later on.

rest/route/github.go
```diff
@@ -583,6 +589,102 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
...
+			var errs []error
+			overriddenPatchComment := fmt.Sprintf("Another [PR](%s) with the same head SHA has ran 'evergreen retry' and overridden this PR's patch. This PR's patch will be canceled and the status reported will be in the context of the other PR.", createGitHubPRLink(pr.Base.User.GetLogin(), pr.Base.Repo.GetName(), pr.GetNumber()))
+			for _, p := range patches {
+				err := gh.sc.AddCommentToPR(ctx, p.GithubPatchData.BaseOwner, p.GithubPatchData.BaseRepo, p.GithubPatchData.PRNumber, overriddenPatchComment)
+				if err != nil {
+					errs = append(errs, err)
+				}
+			}
+
+			overridingPatchComment := fmt.Sprintf("There is an existing patch(s) for this commit SHA that will be cancelled:\n%s\n\nThe status reported will be corresponding to this PR rather than the previous existing ones. If you would like a patch to run for another PR and to cancel this one, comment 'evergreen retry' on the corresponding PR.", gh.createPatchesWithPRLinks(patches))
```
Kimchelly, line 665: nit: patch(s) -> patch(es)

rest/route/github.go
```diff
@@ -583,6 +589,102 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
...
+				"message": "cancelling existing CI on same SHA patches",
+				"owner":   pr.Base.User.GetLogin(),
+				"repo":    pr.Base.Repo.GetName(),
+				"ref":     pr.Head.GetRef(),
+				"pr_num":  pr.GetNumber(),
+			})
+			for _, p := range patches {
+				err = model.CancelPatch(&p, task.AbortInfo{User: evergreen.GithubPatchUser, NewVersion: "", PRClosed: false})
+				if err != nil {
+					err = gh.sc.AddCommentToPR(ctx, pr.Base.User.GetLogin(), pr.Base.Repo.GetName(), pr.GetNumber(), "There was an issue aborting the other patches, please try 'evergreen retry' again.")
```
Kimchelly, line 639: This comment seems fine, but it may result in some confusion since it returns early and doesn't comment about the originating problem first (the overriding patch comment below this). Could we add this comment _after_ posting the comment with the context?

rest/route/github.go
```diff
@@ -583,6 +589,102 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
 		}
 	}
 
+	patches, err := patch.Find(patch.ByGithash(pr.Head.GetSHA()))
+	if err != nil {
+		grip.Info(message.Fields{
```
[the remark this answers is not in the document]
ZackarySantana: I think it should be error

rest/route/github.go
```diff
@@ -583,6 +589,102 @@ func (gh *githubHookApi) AddIntentForPR(pr *github.PullRequest, owner, calledBy
...
+	if len(patches) > 0 {
+		if !overrideExisting {
+			// We want to comment on explaining why a new patch will not be ran.
+			comment := fmt.Sprintf("There is an existing patch(s) for this commit SHA:\n%s\n\nPlease note that the status that is posted is not in the context of this PR but rather the (latest) existing patch and that may affect some tests that may depend on the particular PR. If your tests do not rely on any PR-specific values (like base or head branch name) than your tests will report the same status. If you would like a patch to run in the context of this PR and cancel the other(s), comment 'evergreen retry'.", gh.createPatchesWithPRLinks(patches))
+			grip.Info(message.Fields{
+				"message": "skipping CI on PR due to patch already existing",
+				"owner":   pr.Base.User.GetLogin(),
+				"repo":    pr.Base.Repo.GetName(),
+				"ref":     pr.Head.GetRef(),
+				"pr_num":  pr.GetNumber(),
```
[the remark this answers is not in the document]
ZackarySantana: I reused the links in this info so it just links the PR and patches for quick debugging

ZackarySantana:
I refactored and pulled out the comments to a dedicated file so they can be colocated and not clutter the logic code.

rest/route/github_comments.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/evergreen-ci/evergreen/model/patch"
+	"github.com/google/go-github/v52/github"
+)
+
+type githubComments struct {
+	baseUrl string
+}
+
+func newGithubComments(baseUrl string) *githubComments {
```
Kimchelly, line 14: Small nugget of wisdom - if you don't have a specific reason to use a pointer, generally try to avoid it. Pointers introduce more surface area for nils and panics, and historically, this fueled a paranoid mentality of "but what if it's a nil pointer??" and resulted in lots of extra panic safety checks that might not actually be needed. You can see some examples of this in the `pr` object, where we use `Get<Field>` everywhere because it's full of pointers and we're not sure what fields GitHub is guaranteed to send us/which ones have already been nil checked.

On a further note about pointers and methods - even if some of its method definitions use a pointer receiver (like `(g * githubComments) existingPatches`), it's still okay to return the struct from constructors - Go is smart enough to figure out the magic of working on the pointer to the struct rather than copying the struct with each method. It seems like none of this needs to use pointers, so just returning the struct as-is from `newGithubComments` should be good enough.

docs/Project-Configuration/Github-Integrations.md
```diff
@@ -72,6 +72,10 @@ evergreen refresh
 
 Sometimes Evergreen has trouble sending updated GitHub statuses, so the checks on the PR may not accurately reflect the state of patch on Evergreen. This is especially troublesome when the repository requires passing checks. To re-sync the GitHub checks with Evergreen, comment `evergreen refresh` on the PR.
 
+#### Handling same SHA PR's
+
+GitHub only allows one set of statuses for every commit SHA. If you have a situation where you have two PR's where the HEAD commit is the same, Evergreen will only create a patch for the first one and will make a comment on the second one explaining why. In general, if your tasks do not require the context of the PR (most notably the branch name, which is known to be used in some s3.put tasks), then the status displayed for the PR's will be correct. If the tasks do require the context of the PR, you may comment 'evergreen retry' and force Evergreen to cancel currently running patches in favor of a new one in the context of the PR you have commented on.
```
Kimchelly, line 77: nit: PR's -> PRs

rest/route/github.go
```diff
@@ -603,90 +606,83 @@ func (gh *githubHookApi) AddIntentForPR(ctx context.Context, pr *github.PullRequ
...
+	grip.Error(message.WrapError(catcher.Resolve(), message.Fields{
+		"message": "error informing user(s) that aborting other patches failed",
+		"owner":   pr.Base.User.GetLogin(),
+		"repo":    pr.Base.Repo.GetName(),
+		"ref":     pr.Head.GetRef(),
+		"pr_num":  pr.GetNumber(),
+	}))
+
+	// Abort patches, and if any fail, comment on the PR to inform the user and no-op.
+	for _, p := range patches {
```
Kimchelly, line 670: May be a little neater to abort patches in the same loop as above, just after adding the comment. Would also make sense to continue on error in case any single patch fails to abort, since we still want to try aborting later PR patches in the slice.

rest/route/github.go
```diff
@@ -589,9 +592,9 @@ func (gh *githubHookApi) AddIntentForPR(ctx context.Context, pr *github.PullRequ
 		}
 	}
 
-	patches, err := patch.Find(patch.ByGithash(pr.Head.GetSHA()))
+	conflictingPatches, err := getOtherPatchesWithHash(pr.Head.GetSHA(), *pr.Number)
```
Kimchelly, line 595: nit: `pr.GetNumber()` to reduce nil pointer paranoia

Kimchelly (commented):
I like the code organization changes a lot 💯 

Still have a couple follow-ups, but they're not big.

rest/route/github.go
```diff
@@ -603,90 +606,83 @@ func (gh *githubHookApi) AddIntentForPR(ctx context.Context, pr *github.PullRequ
...
+	}))
+
+	// Abort patches, and if any fail, comment on the PR to inform the user and no-op.
+	for _, p := range patches {
+		err := model.CancelPatch(&p, task.AbortInfo{User: evergreen.GithubPatchUser, NewVersion: "", PRClosed: false})
+		if err == nil {
+			continue
+		}
+		commentErr := gh.sc.AddCommentToPR(ctx, pr.Base.User.GetLogin(), pr.Base.Repo.GetName(), pr.GetNumber(), "There was an issue aborting the other patches, please try 'evergreen retry' again.")
+		grip.Error(message.Fields{
```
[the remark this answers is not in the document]
ZackarySantana: I looked inside of wraperror/fields and passing the commentErr is okay, it handles/expects nil conditions in particular

rest/route/github_comments.go
ZackarySantana: Yeah... I even stored it as a struct instead of a pointer in the gh (githubHookApi) struct. Moved to not use a pointer

rest/route/github.go
ZackarySantana: I combined them and made a comment for the function to explain the design choice. I also put two grip logging, one for comments and one for patch cancels, with different levels since erroring on comments is not as important as erroring on patches.

I could see a case where it might be good to error on comments as well for information, if  you think so I could change it so they are treated the same.

rest/route/github_comments.go
```diff
@@ -0,0 +1,56 @@
...
+
+// overridingPR returns a comment that is posted when there are existing patches for a commit SHA and
+// they will be aborted, so the PR context is of the new PR rather than the existing PR.
+func (g *githubComments) overridingPR(patches []patch.Patch) string {
+	return fmt.Sprintf("There is an existing patch(es) for this commit SHA that will be aborted:\n%s\n\nThe status reported will be corresponding to this PR rather than the previous existing ones. If you would like a patch to run for another PR and to abort this one, comment 'evergreen retry' on the corresponding PR.", g.getLinksForPRPatches(patches))
+}
+
+// overridenPR returns a comment that is posted when a PR's patch is aborted (or changed) in
+// favor of anther patch in a different PR context.
+func (g *githubComments) overridenPR(pr *github.PullRequest) string {
```
Kimchelly, line 32: nit: Sorry, English is hard but this one specifically is "overridden", only "overriding" has a single "d". I think the correct combination is overridden + overriding.

rest/route/github.go
Kimchelly: Eh, splitting the error logs seems fine. LGTM!

Kimchelly (approved):
LGTM mod small nits. And nice job with the PR!

PR #7424, opened 2024-01-18
DEVPROD-1931 Error on module clone failure

bynn:
DEVPROD-1931

### Description
Error instead of warn at module clone failure
Previously reverted because it broke some server projects.


Kimchelly (approved):
🤞

PR #7425, opened 2024-01-18
DEVPROD-726 Add validation for git.get_project using modules

ZackarySantana:
DEVPROD-726

### Description
Ticket details/comments has all the context, but this adds warnings if a user does a git.get_project command and the buildvariant the test in is does not have that module.

### Testing
Manual testing on some custom yamls (like the one linked) and the mongodb/mongo yaml. As well as some unit tests.


validator/project_validator.go
```diff
@@ -1951,6 +1952,58 @@ func validateGenerateTasks(p *model.Project) ValidationErrors {
 	return validateTimesCalledPerTask(p, ts, evergreen.GenerateTasksCommandName, 1, Error)
 }
 
+func validateModuleUsageInGitGetProject(p *model.Project) ValidationErrors {
+	var errs ValidationErrors
```
malikchaya2, line 1956: Lets rename this to ValidationWarnings

malikchaya2 (approved):
approved with nit

validator/project_validator.go
```diff
@@ -1951,6 +1952,58 @@ func validateGenerateTasks(p *model.Project) ValidationErrors {
...
+
+	bvsTaskCmds, _, err := bvsWithTasksThatCallCommand(p, "git.get_project")
+	if err != nil {
+		errs = append(errs, ValidationError{
+			Message: fmt.Sprintf("build variants could not be mapped to tasks that contain 'git.get_project' for project %s", p.Identifier),
+			Level:   Error,
```
malikchaya2, line 1962: This should probably be a warning too

validator/project_validator.go
```diff
@@ -139,6 +139,7 @@ var projectErrorValidators = []projectValidator{
 	validateHostCreates,
 	validateDuplicateBVTasks,
 	validateGenerateTasks,
+	validateModuleUsageInGitGetProject,
```
ablack12, line 142: I think validators that warn should be in projectWarningValidators and errors should be in this list. cc @hadjri to verify.
(I wonder if we should make some sort of linter or test to verify that we keep things in that state? Seems easy to miss and then we'll have to fix everything again)

validator/project_validator.go
ZackarySantana: I don't think we have a validationwarnings struct- other functions which only do warnings have the same nomenclature/struct. What should I do instead?

validator/project_validator.go
ZackarySantana: I think so! I forgot to move it to the warnings rather than the errors check, I'll move it and retest to make sure it still propagates

validator/project_validator.go
ZackarySantana: I think since other functions do as such and there's some reusable logic we use for ValidationErrors to store warnings and errors, I think I'll leave this as is unless you think otherwise

validator/project_validator.go
hadjri: Yup can confirm this would belong in the warning list.

hadjri: Created https://jira.mongodb.org/browse/DEVPROD-3996 re your suggestion @ablack12

PR #7437, opened 2024-01-22
DEVPROD-3671 Add better checks for Bisect Stepback

ZackarySantana:
DEVPROD-3671

### Description
This consolidates the logic for initiating stepback as well as the logic for an existing stepback or not. 
As well, this makes it so when stepback errors, the MarkEnd will not. (aka it will continue and do mission-critical actions).

### Testing
Existing unit tests. Manual testing in staging


model/task_lifecycle.go
```diff
@@ -1309,8 +1309,8 @@ func evalBisectStepback(ctx context.Context, t *task.Task, caller string, stepba
 	// If the stepback info is nil but we reached this point, this must be the first
 	// iteration of stepback.
 	newStepback := t.StepbackInfo == nil && evergreen.IsFailedTaskStatus(t.Status)
-	// If the stepback is not nil, this is an ongoing stepback.
-	existingStepback := t.StepbackInfo != nil
+	// If the stepback is not nil and it's values are not empty, this is an ongoing stepback.
```
Kimchelly, line 1312: grammar nit: its

model/task_lifecycle.go
```diff
@@ -616,7 +616,7 @@ func doBisectStepback(ctx context.Context, t *task.Task) error {
 	// last failing or last passing task.
 	if t.Status == evergreen.TaskSucceeded {
 		s.LastPassingStepbackTaskId = t.Id
-	} else if t.Status == evergreen.TaskFailed {
+	} else if evergreen.IsFailedTaskStatus(t.Status) {
```
Kimchelly, line 619: Maybe L1312 would change the logic, but I don't think this change would actually change how this logic actually runs. `(task.Task).Status` stores the DB model's status, which can only be `evergreen.TaskSucceeded` or `evergreen.TaskFailed` - all other statuses are display statuses, which have meaning in the UI but would never be stored in `(task.Task).Status`. There's [some discussion of this distinction between DB status vs. UI status here](https://github.com/evergreen-ci/evergreen/blob/59404295454cf10be3aead058fbe14fdf10444ce/globals.go#L45-L68) if you're interested.

In addition, it may be worth either:
1. Calling out in the doc comment on `evalStepback` that the `status` passed to `evalStepback` is the _display status_ (i.e. what the UI shows) rather than the DB task status (i.e. `(task.Task).Status`) or
2. If you don't have a specific need to check specific failure types like setup-failed/system-failed (I glanced at linear and bisect and don't see any indication that it needs to check the display status), you can just remove the `status` passed in and just use `(task.Task).Status` where necessary in linear stepback.

Kimchelly (changes requested):
Not sure if you still need extra debugging info as people turn on bisect stepback, but one way you could diagnose further is to add a Grip log after `evalStepback` returns so that if bisect stepback errors, you can track the error trace more easily in Splunk.

model/task_lifecycle.go
ZackarySantana: That makes sense. I don't think either need specific statuses either. I'm going to try debugging more on how it failed for mms then

ZackarySantana: I looked in to it a bit and stepback does use the display status in terms of activation (some [work](https://github.com/evergreen-ci/evergreen/pull/6062) Annie did). I think it makes sense to have bisect stepback activate on the same conditions, and I made it a bit more clear what is happening and why we need the display status

ZackarySantana:
I'm not 100% sure this will fix the issue mms was having with bisect stepback, bc I'm unable to reproduce it- this does add some better checks for bisect stepback and conjoin some logic. It also fixes bisect to have the same activation conditions as linear now.

I think the best plan is to try enabling back in a project (maybe mms) and seeing if any splunk logs from stepback failing, and see why that is (if it continues to fail).

ablack12:
Didn't review this code, but if we aren't certain we've fixed the issue we may want to consider doing https://jira.mongodb.org/browse/DEVPROD-3670 sooner to mitigate issues with it.

model/task_lifecycle.go
```diff
@@ -1251,23 +1256,28 @@ func removeNextMergeTaskDependency(cq commitqueue.CommitQueue, currentIssue stri
...
+// The status passed in is a display status, initially stepback only activates if the task
+// has failed but not on system failure.
 func evalStepback(ctx context.Context, t *task.Task, caller, status string, deactivatePrevious bool) error {
 	s, err := getStepback(t.Id)
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	// A new stepback happens when the display status is failed (not system failure)
+	// and it is not aborted.
+	newStepback := status == evergreen.TaskFailed && !t.Aborted
```
Kimchelly, line 1268: Since this is checking the display status, this looks like it would exclude other display statuses like `evergreen.TaskSetupFailed`/`evergreen.TaskTimedOut` from initiating stepback, but we only want to skip for abort + system failures. Can we have this just exclude aborted tasks and system-failed tasks specifically?

model/task_lifecycle_test.go
```diff
@@ -6319,7 +6319,7 @@ func TestEvalStepbackDeactivatePrevious(t *testing.T) {
 	finishedTask := task.Task{
 		Id:                  "t3",
 		BuildId:             "b3",
-		Status:              evergreen.TaskUndispatched,
+		Status:              evergreen.TaskSucceeded,
```
Kimchelly, line 6322: Why does this test check a different status? Does the (linear) stepback logic behave differently now?

ZackarySantana:
This PR should remove that particular issue with stepback in MarkEnd, but not with other functionalities included in the MarkEnd function.
I think this stepback change can be handled here separately and that ticket just handle all the other functionality and even reconsider how errors are handled in MarkEnd

I was planning on picking up the MarkEnd one once the next quarter started

Kimchelly (changes requested):
One question about the display status logic change, but otherwise looks good! 👍
I'm also fine with just having stepback continue on error for now and addressing more cases for DEVPROD-3670 another time.

re: the enabling question, I'm not sure we would actually want to test on mms since they rely on stepback on a day-to-day basis to detect failure causes - is there a way we can test this in an Evergreen-owned repo or staging first? Maybe intentionally commit a bunch of things and cause some setup failures?

model/task_lifecycle.go
```diff
@@ -784,12 +784,17 @@ func MarkEnd(ctx context.Context, settings *evergreen.Settings, t *task.Task, ca
...
 				return errors.Wrap(err, "getting display task")
 			}
-			err = evalStepback(ctx, t.DisplayTask, caller, t.DisplayTask.Status, deactivatePrevious)
+			err = evalStepback(ctx, t.DisplayTask, caller, status, deactivatePrevious)
 		} else {
 			err = evalStepback(ctx, t, caller, status, deactivatePrevious)
 		}
 		if err != nil {
-			return errors.Wrap(err, "evaluating stepback")
+			grip.Error(message.Fields{
```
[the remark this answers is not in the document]
ZackarySantana: Nice cutdown

model/task_lifecycle.go
ZackarySantana: Hm, I'm not entirely sure what statuses should exclude stepback or not- I simply kept the same activation conditions that were originally for linear stepback.

Looking in globals.go, we have a slice of `TaskSystemFailures` which includes TaskTimedOut- the only two we have that aren't apart of the TaskSystemFailures slice is TaskFailed & TaskSetupFailed.

Should I edit it to activate stepback for the 3 statuses (TaskFailed, TaskSetupFailed, and TaskTimedOut)- or just 2 (TaskFailed, TaskSetupFailed)?

model/task_lifecycle_test.go
ZackarySantana: I'm not sure why originally the status here was set as `TaskUndispatched` but the test later gives evalStepback `TaskSucceeded`- which causes a mismatch of the statuses.

The test is testing if a linear stepback iteration succeeded it should deactivate the previous one.

The reason this wasn't previously erroring before is that we would pass what evalStepback gets in to linear stepback- but now linear stepback just reads it off of the task (L1304).

i.e. it was a test data inconsistency

Kimchelly (commented):
The code change LGTM, mod we should document more explicitly under what conditions Evergreen chooses to initiate stepback.

ZackarySantana:
I'll add some docs outlining that it can only be the TaskFailed status and not our other display statuses, they were different in conditions (I did not do that intentionally originally, it was just from my lack of understanding of display statuses- oops).

docs/Project-Configuration/Stepback-Bisection.md
```diff
@@ -1,5 +1,5 @@
 # Stepback Bisection
-Evergreen's stepback bisection performs stepback by continuously reducing the amount of commits needed to test by half. To enable it, go to the Project Settings page "General" tab and enable the 'Stepback Bisection' flag. For this flag to work, the 'Stepback' flag will also need to be enabled.
+Evergreen's stepback bisection performs stepback by continuously reducing the amount of commits needed to test by half. To enable it, go to the Project Settings page "General" tab and enable the 'Stepback Bisection' flag. For this flag to work, the 'Stepback' flag will also need to be enabled. As well, stepback only activates on tasks that have failed- if the task has any other status (timed out, setup failed, system failed, etc.), stepback will not activate.
```
Kimchelly, line 2: phrasing nit: maybe "stepback only activates on tasks that have failed normally and not for more specific reasons (like setup failed, system failed, etc.)"

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -963,10 +963,10 @@ must be present in the `tasks` array.
 
 ### Stepback
 
-Stepback is set to true if you want to stepback and test earlier commits
-in the case of a failing task. This can be set or unset at the
-top-level, at the build variant level, and for individual tasks (in the task definition or for the
-task within a specific build variant).
+Stepback is set to true if you want to stepback and test earlier commits in the case
+of a failing task (a failing task does not include system failed, setup failed, timed out, etc.).
```
Kimchelly, line 967: "normally failing task" since the others technically are failures as well, they're just more specific failure types.

Kimchelly (approved):
LGTM mod trivial phrasing suggestions.

PR #7438, opened 2024-01-22
DEVPROD-746 Delete unused bvtu level exec timeout

bynn:
DEVPROD-746

### Description
it was never being used and just being passed around so it is being deleted

### Testing
current tests pass


Kimchelly (changes requested):
The code change itself LGTM, just docs questions:
* It looks like the docs still use a bvtu-level exec timeout as [its example configuration](https://github.com/evergreen-ci/evergreen/blob/76e5398afd3db461ce8da972c3ac12502df1d700/docs/Project-Configuration/Project-Configuration-Files.md?plain=1#L227). Can we move the example field somewhere else, like the task definition or something?
* In the docs above, I noticed the field is never documented where you can set it (at the top level or in a task definition), and the other fields have a short description of what they do. Can we document it now or alternatively, make a separate ticket to document `exec_timeout_secs`?

bynn:
TIcket created for documentation + adding example: https://jira.mongodb.org/browse/DEVPROD-4063 (which was maybe more work than just writing the doc haha)

bynn:
deleted the example in bvtu though

Kimchelly (approved):

PR #7439, opened 2024-01-22
DEVPROD-777: document missing archive commands

Kimchelly:
DEVPROD-777

### Description
* Document `archive.auto_extract`, `archive.zip_create`, and `archive.zip_extract` commands.
* Remove some random dead/non-working parameters from the newly-documented archive functions. It looks like they were never implemented, so it's not worth keeping them around.

### Testing
Smoke test already tests these commands, and it still passes.

### DIFF
--- a/config.go
+++ b/config.go
@@ -35,8 +35,9 @@ var (
 	// ClientVersion is the commandline version string used to control auto-updating.
 	ClientVersion = "2024-01-18"
 
-	// Agent version to control agent rollover.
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -35,8 +35,9 @@ var (
 	// ClientVersion is the commandline version string used to control auto-updating.
```
bynn, line 35: nit: should we add the same format for client version

bynn (approved):

PR #7441, opened 2024-01-23
DEVPROD-1723: Handle empty log lines correctly and add knob to turn on/off evergreen logging for projects

julianedwards:
DEVPROD-1723

### Description
- Handle empty log lines correctly in the log service implementation to avoid panics
- Split test log lines correctly and ignore empty lines
- Add temporary knob to enable evergreen logging for a subset of projects

### Testing
Added new unit testing for panics and test log line splitting. Tested in staging that the evergreen logging projects knob is respected.

### Documentation
N/A


taskoutput/task_output.go
```diff
@@ -25,7 +28,7 @@ func InitializeTaskOutput(env evergreen.Environment, opts TaskOptions) *TaskOutp
 	settings := env.Settings()
 
 	output := &TaskOutput{}
-	if settings.LoggerConfig.DefaultLogger != "buildlogger" {
+	if settings.LoggerConfig.DefaultLogger == "evergreen" || utility.StringSliceContains(settings.LoggerConfig.EvergreenLoggerProjects, opts.ProjectID) {
```
ablack12, line 31: Can you use model.EvergreenLogSender here directly or is there an import cycle? (I assume this is temporary in which case no need to re-organize if it does cycle)

julianedwards: It causes an import cycle :(

PR #7443, opened 2024-01-23
DEVPROD-983: remove legacy GitHub token fallbacks from app server

Kimchelly:
DEVPROD-983

### Description
This removes the logic for falling back to the legacy token to send GitHub messages from the app server. Based on Splunk, I have reasonable confidence that **this is not risky**.

It _doesn't_ remove the fallback where the agent can still use the legacy GitHub token to run git clones for tasks. I noticed that there's a fair number of networking errors in the agent REST routes when getting the app token in preparation to run tasks. Since I have lower confidence in that change, I'm breaking the riskier part into a separate follow-up PR.

* Remove fallbacks to use the legacy GitHub token in the app server.
* Fix a test that relies awkwardly on admin settings from three different places (the settings override YAML file, settings from the DB, and the in-memory admin settings). It was only succeeding efore because it relied on the legacy token fallback.
* Fix other test setups that have been relying on GitHub app token creation continuing on error.

### Testing
* Existing unit tests pass.
* Tested in [a staging patch](https://github.com/evergreen-ci/commit-queue-sandbox/pull/655) that the app servers still can send GitHub statuses.
* Checked Splunk logs to verify that app servers haven't had to use the fallback legacy token in the past 7 days.

### Documentation
N/A

### DIFF
--- a/scripts/setup-credentials.sh
+++ b/scripts/setup-credentials.sh
@@ -73,14 +73,6 @@ auth:
 
 github_pr_creator_org: "10gen"
 
-# Do not edit below this line
 expansions:
-  github_app_key: |
+  github_app_key: "$GITHUB_APP_KEY"
 EOF
-
-# Write the GitHub app key to a file for easier formatting
-echo "$GITHUB_APP_KEY" > app_key.txt
-# Linux and MacOS friendly command to add 4 spaces to the start of each line
-sed -i'' -e 's/^/    /' app_key.txt
-# Append the formatted GitHub app key to the creds.yml file
-cat app_key.txt >> creds.yml
--- a/scripts/setup-credentials.sh
+++ b/scripts/setup-credentials.sh
@@ -73,6 +73,14 @@ auth:
 
 github_pr_creator_org: "10gen"
 
+# Do not edit below this line
 expansions:
-  github_app_key: "$GITHUB_APP_KEY"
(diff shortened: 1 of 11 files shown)


scripts/setup-credentials.sh
```diff
@@ -73,14 +73,6 @@ auth:
...
-  github_app_key: |
+  github_app_key: "$GITHUB_APP_KEY"
 EOF
-
-# Write the GitHub app key to a file for easier formatting
-echo "$GITHUB_APP_KEY" > app_key.txt
-# Linux and MacOS friendly command to add 4 spaces to the start of each line
-sed -i'' -e 's/^/    /' app_key.txt
-# Append the formatted GitHub app key to the creds.yml file
-cat app_key.txt >> creds.yml
```
malikchaya2, line 86: were creds.yml and app_key.txt not being used?

malikchaya2 (approved):
Good call on breaking out the riskier changes! 

Approved with one non-blocking question.

scripts/setup-credentials.sh
Kimchelly: I reverted this back to what it originally was - making a YAML file by hand for a script setup is less than ideal but unless we want to convert it to JSON or something, it's needed due to how YAML handles strings with newlines + caring about leading whitespace syntax.

PR #7444, opened 2024-01-24
DEVPROD-826 Expose field to determine if a task has test results in cedar

khelif96:
DEVPROD-826

### Description
Exposes a field that determines if a task has test results or not by reading a cached value instead of reaching out to cedar.

The changes are separated out by commit. 
You can omit reviewing 188d2b7 since that just contains gqlgen generated code from updating the dependency.


graphql/task_resolver.go
```diff
@@ -679,3 +679,13 @@ func (r *taskResolver) VersionMetadata(ctx context.Context, obj *restModel.APITa
...
+
+// !!! WARNING !!!
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
+// one last chance to move it out of harms way if you want. There are two reasons this happens:
+//   - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//     it when you're done.
+//   - You have helper methods in this file. Move them out to keep these resolver files clean.
+func (r *taskResolver) HasTestResults(ctx context.Context, obj *restModel.APITask) (bool, error) {
+	panic(fmt.Errorf("not implemented: HasTestResults - hasTestResults"))
+}
```
sophstad, line 691: This should probably not be included

sophstad (changes requested):
Would you mind including in the description which files actually contain changes? 😅

graphql/task_resolver.go
khelif96: > oof good catch.

khelif96:
> Would you mind noting in the description which files actually contain changes? 😅

Updated the description. The changes were separated by commit.

sophstad (approved):

PR #7447, opened 2024-01-24
DEVPROD-983: clean up GitHub app usages in the agent

Kimchelly:
DEVPROD-983

### Description
Follow-up to #7443. I determined that the remaining usages of the fallback to the legacy GitHub token are not that risky, and reverting would be easy if issues arise.

* Remove remaining usages of in the agent REST routes.
* Fix tests that were relying on the fallback to the legacy GitHub token to pass.

### Testing
Existing tests pass.

### Documentation
N/A


auth/auth.go
```diff
@@ -42,7 +42,7 @@ func LoadUserManager(settings *evergreen.Settings) (gimlet.UserManager, evergree
 	if authConfig.Naive != nil {
 		return makeNaiveManager(authConfig.Naive)
 	}
-	if authConfig.Github != nil {
+	if authConfig.Github != nil && authConfig.Github.ClientId != "" && authConfig.Github.ClientSecret != "" && authConfig.Github.Organization != "" && len(authConfig.Github.Users) > 0 {
```
hadjri, line 45: Can we put these checks into a `Validate` command so we can get a more descriptive error than what we'd end up with [here](https://github.com/evergreen-ci/evergreen/pull/7447/files#diff-f5c9a39fac1719f2a73593994b931d8a874449548e8a3753e1be46e761e503eaR51)?

hadjri (changes requested):

auth/auth.go
Kimchelly: We could I guess, but I really just slapped on this fix to make the smoke test stop erroring (because when the smoke app server starts up, the GitHub app fields are filled in, but not the other stuff). We should just get rid of this logic someday, this is just legacy auth left over from when MongoDB used to change auth systems a lot before it settled on Okta.

hadjri (approved):

auth/auth.go
hadjri: Alright SGTM

Kimchelly: I made DEVPROD-4088 in case we ever want to get rid of our legacy auth.

PR #7452, opened 2024-01-25
DEVPROD-3085 Check for renamed includes when creating diff patch

hadjri:
DEVPROD-3085

### Description
Currently if you try to submit a CLI patch that changes the name of an included config file, the patch will fail to be created because Evergreen is trying to apply a patch to a file that does not yet exist remotely (because it was just renamed to something else).

Modified the `MakePatchedConfig` logic to detect if the patch includes a rename for the remote file in question, and if so, retrieve the old file name remote config, and apply the patch to that instead.
### Testing
Added unit tests and test files.

Tested in staging and confirmed that without this change, if I submit a patch with include file name changes I get the following error: `retrieving file for include 'renamed.yml': patching remote configuration file: running patch command (possibly due to merge conflict on evergreen configuration file):`

Confirmed that after the change the patch goes through with the correct diff, and that the /tmp/evergreen directory on the staging host is clean ([link](https://spruce-staging.corp.mongodb.com/patch/65b2a53d97b1d3630c3f9cc9/configure/changes))

### DIFF
--- a/model/patch_lifecycle.go
+++ b/model/patch_lifecycle.go
@@ -238,7 +238,7 @@ func getPatchedProjectYAML(ctx context.Context, projectRef *ProjectRef, opts *Ge
 	// apply remote configuration patch if needed
 	if p.ShouldPatchFileWithDiff(path) {
 		opts.ReadFileFrom = ReadFromPatchDiff
-		projectFileBytes, err = MakePatchedConfig(ctx, *opts, opts.PatchOpts.env, p, opts.RemotePath, string(projectFileBytes))
+		projectFileBytes, err = MakePatchedConfig(ctx, *opts, string(projectFileBytes))
 		if err != nil {
 			return nil, errors.Wrap(err, "patching remote configuration file")
 		}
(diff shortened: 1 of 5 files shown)


malikchaya2 (approved):
These changes look good. However, it might be a good idea to also test what happens if a new file is created as opposed to an existing file being renamed. Can it find that file?

hadjri:
> These changes look good. However, it might be a good idea to also test what happens if a new file is created as opposed to an existing file being renamed. Can it find that file?

Confirmed it [does work](https://spruce-staging.corp.mongodb.com/patch/65b8198e9dbe3288d7de51f2/configure/changes) with an added include file as well

malikchaya2:
> > These changes look good. However, it might be a good idea to also test what happens if a new file is created as opposed to an existing file being renamed. Can it find that file?
> 
> Confirmed it [does work](https://spruce-staging.corp.mongodb.com/patch/65b8198e9dbe3288d7de51f2/configure/changes) with an added include file as well

awesome, thank you!

PR #7453, opened 2024-01-25
DEVPROD-4025 project variable redaction

ybrill:
[DEVPROD-4025](https://jira.mongodb.org/browse/DEVPROD-4025)

### Description
Redact sensitive default variables, private project variables, and suspicious (non-private) project variables.

### Testing
On [staging](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_echo_variable_patch_f50345d4c091635680061fecb7f18af636a6bf08_65b7fec097b1d3897105fabe_24_01_29_19_39_01/logs?execution=0) the logs had a redacted value for `jonathans_super_secret_variable` which is private and `julian_password` which is not private but has a suspicious pattern.

[Successful race detector patch](https://spruce.mongodb.com/version/65b81f9cd1fe07496530c99a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
### Documentation
Added a line to the expansions section of the configuration file doc.


ybrill:
cc @bsamek since you'd mentioned you might want to look over the PR.
I ended up going with strings.Replacer because the interface seemed more comfortable and maybe it's more efficient 🤷 . I started benchmarking but then thought getting this out the door was a higher priority.

bsamek (approved):
Thanks! LGTM, but @julianedwards should also review, since I'm not familiar with the logging subsystem.

agent/internal/client/interface.go
```diff
@@ -140,6 +140,7 @@ type LoggerConfig struct {
 	Task               []LogOpts
 	SendToGlobalSender bool
 	AWSCredentials     *credentials.Credentials
+	ProjectVars        map[string]string
```
julianedwards, line 143: Should we also pipe through these [global expansions](https://github.com/evergreen-ci/evergreen/blob/main/agent/command/expansion_write.go#L17-L23) here?

julianedwards (changes requested):

agent/internal/client/interface.go
julianedwards: There is also `${__project_aws_ssh_key_value}` in the [list of default expansions](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#default-expansions).

agent/internal/client/redacting_sender.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/mongodb/grip/message"
+	"github.com/mongodb/grip/send"
+)
+
+const redactedVariableTemplate = "<redacted:%s>"
```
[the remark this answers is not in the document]
ybrill: done

agent/internal/client/interface.go
ybrill: Ah, so I know we talked about this yesterday. I should have left a comment explaining my rationale.

Of those expansions, only two (`GlobalGitHubTokenExpansion` and `GithubAppToken`) are populated as default expansions ([here](https://github.com/evergreen-ci/evergreen/blob/ca7e5d6fa44354c8fcf9b878545f3b8dab972aa9/model/project.go#L1133-L1134)). The others are populated by the ec2.assume_role command ([here](https://github.com/evergreen-ci/evergreen/blob/4dcce0dda1269f98a6fbde97aed31d12666e4b37/agent/command/assume_ec2_role.go#L123-L125)).

With regard to the github tokens, I think they're protected from being leaked in logs because they're not exposed as expansions to `shell.exec` or `subprocess.exec` ([here](https://github.com/evergreen-ci/evergreen/blob/66b48dce1b98cbc6d0dee8ce4c8523ea7f5e8a96/agent/command/exec.go#L184-L188)).

With regard to the AWS creds, they are temporary creds that expire after a few minutes, so it seemed okay to leave them. Though we could redact them too if we wanted to.

ybrill: Ah, I didn't read your whole comment.
If you feel like we shouldn't leak them I'll redact them.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -648,6 +648,9 @@ level on the project configuration page or on a build variant level
 within the project. They can be used **as inputs to commands**,
 including shell scripts.
 
+Expansion values defined on the project configurations page are redacted from task logs
+and replaced with \<redacted:expansion_key\>
```
julianedwards, line 652: This needs to be updated with `<REDACTED:expansion_key>`

agent/internal/client/redacting_sender_test.go
```diff
@@ -0,0 +1,58 @@
...
+		substitutions map[string]string
+		inputString   string
+		expected      string
+	}{
+		"MultipleSubstitutions": {
+			substitutions: map[string]string{
+				"secret_key": "secret_val",
+			},
+			inputString: "secret_val secret_val",
+			expected:    "<redacted:secret_key> <redacted:secret_key>",
```
julianedwards, line 24: These tests need to be updated, maybe use `fmt.Sprintf(redactedVariableTemplate, <val>)` instead of hardcoding it?

julianedwards (changes requested):

agent/internal/client/interface.go
julianedwards: >Of those expansions, only two (GlobalGitHubTokenExpansion and GithubAppToken) are populated as default expansions ([here](https://github.com/evergreen-ci/evergreen/blob/ca7e5d6fa44354c8fcf9b878545f3b8dab972aa9/model/project.go#L1133-L1134)). The others are populated by the ec2.assume_role command ([here](https://github.com/evergreen-ci/evergreen/blob/4dcce0dda1269f98a6fbde97aed31d12666e4b37/agent/command/assume_ec2_role.go#L123-L125)).

The github token expansions are only protected from getting set as env vars, but they can still get expanded in the shell script [here](https://github.com/evergreen-ci/evergreen/blob/main/agent/command/shell.go#L253-L254).

We will have to figure out how to add the AWS expansions to the string replacer since they are set dynamically, that can be in follow up work.

ybrill: These are good points.
I've changed the strategy a bit and I think it should now cover all these.

afaict `__project_aws_ssh_key_value` is a project variable so I didn't include it.

docs/Project-Configuration/Project-Configuration-Files.md
ybrill: done

agent/internal/client/redacting_sender_test.go
ybrill: done

ybrill:
Still [works](https://parsley-staging.corp.mongodb.com/evergreen/evg_ubuntu2204_echo_variable_patch_4225ee2a49a89afcd2b3deec7bb1c68c9810262b_65b27e5c97b1d34e31fe84ae_24_01_25_15_29_54/1/task?bookmarks=0,25&shareLine=4) with the new strategy.

julianedwards (approved):
LGTM % race and broken tests

joe-dipilato:
I think our log redaction might be a bit over eager.

See: https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_echo_variable_patch_4225ee2a49a89afcd2b3deec7bb1c68c9810262b_65b27e5c97b1d34e31fe84ae_24_01_25_15_29_54/logs?execution=0&logtype=all

```cpp
[2024/01/25 15:30:30.067] Filesystem     Inodes IUsed IFree IUse% Mounted on
[2024/01/25 15:30:30.067] /dev/root        7.9M  5<redacted:num_hosts>6K  7.<redacted:num_hosts>M    7% /
[2024/01/25 15:30:30.067] tmpfs            2.0M     2  2.0M    1% /dev/shm
[2024/01/25 15:30:30.067] tmpfs            800K   653  800K    1% /run
[2024/01/25 15:30:30.067] tmpfs            2.0M     3  2.0M    1% /run/lock
[2024/01/25 15:30:30.067] /dev/xvda15         0     0     0     - /boot/efi
[2024/01/25 15:30:30.067] /dev/xvdb        150M    <redacted:num_hosts>6  150M    1% /data
[2024/01/25 15:30:30.129] Cleaning up task because task is starting
[2024/01/25 15:30:30.129] Cleaning up processes for task: 'evg_ubuntu220<redacted:num_hosts>_echo_variable_patch_<redacted:num_hosts>225ee2a<redacted:num_hosts>9a89afcd2b3deec7bb1c68c9810262b_65b27e5c97b1d3<redacted:num_hosts>e31fe8<redacted:num_hosts>ae_2<redacted:num_hosts>_01_25_15_29_5<redacted:num_hosts>'.
[2024/01/25 15:30:30.142] Cleaned up processes for task: 'evg_ubuntu220<redacted:num_hosts>_echo_variable_patch_<redacted:num_hosts>225ee2a<redacted:num_hosts>9a89afcd2b3deec7bb1c68c9810262b_65b27e5c97b1d3<redacted:num_hosts>e31fe8<redacted:num_hosts>ae_2<redacted:num_hosts>_01_25_15_29_5<redacted:num_hosts>'.
```

julianedwards:
> I think our log redaction might be a bit over eager.

You are right. If someone has a project variable like `max_attempts=12` any number string `12` in the log output will be replaced with `<REDACTED:max_attempts>`. I think we should just redact private project variables and tell users they need to mark secrets as private.

agent/logging.go
```diff
@@ -167,3 +171,36 @@ func (a *Agent) prepSingleLogger(tc *taskContext, in model.LogOpts, logDir, file
...
+		"private",
+		"pass",
+		"pw",
+	}
+
+	var redactedKeys []string
+	for key := range projectVars {
+		if _, ok := redacted[key]; ok {
+			// Skip since the key is already in the redacted list.
+			continue
```
ybrill, line 190: could it work to not `continue` here and to omit L200-202?

agent/command/assume_ec2_role.go
```diff
@@ -120,9 +120,9 @@ func (r *ec2AssumeRole) Execute(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
-	conf.Expansions.Put(AWSAccessKeyId, credValues.AccessKeyID)
-	conf.Expansions.Put(AWSSecretAccessKey, credValues.SecretAccessKey)
-	conf.Expansions.Put(AWSSessionToken, credValues.SessionToken)
-	conf.Expansions.Put(AWSRoleExpiration, expTime.String())
+	conf.NewExpansions.Put(AWSAccessKeyId, credValues.AccessKeyID)
```
ybrill, line 123: do we need to `Put` into both?

julianedwards: No because it is wraps the same underlying map so doing it once is ok. I think eventually we would want to move everything in the agent to just use the the new `DynamicExpansions`.

agent/logging.go
julianedwards: There is no guarantee that a redacted project variable will have a suspicious pattern so we definitely want the full list of those and then any additional project vars that are not private but have a suspicious pattern.

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -647,6 +647,11 @@ level on the project configuration page or on a build variant level
 within the project. They can be used **as inputs to commands**,
 including shell scripts.
 
+Expansion values defined on the project configurations page that
+    (1) are marked as private OR
+    (2) with key names containing any of the following case-insensitive patterns: `key`, `pass`, `pw`, `token`, `auth`, `private`
```
malikchaya2, line 652: Should we add a blurb about how this shouldn't be relied on and it should be thought of as a last line of defense?

agent/logging.go
```diff
@@ -167,3 +171,36 @@ func (a *Agent) prepSingleLogger(tc *taskContext, in model.LogOpts, logDir, file
...
+		}
+
+		for _, pattern := range suspiciousPatterns {
+			if strings.Contains(strings.ToLower(key), pattern) {
```
malikchaya2, line 194: Why don't we also check the values for suspicious patterns? For example, we can look for things that are often at the start of certain keys like `ghs_` for github.

agent/logging.go
```diff
@@ -116,6 +118,8 @@ func (a *Agent) prepLogger(tc *taskContext, c *model.LoggerConfig, commandName s
 		SendToGlobalSender: a.opts.SendTaskLogsToGlobalSender,
 		AWSCredentials:     pail.CreateAWSCredentials(tc.taskConfig.TaskSync.Key, tc.taskConfig.TaskSync.Secret, ""),
 	}
+	config.Expansions = tc.taskConfig.NewExpansions
+	config.ExpansionsToRedact = redactList(tc.taskConfig.ProjectVars, tc.taskConfig.Redacted)
```
malikchaya2, line 122: We should probably also check during expansions.write and expansions.update if the new expansion needs to be redacted?

malikchaya2 (changes requested):
Would it be possible to add a test to check if old values are maintained in case they get logged at some later time after the expansion was updated? 

For example: When `token : apple` gets updated to `token : banana`, do both `apple` and `banana` get redacted?

julianedwards:
> Would it be possible to add a test to check if old values are maintained in case they get logged at some later time after the expansion was updated?

I will leave that up to the app team to implement since it will require some thought around how to further extend the expansions class. This is really a first pass to introduce redaction since some is better than none.

agent/logging.go
julianedwards: Expansions to redact are already checked in `expansions.write`.

I left the work to redact new private expansions for [DEVPROD-4107](https://jira.mongodb.org/browse/DEVPROD-4107); I imagine that it will be much cleaner to implement that functionality together.

agent/logging.go
julianedwards: To get this out and deployed I think we should just start with this set of patterns and expand as we see fit.

docs/Project-Configuration/Project-Configuration-Files.md
julianedwards: Good idea! Done.

agent/logging.go
malikchaya2: My suggestion is not just to expand the set of patterns, but to scan both keys and values. If getting this out quickly is the priority, would it maybe make sense to do this in a separate ticket?

malikchaya2:
> > Would it be possible to add a test to check if old values are maintained in case they get logged at some later time after the expansion was updated?
> 
> I will leave that up to the app team to implement since it will require some thought around how to further extend the expansions class. This is really a first pass to introduce redaction since some is better than none.

That makes sense. Can you please create a ticket for the follow up work?

malikchaya2 (approved):
Approved after the two comments are addressed.

agent/logging.go
julianedwards: Yep, that can be done here as well. We can add more scanning but the hope is that users will also rotate their secrets and mark their sensitive vars private.

julianedwards:
> That makes sense. Can you please create a ticket for the follow up work? 

[DEVPROD-4025](https://jira.mongodb.org/browse/DEVPROD-4186)

PR #7456, opened 2024-01-25
DEVPROD-2961: remove support for Data-Pipes

julianedwards:
DEVPROD-2961

### Description
We are decommissioning the Data Pipes service. Must deployed after https://github.com/evergreen-ci/evergreen/pull/7461.

### Testing
Tested the admin UI in staging.

### Documentation
N/A


agent/command/perf_send.go
```diff
@@ -68,26 +68,17 @@ func (c *perfSend) Execute(ctx context.Context, comm client.Communicator, logger
 	}
 	c.addEvgData(report, conf)
 
-	// Send data to the Cedar and Data-Pipes services.
+	// Send data to the Cedar service.
 	conn, err := comm.GetCedarGRPCConn(ctx)
 	if err != nil {
 		return errors.Wrap(err, "connecting to Cedar")
 	}
-	dataPipes, err := comm.GetDataPipesConfig(ctx)
```
ybrill, line 76: Is it a concern that an agent that began running a task before the deploy would get a 404 here from an app server deployed to the new code? And if so could it make sense to first remove this call from the agent and then to remove the route in a later commit?

julianedwards: Yeah that is a good call. I will split this into two separate PRs.

ybrill (approved):

PR #7462, opened 2024-01-29
DEVPROD-3861 distinct amboy connection string for Kanopy

ybrill:
[DEVPROD-3861](https://jira.mongodb.org/browse/DEVPROD-3861)

### Description
Add another URL for connecting to amboy when the app is running inside Kanopy. This is necessary because it's a different private link and thus a different URL.

### Testing
It allowed us to set another URL for the instance of evergreen that was running in Kanopy staging.

### DIFF
--- a/config_amboy.go
+++ b/config_amboy.go
@@ -40,8 +40,9 @@ type AmboyConfig struct {
 
 // AmboyDBConfig configures Amboy's database connection.
 type AmboyDBConfig struct {
-	URL      string `bson:"url" json:"url" yaml:"url"`
-	Database string `bson:"database" json:"database" yaml:"database"`
+	URL       string `bson:"url" json:"url" yaml:"url"`
+	KanopyURL string `bson:"kanopy_url" json:"kanopy_url" yaml:"kanopy_url"`
+	Database  string `bson:"database" json:"database" yaml:"database"`
 }
 
 // AmboyRetryConfig represents configuration settings for Amboy's retryability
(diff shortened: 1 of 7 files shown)


config_amboy.go
```diff
@@ -40,8 +40,10 @@ type AmboyConfig struct {
 
 // AmboyDBConfig configures Amboy's database connection.
 type AmboyDBConfig struct {
-	URL      string `bson:"url" json:"url" yaml:"url"`
-	Database string `bson:"database" json:"database" yaml:"database"`
+	URL string `bson:"url" json:"url" yaml:"url"`
+	// TODO (DEVPROD-4163): remove KanopyURL after the cutover to Kanopy.
+	KanopyURL string `bson:"kanopy_url" json:"kanopy_url" yaml:"kanopy_url"`
```
julianedwards, line 45: I think you need to set this field below in `Set` (or else it won't get set from the admin page).

julianedwards (changes requested):

config_amboy.go
ybrill: It seems that [this field](https://github.com/evergreen-ci/evergreen/blob/6de90900d38382fee2cda1f1d2ba8cfd4adceb1b/config_amboy.go#L112) sets the value to whatever is in the `DBConnection` struct.
It worked for me in local-evergreen, fwiw.

julianedwards: Ah I see, that makes sense.

julianedwards (approved):

PR #7463, opened 2024-01-30
DEVPROD-3888 Use custom context for GitHub webhook route

malikchaya2:
DEVPROD-3888

### Description
We ran into some issues that seemed to be caused by the request cancelling which most likely happened due to a client-side timeout. 

This will ignore the request context passed into Run for the GitHub webhook route and instead use a custom context with a 60 second timeout. That will ensure that webhook requests are not affected by the original webhook request context cancelling.

### DIFF
--- a/rest/route/github.go
+++ b/rest/route/github.go
@@ -44,6 +44,9 @@ const (
 	// skipCIDescriptionCharLimit is the maximum number of characters that will
 	// be scanned in the PR description for a skip CI label.
 	skipCIDescriptionCharLimit = 100
+
+	// githubWebhookTimeout is the maximum timeout for processing a GitHub webhook.
+	githubWebhookTimeout = 60 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


rest/route/github.go
```diff
@@ -118,6 +121,9 @@ func (gh *githubHookApi) shouldSkipWebhook(ctx context.Context, owner, repo stri
 }
 
 func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
+	ctx, cancel := context.WithTimeout(context.Background(), githubWebhookTimeout)
```
Kimchelly, line 124: Can we add a comment for future readers about the rationale for ignoring the request timeout (i.e. GitHub occasionally aborts requests early for no obvious reason)?

Kimchelly (approved):
LGTM mod clarifying comment.

PR #7464, opened 2024-01-30
DEVPROD-3944: create job to collect unexpirable host stats

Kimchelly:
DEVPROD-3944

### Description
Create a daily job that will collect estimated uptime stats on unexpirable hosts. This will help diff the cost savings before/after the sleep schedule is rolled out. It intentionally logs to Splunk rather than Honeycomb because Honeycomb's 60 day TTL is too short of a time period.

### Testing
* Added unit tests.
* Tested in staging that it outputted the stats in the expected format.

### Documentation
N/A

### DIFF
--- a/units/crons_remote_hour.go
+++ b/units/crons_remote_hour.go
@@ -57,6 +57,7 @@ func (j *cronsRemoteHourJob) Run(ctx context.Context) {
 		PopulateSSHKeyUpdates(j.env),
 		PopulateDuplicateTaskCheckJobs(),
 		PopulatePodResourceCleanupJobs(),
+		PopulateUnexpirableSpawnHostStatsJob(),
 	}
 
 	queue := j.env.RemoteQueue()
--- a/units/unexpirable_spawnhost_stats.go
+++ b/units/unexpirable_spawnhost_stats.go
@@ -1,0 +1,120 @@
+package units
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/evergreen-ci/evergreen"
+	"github.com/evergreen-ci/evergreen/model/host"
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
+	"github.com/pkg/errors"
+)
+
+const (
+	unexpirableSpawnHostStatsJobName        = "unexpirable-spawnhost-stats"
+	unexpirableSpawnHostStatsJobMaxAttempts = 10
+)
+
+func init() {
+	registry.AddJobType(unexpirableSpawnHostStatsJobName, func() amboy.Job {
+		return makeUnexpirableSpawnHostStatsJob()
+	})
+}
+
+type unexpirableSpawnHostStatsJob struct {
+	job.Base `bson:"metadata" json:"metadata" yaml:"metadata"`
+}
+
+func makeUnexpirableSpawnHostStatsJob() *unexpirableSpawnHostStatsJob {
+	j := &unexpirableSpawnHostStatsJob{
+		Base: job.Base{
+			JobType: amboy.JobType{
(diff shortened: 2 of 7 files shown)


units/crons_remote_hour.go
```diff
@@ -57,6 +57,7 @@ func (j *cronsRemoteHourJob) Run(ctx context.Context) {
 		PopulateSSHKeyUpdates(j.env),
 		PopulateDuplicateTaskCheckJobs(),
 		PopulatePodResourceCleanupJobs(),
+		PopulateUnexpirableSpawnHostStatsJob(),
```
ablack12, line 60: It doesn't really seem like this needs to run more than daily (similar to [DEVPROD-733](https://jira.mongodb.org/browse/DEVPROD-733) i think a bunch of these really don't need to run so much). If it's not too much trouble I'd make a remote_daily, or at least comment this on DEVPROD-733 to be moved, unless we really need it hourly.

units/unexpirable_spawnhost_stats.go
```diff
@@ -0,0 +1,115 @@
...
+// host uptime. For example, it's assumed for simplicity that users don't stop
+// their hosts throughout the day, so if the host is on, it's likely been on the
+// entire day.
+func (j *unexpirableSpawnHostStatsJob) getStats(hosts []host.Host) unexpirableSpawnHostStats {
+	var totalUptime time.Duration
+	uptimeByDistro := map[string]int{}
+	uptimeByInstanceType := map[string]int{}
+
+	for _, h := range hosts {
+		// Estimate that if the host is up now, it's been up all day.
```
ablack12, line 97: This won't be true if we're doing sleep schedules, right? I might be missing something here

units/unexpirable_spawnhost_stats.go
```diff
@@ -0,0 +1,115 @@
...
+
+	hosts, err := host.FindUnexpirableRunning()
+	if err != nil {
+		j.AddRetryableError(errors.Wrap(err, "finding unexpirable running hosts"))
+		return
+	}
+
+	stats := j.getStats(hosts)
+
+	grip.Info(message.Fields{
```
ablack12, line 71: In the ticket we also mentioned logging how many hosts are stopped, is that no longer something we want to do?

units/crons_remote_hour.go
Kimchelly: Being in the hourly cron doesn't necessarily mean that they but they _run_ every hour but that they try to enqueue a job every hour. The timestamp that they use for rounding the job ID determines how often the job actually runs because it produces a conflict on the job `_id` if it's the same time period. If the interval only enqueued once per day (which is what a daily job would do), it may produce some issues because if you _need_ it to run every day, a daily job will try to enqueue at most once per day, so if it fails once, you're done and there's no hope to run it that day.

We could tune it to be once every few hours, but honestly, I think it's fine as-is. The timestamp rounding already ensures that the jobs actually run at the correct cadence, and trying to enqueue a few extra jobs per hour (that won't run) produces negligibly low on the DB.

units/unexpirable_spawnhost_stats.go
Kimchelly: Sleep schedules don't actually exist right now, but yeah, this comment is only relevant for the moment.

units/unexpirable_spawnhost_stats.go
Kimchelly: So I did think about that, but for the sleep schedule, it would probably matter more to distinguish why the host is stopped (whether it's for the nightly sleep schedule vs. whether it's off 24/7). For now, it only stops if the user manually stops it, so there's not much point in just logging the number of stopped hosts until the sleep schedule actually exists.

The downtime isn't greatly important. The uptime is the more important measure just because we need a relative comparison of current uptime vs. uptime after savings measures.

Kimchelly:
Re the discussion of using the user script, I changed this to run hourly instead of daily because stops/starts initiated by a script will be more irregular than a schedule.

ablack12 (approved):

PR #7465, opened 2024-01-30
DEVPROD-3532: Show admin-only distros to admins only

sophstad:
DEVPROD-3532

### Description
- When fetching list of distros, omit admin-only distros if the user is not an admin

### Testing
- Ensure that admin-only distro is returned to a user who has create permissions. It's difficult to test the opposite since our GraphQL tests are run with a superuser, but I was able to verify manually.

### Documentation
N/A


graphql/query_resolver.go
```diff
@@ -167,7 +169,20 @@ func (r *queryResolver) Distros(ctx context.Context, onlySpawnable bool) ([]*res
 		}
 		distros = d
 	}
+
+	userHasDistroCreatePermissions := usr.HasPermission(gimlet.PermissionOpts{
```
ablack12, line 173: Changes look good -- I might recommend moving the bulk of this logic into a function in the distro collection so that you can write a unit test for the "not admin only" case but i do recognize this isn't a big change

ablack12 (approved):

minnakt (approved):

PR #7466, opened 2024-01-30
DEVPROD-806: Remove CloneMethod field from GraphQL

minnakt:
DEVPROD-806

### Description
This PR removes the `CloneMethod` field from GraphQL. It also removes the validation for clone method, which is necessary because we call the validation functions in the `SaveDistro` mutation.

### Testing
- Removed cloneMethod from GraphQL tests


khelif96 (approved):
LGTM!

hadjri (approved):

PR #7468, opened 2024-01-30
DEVPROD-4214 Fix nil settings for definiton creation job

hadjri:
DEVPROD-4214

### Description
Container tasks that operate on new (i.e.non-cached) configurations are currently unable to register new task definitions with AWS due to our amboy job not having its settings populated. Followup from #7361 

### Testing
Confirmed the job works again in staging


Kimchelly (approved):

Kimchelly:
LGTM mod tests passing.

PR #7469, opened 2024-01-31
DEVPROD-4025: Configurable project var redactions

julianedwards:
DEVPROD-4025

### Description
- Add a new logger config field `RedactKeys` to hold the patterns we want to look for for key redaction
- Send redact keys with expansions and vars in agent route
- Add any matching project var keys to the private vars list when setting up the task config
- Redact all private vars and selected default vars from logs

### Testing
Updated unit tests / [tested in staging](https://spruce-staging.corp.mongodb.com/task/evg_ubuntu2204_echo_variable_patch_f50345d4c091635680061fecb7f18af636a6bf08_65b7fec097b1d3897105fabe_24_01_29_19_39_01/logs?execution=2).

### Documentation
N/A

### DIFF
--- a/agent/logging.go
+++ b/agent/logging.go
@@ -118,7 +118,7 @@ func (a *Agent) prepLogger(tc *taskContext, c *model.LoggerConfig, commandName s
 		AWSCredentials:     pail.CreateAWSCredentials(tc.taskConfig.TaskSync.Key, tc.taskConfig.TaskSync.Secret, ""),
 	}
 	config.Expansions = tc.taskConfig.NewExpansions
-	config.ExpansionsToRedact = redactList(tc.taskConfig.Redacted)
+	config.ExpansionsToRedact = getExpansionsToRedact(tc.taskConfig.Redacted)
 
 	defaultLogger := tc.taskConfig.ProjectRef.DefaultLogger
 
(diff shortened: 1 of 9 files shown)


rest/route/agent.go
```diff
@@ -463,6 +463,7 @@ func (h *getExpansionsAndVarsHandler) Run(ctx context.Context) gimlet.Responder
 		Parameters:  map[string]string{},
 		Vars:        map[string]string{},
 		PrivateVars: map[string]bool{},
+		RedactKeys:  h.settings.LoggerConfig.RedactKeys,
```
julianedwards, line 466: We will have to bounce the app servers if we update this list since this is using the cached settings. I think that is fine, but just wanted to point it out.

agent/logging.go
```diff
@@ -172,31 +171,8 @@ func (a *Agent) prepSingleLogger(tc *taskContext, in model.LogOpts, logDir, file
...
-		"key",
-		"secret",
-		"auth",
-		"token",
-		"private",
-		"pass",
-		"pw",
-	}
-
+func redactList(redacted map[string]bool) []string {
```
ablack12, line 174: tiny nit: redactList is a bit confusing since redact is a verb, but we aren't actually doing a redaction here, we're just getting the list of redacted keys (this took me a minute to figure out). Could we rename to getAllRedactedKeys or something?

ablack12: and/or throw a comment on

ablack12 (approved):

agent/logging.go
julianedwards: Done

PR #7477, opened 2024-02-01
DEVPROD-2720: Process HasMatchingTasks aggregation concurrently

minnakt:
DEVPROD-2720

### Description
Right now, the MainlineCommits query loops through every version and calls an expensive aggregation `HasMatchingTasks` on each version sequentially. The goal of this PR is to make it so that we can call `HasMatchingTasks` on multiple versions at the same time and process them concurrently.

In this PR, I created a worker pool of maximum 20 workers that calls `HasMatchingTasks` on the versions.

I ran [this query](https://evergreen-staging.corp.mongodb.com/graphql?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABALICGAlkgDZUIDCEccFKAzgBRFEA6S3AEjiUadRs1ZsA8gAcUFCEjbpSI2kgZMW7WfMVsAhABpufQQCMYFamABqZPBTKpdCpSoBCVm-cfOUrvpGZkQCltZ2Dk4ucm5sAGIQeADieGQyABae3pF%2BMXpKxiFhOb7RAbH6iXgAKmRsANYAklD62RFl-oGFwfyh4T5RXZVK1akQMDIIYHWNykReHUP5cQYhAJREwCHCVOqaEuwcECPzQmpiWpLdbCYDueU3KiVLeRUFbJvbfdwaAB4oAAKZAA5ggpHgwPgAHLwcz4ELcGR4BAAN2BYIhULwsLg8LwiKIqPwbDiW0J3DwEGo1GmAFUZLYSWTvtw2WyyDAUBkkhS2VAUWQUAgahREHzuBQwBKiBQQUgktMZYg2GxQQgZUlsTKUaiKKTFDLJmwUIK4ICqQArBBQFDkn7stkQADuGgJDsdRGREGttpllIQMgg-qIuv1bhDKHqDXtns9CD%2BNq5EY9calIYAvpHHCCwXgmgARbNyvM1AhTEPEvAG-isuPs9Op9lZptEFue9uOqs12Nxznc3mt%2B6dVCcE4fZ7D5bvOLVWbNVpKL4hsD6mTUMgEaFkcWt7hRua9%2BvcVdsdeb7e749shNJgoh7gZer0aYOABKCDYMGo7Afsule5siaQowGwf7yIgdQNMgmaVtOMqdp6U5vAAylGRzjnEk6lNONxjFSkzTPOnxHnGp7nluO4aoBwEoKBjAwKOpH1q0jEoH%2BtGgbBgGovBraIY6AoIEKIpitR9YgqwdQgmwzGOjIoEZAigFRiCCEyo29YqmqYKapCyn1t6vooE0UKoBQABmFAGXGYY1jKB4NGhIFsM5GGnNhryPKc%2BHpBky6AaxTF1teQXsYB3CcWBgECfGUbqa2xqmsJ5pWjadohXGLpuiGRnpSGKJBgVaLhoaKnRnJ8aJrA94Rf%2B3HXqaJb4IWxa5vgZYVjxzKKJVjqacesVskNAmdi2GYgBmQA) for 50 versions on the Apollo GraphQL sandbox and compared the results:
- [Honeycomb trace ](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/sJYjYYp5RYF/trace/sDpYbcZzJjc?fields[]=s_name&fields[]=s_serviceName&span=d1d68dd63108fb83) for the MainlineCommits query without concurrency
- [Honeycomb trace](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/pKuJEGmKhev/trace/t8juVavRKfP?fields[]=s_name&fields[]=s_serviceName&span=676096afdb59114e) for the MainlineCommits query with concurrency

I think you can confirm by looking at the bar lengths and placements that the aggregations are running sequentially vs concurrently (but I'm not really sure about this part).

### Testing
- Existing tests should pass


minnakt:
@bsamek This is ready for review, just keeping it as a draft so other people know not to look at it yet! The function of interest is called `concurrentlyBuildHasMatchingTasksMap`.

graphql/util.go
```diff
@@ -1163,3 +1165,65 @@ func interfaceToMap(ctx context.Context, data interface{}) (map[string]interface
...
+				}
+				output <- HasMatchingTasks{Id: i.Id, HasTasks: hasTasks}
+			}
+		}()
+	}
+	wg.Wait()
+	close(output)
+
+	if catcher.HasErrors() {
+		return werrors.Wrap(catcher.Resolve(), "finding matching tasks")
```
bsamek, line 1206: `werrors` is the same as `errors`, so I think this is incorrectly imported twice.

graphql/util.go
```diff
@@ -1163,3 +1165,65 @@ func interfaceToMap(ctx context.Context, data interface{}) (map[string]interface
 
 	return mapField, nil
 }
+
+type HasMatchingTasks struct {
+	Id       string
+	HasTasks bool
+}
+
+func concurrentlyBuildHasMatchingTasksMap(ctx context.Context, versions []model.Version, opts task.HasMatchingTasksOptions, hasMatchingTasksMap map[string]bool) error {
```
bsamek, line 1174: I would advocate for returning a `map[string]bool` instead of passing one. This way you can return `nil` in the case where you hit any errors. I don't have a particularly strong argument though other than that typically in Go you would expect the results of a function in the return values.

bsamek (changes requested):
The concurrent logic looks good to me. My comments are only nits.

bsamek (approved):
concurrency code lgtm

graphql/util.go
```diff
@@ -1163,3 +1164,67 @@ func interfaceToMap(ctx context.Context, data interface{}) (map[string]interface
 
 	return mapField, nil
 }
+
+type HasMatchingTasks struct {
```
ablack12, line 1168: can this be private? Looks like it's only used in this function.

graphql/util.go
```diff
@@ -1163,3 +1164,67 @@ func interfaceToMap(ctx context.Context, data interface{}) (map[string]interface
...
+	Id       string
+	HasTasks bool
+}
+
+func concurrentlyBuildHasMatchingTasksMap(ctx context.Context, versions []model.Version, opts task.HasMatchingTasksOptions) (map[string]bool, error) {
+	wg := sync.WaitGroup{}
+	input := make(chan model.Version, len(versions))
```
ablack12, line 1175: out of curiosity, why can't we just iterate through versions directly? (May be worth putting a comment so someone doesn't try to clean this up)

graphql/query_resolver.go
```diff
@@ -865,101 +864,85 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error setting version activation status: %s", err.Error()))
-			}
-		}
-		mainlineCommitVersion := MainlineCommitVersion{}
-		shouldCollapse := false
-		if !utility.FromBoolPtr(v.Activated) {
-			shouldCollapse = true
-		} else if isPopulated(buildVariantOptions) && utility.FromBoolPtr(options.ShouldCollapse) {
+
+		var hasMatchingTasksMap map[string]bool
```
ablack12, line 890: I'd recommend renaming to versionHasMatchingTasksMap (or at least commenting -- it's not immediately clear what the subject of the map is, i.e. what "string" corresponds to)

graphql/query_resolver.go
```diff
@@ -865,101 +864,85 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
+		if hasFilters {
+			activeVersions := utility.FilterSlice(versions, func(s model.Version) bool { return utility.FromBoolPtr(s.Activated) })
```
ablack12, line 892: 🚀

graphql/query_resolver.go
ablack12, line 890: Doesn't look like this is used outside of the if statement also so I don't think there's a reason to define this; I think you can just move the function call for concurrentlyBuild out of it's own error if statement so that you can declare directly, as in:
```
hasMatchingTasksMap, err := concurrentlyBuildHasMatchingTasksMap(ctx, activeVersions, opts)
if  err != nil {
}
```

graphql/query_resolver.go
```diff
@@ -865,101 +864,85 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
 				Statuses:                   getValidTaskStatusesFilter(buildVariantOptions.Statuses),
 				IncludeNeverActivatedTasks: true,
 			}
-			hasTasks, err := task.HasMatchingTasks(ctx, v.Id, opts)
-			if err != nil {
-				return nil, InternalServerError.Send(ctx, fmt.Sprintf("Error checking if version has tasks: %s", err.Error()))
-			}
-			if !hasTasks {
-				shouldCollapse = true
+			if hasMatchingTasksMap, err = concurrentlyBuildHasMatchingTasksMap(ctx, activeVersions, opts); err != nil {
```
ablack12, line 899: Do we only call this on active versions because inactive versions will never match?

graphql/query_resolver.go
ablack12: Better yet, it seems like we only need to know what versions _do_ match tasks, not just to have all IDs in the map. I'd suggest just turning this into a versionsMatchingTasksMap (i.e. the concurrentlyBuild function can just not put versions in that have no matching tasks). This is a common usage of "maps as sets"

graphql/query_resolver.go
```diff
@@ -865,101 +864,85 @@ func (r *queryResolver) MainlineCommits(ctx context.Context, options MainlineCom
...
-					lastMainlineCommit.RolledUpVersions = append(lastMainlineCommit.RolledUpVersions, &apiVersion)
-				} else {
-					mainlineCommitVersion.RolledUpVersions = []*restModel.APIVersion{&apiVersion}
-				}
+		// Loop through the current versions to check for matching versions.
+		for _, v := range versions {
+			mainlineCommitVersion := MainlineCommitVersion{}
+			apiVersion := restModel.APIVersion{}
+			apiVersion.BuildFromService(v)
+			versionsCheckedCount += 1
```
ablack12, line 909: nit: ++

ablack12 (changes requested):

graphql/util.go
minnakt: In all of the examples I've seen, it's always been a channel — I think that looping through versions will get the correct result but it will also repeat work (so version A will get processed more than once, which would waste resources). I'll try to add a comment to the best of my understanding HAHA...

bsamek: Looping through the versions directly would mean that every worker processes every version, as opposed to dividing up the work.

graphql/query_resolver.go
minnakt: I think it may have to stay defined outside; I tried moving it inside as described but it threw an error since `hasMatchingTasksMap` is used in the next for-loop block 😔

graphql/util.go
```diff
@@ -1163,3 +1164,67 @@ func interfaceToMap(ctx context.Context, data interface{}) (map[string]interface
...
+
+	// Limit number of parallel requests to the DB.
+	const maxParallel = 20
+	workers := util.Min(maxParallel, len(versions))
+	for i := 0; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for i := range input {
+				hasTasks, err := task.HasMatchingTasks(ctx, i.Id, opts)
```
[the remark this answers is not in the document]
bsamek: Yes, with a context or signaling channel, but I think the complexity isn't worth it here.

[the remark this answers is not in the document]
ablack12: Ah!

minnakt:
The reason why the Spruce tests were failing is because I removed [this code block](https://github.com/evergreen-ci/evergreen/blob/8fe05b3b8f9e14487101ac82e3f308e959b8b2d7/graphql/query_resolver.go#L892-L898) for setting the version's activated field. This field has been baked in for ~2 years so I think it's safe to remove now. To fix the tests, I just added `"activated"` field to our local version data.

graphql/query_resolver.go
minnakt: Do you want me to add an if-statement to only have entries of `[id]: true` in the map? :0 (I can definitely do that I just wasn't sure)

ablack12: yeah exactly, i would suggest just not putting versions in that aren't true, to make this data structure behave more like a set and less like a dictionary

ablack12 (changes requested):
my comment didn't mark me as having reviewed so adding this, would just like to clean up the map logic a bit more -- please reach out if you want to workshop this together!

graphql/query_resolver.go
minnakt: okay tried to clean up the logic here! Lmk what you think!

graphql/query_resolver.go
ablack12, line 899: nit -- preference to not assign values in if statements that you don't intend to use outside of the if statement; I would break this down
```
hasMatchingTasksMap, err = concurrentlyBuildHasMatchingTasksMap(ctx, activeVersions, opts)
if err != nil {  ....
```

ablack12 (approved):
🚀

PR #7478, opened 2024-02-01
DEVPROD-4258 Annotate critical errors in agent with otel spans

bsamek:
DEVPROD-4258

### Description

This annotates all critical errors in the agent with otel spans.

### Testing

N/A

### Documentation

N/A


agent/agent.go
```diff
@@ -340,8 +342,11 @@ func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskRespo
 	}
 
 	if nt.TaskSecret == "" {
+		msg := "task response missing secret"
+		span.SetStatus(codes.Error, msg)
+		span.RecordError(errors.New(msg), trace.WithAttributes(attribute.String("task_id", tc.task.ID)))
```
ybrill, line 347: I've tended to use dot notation for attributes. I've found it enables discovery and helps keep them organized.

agent/agent.go
```diff
@@ -304,6 +304,8 @@ type processNextResponse struct {
 }
 
 func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskResponse, tc *taskContext, needTeardownGroup bool) (processNextResponse, error) {
+	ctx, span := a.tracer.Start(ctx, "process-next-task")
```
ybrill, line 307: If the ctx contains this root span I think that would replace [this span](https://github.com/evergreen-ci/evergreen/blob/e3a0575129fc25e4e627110b547f0f2549d11de2/agent/agent.go#L602) as the root span for a task which would change to how the trace looks for users.

ybrill: Or maybe we could just move that span creation here?

agent/agent.go
bsamek: Done.

agent/agent.go
bsamek: Could you explain what you mean a bit more? From the `Start()` comment it looks like it creates a child span of the span in the `ctx` https://github.com/open-telemetry/opentelemetry-go/blob/279c549cabf6c9b21b074f35d34634d3190f2343/trace/trace.go#L512, so it seemed safe to use.

ybrill: ah sorry.
Like taking [this trace](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/trace/JDFzSHa7BNa?fields[]=s_name&span=18d49e6990f41882), for instance: right now the root span is the `task: 'test-evergreen'` one. If we make a new `process-next-task ` span here it will be a parent to that span and will become the root of the task trace.

bsamek: I made a new context. My thought was that it doesn't make sense to move the span creation here because the task might not actually start, and we could end up with duplicate task information in Honeycomb. We only actually start the task span when we're ready to start the metrics collectors.

bsamek:
Validated that this works in staging https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen-agent/result/nRX589tKBeQ/trace/9aVTVPx2DvT?fields[]=s_name&fields[]=s_serviceName&source=events&span=c4dbaf992cd245c4

agent/agent.go
```diff
@@ -304,13 +304,15 @@ type processNextResponse struct {
 }
 
 func (a *Agent) processNextTask(ctx context.Context, nt *apimodels.NextTaskResponse, tc *taskContext, needTeardownGroup bool) (processNextResponse, error) {
+	processNextTaskCtx, span := a.tracer.Start(ctx, "process-next-task")
```
ybrill, line 307: if we aren't planning to propagate the span down the chain with the ctx we could potentially ignore the returned ctx, like
```go
_, span := a.tracer.Start(ctx, "process-next-task")
```

bsamek: done

PR #7479, opened 2024-02-01
DEVPROD-2712 Use parser project from before generate.tasks for child patches

hadjri:
DEVPROD-2712

### Description
Generate.tasks commands in child patches will fail because the child patch will pull the most recent parser project from the waterfall, which may have already been updated by the mainline generate.tasks command. This causes `cannot redefine tasks` errors on the child patch.

This change:
- Caches a version of the parser project from before it was updated by the generator at the time of updating the parser project, so that a child patch can later reference this cached parser project instead
- If a child patch uses the same repo/project/branch as the parent, use the parent's revision and apply the parent's patch changes
### Testing
Unit tests + confirmed that before this change, running a generate.tasks command in a child patch [will fail](https://spruce-staging.corp.mongodb.com/task/test_trigger_sandbox_git_env_test_my_first_generator_patch_d339e4981591e4e09cf3be33dcf51c8c8124191e_65bd601cb237368ba2ab6441_24_02_02_21_35_24/logs?execution=0) with a `cannot redefine tasks` error, and that post change, the command [succeeds](https://spruce-staging.corp.mongodb.com/version/65bd663f9dbe32bcbcf2964c/downstream-projects?redirect_spruce_users=true).


units/patch_intent.go
```diff
@@ -1076,8 +1090,13 @@ func (j *patchIntentProcessor) buildTriggerPatchDoc(patchDoc *patch.Patch) (*mod
...
 		for _, p := range parentPatch.Patches {
 			if p.ModuleName == "" {
+				moduleName := intent.ParentAsModule
+				if patchDoc.Triggers.SameBranchAsParent {
+					patchDoc.Githash = parentPatch.Githash
+					moduleName = ""
+				}
 				patchDoc.Patches = append(patchDoc.Patches, patch.ModulePatch{
-					ModuleName: intent.ParentAsModule,
+					ModuleName: moduleName,
```
malikchaya2, line 1099: This will add the parent as a module for any child patches that have the same branch, when we were previously not doing that. Do we maybe want to avoid that?

malikchaya2 (changes requested):

model/generate.go
```diff
@@ -184,6 +185,20 @@ func updateParserProject(ctx context.Context, settings *evergreen.Settings, v *V
 		return nil
 	}
 
+	if len(pp.UpdatedByGenerators) == 0 {
```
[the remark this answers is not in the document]
hadjri: Sure, added

units/patch_intent.go
hadjri: The lines above make it such that if the child patch uses the same repo/branch as the parent patch then `moduleName` will be the empty string, meaning those changes will apply to the base project of the patch. I can add a comment making this clearer.

So the parent wont be added as a module, but the parent's patched changes will apply to both the parent and child patches.

malikchaya2: I can't really think of a use case where someone would specifically want a child patch and a parent patch that have the same repo/branch. However, "the parent's patched changes will apply to both the parent and child patches" doesn't feel right, I don't think users expect that to be the case for child patches.

malikchaya2: I can't think of a use case where someone would want a child and parent patch with the same repo/branch. However, "the parent's patched changes will apply to both the parent and child patches" seems like it might break user expectations for child patches. I think users would expect the patch changes to only apply to the parent patch. And for the parent project's changes to be included only if the parent project is set up as a module in the patch trigger alias. 

I can't think of a use case where a parent and child with the same repo/branch would be needed. Are you aware of a use case like that? I wonder if that would indicate an error state or a misconfiguration.

malikchaya2 (changes requested):
waiting on one last comment thread

units/patch_intent.go
hadjri: Sorry to have this thread going on for too long.

> I can't think of a use case where someone would want a child and parent patch with the same repo/branch.

I initially thought the same thing, but this use case is one of the reasons this ticket was created. It turns out mongodb-mongo-master sometimes creates child patches in the `mongodb-mongo-master-nightly` project ([example](https://spruce.mongodb.com/version/65576a9d32f417c7eef6ae38/downstream-projects?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

malikchaya2: Aah very interesting! 
Can you please add a comment to the code explaining this so if people touch that code down the line they know to keep that functionality?

PR #7480, opened 2024-02-01
DEVPROD-4255: reduce noise in Splunk logs

Kimchelly:
DEVPROD-4255

### Description
I pruned some of the biggest contributors to the Splunk log size. I don't think any of these are controversial since they don't drastically change traceability of code, but I addressed each one in a comment.

Post-deploy, will continue monitoring to see if it's still too much logging for Splunk. Will also verify that the Evergreen Splunk dashboard still renders properly.

### Testing
N/A

### Documentation
N/A

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -318,13 +318,6 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 		input.UserData = &userData
 	}
 
-	grip.Debug(message.Fields{
-		"message":       "starting on-demand instance",
-		"args":          input,
-		"host_id":       h.Id,
-		"host_provider": h.Distro.Provider,
-		"distro":        h.Distro.Id,
-	})
 	reservation, err := m.client.RunInstances(ctx, input)
 	if err != nil || reservation == nil {
 		if err == EC2InsufficientCapacityError {
@@ -393,12 +386,6 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 	}
 
 	instance := reservation.Instances[0]
-	grip.Debug(message.Fields{
-		"message":       "started ec2 instance",
-		"host_id":       *instance.InstanceId,
-		"distro":        h.Distro.Id,
-		"host_provider": h.Distro.Provider,
-	})
 	h.Id = *instance.InstanceId
 
 	return nil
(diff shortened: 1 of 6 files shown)


cloud/ec2.go
```diff
@@ -318,13 +318,6 @@ func (m *ec2Manager) spawnOnDemandHost(ctx context.Context, h *host.Host, ec2Set
 		input.UserData = &userData
 	}
 
-	grip.Debug(message.Fields{
```
Kimchelly, line 321: Remove this one + the one below doesn't change traceability because the provisioning create host job already logs when it successfully creates a host, and the EC2 manager already `grip.Error`s in all the error cases.

model/task_queue.go
```diff
@@ -555,17 +555,6 @@ func FindAllTaskQueues() ([]TaskQueue, error) {
 func FindDistroTaskQueue(distroID string) (TaskQueue, error) {
 	queue := TaskQueue{}
 	err := db.FindOneQ(TaskQueuesCollection, db.Query(bson.M{taskQueueDistroKey: distroID}), &queue)
-
-	grip.DebugWhen(err == nil, message.Fields{
```
Kimchelly, line 559: All the debug logs in this file + the scheduler look like leftovers from the tunable scheduler project. I can't find any actively-used dashboard that uses this info.

scheduler/utilization_based_host_allocator.go
```diff
@@ -120,19 +120,6 @@ func UtilizationBasedHostAllocator(ctx context.Context, hostAllocatorData *HostA
 	}
 	numNewHostsToRequest := numNewHostsRequired + numAdditionalHostsToMeetMinimum
 
-	grip.Info(message.Fields{
```
Kimchelly, line 123: I took a look and it looks like this + the `"queue state report"` one are leftovers from SRE dashboards, which seem unused since they don't render properly.

trigger/process.go
```diff
@@ -50,10 +50,10 @@ func NotificationsFromEvent(ctx context.Context, e *event.EventLogEntry) ([]noti
 		grip.Error(message.WrapError(err, msg))
 		return nil, err
 	}
-	grip.Info(msg)
 	if len(subscriptions) == 0 {
 		return nil, nil
 	}
+	grip.Info(msg)
```
Kimchelly, line 56: Changed these to conditionally log only if there's something to actually process. It seems like the majority of Splunk logs are for no-op cases (no subscription, no notification) rather than actually having work to do.

hadjri (approved):

PR #7483, opened 2024-02-02
DEVPROD-3946: add Honeycomb attributes for spawn host modification jobs

Kimchelly:
DEVPROD-3946

### Description
Attach more spawn host attributes to the Amboy job Honeycomb spans (specifically for for spawn host modification jobs). This lets us do things like set up Honeycomb dashboards for stops/starts and filter by host fields like whether it's stopping an unexpirable host, the instance type, the source of the stop/start action (a user pressing the button vs. the sleep schedule), etc.

If the job errors (e.g. because AWS returns an error while stopping the host), Amboy's Honeycomb setup already records those errors and stack traces for us. They're also already logged to Splunk.

### Testing
Stopped/starting a spawn host in staging and verified that [the Honeycomb data is being sent](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen/result/cSoZWBv2MFb) and is queryable.

### Documentation
N/A

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -486,13 +486,13 @@ type ModifySpawnHostSource string
 const (
 	// ModifySpawnHostManual means the spawn host is being modified manually
 	// because a user requested it.
-	ModifySpawnHostManual = "manual"
+	ModifySpawnHostManual ModifySpawnHostSource = "manual"
 	// ModifySpawnHostManual means the spawn host is being modified by the
 	// automatic sleep schedule.
-	ModifySpawnHostSleepSchedule = "sleep_schedule"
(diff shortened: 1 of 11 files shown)


hadjri (changes requested):
Changes LGTM - can we add a couple lines to the spawnhost modify/stop/terminate/start unit tests to check that the source is set as we expect it to? Also we should update the PR title to be descriptive.

Kimchelly:
Updated the title and added some tests - nice catch btw! it turns out I actually forgot to set the type for the global constants and didn't notice until the tests failed.

hadjri (approved):

PR #7484, opened 2024-02-05
DEVPROD-4482: add temporary exception to expansions.write redaction for sys-perf

julianedwards:
DEVPROD-4482

### Description
Add temporary exception to expansions.write redaction for sys-perf.

### Testing
Updated unit tests.

### Documentation
N/A

### DIFF
--- a/agent/command/expansion_write.go
+++ b/agent/command/expansion_write.go
@@ -49,12 +49,12 @@ func (c *expansionsWriter) Execute(ctx context.Context,
 	expansions := map[string]string{}
 	for k, v := range conf.Expansions.Map() {
 		_, ok := conf.Redacted[k]
-		// Users should not be able to use the global github token
-		// expansion as it can result in the breaching of Evergreen's
-		// GitHub API limit. Likewise with AWS expansions.
+		// Redact private variables unless redacted set to true. Always
+		// redact the global GitHub and AWS expansions.
 		if (ok && !c.Redacted && !isPerfProject(conf)) || utility.StringSliceContains(ExpansionsToRedact, k) {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


agent/command/expansion_write.go
```diff
@@ -48,10 +49,10 @@ func (c *expansionsWriter) Execute(ctx context.Context,
...
 	for k, v := range conf.Expansions.Map() {
 		_, ok := conf.Redacted[k]
-		// Users should not be able to use the global github token expansion
-		// as it can result in the breaching of Evergreen's GitHub API limit.
-		// Likewise with AWS expansions.
-		if (ok && !c.Redacted) || utility.StringSliceContains(ExpansionsToRedact, k) {
+		// Users should not be able to use the global github token
+		// expansion as it can result in the breaching of Evergreen's
+		// GitHub API limit. Likewise with AWS expansions.
+		if (ok && !c.Redacted && !isPerfProject(conf)) || utility.StringSliceContains(ExpansionsToRedact, k) {
```
ablack12, line 55: This took me some time to parse, since the comment refers only to I think the second half of the if statement (redacting global expansions) but the first half does the actual verification of private variables, right? Can you update this comment to reflect the current state?

ablack12 (approved):
LGTM with comment improvement. I dropped the ticket itself into the sprint coinciding with March 1st to ensure it didn't get lost, if that's alright.

agent/command/expansion_write.go
julianedwards: Done!

julianedwards: I just simplified it a bit.

PR #7493, opened 2024-02-05
DEVPROD-978 Implement check run limit 

malikchaya2:
DEVPROD-978

### Description
Add checks that stops patches from creating more checkruns than the limit.

### Testing
Added a unit test and tested manually for schedule patch, configure patch, and generate patch.

### DIFF
--- a/model/project_test.go
+++ b/model/project_test.go
@@ -841,6 +841,10 @@ func (s *projectSuite) TestCheckRunCount() {
 				Variant:  "bv_1",
 				TaskName: "a_task_2",
 			},
+			TVPair{
+				Variant:  "bv_1",
+				TaskName: "a_task_3",
+			},
 		},
 	}
 
@@ -850,7 +854,7 @@ func (s *projectSuite) TestCheckRunCount() {
 	variantTasks := []patch.VariantTasks{
 		{
 			Variant: "bv_1",
-			Tasks:   []string{"a_task_1", "a_task_2"},
+			Tasks:   []string{"a_task_1", "a_task_2", "a_task_3"},
 		},
 	}
 	checkRunCount = s.project.GetNumCheckRunsFromVariantTasks(variantTasks)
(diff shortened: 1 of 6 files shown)


units/patch_intent.go
```diff
@@ -391,8 +392,14 @@ func (j *patchIntentProcessor) finishPatch(ctx context.Context, patchDoc *patch.
 	}
 
 	if patchDoc.IsGithubPRPatch() {
+		numCheckRuns := patchedProject.GetNumCheckRunsFromVariantTasks(patchDoc.VariantsTasks)
```
bynn, line 395: it looks like we repeat this section a few times could we make it into a function (unless there's a reason it can't easily be done. It might not be worth it)

model/project_test.go
```diff
@@ -824,6 +830,33 @@ func (s *projectSuite) TestAliasResolution() {
 	s.Empty(displayTaskPairs)
 }
 
+func (s *projectSuite) TestCheckRunCount() {
+	pairs := TaskVariantPairs{
+		ExecTasks: TVPairSet{
+			TVPair{
```
bynn, line 836: can we add some tasks that arent a check run just so we know the test is accurate

model/project_test.go
```diff
@@ -824,6 +830,33 @@ func (s *projectSuite) TestAliasResolution() {
...
+		},
+	}
+
+	checkRunCount := s.project.GetNumCheckRunsFromTaskVariantPairs(&pairs)
+	s.Equal(2, checkRunCount)
+
+	variantTasks := []patch.VariantTasks{
+		{
+			Variant: "bv_1",
+			Tasks:   []string{"a_task_1", "a_task_2"},
```
bynn, line 853: same here

bynn (changes requested):

units/patch_intent.go
malikchaya2: numCheckRuns differs, so we can't easily extract it

model/project_test.go
malikchaya2: done

model/project_test.go
malikchaya2: done

bynn (approved):

PR #7494, opened 2024-02-05
DEVPROD-773 Add blocked tasks to BLOCKED_TASK event

bynn:
DEVPROD-773

### Description
More information was requested on the blocked task event so I added which tasks are unattainable to the event 

### Testing
Now has a blocked_on field 

image

### DIFF
--- a/model/event/task_event.go
+++ b/model/event/task_event.go
@@ -40,13 +40,13 @@ const (
 
 // implements Data
 type TaskEventData struct {
-	Execution int      `bson:"execution" json:"execution"`
-	HostId    string   `bson:"h_id,omitempty" json:"host_id,omitempty"`
-	PodID     string   `bson:"pod_id,omitempty" json:"pod_id,omitempty"`
-	UserId    string   `bson:"u_id,omitempty" json:"user_id,omitempty"`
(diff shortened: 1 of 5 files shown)


khelif96 (approved):
It would be awesome if you could propagate this data to the UI by updating the Graphql query that is returned but if not can you open up a follow-up UI ticket so that we can surface this data.

model/event/task_event.go
```diff
@@ -168,8 +169,8 @@ func LogTaskRestarted(taskId string, execution int, userId string) {
 	logTaskEvent(taskId, TaskRestarted, TaskEventData{Execution: execution, UserId: userId})
 }
 
-func LogTaskBlocked(taskId string, execution int) {
-	logTaskEvent(taskId, TaskBlocked, TaskEventData{Execution: execution})
+func LogTaskBlocked(taskId string, execution int, blockedOn string) {
```
hadjri, line 172: nit: let's doc this exported function

hadjri (approved):

model/event/task_event.go
bynn: doced everything :(

PR #7446, opened 2024-01-24
DEVPROD-972 Create upsertCheckRun function in agent

malikchaya2:
DEVPROD-972

### Description
This adds an agent route and function for upsertCheckRun

### Testing
Added unit tests and tested manually:
https://github.com/evergreen-ci/commit-queue-sandbox/pull/611
https://github.com/evergreen-ci/commit-queue-sandbox/pull/611/checks?check_run_id=20794204696

PR #7495, opened 2024-02-05
DEVPROD-992 Call the upsertCheckRun in the agent

malikchaya2:
DEVPROD-992

### Description
This calls upsertCheckRun in the agent. If the GitHub Check Run Configuration is set to 10 or above, it won't call github.


bynn:
have we tried to call this route manually? I am confused on how it would actually upsert because it does not pass in the checkrun id 

also is there a reason why we don't run the check run unless the limit is more than 10? shouldnt it be more than 0 normally

malikchaya2:
> have we tried to call this route manually? I am confused on how it would actually upsert because it does not pass in the checkrun id
> 
> also is there a reason why we don't run the check run unless the limit is more than 10? shouldnt it be more than 0 normally

it was run manually here: https://github.com/evergreen-ci/evergreen/pull/7446

I'm not running it if the limit is less than 10 because we're not ready to release the feature yet, so I figured that when we are ready to release it we set it to something above 10, and then eventually remove that check. I'm essentially using it like a feature flag.

bynn (approved):

service/templates/admin.html
```diff
@@ -2546,6 +2546,9 @@ <h2 class="modal-title">[[modalTitle]]</h2>
 									</md-button>
 								</md-card-title>
 								<md-card-content>
+										<div class="small" style="height:25px;">
```
ablack12, line 2549: Is it needed to have a feature flag for this since it's a route? Also presumably it should error if the thing isn't "enabled" rather than just return nothing.

ablack12 (changes requested):

service/templates/admin.html
bynn: Since this isn't a route that users can manually call, and has to specify something in their task yaml, I think it makes sense. 

We wouldn't want this feature not being enabled to affect(effect?) the success of a task so I thought it made sense.

ablack12: But < 10 isn't an intuitive check. Could we just make it 0? Like, if the feature is _actually off_ then bail.

malikchaya2: yeah I initially made it 10 because I was playing around with different numbers on staging to test other things, but we can keep it 0 on prod and something else on staging. I'll change it to 0.

rest/route/service.go
```diff
@@ -100,7 +100,7 @@ func AttachHandler(app *gimlet.APIApp, opts HandlerOpts) {
 	app.AddRoute("/task/{task_id}/manifest/load").Version(2).Get().Wrap(requireTask).RouteHandler(makeManifestLoad(settings))
 	app.AddRoute("/task/{task_id}/update_push_status").Version(2).Post().Wrap(requireTask).RouteHandler(makeUpdatePushStatus())
 	app.AddRoute("/task/{task_id}/restart").Version(2).Post().Wrap(requireTask).RouteHandler(makeMarkTaskForRestart())
-	app.AddRoute("/task/{task_id}/upsert_check_run").Version(2).Post().Wrap(requireTask).RouteHandler(makeUpsertCheckRun())
+	app.AddRoute("/task/{task_id}/upsert_check_run").Version(2).Post().Wrap(requireTask).RouteHandler(makeUpsertCheckRun(settings))
```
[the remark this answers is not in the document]
malikchaya2: good call, I changed it to just check_run. 
I think it should be Post because if you call it three times, I believe it'll create three check runs.

agent/agent.go
```diff
@@ -980,6 +980,12 @@ func (a *Agent) finishTask(ctx context.Context, tc *taskContext, status string,
 		grip.Error(errors.Wrap(tc.logger.Flush(flushCtx), "flushing logs"))
 	}
 
+	err := a.upsertCheckRun(ctx, tc)
+	if err != nil {
+		grip.Error(errors.Wrap(err, "upserting checkrun"))
+	}
+	grip.Info("Successfully upserted checkRun.")
```
ablack12, line 987: grip.Info just goes to the agent logs i think. Should use tc.logger.Task().Info if you want this to go to users.

ablack12 (approved):

PR #7499, opened 2024-02-06
hotfix: don't panic if string.lastIndex() doesn't find it 

malikchaya2:
hotfix prod panics


model/artifact/artifact_file.go
```diff
@@ -162,5 +162,8 @@ func EscapeFiles(files []File) []File {
 func escapeFile(path string) string {
 	base := filepath.Base(path)
 	i := strings.LastIndex(path, base)
+	if i == -1 {
```
hadjri, line 165: Let's do a <= just to be safe

hadjri (approved):

PR #7501, opened 2024-02-06
Add comments to EscapeFiles

malikchaya2:
### Description
While hotfixing a panic today we realized that this function would be easier to understand with a function comment.


ZackarySantana:
lgtm!

PR #7503, opened 2024-02-07
DEVPROD-4556: use instance profile for AWS auth on the app servers

julianedwards:
DEVPROD-4556

### Description
This brings us one step closer to removing the AWS static creds from the DB.

- Stop passing through the static DB AWS creds on the app server, instead use the default credential chain
- Remove unused fields (`aws_access_key_id `, `aws_secret_access_key`, `key_name`) from the create host API model and update the user-facing documentation

### Testing
- Updated unit tests
- Created a new IAM role with very limited capabilities and assume that role for the AWS integration tests ([staging patch](https://spruce-staging.corp.mongodb.com/version/65c54c3cb237367806bc34b2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))
- Spawned, paused, added tags, and terminated a user host in staging and ran tasks that spawned hosts

### Documentation
Updated the user-facing documentation for the `host.create` command.


jayakasadev (approved):

docs/Project-Configuration/Project-Commands.md
```diff
@@ -586,10 +586,6 @@ EC2 Parameters:
 
 -   `ami` - For an `ec2` provider, the AMI to start. Must set `ami` or `distro`
     but must not set both.
--   `aws_access_key_id` - AWS access key ID. May set to use a
```
hadjri, line 589: Do we have follow up work for notifying people who have these removed fields in their existing `host.create` commands that they won't do anything anymore?

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/aws/aws-sdk-go v1.49.19
 	github.com/aws/aws-sdk-go-v2 v1.24.1
 	github.com/aws/aws-sdk-go-v2/config v1.26.6
-	github.com/aws/aws-sdk-go-v2/credentials v1.16.16
+	github.com/aws/aws-sdk-go-v2/credentials v1.16.16 // indirect
```
hadjri, line 11: Curious what is this comment for?

docs/Project-Configuration/Project-Commands.md
julianedwards: If someone were using it, it would already be broken since we don't respect these fields—if they are set, we always overwrite them with the `mci` creds on the app servers. No one has been using it for at least 4+ years, so we are good to safely remove it as a feature of the command without notifying anyone.

go.mod
julianedwards: It changed to indirect because I removed evergreen's usage of static credentials, but other packages like `pail` that evergreen imports do still use it.

hadjri: Gotcha, thanks

hadjri (approved):

PR #7504, opened 2024-02-07
DEVPROD-4588 turn on the build-and-push task

ybrill:
[DEVPROD-4588](https://jira.mongodb.org/browse/DEVPROD-4588)

### Description
We're going to start turning on the Kanopy cluster soon. Let's turn this on now so we have a bit of history.

### Testing
We should see it running on the project health when we merge.

### DIFF
--- a/self-tests.yml
+++ b/self-tests.yml
@@ -819,9 +819,9 @@ buildvariants:
     display_name: Build and Push
     run_on:
       - ubuntu2204-small
-    activate: false
     tasks:
       - name: ".build"
+        priority: 100
     expansions:
       GOROOT: /opt/golang/go1.20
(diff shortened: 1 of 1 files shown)


self-tests.yml
```diff
@@ -819,9 +819,9 @@ buildvariants:
     display_name: Build and Push
     run_on:
       - ubuntu2204-small
-    activate: false
     tasks:
       - name: ".build"
+        priority: 100
```
ablack12, line 824: I'd envisioned we only put this priority on the staging task; this may be unnecessary for dists

ybrill: done

ablack12 (approved):

PR #7506, opened 2024-02-07
DEVPROD-3616: Fix bug where projects could not enable performance plugin

minnakt:
DEVPROD-3616

### Description
Currently projects cannot enable the performance plugin because `mergedSection.Identifier` is always empty. This PR changes the code to query for the project identifier instead.

### Testing
- Fixed the corresponding tests & fixed some minor errors (e.g. repos do not have identifiers)
- Tested on staging that I can now enable performance plugin when id and identifier is matching


rest/data/project_settings.go
```diff
@@ -314,10 +314,17 @@ func SaveProjectSettingsForSection(ctx context.Context, projectId string, change
...
+
+		// If we are trying to enable the performance plugin but the project's id and identifier are
+		// different, we should error. The performance plugin requires matching id and identifier.
+		if !mergedBeforeRef.IsPerfEnabled() && mergedSection.IsPerfEnabled() {
+			identifier, err := model.GetIdentifierForProject(projectId)
+			if err != nil {
+				return nil, errors.Wrapf(err, "finding project identifier for project '%s'", projectId)
+			}
+			if projectId != identifier {
+				return nil, errors.Errorf("project '%s' does not have a matching ID and identifier, cannot enable performance plugin", identifier)
```
SupaJoon, line 326: nit: I think this error message will read better:
Failed to enable performance plugin for project {project} because its project ID and identifier are different.

SupaJoon (approved):
This may be worth an App review also

malikchaya2 (approved):

rest/data/project_settings.go
minnakt: I'll update it!

PR #7509, opened 2024-02-07
DEVPROD-4233 Set task ID as attribute

bsamek:
DEVPROD-4233

### Description

Including the name of the task in the span makes it counintuitive to search for
the span, as typically the name of a span is the name of the function that it is
in, and is therefore static. This adds the task.id as an attribute instead of in
the span name.

We may wish to announce in Slack that this change is being made, as it may break
existing queries.

### Testing

None.

### Documentation

None.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -607,7 +607,6 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
 
 	tskCtx = utility.ContextWithAttributes(tskCtx, tc.taskConfig.TaskAttributes())
 	tskCtx, span := a.tracer.Start(tskCtx, "task")
-	span.SetAttributes(attribute.String("task.id", tc.task.ID))
 	defer span.End()
 	tc.traceID = span.SpanContext().TraceID().String()
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -606,7 +606,8 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
 	})
 
 	tskCtx = utility.ContextWithAttributes(tskCtx, tc.taskConfig.TaskAttributes())
-	tskCtx, span := a.tracer.Start(tskCtx, fmt.Sprintf("task: '%s'", tc.taskConfig.Task.DisplayName))
+	tskCtx, span := a.tracer.Start(tskCtx, "task")
+	span.SetAttributes(attribute.String("task.id", tc.task.ID))
 	defer span.End()
 	tc.traceID = span.SpanContext().TraceID().String()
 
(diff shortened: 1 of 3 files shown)


agent/agent.go
```diff
@@ -606,7 +606,8 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
 	})
 
 	tskCtx = utility.ContextWithAttributes(tskCtx, tc.taskConfig.TaskAttributes())
-	tskCtx, span := a.tracer.Start(tskCtx, fmt.Sprintf("task: '%s'", tc.taskConfig.Task.DisplayName))
+	tskCtx, span := a.tracer.Start(tskCtx, "task")
+	span.SetAttributes(attribute.String("task.id", tc.task.ID))
```
ybrill, line 610: There should be a task ID attribute set on the span already (`evergreen.task.id`).
Is it missing from this span?

agent/agent.go
```diff
@@ -606,7 +606,8 @@ func (a *Agent) runTask(ctx context.Context, tcInput *taskContext, nt *apimodels
 	})
 
 	tskCtx = utility.ContextWithAttributes(tskCtx, tc.taskConfig.TaskAttributes())
-	tskCtx, span := a.tracer.Start(tskCtx, fmt.Sprintf("task: '%s'", tc.taskConfig.Task.DisplayName))
+	tskCtx, span := a.tracer.Start(tskCtx, "task")
```
ybrill, line 609: do we want to do the same for the [function spans](https://github.com/evergreen-ci/evergreen/blob/26c086e9bd295bf13d87add1e9f0e5da107ddc35/agent/command.go#L187)?

bsamek: I'm torn about this. On the one hand, it seems like the correct thing to do. On the other hand, it would make viewing a trace harder: https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/result/xGe4zGJjqpi/trace/477vTF73h2v?fields[]=s_name&fields[]=s_serviceName&span=36e4a995c90acfeb. On the other hand, you can add the fields as a column.

I made the change though. Maybe I can ask in Slack before I merge?

agent/agent.go
bsamek: Ah, true, removed.

agent/agent.go
ybrill: Either way's good by me 😄

ybrill (approved):

ablack12 (commented):
No need to wait on my review but does anything need to be added or updated here?
https://docs.devprod.prod.corp.mongodb.com/devprod-docs/observability/Honeycomb/honeycomb-exploring-evergreen-data

bsamek:
> No need to wait on my review but does anything need to be added or updated here? https://docs.devprod.prod.corp.mongodb.com/devprod-docs/observability/Honeycomb/honeycomb-exploring-evergreen-data

I don't think so.

PR #7510, opened 2024-02-07
DEVPROD-4507 Redact secrets from OTEL traces and Splunk logs for all graphql requests

khelif96:
DEVPROD-4507

### Description
This PR does several things.

1. Fixes the OTEL configuration so that we can write traces to honeycomb locally if the API keys and collectors are configured correctly. (I can extract this into a separate pr if desired)
2. Create a directive called `redactSecrets` which is used to flag input fields that contain secrets and we should avoid logging to places such as splunk or honeycomb.
3. Create a script that uses the gqlgen configuration to statically generate a list of fields with the `redactSecrets` directive. This is necessary because directives are not captured in otel traces. The trace spans begin after the directive middleware has already been run.
4. Create a function that recursively iterates through a graphql request variables to determine if a known secret field exists and if it does it will swap it out for the word `REDACTED`. This ensures it will work for all queries of all shapes since the query shape is determined by the client.
5. Apply the above function to the `otelgqlgen` tracer to intercept request variables and replace them.
6. Do the same thing for other places where we log to splunk.
7. Expose these values as part of a graphql resolver so that we redact secrets that are logged to the client side loggers.

Some limitations
1. Evaluating the full GraphQL (GQL) type and path for a redacted field is feasible at code generation time, but not during runtime. This limitation stems from the `gqlgen` library's behavior, where it cannot determine the specific type of a field during query execution. Instead, `gqlgen` recognizes the field as having a `map[string]interface{}` type. This interface represents a set of user-defined variables that do not align with the underlying GraphQL structure. Although `gqlgen` conducts some type evaluation to ensure compatibility, this process is not accessible in a manner that I could utilize. Consequently, the approach I adopted involves identifying the field names designated for redaction and masking any values corresponding to those names, irrespective of their origin within the GraphQL schema.

### Testing
Applied the directive to the field `taskName`
image

image

Splunk logs
image

### Documentation


jayakasadev (approved):

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Sort the fields to ensure consistent output.
+	sort.Sort(sort.StringSlice(fields))
+
+	return generateRedactedFieldsFile(fields)
+}
+
+// generateRedactedFieldsFile generates a file that contains a list of fields that should be redacted in logs. It is used to generate the redacted_fields_gen.go file.
+// The file contains a list of fields that should be redacted in logs, a function to check if a field should be redacted, and a function to redact fields in a map.
+func generateRedactedFieldsFile(fields []string) error {
```
julianedwards, line 36: Instead of generating a file with all the code, would it be possible to just create a file (can be really simple, e.g., newline or comma separated field names) that is read in by this code here? This can be done with an `init` func that that writes the values to the global `RedactedFields` slice.

This will make the code here much easier to understand and test.

graphql/splunktracing.go
```diff
@@ -42,11 +43,23 @@ func (SplunkTracing) InterceptResponse(ctx context.Context, next graphql.Respons
 		end := graphql.Now()
 
 		duration := end.Sub(start)
+		// Deep clone the request variables to avoid modifying the original map.
+		redactedRequestVariables := map[string]interface{}{}
+		registeredTypes := []interface{}{
+			map[string]interface{}{},
+		}
+		err := util.DeepCopy(rc.Variables, &redactedRequestVariables, registeredTypes)
```
julianedwards, line 51: Nit: can you merge this with L53?

graphql/http_handler.go
```diff
@@ -23,7 +30,34 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
...
+		}),
+		otelgqlgen.WithRequestVariablesAttributesBuilder(
+			otelgqlgen.RequestVariablesBuilderFunc(func(requestVariables map[string]interface{}) []attribute.KeyValue {
+				// Deep clone the request variables to avoid modifying the original map.
+				redactedRequestVariables := map[string]interface{}{}
+				registeredTypes := []interface{}{
+					map[string]interface{}{},
+				}
+				// We need to deep copy the request variables to avoid modifying the original map. Since it is passed on to the next middleware.
+				err := util.DeepCopy(requestVariables, &redactedRequestVariables, registeredTypes)
```
julianedwards, line 44: Nit: Combine this with L45:
```
if err := util.DeepCopy(requestVariables, &redactedRequestVariables, registeredTypes); err != nil {
...
}
```

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,117 @@
...
+		if field == fieldName {
+			return true
+		}
+	}
+	return false
+}
+
+// RedactFieldsInMap recursively searches for and redacts fields in a map.
+// Assumes map structure like map[string]interface{} where interface{} can be another map, a slice, or a basic datatype.
+func RedactFieldsInMap(data map[string]interface{}) {
```
julianedwards, line 81: Does this need to be exported?

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	_, err = file.WriteString("}\n")
+	if err != nil {
+		return err
+	}
+
+	_, err = file.WriteString(
+		`
+func IsFieldRedacted(fieldName string) bool {
```
julianedwards, line 70: Does this need to be exported?

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,117 @@
...
+	defer file.Close()
+
+	_, err = file.WriteString(`// Code generated by graphql/redact_secrets_plugin.go DO NOT EDIT.
+package graphql
+
+import (
+	"reflect"
+)
+
+var RedactedFields = []string{
```
julianedwards, line 50: I think this should be unexported.

graphql/http_handler.go
```diff
@@ -23,7 +30,34 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
 	// Send OTEL traces for each request.
 	// Only create spans for resolved fields.
 	srv.Use(otelgqlgen.Middleware(
-		otelgqlgen.WithCreateSpanFromFields(func(fieldCtx *graphql.FieldContext) bool { return fieldCtx.IsResolver }),
+		otelgqlgen.WithCreateSpanFromFields(func(fieldCtx *graphql.FieldContext) bool {
+			return fieldCtx.IsMethod
+		}),
+		otelgqlgen.WithRequestVariablesAttributesBuilder(
+			otelgqlgen.RequestVariablesBuilderFunc(func(requestVariables map[string]interface{}) []attribute.KeyValue {
+				// Deep clone the request variables to avoid modifying the original map.
```
julianedwards, line 38: Where would the map get modified?

julianedwards (changes requested):

SupaJoon (changes requested):
Can you add an integration test for the redacted field query? The returned value should include the name the of the GQL type and the full path to the redacted value.

graphql/redact_secrets_plugin.go
hadjri, line 81: Can we write unit tests for this function?

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, field := range RedactedFields {
+		if field == fieldName {
+			return true
+		}
+	}
+	return false
+}
+
+// RedactFieldsInMap recursively searches for and redacts fields in a map.
+// Assumes map structure like map[string]interface{} where interface{} can be another map, a slice, or a basic datatype.
```
hadjri, line 80: Can we do some input validation to assert this assumption?

hadjri (changes requested):

graphql/http_handler.go
khelif96: So RedactedFieldsInMap does an in place update of the `requestVariables`. When writing this I realized that the otelgqlgen library passes a reference to the actual variables the graphql library uses to execute the underlying resolvers. If we don't create a deep clone of the object we risk updating the request variables the resolvers receive and end up returning broken or invalid data.

graphql/redact_secrets_plugin.go
khelif96: Can you clarify what you mean by this? The `reflect.TypeOf(value).kind()` statements perform the validation to validate the type of fields in the interface.

graphql/redact_secrets_plugin.go
khelif96: Updated this

graphql/redact_secrets_plugin.go
khelif96: This process is a bit intensive to actually generate the file since it involves analyzing the entire graphql schema. It takes about 14 seconds to run the script (`generate_secret_fields.go`). So I think adding that to the start up script everytime someone wants to start the local evergreen server or deploy would be a bit intense.

khelif96: This implementation only requires running the script once when someone make a change to the graphql schema and the value is loaded into memory automatically. 
The rest of the graphql code is codegeneration so this script is just an extension of that.

graphql/http_handler.go
```diff
@@ -23,7 +30,33 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
...
+				if err := util.DeepCopy(requestVariables, &redactedRequestVariables, registeredTypes); err != nil {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message": "failed to deep copy request variables",
+					}))
+					return nil
+				}
+				RedactFieldsInMap(redactedRequestVariables, redactedFields)
+				variables := make([]attribute.KeyValue, 0, len(redactedRequestVariables))
+
+				for k, v := range redactedRequestVariables {
```
ybrill, line 53: Maybve this could just call `otelgqlgen.RequestVariables` with the new map?

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func isFieldRedacted(fieldName string, fieldsToRedact []string) bool {
+	for _, field := range fieldsToRedact {
+		if field == fieldName {
+			return true
+		}
+	}
+	return false
+}
```
ybrill, line 75: would there be a performance boost by using a map[string]bool for `fieldsToRedact`?

graphql/http_handler.go
```diff
@@ -23,7 +30,33 @@ func Handler(apiURL string) func(w http.ResponseWriter, r *http.Request) {
...
+					map[string]interface{}{},
+				}
+				// We need to deep copy the request variables to avoid modifying the original map. Since it is passed on to the next middleware.
+				if err := util.DeepCopy(requestVariables, &redactedRequestVariables, registeredTypes); err != nil {
+					grip.Error(message.WrapError(err, message.Fields{
+						"message": "failed to deep copy request variables",
+					}))
+					return nil
+				}
+				RedactFieldsInMap(redactedRequestVariables, redactedFields)
```
ybrill, line 50: Maybe `RedactFieldsInMap` could return a new map to avoid repeating the DeepCopy logic?

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,111 @@
...
+				RedactFieldsInMap(subMap, fieldsToRedact)
+			}
+		}
+
+		// If the value is a slice, iterate over it and redact fields if elements are maps.
+		if reflect.TypeOf(value).Kind() == reflect.Slice {
+			sliceVal := reflect.ValueOf(value)
+			for i := 0; i < sliceVal.Len(); i++ {
+				elem := sliceVal.Index(i).Interface()
+				if reflect.TypeOf(elem).Kind() == reflect.Map {
```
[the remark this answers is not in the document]
khelif96: I don't think this would be [possible](https://github.com/graphql/graphql-spec/issues/423) in graphql. You can't have a nested slices, The items within a slice need to be objects (maps) or primitive values.

graphql/redact_secrets_plugin.go
julianedwards: Right, but you can instead run the script that analyzes the entire graphql schema once and write the fields to a file, upon startup the `init` func would only read in those fields to the `RedactedFields` slice (not run the entire script). Does that make sense?

graphql/http_handler.go
julianedwards: +1

graphql/schema/types/project_settings.graphql
```diff
@@ -178,7 +178,7 @@ type ProjectEventSettings {
   githubWebhooksEnabled: Boolean!
   projectRef: Project
   subscriptions: [GeneralSubscription!]
-  vars: ProjectVars
+  vars: ProjectVars
```
SupaJoon, line 181:
```suggestion
  vars: ProjectVars
```

SupaJoon (approved):

hadjri (approved):

graphql/redact_secrets_plugin.go
hadjri: Right - that should be sufficient, resolving

graphql/redact_secrets_plugin.go
khelif96: There would especially as we continue to redact new fields

graphql/http_handler.go
khelif96: I moved the DeepCopy within RedactFieldsInMap and made it return a new map.

graphql/http_handler.go
khelif96: Good call!

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,131 @@
...
+	dataCopy := map[string]interface{}{}
+	registeredTypes := []interface{}{
+		map[interface{}]interface{}{},
+		map[string]interface{}{},
+		[]interface{}{},
+		[]util.KeyValuePair{},
+	}
+	err := util.DeepCopy(data, &dataCopy, registeredTypes)
+
+	if err != nil {
```
julianedwards, line 87: Nit: combine this with L85.

julianedwards (approved):
LGTM % small nit

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	for _, field := range fields {
+		_, err = file.WriteString(fmt.Sprintf("\t\"%s\": true,\n", field))
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err = file.WriteString("}\n\n")
```
ybrill, line 63: I would have thought to have just one \n

graphql/redact_secrets_plugin.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func isFieldRedacted(fieldName string, fieldsToRedact map[string]bool) bool {
+	_, ok := fieldsToRedact[fieldName]
+	return ok
+}
```
ybrill, line 73: is there an advantage to checking the field is in the map rather than just checking the value in the map? (since values not in the map will return false as well)

environment.go
```diff
@@ -891,6 +893,11 @@ func (e *envState) initTracer(ctx context.Context, useInternalDNS bool) error {
 		opts = append(opts, otlptracegrpc.WithInsecure())
 	} else {
 		opts = append(opts, otlptracegrpc.WithEndpoint(e.settings.Tracer.CollectorEndpoint))
+		if e.settings.Tracer.CollectorAPIKey != "" {
+			opts = append(opts, otlptracegrpc.WithHeaders(map[string]string{
+				honeycombCollectorHeader: e.settings.Tracer.CollectorAPIKey,
+			}))
+		}
```
ybrill, line 900: maybe we'd also want to do this when `useInternalDNS` is true? On the other hand, maybe we'd never set the header when it's running in k8s. :shrug:

ybrill (approved):

PR #7512, opened 2024-02-08
DEVPROD-3025 Update evaluate command to output sorted build variants

bynn:
DEVPROD-3025

### Description
evaluate has a --diffable flag that will alphabetize tasks and bvs 

### Testing
tested locally 
image

### DIFF
--- a/operations/evaluate.go
+++ b/operations/evaluate.go
@@ -61,7 +61,10 @@ func Evaluate() cli.Command {
 				sortTasksByName := model.ProjectTasksByName(p.Tasks)
 				sort.Sort(sortTasksByName)
 				p.Tasks = sortTasksByName
-				sort.Sort(p.BuildVariants)
+
+				sortBuildVariantsByName := model.BuildVariantsByName(p.BuildVariants)
+				sort.Sort(sortBuildVariantsByName)
+				p.BuildVariants = model.BuildVariants(sortBuildVariantsByName)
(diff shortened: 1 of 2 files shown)


Kimchelly (commented):
I wrote a thing a while back to do this in #6142 (which we only had temporariy), but it sorts by both build variant name and sorts the tasks within the build variant. I wonder if maybe we can reuse that?

operations/evaluate.go
```diff
@@ -50,6 +57,12 @@ func Evaluate() cli.Command {
 			if err != nil {
 				return errors.Wrap(err, "loading project")
 			}
+			if diffable {
+				sortTasksByName := model.ProjectTasksByName(p.Tasks)
+				sort.Sort(sortTasksByName)
+				p.Tasks = sortTasksByName
+				sort.Sort(p.BuildVariants)
```
Kimchelly, line 64: This looks like it would sort by display name, but `ProjectTasksByName` sorts by task name. Can we make them consistent and sort by name (or display name) for both of them? There's a `model.BuildVariantsByName` if it helps.

Kimchelly (approved):
LGTM mod one suggestion to make the sorting consistent on name or display name.

operations/evaluate.go
bynn: they are both by name now!

PR #7513, opened 2024-02-08
DEVPROD-973

bynn:
DEVPROD-973

### Description
calling re-run on a check run will restart the  associated task 
does not update the PR with a comment saying it does because check run events do not return a PR number (will ask github if possible and follow up with a ticket) 

### Testing
calling re-run on stage2 on this [check run](https://github.com/evergreen-ci/commit-queue-sandbox/pull/617/checks?check_run_id=21350875510)
trigger [this task](https://spruce-staging.corp.mongodb.com/task/evg_race_detector_test_thirdparty_c4a5214e9023da31b2ee488f6997f8e9e1295f20_24_02_07_18_14_31/logs?execution=1&sortBy=STATUS&sortDir=ASC) to restart


malikchaya2 (changes requested):
Discussed adding another test in dms, will review after.

malikchaya2 (changes requested):

rest/route/github.go
```diff
@@ -263,11 +264,107 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+	case *github.CheckRunEvent:
+		if event.GetAction() == githubActionRerequested {
+			return gh.handleCheckRunRerequested(ctx, event)
+		}
 	}
 
 	return gimlet.NewJSONResponse(struct{}{})
 }
 
+func (gh *githubHookApi) handleCheckRunRerequested(ctx context.Context, event *github.CheckRunEvent) gimlet.Responder {
```
malikchaya2, line 277: Can we add a comment either here or higher up explaining that this will handle a user clicking the restart button on the check run in GitHub?

rest/route/github.go
```diff
@@ -263,11 +264,107 @@ func (gh *githubHookApi) Run(ctx context.Context) gimlet.Responder {
...
+	owner := event.Repo.Owner.GetLogin()
+	repo := event.Repo.GetName()
```
malikchaya2, line 279: should we check to make sure they are not `""`?

malikchaya2 (approved):
approved with nits

rest/route/github.go
bynn: I think it should be fine because we don't do that anywhere else in this file and we can assume that we always have owner repo from github events

rest/route/github.go
bynn: added

PR #7519, opened 2024-02-08
DEVPROD-4510 Log script before expanding expansions

bsamek:
DEVPROD-4510

### Description

This prints the script for troubleshooting before expansions are expanded.

### Testing

This should be covered by existing tests, and we should validate that a task runs in staging.

### Documentation

None.


ablack12 (approved):
LGTM, you'll have to update the agent version string

bsamek:
Testing in staging.

PR #7520, opened 2024-02-08
DEVPROD-3973: manage persistent DNS names

Kimchelly:
DEVPROD-3973

### Description
This PR 1. assigns persistent DNS records to unexpirable hosts whenever they start up (the first time they're created and every time a stopped one is started) and 2. deletes those records when the host is terminated. Those DNS records can be used to resolve the host's IP address, which makes it more resilient to IP address changes whenever the host reboots.

To reduce the size of this PR, I'm splitting this ticket into two PRs because it still doesn't handle some cases. For example, if a host starts out expirable and is later edited to be unexpirable, that's not handled yet - it'll be handled in the follow-up PR. Furthermore, unexpirable hosts that already exist and don't have a persistent DNS name won't have a persistent DNS name assigned unless they stop/start their host - these existing hosts will all be assigned persistent DNS names in DEVPROD-4040.

It only applies if the required admin settings have been set, so this won't take effect in prod until those admin settings are set.

### Testing
* Added unit tests for new logic.
* Tested with unexpirable hosts in staging to ensure that I could SSH using the persistent DNS name and could also `dig +short` the persistent DNS name and it resolved into the correct IP address, even after rebooting the host.

### Documentation
N/A - this currently isn't visible in the UI nor is it rolled out for all unexpirable hosts yet. Documentation of persistent DNS names will come out with sleep schedule documentation in DEVPROD-4055.

### DIFF
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -789,6 +789,18 @@ func (m *ec2Manager) TerminateInstance(ctx context.Context, h *host.Host, user, 
 	if !IsEC2InstanceID(h.Id) {
 		return errors.Wrap(h.Terminate(ctx, user, fmt.Sprintf("detected invalid instance ID '%s'", h.Id)), "terminating instance in DB")
 	}
+
+	if h.NoExpiration {
+		// Clean up remaining DNS records for unexpirable hosts.
+		grip.Error(message.WrapError(deleteHostPersistentDNSName(ctx, h, m.client), message.Fields{
+			"message":    "could not delete host's persistent DNS name",
+			"op":         "delete",
+			"dashboard":  "evergreen sleep schedule health",
+			"host_id":    h.Id,
+			"started_by": h.StartedBy,
+		}))
+	}
+
 	resp, err := m.client.TerminateInstances(ctx, &ec2.TerminateInstancesInput{
 		InstanceIds: []string{h.Id},
 	})
--- a/cloud/ec2.go
+++ b/cloud/ec2.go
@@ -717,7 +717,7 @@ func (m *ec2Manager) GetInstanceStatuses(ctx context.Context, hosts []host.Host)
 			status := ec2StatusToEvergreenStatus(instance.State.Name)
 			if status == StatusRunning {
 				// cache instance information so we can make fewer calls to AWS's API
-				if err = cacheHostData(ctx, instanceIdToHostMap[hostsToCheck[i]], &instance, m.client); err != nil {
+				if err = cacheHostData(ctx, m.env, instanceIdToHostMap[hostsToCheck[i]], &instance, m.client); err != nil {
 					return nil, errors.Wrapf(err, "caching EC2 host data for host '%s'", hostsToCheck[i])
 				}
 			}
@@ -754,7 +754,7 @@ func (m *ec2Manager) GetInstanceStatus(ctx context.Context, h *host.Host) (Cloud
 
 	if status == StatusRunning {
 		// cache instance information so we can make fewer calls to AWS's API
-		if err = cacheHostData(ctx, h, instance, m.client); err != nil {
+		if err = cacheHostData(ctx, m.env, h, instance, m.client); err != nil {
 			return status, errors.Wrapf(err, "caching EC2 host data for host '%s'", h.Id)
 		}
 	}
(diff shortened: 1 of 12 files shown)


cloud/ec2.go
```diff
@@ -789,6 +789,18 @@ func (m *ec2Manager) TerminateInstance(ctx context.Context, h *host.Host, user,
 	if !IsEC2InstanceID(h.Id) {
 		return errors.Wrap(h.Terminate(ctx, user, fmt.Sprintf("detected invalid instance ID '%s'", h.Id)), "terminating instance in DB")
 	}
+
+	if h.NoExpiration {
+		// Clean up remaining DNS records for unexpirable hosts.
+		grip.Error(message.WrapError(deleteHostPersistentDNSName(ctx, m.env, h, m.client), message.Fields{
+			"message":    "could not delete host's persistent DNS name",
+			"op":         "delete",
+			"dashboard":  "evergreen sleep schedule health",
```
malikchaya2, line 798: I haven't seen the "dashboard" field in logs before. How do you plan to use it?

cloud/ec2.go
```diff
@@ -789,6 +789,18 @@ func (m *ec2Manager) TerminateInstance(ctx context.Context, h *host.Host, user,
 	if !IsEC2InstanceID(h.Id) {
 		return errors.Wrap(h.Terminate(ctx, user, fmt.Sprintf("detected invalid instance ID '%s'", h.Id)), "terminating instance in DB")
 	}
+
+	if h.NoExpiration {
```
malikchaya2, line 793: Just to double check my understanding: hosts that are currently considered unexpirable, once they sign up to follow a sleep schedule will no longer be considered unexpirable, right? And only hosts that have filed for an exception will be unexpirable?

cloud/ec2_util.go
```diff
@@ -310,6 +316,115 @@ func cacheHostData(ctx context.Context, h *host.Host, instance *types.Instance,
...
+	ipAddr := utility.FromStringPtr(instance.PublicIpAddress)
+	if ipAddr == "" {
+		return errors.New("instance did not include an IP address")
+	}
+
+	settings := env.Settings()
+	if settings.Providers.AWS.PersistentDNS.Domain == "" || settings.Providers.AWS.PersistentDNS.HostedZoneID == "" {
+		// TODO (DEVPROD-4040): once all unexpirable hosts have a persistent DNS
+		// name assigned, this should return an error because not having these
+		// settings will prevent persistent DNS names from updating properly.
```
malikchaya2, line 349: Is there a way to assign them when we hit this?

model/host/db.go
```diff
@@ -1351,3 +1353,11 @@ func FindUnexpirableRunning() ([]Host, error) {
 	}
 	return hosts, db.FindAllQ(Collection, db.Query(q), &hosts)
 }
+
+// FindOneByPersistentDNSName returns hosts that have a matching persistent DNS
+// name.
+func FindOneByPersistentDNSName(ctx context.Context, dnsName string) (*Host, error) {
+	return FindOne(ctx, bson.M{
+		PersistentDNSNameKey: dnsName,
```
malikchaya2, line 1361: should we add an index for this or other queries added here?

malikchaya2 (approved):
Approved with some questions.

cloud/ec2.go
Kimchelly: Unexpirable is a separate concept from the sleep schedule - it only means that Evergreen will never terminate the host unless they terminate it themselves (or make it expirable again). The sleep schedule only applies to unexpirable hosts.

cloud/ec2.go
Kimchelly: I'm mostly using it prevent somebody from randomly deciding to delete or modify the log without knowing what effects it'll have. You can see the dashboard in Splunk (it's called "Evergreen Sleep Schedule Health")

cloud/ec2_util.go
Kimchelly: We can't because these are admin settings that are meant to be manually assigned once the feature is ready for users to use. Auto-setting isn't something we do because then we have to hard-code production values into the source code + it would auto-enable the feature when we may not be ready to turn it on.

model/host/db.go
Kimchelly: There's already an index in staging on this key, I was just waiting for an LGTM to duplicate it in prod.

There's no other new queries being introduced in this PR, just updates by host ID.

cloud/ec2.go
malikchaya2: thank you!

PR #7528, opened 2024-02-12
docs: provide better explanation for cron

Kimchelly:
[Slack thread](https://mongodb.slack.com/archives/C0V896UV8/p1707763122311109)

A user asked about its behavior, and I've heard people in the past also be confused about `cron`'s behavior.


ablack12 (approved):
this lgtm; i think we should also really just have a new section bc I think this is a little wordy for a bullet point, so i also collected https://jira.mongodb.org/browse/DEVPROD-4690?focusedId=6088609#comment-6088609 but I think this is still a good improvement right now, and if you think it's enough feel free to close the ticket!

Kimchelly:
Rolling this example into the new section for DEVPROD-4690 SGTM! I think just providing a quick inline example may help resolve enough confusion for now.

PR #7529, opened 2024-02-12
DEVPROD-4063 Document exec timeout

bynn:
DEVPROD-4063

### Description
document clarification

### DIFF
--- a/docs/Project-Configuration/Project-Configuration-Files.md
+++ b/docs/Project-Configuration/Project-Configuration-Files.md
@@ -32,11 +32,11 @@ For example, a couple of tasks might look like:
 ``` yaml
 tasks:
 - name: compile
+  exec_timeout_secs: 20
   commands:
     - command: git.get_project
       params:
         directory: src
-        exec_timeout_secs: 20
     - func: "compile and upload to s3"
 - name: passing_test
   run_on: my_other_distro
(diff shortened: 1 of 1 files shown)


docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -36,6 +36,7 @@ tasks:
     - command: git.get_project
       params:
         directory: src
+        exec_timeout_secs: 20
```
malikchaya2, line 39: Is this set in params or on the command (like on line 37)?

malikchaya2 (changes requested):

docs/Project-Configuration/Project-Configuration-Files.md
bynn: oop yes it should be a little higher

docs/Project-Configuration/Project-Configuration-Files.md
```diff
@@ -32,6 +32,7 @@ For example, a couple of tasks might look like:
 ``` yaml
 tasks:
 - name: compile
+  exec_timeout_secs: 20
```
malikchaya2, line 35: Since this is meant to demonstrate that below "will override the project level exec_timeout defined above for this task", we should probably make this a different value than the one below.

malikchaya2 (approved):
Approved with one requested change.

PR #7530, opened 2024-02-12
DEVPROD-3616: Use mergedBeforeRef instead of querying for identifier

minnakt:
DEVPROD-3616

### Description
Correcting a mistake I made — there's no need to query for identifier because it's already stored in `mergedBeforeRef`.

### Testing
- Tested on staging


Kimchelly (approved):

ablack12 (approved):
Thank you!

PR #7533, opened 2024-02-15
DEVPROD-1739: Fix status filter behavior in getTasksByVersionPipeline

minnakt:
DEVPROD-1739

### Description
Fixes a bug where it's not possible to filter for _different_ status and base status on the task page. (e.g. `opts.Statuses = ["failed"]`, `opts.BaseStatuses = ["success"]`)

This is caused by a logical error in the pipeline. The pipeline first gets all version tasks and their base tasks. Then the pipeline filters out tasks that don't have a status that matches `opts.Statuses`. This is not correct because the pipeline is still operating on the base tasks at this point, so it will remove all base tasks that don't match `opts.Statuses`. `opts.Statuses` should not apply to the base tasks at all.

Moving this stage later in the pipeline fixes the problem because the base task is now embedded in each task and will not match the filter.

### Testing
- On staging
- Added a test that fails without these changes
- Benchmarked the pipeline on Compass; it's a little slower but still performs under 500ms for large server patches. The old pipeline was incorrectly omitting thousands of results which was helping to make it faster.

### DIFF
--- a/model/task/db.go
+++ b/model/task/db.go
@@ -2601,6 +2601,7 @@ func getTasksByVersionPipeline(versionID string, opts GetTasksByVersionOptions) 
 		{"$project": projectOut},
 		{"$match": match},
 	}
+
 	if !opts.IncludeExecutionTasks {
 		pipeline = append(pipeline, bson.M{
 			"$match": bson.M{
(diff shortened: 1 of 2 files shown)


khelif96 (approved):
Great catch 🚀

bynn (approved):

PR #7534, opened 2024-02-15
DEVPROD-4739: fix dependency inclusion for disabled tasks in task groups

Kimchelly:
DEVPROD-4739

### Description
This fixes an issue where if you have a task within a task group that's disabled, then try to schedule it using the schedule patch REST route, it won't schedule the task group or return any error. This normally isn't a problem you encounter when you configure a patch in Spruce because it already auto-filters out disabled tasks from your selection choices.

* Allow task groups with some disabled tasks to be scheduled.
* Continue on error in more cases rather than exit early (to ensure later tasks in a list get scheduled even if there are errors for other tasks).
* Add a TODO for a similar bug (but for requesters instead of disable).

### Testing
* Add unit tests for patch configure route and dependency inclusion for partly-disabled task group tasks.
* Tested manually by submitting [a patch to prod](https://spruce.mongodb.com/patch/65ce36b05623432adc5c7d92/configure/tasks) and then [a patch to staging](https://spruce-staging.corp.mongodb.com/version/65ce37e1a7889000079fc418/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). I used the patch configure route on both to try to schedule the task group and it succeeded in staging once I applied my fix.

### Documentation
N/A

### DIFF
--- a/model/dependency.go
+++ b/model/dependency.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/message"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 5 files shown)


model/dependency.go
```diff
@@ -75,15 +76,29 @@ func (di *dependencyIncluder) handle(pair TVPair, activationInfo *specificActiva
...
 
-	// if the given task is a task group, recurse on each task
+	// For a task group, recurse on each task and add those tasks that should be
+	// included.
 	if tg := di.Project.FindTaskGroup(pair.TaskName); tg != nil {
+		catcher := grip.NewBasicCatcher()
 		for _, t := range tg.Tasks {
 			ok, err := di.handle(TVPair{TaskName: t, Variant: pair.Variant}, activationInfo, generatedVariants, false)
-			if !ok {
-				di.included[pair] = false
```
bynn, line 83: can we keep this line so we don't have to check unschedulable tasks over and over again?

bynn (changes requested):

model/dependency.go
Kimchelly: I read over the logic and I believe that setting `di.included[pair] = false` in every `!ok` case from `di.handle` doesn't actually do anything because it already sets the task inclusion false in every return case where `di.handle` returns false ([here](https://github.com/evergreen-ci/evergreen/blob/86e8b8340c5c7eb8a65575d6797613fcda59c5c2/model/dependency.go#L92-L107)).

Regardless of whether `ok` is used or not, I don't think the task group itself (which is represented by `pair`) needs to be added to `di.included` since it'll never be true. The task group is just a logical grouping mechanism, so it's not a real task you can schedule or depend on (kind of like an execution task is a real task and a display task is just a different way of showing it in the UI).

bynn (approved):

PR #7538, opened 2024-02-15
DEVPROD-728 Log and system fail the task when an agent is re-requesting the same next task a suspicious number of times

malikchaya2:
DEVPROD-728 

### Description
The next task loop in the agent + the coordination with the next task endpoint is likely to cause outages if we change it because if the agent panics, it can repeatedly ask for the next task forever, and the app server will repeatedly give it the same task. 

Repeatedly asking for the same task many times is the most obvious indicator that something is wrong. This system fails the task and logs. An alert will be set up as well so that engineers can manually intervene when it happens, a revert is likely needed. 

### Testing
Added unit tests.


config.go
```diff
@@ -38,7 +38,7 @@ var (
 
 	// Agent version to control agent rollover. The format is the calendar date
 	// (YYYY-MM-DD).
-	AgentVersion = "2024-02-12"
+	AgentVersion = "2024-02-15"
```
Kimchelly, line 41: Does the agent version need to be updated? All the changes look like they're on the app server, not the agent.

model/task/db.go
```diff
@@ -2839,6 +2840,16 @@ func SetHasAnnotations(taskId string, execution int) error {
 	return errors.Wrapf(err, "marking task '%s' as having annotations", taskId)
 }
 
+// SetNumNextTaskDispatches sets the next task count for a task.
+func SetNumNextTaskDispatches(taskId string, execution int, count int) error {
+	err := UpdateOne(
+		ByIdAndExecution(taskId, execution),
+		bson.M{"$set": bson.M{
+			NumNextTaskDispatchesKey: count,
+		}})
```
Kimchelly, line 2849: nit: don't forget to update the in-memory task itself. I don't think it's that important though if you want to ignore this.

model/task/db.go
```diff
@@ -2839,6 +2840,16 @@ func SetHasAnnotations(taskId string, execution int) error {
 	return errors.Wrapf(err, "marking task '%s' as having annotations", taskId)
 }
 
+// SetNumNextTaskDispatches sets the next task count for a task.
+func SetNumNextTaskDispatches(taskId string, execution int, count int) error {
+	err := UpdateOne(
+		ByIdAndExecution(taskId, execution),
+		bson.M{"$set": bson.M{
```
Kimchelly, line 2847: Should this be `$inc` rather than `$set` with an incremented count? `$inc` is slightly safer because it'll increment the counter in a way that's thread-safe.

model/task/db.go
```diff
@@ -2839,6 +2840,16 @@ func SetHasAnnotations(taskId string, execution int) error {
 	return errors.Wrapf(err, "marking task '%s' as having annotations", taskId)
 }
 
+// SetNumNextTaskDispatches sets the next task count for a task.
```
Kimchelly, line 2843: nit: `sets the number of times a host has requested this task and execution as its next task`.

model/task/task.go
```diff
@@ -499,14 +503,19 @@ func (t *Task) IsDispatchable() bool {
...
-	return t.IsHostTask() && t.WillRun()
+	return t.IsHostTask() && t.WillRun() && !t.IsStuckTask()
 }
 
 // IsHostTask returns true if it's a task that runs on hosts.
 func (t *Task) IsHostTask() bool {
 	return (t.ExecutionPlatform == "" || t.ExecutionPlatform == ExecutionPlatformHost) && !t.DisplayOnly
 }
 
+// IsStuckTask returns true if the task has been dispatched over a set number of times.
```
Kimchelly, line 514: nit: `dispatched over the system limit`

rest/route/host_agent.go
```diff
@@ -663,6 +683,15 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
 			continue
 		}
 
+		err = task.SetNumNextTaskDispatches(nextTask.Id, nextTask.Execution, nextTask.NumNextTaskDispatches+1)
```
Kimchelly, line 686: Does this get reset back to 0 when the task restarts?

rest/route/host_agent.go
```diff
@@ -663,6 +683,15 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
...
+		if err != nil {
```
Kimchelly, line 687: nit: L687-692 can be simplified down to `grip.Error(message.WrapError(err, message.Fields { ...}))` because `message.WrapError` already suppresses logging for nil errors.

rest/route/host_agent.go
```diff
@@ -525,6 +525,26 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
...
+				// in that case it will likely cause the same panic again. This resets the task and logs so that the task
+				// doesn't have to  be manually unstuck and an alert can be set up that alerts engineers who can manually
+				// intervene and fix the agent -- likely via a revert.
+				msg := fmt.Sprintf("The agent has re-requested the same task '%d' times.", evergreen.MaxTaskDispatchAttempts)
+				msg += " It's possible that the agent hit a panic and is in a bad state."
+
+				grip.Error(message.Fields{
+					"message":   msg,
+					"distro_id": d.Id,
+					"task_id":   nextTask.Id,
```
Kimchelly, line 543: Can we also include the task execution? May help with debugging.

rest/route/host_agent.go
```diff
@@ -525,6 +525,26 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
 				"host_id":   currentHost.Id,
 			}))
 
+			if nextTask.IsStuckTask() {
+				if err := model.EndAndResetSystemFailedTask(ctx, env.Settings(), nextTask, evergreen.TaskDescriptionStranded); err != nil {
+					return nil, false, errors.Wrap(err, "ending and resetting system failed task")
+				}
+				// The agent requesting the same task over and over again is a sign that the agent may have hit a panic
```
Kimchelly, line 532: I don't think we need to be so specific with what we guess is the cause. The agent could be doing this for other reasons, like having a non-panic-related bug, or something unusual going on in the host that shuts down the agent. I think it would suffice to say something short and simple to the effect of (but maybe not with these exact words) "the agent is expected to run the task once it's been assigned it. If it's asking for to run the same next task over and over after it's been assigned that task before, there's an issue going on".

model/task/task.go
```diff
@@ -499,14 +503,19 @@ func (t *Task) IsDispatchable() bool {
 // IsHostDispatchable returns true if the task should run on a host and can be
 // dispatched.
 func (t *Task) IsHostDispatchable() bool {
-	return t.IsHostTask() && t.WillRun()
+	return t.IsHostTask() && t.WillRun() && !t.IsStuckTask()
```
Kimchelly, line 506: See my comment about `assignNextAvailableTask`, but I don't think this would ever actually run. If the host has already been assigned a task, it'll never hit this case.

rest/route/host_agent.go
```diff
@@ -525,6 +525,26 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
 				"host_id":   currentHost.Id,
 			}))
 
+			if nextTask.IsStuckTask() {
```
Kimchelly, line 528: I'm pretty sure if we want to check for the issue described in the ticket (a host being assigned a task, and then asking for that task over and over again), we have to check it [here in sendBackRunningTask](https://github.com/evergreen-ci/evergreen/blob/8fe05b3b8f9e14487101ac82e3f308e959b8b2d7/rest/route/host_agent.go#L182-L1846) instead of in `assignNextAvailableTask`. If a host has already been assigned a task by the app server, its RunningTask will be set, so that case will just hand the already-assigned task back to the host again, meaning it'll never reach `assignNextAvailableTask` with that same task again.

rest/route/host_agent.go
Kimchelly, line 686: A similar thing to the above comment, but in the case where [the host is already assigned a running task](https://github.com/evergreen-ci/evergreen/blob/8fe05b3b8f9e14487101ac82e3f308e959b8b2d7/rest/route/host_agent.go#L182-L184), we should increment the dispatch count.

rest/route/host_agent.go
```diff
@@ -525,6 +525,26 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
...
+				// doesn't have to  be manually unstuck and an alert can be set up that alerts engineers who can manually
+				// intervene and fix the agent -- likely via a revert.
+				msg := fmt.Sprintf("The agent has re-requested the same task '%d' times.", evergreen.MaxTaskDispatchAttempts)
+				msg += " It's possible that the agent hit a panic and is in a bad state."
+
+				grip.Error(message.Fields{
+					"message":   msg,
+					"distro_id": d.Id,
+					"task_id":   nextTask.Id,
+					"host_id":   currentHost.Id,
```
Kimchelly, line 544: Can we include the agent's revision in this message so we can determine if it's a bad agent deploy?

rest/route/host_agent.go
```diff
@@ -525,6 +525,26 @@ func assignNextAvailableTask(ctx context.Context, env evergreen.Environment, tas
 				"host_id":   currentHost.Id,
 			}))
 
+			if nextTask.IsStuckTask() {
+				if err := model.EndAndResetSystemFailedTask(ctx, env.Settings(), nextTask, evergreen.TaskDescriptionStranded); err != nil {
```
Kimchelly, line 529: I think what we want here is `ClearAndResetStrandedHostTask` rather than `EndAndResetSystemFailedTask` because `EndAndResetSystemFailedTask` doesn't de-assign the task from the host.

Kimchelly (changes requested):

model/task/db.go
```diff
@@ -2866,6 +2867,18 @@ func SetHasAnnotations(taskId string, execution int) error {
 	return errors.Wrapf(err, "marking task '%s' as having annotations", taskId)
 }
 
+// SetNumNextTaskDispatches sets the number of times a host has requested this
+// task and execution as its next task.
+func (t *Task) SetNumNextTaskDispatches() error {
```
Kimchelly, line 2872: Can we call this `IncNumNextTaskDispatches` or something since it's doing a +1 now instead of setting the value?

rest/route/host_agent.go
Kimchelly: Oh I was referring to whether this count gets reset back to 0 if you restart the task. When restarting a task, we basically make a copy of the task and [reset some fields back to a clean state](https://github.com/evergreen-ci/evergreen/blob/21c2cae10e46d135e26b1dcd771add7a4c09658a/model/task/task.go#L2232). So for example, if the number of dispatches is 3 for this task execution, then when the user restarts it, we should make sure that this count resets back to 0 dispatches for the next execution.

rest/route/host_agent.go
```diff
@@ -1085,6 +1091,25 @@ func sendBackRunningTask(ctx context.Context, env evergreen.Environment, h *host
...
+		// task over and over again is a sign that something is wrong.
+		msg := fmt.Sprintf("The agent has re-requested the same task '%d' times.", evergreen.MaxTaskDispatchAttempts)
+		msg += " It's possible that the agent is in a bad state."
+
+		grip.Error(message.Fields{
+			"message":        msg,
+			"task_id":        t.Id,
+			"task_execution": t.Execution,
+			"host_id":        h.Id,
+			"agent_version":  evergreen.AgentVersion,
```
Kimchelly, line 1109: This should log `h.AgentRevision` rather than `evergreen.AgentVersion`. `h.AgentRevision` is the version of the agent currently running on the host, whereas `evergreen.AgentVersion` is the "goal" agent version. They can be different for some time after a deploy if the host hasn't gotten a chance to deploy the newest agent yet, and that difference may be particularly relevant when investigating this log.

rest/route/host_agent_test.go
```diff
@@ -994,6 +995,65 @@ func TestAssignNextAvailableTaskWithDispatcherSettingsVersionLegacy(t *testing.T
 			So(shouldTeardown, ShouldBeFalse)
 			So(t, ShouldBeNil)
 		})
+		Convey("a task's NumNextTaskDispatches should be increased by 1", func() {
```
Kimchelly, line 998: Since the above logic was modified to increment the counter when sending back an already-assigned task, should there be a test for when the task reaches the dispatching limit? Would be good to know that the route fixes the host/task as expected when it hits the limit.

rest/route/host_agent.go
```diff
@@ -1085,6 +1091,25 @@ func sendBackRunningTask(ctx context.Context, env evergreen.Environment, h *host
...
+		})
```
Kimchelly, line 1110: Since it's hit the dispatching limit in this case, is this supposed to return early, or continue with the rest of the logic in the fucntion?

Kimchelly (changes requested):

config.go
malikchaya2: I don't think that's set anymore

rest/route/host_agent.go
```diff
@@ -1085,6 +1091,30 @@ func sendBackRunningTask(ctx context.Context, env evergreen.Environment, h *host
...
+			"message":        msg,
+			"task_id":        t.Id,
+			"task_execution": t.Execution,
+			"host_id":        h.Id,
+			"agent_version":  h.AgentRevision,
+		})
+
+		return gimlet.MakeJSONInternalErrorResponder(gimlet.ErrorResponse{
+			StatusCode: http.StatusInternalServerError,
+			Message:    fmt.Sprintf("The agent has re-requested the task '%s' with execution '%d' '%d' times.", t.Id, t.Execution, evergreen.MaxTaskDispatchAttempts),
```
Kimchelly, line 1114: nit: the surrounding single quotes formatting (`'%d'`) around the number usually isn't necessary. It's only really helpful for string formatting because if the string is empty, it won't show any string at all (so it'll just look like a weird blank space).

rest/route/host_agent_test.go
```diff
@@ -994,6 +1032,65 @@ func TestAssignNextAvailableTaskWithDispatcherSettingsVersionLegacy(t *testing.T
...
+			t, shouldTeardown, err := assignNextAvailableTask(ctx, env, taskQueue, model.NewTaskDispatchService(time.Minute), &theHostWhoCanBoastTheMostRoast, details)
+			So(err, ShouldBeNil)
+			So(shouldTeardown, ShouldBeFalse)
+			So(t, ShouldNotBeNil)
+			taskBFromDb, err := task.FindOneId(stillOkayTask.Id)
+			So(err, ShouldBeNil)
+			So(taskBFromDb, ShouldNotBeNil)
+			So(taskBFromDb.IsStuckTask(), ShouldBeTrue)
+		})
+		Convey("a stuck tasks should not error", func() {
```
Kimchelly, line 1067: I don't think this test is relevant anymore it's already covered by the `AStuckNextTaskShouldError` test case, so may be fine to remove.

Kimchelly (approved):
LGTM mod nits.

PR #7539, opened 2024-02-16
DEVPROD-3812 Clear host secret once it is determined to be unreachable

hadjri:
DEVPROD-3812

### Description
We saw an instance where a host got quarantined, but became unresponsive to the logic where we try to [shut down the agent monitor on a host once it's quarantined](https://gist.github.com/hadjri/fd04881487b1b49924f1eb4b2e449e76#file-host_agent-go-L274-L285).

Since the host was unresponsive (whether it was OOMing or something else), this is what caused an endless loop of `"could not prepare host for agent to exit"` errors to be sent to Splunk.

Made a change such that logs an event every time we fail to stop the agent on a quarantined host, and if that event happens N=10 times in a row, clear the host's secret so its agent will stop being able to communicate with the app server.

### Testing
Added a unit test


model/event/host_event_finder.go
```diff
@@ -85,3 +86,15 @@ func AllRecentHostEventsMatchStatus(ctx context.Context, hostId string, n int, s
 	return true
 
 }
+
+// CountFailedExecuteEvents counts the number of failed execute script events
+// for a host
+func CountFailedExecuteEvents(id string) (int, error) {
+	filter := ResourceTypeKeyIs(ResourceTypeHost)
+	filter[ResourceIdKey] = id
+	filter[TypeKey] = EventHostScriptExecuteFailed
```
Kimchelly, line 95: It seems like we would want this to only look through the N most recent events for the host to look for repeated failures, right? Otherwise, a long-lived static host could accumulate these events over time and will always return lots of events.

rest/route/host_agent.go
```diff
@@ -153,6 +157,17 @@ func (h *hostAgentNextTask) Run(ctx context.Context) gimlet.Responder {
 				"agent":         evergreen.AgentVersion,
 				"current_agent": h.host.AgentRevision,
 			}))
+			event.LogHostScriptExecuteFailed(h.host.Id, err)
```
Kimchelly, line 160: Since this is a pretty specific event case for quarantined hosts, can we use a more specific event log type like `LogHostAgentCleanupFailed` rather than reuse the host script event? There are other host scripts that run which are unrelated to this one.

If we go this route, we also have to file a UI ticket to add it to the host event logs.

rest/route/host_agent.go
```diff
@@ -153,6 +157,17 @@ func (h *hostAgentNextTask) Run(ctx context.Context) gimlet.Responder {
...
+			numExecuteFailedEvents, err := event.CountFailedExecuteEvents(h.host.Id)
+			if err != nil {
+				return gimlet.MakeJSONInternalErrorResponder(err)
+			}
+			if numExecuteFailedEvents > hostUnreachableEventThreshold {
+				err = h.host.CreateSecret(ctx, true)
```
Kimchelly, line 166: nit: can combine L166-167.

rest/route/host_agent.go
```diff
@@ -28,6 +28,10 @@ import (
 // if a host encounters more than this number of system failures, then it should be disabled.
 const consecutiveSystemFailureThreshold = 3
 
+// if we fail to execute a script on a host more than this number of times, we can consider the
```
Kimchelly, line 31: nit: fail to clean up the agent on a quarantined host

Kimchelly (changes requested):

rest/route/host_agent.go
hadjri: I've made a change to instead introduce a new counter field on the host that gets reset to zero when its running task gets cleared.

Kimchelly:
Sorry, I thought about this a little more and remembered that you also have to reset this field back to 0 once the host is unquarantined.

hadjri:
> Sorry, I thought about this a little more and remembered that you also have to reset this attempt counter field back to 0 once the host is unquarantined.

Re-requesting review because of this

model/host/db.go
```diff
@@ -1376,6 +1375,25 @@ func (h *Host) IncrementNumAgentCleanupFailures(ctx context.Context) error {
...
+
+// UnsetNumAgentCleanupFailures unsets the NumAgentCleanupFailures field.
+func (h *Host) UnsetNumAgentCleanupFailures(ctx context.Context) error {
+	err := UpdateOne(
+		ctx,
+		bson.M{
+			IdKey: h.Id,
+		},
+		bson.M{
+			"$unset": bson.M{
```
Kimchelly, line 1393: Might be slightly better to `$set` it explicitly to 0 rather than delete the field entirely since the data model doesn't have `omitempty` for the BSON struct tags.

Kimchelly (approved):
LGTM mod one trivial suggestion.

PR #7541, opened 2024-02-16
DEVPROD-939: Create parsley package and add ParsleyFilters field to DBUser struct 

minnakt:
DEVPROD-939

### Description
This PR adds `ParsleyFilters` to the `User` DB model. 

I wanted to reuse the definition for `ParsleyFilter` that we already have in `model/project_ref.go`, but this would cause an import cycle. For this reason, I moved the definition for `ParsleyFilter` to a new folder called `model/parsley`. Feel free to let me know if there's a better place for this!

### Testing
- Added GraphQL tests to fetch parsley filters for user


graphql/user_resolver.go
```diff
@@ -6,10 +6,28 @@ import (
...
+	"github.com/evergreen-ci/evergreen/model/user"
 	restModel "github.com/evergreen-ci/evergreen/rest/model"
 	"github.com/evergreen-ci/utility"
 )
 
+// ParsleyFilters is the resolver for the parsleyFilters field.
+func (r *userResolver) ParsleyFilters(ctx context.Context, obj *restModel.APIDBUser) ([]*restModel.APIParsleyFilter, error) {
+	parsleyFilters := []*restModel.APIParsleyFilter{}
+
+	usr, err := user.FindOneById(utility.FromStringPtr(obj.UserID))
```
ablack12, line 18: should check if the usr is nil

model/parsley/parsley.go
```diff
@@ -0,0 +1,43 @@
+package parsley
```
ablack12, line 1: I'm typically hesitant to have new packages introduced but i see why this is useful, I think it's separate enough that it won't cause issues

model/parsley/parsley.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"regexp"
+
+	"github.com/mongodb/grip"
+)
+
+// ParsleyFilter represents a filter for the Parsley log viewer. Parsley filters can be defined at
+// the project-level and at the user-level.
+type ParsleyFilter struct {
```
ablack12, line 11: Since this is within a package called Parsley already, it's good practice to just call this struct "Filter" , to reduce naming duplication (`parsley.Filter` as opposed to `parsley.ParsleyFilter`) 
https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name

model/parsley/parsley.go
```diff
@@ -0,0 +1,43 @@
...
+	catcher.NewWhen(p.Expression == "", "filter expression must be non-empty")
+
+	if _, regexErr := regexp.Compile(p.Expression); regexErr != nil {
+		catcher.Wrapf(regexErr, "filter expression '%s' is invalid regexp", p.Expression)
+	}
+
+	return catcher.Resolve()
+}
+
+// ValidateParsleyFilters checks that there are no duplicate expressions among the Parsley filters. It also validates
```
ablack12, line 28: Would it not be okay to have multiple filters with the same expression that have different sensitivity / match? Should we check that those aren't the same as well?

model/parsley/parsley.go
```diff
@@ -0,0 +1,43 @@
...
+// each individual Parsley filter.
+func ValidateParsleyFilters(parsleyFilters []ParsleyFilter) error {
```
ablack12, line 30: similarly, this can just be ValidateFilters

ablack12 (changes requested):

model/parsley/parsley_test.go
```diff
@@ -0,0 +1,63 @@
...
+			ExactMatch:    true,
+		},
+		{
+			Expression:    "def",
+			CaseSensitive: false,
+			ExactMatch:    false,
+		},
+	}
+	err = ValidateParsleyFilters(parsleyFilters)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
minnakt: added!

model/parsley/parsley.go
minnakt: Reading back to my tech design, I'm not sure why I introduced such a strict constraint. Possibly because it was discussed when we reviewed [EVG-18730](https://jira.mongodb.org/browse/EVG-18730) in design sync -- I agree that it seems not necessary though so I removed it

ablack12 (approved):

PR #7542, opened 2024-02-16
DEVPROD-4220 clean up system failure auto restarts

ablack12:
DEVPROD-4220
### Description
I think it makes sense to retain this option long-term so we can temporarily toggle it on in case of build issues or app bugs (such as [DEVPROD-4517](https://jira.mongodb.org/browse/DEVPROD-4517)). Added this as a service flag so it's considered more of a degraded mode option, with one restart max, but I could be convinced to move this to Misc and keep it a configurable integer. 

I also did some drive by cleanup that I'll comment on in-line.

### Testing
Unit tests.


model/task_lifecycle.go
```diff
@@ -293,21 +293,17 @@ func TryResetTask(ctx context.Context, settings *evergreen.Settings, taskId, use
...
-	if evergreen.IsCommitQueueRequester(t.Requester) && evergreen.IsSystemFailedTaskStatus(t.Status) {
-		maxSystemFailedTaskRetries := settings.CommitQueue.MaxSystemFailedTaskRetries
-		if maxSystemFailedTaskRetries > 0 {
-			maxExecution = maxSystemFailedTaskRetries
-		}
-	}
-	// if we've reached the max number of executions for this task, mark it as finished and failed
-	if t.Execution >= maxExecution {
-		// restarting from the UI bypasses the restart cap
-		msg := fmt.Sprintf("task '%s' reached max execution %d: ", t.Id, maxExecution)
```
ablack12, line 305: This logging doesn't seem useful, and seems like something we log when we _do_ want to do this. Unless we intend to be alerting on this behavior, I'm inclined to remove it in favor of better logging practices.

model/task_lifecycle.go
```diff
@@ -293,21 +293,17 @@ func TryResetTask(ctx context.Context, settings *evergreen.Settings, taskId, use
...
-		if origin == evergreen.UIPackage || origin == evergreen.RESTV2Package || origin == evergreen.GithubCheckRun {
```
ablack12, line 306: I probably did this but I see no reason for GithubCheckRuns to receive an exception.

ablack12: Oh no wait I didn't, this is behavior for the new project. I assume this should also be checked for the other checks below as well @bynn ? Should I just add this to UserTriggeredOrigins?

ablack12:
evergreen help

model/task_lifecycle.go
```diff
@@ -293,21 +293,17 @@ func TryResetTask(ctx context.Context, settings *evergreen.Settings, taskId, use
...
-			grip.Debugln(msg, "allowing exception for", user)
-		} else if !t.IsFinished() {
+	// For system failures, we restart once for tasks on their first execution, if configured.
+	if evergreen.IsSystemFailedTaskStatus(t.Status) && t.Execution == 0 &&
```
hadjri, line 297: I think it would make more sense to have `evergreen.IsSystemFailedTaskStatus(t.Status) && t.Execution == 0 && settings.ServiceFlags.SystemFailedTaskRestartDisabled` be its own bool value to condition on at L304, rather than changing the maxExecution value to one.

I think this would be easier to understand since the desicion we're making here is a binary one (i.e. we are either restarting it once or we aren't)

hadjri (changes requested):
Since this is more of a degraded mode option, I think it makes sense to keep it as a boolean flag rather than a configurable int, for simplicity 👍

model/task_lifecycle.go
ablack12: I thought about that, but I think that's actually not what this boolean means with the way it's implemented -- really it only restarts if the _first_ execution failed. If we just check the boolean at 304 without considering execution or adding some new state tracking, we'll always restart bc we won't know if we've auto-restarted once or not.

I'm not inclined to add state tracking for this since ideally we usually won't be in this degraded mode.

model/task_lifecycle.go
bynn: Yes, since this is manually triggered by users clicking a button on the github UI!

model/task_lifecycle.go
hadjri: But I think I think we should be setting maxExecution = 1 for all system failed tasks and not only for tasks where `t.Execution == 0` correct?

hadjri (changes requested):
Sorry adding comment to indicate review

model/task_lifecycle.go
ablack12: Also I think the way we're checking system failure here is outdated, but hasn't mattered since https://github.com/evergreen-ci/evergreen/blob/main/model/task_lifecycle.go#L2223 does a check also

ablack12: I'm actually not certain we need this additional check but i think it doesn't hurt

ablack12:
evergreen help

PR #7543, opened 2024-02-16
DEVPROD-966 Save check run ID on Task

malikchaya2:
[DEVPROD-966]

### Description
Save check run ID on Task. This will be need for future executions when the same check run will be updated. 

### Testing
This will be tested as part of DEVPROD-970


model/task/task.go
```diff
@@ -179,6 +179,10 @@ type Task struct {
 	// CheckRunPath is a local file path to an output json file for the checkrun.
 	CheckRunPath string `bson:"check_run_path" json:"check_run_path"`
 
+	// CheckRunId is the id for the checkrun that was created in github.
+	// This is used to update the checkrun for future executions of the task.
+	CheckRunId *int `bson:"check_run_id,omitempty" json:"check_run_id,omitempty"`
```
Kimchelly, line 184: Can the check run ID be 0? If it's impossible for it to be 0, I feel like not using a pointer (and just checking `CheckRunId == 0` instead) might be slightly preferable to lots of nil checking when we're not sure if it's set.

model/task/task.go
```diff
@@ -3448,6 +3452,21 @@ func (t *Task) SetDisplayTaskID(id string) error {
 		}}))
 }
 
+// SetCheckRunId sets the checkRunId for the task
+func (t *Task) SetCheckRunId(checkRunId int) error {
+	t.CheckRunId = utility.ToIntPtr(checkRunId)
+	return UpdateOne(
```
Kimchelly, line 3458: It's slightly more preferable to `UpdateOne` before setting the field in memory:
```
if err := UpdateOne(<update>...); err != nil {
    return err
}

t.CheckRunId = checkRunId
return nil
```
This avoids a general issue where the task in-memory (i.e. `t`) has a different state from the DB if the update fails. It's caused some bugs in the past where an update failed but `t` was updated, so the logic in the outer function acted on wrong information.

model/task/task.go
```diff
@@ -3448,6 +3452,21 @@ func (t *Task) SetDisplayTaskID(id string) error {
 		}}))
 }
 
+// SetCheckRunId sets the checkRunId for the task
+func (t *Task) SetCheckRunId(checkRunId int) error {
```
Kimchelly, line 3456: Is this supposed to be unit tested? Or is this also going to be tested as part of DEVPROD-970?

Or is the `checkRunHandler` already tested, maybe? Can we add to that test?

rest/route/agent.go
```diff
@@ -1502,5 +1504,8 @@ func (h *checkRunHandler) Run(ctx context.Context) gimlet.Responder {
 		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "setting patch parameters"))
 	}
 
+	checkRunInt := int(utility.FromInt64Ptr(checkRun.ID))
+	err = t.SetCheckRunId(checkRunInt)
```
Kimchelly, line 1508: Is this error supposed to do something?

Kimchelly (changes requested):

model/task/task.go
malikchaya2: I don't know... 
I want to assume that not, but all the docs say is: 
image

and they use the example of ` "id": 4,` in a response. So I don't feel like I can rely on them not being 0.

model/task/task.go
malikchaya2: added a unit test

model/task/task_test.go
```diff
@@ -3496,6 +3496,23 @@ func TestAddParentDisplayTasks(t *testing.T) {
...
+	}
+
+	assert.NoError(t, t1.Insert())
+	assert.NoError(t, t1.SetCheckRunId(12345))
+
+	var err error
+	t1, err = FindOneId(t1.Id)
+	assert.NoError(t, err)
+
+	assert.Equal(t, t1.CheckRunId, utility.ToIntPtr(12345))
```
Kimchelly, line 3512: nit: it's usually considered better to compare values (which seems like what you're trying to do) rather than pointers because pointers are just meaningless addresses, so if go so happened to decide to give them the same address, it's not something we should rely on. This should instead be:
```go
assert.Equal(t, 12345, utility.FromIntPtr(t1.CheckRunId))
```
(also note the flipped arguments because the first value is always the expected value and the second is the actual value).

model/task/task_test.go
```diff
@@ -3496,6 +3496,23 @@ func TestAddParentDisplayTasks(t *testing.T) {
...
+	require.NoError(t, db.Clear(Collection))
+	t1 := &Task{
+		Id: "t1",
+	}
+
+	assert.NoError(t, t1.Insert())
+	assert.NoError(t, t1.SetCheckRunId(12345))
+
+	var err error
+	t1, err = FindOneId(t1.Id)
```
Kimchelly, line 3509: Have to `require.NotNil(t, t1)` or else this test can panic if it doesn't find the task.

rest/route/agent.go
```diff
@@ -1496,16 +1496,25 @@ func (h *checkRunHandler) Run(ctx context.Context) gimlet.Responder {
...
 			"task_id": t.Id,
 		})
-		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "setting patch parameters"))
+		return gimlet.MakeJSONInternalErrorResponder(errors.Wrap(err, "creating check run"))
 	}
 
 	checkRunInt := int(utility.FromInt64Ptr(checkRun.ID))
-	err = t.SetCheckRunId(checkRunInt)
+	if err = t.SetCheckRunId(checkRunInt); err != nil {
+		errorMessage := fmt.Sprintf("setting checkRun ID on task: %s", t.Id)
```
Kimchelly, line 1510: nit: seems like the grip error log already has the task ID, so it could be simplified down to:
```go
err = errors.Wrap(err, "setting check run ID on task")
grip.Error(message.WrapError(err, message.Fields{ /* task ID goes here */ }))
return gimlet.MakeJSONInternalErrorResponder(err)
```

Kimchelly (approved):
LGTM mod nits.

PR #7550, opened 2024-02-20
DEVPROD-2251 Add JSON output flag for evergreen patch command

ZackarySantana:
DEVPROD-2251

### Description
The flag requires that the user also include the skip-confirm flag (parsing the output with the confirmation dialogs we prompt would be difficult to support and just adds overhead for users).

As well, this consolidates the list-patches and patch command to use the same code when generating output. This also touches some of the commit-queue command because it updates a function signature.

### Testing
I ran:
* `evergreen list-patches` gives generic output
* `evergreen list-patches -j` gives JSON output (an array of API Patches)

* `evergreen patch` and going through the confirmations, gives generic output
* `evergreen patch -j` gives an error stating it must be used with --skip_confirm
* `evergreen patch -j --skip_confirm` gives JSON output (a single API patch)

### Documentation
Flag usage (no other docs needed)

### DIFF
--- a/operations/commit_queue.go
+++ b/operations/commit_queue.go
@@ -448,12 +448,12 @@ func backport() cli.Command {
 				return errors.Wrap(err, "uploading backport patch")
 			}
 
-			o := outputPatchParams{
-				patch:  backportPatch,
-				uiHost: uiV2,
+			params := outputPatchParams{
+				patches: []patch.Patch{*backportPatch},
+				uiHost:  uiV2,
 			}
 
-			if err = patchParams.displayPatch(ac, o); err != nil {
+			if err = patchParams.displayPatch(ac, params); err != nil {
 				return errors.Wrap(err, "getting result display")
 			}
 
@@ -518,11 +518,11 @@ func listCLICommitQueueItem(item restModel.APICommitQueueItem, ac *legacyClient,
 	if p.Author != "" {
 		grip.Infof("Author: %s", p.Author)
 	}
-	o := outputPatchParams{
-		patch:  p,
-		uiHost: uiServerHost,
+	params := outputPatchParams{
+		patches: []patch.Patch{*p},
+		uiHost:  uiServerHost,
 	}
-	disp, err := getPatchDisplay(ac, o)
+	disp, err := getPatchDisplay(ac, params)
 	if err != nil {
 		grip.Error(errors.Wrapf(err, "getting patch display summary for patch '%s'", p.Id.Hex()))
 		return
@@ -648,19 +648,19 @@ func (p *mergeParams) uploadMergePatch(conf *ClientSettings, ac *legacyClient, u
 	if err = patchParams.validateSubmission(diffData); err != nil {
 		return err
 	}
-	patch, err := patchParams.createPatch(ac, diffData)
+	newPatch, err := patchParams.createPatch(ac, diffData)
 	if err != nil {
 		return err
 	}
-	o := outputPatchParams{
-		patch:  patch,
-		uiHost: uiV2Url,
+	params := outputPatchParams{
+		patches: []patch.Patch{*newPatch},
+		uiHost:  uiV2Url,
 	}
-	if err = patchParams.displayPatch(ac, o); err != nil {
+	if err = patchParams.displayPatch(ac, params); err != nil {
 		grip.Error("Patch information cannot be displayed.")
 	}
 
-	p.id = patch.Id.Hex()
+	p.id = newPatch.Id.Hex()
 	patchParams.setDefaultProject(conf)
 
 	return nil
@@ -773,10 +773,10 @@ func getAPICommitQueuePatchDisplay(ac *legacyClient, apiPatch *restModel.APIPatc
 		return "", errors.Wrap(err, "converting patch to service model")
 	}
 
-	o := outputPatchParams{
-		patch:     &servicePatch,
+	params := outputPatchParams{
+		patches:   []patch.Patch{servicePatch},
 		summarize: summarize,
 		uiHost:    uiHost,
 	}
-	return getPatchDisplay(ac, o)
+	return getPatchDisplay(ac, params)
 }
--- a/operations/commit_queue.go
+++ b/operations/commit_queue.go
@@ -448,7 +448,12 @@ func backport() cli.Command {
 				return errors.Wrap(err, "uploading backport patch")
 			}
 
-			if err = patchParams.displayPatch(ac, backportPatch, uiV2, false); err != nil {
+			o := outputPatchParams{
+				patch:  backportPatch,
+				uiHost: uiV2,
+			}
+
+			if err = patchParams.displayPatch(ac, o); err != nil {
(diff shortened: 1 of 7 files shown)


operations/commit_queue.go
```diff
@@ -448,7 +448,12 @@ func backport() cli.Command {
 				return errors.Wrap(err, "uploading backport patch")
 			}
 
-			if err = patchParams.displayPatch(ac, backportPatch, uiV2, false); err != nil {
+			o := outputPatchParams{
```
ablack12, line 451: nit: can you call this params instead? `o` isn't a typical variable name (trying to figure out why I think this -- typically variables should be descriptive, so "params" is closer, or if it's going to be abbreviated  down to one letter, it should be on the noun, not the descriptive word so "p" is closer than "o")

operations/patch_util.go
```diff
@@ -174,13 +176,15 @@ func (p *patchParams) validateSubmission(diffData *localDiff) error {
...
-	patchDisp, err := getPatchDisplay(ac, newPatch, p.ShowSummary, uiHost, isCommitQueuePatch)
+func (p *patchParams) displayPatch(ac *legacyClient, o outputPatchParams) error {
+	patchDisp, err := getPatchDisplay(ac, o)
 	if err != nil {
 		return err
 	}
 
-	grip.Info("Patch successfully created.")
+	if !o.outputJSON {
+		grip.Info("Patch successfully created.")
```
ablack12, line 186: you can use grip.InfoWhen here

operations/patch_util.go
```diff
@@ -524,23 +528,74 @@ func validatePatchSize(diff *localDiff, allowLarge bool) error {
 	return nil
 }
 
+type outputPatchParams struct {
+	patches []patch.Patch
```
ablack12, line 532: I'm not a huge fan of having both patches and patch be options for these params. I think it can lead to bugs and has strange implications, like below where you have to assign "patch" for each call of getPatchDisplay to essentially point to what you want.

While I'm all for modularizing functions, getPatchesDisplay doesn't do _that_ much more than getPatchDisplay (actually it really just seems to do the same thing? It loops through the patches and then calls the same thing as 591-598 it looks like).

Can these functions just be combined into one getPatchesDisplay? A list of one patch can be passed in for the `evergreen patch` command.

ablack12: Feel free to reach out if you want to discuss this more

operations/patch.go
```diff
@@ -232,7 +238,9 @@ func Patch() cli.Command {
 						continue
 					}
 					if err = addModuleToPatch(params, args, conf, newPatch, &module, modulePath); err != nil {
-						grip.Errorf("Error adding module '%s' to patch: %s", module.Name, err)
+						if !outputJSON {
```
ablack12, line 241: grip.ErrorWhen

operations/flags.go
```diff
@@ -158,6 +158,13 @@ func addSyncStatusesFlag(flags ...cli.Flag) []cli.Flag {
 	})
 }
 
+func addJSONOutputFlag(flags ...cli.Flag) []cli.Flag {
+	return append(flags, cli.BoolFlag{
+		Name:  joinFlagNames(jsonFlagName, "j"),
+		Usage: "output JSON instead of text",
```
ablack12, line 164: I think it would be good if the patch version of this flag mentioned that warnings are suppressed and that it has to be used with -y, so we may not want to have a variable for this flag

ablack12 (changes requested):

operations/commit_queue.go
ZackarySantana: My only concern was that in this particular file we call the 'patchParams' as 'params'- But the usage is pretty small so I actually think it won't be confusing- i'll switch it

operations/patch_util.go
ZackarySantana: That's fair. The only problem is that output json and outputing our generic output is a little different. When output json for a single or multiple, it should be either an array or a single object. While our generic output just has the same output repeated. I'll consolidate these two and make it have some checks I think

operations/flags.go
ZackarySantana: That makes sense, I'll switch it to be standalone and include that

operations/commit_queue.go
ablack12: in that case we could go so far as to say outputParams for the variable -- better to have clarity than keep it short

operations/patch_util.go
```diff
@@ -174,24 +176,26 @@ func (p *patchParams) validateSubmission(diffData *localDiff) error {
 	return nil
 }
 
-func (p *patchParams) displayPatch(ac *legacyClient, newPatch *patch.Patch, uiHost string, isCommitQueuePatch bool) error {
-	patchDisp, err := getPatchDisplay(ac, newPatch, p.ShowSummary, uiHost, isCommitQueuePatch)
+// displayPatch outputs the given patch(es) to the user. If there is only one patch,
+// and browse is true, it will open the patch in the user's default web browser.
+func (p *patchParams) displayPatch(ac *legacyClient, o outputPatchParams) error {
```
ablack12, line 181: function signatures should also use p or params instead of o

operations/patch.go
```diff
@@ -42,6 +43,10 @@ func getPatchFlags(flags ...cli.Flag) []cli.Flag {
 		addUncommittedChangesFlag(),
 		addReuseFlags(),
 		addPreserveCommitsFlag(
+			cli.BoolFlag{
+				Name:  joinFlagNames(jsonFlagName, "j"),
+				Usage: "outputs the patch as a JSON object and surpresses warnings, must be used with --skip-confirm",
```
ablack12, line 48: nit: *suppresses

operations/patch.go
```diff
@@ -89,8 +94,8 @@ func Patch() cli.Command {
 			setPlainLogger,
 			mutuallyExclusiveArgs(false, patchDescriptionFlagName, autoDescriptionFlag),
 			mutuallyExclusiveArgs(false, preserveCommitsFlag, uncommittedChangesFlag),
-			mutuallyExclusiveArgs(false, repeatDefinitionFlag, repeatPatchIdFlag,
-				repeatFailedDefinitionFlag),
+			mutuallyExclusiveArgs(false, repeatDefinitionFlag, repeatPatchIdFlag, repeatFailedDefinitionFlag),
+			conditionallyRequiredArgs(jsonFlagName, skipConfirmFlagName),
```
ablack12, line 98: I don't feel too strongly about this but could we just enforce this ourselves? I think users would expect confirmation to be skipped, and we can mention it in the help text. Code-wise we can just set skipConfirm to true if the json flag is set so it won't change much.

ablack12 (approved):
Approving as i'm out tomorrow but if you think the comments need another look i'll be back monday or you can pull in Malik!

operations/patch.go
ZackarySantana: Hm that's true, my thinking was that if we do do that, we can't inform the user that we have done so. I'll switch it to silently set it since I agree that it would be an expectation!

PR #7554, opened 2024-02-20
DEVPROD-3246 Add host/pod id to agent panics

ZackarySantana:
DEVPROD-3246

### Description
Adds the host id to the agent panic logs to help ops

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1189,6 +1189,7 @@ func (a *Agent) logPanic(logger client.LoggerProducer, pErr, originalErr error, 
 		logMsg := message.Fields{
 			"message":   "programmatic error: Evergreen agent hit panic",
 			"operation": op,
+			"host_id":   a.ec2InstanceID,
 		}
 		logger.Task().Error(logMsg)
 	}
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1189,7 +1189,9 @@ func (a *Agent) logPanic(logger client.LoggerProducer, pErr, originalErr error, 
 		logMsg := message.Fields{
 			"message":   "programmatic error: Evergreen agent hit panic",
 			"operation": op,
-			"host_id":   a.ec2InstanceID,
+			"host_id":   a.opts.HostID,
+			"pod_id":    a.opts.PodID,
+			"ec2_id":    a.ec2InstanceID,
 		}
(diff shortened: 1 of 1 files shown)


agent/agent.go
```diff
@@ -1189,6 +1189,7 @@ func (a *Agent) logPanic(logger client.LoggerProducer, pErr, originalErr error,
 		logMsg := message.Fields{
 			"message":   "programmatic error: Evergreen agent hit panic",
 			"operation": op,
+			"host_id":   a.ec2InstanceID,
```
Kimchelly, line 1192: While this is generally fine, it would be better to pass [the agent's host ID](https://github.com/evergreen-ci/evergreen/blob/70172791d8c8b43b56506502ea75aaef7d204f08/agent/agent.go#L59) here because 1. the EC2 instance ID only applies to EC2-spawnable hosts (so it won't be set for MacOS) and 2. we aren't guaranteed to be able to get `a.ec2InstanceID` since it's we have to `GET` it from an endpoint, whereas the host ID I linked is always populated when the agent starts since it's required to auth back to Evergreen.

Kimchelly (approved):
The proposed change LGTM mod 1. a recommendation to use an alternative field that covers more cases + 2. bump the agent version.

agent/agent.go
ZackarySantana: I kept the ec2 id (renamed to ec2_id), added the host_id and pod_id. I'm not sure if this has redundant id's (I haven't done much debugging with our agents) but looking at the options struct it seemed like it may help.

Lmk if they look good to you!

Kimchelly: The EC2 instance ID is the same as the host ID if it's an EC2 host, so I don't think we need `ec2_id` here. As for host ID vs pod ID, they're mutually exclusive, so I think maybe we can just add some conditional logic to only add them to the log if they're non-empty. Something like:
```go
logMsg := message.Fields { /* the original stuff... */ }
if a.opts.HostID != "" {
    logMsg["host_id"] = a.opts.hostID
}
if a.opts.PodID != "" {
    logMsg["pod_id"] = a.opts.PodID
}
```

ZackarySantana: Ah makes sense! I thought they may be the same but wasn't sure. I'll add that logic in ty!

PR #7555, opened 2024-02-20
DEVPROD-3873 Allow container image to use project variable

hadjri:
DEVPROD-3873

### Description
Allow container image URIs to be expanded from project variables. This will allow staging and prod to reference their own ECR repositories and get the staging container variant green again.
### Testing
Tested in staging. The PR checks here can't run, so I scheduled [this patch](https://spruce.mongodb.com/version/65d7b8989ccd4eb556cbb23a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) in prod to represent the build state

### DIFF
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2884,7 +2884,7 @@ func ValidateContainers(ecsConf evergreen.ECSConfig, pRef *ProjectRef, container
 		image := container.Image
 		if expansions != nil {
 			image, err = expansions.ExpandString(container.Image)
-			catcher.Add(errors.Wrap(err, "expanding container image"))
+			catcher.Wrap(err, "expanding container image")
 		}
 		catcher.Add(container.System.Validate())
 		if container.Resources != nil {
--- a/model/project_ref.go
+++ b/model/project_ref.go
@@ -2870,7 +2870,22 @@ func GetMessageForPatch(patchID string) (string, error) {
 // on the project admin page.
 func ValidateContainers(ecsConf evergreen.ECSConfig, pRef *ProjectRef, containers []Container) error {
 	catcher := grip.NewSimpleCatcher()
+
+	projVars, err := FindMergedProjectVars(pRef.Id)
+	if err != nil {
+		return errors.Wrapf(err, "getting project vars for project '%s'", pRef.Id)
+	}
+	var expansions *util.Expansions
+	if projVars != nil {
+		expansions = util.NewExpansions(projVars.Vars)
+	}
+
 	for _, container := range containers {
+		image := container.Image
+		if expansions != nil {
+			image, err = expansions.ExpandString(container.Image)
+			catcher.Add(errors.Wrap(err, "expanding container image"))
+		}
 		catcher.Add(container.System.Validate())
 		if container.Resources != nil {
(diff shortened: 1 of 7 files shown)


model/project_ref.go
```diff
@@ -2870,7 +2870,22 @@ func GetMessageForPatch(patchID string) (string, error) {
...
+	var expansions *util.Expansions
+	if projVars != nil {
+		expansions = util.NewExpansions(projVars.Vars)
+	}
+
 	for _, container := range containers {
+		image := container.Image
+		if expansions != nil {
+			image, err = expansions.ExpandString(container.Image)
+			catcher.Add(errors.Wrap(err, "expanding container image"))
```
Kimchelly, line 2887: nit: this can be simplified down to `catcher.Wrap(err, "expanding container image")`

units/pod_allocator.go
```diff
@@ -231,6 +233,16 @@ func (j *podAllocatorJob) populate(ctx context.Context) error {
 		j.pRef = pRef
 	}
 
+	if j.expansions == nil {
```
Kimchelly, line 236: nit: if we don't plan to use `j.expansions` outside of this populate function, can we just make it a local variable like how it's done in `model/project_ref.go`? These expansions seem like they're meant for a pretty specific purpose (which is to expand the container image).

Or did you plan to test the expansion behavior and this field makes it easier?

validator/project_validator_test.go
```diff
@@ -4394,6 +4394,7 @@ func TestValidateContainers(t *testing.T) {
 			}
 
 			ref := &model.ProjectRef{
+				Id:         "proj",
```
Kimchelly, line 4397: GH doesn't let me comment on unchanged lines, but should L4381 also clear the project vars collection?

Kimchelly (changes requested):
* Should the `Image` field itself have the `plugin:"expand"` struct tag? It's not actually necessary to make the logic work since this just calls `ExpandString` directly, but it would be consistent with other expandable fields in project commands and might make it a bit clearer that the value is special and can be expanded.
* Should we document that container images can be expanded from the project vars? At least for ourselves so we know it's intentional behavior.

units/pod_allocator.go
hadjri: I did this as an attempt to group the vars fetching with all of the other DB operations this job uses in its `populate` function for organizational purposes, but since it's used in only one spot like you said I'm fine to change it to be a local variable.

validator/project_validator_test.go
hadjri: Did this + modified this test to assert that the image expansion works correctly

hadjri:
> Should the Image field itself have the plugin:"expand" struct tag? It's not actually necessary to make the logic work since this just calls ExpandString directly, but it would be consistent with other expandable fields in project commands and might make it a bit clearer that the value is special and can be expanded.

Makes sense, added the plugin tag or Image.

> Should we document that container images can be expanded from the project vars? At least for ourselves so we know it's intentional behavior.

Made a change to the docs that uses an expansion for the image as an example and added a note specifying that the field can be expanded.

Kimchelly (approved):
Nice job!

PR #7561, opened 2024-02-21
DEVPROD-4616 Add host service password to private variables

hadjri:
DEVPROD-4616

### Description
Hosts' service passwords were being leaked in the system logs on windows distros. Changed the get expansions and vars route to retrieve the host's (if any) service password and add it as a private var so that the redacting sender in the agent will know to redact it from the logs.

### Testing
Updated unit test to check that the new expansion is set as a private variable when retrieving expansions and vars for a task.

Tested in [staging](https://evergreen-staging.corp.mongodb.com/task_log_raw/test_trigger_sandbox_ubuntu2004_patch_only_task_patch_25dd3ef48386f68d6dfd619d0f17b5cc6e8f8516_65d65805499f72000734afbe_24_02_21_20_07_50/2?type=ALL#L3521) that the password is now properly redacted.


rest/route/agent.go
```diff
@@ -448,6 +448,10 @@ func (h *getExpansionsAndVarsHandler) Run(ctx context.Context) gimlet.Responder
 			res.PrivateVars = projectVars.PrivateVars
 		}
 	}
+	if foundHost != nil {
+		res.Vars[evergreen.HostServicePasswordExpansion] = foundHost.ServicePassword
```
ablack12, line 452: This seems really niche. Why did we need to do this? Did this come from a project variable added on the distro page? (Do you have an example?) Can we just add the option to make those private?

ablack12: Or can we add this variable to the admin list of things we redact as opposed to hardcoding here?

ablack12 (changes requested):

rest/route/agent.go
hadjri: For windows hosts, a random string `ServicePassword` field [gets set for them](https://gist.github.com/hadjri/134be02df6d49577ea8a818f7d776fee#file-hostutil-go-L538-L548), which gets passed as a [parameter](https://gist.github.com/hadjri/52e2ede38dec91a0101d35a4ea08cc70#file-hostutil-go-L293) to a jasper startup command on the host.  When this command runs on the host it reveals the password in the system logs.

So the issue is that this password isn't a project variable set anywhere, but a field on the host document itself and is unique and dynamically generated for every windows host. The way we redact things in the logs is through variables that are either explicitly marked as private or regex match with some `redact_keys` we have set on the admin page which you mentioned. So although this is niche and only applies to windows hosts, putting the host password in a private variable in this REST route is the only straightforward way I can think to achieve redacting this.

ablack12: Okay I see the problem. It might be more consistent to add to [the default expansions](https://github.com/evergreen-ci/evergreen/blob/ae19a2e10b32961985811d525f1ef06ff687fc51/model/project.go#L1108) and then adding adding explicitly to https://github.com/evergreen-ci/evergreen/blob/a2d4d5ad98bb8d6bd5620d8fa88c589fb1060f6a/agent/command/expansion_write.go#L18 , or at least adding a good comment here to explain -- I'm just concerned that this is pretty niche, and we're going to forget what's going on.

I'm also curious where we're leaking this, shouldn't it be from our own code? Should we make any additional adjustments? (Like should we not log here? https://github.com/evergreen-ci/evergreen/blob/ba45f61e3337a856f6a7aaef4ebd553be2980094/model/host/hostutil.go#L558)

ablack12 (changes requested):
Oops adding another comment to indicate reviewed

rest/route/agent.go
hadjri: Sounds good, moved to this ExpansionsToRedact list and into the default expansions. Confirmed in [staging](https://evergreen-staging.corp.mongodb.com/task_log_raw/test_trigger_sandbox_ubuntu2004_patch_only_task_patch_25dd3ef48386f68d6dfd619d0f17b5cc6e8f8516_65d65805499f72000734afbe_24_02_21_20_07_50/3?type=ALL) that this works. Per your second comment, I think it's fine because this will get auto-redacted in the logs now with this change. The original place we were leaking this secret was this [line](https://gist.github.com/hadjri/edd383260e7f6282290db0faa0ed7fb5#file-hostutil-go-L293).

ablack12: hmm okay I think auto-redaction is good but we should try not to depend on it; I was hoping we could do something [like this](https://github.com/evergreen-ci/evergreen/blob/26c086e9bd295bf13d87add1e9f0e5da107ddc35/agent/command/git.go#L241) but I think this case might be more nested

ablack12 (approved):

PR #7562, opened 2024-02-22
DEVPROD-911 Warn when tasks and variants are not specified in CLI

bynn:
DEVPROD-911 🚨 🚨 🚨 

### Description
CLI now warns if you pass in a task without a variant or vise versa

### Testing
local cli 
image

### Documentation
Documented that both tasks and variants must explicitly be specified


docs/CLI.md
```diff
@@ -41,6 +41,13 @@ Tasks and variants can also be specified using the regex variants(-rv) and regex
 evergreen patch --regex_variants "enterprise.*" --rt "test-.*"
 ```
 
+Tasks and variants must _both_ be specified:
+```bash
+evergreen patch --regex_variants "enterprise.*" // not valid, will not select any tasks
+evergreen patch -t unittest // not valid, will not select any tasks
+evergreen patch --regex_variants "enterprise.*" -t unittest // valid
```
malikchaya2, line 48: nit, can we use `regex_tasks` instead of `-t` to match it to `regex_variants`?

docs/CLI.md
```diff
@@ -41,6 +41,13 @@ Tasks and variants can also be specified using the regex variants(-rv) and regex
 evergreen patch --regex_variants "enterprise.*" --rt "test-.*"
 ```
 
+Tasks and variants must _both_ be specified:
```
malikchaya2, line 44: Can we say something like "When specifying tasks and variants" or something like that at the beginning? Because it doesn't have to be specified, they can specify nothing and just do `evergreen patch -u`.

malikchaya2 (approved):
approved with nits

docs/CLI.md
bynn: I purposely didn't match them to show that this works even if you mix regex tasks and variants. I'll add another example with both regex tasks and variants though.

PR #7563, opened 2024-02-22
DEVPROD-4989: rollover agent for new mci key

julianedwards:
DEVPROD-4989

### Description
We rotated the key so we need to rollover the agents to get the new key.

### Testing
N/A

### Documentation
N/A


Kimchelly (approved):
Worth mentioning that even though there's no code change, what we really need is for the agents to restart to get the latest agent setup data. It only gets this data when it first starts up.

PR #7564, opened 2024-02-22
DEVPROD-3947: add validation and calculate next stop/start time

Kimchelly:
DEVPROD-3947

### Description
I apologize in advance, the code change itself isn't actually that long (most of the diff is tests) but the algorithm is kind of tricky. Feel free to read the test cases or ask questions if you need clarification.

* Add validation function to ensure sleep schedule provided by user is sensible and follows the basic requirements.
* Add helper functions to calculate the next scheduled stop/start times. This uses [the same cron library](https://pkg.go.dev/github.com/robfig/cron) that we use for implementing the `cron` YAML field to help with the future time calculations.

### Testing
Added lots of unit tests for validation and calculation correctness.

### Documentation
N/A

### DIFF
--- a/model/host/host.go
+++ b/model/host/host.go
@@ -402,9 +402,9 @@ func (i *SleepScheduleInfo) Validate() error {
 		uniqueWeekdays[weekday] = struct{}{}
 	}
 
-	const minNumHoursPerWeek = 24 * time.Hour
+	const minNumHoursPerWeek = utility.Day
 
-	weeklySleep := 24 * time.Hour * time.Duration(len(i.WholeWeekdaysOff))
+	weeklySleep := utility.Day * time.Duration(len(i.WholeWeekdaysOff))
 	if i.DailyStopTime != "" && i.DailyStartTime != "" {
 		// Add up how much time is hypothetically spent sleeping for the daily
 		// sleep schedule.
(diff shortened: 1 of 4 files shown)


model/host/host.go
```diff
@@ -3370,3 +3475,233 @@ func (h *Host) UnsetPersistentDNSInfo(ctx context.Context) error {
...
+func (h *Host) GetNextScheduledStopTime(now time.Time) (time.Time, error) {
+	if h.SleepSchedule.PermanentlyExempt || h.SleepSchedule.ShouldKeepOff {
+		return SleepScheduleSentinelTime, nil
+	}
+
+	if err := h.SleepSchedule.Validate(); err != nil {
+		return time.Time{}, errors.Wrap(err, "invalid sleep schedule")
+	}
+
+	userTimeZone, err := time.LoadLocation(h.SleepSchedule.TimeZone)
```
hadjri, line 3497: We could consider putting a `*time.Location` on the  sleep schedule info itself that gets populated during validate because validate is already running and `LoadLocation` and returning an error if it fails

model/host/host.go
```diff
@@ -3370,3 +3475,233 @@ func (h *Host) UnsetPersistentDNSInfo(ctx context.Context) error {
...
+	// It's critical to do all the time calculations in the user's time zone and
+	// not in UTC. If the time calculations were all in UTC, they wouldn't
+	// account for timezone-specific quirks like daylight savings time changes.
+	calculateTimeAfter = calculateTimeAfter.In(userTimeZone)
+
+	// Partition the weekdays into the first day in a series of whole days off
+	// (if any) and the weekdays when the daily sleep schedule applies (if any).
+	var dailyWeekdays []string
+	var firstContiguousWeekdayOff []string
+	for _, weekday := range []time.Weekday{time.Sunday, time.Monday, time.Tuesday, time.Wednesday, time.Thursday, time.Friday, time.Saturday} {
```
hadjri, line 3539: We can make this a constant since it's used twice

hadjri (changes requested):

model/host/host.go
Kimchelly: Sure, I made some constants in utility and used them here.

hadjri (approved):

PR #7566, opened 2024-02-23
DEVPROD-2299 Add CI task for go vulnerabilities

ZackarySantana:
DEVPROD-2299

### Description
This adds a go task for go vulnerabilities. It adds a script for users to run locally if they would like and a make command to help with it

### Testing
Manual and PR patches


makefile
```diff
@@ -361,6 +361,9 @@ gqlgen:
 	$(gobin) run github.com/99designs/gqlgen generate
 	$(gobin) run cmd/gqlgen/generate_secret_fields.go
 
+govul-install:
```
bsamek, line 364: For discoverability (some shells will autocomplete `make` targets, and people may read the `makefile` themselves), I would suggest adding a target to run `scripts/check-go-vulnerabilities.sh`, and calling that `make` target from `self-tests.yml`.

scripts/check-go-vulnerabiltiies.sh
```diff
@@ -0,0 +1,33 @@
...
+#!/usr/bin/env bash
+
+# This script checks for go vulnerabilities.
+
+govul="govulncheck"
+
+# Test if govulncheck is installed as a binary or a go module.
+if command -v govul &> /dev/null; then
+    # If the user has installed swaggo as a binary, use the binary.
+    swaggo=$(command -v govul)
```
bsamek, line 11: Is the `swaggo` here from a copy/paste maybe?

bsamek, line 11: Is `govul` ever a binary?

bsamek (changes requested):

scripts/check-go-vulnerabiltiies.sh
ZackarySantana: Yup should be switched to govul!

ZackarySantana: No, this should be either the variable or hardcoded. I'll make it hard coded for clarity

makefile
ZackarySantana: I'm not convinced about adding a target for it since to run it locally, it would be the same as all the other verify scripts. The other verify scripts also do not have a target (only one does and it's instead a go command rather than a shell script). I think I should either add all the scripts to the makefile and rewrite the self-tests to use those targets or write none to keep consistency. I'm leaning towards not because it adds length to the makefile.

Alternatively I thought about using a wildcard syntax in the makefile since most follow a "verify-" and then adding only a couple commands but it wouldn't have great autocomplete then. What do you think?

bsamek: That (not adding to the makefile) sounds reasonable to me.

scripts/check-go-vulnerabilities.sh
```diff
@@ -0,0 +1,33 @@
+
```
bsamek, line 1: In case you didn't see, this script is working locally for me, but is failing in CI.

scripts/check-go-vulnerabilities.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+# If govulncheck is not installed, exit with an error.
+if ! command -v "$govul" &> /dev/null; then
+    echo "govulncheck is not installed."
+    exit 1
+fi
+
+result=$($govul ./...)
+if [ $? -eq 0 ]; then
```
bsamek, line 25: It might be nice if this prints a message so that users know no vulnerabilities were found. This distinguishes it from the script crashing or exiting without a useful error message.

bsamek (changes requested):

scripts/check-go-vulnerabilities.sh
ZackarySantana: Yeah... I've been trying to debug it but I'm really not sure why it's giving the error in CI. I tested with the same exact go version and according to the logs it has the error in stderr but then it says "no vulnerabilities found" in stdout. If you have some insight on the tool please lmk! If not no worries, I will continue to debug 🔨

scripts/check-go-vulnerabilities.sh
```diff
@@ -0,0 +1,33 @@
...
+    govul=$(command -v govulncheck)
+elif [ -n "$GOPATH" ] && [ -f "$GOPATH/bin/govulncheck" ]; then
+    govul="$GOPATH/bin/govulncheck"
+elif [ -n "$GOBIN" ] && [ -f "$GOBIN/govulncheck" ]; then
+    govul="$GOBIN/govulncheck"
+fi
+
+# If govulncheck is not installed, exit with an error.
+if ! command -v "$govul" &> /dev/null; then
+    echo "govulncheck is not installed."
```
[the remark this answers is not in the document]
ZackarySantana: I agree, changing it!

scripts/check-go-vulnerabilities.sh
ZackarySantana: I also changed it for the swaggo fmt file to be more consistent/helpful

scripts/check-go-vulnerabilities.sh
bsamek: I wonder if it is some weird shell thing. Maybe try `set -x` at the top of the script to verify that it's doing what you expect?

ZackarySantana:
I figured it out, the go binary govulncheck was pulling from the path was an old version (1.7) which did not have the -compile flag on the 'go list' subcommand (combed through the go versions and the commands- had to go _far_ back).

I added an export line and it works as intended in CI now.

ZackarySantana:
I'm not worried about the failing parsely and spruce tasks, this should have absolutely no effect on them

bsamek (approved):

PR #7567, opened 2024-02-23
DEVPROD-5081 Update Deepcopy to use JSON copy method instead of gob.

khelif96:
DEVPROD-5081

### Description
The `util.DeepCopy` method used [gob](https://pkg.go.dev/encoding/gob)  to encode/decode structs to allow for copying go structs. This however required each struct to be registered with the gob method in order for gob to be able to copy it. This unfortunately caused errors when used alongside the graphql request secret redaction since there were many different possibilities of structs that could be passed in that would need to be copied. This PR updates the `util.DeepCopy` method to use a more flexible json Marshall/Unmarshal method to copy structs.

### DIFF
--- a/environment.go
+++ b/environment.go
@@ -1050,13 +1050,7 @@ func (e *envState) SaveConfig(ctx context.Context) error {
 	// this is a hacky workaround to any plugins that have fields that are maps, since
 	// deserializing these fields from yaml does not maintain the typing information
 	var copy Settings
-	registeredTypes := []interface{}{
-		map[interface{}]interface{}{},
-		map[string]interface{}{},
-		[]interface{}{},
-		[]util.KeyValuePair{},
-	}
-	err := util.DeepCopy(*e.settings, &copy, registeredTypes)
+	err := util.DeepCopy(*e.settings, &copy)
 	if err != nil {
 		return errors.Wrap(err, "copying settings")
 	}
--- a/environment.go
+++ b/environment.go
@@ -2,7 +2,6 @@ package evergreen
 
 import (
 	"context"
-	"encoding/gob"
 	"fmt"
 	"math"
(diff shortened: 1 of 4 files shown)


util/copy.go
```diff
@@ -1,25 +1,20 @@
...
 
 import (
-	"bytes"
-	"encoding/gob"
+	"encoding/json"
 
 	"github.com/pkg/errors"
 )
 
 // DeepCopy makes a deep copy of the src value into the copy params
```
Kimchelly, line 9: Can we also call out that it only copies exported fields? Unexported (i.e. lowercase) fields won't be copied.

environment.go
```diff
@@ -1050,13 +1050,7 @@ func (e *envState) SaveConfig(ctx context.Context) error {
...
-		map[interface{}]interface{}{},
-		map[string]interface{}{},
-		[]interface{}{},
-		[]util.KeyValuePair{},
-	}
-	err := util.DeepCopy(*e.settings, &copy, registeredTypes)
+	err := util.DeepCopy(*e.settings, &copy)
 	if err != nil {
 		return errors.Wrap(err, "copying settings")
 	}
```
Kimchelly, line 1056: Is the `gob.Register` on L1064 still needed if we're not using gob encoding anymore? It seems like, aside from the deep copy, all of the extra logic here (from the "hacky workaround" comment above to the special handling of `Plugins` below) isn't needed if we don't have to deal with registering the types specially. We could remove the comment and all the special plugin-handling logic below if they aren't relevant.

util/copy.go
khelif96: Good catch I wasn't aware this was a limitation.

Kimchelly (approved):

